PR #639, opened 2016-06-13
Make sequencing abstract

perryibm:
Remove CLI manual use of pipe.js and use of parameters and use a new sequence kind and components field.

Use SequenceExec for sequences but with a NodejsExec-like implementation for now where the compoments meta-data is converted to parameters.

### DIFF
--- a/common/scala/src/whisk/core/entity/Exec.scala
+++ b/common/scala/src/whisk/core/entity/Exec.scala
@@ -19,7 +19,7 @@ package whisk.core.entity
 import scala.collection.JavaConversions.asScalaSet
 import spray.json.DefaultJsonProtocol
 import spray.json.RootJsonFormat
-import spray.json.{ JsValue, JsObject, JsString }
+import spray.json.{ JsValue, JsObject, JsString, JsArray }
 import spray.json.DeserializationException
 import whisk.core.entity.ArgNormalizer.trim
 import spray.json.pimpString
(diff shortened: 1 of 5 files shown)


common/scala/src/whisk/core/entity/Exec.scala
```diff
@@ -76,6 +76,10 @@ protected[core] case class JavaExec(jar: String, main: String) extends Exec(Exec
 
 protected[core] case class BlackBoxExec(image: String) extends Exec(Exec.BLACKBOX)
 
+protected[core] case class SequenceExec(code: String, components: Vector[String]) extends Exec(Exec.SEQUENCE) {
+    val image = "whisk/nodejsaction"
```
rabbah: :/

common/scala/src/whisk/core/entity/Exec.scala
```diff
@@ -76,6 +76,10 @@ protected[core] case class JavaExec(jar: String, main: String) extends Exec(Exec
 
 protected[core] case class BlackBoxExec(image: String) extends Exec(Exec.BLACKBOX)
 
+protected[core] case class SequenceExec(code: String, components: Vector[String]) extends Exec(Exec.SEQUENCE) {
```
rabbah: `Vector[Namespace]` instead?

perryibm: I was going to leave it as a string for this round as the underlying implementation is still with pipe.js.  Next round, Vector[String] becomes Vector[Namespace] and code field disappears.

common/scala/src/whisk/core/entity/Exec.scala
perryibm: Yes - as below.

common/scala/src/whisk/core/entity/Exec.scala
```diff
@@ -133,6 +141,17 @@ protected[core] object Exec
                         case _                => None
                     }
                     NodeJSExec(code, init)
+                case Exec.SEQUENCE =>
```
rabbah: thoughts for pedantic check for defined `code` and reject?

rabbah:
Replaced with #709.

Issue #365, opened 2016-05-09
Question&Answer (Dialog) package


csantanapr:
We could use as an example the [Watson Dialog API](http://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/dialog.html)

lzbj:
@csantanapr ,could you please assign this task to me?

lzbj:
The latest recommended soluitons here:
http://watson-qa-demo.mybluemix.net/
https://developer.ibm.com/watson/blog/2015/11/11/watson-question-and-answer-service-to-be-withdrawn/

csantanapr:
Good catch, then instead we can create actions using the replacements services: Natural Language Classifier, Dialog, Retrieve and Rank, and Document Conversion. 

What do you think we use this issue to implement Dialog and create new issues for the other ones?

lzbj:
Thanks, yes, agree with you. Then maybe we can think of using of those created actions in a pipline to achieve some more complex apps, such as Q&A etc. Another solution is to utilize several Watson api in one action to compose a complex app, but in my opinion, the former one is more flexible.

csantanapr:
@lzbj I agree that's we have sequences to create more complex scenarios, but we should have generic actions in a package that user can copy/clone and adjust and create their own sequence or trigger an event from inside an action to make a rule run with another actions

lzbj:
Yes, sure.

lzbj:
@csantanapr , could you please help create 4 issues with Package label, that associated with replacements services as you mentioned above and assign to me? thank you.

csantanapr:
@lzbj I converted this issue to be for Dialog action, and created 3 more 587, 588, 589
Take into account that each one should have proper tests, and docs cover
I suggest to work on one first, this issue is already assigned to you, as you finish this one and is ready to pick the next one then assign at that time.

lzbj:
@csantanapr Thank you, yes, I'll do as you pointed out.

lzbj:
https://github.com/openwhisk/openwhisk/pull/641
@csantanapr , could you help review it when available, thank you.

markusthoemmes:
@csantanapr still valid/in the right place?

markusthoemmes:
Issue moved to [openwhisk/openwhisk-catalog #209](https://github.com/openwhisk/openwhisk-catalog/issues/209) via [**ZenHub**](https://www.zenhub.com/)

PR #641, opened 2016-06-14
[WIP] Add Watson Dialog package.

lzbj:
For issue #365.

### DIFF
--- a/ansible/templates/whisk.properties.j2
+++ b/ansible/templates/whisk.properties.j2
@@ -53,4 +53,4 @@ python.27=python
 use.cli.download=false
 nginx.conf.dir={{ nginx_conf_dir }}
 testing.auth={{ openwhisk_home }}/config/keys/auth.guest
-vcap.services.file=
+vcap.services.file={{ openwhisk_home }}/tests/credentials/vcap_services.json
--- a/catalog/watson/dialog.js
+++ b/catalog/watson/dialog.js
@@ -1,0 +1,100 @@
+var watson = require('watson-developer-cloud')
+var fs = require('fs')
+
+/**
+ * conversation between virtual agents and users
+ * see https://www.ibm.com/smarterplanet/us/en/ibmwatson/developercloud/dialog/api/v1/
+ * param username The Watson service username.
+ * param password The Watson service password.
+ * param dialogname The name of the dialog to create.
+ * param payload The content of file used to create the dialog.
+ * param tempfile The dialog template file used to create the dialog.
+ * param inputs The input of the user.
+ */
+function main(params) {
+    var username = params.username;
+    var password = params.password;
+    var dialogname = params.dialogname || 'my-dialog';
+    var tempfile = params.tempfile;
+    var payload = params.payload;
+    var dialogId = '';
+    var inputs = params.inputs;
+    var dialog_service = watson.dialog({
+        username: username,
+        password: password,
+        version: 'v1'
+    });
+
+    if (!tempfile && !payload) {
+        whisk.error('you need to give the payload or tempfile path to create dialog');
+    }
+
+
+    var parameters = {};
+    var writeFile = 'temp_file.xml';
+    // tempfile is the first choice.
+    if (tempfile) {
+        parameters = {
+            name: dialogname,
+            file: fs.createReadStream(tempfile)
+        };
+    } else {
+        //write a temp file.
+        var bu = Buffer(payload);
+        var writestream = fs.createWriteStream(writeFile);
+        writestream.once('open', function() {
+            writestream.write(bu);
+        });
+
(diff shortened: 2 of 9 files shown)


catalog/watson/dialog.js
```diff
@@ -0,0 +1,63 @@
...
+        version: 'v1'
+    });
+
+    var params = {
+        name: dialogname,
+        file: fs.createReadStream(tempfile)
+    };
+
+    //create a dialog first
+    dialog_service.createDialog(params, function(err, dialog) {
```
rabbah: is the intention that `conversation` below is reached even if `createDialog` fails?

lzbj: The conversation is based on the dialog creation, if the dialog creates failed, the action is failed.

rabbah: is the call to `createDialog` asynchronous? This is suggested by the callback. In which case `dialogId` is potentially undefined and you're starting `conversation` even though `createDialog` may not have completed or returned an error.

lzbj: Yes, correct, moved the conversation after dialog create success. Thank you rabbah.

csantanapr:
Looks good can you attach the file pizza.xml so I can try it out. 
maybe we can have a short version of this file in the docs this it's easier for an user to give it a quick try

Also please add unit tests we want to have all catalog content tested with Travis CI

csantanapr:
@lzbj I added WIP to the title to indicate that this is still in progress and the only left are the test cases for the new action.

lzbj:
@csantanapr , thanks, yes, I'll add the test cases.

csantanapr:
HI @lzbj please work with @houshengbo on debugging travis build failure.
I see you added the vcap encrypted file, the file can't be decrypted here:https://travis-ci.org/openwhisk/openwhisk/builds/139972366

Also I think we need to save the not encrypted .json file some where that only us can access not public, give it to @houshengbo this will allow us to update it in the future/

@houshengbo is working on a new repo, and I think the tests and code for catalog will move there https://github.com/openwhisk/openwhisk-catalog

lzbj:
@csantanapr , yes, I'll try to figure it out with help of @houshengbo . Thank you.

tests/credentials/vcap_services_template.json
```diff
@@ -0,0 +1,14 @@
+{
```
rabbah: @nwspeete-ibm can you comment on this?

tests/src/packages/WatsonTests.java
```diff
@@ -0,0 +1,72 @@
+/*
```
rabbah: Please rewrite this test in scala. See other Watson tests for example.

ansible/templates/whisk.properties.j2
```diff
@@ -52,4 +52,4 @@ python.27=python
 use.cli.download=false
 nginx.conf.dir={{ nginx_conf_dir }}
 testing.auth={{ openwhisk_home }}/config/keys/auth.guest
-vcap.services.file=
+vcap.services.file={{openwhisk_home}}/tests/credentials/vcap_services.json
```
rabbah: @nwspeete-ibm @csantanapr does this align with other Watson packages requiring vcap?

rabbah: Also should we use a different term rather than vcap per @mbehrendt's suggestion?

vcap_services.json.enc
```diff
@@ -0,0 +1,2 @@
+p���'+=/�c�L���>�����f͠W�Fʎ���?1ֱ�_���#oހ�̔��aJzjxʝ�7�������I�[�d3� [?^�5�b���R?FR@d���� 9M�uaO�x�@a��;����=��������8q�@��I���+�Y��)�L�:���u����
```
rabbah: Even encrypted this file doesn't belong in openwhisk. Others can't run these tests with this file.

rabbah:
pizza.xml surely doesn't belong in docs. If needed for tests put it in tests/dat.

tests/src/packages/WatsonTests.java
lzbj: Ok, got it thank you.

nwspeete-ibm:
Dialog will require its own VCAP credentials to be stored into blue. They can be retrieved using the Util (see other watson scala tests for example). Be sure to set VCAP_SERVICES property when using open.

lzbj:
stop commit to openwhisk, switch to openwhisk-catalog repo.

csantanapr:
Please submit new catalog content to the new repo [openwhisk-catalog](https://github.com/openwhisk/openwhisk-catalog)

I can help create encrypted travis file with credentials

lzbj:
Correct, thank you @csantanapr

PR #642, opened 2016-06-14
Added message to indicate a timeout of an action

markusthoemmes:
Added message to indicate a timeout of an action in the activation record.

Signed-off-by: Christian Bickel <christian.bickel@de.ibm.com>

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala
@@ -52,10 +52,10 @@ protected[core] object ActivationResponse extends DefaultJsonProtocol {
     /* The field name that is universally recognized as the marker of an error, from the application or otherwise. */
     val ERROR_FIELD: String = "error"
 
-    private val Success          = 0
-    private val ApplicationError = 1
-    private val ContainerError   = 2
-    private val WhiskError       = 3
+    val Success          = 0
+    val ApplicationError = 1
+    val ContainerError   = 2
+    val WhiskError       = 3
 
     private def messageForCode(code: Int) = {
         require(code >= 0 && code <= 3)
--- a/common/scala/src/main/scala/whisk/core/entity/TimeLimit.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/TimeLimit.scala
@@ -46,7 +46,7 @@ protected[entity] class TimeLimit private (val duration: FiniteDuration) extends
 protected[core] object TimeLimit extends ArgNormalizer[TimeLimit] {
     protected[entity] val MIN_DURATION = 100 milliseconds
     protected[entity] val MAX_DURATION = 5 minutes
-    protected[entity] val STD_DURATION = 1 minutes
+    protected[entity] val STD_DURATION = 1 minute
 
     /** Gets TimeLimit with default duration */
     protected[core] def apply(): TimeLimit = new TimeLimit(STD_DURATION)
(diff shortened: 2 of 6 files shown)


rabbah:
did you move the tests from blue that do the same?

common/scala/src/whisk/core/entity/ActivationResult.scala
```diff
@@ -51,10 +50,10 @@ protected[core] object ActivationResponse extends DefaultJsonProtocol {
     /* The field name that is universally recognized as the marker of an error, from the application or otherwise. */
     val ERROR_FIELD: String = "error"
 
-    private val Success          = 0
-    private val ApplicationError = 1
-    private val ContainerError   = 2
-    private val WhiskError       = 3
+    val Success = 0
```
rabbah: nit pick - realign? i know it's annoying because i couldn't get auto format to todo the right thing,

core/dispatcher/src/whisk/core/container/WhiskContainer.scala
```diff
@@ -127,7 +128,8 @@ class WhiskContainer(
                 val http = new HttpUtils(connection, host)
                 val (code, bytes) = http.dopost(endpoint, msg, Map(), timeout)
                 Try { connection.close() }
-                Some(code, new String(bytes, "UTF-8"))
+                val returnCode = if (code == -1) ActivationResponse.ContainerError else code
```
rabbah: why?

core/dispatcher/src/whisk/core/invoker/Invoker.scala
```diff
@@ -275,7 +275,8 @@ class Invoker(
                         info(this, s"posted completion of activation ${msg.activationId}")
                     }
                     // We return the container last as that is slow and we want the activation to logically finish fast
-                    pool.putBack(con, failedInit)
+                    val removeContainer = failedInit || response.exists { _._1 == ActivationResponse.ContainerError }
```
rabbah: why?

markusthoemmes: If an action times out, and we just pause the container, on unpause (reuse) the action might actually still be running in the background, finishing the work it wasn't able to complete before it was paused. It might make sense to remove the container everytime there was a container, whisk or application error.

core/dispatcher/src/whisk/core/container/WhiskContainer.scala
markusthoemmes: See below, this is to normalize the HttpClient interface, which yields magic number `-1` with invoker internal error codes.

common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala
```diff
@@ -52,10 +52,10 @@ protected[core] object ActivationResponse extends DefaultJsonProtocol {
...
     val ERROR_FIELD: String = "error"
 
-    private val Success          = 0
-    private val ApplicationError = 1
-    private val ContainerError   = 2
-    private val WhiskError       = 3
+    val Success          = 0
+    val ApplicationError = 1
+    val ContainerError   = 2
+    val WhiskError       = 3
```
rabbah: why expose these? there are ` isSuccess` etc message to use, not sufficient?

markusthoemmes: We used these as constants to return (see https://github.com/openwhisk/openwhisk/pull/642/files#diff-2850946ee2d35de3b057e40f10f1887fR131 for example), so we wouldn't need magic numbers.

perryibm:
rebased/merged

Issue #643, opened 2016-06-14
Eclipse (Scala IDE) project files don't work


tuler:
I'm trying to generate project files for Scala IDE using the eclipse gradle plugin, but the generated files don't work. Controller project is linked to dispatcher src.

I executed:
gradle cleanEclipse
gradle eclipse

I would move the source code into a standard structure (src/main/scala, src/main/resources), as described at:
https://docs.gradle.org/current/userguide/scala_plugin.html#N1664C

If this is the way to go I can open a pull request.

markusthoemmes:
I don't see any objections on why we shouldn't follow the convention. Pull request very welcome!

The link to certain projects (i.e. loadbalancer) is "works as designed". Which problems do you experience with the files otherwise?

tuler:
loadBalancer is not linked at controller project (as it should).
controller project is linked to dispatcher src (instead of controller src)

I'm using Scala IDE 4.4.1
Build id: 4.4.1-vfinal-2016-05-04T11:16:00Z-Typesafe

rabbah:
The link to dispatcher src is not necessary and about to get removed in #578. I agree the linking should be at project level otherwise instead of src if that's the suggestion.

tuler:
#646 fixes this.

PR #646, opened 2016-06-14
Standard folder structure

tuler:
This moves src files to a standard project structure as described by the [scala gradle plugin](https://docs.gradle.org/current/userguide/scala_plugin.html#N1664C).

This was raised in issue #643 

I would still change thinks like:

- rename package "whisk" to "openwhisk" or "com.openwhisk". Whisk.com is an existing business, and there is even a organization "whisk" at docker hub.

- rename project "scala" to "core" or "common"

I'm not sure if changes are required to ansible scripts or CI/travis scripts.


core/controller/build.gradle
```diff
@@ -10,23 +10,11 @@ repositories {
...
-            srcDir 'src/resources'
-        }
-    }
-}
-
 dependencies {
     compile 'org.scala-lang:scala-library:2.11.7'
     compile project(':common:scala')
+    compile project(':core:loadBalancer')
+    compile project(':core:dispatcher')
```
rabbah: fairly sure this is not necessary

settings.gradle
```diff
@@ -2,6 +2,7 @@ include 'common:scala'
 
 include 'core:controller'
 include 'core:dispatcher'
+include 'core:loadBalancer'
```
rabbah: @markusthoemmes more reason to merge loadbalance source with controller?

common/scala/src/main/scala/whisk/core/entity/ActivationLogs.scala
```diff
@@ -28,7 +28,7 @@ import spray.json.RootJsonFormat
 import spray.json.deserializationError
 import spray.json.pimpAny
 
-protected[core] case class ActivationLogs(private val logs: Vector[String] = Vector()) extends AnyVal {
+protected[core] case class ActivationLogs(val logs: Vector[String] = Vector()) extends AnyVal {
```
rabbah: private -> public orthogonal change?

core/controller/build.gradle
tuler: If I remove that Backend.scala and Controller.scala don't compile. They depend on LoadBalancerService.

common/scala/src/main/scala/whisk/core/entity/ActivationLogs.scala
tuler: Without this I get the following warning:
sbt-api: Unhandled type class scala.reflect.internal.Types$MethodType : ($this: whisk.core.entity.ActivationLogs)Vector[String]

And this triggers an error in the dependent projects.

I found these related issues:
http://stackoverflow.com/questions/29720289/warning-about-unhandled-type-class-during-compilation
http://www.scriptscoop2.com/t/3557fd15ab2c/scala-warning-about-unhandled-type-class-during-compilation.html

Seems like a compiler bug.

core/controller/build.gradle
markusthoemmes: I think @rabbah is just talking about the dispatcher dependency. Does it yield the same?

tuler:
Build broken, not sure why :-(

markusthoemmes:
Doesn't look related to your change, restarted the build.

settings.gradle
markusthoemmes: Definitly! @tuler would you mind to copy the code from `loadbalancer` to the controller (they have been merged which is why there was no `build.gradle` left for it). It should then be safe to delete the whole loadbalancer folder I think.

markusthoemmes:
@tuler FYI we have the policy of a linear git-history, meaning: no merge commits. Your pull-request reminded me that we haven't documented that, which we have now, see https://github.com/openwhisk/openwhisk/wiki/Contributing:-Git-guidelines.

You don't need to mind for this pull-request, we will just squash on merge, but for future pull-request it'll be easier for us if `rebase` is used to keep the pull-request on track instead of `merge`.

rabbah:
PG 111.

rabbah:
Rebased, squashed and merged as https://github.com/openwhisk/openwhisk/commit/3f50c7b990145cebbd269a5978b7badd84123c5e.

Eclipse users should `gradle cleanEclipse` to remove old projects and reinit with `gradle eclipse`.

rabbah:
Thanks @tuler for the contribution.

csantanapr:
I believe this is merged now with 3f50c7b990145cebbd269a5978b7badd84123c5e
Thanks @tuler for tine contrib. 👍

PR #651, opened 2016-06-14
WIP added .gitattributes to set eol=lf

### LINKED ISSUE #633: windows platform issue: "\r" command not found
When following the  guide executing vagrant up(for cloudant case) for custom setup , there are issues:

createImmortalDBs.sh
`/bin/bash^M: bad interpreter: No such file or directory`

`/home/vagrant/openwhisk/tools/db/../../config/dbSetup.sh: line 5: $'\r': command not found
 default: Unrecognized OPEN_WHISK_DB_PROVIDER: ''`

....

This is because many script failed because of issue  "\r" command not found. Usually users will use Cygwin ,cmd or other terminal to run the command and should perform dos2unix(convert  dos format to unix format) to overcome this issue.

I use this command as a workround and these issues were solved.

```
$ pwd
$ /c/workspace/openwhisk
$ find ./ -name "*.sh" -print | xargs  dos2unix
```

I think it is good to write in the document and windows users will get the solution when they meet the issue.

csantanapr:
Fixes #633

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+* text eol=lf
+
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,11 @@
-* text eol=lf
+# Auto detect text files and perform LF normalization
+# http://davidlaing.com/2012/09/19/customise-your-gitattributes-to-become-a-git-ninja/
+* text=auto
+
+*.sh text eol=lf
+*.scala text
+
+*.jar binary
+*.png binary
+
 


.gitattributes
```diff
@@ -0,0 +1,2 @@
+* text eol=lf
```
csantanapr: @rabbah What do you think too global? Or should we make it scope it to `*.sh` ?

rabbah: I defer to @psuter.

psuter: Not an expert on this (just read the docs). Could we have:

    * text=auto
    *.sh text
    *.scala text
    [...]
    *.jar binary
    *.png binary

Etc.? I'd rather have an explicit list of extensions than open ourselves to surprises (e.g. when one of us upgrades Git).

.gitattributes
```diff
@@ -1,2 +1,11 @@
...
+# Auto detect text files and perform LF normalization
+# http://davidlaing.com/2012/09/19/customise-your-gitattributes-to-become-a-git-ninja/
+* text=auto
+
+*.sh text eol=lf
+*.scala text
+
+*.jar binary
+*.png binary
+
```
csantanapr: @psuter What about this one? Good to go?

csantanapr:
@psuter review and merge.

psuter:
Tweaked and merged as e11a92d626ef0c171ef39d58ea78738d51e977ed.

PR #654, opened 2016-06-15
make subjects db environment-specific

lehoanganh:
Signed-off-by: Dominik Jall <djall@de.ibm.com>


lehoanganh:
FYI: @domdom82 @jeremiaswerner

domdom82:
looks good. background - we have different authdbs in blue per environment. now described as "auth_db" (following guideline where _ notation is env specific and . notation is global). To make this look the same in open we have added this PR.

PR #652, opened 2016-06-15
[Ready] Supercedes #561 - Fix Docker Action SDK

rabbah:
Supercedes #561.

Fix all parts of the docker skeleton action which were completely broken.

- Check `clientApp` returned a JSON parsable object and if not, report error and log a message.
- Update example to show how params are received and return a proper JSON.
- Fix space in path in various build scripts.
- Add gradle build for docker skeleton - to enable testing locally.
- Replace broken docker action tests with test that verifies the docker skeleton can tolerate an init message and can receive/respond with JSON objects. This test suite is not complete in that the docker skeleton hardcodes an example and the tests are specific to that example so cannot test that failure modes are correct yet.

PR #657, opened 2016-06-15
Adding action images as dependencies for the test task

markusthoemmes:
Adding action images as dependencies for the test task so they are guaranteed to be there when the testsuite runs.


rabbah:
Going to hold pending #652 (will need to update for one more dependence: the docker sdk action).

rabbah:
Rebased and merged as https://github.com/openwhisk/openwhisk/commit/386a45110c7eed940056471a03ad9b8640b8a9f4.

PR #660, opened 2016-06-15
fix gateway timeout to controller 

rabbah:
Fix gateway timeout to controller - wait up 5 minutes which is max allowed action duration.

### DIFF
--- a/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/ansible/roles/nginx/templates/nginx.conf.j2
@@ -37,6 +37,7 @@ http {
         }
         location /api/v1 {
             proxy_pass http://{{ groups['controllers']|first }}:{{ controller.port }};
+            proxy_read_timeout 310s;
         }
         location /openwhisk-0.1.0.tar.gz {
             root /etc/nginx;


ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -37,6 +37,7 @@ http {
         }
         location /api/v1 {
             proxy_pass http://{{ groups['controllers']|first }}:{{ controller.port }};
+            proxy_read_timeout 300s;
```
rabbah: default is 60s.

rabbah:
@lehoanganh @domdom82 FYI

ansible/roles/nginx/templates/nginx.conf.j2
markusthoemmes: Do we need little bit of slack, as the action, even when called blocking will return in 5 minutes + the time needed to propagate stuff?

rabbah: 305s?

markusthoemmes: Something along these lines, 305 should be enough, 310 to be sure?

rabbah: btw controller grants no more than 5s grace.

markusthoemmes: So it's going to be 305 (including grace) + some nanoseconds at least to get the request back to nginx so it doesn't say: Gateway Timeout?

rabbah:
rebased.

sjfink:
rebased and merged

PR #669, opened 2016-06-16
Allow ant run to use a test selector

rabbah:
since we run tests via gradle but at the top level using ant, allow test selector using `-Dwhisktests.prefix` for convenience. Example use: `ant run -Dwhisktests.prefix="system.basic.Wsk*"` runs `WskBasicTests` and `WskSdkTests` from `system.basic` package.

### DIFF
--- a/config/config.xml
+++ b/config/config.xml
@@ -11,12 +11,12 @@
 
     <!-- required properties with default values -->
     <!-- where all build artifacts go -->
-    <property name="build.dir" value="${basedir}/build" />
-    <property name="cli.build.dir" value="${basedir}/build/tools/cli" />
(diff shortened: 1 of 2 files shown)


csantanapr:
what about `--no-rebuild` does it help re-running tests faster?

rabbah:
In a highly scientific study, `--no-rebuild` saved 4 seconds.

csantanapr:
haha you want me to merge this now?

rabbah:
I was going to use this for an experiment but I will just use gradle directly. Since we want to continue to decimate `ant` from our build, this was perhaps counter productive.

PR #670, opened 2016-06-16
[Ready] Replaced ConsulKV by ConsulClient in Entitlement

psuter:
Also replaces the infinite-polling thread by a `Scheduler`-produced actor.

Other assorted changes, such as pushing `ActorContext` implicit arguments throughout the system, and removing implicit `ExecutionContext` args when it can be obtained from the `ActorSystem`.

After this change, there is still a use of `ConsulKV` in `Config`, but it's once, at construction time only. (There are other uses in the invoker, still.)

This mimics the work done in 349da264f7b9def36b0d7dc189ad7170082e8dc1.

### DIFF
--- a/core/controller/src/whisk/core/entitlement/Entitlement.scala
+++ b/core/controller/src/whisk/core/entitlement/Entitlement.scala
@@ -17,17 +17,23 @@
 package whisk.core.entitlement
 
 import scala.collection.concurrent.TrieMap
-import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.Promise
-import scala.util.Failure
-import scala.util.Success
+import scala.concurrent.duration.DurationInt
+import scala.language.postfixOps
+import scala.util.{ Try, Success, Failure }
+
 import Privilege.REJECT
+import akka.actor.ActorSystem
 import spray.http.StatusCodes.ClientError
 import spray.http.StatusCodes.TooManyRequests
 import spray.json.JsBoolean
-import whisk.common.ConsulKV
+import spray.json.pimpString
+import whisk.common.ConsulClient
+import whisk.common.ConsulKV.LoadBalancerKeys
 import whisk.common.Logging
+import whisk.common.LoggingMarkers
+import whisk.common.Scheduler
 import whisk.common.TransactionId
 import whisk.common.Verbosity
 import whisk.core.WhiskConfig
@@ -35,8 +41,6 @@ import whisk.core.entity.Namespace
 import whisk.core.entity.Parameters
 import whisk.core.entity.Subject
 import whisk.http.ErrorResponse
-import scala.language.postfixOps
-import whisk.common.LoggingMarkers
 
 package object types {
     type Entitlements = TrieMap[(Subject, String), Set[Privilege]]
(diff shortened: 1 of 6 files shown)


rabbah:
FYI @markusthoemmes.

psuter:
I kind of wanted to make sure it passes the tests before bothering
@markusthoemmes with this :)

psuter:
IPG 84

core/controller/src/whisk/core/entitlement/Entitlement.scala
```diff
@@ -72,35 +78,33 @@ protected[core] object EntitlementService {
...
+    /** query the KV store this often */
+    private val overloadCheckPeriod = 10.seconds
+
+    private val kvClient = new ConsulClient(config.consulServer)
+
+    private var counted: Boolean = false
+    Scheduler.scheduleWaitAtLeast(overloadCheckPeriod) { () =>
+        kvClient.get(LoadBalancerKeys.overloadKey).map { isOverloaded =>
+            Try(isOverloaded.parseJson) match {
+                case Success(JsBoolean(v)) if !counted || loadbalancerOverload != v =>
```
markusthoemmes: Is the `counted` necessary? I understand that it forces an update when running the first time, but do we really need to do that? Couldn't we live with just the default `loadBalancerOverload` of `false`? @perryibm ?

core/controller/src/whisk/core/entitlement/Entitlement.scala
```diff
@@ -72,35 +78,33 @@ protected[core] object EntitlementService {
...
-                    case JsBoolean(v) => if ((count == 0) || loadbalancerOverload != v) {
-                        loadbalancerOverload = v
-                        info(this, s"EntitlementService: loadbalancerOverload = ${v}")
-                        count = count + 1
-                    }
-                    case _ => ()
-                }
-                Thread.sleep(overloadCheckPeriodMillis)
+    /** query the KV store this often */
+    private val overloadCheckPeriod = 10.seconds
```
markusthoemmes: Specific reason why you not use `postfixOps` and write `10 seconds`?

core/controller/src/whisk/core/entitlement/Entitlement.scala
```diff
@@ -72,35 +78,33 @@ protected[core] object EntitlementService {
...
+    private var counted: Boolean = false
+    Scheduler.scheduleWaitAtLeast(overloadCheckPeriod) { () =>
+        kvClient.get(LoadBalancerKeys.overloadKey).map { isOverloaded =>
+            Try(isOverloaded.parseJson) match {
+                case Success(JsBoolean(v)) if !counted || loadbalancerOverload != v =>
+                    loadbalancerOverload = v
+                    info(this, s"EntitlementService: loadbalancerOverload = ${v}")
+                    counted = true
+
+                case _ => ;
```
markusthoemmes: Semicolon? The horror!

core/controller/src/whisk/core/entitlement/LocalEntitlement.scala
```diff
@@ -31,9 +32,11 @@ private object LocalEntitlementService {
 
 protected[core] class LocalEntitlementService(
     private val config : WhiskConfig)
-    (implicit ec: ExecutionContext)
+    (implicit actorSystem: ActorSystem)
     extends EntitlementService(config) {
 
+    private implicit val executionContext = actorSystem.dispatcher
```
markusthoemmes: Seems like the `LocalEntitlementService` doesn't need an actor system, so rather just keep the `ExecutionContext` here?

core/controller/src/whisk/core/entitlement/Entitlement.scala
```diff
@@ -37,6 +42,7 @@ import whisk.core.entity.Subject
 import whisk.http.ErrorResponse
 import scala.language.postfixOps
 import whisk.common.LoggingMarkers
+import whisk.common.Scheduler
```
markusthoemmes: Reorg imports?

core/controller/src/whisk/core/entitlement/Entitlement.scala
psuter: I tried to strictly follow what was there before; there was a `count` field being incremented and tested against `0`, so at least my version won't overflow :)

core/controller/src/whisk/core/entitlement/Entitlement.scala
psuter: Personal taste? I find postfix operators make things harder to parse visually in general.

core/controller/src/whisk/core/entitlement/Entitlement.scala
psuter: If you think that's clearer I can change. Note that you'll still have an `if` in the closure.

rabbah: i would put a comment if the `;` is removed

psuter:
Prettier, based on feedback. IPG 101 eventually.

markusthoemmes:
LGTM, memo to myself to merge.

markusthoemmes:
rebased and merged as 24d7fc6ab674af9e702b3d00c84b430465f02747

Issue #672, opened 2016-06-16
docker-machine: not found   Persistent couch db


lfzCarlosC:
```
-----host machine-------
$ ./hello
...........
{
    "message": "hello"
}
$  vagrant ssh

------vm environment    intend to setup a persistant couch db------
$ export OW_DB=Cloudant
$ export OW_DB_USERNAME=admin
$ export OW_DB_PASSWORD=password
$ export OW_DB_PROTOCOL=https
$ export OW_DB_HOST=http
$ export OW_DB_PORT=6000
ansible-playbook setup.yml


PLAY [localhost] **************************************************************


TASK [setup] ******************************************************************

Thursday 16 June 2016  10:42:16 +0000 (0:00:00.038)       0:00:00.038 *********

ok: [localhost]

TASK [find the ip of docker-machine] ******************************************

Thursday 16 June 2016  10:42:19 +0000 (0:00:03.213)       0:00:03.251 *********

fatal: [localhost -> localhost]: FAILED! => {"changed": true, "cmd": "\"docker-
achine\" \"ip\" \"whisk\"", "delta": "0:00:00.003987", "end": "2016-06-16 10:42
19.515486", "failed": true, "rc": 127, "start": "2016-06-16 10:42:19.511499", "
tderr": "/bin/sh: 1: docker-machine: not found", "stdout": "", "stdout_lines":
], "warnings": []}
...ignoring

TASK [set_fact] ***************************************************************

Thursday 16 June 2016  10:42:19 +0000 (0:00:00.203)       0:00:03.454 *********

ok: [localhost]

TASK [gen hosts for docker-machine] *******************************************

Thursday 16 June 2016  10:42:19 +0000 (0:00:00.042)       0:00:03.497 *********

skipping: [localhost]

TASK [prepare db_local.ini] ***************************************************

Thursday 16 June 2016  10:42:19 +0000 (0:00:00.032)       0:00:03.529 *********



ok: [localhost -> localhost]

PLAY RECAP ********************************************************************

localhost                  : ok=4    changed=0    unreachable=0    failed=0

Thursday 16 June 2016  10:42:19 +0000 (0:00:00.367)       0:00:03.897 *********

===============================================================================

setup ------------------------------------------------------------------- 3.21s
prepare db_local.ini ---------------------------------------------------- 0.37s
find the ip of docker-machine ------------------------------------------- 0.20s
set_fact ---------------------------------------------------------------- 0.04s
[...]

PR #688, opened 2016-06-17
issue #672: improve ansible setup.yml

lehoanganh:
* only use docker-machine commands in mac env. 
* add db host on-the-fly to ansible runtime once mac env is used. 
* improve README.md

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -25,7 +25,7 @@ brew install python
 pip install ansible==2.0.2.0
 
 cd ansible
-ansible-playbook setup.yml [-e docker_machine_name=whisk]
+ansible-playbook -i environments/mac setup.yml [-e docker_machine_name=whisk]
 ```
 
 **Hint:** If you omit the optional `-e docker_machine_name` parameter, it will default to "whisk".  
@@ -72,7 +72,7 @@ After the playbook is done you should see a file called `db_local.ini` in your a
 If you want to use the ephemeral CouchDB, run this command
 
 ```
-ansible-playbook setup.yml
+ansible-playbook -i environments/<environment> setup.yml
 ```
 
 #####  Persistent CouchDB
@@ -87,7 +87,7 @@ export OW_DB_PROTOCOL=<your couchdb protocol>
 export OW_DB_HOST=<your couchdb host>
 export OW_DB_PORT=<your couchdb port>
 
-ansible-playbook setup.yml
+ansible-playbook -i environments/<environment> setup.yml
 ```
 
 ##### Cloudant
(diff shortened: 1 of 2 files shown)


tuler:
Support docker for mac (beta) would also be great.

ansible/README.md
```diff
@@ -25,7 +25,7 @@ brew install python
 pip install ansible==2.0.2.0
 
 cd ansible
-ansible-playbook setup.yml [-e docker_machine_name=whisk]
+ansible-playbook -i environments/mac setup.yml [-e docker_machine_name=whisk]
```
csantanapr: @lehoanganh  Is this suppose to be prereq.yml?

lehoanganh: no, it's for setup.yml too. `setup.yml` requires a specific environment for now, cause it needs the ip of db group which is different in mac and local. Once we can unite mac and local env, we still have to input the env here.

lehoanganh: @tuler : I'd love to support it, if I could use it :). Until now, no reaction from docker for trying the new docker for mac.

csantanapr: Ah I think what's going on, and someone else hit the same problem in my team. 
I think part of the problem is the naming of the steps we "prereq" and "setup"  and it's hard to remember which what order they need to run. 
I will continue to use it as is but might come up with a simpler naming and ordering. Right people need to pay attention to the order and run setup and then prereq

I'm not questioning on what each do, just giving feedback that folks are getting both names mixed

ansible/setup.yml
```diff
@@ -6,14 +6,20 @@
...
       docker_machine_ip: "{{ result.stdout }}"
-  
+    when: "'environments/mac' in inventory_dir"
+
   - name: gen hosts for docker-machine 
     local_action: template src="{{playbook_dir}}/environments/mac/hosts.j2.ini" dest="{{ playbook_dir }}/environments/mac/hosts"
-    when: result.rc == 0
+    when: "'environments/mac' in inventory_dir"
+
+  - name: add new db host on-the-fly
```
markusthoemmes: Add a comment as to why this is needed (trick you mentioned to make the next lines work)

lehoanganh: line 20 is for a good reason. Because at the beginning, we don't have a hosts file for mac, which is needed for generating the `db_local.ini`. That's why line 20 adds the newly generated db host on-the-fly into ansible environment.

ansible/README.md
rabbah: @lehoanganh is it an error to run ansible-playbook setup.yml without an environment then? It appears to have the same effect compared to specifying the environment.

rabbah:
Rebased and merged as https://github.com/openwhisk/openwhisk/commit/8bc094ac7a5c84bc44a0d77cae36654e27edc32b.

PR #704, opened 2016-06-20
Run scanCode in travis builds

markusthoemmes:
We need to catch tabs/non-compliant files in the build process again.


rabbah:
Rebased and merged as https://github.com/openwhisk/openwhisk/commit/dade50b2a6f213282cb225c1c5def88e8cd80f27.

PR #706, opened 2016-06-20
Add Go CLI to OpenWhisk - Defect Fixes

dubeejw:
- Defect fixes

### DIFF
--- a/tools/go-cli/go-whisk-cli/commands/trigger.go
+++ b/tools/go-cli/go-whisk-cli/commands/trigger.go
@@ -66,7 +66,7 @@ var triggerFireCmd = &cobra.Command{
         }
         client.Namespace = qName.namespace
 
-        payload := map[string]interface{}{}
+        var payload *json.RawMessage
 
         if len(flags.common.param) > 0 {
             parameters, err := parseParameters(flags.common.param)
@@ -77,9 +77,12 @@ var triggerFireCmd = &cobra.Command{
                 return werr
             }
 
-            for _, param := range parameters {
-                payload[param.Key] = param.Value
-            }
+            payload = parameters
+        }
+
+        if payload == nil {
+            data := []byte("{}")
+            payload = (*json.RawMessage)(&data)
         }
 
         if len(args) == 2 {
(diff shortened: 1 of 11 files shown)


tools/go-cli/go-whisk-cli/commands/trigger.go
```diff
@@ -512,6 +563,32 @@ func createFeed (triggerName string, FullFeedName string, parameters []string) e
...
 }
 
+func deleteFeed (triggerName string, FullFeedName string, parameters []string) error {
+    var originalParams = flags.common.param
+//changes?blocking=true&result=false
+//url = 'https://%(apibase)s/namespaces/%(namespace)s/actions/%(name)s?blocking=%(blocking)s&result=%(result)s' % {
+
+    ///dummyCloudantPackage/changes?blocking=true&result=false
+
+    //dummyCloudantTrigger-1466181193417?blocking=true
```
markusthoemmes: Want to remove these comments?

csantanapr:
@mdeuser could you help review this PR? Once you give me a 👍 I will merge.

mdeuser:
@dubeejw - A few questions/comments.
- Should annotations be processed similar to the parameters?  I think both can take JSON
- Regarding the parseParameters() and parseParametersArray() functions - can a single parseParameters() function handle the parsing such that the command handler is not entirely aware of the parameter value format?
- Why is a '+' (plus) character replaced with a ' ' (space) character in the action name, package name, rule name, and trigger name after the name is URL escaped?  I see that the url.QueryUnescape() does the inverse of ur.QueryEscape(), but it converts '+' to ' ' (space).

dubeejw:
@mdeuser,

url.QueryEspace replace spaces with plus characters. In comparison, the Python CLI never replaces spaces for the action, package, rule, and trigger names. Replacing the plus characters with spaces in needed in the Go CLI to achieve parity with the Python CLI. Also, if the spaces are not preserved, automated tests fail.

The spaces have to be added back because the the automation expects to see the space when it reads responses from the CLI. This is due to the fact that the Python CLI treats spaces in the same manner.

mdeuser:
@dubeejw - Ah.  I think I see the problem.  The resource name (i.e. package name, action name, etc) is actually part of the URL path, not part of the URL query.  The URL escaping process is different depending on whether the string to escape will be part of the URL path or the URL query.  In a URL path, a space is escaped with %20; in a URL query, a space is escaped with a '+'.
http://stackoverflow.com/questions/1634271/url-encoding-the-space-character-or-20

The url.QueryEscape() function should not be used on these name strings when they are part of the URL path and not in a URL query

mdeuser:
I opened issues for my comments to be addressed outside of this PR.  PR should work as-is, but issues should be considered.
LGTM

csantanapr:
Rebase and merged 20178d6b88c18ad5abbcf373e101b41d04952388

PR #707, opened 2016-06-20
Add Go CLI to OpenWhisk - Gradle

dubeejw:
- Build Go CLI with Gradle

### DIFF
--- a/tools/go-cli/build.gradle
+++ b/tools/go-cli/build.gradle
@@ -1,0 +1,67 @@
+ext.dockerImageName = "whisk/go-cli"
+ext.dockerContainerName = "go-cli"
+apply from: "../../docker.gradle"
+
+// Detects if the local OS is Unix. If so, ant.properties.family is set to linux.
+ant.condition(property: "family", value: "linux") {
+    os(family: "unix")
+}
+
+// Detects if the local OS is Mac. If so, ant.properties.family is set to mac.
+ant.condition(property: "family", value: "mac") {
+    os(family: "mac")
+}
+
+// Detects if the local CPU architecture is amd64. If so, ant.properties.arch is set to amd64.
+ant.condition(property: "arch", value: "amd64") {
+    os(arch: "amd64")
+}
+
+// Detects if the local CPU architecture is x86_64. If so, ant.properties.arch is set to amd64.
+ant.condition(property: "arch", value: "amd64") {
+    os(arch: "x86_64")
+}
+
+// Detects if the local CPU architecture is 386. If so, ant.properties.arch is set to 386.
+ant.condition(property: "arch", value: "386") {
+    os(arch: "386")
+}
+
+// Detects if the local CPU architecture is i386. If so, ant.properties.arch is set to 386.
+ant.condition(property: "arch", value: "386") {
+    os(arch: "i386")
+}
+
+// Detects if the local CPU architecture is x86_32. If so, ant.properties.arch is set to 386.
+ant.condition(property: "arch", value: "386") {
+    os(arch: "x86_32")
+}
+
+task removeBinary(type: Delete) {
+    delete "${projectDir}/../../bin/go-cli"
+}
+
+task distBinary(dependsOn: [removeBinary, distDocker]) << {
+    run(dockerBinary + ["rm", "-f", dockerContainerName], true)
+    run(["mkdir", "${projectDir}/../../bin/go-cli/"])
+    run(dockerBinary + ["run", "--name", dockerContainerName, dockerImageName + ":" + dockerImageTag])
+
+    if (ant.properties.family in ["linux", "mac"] && ant.properties.arch in ["amd64", "386"]) {
+        run(dockerBinary + ["cp", dockerContainerName +
+            ":/src/github.com/go-whisk-cli/${ant.properties.family}/${ant.properties.arch}/wsk",
+            "${projectDir}/../../bin/go-cli/wsk"])
+    }
+
+    run(dockerBinary + ["rm", "-f", dockerContainerName])
+}
+
+distDocker.finalizedBy([removeBinary, distBinary])
+
+def run(cmd, ignoreError = false) {
+    println("Executing '${cmd.join(" ")}'")
+    def proc = cmd.execute()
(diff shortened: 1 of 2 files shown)


tools/go-cli/build.gradle
```diff
@@ -0,0 +1,75 @@
...
+task removeBinary(type: Delete) {
+    delete "${projectDir}/../../bin/go-cli"
+    delete "${projectDir}/wsk"
+}
+
+task distBinary(dependsOn: [removeBinary, distDocker]) << {
+
+    run(dockerBinary + ["rm", "-f", dockerContainerName], true)
+
+    run(["rm", "-rf", "${projectDir}/../../bin/go-cli/"], true)
```
markusthoemmes: This seems to be redundant with `removeBinary`. Is this possible to achieve without mangling with directories outside of the `go-cli` project's directory?

tools/go-cli/build.gradle
```diff
@@ -0,0 +1,75 @@
+ext.dockerImageName = "whisk/go-cli-test"
+ext.dockerContainerName = "go-cli-test"
```
markusthoemmes: Why is this called `go-cli-test`?

tools/go-cli/build.gradle
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }
+
+    if (ant.properties.family == "mac") {
+        if (ant.properties.arch == "amd64") {
+            run(dockerBinary + [dockerBinary, "cp", dockerContainerName + ":/src/github.com/go-whisk-cli/mac/amd64/wsk", "${projectDir}/../../bin/go-cli/wsk"])
+        } else if (ant.properties.arch == "386") {
+            run(dockerBinary + [dockerBinary, "cp", dockerContainerName + ":/src/github.com/go-whisk-cli/mac/386/wsk", "${projectDir}/../../bin/go-cli/wsk"])
+        }
+    }
```
markusthoemmes: Can both of these blocks be shortened? The `docker` calls seem to contain quite a bit of duplication. Groovy supports `in` notation for lists etc. so i can imagine you can build something like:

```
if(ant.properties.family in ["linux", "mac"] && ant.properties.arch in ["amd64", "386"]) {
    run(dockerBinary + [dockerBinary, "cp", dockerContainerName + ":/src/github.com/go-whisk-cli/${ant.properties.family}/${ant.properties.arch}/wsk", "${projectDir}/../../bin/go-cli/wsk"])
}
```

Very nice catch with the availability of `dockerBinary` though, wasn't aware of that 👍

tools/go-cli/build.gradle
dubeejw: We could create a symlink in the openwhisk/bin folder like there is for the Python CLI. It appears that the only reason the Python CLI does this is because Python doesn't generate binaries. I'm not sure that the extra indirection is needed.

tools/go-cli/build.gradle
```diff
@@ -0,0 +1,75 @@
...
+    os(arch: "386")
+}
+
+ant.condition(property: "arch", value: "386") {
+    os(arch: "i386")
+}
+
+ant.condition(property: "arch", value: "386") {
+    os(arch: "x86_32")
+}
```
markusthoemmes: There is a plugin to detect architecture and operating system https://github.com/google/osdetector-gradle-plugin.

dubeejw: @mdeuser, @csantanapr, would we still have time to get legal clearance on the package Markus mentioned?

tools/go-cli/build.gradle
```diff
@@ -0,0 +1,60 @@
...
+
+task removeBinary(type: Delete) {
+    delete "${projectDir}/../../bin/go-cli"
+}
+
+task distBinary(dependsOn: [removeBinary, distDocker]) << {
+    run(dockerBinary + ["rm", "-f", dockerContainerName], true)
+    run(["mkdir", "${projectDir}/../../bin/go-cli/"])
```
markusthoemmes: Can we still put the binaries in `/build` and then have a `Copy` task creating the folder and moving the binary out? It'ld then be more analogous to the existing `cli` gradle file + moving the correct binary out.

dubeejw: The Python CLI moves the Python source code to the build directory to compress it using Gradle. We are compressing the Go CLI binaries directly in the Docker image.

markusthoemmes: Well, yeah, but it also puts the outcome of the `distTar` task in `build/distributions` which is idiomatic for gradle. From there on, we can continue with copying the stuff around to wherever we need it, keeping the `distBinary`  task itself gradle idiomatic.

mdeuser:
@dubeejw - I believe we can update the IBM Go CLI legal approval package.  The above mentioned gradle plugin is only used at build time (i.e. we don't ship it) and it's Apache 2.0 licensed which is one of the friendlier OSS licenses.  Just wondering, what if we just copy all Go binaries for all platforms and architectures, and let the final phase of distribution determine what is available.

tools/go-cli/build.gradle
csantanapr: @dubeejw This is an open source project, time if infinite 😄 
I did a quick look at the LICENSE for the dependency and it's Apache 2.0, so it should be fine and compatible with this open source project.

markusthoemmes: If we want to defer that change: please add a comment above the ant calls describing that they are used to detect the environment you're running on and set `ant.property.arch` and `ant.property.family` to the respective values of your system

dubeejw:
Create issues https://github.com/openwhisk/openwhisk/issues/724, and https://github.com/openwhisk/openwhisk/issues/723 for the suggestions provided by Markus.

csantanapr:
@mdeuser @dubeejw So this one ready to merge (i.e. no more changes)?
And handle the suggestions by @markusthoemmes in the new issues

dubeejw:
@csantanapr, correct.

csantanapr:
@dubeejw I will wait before I merge for a green Travis build on the commit https://travis-ci.org/openwhisk/openwhisk/builds/139308715

PR #711, opened 2016-06-21
Add address of edge host to WhiskConfig

ChrBi:
This PR adds the address of the edge host to WhiskConfig.
Before this change, it was only possible to get the edge host including a port, but not the address itself.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
+++ b/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
@@ -53,6 +53,8 @@ class WhiskConfig(
     val loadbalancerHost = this(WhiskConfig.loadbalancerHostName) + ":" + this(WhiskConfig.loadbalancerHostPort)
     val messagehubtriggerHost = this(WhiskConfig.messagehubtriggerHostName) + ":" + this(WhiskConfig.messagehubtriggerHostPort)
 
+    val edgeHostName = this(WhiskConfig.edgeHostName)
+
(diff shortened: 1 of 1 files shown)


common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -192,4 +194,6 @@ object WhiskConfig {
     val monitorHost = Map(monitorHostName -> null, monitorHostPort -> null)
     val zookeeperHost = Map(zookeeperHostName -> null, zookeeperHostPort -> null)
     val entitlementHost = Map(entitlementHostName -> "", entitlementHostPort -> "")
+
+    val edgeAddress = Map(edgeHostName -> null)
```
markusthoemmes: This is unneeded?

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -53,6 +53,8 @@ class WhiskConfig(
     val loadbalancerHost = this(WhiskConfig.loadbalancerHostName) + ":" + this(WhiskConfig.loadbalancerHostPort)
     val messagehubtriggerHost = this(WhiskConfig.messagehubtriggerHostName) + ":" + this(WhiskConfig.messagehubtriggerHostPort)
 
+    val edgeAdress = this(WhiskConfig.edgeHostName)
```
markusthoemmes: Name it `edgeHostName` to follow conventions of the object below

PR #714, opened 2016-06-21
'wsk property get' now displays all properties and their values.  Also updated .gitattributes for .py files'

mdeuser:
'wsk property get' now displays all properties and their values
Also updated .gitattributes for .py files and wsk bash/python files without .py/.sh extensions
  - These must retain LF EOL because the python CLI is built on vagrant/linux on Windows

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,7 +9,7 @@
 *.java          text
 *.js            text
 *.md            text
-*.py            text
+*.py            text eol=lf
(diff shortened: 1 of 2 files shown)


mdeuser:
@dubeejw @csantanapr - Ready for review (James - code, Carlos - .gitattributes)

dubeejw:
:+1:

Issue #577, opened 2016-06-06
Go CLI distribution


csantanapr:
Related to new Go-CLI  https://github.com/openwhisk/openwhisk/pull/238

The design for python CLI is to be available in `https://$apihost/cli/download`
A user will install using `$ sudo pip install --upgrade https://new-console.ng.bluemix.net/openwhisk/cli/download`

For Go-CLI we should have a good deprecation period and have both available, making transition smooth.

Have the go cli binaries available at `https://$apihost/cli/go/download`
The `go/download` directory will have the build out per platform and cpu architecture
```
linux\...
mac\...
windows\...
```
Nginx should be configure to list directory content
```
location cli/go/download {
               autoindex on;
        }
```
This allow user to navigate to find the download he wants.

Also something like a website bluemix can have direct links:

Some examples of direct url downloads
`https://$apihost/cli/go/download/mac/amd64/wsk`
`https://$apihost/cli/go/download/windows/amd64/wsk`
`https://$apihost/cli/go/download/linux/386/wsk`
`https://$apihost/cli/go/download/linux/amd64/wsk`

All the builds being generated today:
```
.
├── linux
│   ├── 386
│   │   └── wsk
│   ├── amd64
│   │   └── wsk
│   ├── arm
│   │   └── wsk
│   ├── arm64
│   │   └── wsk
│   ├── mips64
│   │   └── wsk
│   ├── mips64le
│   │   └── wsk
│   ├── ppc64
│   │   └── wsk
│   └── ppc64le
│       └── wsk
├── mac
│   ├── 386
│   │   └── wsk
│   ├── amd64
│   │   └── wsk
│   ├── arm
│   │   └── wsk
│   └── arm64
│       └── wsk
└── windows
    ├── 386
    │   └── wsk.exe
    └── amd64
        └── wsk.exe

```

mdeuser:
@nauerz-ibm previously mentioned that the wsk CLI should have parity with the Bluemix CLI's architecture/CPU support.  The Bluemix CLI download page http://clis.ng.bluemix.net/ui/home.html lists the following platform/architectures:

- Mac OS X 
- Linux 64-bit (need to confirm CPU)
- Windows 64-bit

Bluemix provides an installer for the Mac and Windows and a tgz for Linux.  Currently the OpenWhisk CLI is pretty simple, so perhaps just downloading the binary is sufficient.

csantanapr:
[...]

PR #726, opened 2016-06-21
Add Go CLI to OpenWhisk - Ansible

dubeejw:
- Setup Ansible so that Go CLI can be published with nginx (not included)

### DIFF
--- a/ansible/environments/mac/group_vars/all
+++ b/ansible/environments/mac/group_vars/all
@@ -3,6 +3,7 @@ nginx_conf_dir: /Users/Shared/nginx
 consul_conf_dir: /Users/Shared/consul
 whisk_logs_dir: /Users/Shared/wsklogs
 cli_conf_dir: /Users/Shared/cli
+go_cli_conf_dir: /Users/Shared/go-cli
 docker_registry: ""
 docker_dns: ""
 
--- a/ansible/roles/go-cli/tasks/deploy.yml
+++ b/ansible/roles/go-cli/tasks/deploy.yml
@@ -1,0 +1,20 @@
+---
+# Tasks for handling Go CLI customization and publishing
+
+- name: "Ensure go-cli directory exists"
+  file:
+    path: "{{ go_cli_conf_dir }}"
+    state: directory
+
+- name: "Pull the {{ docker_image_tag }} image of go-cli"
+  shell: "docker pull {{ docker_registry }}whisk/go-cli:{{ docker_image_tag }}"
+  when: docker_registry != ""
+
+# TODO: Publish compressed binaries using nginx
+
+- name: remove go-cli container after publishing
+  docker:
+    name: go-cli
+    image: "{{ docker_registry }}whisk/go-cli:{{ docker_image_tag }}"
+    state: absent
+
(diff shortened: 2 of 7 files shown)


dubeejw:
@mdeuser, please review.

mdeuser:
@dubeejw - I can take a look, but we should have one of the ansible experts eyes on this as well.

dubeejw:
@domdom82, can you take a look at this?

ansible/environments/mac/group_vars/all
```diff
@@ -3,6 +3,7 @@ nginx_conf_dir: /Users/Shared/nginx
 consul_conf_dir: /Users/Shared/consul
 whisk_logs_dir: /Users/Shared/wsklogs
 cli_conf_dir: /Users/Shared/cli
+go_cli_conf_dir: /tmp/go-cli
```
lehoanganh: pls place it under shared folder `/Users/Shared` like other folders. For mac, it's diffent.

ansible/roles/go-cli/tasks/deploy.yml
```diff
@@ -0,0 +1,20 @@
...
+- name: "Ensure go-cli directory exists"
+  file:
+    path: "{{ go_cli_conf_dir }}"
+    state: directory
+
+- name: "Pull the {{ docker_image_tag }} image of go-cli"
+  shell: "docker pull {{ docker_registry }}whisk/go-cli:{{ docker_image_tag }}"
+  when: docker_registry != ""
+
+# TODO: Publish compressed binaries using nginx
```
lehoanganh: you don't want to publish the tarball here?

ansible/roles/go-cli/templates/default.props.j2
```diff
@@ -0,0 +1,2 @@
+CLI_API_HOST={{ cli_api_host | default(groups['edge']) | first }}
```
lehoanganh: the go-cli needs also `default.props`?

ansible/roles/go-cli/tasks/deploy.yml
dubeejw: Why not? The tarballs will be in the Docker container at this point.

dubeejw: @lehoanganh, at this point in the Ansible file for the Python CLI, a generated tarball is moved to the nginx config directory. This is what we want to do here.

dubeejw: @lehoanganh, we have @houshengbo working on the publishing aspect mentioned here.

lehoanganh: I try to understand the process. The `go-cli` would do the same process like python cli does?

If so, at this comment

```
# TODO: Publish compressed binaries using nginx
```

Why don't we have steps to get the tarball in docker container and place it somewhere under nginx?

dubeejw: Vincent will be working on that part. He'll add it in afterward.

dubeejw: Plus the process isn't as simple because we have numerous tar/zip files now.

dubeejw: See https://github.com/openwhisk/openwhisk/issues/577

PR #722, opened 2016-06-21
Wait for Kafka and Zookeeper using the designated method

psuter:
This ports some of the functionality of isAlive to the Ansible deployment scripts. Before the change, Ansible only waits for the port to be bound, which does not guarantee that the services are actually available.

~~Appears to fix #713.~~

FYI @rabbah @perryibm

Issue #713, opened 2016-06-21
excessive delay on new deployment where invoker does not appear to receive kafka message from controller


rabbah:
Observed on master branch, running locally on a mac with docker-machine.
Steps to reproduce:

```
> redo setup prereqs couchdb initidb deploy
> wsk action invoke /whisk.system/samples/echo -u $(cat ansible/files/auth.guest)
```

Watch controller logs to confirm that kafka message posted to message bug:
```
> wskadmin syslog get 29 controller
...
[2016-06-21T14:50:18.584Z] [INFO] [#tid_29] [LoadBalancerService] user has 1 activations posted. Posted to invoke0[0][0]
```

Watch invoker logs and confirm that somewhere between 4-5 minutes later, the invoker should pick up the kafka message and invoke the action:
```
> wskadmin syslog get 29 invoker0 # repeat several minutes later
[2016-06-21T14:54:45.791Z] [INFO] [#tid_29] [Dispatcher] matching message to 1 handlers: invoker
```

@psuter @perryibm

rabbah:
@lehoanganh @domdom82 Please review the commit above (https://github.com/psuter/openwhisk/commit/7f8acd7def29f99d75de204fecd1a1e2ed6c8df4). The ansible roles for kafka and zookeeper were not correct and did not preserve the semantics of [isAlive](tools/health/isAlive). This caused some combination of zookeeper/kafka misconfiguration that surfaces in the invoker with the symptom described above. This symptom seems to resolve itself after several minutes. We have not yet completely explained the issue but have strong evidence that restoring `isAlive` semantics avoid the problem. For pedagogy, we should further explain the observed effect.

The reason this symptom does not surface in Travis CI or local tests (unless you repeat the steps outlined above) is that there are several minutes worth of tests that run before the first test that required the invoker to be involved.

All roles should be checked for semantic equivalence with `isAlive`. @psuter and I fixed the two immediate ones that are impacting/blocking work related to streaming db support (needed for zip file uploads and the like). You should verify the other components are not equally giving false positive "is alive" responses.

FYI @jeremiaswerner

rabbah:
Reopened - observing a similar issue with controller on active ack.

rabbah:
Bumping zookeeper/kafka gap hid the issue again. This is unsatisfactory - need to dig and understand what's going on with zk/kafka at startup.

@perryibm

lehoanganh:
you meant, if the changes in open, the blue deployment is stil failed?

[...]

PR #727, opened 2016-06-22
WIP: restore isAlive behavior for consul liveness check (issue #713).

rabbah:
FYI @psuter 

Will pile this on as part of #722.


rabbah:
Replaced with newer PR.

PR #736, opened 2016-06-22
Provide Insecure Mode for Go CLI

dubeejw:
- Allows for SSL certificate checking to be enabled, or disabled through a
  --insecure flag

### DIFF
--- a/tools/go-cli/go-whisk-cli/commands/commands.go
+++ b/tools/go-cli/go-whisk-cli/commands/commands.go
@@ -24,9 +24,46 @@ import (
 
     "../../go-whisk/whisk"
     "errors"
+    "github.com/spf13/cobra"
 )
 
 var client *whisk.Client
+
+func setupClientConfig(cmd *cobra.Command, args []string) (error){
+    var apiHostBaseUrl = fmt.Sprintf("https://%s/api/", Properties.APIHost)
+
+    baseURL, err := url.Parse(apiHostBaseUrl)
+
+    if err != nil {
+        whisk.Debug(whisk.DbgError, "url.Parse(%s) error: %s\n", apiHostBaseUrl, err)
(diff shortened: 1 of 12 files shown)


dubeejw:
@mdeuser, please review.

mdeuser:
@dubeejw - some comments
- IMHO - Use cobra as much as possible for argument processing.  Piggybacking moving the httpClient initialization away from init time... the httpClient initialization can be contained in a single `Pre*Run` cobra command function.  See https://github.com/spf13/cobra#prerun-or-postrun-hooks.  This way, commands not needing httpClient support would not have the overhead of initializing a httpClient.
- I recommend that parseArgs() focus solely on parsing arguments and that the setup() (was init()) function should be called outside of parseArgs()
- Remove the `\n` character from the errMsg strings - the string used to create the error objects.  See commands.go line 37

dubeejw:
@mdeuser, ready for a new review. I need to run it through the automation before it gets merged.

dubeejw:
All tests passed.

tools/go-cli/go-whisk-cli/commands/commands.go
```diff
@@ -24,10 +24,47 @@ import (
...
+    clientConfig := &whisk.Config{
+        AuthToken:  Properties.Auth,
+        Namespace:  Properties.Namespace,
+        BaseURL:    baseURL,
+        Version:    Properties.APIVersion,
+        Insecure:   flags.global.insecure,
+    }
+
+    // Setup client
+    client, err = whisk.NewClient(http.DefaultClient, clientConfig)
```
mdeuser: @dubeejw - The init() method also initializes the client variable with a new NewClient object.  Is this still needed?

dubeejw:
@mdeuser, how about now?

csantanapr:
rebase and merged 0e92f6cace3ab8d2fe8b7f8178c4603f450a7537

PR #741, opened 2016-06-22
Set CLI build version at build-time using gradle

mdeuser:
Set CLI build version using gradle + ansible generated properties file
Go CLI build version now comes from same default.properties as the python CLI
Also fix CLI build version logic cuz Go code was not getting build version from docker

### DIFF
--- a/docker.gradle
+++ b/docker.gradle
@@ -12,6 +12,7 @@
  * - dockerTimeout (optional, default 840): Timeout for docker operations in seconds
  * - dockerRetries (optional, default 3): How many times to retry docker operations
  * - dockerBinary (optional, default 'docker'): The binary to execute docker commands
+ * - dockerBuildArgs (options, default ''): Project specific custom docker build arguments
  * - dockerHost (optional): The docker host to run commands on, default behaviour is
  *       docker's own DOCKER_HOST environment variable
  */
@@ -22,14 +23,20 @@ ext {
     dockerTimeout = project.hasProperty('dockerTimeout') ? dockerTimeout.toInteger() : 840
     dockerRetries = project.hasProperty('dockerRetries') ? dockerRetries.toInteger() : 3
     dockerBinary = project.hasProperty('dockerBinary') ? [dockerBinary] : ['docker']
+    dockerBuildArg = ['build']
 }
 
 if(project.hasProperty('dockerHost')) {
     dockerBinary += ['--host', project.dockerHost]
 }
 
+if(project.hasProperty('dockerBuildArgs')) {
+    dockerBuildArg += ['--build-arg', project.dockerBuildArgs]
+}
+
 task distDocker << {
-    def cmd = dockerBinary + ['build', '-t', dockerImageName, project.buildscript.sourceFile.getParentFile().getAbsolutePath()]
+    def cmd
+    cmd = dockerBinary + dockerBuildArg + ['-t', dockerImageName, project.buildscript.sourceFile.getParentFile().getAbsolutePath()]
     retry(cmd, dockerRetries, dockerTimeout)
 }
 task tagImage << {
--- a/tools/go-cli/build.gradle
+++ b/tools/go-cli/build.gradle
@@ -1,5 +1,6 @@
 ext.dockerImageName = "whisk/go-cli"
 ext.dockerContainerName = "go-cli"
+ext.dockerBuildArgs = getDockerBuildArgs()
 apply from: "../../docker.gradle"
 
 // Detects if the local OS is Unix. If so, ant.properties.family is set to linux.
@@ -35,6 +36,13 @@ ant.condition(property: "arch", value: "386") {
 // Detects if the local CPU architecture is x86_32. If so, ant.properties.arch is set to 386.
 ant.condition(property: "arch", value: "386") {
     os(arch: "x86_32")
+}
+
+// Returns the Go CLI docker build args
+def getDockerBuildArgs() {
+    java.text.SimpleDateFormat sdf = new java.text.SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ")
+    String dateStr = sdf.format(new Date())
+    return "BUILDTIME="+dateStr
 }
 
 task removeBinary(type: Delete) {
--- a/tools/go-cli/go-whisk-cli/main.go
+++ b/tools/go-cli/go-whisk-cli/main.go
@@ -25,12 +25,20 @@ import (
     "../go-whisk-cli/commands"
 )
 
+// CLI_BUILD_TIME holds the time of the CLI build.  During gradle builds,
+// this value will be overwritten via the command:
+//     go build -ldflags "-X main.CLI_BUILD_TIME=nnnnn"   // nnnnn is the new timestamp
+var CLI_BUILD_TIME string = "not set"
+
 var cliDebug = os.Getenv("WSK_CLI_DEBUG")  // Useful for tracing init() code
 
 func init() {
     if len(cliDebug) > 0 {
         whisk.SetDebug(true)
     }
+
+    // Rest of CLI uses the Properties struct, so set the build time there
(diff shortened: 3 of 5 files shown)


mdeuser:
@dubeejw - Would you kindly review.  Thanks!

tools/go-cli/build.gradle
```diff
@@ -37,6 +38,19 @@ ant.condition(property: "arch", value: "386") {
     os(arch: "x86_32")
 }
 
+// Sets the ext.dockerBuildArgs from the whisk properties file that contains the build timestamp
+def setDockerBuildArgs() {
+    Properties props = new Properties()
+    String propsFileName = project.projectDir.getPath()+"/default.props"
+    props.load(new FileInputStream(propsFileName))
+    props.each { property ->
+        project.ext.set(property.key, property.value)
```
markusthoemmes: Is that needed, couldn't find another use of it

tools/go-cli/build.gradle
```diff
@@ -37,6 +38,19 @@ ant.condition(property: "arch", value: "386") {
     os(arch: "x86_32")
 }
 
+// Sets the ext.dockerBuildArgs from the whisk properties file that contains the build timestamp
+def setDockerBuildArgs() {
+    Properties props = new Properties()
+    String propsFileName = project.projectDir.getPath()+"/default.props"
+    props.load(new FileInputStream(propsFileName))
+    props.each { property ->
```
markusthoemmes: Can't the whole loop be rewritten with one `props.find { key, value -> key == "WHISK_VERSION_DATE") }`?

docker.gradle
```diff
@@ -22,14 +23,21 @@ ext {
...
 
 task distDocker << {
-    def cmd = dockerBinary + ['build', '-t', dockerImageName, project.buildscript.sourceFile.getParentFile().getAbsolutePath()]
+    def cmd
+    if (dockerBuildArgs != '') {
+        cmd = dockerBinary + ['build', '--build-arg', dockerBuildArgs, '-t', dockerImageName, project.buildscript.sourceFile.getParentFile().getAbsolutePath()]
+    }
+    else {
+        cmd = dockerBinary + ['build', '-t', dockerImageName, project.buildscript.sourceFile.getParentFile().getAbsolutePath()]
+    }
```
markusthoemmes: You might want to follow the same pattern as with `dockerHost`.  Check if the variable is set (don't define a default) and then create a new variable `dockerBuildArgsArgument` which holds the complete `--build-arg test1 test2`. That prevents us from having to nest the `if`s in the `distDocker` block more and more.

mdeuser: @markusthoemmes - good idea.  I see what was done above with dockerHost, and I'll have the build args follow the same process

tools/go-cli/build.gradle
mdeuser: @markusthoemmes - Yeah, the looping code was there for debug/trial.  As a Dictionary, it can be directly indexed with the key of interest.  I'll make that change.

tools/go-cli/build.gradle
mdeuser: @markusthoemmes - Getting my feet wet with gradle.  I now think all that needs be set is the ext.dockerBuildArgs.

tools/go-cli/build.gradle
markusthoemmes: ugh, you're right. even my idea was way too complicated thinking :D.

docker.gradle
mdeuser: @markusthoemmes - Oh.  I had been thinking that ansible was called prior to the gradle build.  If that's not the case, this logic is moot.  It looks like the build timestamp is set in the whisk.properties during the gradle build.  I'll use that instead - so, no need for any ansible updates re: setting this property in a default.props file (like the python CLI does)

markusthoemmes: The only dependence we have in gradle towards whisk.properties currently is for tests. All other tasks should ideally not be dependent on anything. For the python CLI, gradle builds the container, then it is started by ansible, populated with the proper default.props and then the `tar.gz` files are generated. Maybe we can establish the same approach here?

mdeuser: Currently, the CLI build version is baked into the Go CLI at build time.  No need to read an external props file for that value.  If I understand correctly, the whisk.properties is the basis for the default.props used by the python CLI.  If so, perhaps the Go CLI can leverage whisk.properties as well - just during build time

rabbah: Less dependence on props file is good.

tools/go-cli/go-whisk-cli/main.go
```diff
@@ -20,17 +20,30 @@ import (
...
 
 func init() {
     if len(cliDebug) > 0 {
         whisk.SetDebug(true)
     }
+
+    // Initialize a default CLI build timestamp value only if the value was not
+    // set via the go linker flags.
+    if CLI_BUILD_TIME == "not set" {
+        CLI_BUILD_TIME = time.Now().Format("2006-01-02T15:04:05-07:00")
```
dubeejw: Why's the CLI_BUILD_TIME  variable being set to a hardcoded value?

dubeejw: I'm not sure we should put a fake timestamp there. If for some reason a defect arose in the Docker build process preventing the timestamp from being written into the CLI_BUILD_TIME variable, we would never know. Without writing a default timestamp the automation would catch such a defect.

mdeuser: There's actually not a hardcoded timestamp.  It sure looks that way due to the way the Go time format string must be specified.  Unlike in Java, where the date format string looks something like "yyyy-MM-dd....", Go requires an actual arbitrary date and time.
http://stackoverflow.com/questions/20234104/how-to-format-current-time-using-a-yyyymmddhhmmss-format

The actual default timestamp is set to "now" - the time the init() function is called.  Yeah, I debated a bit about this default timestamp.  I wanted local builds to have a timestamp, but you are right in that it might give a false positive to the production builds

rabbah: How about 1 January 1970 ... (epoch start) then.

tools/go-cli/build.gradle
```diff
@@ -37,6 +38,13 @@ ant.condition(property: "arch", value: "386") {
     os(arch: "x86_32")
 }
 
+// Sets the ext.dockerBuildArgs with the build build timestamp
+def setDockerBuildArgs() {
+    java.text.SimpleDateFormat sdf = new java.text.SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ")
+    String dateStr = sdf.format(new Date())
+    ext.dockerBuildArgs = "BUILDTIME="+dateStr
+}
```
markusthoemmes: Could we change that function to actually `return` the values needed to pass to `ext.dockerBuildArgs` and assign them above? Makes the code easier to reason about.

tools/go-cli/go-whisk-cli/main.go
mdeuser: I'll just leave that local-build default as "not set", so no worries about the confusing Go date format string.  For more clarity, the format string can be in an arbitrary order, but the actual values for the year, etc fields, must be a specific numeric value.  For example, to format the year as a 4 digit year, `2006` must be used, for a 2 digit year `06` must be used.  Go uses numbers instead of alpha chars like 'yyyy' for many of the date format components.

tools/go-cli/build.gradle
mdeuser: Yes, that will make it more readable.

mdeuser:
@dubeejw @markusthoemmes - Thanks for your comments.  I think have have addressed them all now ;-)

dubeejw:
:+1:

csantanapr:
rebase and merged bf409c1aabc0087cbca0a65ae19683b1d07437f1

PR #742, opened 2016-06-23
Adds github feed deletion (issue #388).

### LINKED ISSUE #388: Implement the delete of webhook for github package
the Git package creates a webhook but is not deleting it. There is a missing correlation between the trigger and the webhook.

lzbj:
List all the existing triggers, delete the trigger with
the url that contains the specified trigger name.

Fixes #388

### DIFF
--- a/catalog/github/webhook.js
+++ b/catalog/github/webhook.js
@@ -27,12 +27,11 @@ function main(params) {
     }
   }
 
-  var endpoint = 'openwhisk.ng.bluemix.net';
   var lifecycleEvent = params.lifecycleEvent;
   var triggerName = params.triggerName.split("/");
 
     // URL of the whisk system. The calls of github will go here.
-  var whiskCallbackUrl = 'https://' + whisk.getAuthKey() + "@" + endpoint + '/api/v1/namespaces/' + encodeURIComponent(triggerName[1]) + '/triggers/' + encodeURIComponent(triggerName[2]);
+  var whiskCallbackUrl = 'https://' + whisk.getAuthKey() + "@" + params.endpoint + '/api/v1/namespaces/' + encodeURIComponent(triggerName[1]) + '/triggers/' + encodeURIComponent(triggerName[2]);
 
     // The URL to create the webhook on Github
   var registrationEndpoint = 'https://api.github.com/repos/' + (organization ? organization : username) + '/' + repository + '/hooks';
@@ -87,6 +86,61 @@ function main(params) {
       }
     });
 
+    return whisk.async();
+  } else if(lifecycleEvent === 'DELETE') {
+    //list all the existing webhooks first.
+    var options = {
+        method: 'GET',
+        url: registrationEndpoint,
+        headers: {
+            'Content-Type': 'application/json',
+            'Authorization': authorizationHeader,
+            'User-Agent': 'whisk'
+        }
+    }
+
+    request(options, function(error, response, body){
+      if (error){
+        whisk.error({
+          response: response,
+          error:error,
+          body:body
+        });
+      } else {
+        for(i=0; i<body.length;i++){
+            if (body[i].config.url == decodeURIComponent(whiskCallbackUrl)) {
+                var options = {
+                    method: 'DELETE',
+                    url: body[i].url,
+                    headers: {
+                        'Content-Type': 'application/json',
+                        'Authorization': authorizationHeader,
+                        'User-Agent': 'whisk'
+                    }
+                }
+
+                request(options, function(error, response, body) {
+                    if (error) {
+                        whisk.error({
+                          response: response,
+                          error:error,
+                          body:body
+                        });
+                    } else {
+                        console.log("Status code: " + response.statusCode);
+                        if(response.statusCode >= 400) {
+                            console.log("Response from Github: " + body);
+                            whisk.error({
+                                statusCode: response.statusCode,
+                                response: body
+                            });
+                        }
+                    }
+                });
+            }
+        }
+      }
+    });
     return whisk.async();
   }
 
--- a/catalog/installGit.sh
+++ b/catalog/installGit.sh
@@ -12,6 +12,7 @@ source "$CATALOG_HOME/util.sh"
 echo Installing Git package.
 
 createPackage github \
+    -p endpoint "openwhisk.ng.bluemix.net"
     -a description "Package which contains actions and feeds to interact with Github"
 
 waitForAll


lzbj:
Just for a temp code review, I'll add unitary test and other necessary file changes if the solution is acceptable.

catalog/github/webhook.js
```diff
@@ -7,58 +7,106 @@ var request = require('request');
...
+ *  @param {string} endpoint - The endpoint the openwhisk deployed on
  *  @return {object} whisk async
  */
 function main(params) {
-  var username = params.username;
-  var repository = params.repository;
-  var accessToken = params.accessToken;
+    var username = params.username;
+    var repository = params.repository;
+    var accessToken = params.accessToken;
```
rabbah: Remove endpoint here instead bind as a Paramater on package for default.

catalog/github/webhook.js
```diff
@@ -7,58 +7,106 @@ var request = require('request');
...
+    if (lifecycleEvent === 'CREATE') {
+        var events = params.events.split(',');
 
-    var body = {
-      name: 'web',
-      active: true,
-      events: events,
-      config: {
-        url: whiskCallbackUrl,
-        content_type: 'json'
```
rabbah: Either return in if block above with return whisk.error or guard next code block with else.

rabbah: Applied everywhere there is error or done call.

catalog/github/webhook.js
```diff
@@ -7,58 +7,106 @@ var request = require('request');
...
+                        } else {
+                            console.log("Status code: " + response.statusCode);
+                        }
+                    })
+
+                }
+            }
+        });
+    }
+    return whisk.done();
```
rabbah: Is this in the wrong place?

catalog/github/webhook.js
lzbj: I'm not very clear, could you please give a little more info, thank you.

lzbj: @rabbah ,I'm not very clear, could you please give a little more info, thank you.

csantanapr:
@lzbj @houshengbo Are you actively working on this, its very important to addresses important github package bug.

cc @bjustin-ibm

catalog/github/webhook.js
bjustin-ibm: @lzbj I believe what Rodric is suggesting is that, instead of setting the default endpoint value in code, you should specify the value when the package is created.  Please see [this example](https://github.com/openwhisk/openwhisk/blob/master/catalog/installWeather.sh#L15) in the script which installs the weather package. Then your code can just refer to `params.endpoint` with confidence that if the parameter was not explicitly set when creating the feed, then the value will already be the default.

@rabbah Please confirm this was your intention.

lzbj:
@csantanapr ,yes, I will try to submit a pr according to latest comments.

catalog/github/webhook.js
lzbj: @bjustin-ibm , thank you, I'll update it.

lzbj:
@rabbah @bjustin-ibm , please help review, thank you.

catalog/installGit.sh
```diff
@@ -20,7 +20,7 @@ install "$CATALOG_HOME/github/webhook.js" \
     github/webhook \
     -a feed true \
     -a description 'Creates a webhook on github to be notified on selected changes' \
-    -a parameters '[ {"name":"username", "required":true, "bindTime":true}, {"name":"repository", "required":true, "bindTime":true}, {"name":"accessToken", "required":true, "bindTime":true},{"name":"events", "required":true} ]' \
+    -a parameters '[ {"name":"username", "required":true, "bindTime":true}, {"name":"repository", "required":true, "bindTime":true}, {"name":"accessToken", "required":true, "bindTime":true}, {"name":"endpoint", "required":false, "bindTime:true}, {"name":"events", "required":true} ]' \
```
bjustin-ibm: @lzbj I believe this is not quite what @rabbah was suggesting.  Please see e0a628ea6fc851c85a381f9618737012c7e482c3 for how to set and use the default value as a parameter on the package itself.

catalog/github/webhook.js
```diff
@@ -88,6 +88,66 @@ function main(params) {
...
     return whisk.async();
+  } else if(lifecycleEvent === 'DELETE') {
+    //list all the existing webhooks first.
+    var options = {
+        method: 'GET',
+        url: registrationEndpoint,
+        headers: {
+            'Content-Type': 'application/json',
+            'Authorization': authorizationHeader,
+            'User-Agent', 'whisk'
```
bjustin-ibm: @lzbj Syntax error. Should be a colon instead of a comma. Please ensure that your code runs as expected before putting up for review.

catalog/github/webhook.js
```diff
@@ -88,6 +88,66 @@ function main(params) {
...
+          body:body
+        });
+      } else {
+        for(i=0; i<body.length: i++){
+            var urlpath = encodeURIComponent(body[i].config.url);
+            var pathname = url.parse(urlpath).pathname;
+            //if the url contains the triggerName[1], then delete the webhook
+            //so this delete all the trigger start with triggerName[1] such as
+            // trigger1, trigger2.
+            if (pathname.search(encodeURIComponent(triggerName[1])) != -1) {
```
bjustin-ibm: This logic can easily result in false-positives. For example, if the trigger name contains "trigger" then every GH webhook will match and be removed. Since the OW trigger endpoint url is formalized, you could instead check the trigger name against the url path component that follows the "triggers" segment.

bjustin-ibm: Which brings up an interesting idea... Instead of parsing the webhook's `config.url` property to try to pick out the namespace and trigger name, I wonder if you could simply compare it for equality with the `whiskCallbackUrl` variable?

lzbj: Yes, good idea, I'll test it.

catalog/github/webhook.js
lzbj: @bjustin-ibm , thank you, yes, I'll do a test next before propose for review.

catalog/github/webhook.js
lzbj: @bjustin-ibm , is the latest change is what you suggested? thank you.

catalog/installGit.sh
```diff
@@ -12,14 +12,15 @@ source "$CATALOG_HOME/util.sh"
...
+    -p endpoint "openwhisk.ng.bluemix.net"
     -a description "Package which contains actions and feeds to interact with Github"
 
 waitForAll
 
 install "$CATALOG_HOME/github/webhook.js" \
     github/webhook \
     -a feed true \
-    -a description 'Creates a webhook on github to be notified on selected changes' \
+    -a description 'Creates a webhook on github to be notified on selected changes or deteles the webhook' \
```
starpit: misspelling: deteles -> deletes

lzbj: thank you @starpit

catalog/installGit.sh
```diff
@@ -12,14 +12,15 @@ source "$CATALOG_HOME/util.sh"
...
+    -p endpoint "openwhisk.ng.bluemix.net"
     -a description "Package which contains actions and feeds to interact with Github"
 
 waitForAll
 
 install "$CATALOG_HOME/github/webhook.js" \
     github/webhook \
     -a feed true \
-    -a description 'Creates a webhook on github to be notified on selected changes' \
+    -a description 'Creates a webhook on github to be notified on selected changes or deletes the webhook' \
```
starpit: i don't think we should document the "Deletes the web hook" part, as this is an internal part of open whisk, not something relevant to end users. in other words, when they delete the trigger, the webhook will be deleted, this is just open whisk functioning as it should..

the intent of this description field is to help end users understand the offering.

bjustin-ibm:
@lzbj Thank you for contributing this work. The GitHub trigger code has been moved to the [openwhisk/openwhisk-catalog](https://github.com/openwhisk/openwhisk-catalog) repo. I have migrated your changes over to that repo as well as added fixes and additional error handling. Please follow along at openwhisk/openwhisk-catalog#139

PR #749, opened 2016-06-23
redo to support custom targets

rabbah:
Added ability to deploy (via ansible playbooks) targets using inferred YAML file if component is not one of known (builtin) targets.


sjfink:
rebased and pushed.

PR #753, opened 2016-06-24
[Ready] Tweaked entitlement checks to bypass entitlement service entirely for default namespace

rabbah:
Tweaked entitlement checks so that the default namespace is always checked and does not require a query from an entitlement service even if such service exists. Only if the implicit entitlement fails is the service queried for additional namespaces to check.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/entitlement/RemoteEntitlement.scala
+++ b/core/controller/src/main/scala/whisk/core/entitlement/RemoteEntitlement.scala
@@ -45,7 +45,7 @@ import whisk.core.entity.Subject
 import whisk.core.WhiskConfig
 
 protected[core] class RemoteEntitlementService(
-    private val config : WhiskConfig,
+    private val config: WhiskConfig,
     private val timeout: FiniteDuration = 5 seconds,
     private implicit val actorSystem: ActorSystem,
     private implicit val ec: ExecutionContext)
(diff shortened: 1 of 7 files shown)


rabbah:
PPG 681 approved.

markusthoemmes:
LGTM, memo to self to merge ASAP

core/controller/src/main/scala/whisk/core/entitlement/RemoteEntitlement.scala
```diff
@@ -64,7 +64,7 @@ protected[core] class RemoteEntitlementService(
             addHeader("X-Transaction-Id", transid.toString())
             ~> sendReceive
             ~> unmarshal[List[String]])
-        pipeline(Get(url))
+        pipeline(Get(url)) map { _.toSet }
```
markusthoemmes: Could unmarshal to a `Set` directly to make the extra step obsolete?

rabbah: https://github.com/openwhisk/openwhisk/pull/753/commits/330835ec0bbfa96376d12fa1626a4ebfd03f8339

markusthoemmes:
rebased and merged in 7360b84498e024b2c1d71767d3460af5efbbc308

PR #763, opened 2016-06-24
Dumbing down the terminal to keep Gradle quiet

psuter:
Affects TravisCI only. Pretends the terminal is crippled; disables the `> Loading` CR-animations that make the log hard to read. 

~~@domdom82 you had the last change in the file :)~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,10 @@ scala:
 services:
   - docker
 
+env:
+  global:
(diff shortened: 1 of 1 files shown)


rabbah:
Rebased and merged as https://github.com/openwhisk/openwhisk/commit/ac3b9f5c6743f2a17e695829944a34eefb0815f1.

PR #770, opened 2016-06-26
Update Invalid Argument Handling for Actions

dubeejw:
- Improve error handling when an invalid number of arguments are passed to action commands
- Display an error when an invalid kind is passed when creating, or updating an action
- Display an error when an invalid file extension is used to create, or update an action
- Remove colors

### DIFF
--- a/tests/src/system/basic/WskBasicTests.scala
+++ b/tests/src/system/basic/WskBasicTests.scala
@@ -210,6 +210,18 @@ class WskBasicTests
             wsk.action.delete(name, expectedExitCode = CONFLICT)
     }
 
+    it should "reject delete of an action without an action name" in {
+        val stderr = wsk.cli(Seq("action", "delete"), expectedExitCode = ERROR_EXIT).stderr
+        stderr should include("error: Invalid argument(s). An action name is required.")
+        stderr should include("Run 'wsk --help' for usage.")
+    }
+
+    it should "reject delete of an action with an invalid argument" in {
+        val stderr = wsk.cli(Seq("action", "delete", "actionName", "invalidArg"), expectedExitCode = ERROR_EXIT).stderr
+        stderr should include("error: Invalid argument(s): invalidArg")
+        stderr should include("Run 'wsk --help' for usage.")
+    }
+
     it should "reject creating entities with invalid names" in withAssetCleaner(wskprops) {
         (wp, assetHelper) =>
             val names = Seq(
@@ -252,6 +264,18 @@ class WskBasicTests
         }
     }
 
+    it should "reject get of an action without an action name" in {
+        val stderr = wsk.cli(Seq("action", "get"), expectedExitCode = ERROR_EXIT).stderr
+        stderr should include("error: Invalid argument(s). An action name is required.")
+        stderr should include("Run 'wsk --help' for usage.")
+    }
+
+    it should "reject get of an action with an invalid argument" in {
+        val stderr = wsk.cli(Seq("action", "get", "actionName", "invalidArg"), expectedExitCode = ERROR_EXIT).stderr
+        stderr should include("error: Invalid argument(s): invalidArg")
+        stderr should include("Run 'wsk --help' for usage.")
+    }
+
     it should "reject update action in shared package not owned by authkey" in {
         wsk.action.create("/whisk.system/util/cat", None,
             update = true, shared = Some(true), expectedExitCode = FORBIDDEN)
@@ -292,6 +316,12 @@ class WskBasicTests
         val result = wsk.action.list(Some("/whisk.system/util")).stdout
         result should include regex ("""/whisk.system/util/head\s+shared""")
         result should include regex ("""/whisk.system/util/date\s+shared""")
+    }
+
+    it should "reject list of an action with an invalid argument" in {
+        val stderr = wsk.cli(Seq("action", "list", "actionName", "invalidArg"), expectedExitCode = ERROR_EXIT).stderr
+        stderr should include("error: Invalid argument(s): invalidArg")
+        stderr should include("Run 'wsk --help' for usage.")
     }
 
     it should "create, update, get and list a package" in withAssetCleaner(wskprops) {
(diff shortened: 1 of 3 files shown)


tools/go-cli/go-whisk-cli/commands/action.go
```diff
@@ -46,45 +44,28 @@ var actionCmd = &cobra.Command{
...
+
+        if len(args) < 2 {
+            whisk.Debug(whisk.DbgError, "Action create command must have at least two arguments\n")
+            errMsg := fmt.Sprintf("Invalid argument(s). An action name, and artifact are required.")
+            whiskErr := whisk.MakeWskError(errors.New(errMsg), whisk.EXITCODE_ERR_GENERAL,
+                whisk.DISPLAY_MSG, whisk.DISPLAY_USAGE)
+            return whiskErr
+        } else if len(args) > 2 {
+            whisk.Debug(whisk.DbgError, "Action create command must not have more than two arguments\n")
+            errMsg := fmt.Sprintf("Invalid argument(s): %s.", strings.Join(args[2:], ", "))
```
mdeuser: I'm not sure if this error message is for python compatibility.. If not, I think it would be helpful to note that there are too many arguments.

tools/go-cli/go-whisk-cli/commands/action.go
```diff
@@ -230,17 +199,13 @@ var actionInvokeCmd = &cobra.Command{
...
         if flags.common.blocking && flags.action.result {
-            //printJSON(activation.Response.Result)
             printJsonNoColor(activation.Response.Result)
         } else if flags.common.blocking {
-            fmt.Printf("%s invoked /%s/%s with id %s\n", color.GreenString("ok:"), boldString(qName.namespace), boldString(qName.entityName),
-                boldString(activation.ActivationID))
-            boldPrintf("response:\n")
-            //printJSON(activation)
+            fmt.Printf("ok: invoked /%s/%s with id %s\n", qName.namespace, qName.entityName, activation.ActivationID)
+            fmt.Println("response:")
```
mdeuser: Missing '\n' ?

tools/go-cli/go-whisk-cli/commands/action.go
```diff
@@ -592,10 +563,12 @@ func parseAction(cmd *cobra.Command, args []string) (*whisk.Action, bool, error)
...
 
     } else if artifact != "" {
-        stat, err := os.Stat(artifact)
+        ext := filepath.Ext(artifact)
+
+        _, err := os.Stat(artifact)
         if err != nil {
             whisk.Debug(whisk.DbgError, "os.Stat(%s) error: %s\n", artifact, err)
-            errMsg := fmt.Sprintf("File '%s' does not exist: %s", artifact, err)
+            errMsg := fmt.Sprintf("File '%s' does not exist.", artifact)
```
mdeuser: os.Stat() errors can mean either 1) the path does not exist or 2) some other error.
http://stackoverflow.com/questions/10510691/how-to-check-whether-a-file-or-directory-denoted-by-a-path-exists-in-golang

tools/go-cli/go-whisk-cli/commands/action.go
```diff
@@ -730,6 +709,7 @@ func init() {
     actionUpdateCmd.Flags().BoolVar(&flags.action.docker, "docker", false, "treat artifact as docker image path on dockerhub")
     actionUpdateCmd.Flags().BoolVar(&flags.action.copy, "copy", false, "treat artifact as the name of an existing action")
     actionUpdateCmd.Flags().BoolVar(&flags.action.sequence, "sequence", false, "treat artifact as comma separated sequence of actions to invoke")
+    actionUpdateCmd.Flags().StringVar(&flags.action.kind, "kind", "", "the kind of the action runtime (example: swift:3)")
```
mdeuser: What are the supported kind types? It would be helpful to include the exact valid kind strings in the --kind help.

tools/go-cli/go-whisk-cli/commands/action.go
dubeejw: Replaced Printf with Println.

tools/go-cli/go-whisk-cli/commands/action.go
```diff
@@ -621,13 +603,19 @@ func parseAction(cmd *cobra.Command, args []string) (*whisk.Action, bool, error)
         } else if flags.action.kind == "nodejs:6" || flags.action.kind == "nodejs:6.0" ||
             flags.action.kind == "nodejs:6.0.0" {
             action.Exec.Kind = "nodejs:6"
-        } else if matched, _ := regexp.MatchString(".swift$", stat.Name()); matched {
+        } else if len(flags.action.kind) > 0 {
```
mdeuser: It seems that python and java runtime are supported in the code below that examines the extension (.py, .jar).  Should there be --kind values of `python` and `java`?

tools/go-cli/go-whisk-cli/commands/action.go
```diff
@@ -637,8 +625,8 @@ func parseAction(cmd *cobra.Command, args []string) (*whisk.Action, bool, error)
                 return nil, sharedSet, err
             }
         } else {
-            whisk.Debug(whisk.DbgError, "--kind argument '%s' is invalid\n", flags.action.kind)
-            errMsg := fmt.Sprintf("'%s' is not a supported action runtime", flags.action.kind)
+            whisk.Debug(whisk.DbgError, "Action runtime exention '%s' is not supported\n", ext)
```
mdeuser: I'm not sure we're consistent wrt supported file extensions.  If the --kind flag has a valid kind value, it seems that the extension is not checked, so the file could have any extension or none if the --kind is either swift:3 or nodes:6

tools/go-cli/go-whisk-cli/commands/action.go
dubeejw: I think it would be inaccurate to say that there has been too many arguments passed here. This case only considers how many arguments were passed without the -- prefix. Therefore, we would be telling the user that her, or she has supplied too many arguments without counting how many arguments were passed using the -- prefix.

tools/go-cli/go-whisk-cli/commands/action.go
dubeejw: There are multiple runtimes for Swift, and NodeJs. The --kind argument is needed for those to specify which version to use. Where as Python, and Java only have one runtime a piece.

tools/go-cli/go-whisk-cli/commands/action.go
dubeejw: I believe there will be .zip file support in the future. That means a user would be able to specify a --kind swift, and upload a zip file.

dubeejw: Until then we could add an extra check matching the kind with the file extension.

mdeuser: Perhaps there's a mismatch between the --lib help text `must be a gzipped tar file` and this code which unpacks either a .tar or a .gzip, but not a .tgz.

dubeejw: Creating actions with the --lib flag is no longer supported. I will remove the logic.

rabbah: 👍 we'll bring this back when the support is there again.

dubeejw:
Fixes https://github.com/openwhisk/openwhisk/issues/819

dubeejw:
IPG 359.

dubeejw:
I've included more tests. @mdeuser, please take another look.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -168,6 +168,16 @@ class WskBasicTests
             wsk.action.delete(name, expectedExitCode = CONFLICT)
     }
 
+    it should "reject delete of an action without an action name" in {
+        wsk.cli(Seq("action", "delete"), expectedExitCode = ERROR_EXIT).
+          stderr should include("error: Invalid argument(s). An action name is required.")
```
mdeuser: For this and the other tests, is the cli usage string also expected?

rabbah: It gets annoying I think if we always print usage information. I suggest we should do that deliberately in a few cases only.

dubeejw:
IPG 396.

dubeejw:
PP 893.

dubeejw:
@mdeuser, please merge if there are no additional comments.

csantanapr:
rebase and merge 7c04bd20407be6b72c842ceeda220417d9f5c9d5

PR #772, opened 2016-06-27
Update Go CLI Package Argument Errors

dubeejw:
- Display more information about invalid arguments

### DIFF
--- a/tests/src/system/basic/WskBasicTests.scala
+++ b/tests/src/system/basic/WskBasicTests.scala
@@ -337,6 +337,87 @@ class WskBasicTests
             stdout should not include regex(""""key": "xxx"""")
     }
 
+    it should "reject create of a package without a package name" in {
+        val stderr = wsk.cli(Seq("package", "create"), expectedExitCode = ERROR_EXIT).stderr
+        stderr should include("error: Invalid argument(s). A package name is required.")
+        stderr should include("Run 'wsk --help' for usage.")
+    }
+
+    it should "reject create of a package with an invalid argument" in {
+        val stderr = wsk.cli(Seq("package", "create", "packageName", "invalidArg"),
+            expectedExitCode = ERROR_EXIT).stderr
+        stderr should include("error: Invalid argument(s): invalidArg")
+        stderr should include("Run 'wsk --help' for usage.")
+    }
+
+    it should "reject update of a package without a package name" in {
+        val stderr = wsk.cli(Seq("package", "update"), expectedExitCode = ERROR_EXIT).stderr
+        stderr should include("error: Invalid argument(s). A package name is required.")
+        stderr should include("Run 'wsk --help' for usage.")
+    }
+
+    it should "reject update of a package with an invalid argument" in {
+        val stderr = wsk.cli(Seq("package", "update", "packageName", "invalidArg"),
+            expectedExitCode = ERROR_EXIT).stderr
+        stderr should include("error: Invalid argument(s): invalidArg")
+        stderr should include("Run 'wsk --help' for usage.")
+    }
+
+    it should "reject get of a package without a package name" in {
+        val stderr = wsk.cli(Seq("package", "get"), expectedExitCode = ERROR_EXIT).stderr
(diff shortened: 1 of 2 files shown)


tools/go-cli/go-whisk-cli/commands/package.go
```diff
@@ -76,11 +40,19 @@ var packageBindCmd = &cobra.Command{
...
+
+        if len(args) < 2 {
+            whisk.Debug(whisk.DbgError, "Package bind command must have at least one argument\n")
+            errMsg := fmt.Sprintf("Invalid argument(s). A package name, and binding name are required.")
+            whiskErr := whisk.MakeWskError(errors.New(errMsg), whisk.EXITCODE_ERR_GENERAL,
+                whisk.DISPLAY_MSG, whisk.DISPLAY_USAGE)
+            return whiskErr
+        } else if len(args) > 2 {
+            whisk.Debug(whisk.DbgError, "Package bind command must not have more than two arguments\n")
+            errMsg := fmt.Sprintf("Invalid argument(s): %s.", strings.Join(args[2:], ", "))
```
mdeuser: How about including the reason why the argument are invalid (i.e. too many arguments)? Otherwise, one might think the arguments might just be misspelled or something like that.

dubeejw: I think it would be inaccurate to say that there has been too many arguments passed here. This case only considers how many arguments were passed without the -- prefix. Therefore, we would be telling the user that her, or she has supplied too many arguments without counting how many arguments were passed using the -- prefix.

mdeuser: Current error message regarding number of expected arguments looks good.

dubeejw:
@mdeuser, added tests. Please take another look.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -274,6 +274,71 @@ class WskBasicTests
...
+          stderr should include("error: Invalid argument(s). A package name is required.")
+    }
+
+    it should "reject update of a package with an invalid argument" in {
+        wsk.cli(Seq("package", "update", "packageName", "invalidArg"), expectedExitCode = ERROR_EXIT).
+          stderr should include("error: Invalid argument(s): invalidArg")
+    }
+
+    it should "reject get of a package without a package name" in {
+        wsk.cli(Seq("package", "update"), expectedExitCode = ERROR_EXIT).
```
rabbah: another untested feature of the python CLI - a get with an additional argument will project that one argument from the response; this is handy if you just want to see one property form an otherwise large record:

```
> wsk package get p parameters
ok: got package p, projecting parameters
[]
```

this works for all the entities in the python cli.

```
> wsk package get p bogus
ok: got package p, but it does not contain property bogus
```

mdeuser:
Minor comment.  The help text for the `package list` command should be updated now that the `namespace` argument is optional.  From `wsk package list <namespace string> [flags]` to something like `wsk package list [namespace] [flags]`.  I've not yet discovered how to display per-argument help text, so that argument label may need some bolstering to provide missing details (i.e. value format)

mbehrendt:
@mdeuser do we have the optional `namespace` parameter now for all `wsk <action/trigger/rule/activation> list` commands?

dubeejw:
@mbehrendt, the namespace is optional for all the mentioned commands. The help for the Go CLI needs to be updated to reflect the optional argument.

mdeuser:
@mbehrendt - Thanks for the the `activation list` support is not there yet.  Yes, the help syntax for several commands need updating as well.

mbehrendt:
sounds good -- thank you

rabbah:
A namespace for activations should be ignored as all activations for a subject go to one namespace.

mdeuser:
Updates look good to me.

dubeejw:
@mdeuser, please merge.

csantanapr:
rebase and merge 22bc79764f63a6524bc181bddcb5999951c70886

PR #773, opened 2016-06-27
Bump kafka to 0.9.0.1 to fix high cpu utilization

### LINKED ISSUE #701: Excessive CPU usage during no load using "vagrant/simple"?
Following the [Quick Start](https://github.com/openwhisk/openwhisk#quick-start) guidelines, I have successfully set up a local instance of the OpenWhisk platform.

**Looking at the VM instance, I've noticed the system is under high CPU load at all times, even when the system is idle.**

Here's the output from _top_ on the system. 
ow
Those top four processes (10264, 10216, 10990 & 11772) remain at similar CPU usage whilst the system is idle. 

10264 is the Kafka process, 10216 is docker-proxy for the 9092 (kafka's port), the other two processes are the controller and invoker instances.

I've increased the VM memory from 4GB to 8GB with no difference.

Is this expected behaviour?

danielfederschmidt:
This PR resolves #701 by bumping kafka to the next minor version because investigation showed that a fix provided by this version resolves this issue.

> [KAFKA-3159] - Kafka consumer 0.9.0.0 client poll is very CPU intensive under certain conditions

### DIFF
--- a/services/kafka/Dockerfile
+++ b/services/kafka/Dockerfile
@@ -19,7 +19,10 @@ RUN curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-
 
 # Install Kafka Stuff
 RUN mkdir /kafka /data
-ENV KAFKA_RELEASE_ARCHIVE kafka_2.11-0.9.0.0.tgz
+
+ENV KAFKA_RELEASE 0.9.0.1
+ENV KAFKA_SCALA_VERSION 2.11
+ENV KAFKA_RELEASE_ARCHIVE kafka_${KAFKA_SCALA_VERSION}-${KAFKA_RELEASE}.tgz
 
 # Download Kafka binary distribution
 #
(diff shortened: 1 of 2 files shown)


services/kafka/Dockerfile
```diff
@@ -28,8 +28,8 @@ ENV KAFKA_RELEASE_ARCHIVE kafka_2.11-0.9.0.0.tgz
 # Install Kafka to /kafka
 
 WORKDIR /tmp
-RUN wget --no-verbose http://www.us.apache.org/dist/kafka/0.9.0.0/${KAFKA_RELEASE_ARCHIVE} && \
-    wget --no-verbose https://dist.apache.org/repos/dist/release/kafka/0.9.0.0/${KAFKA_RELEASE_ARCHIVE}.md5  && \
+RUN wget --no-verbose http://www.us.apache.org/dist/kafka/0.9.0.1/${KAFKA_RELEASE_ARCHIVE} && \
+    wget --no-verbose https://dist.apache.org/repos/dist/release/kafka/0.9.0.1/${KAFKA_RELEASE_ARCHIVE}.md5  && \
```
markusthoemmes: While we're at it: How about adding an ENV var above which contains the Kafka version like

`KAFKA_VERSION 0.9.0.1` and use it both in the `RELEASE_ARCHIVE` and the urls?

danielfederschmidt: yeah, I noticed that as well - I'll change that.

perryibm:
Change the comment at the top also to say 0.9.0.1.

perryibm:
This does not fix the high CPU problem and @danielfederschmidt double-checked and confirmed that it does not fix the high CPU problem.

rabbah:
superceded by #796.

PR #776, opened 2016-06-27
Updating Mac instructions to install docker 1.9.1 locally

markusthoemmes:
This is for it to be consistent with docker-machine's setup.

### DIFF
--- a/tools/macos/README.md
+++ b/tools/macos/README.md
@@ -24,8 +24,8 @@ echo '
 brew tap caskroom/cask
 # install virtualbox
 brew cask install virtualbox
-# install docker 1.10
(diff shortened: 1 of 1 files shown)


csantanapr:
I had problems with docker 1.9.1
I think we should use latest GA version of docker on native mac 1.11.2
When I switched to 1.11.2 I had better luck stopping and cleaning kafka container.

csantanapr:
It looks like it was a kafka problem and not docker, we will stick with 1.9.1 to have a good devops story, sicne 1.9.1 its what we use to deploy.

csantanapr:
rebase and merged a0b4c1dc112ffcad8a9bb33601de7d19d164fcff

Issue #748, opened 2016-06-23
 Circular dependency of openwhisk and openwhisk-catalog


houshengbo:
Since the packages and install scripts are moved into openwhisk-catalog, how shall we cope with the dependencies between openwhisk and openwhisk-catalog?

If we still keep an install script in openwhisk, each time we run it, it git clone the openwhisk-catalog repo and install the packages.
If we run the install script in openwhisk-catalog, it will clone the openwhisk repo and install the packages.

This situation sustains in the Travis CI as well. openwhisk and openwhisk-catalog both need Travis CI to run the tests on each PRs to ensure the code quality. However, each of them needs one another's repo to install the catalog.

We need to break the circular dependency in either of the projects.

houshengbo:
The pair issue in openwhisk-catalog: https://github.com/openwhisk/openwhisk-catalog/issues/24

houshengbo:
I personally suggest to remove the install script in openwhisk. The openwhisk-catalog will pull openwhisk repo for the catalog installation tests each time the code is changed.

I also prefer to split catalog installation from the installation of openwhisk. Now we are trying to install everything when we install openwhisk, which I think there is a potential to improve.

Installation of OpenWhisk can be abstracted into a general project or somewhere we feel comfortable (like the role of devstack plays in openstack). It may install the openwhisk core(we can define late what it means), it may install the catalogs, it may install third party action... The role is to be responsible for deployment.

tuler:
I like the idea of keeping the installation of OpenWhisk as a separate entity. The current ansible approach is confusing, and it supports only simple and specific deployment scenarios. In addition to that the building and installation are too connected. We should be capable of building and producing artifacts (as docker images) and use the same images for different deployment scenarios.

markusthoemmes:
@tuler in which way are building and deployment connected currently? If you find the current approach confusing, feedback will be very welcome so we can improve the processes.

houshengbo:
From the OpenWhisk's perspective, to break the dependency on the installation of the catalog packages, I propose to work on the following item(s):

*Remove or replace the catalog installation from the test in OpenWhisk. By far, OpenWhisk installs all the packages by default and there are tests regarding all the package installation. We should either remove them out as well or replace them with some fixtures or mocks. This could be the first step to break the catalog dependencies in OpenWhisk. After this, the Travis CI build to check each PR won't break at the catalog package check.

tuler:
[...]

PR #720, opened 2016-06-21
[WIP] Add the script to install the catalog from openwhisk-catalog

houshengbo:
Since all the packages have been relocated to openwhisk-catalog,
this patch adds a script under openwhisk to pull all the packages
from openwhisk-catalog and install them.

Closes-Bug: #719

PR #780, opened 2016-06-27
[WIP]: Remove the catalog installation and the tests

houshengbo:
This patch targets to remove the catalog installation from OpenWhisk,
and the tests for them, so that OpenWhisk will not install catalog
by default. Instead, the tests need to be created in OenWhisk-catalog,
which is hosting the new location for all the catalog.

Partial-Closes-Bug: #748


houshengbo:
@csantanapr Yes.
I am narrowing down the scope for the tests now into:
```
#       deleted:    tests/src/packages/SamplesTests.java
#       deleted:    tests/src/packages/SlackTests.java
#       deleted:    tests/src/packages/UtilsTests.scala
#       deleted:    tests/src/packages/websocket/WebSocketTests.scala
#       **deleted:    tests/src/services/HeadersTests.scala**
#       deleted:    tests/src/system/basic/CLIActionTests.java
#       deleted:    tests/src/system/basic/CLIJavaTests.scala
#       deleted:    tests/src/system/basic/CLIPythonTests.scala
#       deleted:    tests/src/system/basic/CLIRuleTests.java
#       deleted:    tests/src/system/basic/CLISequentialTests.java
#       deleted:    tests/src/system/basic/CLISwiftTests.scala
#       deleted:    tests/src/system/basic/ConsoleTests.scala
#       deleted:    tests/src/system/basic/PackageTests.scala
#       deleted:    tests/src/system/basic/Swift3WhiskObjectTests.scala
#       deleted:    tests/src/system/basic/WskActionSequenceTests.scala
#       deleted:    tests/src/system/basic/WskBasicTests.scala
#       deleted:    tests/src/system/rest/ActionSchemaTests.scala
#       **deleted:    tests/src/whisk/core/container/test/ContainerPoolTests.scala**
```

I still need to look at HeadersTests.scala and ContainerPoolTests.scala, because they are not quite related to catalog, but will fail without catalog installation.

rabbah:
There are several of these that should NOT be moved. Anything in whisk/core should not move. Anything in system/baisc should not be moved. If these tests fails without catalog, then they should be made to work (by adding whatever assets they need to the tests/dat directory and creating the required assets as part of the test harness).

rabbah:
May I suggest that instead of whole sale moving packages and tests suites that you do this incrementally? Move exactly 1 package and its corresponding suite at a time. You''ll be thankful later to do it this way.

houshengbo:
@rabbah @csantanapr 
There are many test cases in system/basic related to the actions created by installCatalog. They will fail when openwhisk does not install the catalog. Since all the catalog is moved to openwhisk-catalog, the test cases deleted here in openwhisk, will be available in openwhisk-catalog. 

I will keep this PR alive, because openwhisk-catalog can refer to it for empty catalog. 
I can do the deletion of catalog incrementally.

markusthoemmes:
We should not lose most of the tests you pointed out. They should rather be refactored to not rely on the catalog anymore (for the ones it makes sense obviously)

rabbah:
@houshengbo this seems like it should be combined with you other pr #720.

houshengbo:
Let me close it first. I will come up a new script when openwhisk-catalog is reliable enough.

PR #78, opened 2016-03-14
Support Promises/A+ in nodeJsActions

Samze:
This change adds support for returning promises to node.js actions.

It can be used like this.
```
function main() {
   return new Promise(function (resolve, reject) {
       setTimeout(function() {
           resolve({done: true});
       }, 100);
   })
}
```

You can also reject promises, causing them to be wrapped in an error.
```
function main() {
   return new Promise(function (resolve, reject) {
       setTimeout(function() {
           reject({msg: "oops"});
       }, 100);
   })
}
```
Returns this `{ error: { msg: "oops" } }`

This fix uses the Bluebird promise library for now until a newer node versions is implemented. At which point we can remove the dependency and use ES6 promises. They both have the same resolve api so this will be a server change only and transparent to any users.

Note this commit makes use of the fact that calling Promise.resolve on a non-promise type (value/undefined) returns that value.

(See spec)
http://www.ecma-international.org/ecma-262/6.0/#sec-promise-objects

This fix resolves #56

PR #781, opened 2016-06-27
Support promises for Node.js

jasonpet:
Add support for promises.
JSHint cleanup as well.

### DIFF
--- a/core/nodejsActionBase/runner.js
+++ b/core/nodejsActionBase/runner.js
@@ -36,13 +36,13 @@ function NodeActionRunner(message, whisk) {
         return;
     }
 
-    var hrStart = undefined;
+    var hrStart;
     var callback = { };
 
     function closeAndReturn(result) {
         var hrEnd = process.hrtime();
 
-        if (typeof callback.closedConnection == 'undefined') {
+        if (typeof callback.closedConnection === 'undefined') {
             callback.closedConnection = true;
             callback.next({
                 result  : result,
@@ -71,13 +71,20 @@ function NodeActionRunner(message, whisk) {
         var result = this.userScriptMain(args);
 
         if (result !== this.whisk.async()) {
-            // This happens, e.g. if you just have "return;"
-            if (typeof result == 'undefined') {
-                result = {};
-            }
-            closeAndReturn(result);
+            var self = this;
+            Promise.resolve(result) //Non-promises/undefined instantly resolve.
+                .then(function(resolvedResult) {
+                    // This happens, e.g. if you just have "return;"
+                    if (typeof resolvedResult === 'undefined') {
+                        resolvedResult = {};
+                    }
+                    closeAndReturn(resolvedResult);
+                }, function (error) {
+                        self.whisk.error(error);
+                    }
+                );
         }
-    }
+    };
 }
 
 /**
@@ -107,7 +114,7 @@ function modWhisk(whisk, callback, closeAndReturn) {
             // call it directly.
             console.log('Warning: whisk.done() or whisk.error() called more than once.');
         }
-    }
+    };
     return whisk;
 }
 
@@ -118,7 +125,7 @@ function duration(start, end) {
 
 function getname(name) {
     name = typeof name === 'string' ? name.trim() : '';
-    name = name != '' ? name : 'Anonymous User Action';
+    name = name !== '' ? name : 'Anonymous User Action';
     return name;
 }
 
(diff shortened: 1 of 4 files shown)


jasonpet:
whisk.done() and whisk.async() will be removed for nodejs 6 as part of this pr since they are not necessary now that promises are supported.  However, with the change being made to make nodejs 6 the default in https://github.com/openwhisk/openwhisk/pull/738 all async js actions (such as pipe.js etc.....) installed by the install scripts will fail when run.  This is because the install scripts do not specify the kind so nodejs 6 will be used once it becomes the default.  As part of this pr I will update the install scripts to specify the kind to use nodejs for all of the async js actions.  I will will open a new pr to handle converting our existing async js actions to use promises.

sjfink:
I suggest we `deprecate` async and done rather than remove them.   That is, just leave them there for now and document that people should start using Promises instead.

jasonpet:
pg build 212

jasonpet:
@psuter - assigning to you since you own #78 as well.  I made some simple tweaks to the commit from #78 so it would apply to the current code base.

jasonpet:
@csantanapr - I will open a new issue to document that promises can now be used instead of async and done as Steve had suggested.  Where should we document this?  I see doc references for using async in references.md and actions.md.  Is this where the promises documentation should go?  Should we remove the references to async in the doc?

jasonpet:
I am going to add a simple sample action showing how an action can return a promise and add a test that uses the sample.

csantanapr:
actions.md add simple example with promises, 
in references.md mark async and done as deprecated and provide more in depth details about promises (i.e. how to use bluebird, q, etc..) @bjustin-ibm might have better ideas on how to structure this.

psuter:
Looks good to me, will merge when we have the example in docs+tests. Feel free to use additional commits as the blame shouldn't fall on the original author when the tests fail :)  As a point of curiosity, since the change is in the common container, does this mean we can use, e.g. bluebird packaged with webpack, and work with promises in the Node 0.9 container?

bjustin-ibm:
> (i.e. how to use bluebird, q, etc..)

It looks like I misunderstood the Promises documentation regarding the use of other libraries, and you do not get toll-free bridging for other promises libraries.

@psuter You are free to use other promise packages in the Node 0.12 runtime, but in the end, you will need to utilize `whisk.async()` and `whisk.done()` to inform the OW runtime what you are up to.  Similarly, in the Node 6 runtime, you will ultimately need to return an ES6 Promise, even if you are using another promise library in your code.

jasonpet:
Native promises are supported in the Node 0.12.14 container so there is no need to use bluebird in that container.

jasonpet:
I added a sample for Promises and a test for the sample as well.  I also found that the existing tests I had created to test Node.js 0.12.14 and Node.js 6 are both testing Node.js 6 now that the default Node has changed.  I updated the tests to specify the --kind for testing Node.js 0.12.14.

jasonpet:
Created issue https://github.com/openwhisk/openwhisk/issues/825 to handle the documentation for Promises.

core/nodejsActionBase/runner.js
```diff
@@ -71,13 +71,20 @@ function NodeActionRunner(message, whisk) {
...
+            var self = this;
+            Promise.resolve(result) //Non-promises/undefined instantly resolve.
+                .then(function(resolvedResult) {
+                    // This happens, e.g. if you just have "return;"
+                    if (typeof resolvedResult === 'undefined') {
+                        resolvedResult = {};
+                    }
+                    closeAndReturn(resolvedResult);
+                }, function (error) {
+                        self.whisk.error(error);
```
rabbah: is there a test case to cover the error case that reaches this path?

tests/src/actionContainers/NodeJsActionContainerTests.scala
```diff
@@ -357,4 +357,50 @@ class NodeJsActionContainerTests extends FlatSpec with Matchers {
         filtered(out).trim shouldBe empty
         filtered(err).trim shouldBe empty
     }
+
+    it should "support resolved promises" in {
+      val (out, err) = withNodeJsContainer { c =>
+      val code = """
```
rabbah: is line 364 deliberately aligned one character to the left?

core/nodejsActionBase/runner.js
jasonpet: yes, there is an action container test in this PR that rejects the Promise and will hit this path.

jasonpet:
pg build 251 ran with latest updates.  @psuter - this should be ready to merge now.

psuter:
Glad we now have this in!

sjfink:
@jasonpet in addition to the official documentation, how about a blog post on this?

Samze:
Great to see this made it in :)

PR #789, opened 2016-06-28
Allow null values in Consul Client on recurse fetching

ChrBi:
For some keys in WhiskPropertys, there is the value `null` saved in Consul.
On trying to `getRecurse` on the root (""), this leads to errors on converting the entry to `ConsulEntry`.
This fix removes this problem and returns an empty String if the value in consul was `null`.

### DIFF
--- a/common/scala/src/main/scala/whisk/common/ConsulClient.scala
+++ b/common/scala/src/main/scala/whisk/common/ConsulClient.scala
@@ -36,9 +36,12 @@ import spray.http.Uri.Path
 import spray.http.Uri.Query
 import spray.httpx.SprayJsonSupport.sprayJsonUnmarshaller
 import spray.json.DefaultJsonProtocol
+import java.util.NoSuchElementException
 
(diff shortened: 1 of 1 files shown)


common/scala/src/main/scala/whisk/common/ConsulClient.scala
```diff
@@ -74,7 +76,7 @@ class ConsulClient(host: String)(implicit val actorSystem: ActorSystem) {
      */
     def get(key: String): Future[String] = {
         listPipeline(Get(uriWithKey(key))) map {
-            _.head.decodedValue
+            _.head.decodedValue.get
```
rabbah: why is this get safe?

markusthoemmes: Because the Future should fail if it can't get the entry.

markusthoemmes:
rebased and merged in 7454e01cf035053813690a619fb0f9ce163e4718

PR #797, opened 2016-06-28
CLI to failure when no authorization key configured/provided (prior to sending HTTP request)

### LINKED ISSUE #778: Go CLI Should Not Send Request if Auth Key is Not Set
Currently the Go CLI sends request if an auth key is not set. This request is rejected by the Whisk backend, and an error is displayed to the user. This logic should be reworked so that the Go CLI does not send the request if the auth key is not set, and display an error to the user.

Tests must be updated to handle this change.

See comments: https://github.com/openwhisk/openwhisk/pull/614#discussion-diff-68622642

mdeuser:
Check that the auth key is set prior to sending the HTTP request
Fail the CLI if the require auth key is not set
Remove auth key from `sdk install` as it is not needed
Refactor the namespace code to use the common HTTP client code
Fix incorrect argument help text
Fixes #778

### DIFF
--- a/tests/src/common/Wsk.scala
+++ b/tests/src/common/Wsk.scala
@@ -78,12 +78,12 @@ case class WskProps(
 }
 
 class Wsk(override val usePythonCLI: Boolean = true) extends RunWskCmd {
-    implicit val action = new WskAction()
-    implicit val trigger = new WskTrigger()
-    implicit val rule = new WskRule()
-    implicit val activation = new WskActivation()
-    implicit val pkg = new WskPackage()
-    implicit val namespace = new WskNamespace()
+    implicit val action = new WskAction(usePythonCLI)
+    implicit val trigger = new WskTrigger(usePythonCLI)
+    implicit val rule = new WskRule(usePythonCLI)
+    implicit val activation = new WskActivation(usePythonCLI)
+    implicit val pkg = new WskPackage(usePythonCLI)
+    implicit val namespace = new WskNamespace(usePythonCLI)
 }
 
 trait FullyQualifiedNames {
@@ -202,13 +202,14 @@ trait HasActivation {
     }
 }
 
-class WskAction()
+class WskAction(override val usePythonCLI: Boolean = true)
     extends RunWskCmd
     with ListOrGetFromCollection
     with DeleteFromCollection
     with HasActivation {
 
     override protected val noun = "action"
+    override def baseCommand = Wsk.baseCommand(usePythonCLI)
 
     /**
      * Creates action. Parameters mirror those available in the CLI.
@@ -267,13 +268,14 @@ class WskAction()
     }
 }
 
-class WskTrigger()
+class WskTrigger(override val usePythonCLI: Boolean = true)
     extends RunWskCmd
     with ListOrGetFromCollection
     with DeleteFromCollection
     with HasActivation {
 
     override protected val noun = "trigger"
+    override def baseCommand = Wsk.baseCommand(usePythonCLI)
 
     /**
      * Creates trigger. Parameters mirror those available in the CLI.
@@ -317,13 +319,14 @@ class WskTrigger()
     }
 }
 
-class WskRule()
+class WskRule(override val usePythonCLI: Boolean = true)
     extends RunWskCmd
     with ListOrGetFromCollection
     with DeleteFromCollection
     with WaitFor {
 
     override protected val noun = "rule"
+    override def baseCommand = Wsk.baseCommand(usePythonCLI)
 
     /**
      * Creates rule. Parameters mirror those available in the CLI.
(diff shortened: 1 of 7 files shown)


mdeuser:
@dubeejw - Ready for review.  Thanks!

mdeuser:
@dubeejw - rebased against master branch; openwhisk CLI tests all pass

dubeejw:
@mdeuser, do all the tests pass when you flip the switch to use the Go CLI in WskBasicTests.scala?

dubeejw:
Also it looks like there are other changes in here that don't relate to the auth key issue?

mdeuser:
It looks like my comments in the commit all disappeared during the rebase.  The only change not related to the auth check is a change to the property command help text that I noticed was incorrect while using `wsk property` in testing wsk without any --auth setting.

mdeuser:
@dubeejw - Good point.  All 400+ tests for both python and Go CLI pass in my local env.

mdeuser:
@dubeejw - What exactly is the way to flip the test switch (Python -> Go CLI)?  I thought I changed it correctly, but now I'm suspicious.  Let's Hold off on merge until we have a chance to talk tomorrow.

mdeuser:
PG 228.
Added back code in WskBasicTests.scala to propogate the usePythonCLI attribute into the other command specific classes.  That fixed the issue I was seeing with the CLI switching from GO to Python mid execution of WskBasicTests.  Tests all now pass locally with Python and Go enabled in WskBasicTests.
@dubeejw - ready for re-review.

dubeejw:
@mdeuser, build failed due to trailing whitespace.

tests/src/common/Wsk.scala
```diff
@@ -641,11 +646,12 @@ class WskNamespace()
     }
 }
 
-class WskPackage()
+class WskPackage(override val usePythonCLI: Boolean = true) 
     extends RunWskCmd
     with ListOrGetFromCollection
     with DeleteFromCollection {
     override protected val noun = "package"
+    override def baseCommand = Wsk.baseCommand(usePythonCLI)
```
rabbah: I suggest also making the sealed trait's property abstract (that way there can't be confusion)

```
diff --git a/tests/src/common/Wsk.scala b/tests/src/common/Wsk.scala
index 08f3bba..a3a3b6c 100644
--- a/tests/src/common/Wsk.scala
+++ b/tests/src/common/Wsk.scala
@@ -787,7 +787,7 @@ object Wsk {
 
 sealed trait RunWskCmd {
 
-    val usePythonCLI: Boolean = true
+    val usePythonCLI: Boolean
 
```

tests/src/common/Wsk.scala
```diff
@@ -202,13 +202,14 @@ trait HasActivation {
     }
 }
 
-class WskAction()
+class WskAction(override val usePythonCLI: Boolean = true)
```
dubeejw: Trailing whitespace.

tests/src/common/Wsk.scala
```diff
@@ -267,13 +268,14 @@ class WskAction()
     }
 }
 
-class WskTrigger()
+class WskTrigger(override val usePythonCLI: Boolean = true)
```
dubeejw: Trailing whitespace.

tests/src/common/Wsk.scala
```diff
@@ -317,13 +319,14 @@ class WskTrigger()
     }
 }
 
-class WskRule()
+class WskRule(override val usePythonCLI: Boolean = true)
```
dubeejw: Trailing whitespace.

tests/src/common/Wsk.scala
```diff
@@ -432,12 +435,13 @@ class WskRule()
     }
 }
 
-class WskActivation()
+class WskActivation(override val usePythonCLI: Boolean = true)
```
dubeejw: Trailing whitespace.

tests/src/common/Wsk.scala
```diff
@@ -608,11 +612,12 @@ class WskActivation()
     private case class PartialResult(ids: Seq[String]) extends Throwable
 }
 
-class WskNamespace()
+class WskNamespace(override val usePythonCLI: Boolean = true)
```
dubeejw: Trailing whitespace.

tests/src/common/Wsk.scala
```diff
@@ -641,11 +646,12 @@ class WskNamespace()
     }
 }
 
-class WskPackage()
+class WskPackage(override val usePythonCLI: Boolean = true)
```
dubeejw: Trailing whitespace.

tests/src/common/Wsk.scala
rabbah: @csantanapr we should document the git commit hook with runs scan code for devs to adopt.

mdeuser:
Travis is now happy.  Since there was a rebase, a new PG 264 is being run

PR #807, opened 2016-06-29
[Review] Go CLI Does Not Create Action with Empty File

dubeejw:
- Ensure code property always exists even if it is an empty string
- Add test case to ensure actions can be create with empty files
- Add test case to invoke an empty action

### DIFF
--- a/tests/src/system/basic/CLIActionTests.java
+++ b/tests/src/system/basic/CLIActionTests.java
@@ -329,6 +329,35 @@ public class CLIActionTests {
     }
 
     @Test(timeout=120*1000)
+    public void createActionWithEmptyFile() throws Exception {
+        String action = "createActionWithEmptyFile";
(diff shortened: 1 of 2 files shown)


dubeejw:
@mdeuser, ready for review.

dubeejw:
IPG 227.

dubeejw:
Please review.

mdeuser:
@rabbah - What is the expected behavior when an action with no code is invoked?  Just wondering if that should also be part of this test.

rabbah:
there should be a test for invoking a malformed action (an empty action is considered malformed).

dubeejw:
Test added for invoking an empty action.

dubeejw:
IPG 237.

rabbah:
Rebased and merged as https://github.com/openwhisk/openwhisk/commit/51817283c204c055456646c8bfb536ab78d2fdf1.

PR #810, opened 2016-06-29
[WIP] Do Not Take Arguments for --shared Flag (Go CLI)

dubeejw:
- Remove logic for --shared flag to take arguments
- If --shared is present in a command, the entity will be public.
  Otherwise, the entity will not be publically shared

### DIFF
--- a/tests/src/common/WskCli.java
+++ b/tests/src/common/WskCli.java
@@ -367,7 +367,7 @@ public class WskCli {
         }
 
         if (shared) {
-            cmd = Util.concat(cmd, new String[] { "--shared", "yes"});
+            cmd = Util.concat(cmd, new String[] { "--shared" });
         }
 
         RunResult result = cli(expectedCode, cmd);
--- a/tools/go-cli/go-whisk-cli/commands/action.go
+++ b/tools/go-cli/go-whisk-cli/commands/action.go
@@ -85,7 +85,7 @@ var actionCreateCmd = &cobra.Command{
     SilenceErrors:  true,
     PreRunE: setupClientConfig,
     RunE: func(cmd *cobra.Command, args []string) error {
-        action, sharedSet, err := parseAction(cmd, args)
+        action, err := parseAction(cmd, args)
         if err != nil {
             whisk.Debug(whisk.DbgError, "parseAction(%s, %s) error: %s\n", cmd, args, err)
             errMsg := fmt.Sprintf("Unable to parse action command arguments: %s", err)
@@ -94,9 +94,9 @@ var actionCreateCmd = &cobra.Command{
             return whiskErr
         }
 
-        _, _, err = client.Actions.Insert(action, sharedSet, false)
+        _, _, err = client.Actions.Insert(action, false)
         if err != nil {
-            whisk.Debug(whisk.DbgError, "client.Actions.Insert(%#v, %t, false) error: %s\n", action, sharedSet, err)
+            whisk.Debug(whisk.DbgError, "client.Actions.Insert(%#v, false) error: %s\n", action, err)
             errMsg := fmt.Sprintf("Unable to create action: %s", err)
             whiskErr := whisk.MakeWskErrorFromWskError(errors.New(errMsg), err, whisk.EXITCODE_ERR_NETWORK,
                 whisk.DISPLAY_MSG, whisk.NO_DISPLAY_USAGE)
@@ -147,7 +147,7 @@ var actionUpdateCmd = &cobra.Command{
     SilenceErrors:  true,
     PreRunE: setupClientConfig,
     RunE: func(cmd *cobra.Command, args []string) error {
-        action, sharedSet, err := parseAction(cmd, args)
+        action, err := parseAction(cmd, args)
         if err != nil {
             whisk.Debug(whisk.DbgError, "parseAction(%s, %s) error: %s\n", cmd, args, err)
             errMsg := fmt.Sprintf("Unable to parse action command arguments: %s", err)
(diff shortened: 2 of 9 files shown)


dubeejw:
@mdeuser, please review.

dubeejw:
IPG 222

dubeejw:
@mdeuser, please review.

tools/go-cli/go-whisk-cli/commands/action.go
```diff
@@ -720,7 +709,7 @@ func init() {
     actionCreateCmd.Flags().BoolVar(&flags.action.copy, "copy", false, "treat artifact as the name of an existing action")
     actionCreateCmd.Flags().BoolVar(&flags.action.sequence, "sequence", false, "treat artifact as comma separated sequence of actions to invoke")
     actionCreateCmd.Flags().StringVar(&flags.action.kind, "kind", "", "the kind of the action runtime (example: swift:3)")
-    actionCreateCmd.Flags().StringVar(&flags.action.shared, "shared", "", "shared action (default: private)")
+    actionCreateCmd.Flags().BoolVar(&flags.common.shared, "shared", false, "make an action publically accessible")
```
rabbah: publically -> publicly

tools/go-cli/go-whisk-cli/commands/action.go
```diff
@@ -731,7 +720,7 @@ func init() {
     actionUpdateCmd.Flags().BoolVar(&flags.action.docker, "docker", false, "treat artifact as docker image path on dockerhub")
     actionUpdateCmd.Flags().BoolVar(&flags.action.copy, "copy", false, "treat artifact as the name of an existing action")
     actionUpdateCmd.Flags().BoolVar(&flags.action.sequence, "sequence", false, "treat artifact as comma separated sequence of actions to invoke")
-    actionUpdateCmd.Flags().StringVar(&flags.action.shared, "shared", "", "shared action (default: private)")
+    actionUpdateCmd.Flags().BoolVar(&flags.common.shared, "shared", false, "make an action publically accessible")
```
rabbah: publically -> publicly

tests/src/common/WskCli.java
```diff
@@ -367,7 +367,7 @@ public String createAction(int expectedCode, String name, String artifact, Strin
         }
 
         if (shared) {
-            cmd = Util.concat(cmd, new String[] { "--shared", "yes"});
+            cmd = Util.concat(cmd, new String[] { "--shared" });
```
rabbah: missed removing the "yes"/"no" from Wsk.scala.

tools/go-cli/go-whisk-cli/commands/package.go
```diff
@@ -185,16 +184,6 @@ var packageCreateCmd = &cobra.Command{
         }
         client.Namespace = qName.namespace
 
-        if (flags.common.shared == "yes") {
```
rabbah: I had understood the plan as making yes/no optional. I believe with this change, once you make the package public (or other types), you cannot make it private. Is that a correct statement?

Note that the python CLI allows yes/no optionally but requires --shared no to make the package private.
To wit, we need a test that makes a public package private via CLI.

dubeejw: The package we are using to parse arguments does not allow flags to to arguments, and not take arguments. When --shared is not present in a command, the entity will be set to not be shared.

dubeejw: shared

rabbah:
So I create a package and make it public. Then I update the package and don't provide --shared, it becomes private?

dubeejw:
Exactly. I can close this PR, and leave it the way it is to accept arguments.

rabbah:
I'm inclined to say yes - leave as is. Let others weigh in.

tools/go-cli/go-whisk-cli/commands/package.go
mdeuser: @rabbah - Do you think going forward this shared argument will eventually take additional states besides today's binary public|private values?  IMHO, I think we'll see additional shared scopes, including custom ones, in the future.  If so, then perhaps we should move this argument away from being a true|false state and more of a select-a-shared-scoped kind of argument.

tools/go-cli/go-whisk-cli/commands/action.go
mdeuser: maybe also add what the default value is.. something like `(default: action is private)`.. maybe shorter `(default: private)`

tools/go-cli/go-whisk-cli/commands/package.go
```diff
@@ -598,19 +550,19 @@ func init() {
     packageCreateCmd.Flags().StringSliceVarP(&flags.common.annotation, "annotation", "a", []string{}, "annotations")
     packageCreateCmd.Flags().StringSliceVarP(&flags.common.param, "param", "p", []string{}, "default parameters")
     packageCreateCmd.Flags().StringVarP(&flags.xPackage.serviceGUID, "service_guid", "s", "", "a unique identifier of the service")
-    packageCreateCmd.Flags().StringVar(&flags.common.shared, "shared", "" , "shared action (default: private)")
+    packageCreateCmd.Flags().BoolVar(&flags.common.shared, "shared", false, "make a package publically accessible")
```
mdeuser: "make a package.." -> "make package...(default: private)" (remove the 'a' indefinite article).  Same comment for other `shared` help text

mdeuser:
Should the whisk/trigger.go have the following update similar to the action, package and rule modules?
In the `Trigger` struct used to encode the JSON sent to the server:
```
Publish bool `json:"publish,omitempty"`
```
change to

```
Publish bool `json:"publish"`  // remove the omitempty
```

tools/go-cli/go-whisk-cli/commands/package.go
rabbah: fine grained entitlement is a feature we want to surface, eventually.

Issue #816, opened 2016-06-30
cannot update an action with just parameters 


rabbah:
## Steps to reproduce the issue:

```
1.  touch foo.js
2.  wsk action create foo foo.js -p a A
3.  wsk action update foo -p b B
4.  wsk action get foo parameters
```

## Provide the expected results and outputs:

```
ok: got action foo, projecting parameters
[
    {
        "key": "b",
        "value": "B"
    }
]
```

## Provide the actual results and outputs:

```
ok: got action foo, projecting parameters
[]
```

The test coverage was missing a case of parameter update without an `exec` defined. Also we are missing tests for parameter update from CLI, and tests missing for projecting properties (these are not supported by go CLI).

@mdeuser @dubeejw FYI.

rabbah:
Closed by #818.

PR #818, opened 2016-07-01
[Ready] Fix issue #816. 

rabbah:
When updating an action that is already a sequence or when updating an action with a sequence, set parameters explicitly. If new action is not a sequence or if the type of the existing action is not a sequence, preserve new parameters or use pre-existing parameters.

This logic will be removed when sequences are fully converted to a first class entity type. In the previous incremental commit toward this goal, I didn't quite get the logic correct on the update operation. I added more tests to cover the different use cases, including a test we did not have for using the CLI to do an action update defining only new parameters.

@perryibm @markusthoemmes @ChrBi for your reviewing pleasure.
FYI @dubeejw @mdeuser note new test (passed go cli).

PPG 736 approved.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Actions.scala
@@ -391,14 +391,24 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
         // exec is given.
         val parameters = content.exec map {
             case seq: SequenceExec => Parameters("_actions", JsArray(seq.components map { JsString(_) }))
-            case _                 => content.parameters getOrElse Parameters()
+            case _ => content.parameters getOrElse {
+                action.exec match {
+                    case seq: SequenceExec => Parameters()
+                    case _                 => action.parameters
+                }
+            }
+        } getOrElse {
+            action.exec match {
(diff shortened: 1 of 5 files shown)


core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -391,14 +391,24 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+                action.exec match {
+                    case seq: SequenceExec => Parameters()
+                    case _                 => action.parameters
+                }
+            }
+        } getOrElse {
+            action.exec match {
+                case seq: SequenceExec => action.parameters // discard content.parameters
+                case _                 => content.parameters getOrElse action.parameters
+            }
```
ChrBi: I'dd suggest to add a few comments to see with one few, in which edge case which Parameters are set and why.

rabbah:
Pushed comment and package name fix @ChrBi. Squash before merge.

rabbah:
Rebased and merged as https://github.com/openwhisk/openwhisk/commit/d95e0698d3a20a98bbf385325cdee41e28cd1da5.

PR #765, opened 2016-06-24
[Review] `akka-http` clients for CouchDB and Consul

psuter:
@rabbah @markusthoemmes for your reading pleasure

PR #824, opened 2016-07-01
[Ready] Fewer `ActorSystem`s

psuter:
This PR is built on top of #765 because it assumes Akka 2.4.7 (and therefore shouldn't be merged before it). Only the last two commits are specific to this PR.

Makes it more explicit where `ActorSystem`s are created. Eliminates the creation of one spurious system in the invoker. Generalizes the practice of retrieving `ExecutionContext`s from `ActorSystem`s when available.

Taking the actor system out of `BasicHttpService` is also a required step towards having cohabiting Spray and Akka HTTP API endpoints.

### DIFF
--- a/core/dispatcher/src/main/scala/whisk/core/dispatcher/Dispatcher.scala
+++ b/core/dispatcher/src/main/scala/whisk/core/dispatcher/Dispatcher.scala
@@ -19,6 +19,7 @@ package whisk.core.dispatcher
 import scala.collection.concurrent.TrieMap
 import scala.collection.parallel.mutable.ParTrieMap
 import scala.concurrent.Future
+import scala.concurrent.ExecutionContext
 import scala.util.Failure
 import scala.util.Success
 import scala.util.matching.Regex.Match
@@ -87,6 +88,8 @@ trait Registrar {
 trait MessageDispatcher extends Registrar with Logging {
     consumer: MessageConsumer =>
 
+    val executionContext: ExecutionContext
+
     /**
      * Starts a listener to consume message from connector (e.g., kafka).
      * The listener consumes messages from the bus using a streaming consumer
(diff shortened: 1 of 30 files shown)


core/dispatcher/src/main/scala/whisk/core/dispatcher/Dispatcher.scala
```diff
@@ -191,8 +195,6 @@ object Dispatcher extends Logging {
     def requiredProperties =
         Map(servicePort -> 8080.toString()) ++ kafkaHost
 
-    val executionContext = ExecutionContextFactory.makeCachedThreadPoolExecutionContext()
```
psuter: Did we have any evidence that this is superior to the default? @rabbah @perryibm

rabbah: This pool is currently semantically important. We can't remove it until the work on back pressure is merged.

psuter: I assume because in theory you may need an unbounded number of concurrent executors? I can amend the PR to keep a cached pool and make it the underlying context of the `ActorSystem` (just like we have now), but still have a single context (like this PR is introducing).

psuter: Now constructing the `ActorSystem` with the aforementioned execution context in `InvokerService.main`.

tests/src/services/KafkaConnectorTests.scala
```diff
@@ -40,7 +40,7 @@ import scala.language.postfixOps
 @RunWith(classOf[JUnitRunner])
 class KafkaConnectorTests extends FlatSpec with Matchers with BeforeAndAfter {
     implicit val transid = TransactionId.testing
-    implicit val ec = ExecutionContextFactory.makeCustomThreadPoolExecutionContext(poolSize = 4)
+    implicit val ec = ExecutionContextFactory.makeCachedThreadPoolExecutionContext()
```
psuter: I may have missed the justification for having a custom context here? Now the test uses the same type of `ExecutionContext` as the production code.

psuter:
PPG 760

psuter:
Includes all changes from #765.

psuter:
PPG 766 for the freshly rebased version.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/compare/5295f25514e4...1ebc8b261351.

PR #836, opened 2016-07-02
Remove ant files

rabbah:
These are no longer needed and docs no longer reference `ant`.


markusthoemmes:
joyfully rebased and merged in 7c0f6d17e1078a45d376fef33e4ed3becd087913

PR #844, opened 2016-07-04
Adding two mirrors of tinycorelinux if the default repo is down

fxulusoy:


rabbah:
should we add these mirrors automatically as part of the tweaking script?

csantanapr:
@fxulusoy I agree with @rabbah 
I think it will be better to address the problem in the script tweak-dockermachine.sh
by adding the mirror, or just ping which repo is up and use that one.

csantanapr:
bump @fxulusoy

fxulusoy:
@csantanapr I will add the mirrors  in the script tweak-dockermachine.sh

csantanapr:
rebase and merged 83227e2ee617965e1208f67a12bed2de96f37ba0

PR #853, opened 2016-07-06
Limit size of the logs, that are written in user containers

markusthoemmes:
Let the activation fail, if the limit is exceededed.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/Limits.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Limits.scala
@@ -40,14 +40,16 @@ protected[entity] abstract class Limits {
 
 /**
  * Limits on a specific action. Includes the following properties
- * { timeout: maximum duration in msecs an action is allowed to consume in [100 msecs, 1 minute]
- *   memory: maximum memory in megabytes an action is allowed to consume in [128, 8192]
+ * { timeout: maximum duration in msecs an action is allowed to consume in [100 msecs, 5 minutes]
+ *   memory: maximum memory in megabytes an action is allowed to consume in [128, 512]
  * }
  *
  * @param timeout the duration in milliseconds, assured to be non-null because it is a value
  * @param memory the memory limit in megabytes, assured to be non-null because it is a value
+ * @param logs the limit for logs written by the container being written to the database.
+ *     <code>Option</code> for database schema migration
  */
-protected[core] case class ActionLimits protected[core] (timeout: TimeLimit, memory: MemoryLimit) extends Limits {
+protected[core] case class ActionLimits protected[core] (timeout: TimeLimit, memory: MemoryLimit, logs: Option[LogLimit] = Some(LogLimit())) extends Limits {
     override protected[entity] def toJson = ActionLimits.serdes.write(this)
 }
 
@@ -65,28 +67,15 @@ protected[core] object ActionLimits
     /** Creates a ActionLimits instance with default duration and memory limit. */
     protected[core] def apply(): ActionLimits = ActionLimits(TimeLimit(), MemoryLimit())
 
-    /**
-     * Creates a ActionLimits instance with given duration and default memory limit.
-     *
-     * @param duration the duration for the action, must not be null and must be within allowed limits
-     * @return corresponding ActionLimits
-     * @throws IllegalArgumentException if expected properties are missing or not valid
-     */
-    @throws[IllegalArgumentException]
-    protected[core] def apply(duration: FiniteDuration): ActionLimits = ActionLimits(TimeLimit(duration), MemoryLimit())
+    override protected[core] implicit val serdes = new RootJsonFormat[ActionLimits] {
+        val helper = jsonFormat3(ActionLimits.apply)
 
-    /**
-     * Creates an ActionsLimits instance from a duration and memory limit.
-     *
-     * @param duration the duration for the action, must not be null and must be within allowed limits
-     * @param megabytes the memory limit in megabytes, must be within allowed limits
-     * @return corresponding ActionLimits
-     * @throws IllegalArgumentException if expected properties are missing or not valid
-     */
-    @throws[IllegalArgumentException]
-    protected[entity] def !(duration: FiniteDuration, megabytes: Int): ActionLimits = ActionLimits(TimeLimit(duration), MemoryLimit(megabytes))
-
-    override protected[core] implicit val serdes = jsonFormat2(ActionLimits.apply)
+        def read(value: JsValue) = {
+            val inter = helper.read(value)
+            inter.copy(logs = Some(inter.logs.getOrElse(LogLimit())))
+        }
+        def write(a: ActionLimits) = helper.write(a.copy(logs = Some(a.logs.getOrElse(LogLimit()))))
+    }
 }
 
 protected[core] object TriggerLimits
--- a/common/scala/src/main/scala/whisk/core/entity/Limits.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Limits.scala
@@ -49,7 +49,7 @@ protected[entity] abstract class Limits {
  * @param logs the limit for logs written by the container being written to the database.
  *     <code>Option</code> for database schema migration
  */
-protected[core] case class ActionLimits protected[core] (timeout: TimeLimit, memory: MemoryLimit, logs: Option[LogLimit] = Some(LogLimit())) extends Limits {
+protected[core] case class ActionLimits protected[core] (timeout: TimeLimit, memory: MemoryLimit, logs: LogLimit = LogLimit()) extends Limits {
     override protected[entity] def toJson = ActionLimits.serdes.write(this)
 }
 
@@ -71,10 +71,13 @@ protected[core] object ActionLimits
         val helper = jsonFormat3(ActionLimits.apply)
 
         def read(value: JsValue) = {
-            val inter = helper.read(value)
-            inter.copy(logs = Some(inter.logs.getOrElse(LogLimit())))
+            val obj = Try { value.asJsObject.convertTo[Map[String, JsValue]] }.getOrElse(deserializationError("no valid json object passed"))
+            val time = TimeLimit.serdes.read(obj.get("timeout").getOrElse(deserializationError("'timeout' is missing")))
+            val memory = MemoryLimit.serdes.read(obj.get("memory").getOrElse(deserializationError("'memory' is missing")))
+            val logs = obj.get("logs").map { LogLimit.serdes.read(_) }.getOrElse(LogLimit())
+            ActionLimits(time, memory, logs)
         }
-        def write(a: ActionLimits) = helper.write(a.copy(logs = Some(a.logs.getOrElse(LogLimit()))))
+        def write(a: ActionLimits) = helper.write(a)
     }
 }
 
--- a/common/scala/src/main/scala/whisk/core/entity/LogLimit.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/LogLimit.scala
@@ -22,6 +22,8 @@ import spray.json.JsNumber
 import spray.json.JsValue
 import spray.json.RootJsonFormat
 import spray.json.deserializationError
+import scala.util.Success
+import scala.util.Failure
 
 /**
  * LogLimit encapsulates allowed amount of logs written by an action.
@@ -29,6 +31,9 @@ import spray.json.deserializationError
  * It is a value type (hence == is .equals, immutable and cannot be assigned null).
  * The constructor is private so that argument requirements are checked and normalized
  * before creating a new instance.
+ *
+ * FIXME: Int because of JSON deserializer vs. <code>ByteSize</code> and compatibility
+ * with <code>MemoryLimit</code>
  *
  * @param megabytes the memory limit in megabytes for the action
  */
@@ -43,15 +48,15 @@ protected[core] object LogLimit extends ArgNormalizer[LogLimit] {
     protected[core] def apply(): LogLimit = new LogLimit(STD_LOGSIZE)
 
     /**
-     * Creates LogLimit for limit. Only the default limit is allowed currently
+     * Creates LogLimit for limit. Only the default limit is allowed currently.
      *
      * @param megabytes the limit in megabytes, must be within permissible range
      * @return LogLimit with limit set
      * @throws IllegalArgumentException if limit does not conform to requirements
      */
     @throws[IllegalArgumentException]
-    protected[core] def apply(megabytes: Int): LogLimit = {
-        require(megabytes == STD_LOGSIZE, "only standard log limit allowed")
+    private def apply(megabytes: Int): LogLimit = {
+        require(megabytes == STD_LOGSIZE, s"only standard log limit of '$STD_LOGSIZE' (megabytes) allowed")
         new LogLimit(megabytes);
     }
 
@@ -62,6 +67,10 @@ protected[core] object LogLimit extends ArgNormalizer[LogLimit] {
             val JsNumber(mb) = value
             require(mb.isWhole(), "log limit must be whole number")
             LogLimit(mb.intValue)
-        } getOrElse deserializationError("log limit malformed")
+        } match {
+            case Success(limit)                       => limit
+            case Failure(e: IllegalArgumentException) => deserializationError(e.getMessage, e)
+            case Failure(e: Throwable)                => deserializationError("log limit malformed", e)
+        }
     }
 }
--- a/common/scala/src/main/scala/whisk/core/entity/LogLimit.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/LogLimit.scala
@@ -24,8 +24,7 @@ import spray.json.RootJsonFormat
 import spray.json.deserializationError
 
 /**
- * MemoyLimit encapsulates allowed memory for an action. The limit must be within a
- * permissible range (currently [128MB, 512MB]).
+ * LogLimit encapsulates allowed amount of logs written by an action.
  *
  * It is a value type (hence == is .equals, immutable and cannot be assigned null).
  * The constructor is private so that argument requirements are checked and normalized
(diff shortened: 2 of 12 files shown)


common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -83,6 +84,8 @@ class WhiskConfig(
     val mainDockerEndpoint = this(WhiskConfig.mainDockerEndpoint)
     val routerDockerEndpoint = this(WhiskConfig.routerDockerEndpoint)
     val elkDockerEndpoint = this(WhiskConfig.elkDockerEndpoint)
+
+    lazy val logSizeLimit = ObjectSize.fromString(this(WhiskConfig.logSizeLimit))
```
rabbah: why lazy?

core/dispatcher/src/main/scala/whisk/core/container/Container.scala
```diff
@@ -34,6 +35,7 @@ class Container(
     val image: String,
     network: String,
     pull: Boolean = false,
+    logLimits: ObjectSize,
```
rabbah: why not make logLimits part of ActionLimits?

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -461,6 +451,8 @@ class Invoker(
         val duration = Duration.fromNanos(java.time.Duration.between(interval._1, interval._2).toNanos)
         val activationResponse = if (duration >= timeout) {
             ActivationResponse.applicationError(s"action exceeded its time limits of ${timeout.toMillis} milliseconds")
+        } else if (log == JsArray("log limit exceeded".toJson)) {
```
rabbah: why not truncate the logs instead of failing the action? considered that as an alternate option?

markusthoemmes: Yep, both ways are certainly possible. It's a matter of deciding which way we want to go. I agree that the result is independent. One could also think of it as a measure of penalty against a misbehaving action.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
markusthoemmes: If the value is not set (and not even required by the class instantiating WhiskConfig) an expection is thrown by the `fromString` method.

perryibm:
@rabbah brings up an important point but I think there are actually two issues.  I agree on (2) and in-between on (1).

1.  Should the action be failed if it exceeds the quota?  I am ambivalent with it failing so the user is alerted in some way.  Alternatives are to consider error codes that signify something unusual has happened but not a failure.  We may need this anyway to alert the user that the action required a retry.

2. Should the log be truncated if it exceeds the quota?  I would argue no for two reasons.  First, they are allowed to use space up to the quota so why would we impose a severe penalty for going one byte and not just allow them to have the log up to the quota?  Second, turning the entire log into the string `log limit exceeded` would leave no clues in the log about why the log became so voluminous.

rabbah:
we already have a "status code" - we can add to it to reflect a warning, log size exceeded status, for example.

markusthoemmes:
@perryibm very valid points, thanks for the feedback.

The error model is in fact a bit ambiguous here. I'll take the failure out again and implement the "cut after limit" idea. I'll add the line "Logs have exceeded their limit" as the last line in the log.

markusthoemmes:
Took out the log-rotation mechanisms. Only Limiting the size written to cloudant now. Could you re-review @rabbah @perryibm?

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -311,27 +316,49 @@ class Invoker(
...
+     * Takes the first bytes of a list of strings. If the number of bytes would cut an entry
+     * it is omitted entirely.
+     *
+     * @param strings The list of strings
+     * @param bytes number of
+     */
+    def truncateLogs(strings: Vector[String], sizeLimit: ObjectSize, written: ObjectSize = 0 bytes): Vector[String] = {
+        val stringStream = strings.toStream
+        val cummulativeSizes = stringStream.scanLeft(0.bytes) { (acc, current) => acc + current.sizeInBytes }.tail
+        val truncatedLogs = stringStream.zip(cummulativeSizes).takeWhile(_._2 < sizeLimit).map(_._1).toVector
```
perryibm: It's functional but takes 5 passes on something that's potentially "large". What about one scanLeft and make the accumulator the vector and the cumulative size.

markusthoemmes: Consulted @psuter on that one because of the specific issue you mentioned. It uses Streams so there are only as many elements computed (lazily) as needed to find the maximum size of logs in this case.

common/scala/src/main/scala/whisk/core/entity/LogLimit.scala
```diff
@@ -0,0 +1,67 @@
...
+ * @param megabytes the memory limit in megabytes for the action
+ */
+protected[entity] class LogLimit private (val megabytes: Int) extends AnyVal {
+    protected[core] def apply() = megabytes
+}
+
+protected[core] object LogLimit extends ArgNormalizer[LogLimit] {
+    protected[core] val STD_LOGSIZE = 10 // MB
+
+    /** Gets TimeLimit with default duration */
```
rabbah: bad comment - copy/paste fail :)

common/scala/src/main/scala/whisk/core/entity/LogLimit.scala
```diff
@@ -0,0 +1,67 @@
...
+    protected[core] def apply(): LogLimit = new LogLimit(STD_LOGSIZE)
+
+    /**
+     * Creates LogLimit for limit, iff limit is within permissible range.
+     *
```
rabbah: comment says permissible range but require statement below doesn't allow for a range.

common/scala/src/main/scala/whisk/core/entity/MemoryLimit.scala
```diff
@@ -27,7 +25,7 @@ import spray.json.deserializationError
 
 /**
  * MemoyLimit encapsulates allowed memory for an action. The limit must be within a
- * permissible range (currently [128MB, 8192MB]).
+ * permissible range (currently [128MB, 512MB]).
```
rabbah: makes consistent with: https://github.com/openwhisk/openwhisk/blob/master/docs/reference.md#per-action-memory-mb-default-256mb

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -332,27 +337,49 @@ class Invoker(
      *
      * Note: Updates the container's log cursor to indicate consumption of log.
      */
-    private def getContainerLogs(con: WhiskContainer, tries: Int = LogRetryCount)(implicit transid: TransactionId): JsArray = {
+    private def getContainerLogs(con: WhiskContainer, limits: ActionLimits, tries: Int = LogRetryCount)(implicit transid: TransactionId): JsArray = {
         val size = con.getLogSize(runningInContainer)
+        val maxSize = limits.logs.map { _.megabytes }.getOrElse(Int.MaxValue)
```
rabbah: really? `getOrElse(default log limit makes more sense to me)`. this grandfathers actions to unbounded log length, is this intended?

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -332,27 +337,49 @@ class Invoker(
...
-            var isSwift = con.image == swiftImageName || con.image == swift3ImageName
-            val (complete, logArray) = processJsonDriverLogContents(rawLog, isNodeJs || isSwift)
+            val isSwift = con.image == swiftImageName || con.image == swift3ImageName
+            val (complete, logArray) = processJsonDriverLogContents(rawLog, isNodeJs || isSwift, maxSize megabytes)
+
             if (tries > 0 && !complete) {
                 info(this, s"log cursor advanced but missing sentinel, trying $tries more times")
                 Thread.sleep(LogRetry)
-                getContainerLogs(con, tries - 1)
+                getContainerLogs(con, limits, tries - 1)
```
rabbah: should this pass LogLimit instead - since the other limits are not relevant to getContainerLogs.

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -332,27 +337,49 @@ class Invoker(
...
+    }
+
+    /**
+     * Takes the first bytes of a list of strings. If the number of bytes would cut an entry
+     * it is omitted entirely.
+     *
+     * @param strings The list of strings
+     * @param sizeLimit The limit of strings to get
+     */
+    def truncateLogs(strings: Vector[String], sizeLimit: ObjectSize): Vector[String] = {
```
rabbah: `sizeLimit` -> `LogLimit` then I think `ObjectSize` while elegant becomes unnecessary... thoughts?

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -396,7 +424,7 @@ class Invoker(
 
     private def formatLog(time_stream_msg: (String, String, String)) = {
```
rabbah: move this method inside `processJsonDriverLogContents`?

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -332,27 +337,49 @@ class Invoker(
...
         val advanced = size != con.lastLogSize
         if (tries <= 0 || advanced) {
             val rawLogBytes = con.getDockerLogContent(con.lastLogSize, size, runningInContainer)
             val rawLog = new String(rawLogBytes, "UTF-8")
+
             val isNodeJs = con.image == nodejsImageName || con.image == nodejs6ImageName
-            var isSwift = con.image == swiftImageName || con.image == swift3ImageName
-            val (complete, logArray) = processJsonDriverLogContents(rawLog, isNodeJs || isSwift)
+            val isSwift = con.image == swiftImageName || con.image == swift3ImageName
+            val (complete, logArray) = processJsonDriverLogContents(rawLog, isNodeJs || isSwift, maxSize megabytes)
```
rabbah: `maxSize` is only used here - just pass the limit to `processJsonDriverLogContents`.

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -332,27 +337,49 @@ class Invoker(
      *
      * Note: Updates the container's log cursor to indicate consumption of log.
      */
-    private def getContainerLogs(con: WhiskContainer, tries: Int = LogRetryCount)(implicit transid: TransactionId): JsArray = {
+    private def getContainerLogs(con: WhiskContainer, limits: ActionLimits, tries: Int = LogRetryCount)(implicit transid: TransactionId): JsArray = {
         val size = con.getLogSize(runningInContainer)
+        val maxSize = limits.logs.map { _.megabytes }.getOrElse(Int.MaxValue)
         val advanced = size != con.lastLogSize
         if (tries <= 0 || advanced) {
```
rabbah: and not over log limit?

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -332,27 +337,49 @@ class Invoker(
...
             val rawLogBytes = con.getDockerLogContent(con.lastLogSize, size, runningInContainer)
             val rawLog = new String(rawLogBytes, "UTF-8")
+
             val isNodeJs = con.image == nodejsImageName || con.image == nodejs6ImageName
-            var isSwift = con.image == swiftImageName || con.image == swift3ImageName
-            val (complete, logArray) = processJsonDriverLogContents(rawLog, isNodeJs || isSwift)
+            val isSwift = con.image == swiftImageName || con.image == swift3ImageName
+            val (complete, logArray) = processJsonDriverLogContents(rawLog, isNodeJs || isSwift, maxSize megabytes)
+
             if (tries > 0 && !complete) {
```
rabbah: and not over size limit?

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -380,13 +407,14 @@ class Invoker(
                             None
                     }
             }
+
             if (requireSentinel) {
                 val (sentinels, regulars) = records.partition(_._3 == LOG_ACTIVATION_SENTINEL)
                 val hasOut = sentinels.exists(_._2 == "stdout")
                 val hasErr = sentinels.exists(_._2 == "stderr")
-                (hasOut && hasErr, JsArray(regulars map formatLog))
+                (hasOut && hasErr, JsArray(truncateLogs(regulars.map(formatLog), maxSize).map(_.toJson)))
```
rabbah: the first part of the tuple should also compute if max size is reached.

tests/src/system/basic/ActionTests.scala
```diff
@@ -76,6 +83,41 @@ class ActionTests
             withActivation(wsk.activation, run) {
                 _.fields("response").asJsObject.fields("result").toString should include(
                     """[OK] message terminated successfully""")
+
+            }
+    }
+
+    it should "fail with Application error, if log size exceeds its limit" in withAssetCleaner(wskprops) {
```
rabbah: update test name.

tests/src/whisk/common/SchedulerTests.scala
```diff
@@ -94,7 +93,7 @@ class SchedulerTests extends FlatSpec with Matchers {
         scheduled ! PoisonPill
 
         val differences = calculateDifferences(calls)
-        all(differences) should be >= timeBetweenCalls
+        assert(differences.forall(_ >= timeBetweenCalls), s"$differences have not been greater than $timeBetweenCalls")
```
rabbah: boo (i know).

tests/src/whisk/core/entity/test/SchemaTests.scala
```diff
@@ -314,8 +315,8 @@ class SchemaTests extends FlatSpec with BeforeAndAfter {
 
     it should "properly deserialize JSON" in {
         val json = Seq[JsValue](
-            JsObject("timeout" -> TimeLimit.STD_DURATION.toMillis.toInt.toJson, "memory" -> MemoryLimit.STD_MEMORY.toJson),
-            JsObject("timeout" -> TimeLimit.STD_DURATION.toMillis.toInt.toJson, "memory" -> MemoryLimit.STD_MEMORY.toJson, "foo" -> "bar".toJson))
+            JsObject("timeout" -> TimeLimit.STD_DURATION.toMillis.toInt.toJson, "memory" -> MemoryLimit.STD_MEMORY.toJson, "logs" -> LogLimit.STD_LOGSIZE.toJson),
+            JsObject("timeout" -> TimeLimit.STD_DURATION.toMillis.toInt.toJson, "memory" -> MemoryLimit.STD_MEMORY.toJson, "logs" -> LogLimit.STD_LOGSIZE.toJson, "foo" -> "bar".toJson))
```
rabbah: a test where it will serdes correctly even when `logs` is missing?

common/scala/src/main/scala/whisk/core/entity/Limits.scala
```diff
@@ -47,7 +47,7 @@ protected[entity] abstract class Limits {
  * @param timeout the duration in milliseconds, assured to be non-null because it is a value
  * @param memory the memory limit in megabytes, assured to be non-null because it is a value
  */
-protected[core] case class ActionLimits protected[core] (timeout: TimeLimit, memory: MemoryLimit) extends Limits {
+protected[core] case class ActionLimits protected[core] (timeout: TimeLimit, memory: MemoryLimit, logs: Option[LogLimit] = Some(LogLimit())) extends Limits {
```
rabbah: update javadoc.
add comment that log limit is an option for backward compatibility/schema migration.

rabbah:
Nicely done. As you respond to my comments, if there are additional commits please don't squash them (so we can attribute commits to the comments/response).

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -369,15 +368,15 @@ class Invoker(
...
      * @param strings The list of strings
-     * @param sizeLimit The limit of strings to get
+     * @param limit The limit of strings to get
      */
-    def truncateLogs(strings: Vector[String], sizeLimit: ObjectSize): Vector[String] = {
+    def truncateLogs(strings: Vector[String], limit: LogLimit): Vector[String] = {
         val stringStream = strings.toStream
         val cummulativeSizes = stringStream.scanLeft(0.bytes) { (acc, current) => acc + current.sizeInBytes }.tail
-        val truncatedLogs = stringStream.zip(cummulativeSizes).takeWhile(_._2 < sizeLimit).map(_._1).toVector
+        val truncatedLogs = stringStream.zip(cummulativeSizes).takeWhile(_._2 < limit.megabytes.megabytes).map(_._1).toVector
```
rabbah: can you add an apply method to that limit.megabytes() eschews the double megabytes.megabytes?

common/scala/src/main/scala/whisk/core/entity/Limits.scala
```diff
@@ -67,7 +67,15 @@ protected[core] object ActionLimits
     /** Creates a ActionLimits instance with default duration and memory limit. */
     protected[core] def apply(): ActionLimits = ActionLimits(TimeLimit(), MemoryLimit())
 
-    override protected[core] implicit val serdes = jsonFormat3(ActionLimits.apply)
+    override protected[core] implicit val serdes = new RootJsonFormat[ActionLimits] {
+        val helper = jsonFormat3(ActionLimits.apply)
+
+        def read(value: JsValue) = {
+            val inter = helper.read(value)
+            inter.copy(logs = Some(inter.logs.getOrElse(LogLimit())))
```
rabbah: the dual  - since you have a custom serdes - is to make ActionLimit use a definite LogLimit (not an Option) and tolerate the missing LogLimit in the serializer.

markusthoemmes:
PPG 876 was green

common/scala/src/main/scala/whisk/core/entity/LogLimit.scala
```diff
@@ -0,0 +1,67 @@
...
+}
+
+protected[core] object LogLimit extends ArgNormalizer[LogLimit] {
+    protected[core] val STD_LOGSIZE = 10 // MB
+
+    /** Gets LogLimit with default log limit */
+    protected[core] def apply(): LogLimit = new LogLimit(STD_LOGSIZE)
+
+    /**
+     * Creates LogLimit for limit. Only the default limit is allowed currently
```
rabbah: add `.` at end of comment.

common/scala/src/main/scala/whisk/core/entity/LogLimit.scala
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @param megabytes the limit in megabytes, must be within permissible range
+     * @return LogLimit with limit set
+     * @throws IllegalArgumentException if limit does not conform to requirements
+     */
+    @throws[IllegalArgumentException]
+    protected[core] def apply(megabytes: Int): LogLimit = {
+        require(megabytes == STD_LOGSIZE, "only standard log limit allowed")
```
rabbah: show the standard log limit in error message?

common/scala/src/main/scala/whisk/core/entity/LogLimit.scala
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    override protected[core] implicit val serdes = new RootJsonFormat[LogLimit] {
+        def write(m: LogLimit) = JsNumber(m.megabytes)
+
+        def read(value: JsValue) = Try {
+            val JsNumber(mb) = value
+            require(mb.isWhole(), "log limit must be whole number")
+            LogLimit(mb.intValue)
+        } getOrElse deserializationError("log limit malformed")
```
rabbah: note that the inner error message is traded here for more generic malformed message (just pointing it out in case it was accidental).

tests/src/system/basic/ActionTests.scala
```diff
@@ -76,6 +83,43 @@ class ActionTests
...
+            val name = "TestActionCausingExceededLogs"
+            assetHelper.withCleaner(wsk.action, name, confirmDelete = true) {
+                val actionName = TestUtils.getTestActionFilename("dosLogs.js")
+                (action, _) => action.create(name, Some(actionName))
+            }
+
+            val run = wsk.action.invoke(name)
+            withActivation(wsk.activation, run) { response =>
+                response.fields("response").asJsObject.fields("result").toString should include(s""""msg":1""")
+                response.fields("response").asJsObject.fields("status").toString should include("success")
```
rabbah: check the log message exists and only the log message?

tests/src/system/basic/ActionTests.scala
```diff
@@ -76,6 +83,43 @@ class ActionTests
...
+            val allowedSize = 10 megabytes
+            val writing = allowedSize + 2.megabytes
+            val characters = writing.toBytes / 2 // a character takes 2 bytes
+
+            val run = wsk.action.invoke(name, Map("payload" -> characters.toJson))
+            withActivation(wsk.activation, run) { response =>
+                val lines = response.fields("logs").convertTo[List[String]]
+                lines.last should be("Logs have been truncated because they exceeded the limit of 10 megabytes")
+                // dropping 39 characters (timestamp + streamname)
+                lines.dropRight(1).map(_.drop(39)).mkString.sizeInBytes should be < allowedSize
```
rabbah: should be `<=`

common/scala/src/main/scala/whisk/core/entity/LogLimit.scala
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * LogLimit encapsulates allowed amount of logs written by an action.
+ *
+ * It is a value type (hence == is .equals, immutable and cannot be assigned null).
+ * The constructor is private so that argument requirements are checked and normalized
+ * before creating a new instance.
+ *
+ * @param megabytes the memory limit in megabytes for the action
+ */
+protected[core] class LogLimit private (val megabytes: Int) extends AnyVal {
```
rabbah: maybe a comment here that it's int for json serializer (vs the new size types).

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/b17eda57984d579ebfe05c0e738cd7913f3f3bf2.

PR #858, opened 2016-07-06
Fixes issue 843

rabbah:
@fxulusoy want to give this a try? 
@domdom82 @lehoanganh please review - I added the essential bits from the prereq playbook.

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -1,10 +1,10 @@
-Deploying Openwhisk using Ansible
+Deploying OpenWhisk using Ansible
 =========
 
 
 ### Getting started
 
-If you want to deploy Openwhisk locally using Ansible, you first need to install Ansible on your development environment:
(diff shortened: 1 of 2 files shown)


fxulusoy:
@rabbah I have tried many reboots, pause/start vm and it works. The changes LGTM

domdom82:
looks ok although it has the downside of managing the prereqs in two places now. if we up the version or add more packages, we must not forget to update the tweak script, too. on the other hand I don't see any other way without tainting the common ansible playbooks with `if mac then xxx` exceptions.

rabbah:
valid point - I'm biased toward convenience. The playbook is still valid and it's harmless to rerun even if the tweak is applied. With automated testing for the docker machine (or with docker native for max) this might become moot.

fxulusoy:
Merged. https://github.com/openwhisk/openwhisk/commit/ea03638fa2106bafe1b313f4b34e8fd4831bd702

Issue #719, opened 2016-06-21
Add the sh script to install the openwhisk-catalog


houshengbo:
Since openwhisk-catalog is ready to use, a script should be added into openwhisk to install all the catalog from openwhisk-catalog.

rabbah:
@nwspeete-ibm this is related to the push notification package. Is there a reusable solution?

FYI @csantanapr.

csantanapr:
I think we need to wait to see what can be consolidated and refactor.

For now I would suggest to

1. Re-use the file name `catalog/installCatalog.sh`
The new content will be something like 
````
git clone https://github.com/openwhisk/openwhisk-catalog
cd  openwhisk-catalog
/.install.sh
```
2. Move the contents of `catalog/installCatalog.sh` into new `openwhisk-catalog/install.sh`

houshengbo:
I can definable change to the re-usable way. We do not need to create the install.sh, because I have already made the installcatalog.sh under openwhisk-catalog/packages, able to call and install all the packages.

csantanapr:
Closing with #885

PR #860, opened 2016-07-06
Make the installation of catalog optional

houshengbo:
As a non-core component, catalog installation should be optional.
Currently, the catalog is installed by default together with
openwhisk. This patch adds a parameter named catalog_source to
determine where and whether to install the catalog.

If catalog_source is "openwhisk": install the catalog from openwhisk.
If it is "openwhisk-catalog", install the catalog from openwhisk-catalog.
If it is set to "none", do not install catalog by default.

Closes-Bug: #719


houshengbo:
@rabbah @markusthoemmes @lehoanganh 
This patch adds a argument to determine whether and where openwhisk.yml can install the catalog. It lays a good foundation to transition from openwhisk/catalog to openwhisk-catalog.

Issue #661, opened 2016-06-15
Language runtimes need to send a sentinel marker to mark the end of an activation


perryibm:
Currently, only nodejs does this and so we are at risk of losing some output in case docker log flushing is tardy.  We mitigate this with a sleep (around 100ms I believe) which is slow and still not correct.  We should fix this for all our languages - starting with Swift/Python.

Once the marker is in place, we can enable support for it in the invoker which must happen at the same time as generating the markers.  Markers should be generated on both stdin and stdout and both streams should be flushed after the markers are generated.

paulcastro:
PG 675

ioana-blue:
As far as I understand, there is no test in place for this, correct? Once we add the markers, @perryibm will add code in the Invoker to deal with this, right?

paulcastro:
For Swift, I had to modify the existing tests since they were expecting stdout to be empty.  That may not be necessary for Python?

If @perryibm is busy I can help add the code to the Invoker since I just did it for Swift.

rabbah:
Go for it :+1:.

PR #861, opened 2016-07-07
Issue #661 Add sentinel marker to Swift and Swift3 runtimes, remove u…

paulcastro:
…nnecessary Xlinker flags in swift build command


core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -315,7 +319,8 @@ class Invoker(
             val rawLogBytes = con.getDockerLogContent(con.lastLogSize, size, runningInContainer)
             val rawLog = new String(rawLogBytes, "UTF-8")
             val isNodeJs = con.image == nodejsImageName || con.image == nodejs6ImageName
-            val (complete, logArray) = processJsonDriverLogContents(rawLog, isNodeJs)
+            var isSwift = con.image == swiftImageName || con.image == swift3ImageName
```
markusthoemmes: why is this a `var`? Does it need to be mutable?

PR #868, opened 2016-07-08
Optimize completion path of active ack

perryibm:
Currently we fetch the logs before making an activation record which we send back.  Instead, since active ack does not carry logs anyway, we can send back a log-less activation record first.  Then get the logs, make a more complete activation record, and records this in the database.  PPG811 passed


rabbah:
proposing this patch to your pr: https://github.com/openwhisk/openwhisk/commit/5d4c16532ad14e1c81a67e047ff5f082a2a1b11f

perryibm:
Squashed and retesting

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/69a23f9a85693b9b57f7a7160815d02ac6bb0b7b.

Issue #804, opened 2016-06-29
go CLI cannot create docker action


rabbah:
## Environment details:

* mac + docker-machine

## Steps to reproduce the issue:

1. build and deploy locally
2. set API host to local deployment
3.  ./bin/go-cli/wsk action create example some/image --docker  

## Provide the expected results and outputs:

```
> ./bin/wsk action create example some/image --docker  # using python CLI
ok: created action example
```

## Provide the actual results and outputs:

```
> ./bin/go-cli/wsk action create example some/image --docker -i
error: Unable to create action: error: Request failure: error: The request content was malformed:
'kind' must be a string defined in 'exec' (code 86)
```

It looks like we're missing a CLI test for docker actions.

@mdeuser @dubeejw

dubeejw:
https://github.com/openwhisk/openwhisk/pull/795

PR #869, opened 2016-07-08
Fix Docker Actions in Go CLI

dubeejw:
- Set "kind" to "blackbox" when creating an action with a Docker container
- Create a CLI test that creates actions with Docker containers

### DIFF
--- a/tests/src/system/basic/WskBasicTests.scala
+++ b/tests/src/system/basic/WskBasicTests.scala
@@ -335,24 +335,6 @@ class WskBasicTests
             wsk.action.create("updateMissingFile", Some("notfound"), update = true, expectedExitCode = MISUSE_EXIT)
     }
 
-    ignore should "create, and invoke an action that utilizes a docker container" in withAssetCleaner(wskprops) {
-        val name = "dockerContainer"
-        (wp, assetHelper) =>
-            assetHelper.withCleaner(wsk.action, name) {
-                // this docker image will be need to be pulled from dockerhub and hence has to be published there first
-                (action, _) => action.create(name, Some("whisk/dockerskeleton"), kind = Some("docker"))
-            }
-
-            val args = Map("payload" -> "test".toJson)
-            val run = wsk.action.invoke(name, args)
-            withActivation(wsk.activation, run) {
-                activation =>
-                    val result = activation.fields("response").asJsObject.fields("result").asJsObject
-                    result.fields("args") shouldBe args.toJson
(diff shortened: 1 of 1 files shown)


dubeejw:
https://github.com/openwhisk/openwhisk/issues/804

rabbah:
`with container` is not necessary here - that's for testing the container directly (vs the intended test with is an integration test through the CLI).

hence remove it:

```
-    val dockerSdkContainerImageName = "whisk/dockerskeleton"
-
-    // Helpers specific to dockerskeleton
-    def withDockerSdkContainer(code: ActionContainer => Unit) = {
-        withContainer(dockerSdkContainerImageName)(code)
-    }
-
     it should "create, and invoke an action that utilizes a docker container" in withAssetCleaner(wskprops) {
         (wp, assetHelper) =>
-            val (out, err) = withDockerSdkContainer { c =>
-                val name = "dockerContainer"
-                assetHelper.withCleaner(wsk.action, name) {
-                    (action, _) => action.create(name, Some("whisk/dockerskeleton"), kind = Some("docker"))
-                }
-
-                val run = wsk.action.invoke(name, Map("payload" -> "test".toJson), blocking = true, result = true)
-                run.stdout should include regex (""""msg": "Hello from arbitrary C program!"""")
-                run.stdout should include regex (""""payload": "test"""")
+            val name = "dockerContainer"
+                (action, _) => action.create(name, Some("whisk/dockerskeleton"), kind = Some("docker"))
             }
 
-            out.trim shouldBe empty
-            err.trim shouldBe empty
+            val run = wsk.action.invoke(name, Map("payload" -> "test".toJson), blocking = true, result = true)
+            run.stdout should include regex (""""msg": "Hello from arbitrary C program!"""")
+            run.stdout should include regex (""""payload": "test"""")
```

dubeejw:
IPG 290.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/13ece35b1fc0359917f60b354c0f1f1f6dce75e9.

rabbah:
I had to revert this test: https://github.com/openwhisk/openwhisk/commit/1e0166750ce444334be7ea87d7d9ce32ea6d9d9d. The IPG did not pass (my fault, looked up the wrong one). The test fails because the invoker will attempt to pull the action from docker hub which will fail.

We could push that image locally to the invoker but that means changing the registry also. We can push it to docker hub but then we have to rename it to openwhisk/dockerskeleton.

dubeejw:
Let's push the image to Docker hub? I don't have the credentials for the openwhisk account though.

PR #872, opened 2016-07-09
Kafka consumer throttling in invoker

### LINKED ISSUE #143: Kafka backpressure
Currently the invoker absorbs kafka messages as fast as possible without heed to what it can truly handle.  This is bad because metrics based on kafka queues are misleading and also prevents RAS as the absorbed (yet unprocessed) messages are lost.  The load balancer is also potentially confused about how things are progressing.  Finally, the invoker can start too many activations and start thrashing and/or disrupt fairness/QoS.

The plan is to upgrade to the latest Kafka bindings which afford us to control absorption rate and then throttle this based on availability determined from the ContainerPool.  Heuristically, we probably don't want to completely throttle until there is slack but rather absorb just a few extra so activity remains pipelined.

rabbah:
Introduced an actor polls the message bus for new messages and dispatches them to the given
handler. The actor tracks the number of messages dispatched and will not dispatch new
messages until some number of them are acknowledged.

This is used by the invoker to pull messages from the message bus and apply back pressure
when the invoker does not have resources to complete processing messages (i.e., no containers
are available to run new actions).

When the invoker releases resources (by reclaiming containers) it will send a message
to this actor which will then attempt to fill the pipeline with new messages.

The actor tries to fill the pipeline with additional messages while the number
of outstanding requests is below the pipeline fill threshold.

IPG 288 approved. Closes #143.


common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -99,9 +84,10 @@ class KafkaConsumerConnector(
         val props = new Properties
         props.put(ConsumerConfig.GROUP_ID_CONFIG, groupid)
         props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, kafkahost)
-        props.put(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG, "10000");
+        props.put(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG, 60000.toString);
         props.put(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, true.toString)
-        props.put(ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG, "10000");
+        props.put(ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG, 10000.toString);
+        props.put(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, maxMessagePerPoll.toString)
```
rabbah: using kafka 10 feature to provide a bound on max records returned from poll per partition.

core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -102,7 +102,7 @@ class ContainerPool(
     def maxActive_=(value: Int): Unit = _maxActive = Math.max(0, value)
 
     def resetMaxIdle() = _maxIdle = defaultMaxIdle
-    def resetMaxActive() = _maxActive = defaultMaxActive
+    def resetMaxActive() = _maxActive = ContainerPool.defaultMaxActive
```
rabbah: moved to singleton to it is accessible in invoker (to control depth of pipeline).

core/dispatcher/src/main/scala/whisk/core/dispatcher/ActivationFeed.scala
```diff
@@ -0,0 +1,98 @@
...
+    /** Pulls new messages from the message bus. */
+    case class FillQueueWithMessages()
+
+    /** Indicates resources are available because transaction completed, may cause pipeline fill. */
+    case class ContainerReleased(tid: TransactionId)
+
+    /** Indicate resources are available because transaction failed, may cause pipeline fill. */
+    case class FailedActivation(tid: TransactionId)
+}
+
```
rabbah: the throttling is applied by this class/actor.

core/dispatcher/src/main/scala/whisk/core/dispatcher/Dispatcher.scala
```diff
@@ -102,57 +87,42 @@ trait MessageDispatcher extends Registrar with Logging {
...
             info(this, s"activeCount = $count while handling ${rule.name}")
-            rule.doit(topic, msg, matches)
+            rule.doit(topic, msg, matches) // returns a future which is flat-mapped to hang onComplete
         } flatMap (identity) onComplete {
             case Success(a) => info(this, s"activeCount = ${counter.prev()} after handling $rule")
             case Failure(t) => error(this, s"activeCount = ${counter.prev()} ${errorMsg(rule, t)}")
         }
     }
 
-    private def inform(matchers: ParTrieMap[String, DispatchRule])(implicit transid: TransactionId) = {
```
rabbah: remove Par collection since there is only one invoker (and generally one consumer per dispatcher).
it's arguable if we should flatten this further - i opted to defer for now.

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -258,7 +277,7 @@ class Invoker(
                     case None => (false, con.run(params, msg.meta, auth.compact, timeoutMillis,
                         action.fullyQualifiedName, msg.activationId.toString)) // cached
                     case Some((start, end, Some((200, _)))) => { // successful init
-                        // Try a little slack for json log driver to process
+                        // TODO:  @perryibm update comment if this is still necessary else remove
```
rabbah: I'm fairly sure we can nuke this. But deferred from this PR.

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -99,9 +84,10 @@ class KafkaConsumerConnector(
         val props = new Properties
         props.put(ConsumerConfig.GROUP_ID_CONFIG, groupid)
         props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, kafkahost)
-        props.put(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG, "10000");
+        props.put(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG, 10000.toString)
         props.put(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, true.toString)
-        props.put(ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG, "10000");
+        props.put(ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG, 10000.toString)
+        props.put(ConsumerConfig.MAX_POLL_RECORDS_CONFIG, maxMessagePerPoll.toString)
```
rabbah: using kafka 10 feature to provide a bound on max records returned from poll per partition.
default set to preserve existing behavior (return all records).

core/dispatcher/src/main/scala/whisk/core/dispatcher/Dispatcher.scala
```diff
@@ -165,41 +134,56 @@ trait MessageDispatcher extends Registrar with Logging {
...
+     * @param handler is the message handler to add override
+     * @param replace indicates whether a new handler should replace an older handler by the same name
+     * @return an option dispatch rule, the previous value of the rule if any
+     */
+    def addHandler(handler: DispatchRule, replace: Boolean): Option[DispatchRule] = {
+        if (handler != null && handler.isValid) {
+            if (replace) handlers.put(handler.name, handler)
+            else handlers.putIfAbsent(handler.name, handler)
+        } else None
+    }
```
markusthoemmes: `null` 😱 

This seems to swallow errors if the creation of for example `Invoker` fails? Shouldn't we handle at least the `isValid` case outside of this method?

tests/src/services/KafkaConnectorTests.scala
```diff
@@ -62,7 +62,7 @@ class KafkaConnectorTests extends FlatSpec with Matchers with BeforeAndAfter {
               val message = new Message { override val serialize = Calendar.getInstance().getTime().toString }
               val start = java.lang.System.currentTimeMillis
               val sent = Await.result(producer.send(topic, message), 10 seconds)
-              val received = consumer.getMessages(10 seconds).map { r => new String(r.value, "utf-8") }
+              val received = consumer.getMessages(10 seconds).map { r => new String(r._4, "utf-8") }
```
markusthoemmes: Deconstruct the tuple for more readability.

core/dispatcher/src/main/scala/whisk/core/dispatcher/ActivationFeed.scala
```diff
@@ -0,0 +1,100 @@
...
+        case FillQueueWithMessages => Try {
+            val messages = consumer.getMessages(longpollDuration) // grab next batch of messages
+            val count = messages.size
+            messages foreach {
+                case (topic, partition, offset, bytes) =>
+                    logging.info(this, s"processing $topic[$partition][$offset ($count)]")(TransactionId.dispatcher)
+                    pipelineOccupancy += 1
+                    handler(topic, bytes)
+            }
+            fill()
```
psuter: If I understand correctly, this is basically `while(true)`, and will thus also always be using a thread (because `consumer.getMessages` is blocking)? However, because the next iteration of the "loop" is triggered by an actor, other tasks may be scheduled in-between, so in theory it should even work with a thread pool of size 1?

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
perryibm: yes, let's keep orthogonal semantic changes to other PRs in case there are issues

core/dispatcher/src/main/scala/whisk/core/dispatcher/Dispatcher.scala
perryibm: null isn't great but no worse than isValid being false.   Why would a handler be invalid and is it possible to make it statically impossible?  We should at least issue a warning for any bad case here.

core/dispatcher/src/main/scala/whisk/core/dispatcher/ActivationFeed.scala
perryibm: It would "work" with a thread pool of size 1 but not robustly if you mean "pool" generally.  If it's really a pool, that means someone else is part of the pool and they could get blocked for a long time on that call.

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -48,39 +50,22 @@ class KafkaConsumerConnector(
...
-                                info(self, s"processing $topic[$partition][$offset ($count)]")
-                                val consumed = process(topic, msg)
-                            }
-                            commit()
-                    } match {
-                        case Failure(t) => error(self, s"while polling: $t")
-                        case _          => // ok
+                    Try { getMessages() } match {
+                        case Success(records) =>
+                            records foreach { case (t, p, o, b) => process(t, p, o, b) }
```
perryibm: Check out `Function.tupled`

core/dispatcher/src/main/scala/whisk/core/dispatcher/ActivationFeed.scala
```diff
@@ -0,0 +1,100 @@
...
+    logging: Logging,
+    consumer: MessageConsumer,
+    maxPipelineDepth: Int,
+    longpollDuration: FiniteDuration,
+    handler: (String, Array[Byte]) => Any)
+    extends Actor {
+    import ActivationFeed.ActivationNotification
+    import ActivationFeed.FillQueueWithMessages
+
+    require(maxPipelineDepth >= 2 && maxPipelineDepth % 2 == 0, "max pipeline depth must be even and >= 2")
```
perryibm: Why does it have to be even?

core/dispatcher/src/main/scala/whisk/core/dispatcher/ActivationFeed.scala
```diff
@@ -0,0 +1,100 @@
...
+    require(consumer.maxMessagePerPoll <= maxPipelineDepth, "consumer may yield more messages per poll than permitted by max depth")
```
perryibm: Don't you mean "may not"?

core/dispatcher/src/main/scala/whisk/core/dispatcher/ActivationFeed.scala
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        case _: ActivationNotification =>
+            pipelineOccupancy -= 1
+            fill()
+    }
+
+    private def fill() = {
+        if (pipelineOccupancy <= pipelineFillThreshold) {
+            logging.debug(this, s"filling activation pipeline: $pipelineOccupancy <= $pipelineFillThreshold")(TransactionId.dispatcher)
```
perryibm: Shouldn't it be `<` before it will actually fill?

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -232,6 +259,7 @@ class Invoker(
                 case None => {
                     activationCounter.next() // this is the global invoker counter
                     incrementUserActivationCounter(tran.msg.subject)
+                    tran.releaseResource(FailedActivation(transid))
```
perryibm: This is not a failed activation path but the path taken for the first one to record the activation completed.

core/dispatcher/src/main/scala/whisk/core/dispatcher/ActivationFeed.scala
rabbah: It doesn't - left over from a previous iteration - will remove it. The only constraint is the one below.

core/dispatcher/src/main/scala/whisk/core/dispatcher/Dispatcher.scala
rabbah: note: this is NOT new code; i just moved it down in the file.

that said, historically this was used in the activator which added and remove listeners (for rule matching) dynamically.

we have a dispatcher with exactly one handler - which is why i made the comment above that we should flatten "this" (class) further.

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
rabbah: either we release the resource because of a FailedActivation, or, earlier when the container is put back. Note that in releaseResource, there is a one time flag that checks if the message was already sent. If the releaseResource call on line 262 succeeds it's because the activation failed to return the container (which generally means the future failed).

core/dispatcher/src/main/scala/whisk/core/dispatcher/ActivationFeed.scala
rabbah: `pipelineFillThreshold = maxPipelineDepth - consumer.maxMessagePerPoll`
`pipelineFillThreshold + consumer.maxMessagePerPoll =
maxPipelineDepth - consumer.maxMessagePerPoll + consumer.maxMessagePerPoll = maxPipelineDepth`

core/dispatcher/src/main/scala/whisk/core/dispatcher/ActivationFeed.scala
rabbah: the number of threads in the invoker should be: N+2 where N is max parallel containers (maxdepth/2), and two threads each for kafka connector and this kafka consumer.

core/dispatcher/src/main/scala/whisk/core/dispatcher/Dispatcher.scala
rabbah: i'm agreeable to removing the registrator altogether as part of this pull request - if we're not overloading a topic, we don't need the inner path matching anymore.

rabbah:
PPG 840 approved.

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -48,39 +50,27 @@ class KafkaConsumerConnector(
      * duration.
      *
      * @param duration the maximum duration for the long poll
-     * @param syncCommit update offsets on the topic immediately.  Default false as we use auto-commit.
      */
-    def getMessages(duration: Duration = 500 milliseconds, syncCommit: Boolean = false) = {
+    override def peek(duration: Duration = 500 milliseconds) = {
```
rabbah: renamed to peek - to emphasize need for commit() (even with kafka auto commit, want to commit offsets when we have processed messages eagerly).

rabbah:
Rebased and ready to merge.

PR #873, opened 2016-07-11
Refactor Go CLI error messages so that a single "error:" appears at the start of the message

### LINKED ISSUE #806: CLI error message contains multiple occurences of the word "error"
## Environment details:

* local deployment: vagrant on Windows   
* docker 1.9.1, vagrant 1.8.8, ubuntu (from vagrant)

1. build and deploy locally
2. set API host to local deployment
3. ./bin/go-cli/wsk action list --apihost invalid.host.name.com (or another failing command)

An error message that's readable, without multiple unnecessary "error" strings inside the message
```
error: Unable to list action(s): Request failure: Get https://bad.missing.com/api/v1/namespaces/mdeuser_dev/actions?limit=30&skip=0: dial tcp: lookup bad.missing.com: getaddrinfow: No such host is known.
```

```
error: Unable to list action(s): error: Request failure: error: Get https://bad.missing.com/api/v1/namespaces/mdeuser_dev/actions?limit=30&skip=0: dial tcp: lookup bad.missing.com: getaddrinfow: No such host is known.
```

* This issue affects more than just the above example

mdeuser:
Refactor error messages so that a single "error:" appears at the start of the message
Fixes #806

### DIFF
--- a/tools/go-cli/go-whisk-cli/commands/property.go
+++ b/tools/go-cli/go-whisk-cli/commands/property.go
@@ -317,7 +317,7 @@ func getPropertiesFilePath() (propsFilePath string, werr error) {
 
         if err != nil {
             whisk.Debug(whisk.DbgError, "homedir.Expand(%s) failed: %s\n", Properties.PropsFile, err)
-            errStr := fmt.Sprintf("Unable to locate properties file '%s'; error %s", Properties.PropsFile, err)
+            errStr := fmt.Sprintf("Unable to locate properties file '%s': %s", Properties.PropsFile, err)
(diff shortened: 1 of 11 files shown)


mdeuser:
@dubeejw - Ready for review

mdeuser:
IPG 300

dubeejw:
@mdeuser, instead of adding the original error message to the end of each new error message, why don't we just concatenate the old error message to the new one inside the MakeWskErrorFromWskError() method?

mdeuser:
@dubeejw - Thanks for the review.  Here are a few reasons for keeping the message concatenation processing as it is.  Encapsulation - each error object is responsible and stores it's own error message; only pull in another object's error information when it is needed.  Secondly, while memory consumption is not really an issue for the CLI right now, concatenation the sub-error message would increase memory consumption for errors that are made up of a series of error objects.  Also, there's not much performance savings by concatenating each message during the error object construction since the CLI only calls toString() once.

PR #874, opened 2016-07-12
Fix formatting for command examples.

pparkkin:
I noticed the formatting for the commands in the Ubuntu setup doc was a little funky, so I decided to fix it.


rabbah:
Will merge as part of #894. Thanks for the path @pparkkin.

PR #877, opened 2016-07-13
Save Markers in map to calculate timediff

ChrBi:
Signed-off-by: Markus Thoemmes <markus.thoemmes@de.ibm.com>

This PR adds the ability to calculate the time difference between markers in the core components.
We decided to calculate the time between all markers, because we agreed on calculate the time between nearly all the markers.


ChrBi:
Won't be solved like this.

PR #883, opened 2016-07-13
Flip the Switch to Install the Catalog with the Go CLI

dubeejw:
- Use the Go CLI to install the default catalog instead of the Python CLI
- Fix error occurring in installWeather.sh when creating the Weather package
- Fix annotation formatting in installWeather.sh, and installSystem.sh
- Do not use a property file if WSK_CONFIG_FILE is set, but blank

### DIFF
--- a/catalog/util.sh
+++ b/catalog/util.sh
@@ -15,7 +15,7 @@ if [ ! -f "$WHISKPROPS_FILE" ]; then
 fi
 EDGE_HOST=`fgrep edge.host= "$WHISKPROPS_FILE" | cut -d'=' -f2`
 
-USE_PYTHON_CLI=true
+USE_PYTHON_CLI=false
 
 function createPackage() {
     PACKAGE_NAME=$1
--- a/catalog/installSystem.sh
+++ b/catalog/installSystem.sh
@@ -32,9 +32,9 @@ install "$CATALOG_HOME/utils/cat.js" \
 install "$CATALOG_HOME/utils/split.js" \
      util/split \
      -a description 'Split a string into an array' \
-     -a parameters '[{"name": "payload", "required":true, "description":"A string"}], { "name": "separator", "required": false, "description": "The character, or the regular expression, to use for splitting the string" }]' \
+     -a parameters '[{"name": "payload", "required":true, "description":"A string"}, { "name": "separator", "required": false, "description": "The character, or the regular expression, to use for splitting the string" }]' \
      -a sampleInput '{ "payload": "one,two,three", "separator": "," }' \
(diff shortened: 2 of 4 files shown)


dubeejw:
IPG 306.

dubeejw:
IPG 320.

dubeejw:
IPG 341.

dubeejw:
IPG 341

dubeejw:
Don't merge yet.

rabbah:
We need to test that params and annotations have the expected type. :)

dubeejw:
Parameters are fine. The annotations will be the issue.

jasonpet:
Opened https://github.com/openwhisk/openwhisk/issues/952 for issue with annotations

dubeejw:
Depends on https://github.com/openwhisk/openwhisk/pull/953.

csantanapr:
@dubeejw is this wip?

rabbah:
This was blocked and now pending new pg.

dubeejw:
@rabbah, @csantanapr, PG1 128. Should be ready for merge now.

dubeejw:
PG1 133.

rabbah:
PG1-133 approved. Feeling lucky?

catalog/util.sh
```diff
@@ -25,7 +25,7 @@ function createPackage() {
     else
         CMD_ARRAY=("$OPENWHISK_HOME/bin/go-cli/wsk" -i --apihost "$EDGE_HOST" package update --auth "$AUTH_KEY" --shared yes "$PACKAGE_NAME" "${REST[@]}")
     fi
-    export WSK_CONFIG_FILE= # override local property file to avoid namespace clashes
+    export WSK_CONFIG_FILE=None # override local property file to avoid namespace clashes
```
rabbah: why is None required here? the alternative won't work with go cli?

rabbah: I don't see that we have a test for this (would require saving ~/.wskprops, creating a new one with a namespace set, running a wsk command with WSK_CONFIG_FILE override, and confirming that namespace from wskprops is not read, then restoring ~/.wskprops (unconditionally)).

dubeejw: @rabbah, I fixed the above logic to not use a config file if the WSK_CONFIG_FILE is set to an empty value.

For the test we could set the config file to a blank value then try to do a property set, which should fail since no property file is defined.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/b7f8c330b63cfd795e03de328a6082acfad9cc93.

Issue #815, opened 2016-06-30
go CLI does not show response.result.error


jasonpet:
The response.result.error field is not being shown in the go CLI.  You can verify this by calling whisk.error() in an async action.  The go CLI simply shows the following: 
error: Unable to invoke action <action name>: error: Request failure: error:  (code 0)

mdeuser:
@rabbah - What is the expected HTTP response status code for an action invocation of an action that calls whisk.error() ?

PR #884, opened 2016-07-13
Arguments are not Parsed Properly for Package Refresh

dubeejw:
- Update argument parsing for package refresh to handle no arguments
- Use the namespace from the property file, or _ if a namespace is not provided when parsing the qualified name

### DIFF
--- a/tools/go-cli/go-whisk-cli/commands/package.go
+++ b/tools/go-cli/go-whisk-cli/commands/package.go
@@ -20,6 +20,7 @@ import (
     "errors"
     "fmt"
     "net/http"
+    "strings"
 
     "../../go-whisk/whisk"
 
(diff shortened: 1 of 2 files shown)


dubeejw:
IPG 307.

dubeejw:
Please review.

dubeejw:
Any comments on this one?

mdeuser:
@dubeejw - What issue is associated with this PR?

mdeuser:
@dubeejw - The help syntax should also change to reflect the optional nature of the argument.
<namespace string> -> [namespace string]

dubeejw:
Help updated.

rabbah:
`wsk package refresh` in an openwhisk deployment - from an end user perspective - will appear to be a bug in the CLI/backend because it's making a POST to the packages API. This route should be intercepted by nginx and an error (method not yet supported?) reported instead. @markusthoemmes thoughts on this?

markusthoemmes:
@rabbah Yes, that's exactly the behaviour we had on our old CLI. We had that behaviour once, but checking the history of our nginx we seem to have lost it at some point? There used to be a route in there intercepting as you described.

rabbah:
maybe it was lost during ansible migration?

markusthoemmes:
Already verified, it was gone before.

dubeejw:
Fixes https://github.com/openwhisk/openwhisk/issues/815

PR #885, opened 2016-07-13
Make the catalog able to install from openwhisk-catalog repos

houshengbo:
Currently, the catalog is installed by default together with
openwhisk. This patch adds a parameter named catalog_source to
determine whether to install the catalog from openwhisk-catalog.

If it is "catalog-repos", install the catalog from all the available
catalog repos. If it is set to any value except "catalog-repos",
openwhisk will not install catalog from the openwhisk-catalog.
The command to install the catalog from openwhisk-catalog is:
"ansible-playbook -i environments/local postdeploy.yml -e catalog_source=catalog-repos".

The script installCatalog.sh in the catalog needs to be able to
parse catalog_auth_key, api_host and catalog_namespace as the arguments.

Closes-Bug: #719

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -48,3 +48,47 @@ db:
   authkeys:
   - guest
   - whisk.system
+
+# The default value for catalog_source is empty, meaning that
+# openwhisk will not install the catalog from the catalog repositories.
+# If catalog_source is not set to "catalog-repos", the other variables
+# like catalog_namespace, catalog_auth_key and catalog_repos will not
+# be effective.
+catalog_source: ""
+
+# The default name space is /whisk.system. The catalog namespace must begin with a slash "/".
+catalog_namespace: "/whisk.system"
+
+# The catalog_auth_key is used to determine the secret key to authenticate the openwhisk service.
+# The value for this variable can be set to either the secret key itself or the file, which
+# saves the secret key.
+# By default, we take the key from {{ openwhisk_home }}/ansible/files/auth.whisk.system.
+catalog_auth_key: "{{ openwhisk_home }}/ansible/files/auth.whisk.system"
+
+# The catalog_repos is used to specify all the catalog names and repository URLs,
+# so that openwhisk knows where to download the catalog and install them. The key
+# specifies the catalog name and the url saves the URL of the repository. The location
+# specifies the location to save the code of the catalog. The version specifies the hash
+# of the commit to be cloned. If it is omit or set to HEAD, the latest commit will be
+# selected. The repo_update specifies whether to retrieve new revisions from the origin
+# repository and the default value is yes, meaning that it will retrieve the new
+# revisions. The keys url and location are mandatory and the keys version and repo_update
+# are optional. To add a new repository, please follow the template by adding:
+#
+# catalog_repos:
+#   ...
+#   <catalog-name>:
+#     url: <URL of repository>, mandatory.
+#     location: <local location to save the catalog>, mandatory.
+#     version: <hash of the commit>, optional, default to HEAD.
+#     repo_update: <whether to retrieve new revisions from the origin repository>,
+#                  optional, default to yes. Yes means to retrieve the new revisions, and
+#                  no means not to retrieve the new revisions.
+#
+catalog_repos:
+  openwhisk-catalog:
+    url: https://github.com/openwhisk/openwhisk-catalog.git
+    # Set the local location as the same level as openwhisk home, but it can be changed.
+    location: "{{ openwhisk_home }}/../openwhisk-catalog"
+    version: "HEAD"
+    repo_update: "yes"
--- a/ansible/postdeploy.yml
+++ b/ansible/postdeploy.yml
@@ -6,4 +6,8 @@
 - hosts: ansible
   tasks:
     - include: tasks/installCatalog.yml
-      when: mode == "deploy"
+      when: (mode == "deploy" and catalog_source != "catalog-repos")
+      
+    - include: tasks/installOpenwhiskCatalog.yml
+      when: (mode == "deploy" and catalog_source == "catalog-repos")
+      with_dict: "{{ catalog_repos }}"
--- a/ansible/tasks/installOpenwhiskCatalog.yml
+++ b/ansible/tasks/installOpenwhiskCatalog.yml
@@ -1,0 +1,34 @@
+---
+# This task will install the standard actions and packages available in openwhisk-catalog repos.
+
+- set_fact:
+    catalog_location={{ item.value.location }}
+    catalog_repo_url={{ item.value.url }}
+    api_host={{ groups['edge'] | first }}
+    version="HEAD"
+    repo_update="yes"
+
+- set_fact:
+    version={{ item.value.version }}
+  when: item.value.version is defined
+
+- set_fact:
+    repo_update={{ item.value.repo_update }}
+  when: item.value.repo_update is defined
+
+- name: "ensure catalog_location directory exists"
+  file:
+    path: "{{ catalog_location }}"
+    state: directory
+
+- name: download the catalog repository to the catalog location if necessary
+  git:
+    repo: "{{ catalog_repo_url }}"
+    dest: "{{ catalog_location }}"
+    update: "{{ repo_update }}"
+    version: "{{ version }}"
+
+- name: install the catalog from the catalog location
+  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ catalog_namespace }} chdir="{{ catalog_location }}/packages"
+  environment:
+    OPENWHISK_HOME: "{{ openwhisk_home }}"


houshengbo:
I need people's reviews an many as possible, since this is very crucial PR for us to install the catalog from openwhisk-catalog.

ansible/postdeploy.yml
```diff
@@ -6,4 +6,9 @@
 - hosts: ansible
   tasks:
     - include: tasks/installCatalog.yml
-      when: mode == "deploy"
\ No newline at end of file
+      when: mode == "deploy"
+
+- hosts: ansible
+  tasks:
+    - include: tasks/installOpenwhiskCatalog.yml
```
rabbah: Should this be a comma separated list of GitHub repos?

ansible/tasks/installOpenwhiskCatalog.yml
```diff
@@ -0,0 +1,22 @@
+---
+# This task will install the standard actions and packages available in openwhisk-catalog.
+
+- name: the openwhisk-catalog URL and the location to download
+  set_fact:
+    openwhisk_catalog_url: "https://github.com/openwhisk/openwhisk-catalog.git"
+    openwhisk_catalog_home: "{{ openwhisk_home }}/../openwhisk-catalog"
```
rabbah: Should not hardcore where to install the catalog. How about an variable with this as a default.

ansible/tasks/installOpenwhiskCatalog.yml
```diff
@@ -0,0 +1,22 @@
...
+ 
+- name: download the openwhisk-catalog to the openwhisk-catalog location if necessary
+  git:
+    repo: "{{ openwhisk_catalog_url }}"
+    dest: "{{ openwhisk_catalog_home }}"
+    depth: 1
+  when: openwhisk_catalog_path.stat.isdir is not defined
+
+- name: install the catalog from openwhisk-catalog
+  local_action: shell "./installCatalog.sh" chdir="{{ openwhisk_catalog_home }}/packages"
```
rabbah: Can we pass some cli commands here: auth key, namespace, API host?

ansible/postdeploy.yml
houshengbo: @rabbah what do you mean by "list of GitHub repos"?
openwhisk-catalog is just one repo. I am not sure if I catch you. Could you elaborate more? Thank you.

ansible/tasks/installOpenwhiskCatalog.yml
houshengbo: I will change that.

ansible/tasks/installOpenwhiskCatalog.yml
houshengbo: I can change the script installCatalog.sh in openwhisk-catalog to receive these params.

ansible/postdeploy.yml
rabbah: it is one repo now - it won't always be one repo, plus we want this to be a model for others so if someone else contributes a repo with packages, we should be able to install them just the same.

houshengbo: yes. I can switch to a more general approach to accept multiple repos.

houshengbo:
There is a WIP PR in openwhisk-catalog, resolving the parameters' parsing: https://github.com/openwhisk/openwhisk-catalog/pull/91.

ansible/tasks/installOpenwhiskCatalog.yml
```diff
@@ -0,0 +1,22 @@
+---
+# This task will install the standard actions and packages available in openwhisk-catalog repos.
+
+- set_fact:
+    catalog_location={{ openwhisk_home }}/../{{ item.key }}
```
jasonpet: I think the user should be able to put openwhisk-catalog where they want.  We could use an environment variable here that points to the user's openwhisk-catalog.

ansible/postdeploy.yml
```diff
@@ -6,4 +6,8 @@
 - hosts: ansible
   tasks:
     - include: tasks/installCatalog.yml
-      when: mode == "deploy"
\ No newline at end of file
+      when: mode == "deploy"
+
```
jasonpet: It looks like the catalog is being installed from both openwhisk and openwhisk-catalog.  Is this by design?  Currently, these look identical so we won't really know if they are being installed properly from openwhisk-catalog if they had just been installed by openwhisk.  Seems to defeat the purpose of verifying that the new install is working properly.

ansible/tasks/installOpenwhiskCatalog.yml
```diff
@@ -0,0 +1,22 @@
...
+
+- stat: path="{{ catalog_location }}"
+  register: catalog_path
+ 
+- name: download the catalog repository to the catalog location if necessary
+  git:
+    repo: "{{ catalog_repo_url }}"
+    dest: "{{ catalog_location }}"
+    depth: 1
+  when: catalog_path.stat.isdir is not defined
```
jasonpet: Does the above line mean that it will only download the repo if my catalog_location directory does not already exist?

houshengbo: Yes. It the catalog directory already exists, we will skip the process of downloading the catalog repository.

ansible/postdeploy.yml
houshengbo: The ultimate goal is that the catalog installation scripts are removed out of openwhisk. However, the current phase is a transition state. We need to keep both of them tentatively, otherwise if we miss either of them, we cannot make sure all the actions are installed.

For the actions installed from the openwhisk-catalog, their test cases will be verified in the travis ci of the openwhisk-catalog, not in the openwhisk. The travis CI in openwhisk-catalog is responsible to guarantee the quality of the actions installed from the catalog.

ansible/group_vars/all
```diff
@@ -48,3 +48,40 @@ db:
...
+
+# The default value for catalog_source is empty, meaning that
+# openwhisk will not install the catalog from the catalog repositories.
+# If catalog_source is not set to "catalog-repos", the other variables
+# like namespace, api_host, auth_key_file and catalog_repos will not
+# be effective.
+catalog_source: ""
+
+# The default name space is /whisk.system. The namespace must begin with a slash "/".
+namespace: "/whisk.system"
```
rabbah: this is the namespace for installing the catalog? if so, perhaps it should be called catalog_namespace?

ansible/group_vars/all
```diff
@@ -48,3 +48,40 @@ db:
...
+
+# The api_host is used to specify the IP address of the host, where
+# openwhisk edge host is running.
+api_host: "172.17.0.1"
```
rabbah: @lehoanganh can this be retrieved in a way that is specific to the env?

ansible/group_vars/all
```diff
@@ -48,3 +48,40 @@ db:
...
+
+# The auth_key_file is used to determine where the authentication key is saved.
+# By default, we take the key from {{ openwhisk_home }}/config/keys/auth.whisk.system.
+auth_key_file: "{{ openwhisk_home }}/config/keys/auth.whisk.system"
```
rabbah: this is the key for installing the catalog, as such suggest we call it catalog_authkey.

jasonpet:
@houshengbo - we don't want to forget to update the ansible scripts for blue.  Those using blue (like me) will not get the new catalog installed when running ansible from blue.  I can create the equivalent blue ansible scripts for you if you want since you don't have a blue environment.

houshengbo:
@jasonpet it will be great that you can help me create the equivalent ansible scripts for blue. thanks.

ansible/group_vars/all
houshengbo: will change it.

ansible/group_vars/all
houshengbo: will change it.

ansible/group_vars/all
houshengbo: @rabbah I have found out the way to pick the value of the edge host in the yml file, so I can remove this variable.

houshengbo:
@rabbah @jasonpet @csantanapr @daisy-ycguo @lzbj 
I have figured out a way to directly pick up the value of edge host as the api host argument to install the catalog from other repos. The latest patch has conducted this change.

jasonpet:
Your changes look good to me.  I tried it all out and it worked for me.

ansible/group_vars/all
```diff
@@ -48,3 +48,37 @@ db:
...
+#
+# catalog_repos:
+#   ...
+#   <catalog-name>:
+#     url: <URL of repository>
+#     location: <local location to save the catalog>
+#
+catalog_repos:
+  openwhisk-catalog:
+    url: https://github.com/openwhisk/openwhisk-catalog.git
```
rabbah: we should include a hash to checkout a specific version of the repo.

ansible/group_vars/all
```diff
@@ -48,3 +48,37 @@ db:
...
+    # Set the local location as the same level as openwhisk home, but it can be changed.
+    location: "{{ openwhisk_home }}/../openwhisk-catalog"
```
rabbah: is this directory created if it doesn't exist?

houshengbo: The location will be checked the existence.
If it exists, use this existing folder for the catalog.
If not, it will be created and download the catalog into it.

ansible/group_vars/all
houshengbo: Do you mean we can expect a commit hash like "89915b4cc0810a9c9e67b3706a2850c58120cf75", so that we can checkout and use?

ansible/group_vars/all
```diff
@@ -48,3 +48,41 @@ db:
...
+# By default, we take the key from {{ openwhisk_home }}/ansible/files/auth.whisk.system.
+catalog_auth_key: "{{ openwhisk_home }}/ansible/files/auth.whisk.system"
+
+# The catalog_repos is used to specify all the catalog names and repository URLs,
+# so that openwhisk knows where to download the catalog and install them. The key
+# specifies the catalog name and the url saves the URL of the repository. The location
+# specifies the location to save the code of the catalog. The version specifies the hash
+# of the commit to be cloned. If it is HEAD, the latest commit will be selected.
+# Too add a new repository, please follow the template by adding:
+#
```
jasonpet: Should be To, not Too.

ansible/tasks/installOpenwhiskCatalog.yml
```diff
@@ -0,0 +1,31 @@
...
+  file:
+    path: "{{ catalog_location }}"
+    state: directory
+
+- name: download the catalog repository to the catalog location if necessary
+  git:
+    repo: "{{ catalog_repo_url }}"
+    dest: "{{ catalog_location }}"
+    version: "{{ version }}"
+
```
jasonpet: I see you changed this to always do the git even if you already have the repo in your workspace.  This solves the issue with Jenkins builds that do not clean up the workspace.  However, as a developer I do not want a git update happening on my catalog repo everytime I do a deploy.  Please create an ansible flag I can use to do an install catalog only (no git clone or update).

csantanapr:
@houshengbo take into account that I merged, this only to accelerate the migration to openwhisk-catalog. this PR doesn't affect current documentation or travis CI. So it doesn't have any functional impact yet.

In the future I would like to see complete PRs being submitted, and not work in progress.

PR #891, opened 2016-07-14
fix test case error message regex test

mdeuser:
Go CLI is no longer returning "error:" as prefix to the server error message


mdeuser:
@dubeejw - Ready for review

dubeejw:
👍

dubeejw:
PPG 846.

PR #893, opened 2016-07-14
Github webhook enhancements

### LINKED ISSUE #886: two github.webhook error handling bugs
github.webhook has two bugs with regards to error handling. here is the relevant source:

```
      if (error) {
        whisk.error();
      }
      console.log("Status code: " + response.statusCode);
      whisk.done({response: body});
```

issue 1: it should be `return whisk.error();`

issue 2: there is a missing clause, to handle non-200 responses from github. proposed code:

```
      if (error) {
        return whisk.error({response: "Internal Error"});
      }
      console.log("Status code: " + response.statusCode);
      if (response.statusCode >= 400) { /* or != 200? */
        return whisk.error({response: body});
      } else {
        return whisk.done({response: body});
      }
```

@bjustin-ibm raised a related UI issue: https://github.ibm.com/BlueMix-Fabric/Bluemix-Whisk-UI/issues/473

bjustin-ibm:
Resolves #886 #910 

Call `whisk.error()` with the actual error object received.
Call `whisk.error()` when the status code from Github is something other than `200`.

Allow the `repository` parameter to optionally be specified as `<organization>`/`<repoName>` so that it can be used to track repositories that aren't directly owned by the user.

### DIFF
--- a/catalog/github/webhook.js
+++ b/catalog/github/webhook.js
@@ -53,12 +53,29 @@ function main(params) {
 
     request(options, function(error, response, body) {
       if (error) {
-        whisk.error();
+        whisk.error({
+          response: response,
+          error: error,
+          body: body
+        });
+      } else {
+        console.log("Status code: " + response.statusCode);
+
+        if(response.statusCode >= 400) {
+          console.log("Response from Github: " + body);
+          whisk.error({
+            statusCode: response.statusCode,
+            response: body
+          });
+        } else {
+          whisk.done({response: body});
+        }
       }
-      console.log("Status code: " + response.statusCode);
-      whisk.done({response: body});
     });
+
     return whisk.async();
   }
+
+  // some lifecycleEvent for which there is nothing to do here
   return whisk.done();
 }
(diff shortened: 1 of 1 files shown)


bjustin-ibm:
@psuter @rabbah @starpit Would you mind having a look at this? I believe all the `returns` are in order: One when `lifeCycleEvent` is `CREATE` and one for all other cases.

catalog/github/webhook.js
```diff
@@ -53,12 +53,29 @@ function main(params) {
...
+          });
+        } else {
+          whisk.done({response: body});
+        }
       }
-      console.log("Status code: " + response.statusCode);
-      whisk.done({response: body});
     });
+
     return whisk.async();
```
psuter: I feel bad for asking, given how long I've left this open, but could we use a `Promise`? I think we're already operating internally with the mental model that `.async()` and friends are on their way to being marked deprecated?

psuter:
Merged as 9fceba2ee3f9a06e6466f6fe71f2d1235c4dbde1.

catalog/github/webhook.js
bjustin-ibm: I totally agree, but I would prefer to handle this in a separate Issue/PR. This PR already has two unrelated changes put together.

PR #901, opened 2016-07-15
WIP: another test PR

bjustin-ibm:
…envelope

OpenWhisk actions are expected to declare a global "main" function
However, this does not work well if you want to develop multiple actions in a single Xcode target, as you cannot re-declare a global main function.
This solution suggests that users working in Xcode may optionally declare their main function inside a struct, in order to solve the namespacing issue at compliation time.
Just before the .swift code is run, if a struct declaration is found, peel away the struct to reveal the delicious main function contained inside.

### DIFF
--- a/core/swiftAction/proxy.py
+++ b/core/swiftAction/proxy.py
@@ -43,8 +43,50 @@ def init():
     message = message.get("value", {})
 
     if "code" in message:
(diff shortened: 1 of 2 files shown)


bjustin-ibm:
Please ignore.

PR #906, opened 2016-07-16
BUG: Unsafe parameter passing in Python action

epatters:
The line is incorrect because JSON is not valid Python, e.g., `null` is not defined in Python.

### DIFF
--- a/core/pythonAction/pythonaction.py
+++ b/core/pythonAction/pythonaction.py
@@ -61,7 +61,7 @@ def run():
     result = None
     try:
         exec(flask.g, namespace)
-        exec("param = " + json.dumps(value), namespace)
+        namespace["param"] = value
         exec("fun = main(param)", namespace)
         result = namespace['fun']
     except Exception:


epatters:
@ioana-blue, could you help out with a unit test when you get a chance? I don't speak Scala.

It should be enough to pass `{"foo": null}` as inputs.

rabbah:
no serious Scala speak necessary: see PythonActionContainerTests.scala, copy/paste of "support valid flows" test in this case will get you most of the way there ;) if you want to try.

ioana-blue:
I'll take a look at this PR and try to merge it. I'll work separately on a test case. I don't think it's worth the overhead for Evan to learn how to write tests in our infrastructure, he's really productive with the tasks he needs to focus on. I'm removing the wip label.

rabbah:
Add the test on top of this commit? It's really just a copy/paste and small edit - we should not merge a fix without a test ;)

ioana-blue:
Ok, I'll try to add the test here.

ioana-blue:
Per @epatters suggestion, I created my own branch and added his one-line fix. I created a test, thanks @rabbah for suggesting the inspiration :)

ioana-blue:
Closing subsumed by #950

PR #913, opened 2016-07-18
Do Not Display Error Prefix by Default

dubeejw:
- Disable the default error prefix on error objects


dubeejw:
@mdeuser, can you review this real quick?

mdeuser:
@dubeejw - The "error:" prefix should not be displayed by the WskError Error() function.  The "error:" string prefix is controlled in main.go
```
        // If the err msg should be displayed to the console and it has not already been
        // displayed, display it now.
        var errMsgPrefix string = "error: "
        if (exitCode == 0) {
            errMsgPrefix = ""
        }
        if displayMsg && !msgDisplayed {
            fmt.Fprintf(os.Stderr, "%s%s\n", errMsgPrefix, err)
        }
```

Issue #887, opened 2016-07-14
go cli does not separate a long package (binding) name with a space from "private"


rabbah:
## Steps to reproduce the issue:

1.   `wsk package create aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`
2.   `wsk package list`

## Provide the expected results and outputs:

```
/{namespace}/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa private        
```

note the space before `private`

## Provide the actual results and outputs:

```
/{namespace}/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaprivate
```

note the lack of space before `private` - making for a subtle error when copy pasting package names into say the iOS SDK.

@mdeuser @dubeejw

mdeuser:
@rabbah - is there a need for the `private` column to be at a particular location (i.e. column 70 vs 71)?

rabbah:
no - the requirement i would say is that it shouldn't appear to be part of the string appearing in an adjacent column.

PR #914, opened 2016-07-18
Always Separate a Listed Name from its Published Type

dubeejw:
- Always add a space after an action, package, rule, or trigger name to separate it from its published type regardless if the name length


dubeejw:
Fixes #887.

dubeejw:
Ready for review.

Issue #917, opened 2016-07-18
Valid activation ids with addtional characters appended are treated as valid activation ids


mdeuser:
## Environment details:

* local deployment, native Windows, Bluemix

## Steps to reproduce the issue:

1.   `wsk action create test ..\..\..\catalog\samples\hello.js`
2.   `wsk action invoke test`
3.   Note the activation id
4.   `wsk activation get <activationid>`
5.   `wsk activation get <activationid>1234567890`  - appending 1234567890

## Provide the expected results and outputs:

Step 4 - activation id in this test run output is `a559d854f6f943f0930e4d512abdeaac`
```
ok: got activation a559d854f6f943f0930e4d512abdeaac
{
    "namespace": "mdeuser@us.ibm.com",
    "name": "test",
    "version": "0.0.1",
    "subject": "mdeuser@us.ibm.com",
    "activationId": "a559d854f6f943f0930e4d512abdeaac",
    "start": 1468876430035,
    "end": 1468876430514,
    "response": {
        "status": "success",
        "success": true
    },
    "logs": [
        "2016-07-18T21:13:50.507146226Z stdout: hello undefined!"
    ]
}
```
Step 5.
Expected an error stating that the activation id was invalid

## Provide the actual results and outputs:

Step 5 output
```
ok: got activation a559d854f6f943f0930e4d512abdeaac1234567890
{
    "namespace": "mdeuser@us.ibm.com",
    "name": "test",
    "version": "0.0.1",
    "subject": "mdeuser@us.ibm.com",
    "activationId": "a559d854f6f943f0930e4d512abdeaac",
    "start": 1468876430035,
    "end": 1468876430514,
    "response": {
        "status": "success",
        "success": true
    },
    "logs": [
        "2016-07-18T21:13:50.507146226Z stdout: hello undefined!"
    ]
}
```

## Additional information you deem important:
* The activation ID with the appended characters is sent in the GET HTTP request.  It appears that anything after the first 32 characters is truncated by the server.

rabbah:
Fixed by #918.

PR #918, opened 2016-07-18
Fixes issue #917

rabbah:
Do not accept activation id longer than 32 chars if '-' is missing. Add tests.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/ActivationId.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/ActivationId.scala
@@ -18,16 +18,19 @@ package whisk.core.entity
 
 import java.math.BigInteger
 
+import scala.language.postfixOps
+import scala.util.Failure
+import scala.util.Success
 import scala.util.Try
 
 import spray.json.DefaultJsonProtocol.StringJsonFormat
+import spray.json.DeserializationException
 import spray.json.JsObject
 import spray.json.JsString
 import spray.json.JsValue
 import spray.json.RootJsonFormat
 import spray.json.deserializationError
 import spray.json.pimpAny
-import scala.language.postfixOps
 
 /**
  * An activation id, is a unique id assigned to activations (invoke action or fire trigger).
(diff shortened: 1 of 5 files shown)


mdeuser:
This fix now results in `invalid activation id` being returned from the server.

I noticed other error message strings in the commit that include additional information that indicates more specifically about what is wrong with the id  - i.e.  `activation id is malformed (too long)`,  `activation id is malformed (too short)`.   Was this additional information expected to be sent in the server response?  Would be nice..

rabbah:
yes the message will come back as part of the API response. i can add an additional test for that.

mdeuser:
Should my local test have seen the `activation id is malformed (too long)` error message from the server?  My test run resulted in the `invalid activation id` message being received.

rabbah:
Addressed your comment @mdeuser in https://github.com/openwhisk/openwhisk/pull/918/commits/445e81b8bba387103e11edad277ac275e309c2cf.

rabbah:
IPG 350 approved.

rabbah:
Rebased and squashed.

common/scala/src/main/scala/whisk/core/entity/ActivationId.scala
```diff
@@ -88,13 +91,23 @@ protected[core] object ActivationId extends ArgNormalizer[ActivationId] {
         def read(value: JsValue) = Try {
             val JsString(s) = value
```
psuter: Can the id ever contain only numbers, be decoded into `JsNumber`, and have this throw a `MatchError`?

rabbah: The schema dictated the activation id is a string. Will add test though and harden for JsNumber. Thanks.

rabbah: Addressed in https://github.com/openwhisk/openwhisk/pull/918/commits/c057031eb1661743eb87539d424b97959544b102

common/scala/src/main/scala/whisk/core/entity/ActivationId.scala
```diff
@@ -89,25 +90,44 @@ protected[core] object ActivationId extends ArgNormalizer[ActivationId] {
...
             }
         } match {
             case Success(a)                                 => a
             case Failure(DeserializationException(t, _, _)) => deserializationError(t)
             case Failure(t)                                 => deserializationError("activation id is malformed")
         }
     }
+
+    private def bigIntToActivationId(n: BigInt): ActivationId = {
+        // print the bigint using base 10 than convert to base 16
```
markusthoemmes: *then

psuter:
Merged as 8812ced660f568c79ca8511cbae63eadcbb19f95.

PR #920, opened 2016-07-19
Documentation updates for Promises support

jasonpet:
Updating whisk.async() references in actions.md and reference.md and replacing with Promises.
Also updating the Node.js runtime documentation to include Node.js 6 and its npm packages (versions will now be included as well).  Documentation will also mention that Node.js 6 is now the default Node runtime and specify how the kind flag can be used to specify a specific Node runtime.

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -167,22 +167,27 @@ Rather than pass all the parameters to an action every time, you can bind certai
 
 ### Creating asynchronous actions
 
-JavaScript functions that continue execution in a callback function might need to return the activation result after the `main` function has returned. You can accomplish this using the `whisk.async()` and `whisk.done()` functions in your action.
+JavaScript functions that continue execution in a callback function might need to return the activation result after the `main` function has returned. You can accomplish this by returning a Promise in your action.
 
 1. Save the following content in a file called `asyncAction.js`.
 
   ```
-  function main() {
-      setTimeout(function() {
-          return whisk.done({done: true});
-      }, 20000);
-      return whisk.async();
-  }
+  function main(args) {
+       return new Promise(function(resolve, reject) {
+         setTimeout(function() {
+           resolve({ done: true });
+         }, 2000);
+      })
+   }
   ```
 
-  Notice that the `main` function returns immediately, and the `whisk.async()` return value indicates that this activation should continue running.
+  Notice that the `main` function returns a Promise, which indicates that the activation hasn't completed yet, but is expected to in the future.
 
-  The `setTimeout()` JavaScript function in this case waits for twenty seconds before calling the callback function, where the call to `whisk.done()` indicates that the activation is complete.
+  The `setTimeout()` JavaScript function in this case waits for twenty seconds before calling the callback function.  This represents the asynchronous code and goes inside the Promise's callback function. 
+  
+  The Promise's callback takes two arguments, resolve and reject, which are both functions.  The call to `resolve()` fulfills the Promise and indicates that the activation has completed normally.
+  
+  A call to `reject()` can be used to reject the Promise and signal that the activation has completed abnormally.
 
 2. Run the following commands to create the action and invoke it:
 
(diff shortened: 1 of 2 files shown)


docs/actions.md
```diff
@@ -167,22 +167,25 @@ Rather than pass all the parameters to an action every time, you can bind certai
...
+   }
   ```
 
-  Notice that the `main` function returns immediately, and the `whisk.async()` return value indicates that this activation should continue running.
+  Notice that the `main` function returns a Promise, which indicates that the activation hasn't completed yet, but is expected to in the future.
 
-  The `setTimeout()` JavaScript function in this case waits for twenty seconds before calling the callback function, where the call to `whisk.done()` indicates that the activation is complete.
+  The `setTimeout()` JavaScript function in this case waits for twenty seconds before calling the callback function.  This represents the asynchronous code and goes inside the Promise's callback function. 
+  
+  The Promise's callback takes two arguments, resolve and reject, which are both functions.  The call to `resolve()` fulfills the Promise and indicates that the activation is complete.
```
bjustin-ibm: For completeness, you should probably mention what `reject()` is for.

docs/actions.md
```diff
@@ -239,21 +242,21 @@ This example invokes a Yahoo Weather service to get the current conditions at a
...
-            var output = 'It is ' + temperature + ' degrees in ' + location + ' and ' + text;
-            whisk.done({msg: output});
+        return new Promise(function(resolve, reject) {
+            request.get(url, function(error, response, body) {
+                var condition = JSON.parse(body).query.results.channel.item.condition;
+                var text = condition.text;
+                var temperature = condition.temp;
+                var output = 'It is ' + temperature + ' degrees in ' + location + ' and ' + text;
+                resolve({msg: output});
+            });
```
bjustin-ibm: Perhaps add simple error handling to demonstrate `reject()` like
```javascript
if(error) {
    reject(error);
} else {
    // the code you currently show as the request callback code
}
```

docs/reference.md
```diff
@@ -133,7 +133,7 @@ It is common for JavaScript functions to continue execution in a callback functi
 A JavaScript action's activation is **synchronous** if the main function exits under one of the following conditions:
 
 - The main function exits without executing a ```return``` statement.
-- The main function exits by executing a ```return``` statement that returns any value *except* ```whisk.async()```.
+- The main function exits by executing a ```return``` statement that returns any value *except* a Promise or ```whisk.async()```.
```
bjustin-ibm: Seems like we shouldn't mention `whisk.async()` here.

docs/reference.md
```diff
@@ -224,48 +237,91 @@ The signature for `next` is `function(error, activation)`, where:
...
-- tmp
-- watson-developer-cloud
-- when
-- ws
-- xml2js
-- xmlhttprequest
-- yauzl
+### Node.js runtime environments
+
+JavaScript actions are executed by default in a Node.js version 6.2.0 environment.  The 6.2.0 environment will also be used for an action if the kind flag is explicitly specified with a value of 'nodejs:6' when creating/updating the action.
```
bjustin-ibm: Maybe format "kind" with back ticks like `--kind` so that it is clear this is a CLI argument.

jasonpet:
@bjustin-ibm thanks for the feedback.  I made some updates.

docs/actions.md
```diff
@@ -167,22 +167,27 @@ Rather than pass all the parameters to an action every time, you can bind certai
 
 ### Creating asynchronous actions
 
-JavaScript functions that continue execution in a callback function might need to return the activation result after the `main` function has returned. You can accomplish this using the `whisk.async()` and `whisk.done()` functions in your action.
+JavaScript functions that continue execution in a callback function might need to return the activation result after the `main` function has returned. You can accomplish this by returning a Promise in your action.
```
bjustin-ibm: @jasonpet When I merge this, I'm going to change this line to "JavaScript functions that run asynchronously may need to return the activation result after the `main` function has returned."  Let me know if you disagree.

bjustin-ibm:
Merged as 1710dd04f6ebc4b0d9c6bed07058e7ef7e295f36

Issue #921, opened 2016-07-19
Alarm triggers created with Go CLI never fire


bjustin-ibm:
The command issued (with both the Python and Go CLIs) was:
```$ wsk trigger create triggerName -f /whisk.system/alarms/alarm -p cron '0 */1 * * * 1-5'```

The key seems to be that the `authKey` is not passed along when creating the trigger.  For example, if I create a trigger with the Python CLI, the activation contains something like:

```json
{
    "logs": [
        "2016-07-19T19:46:12.351847028Z stdout: alarm:  { authKey: 'MY_ACTUAL_AUTH_KEY',",
        "2016-07-19T19:46:12.351887265Z stdout: trigger_payload: '',",
        "2016-07-19T19:46:12.35189886Z  stdout: cron: '0 */1 * * * 1-5',",
        "2016-07-19T19:46:12.351909522Z stdout: package_endpoint: '10.143.15.101:11001',",
        "2016-07-19T19:46:12.351919985Z stdout: lifecycleEvent: 'CREATE',",
        "2016-07-19T19:46:12.351930575Z stdout: triggerName: '/bjustin@us.ibm.com_dev/deleteme' }",
        "2016-07-19T19:46:12.510194615Z stdout: alarm: done http request",
        "2016-07-19T19:46:12.510568729Z stdout: { ok: 'your trigger was created successfully' }"
    ],
```

But creating the same trigger with the Go CLI results in something like:

```json
    "logs": [
        "2016-07-19T19:46:12.351847028Z stdout: alarm:  { authKey: '',",
        "2016-07-19T19:46:12.351887265Z stdout: trigger_payload: '',",
        "2016-07-19T19:46:12.35189886Z  stdout: cron: '0 */1 * * * 1-5',",
        "2016-07-19T19:46:12.351909522Z stdout: package_endpoint: '10.143.15.101:11001',",
        "2016-07-19T19:46:12.351919985Z stdout: lifecycleEvent: 'CREATE',",
        "2016-07-19T19:46:12.351930575Z stdout: triggerName: '/bjustin@us.ibm.com_dev/deleteme' }",
        "2016-07-19T19:46:12.510194615Z stdout: alarm: done http request",
        "2016-07-19T19:46:12.510568729Z stdout: { ok: 'your trigger was created successfully' }"
    ],
```

mdeuser:
@bjustin-ibm - Can you provide the exact command(s) you used?  Thanks!

[...]

PR #922, opened 2016-07-19
Include Auth Key from Properties when Creating and Deleting Triggers

dubeejw:
- If the auth key was not specified via the command line, it would not be sent when triggers were being created, or deleted

### DIFF
--- a/tests/src/system/basic/WskBasicTests.scala
+++ b/tests/src/system/basic/WskBasicTests.scala
@@ -466,6 +466,19 @@ class WskBasicTests
             wsk.trigger.list().stdout should include(name)
     }
 
+    it should "create a trigger using property file" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+            val name = "listTriggers"
+            val tmpProps = File.createTempFile("wskprops", ".tmp")
+            val env = Map("WSK_CONFIG_FILE" -> tmpProps.getAbsolutePath())
+            wsk.cli(Seq("property", "set", "--auth", wp.authKey) ++ wskprops.overrides, env = env)
+            assetHelper.withCleaner(wsk.trigger, name) {
+                (trigger, _) =>
+                    wsk.cli(Seq("-i", "trigger", "create", name), env = env)
+            }
+            tmpProps.delete()
+    }
+
     it should "create, and list a trigger with a long name" in withAssetCleaner(wskprops) {
         (wp, assetHelper) =>
             val name = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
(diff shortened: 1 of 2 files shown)


dubeejw:
Fixes https://github.com/openwhisk/openwhisk/issues/921

dubeejw:
Can you try this out, @bjustin-ibm?

dubeejw:
IPG 348.

bjustin-ibm:
> Can you try this out, @bjustin-ibm?

It works! Thanks.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -466,6 +466,19 @@ class WskBasicTests
             wsk.trigger.list().stdout should include(name)
     }
 
+    it should "create a trigger using property file" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+            val name = "listTriggers"
+            val tmpProps = File.createTempFile("wskprops", ".tmp")
+            val env = Map("WSK_CONFIG_FILE" -> tmpProps.getAbsolutePath())
+            wsk.cli(Seq("property", "set", "--auth", wp.authKey) ++ wskprops.overrides, env = env)
```
rabbah: `wskprops.overrides` are intended to precede the sub command whereas in the above they come after - is this intentions? note this is what happens in this case: `wsk property set --auth xxx --apihost yyy --auth xxx` since `auth` is one of the overrides.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -466,6 +466,19 @@ class WskBasicTests
...
+            assetHelper.withCleaner(wsk.trigger, name) {
+                (trigger, _) =>
+                    wsk.cli(Seq("-i", "trigger", "create", name), env = env)
```
rabbah: how is this test testing that the auth key is passed to the feed?

tests/src/system/basic/WskBasicTests.scala
rabbah: ignore invalid comment.

tests/src/system/basic/WskBasicTests.scala
markusthoemmes: This should assert an exit code of `0` to assure the trigger has been created properly.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -466,6 +466,19 @@ class WskBasicTests
             wsk.trigger.list().stdout should include(name)
     }
 
+    it should "create a trigger using property file" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+            val name = "listTriggers"
```
markusthoemmes: Choose a more speaking name for the trigger to make logs easier to parse in case of test failures.

markusthoemmes:
Isn't this test supposed to test if the feed action called on trigger creation gets the `authKey` parameter passed to it? This test doesn't assert that at all.

PR #927, opened 2016-07-20
API host property value will default to "" (empty string) if the API host value is not configured

### LINKED ISSUE #803: go CLI assumed api host is bluemix when no api host is set for a local deployment
## Environment details:

* mac + docker-machine

-  `redo couchdb initdb deploy` which runs these commands:
```
ant clean
ansible-playbook -i environments/mac couchdb.yml
ansible-playbook -i environments/mac initdb.yml
gradlew distDocker --parallel -PdockerHost=192.168.99.100:4243
ansible-playbook -i environments/mac openwhisk.yml
```
- `./bin/go-cli/wsk property get --apihost --apibuild --apibuildno`

```
whisk API host		192.168.99.100
whisk API build		2016-06-29T04:51:10Z
whisk API buildno	latest
```

```
whisk API host		 openwhisk.ng.bluemix.net
whisk API build		 2016-06-27T18:51:10Z
whisk API build number	 jenkins-mainWhisk-18130
```

The API host should be pointing to my local deployment. No API host would be better than pointing me to a external/public deployment. At least then the CLI should require me to do an initialization.

This is surely related to the loss of the property file and #783.

@mdeuser @dubeejw

mdeuser:
- Use a default api host value of "" (empty string) when API host value is not configured
- --apihost argument still overrides configured apihost value (including the default value)

Fixes #803

### DIFF
--- a/tests/src/common/TestUtils.java
+++ b/tests/src/common/TestUtils.java
@@ -333,6 +333,12 @@ public class TestUtils {
         return rr;
     }
 
+    /*
+     * WARNING: Consider using the WSK_CONFIG_FILE environment variable in
+     *          tests that will be manipulating the CLI property values.
+     *          Use this method only after determining WSK_CONFIG_FILE will
+     *          not work for the test case.
+     */
     public static File backupWskProps() throws IOException {
         String homedir = System.getProperty("user.home");
         Path wskpropsPath = FileSystems.getDefault().getPath(homedir, ".wskprops");
@@ -347,6 +353,12 @@ public class TestUtils {
         return tempfilePath.toFile();
     }
 
+    /*
+     * WARNING: Consider using the WSK_CONFIG_FILE environment variable in
+     *          tests that will be manipulating the CLI property values.
+     *          Use this method only after determining WSK_CONFIG_FILE will
+     *          not work for the test case.
+     */
     public static void restoreWskProps(File backupWskProps) throws IOException {
         String homedir = System.getProperty("user.home");
         Path wskpropsPath = FileSystems.getDefault().getPath(homedir, ".wskprops");
--- a/tests/src/common/TestUtils.java
+++ b/tests/src/common/TestUtils.java
@@ -21,6 +21,11 @@ import static org.junit.Assert.assertTrue;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.HashMap;
@@ -28,6 +33,7 @@ import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.Collections;
+import java.util.UUID;
 
 import com.google.gson.JsonObject;
 import com.google.gson.JsonArray;
@@ -325,5 +331,35 @@ public class TestUtils {
         RunResult rr = new RunResult(exitCode, new String(bl.getStdout()), new String(bl.getStderr()));
         rr.validateExitCode(expectedExitCode);
         return rr;
+    }
+
+    public static File backupWskProps() throws IOException {
+        String homedir = System.getProperty("user.home");
+        Path wskpropsPath = FileSystems.getDefault().getPath(homedir, ".wskprops");
+        String tempfileName =  UUID.randomUUID().toString()+".wskprops";
+        Path tempfilePath = FileSystems.getDefault().getPath(homedir, tempfileName);
+        try {
+            Files.copy(wskpropsPath, tempfilePath, StandardCopyOption.REPLACE_EXISTING );
+        }
+        catch (IOException e) {
+            throw e;
+        }
+        return tempfilePath.toFile();
+    }
+
+    public static void restoreWskProps(File backupWskProps) throws IOException {
+        String homedir = System.getProperty("user.home");
+        Path wskpropsPath = FileSystems.getDefault().getPath(homedir, ".wskprops");
+        try {
+            Files.copy(backupWskProps.toPath(), wskpropsPath, StandardCopyOption.REPLACE_EXISTING );
+        }
+        catch (IOException e) {
+            throw e;
+        }
+    }
+
+    public static File getWskPropsFile() {
+        String homedir = System.getProperty("user.home");
+        return new File(homedir + File.separator + ".wskprops");
     }
 }
--- a/tests/src/system/basic/WskBasicTests.scala
+++ b/tests/src/system/basic/WskBasicTests.scala
@@ -90,9 +90,38 @@ class WskBasicTests
         stdout should include regex ("""(?i)whisk API version\s+v1""")
     }
 
+    it should "validate default property values" in {
+        val tmpwskprops = File.createTempFile("wskprops", ".tmp")
+        val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
+        try {
+            wsk.cli(Seq("property", "unset", "--auth", "--apihost", "--apiversion", "--namespace"), env = env)
+            var stdout = wsk.cli(Seq("property", "get", "--auth"), env = env).stdout
+            stdout should include regex ("""(?i)whisk auth\s*$""")      // default = empty string
+            stdout = wsk.cli(Seq("property", "get", "--apihost"), env = env).stdout
+            stdout should include regex ("""(?i)whisk API host\s*$""")  // default = empty string
+            stdout = wsk.cli(Seq("property", "get", "--namespace"), env = env).stdout
+            stdout should include regex ("""(?i)whisk namespace\s*_$""")// default = _
+        }
+        finally {
+            tmpwskprops.delete()
+        }
+    }
+
     it should "show api build version" in {
-        val stdout = wsk.cli(wskprops.overrides ++ Seq("property", "get", "--apibuild")).stdout
-        stdout should include regex ("""(?i)whisk API build\s+201.*""")
+        //val wskpropsBackup = TestUtils.backupWskProps()
+        val tmpwskprops = File.createTempFile("wskprops", ".tmp")
+        val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
+        wsk.cli(wskprops.overrides ++ Seq("property", "set"), env = env )
+        val stdout = wsk.cli(Seq("property", "get", "--apibuild", "-i"), env = env).stdout
+        try {
+            stdout should include regex ("""(?i)whisk API build\s+201.*""")
+        }
+        finally {
+            tmpwskprops.delete()
+            //TestUtils.restoreWskProps(wskpropsBackup)
+            //wskpropsBackup.delete()
+        }
+
     }
 
     it should "fail to show api build when setting apihost to bogus value" in {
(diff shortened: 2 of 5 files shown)


mdeuser:
@dubeejw - Ready for review please.

mdeuser:
PG 353 passed

rabbah:
We should hold this until the related Pr is accepted so that you can enable as well the two tests in wsk basic tests that failed with go cli.

tools/go-cli/go-whisk-cli/commands/property.go
```diff
@@ -42,7 +42,7 @@ var Properties struct {
 }
 
 const DefaultAuth       string = ""
-const DefaultAPIHost    string = "openwhisk.ng.bluemix.net"
+const DefaultAPIHost    string = ""
```
dubeejw: Should there be a test that ensure that the API host is not set by default?

rabbah:
Also please check the docs for configuring the cli in openwhisk to make sure it specifies that setting the API host is now required.

It might be convenient we can add an ansible target to do this.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -88,13 +88,17 @@ class WskBasicTests
     }
 
     it should "show api build version" in {
+        wsk.cli(wskprops.overrides ++ Seq("property", "set"))
```
dubeejw: Does this set the properties like so: 
```wsk --apihost someHost --auth someAuth property set```

File: tools/go-cli/go-whisk-cli/commands/property.go (hunk 1 above)
Comment (mdeuser) [inline tools/go-cli/go-whisk-cli/commands/property.go]: Will add a test validating auth, apihost, and namespace default values

**Comment by rabbah** [comment]
I retitled #924 - I don't think this fix addresses that bug (the way to check is to try the new tests).

File: tests/src/system/basic/WskBasicTests.scala [hunk 3]
```diff
@@ -87,14 +87,27 @@ class WskBasicTests
...
+        var stdout = wsk.cli(Seq("property", "get", "--auth")).stdout
+        stdout should include regex ("""(?i)whisk auth\s*$""")      // default = empty string
+        stdout = wsk.cli(Seq("property", "get", "--apihost")).stdout
+        stdout should include regex ("""(?i)whisk API host\s*$""")  // default = empty string
+        stdout = wsk.cli(Seq("property", "get", "--namespace")).stdout
+        stdout should include regex ("""(?i)whisk namespace\s*_$""")// default = _
+    }
+
     it should "show api build version" in {
+        wsk.cli(wskprops.overrides ++ Seq("property", "set"))
```
dubeejw: Does this test rely on auth, apihost, apiversion, or namespace?

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -87,14 +87,27 @@ class WskBasicTests
...
         val stdout = wsk.cli(wskprops.overrides ++ Seq("property", "get", "--apibuild")).stdout
         stdout should include regex ("""(?i)whisk API build\s+201.*""")
+        wsk.cli(Seq("property", "unset", "--auth", "--apihost", "--apiversion", "--namespace"))
```
dubeejw: If the tests are ran locally won't this unset the values for each of these arguments for the local wsk instance?

tests/src/system/basic/WskBasicTests.scala
rabbah: The overrides should not change the set properties.

tests/src/system/basic/WskBasicTests.scala
dubeejw: Maybe we should read the old properties, modify them, then restore the old ones.

rabbah: If you set a temporary prop file then you don't have to worry about that.

dubeejw: @mdeuser, I think we can use the WSK_CONFIG_FILE environment variable to set a temp property file.

tests/src/system/basic/WskBasicTests.scala
mdeuser: The `show api build version` and `show api build number` tests only need valid `apihost` and `apiversion` values to be configured in the properties file.  Both those values are set via this wsk.cli() call.  Note that the default `--apiversion` value is actually the same value as the `--apiversion` value being set, so explicitly setting that value is not required.

tests/src/system/basic/WskBasicTests.scala
mdeuser: IMHO, each test should not assume there are any particular values in the .wskprops file.

On a fresh system, the file should be empty - but is set with the default values upon the first wsk CLI invocation.  Currently, most tests explicitly set the `--auth` and other properties on the command line so those test do no rely on what is in the .wskprops file.

If a test needs a particular value actually in the .wskprops file, the test should explicitly set that value and then run the test.  After the test, unsetting the value back to the default seems preferred.

rabbah: All of the Scala tests override the setting for the property file, API host, namespace and auth. So you are right they don't have nor should they have any dependencies on a local file. The Java tests aren't so sound in that there are setting they may not be overriding completely.

mdeuser: I'll add a couple helper methods in the TestUtils class to backup and restore the .wskprops file.  We can use these instead of the WSK_CONFIG_FILE env.  We may need to fix other Java tests that are more intrusive..

rabbah: WSK_CONFIG_FILE allows you to avoid saving .wskprops altogether - why bother saving it? The tests shouldn't touch the user's installed wskprops since it can be avoided.

mdeuser:
@dubee - Tests now using .wksprops backup/restore instead of WSK_CONFIG_FILE.  Ready for re-review.  I also refactored existing tests that were using that env variable for private property settings.

@rabbah - Regarding docs, the docs have not been updated to reflect the addition of the Go CLI (we're still rolling out our tests and fixing "blocking" issues).  I'll create an issue to make sure the docs are updated when we are ready to make this CLI officially available.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -87,48 +87,69 @@ class WskBasicTests
         stdout should include regex ("""(?i)whisk API version\s+v1""")
     }
 
+    it should "validate default property values" in {
+        val wskpropsBackup = TestUtils.backupWskProps()
```
rabbah: Sorry to keep repeating the same thing - but why do this vs 
`val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())` which is used in other tests that much around with saved properties?

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -87,48 +87,69 @@ class WskBasicTests
...
+        wsk.cli(Seq("property", "unset", "--auth", "--apihost", "--apiversion", "--namespace"))
+        var stdout = wsk.cli(Seq("property", "get", "--auth")).stdout
+        stdout should include regex ("""(?i)whisk auth\s*$""")      // default = empty string
+        stdout = wsk.cli(Seq("property", "get", "--apihost")).stdout
+        stdout should include regex ("""(?i)whisk API host\s*$""")  // default = empty string
+        stdout = wsk.cli(Seq("property", "get", "--namespace")).stdout
+        stdout should include regex ("""(?i)whisk namespace\s*_$""")// default = _
+        TestUtils.restoreWskProps(wskpropsBackup)
+        wskpropsBackup.delete()
```
rabbah: If the test fails anywhere before this, you will not restore the saved file. Another reason to avoid mucking around with it.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -87,48 +87,69 @@ class WskBasicTests
...
+    }
+
     it should "show api build version" in {
-        val stdout = wsk.cli(wskprops.overrides ++ Seq("property", "get", "--apibuild")).stdout
+        val wskpropsBackup = TestUtils.backupWskProps()
+        wsk.cli(wskprops.overrides ++ Seq("property", "set"))
```
rabbah: this should have no effect: `wsk property set` is a no op. This may be a point of divergence between the semantics of arguments and their position in python vs go. In python `wsk --auth xxx property set` is not the same as `wsk property set --auth xxx`. In go cli, from this test, I'm guess that they are the same.

tests/src/system/basic/WskBasicTests.scala
mdeuser: A while ago, Carlos and I had a discussion about the use of WSK_CONFIG_FILE.  The basic issue boiled down to opening the door to the possibility of a customer saving .wskprops files under source control (i.e. staging configuration, production configuration, etc) coupled with the possible future addition of the secure/insecure setting being added to this properties file.  We wanted to make sure deployments did not result in the CLI accidentally being configured for insecure operation.  So, the thought was that this env variable would eventually go away; so I was making preparations for that day.

tests/src/system/basic/WskBasicTests.scala
mdeuser: Good point.  Would a `finally` block handle this situation?

tests/src/system/basic/WskBasicTests.scala
mdeuser: Some of the CLI arguments, like --auth, --apihost, --apiversion and others, are global and defined against the wsk command (these flags each take an argument).  The `wsk property` command overrides these flags (these flags do not take arguments).  That's currently why a `wsk property get --apibuild --apihost x.y.z` command does not obtain the API build value from the x.y.z whisk host.  I'll need to look at this more to see if we can support both flags simultaneously.

tests/src/system/basic/WskBasicTests.scala
rabbah: I don't anticipate that we'll remove it or if we do, we will replace it with a command line argument that points to a wskprop file --- I added this support for the following reasons: I have several config files for multiple scenarios/deployments and testing. So I rapidly switch profiles this way.

I think it's a mistake for tests to muck around with user files - we should avoid it. The concern about committing wskprops into source control could be mitigated in other ways. For example, we can mandate that the prop file is always called `.wskprops` and the value of WSK_CONFIG_FILE becomes the path to that file. Then a gitignore file could do the rest.

mdeuser: Yeah, I agree that the tests should leave no footprints in the user wsk CLI configuration.  I like the cli argument that specifies the property file.  With the decision on external configuration file handling outstanding, I'll switch the tests back to using the existing env variable

mdeuser:
@rabbah - Ready for merge.  Travis finally passed.  PG 366 was successful (minus the trigger failure fix).

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -90,9 +90,38 @@ class WskBasicTests
         stdout should include regex ("""(?i)whisk API version\s+v1""")
     }
 
+    it should "validate default property values" in {
+        val tmpwskprops = File.createTempFile("wskprops", ".tmp")
+        val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
+        try {
+            wsk.cli(Seq("property", "unset", "--auth", "--apihost", "--apiversion", "--namespace"), env = env)
```
rabbah: since new file, dont need unset - in other words this is a noop - yes?

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -90,9 +90,38 @@ class WskBasicTests
...
+        }
+    }
+
     it should "show api build version" in {
-        val stdout = wsk.cli(wskprops.overrides ++ Seq("property", "get", "--apibuild")).stdout
-        stdout should include regex ("""(?i)whisk API build\s+201.*""")
+        //val wskpropsBackup = TestUtils.backupWskProps()
+        val tmpwskprops = File.createTempFile("wskprops", ".tmp")
+        val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
+        wsk.cli(wskprops.overrides ++ Seq("property", "set"), env = env )
```
rabbah: I don't like this because it's doing: `wsk -i --apihost xxx --apiversion yyy property set` which to me seems like a noop - rather, i would prefer if it was `wsk property set --apihost xxx --apiversion yyy`.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -90,9 +90,38 @@ class WskBasicTests
...
+            tmpwskprops.delete()
+        }
+    }
+
     it should "show api build version" in {
-        val stdout = wsk.cli(wskprops.overrides ++ Seq("property", "get", "--apibuild")).stdout
-        stdout should include regex ("""(?i)whisk API build\s+201.*""")
+        //val wskpropsBackup = TestUtils.backupWskProps()
+        val tmpwskprops = File.createTempFile("wskprops", ".tmp")
+        val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
```
rabbah: We could factor this into a new `withConfigFile` pattern to reduce the number of times we end up repeating create file, set env map, and delete file on error. This reduces the changes of forgetting a finally also.

mdeuser: @rabbah - I opened #944 for this test case refactoring we'll need to dig a bit into scala to do that; so it won't be super quick; but our team needs to learn it though.  OK to merge this PR now with the un-refactored test code?

tests/src/common/TestUtils.java
```diff
@@ -326,4 +332,34 @@ public static RunResult runCmd(int expectedExitCode, int timeoutMilli, File dir,
         rr.validateExitCode(expectedExitCode);
         return rr;
     }
+
```
rabbah: should there be  comment or warning to these methods since they are not used - they encourage doing things by copy/restore of the prop file which i think we want to discourage.

tests/src/system/basic/WskBasicTests.scala
rabbah: an example of test for #962.

tests/src/system/basic/WskBasicTests.scala
rabbah: ok to rewrite this as: 

` +        wsk.cli(Seq("property", "set") ++ wskprops.overrides, env = env ) `

i understand that the apihost (part of override) are global and can be used anywhere unlike python's stricter arg parsing but i find an empty property set confusing.

mdeuser: Just wondering.. is the order of the sequence concatenation important to scala (i.e. `Seq("property", "set") ++ wskprops.overrides` vs `wskprops.overrides ++ Seq("property", "set")` ) ?

In this particular test, we're using wskprops (not tmpwksprops), so I believe the overrides set should not be empty.

tests/src/common/TestUtils.java
mdeuser: i'll add that comment.  i wanted to keep these debugged utility methods hanging around, just in case they are useful at some later point.

tests/src/system/basic/WskBasicTests.scala
rabbah: yes - order matters for a sequence. I agree putting the overrides later might look confusing so perhaps should just set the properties explicitly for apihost (which is what you get out of overrides).

rabbah:
Decided that given the semantics of overrides in the go CLI not to make the change implied above.

Rebased and merged as https://github.com/openwhisk/openwhisk/commit/3c95e7d3857dbf889b4fe34ae78c45654af0e31c.

PR #939, opened 2016-07-25
The test should be moved into another package

fxulusoy:
"show api build using http apihost" test in WskBasicTests is failing.


rabbah:
This test was specifically added to test http. Hence we should just move it to a different package.

fxulusoy:
@rabbah if it is not okay to ignore for now or where can I move it?

rabbah:
suggest establishing new package `whisk.cli` and move it there - if looking for expedient solution then just ignore and we'll move it later.

fxulusoy:
@rabbah Let's ignore for now.

PR #943, opened 2016-07-25
Ignore experimental Swift3 tests

fxulusoy:
Some of these tests are failed sometimes and breaks the CI pipeline. Ignoring the tests was agreed.


paulcastro:
We should also ignore the test in Swift3ActionContainerTests in tests/src/actionContainers

PR #945, opened 2016-07-25
Remove reference to whisk.done() in synchronous action

jasonpet:
There is a reference to whisk.done(), which is used to signal normal completion of an asynchronous action, in an example of a synchronous action.  Multiple users have mentioned that this is confusing.


bjustin-ibm:
@jasonpet Now I wonder whether this example should be merged with the one above it. I'm not sure I see the value in having these two separate synchronous actions.

jasonpet:
@bjustin-ibm - I merged the two examples.

bjustin-ibm:
Merged as e4770a330ceac0ab0272ec975a366ec8cc92a097

Issue #493, opened 2016-05-23
python code with C-style comments gives "action did not return a JSON object" error


starpit:
expected: "parse error"
instead received: "action did not return a JSON object" as the error message

ioana-blue:
I'm guessing the compilation fails and, as a result, when the action is run, an error is thrown, not the expected json is returned and the error above is sent to the user. Not sure how to fix this. Perhaps compile the code ahead of time and check it compiled properly and if that's not the case try to send a different error as part of initialization?

rabbah:
can't you wrap with a try and report error as response?

starpit:
i think this could work? if i recall, the "ugly whisk stack trace" part of the error response gave an indication of the problem, it's just the top level error message whisk returns is deceptive.

ioana-blue:
there is probably a try already. it might work, but not sure how to differentiate between compilation error or other type of error. if we just want to capture "error with code", that's probably easier.

starpit:
hmm, now i'm getting a different response. i'll continue to try to replicate my original error message. this one is pretty close, though it'd be nice to find a way to avoid displaying internal stack traces to the user.

```
{
  "error": "Initialization has failed due to: SyntaxError: Unexpected token ILLEGAL\n    at new NodeActionRunner (/nodejsAction/runner.js:32:21)\n    at doInit (/nodejsAction/src/service.js:129:22)\n    at initCode (/nodejsAction/src/service.js:62:39)\n    at safeEndpoint (/nodejsAction/app.js:57:13)\n    at Layer.handle [as handle_request] (/nodejsAction/node_modules/express/lib/router/layer.js:82:5)\n    at next (/nodejsAction/node_modules/express/lib/router/route.js:110:13)\n    at Route.dispatch (/nodejsAction/node_modules/express/lib/router/route.js:91:3)\n    at Layer.handle [as handle_request] (/nodejsAction/node_modules/express/lib/router/layer.js:82:5)\n    at /nodejsAction/node_modules/express/lib/router/index.js:267:22\n    at Function.proto.process_params (/nodejsAction/node_modules/express/lib/router/index.js:321:12)"
}
```

starpit:
here is the example code i used for my previous comment:
```
/**
  *
  * main() will be invoked when you Run This Action.
  * 
  * @param Whisk actions accept a single parameter,
  *        which must be a JSON object.
  *
  * @return The return value must also be JSON.
  *         It will be the output of this action.
  *
  */
import sys

[...]

PR #950, opened 2016-07-25
Fix bug in passing parameters to the python action

ioana-blue:
Fixed bug and added test to chec a couple of jsons, one with null

Subsumes #906

### DIFF
--- a/core/pythonAction/pythonaction.py
+++ b/core/pythonAction/pythonaction.py
@@ -35,8 +35,16 @@ def init():
 
     message = payload.get("value", {})
     if "code" in message:
-        # store the code
-        flask.g = message["code"]
+        # store the compiled code
+        try:
+            flask.g = compile(message["code"], filename = '<string>', mode = 'exec')
+        except Exception:
+            flask.g = None
+            traceback.print_exc(file = sys.stderr)
+            sys.stderr.flush()
+            response = flask.jsonify({"error": "the action failed to compile. See logs for details." })
+            response.status_code = 502
+            return response
         return ('OK', 200)
     else:
         flask.abort(403)
@@ -56,12 +64,18 @@ def run():
     if not isinstance(value, dict):
         flask.abort(403)
 
+    if flask.g is None:
+        # no code to execute
+        response = flask.jsonify({"error": "No code to execute (init failed). See logs for details." })
+        response.status_code = 502
+        return response
+
     # initialize the namespace for the execution
     namespace = {}
     result = None
     try:
+        namespace['param'] = value
         exec(flask.g, namespace)
-        exec("param = " + json.dumps(value), namespace)
         exec("fun = main(param)", namespace)
         result = namespace['fun']
     except Exception:
@@ -73,7 +87,7 @@ def run():
         response.status_code = 200
         return response
     else:
-        response = flask.jsonify({ "error": "the action did not return a dictionary", "action_output": result })
+        response = flask.jsonify({"error": "The action did not return a dictionary", "action_output": result })
         response.status_code = 502
         return response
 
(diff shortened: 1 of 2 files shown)


ioana-blue:
This is ready for review & merge.

rabbah:
I'm not sure that we need a complex test for null - here's a suggested alternative:

https://github.com/openwhisk/openwhisk/compare/master...rabbah:pr-906?expand=1

core/pythonAction/pythonaction.py
```diff
@@ -61,7 +61,7 @@ def run():
     result = None
     try:
         exec(flask.g, namespace)
```
rabbah: why isn't this part of init: `exec(flask.g, namespace)` - wouldn't that save the eval overhead on every run?

ioana-blue: I think this evolved. At some point there was an eval part of init. I'll fix it part of this PR.

ioana-blue:
One of your links above doesn't work for me, but I get what you're saying. Fixing.

core/pythonAction/pythonaction.py
ioana-blue: Actually, I think this statement in particular is not needed (subsumed by the following exec), which is really a bug and not only an optimization issue. Good catch!

ioana-blue:
@rr I addressed your comments and fixes are in place. Please take a careful look since I fixed another issue with providing feedback during init for when compilation fails (I think there was an issue opened for this, maybe from Nick). moving this to review.

tests/src/actionContainers/PythonActionContainerTests.scala
```diff
@@ -73,6 +72,30 @@ class PythonActionContainerTests extends FlatSpec
...
+            val (initCode, _) = c.init(initPayload(code))
+            initCode should be(200)
+
+            val argss = List(
+                JsObject("user" -> JsNull),
+                JsObject("user" -> JsString("Momo")))
+
+            for (args <- argss) {
+                val (runCode, out) = c.run(runPayload(args))
+                println(out)
```
rabbah: remove println.

core/pythonAction/pythonaction.py
```diff
@@ -35,8 +35,16 @@ def init():
...
-        # store the code
-        flask.g = message["code"]
+        # store the compiled code
+        try:
+            flask.g = compile(message["code"], filename = '<string>', mode = 'exec')
+        except Exception:
+            flask.g = None
+            traceback.print_exc(file = sys.stderr)
+            sys.stderr.flush()
+            response = flask.jsonify({"error": "the action failed to compile. See logs for details." })
```
rabbah: the -> The (since you capitalized elsewhere).

core/pythonAction/pythonaction.py
```diff
@@ -56,12 +64,18 @@ def run():
     if not isinstance(value, dict):
         flask.abort(403)
 
+    if flask.g is None:
+        # no code to execute
+        response = flask.jsonify({"error": "No code to execute (init failed). See logs for details." })
```
rabbah: init -> compilation? since you called it compilation above.

rabbah: this should never happen btw (ie if you're called when init failed, there's an internal system error).

rabbah:
Can you link to the issue mentioned above (in the commit and this pr).
PG?

rabbah:
Not sure if you meant this one: https://github.com/openwhisk/openwhisk/issues/493...?

core/pythonAction/pythonaction.py
```diff
@@ -73,7 +87,7 @@ def run():
         response.status_code = 200
         return response
     else:
-        response = flask.jsonify({ "error": "the action did not return a dictionary", "action_output": result })
+        response = flask.jsonify({"error": "The action did not return a dictionary", "action_output": result })
```
rabbah: missing period.

Issue #946, opened 2016-07-25
the sampleInput of util.split is not valid JSON


starpit:
the sampleInput for split is not JSON it is a string. compare these two:

sort has proper JSON:
![screen shot 2016-07-25 at 2 58 21 pm](https://cloud.githubusercontent.com/assets/4741620/17113378/8557fec6-5278-11e6-98d1-e29ad36136c7.png)

and split does not:
![screen shot 2016-07-25 at 2 58 18 pm](https://cloud.githubusercontent.com/assets/4741620/17113380/89648fa2-5278-11e6-9542-0f36c046a27b.png)

csantanapr:
@jasonpet can you take a look?

jasonpet:
sure

starpit:
there is a missing close quote in the description annotation

starpit:
see this one, from way back: https://github.com/openwhisk/openwhisk/issues/137

the backend is accepting invalid json, and, worse, stringifying it.

csantanapr:
@dubeejw why you close this issue?

jasonpet:
@csantanapr - he closed it because he merged the fix in already.

csantanapr:
@jasonpet Thanks for the info. @dubeejw When we close issues because the PR is merged, It's good to add a message like "closing merged with commit xxxxxxx"  this way give reason but also can find the commit that fixed the issue, then we can track when it makes it to production.

csantanapr:
@jasonpet What's the PR that fixed this issue? I was not able to see a mentioned reference here.

Yes I know I know, if I look very hard I eventually will find it 😃

jasonpet:
PR - https://github.com/openwhisk/openwhisk/pull/951

PR #951, opened 2016-07-25
Invalid syntax in annotation for split action

jasonpet:
Invalid syntax in annotation is causing wsk cli to create the input as a String instead of JSON.


csantanapr:
Fixes #946

Issue #952, opened 2016-07-25
go CLI does not have the expected type for annotation values


jasonpet:
The go CLI is setting the value type as a String even when the expected type should be JSON.

jasonpet:
Example:
 wsk -i action get /whisk.system/samples/wordCount

  "key": "sampleInput",
   "value": "{ \"payload\": \"Five fuzzy felines\"}"

Issue #137, opened 2016-03-29
wsk "-a" accepts invalid JSON and turns it into a string annotation


starpit:
we just merged in a breaking change to installWatson.sh:

```
install 'watson/languageId.js'      watson/languageId \
    -a description 'Identify language' \
    -a parameters '[ {"name":"username", "required":true, "bindTime":true}, {"name":"password", "required":true, "type":"password", "bindTime":t\
rue}, {"name":"payload", "required":true} ]' \
    -a sampleInput '{"payload": "Bonjour", "username":"XXX", "password":"XXX"}' \
    -a sampleOutput '{"language": "French", confidence: 1}'
```

the sampleOutput annotation is missing the "quotes" around confidence. wsk turns this into a string annotation, which was not what the author intended.

Option 1: change the meaning of -a to mean "strict JSON", i.e. the default will be to fail if the annotation is not parseable as JSON; add a --nonstrict top-level switch, or an -na annotation-level switch to allow the current behavior

Option 2: add -ja which means fail if the annotation is not parseable as JSON; this option leaves the default with our current relaxed approach to parsing the annotations

sjfink:
In the name of simplicity, I'd vote for Option 1 ... simply fail if it's not valid JSON.   I wouldn't even add a --nonstrict flag.

Other opinions?

jasonpet:
I found that packages are still being installed using python cli.  When I switch to use go cli the input is always a String instead of JSON (even for valid annotations with no syntax errors)

starpit:
@jasonpet we should open a go cli defect, i think?

jasonpet:
yeah, we definitely cannot merge this yet: https://github.com/openwhisk/openwhisk/pull/883.
It will break the UI.  I let James Dubee know already.
@dubeejw - do we have a defect for this already?  If not I can create one.

rabbah:
No we don't yet.

jasonpet:
Opened #952 for issue with annotation values using go cli.

mdeuser:
For clarity regarding the format of the key string and value string, let's consider both of these valuea for the command `wsk action create test test.js -a k1 v1`

1. k1, the key.  This will eventually be folded into JSON prior to the REST endpoint call.  Currently, the user is not expected to provide k1 in valid JSON format, and the CLI will handle the conversion of the k1 string into a valid JSON string.  Conversion involves escaping embedded `/` and `"` characters, wrapping the entire value in double quotes.
[...]

PR #953, opened 2016-07-25
Annotations Must be Parsed as Raw JSON

dubeejw:
- Update Go CLI to handle annotations as raw JSON instead of as strings
  - Use the same path that parameter parsing does
- Refactor printing of package summary
- Add tests to validate annotation, and parameter parsing
- Add negative tests to ensure arguments are passed to annotations, and parameters

### DIFF
--- a/tests/src/system/basic/WskBasicTests.scala
+++ b/tests/src/system/basic/WskBasicTests.scala
@@ -23,6 +23,7 @@ import org.apache.commons.io.FileUtils
 import org.junit.runner.RunWith
 import org.scalatest.junit.JUnitRunner
 
+import spray.json._
 import spray.json.DefaultJsonProtocol._
 import spray.json.pimpAny
 import spray.json.pimpString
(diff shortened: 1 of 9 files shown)


dubeejw:
Fixes https://github.com/openwhisk/openwhisk/issues/952

dubeejw:
PPG 896.

rabbah:
we should add basic tests which create annotations (and params) of various types and makes sure that the get result matches intended type.

dubeejw:
Was thinking the same thing. We need annotations tests for actions, packages, and triggers.

rabbah:
Is the annotation parsing in the go cli using the same path as parameters (no new coded needed? I will read the code, of course, and could probably answer my own question later).

dubeejw:
The annotation parsing is using the same path as the parameter passing now. Parameter passing should be working well. I had to rewrite that awhile ago to get all the CLI tests to pass.

dubeejw:
PPG 904.

dubeejw:
@mdeuser, ready for review.

rabbah:
Tests look good on first pass - one tip: rather than repeating the annotation/parameter json in several test, you could lift them into a function - same for the checks/regex matches. On the latter, there might be a helper which will turn the wsk cli response into a json object which will then allow you to just do JsValue equality (much simpler).

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -291,6 +292,72 @@ class WskBasicTests
...
+            val stderr = wsk.cli(wskprops.overrides ++ Seq("package", "create", "packageName", "--auth", wp.authKey,
+                "-p"), expectedExitCode = ERROR_EXIT).stderr
+            stderr should include("Parameter arguments must be a key value pair")
+    }
+
+    it should "not create a package when an annotation contains invalid JSON" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+            val stderr = wsk.cli(wskprops.overrides ++ Seq("package", "create", "packageName", "--auth", wp.authKey,
+                "-a", "key", "{\"invalidObj\", }"), expectedExitCode = ERROR_EXIT).stderr
+            stderr should include("Error encoding request body")
```
mdeuser: Customer facing error message is a bit cryptic.  Is it possible to have an error message that is more clear (i.e. "The annotation value is not valid JSON: <json encoding error details>")?

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -291,6 +292,72 @@ class WskBasicTests
...
+    }
+
+    it should "not create a package when a parameter contains invalid JSON" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+            val stderr = wsk.cli(wskprops.overrides ++ Seq("package", "create", "packageName", "--auth", wp.authKey,
+                "-p", "key", "{\"invalidObj\", }"), expectedExitCode = ERROR_EXIT).stderr
+            stderr should include("Error encoding request body")
```
mdeuser: Same comment for -p values that are invalid JSON - more specific error message.

mdeuser:
Looking at the following code snippet from `convertArgumentInputToJSON()`:
```
        if ((len(args[i + 1]) > 0) && (isValidJSON(args[i + 1]) || (args[i + 1][0] == '[' && args[i + 1][len(args[i + 1]) - 1] == ']') ||
        (args[i + 1][0] == '"' && args[i + 1][len(args[i + 1]) - 1] == '"'))) {
```
it might be more clear to use regex patterns for the [xxx] and "xxx" tests

Minor restructure comment:  combine the above condition tests with the Atoi() condition test to reduce code (i.e. drop additional `else`, remove the extra `res = res +` statement)

dubeejw:
@mdeuser, in regard to your comment about the invalid JSON error message, all invalid JSON will be assumed to be a string now. That means there will no longer be an error message for invalid JSON.

dubeejw:
P1P 70.

dubeejw:
@mdeuser, any more comments on this one?

mdeuser:
@dubeejw - Oh.. I was waiting for the final code updates.  Is that all complete now?

tools/go-cli/go-whisk-cli/commands/util.go
```diff
@@ -99,177 +98,75 @@ func parseQualifiedName(name string) (qName qualifiedName, err error) {
...
+            "getJSONKeyValueFromArguments(): Assuming string for input as it is not valid JSON: %s\n", value)
+        value = fmt.Sprintf("\"%s\"", getEscapedString(value))
     }
 
-    annotations = whisk.Annotations(parsedArgs)
+    return key, value
+}
 
-    return annotations, nil
+func getEscapedString(value string) (string) {
```
mdeuser: @dubeejw - From the JSON specification (http://www.json.org/), the only allowable times a '\' may exist in a string value is when that '\' is followed by a:  ", \, /, b, f, n, r, t, or u character - for a total of 9 special characters.   If this function is to correctly handle these special characters in a JSON string, then \/ and \u should also be handled.

That said, while I may not completely understand the context for this '\' character handling, why don't we just escape all appearances of the '\' character?  Does whatever is consuming the '\' character and forcing the '\' to be escaped in these explicit conditions also need all other appearances of the '\' to be escaped?

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -583,6 +730,86 @@ class WskBasicTests
...
+            }
+
+            val stdout = wsk.trigger.get(name).stdout
+            assert(stdout.startsWith(s"ok: got trigger $name\n"))
+
+            val resJSON = stdout.substring(stdout.indexOf("\n") + 1).parseJson.asJsObject
+            assert(resJSON.fields("parameters") == getEscapedJSONTestArgOutput)
+    }
+
+    it should "create a trigger with the proper annotation escapes" in withAssetCleaner(wskprops) {
```
mdeuser: also try a parameter or annotation string value containing non-ascii characters, such as "日本語"

tools/go-cli/go-whisk-cli/commands/util.go
dubeejw: This function is escaping actual whitespace characters, backslashes, and qoutes to have parity with the Python CLI.

rabbah: It may be worth saying that where it makes sense to change things/add or remove features to do things better than the Python cli, we should.

mdeuser: Consider the command `wsk action create test t.js -p key value`.  Are    While the `key` and `value` strings are eventually embedded inside JSON, is there a requirement that these values be JSON compliant on the CLI invocation by the user?

dubeejw: If the user desires to pass JSON as an value, the input would have to comply to JSON standards. In the case where the user wants to just pass a string, the input does not need to conform to JSON standards.

rabbah: https://github.com/openwhisk/openwhisk/issues/137.

mdeuser:
I'm taking over this PR while James is out.

bjustin-ibm:
@mdeuser Has this PR been superseded by #1007? If so, please close this one.

mdeuser:
Yes, PR #1007 obviates the need for this PR.  Closing

PR #954, opened 2016-07-26
Make logsize settable for each action

markusthoemmes:
Adds the logsize as a limit that is settable per action and thus included in the quota for each action.

@mdeuser please review the go-cli parts

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -406,7 +406,7 @@ OpenWhisk has a few system limits, including how much memory an action uses and 
 | ----- | ----------- | ------------ | -----| ------- |
 | timeout | a container is not allowed to run longer than N milliseconds | per action |  milliseconds | 60000 |
 | memory | a container is not allowed to allocate more than N MB of memory | per action | MB | 256 |
-| logs | a container is not allowed to write mor than N MB to stdout | per action | MB | 10 |
+| logs | a container is not allowed to write more than N MB to stdout | per action | MB | 10 |
 | concurrent | it's not allowed to have more than N concurrent activations per namespace | per namespace | number | 100 |
 | minuteRate | a user cannot invoke more than this many actions per minute | per user | number | 120 |
 | hourRate | a user cannot invoke more than this many actions per hour | per user | number | 3600 |
@@ -422,9 +422,9 @@ OpenWhisk has a few system limits, including how much memory an action uses and 
 * A container cannot have more memory allocated than the limit.
 
 ### Per action logs (MB) (Default: 10MB)
-* The log limit M is in the range from [0MB..10MB] and is set per action in MB.
-* A user can change the limit when creating the action.
-* If a container writes more logs than the limit, the activation will contain the logs until that limit and truncated afterwards. The last logline will contain warning.
+* The log limit N is in the range [0MB..10MB] and is set per action.
+* A user can change the limit when creating or updating the action.
+* Logs that exceed the set limit are truncated and a warning is added as the last output of the activation to indicate that the activation exceeded the set log limit.
 
 ### Per action artifact (MB) (Fixed: 1MB)
 * The maximum code size for the action is 1MB.
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -406,6 +406,7 @@ OpenWhisk has a few system limits, including how much memory an action uses and 
 | ----- | ----------- | ------------ | -----| ------- |
 | timeout | a container is not allowed to run longer than N milliseconds | per action |  milliseconds | 60000 |
 | memory | a container is not allowed to allocate more than N MB of memory | per action | MB | 256 |
+| logs | a container is not allowed to write mor than N MB to stdout | per action | MB | 10 |
 | concurrent | it's not allowed to have more than N concurrent activations per namespace | per namespace | number | 100 |
 | minuteRate | a user cannot invoke more than this many actions per minute | per user | number | 120 |
 | hourRate | a user cannot invoke more than this many actions per hour | per user | number | 3600 |
@@ -419,6 +420,11 @@ OpenWhisk has a few system limits, including how much memory an action uses and 
 * The memory limit M is in the range from [128MB..512MB] and is set per action in MB.
 * A user can change the limit when creating the action.
 * A container cannot have more memory allocated than the limit.
+
+### Per action logs (MB) (Default: 10MB)
+* The log limit M is in the range from [0MB..10MB] and is set per action in MB.
+* A user can change the limit when creating the action.
+* If a container writes more logs than the limit, the activation will contain the logs until that limit and truncated afterwards. The last logline will contain warning.
 
 ### Per action artifact (MB) (Fixed: 1MB)
 * The maximum code size for the action is 1MB.
--- a/tests/src/common/Wsk.scala
+++ b/tests/src/common/Wsk.scala
@@ -41,6 +41,7 @@ import spray.json.JsValue
 import spray.json.pimpString
 import whisk.utils.retry
 import java.time.Instant
+import whisk.core.entity.ByteSize
 
 /**
  * Provide Scala bindings for the whisk CLI.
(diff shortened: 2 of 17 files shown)


markusthoemmes:
PPG 897

tools/go-cli/go-whisk/whisk/shared.go
```diff
@@ -38,4 +38,5 @@ type Parameters *json.RawMessage
 type Limits struct {
     Timeout int `json:"timeout,omitempty"`
     Memory  int `json:"memory,omitempty"`
+    Logsize int `json:"logs"`
```
mdeuser: IMHO, the JSON label "logsize" would better fit the purpose.
Add "omitempty" so that a `wsk action update` won't accidentally reset this value to the 0

markusthoemmes: `omitempty` is not feasible though, it will turn a `0` (perfectly valid) to an empty object. Have a solution for that?

mdeuser:
Are there any doc updates providing more details about this feature?
- What happens when the log limit is reached (i.e. log file wraps, no logging, log file is reset, etc)?
- Default log size when none is set?

rabbah:
This is a new limit we need do document - @csantana FYI.

tools/go-cli/go-whisk/whisk/shared.go
mdeuser: Ah.. The default values have been changed from 0 to -1.  What will the server do with -1 values on both `wsk action create` and `wsk action update`?

markusthoemmes: @ChrBi and me played a bit with it. They are not sent at all (as they weren't with `0` set). Note that the server will decline `0` for memory and timeout limits.

mdeuser: The limits struct is instantiated when any of the memory, timeout or logsize arg is used.  Any arg not set will default to 0.  When the `omitempty` is removed, the default 0 value will be sent.  Can you try this test.
```
touch a.js
wsk action create tempaction a.js --logsize 10
wsk action get tempaction
wsk action update tempaction --memory 512  // leaving logsize alone
wsk action get tempaction    // Did the logsize setting change?
```

markusthoemmes: Cheers for pointing that out. We haven't tried that. The unfortunate is: `0` is overloaded as the "empty" value for `int`. Investigating...

docs/reference.md
```diff
@@ -420,6 +421,11 @@ OpenWhisk has a few system limits, including how much memory an action uses and
 * A user can change the limit when creating the action.
 * A container cannot have more memory allocated than the limit.
 
+### Per action logs (MB) (Default: 10MB)
+* The log limit M is in the range from [0MB..10MB] and is set per action in MB.
+* A user can change the limit when creating the action.
+* If a container writes more logs than the limit, the activation will contain the logs until that limit and truncated afterwards. The last logline will contain warning.
```
mdeuser: - Just confirming.  "truncated afterward" means that no additional log entries will be added to the log file ?
- What happens when the size is set to 0MB?

markusthoemmes: 1. Right, my wording might be off, suggestions?
2. Setting it to 0 will either result in an empty log (nothing written) or in just the warning of truncation

rabbah: While updating the limits, should also add the bits for
exec/annotations/parameters.

docs/reference.md
```diff
@@ -406,6 +406,7 @@ OpenWhisk has a few system limits, including how much memory an action uses and
 | ----- | ----------- | ------------ | -----| ------- |
 | timeout | a container is not allowed to run longer than N milliseconds | per action |  milliseconds | 60000 |
 | memory | a container is not allowed to allocate more than N MB of memory | per action | MB | 256 |
+| logs | a container is not allowed to write mor than N MB to stdout | per action | MB | 10 |
```
rabbah: mor -> more

docs/reference.md
```diff
@@ -420,6 +421,11 @@ OpenWhisk has a few system limits, including how much memory an action uses and
 * A user can change the limit when creating the action.
 * A container cannot have more memory allocated than the limit.
 
+### Per action logs (MB) (Default: 10MB)
+* The log limit M is in the range from [0MB..10MB] and is set per action in MB.
```
rabbah: using M here, N above. suggest this:
 +* The log limit N is in the range [0MB..10MB] and is set per action.

docs/reference.md
```diff
@@ -420,6 +421,11 @@ OpenWhisk has a few system limits, including how much memory an action uses and
...
+* A user can change the limit when creating the action.
```
rabbah: or updating.

docs/reference.md
rabbah: suggested alternative: Logs that exceed the set limit are truncated and a warning is added as the last output of the activation to indicate that the activation exceeded the set log limit.

tests/src/common/Wsk.scala
```diff
@@ -226,6 +225,7 @@ class WskAction(override val usePythonCLI: Boolean = false)
         annotations: Map[String, JsValue] = Map(),
         timeout: Option[Duration] = None,
         memory: Option[Int] = None,
+        logsize: Option[Int] = None,
```
rabbah: you could make this take Size instead of Int (since you're using `0 MB` in the test with `.toInt`.

tools/cli/wskaction.py
```diff
@@ -95,7 +97,8 @@ def create(self, args, props, update):
             if args.param:
                 payload['parameters'] = getParams(args)
             # API will accept limits == {} as limits not specified on an update
-            if args.timeout or args.memory:
+            print args.logsize
```
rabbah: remove.

tools/cli/wskaction.py
```diff
@@ -95,7 +97,8 @@ def create(self, args, props, update):
...
+            if args.timeout or args.memory or not args.logsize is None:
```
rabbah: should these all be `is None` vs falsy comparisons?

rabbah:
cli tests for the corner cases (create action and get limits sufficient to make sure covered).

tools/go-cli/go-whisk-cli/commands/action.go
```diff
@@ -506,13 +506,22 @@ func parseAction(cmd *cobra.Command, args []string) (*whisk.Action, bool, error)
...
     // Only include the memory and timeout limit if set
-    if flags.action.memory > 0 || flags.action.timeout > 0 {
+    if flags.action.memory > -1 || flags.action.timeout > -1 || flags.action.logsize > -1 {
         limits = new(whisk.Limits)
-        if flags.action.memory > 0 {
-            limits.Memory = flags.action.memory
+        if flags.action.memory > -1 {
+            limits.Memory = &flags.action.memory
+        } else {
+            limits.Memory = nil
```
mdeuser: Allocated pointers are initialized to nil, so there's actually no need to explicitly set these limits.* pointers to nil.

markusthoemmes: Makes sense, thanks!

rabbah:
Ready modulo corner case tests for CLI which I volunteered to write.

docs/reference.md
```diff
@@ -420,6 +421,11 @@ OpenWhisk has a few system limits, including how much memory an action uses and
 * A user can change the limit when creating the action.
 * A container cannot have more memory allocated than the limit.
 
+### Per action logs (MB) (Default: 10MB)
+* The log limit N is in the range [0MB..10MB] and is set per action.
+* A user can change the limit when creating or updating the action.
+* Logs that exceed the set limit are truncated and a warning is added as the last output of the activation to indicate that the activation exceeded the set log limit.
+
```
mdeuser: Add doc clarifying behavior of log when size limit is set to 0.

rabbah:
Squashed rebased and merged as https://github.com/openwhisk/openwhisk/commit/f0dab306bd8e7bfc9fcafde0467af985940131c9.

rabbah:
@mdeuser I thought it was clear what happens (ie the logs are truncated even with 0MB) - do you have suggested alternative language?

mdeuser:
Since a 0MB log size limit is an unusual setting, I thought it might need explicit doc.  In other products, I've seen the use of 0 to mean "unlimited".

PR #955, opened 2016-07-26
Watson annotation updates

jasonpet:
Fix for some annotation errors found when using the Watson actions from the UI


dubeejw:
@jasonpet, looks good. Can you rebase?

PR #959, opened 2016-07-27
Add catalog-machine to WhiskConfig

ChrBi:
Add catalog-machine to WhiskConfig.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
+++ b/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
@@ -84,6 +84,7 @@ class WhiskConfig(
     val mainDockerEndpoint = this(WhiskConfig.mainDockerEndpoint)
     val routerDockerEndpoint = this(WhiskConfig.routerDockerEndpoint)
     val elkDockerEndpoint = this(WhiskConfig.elkDockerEndpoint)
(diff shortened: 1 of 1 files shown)


common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -158,6 +159,7 @@ object WhiskConfig {
     val mainDockerEndpoint = "main.docker.endpoint"
     val routerDockerEndpoint = "router.docker.endpoint"
     val elkDockerEndpoint = "elk.docker.endpoint"
+    val catalogDockerEndpoint = "catalog.docker.endpoint"
```
rabbah: this will look like dead code in openwhisk since it's not used in openwhisk.
is there a blue whiskconfig that is a better home for this?

rabbah: a more general way perhaps of allow other properties is to provide a getter on a string directly.

rabbah:
Opened an issue in blue to refactor this config class into open/blue.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/e559b029ffdeeb9342cc7c085622a6e68998b993.

PR #960, opened 2016-07-27
Allow invoker to run pull concurrently with other docker ops

perryibm:
Lift the pull operation from the Container ctor chain into the container pool.

Send docker pull operation to a different guard so that pull operations are serialized
among themselves but can run concurrently with non-pull operations.  (PPG 920)


rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/d35288d5705b52404713e1b85eef4edf77549996.

Issue #586, opened 2016-06-07
CLI should support http proxy


csantanapr:
The CLI connects directly to the API HOST on default port 443
But if the user needs to use an http proxy (very typical inside corporations) the CLI needs to provide a way for the user to use the WSK CLI.

We should take same approach as CF CLI, using an environment variable
https_proxy

```
If you have an HTTP proxy server on your network between a host running the cf CLI and your Cloud Foundry API endpoint, you must set https_proxy with the hostname or IP address of the proxy server.

The https_proxy environment variable holds the hostname or IP address of your proxy server.

https_proxy is a standard environment variable. Like any environment variable, the specific steps you use to set it depends on your operating system.

Format of https_proxy
https_proxy is set with hostname or IP address of the proxy server in URL format: https_proxy=http://proxy.example.com

If the proxy server requires a user name and password, include the credentials: https_proxy=http://username:password@proxy.example.com

If the proxy server uses a port other than 80, include the port number: https_proxy=http://username:password@proxy.example.com:8080


```

csantanapr:
Some more notes
It looks like http_proxy is something that the golang http library handles, user can use both HTTP_PROXY or http_proxy

If we use the golang net http library to connect to openwhisk API it should be easy to implement, but as always we will need to write some test cases to have confident in the integration.

cc @dubeejw @mdeuser

rabbah:
Closed by #612.

rabbah:
The Go CLI may not have this feature, and #612 did not add a test when this feature was added to the Python CLI.

@mdeuser @dubeejw FYI.

rabbah:
Added for Python CLI with tests as part of #925 as I ran into the need for this. There is a tangentially related issue for the Go CLI in #924.

dubeejw:
It looks like setting an environment variable for HTTP_PROXY, or HTTPS_PROXY will cause the Go CLI to use a proxy. Would need a proxy to verify this though.

rabbah:
```
Starting test Wsk CLI usage should show api build using http apihost at 2016-07-30 11:07:53.774
expected exit code = 0
exit code = 3
stdout: whisk API host		 
whisk API build		 Unknown

stderr: error: Unable to obtain API build information: Request failure: Get https:///api/v1: http: no Host in request URL
```

csantanapr:
@dubeejw @mdeuser supporting http proxy very important to users behind corporate firewall.

dubeejw:
[...]

PR #962, opened 2016-07-28
Split CLI tests into basic usage and integration tests.

rabbah:
@fxulusoy FYI this is a result of ignoring a test the `http` test.
@mdeuser @dubeejw I split the basic cli tests into one that are integration tests and ones that are mostly basic cli usage - i expect we can put other tests in the new suite over time.

### DIFF
--- a/tests/src/whisk/core/entity/test/SchemaTests.scala
+++ b/tests/src/whisk/core/entity/test/SchemaTests.scala
@@ -16,7 +16,9 @@
 
 package whisk.core.entity.test
 
+import scala.BigInt
 import scala.Vector
+import scala.concurrent.duration.DurationInt
 import scala.language.postfixOps
 import scala.language.reflectiveCalls
 import scala.util.Try
@@ -24,6 +26,7 @@ import scala.util.Try
 import org.junit.runner.RunWith
 import org.scalatest.BeforeAndAfter
 import org.scalatest.FlatSpec
+import org.scalatest.Matchers
 import org.scalatest.junit.JUnitRunner
 
 import spray.json.DefaultJsonProtocol._
@@ -36,6 +39,7 @@ import spray.json.JsString
 import spray.json.JsValue
 import spray.json.pimpAny
 import spray.json.pimpString
+
 import whisk.core.entity.ActionLimits
 import whisk.core.entity.ActivationId
 import whisk.core.entity.AuthKey
@@ -43,15 +47,15 @@ import whisk.core.entity.DocId
 import whisk.core.entity.DocInfo
 import whisk.core.entity.EntityName
 import whisk.core.entity.Exec
+import whisk.core.entity.LogLimit
 import whisk.core.entity.MemoryLimit
 import whisk.core.entity.Namespace
 import whisk.core.entity.Parameters
 import whisk.core.entity.Secret
 import whisk.core.entity.SemVer
 import whisk.core.entity.TimeLimit
-import whisk.core.entity.LogLimit
 import whisk.core.entity.UUID
-import org.scalatest.Matchers
+import whisk.core.entity.size.SizeInt
 
 @RunWith(classOf[JUnitRunner])
 class SchemaTests extends FlatSpec with BeforeAndAfter with Matchers {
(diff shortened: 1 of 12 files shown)


rabbah:
@ChrBi @markusthoemmes a few changes for your review.

rabbah:
@mdeuser @dubeejw note and review the changes in restructuring the basic cli tests - i moved some out of the system package and into a new cli test package intended to test basic operations and corner cases using the cli. not all of the tests in the latter can run without a deployed backend (yet).

tests/src/whisk/core/entity/test/SchemaTests.scala
```diff
@@ -317,9 +321,9 @@ class SchemaTests extends FlatSpec with BeforeAndAfter with Matchers {
 
     it should "properly deserialize JSON" in {
         val json = Seq[JsValue](
-            JsObject("timeout" -> TimeLimit.STD_DURATION.toMillis.toInt.toJson, "memory" -> MemoryLimit.STD_MEMORY.toJson, "logs" -> LogLimit.STD_LOGSIZE.toJson),
-            JsObject("timeout" -> TimeLimit.STD_DURATION.toMillis.toInt.toJson, "memory" -> MemoryLimit.STD_MEMORY.toJson, "logs" -> LogLimit.STD_LOGSIZE.toJson, "foo" -> "bar".toJson),
-            JsObject("timeout" -> TimeLimit.STD_DURATION.toMillis.toInt.toJson, "memory" -> MemoryLimit.STD_MEMORY.toJson))
+            JsObject("timeout" -> TimeLimit.STD_DURATION.toMillis.toInt.toJson, "memory" -> MemoryLimit.STD_MEMORY.size.toInt.toJson, "logs" -> LogLimit.STD_LOGSIZE.size.toInt.toJson),
```
markusthoemmes: Use `toMB` instead of `size` to force this to return the value in MB (as expected by the backend)

tests/src/whisk/core/entity/test/SchemaTests.scala
```diff
@@ -335,13 +339,13 @@ class SchemaTests extends FlatSpec with BeforeAndAfter with Matchers {
             JsObject(),
             JsNull,
             JsObject("timeout" -> TimeLimit.STD_DURATION.toMillis.toInt.toJson),
-            JsObject("memory" -> MemoryLimit.STD_MEMORY.toJson),
-            JsObject("logs" -> (LogLimit.STD_LOGSIZE + 1).toJson),
-            JsObject("TIMEOUT" -> TimeLimit.STD_DURATION.toMillis.toInt.toJson, "MEMORY" -> MemoryLimit.STD_MEMORY.toJson),
-            JsObject("timeout" -> (TimeLimit.STD_DURATION.toMillis.toDouble + .01).toJson, "memory" -> (MemoryLimit.STD_MEMORY.toDouble + .01).toJson),
+            JsObject("memory" -> MemoryLimit.STD_MEMORY.size.toInt.toJson),
+            JsObject("logs" -> (LogLimit.STD_LOGSIZE.size.toInt + 1).toJson),
```
markusthoemmes: see above

tests/src/whisk/core/entity/test/SchemaTests.scala
```diff
@@ -335,13 +339,13 @@ class SchemaTests extends FlatSpec with BeforeAndAfter with Matchers {
...
+            JsObject("TIMEOUT" -> TimeLimit.STD_DURATION.toMillis.toInt.toJson, "MEMORY" -> MemoryLimit.STD_MEMORY.size.toInt.toJson),
+            JsObject("timeout" -> (TimeLimit.STD_DURATION.toMillis.toDouble + .01).toJson, "memory" -> (MemoryLimit.STD_MEMORY.size.toDouble + .01).toJson),
             JsObject("timeout" -> null, "memory" -> null),
             JsObject("timeout" -> JsNull, "memory" -> JsNull),
-            JsObject("timeout" -> TimeLimit.STD_DURATION.toMillis.toString.toJson, "memory" -> MemoryLimit.STD_MEMORY.toString.toJson))
+            JsObject("timeout" -> TimeLimit.STD_DURATION.toMillis.toString.toJson, "memory" -> MemoryLimit.STD_MEMORY.size.toInt.toString.toJson))
```
markusthoemmes: See above (apply to all instances)

markusthoemmes:
Great job, looks good to me modulo comments.

tests/src/whisk/core/entity/test/SchemaTests.scala
rabbah: good point - i used size because i assumed/knew it was MB already but that may change.

mdeuser:
@rabbah - Thanks for the heads up.  When adding new test cases in the future, what litmus test can be applied to determine whether the test case falls into the "integration" camp or not?

rabbah:
new IPG 386 and PPG 939.

rabbah:
@mdeuser I split the tests based on what criteria and invariant were being checked. Basic usage, testing parameter corner cases,  validity of arguments and the like I put in the new suite. Tests which are verifying the CRUD+activate life cycles or package operations I left in the `system` package. We can evolve a better criteria for where to locate CLI tests.

rabbah:
Going to rebase and resolve conflicts following merge of #927.

rabbah:
@mdeuser @dubeejw I rebased and resolved conflicts from the tests added in #927. In the process I eliminated one test which was not redundant (checking that unset of multiple values works as expected).

One test is still using python CLI (down from 2) because the go CLI assumes https always even if api host is set explicitly. Can you address this?

rabbah:
@mdeuser @dubeejw I rebased and resolved conflicts from the tests added in #927. In the process I discovered that doing `wsk --apihost <valid host> property get --apibuild` is not working and neither is the test which tries to access controller via `http`. Can you address these separately (https://github.com/openwhisk/openwhisk/issues/586) or send a pull request to my branch and I can merge here.

I also eliminated one test that is redundant (checking that unset of multiple values works as expected).

rabbah:
PPG 945.

rabbah:
Rebased and squashed.

mdeuser:
@rr - You are correct in that that doing `go-cli/wsk --apihost <valid host> property get --apibuild` will not behave like the Python API.

The hard Go CLI limitation is support for argument overloading - assigning the same named argument to the root command and also to a sub-command.  In this example, not only is the argument assigned twice, but it has different characteristics.  The `--apihost` argument assigned to the `wsk` root command takes a value while the `--apihost` argument assigned to the `property get` command does not.  The Go CLI argument package permits the assignment of the same argument name to different sub-commands as well as the root command; however, for any particular CLI invocation, all arguments are merged such that the argument is applied only to the last sub-command.

If we want the `wsk property get` command to take a apihost argument as the target host, the argument label will need to change.  That said, I think the better solution may be to separate the existing `wsk property` command into two commands - one for configuring the CLI and another for accessing whisk server values.  This would remove the `--apihost` argument overload conflict.

rabbah:
Making it too commands makes it inconvenient to check build versions against different deployments (now it's two steps). What about if get didn't accept arguments with -- but instead a set of predefined values so instead of --apihost it becomes apihost?

rabbah:
It looks like we will need to address game outstanding issue separately (for http, missing host, and property get). In which case shall we merge this to get the suite split and accept that three tests will continue to run against the Python cli until a resolution is in place for the go cli?

mdeuser:
Rebased and merged.  PG 79 is now running against latest blue/open

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/461d9a8fd67113bafd09ff4cfd09cc67780b219a.

Issue #958, opened 2016-07-27
Document limits for exec, parameters and annotations


markusthoemmes:
We recently introduced limits for those. We need to make sure they are properly documented.

@csantanapr fyi

csantanapr:
@ChrBi please add the limit information in the [reference.md](https://github.com/openwhisk/openwhisk/blob/master/docs/reference.md)

rabbah:
@cbickel done?

PR #963, opened 2016-07-28
Add documentation for exec-, parameter- and annotation-limits #958

ChrBi:
Documentation for the Code, annotation and parameter limit.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -151,7 +151,7 @@ function main(params) {
 ```
 
 A JavaScript action's activation is **asynchronous** if the main function exits by returning a Promise.  In this case the system assumes that the action is still running, until the Promise has been fulfilled or rejected.
-Start by instantiating a new Promise object and passing it a callback function. The callback takes two arguments, resolve and reject, which are both functions. All your asynchronous code goes inside that callback. 
+Start by instantiating a new Promise object and passing it a callback function. The callback takes two arguments, resolve and reject, which are both functions. All your asynchronous code goes inside that callback.
 
 Here is an example on how to fulfill a Promise by calling the resolve function.
 
@@ -183,7 +183,7 @@ It is possible an action is synchronous on some inputs and asynchronous on other
   function main(params) {
       if (params.payload) {
          // asynchronous activation
-         return new Promise(function(resolve, reject) { 
+         return new Promise(function(resolve, reject) {
                 setTimeout(function() {
                   resolve({ done: true });
                 }, 100);
@@ -408,6 +408,8 @@ OpenWhisk has a few system limits, including how much memory an action uses and 
 | concurrent | it's not allowed to have more than N concurrent activations per namespace | per namespace | number | 100 |
 | minuteRate | a user cannot invoke more than this many actions per minute | per user | number | 120 |
 | hourRate | a user cannot invoke more than this many actions per hour | per user | number | 3600 |
+| codeSize | the maximum size of the actioncode | not configurable, limit per action | MB | 48 |
+| parameters | the maximum size of the paramters that can be attached | not configurable, limit per action/package/trigger | MB | 1 |
 
 ### Per action timeout (ms) (Default: 60s)
 * The timeout limit N is in the range [100ms..300000ms] and is set per action in milliseconds.
@@ -424,8 +426,8 @@ OpenWhisk has a few system limits, including how much memory an action uses and 
 * A user can change the limit when creating or updating the action.
 * Logs that exceed the set limit are truncated and a warning is added as the last output of the activation to indicate that the activation exceeded the set log limit.
 
-### Per action artifact (MB) (Fixed: 1MB)
-* The maximum code size for the action is 1MB.
+### Per action artifact (MB) (Fixed: 48MB)
+* The maximum code size for the action is 48MB.
 * It is recommended for a JavaScript action to use a tool to concatenate all source code including dependencies into a single bundled file.
 
 ### Per activation payload size (MB) (Fixed: 1MB)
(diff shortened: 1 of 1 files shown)


docs/reference.md
```diff
@@ -430,6 +433,21 @@ OpenWhisk has a few system limits, including how much memory an action uses and
...
+
+### Size of the parameters (Fixed: 1MB)
+* The site limit for the parameters of an action/package/trigger is 1MB.
+* The limit cannot be changed by the user.
+* An entity with too big parameters will be rejected on trying to create or update it.
+
+### Size of the annotations (Fixed: 1MB)
+* The size limit for the annotations of an action/package/rule/trigger is 1MB.
+* The limit cannot be changed by the user.
+* An entity with too big parameters will be rejected on trying to create or updating it.
```
markusthoemmes: update

docs/reference.md
```diff
@@ -430,6 +433,21 @@ OpenWhisk has a few system limits, including how much memory an action uses and
 * A user cannot change this limit when creating the action.
 * A CLI call that exceeds this limit receives an error code corresponding to TOO_MANY_REQUESTS.
 
+### Size of the code per action (Fixed: 48MB)
+* The size limit for the code for an action is 48MB.
+* The limit cannot be changed by the user.
+* An action with too much code will be rejected on trying to create or update it.
+
+### Size of the parameters (Fixed: 1MB)
+* The site limit for the parameters of an action/package/trigger is 1MB.
```
markusthoemmes: size

docs/reference.md
```diff
@@ -430,6 +433,21 @@ OpenWhisk has a few system limits, including how much memory an action uses and
 * A user cannot change this limit when creating the action.
 * A CLI call that exceeds this limit receives an error code corresponding to TOO_MANY_REQUESTS.
 
+### Size of the code per action (Fixed: 48MB)
+* The size limit for the code for an action is 48MB.
```
markusthoemmes: for the code of an action

docs/reference.md
```diff
@@ -402,6 +402,9 @@ OpenWhisk has a few system limits, including how much memory an action uses and
 | concurrent | it's not allowed to have more than N concurrent activations per namespace | per namespace | number | 100 |
 | minuteRate | a user cannot invoke more than this many actions per minute | per user | number | 120 |
 | hourRate | a user cannot invoke more than this many actions per hour | per user | number | 3600 |
+| codeSize | the maximum size of the actioncode | not configurable, limit is for each action | MB | 48 |
+| parameterLimit | the maximum size of the paramters that can be attached | not configurable, limit is for each action/package/trigger | MB | 1 |
```
markusthoemmes: parameters

docs/reference.md
```diff
@@ -402,6 +402,9 @@ OpenWhisk has a few system limits, including how much memory an action uses and
...
+| annotationLimit | the maximum size of the annotations that can be attached | not configurable, limit os for each action/package/rule/trigger | MB | 1 |
```
markusthoemmes: is

docs/reference.md
```diff
@@ -402,6 +402,9 @@ OpenWhisk has a few system limits, including how much memory an action uses and
 | concurrent | it's not allowed to have more than N concurrent activations per namespace | per namespace | number | 100 |
 | minuteRate | a user cannot invoke more than this many actions per minute | per user | number | 120 |
 | hourRate | a user cannot invoke more than this many actions per hour | per user | number | 3600 |
+| codeSize | the maximum size of the actioncode | not configurable, limit is for each action | MB | 48 |
```
markusthoemmes: limit per action

docs/reference.md
```diff
@@ -430,6 +433,21 @@ OpenWhisk has a few system limits, including how much memory an action uses and
 * A user cannot change this limit when creating the action.
 * A CLI call that exceeds this limit receives an error code corresponding to TOO_MANY_REQUESTS.
 
+### Size of the code per action (Fixed: 48MB)
+* The size limit for the code of an action is 48MB.
+* The limit cannot be changed by the user.
+* An action with too much code will be rejected on trying to create or update it.
+
+### Size of the parameters (Fixed: 1MB)
+* The size limit for the parameters of an action/package/trigger is 1MB.
```
csantanapr: Not very clear what you are referring here.
You mean at creating time vs. activation time (invoke or fire) ?
For example when creating a new action with the CLI content for --param arguments can't be larger than 1MB?

docs/reference.md
```diff
@@ -430,6 +433,21 @@ OpenWhisk has a few system limits, including how much memory an action uses and
...
+* The limit cannot be changed by the user.
+* An entity with too big parameters will be rejected on trying to create or update it.
+
+### Size of the annotations (Fixed: 1MB)
```
csantanapr: I don't think we should document annotations, since we don't document what they are and they are for.
I think once we doc annotations, then the limitation can be added to the docs.

docs/reference.md
```diff
@@ -430,6 +433,21 @@ OpenWhisk has a few system limits, including how much memory an action uses and
 * A user cannot change this limit when creating the action.
 * A CLI call that exceeds this limit receives an error code corresponding to TOO_MANY_REQUESTS.
 
+### Size of the code per action (Fixed: 48MB)
```
csantanapr: I think this means that you need to remove the document the limitation that we already have about action artifacts cant' be larger than 1MB, if I understand this limitation correctly

today we have this in the doc both table and text:
Per action artifact (MB) (Fixed: 1MB)

The maximum code size for the action is 1MB.
It is recommended for a JavaScript action to use a tool to concatenate all source code including dependencies into a single bundled file.

csantanapr:
@ChrBi I review the changes, please address at your convenience

docs/reference.md
rabbah: There is a bug/feature (take your pick) in the nodejs proxy which limits the body payload to 1MB - this should be removed: https://github.com/openwhisk/openwhisk/blob/master/core/nodejsActionBase/app.js#L36 as it conflates the size of the exec and arguments unnecessarily.

docs/reference.md
rabbah: Do we have an open issue to document annotations - I did not find it.

docs/reference.md
rabbah: good point - applies to the sum (1MB is total which means create/update+invoke).

csantanapr: Yes we have: #973  😸

docs/reference.md
csantanapr: @rabbah yes we have #973 😸

docs/reference.md
ChrBi: The limit is only for PUT requests. So the limit is only on creating/updating the entity.

csantanapr:
rebase and merged fdd10ae2f63ed8d9cab89c27e89a6dab37e54193

Issue #967, opened 2016-07-28
upgrade to ansible latest version


houshengbo:
## Steps to reproduce the issue:

For each PR submitted to openwhisk, the travis ci will running into a issue that db_local.ini is missing.

## Additional information you deem important:
* we can specify the ansible version to 2.0.2.0 to make sure everything works smooth.

rabbah:
@lehoanganh can we upgrade required ansible to at least v2.1.0.0? the latest release might fix #632 but requires other changes to our playbooks.

lehoanganh:
I'm trying the latest version. The drawback of pinned version with apt-get is, the Ansible team delete the old version once they publish the new version in the repo. We were also facing this problem in the past where the pinned version was not found via apt-get.

rabbah:
so the pip path while results in a longer install step might be best in general?

lehoanganh:
or we can pin the version, if the problem occurs again, fix it with the new version. With such process, we can control the installed version.

lehoanganh:
An example for you
https://travis-ci.org/lehoanganh/openwhisk/builds/148289322

Try this pull: https://github.com/openwhisk/openwhisk/pull/968/files

```
E: Version '2.1.0.0-1*' for 'ansible' was not found
```

```
lha@ubuntu  ~  sudo apt-cache policy ansible
ansible:
  Installed: (none)
  Candidate: 2.1.1.0-1ppa~trusty
  Version table:
     2.1.1.0-1ppa~trusty 0
        500 http://ppa.launchpad.net/ansible/ansible/ubuntu/ trusty/main amd64 Packages
     1.7.2+dfsg-1~ubuntu14.04.1 0
        100 http://us.archive.ubuntu.com/ubuntu/ trusty-backports/universe amd64 Packages
     1.5.4+dfsg-1 0
        500 http://us.archive.ubuntu.com/ubuntu/ trusty/universe amd64 Packages
```

`2.1.0.0-1` was deleted from repo

rabbah:
so pip install 2.0.0.2 or 2.1.0.0?

lehoanganh:
`pip install --user ansible==2.0.2.0` worked. https://travis-ci.org/lehoanganh/openwhisk/builds/148293667

lehoanganh:
pip persists all versions of a package

```
lha@dyn-9-152-96-140  ~/Development/hoang/openwhisk   test-travis  pip install ansible==1111111111111
Collecting ansible==1111111111111
[...]

Issue #530, opened 2016-05-27
setup failing to install ansible using ubuntu-setup/ansible.sh


csantanapr:
Affects vagrant and stand-alone ubuntu
```
Reading state information... Done
E: Version '2.0.2.0-1ppa~trusty' for 'ansible' was not found
sudo: pip: command not found
The program 'ansible' is currently not installed. To run 'ansible' please ask your administrator to install the package 'ansible'
The program 'ansible-playbook' is currently not installed. To run 'ansible-playbook' please ask your administrator to install the package 'ansible'
```

csantanapr:
Root cause is that we specify to strict version of ansible `2.0.2.0-1ppa~trusty`
this version is not longer available, 
It should be more relax the version like `2.*`

domdom82:
actually it needs to be at least 2.0.2.0 I wonder why they keep throwing out previous versions. makes version pinning impossible.

csantanapr:
Using 2.* installs 2.1.0.0 which it works for us

See [here](https://launchpad.net/~ansible/+archive/ubuntu/ansible)

domdom82:
I guess we will be safe. We had problems with 2.0.1.0 that were fixed in 2.0.2.0. So as long as they don't screw up and put up 2.0.1.0 again using "2.*" will work ok. I wonder why they don't just keep old versions for some time longer.

PR #968, opened 2016-07-29
Use ansible 2.0.2.0 as the version to install openwhisk

houshengbo:
Closes-Bug: #967

### DIFF
--- a/tools/travis/setup.sh
+++ b/tools/travis/setup.sh
@@ -13,4 +13,10 @@ sudo service docker restart
 # Python stuff
 sudo apt-get -y install python-pip
 pip install --user jsonschema
-pip install --user ansible
+
+# Ansible stuff
+sudo apt-get install -y software-properties-common
+sudo apt-add-repository -y ppa:ansible/ansible2.1
+sudo apt-get update
+sudo apt-get install -y ansible=2.1.0.0-1*
+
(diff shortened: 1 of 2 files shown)


rabbah:
In https://github.com/openwhisk/openwhisk/issues/530 the change was made to relax the installed version of ansible and in https://github.com/houshengbo/openwhisk/commit/7ab9a47c8c17cfa0adbd0549c5b66bcb601871f9#diff-eb553454a207e03d414e88d99e504148 the installed ansible was done via apt-get instead of pip (because "pip install on vagrant requires lots of dev packages as it wants to compile deps. we don't want that.").

so we may want to install both via apt-get but install a version earlier than the latest which was just released.

@csantanapr @houshengbo

rabbah:
`sudo apt-get install -y ansible=2.1.0.0-*` might be the stop gap.

houshengbo:
vagrant is calling source all.sh under ubuntu setup. 
sudo apt-get install -y ansible=2.1.0.0-1* and 2.1.0.0-1ppa~trusty do not work well. It always shows unable to find ansible=2.1.0.0-1* or 2.1.0.0-1ppa~trusty.

pip can work well in installing 2.1.0.0.

rabbah:
Reverted to using pip.

Merged as https://github.com/openwhisk/openwhisk/commit/36ad900f7a81cb835e732529a5f0a69429067dd3.

PR #972, opened 2016-07-29
Compute cores dynamically, Impose CPU limit. Multiple containers per core.

perryibm:
Compute number of cores dynamically.  Use cpu sharing to limit cpu resource.  Map multiple containers per core.  PPG 937+938.


core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -633,8 +638,33 @@ class ContainerPool(
...
+    /*
+     * The total number of containers is simply the number of cores dilated by the cpu sharing.
+     */
+    def getDefaultMaxActive() = getNumberOfCore() * ContainerPool.shareFactor
+
+    /*
+     * Retrieve the number of cores on the host machine - defaulting to defaultNumCore if unknown.
+     */
+    private def getNumberOfCore(): Int = try {
+        val (stdout, stderr, exitCode) = SimpleExec.syncRunCmd(Array("/usr/bin/nproc"))(TransactionId.invoker)
```
psuter: Will this give a different result than `Runtime.getRuntime().availableProcessors()` ?

core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -633,8 +638,33 @@ class ContainerPool(
...
     def requiredProperties = Map(selfDockerEndpoint -> "localhost") ++ Map(dockerImageTag -> "latest") ++ Map(invokerContainerNetwork -> "bridge") ++ Map(invokerContainerPolicy -> "")
     type RunResult = (Instant, Instant, Option[(Int, String)])
-    val defaultMaxActive = 4
+
+    private val defaultNumCore = 4  // The number of cores if it cannot be determined dynamically with nproc
+    // The shareFactor indicates the number of containers that would share a single core, on average.
+    // cpuShare is a docker option (-c) whereby a container's CPU access is limited.
+    // A value of 1024 is the full share so a strict resource division with a shareFactor of 2 would yield 512.
+    // On an idle/underloaded system, a container will still get to use underutilized CPU shares.
+    val shareFactor = 2
```
jeremiaswerner: would it make sense to make the shareFactor configurable to run tests with different parameters?

jeremiaswerner:
@danielfederschmidt fyi

core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
markusthoemmes: Also could this be expressed with a `Try { } match ...` for "scalaism" and a dropped `return`?

core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -633,8 +638,33 @@ class ContainerPool(
...
+    val cpuShare = (1024.0 / getDefaultMaxActive()).toInt
+
+    /*
+     * The total number of containers is simply the number of cores dilated by the cpu sharing.
+     */
+    def getDefaultMaxActive() = getNumberOfCore() * ContainerPool.shareFactor
```
markusthoemmes: It's convention in Scala to drop `()` functions without a sideeffect. One usually also doesn't use `get` for a getter but rather drops the "get". That'll even make the changes above unnecessary, as this method would then be called `defaultMaxActive`

core/dispatcher/src/main/scala/whisk/core/container/ContainerUtils.scala
```diff
@@ -67,14 +67,15 @@ trait ContainerUtils extends Logging {
      */
     def makeContainer(name: Option[String], image: String, network: String, env: Map[String, String], args: Array[String], limits: ActionLimits, policy: Option[String])(implicit transid: TransactionId): ContainerId = {
         val nameOption = if (name isDefined) Array("--name", name.getOrElse("")) else Array.empty[String]
+        val cpuArg = Array("-c", s"${ContainerPool.cpuShare}")
```
markusthoemmes: Why the string formatting? You could write just `ContainerPool.cpuShare.toString`

core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
rabbah: you mean via "whisk config"?

core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
perryibm: @psuter  Yes, you are right though that reminds me that actually both are incorrect and my code was a failed attempt to get the right answer.  The problem is that the number of cores available to the JVM is not the same as the host since the invoker's container could itself be limited.  So my choice is to assume that is has not been limited (and use your code) or else to get "nproc" to run outside the container.   The latter is more correct but we have no way to do this without messing with deployment.

core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
perryibm: The reason it is implemented this way was as a suggestion (perhaps this is premature) that the method, while not side-effecting, is not a pure function as the cpu set could change.

core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
perryibm: Yes, I think we could consider making this a deployment parameter that is stored in consul and adjustable so that if an invoker is restarted, it will grab the current shareFactor value.  That would make this value here a backup value in case someone writes a garbage value into consul.

rabbah: use whisk config then :)

core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
psuter: Thanks for the detailed clarification. Given how unreliable all methods are, I support the opinion expressed by others to make this part of the config. I think it's acceptable (at least as a first step) to assume homogeneous invokers and specify the number of cores only once? (As opposed to: `invoker.0.numcores = .., invoker.1.numcores = ...`)

rabbah:
@perryibm can you confirm that this blocked by #1016.

rabbah:
pending pg @perryibm ?

rabbah:
Needs conflict resolution.

Issue #976, opened 2016-07-31
stem cell containers for nodejs actions should use nodejs 6


rabbah:
The stem cell containers that the invoker pre-warms are not for nodejs 6 which is now the default node runtime. Hence we are not getting good utilization out of these stem cell containers.

PR #977, opened 2016-07-31
Change pre-warm container image (issue #976).

rabbah:
Change image for pre-warmed containers from NodeJs to NodeJs6 since the latter is the default for Node actions now.

PPG 947 approved.


perryibm:
rebased and merged

Issue #932, opened 2016-07-21
Port uses of async/done/error to Promises


psuter:
For JavaScript feeds, actions, etc., as `Promise`s are the way forward.

An exception should be made for tests that explicitly exercise the `async`/`done`/`error` patterns as long as they are not deprecated.

To get a crude estimate of the changes required (as of 9fceba2ee3f9a06e6466f6fe71f2d1235c4dbde1):

    $ grep -lrI --include '*.js' async .
    ./catalog/github/webhook.js
    ./catalog/samples/countdown.js
    ./catalog/samples/curl.js
    ./catalog/samples/helloAsync.js
    ./catalog/samples/helloPromises.js
    ./catalog/samples/wcbin.js
    ./catalog/slack/post.js
    ./catalog/utils/pipe.js
    ./catalog/watson/languageId.js
    ./catalog/watson/speechToText.js
    ./catalog/watson/textToSpeech.js
    ./catalog/watson/translate.js
    ./catalog/weather/forecast.js
    ./catalog/websocket/sendWebSocketMessageAction.js
    ./core/nodejsActionBase/runner.js
    ./core/nodejsActionBase/src/whisk.js
    ./core/nodejsActionBase/test.js
    ./tests/dat/actions/helloAsyncDoneTwice.js
    ./tests/dat/actions/ping.js

FYI @bjustin-ibm @jasonpet

jasonpet:
We should make these changes in openwhisk-catalog for all actions that have been moved there.

bjustin-ibm:
@jasonpet do you want to take this, or perhaps split the chore between us?

jasonpet:
We can split it.  I will do it for all of the actions that will now exist in openwhisk-catalog if you can do it for the ones staying behind.

bjustin-ibm:
Sounds good, can you give me a concrete list of the ones that are staying?

psuter:
If you could also make a note of which ones are actually tested? (hopefully all of them)

jasonpet:
@houshengbo - do you have a list already of all the actions that are actually tested?

PR #996, opened 2016-08-02
Have whisk.invoke() and whisk.trigger() optionally return a Promise

### LINKED ISSUE #987: Should whisk.invoke() return a Promise?
I think the answer is definitely "yes" but not breaking existing action code makes it a little tricky. If I understand correctly, `whisk.invoke()` is currently always asynchronous and returns nothing.  If you specify a callback function as the `next` parameter, that callback will get called at some later time, depending on the value of the `blocking` parameter.

I propose that if `next` is left unspecified, then `whisk.invoke()` should return a Promise. If an error occurs, the promise is rejected with an error object (the same that would be passed to the callback, had `next` been specified). The promise is resolved on the activation in the same way that the callback would be called - as soon as the activation is created if `blocking` is false, or with the result if `blocking` is true.

### Non-blocking example
```javascript
whisk.invoke({
    name: 'myAction',
    blocking: false
  })
  .then(function (activation) {
      // activation completed successfully, but not necessarily finished
  })
  .catch(function (reason) {
      // activation was not created
      console.log('An error has occured ' + reason.error);
  });
```

### Blocking example
```javascript
whisk.invoke({
    name: 'myAction',
    blocking: true
  })
  .then(function (activation) {
      // activation completed successfully, activation contains the result
      console.log('Activation ' + activation.activationId + ' completed successfully and here is the result ' + activation.result);
  })
  .catch(function (reason) {
      console.log('An error has occured ' + reason.error);

      if(reason.activation) {
        console.log('Please check activation ' + reason.activation.activationId + ' for details.');
      } else {
        console.log('Failed to create activation.');
      }
  });
```

By doing it this way, existing actions do not need to change.  If an existing use of `whisk.invoke()` specifies a `next` parameter, the behavior is exactly the same - `whisk.invoke()` returns nothing, the callback is called as before. If they do not specify a `next` parameter, then their existing action code does not need to change because it already ignores what is returned (nothing) from calling `whisk.invoke()`.

bjustin-ibm:
Resolves #987

### DIFF
--- a/core/nodejsActionBase/src/whisk.js
+++ b/core/nodejsActionBase/src/whisk.js
@@ -103,31 +103,22 @@ function Whisk(apihost, logger) {
         // { activationId }.
         //
         // Project the appropriate value based on invoke mode and HTTP statusCode
-        // where success is 200, and anything else is a failure.
-        post(packet, logger, function(error, response, body) {
-            logger && logger.info('[whisk]', '[invoke]', 'response body', body);
+        // where success is 200 or 202, and anything else is a failure.
+        var acceptStatusCode = function(statusCode) {
+          return (statusCode === 200 || statusCode === 202);
+        };
+        return post(packet, logger, acceptStatusCode)
+          .then(function (runResult) {
+              logger && logger.info('[whisk]', '[invoke]', runResult.error, runResult.activation);
 
-            var activation = undefined;
-
-            error = error || undefined;
-            response = response || {};
-            body = body || {};
-
-            if (response.statusCode == 200 || response.statusCode == 202) {
-                activation = {
-                   activationId: body.activationId, // id always present
-                   result: (body.response || {}).result // may not exist if non-blocking
-                };
-            } else if (!error) {
-                // activation failed, set error to API host error response.
-                error = body.error + ' (' + body.errorCode + ')';
-            }
-
-            if (next) {
-                logger && logger.info('[whisk]', '[invoke]', 'next', error, activation);
-                next(error, activation);
-            }
-        });
+              if (next) {
+                  next(runResult.error, runResult.activation);
+              } else if (!runResult.error) {
+                  return runResult.activation;
+              } else {
+                  return Promise.reject(runResult);
+              }
+          });
     }
 
     /**
@@ -159,27 +150,21 @@ function Whisk(apihost, logger) {
         //
         // Project the appropriate value based on HTTP statusCode
         // where success is 200, and anything else is a failure.
-        post(packet, logger, function(error, response, body) {
-            logger && logger.info('[whisk]', '[trigger]', 'response body', body);
+        var acceptStatusCode = function(statusCode) {
+          return statusCode === 200;
+        };
+        return post(packet, logger, acceptStatusCode)
+          .then(function (runResult) {
+              logger && logger.info('[whisk]', '[trigger]', runResult.error, runResult.activation);
 
-            var activation = undefined;
-
-            error = error || undefined;
-            response = response || {};
-            body = body || {};
-
-            if (response.statusCode == 200) {
-                activation = { activation: body.activationId }; // id always present
-            } else if (!error) {
-                // activation failed, set error to API host error response.
-                error = body.error + ' (' + body.errorCode + ')';
-            }
-
-            if (next) {
-                logger && logger.info('[whisk]', '[trigger]', 'next', error, activation);
-                next(error, activation);
-            }
-        });
+              if (next) {
+                  next(runResult.error, runResult.activation);
+              } else if (!runResult.error) {
+                  return runResult.activation;
+              } else {
+                  return Promise.reject(runResult);
+              }
+          });
     }
 
     /**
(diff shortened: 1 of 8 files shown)


bjustin-ibm:
All NodeJS* action container tests pass now (at least locally). Time to write some new tests that make use of the returned promise.

tests/src/system/basic/WskBasicNodeTests.scala
```diff
@@ -80,6 +81,101 @@ class WskBasicNodeTests
             }
     }
 
+    it should "Ensure that whisk.invoke() returns a promise" in withAssetCleaner(wskprops) {
+        val expectedDuration = 3 seconds
+
+        (wp, assetHelper) =>
+            val asyncName = "asyncAction"
```
rabbah: is asyncAction needed?

tests/src/system/basic/WskBasicNodeTests.scala
```diff
@@ -80,6 +81,101 @@ class WskBasicNodeTests
...
+                    val duration = System.currentTimeMillis() - start
+                    duration should be >= expectedDuration.toMillis
+            }
+    }
+
+    it should "Ensure that whisk.invoke() still uses a callback when provided one" in withAssetCleaner(wskprops) {
+        val expectedDuration = 3 seconds
+
+        (wp, assetHelper) =>
+            val asyncName = "asyncAction"
```
rabbah: is asyncAction needed?

tests/src/system/basic/WskBasicNodeTests.scala
```diff
@@ -80,6 +81,101 @@ class WskBasicNodeTests
...
+            val asyncAction = Some(TestUtils.getTestActionFilename("threeSecondRule.js"))
```
rabbah: what's the rationale for calling this a 3s rule file (wondering if it should we have a clearer filename)?

bjustin-ibm: I may have added the following comment to threeSecondRule.js after you made this comment. Let me know if it is still unclear:

```javascript
/*
 * After 3 seconds, either resolve or reject depending on the value
 * of the resolveOrReject parameter.
 *
 * This craziness is so that the whisk.invoke() promise support can be
 * thoroughly tested (for invocations that both resolve and reject).
 */
```

tests/src/system/basic/WskBasicNodeTests.scala
bjustin-ibm: Do I not need to provide the action name to the cleaner and `action.create`?

bjustin-ibm: I see now... the name of the action should have been "ThreeSecondRule" to match the action invoked in `invokePromise.js` and `invokeCallback.js`

bjustin-ibm:
@csantanapr Please review the documentation update.

csantanapr:
>@csantanapr Please review the documentation update.
I'm not allowed to review anything else? 😿

bjustin-ibm:
> I'm not allowed to review anything else? 😿 As a javascript fanboy I feel hurt

Ha! Review all you want 👀

core/nodejsActionBase/src/whisk.js
```diff
@@ -104,28 +104,42 @@ function Whisk(apihost, logger) {
...
-                   result: (body.response || {}).result // may not exist if non-blocking
-                };
-            } else if (!error) {
-                // activation failed, set error to API host error response.
-                error = body.error + ' (' + body.errorCode + ')';
-            }
+              if (response.statusCode == 200 || response.statusCode == 202) {
+                  activation = {
+                     activationId: body.activationId, // id always present
+                     result: (body.response || {}).result // may not exist if non-blocking
```
psuter: This is tricky (and not necessarily a problem introduced by this revision): if we set `result` to `{}` when it's unavailable, how can the user differentiate between a blocking call that ended up returning before the result was available (a possibility documented in the REST API), and a blocking call that did return an empty object as the result?

psuter:
I'd like to request that we apply the same upgrade to `whisk.trigger()`.

csantanapr:
LGTM, same as @psuter let's do `whisk.trigger()` also

bjustin-ibm:
> I'd like to request that we apply the same upgrade to whisk.trigger().

I totally agree. Can we handle that in a separate PR though? #932 is waiting for the `invoke()` enhancement.

psuter:
I don't think any of those are urgent? I'd rather have two consistent points (before and after the optional `next`).

bjustin-ibm:
> I'd rather have two consistent points (before and after the optional next).

Sounds reasonable.  I'll get on it.

bjustin-ibm:
@csantanapr @psuter Both `whisk.trigger()` and `whisk.invoke()` are now implemented to return a promise. Have at it!

csantanapr:
@bjustin-ibm changes LGTM, 👍  for consolidation code in `post` method

psuter:
Ready when squashed as far as I'm concerned :+1:

bjustin-ibm:
@psuter squashed and ready!

csantanapr:
PG 109

tests/src/system/basic/WskBasicNodeTests.scala
```diff
@@ -80,6 +82,171 @@ class WskBasicNodeTests
...
+            assetHelper.withCleaner(wsk.action, nameOfActionThatTriggers) {
+                (action, _) =>
+                    action.create(nameOfActionThatTriggers, actionThatTriggers)
+            }
+
+            // this is expected to fail this time because we have not yet created the trigger
+            val runReject = wsk.action.invoke(nameOfActionThatTriggers, Map("shouldSucceed" -> false.toJson))
+            withActivation(wsk.activation, runReject) {
+                activation =>
+                    activation.getFieldPath("response", "success") shouldBe Some(false.toJson)
```
bjustin-ibm: @csantanapr This test fails because `triggerPromise.js` is set up to resolve or reject based on the value of the `shouldSucceed` parameter. In this case, `shouldSucceed` is `false` (meaning I expect the trigger to fail because it doesn't exist yet), and the trigger fails (as expected) so `triggerPromise.js` _resolves_ with a message of "Trigger failed as expected." But here you can see the test is expecting the action to fail.

Obviously this is way too confusing 😉 

I will simplify the test actions to always reject when the trigger fails, and always resolve when the trigger succeeds. This will be much simpler and easy to understand, while still allowing me to test all paths.

csantanapr:
PG 113

psuter:
Merged as e3c57b0ddbc2fa811878a9398c483d1613399d40

PR #990, opened 2016-08-02
Prevent invoker from not fetching messages anymore after exceptions o…

ChrBi:
…n commit and increase session timeout #2306

Signed-off-by: Jeremias Werner <JEREWERN@de.ibm.com>

PPG 956 is green

PR #997, opened 2016-08-03
Catch expections in ActivationFeed due to Kafka poll/commit failure and retry (Supersedes #990)

rabbah:
Modify activation feed so that we grab next batch of messages and commit offsets immediately, essentially marking the activation has having satisfied "at most once" semantics (this is the point at which the activation is considered started); if the commit fails, then messages peeked are peeked again on the next poll.

While the commit is synchronous and will block until it completes, at steady state with enough buffering (i.e., maxPipelineDepth > maxPeek), the latency of the commit should be masked.

Add kafka test where we intentionally fail the commit and then confirm recovery.

Supersedes #990.
PPG 960 approved.

### DIFF
--- a/common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
+++ b/common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
@@ -22,6 +22,7 @@ import scala.collection.JavaConversions.iterableAsScalaIterable
 import scala.collection.JavaConversions.seqAsJavaList
 import scala.concurrent.duration.Duration
 import scala.concurrent.duration.DurationInt
+import scala.concurrent.duration.FiniteDuration
 import scala.language.postfixOps
 import scala.util.Failure
 import scala.util.Success
@@ -40,7 +41,9 @@ class KafkaConsumerConnector(
     groupid: String,
     topic: String,
     override val maxPeek: Int = Int.MaxValue,
-    readeos: Boolean = true)
+    readeos: Boolean = true,
+    sessionTimeout: FiniteDuration = 30 seconds,
+    autoCommitInterval: FiniteDuration = 10 seconds)
     extends MessageConsumer
     with Logging
     with TransactionCounter {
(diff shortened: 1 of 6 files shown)


common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -90,9 +93,10 @@ class KafkaConsumerConnector(
         val props = new Properties
         props.put(ConsumerConfig.GROUP_ID_CONFIG, groupid)
         props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, kafkahost)
-        props.put(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG, 30000.toString)
+        props.put(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG, sessionTimeout.toMillis.toString)
+        props.put(ConsumerConfig.HEARTBEAT_INTERVAL_MS_CONFIG, (sessionTimeout.toMillis / 3).toString)
```
ChrBi: The default of this value is 3000 milliseconds? Should we really increase it to 10.000 milliseconds?

core/dispatcher/src/main/scala/whisk/core/dispatcher/ActivationFeed.scala
```diff
@@ -74,16 +76,28 @@ protected class ActivationFeed(
...
+                    // Grab next batch of messages and commit offsets immediately
+                    // essentially marking the activation as having satisfied "at most once"
+                    // semantics (this is the point at which the activation is considered started).
+                    // If the commit fails, then messages peeked are peeked again on the next poll.
+                    // While the commit is synchronous and will block until it completes, at steady
+                    // state with enough buffering (i.e., maxPipelineDepth > maxPeek), the latency
+                    // of the commit should be masked.
+                    val records = consumer.peek(longpollDuration)
+                    consumer.commit()
+                    (records, records.size)
```
markusthoemmes: Why return a tuple, seems redundant?

PR #1001, opened 2016-08-03
Accept query doc as JSON object rather than serialized String

psuter:
Avoids potential double-encoding.


csantanapr:
@psuter What uses this class? Maybe this is a class that doesn't belong in open and it's for internal use?
Was not able to find any references to this code in the repo.

csantanapr:
Only applicable to cloudant? Or it works with CouchDB also?

psuter:
The class is useful if you write a client to retrieve tailored information out of a database.

csantanapr:
@rabbah please review and merge.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/f33437a6f9dc9d775114cc1f269de5cf1c19d54f.

PR #1003, opened 2016-08-04
Flip the catalog installation to the openwhisk-catalog repo

houshengbo:
Currently, all the actions in openwhisk-catalog have been synchronized
with the ones in openwhisk. The test coverage of actions is equal to
openwhisk. We are technically able to switch from openwhisk/catalog to the
openwhisk-catalog.

### DIFF
--- a/tests/dat/actions/cat.js
+++ b/tests/dat/actions/cat.js
@@ -1,0 +1,13 @@
+/**
+ * Equivalent to unix cat command.
+ * Return all the lines in an array. All other fields in the input message are stripped.
+ * @param lines An array of strings.
+ */
+function main(msg) {
+    var lines = msg.lines || [];
+    var retn = {lines: lines, payload: lines.join("\n")};
+    console.log('cat: returning ' + JSON.stringify(retn));
+    return retn;
+}
+
+
--- a/tests/dat/actions/head.js
+++ b/tests/dat/actions/head.js
@@ -1,0 +1,14 @@
+/**
+ * Return the first num lines of an array.
+ * @param lines An array of strings.
+ * @param num Number of lines to return.
+ */
+function main(msg) {
+    var lines = msg.lines || [];
+    var num = msg.num || 1;
+    var head = lines.slice(0, num);
+    console.log('head get first ' + num + ' lines of ' + lines + ': ' + head);
+    return {lines: head, num: num};
+}
+
+
--- a/tests/dat/actions/sort.js
+++ b/tests/dat/actions/sort.js
@@ -1,0 +1,14 @@
+/**
+ * Sort a set of lines.
+ * @param lines An array of strings to sort.
+ */
+function main(msg) {
+    var lines = msg.lines || [];
+    //console.log('sort got ' + lines.length + ' lines');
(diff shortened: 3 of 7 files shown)


houshengbo:
@rabbah @jasonpet @csantanapr 
I think the openwhisk-catalog is ready, this PR can be the first step to switch to openwhisk-catalog. After this one, actions can be removed gradually in openwhisk, and test cases will be changed accordingly.

jasonpet:
@houshengbo - other builds outside of travis run these tests as well.  If you change the package name of the tests from util to utils these others builds, which will not be flipped as part of this PR will fail.  Please contact me via slack if you do not now about the other builds I am referring to.  In my opinion these tests either need to move to the openwhisk-catalog repo or install the actions themselves via files in the tests/dat/actions folder.  That way tests in open would not need to depend on openwhisk-catalog being installed.

jasonpet:
@houshengbo - I created a fix for WskActionSequenceTests.  You can pull my commit:
https://github.com/jasonpet/openwhisk/commit/260e5b346b420a6ab74ae35b2c6655054758d87c

houshengbo:
@jasonpet Thank you. I have pulled the branch and update this PR.
I will continue to fix the WskBasicTests.

houshengbo:
@jasonpet @csantanapr @rabbah 

I found out that test cases tests/src/system/basic/WskBasicTests.scala are different from in  tests/src/system/basic/WskActionSequenceTests.scala. 

I was trying to creating the action before calling the actions themselves, but I did  not succeed.
wskprops uses the guest auth key and "_" as the namespace. If I manually change it into the system auth key and whisk.system as the namespace, it tells me I am unable to create the action, since HTTP only expects "GET"; if I contribute to use the guest auth key and the namespace "_", I can still delete the action even if I change the auth key to something else. I am unable to create the valid test case for 
```
    it should "reject deleting action in shared package not owned by authkey" in {
        wsk.action.get("/whisk.system/util/cat") // make sure it exists
        wsk.action.delete("/whisk.system/util/cat", expectedExitCode = FORBIDDEN)
    }
```
and the similar ones.

I have not found any other ways to do it except changing util to utils, because we change this name in catalog repos. I am not sure if we are OK with this approach so far.

csantanapr:
I don't follow `HTTP only expects "GET"`  ?
when the cut is done, tests should not depend on artifacts (catago/system) to be installed first. 
The openwhisk runtime should not depend on a catalog to be installed to be able to work correctly. 
Looking at the tests cases it look they belong to in openwhisk and not catalog.
You will need to configure the tests to install a dummy/fixture package to be able to run the basic tests that verify that an user can't delete or modify a shared package without having the correct auth key. For the sequences it's just testing that sequences can be created with actions and the the sequence can be executed correctly, also can be fixture actions it's not important the functionality of the actions, but that they can be chained usable by tests.

I would suggest get together with @jasonpet or @bjustin-ibm they are familiar with the test utilities classes, there might be a need to create some helper classes to install/uninstall system packages as fixtures to be use by the test cases.

houshengbo:
Sorry, I messed up this one. No idea why it is not updated after I submit new patches.

PR #1005, opened 2016-08-04
Reduce frequency of posting user activation counts

perryibm:
Change KVReporter's input thunk so that it accepts a count.  Use count to reduce the rate at which user activation counts are written out.  PG 970


rabbah:
Rebased and merged as https://github.com/openwhisk/openwhisk/commit/c493f50c93b1a5560715bba0e77d4dedc6278b08.

PR #1007, opened 2016-08-04
Annotations Must be Parsed as Raw JSON

### LINKED ISSUE #717: Parse annotation values similar to parameters
See https://github.com/openwhisk/openwhisk/pull/706#issuecomment-227319149

Confirm annotation value format(s) is the same as with parameter value formats - both from the CLI and in the payload being sent to the controller.

### LINKED ISSUE #718: Refactor parameter parsing functions
See https://github.com/openwhisk/openwhisk/pull/706#issuecomment-227319149; second bullet

The goal is for the command parsing code to not have to know about the allowable parameter value formats and the expectations of the associated payload.  Hide of of the format sensitive processing inside utility functions.  This will streamline the command processing logic.

### LINKED ISSUE #952: go CLI does not have the expected type for annotation values

mdeuser:
Annotations Must be Parsed as Raw JSON

parameter/annotation `key` input value is a string.  It will be converted into a JSON string doing
`\` and `"` escaping.  If it's not a valid JSON string after that, an error is returned.

parameter/annotation `value` input value is expected to be valid JSON.  If it is not valid JSON,
and it is NOT surrounded by either {} or [], it will be surrounded with double quotes making it a JSON
string.  In all cases, if the final value is still not valid JSON, an error is returned.

Replaces PR #953 
Fixes #952 
Fixes #717 
Fixes #718

### DIFF
--- a/tests/src/system/basic/WskBasicTests.scala
+++ b/tests/src/system/basic/WskBasicTests.scala
@@ -30,6 +30,7 @@ import common.TestUtils
 import common.TestUtils.ANY_ERROR_EXIT
 import common.TestUtils.BAD_REQUEST
 import common.TestUtils.CONFLICT
+import common.TestUtils.ERROR_EXIT
 import common.TestUtils.FORBIDDEN
 import common.TestUtils.NOTALLOWED
 import common.TestUtils.NOT_FOUND
@@ -39,6 +40,7 @@ import common.TestUtils.UNAUTHORIZED
 import common.Wsk
 import common.WskProps
 import common.WskTestHelpers
+import spray.json._
 import spray.json.DefaultJsonProtocol._
 import spray.json.JsObject
 import spray.json.pimpAny
@@ -332,6 +334,7 @@ class WskBasicTests
             }
     }
 
+
     behavior of "Wsk Rule CLI"
 
     it should "create rule, get rule, update rule and list rule" in withAssetCleaner(wskprops) {
@@ -372,5 +375,4 @@ class WskBasicTests
         wsk.namespace.get(expectedExitCode = SUCCESS_EXIT)(WskProps()).
             stdout should include("default")
     }
-
 }
--- a/tests/src/common/Wsk.scala
+++ b/tests/src/common/Wsk.scala
@@ -601,8 +601,7 @@ class WskActivation(override val usePythonCLI: Boolean = false)
                 Try {
                     // strip first line and convert the rest to JsObject
                     assert(stdout.startsWith("ok: got activation"))
-                    val firstNewline = stdout.indexOf("\n")
-                    stdout.substring(firstNewline + 1).parseJson.asJsObject
+                    parseJsonString(stdout)
                 } map {
                     Right(_)
                 } getOrElse Left(s"cannot parse activation from '$stdout'")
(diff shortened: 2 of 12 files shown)


mdeuser:
PG 84 - passed.
PG 92 - in progress (for updated test case with non-ascii)

rabbah:
I suggest we move these tests to `WskBasicCliUsageTests` as they test CLI functional and don't need to run as part of system integration tests.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -166,6 +168,86 @@ class WskBasicTests
...
+            assetHelper.withCleaner(wsk.pkg, name) {
+                (pkg, _) =>
+                    pkg.create(name, annotations = getValidJSONTestArgInput)
+            }
+
+            val stdout = wsk.pkg.get(name).stdout
+            assert(stdout.startsWith(s"ok: got package $name\n"))
+
+            val resJSON = stdout.substring(stdout.indexOf("\n") + 1).parseJson.asJsObject
+            assert(resJSON.fields("annotations") == getValidJSONTestArgOutput)
```
rabbah: I suggest you do not use asserts and instead use scalatest matchers (which we use throughout this suite and others for examples, or http://www.scalatest.org/user_guide/using_matchers). They generate better error messages in general and read better.

e.g. `resJson.fields("annotations") shouldBe(getValidJSONTestArgOutput)` and if the two end up not matching you'll get a nice error messages showing the two sides of the comparison. 

Also since you use this more than once, suggest we make it a utility in `Wsk`:
` +            val resJSON = stdout.substring(stdout.indexOf("\n") + 1).parseJson.asJsObject `
In fact there are at least two cases of this in `Wsk.scala` already to consolidate then.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -166,6 +168,86 @@ class WskBasicTests
...
+        (wp, assetHelper) =>
+            val name = "packageParameters"
+
+            assetHelper.withCleaner(wsk.pkg, name) {
+                (pkg, _) =>
+                    pkg.create(name, parameters = getValidJSONTestArgInput)
+            }
+
+            val stdout = wsk.pkg.get(name).stdout
+            assert(stdout.startsWith(s"ok: got package $name\n"))
```
rabbah: you could move this into `WskPackage` and guard with `if (expectedExitCode == SUCCESS) then assert`.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -166,6 +168,86 @@ class WskBasicTests
...
+
+            val resJSON = stdout.substring(stdout.indexOf("\n") + 1).parseJson.asJsObject
+            assert(resJSON.fields("parameters") == getValidJSONTestArgOutput)
+    }
+
+    it should "not create a package when -a is specified without arguments" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
```
rabbah: use `assetHelper.withCleaner(wsk.pkg, name, confirmDelete = false) { ...` in case test fails, asset cleaner deletes the created entity.

mdeuser:
PG 103 - for test case changes (comments).   PG "passed" (only issue was due to timing related failure regarding activation record accounting in an alarm feed test - not related to the affected tests).

mdeuser:
@rabbah - Ready for merge.  If there are other follow up test case related comments/issues, I'd like to handle those via a separate issue as this fix is blocking the CLI from being beta ready.   Thanks!

Issue #636, opened 2016-06-13
Enhancements for the Rules API


markusthoemmes:
With the latest refactoring of Activator into Controller, much of the groundwork has been done to implement some enhancements for the Rules API.
- [x] Ability to add actions to a rule that are not in the same namespace as the rule itself (including actions in packages, #239)
- [x] Make rules enabled by default
- [ ] List all rules (actions) associated with a trigger (#324)
- [ ] Allow for anonymous rules (#833)

starpit:
i'm working on rules enabled by default

markusthoemmes:
Covered in more precise issues.

PR #1008, opened 2016-08-05
change default rule creation behavior to have status=ACTIVE

starpit:
see #636

PG1 85 approved.

### DIFF
--- a/ansible/README_DISTRIBUTED.md
+++ b/ansible/README_DISTRIBUTED.md
@@ -1,0 +1,98 @@
+Deploying Distributed OpenWhisk using Ansible
+=========
+
+**Caveat:** All Ansible commands are meant to be executed from the `ansible` directory.
+This is important because that's where `ansible.cfg` is located which contains generic settings that are needed for the remaining steps.
+
+By default, if you omit the `-i` parameter in Ansible commands, the `local` environment will be used.
+
+In all instructions, replace `<openwhisk_home>` with the base directory of your OpenWhisk source tree. e.g., `openwhisk`.
+
+Login to your bootstrapper VM. Your local machine can act as the bootstrapping machine as well, if it can connect to the VMs deployed in your IaaS (Infrastructure as a Service platform).
+
+#### Distributed Deployment using OpenStack as IaaS
+
+This installs modules and packages to manage cloud instances via Ansible.
+
+```
+sudo apt-get -y install python-setuptools python-dev libssl-dev
+sudo pip install shade pytz positional appdirs monotonic rfc3986
+sudo apt-get install python-novaclient
+sudo pip install six --upgrade
+```
+If you would like the environment instances and hosts file to be generated and managed by Ansible, set values for the following keys using environment variables. These values can be pulled from the Openstack UI (`https://${openstack_dashboard_url}/project/access_and_security/`) as an [RC](http://docs.openstack.org/user-guide/common/cli-set-environment-variables-using-openstack-rc.html) file.
+
+```
+export OS_FLAVOR=m1.medium
+export OS_IMAGE=Ubuntu14.04-1Nic
+export OS_KEY_NAME=key_name
+export OS_NET_NAME=network_name
+export OS_NET_ID=e489dcf2-4601-4809-a459-e3821a95d23a
+export OS_USERNAME=abcxyz
+export OS_PASSWORD=*******
+export OS_PROJECT_NAME=OpenWhisk
+export OS_SECURITY_GROUPS=sec_group
+
+## Keystone v2
+export OS_AUTH_URL=https://OpenStack_URL:5000/v2.0
+export OS_TENANT_NAME="OpenWhisk"
+export OS_TENANT_ID=a9e6a61ab914455cb4329592d5733325
+
+## Keystone v3
+export OS_AUTH_URL=https://OpenStack_URL:5000/v3
+export OS_PROJECT_ID=a9e6a61ab914455cb4329592d5733325
+export OS_USER_DOMAIN_NAME="domain"
+```
+#### Setup and provision OpenWhisk component VMs
+
+- Set a value for the default ssh user in the defaults section of the `ansible.cfg` file.
+```
+[defaults]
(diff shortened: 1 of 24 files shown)


csantanapr:
@starpit This changes the default state to enable/active when a rule is created right?

This means we need two more things:
1. Update docs (example says that it needs to use --enable so the rule is enable on creation)
https://github.com/openwhisk/openwhisk/blob/master/docs/triggers_rules.md
2. With this change CLI users will not be able to create a rule that is disable anymore, the `--enable` flag on create doesn't make sense anymore, it means a new flag needs to be introduce (i.e. --disable) for `wsk rule create --disable rulename trigger action`

```
$ wsk rule create -h
usage: wsk rule create [-h] [-u AUTH] [--shared [{yes,no}]] [--enable]
                       name trigger action

positional arguments:
  name                  the name of the rule
  trigger               the trigger
  action                the action

optional arguments:
  -h, --help            show this help message and exit
  -u AUTH, --auth AUTH  authorization key
  --shared [{yes,no}]   shared action (default: private)
  --enable              enable rule after creating it
```

cc @mdeuser

starpit:
@rabbah suggests handling the client update separately. How generally do you handle docs changes? Rolled in, or separate PR (I would imagine that docs updates have a distinct set of reviewers, which led me to limit this to the controller change, but am happy to go with the standard practice)

csantanapr:
Yeah that's OK other team members can take care of the doc and go-CLI changes. 

I was trying to note what were the impacts of these PR, and the fact that CLI changes have not being discuss before. 

Any other impacts from this PR?

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -348,14 +348,14 @@ class WskBasicTests
             assetHelper.withCleaner(wsk.rule, ruleName) {
                 (rule, name) =>
                     rule.create(name, trigger = triggerName, action = actionName)
-                    rule.create(name, trigger = triggerName, action = actionName, update = true)
+                    rule.create(name, trigger = triggerName, action = actionName, update = true, expectedExitCode = CONFLICT)
```
markusthoemmes: Why is this supposed to be a conflict?

starpit: now that rules are enabled by default, the second "update" operation will result in a CONFLICT. we have a couple of options here.

1. as i did, perform the update and expect a conflict. this will check that the first (creation) operation indeed enabled the rule

2. add a disable call in-between, and revert the update line to expect success

thoughts?

starpit:
@csantanapr thanks for the help with this! i agree with the need to update the docs and the CLI. the third piece here is the UI. i have already discussed this with the UI team.

mdeuser:
Is there an issue for the required CLI updates?

starpit:
@mdeuser not yet. we should discuss the terminology there. e.g. is it --no-enable? --no-start (this is the terminology of cf)?

csantanapr:
@starpit @mdeuser I created issues #1014 (docs) and #1015 (go-cli)

My proposal for CLI changes are in #1015 
No need for a `--no-enable`, user can just run `wsk rule disable` after creation making things more simple.

tests/src/system/basic/WskBasicTests.scala
markusthoemmes: But why is it a conflict in the first place? Updating has nothing to do with activating/deactivating (Update is PUT, activating/.. is POST).

markusthoemmes:
I'ld suggest to at least have the doc changes in this PR. Having self-contained changes in a PR is pure goodness.

starpit:
but the docs changes refer to the CLI (after all, this is not an API schema updates, this is a semantic update). so by this reasoning, we also need to roll he CLI changes into this PR?

tests/src/system/basic/WskBasicTests.scala
starpit: one cannot update an active sequence, i believe?

markusthoemmes: ah, that makes sense yeah

starpit:
@markusthoemmes i just pushed a change to the docs. for sequential updates like this, do i need to rebase the two commits into one?

csantanapr:
I would preferred to have everything in one PR as general rule (code, tests, doc), but that's just a preference.
If @starpit to include the doc and CLI updates in this PR, that's OK by me.

rabbah:
The API currently ignores the "status" property when creating/updating a rule (PUT). An alternative strategy is to allow PUT to specify the desired state of the rule. The default can still be enabled but this would at least allow one to create a rule in an inactive state which we otherwise can't do after this change.

If we were to do that then --enable in the CLI can be set as the default with a --disable added for symmetry.

starpit:
i'm trying to think of cases where an atomic create-but-disabled is important. if you aren't ready for your rule to be live, then don't create it.

cf's "no-start" option is important for them, because their push command doesn't not accept all of the necessary arguments for starting. e.g. environment variables and service bindings. so you push the app, which then lets you associate env vars and services... then you start it.

i don't see any reason why we would need to have that kind of two-phase startup, for openwhisk?

rabbah:
I haven't come up with a compelling reason.

So if a rule is created always in enabled state, is the required CLI change to just remove `--enable`?

csantanapr:
Yes @rabbah check #1015

tests/src/system/basic/WskBasicTests.scala
markusthoemmes: So i think to preserve what this test is supposed to assert we should implement option 2 you mentioned. Thoughts?

rabbah: As an additional tests, sure (since there's no way to create a test that's disabled).

starpit:
ok, i've updated the WskBasicTests rule test. it now

1. tries to update the rule, and expect CONFLICT
2. disables the rule
3. tries to update the rule, and expect SUCCESS

starpit:
PG1 99 passed

PR #1010, opened 2016-08-05
Remove services:registrator in gradle setting

houshengbo:
Since services:registrator is removed, we need to remove from the
gradle setting. If it is still there, import openwhisk project
as a gradle project will raise an error in eclipse.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -12,7 +12,6 @@ include 'core:javaAction'
 
 include 'services:consul'
 include 'services:kafka'
-include 'services:registrator'
 include 'services:zookeeper'
 
 include 'tools:cli'


rabbah:
Deferring to @markusthoemmes.

markusthoemmes:
LGTM

ioana-blue:
@markusthoemmes I lost track of the story with registrators and consul health. Do we still use registrators but pull some public image?

markusthoemmes:
@ioana-blue yep, on spot. We don't build the image ourselves anymore (we never really did i think). This just removes a dead link.

csantanapr:
merged 8bea78bd9a614516ef51d348af5cacaac0799503

PR #1016, opened 2016-08-05
Add whisk config parameters related to invoker core and core sharing

perryibm:
This is just adding parameters without use yet.  That will come after this when the other PR is adjusted to use this. PG 83


rabbah:
I think these should go in here instead: https://github.com/openwhisk/openwhisk/blob/master/ansible/group_vars/all.

@domdom82 @lehoanganh

rabbah:
Moving back to `wip` to address comment above.

markusthoemmes:
@rabbah is right, these should go to `group_vars/all` to be overridable per environment.

markusthoemmes:
rebased and merged in d3a297d143827eff6a40c46c59e7cf6c455bc738

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -216,6 +216,8 @@ object WhiskConfig extends Logging {
 
     val invokerContainerNetwork = "invoker.container.network"
     val invokerContainerPolicy = "invoker.container.policy"
+    val invokerNumCore = "invoker.numcore"
+    val invokerCoreShare = "invoker.coreshare"
```
markusthoemmes: I think these are not enough, you'll need a "getter" above in that class. Please augment that to the PR actually using these values. Missed it before merging.

PR #1023, opened 2016-08-08
Make image prefix configurable to be able to specify different docker…

markusthoemmes:
… repositories

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/Exec.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Exec.scala
@@ -103,18 +103,18 @@ protected[core] object Exec
     private lazy val b64decoder = Base64.getDecoder()
 
     // The possible values of the JSON 'kind' field.
-    protected[core] val NODEJS   = "nodejs"
-    protected[core] val NODEJS6  = "nodejs:6"
-    protected[core] val PYTHON   = "python"
-    protected[core] val SWIFT    = "swift"
-    protected[core] val SWIFT3   = "swift:3"
-    protected[core] val JAVA     = "java"
+    protected[core] val NODEJS = "nodejs"
+    protected[core] val NODEJS6 = "nodejs:6"
+    protected[core] val PYTHON = "python"
+    protected[core] val SWIFT = "swift"
+    protected[core] val SWIFT3 = "swift:3"
+    protected[core] val JAVA = "java"
     protected[core] val BLACKBOX = "blackbox"
     protected[core] val SEQUENCE = "sequence"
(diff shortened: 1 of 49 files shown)


markusthoemmes:
PG1 218 was good.

rabbah:
What about these [tests](https://github.com/openwhisk/openwhisk/tree/master/tests/src/actionContainers) which assume the prefix is `whisk`.

markusthoemmes:
Thanks for pointing that out @rabbah. Implemented the whole change (gradle + ansible + invoker). The last commit can be removed from the PR.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -103,18 +103,18 @@ protected[core] object Exec
...
-    protected[core] val PYTHON   = "python"
-    protected[core] val SWIFT    = "swift"
-    protected[core] val SWIFT3   = "swift:3"
-    protected[core] val JAVA     = "java"
+    protected[core] val NODEJS = "nodejs"
+    protected[core] val NODEJS6 = "nodejs:6"
+    protected[core] val PYTHON = "python"
+    protected[core] val SWIFT = "swift"
+    protected[core] val SWIFT3 = "swift:3"
+    protected[core] val JAVA = "java"
```
rabbah: now i feel @psuter's pain.

core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -493,13 +493,13 @@ class ContainerPool(
         val network = config.invokerContainerNetwork
         val policy = config.invokerContainerPolicy
         val env = getContainerEnvironment()
-        val pull = !imageName.contains("whisk/")
+        val pull = !imageName.contains(config.dockerImagePrefix + "/")
```
rabbah: since `action: WhiskAction` is in scope, should this be replaced by `action.exec is blackbox` instead?

markusthoemmes: action is not in scope here and can't easily put in scope, as this method is used to run stemcell containers without an action being set.

rabbah: you're right - got fooled by the collapsed lines.

rabbah:
Squashed and Merged as https://github.com/openwhisk/openwhisk/commit/6c65b189a673aefd5bae0a5c5a8d7221856c0109.

PR #1024, opened 2016-08-08
change default rule creation behavior to have status=ACTIVE 

starpit:
reopening this as a new PR (see #1008 for historical discussion) due to some git issues on my part.

This addresses #636, by having rules enabled by default, upon creation. It also updates the tests as needed, along with the documentation.

PG1 99 and 101 passed.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Rules.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Rules.scala
@@ -109,7 +109,7 @@ trait WhiskRulesApi extends WhiskCollectionAPI {
                 val docid = DocId(WhiskEntity.qualifiedName(namespace, name))
                 putEntity(WhiskRule, entityStore, docid, overwrite, update(content) _, () => { create(content, namespace, name) },
                     postProcess = Some { rule: WhiskRule =>
-                        completeAsRuleResponse(rule, Status.INACTIVE)
+                        completeAsRuleResponse(rule, Status.ACTIVE)
                     })
             }
         }
@@ -287,7 +287,7 @@ trait WhiskRulesApi extends WhiskCollectionAPI {
                         content.publish getOrElse false,
                         content.annotations getOrElse Parameters())
 
-                    val triggerLink = ReducedRule(Namespace(actionName), Status.INACTIVE)
+                    val triggerLink = ReducedRule(Namespace(actionName), Status.ACTIVE)
                     val saveRule = WhiskTrigger.put(entityStore, trigger.addRule(Namespace(ruleName), triggerLink)) onComplete {
                         case Success(_) => predicate.success(rule)
                         case Failure(t) => predicate.failure(t)
--- a/docs/triggers_rules.md
+++ b/docs/triggers_rules.md
@@ -104,9 +104,14 @@ As an example, create a rule that calls the hello action whenever a location upd
   $ wsk action update hello hello.js
   ```
 
-3. Create and enable the rule. The three parameters are the name of the rule, the trigger, and the action.
+3. Create the rule. Note that the rule will be enabled upon creation, meaning that it will be immediately available to respond to activations of your trigger. The three parameters are the name of the rule, the trigger, and the action.
   ```
-  $ wsk rule create --enable myRule locationUpdate hello
+  $ wsk rule create myRule locationUpdate hello
+  ```
+
+At any time, you can choose to disable a rule.
+  ```
+  $ wsk rule disable myRule
   ```
 
 4. Fire the locationUpdate trigger. Each time that you fire an event, the hello action is called with the event parameters.
--- a/tests/src/common/Wsk.scala
+++ b/tests/src/common/Wsk.scala
@@ -346,19 +346,13 @@ class WskRule(override val usePythonCLI: Boolean = false)
         annotations: Map[String, JsValue] = Map(),
         shared: Option[Boolean] = None,
         update: Boolean = false,
-        enable: Boolean = false,
         expectedExitCode: Int = SUCCESS_EXIT)(
             implicit wp: WskProps): RunResult = {
         val params = Seq(noun, if (!update) "create" else "update", "--auth", wp.authKey, fqn(name), (trigger), (action)) ++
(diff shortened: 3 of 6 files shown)


docs/triggers_rules.md
```diff
@@ -104,9 +104,14 @@ As an example, create a rule that calls the hello action whenever a location upd
   $ wsk action update hello hello.js
   ```
 
-3. Create and enable the rule. The three parameters are the name of the rule, the trigger, and the action.
+3. Create the rule. Note that the rule will be enabled upon creation, meaning that it will be immediately available to respond to activations of your trigger. The three parameters are the name of the rule, the trigger, and the action.
```
rabbah: will be -> is?

tests/src/common/Wsk.scala
```diff
@@ -354,7 +354,7 @@ class WskRule(override val usePythonCLI: Boolean = false)
             { if (enable) Seq("--enable") else Seq() } ++
             { shared map { s => Seq("--shared", if (s) "yes" else "no") } getOrElse Seq() }
         val result = cli(wp.overrides ++ params, expectedExitCode)
-        assert(result.stdout.contains("ok:"), result)
+        if (expectedExitCode == SUCCESS_EXIT) assert(result.stdout.contains("ok:"), result)
         if (enable) {
```
rabbah: this logic no longer necessary (removing the poll will speed things up).

tests/src/common/Wsk.scala
```diff
@@ -354,7 +354,7 @@ class WskRule(override val usePythonCLI: Boolean = false)
...
             val b = waitfor(() => checkRuleState(name, active = true), totalWait = 30 seconds)
```
rabbah: a comment in then is warranted i think since this is mostly unnecessary now.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -348,6 +348,14 @@ class WskBasicTests
             assetHelper.withCleaner(wsk.rule, ruleName) {
                 (rule, name) =>
                     rule.create(name, trigger = triggerName, action = actionName)
+
+                    // to validate that the rule was created enabled, we do an update and expect CONFLICT
+                    rule.create(name, trigger = triggerName, action = actionName, update = true, expectedExitCode = CONFLICT)
```
rabbah: since these are the "tests", they should be outside the asset cleaner.

rabbah:
This should also be removed: https://github.com/openwhisk/openwhisk/blob/master/tests/src/common/WskCli.java#L475.

tests/src/common/Wsk.scala
rabbah: This should also be removed: https://github.com/openwhisk/openwhisk/blob/master/tests/src/common/WskCli.java#L475.

starpit: i had to add that if guard to make the tests pass. are you saying that upstream/master no longer requires this?

docs/triggers_rules.md
starpit: i think will be is proper here, because we're describing the possible future. i.e. this sentence occurs before the example command line invocation.

tests/src/system/basic/WskBasicTests.scala
starpit: i'm not sure i follow. prior to my change, both tests were within the cleaner, i just added a few more. are you saying that the original code wasn't right, either?

markusthoemmes: I'ld even add: "expecting rule update to fail with conflict, because activated rules shall not be updated"

tests/src/common/Wsk.scala
rabbah: > i had to add that if guard to make the tests pass. are you saying that upstream/master no longer requires this?

Then we have a problem. The predicate and poll loop is waiting for rule to be enabled but create has finished implying the poll is not necessary. Or did I miss something?

starpit: are we talking about the same line: i was talking about my change, i.e. the `if (expectedExitCode == SUCCESS_EXIT)` ? are you talking about the `if (enable)` line (i think you so)?

if so, then yes, i agree we should remove that if (enable) part.

rabbah: Indeed referring to if(enable) {}.

tests/src/system/basic/WskBasicTests.scala
rabbah: The asset cleaner should wrap asset creation. So yes the prev code was also doing it "wrong".

starpit: i don't think we can use expectedExitCode = ANY_EXIT_CODE, because the rule creation logic needs to know whether it should look for "ok" on stdout. either we need to eliminate the stdout check, or we need to (somehow) information rule.create of the expected failure

starpit:
the updates are in. PG1 107 clean.

core/controller/src/main/scala/whisk/core/controller/Rules.scala
```diff
@@ -287,7 +287,7 @@ trait WhiskRulesApi extends WhiskCollectionAPI {
                         content.publish getOrElse false,
                         content.annotations getOrElse Parameters())
 
-                    val triggerLink = ReducedRule(Namespace(actionName), Status.INACTIVE)
+                    val triggerLink = ReducedRule(Namespace(actionName), Status.ACTIVE)
                     val saveRule = WhiskTrigger.put(entityStore, trigger.addRule(Namespace(ruleName), triggerLink)) onComplete {
                         case Success(_) => predicate.success(rule)
                         case Failure(t) => predicate.failure(t)
```
rabbah: conferring with @markusthoemmes we propose changing this line to:

```
-                        case Failure(t) => predicate.failure(t)
+                        case Failure(t) => predicate.failure {
+                            RejectRequest(Conflict, "Rule was created but remains inactive (auto-enable failed, try again).")
+                        }
```

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/f580a0d5dd98b69956165e6e1e2373010a197080.

Issue #1027, opened 2016-08-09
Travis CI constantly fails. on docker run couchdb


houshengbo:
Travis ci raises this error constantly:

```
TASK [couchdb : (re)start CouchDB] *********************************************

Tuesday 09 August 2016  17:48:32 +0000 (0:00:00.085)       0:00:01.090 ******** 

fatal: [172.17.42.1]: FAILED! => {"changed": false, "failed": true, "msg": "ConnectionError(ProtocolError('Connection aborted.', error(111, 'Connection refused')),)"}
```

csantanapr:
sounds like docker-engine problems in travis VM

csantanapr:
Here is one example:
https://travis-ci.org/openwhisk/openwhisk/builds/150994577

csantanapr:
it will be useful to see the version of docker travis is using, we should stick to 1.9.1

markusthoemmes:
The docker version is printed in the log: `Docker version 1.9.1, build a34a1d5`.

We do stick to 1.9.1, docker is setup in `tools/travis/setup.sh`

csantanapr:
We sent email to Travis support to report the problem.

csantanapr:
To stay tune you can create a github trigger in OpenWhisk 😸

csantanapr:
Closing with PR #1030

PR #1030, opened 2016-08-10
Adapt docker installation to new travis images

markusthoemmes:
- Remove travis environment, no longer needed

Signed-off-by: Dominik Jall <djall@de.ibm.com>


tools/travis/setup.sh
```diff
@@ -1,15 +1,13 @@
 #!/bin/bash
 
+sudo gpasswd -a travis docker
+sudo -E bash -c 'echo '\''DOCKER_OPTS="-H tcp://0.0.0.0:4243 -H unix:///var/run/docker.sock --api-enable-cors --storage-driver=aufs"'\'' > /etc/default/docker'
+
 # Docker
 sudo apt-get -y update -qq
-sudo apt-get -o Dpkg::Options::="--force-confnew" --force-yes -y install docker-engine=1.9.1-0~trusty
+sudo apt-get -o Dpkg::Options::="--force-confold" --force-yes -y install docker-engine=1.9.1-0~trusty
```
csantanapr: Is the only  fix is `--force-confold`?  Or also we need to use `environment/local` also on top of it?

tools/travis/build.sh
```diff
@@ -11,7 +11,7 @@ tools/build/scanCode.py .
 
 cd $ROOTDIR/ansible
 
-ANSIBLE_CMD="ansible-playbook -i environments/travis"
+ANSIBLE_CMD="ansible-playbook -i environments/local"
```
csantanapr: Is this required, or we changed our minds to re-use `local` for travis also?

As a side note, I think using "local" as the name of the environment now becomes even more vague.
I think we should rename it to `environment/ubuntu-single-box`

markusthoemmes: `travis` environment is no longer needed. Travis changed the docker gateway to be `172.17.0.1` so I figured, having two identical environments feels wrong.

`local` simply means: install to the local docker daemon which i think is fine.

tools/travis/setup.sh
markusthoemmes: Reordering, `--force-confold` and using a different gateway address was the fix. All is needed.

csantanapr:
Fixes #1027

tools/travis/build.sh
csantanapr: Ah good to know, it would be good in the future to dynamically get this value, or maybe find a way to just use localhost with any port required expose on localhost.

The PR is green in Travis, so no more comments, PR is ready

csantanapr:
rebase and merge with 0a4b14a9ee2dadae62ac4fd79fe23841b6d5165c

PR #1031, opened 2016-08-10
add option to print verbose CLI output to stdout during tests.

domdom82:
this will greatly increase verbosity of CLI tests if needed. currently applied to CLIRuleTests only for debugging. can be turned on for other tests as well. we could also pass it in as a gradle property on demand.

### DIFF
--- a/tests/src/common/TestUtils.java
+++ b/tests/src/common/TestUtils.java
@@ -328,7 +328,9 @@ public class TestUtils {
 
         int exitCode = bl.launch(timeoutMilli);
         RunResult rr = new RunResult(exitCode, new String(bl.getStdout()), new String(bl.getStderr()));
-        logger.info("RunResult: " + rr);
+        if (logger != null) {
+            logger.info("RunResult: " + rr);
+        }
         rr.validateExitCode(expectedExitCode);
         return rr;
     }
--- a/tests/src/common/TestUtils.java
+++ b/tests/src/common/TestUtils.java
@@ -72,7 +72,7 @@ public class TestUtils {
 
 
     static {
-        logger.setLevel(Level.WARNING);
+        logger.setLevel(Level.INFO);
         System.out.println("JUnit version is: " + Version.id());
     }
 
@@ -328,6 +328,7 @@ public class TestUtils {
 
         int exitCode = bl.launch(timeoutMilli);
         RunResult rr = new RunResult(exitCode, new String(bl.getStdout()), new String(bl.getStderr()));
+        logger.info("RunResult: " + rr);
         rr.validateExitCode(expectedExitCode);
         return rr;
     }


tests/src/system/basic/CLIRuleTests.java
```diff
@@ -65,9 +66,9 @@ public static void setUp() throws Exception {
     @Test
     public void rule1to1() throws Exception {
         try {
-        	wsk.sanitize(Rule, "R_121");
-        	wsk.sanitize(Action, "A_121");
-        	wsk.sanitize(Trigger, "T_121");
+            wsk.sanitize(Rule, "R_121");
```
rabbah: format changed incorrectly.

markusthoemmes: are there tabs in the current code?

tests/src/common/WskCli.java
```diff
@@ -976,7 +982,14 @@ public RunResult cli(boolean verbose, int expectedExitCode, String... params) th
     public RunResult cli(boolean verbose, int expectedExitCode, File workingDir, String... params) throws IllegalArgumentException, IOException {
         String[] cmd = verbose ? Util.concat(baseCmd, "--verbose") : baseCmd;
         String[] args = Util.concat(cmd, "-i", "--apihost", WhiskProperties.getEdgeHost());
-        RunResult rr = TestUtils.runCmd(DONTCARE_EXIT, workingDir, TestUtils.logger, this.env, Util.concat(args, params));
+        String[] finalArgs = Util.concat(args, params);
+        if (this.verbose == true) {
+            System.out.println("CLI RUNNING WITH ARGS:\n" + Arrays.toString(finalArgs));
```
rabbah: this you can get by setting the `logger` verbosity to `info` vs the default `warning` level.

tests/src/common/WskCli.java
```diff
@@ -976,7 +982,14 @@ public RunResult cli(boolean verbose, int expectedExitCode, String... params) th
...
+        }
+        RunResult rr = TestUtils.runCmd(DONTCARE_EXIT, workingDir, TestUtils.logger, this.env, finalArgs);
+        if (this.verbose == true) {
+            System.out.println("CLI RUN RESULT:\n" + rr);
```
rabbah: consider printing to `logger.debug` instead.

tests/src/system/basic/CLIRuleTests.java
rabbah: surely if there are tabs now, either we're not running scan code correctly, or there are no tabs.

rabbah: edited my comment.

rabbah:
I think modifying the `assert` is probably better (I normalized the ones in CLIActionTests as follows:

```
                assertTrue("Expected '" + exp + "' which is missing in log for activation " + activationId, present);
```

so that you always have the activation ID. I think this is likely sufficient to do for CLIRulesTests instead of the verbose flag.

csantanapr:
rebase to make Travis happy again related to #1030

domdom82:
@rabbah The tests did get an activation id but they actually check the log for a string with `logsForActivationContain` which made the assertion fail. The first assertion which got the activation id passed. So in this case only getting the activation id would not have helped.

rabbah:
From the jenkins job output, you can't tell what the activation id - usually just knowing this is enough to RCA.

domdom82:
made this way simpler by tweaking log levels as @rabbah recommended. also added a log statement to log the CLI result, which is enough to get all you need for debugging.

domdom82:
Apparently, logger can be null..?
```
system.rest.JsonSchemaTests > JSON schema validator should accept a correct object FAILED
    java.lang.NullPointerException
        at common.TestUtils.runCmd(TestUtils.java:331)
        at common.TestUtils.runCmd(TestUtils.java:307)
        at common.TestUtils.runQuietly(TestUtils.java:299)
```

Adding a null check.

domdom82:
@rabbah pls let me know when LGTM so I can squash and rebase.

rabbah:
I'll do both - no worries. We're accepting that this will make our test outputs more verbose, yes?

rabbah:
Is there a PG?

domdom82:
https://openwhisk-jenkins.swg-devops.com:8443/job/Playground1/135/

I'd rather accept more log output and disk usage than not knowing what went wrong when I get paged :-)

rabbah:
Will merge when PG approves.

domdom82:
PG failed while pulling swift image. weird.

Relaunched https://openwhisk-jenkins.swg-devops.com:8443/job/Playground1/137/

ChrBi:
Merged in 05fd5f4

PR #1033, opened 2016-08-10
Make task print when its called

markusthoemmes:
omitting `<<` in a plain task causes the statements to be executed during gradles initialize phase as opposed to when its finally called


csantanapr:
LGTM

csantanapr:
rebase and merge dfd1e4e5b6eca8889d87eaf09395c581ab11670e

PR #1038, opened 2016-08-11
One proxy for swift(3), python and docker actions

### LINKED ISSUE #518: proxy.py in Swift runtime does not always flush output streams
It has been seen that some activations do not have complete `logs`, even though the activations were successful.  Additionally, this behavior is not consistent, and causes issues for unit tests that examine the activation logs.

This seems to be because `proxy.py` may not be fulling flushing what it prints to `stderr` and `stdout`.

### LINKED ISSUE #661: Language runtimes need to send a sentinel marker to mark the end of an activation

### LINKED ISSUE #691: Consolidate web server used in built-in runtimes where applicable
The current state of affairs is that each runtime has it's own copy of a web server that is used to handle the `/init` and `/run` calls. To make matters worse, this same functionality has been implemented separately in NodeJS, Python, and Java.

Let's take the time to consolidate this common functionality so it can be re-used among all the built-in runtimes.  As discussed in #690, this could be accomplished by publishing a common Docker base image not only for all built-in runtimes, but also for users who need their own custom environments.

### LINKED ISSUE #982: Remove limitation of 1MB code for JS-actions
https://github.com/openwhisk/openwhisk/blob/master/core/nodejsActionBase/app.js#L36

There's a limit, that prohibits actions that are greater than 1MB. Since we check the size of actions on creating them (and have a limit of 48MB) this limit is obsolete an can be removed.

### LINKED ISSUE #1012: all (console) output from a docker action is assumed to be part of the action result
Looking at the docker action proxy, it considers all output produced on stdout and stderr as part of the action result:

```
proc.stdout.on('data', function(data) { output += data; });
proc.stderr.on('data', function(data) { output += data; });
var result = JSON.parse(output);
```

This prevents a docker action from having both logs and a result. Instead, the proxy should perhaps use the same convention as with Swift actions - the last line of the output is treated as the result and everything else is a log message.

rabbah:
Adds "action proxy" a new base image for actions including swift and "black box" using a common proxy server coded in python and using flask. The proxy implements the required init and run routes. The container include a sample shell script making it useful as a stand alone example action. It also allows arbitrary shell scripts (python, perl, bash, etc.) to run by initializing the internal action code on init. The run methods always prints the required sentinels, making actions that inherit from this proxy compliant with the invoker's expectations for extracting user log streams.
    
Elaborates bash example with comments and showing logging as well as receiving action input.
Rebase swift/3 and python to use canonical image (which also becomes the docker skeleton).
Add Python actions to set of sentinelled actions for which the invoker will wait for marker while fetching logs.
    
Fixes #1012.
Fixes #691.
Fixes #661.
Fixes #518.
Fixed #982.

@psuter @ioana-blue @bjustin-ibm @perryibm for your review. The commits are grouped and are progressive, so that it's easier to process.


core/actionProxy/README.md
```diff
@@ -12,17 +12,21 @@ The script `buildAndPush.sh` is provided for your convenience. The following com
...
+
 # invoke created action
-> wsk action invoke dockerSkeletonExample --blocking
+wsk action invoke dockerSkeletonExample --blocking
 ```
 
-The binary to execute must be located in the `client` folder. The name of the binary
-(or shell script as in the standard bundle) must be `action`.
+The binary to execute must be located in the `/action` folder. The name of the binary
+(or shell script as in the standard bundle) must be `/action/a.out`.
```
bjustin-ibm: To my brain `a.out` means "binary output from a compiler". This name may be confusing despite the comment saying that shell scripts are also allowed. Perhaps a more generic name like `/action/exec` would help?

Also, wouldn't any executable file be acceptable? It seems like calling out only shell scripts may seem to exclude things like Python scripts.

How about:
> The executable file must be located in the `/acton` folder. The name of the executable must be `/action/exec` and can be any file with executable permissions.

bjustin-ibm:
Are `core/actionProxy/init.py` and `core/actionProxy/run.py` needed?

rabbah:
Added https://github.com/openwhisk/openwhisk/pull/1038/commits/da8a138bed947bc94240aac859656b7a9d5a6105 to address feedback from @bjustin-ibm.

rabbah:
PG1-158 approved.

bjustin-ibm:
@rabbah As far as I can tell, the code changes seem solid. However, what seems to be missing is documentation describing this new system. It seems that individual action runners have certain obligations and opportunities to customize the way their container runs by implementing one or more of the lifecycle functions defined by the base ActionRunner (e.g. `epilogue`, `verify`, `build`, etc.). However, I don't see any documentation describing these aspects.

I think it would be valuable to have either comments in `actionproxy.py` describing the lifecycle and customization points or, perhaps better, a separate document describing how to create a new action container.

core/dispatcher/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -318,10 +319,10 @@ class Invoker(
     private val LogRetryCount = 15
     private val LogRetry = 100 // millis
     private val LOG_ACTIVATION_SENTINEL = "XXX_THE_END_OF_A_WHISK_ACTIVATION_XXX"
-    private val nodejsImageName = WhiskAction.containerImageName(NodeJSExec("", None), config.dockerRegistry, config.dockerImageTag)
-    private val nodejs6ImageName = WhiskAction.containerImageName(NodeJS6Exec("", None), config.dockerRegistry, config.dockerImageTag)
-    private val swiftImageName = WhiskAction.containerImageName(SwiftExec(""), config.dockerRegistry, config.dockerImageTag)
-    private val swift3ImageName = WhiskAction.containerImageName(Swift3Exec(""), config.dockerRegistry, config.dockerImageTag)
+    private val sentinelledActions = {
```
psuter: The creation of this set looks like a big hack to me. Can we replace by a `def` that checks the type and use that instead of `.contains` below?

tests/src/actionContainers/ActionProxyContainerTests.scala
```diff
@@ -0,0 +1,174 @@
...
+
+    def withPythonContainer(code: ActionContainer => Unit) = withContainer("whisk/dockerskeleton")(code)
+
+    behavior of "whisk/dockerskeleton"
+
+    it should "run sample without init" in {
+        val (out, err) = withPythonContainer { c =>
+            val (runCode, out) = c.run(JsNull)
+            runCode should be(200)
+            out should be(Some(JsObject("msg" -> JsString("Hello from bash script!"))))
```
psuter: Not sure why we'd want to consider that default behavior an important unit test?

tests/src/actionContainers/ActionProxyContainerTests.scala
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        checkStreams(out, err, {
+            case (o, _) => o should include("Actions may log to stdout")
+        })
+    }
+
+    it should "run sample with init that does nothing" in {
```
psuter: What's the justification for this not being an error but falling back to the default case?

tests/src/actionContainers/ActionProxyContainerTests.scala
rabbah: You mean specifically why check the result object instead of the status code itself?
Partly my rationale is that this image will end up on dockerhub and is the built in same.
I'm agreeable to removing line 40 if that's what you mean.

tests/src/actionContainers/ActionProxyContainerTests.scala
rabbah: I'll frame this differently: should the skeleton have a built in sample as the default action exec to run or not? If the latter, then these become errors.

psuter: There are two points here (and in the related comment above):

- should the image have a default, built-in action? I would have thought not, but I can see that perhaps the goal is to make errors user-friendly when someone extends the image but, e.g., puts their executable at the wrong location? In that case, the message should perhaps be "this is the default action", or something in that vein? (see for instance the Apache 2 default page that includes instructions on how to make it go away)

- should POSTing `null` to `/init` be valid/invalid/undefined? I don't see why we would want a test that makes it valid. I see that the comments in the proxy say this will be detected later, but falling back to the default action will just mask the error in a strange way IMHO.

rabbah: OK - perhaps i went a bit too far in consolidating the base docker image and the docker skeleton for dockerhub/example. The latter could extend the base image and add the default action (for which there will be an independent test). In that case I'd accept your critique and suggestions and will address accordingly.

rabbah:
@psuter Addressed your comment re BIG hack here: https://github.com/openwhisk/openwhisk/pull/1038/commits/1d1fc6bee561c2c7b58afa173df4b67be3f0a933.

rabbah:
@bjustin-ibm addressed your request here: https://github.com/openwhisk/openwhisk/pull/1038/commits/e3063b67476c57f5b41cfbe158f614778de8fdca.

rabbah:
@psuter replaced "default" stub here: https://github.com/openwhisk/openwhisk/pull/1038/commits/d4147d11e4aaabae81a36b0350cd21c53dcb4877. Made the docker example self-contained and a derived image.

bjustin-ibm:
@rabbah I made some minor edits to the README and opened a PR directly to your fork here: https://github.com/rabbah/openwhisk/pull/1

rabbah:
Adding tests to confirm actions have expected environment available. Also going to refactor some of the common tests to reduce clone and own.

rabbah:
Commits reorganized and squashed. PG1-233 pending.

rabbah:
Commits reorganized and squashed. PG1-238 approved.

rabbah:
Need to update https://github.com/openwhisk/openwhisk/blob/master/docs/actions.md#creating-docker-actions

PR #1040, opened 2016-08-11
Fix concurrent user activation throttler

markusthoemmes:
- ConsulClient failed silently and with the wrong exception
- A failure in one call (not even related to the other calls) tore down the entire update logic

### DIFF
--- a/core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
+++ b/core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
@@ -68,6 +68,12 @@ protected[core] case class Resource(
 }
 
 protected[core] object EntitlementService {
+    val requiredProperties = WhiskConfig.consulServer ++ Map(
+        WhiskConfig.actionInvokePerMinuteLimit -> null,
+        WhiskConfig.actionInvokePerHourLimit -> null,
+        WhiskConfig.triggerFirePerMinuteLimit -> null,
+        WhiskConfig.actionInvokeConcurrentLimit -> null,
+        WhiskConfig.triggerFirePerHourLimit -> null)
     /**
      * The default list of namespaces for a subject.
      */
@@ -83,8 +89,9 @@ protected[core] abstract class EntitlementService(config: WhiskConfig)(
     private implicit val executionContext = actorSystem.dispatcher
 
     private var loadbalancerOverload: Option[Boolean] = None
-    private val invokeRateThrottler = new RateThrottler(config, 120, 3600)
-    private val triggerRateThrottler = new RateThrottler(config, 60, 720)
+    private val invokeRateThrottler = new RateThrottler(config.actionInvokePerMinuteLimit.toInt, config.actionInvokePerHourLimit.toInt)
+    private val triggerRateThrottler = new RateThrottler(config.triggerFirePerMinuteLimit.toInt, config.triggerFirePerHourLimit.toInt)
+    private val concurrentInvokeThrottler = new ActivationThrottler(config.consulServer, config.actionInvokeConcurrentLimit.toInt)
 
     /** query the KV store this often */
     private val overloadCheckPeriod = 10.seconds
@@ -163,6 +170,8 @@ protected[core] abstract class EntitlementService(config: WhiskConfig)(
     protected[core] def check(subject: Subject, right: Privilege, resource: Resource)(implicit transid: TransactionId): Future[Boolean] = {
         checkSystemOverload(subject, right, resource) orElse {
             checkUserThrottle(subject, right, resource)
+        } orElse {
+            checkConcurrentUserThrottle(subject, right, resource)
         } getOrElse {
             val promise = Promise[Boolean]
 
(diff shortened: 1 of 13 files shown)


markusthoemmes:
Note to self: Try to optimize the numbers of action invocations by stopping as soon as they get throttled.

core/controller/src/main/scala/whisk/core/loadBalancer/ActivationThrottle.scala
```diff
@@ -120,22 +120,25 @@ class ActivationThrottle(config: WhiskConfig)(
...
-        for {
-            concurrencyLimits <- getConcurrencyLimits
+        // Fall back to the current version of the limits
+        val concurrency = getConcurrencyLimits.recover {
+            case _ => userActivationLimits
+        }.map { concurrencyLimits =>
+            userActivationLimits = concurrencyLimits
+        }
+
+        val activations = for {
```
psuter: No fallback for this case?

markusthoemmes: If one of both calls fail, the calculation would be invalid anyway.

Your comment made me notice though, that above "fallback" is unnecessary. Removing.

markusthoemmes:
@perryibm please take a look at the test specifically

perryibm:
3 suggestions:

1) Move the throttle window `1.minute` to a constant and group it with the 60.   The time window of 1 minute and 60 firings are whisk limit properties and these can be adjusted as needed if we start changing the way we throttle.
2) The parameter maximumInvoke of 120 is really a testing parameter and perhaps shoud be restated as 2*maximumFirings to show that we decide to go at twice the throttle rate to trigger it.
3) Once something is throttled, it would be great if you could stop the Future's that have not been fired from occurring.  One way would be through a condition variable but I leave it to you to do it or not.  It would be good to know how long this test takes on average to complete.  In a perfect world, it would be 1 minute.  When one gets unlucky, it would be 2 minutes if we coded things well.  Since you've gone to the trouble if optimizing, can you indicate what you usually get?

markusthoemmes:
Thanks @perry, i'll have a look at the constants.

To 3: Already did that, check the last commit. It stops as soon as a request is throttled.

rabbah:
Can we lift the rate limits out of the throttle and parameterize this test accordingly --- If we change the limits upward, will this test catch the lack of throttling?

markusthoemmes:
@perryibm in the tests take somewhere between 4 to 5 minutes.

rabbah:
Can we move these bits from LoadBalancerToKafka to the RateThorttle?

```
                val userCount = activationThrottle.countForNamespace(subject)
                val userLimit = activationThrottle.limitForNamespace(subject)
                info(this, s"(DoS) current activation count for '$subject': $userCount (limit=$userLimit)")
                if (userCount > userLimit) {
                    info(this, s"(DoS) '$subject' maxed concurrent invocations")
                    Future.successful(throttleError)
```

tests/src/limits/ThrottleTests.scala
```diff
@@ -0,0 +1,200 @@
...
+
+    /**
+     * Extracts the number of throttled results from a sequence of <code>RunResult</code>
+     *
+     * @param results the sequence of results
+     * @param message the message to determine the type of throttling
+     */
+    def throttledActivations(results: List[RunResult], message: String) = results.map { result =>
+        result.stderr
+    }.count(_.contains(message))
```
markusthoemmes: Memo to self: Rewrite as `results.count(_.stderr.contains(message))`

tests/src/limits/ThrottleTests.scala
```diff
@@ -0,0 +1,167 @@
...
+            }
+
+            // invokes per minute * 2 because the current minute could advance which resets the throttle
+            val results = List.fill(maximumInvokesPerMinute * 2 + 1)("testWord").par.map { payload =>
+                wsk.action.invoke(name, Map("payload" -> payload.toJson), expectedExitCode = DONTCARE_EXIT)
+            }
+            val afterInvokes = Instant.now
+
+            waitForActivations(results)
+            throttledActivations(results, rateMessage) should be > 0
```
rabbah: why not check the exit code is "throttled"?

tests/src/limits/ThrottleTests.scala
```diff
@@ -0,0 +1,167 @@
...
+
+    /**
+     * Extracts the number of throttled results from a sequence of <code>RunResult</code>
+     *
+     * @param results the sequence of results
+     * @param message the message to determine the type of throttling
+     */
+    def throttledActivations(results: ParSeq[RunResult], message: String) = results.map { result =>
+        result.stderr
+    }.count(_.contains(message))
```
rabbah: contains message && exit code is "throttled" for completeness?

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -90,6 +91,7 @@ protected[core] abstract class EntitlementService(config: WhiskConfig)(
     private var loadbalancerOverload: Option[Boolean] = None
     private val invokeRateThrottler = new RateThrottler(config.actionInvokePerMinuteLimit.toInt, config.actionInvokePerHourLimit.toInt)
     private val triggerRateThrottler = new RateThrottler(config.triggerFirePerMinuteLimit.toInt, config.triggerFirePerHourLimit.toInt)
+    private val concurrentInvokeThrottler = new ActivationThrottler(config.consulServer, config.actionInvokeConcurrentLimit.toLong)
```
rabbah: long vs int?

markusthoemmes:
@rabbah moved the code into the entitlement package, factored in the "long vs int" comment. Squashed to what i think are the 3 orthogonal bits in this PR.

markusthoemmes:
PG1 219 is good.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/compare/e5cf3c1806b8...2895a9626c84.

PR #1046, opened 2016-08-12
Fix logging bits

### LINKED ISSUE #1044: adjust ERROR log levels in core components
Adjust log level on the following log messages from ERROR to INFO:

- [x] `[ERROR][CouchDbRestStore] [marker:database_getDocument_error] [GET] failure: 'requirement failed: document type class whisk.core.entity.WhiskTrigger did not match expected type class whisk.core.entity.WhiskAction.'`

this is a user error that arises when retrieving an entity with the wrong type.

`[ERROR] [KafkaProducerConnector] sending message on topic '...' failed: org.apache.kafka.common.errors.RecordTooLargeException: The message is xxx bytes when serialized which is larger than the maximum request size you have configured with the max.request.size configuration.` (see https://github.com/openwhisk/openwhisk/issues/1044#issuecomment-239485437)

this can arise if sending a large payload during invoke - should adjust the log level but also short circuit the post and reject earlier in the route.

- [x] `[ERROR] [RemoteEntitlementService] [marker:controller_checkEntitlement_error] failed while checking entitlement: Status: 401 Unauthorized`
- [x] `[ERROR] [NamespacesApi] [GET] namespaces failed: Status: 401 Unauthorized`
- [x] `[ERROR][PackagesApi] [LIST] namespaces lookup failed: Status: 401 Unauthorized`
- [x] `[ERROR] [PackageCollection] entitlement check on package failed: not found on 'get'`

authorization errors should be at INFO level.

- [x] `[ERROR] [RulesApi] [POST] rule update failed: not found on 'get'`
- [x] `[ERROR] [ActionsApi] [DEL] entity failed: Unexpected http response code: 409 Conflict`
- [x] `[ERROR] [ActionsApi] [POST] action activation failed: org.apache.kafka.common.errors.RecordTooLargeException:`

these are likely user error (e.g., deleting while updating it) and should be at INFO level.

### LINKED ISSUE #1045: Controller logs not fully following format
The controller has some logs like these:

```
sslConfig.config.loose.disableSNI = false
```

They don't follow our log format. We should either silence them or adjust them to fit the layout.

markusthoemmes:
- Adding documentation for Logging trait
- Removing unnecessary part of the log
- Getting rid of newlines in log messages
- Updating akka to get rid of an unnecessary log message on stdout
- Fix Package error message
- Surface proper message when rule update failed due to trigger absence
- Adding new exception type, fixing entity conformance logs
- Report invoke being too big properly, needs to be short-circuited
- Refactoring remote entitlement service
   - more dense code
   - adjust log levels related to entitlement
   - surface errors reported by the remote entitlement service
- Fix loglevel of conflicts on delete
- Fix loglevel for trigger fire iff action was deleted

Fixes #1045 
Fixes #1044

Every commit fixes exactly one thing for easier reviewability.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/ApiUtils.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/ApiUtils.scala
@@ -71,13 +71,13 @@ protected sealed trait Messages {
 }
 
 /** An exception to throw inside a Predicate future. */
-protected[controller] case class RejectRequest(code: ClientError, message: Option[ErrorResponse]) extends Throwable
+protected[core] case class RejectRequest(code: ClientError, message: Option[ErrorResponse]) extends Throwable
 
-protected[controller] object RejectRequest {
-    protected[controller] def apply(code: ClientError, m: String)(implicit transid: TransactionId): RejectRequest = {
+protected[core] object RejectRequest {
+    protected[core] def apply(code: ClientError, m: String)(implicit transid: TransactionId): RejectRequest = {
         RejectRequest(code, Some(ErrorResponse(m, transid)))
     }
-    protected[controller] def apply(code: ClientError, t: Throwable)(implicit transid: TransactionId): RejectRequest = {
+    protected[core] def apply(code: ClientError, t: Throwable)(implicit transid: TransactionId): RejectRequest = {
         val reason = t.getMessage
         RejectRequest(code, if (reason != null) reason else "Rejected")
     }
--- a/core/controller/src/main/scala/whisk/core/controller/ApiUtils.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/ApiUtils.scala
@@ -57,6 +57,7 @@ import whisk.http.ErrorResponse
 import whisk.http.ErrorResponse.{ terminate }
 import spray.http.StatusCodes
 import spray.json.JsBoolean
+import whisk.core.database.DocumentTypeMismatchException
 
 protected sealed trait Messages {
     /** Standard message for reporting resource conflicts */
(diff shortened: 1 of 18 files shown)


markusthoemmes:
Memo to self: Implement tests for:
- Rule activation after deleting the trigger associated should surface a 404 "Only rules with existing triggers can be activated"
- GET on an action/trigger/rule should return a 409 Conflict if the name is tied to an entity of different type
- POST to action/trigger with a payload being too big should result in 413

core/controller/src/main/scala/whisk/core/controller/ApiUtils.scala
```diff
@@ -174,7 +175,7 @@ trait ReadOps extends Directives with Messages with Logging {
             case Failure(t: NoDocumentException) =>
                 info(this, s"[GET] entity does not exist")
                 terminate(NotFound)
-            case Failure(t: IllegalArgumentException) =>
```
rabbah: some of the operations e.g. `get` may still throw illegal argument exception because of `requires` in their constructors.

core/controller/src/main/scala/whisk/core/controller/AuthorizedRouteDispatcher.scala
```diff
@@ -85,8 +85,8 @@ trait BasicAuthorizedRouteProvider extends Directives with Logging {
         next: () => RequestContext => Unit)(
             implicit transid: TransactionId): RequestContext => Unit = {
         onComplete(entitlementService.check(user, right, resource)) {
-            case Success(entitment) =>
```
rabbah: doh.

core/controller/src/main/scala/whisk/core/controller/ApiUtils.scala
markusthoemmes: It'll fall through to the `t: Throwable`

markusthoemmes:
PG1 174 was green

PR #1048, opened 2016-08-12
Enabling strict scala compilation

markusthoemmes:
Many "bad habits" can be caught by the scala compiler. This embraces that.

- make warnings fail compilation
- make unused-imports fail compilation
- make dead-code fail compilation


csantanapr:
@markusthoemmes I was thinking doing something similar for all our javascript in our repo (i.e. using a javascript linter as a small step to start). What do you think?

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -163,8 +158,6 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](
                     // FIXME remove mutability from appropriate classes now that it is no longer required by GSON.
                     deserialized.asInstanceOf[WhiskDocument].revision(DocRevision(responseRev))
 
-                    deserialized
```
markusthoemmes: @psuter can you confirm this is equivalent to what it was before?

markusthoemmes:
@csantanapr yes please, that is pure goodness i believe.

common/scala/build.gradle
```diff
@@ -32,3 +32,7 @@ dependencies {
     compile 'org.apache.kafka:kafka-clients:0.10.0.0'
     compile 'org.apache.httpcomponents:httpclient:4.4.1'
 }
+
+tasks.withType(ScalaCompile) {
+    scalaCompileOptions.additionalParameters = ['-feature', '-deprecation', '-Xfatal-warnings', '-Ywarn-unused-import']
```
psuter: My 2¢:
  * add `-unchecked`
  * we use `postfixOps` so much that I would add `-language:postfixOps`, or the imports are going to get annoying very fast

common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala
```diff
@@ -77,11 +74,11 @@ protected[core] object ActivationResponse extends DefaultJsonProtocol {
...
-    protected[core] def applicationError(errorMsg: String)    = error(ApplicationError, JsString(errorMsg))
-    protected[core] def containerError(errorValue: JsValue)   = error(ContainerError, errorValue)
-    protected[core] def containerError(errorMsg: String)      = error(ContainerError, JsString(errorMsg))
-    protected[core] def whiskError(errorValue: JsValue)       = error(WhiskError, errorValue)
-    protected[core] def whiskError(errorMsg: String)          = error(WhiskError, JsString(errorMsg))
+    protected[core] def applicationError(errorMsg: String) = error(ApplicationError, JsString(errorMsg))
+    protected[core] def containerError(errorValue: JsValue) = error(ContainerError, errorValue)
+    protected[core] def containerError(errorMsg: String) = error(ContainerError, JsString(errorMsg))
+    protected[core] def whiskError(errorValue: JsValue) = error(WhiskError, errorValue)
+    protected[core] def whiskError(errorMsg: String) = error(WhiskError, JsString(errorMsg))
```
psuter: FWIW I find this much less readable that previously (and similarly for other whitespace refactorings in this commit). Can we convince the Eclipse formatter that sometimes we do know better?

tests/src/actionContainers/JavaActionContainerTests.scala
```diff
@@ -238,32 +236,29 @@ object JarBuilder {
     import java.util.Base64;
```
psuter: I can't believe you let that `;` get away ;)

common/scala/src/main/scala/whisk/core/entity/WhiskStore.scala
```diff
@@ -243,10 +243,8 @@ object WhiskEntityQueries {
         upto: Option[Instant] = None,
         convert: Option[JsObject => Try[T]])(
             implicit transid: TransactionId): Future[Either[List[JsObject], List[T]]] = {
-        implicit val ec = db.executionContext
         val startKey = List(since map { _.toEpochMilli } getOrElse 0)
         val endKey = List(upto map { _.toEpochMilli } getOrElse TOP, TOP)
-        val includeDocs = convert.isDefined
```
psuter: This is slightly worrisome? @rabbah may want to take a look.

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
psuter: It is, but it's now IMHO less readable than before -- logically, first line applies a side effect, second line returns.

Case in point: you couldn't tell.

common/scala/build.gradle
rabbah: if postfixOps is added, shall we nuke all the imports then?
@markusthoemmes will these compiler settings end up in the eclipse project setting also?

common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala
rabbah: there is an setting that gets the editor to align some tokens (e.g., `=>`), maybe there's another for `=` --- worth checking.

tests/src/actionContainers/JavaActionContainerTests.scala
rabbah: the horror!

common/scala/src/main/scala/whisk/core/entity/WhiskStore.scala
rabbah: This probably came up as an `unused variable` - it use to replace `convert` in the next line (but we've disabled full docs some time ago - and until streaming is added, I don't see us enabling this. So I think it's OK.

markusthoemmes:
Incooperated changes suggested @psuter

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -36,38 +32,27 @@ import spray.http.HttpMethods.PUT
...
-import whisk.core.WhiskConfig
 import whisk.core.database.NoDocumentException
 import whisk.core.entity.ActionLimits
 import whisk.core.entity.ActivationId
-import whisk.core.entity.ActivationResponse
 import whisk.core.entity.DocId
 import whisk.core.entity.DocInfo
 import whisk.core.entity.EntityName
-import whisk.core.entity.{ Exec, SequenceExec }
+import whisk.core.entity.{ SequenceExec }
```
psuter: This looks silly :) (and many similar ones)

common/scala/build.gradle
rabbah: can these be consolidated to one setting file (rather than repeat in each scala project)?

common/scala/src/main/scala/whisk/core/database/CouchDbRestClient.scala
```diff
@@ -100,12 +100,12 @@ class CouchDbRestClient protected (system: ActorSystem, protocol: String, host:
         }
 
         futureBody.map { body =>
-            HttpRequest(method = method, uri = uri, headers = headers, entity = body)
+            HttpRequest(method=method, uri=uri, headers=headers, entity=body)
```
rabbah: I prefer the old version better (with spaces left/right of `=`.
Are you applying all this formatting automatically - if so, will there be an updated eclipse file?

markusthoemmes: That's @psuter's scala style. Mine is the one with the spaces.

markusthoemmes:
@rabbah your suggestions have made it to the code. Refactored the scala version to be centrally defined in the same go.

common/scala/src/main/scala/whisk/core/database/CouchDbRestClient.scala
psuter: I made the change (which really is a revert of another commit in this PR) with the intent of separating two concerns:

1. enforcing certain compiler-checkable properties
2. enforcing style rules

I think this PR should be about 1, and only 1. We cannot currently enforce style so I'd rather that we don't introduce all these changes.

The reason the PR altered the style in so many places is because @markusthoemmes apparently configures his editor to format on save. I'm happy to do the same but before we do that we need a way to do it consistently, and we need to agree on what the preferred style is.

Issue #1043, opened 2016-08-12
KafkaConsumerConnector onMessage needs to wrap commit with a try


rabbah:
The onMessage consumer will abort if an exception is thrown during offset `commit`. The solution adopted in the `ActivationFeed` should be applied to `onMessage` as well. Marking against controller since it is using `onMessage`.

PR #1049, opened 2016-08-12
Prevent KafkaConsumerConnector from Commit-Exceptions #1043

cbickel:
PG1#131 started

### DIFF
--- a/common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
+++ b/common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
@@ -70,12 +70,10 @@ class KafkaConsumerConnector(
                 while (!disconnect) {
                     Try {
                         // Grab next batch of messages and commit offsets immediately
-                        // If the commit fails, the message will not be processed and fetched again in the next run
-                        val records = peek()
+                        // Process the message and commit afterwards. Otherwise the message would not be fetched again.
+                        // It won't be processed twice (tested in "KafkaConnectorTest")
+                        peek() foreach { process.tupled(_) }
                         commit()
-                        records
-                    } map { records =>
-                        records foreach { process.tupled(_) }
                     } recover {
                         case e: CommitFailedException => error(self, s"failed to commit to kafka: ${e.getMessage}")
                         case e: Throwable             => error(self, s"exception while pulling new records: ${e.getMessage}")
--- a/common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
+++ b/common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
@@ -70,10 +70,12 @@ class KafkaConsumerConnector(
                 while (!disconnect) {
                     Try {
                         // Grab next batch of messages and commit offsets immediately
-                        // Process the message and commit afterwards. Otherwise the message would not be fetched again.
                         // It won't be processed twice (tested in "KafkaConnectorTest")
-                        peek() foreach { process.tupled(_) }
+                        val messages = peek()
                         commit()
+                        messages
+                    } map {
+                        _.foreach { process.tupled(_) }
                     } recover {
                         case e: CommitFailedException => error(self, s"failed to commit to kafka: ${e.getMessage}")
                         case e: Throwable             => error(self, s"exception while pulling new records: ${e.getMessage}")
--- a/common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
+++ b/common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
@@ -24,8 +24,6 @@ import scala.concurrent.duration.Duration
 import scala.concurrent.duration.DurationInt
 import scala.concurrent.duration.FiniteDuration
 import scala.language.postfixOps
-import scala.util.Failure
-import scala.util.Success
 import scala.util.Try
 
 import org.apache.kafka.clients.consumer.ConsumerConfig
@@ -35,6 +33,7 @@ import org.apache.kafka.common.serialization.ByteArrayDeserializer
 import whisk.common.Logging
 import whisk.common.TransactionCounter
 import whisk.core.connector.MessageConsumer
+import org.apache.kafka.clients.consumer.CommitFailedException
 
 class KafkaConsumerConnector(
     kafkahost: String,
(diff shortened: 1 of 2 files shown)


common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -69,13 +70,18 @@ class KafkaConsumerConnector(
...
-                        case Failure(t) =>
-                            error(self, s"while polling: $t")
+                    Try {
+                        // Grab next batch of messages and commit offsets immediately
+                        // If the commit fails, the message will not be processed and fetched again in the next run
+                        val records = peek()
+                        commit()
+                        records
+                    } map { records =>
+                        records foreach { process.tupled(_) }
```
rabbah: put on one lines or use `_`? I think it's pretty clear the code is processing records.

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -69,13 +70,18 @@ class KafkaConsumerConnector(
...
-                        case Success(records) =>
-                            records foreach { process.tupled(_) }
-                        case Failure(t) =>
-                            error(self, s"while polling: $t")
+                    Try {
+                        // Grab next batch of messages and commit offsets immediately
+                        // If the commit fails, the message will not be processed and fetched again in the next run
+                        val records = peek()
+                        commit()
+                        records
```
rabbah: I'm thinking that since we're in fact not using split phase, we should create a `pop` method in the consumer which does `peek() and commit()` and throw the exception from either which is wrapped by the `Try` here and in `ActivationFeed`.

tests/src/services/KafkaConnectorTests.scala
```diff
@@ -108,4 +111,44 @@ class KafkaConnectorTests extends FlatSpec with Matchers with BeforeAndAfterAll
...
+        })
+        // The messages are collected in another Thread. Give it some time to receive the message.
+        Thread.sleep(5000)
+        consumer.throwCommitException = false
+        // Give the background-thread some time to get the message and commit that it is collected.
+        Thread.sleep(5000)
+
+        stream.toString should include("exception while pulling new records: commit failed")
+
+        printstream.close()
```
rabbah: wrap with `finally`.

tests/src/services/KafkaConnectorTests.scala
```diff
@@ -108,4 +111,44 @@ class KafkaConnectorTests extends FlatSpec with Matchers with BeforeAndAfterAll
...
+        stream.close()
+    }
```
rabbah: wrap `close()` with `finally`.

tests/src/services/KafkaConnectorTests.scala
```diff
@@ -108,4 +111,44 @@ class KafkaConnectorTests extends FlatSpec with Matchers with BeforeAndAfterAll
...
+            println("message recieved")
+        })
+        // The messages are collected in another Thread. Give it some time to receive the message.
+        Thread.sleep(5000)
+        consumer.throwCommitException = false
+        // Give the background-thread some time to get the message and commit that it is collected.
+        Thread.sleep(5000)
+
+        stream.toString should include("exception while pulling new records: commit failed")
+
```
rabbah: also test that stream contains message received?

tests/src/services/KafkaConnectorTests.scala
```diff
@@ -108,4 +111,44 @@ class KafkaConnectorTests extends FlatSpec with Matchers with BeforeAndAfterAll
...
+        consumer.outputStream = printstream
+
+        val message = new Message { override val serialize = Calendar.getInstance().getTime().toString }
+        val sent = Await.result(producer.send(topic, message), 10 seconds)
+        consumer.throwCommitException = true
+        consumer.onMessage((topic, partition, offset, bytes) => {
+            println("message recieved")
+        })
+        // The messages are collected in another Thread. Give it some time to receive the message.
+        Thread.sleep(5000)
```
rabbah: instead of sleeps, how about a retry around the test matchers?

rabbah:
Reviewed - some things to address, kick back to review when ready.

ChrBi:
On writing the test I had the problem, that the message has not been fetched again from Kafka if the commit failed. I also did some experiments with the “enabale.auto.commit” setting, but it still didn’t work.
So I changed the order of committing and processing again and added a check to the test, that the message is only processed once.

PG 206 started.

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -69,13 +68,16 @@ class KafkaConsumerConnector(
...
-                        case Success(records) =>
-                            records foreach { process.tupled(_) }
-                        case Failure(t) =>
-                            error(self, s"while polling: $t")
+                    Try {
+                        // Grab next batch of messages and commit offsets immediately
+                        // Process the message and commit afterwards. Otherwise the message would not be fetched again.
+                        // It won't be processed twice (tested in "KafkaConnectorTest")
+                        peek() foreach { process.tupled(_) }
+                        commit()
```
rabbah: comment says commit offsets immediately - but this is immediately. In fact what is different about the code path vs previous are the `recover` statements.

how does this not causes messages to be reprocessed if commit fails?

markusthoemmes:
Interesting read on the matter: http://manuel.bernhardt.io/2016/08/09/akka-anti-patterns-flat-actor-hierarchies-or-mixing-business-logic-and-failure-handling/. Relevant docu: http://doc.akka.io/docs/akka/2.4.9/general/supervision.html#supervision

Maybe this exact problem should be solved by supervision (e.g. restarting the connector completely)?

tests/src/services/KafkaConnectorTests.scala
```diff
@@ -107,4 +112,49 @@ class KafkaConnectorTests extends FlatSpec with Matchers with BeforeAndAfterAll
...
+}
+
+class TestKafkaConsumerConnector(
+    kafkahost: String,
+    groupid: String,
+    topic: String,
+    sessionTimeout: FiniteDuration) extends KafkaConsumerConnector(kafkahost, groupid, topic, sessionTimeout = sessionTimeout) {
+
+    override def commit() = {
+        if (throwCommitException) {
```
rabbah: Take a look at this test in `KafkaConnectorTests: it should "send and receive a kafka message even after session timeout"` as a way of actually getting commit to throw a real exception.

rabbah:
@cbickel are you still working on this or shall i take it?

cbickel:
Sry, I didn't have time to work on this the last days. If you have time at the moment, you can take it.

cbickel:
PG1#556 is running

cbickel:
PG2#263 is green

Issue #136, opened 2016-03-29
"failed to fetch action whisk.system/samples/greeting: Object Not Found"


DamianCummins:
Seeing the following after completing the OpenWhisk configure cli instructions on Bluemix:

root@96eb30f5db81:/opt# wsk action get --summary /whisk.system/samples/greeting
action /whisk.system/samples/greeting: Print a friendly greeting
   (params: name place)

root@96eb30f5db81:/opt# wsk action invoke --blocking --result /whisk.system/samples/greeting
error: {
    "activationId": "358991740954436e9fc385e351990130",
    "annotations": [],
    "end": 1459261038802,
    "logs": [],
    "name": "samples",
    "namespace": "dcummin3@uk.ibm.com",
    "publish": false,
    "response": {
        "result": {
            "error": "failed to fetch action whisk.system/samples/greeting: Object Not Found"
        },
        "status": "action developer error",
        "success": false
    },
    "start": 1459261038802,
    "subject": "dcummin3@uk.ibm.com",
    "version": "0.0.1"
}

Any idea why I would be able to get the summary but not invoke the action?
This also appears to happen for watson/translate action after I've bound to the service.

Running on: Linux version 4.1.19-boot2docker (gcc version 4.9.2 (Debian 4.9.2-10) )

DamianCummins:
Note, I am able to create and invoke my own actions.

rabbah:
Thanks for reporting this. I will have a look.

rabbah:
I have confirmed this is an issue and will report back when a fix is deployed.

DamianCummins:
@rabbah Great, thanks for the prompt response.

rabbah:
You may try again - should work for you this time.

DamianCummins:
@rabbah yep that worked. Thanks.

PR #1051, opened 2016-08-12
Fixes weather forecast action using deprecated API.

jthomas:
Weather service used by the forecast action has recently released a
new API version. This has broken the current Action source, which
used the old URL structure. Updating the code to use the new URL
paths and parameters.

### DIFF
--- a/catalog/weather/forecast.js
+++ b/catalog/weather/forecast.js
@@ -22,7 +22,7 @@ function main(params) {
     var units = params.units || 'm';
     var timePeriod = params.timePeriod || '10day';
     var url = 'https://twcservice.mybluemix.net/api/weather/v1/geocode/' + lat + '/' + lon;
-    var qs = {language: language, units: units}; 
+    var qs = {language: language, units: units};
 
     switch(timePeriod) {
         case '48hour':
--- a/catalog/weather/forecast.js
+++ b/catalog/weather/forecast.js
@@ -21,30 +21,28 @@ function main(params) {
     var language = params.language || 'en-US';
     var units = params.units || 'm';
     var timePeriod = params.timePeriod || '10day';
-    var timeURL ='/forecast/daily/10day';
+    var url = 'https://twcservice.mybluemix.net/api/weather/v1/geocode/' + lat + '/' + lon;
+    var qs = {language: language, units: units}; 
 
     switch(timePeriod) {
-        case '10day':
-            timeURL = '/forecast/daily/10day';
-            break;
-   	    case '24hour':
-            timeURL = '/forecast/hourly/24hour';
+        case '48hour':
+            url += '/forecast/hourly/48hour.json';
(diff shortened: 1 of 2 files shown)


jasonpet:
@jthomas - what do you mean by broken?  Do you mean a deprecated warning is thrown?  The weather tests are passing.

jasonpet:
Please reference a pg build when this is ready to be merged.

rabbah:
I think we'll have to run a pg on his behalf @jasonpet.

rabbah:
I think we'll have to run a pg on his behalf @jasonpet.
We will need to mirror this against the catalog, yes? In which case, is Travis testing the catalog yet?

jasonpet:
@rabbah - sure I can run one, but first we need him to fix up the trailing whitespaces issue.

jasonpet:
Yeah, I can update the action over in the openwhisk-catalog to mirror this.  The test for weather however, does not exist in open.  Will only run as part of pg build.

catalog/weather/forecast.js
```diff
@@ -21,30 +21,28 @@ function main(params) {
     var language = params.language || 'en-US';
     var units = params.units || 'm';
     var timePeriod = params.timePeriod || '10day';
-    var timeURL ='/forecast/daily/10day';
+    var url = 'https://twcservice.mybluemix.net/api/weather/v1/geocode/' + lat + '/' + lon;
```
csantanapr: So'm confuse if they moved to a new API, why it went from `weather/v2` instead of `weather/v1` seems backward

csantanapr: Never mind I see the test now

> https://new-console.ng.bluemix.net/docs/services/InsightsWeather/weather_rest_apis.html

> This service is deprecated: All instances of this service are deprecated. The revised service is available in Bluemix as Weather Company Data for IBM Bluemix.
Current apps that use the Insights for Weather APIs will continue to work without any modifications for 90 days after the Weather Company data service is released on July 1 2016. To take advantage of the newly added APIs and the improved pricing model, it is in your best interest to migrate to one of the new plans.

csantanapr:
New API documented here as a reference:
https://new-console.ng.bluemix.net/docs/services/Weather/weather_rest_apis.html#url_construction

We should add that url in the comments as a quick reference.

jasonpet:
I will have to create a new bluemix Weather Company Data service and add new entry into vcap file before we will even get a pg build to pass.  The existing Weather Insights service will not work with the updated action.  Once the vcap updates get committed with the new service info I will kick off a pg build.  @jthomas can you please remove the trailing whitespace line in forcast.js for this PR.

jasonpet:
pg build #136

csantanapr:
rebase and merged 4e7cb7e397129cca2cd4920434d35f4231b3e92a

PR #1053, opened 2016-08-12
Correct the endmarker for blocking activations

ChrBi:
Before this change, the logmarker has been written before the Future was finished. But it has to be written when the response goes back to the user.


rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/77b711057eb7779ead8b077876f4139beb6d4d2f.

PR #1056, opened 2016-08-12
Removing deployment variable (prod/open)

animeshsingh:
Also including couple of README changes. Was trying to squash - but not going the right was I believe - please check

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -47,8 +47,9 @@ linux:
 
 docker_image_tag: latest
 docker:
-  #user: The user to install docker for. Defaults to the ansible user if not set. This will be the user who is able to run 
-  #docker commands on a machine setup with prereq_build.yml
+  # The user to install docker for. Defaults to the ansible user if not set. This will be the user who is able to run 
+  # docker commands on a machine setup with prereq_build.yml
+  #user:
   version: 1.9.1-0~trusty
   storagedriver: overlay
   port: 4243
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -47,7 +47,8 @@ linux:
 
 docker_image_tag: latest
 docker:
-  #user:
+  #user: The user to install docker for. Defaults to the ansible user if not set. This will be the user who is able to run 
+  #docker commands on a machine setup with prereq_build.yml
   version: 1.9.1-0~trusty
   storagedriver: overlay
   port: 4243
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -1,10 +1,6 @@
 mode: deploy
 prompt_user: true
-
 openwhisk_home: "{{ lookup('env', 'OPENWHISK_HOME')|default(playbook_dir + '/..', true) }}"
-
-# Use this parameter to determine the type of OpenWhisk deploy (e.g., development, production)
-deployment: prod
 
 whisk:
   version:
@@ -51,7 +47,7 @@ linux:
 
 docker_image_tag: latest
 docker:
-  user:
+  #user:
   version: 1.9.1-0~trusty
   storagedriver: overlay
   port: 4243
--- a/ansible/roles/whiskhost/tasks/deploy.yml
+++ b/ansible/roles/whiskhost/tasks/deploy.yml
@@ -2,6 +2,6 @@
 # Add user and patch kernel.
 
 - include: addUser.yml
-  when: "'prod' in deployment"
+  when: docker.user is defined
 
 - include: patchKernel.yml
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ allow you to configure OpenWhisk to use a persistent data store via Cloudant or 
 These steps were tested on Mac OS X El Capitan, Ubuntu 14.04.3 LTS and Windows using Vagrant.
 
 * [Vagrant on Mac, Windows PC, or GNU/Linux](tools/vagrant/README.md)
-* [OpenWhisk on Ubuntu natively](tools/ubuntu-setup/README.md)
+* [OpenWhisk on Ubuntu Server(s) natively](tools/ubuntu-setup/README.md)
 
 ### Learn concepts and commands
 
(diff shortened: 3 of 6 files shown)


rabbah:
@domdom82 @lehoanganh for your reviewing pleasure.

ansible/group_vars/all
```diff
@@ -51,7 +47,7 @@ linux:
 
 docker_image_tag: latest
 docker:
-  user:
+  #user:
```
markusthoemmes: remove that line if its not needed?

ansible/roles/whiskhost/tasks/deploy.yml
```diff
@@ -2,6 +2,6 @@
 # Add user and patch kernel.
 
 - include: addUser.yml
-  when: "'prod' in deployment"
+  when: docker.user is defined
```
markusthoemmes: Add documentation for this setting.

Also wondering: @domdom82 mentioned it to be fairly "hard" to override puncutated values in environment specific `group_vars/all` (where I believe this should be set?). I think that's why we opted for `docker_user` style variables for convenience. @domdom82  comment?

animeshsingh: No documentation needed in open source code, as user is not expected to do anything

markusthoemmes: Where does `docker.user` come from then?

animeshsingh: Well if you need to use same for internal deployment, you can use the variable to set to whatever user we are using.

markusthoemmes: Is it overridable in an environment or do i need to manipulate the global `group_vars/all`?

animeshsingh: I am not sure - we havent tested overriding it.

markusthoemmes: Right, let's wait for @domdom82 or @lehoanganh to chime in, they'll know.

Can we add at least a linecomment for that variable in `group_vars/all` to describe what it does? Something like

```
# The user to install docker for. Defaults to the ansible user iff not set. This will be the 
# user who is able to run docker commands on a machine setup with prereq_build.yml
```

animeshsingh: Done

markusthoemmes:
@lehoanganh @domdom82 comments or thumbs up please.

animeshsingh:
@domdom82 - can we get this PR in please?

ansible/roles/whiskhost/tasks/deploy.yml
lehoanganh: Conventionally we use the following form in the **global** `group_vars/all`

```
docker
  user: foo
```

and this following form in **environment** `groups_var/all`

```
docker_user: foo
```

with the intention that vars in environments can be overridden from the CLI with `-e docker_user=bar`.

The global var could also be overridden (but preferably not), however it has be passed to the CLI as a *multi-hirarchy JSON format*   which is not human friendly.

domdom82: @lehoanganh is right about that. it should be similar to the docker_image_tag param except it does not have a default value. I would also prefer to set it on the environment even though for us the user is probably going to be the same everywhere.

rabbah:
@domdom82 @lehoanganh do you expect this is a value should be overridable and if so which is the more appropriate convention? It's not clear from your comments if you're agreeing that it should be or just clarifying the conventions and cli/env override mechanisms.

lehoanganh:
my 2 cents

```
ansible-playbook -i environments/mac properties.yml

PLAY [ansible] *****************************************************************

TASK [setup] *******************************************************************
Tuesday 23 August 2016  10:47:39 +0200 (0:00:00.026)       0:00:00.026 ********
ok: [ansible]

TASK [debug] *******************************************************************
Tuesday 23 August 2016  10:47:39 +0200 (0:00:00.483)       0:00:00.509 ********
ok: [ansible] => {
    "docker.user": "foo"
}
```

```
ansible-playbook -i environments/mac properties.yml -e '{"docker": {"user": bar}}'

PLAY [ansible] *****************************************************************

TASK [setup] *******************************************************************
Tuesday 23 August 2016  10:48:43 +0200 (0:00:00.026)       0:00:00.026 ********
ok: [ansible]

TASK [debug] *******************************************************************
Tuesday 23 August 2016  10:48:44 +0200 (0:00:00.482)       0:00:00.508 ********
ok: [ansible] => {
    "docker.user": "bar"
}
```

overwritting a **global** var via CLI like above is not human-friendly. If it should be overriden, putting it in an environment maybe a better choice, IMHO it's not global properties where every deployment should need it.

ansible/roles/whiskhost/tasks/deploy.yml
animeshsingh: With one instance of the code pulled in, one deployment of Whisk is going to be there. As such, it should be part of global deployment, and it doesn't buy us anything moving the checks to under environment.

animeshsingh:
@lehoanganh, as commented above, with one instance of the code pulled in, one deployment of OpenWhisk is going to be there. As such, it should be part of global deployment, and it doesn't buy us anything moving the checks to under environment, just so that its easier to override

animeshsingh:
Also why do we want to confuse someone who is going to deploy this as to how many places this user has to be changed? the logix currently looks in one place, and if we change it to a convoluted logic for looking for user defined in multiple palces - its going to be confusing for someone not familiar with code anatomy - in my view this doesnt buy as anything substantive

domdom82:
Since we can still use the global `group_vars/all` in our deployment and our `docker.user` unlikely to change anytime soon per environment I tend to agree with @animeshsingh for now. If we experience problems later, this is not going to be hard to change anyway. So LGTM.

markusthoemmes:
rebased, squashed and merged 153129c6c3b1fd71d5359b3a308217737d495481

PR #1060, opened 2016-08-15
Enhance debugging for failed tests that assert on log-message and pri…

ChrBi:
…nt activationId for CLIRuleTests

Enhance logging to debug failed Test cases.


rabbah:
This is a duplicate pr.

PR #1062, opened 2016-08-15
Add since parameter to WskCli to get only activation from the current…

ChrBi:
… test

PG 164 is started.

### DIFF
--- a/tests/src/common/WskCli.java
+++ b/tests/src/common/WskCli.java
@@ -599,10 +599,12 @@ public class WskCli {
      *
      * @param name
      *            name of the action.
+     * @param since
+     * 			  since parameter for cli. Only activations after this timestamp will be returned.
      * @return the activation ids
      */
-    public List<String> getActivations(String name) throws IOException {
-        return getActivations(name, 0);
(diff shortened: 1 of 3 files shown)


rabbah:
system.basic.CLIRuleTests > rule2to2 FAILED

ChrBi:
That was my mistake. PG 164 is green now.

tests/src/common/WskCli.java
```diff
@@ -838,14 +852,16 @@ private boolean logsForActionContain(String action, String sRegex, int initialWa
      *            name of the action.
      * @param w
      *            the string to search for
+     * @param since
+     * 			  since parameter for cli. Only activations after this timestamp will be returned.
```
markusthoemmes: There are tabs in there. (several other places aswell). Can you doublecheck why scanCode doesn't work?

rabbah: scan code checks for leading tabs, not ones since a comment (or within a line) @psuter FYI.

rabbah:
Tabs removed and merged as https://github.com/openwhisk/openwhisk/commit/43a0b306e723363a9a2d87569c56059d6a44e0e8.

PR #1069, opened 2016-08-16
Exclude SwiftTest from CLi tests

cbickel:
Signed-off-by: Hoang Anh Le <hoang@de.ibm.com>

PG 168 is green

### DIFF
--- a/tests/src/whisk/core/cli/test/SwiftTests.scala
+++ b/tests/src/whisk/core/cli/test/SwiftTests.scala
@@ -14,15 +14,13 @@
  * limitations under the License.
  */
 
-package system.basic
+package whisk.core.cli.test
 
 import scala.concurrent.duration.DurationInt
 import scala.language.postfixOps
-
 import org.junit.runner.RunWith
 import org.scalatest.Matchers
 import org.scalatest.junit.JUnitRunner
-
 import common.TestHelpers
 import common.TestUtils
 import common.Wsk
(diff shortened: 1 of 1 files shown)


rabbah:
@ChrBi @lehoanganh can you explain the rationale for this change?

ChrBi:
This test failed in the past a few times. On looking into it, we realized that the action was killed after one minute in the invoker. That's why we increased the timeout-limit of the action in this test.

sjfink:
Do we have any reason to believe 2 minutes is better than 1?

It seems more likely that the action is hanging rather than just slow, right?

perryibm:
@paulcastro  Paul and I think that increasing from 1 to 2 minute will not make a difference.

Has it been observed to take close to 1 minute and work?

ChrBi:
The problem we had here was that the action took longer than one minute in usercontainer. So it haas been stopped by the invoker.
If the action was only blocked or crached, how can we find out, why it was blocked and what's the reason for this?

Here is a failed test from CI build 280
```
system.basic.CLISwiftTests > Swift Actions should invoke a swift action FAILED
    org.scalatest.exceptions.TestFailedException: Test duration exceeds expectation (ms) 31922 was not less than or equal to 30000
```
In this example the action took longer than 30 seconds, but it didn't stuck. Isn't it possible that it could take more than a minute?

But if we all still say, that two minutes don't help, I will revert the changes and only rename the class, as discussed with @jeremiaswerner

rabbah:
Is it sufficient to rename the suite but not move it to another package in this case?

cbickel:
You are right, it is more a CLI test. I moved it.
New PG1#261 is running

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/112c42dbb9889bcb2474e286ade7ecf6e5f4109c.

Issue #1067, opened 2016-08-16
phase ordering of posting message to invoker and waiting for active ack response


rabbah:
The controller phase ordering for invoking a blocking action and waiting for the active ack response: 1. post to kafka, 2. wait for active ack, 3. fall over to db if necessary, 4. terminate response. But it's possible for 2 to occur before the controller has registered the listener, so the phase ordering should account for a response that may be received by the controller before the activation id listener is created.

Specifically, this is related to `queryActivationResponse` occurring after `performLoadBalancerRequest` (chained two future callbacks later) in `Actions.scala`.

@ChrBi @perryibm

PR #1071, opened 2016-08-16
issue #1067 Let the controller first listen to the active ack of the …

cbickel:
…invoker before sending the action to Kafka

PG 169 started.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Actions.scala
@@ -434,22 +434,21 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
             implicit transid: TransactionId): Future[(ActivationId, Option[WhiskActivation])] = {
         // merge package parameters with action (action parameters supersede), then merge in payload
         val args = { env map { _ ++ action.parameters } getOrElse action.parameters } merge payload
-        val message = Message(transid, s"/actions/invoke/${action.namespace}/${action.name}/${action.rev}", user, ActivationId(), args)
 
-        info(this, s"[POST] action activation id: ${message.activationId}")
-
-        val promise = Promise[Option[WhiskActivation]]
+        val promise = Promise[WhiskActivation]
         val duration = action.limits.timeout()
         val timeout = duration + blockingInvokeGrace
-        if (blocking) {
-            val docid = DocId(WhiskEntity.qualifiedName(user.namespace, message.activationId))
+        val activationId = if (blocking) {
             info(this, s"[POST] action activation will block on result up to $timeout ($duration + $blockingInvokeGrace grace)")
             // Start listening for finished activations before sending the action to invoker
             // This avoids, that the invoker is faster with the finished message than the controller knows that it has to listen to it.
             // If the post of LB to Kafka fails, the activationId will be deleted after 30 seconds from the map, the controller waits for.
             // Letting the promise fail on a failure on posting to Kafka is not necessary, because nothing will done with it afterwards.
-            pollLocalForResult(docid.asDocInfo, message.activationId, promise)
-        }
+            pollLocalForResult(user.namespace, ActivationId(), promise)
+        } else ActivationId()
+
+        val message = Message(transid, s"/actions/invoke/${action.namespace}/${action.name}/${action.rev}", user, activationId, args)
+        info(this, s"[POST] action activation id: ${message.activationId}")
 
         performLoadBalancerRequest(INVOKER, message, transid) flatMap { response =>
             response.id match {
@@ -467,8 +466,8 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
             }
         } flatMap { activationId =>
             if (blocking) {
-                val response = promise.future map {
-                    (activationId, _)
+                val response = promise.future map { activation =>
+                    (activationId, Some(activation))
                 } withTimeout (timeout, new BlockingInvokeTimeout(activationId))
                 response onComplete {
                     case Success(_) =>
@@ -496,20 +495,29 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
      * a database operation to obtain the canonical answer.
      */
     private def pollLocalForResult(
-        docid: DocInfo,
+        namespace: Namespace,
         activationId: ActivationId,
-        promise: Promise[Option[WhiskActivation]])(
-            implicit transid: TransactionId): Unit = {
-        queryActivationResponse(activationId, transid) map {
-            activation => promise.trySuccess { Some(activation) }
-        } onFailure {
+        promise: Promise[WhiskActivation])(
+            implicit transid: TransactionId): ActivationId = {
+
+        val activeAckPromise = Promise[WhiskActivation]
+        val activationIdToSearch = queryActivationResponse(activationId, transid, activeAckPromise)
+
+        activeAckPromise.future map { activation =>
+            promise.trySuccess { activation }
+            activation.activationId
+        } recover {
             case t: TimeoutException =>
                 info(this, s"[POST] switching to poll db, active ack expired")
-                pollDbForResult(docid, activationId, promise)
+                val docid = DocId(WhiskEntity.qualifiedName(namespace, activationIdToSearch)).asDocInfo
+                pollDbForResult(docid, activationIdToSearch, promise)
             case t: Throwable =>
                 error(this, s"[POST] switching to poll db, active ack exception: ${t.getMessage}")
-                pollDbForResult(docid, activationId, promise)
+                val docid = DocId(WhiskEntity.qualifiedName(namespace, activationIdToSearch)).asDocInfo
+                pollDbForResult(docid, activationIdToSearch, promise)
         }
+
+        activationIdToSearch
     }
 
     /**
@@ -521,11 +529,11 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
     private def pollDbForResult(
         docid: DocInfo,
         activationId: ActivationId,
-        promise: Promise[Option[WhiskActivation]])(
+        promise: Promise[WhiskActivation])(
             implicit transid: TransactionId): Unit = {
         if (!promise.isCompleted) {
             WhiskActivation.get(activationStore, docid) map {
-                activation => promise.trySuccess { Some(activation) } // activation may have logs, do not strip them
+                activation => promise.trySuccess { activation } // activation may have logs, do not strip them
             } onFailure {
                 case e: NoDocumentException =>
                     Thread.sleep(500)
--- a/core/controller/src/main/scala/whisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Actions.scala
@@ -17,7 +17,6 @@
 package whisk.core.controller
 
 import scala.concurrent.Future
-import scala.concurrent.Promise
 import scala.concurrent.TimeoutException
 import scala.concurrent.duration.DurationInt
 import scala.util.Failure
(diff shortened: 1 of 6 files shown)


core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -450,50 +450,55 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+        val duration = action.limits.timeout()
+        val timeout = duration + blockingInvokeGrace
+        if (blocking) {
+            val docid = DocId(WhiskEntity.qualifiedName(user.namespace, message.activationId))
+            info(this, s"[POST] action activation will block on result up to $timeout ($duration + $blockingInvokeGrace grace)")
+            // Start listening for finished activations before sending the action to invoker
+            // This avoids, that the invoker is faster with the finished message than the controller knows that it has to listen to it.
+            // If the post of LB to Kafka fails, the activationId will be deleted after 30 seconds from the map, the controller waits for.
+            // Letting the promise fail on a failure on posting to Kafka is not necessary, because nothing will done with it afterwards.
+            pollLocalForResult(docid.asDocInfo, message.activationId, promise)
```
rabbah: Create the active ack listener eagerly to avoid a race between establishing the listener and the invoker actually sending an active ack for this action. This prevents an active ack response for the activation being dropped because the listener is not yet active (this can happen if the response is faster coming back to the controller than the future callbacks getting a chance to run) and forcing the db polling fall back to kick in.

rabbah:
Suggested alternative comment for you to polish.

rabbah:
Travis failures:
```
hisk.core.controller.test.ActionsApiTests > Actions API should invoke an action, blocking FAILED
whisk.core.controller.test.ActionsApiTests > Actions API should invoke a blocking action and return error response when activation fails FAILED
```
are related to this change.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -134,7 +132,11 @@ class LoadBalancerService(config: WhiskConfig, verbosity: LogLevel)(
                     if (failedit) info(this, "active response timed out")
                 }
             })
-        }).future
+        }).future.map { activation =>
+            promise.success(activation)
```
markusthoemmes: `trySuccess`?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -491,20 +495,29 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
     private def pollLocalForResult(
-        docid: DocInfo,
+        namespace: Namespace,
         activationId: ActivationId,
-        promise: Promise[Option[WhiskActivation]])(
-            implicit transid: TransactionId): Unit = {
-        queryActivationResponse(activationId, transid) map {
-            activation => promise.trySuccess { Some(activation) }
-        } onFailure {
+        promise: Promise[WhiskActivation])(
```
markusthoemmes: As discussed in person: Can we try to get rid of these `Promise` passing `Unit` returning methods in favor of `Future` composition?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -434,53 +434,57 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+        val promise = Promise[WhiskActivation]
+        val duration = action.limits.timeout()
+        val timeout = duration + blockingInvokeGrace
+        val activationId = if (blocking) {
+            info(this, s"[POST] action activation will block on result up to $timeout ($duration + $blockingInvokeGrace grace)")
+            // Start listening for finished activations before sending the action to invoker
+            // This avoids, that the invoker is faster with the finished message than the controller knows that it has to listen to it.
+            // If the post of LB to Kafka fails, the activationId will be deleted after 30 seconds from the map, the controller waits for.
+            // Letting the promise fail on a failure on posting to Kafka is not necessary, because nothing will done with it afterwards.
+            pollLocalForResult(user.namespace, ActivationId(), promise)
```
markusthoemmes: As noted in personal discussion:
I found it hard to grasp the code because we use `Unit` returning functions which resolve a `Promise` at some point. This reads like it blocks the code until the result is there (which makes no sense at this point).

I'ld prefer to use `Future` composition if possible so it becomes more apparent that this is in fact returning a Future which is resolved as soon as the ack-message is there.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -434,53 +434,57 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
-                    response // will either complete with activation or fail with timeout
-                } else Future {
-                    // Duration of the non-blocking activation in Controller.
-                    // We use the start time of the tid instead of a startMarker to avoid passing the start marker around.
-                    transid.finished(this, StartMarker(transid.meta.start, LoggingMarkers.CONTROLLER_ACTIVATION))
-                    (activationId, None)
+            }
+        } flatMap { activationId =>
+            if (blocking) {
+                val response = promise.future map { activation =>
```
markusthoemmes: Would it make sense to decompose `pollLocalForResult` and `pollDbForResult` and run the latter explicitly here? Or alternatively fold the last fallback into `pollForResult` so we have

1. wait for ack
2. fallback to db
3. timeout

in one place of code.

ChrBi:
PG1#226 started
Test to check if failover to active polling and test if timeout is working are added.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -490,21 +495,35 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
-            implicit transid: TransactionId): Unit = {
-        queryActivationResponse(activationId, transid) map {
-            activation => promise.trySuccess { Some(activation) }
-        } onFailure {
+        timeout: FiniteDuration)(
+            implicit transid: TransactionId): (ActivationId, Future[WhiskActivation]) = {
+
+        val (activationIdToSearch, activeCallback) = queryActivationResponse(activationId, transid)
+
+        val promise = Promise[WhiskActivation]
```
markusthoemmes: Can we get rid of this promise aswell? It doesn't strike me as necessary to have the promise passed to `pollDbForResult`.

`recoverWith` might be helpful

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -434,53 +435,57 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+                    } else {
+                        error(this, s"[POST] action activation failed: ${response.error.getOrElse("??")}")
+                        Future failed new IllegalStateException(s"activation failed with error: ${response.error.getOrElse("??")}")
+                    }
+            }
+        } flatMap { activationId =>
+            blockingCallback match {
+                case Some(callback) => {
+                    // Blocking invocation
+                    callback map { activation =>
```
markusthoemmes: What happens in the error case (i.e. timeout)?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -434,53 +435,57 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+        val (activationId, blockingCallback) = if (blocking) {
+            val timeout = duration + blockingInvokeGrace
+            info(this, s"[POST] action activation will block on result up to $timeout ($duration + $blockingInvokeGrace grace)")
+            // Create the active ack listener eaerly to avoid a race between establishing the listener and the invoker
+            // actually sending an active ack for this action. This prevents an active ack response for the activation
+            // being dropped because the listener is not yet active (this can happen if the response is faster coming back
+            // to the controller than the future callbacks getting a chance to run) and forcing the db polling fall back to kick in.
+            val (aid, callback) = waitForActivationResponse(user.namespace, ActivationId(), timeout)
+            (aid, Some(callback))
+        } else (ActivationId(), None)
```
markusthoemmes: The `ActivationId()` is independent from blocking or non-blocking? Define above the block?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -490,21 +495,35 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+    private def waitForActivationResponse(
+        namespace: Namespace,
         activationId: ActivationId,
-        promise: Promise[Option[WhiskActivation]])(
-            implicit transid: TransactionId): Unit = {
-        queryActivationResponse(activationId, transid) map {
-            activation => promise.trySuccess { Some(activation) }
-        } onFailure {
+        timeout: FiniteDuration)(
+            implicit transid: TransactionId): (ActivationId, Future[WhiskActivation]) = {
```
markusthoemmes: Why does this need to respond with the `ActivationId` if it already gets passed one?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -516,11 +535,11 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
     private def pollDbForResult(
         docid: DocInfo,
         activationId: ActivationId,
-        promise: Promise[Option[WhiskActivation]])(
+        promise: Promise[WhiskActivation])(
```
markusthoemmes: Can the effect of this be achieved through Future composition?

rabbah:
Is the threading of `ActivationId` done for testing? (Many of @markusthoemmes' comments above are related to this.) If so, a cleaner option is to define an activation id factory in the backend services trait which is called to get a new activation id instead of ActivationId(). Then the test will override this trait to issue an activation id as needed for tests.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
cbickel: We need to return the activationId again, because of the unitTests.
In the test an activation is created in the db with an predefined activationId. `queryActivationResponse` returns this predefined ActivationId and to let the test pass, the action has to look in the db for the activationId, that has been written by the test.
Before this change, `performLoadBalancerRequest` had this functunality.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
cbickel: On a timeout the Future `callback` is failing.
On a timeout, in line 524 a `BlockingInvokeTimeout` is thrown. This will be passed through and catched in line 276.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
cbickel: For the unit-tests, a different ActivationId will be returned from `waitForActivationResponse`.
In this case we have to work with another Id than the generated one before.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -434,53 +434,57 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+
+        performLoadBalancerRequest(INVOKER, message, transid) flatMap { response =>
+            response.id match {
+                case Some(activationId) =>
+                    Future successful activationId
+                case None =>
+                    if (response.error.getOrElse("??").equals("too many concurrent activations")) {
+                        // DoS throttle
+                        warn(this, s"[POST] action activation rejected: ${response.error.getOrElse("??")}")
+                        Future failed new TooManyActivationException("too many concurrent activations")
```
markusthoemmes: This is dead code i believe. The `ActivationThrottle` emitting this error response is part of the Entitlement in the controller now. It will be short-circuited before it even reaches here.

PR #1075, opened 2016-08-17
Remove Pipe Dependency from Action Sequences

### LINKED ISSUE #1057: Why is system/pipe hard coded in action.go?
I think this is a kind of issue that we hard coded "system/pipe" at Line 533 of openwhisk/tools/go-cli/go-whisk-cli/commands/action.go.

If this action is missing, some of our test cases have to fail, because they are unable to create the sequence of actions. All the test cases we have now cannot access the system auth key, which means they are unable to create fixture for this system/pipe action.

I propose to change this way of implementation, at least by replacing it with a parameter/variable(we can keep the system/pipe action as default), so that we can specify a "mocking pipe" action in the test cases.

dubeejw:
- Remove the dependency on the system pipe when creating actions
- Pass action artifacts to the backend as a component field instead of as parameters

Fixes #1057

### DIFF
--- a/tools/go-cli/go-whisk-cli/commands/action.go
+++ b/tools/go-cli/go-whisk-cli/commands/action.go
@@ -528,53 +528,9 @@ func parseAction(cmd *cobra.Command, args []string) (*whisk.Action, bool, error)
     client.Namespace = qName.namespace
     action.Exec = existingAction.Exec
   } else if flags.action.sequence {
-    currentNamespace := client.Config.Namespace
-    client.Config.Namespace = "whisk.system"
-
-    pipeAction, _, err := client.Actions.Get("system/pipe")
-    if err != nil {
-      whisk.Debug(whisk.DbgError, "client.Actions.Get(%s) error: %s\n", "system/pipe", err)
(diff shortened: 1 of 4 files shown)


tools/go-cli/go-whisk-cli/commands/action.go
```diff
@@ -531,22 +531,23 @@ func parseAction(cmd *cobra.Command, args []string) (*whisk.Action, bool, error)
     currentNamespace := client.Config.Namespace
     client.Config.Namespace = "whisk.system"
 
-    pipeAction, _, err := client.Actions.Get("system/pipe")
+    /*pipeAction, _, err := client.Actions.Get("system/pipe")
```
rabbah: just remove this code? it's not coming back.

rabbah:
Can you add the test alluded to here: https://github.com/openwhisk/openwhisk/issues/1057#issuecomment-240458764?

dubeejw:
Will do.

dubeejw:
PG1 214.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/e5cf3c1806b8c1ed2f3dfaba24f9aacedfe81f90.

Issue #1073, opened 2016-08-16
Change the Invoker Health protocol


markusthoemmes:
Today, we have a "manual" health protocol between the controller and the invokers. Every invoker writes his current timestamp (in an endless loop, every 2 seconds) to the consul Key/Value store. This is putting a lot of load onto the KV store.

Instead, we're proposing to use consul's service list and health checks consul itself is sending to our services to determine wether the components are up. Bonus points for having a more meaningful `/ping` endpoint which actually does checks for the component (for example the controller checking its connectivity)

rabbah:
Is this still valid? (Rest is covered by #1711?)

Issue #903, opened 2016-07-15
Consolidate Consul clients


psuter:
OpenWhisk uses several parts of the [Consul client API](https://www.consul.io/docs/agent/http.html), but the functionality is scattered across several clients:
- `whisk.common.ConsulClient`, [key-value store API](https://www.consul.io/docs/agent/http/kv.html), using `akka-http` (non blocking)
- `whisk.common.ConsulKV`, [key-value store API](https://www.consul.io/docs/agent/http/kv.html), based on the deprecated `HttpUtils` (blocking). Used at this point only at load-time for `Config` objects.
- `whisk.common.ConsulServiceCheck`, [catalog](https://www.consul.io/docs/agent/http/catalog.html) and [health](https://www.consul.io/docs/agent/http/health.html) APIs, based on `HttpUtils` (blocking), used for monitoring.

PR #1076, opened 2016-08-17
Extend ConsulClient to support catalog and health API

markusthoemmes:
This will be needed for future work on #1073 and also covers some deployment semantics via tests (asserts that consul reports everything up after the deployment)


psuter:
Related to #903.

common/scala/src/main/scala/whisk/common/ConsulClient.scala
```diff
@@ -78,8 +92,8 @@ class ConsulClient(hostAndPort: String)(implicit val actorSystem: ActorSystem) {
                 uri = uriWithKey(key)
             )
         ) flatMap { response =>
-            Unmarshal(response.entity).to[List[ConsulEntry]]
-        } map { _.head.decodedValue.getOrElse(throw new NoSuchElementException()) }
+                Unmarshal(response.entity).to[List[ConsulEntry]]
```
psuter: (crying silently)

tests/src/whisk/consul/ConsulHealthTests.scala
```diff
@@ -0,0 +1,57 @@
+/*
```
psuter: Why a different file?

tests/src/whisk/consul/ConsulClientTests.scala
```diff
@@ -73,13 +73,22 @@ class ConsulClientTests extends FlatSpec with ScalaFutures with Matchers with Ws
...
     }
 
-    it should "drop the first part of the key" in {
+    "Consul Catalog client" should "return a list of all services" in {
+        consul.catalog.services().futureValue.size should be > 0
+    }
+
+    "Consul Health client" should "return a list of health results" in {
+        val services = consul.health.service("bogus").futureValue
+        services shouldBe List.empty
```
psuter: Test the positive case too? or is that too much setup?

psuter:
Code looks good. Can we take the opportunity to remove the synchronous clients? IIRC last time I did it locally it "only" meant making sure an `ActorSystem` was available by the time the first configuration object is created.

tests/src/whisk/consul/ConsulClientTests.scala
markusthoemmes: Thought about that aswell. It means implementing methods we don't need though.

common/scala/src/main/scala/whisk/common/ConsulClient.scala
rabbah: we need to fix the white space wars!

tests/src/whisk/consul/ConsulHealthTests.scala
markusthoemmes: This test is supposed to test a full openwhisk deployment with services registered and so on. The other test is supposed to "unit" test just the client in conjunction with a consul-server.

markusthoemmes:
@psuter `ConsulKV` nuked!

tests/src/whisk/consul/ConsulClientTests.scala
markusthoemmes: Convinced myself that I shall not be lazy. Will implement the rest as well.

markusthoemmes:
@psuter testcases added

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -96,6 +96,8 @@ object Controller {
     }
 
     def main(args: Array[String]): Unit = {
+        implicit val system = ActorSystem("controller-actor-system")
```
psuter: I believe this changes the behavior if the config is not valid? (hangs because no one terminates the `ActorSystem`)

rabbah: > I believe this changes the behavior if the config is not valid? (hangs because no one terminates the ActorSystem)
> 
I concur. It should get terminated. The controller will not be healthy and won't respond to ping though.

PR #1079, opened 2016-08-17
Sets the context classloader before running Java actions

### LINKED ISSUE #1020: Wrong Context Class Loader in Java actions
The Context Class Loader visible within Java actions is the one that loads the wrapper/proxy code (as opposed to the JarLoader that loads the action class itself). This is both a potential security hole (exposing Whisk internals to the Java action code), and may cause the action code that actually tries to use the context class loader to misbehave (e.g., an attempt to use it in order to load additional classes from the provided action jar would cause ClassNotFound exception).

The solution seems to be to set the context class loader (before invoking the main method of the action class) to the same class loader that is used to load the action jar (and potentially to restore it to the original one after main method is over).

`Thread.currentThread().setContextClassLoader(loader)`

psuter:
Accompanying unit tests. Closes #1020.

FYI @glikson @rabbah

### DIFF
--- a/core/javaAction/proxy/src/main/java/openwhisk/java/action/Proxy.java
+++ b/core/javaAction/proxy/src/main/java/openwhisk/java/action/Proxy.java
@@ -111,6 +111,7 @@ public class Proxy {
                 return;
             }
 
+            ClassLoader cl = Thread.currentThread().getContextClassLoader();
             SecurityManager sm = System.getSecurityManager();
 
             try {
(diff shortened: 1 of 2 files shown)


glikson:
Looks good, thanks!

psuter:
PG1 205 in progress

markusthoemmes:
rebased and merged in f19c13f34fae271731e2daa1e03292ce3075bd95

PR #1097, opened 2016-08-20
update slack.post to support token-based auth

### LINKED ISSUE #1091: slack.post should support passing through access token
right now, slack.post only supports the incoming webhook style of posting a slack message. in order to support the `chat.postMessage` API https://api.slack.com/methods/chat.postMessage, we will need to update slack.post in order to pass through a `token` field (the access token, probably acquired via a separate oauth exchange)

the rest of slack.post should handle the webhook-versus-postMessage distinction with grace. observe that both protocols require channel and text parameters, and both taken an optional icon_emoji parameter. i.e. the only distinction between the two is `username` versus `token`.

starpit:
This PR updates slack.post to support slack's oauth token style of authentication. If `params.token` is defined, the post assumes it is using token style authentication.

The only minor complexity here lies in the format of the body that slack expects, combined with some limitations of the formData npm that the request npm uses. The (preexisting) webhook-style API expects the formData to have the form `{ payload: body }`, whereas the slack API expects it to be of the form `body`, i.e. no payload wrapper. On top of this, the formData npm cannot tolerate nested objects -- this is why the slack.post code has always called `JSON.stringify` on the body. This is not needed (and causes harm) for the slack API-style of invocation.

Next, I removed the icon_emoji from the body if it is not defined by params. formData also barfs on `undefined`. This required updating the test to remove the expectation of seeing icon_emoji in the output.

The last change to the post.js code: I pulled the `console.log` calls so that they'd only execute in the webhook code path, in order to avoid leaking an unredacted auth token to the logs.

I also updated the docs/catalog.md to describe the new token-based auth support.

Finally, I updated installSlack  to add description and doclink fields. Here, I hoisted the bindTime parameters from the action annotation to the level of a package annotation. This has been a long-standing issue with the slack metadata that forced the Bluemix UI code to have some hacks: i.e. when prompting the user for a package binding, we had to hackishly look at the action annotations, even though the user had requested to make a package binding -- an operation independent of any action.

closes #1091 
PG1 229 [clean]

### DIFF
--- a/catalog/slack/post.js
+++ b/catalog/slack/post.js
@@ -14,9 +14,32 @@ function main(params) {
   var body = {
     channel: params.channel,
     username: params.username || 'Simple Message Bot',
-    text: params.text,
-    icon_emoji: params.icon_emoji
+    text: params.text
   };
+
+  if (params.icon_emoji) {
+    // guard against sending icon_emoji: undefined
+    body.icon_emoji = params.icon_emoji;
+  }
+
+  if (params.token) {
+    //
+    // this allows us to support /api/chat.postMessage
+    // e.g. users can pass params.url = https://slack.com/api/chat.postMessage
+    //                 and params.token = <their auth token>
+    //
+    body.token = params.token;
+  } else {
+    //
+    // the webhook api expects a nested payload
+    //
+    // notice that we need to stringify; this is due to limitations
+    // of the formData npm: it does not handle nested objects
+    //
+    body = {
+      payload: JSON.stringify(body)
+    };
+  }
 
   if (params.attachments) {
     body.attachments = params.attachments;
@@ -27,9 +50,7 @@ function main(params) {
 
   request.post({
     url: params.url,
-    formData: {
-      payload: JSON.stringify(body)
-    }
+    formData: body
   }, function(err, res, body) {
     if (err) {
       console.log('error: ', err, body);
--- a/catalog/slack/post.js
+++ b/catalog/slack/post.js
@@ -36,6 +36,9 @@ function main(params) {
     // notice that we need to stringify; this is due to limitations
     // of the formData npm: it does not handle nested objects
     //
+    console.log(body);
+    console.log("to: " + params.url);
+
     body = {
       payload: JSON.stringify(body)
     };
@@ -44,9 +47,6 @@ function main(params) {
   if (params.attachments) {
     body.attachments = params.attachments;
   }
-
-  console.log(body);
-  console.log("to: " + params.url);
 
   request.post({
     url: params.url,
--- a/docs/catalog.md
+++ b/docs/catalog.md
@@ -474,6 +474,7 @@ The `/whisk.system/slack/post` action posts a message to a specified Slack chann
 - `channel`: The Slack channel to post the message to.
 - `username`: The name to post the message as.
 - `text`: A message to post.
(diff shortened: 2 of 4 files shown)


catalog/slack/post.js
```diff
@@ -14,22 +14,43 @@ function main(params) {
...
+    //
+    body.token = params.token;
+  } else {
+    //
+    // the webhook api expects a nested payload
+    //
+    // notice that we need to stringify; this is due to limitations
+    // of the formData npm: it does not handle nested objects
+    //
+    console.log(body);
```
rabbah: intended to keep this?

starpit: for the webhook-style, it seems OK to leave it in. plus, as of now, the createSlackMessage test verifies the result by inspecting this log output.

starpit:
one unintended consequence of this PR: `post` becomes more than just posting message. the user can pass any api method as the url. we may wish to leverage this in the future.

csantanapr:
@starpit : 

1. we are at the moment on dual maintenance with "openwhisk-catalog" repo, need to submit PR to keep both repos in sync for now
https://github.com/openwhisk/openwhisk-catalog/blob/master/packages/slack/post.js
https://github.com/openwhisk/openwhisk-catalog/blob/master/tests/src/packages/slack/SlackTests.scala
FYI @jasonpet @houshengbo 

2. It's missing documentation updates for the new `auth` parameter
https://github.com/openwhisk/openwhisk/blob/master/docs/catalog.md#using-the-slack-package

3. It's missing parameter and annotation info for the new parameter `auth`
https://github.com/openwhisk/openwhisk/blob/master/catalog/installSlack.sh#L19

starpit:
@csantanapr thanks for the feedback! i have updated the PR for the main openwhisk repo, and amended the description to reflect the changes (updated docs; and updated the installSlack script so that the bindTime parameters are package annotations rather than action annotations (this has been a long-standing bug in the slack metadata); plus i added some descriptions and doclink fields

i'll look into the other repo now.

starpit:
@csantanapr i've created a parallel PR in openwhisk-catalog https://github.com/openwhisk/openwhisk-catalog/pull/122

csantanapr:
@starpit PG?

starpit:
i ran a PG. the only updates since then have been for documentation. but i'm happy to fire another one off it it helps: PG1 280

starpit:
@csantanapr PG1 280 is now clean. here is the corresponding openwhisk-catalog item: https://github.com/openwhisk/openwhisk-catalog/pull/122

docs/catalog.md
```diff
@@ -493,6 +493,9 @@ The following is an example of configuring Slack, creating a package binding, an
   $ wsk action invoke mySlack/post --blocking --result --param text 'Hello from OpenWhisk!'
   ```
 
+### Using the Slack token-based API
```
csantanapr: @starpit  please add `token` to the list of parameters above

>The parameters are as follows:
url: The Slack webhook URL.
channel: The Slack channel to post the message to.
username: The name to post the message as.
text: A message to post.

csantanapr:
other than adding `token` to the list of parameters LGTM.
@jasonpet can you review and if it's good go ahead and merge

csantanapr:
@starpit sorry for asking for a PG in https://github.com/openwhisk/openwhisk/pull/1097#issuecomment-241961419
I missed it when I scan the PR, it was at the top

jasonpet:
LGTM

csantanapr:
rebase and merge 429a53b7a2e5d5bcf919e34cffd0ab4c372b162e

PR #1103, opened 2016-08-21
add field level descriptions and doclink to github webhook (closes 1102)

### LINKED ISSUE #1102: github.webhook is missing field-level descriptions
in addition to adding some simple but helpful field-level descriptions. i suggest that we pioneer here a `doclink` field. this will be particularly helpful for the accessToken and events fields, which are quite cryptic on their own.

also, the main description field inaccurately renders GitHub as "github".

starpit:
This PR does the following:
1) fixes the main description to render the provider name more accurately as "GitHub", rather than the previous "github"
2) adds field-level descriptions to web hook
3) adds a `doclink` attribute to the accessToken and events fields

closes #1102
PG1 243 [clean]

### DIFF
--- a/catalog/installGit.sh
+++ b/catalog/installGit.sh
@@ -20,7 +20,7 @@ install "$CATALOG_HOME/github/webhook.js" \
     github/webhook \
     -a feed true \
     -a description 'Creates a webhook on GitHub to be notified on selected changes' \
-    -a parameters '[ {"name":"username", "required":true, "bindTime":true, "description": "Your GitHub username"}, {"name":"repository", "required":true, "bindTime":true, "description": "The name of your GitHub repository"}, {"name":"accessToken", "required":true, "bindTime":true, "description": "A webhook or personal token", "doclink": "https://github.com/settings/tokens/new"},{"name":"events", "required":true, "doclink": "https://developer.github.com/webhooks/#events"} ]' \
+    -a parameters '[ {"name":"username", "required":true, "bindTime":true, "description": "Your GitHub username or organization"}, {"name":"repository", "required":true, "bindTime":true, "description": "The name of your GitHub repository"}, {"name":"accessToken", "required":true, "bindTime":true, "description": "A webhook or personal token", "doclink": "https://github.com/settings/tokens/new"},{"name":"events", "required":true, "doclink": "https://developer.github.com/webhooks/#events"} ]' \
     -a sampleInput '{"username":"whisk", "repository":"WhiskRepository", "accessToken":"123ABCXYZ", "events": "push,commit,delete"}'
 
 waitForAll
--- a/catalog/installGit.sh
+++ b/catalog/installGit.sh
@@ -20,7 +20,7 @@ install "$CATALOG_HOME/github/webhook.js" \
     github/webhook \
     -a feed true \
     -a description 'Creates a webhook on GitHub to be notified on selected changes' \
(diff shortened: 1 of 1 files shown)


catalog/installGit.sh
```diff
@@ -19,8 +19,8 @@ waitForAll
 install "$CATALOG_HOME/github/webhook.js" \
     github/webhook \
     -a feed true \
-    -a description 'Creates a webhook on github to be notified on selected changes' \
-    -a parameters '[ {"name":"username", "required":true, "bindTime":true}, {"name":"repository", "required":true, "bindTime":true}, {"name":"accessToken", "required":true, "bindTime":true},{"name":"events", "required":true} ]' \
+    -a description 'Creates a webhook on GitHub to be notified on selected changes' \
+    -a parameters '[ {"name":"username", "required":true, "bindTime":true, "description": "Your GitHub username"}, {"name":"repository", "required":true, "bindTime":true, "description": "The name of your GitHub repository"}, {"name":"accessToken", "required":true, "bindTime":true, "description": "A webhook or personal token", "doclink": "https://github.com/settings/tokens/new"},{"name":"events", "required":true, "doclink": "https://developer.github.com/webhooks/#events"} ]' \
```
csantanapr: Should this description include organization?
"The Github username or organization"
What you think @bjustin-ibm ?

starpit: i think this is correct, i.e. that the "username" field can also be your organization name

starpit:
@csantanapr i have updated the PR as you have suggested, i.e. to mention that the username field can be your GitHub organization name, too.

catalog/installGit.sh
bjustin-ibm: As currently implemented, the `username` is the default "organization" when constructing the URL to the repo. However, if the `repository` value contains a `/` then it is interpreted as `<org>/<repo>`. 

This was done to allow the user to create a webhook on a repo that isn't a personal fork, without introducing an explicit new parameter called `organization` (which is unnecessary and confusing for personal forks).

This behavior is probably worth calling out in the `repository` parameter description.

bjustin-ibm: > Should this description include organization?
> "The Github username or organization"

This is actually not correct as a description for `username`. The `username` field is *always* used for authentication. Putting a value of an organization here will mess up authentication.

bjustin-ibm:
@starpit Sorry for chiming in late, but @csantanapr's suggestion about the `username` field is not correct. Please review these new comments I just posted:

https://github.com/openwhisk/openwhisk/pull/1103#discussion_r75861966

csantanapr:
Thanks Justin, see what I mean ?
I was not sure how repos under orga are supported. 
So it means that repository should be in the form of [org/]reponame were is  org is optional

starpit:
is there anything more you need from me on this?

bjustin-ibm:
> is there anything more you need from me on this?

@starpit I think it would be great to mention how the `repository` value can optionally include the organization. Apart from that, this LGTM.

starpit:
how about we put this in the sampleInput? right now, it reads `"repository": "WhiskRepository"`

how about changing this to `"repository": "[myOrganization/]myRepository"`

bjustin-ibm:
The square brackets were (I'm supposing) intended by Carlos to show that it is optional, and is not actually part of the value. In this case, the sample input should be `"repository": "myOrganization/myRepository"`.

However, I worry that using this sample input might imply that the organization is required.

How about changing the `repository` description to 
>The name of your GitHub repository. Optionally this can be prefixed with the GitHub organization followed by a slash (e.g. myOrganization/myRepository).

Too long?

starpit:
i've updated the PR in response to the comments. here is an idea of how these updates might be reflected in a UI: 
screen shot 2016-08-24 at 1 38 45 pm

bjustin-ibm:
rebased/squashed/merged as 328b9b138455ec77df023dcf029c8549db0d0ed4

Thanks @starpit!

csantanapr:
@starpit remember that we are in dual mode and need to sync with "openwhisk-catalog"

PR #1108, opened 2016-08-23
Attached jars

psuter:
Working implementation of attachments for jars in Java actions. For review only at this point to discuss design etc. I'm only moderately satisfied for instance with my attempt at separating attachment API methods in `AttachmentStore`, for instance (how long can we really pretend that `ArtifactStore` is db-agnostic?).

It's likely that supporting attachments for other kinds of actions will lead to at least some refactoring to isolate common patterns. The `Attachment[T]` class is already an example of that.

Note that this PR is fully backwards compatible: the jar is always retrieved to build the Java action entity. As a consequence, it does not e.g. involve streaming beyond the DB <-> entity interactions.


tests/src/whisk/core/database/test/CouchDbRestClientTests.scala
```diff
@@ -75,48 +87,96 @@ class CouchDbRestClientTests extends FlatSpec
...
+        val doc = JsObject("greeting" -> JsString("hello"))
+        val attachmentName = "misc"
+        val attachmentType = ContentTypes.`text/plain(UTF-8)`
+        val attachment = ("""
+            | There once was an OpenWhisk action
+            | Without an exec as companion
+            |   It's not in the document
+            |   It is an attachment!
+            | Stream it for fast invocation.
+        """).stripMargin
```
sjfink: This poem is awful :-1:

psuter: Could be worse: https://en.wikipedia.org/wiki/Vogon#Poetry

Happy to replace with something more prosaic in any case.

rabbah: there once as an action
it couldn't stream its bytes
now it has an attachment
and that's alright
ready for some fun?

sjfink: There once was an action named foo,
With no exec, he didn't know what to do!
   But don't panic my friend,
   Just stream it till the end!
And all's well and good.   Phew!

sjfink:
Should this be moved to 'review' from 'wip'?

psuter:
set to `review`

common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
```diff
@@ -194,4 +219,28 @@ trait DocumentFactory[W] extends InMemoryCache[W] {
...
+        Try {
+            require(db != null, "db defined")
+            require(doc != null, "doc undefined")
+        } map { _ =>
+            val sink = StreamConverters.fromOutputStream(() => outputStream)
+            db.readAttachment[IOResult](doc, attachmentName, sink).map {
+                case (_, r) =>
+                    if (!r.wasSuccessful) {
+                        // FIXME...
+                        // Figure out whether OutputStreams are even a decent model.
```
rabbah: Do we need an issue for this?

common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
```diff
@@ -149,13 +157,30 @@ trait DocumentFactory[W] extends InMemoryCache[W] {
...
     }
 
+    def attach[Wsuper >: W](db: AttachmentStore[Wsuper], doc: DocInfo, attachmentName: String, contentType: ContentType, bytes: InputStream)(
+        implicit transid: TransactionId): Future[DocInfo] = {
+
+        Try {
+            require(db != null, "db undefined")
+            require(doc != null, "doc undefined")
+        } map { _ =>
+            implicit val logger: Logging = db.asInstanceOf[ArtifactStore[_]]
```
rabbah: is the cast necessary?

rabbah:
@psuter and I agreed to collapse the ArtifactStore and AttachmentStore abstraction as part of this pull request. That this interface will work for other stores (e.g., relational) is already questionable as there are Couch details that permeate once you set aside the put/get/del methods.

psuter:
Git surgery has completed. Travis and PG1 403 will tell how the patient is doing.

PR #1109, opened 2016-08-23
Initial revision of documentation for how to create feeds.

sjfink:


docs/feeds.md
```diff
@@ -0,0 +1,105 @@
...
+There are at least 3 architectural patterns for creating a feed: **Hooks**, **Polling** and **Connections**.
+
+## Hooks
+In the *Hooks* pattern, we set up a feed using a [webhook](https://en.wikipedia.org/wiki/Webhook) facility exposed by another service.   In this strategy, we configure a webhook on an external service to POST directly to a URL to fire a trigger.  This is by far the easiest and most attractive option for implementing low-frequency feeds.
+
+<!-- The github feed is implemented using webhooks.  Put a link here when we have the open repo ready -->
+
+## Polling
+In the "Polling" pattern, we arrange for a OpenWhisk *action* to poll an endpoint periodically to fetch new data.
+This is easy to build, but is low performance, and is limited by the polling interval.
```
glikson: This is somewhat argumentative (and leaving quite a bit room for potential misinterpretation).. consider rephrasing.

docs/feeds.md
```diff
@@ -0,0 +1,105 @@
...
+
+There are at least 3 architectural patterns for creating a feed: **Hooks**, **Polling** and **Connections**.
+
+## Hooks
+In the *Hooks* pattern, we set up a feed using a [webhook](https://en.wikipedia.org/wiki/Webhook) facility exposed by another service.   In this strategy, we configure a webhook on an external service to POST directly to a URL to fire a trigger.  This is by far the easiest and most attractive option for implementing low-frequency feeds.
+
+<!-- The github feed is implemented using webhooks.  Put a link here when we have the open repo ready -->
+
+## Polling
+In the "Polling" pattern, we arrange for a OpenWhisk *action* to poll an endpoint periodically to fetch new data.
```
glikson: *an* OpenWhisk action

docs/feeds.md
```diff
@@ -0,0 +1,105 @@
...
+This is easy to build, but is low performance, and is limited by the polling interval.
+
+## Connections
+In the "Connections"  pattern, we stand up a separate service somewhere that maintains a persistent connection to a feed source.    The connection based implementation might interact with a service endpoint via long polling, or to set up a push notification.
```
glikson: Consider choosing a different name for this pattern. Maybe "bridge"? If I understand correctly, the essence is that there is a separate runtime that monitors the external system (one way or another). Type of connection is an implementation detail.

sjfink: Hmm.   The essence I'm trying to convey is that some non-serverless, long running server has to stay up to maintain a persistent connection.     Hence the term "connection".  ?   Other opinions on connection vs. bridge?

docs/feeds.md
```diff
@@ -0,0 +1,106 @@
...
+#  Implementing Feed Actions
+
+The *feed action* is a normal OpenWhisk *action*, but it should accept the following parameters:
+* **lifecycleEvent**: one of 'CREATE', 'DELETE', 'PAUSE', or 'UNPAUSE'
+* **triggerName**: the fully-qualified name of the trigger which contains events produced from this feed.
+* **authKey**: the Basic auth credentials of the OpenWhisk user who owns the trigger just mentioned
+
+The feed action can also accept any other parameters it needs to manage the feed.  For example the cloudant changes feed action expects to receive parameters including *'dbname'*, *'username'*, etc.
+
+When the user creates a trigger from tthe CLI with the **--feed** parameter, the system automatically invokes the feed action with the appropriate parameters.
```
csantanapr: typo `tthe CLI`

docs/feeds.md
```diff
@@ -0,0 +1,106 @@
...
+
+`wsk action invoke mycloudant/changes -p lifecycleEvent CREATE -p triggerName T -p authKey <userAuthKey> -p password <password value from mycloudant binding> -p username <username value from mycloudant binding> -p dbName mytype`
+
+The feed action named *changes* takes these parameters, and is expected to take whatever action is necessary to set up a stream of events from Cloudant, with the appropriate configuration, directed to the trigger *T*.    
+
+For the Cloudant *changes* feed, the action happens to talk directly to a *cloudant trigger* service we've implemented with a connection-based architecture.   We'll discuss the other architectures below.
+
+A similar feed action protocol occurs for `wsk trigger delete`.    
+
+We have not yet implemented pause and unpause, but they will be similar.
```
csantanapr: Do we document in official docs  we publish things we have not implemented yet?

docs/feeds.md
```diff
@@ -0,0 +1,106 @@
...
+
+# Implementing Feeds with Hooks
+
+Setting up a feed via a hook is *by far* the easiest way to prototype a new feed,
```
csantanapr: *by far* the easiest, according to who? a PH.D? :-)
I think this should be written in affirmative way. And why talk about prototypes? people would think they should not be used for implementing a feed as it's only useful for doing "prototypes".

I would just remove the sentence

docs/feeds.md
```diff
@@ -0,0 +1,106 @@
...
+if the even producer supports a webhook
```
csantanapr: typo `even producer`

docs/feeds.md
```diff
@@ -0,0 +1,106 @@
...
+facility.
+
+
+With this method there is _no need_ to stand up any persistent service outside of OpenWhisk.  All feed management happens naturally though stateless OpenWhisk *feed actions*, which negotiate
+directly with a third part webhook API.
+
+When invoked with `CREATE`, the feed action simply installs a webhook for some other service, asking the remote service to POST notifications to the appropriate `fireTrigger` URL in OpenWhisk.
+
+The webhook should be directed to send notifications to a URL such as:
+`POST /namespaces/{namespace}/triggers/{triggerName}`
```
csantanapr: I would add that the JSON body of the HTTP POST will be the parameters use to fire the trigger, that will be forward to the actions associated via rules.

docs/feeds.md
```diff
@@ -0,0 +1,106 @@
...
+
+The feed action named *changes* takes these parameters, and is expected to take whatever action is necessary to set up a stream of events from Cloudant, with the appropriate configuration, directed to the trigger *T*.    
+
+For the Cloudant *changes* feed, the action happens to talk directly to a *cloudant trigger* service we've implemented with a connection-based architecture.   We'll discuss the other architectures below.
+
+A similar feed action protocol occurs for `wsk trigger delete`.    
+
+We have not yet implemented pause and unpause, but they will be similar.
+
+# Implementing Feeds with Hooks
```
csantanapr: Should we use `with Webhooks` ?

docs/feeds.md
```diff
@@ -0,0 +1,106 @@
...
+
+# Implementing Feeds with Polling
+
+It is possible to set up an OpenWhisk *action* to poll a feed source entirely within OpenWhisk, without the need to stand up any persistent connections or external service.
+
+For feeds where a webhook is not available, but do not need high-volume or low latency response times, polling is an attractive option.
+
+To set up a polling-based feed, the feed action takes the following steps when called for `CREATE`:
+1.   The feed action sets up a periodic trigger (*T*) with the desired frequency, using the `whisk.system/alarms` feed.
+2.   The feed developer creates a `pollMyService' action which simply polls the remote service and returns any new events.
```
csantanapr: typo: single quote `'` instead of back tick ` at the end of pollMyService

docs/feeds.md
```diff
@@ -0,0 +1,106 @@
...
+3.  The feed action sets up a *rule*
+*T -> pollMyService*.
+
+That's it.   This procedure implements a polling-based trigger entirely using OpenWhisk actions, without any need for a separate service.
```
csantanapr: Don't think we need the "That's it."  portion.

docs/feeds.md
```diff
@@ -0,0 +1,106 @@
...
+
+It is possible to set up an OpenWhisk *action* to poll a feed source entirely within OpenWhisk, without the need to stand up any persistent connections or external service.
+
+For feeds where a webhook is not available, but do not need high-volume or low latency response times, polling is an attractive option.
+
+To set up a polling-based feed, the feed action takes the following steps when called for `CREATE`:
+1.   The feed action sets up a periodic trigger (*T*) with the desired frequency, using the `whisk.system/alarms` feed.
+2.   The feed developer creates a `pollMyService' action which simply polls the remote service and returns any new events.
+3.  The feed action sets up a *rule*
+*T -> pollMyService*.
```
csantanapr: I think this should be in one sentence in the 3. bullet

docs/feeds.md
```diff
@@ -0,0 +1,106 @@
...
+
+That's it.   This procedure implements a polling-based trigger entirely using OpenWhisk actions, without any need for a separate service.
+
+# Implementing Feeds via Connections
+
+The previous 2 architectural choices are simple and easy to implement. However, if you want a high-performance feed, there is no substitute for persistent connections and long-polling or similar techniques.
+
+Since OpenWhisk actions are stateless,  an action cannot maintain a persistent connection to a third party. Instead, we are forced to stand up a separate service (outside of OpenWhisk) that runs all the time.   We call these *provider services*.  A provider service can maintain connections to third party event sources that support long polling or other connection-based notifications.
```
csantanapr: I would rephrase the `forced` portion
We are never forced, we just have trade offs :-)

csantanapr: Since OpenWhisk actions are stateless,  an action cannot maintain a persistent connection to a third party.
I would say that cannot a persistent connection not because they are stateless but because they have a limit on how long they can run.
Maybe:
Since OpenWhisk actions are designed to be short lived they can't maintain a persistent connection to a third party service. To implement a feed based on long running process you would required to use a service outside of OpenWhisk that runs all the time. We call these *event provider* services.

docs/feeds.md
```diff
@@ -0,0 +1,106 @@
...
+The provider service should provide a REST API that allows the OpenWhisk *feed action* to control the feed.   The provider service acts as a proxy between the event provider and OpenWhisk -- when it receives events from the third party, it sends them on to OpenWhisk by firing a trigger.
+
+The Cloudant *changes* feed is the canonical example -- it stands up a `cloudanttrigger` service which mediates between Cloudant notifications over a persistent connection, and OpenWhisk triggers.
+<!-- TODO: add a reference to the open source implementation -->
+
+The *alarm* feed is implemented with a similar pattern.
+
+The connection-based architecture is the highest performance option -- but it's far more difficult to operate and maintain than the polling and hook architectures.   The provider service must be production-quality: it must be highly-available and fault-tolerant.
```
csantanapr: This sentence has too much internal info that user doesn't care. 

The connection-based architecture is the highest performance option.
We already established at the beginning of the section that was the high performance feed.
So no need to repeat here.
 
-- but it's far more complex to operate and maintain than the polling and hook architectures   The provider service must be production-quality: it must be highly-available and fault-tolerant. 

Everything needs to be production-quality so not much explanation why is called out here. 
You can say the same as webhooks, if they don't have a reliable way or hihg performance way to call webhooks they will have problems.

Maybe what you want to tell users  that if your are going to use a third party service that uses the connection-based architecture or you going to implemented on your own that for large amount of traffic you need to take into consideration performance and throttle implications to have a high quality event stream into OpenWhisk.

csantanapr:
@sjfink I did a first pass and provided comments

docs/feeds.md
sjfink: I guess I prefer the more general term "hook" to cover more possible cases, but I don't feel that strongly.

csantanapr: Yeah I was also on the fence, because it can be anything. But I think using the term Webhook helps with SEO and page search. and for "web" it means http based hooks and most common way that people refer to.

docs/feeds.md
```diff
@@ -85,22 +82,23 @@ For feeds where a webhook is not available, but do not need high-volume or low l
...
 The provider service should provide a REST API that allows the OpenWhisk *feed action* to control the feed.   The provider service acts as a proxy between the event provider and OpenWhisk -- when it receives events from the third party, it sends them on to OpenWhisk by firing a trigger.
 
 The Cloudant *changes* feed is the canonical example -- it stands up a `cloudanttrigger` service which mediates between Cloudant notifications over a persistent connection, and OpenWhisk triggers.
 <!-- TODO: add a reference to the open source implementation -->
 
 The *alarm* feed is implemented with a similar pattern.
 
-The connection-based architecture is the highest performance option -- but it's far more difficult to operate and maintain than the polling and hook architectures.   The provider service must be production-quality: it must be highly-available and fault-tolerant.    
+The connection-based architecture is the highest performance option -- 
+but imposes more overhead on operations compared to the polling and hook architectures.
```
csantanapr: Changes LGTM, you tell me if you want more 👀  to review it or call ready for press.

sjfink:
rebased and squashed -- ready to go.

csantanapr:
merged 8563176ac922744a73c7f638e45f58e613792f15

PR #1115, opened 2016-08-24
Tighten the policy to pull images when creating a container

markusthoemmes:
String checking is bad, type-checking is good!

### DIFF
--- a/core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
+++ b/core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
@@ -44,6 +44,7 @@ import whisk.core.entity.WhiskEntityStore
 import whisk.core.entity.NodeJS6Exec
 import akka.event.Logging.LogLevel
 import akka.event.Logging.InfoLevel
+import whisk.core.entity.BlackBoxExec
 
 /**
  * A thread-safe container pool that internalizes container creation/teardown and allows users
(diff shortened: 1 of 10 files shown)


lehoanganh:
it means, for each invoking docker action, invoker will try to pull the container, and doesn't care if the image is there locally or not?

markusthoemmes:
Yes, that's semantically preserving what it did before. Docker will fast forward though, and almost immediatly notice that the layers are already there.

lehoanganh:
2 seconds for the check ;)

fxulusoy:
Checking the image manually whether it is there or not does cost some time as well :)

markusthoemmes:
@lehoanganh because an explicit `pull` at least checks, if the latest version is around.

core/dispatcher/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -483,19 +484,18 @@ class ContainerPool(
         val key = makeKey(action, auth)
         val warmedContainer = if (limits.memory == defaultMemoryLimit && imageName == nodeImageName) getWarmNodejsContainer(key) else None
         val containerName = makeContainerName(action)
-        val con = warmedContainer getOrElse makeGeneralContainer(key, containerName, imageName, limits)
+        val con = warmedContainer getOrElse makeGeneralContainer(key, containerName, imageName, limits, action.exec.isInstanceOf[BlackBoxExec])
```
cbickel: Not sure what's cheaper, but `action.exec.kind == Exec.blackbox` would be another solution.

jasonpet:
see pg2 9

rabbah:
Squashed and merged as https://github.com/openwhisk/openwhisk/compare/429a53b7a2e5...60d1bad5ea0b.

PR #1116, opened 2016-08-24
Controller publishes digested user activation concurrency

perryibm:
This PR causes the controller to publish its notion of activation concurrency level per user.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
+++ b/core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
@@ -21,8 +21,9 @@ import scala.concurrent.duration.DurationInt
 
 import akka.actor.ActorSystem
 import spray.json.DefaultJsonProtocol._
-import spray.json.pimpString
+import spray.json._
 import whisk.common.ConsulClient
+import whisk.common.ConsulKV.ControllerKeys
 import whisk.common.ConsulKV.InvokerKeys
 import whisk.common.ConsulKV.LoadBalancerKeys
 import whisk.common.Logging
@@ -49,7 +50,7 @@ class ActivationThrottler(consulServer: String, concurrencyLimit: Int)(
      */
     private var userActivationCounter = Map.empty[String, Long]
 
-    private val healthCheckInterval = 2.seconds
+    private val healthCheckInterval = 5.seconds
     private val consul = new ConsulClient(consulServer)
 
     /**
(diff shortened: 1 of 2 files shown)


core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
```diff
@@ -90,6 +92,21 @@ class ActivationThrottler(consulServer: String, concurrencyLimit: Int)(
...
+    private def publishUserConcurrentActivation() = {
+        val subjects = userActivationCounter.keySet toList
+        val groups = subjects.groupBy { user => user.substring(0, 1) } // Any sort of partitioning will be ok for monitoring
+        groups.keySet map { prefix =>
+            val key = ControllerKeys.userActivationCountKey + "/" + prefix
+            val users = groups.getOrElse(prefix, Set())
+            val items = users map { u => (u, JsNumber(userActivationCounter.get(u) getOrElse 0L)) }
+            consul.kv.put(key, JsObject(items toMap).compactPrint)
+        }
+    }
```
markusthoemmes: This should be equivalent to:

```
userActivationCounter.groupBy(_._1.take(1)).map {
  case (prefix, items) =>
    val key = ControllerKeys.userActivationCountKey + "/" + prefix
    consul.kv.put(key, items.toJson.compactPrint)
}
```

(not tried to compile)

core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
```diff
@@ -100,6 +117,8 @@ class ActivationThrottler(consulServer: String, concurrencyLimit: Int)(
                     val loadbalancerCount = loadbalancerActivationCount(subject)
                     subject -> (loadbalancerCount - invokerCount)
             }
+            publishUserConcurrentActivation()
```
markusthoemmes: Given that above method returns a `List[Future[Unit]]` this will not "block" the scheduler long enough and could potentially lead to overlapping calls.

Can we change `publishUserConcurrentActivation` to return a `Future[Unit]`?

core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
```diff
@@ -101,5 +115,7 @@ class ActivationThrottler(consulServer: String, concurrencyLimit: Int)(
                     subject -> (loadbalancerCount - invokerCount)
             }
         }
+        Future.sequence(publishUserConcurrentActivation())
```
markusthoemmes: Minor nit, but should the `Future.sequence` move to the `publishUserConcurrentActivation` function?
`List[Future[Unit]]` doesn't really make sense as a return type.

perryibm:
Adjusted.  Testing PG 339

markusthoemmes:
rebased and merged in a0dd8458d03adf63c37e73373e54a420397e57d3

PR #1120, opened 2016-08-24
Supercedes 1071 - lifts active ack processing to avoid race

### LINKED ISSUE #1067: phase ordering of posting message to invoker and waiting for active ack response

### LINKED ISSUE #1140: Convert all blocking activation requests to non-blocking after 60 seconds

rabbah:
Lift listener for active ack responses in action activations before the post to loadbalancer - this avoids a race where the response comes back before the listener becomes active. Use activation id generator to allow testing of blocking invokes. Adds tests for active path (previously missing). Refactors the logic for active ack processing.

Also in this pull request:
1. fix reading from consul in whisk config: allow read from consul to fail if consul is not responsive as long as required properties are set.
2. convert all blocking invokes to non-blocking after 60 seconds (regardless of the action timeout limit).

Thanks to @markusthoemmes and @cbickel for feedback.

Replaces #1071.
Fixes #1067.
Fixes #1140.


cbickel:
LGTM

rabbah:
PG2 13 approved.

perryibm:
rebased/merged

PR #1137, opened 2016-08-26
sync docs with bluemix

csantanapr:
minor changes to make sync docs better going forward.
please review and merge @bjustin-ibm 
make any changes you find as you do the merge

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -265,6 +265,7 @@ This example invokes a Yahoo Weather service to get the current conditions at a 
 
   This example also shows the need for asynchronous actions. The action returns a Promise to indicate that the result of this action is not available yet when the function returns. Instead, the result is available in the `request` callback after the HTTP call completes, and is passed as an argument to the `resolve()` function.
 
+
 2. Run the following commands to create the action and invoke it:
   ```
   $ wsk action create weather weather.js
@@ -480,6 +481,7 @@ $ wsk action invoke --blocking --result helloJava --param name World
 
 **Note:** If the JAR file has more than one class with a main method matching required signature, the CLI tool uses the first one reported by `jar -tf`.
 
+
 ## Creating Docker actions
 
 With OpenWhisk Docker actions, you can write your actions in any language.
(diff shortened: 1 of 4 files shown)


rabbah:
> three steps are required to set up a custom binary for use by OpenWhisk

Would become 5 steps. That's why I removed 4-5.

csantanapr:
>Would become 5 steps. That's why I removed 4-5.
Why not remove step 2 and 3, and just leave it at one single step that would be more awesome.

I will submit an update and remove the numbers for step 4 and 5. I leave the steps numberless

bjustin-ibm:
@csantanapr @rabbah I see the "ready" tag but [the last comment](https://github.com/openwhisk/openwhisk/pull/1137#issuecomment-242890201) seems like there is still a question to be resolved.

csantanapr:
@bjustin-ibm sorry I made the change to address @rabbah comment, but I made the mistake to squashed it.
You can see now that there is no step 4, or 5 in the docker section.
@rabbah can you confirm you are OK now?

docs/actions.md
```diff
@@ -567,16 +570,16 @@ The action may be invoked as any other OpenWhisk action.
   }
   ```
 
-If you update the Docker image, run `buildAndPush.sh` to refresh the image on Docker Hub, then run `wsk action update` to make the system to fetch the new image. New invocations will start using the new image and not a warm image with the old code.
-
+  To update the Docker image, run `buildAndPush.sh` to refresh the image on Docker Hub, then run `wsk action update` to make the system to fetch the new image. New invocations will start using the new image and not a warm image with the old code.
```
rabbah: This may be interpreted to mean that in the absence of `update` the system uses the old image. That's not the case if for example the image uses the `latest` tag. Every cold invoke will pull the latest tag even if action is not updated.

csantanapr: @rabbah Please provide the text you would like to see here to make clear to the user.

My impression was that the only thing users need to worry about when they have a new image for the same action was to run `wsk action update`, but it looks they have to do something else?

csantanapr: Ah I get it now. It's a guarantee then, when the user does a `wsk action update` it's the telling the system, is telling the system that any new invocations force a docker pull and pull down new `latest` even if there is a container already running with the `init` already performed.

csantanapr: @rabbah How about something like this:

To update the Docker action, run buildAndPush.sh to refresh the image on Docker Hub, this will allow next time the system pulls your Docker image to run the new code for your action. If there are no warm containers any new invocations will use the new Docker image. Take into account that if there is a warm container using a previous version of your Docker Image it will continue to use this unless you run `wsk action update`, this will indicate to the system that for any new invocations force a `docekr pull` resulting on pulling a your new Docker image.

csantanapr:
@rabbah I don't know what else todo with this PR, so go ahead and do what you think is best with it.

rabbah:
OK thanks for the tag - the title says "sync docs with blue" - is this syncing with bluemix doc?

docs/actions.md
rabbah: The comment above is a good one - should be in the text.

rabbah:
Also I'm confusing (maybe more so by the title). Ignoring white space there's only one change: https://github.com/openwhisk/openwhisk/pull/1137/files?w=1.

The one comment that nice explains the docker container update/lifecycle I assume I can put in the doc. Can you clarify though what we're synchronizing against?

csantanapr:
sync with bluemix docs, having this small changes on spaces and and new lines, make my work to manually sync the content between bluemix and openwhisk easier for me.

rabbah:
thanks! ok to add the paragraph you wrote in the comment about update?

csantanapr:
@rabbah that's why I thought this PR was going to be a no brainer, because it was just white spaces and and end of lines nothing else. Nothing about new content, nothing about docker that could be done in another issue/PR.
That's why I'm bit frustrated

rabbah:
OK so you want to leave it as is? OK with me. But your explanation is a good one, shame to lose it.

PR #1139, opened 2016-08-27
Change response code for non-existent package to not found from not allowed

rabbah:
The entitlement check for packages failed client requests with not authorized responses
when checking a package that does not exist. This served to neither confirm or deny the
existence of the package --- but has been confusing. Changed the response to not found should
the package in fact not exist. This removes an inconsistency as well between the local and
remote entitlement flows.

PG2-20 approved.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/entitlement/PackageCollection.scala
+++ b/core/controller/src/main/scala/whisk/core/entitlement/PackageCollection.scala
@@ -18,14 +18,16 @@ package whisk.core.entitlement
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+
+import spray.http.StatusCodes.NotFound
 import whisk.common.TransactionId
+import whisk.core.controller.RejectRequest
+import whisk.core.database.NoDocumentException
 import whisk.core.entity.DocId
+import whisk.core.entity.EntityName
+import whisk.core.entity.WhiskEntity
 import whisk.core.entity.WhiskPackage
 import whisk.core.entity.types.EntityStore
-import scala.concurrent.Promise
-import whisk.core.entity.WhiskEntity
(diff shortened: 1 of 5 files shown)


core/controller/src/main/scala/whisk/core/entitlement/PackageCollection.scala
```diff
@@ -104,7 +107,14 @@ class PackageCollection(entityStore: EntityStore) extends Collection(Collection.
         } onFailure {
             case t: NoDocumentException =>
                 info(this, s"the package does not exist")
-                promise.success(false)
+                // if owner, reject with not found, otherwise reject fail the promise
+                // to reject with unauthorized (this prevents leaking information about
+                // packages in other namespaces
+                if (isOwner) {
+                    promise.failure(RejectRequest(NotFound))
```
markusthoemmes: I think, this time we can really get around using a Promise for this and achieve the same using Future composition?

rabbah: Yes but why?

markusthoemmes: Because Future composition generally makes for easier to read code and makes for more deterministic code.

Maybe this clarifies the point: to me `Promise` is to `Future` as `var` to `val` (regarding determinism in the code).

rabbah:
New PG2-22 pending.

rabbah:
PG2-22 approves.

rabbah:
new pg pending.

rabbah:
PG2-33 approved.

markusthoemmes:
rebased and merged in ba51b5c032ee39de36fef7db7a9f3ba806cae2e4

PR #1141, opened 2016-08-27
Rebase action proxy (dockerskeleton) on alpine image

rabbah:
Reduce image size for docker actions skeleton from 619.8MB to 86.78MB.
This makes the example docker image size equally smaller.
Use the base image for python actions but not Swift since the latter not yet ready for a diet.

PG1-321 approved (node failing test is due to #1142 and not this pr).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,7 @@ node_modules
 ansible/environments/mac/hosts
 ansible/db_local.ini*
 ansible/tmp/*
+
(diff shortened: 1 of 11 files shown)


csantanapr:
LGTM

csantanapr:
With this change Inwill push a new dockerskeleton to docker hub and then create public repos using dockerskeleton as base for the other examples we have on docker hub

PR #1146, opened 2016-08-28
Move note about not sharing sensitive information inside the comment block

rabbah:
when opening new issues, this line would appear outside the comment block.

> note: obfuscate all passwords, credentials or other confidential information

and hence appear in the issue description. Move it instead inside the template comment block and put it right at the top.


csantanapr:
rebase and merge d2cf98e5b5e3b6eddc4253761fc9bc9de03c42be

PR #1150, opened 2016-08-28
Remove sleeps

rabbah:
Remove unnecessary sleeps in invoker:
- a 3s sleep before calling /init for black box containers is not necessary since the HTTP call will retry.
- slack sleeps after init also not necessary since if container respond to inits, it's ready to run.
- sleep during nio channel draining should not be necessary.

And adds tests using blackbox actions for containers that misbehave to confirm they will timeout with expected exception. Three cases are tested:

1. blackbox that doesn't implement a proper proxy (doesn't run a proxy at all)
2. blackbox that doesn't respond to init (implements /init but does not respond)
3. blackbox that doesn't respond to run (implements /run but does not respond)

### DIFF
--- a/tests/src/actionContainers/DockerExampleContainerTests.scala
+++ b/tests/src/actionContainers/DockerExampleContainerTests.scala
@@ -16,15 +16,17 @@
 
 package actionContainers
 
+import java.util.concurrent.TimeoutException
+
 import org.junit.runner.RunWith
 import org.scalatest.junit.JUnitRunner
 
 import ActionContainer.withContainer
+import common.WskActorSystem
 import spray.json.JsNumber
 import spray.json.JsObject
 import spray.json.JsString
-
-import common.WskActorSystem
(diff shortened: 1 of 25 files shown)


tests/src/actionContainers/DockerExampleContainerTests.scala
```diff
@@ -83,4 +85,64 @@ class DockerExampleContainerTests extends ActionProxyContainerTestUtils with Wsk
             case (o, _) => o should include("This is an example log message from an arbitrary C program!")
         }, 2)
     }
+
+    behavior of "bad containers"
+
+    it should "timeout init with exception" in {
+        val (out, err) = withContainer("badaction") { c =>
+            an[TimeoutException] should be thrownBy {
```
markusthoemmes: `a[TimeoutException]` (multiple occurences)

rabbah:
pg2-53 approved.

PR #1154, opened 2016-08-29
Wrap consul read of configuration with Try 

rabbah:
Allows API unit tests to run once again without a deployment by allowing consul read to fail during configuration - the invariant that all required properties must be defined for config to be valid is not changed.


perryibm:
pg2 - 65

rabbah:
pg2-69 approved.

perryibm:
rebase/merged

PR #1156, opened 2016-08-29
All blocking invokes convert to non-blocking after 60 seconds 

### LINKED ISSUE #1140: Convert all blocking activation requests to non-blocking after 60 seconds

rabbah:
Regardless of action timeout, controller converts blocking requests to non-blocking after min(60 seconds, action timeout). The API responds with the activation id and 202. Nginx conf is adjusted accordingly (70 seconds = 60 + 10 seconds to allow controller to terminate connection vs nginx).

Fixes #1140.
Blocked by #1120 (must be merged first).


rabbah:
PG2-45 approved.

perryibm:
pg2 - 66

perryibm:
merged

PR #1158, opened 2016-08-29
preping for new location of echo in utils/echo

csantanapr:


fxulusoy:
@csantanapr Why don't you fix the other files (especially docs) which reference to samples/echo?

csantanapr:
@fxulusoy That's a bigger effort, that will take care when all changes that need to be `catalog.md` doc in the `openwhisk-catalog` repo as part of the transition. When the `catalog` folder get's removed from this repo, then the `catalog.md` in this repo would have the new content based on `openwhisk-catalog`

csantanapr:
PG2:42 running...

csantanapr:
@jasonpet Travis is breaking because `utils` package doesn't exist, only `util`.
I don't think would be a good idea to have two packages listing. As we discuss I think best thing to do is to implement a temporary `samples/echo` in `openwhisk-catalog` until we make the complete switched over.

PR #1161, opened 2016-08-30
Tweak citool.

rabbah:
Small tweak to citool: Detect if tests ran, if not, report message as such rather than a misleading all tests passed message.


psuter:
Merged as 7b2425fc5debf5fc22a4cfc33f0a66517b1f3159

PR #1166, opened 2016-08-31
native binaries with dockerskeleton

rabbah:
Allow docker skeleton to be initialized by arbitrary gzipper tar ball as long as it is executable within the openwhisk/dockerskeleton container. This has a few advantages: 1. don't need to pull image from dockerhub (unpacking a tar ball should be faster but I did not measure it yet), 2. `exec` can be precompiled making cold start faster (same comment re measuring), 3. could keep the container image small (don't need to install gcc, swiftc, etc. to compile various flavors of native executables). Of course, the executable must be statically linked if it has library dependencies. For more complex requirements, one may still use the more general docker actions. As such, this is NOT a replacement for general docker actions.

As a general optimization, the dockerskeleton images should be pushed to the invoker anyway to save fetching layers for blackbox actions that use the openwhisk docker action sdk.

In anticipation of action attachments, this is modeled after the java jars.

Here's an example of how to use it:

```
on host:
# cd /Users/Shared
# docker run -v /Users/Shared/:/workspace -it dockerskeleton bash

now inside container
# cd /workspace
# apk add --no-cache bzip2-dev gcc libc-dev
# echo '#include <stdio.h>
int main(int argc, char *argv[]) { printf("hello blackbox\n"); printf("%s\n", argv[1]); }' > bb.c
# gcc -o exec -static bb.c
# tar zcf bb.tgz exec

exit container to return to host /Users/Shared
# wsk action create bb openwhisk/dockerskeleton --docker --lib bb.tgz
ok: created action bb
# wsk action invoke bb --blocking --result -p msg hello
{
    "msg": "hello"
}
```

One caveat is that we exempt blackbox actions from log sentinel filtering but this is a case where we should filter the sentinel out. In fact this will be the case for any blackbox action that uses the docker action sdk. Otherwise the logs will look like this:

```
"logs": [
        "2016-08-31T03:45:48.44970514Z  stderr: XXX_THE_END_OF_A_WHISK_ACTIVATION_XXX",
        "2016-08-31T03:45:48.449739089Z stdout: hello blackbox",
        "2016-08-31T03:45:48.449742919Z stdout: XXX_THE_END_OF_A_WHISK_ACTIVATION_XXX"
]
```

It's also worth noting that in general this extends the `initialization` support to black box actions - while currently on initialization, the container does not receive any data, it will now receive whatever payload is attached to the action via `--lib`.

FYI @psuter, @paulcastro.


sjfink:
How are dynamically linked libraries going to work?

rabbah:
must be statically linked - added to the description.

sjfink:
so we need the other facility too right?   For software that depends on dlls?

tests/src/actionContainers/ActionProxyContainerTests.scala
```diff
@@ -151,6 +155,25 @@ class ActionProxyContainerTests extends BasicActionRunnerTests with WskActorSyst
         })
     }
 
+    it should "extract and run a compatible tgz exec" in {
+        val tgz = FileUtils.readFileToByteArray(new File(TestUtils.getTestActionFilename("blackbox.tgz")))
```
sjfink: Should we support .zip format?

mbehrendt: i think we should

rabbah:
pg2-78 approved.

tests/src/actionContainers/ActionProxyContainerTests.scala
rabbah: for general black box containers, the archive format is up to the user - it need not be either gzip or zip; the only constraint is that it fits without our set limits.

for these native activation containers, since the binary must be built inside a compatible container (namely the whisk/dockerskeleton), i see less value in adding zip support - there should be just one format for ease of maintenance.

rabbah:
rebased with new pg2/147.

perryibm:
Reminder:  update go cli (python gone)

rabbah:
@dubeejw I recall in previous revs of the CLI in Go there was support for --lib. Is that correct? Is it worth restoring the commit or is this an "easy" enough change that can be done fresh?

rabbah:
CLI "--lib" support may end up part of #1403 and could be reused here.

Since the zip file uploads for the other runtimes is centered around zip, I will change from tgz to zip so the formats are consistent (if we're only going to support one now it should be the same one).

dubeejw:
I believe we removed the --lib support from the CLI. It will have to be added back.

rabbah:
Change the model to follow zip file uploads used in nodejs and dropped tgz support in favor of having just one format. This change the flow slightly to:


```
on host:
# cd /Users/Shared
# docker run -v /Users/Shared/:/workspace -it dockerskeleton bash

now inside container
# cd /workspace
# apk add --no-cache bzip2-dev gcc libc-dev
# echo '#include <stdio.h>
int main(int argc, char *argv[]) { printf("hello blackbox\n"); printf("%s\n", argv[1]); }' > bb.c
# gcc -o exec -static bb.c
# zip bb.tgz exec

exit container to return to host /Users/Shared
# wsk action create bb --docker bb.tgz
ok: created action bb
# wsk action invoke bb --blocking --result -p msg hello
{
    "msg": "hello"
}
```

Because the CLI fills in the docker image and it's our own, then we know we can 1. skip puling the image, and 2. the logs will be sentinelled.

```
"logs": [
        "2016-08-31T03:45:48.449739089Z stdout: hello blackbox",
]
```

rabbah:
pg1/787 pending.

rabbah:
pg2/521.

psuter:
Do we expect to document this as officially supported? If so, it would be good to have an example that does _something_ with the incoming JSON.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -93,9 +94,18 @@ protected[core] case class JavaExec(code: Attachment[String], main: String) exte
...
 protected[core] case class PythonExec(code: String) extends CodeExec[String](Exec.PYTHON)
 
-protected[core] case class BlackBoxExec(image: String) extends Exec(Exec.BLACKBOX) {
-    def size = image sizeInBytes
-    override val sentinelledLogs = false
+/**
+ * @param image the image name
+ * @param code an optional script or zip archive (as base64 encoded) string
+ */
+protected[core] case class BlackBoxExec(override val image: String, code: Option[String]) extends CodeExec[Option[String]](Exec.BLACKBOX) {
```
psuter: There is no invariant that `(image == Exec.BLACKBOX_SKELETON) != code.isDefined`, right?

This means, I guess, that users can build their own skeleton and feed it a binary? Is this a { desired, tested, documented } feature?

I realize they may have to bypass the CLI to do that.

rabbah: it is tested - i also think it is desirable.
it is not documented (neither is the intended feature yet beyond this pull request comments - I can add the bits to the actions docs.)

psuter:
No one else reviewing this? LGTM

perryibm:
I am looking at it now and getting ready to push it.  Just wrapping my head around a few tests before doing so.

perryibm:
merged

PR #1178, opened 2016-08-31
update reference on docker skeleton to use alpine image

### LINKED ISSUE #1177: update docs reference for docker skeleton to refer to alpine image
The current docs in [reference.md#docker-actions](https://github.com/openwhisk/openwhisk/blob/master/docs/reference.md#docker-actions)
contains the following

> Docker actions run a user-supplied binary in a Docker container. The binary runs in a Docker image based on Ubuntu 14.04 LTD, so the binary must be compatible with this distribution.

Now the dockerSkeleton is based on [python:2.7.12-alpine](https://hub.docker.com/r/library/python)

Do we need to reflect this change to users some how.
They can use Ubuntu (large) or Alpine (small) one

csantanapr:
Fixes #1177 

@rabbah should we mention just alpine or say that they can use any, or maybe don't mention anything about the specific base image and put more info on the protocol contract that image should follow.

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -265,6 +265,7 @@ This example invokes a Yahoo Weather service to get the current conditions at a 
 
   This example also shows the need for asynchronous actions. The action returns a Promise to indicate that the result of this action is not available yet when the function returns. Instead, the result is available in the `request` callback after the HTTP call completes, and is passed as an argument to the `resolve()` function.
 
+
 2. Run the following commands to create the action and invoke it:
   ```
   $ wsk action create weather weather.js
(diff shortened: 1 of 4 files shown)


rabbah:
Knowing the base distro is good for those who want to use it. Of course users bringing their own containers can bring whatever distro they like.

csantanapr:
I will add the following text to docker action on how to update

To update the Docker action, run buildAndPush.sh to refresh the image on Docker Hub, this will allow next time the system pulls your Docker image to run the new code for your action. If there are no warm containers any new invocations will use the new Docker image. Take into account that if there is a warm container using a previous version of your Docker Image it will continue to use this unless you run wsk action update, this will indicate to the system that for any new invocations force a docekr pull resulting on pulling your new Docker image.

csantanapr:
rebase and merged 369927b8552d6ac2948b42926401fde9fee97189

PR #1179, opened 2016-09-01
Move starting counter to where it should be.

rabbah:
Fixes two problems:

1. starting container counter incorrectly incremented when action gets a stem cell container; this should not be the case and over accounts for active containers. Moved the counter into the docker serializer where it will correctly account for the new container. The counter is inside the synchronized region rather than outside may allow more actions to try and make progress but will these will serialize on the docker lock anyway.

2. the activation counting had a race and was off by ~1 (approximately because the initialization was wrong and how many counters are created depends on the number of concurrent activations for a specific user).


rabbah:
@psuter will submit a separate pr that covers the change form the last commit in this request to keep some separations between the concerns addressed by this pr.

PR #1181, opened 2016-09-01
Throw NoSuchElementException in Consul, if 404 is returned on

cbickel:
Add NoSuchElementException to `getRecurse`.
This is also thrown on simple get.


cbickel:
PG1#359: only throttle test failed

cbickel:
PG1#361 is green

perryibm:
rebases/merged

Issue #1183, opened 2016-09-01
Go cli not working anymore on German MAC


cbickel:
@dubeejw Since the merges of the i18n of the go-cli, my deployment is not working anymore, because the go-cli exits on any commands with the following error:

```
panic: no supported languages found []string{"de_DE"}
```

## Environment details:

* local deployment (Mac)

## Steps to reproduce the issue:

1. Execute any command in go-cli or do bluewhisk deployment. 

## Provide the expected results and outputs:
The expected output is, that the command works like before the i18n.

## Provide the actual results and outputs:

```
panic: no supported languages found []string{"de_DE"}

goroutine 1 [running]:
panic(0x4fd580, 0xc820013540)
       	/usr/local/go/src/runtime/panic.go:481 +0x3e6
github.com/nicksnyder/go-i18n/i18n/bundle.(*Bundle).MustTfunc(0xc8200127d0, 0x5bb838, 0x5, 0x0, 0x0, 0x0, 0xe)
       	/src/github.com/nicksnyder/go-i18n/i18n/bundle/bundle.go:156 +0xa4
github.com/nicksnyder/go-i18n/i18n.MustTfunc(0x5bb838, 0x5, 0x0, 0x0, 0x0, 0x5)
       	/src/github.com/nicksnyder/go-i18n/i18n/i18n.go:130 +0x55
_/src/github.com/go-whisk/wski18n.InitWithLocale(0x5bb838, 0x5)
       	/src/github.com/go-whisk/wski18n/i18n.go:90 +0x9c
_/src/github.com/go-whisk/wski18n.Init(0x9515e0, 0x867008, 0x0, 0x0)
       	/src/github.com/go-whisk/wski18n/i18n.go:81 +0x59
_/src/github.com/go-whisk/wski18n.init.1()
       	/src/github.com/go-whisk/wski18n/i18n.go:57 +0x5c
_/src/github.com/go-whisk/wski18n.init()
       	/src/github.com/go-whisk/wski18n/i18n_resources.go:445 +0xf32
_/src/github.com/go-whisk/whisk.init()
       	/src/github.com/go-whisk/whisk/wskerror.go:151 +0x53
main.init()
       	/src/github.com/go-whisk-cli/main.go:100 +0x5b
```

mdeuser:
@dubeejw - As a stop gap until the CLI PII is fully globalized, in the Locale() function, just return the DEFAULT_LOCALE (in both go-whisk and go-whisk-cli).

PR #1185, opened 2016-09-01
Always use Default Locale

dubeejw:
- Use the default language en_US until externalized messages are translated
- Fixes: https://github.com/openwhisk/openwhisk/issues/1183
- PG2 80

### DIFF
--- a/tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
+++ b/tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
@@ -89,6 +89,12 @@ class WskBasicUsageTests
         }
     }
 
+    it should "show help and usage info using the default language" in {
+        val env = Map("LANG" -> "de_DE")
+        // Call will fail with exit code 2 if language not supported
+        wsk.cli(Seq("-h"), env = env)
+    }
+
     it should "show cli build version" in {
         val stdout = wsk.cli(Seq("property", "get", "--cliversion")).stdout
         stdout should include regex ("""(?i)whisk CLI version\s+201.*""")
--- a/tools/go-cli/go-whisk-cli/wski18n/i18n.go
+++ b/tools/go-cli/go-whisk-cli/wski18n/i18n.go
@@ -63,7 +63,8 @@ func CurLocale() string {
 
 func Locale(detector Detector) string {
 
-    sysLocale := normalize(detector.DetectLocale())
+    // Use default locale until strings are translated
+    /*sysLocale := normalize(detector.DetectLocale())
     if isSupported(sysLocale) {
         return sysLocale
     }
@@ -71,7 +72,7 @@ func Locale(detector Detector) string {
     locale := defaultLocaleForLang(detector.DetectLanguage())
     if locale != "" {
         return locale
-    }
+    }*/
 
     return DEFAULT_LOCALE
 }
--- a/tools/go-cli/go-whisk/wski18n/i18n.go
+++ b/tools/go-cli/go-whisk/wski18n/i18n.go
@@ -63,7 +63,8 @@ func CurLocale() string {
 
 func Locale(detector Detector) string {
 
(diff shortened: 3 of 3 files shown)


dubeejw:
@mdeuser, ready for review.

tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -73,8 +73,9 @@ class WskBasicUsageTests
         Wsk.exists(wsk.usePythonCLI)
     }
 
-    it should "show help and usage info" in {
-        val stdout = wsk.cli(Seq("-h")).stdout
+    it should "show help and usage info using the default language" in {
+        val env = Map("LANG" -> "de_DE")
```
mdeuser: When strings are fully translated and de_DE is supported, this test case fail since it explicitly looks for English strings in the output.

mdeuser:
@dubeejw - Related to externalized strings... `pt_BR.all.JSON` file name s/b `pt_BR.all.json`.

tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
dubeejw: True, we can remove any testing on this PR.

dubeejw:
Not sure what you mean here?

mdeuser:
@dubeejw - the pt_BR file extension is currently uppercase, while all other *.all.json files have `.json` in lowercase.

tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
mdeuser: Perhaps add a second test for the de_DE locale behavior validation?

dubeejw: Another test would also fail if/when translation occurs.

rabbah: That's what the test is for, no? When we have German strings, we'll have to update the test.

mdeuser: This is a modification to an original test case.  The orginal test was to confirm the existence of the help and usage.  Since the test case is hard coded with English, the test uses the default locale of English.  This test will not change when non-English translations are ready; it's an English only test.

With the additional of a non-English locale, the test case is now testing that English help and usage appears when a non-English locale is specified.  This is a short term test and will need updating when non-English translations are available.  IMHO, this test is separate from the original test intent.

dubeejw:
@mdeuser, new tests added.

dubeejw:
@mdeuser, updated tests again to make what is happening clearer.

rabbah:
deferring to @cbickel on this one.

tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -89,6 +89,12 @@ class WskBasicUsageTests
         }
     }
 
+    it should "show help and usage info using the default language" in {
+        val env = Map("LANG" -> "de_DE")
+        // Call will fail with exit code 2 if language not supported
```
mdeuser: @dubeejw - If the locale is not supported, we should log the warning and fall back to English.

dubeejw: @mdeuser, no need to log a message since the messages aren't translated.

cbickel:
merged in b0d4f07

tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
mdeuser: @dubeejw - Sorry for the confusion.  I was referring to the comment that says the CLI will `fail with an exit code of 2 if language is not supported`.  In this scenario, the CLI should muddle on in English.

PR #1191, opened 2016-09-02
Localize `isOverload` check to ActivationThrottle

markusthoemmes:
Reduces the load on consul by not unnecessarily getting system-overload information out of it. This information is already in the controller anyway.

- Some refactoring around loadbalancer health logging
- Reducing loglevel of one log to info (was warn)

### DIFF
--- a/ansible/templates/whisk.properties.j2
+++ b/ansible/templates/whisk.properties.j2
@@ -17,10 +17,12 @@ whisk.ssl.challenge=openwhisk
 defaultLimits.actions.invokes.perMinute={{ defaultLimits.actions.invokes.perMinute }}
 defaultLimits.actions.invokes.concurrent={{ defaultLimits.actions.invokes.concurrent }}
 defaultLimits.triggers.fires.perMinute={{ defaultLimits.triggers.fires.perMinute }}
+defaultLimits.actions.invokes.concurrentInSystem={{ defaultLimits.actions.invokes.concurrentInSystem }}
 
 {% if limits is defined %}
 limits.actions.invokes.perMinute={{ limits.actions.invokes.perMinute }}
 limits.actions.invokes.concurrent={{ limits.actions.invokes.concurrent }}
+limits.actions.invokes.concurrentInSystem={{ limits.actions.invokes.concurrentInSystem }}
 limits.triggers.fires.perMinute={{ limits.triggers.fires.perMinute }}
 {% endif %}
 
--- a/core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
+++ b/core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
@@ -34,14 +34,14 @@ import whisk.core.entity.Subject
  * Determines user limits and activation counts as seen by the invoker and the loadbalancer
  * in a scheduled, repeating task for other services to get the cached information to be able
  * to calculate and determine whether the namespace currently invoking a new action should
- * be allowed to do so
+ * be allowed to do so.
  *
  * @param config containing the config information needed (consulServer)
  */
-class ActivationThrottler(consulServer: String, concurrencyLimit: Int)(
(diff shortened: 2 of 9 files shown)


core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
```diff
@@ -59,7 +59,12 @@ class ActivationThrottler(consulServer: String, concurrencyLimit: Int)(
     def check(subject: Subject): Boolean = userActivationCounter.getOrElse(subject(), 0L) < concurrencyLimit
 
     /**
-     * Returns the per-namespace activation count as seen by the loadbalancer
+     * Checks whether the system is in a generally overloaded state.
+     */
+    def isOverloaded = userActivationCounter.values.sum > 5000
```
rabbah: We should move this to ansible config - since its deployment sensitive and we've consolidated the other throttle related params.

markusthoemmes: Right! Will fix.

markusthoemmes:
I rebased and adjusted to @rabbah's comment. @perry and @rabbah please rereview as needed.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -72,30 +71,31 @@ class LoadBalancerService(config: WhiskConfig, verbosity: LogLevel)(
...
-            val completedCount = completedCounts.foldLeft(0)(_ + _)
+            val completedCount = invokerCounts.values.sum
+
             val inFlight = issuedCount - completedCount
-            val overload = JsBoolean(overloadThreshold > 0 && inFlight >= overloadThreshold)
+
+            val health = invokerHealth.getInvokerHealth()
             if (count % 10 == 0) {
                 implicit val sid = TransactionId.loadbalancer
-                info(this, s"In flight: $issuedCount - $completedCount = $inFlight $overload")
```
rabbah: do we log "overloaded" anywhere?

perryibm:
Please add a logging statement showing the global limit to ease debugging.  Localizing is good.  However, we are summing over the user map for every activation so if there are a lot of users this might become a bottleneck.  Probably not worth worrying about as even 10K users would still cause it to take maybe 100K cycles (since L2 cache hit is about 10/cycle) which would be about 0.05ms.  Of course, as we scale up anything that is O(users) on every invocation will be unsustainable.

markusthoemmes:
@perryibm very valid concern. I expect we can soften this once the loadbalancer itself keeps track of the requests so we'ld have a local "overallCount" aswell? Can refactor then?

Added the log.

perryibm:
LGTM

markusthoemmes:
Rebased and PG1 5​50.

perryibm:
Waiting a few rounds while recent merges clear and will merge this shortly.

perryibm:
rebased and merged

ansible/templates/whisk.properties.j2
```diff
@@ -17,10 +17,12 @@ whisk.ssl.challenge=openwhisk
 defaultLimits.actions.invokes.perMinute={{ defaultLimits.actions.invokes.perMinute }}
 defaultLimits.actions.invokes.concurrent={{ defaultLimits.actions.invokes.concurrent }}
 defaultLimits.triggers.fires.perMinute={{ defaultLimits.triggers.fires.perMinute }}
+defaultLimits.actions.invokes.concurrentInSystem={{ defaultLimits.actions.invokes.concurrentInSystem }}
 
 {% if limits is defined %}
 limits.actions.invokes.perMinute={{ limits.actions.invokes.perMinute }}
 limits.actions.invokes.concurrent={{ limits.actions.invokes.concurrent }}
+limits.actions.invokes.concurrentInSystem={{ limits.actions.invokes.concurrentInSystem }}
```
rabbah: if limits is defined then all the values must be defined.
the omission of the `mac` environment limits breaks mac devs.

Issue #1190, opened 2016-09-01
pip install 'requests' in Python actions Dockerfile


iainhouston:
At 9:00 this morning the action was running fine - with a python action created no more than 24 hours ago at which time the python program was able to import `requests==2.11.1` in the Openwhisk runtime context.  

After 9:00am I updated the action and now the Openwhisk runtime fails to import `requests`   

`requests`  is rather crucial  to do `curl`-like things

## Environment details:

* local deployment, vagrant, native ubuntu, Mac OS, Bluemix, ...   
* version of docker, vagrant, ubuntu, ...

## Steps to reproduce the issue:

1.   Create an action written in python that uses the `requests` library
2.   Test the action locally
3.   `wsk action invoke pyRequiresRequests --blocking --result `

## Provide the expected results and outputs:

At 9:00 am
```
sitemonitor 4e213863511a40499f081b35c05e2ccd
{
    "text": "All sites OK"
}
```

## Provide the actual results and outputs:

More recently

```
sitemonitor 98325c35ed394f6fbcc0fd94303f5642
{
    "error": "The action did not return a dictionary."
}
Logs:
-01T20:40:59.620303013Z stderr: Traceback (most recent call last):
-01T20:40:59.620386383Z stderr: File "pythonrunner.py", line 50, in run
-01T20:40:59.620406898Z stderr: exec(self.fn, namespace)
-01T20:40:59.620609563Z stderr: File "action", line 4, in <module>
-01T20:40:59.620803581Z stderr: ImportError: No module named requests
1 September 2016 21:40:59 BST
Duration: 252ms
```

## Additional information you deem important:
* issue happens recently  
* I didn't notice the `pip install requests` in the Dockerfile previously ... but *something* must have changed to make the import to be failing now

markusthoemmes:
I believe this is due to recent changes to the python action container and a promotion of a new OpenWhisk version to Bluemix.

rabbah:
Fixed by above commit.

iainduncani:
@markusthoemmes I presume that question was meant for @iainhouston, I've never used python in Whisk!

PR #1192, opened 2016-09-02
Adding `requests` to the python container

markusthoemmes:
- Add the requests module to the python container
- Add tests to assert that module installation for the python container works as expected.

Solves #1190

### DIFF
--- a/core/pythonAction/Dockerfile
+++ b/core/pythonAction/Dockerfile
@@ -3,6 +3,9 @@ FROM dockerskeleton
 
 RUN apk add --no-cache python-dev
 
+# Install common modules for python
+RUN pip install requests
+
 ENV FLASK_PROXY_PORT 8080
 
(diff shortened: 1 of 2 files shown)


rabbah:
Any other packages we lost/should include?

psuter:
Here's my personal list:

    pip install beautifulsoup
    pip install httplib2
    pip install lxml
    pip install python-dateutil
    pip install scrapy
    pip install simplejson
    pip install twisted

rabbah:
@markusthoemmes added a simple test for `request`. I suggest we augment the test to require the installed packages but do not necessarily need to test each one of that.

rabbah:
Added @psuter's list with some more tests here: https://github.com/openwhisk/openwhisk/commit/d2df2c2466786a58f1b9170f8cc2074f865ba7e3.

PR #1194, opened 2016-09-02
Add fullyqualified action names and limits to activation (Only for actions, not for triggers and rules)

cbickel:
Add view to return all activations that have been triggered  by actions


cbickel:
PG2#75 is running

cbickel:
closed for https://github.com/openwhisk/openwhisk/pull/1233

PR #1197, opened 2016-09-02
Removed pool reference from [Whisk]Containers

psuter:
The main reason for keeping a reference to the pool from within the container was access to logs, which require pool-specific knowledge.  However every client accessing logs has a reference to the pool anyway and can use it.


psuter:
PG2 81

PR #1199, opened 2016-09-02
Further explain request/response activations.

### LINKED ISSUE #1140: Convert all blocking activation requests to non-blocking after 60 seconds

rabbah:
Related to #1156.
Closes #1140.


csantanapr:
rebase and merged b3edff88552a6796ab2b05dcb827c6e59bb04f14

PR #1200, opened 2016-09-03
Test retries.

rabbah:
Wrap API tests on list operations with retry for eventual consistency.


mdeuser:
LGTM.  I noticed that there are some `list` related package tests that don't have the `whisk.utils.retry` decoration.. For example `list all public packages in explicit namespace excluding bindings`.  Is retry not needed for that one?

rabbah:
Will add it - missed them.

rabbah:
Now covered the packages tests as well.

rabbah:
Haven't seen this failures in some time. Going to close.

PR #1202, opened 2016-09-03
additional python packages in base image for python actions

rabbah:
adds python packages suggested by @psuter with additional tests.

pg2-84 approved.
replaces #1192

### DIFF
--- a/core/actionProxy/Dockerfile
+++ b/core/actionProxy/Dockerfile
@@ -7,7 +7,7 @@ RUN apk add --no-cache bash \
         bzip2-dev \
         gcc \
         libc-dev \
-  && pip install --no-cache-dir gevent flask \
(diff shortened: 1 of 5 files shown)


markusthoemmes:
Do we really need to test all the packages we include? My initial tests idea was to verify that the installation with `pip` works as intended. I'ld think that's sufficient. We don't do it for node.js either.

csantanapr:
Let's ping a specific version for this packages and also need to be documented for users which version of Python it's installed along side which packages with respective versions

rabbah:
It was primarily to make sure we don't accidentally remove a dependence after pip install.

psuter:
Merged as 313c67174980df4e19af65c38388c69dfad69d11

Issue #1165, opened 2016-08-30
Remove the catalog and catalog tests


houshengbo:
After the dependency on installed catalog is removed, we should remove the catalog folder and related test cases. This issue should be done after https://github.com/openwhisk/openwhisk/issues/1164 is fixed.

PR #1208, opened 2016-09-06
Remove the catalog

houshengbo:
Closes-Bug: #1165


jasonpet:
You cannot remove util.sh.  This is currently needed by other repositories.

catalog/util.sh
```diff
@@ -1,74 +0,0 @@
-#!/bin/bash
```
jasonpet: Removing util.sh will break other repositories during their installs.

jasonpet:
I do not want multiple copies of util.sh in each new repository.  We already ran into an issue where the util.sh from openwhisk repo and the new catalog repo were not in sync.  We should only have one and reference that from the other repositories.

jasonpet:
@houshengbo - we need to do one last check to make sure everything is in sync between the catalog repository and all of these files we are deleting.

houshengbo:
@jasonpet This one is ready as well.

jasonpet:
pg1#419

houshengbo:
@jasonpet The catalog repo has been synchronized with openwhisk catalog, so the catalog under openwhisk is able to remove.

PR #1105, opened 2016-08-22
Add notice about explicit passing in action sequence

hassenius:
Add some clarification of the unix pipe like behaviour of action sequence.

PR #1209, opened 2016-09-06
make clear how parameters in sequences get pass

csantanapr:
supersede #1105

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -333,9 +333,13 @@ Several utility actions are provided in a package called `/whisk.system/util` th
   }
   ```
 
-  In the result, you see that the lines are sorted.
+  In the result, you see that the lines are sorted, this is the result of the last action in the sequence.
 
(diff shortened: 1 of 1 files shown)


docs/actions.md
```diff
@@ -333,9 +333,13 @@ Several utility actions are provided in a package called `/whisk.system/util` th
   }
   ```
 
-  In the result, you see that the lines are sorted.
+  In the result, you see that the lines are sorted, this is the result of the last action in the sequence.
 
-**Note**: For more information about invoking action sequences with multiple named parameters, see [Setting default parameters](./actions.md#setting-default-parameters)
+  The first action in the sequence receives the parameters of the invocation, the JSON object return from the first action becomes the input
```
rabbah: JSON object return -> JSON object return_ed_

docs/actions.md
```diff
@@ -333,9 +333,13 @@ Several utility actions are provided in a package called `/whisk.system/util` th
...
+  parameters to the second action, if you already defined default parameters for the second action those will also be passed, take into account
```
rabbah: parameters -> parameter

PR #1219, opened 2016-09-07
Fix the typo of fullActioName

houshengbo:
This patch changes the wrong fullActioName into fullQualifiedName.

### DIFF
--- a/tests/src/system/basic/WskActionTests.scala
+++ b/tests/src/system/basic/WskActionTests.scala
@@ -241,24 +241,9 @@ class WskActionTests
 
     it should "reject an invoke with the wrong parameters set" in withAssetCleaner(wskprops) {
         (wp, assetHelper) =>
-            val name = "ping"
-            val samplePackage = "samples"
-            val sampleAction = "helloWorld"
-            val fullActioName = s"/$guestNamespace/$samplePackage/$sampleAction"
-            assetHelper.withCleaner(wsk.action, name) {
-                (action, _) => action.create(name, Some(TestUtils.getTestActionFilename("ping.js")))
-            }
-            assetHelper.withCleaner(wsk.pkg, samplePackage) {
-                (pkg, _) => pkg.create(samplePackage, shared = Some(true))(wp)
-            }
-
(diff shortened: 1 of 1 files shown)


tests/src/system/basic/WskActionTests.scala
```diff
@@ -244,20 +244,20 @@ class WskActionTests
             val name = "ping"
             val samplePackage = "samples"
             val sampleAction = "helloWorld"
-            val fullActioName = s"/$guestNamespace/$samplePackage/$sampleAction"
+            val fullQualifiedName = s"/$guestNamespace/$samplePackage/$sampleAction"
```
rabbah: I think I noted this in another comment - if not, here goes again:
I don't think we need to create any assets here.
The test should fail with a ERROR exit code (1) so if the CLI permits the operation, it would otherwise fail with 148.

tests/src/system/basic/WskActionTests.scala
```diff
@@ -241,23 +241,9 @@ class WskActionTests
...
-            }
-
-            assetHelper.withCleaner(wsk.action, fullActioName) {
-                val file = Some(TestUtils.getTestActionFilename("empty.js"))
-                (action, _) => action.create(fullActioName, file, shared = Some(true))(wp)
-            }
+            val fullQualifiedName = s"/$guestNamespace/samples/helloWorld"
             val payload = "bob"
-            val rr = wsk.cli(Seq("action", "invoke", fullActioName, payload), expectedExitCode = TestUtils.ANY_ERROR_EXIT)
+            val rr = wsk.cli(Seq("action", "invoke", fullQualifiedName, payload), expectedExitCode = TestUtils.ANY_ERROR_EXIT)
```
rabbah: since `wsk.cli` validates the exit code, you could use `TestUtils.ERROR_EXIT` instead and omit line 247.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/63d483a9acf9ee0fe416475a5254cd36672fda7c.

PR #1233, opened 2016-09-09
Add fully-qualified action name and action limits to annotations of activations

cbickel:
In addition, a cloudant view is provided, that returns only action-activations.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
@@ -83,7 +83,6 @@ import whisk.core.entity.BlackBoxExec
 import whisk.core.entity.DocId
 import whisk.core.entity.DocInfo
 import whisk.core.entity.DocRevision
-import whisk.core.entity.EntityName
 import whisk.core.entity.EntityPath
 import whisk.core.entity.LogLimit
 import whisk.core.entity.SemVer
@@ -98,6 +97,9 @@ import whisk.core.entity.size.SizeLong
 import whisk.core.entity.size.SizeString
 import whisk.http.BasicHttpService
 import whisk.utils.ExecutionContextFactory
+import whisk.common.Logging
+import whisk.core.entity.Parameters
+import whisk.core.entity.ActionLimits
 
 /**
  * A kafka message handler that invokes actions as directed by message on topic "/actions/invoke".
@@ -186,11 +188,11 @@ class Invoker(
         implicit transid: TransactionId): Unit = {
         error(this, errorMsg)
         val msg = tran.msg
-        val name = EntityName(actionDocInfo.id().split(EntityPath.PATHSEP)(1))
+        val name = EntityPath(actionDocInfo.id())
         val version = SemVer() // TODO: this is wrong, when the semver is passed from controller, fix this
         val response = ActivationResponse.whiskError(errorMsg)
         val interval = computeActivationInterval(tran)
-        val activation = makeWhiskActivation(msg, name, version, response, interval)
+        val activation = makeWhiskActivation(msg, name, version, response, interval, None)
         completeTransaction(tran, activation, FailedActivation(transid))
     }
 
@@ -253,7 +255,7 @@ class Invoker(
 
                     val activationInterval = computeActivationInterval(tran)
                     val activationResponse = getActivationResponse(activationInterval, action.limits.timeout.duration, response, failedInit)
-                    val activationResult = makeWhiskActivation(msg, action.name, action.version, activationResponse, activationInterval)
+                    val activationResult = makeWhiskActivation(msg, EntityPath(action.docid.id), action.version, activationResponse, activationInterval, Some(action.limits))
                     val completeMsg = CompletionMessage(transid, activationResult)
 
                     producer.send("completed", completeMsg) map { status =>
@@ -278,7 +280,7 @@ class Invoker(
                     ActivationResponse.whiskError("error starting container to run action")
                 }
                 val interval = computeActivationInterval(tran)
-                val activation = makeWhiskActivation(msg, action.name, action.version, response, interval)
+                val activation = makeWhiskActivation(msg, EntityPath(action.docid.id), action.version, response, interval, Some(action.limits))
                 completeTransaction(tran, activation, FailedActivation(transid))
             }
         }
@@ -432,13 +434,14 @@ class Invoker(
      */
     private def makeWhiskActivation(
         msg: Message,
-        actionName: EntityName,
+        actionName: EntityPath,
         actionVersion: SemVer,
         activationResponse: ActivationResponse,
-        interval: Interval) = {
+        interval: Interval,
+        limits: Option[ActionLimits]) = {
         WhiskActivation(
             namespace = msg.activationNamespace,
-            name = actionName,
+            name = actionName.last,
             version = actionVersion,
             publish = false,
             subject = msg.subject,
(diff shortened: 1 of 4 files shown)


common/scala/src/main/scala/whisk/core/entity/Limits.scala
```diff
@@ -62,7 +62,7 @@ protected[core] object ActionLimits
     /** Creates a ActionLimits instance with default duration, memory and log limits. */
     protected[core] def apply(): ActionLimits = ActionLimits(TimeLimit(), MemoryLimit(), LogLimit())
 
-    override protected[core] implicit val serdes = new RootJsonFormat[ActionLimits] {
+    override implicit val serdes = new RootJsonFormat[ActionLimits] {
```
rabbah: do we need the relaxed protection?

common/scala/src/main/scala/whisk/core/entity/Parameter.scala
```diff
@@ -87,7 +87,7 @@ protected[core] class Parameters protected[entity] (
     /**
      * Retrieves parameter by name if it exists.
      */
-    protected[core] def apply(p: String) = Try { params(new ParameterName(p)) } map { _() } toOption
+    def apply(p: String) = Try { params(new ParameterName(p)) } map { _() } toOption
```
rabbah: same.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -221,11 +223,11 @@ class Invoker(
         implicit transid: TransactionId): Unit = {
         error(this, errorMsg)
         val msg = tran.msg
-        val name = EntityName(actionDocInfo.id().split(EntityPath.PATHSEP)(1))
+        val name = EntityPath(actionDocInfo.id.id)
```
rabbah: I thought we agreed to leave this as is. Instead, fix the namespace value when creating the activation to not be `subject`.

cbickel: Yes, we agreed, that we leave the name in the WhiskActivation as EntityName.
But I changed the type, that is passed to `makeWhiskActivation()` from EntityName to EntityPath, to pass only one variable around and not the name and the fqdn.
At the place, where the name is written to the activation, this name is still EntityName.
I only add the fqdn to the annotations.

cbickel:
PG1#404 is running

rabbah:
we'll need a test for the this - the go CLI appears to ignore the annotations on activations.

@dubeejw

lehoanganh:
PG1#426

lehoanganh:
@rabbah PG426 was good. The failed test case was not related with this code base. Issue 2533 covers it.

lehoanganh:
or PG445 also approved.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -462,13 +463,14 @@ class Invoker(
...
+        actionName: EntityPath,
         actionVersion: SemVer,
         activationResponse: ActivationResponse,
-        interval: Interval) = {
+        interval: Interval,
+        limits: Option[ActionLimits]) = {
         WhiskActivation(
             namespace = msg.subject.namespace,
-            name = actionName,
+            name = actionName.last,
```
rabbah: we agree that in the future, name will be the fully qualified name but this is an invasive change to make now and deferring it; as such an annotation is added to contain the fully qualified name as an intermediate point.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -477,7 +479,8 @@ class Invoker(
             start = interval.start,
             end = interval.end,
             response = activationResponse,
-            logs = ActivationLogs())
+            logs = ActivationLogs(),
+            annotations = Parameters("limits", limits.toJson) ++ Parameters("fqdn", actionName.toJson))
```
rabbah: fqdn -> path?

tools/db/loadTransientDBViews.sh
```diff
@@ -71,6 +71,9 @@ function view() {
...
+        },
+        "actionActivations": {
+          "map": "function (doc) {\n  var PATHSEP = \"/\";\n\n  var isPackage = function (doc) {  return (doc.binding !== undefined) };\n  var isAction = function (doc) { return (doc.exec !== undefined) };\n  var isTrigger = function (doc) { return (doc.exec === undefined && doc.binding === undefined && doc.parameters !== undefined) };\n  var isRule = function (doc) {  return (doc.trigger !== undefined) };\n  var isActivation = function (doc) { return (doc.activationId !== undefined) };\n\n  if (isActivation(doc) && doc.end && doc.start && doc.start < doc.end) try {\n    emit([doc.end], doc.name);\n  } catch (e) {}\n}"
```
rabbah: shouldn't this emit the "fqdn"?

lehoanganh: no, the "fqdn" should still be there in the annotations so that in the next step "other process" can find the fqdn of the action which caused this activation.

 Actually, the view does likely the same thing as "activations" view. But one can inout the time period by executing this view.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
lehoanganh: ?

rabbah: suggesting that you rename `fqdn` to something more clear.

lehoanganh: ok dokie

cbickel:
Tests for view and for the annotations is written.
New PG2#196 is running.

dubeejw:
The CLI updates look good.

cbickel:
The cloudant view is removed and PG2#203 is running.

tests/src/system/basic/WskActionTests.scala
```diff
@@ -264,6 +267,27 @@ class WskActionTests
...
+
+            val run = wsk.action.invoke(name, Map("payload" -> testString.toJson))
+            withActivation(wsk.activation, run) {
+                activation =>
+                    activation.response.status shouldBe "success"
+                    val annotations = activation.annotations.get.mkString(", ")
+                    annotations should include("\"limits\"")
+                    annotations should include(memoryLimit.size.toString)
+                    annotations should include("\"path\"")
+                    annotations should include(name)
```
markusthoemmes: Can these checks be tightened? I.e. `annotations should contain(JsObject("key" -> "limits", "value" -> "something"))` (won't compile but you get the idea)

tools/cli/go-whisk/whisk/activation.go
```diff
@@ -41,6 +42,7 @@ type Activation struct {
     End          int64  `json:"end"`                    // Since a 0 is a valid value from server, don't omit
     Response     `json:"response,omitempty"`
     Logs         []string `json:"logs,omitempty"`
+    Annotations *json.RawMessage    `json:"annotations,omitempty"`
```
markusthoemmes: nit: fix the whitespacing? just remove it down to 1 here or adjust every line to match.

tests/src/system/basic/WskActionTests.scala
rabbah: `limits` should be `ActionLimits().toJson`.

rabbah:
I moved the test and rewrote the check. Merged as https://github.com/openwhisk/openwhisk/commit/a9ebadfa454434e4998a39dfb7a5b82159cac1f1.

PR #1234, opened 2016-09-09
adding support for openstack persistent volumes

kkbankol-ibm:
These proposed changes enhance the distributed ansible playbooks with the ability to create and attach a persistent disk to the couchdb instance. The size of the couchdb disk (in GB) can be set either as an environment variable (OS_WSK_DB_VOLUME=15) or in the instances object from environments/distributed/group_vars/all.

### DIFF
--- a/ansible/environments/distributed/group_vars/all
+++ b/ansible/environments/distributed/group_vars/all
@@ -27,11 +27,13 @@ registry:
     threshold: 80
     keeptags: 5
 
+# openstack vars
+# TODO, should move these to a role and create a designated "defaults.yml" file
 rc_file:
 #net_name: "{{ lookup('env','OS_NET_NAME') }}"
 net_id: "{{ lookup('env','OS_NET_ID') }}"
 image: "{{ lookup('env','OS_IMAGE') }}"
-flavor: "{{ lookup('env','OS_FLAVOR') }}"
+default_flavor: "{{ lookup('env','OS_FLAVOR') | default('m1.small') }}"
 auth:
   auth_url: "{{ lookup('env','OS_AUTH_URL') }}"
   username: "{{ lookup('env','OS_USERNAME') }}"
@@ -40,6 +42,8 @@ auth:
 key_name: "{{ lookup('env','OS_KEY_NAME') }}"
 security_groups: "{{ lookup('env','OS_SECURITY_GROUPS') }}"
 
+# list of instances.
+# TODO, move each object to designated group_vars file
 instances:
   - name: registry
     num_instances: 1
@@ -69,3 +73,7 @@ instances:
     num_instances: 1
     flavor:
     volume:
+      name:
+      size:
+      fstype:
+      fsmount:
--- a/ansible/environments/distributed/group_vars/all
+++ b/ansible/environments/distributed/group_vars/all
@@ -68,3 +68,4 @@ instances:
   - name: db
     num_instances: 1
     flavor:
+    volume:
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -72,7 +72,7 @@ docker:
   restart:
     policy: always
 
-cli_path: "{{ openwhisk_home }}/bin/wsk"
+cli_path: "{{ openwhisk_home }}/bin/go-cli/wsk"
 
 # The default name space is /whisk.system. The catalog namespace must begin with a slash "/".
 catalog_namespace: "/whisk.system"
--- a/ansible/provision_env_dist.yml
+++ b/ansible/provision_env_dist.yml
@@ -6,11 +6,11 @@
     #- name: Load rc file if provided
     #  shell: . rc_file
 
-    # TODO, not sure if this is necessary
+    # TODO, ensure this is necessary
     - include_vars: environments/distributed/group_vars/all
 
     - name: add header to hosts file
-      copy: 
+      copy:
         content: "; the first parameter in a host is the inventory_hostname which has to be\n; either an ip\n; or a resolvable hostname\n\n; used for local actions only\nansible ansible_connection=local\n"
         dest: "{{inventory_dir}}/hosts"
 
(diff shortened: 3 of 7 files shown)


ansible/roles/couchdb/tasks/deploy.yml
```diff
@@ -5,6 +5,21 @@
...
+  shell: df -h
+  register: disk_status
+
+- name: (re)start CouchDB
+  docker:
+    name: couchdb
+    image: couchdb
+    state: reloaded
+    restart_policy: "{{ docker.restart.policy }}"
+    volumes: "/var/couchdb:/usr/local/var/lib/couchdb"
```
markusthoemmes: As I read this, only this line is different from the "normal" "(re)start CouchDB". Can this be solved in a more DRY way?

kkbankol-ibm: hey @markusthoemmes , i've just added and pushed a fix for that, please let me know if there's anything else I need to update

ansible/environments/distributed/group_vars/all
```diff
@@ -70,3 +70,4 @@ instances:
   - name: db
     num_instances: 1
     flavor:
+    volume:
```
domdom82: See comments regarding disk sizes below. The same works for instance flavors here.

Instead of describing specific host groups in the `group_vars/all` file which is applied to all hosts, you could provide individual group_vars files for each host group which would make this list obsolete.

e.g.

group_vars/edge
```
flavor: small
```

group_vars/controllers
```
flavor: medium
```

etc.

the `num_instances` property is redundant because the number of hosts is already defined in the `hosts` inventory file per each group.

for provisioning you would have a playbook that simply targets all hosts and uses their flavor to provision the right size:

```
- hosts: all
  tasks:
    - create volume
    - provision host
    - attach volume
    ....
```

each task could simply access variables as defined in the proper group_vars/xxx file without the need to `include_vars` and weird `set_facts` for the instance name. The instance name is already defined in your `hosts` inventory file. It is `inventory_hostname`. Why define it a second time?

ansible/provision_env_dist.yml
```diff
@@ -6,22 +6,53 @@
...
       with_items: instances
 
-    - include_vars: environments/distributed/group_vars/all
-
     - meta: refresh_inventory
 
+    - name: refresh vars
+      include_vars: environments/distributed/group_vars/all
+
+# TODO, add support to loop through instances, configure all vms that have cinder disks (kafka, elk, etc). Hardcoding to db for now
```
animeshsingh: Like the idea of redesign - we can revisit it later in a future iteration

ansible/environments/distributed/group_vars/all
animeshsingh: Should fall in redesign category above - to be revisited

ansible/provision_env_dist.yml
```diff
@@ -6,22 +6,53 @@
...
+        db_instance: "{{instances | selectattr('name', 'equalto', 'db') | list | first}}"
+    - name: Set boolean flag to determine if volume is to be configured or not. Skip if volume value is not set in both env and group_vars, or if it has already been configured
+      set_fact:
+        configure_volume: "{{((db_volume_env is defined and db_volume_env != '') or (db_instance.volume is defined and db_instance.volume != '')) and ('/var/couchdb' not in disks.stdout) and (volume is defined)}}"
+    - name: Create mount folder
+      shell: mkdir -p /var/couchdb
+      become: true
+      when: configure_volume
+    - name: Partition persistent disk
+      shell: mkfs.ext3 {{volume}} 2>/dev/null
```
animeshsingh: @kkbankol-ibm please make filesystem time configurable

ansible/provision_env_dist.yml
```diff
@@ -6,22 +6,53 @@
...
+      become: true
+      when: configure_volume
+    - name: Mount persistent disk
+      shell: mount -t ext3 {{volume}} /var/couchdb
```
animeshsingh: Agreed

ansible/environments/distributed/group_vars/all
kkbankol-ibm: When the user first runs the provision_env_dist playbook, the Openwhisk VMs do not exist yet in the targeted IAAS, therefore they will not have a valid hosts file yet. The provision playbook iterates through the instances list, makes Openstack api calls for each instance object, and generates the hosts file dynamically based on the information returned by the api results. The num_instances and name properties are required here to define the group names that'll be placed in the hosts file, as well as the number of instances that will be in each group. This provision playbook is optional if the user wants to manage the instances and hosts file themselves.

I've removed the include_vars calls, looks like those weren't needed.
I agree, there is a need for redesign so we can get rid of those tricky set_facts calls, can probably start on that later this week. 
I've made adjustments based on your other suggestions.

domdom82: @kkbankol-ibm agreed this is a tricky part. off the top of my head I see two options to solve this:
1. we could work with a template of the hosts file that only describes the groups and how many servers there are in each group. the playbook would read the template, provision the VMs and then generate the actual hosts file from the template and fill in the IPs. I mean the playbook at the moment pretty much does exactly that: it fills the header, then adds the hosts, then puts the registry entry at the bottom - it writes a template.
For reference you could check out the hosts template we use for the mac env. It finds out the IP of the docker-machine VM and fills it in.

2. A separate approach could be to not use a statically generated template at all. I haven't tried it yet but I know if you put in a shell or python script in the environment dir, it will be called to produce a JSON array of hosts. there are plenty of pre-canned inventory scripts already there - OpenStack included.
Check this out http://docs.ansible.com/ansible/intro_dynamic_inventory.html#example-openstack-external-inventory-script

Other than that, I did approve so you are good to merge. Just open a secondary issue to tackle the design changes and we're good.

ansible/templates/hosts
```diff
@@ -0,0 +1,13 @@
+; the first parameter in a host is the inventory_hostname which has to be
```
domdom82: if this is meant for distributed it should go into the distributed env. 
also it should end in .j2 as it is jinja2 template script that eventually becomes a hosts file.

rabbah:
I will verify and merge this.

ansible/group_vars/all
```diff
@@ -73,7 +73,7 @@ docker:
   restart:
     policy: always
 
-cli_path: "{{ openwhisk_home }}/bin/wsk"
+cli_path: "{{ openwhisk_home }}/bin/go-cli/wsk"
```
rabbah: outdated - will fix.

ansible/templates/hosts
rabbah: @domdom82 what about this comment?

rabbah:
The change to the `couchdb` deploy task is occasionally failing on a mac with docker-machine:

```
fatal: [192.168.99.100]: FAILED! => {"changed": true, "cmd": "df -h", "delta": "0:00:00.003609", "end": "2016-09-22 00:00:57.963466", "failed": true, "rc": 1, "start": "2016-09-22 00:00:57.959857", "stderr": "df: /mnt/sda1/var/lib/docker/aufs/mnt/2f35659bdca922697e809d67d1007421d17bf0b938081217739271ae57c0fd60: Permission denied\ndf: /mnt/sda1/var/lib/docker/containers/5a46ae627b8d586c2a86955d3cb1cf3aa97155fa024fae26954b89e5298c5d6b/shm: Permission denied\ndf: /var/run/docker/netns/50a4a5d6d53e: Permission denied" ...
```

ansible/templates/hosts
kkbankol-ibm: fixed, template has been moved to ansible/environments/distributed/templates/hosts.j2

kkbankol-ibm:
@rabbah , hmm, strange. haven't seen that permissions error yet, perhaps docker-machine doesn't allow the user to run that df command. my last push added a conditional to only run that df -h command if a volume has been created/mounted and referenced in the hosts file

animeshsingh:
@rabbah good to go?

rabbah:
Last patch resolved restart of couchdb.

Merged as https://github.com/openwhisk/openwhisk/commit/bd12935a71eabce39ca39fcf60e0006836bd2f63.

ansible/templates/hosts
domdom82: @rabbah I wanted to comment on the whole file. is this possible? didn't find a button for it so I took the first line.

rabbah: the file was moved as you suggested.

PR #1235, opened 2016-09-09
Delete hourly limits and add override mechanism for deployment specfic limits

perryibm:
While ansible does not have an override mechanism, this can be accomplished by grouping keys together in WhiskConfig.  PG1-408

### DIFF
--- a/common/scala/src/main/scala/whisk/common/Config.scala
+++ b/common/scala/src/main/scala/whisk/common/Config.scala
@@ -16,6 +16,8 @@
 
 package whisk.common
 
+import scala.util.Try
+
 /**
  * A set of properties which define a configuration.
  *
@@ -39,21 +41,19 @@ package whisk.common
  *
  * @param propertiesFile a file which defines the properties
  */
-class Config(requiredProperties: Map[String, String]) {
+class Config(requiredProperties: Map[String, String], optionalProperties: Set[String] = Set()) extends Logging {
 
     def isValid: Boolean = valid
 
     /**
      * Gets value for key if it exists else the empty string.
+     * The value of the override key will instead be returned if its value is present in the map.
      *
      * @param key to lookup
+     * @param overrideKey the property whose value will be returned if the map contains the override key.
      * @return value for the key or the empty string if the key does not have a value/does not exist
      */
-    def apply(key: String): String = try {
-        settings(key)
-    } catch {
-        case _: Throwable => ""
-    }
+    def apply(key: String, overrideKey: String = ""): String = Try(settings(overrideKey)).orElse(Try(settings(key))).getOrElse("")
 
     /**
      * Returns the value of a given key.
--- a/core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
+++ b/core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
@@ -41,6 +41,8 @@ import whisk.core.entity.Subject
 class ActivationThrottler(consulServer: String, concurrencyLimit: Int)(
     implicit val system: ActorSystem) extends Logging {
 
+    info(this, s"concurrencyLimit = $concurrencyLimit")
+
     implicit private val executionContext = system.dispatcher
 
     /**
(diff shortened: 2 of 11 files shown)


core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
```diff
@@ -41,6 +41,8 @@ import whisk.core.entity.Subject
 class ActivationThrottler(consulServer: String, concurrencyLimit: Int)(
     implicit val system: ActorSystem) extends Logging {
 
+    info(this, s"ActivationThrottler.concurrencyLimit = $concurrencyLimit")
```
rabbah: `this` will print `ActivationThrottler`.

core/controller/src/main/scala/whisk/core/entitlement/RateThrottler.scala
```diff
@@ -28,8 +28,9 @@ import akka.event.Logging.InfoLevel
  *
  * For now, we throttle only at a 1-minute granularity.
  */
-class RateThrottler(maxPerMinute: Int,
-                    maxPerHour: Int) extends Logging {
+class RateThrottler(maxPerMinute: Int) extends Logging {
+
+    info(this, s"RateThrottler.maxPerMinute = $maxPerMinute")
```
rabbah: `RateThrottler` not necessary. Will change when merging.

perryibm:
PG1-483 (and 484) show it working.  Note that the throttle test case for concurrent checking now performs exactly N+1 invokes and is thus "tight".

tests/src/limits/ThrottleTests.scala
```diff
@@ -55,13 +55,27 @@ class ThrottleTests
...
     val concurrencyMessage = "The user has sent too many requests in a given amount of time."
 
+    /*
+     * Retrieve a numeric limit for the key from the property set.  If the overrideKey is present, use that.
+     */
+    def getLimit(key: String, overrideKey: String) = try {
+        WhiskProperties.getProperty(overrideKey).toInt
+    } catch {
+        case _: Throwable => WhiskProperties.getProperty(key).toInt
+    }
```
markusthoemmes: I think, using the "functional" way, a `Try` makes this more readable (and idiomatic).

```
def getLimit(key: String, overrideKey: String) = Try {
    WhiskProperties.getProperty(overrideKey).toInt
} getOrElse {
    WhiskProperties.getProperty(key).toInt
}
```

markusthoemmes:
In general: Has it been verified that this override approach is necessary and cannot completely be done in ansible? @lehoanganh @domdom82 ?

perryibm:
Yes, I have talked to both of them about this.

tests/src/limits/ThrottleTests.scala
perryibm: Did you mean `...orElse....getOrElse`  since `value` may also fail?

perryibm:
retesting one last time at PG1-487

perryibm:
Last travis build got stuck building swift image. Re-trying.

markusthoemmes:
In a quick chat with @domdom82 and @lehoanganh they stated that this would be possible by removing the global "default" and use a specific limit for each deployment. Would it make sense to do that instead?

rabbah:
We considered this. Since the limits are overridden to a lower value in only a very small number of deployments (1 currently), it makes more sense to do it this way rather than modify `N` environment files which all will have the same value.

common/scala/src/main/scala/whisk/common/Config.scala
```diff
@@ -39,20 +39,26 @@ package whisk.common
...
      * Gets value for key if it exists else the empty string.
+     * The value of the override key will instead be returned if its value is present in the map.
      *
      * @param key to lookup
+     * @param overrideKey the property whose value will be returned if the map contains the override key.
      * @return value for the key or the empty string if the key does not have a value/does not exist
      */
-    def apply(key: String): String = try {
-        settings(key)
+    def apply(key: String, overrideKey: String = ""): String = try {
```
markusthoemmes: Is this the same as

`Try(settings(overrideKey)).orElse(Try(settings(key)).getOrElse("")` ?

tests/src/limits/ThrottleTests.scala
perryibm: yep - done - i misread it the first time thinking it was for some other file

perryibm:
Travis now passing the final throttle test.  Redoing PG just in case - PG1-501.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/dbc0e8f997fe67d3624ce9d02de287fd76d53d96.

Issue #1015, opened 2016-08-05
Update CLI to reflect that rules are now enable/active on creation


csantanapr:
The backend will be change so rules on creation will be active/enable by default #1008 
The CLI will need to be updated to change this.

This change is based on user feedback, that they always expect rules to be active/enable on creation time.

The proposal is to remove the flag `--enable` from the `wsk rule create` command.
Not introduce a a new flag (i.e. ``--not-enable`) as this is confusing, and not a feature requested by users.
Is user wants to disable a rule when creating they can run the disable command `wsk rule disable`
For example 
```
wsk rule create rule1 trigger1 action1
wsk rule disable rule1
```
To re-enable the rule
```
wsk rule enable rule1
```

This depends on #1008

csantanapr:
If after a while there is feedback on requesting some type of flag `--not-enable` then this can be added to the CLI, but for now I prefer to not implement until we hear feedback from real users using the new behavior.

I know choosing default values are a pain to get right, but waiting some time and  feedback/analytics is a good way to land in correct place.

dubeejw:
@csantanapr, will we also need a command to enable a rule after it has been disabled? Ex:

```wsk rule enable rule1```

csantanapr:
@dubeejw enable and disable command already exist today

rabbah:
To summarize: we agreed to remove `--enable` from the CLI for `rules` and do nothing else.

dubeejw:
@rabbah, I don't believe ```rule update``` currently has the ```--enable``` flag.

PR #1236, opened 2016-09-09
Remove Enable Flag from Rule Create

dubeejw:
- Remove the --enable flag from the "rule create" command as rules are enabled by default now

Fixes https://github.com/openwhisk/openwhisk/issues/1015
PG2 111


csantanapr:
merged d2d8cfc15bb0da7cfead58d588e8968d5a3579d1

PR #1239, opened 2016-09-11
Fixes ContainerName in the Invoker

markusthoemmes:
Logs extracted from the Invoker are currently named after JVM object addresses. This restores old behaviour.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/container/Container.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/Container.scala
@@ -49,7 +49,7 @@ class Container(
     implicit var transid = originalId
 
     val id = Container.idCounter.next()
-    val name = containerName.getOrElse("anon")
+    val name = containerName.getOrElse(ContainerName.fromString("anon"))
 
     val (containerId, containerHostAndPort) = bringup(containerName, image, network, cpuShare, env, args, limits, policy)
 
--- a/core/invoker/src/main/scala/whisk/core/container/package.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/package.scala
@@ -83,7 +83,9 @@ package object container {
         override def toString() = s"$host:$port"
     }
 
-    sealed abstract class ContainerIdentifier(val id: String)
+    sealed abstract class ContainerIdentifier(val id: String) {
+        override def toString() = id
+    }
     class ContainerName(val name: String) extends ContainerIdentifier(name)
     class ContainerHash(val hash: String) extends ContainerIdentifier(hash)
 


core/invoker/src/main/scala/whisk/core/container/Container.scala
```diff
@@ -49,7 +49,7 @@ class Container(
     implicit var transid = originalId
 
     val id = Container.idCounter.next()
-    val name = containerName.getOrElse("anon")
+    val name = containerName.getOrElse(ContainerName.fromString("anon"))
```
psuter: Good catch, but I think the fix should be `containerName.map(_.toString).getOrElse("anon")`. We shouldn't create an instance of the typed `ContainerName` from the gargabe value `anon`.

core/invoker/src/main/scala/whisk/core/container/package.scala
```diff
@@ -83,7 +83,9 @@ package object container {
         override def toString() = s"$host:$port"
     }
 
-    sealed abstract class ContainerIdentifier(val id: String)
```
psuter: I would prefer that we avoided that (to make sure we never use `String`s when something else makes more semantic sense. It's on me that I didn't catch all the required `.name`, though.

psuter: (e.g. around line 57 in `Container.scala`:

```def details: String = {
        val name = containerName getOrElse "??"```

...the `val` can probably just go, now that it's defined outside (or be replaced with `containerName.map(_.name).getOrElse("??")` if we really prefer that to `anon` in this context)

markusthoemmes: But isn't changing the `val name` in `Container` to `String` exactly the opposite of "make sure to never use `String`s when something else makes more semantic sense"?

rabbah: I don't follow - isn't the type of name a string?


-r

markusthoemmes: The types are:

1. `Container.containerName` = `Option[ContainerName]`
2. `Container.name` = `ContainerName` ( @psuter  proposes to change to `String`)
3. `ContainerName.name` = `String`

psuter: (Sorry I missed the conversation until now, not sure why.) `Container.name` was and is supposed to be a `String`. By my mistake it became an `Any` but it's really supposed to be a `String` for when you want to generate other `String`s out of it and can't live with it being optional.

psuter:
Superseeded by #1305.

PR #1241, opened 2016-09-12
Add timing information to build to be able to diagnose travis buildtimes

markusthoemmes:
- Add build timing to be able to see what slows the build down
- Refactored docker.gradle


markusthoemmes:
PG1 4​29 approved

PR #1242, opened 2016-09-12
a little preparatory code for packaged actions in rules

sjfink:
no change to currently running tests -- should not need pg


rabbah:
Squashed and merged as https://github.com/openwhisk/openwhisk/commit/ff76045158172e1558a3d3ab172137ebc47ba602.

Issue #1248, opened 2016-09-13
error: Unable to invoke action using CLI when reading from file input


csantanapr:
sequence tutorial documented doesn't work with go-cli

same tutorial works with old python cli
1. Create an action sequence so that the result of one action is passed as an argument to the next action.
   
   ```
   $ wsk action create myAction --sequence /whisk.system/util/split,/whisk.system/util/sort
   ```
   
   This action sequence converts some lines of text to an array, and sorts the lines.
2. Before you invoke the action sequence, create a text file called 'haiku.txt' with a few lines of text:
   
   ```
   Over-ripe sushi,
   The Master
   Is full of regret.
   ```

Using go cli

```
$ wsk action invoke --blocking --result myAction --param payload "$(cat haiku.txt)"
error: Unable to invoke action 'myAction': Unable to create HTTP request for POST 'actions/myAction?blocking=true': Error encoding request body: json: error calling MarshalJSON for type *json.RawMessage: invalid character '\n' in string literal
```

Expected output:
4. Invoke the action:

```
  $ wsk action invoke --blocking --result myAction --param payload "$(cat haiku.txt)"
```

```
  {
      "length": 3,
      "lines": [
          "Is full of regret.",
          "Over-ripe sushi,",
          "The Master"
      ]
  }
```

rabbah:
> error: Unable to invoke action 'myAction': Unable to create HTTP request for POST 'actions/myAction?blocking=true': Error encoding request body: json: error calling MarshalJSON for type *json.RawMessage: invalid character '\n' in string literal
> 

The cli should do a better job reporting errors. Three levels of "[e|E]rror" and multiple : is confusing and ugly; surely can do better on the message and its formatting.

csantanapr:
Issue moved to [apache/incubator-openwhisk-cli #150](https://github.com/apache/incubator-openwhisk-cli/issues/150) via [**ZenHub**](https://www.zenhub.com/)

Issue #1129, opened 2016-08-26
buildAndPush.sh from dockerSkeleton is not executable


csantanapr:
## Environment details:
- Mac El Capitan
- whisk CLI version         2016-08-25T15:59:36+00:00
- whisk API build       2016-08-25T17:04:25Z
- whisk API build number    whisk-build-253
## Steps to reproduce the issue:
1.   wsk sdk install docker
2.   cd dockerSkeleton
3.   ./buildAndPush.sh csantanapr/blackboxdemo
## Provide the expected results and outputs:

```
script to run with success
```
## Provide the actual results and outputs:

```
-bash: ./buildAndPush.sh: Permission denied
```
## Additional information you deem important:

$ ls -l
total 32
-rw-r--r--  1 csantanapr  staff   281 Aug 26 09:06 Dockerfile
-rw-r--r--  1 csantanapr  staff  1096 Aug 26 09:06 README.md
**-rw-r--r--  1 csantanapr  staff   507 Aug 26 09:06 buildAndPush.sh**
-rw-r--r--  1 csantanapr  staff   625 Aug 26 09:06 example.c

csantanapr:
FYI @rabbah

rabbah:
This appears to be a problem with the `go-cli`:

```
> go-cli/wsk --apihost 192.168.99.100 -i sdk install docker 
The docker skeleton is now installed at the current directory.
> ls -l dockerSkeleton/buildAndPush.sh 
-rw-r--r--  ... dockerSkeleton/buildAndPush.sh
> rm -rf dockerSkeleton/
> wsk --apihost 192.168.99.100 -i sdk install docker
The docker skeleton is now installed at the current directory.
> ls -l dockerSkeleton/buildAndPush.sh 
-rwxr-xr-x  ... dockerSkeleton/buildAndPush.sh
```

@dubeejw @mdeuser

csantanapr:
Please update doc when this one get's resolve
We have this workaround in the docs `chmod +x buildAndPush.sh`

rabbah:
Fixed by @dubeejw.

PR #1249, opened 2016-09-13
Update docs 0912 for action tutorials

csantanapr:
Addressed multiple issues
- update sample to reflect correct value for 20sec
- Fix python sample
- rename sequence action name
- workaround for sequence not working with CLI #1248 
- workaround for user to see correct Dockerfile to use, will remove once sdk updated

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -178,7 +178,7 @@ JavaScript functions that run asynchronously may need to return the activation r
        return new Promise(function(resolve, reject) {
          setTimeout(function() {
            resolve({ done: true });
-         }, 2000);
+         }, 20000);
       })
    }
   ```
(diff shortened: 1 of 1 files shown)


rabbah:
> update sample to reflect correct value for 20sec
How about changing the text to two seconds instead - for an example waiting 20 seconds is not necessary or ideal.

rabbah:
> workaround for user to see correct Dockerfile to use, will remove once sdk updated

sdk is updated - do you mean promoted on Bluemix?

rabbah:
> + $ chmod +x buildAndPush.sh

This is also a workaround against a known go cli bug - will link it so that we remove this when the issue is fixed.

docs/actions.md
```diff
@@ -549,6 +553,31 @@ For the instructions that follow, assume that the Docker user ID is `janesmith`
...
+          gcc \
+          libc-dev \
+  ### Compile source file(s)
+  && cd /action; gcc -o exec example.c \
+  && apk del .build-deps
+
+  CMD ["/bin/bash", "-c", "cd actionProxy && python -u actionproxy.py"]  
+  ```
+  ```
+  $ chmod +x buildAndPush.sh
```
rabbah: issue https://github.com/openwhisk/openwhisk/issues/1129.

csantanapr:
@rabbah review the last two commits.
Now that docker sdk it's promoted workaround no longer need it.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/2045795e1c4f2d9c1e7237a9d8716ec921945428.

PR #1254, opened 2016-09-13
Reify types in ActivationMessage

rabbah:
Step 1 of several to come to allow activations to be stored in a specific namespace.

1. remove internal path in activation message
2. remove dispatcher matching of internal path to handlers (since there is only 1 now per dispatcher)
3. make fully qualified entity name a triple that includes an optional version
4. thread activation namespace from controller to the invoker (where the activation namespace is semantically preserving)


common/scala/src/main/scala/whisk/core/entity/EntityPath.scala
```diff
@@ -155,8 +156,20 @@ protected[core] object EntityName {
...
+ * - Version: a unique version for the resource
  *
- * Note: by convention, a "fully qualified name" would additionally include a version
+ * The version is not a SemVer (yet) because it semantic versioning of entities
+ * is not enforced. Instead this will be a context specific version identifier.
  */
-protected[core] case class EntityQName(path:EntityPath, name:EntityName)
+protected[core] case class FullyQualifiedEntityName(path: EntityPath, name: EntityName, version: Option[String]) {
+    override def toString = {
+        s"$path${EntityPath.PATHSEP}$name${version.map { v => s"@$v" }.getOrElse("")}"
```
markusthoemmes: String interpolation actually looks very clunky in this case.

`path + EntityPath.PATHSEP + name + version.map("@" + _).getOrElse("")` is a little easier on the eye i think?

core/invoker/src/main/scala/whisk/core/dispatcher/MessageHandler.scala
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Runs handler for a Kafka message. This method is run inside a future.
+     * If the method fails with an exception, the exception completes
+     * the wrapping future within which the method is run.
+     *
+     * @param msg the Message object to process
+     * @param transid the transaction id for the Kafka message
+     * @return Future that executes the handler
+     */
+    def doit(msg: Message)(implicit transid: TransactionId): Future[Any]
```
markusthoemmes: Can we rename this to `onMessage`?

rabbah:
PG2/133 approved.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerToKafka.scala
```diff
@@ -46,25 +46,22 @@ trait LoadBalancerToKafka extends Logging {
...
-     * @return msg to return in HTTP response
+     * @param msg the activation message to publish on an invoker topic
+     * @param transid the transaction id for the request
+     * @return result of publishing the message as a Future
      */
-    def doPublish(component: String, msg: Message)(implicit transid: TransactionId): Future[Unit] = {
-        getTopic(component, msg) match {
+    def publish(msg: Message)(implicit transid: TransactionId): Future[Unit] = {
+        getInvoker(msg) map {
+            i => (i, Message.invoker(i))
```
markusthoemmes: Why the indirection with a `Tuple2` and not just `getInvoker(msg) match ...` and build the topicname in the `Some` case?

markusthoemmes:
rebased squashed and merged in 841821e21690e4f812ab3d67ff7578cfbe1801a0

PR #1255, opened 2016-09-13
Add the deprecation script

houshengbo:
This patch adds a script to deprecate the packages and actions,
which have changed their names after the transition to openwhisk-catalog.

The package util has renamed to utils, and echo action has moved from
samples/echo to utils/echo.

### DIFF
--- a/catalog/deprecate.sh
+++ b/catalog/deprecate.sh
@@ -1,0 +1,99 @@
+#!/bin/bash
+#
+# This script is used to deprecate the packages and actions, which have their names changed
+# after the installation of the catalog repository.
+#
+
+# The first argument is the catalog authentication key, which can be passed via either
+# a file or the key itself.
+SCRIPTDIR="$(cd $(dirname "$0")/ && pwd)"
+OPENWHISK_HOME="$SCRIPTDIR/.."
+CATALOG_AUTH_KEY=${1:-"$OPENWHISK_HOME/ansible/files/auth.whisk.system"}
+
+# If the auth key file exists, read the key in the file. Otherwise, take the
+# first argument as the key itself.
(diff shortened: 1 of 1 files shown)


catalog/deprecate.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
+#
+# use the command line interface to install standard actions deployed
```
rabbah: comment is out of date.

catalog/deprecate.sh
```diff
@@ -0,0 +1,99 @@
...
+# Make sure that the api_host is not empty.
+: ${API_HOST:?"API_HOST must be set and non-empty"}
+
+# The api host is passed as the third argument. If it is not provided, take "/whisk.system"
+# as the default value.
+WHISK_NAMESPACE=${3:-"/whisk.system"}
+
+# If the WHISK_NAMESPACE does not start with a forward slash, add it.
+if [[ $WHISK_NAMESPACE != \/* ]] ; then
+    WHISK_NAMESPACE="/$catalog_namespace"
```
rabbah: `$catalog_namespace` not defined?

catalog/deprecate.sh
```diff
@@ -0,0 +1,99 @@
...
+        if [ $STATUS -ne 0 ]
+        then
+            let ERRORS=ERRORS+1
+        fi
+    done
+    PIDS=()
+}
+
+echo Deprecating the open catalog package util
+addPackageAnnotation util \
```
rabbah: @csantanapr here's a use case for update without create!

rabbah:
Updated path for go-cli. Merged as https://github.com/openwhisk/openwhisk/commit/ca4992f870a25b5ca980fea9a85b42cc864fefaa.

PR #1258, opened 2016-09-14
Use new Identity in API router, authentication and authorization

rabbah:
Introduces the foundation for limited scope keys: an activation only key with no CRUD rights. Commits are incremental and self-contained, working up to the point of using the new Identity type to reject requests to CRUD resources with activation keys.

This builds on #1254 (first two commits will be dropped from this pull request).

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/AuthKey.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/AuthKey.scala
@@ -16,6 +16,13 @@
 
 package whisk.core.entity
 
+import scala.util.Try
+
+import spray.json.JsString
+import spray.json.JsValue
+import spray.json.RootJsonFormat
+import spray.json.deserializationError
+
 /**
  * Authentication key, consisting of a UUID and Secret.
  *
@@ -29,7 +36,8 @@ protected[core] class AuthKey private (private val k: (UUID, Secret)) extends An
     def uuid = k._1
     def key = k._2
     def revoke = new AuthKey(uuid, Secret())
-    override def toString = s"${uuid()}:${key()}"
+    def compact = s"$uuid:$key"
+    override def toString = compact
 }
 
 protected[core] object AuthKey {
@@ -77,5 +85,14 @@ protected[core] object AuthKey {
         val k = if (parts.size >= 1) parts(0).trim else null
         val v = if (parts.size == 2) parts(1).trim else null
         (k, v)
+    }
+
+    protected[core] implicit val serdes = new RootJsonFormat[AuthKey] {
+        def write(k: AuthKey) = JsString(k.toString)
+
+        def read(value: JsValue) = Try {
+            val JsString(s) = value
+            AuthKey(s)
+        } getOrElse deserializationError("authorization key malformed")
     }
 }
--- a/common/scala/src/main/scala/whisk/core/entity/Subject.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Subject.scala
@@ -17,15 +17,18 @@
 package whisk.core.entity
 
 import scala.util.Try
+
+import spray.json.JsString
 import spray.json.JsValue
 import spray.json.RootJsonFormat
-import spray.json.JsString
 import spray.json.deserializationError
+import whisk.core.entitlement.Privilege
+import whisk.core.entitlement.Privilege.Privilege
 
 protected[core] class Subject private (private val subject: String) extends AnyVal {
     protected[core] def apply() = subject
-    protected[core] def namespace = EntityPath(subject)
     protected[entity] def toJson = JsString(subject)
+    protected[core] def toIdentity(authkey: AuthKey) = Identity(this, EntityName(subject), authkey, Privilege.ALL)
     override def toString = subject
 }
 
@@ -66,3 +69,5 @@ protected[core] object Subject extends ArgNormalizer[Subject] {
 
     private val rand = new scala.util.Random()
 }
+
+protected[core] case class Identity(subject: Subject, namespace: EntityName, authkey: AuthKey, rights: Set[Privilege])
--- a/core/controller/src/main/scala/whisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Actions.scala
@@ -24,6 +24,7 @@ import scala.concurrent.duration.FiniteDuration
 import scala.util.Failure
 import scala.util.Success
 import scala.language.postfixOps
+
 import akka.actor.ActorSystem
 import spray.http.HttpMethod
 import spray.http.HttpMethods.DELETE
@@ -41,9 +42,12 @@ import spray.httpx.SprayJsonSupport.sprayJsonUnmarshaller
 import spray.json.DefaultJsonProtocol.RootJsObjectFormat
 import spray.json.{ JsArray, JsObject, JsString }
 import spray.routing.RequestContext
+import org.apache.kafka.common.errors.RecordTooLargeException
+
 import whisk.common.LoggingMarkers
 import whisk.common.StartMarker
 import whisk.common.TransactionId
+import whisk.common.PrintStreamEmitter
 import whisk.core.database.NoDocumentException
 import whisk.core.entity.ActionLimits
 import whisk.core.entity.ActivationId
@@ -67,18 +71,15 @@ import whisk.core.entity.types.ActivationStore
 import whisk.core.entity.types.EntityStore
 import whisk.core.entitlement.Collection
 import whisk.core.entitlement.Privilege
-import whisk.core.entity.WhiskAuth
 import whisk.core.connector.{ ActivationMessage => Message }
 import whisk.core.entitlement.Resource
 import whisk.core.entity.WhiskPackage
 import whisk.core.entity.Binding
 import whisk.core.entity.Subject
-import whisk.http.ErrorResponse.terminate
-import whisk.common.PrintStreamEmitter
-import org.apache.kafka.common.errors.RecordTooLargeException
-import whisk.utils.ExecutionContextFactory.FutureExtensions
-import scala.concurrent.duration.FiniteDuration
 import whisk.core.entity.FullyQualifiedEntityName
+import whisk.core.entity.Identity
+import whisk.http.ErrorResponse.terminate
+import whisk.utils.ExecutionContextFactory.FutureExtensions
 
 /**
  * A singleton object which defines the properties that must be present in a configuration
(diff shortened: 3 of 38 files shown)


core/controller/src/main/scala/whisk/core/entitlement/RemoteEntitlement.scala
```diff
@@ -22,12 +22,14 @@ import scala.concurrent.duration.DurationInt
...
 import scala.util.Try
 
+import Privilege.Privilege
 import akka.actor.ActorSystem
 import spray.client.pipelining.Get
 import spray.client.pipelining.Post
 import spray.client.pipelining.WithTransformerConcatenation
 import spray.client.pipelining.addHeader
 import spray.client.pipelining.sendReceive
+import spray.client.pipelining.sendReceive$default$3
```
markusthoemmes: `import spray.client.pipelining._`?

core/controller/src/main/scala/whisk/core/entitlement/RemoteEntitlement.scala
```diff
@@ -37,12 +39,13 @@ import spray.http.StatusCodes.Unauthorized
 import spray.http.Uri
 import spray.httpx.SprayJsonSupport.sprayJsonUnmarshaller
 import spray.httpx.UnsuccessfulResponseException
-import spray.json._
+import spray.json.DefaultJsonProtocol
 import spray.json.DefaultJsonProtocol._
+import spray.json.pimpString
```
markusthoemmes: Keep the imports wildcarded?

common/scala/src/main/scala/whisk/core/entity/Subject.scala
```diff
@@ -17,10 +17,12 @@
 package whisk.core.entity
 
 import scala.util.Try
+
+import spray.json.JsString
 import spray.json.JsValue
 import spray.json.RootJsonFormat
-import spray.json.JsString
 import spray.json.deserializationError
```
markusthoemmes: `import spray.json._`

common/scala/src/main/scala/whisk/core/entity/AuthKey.scala
```diff
@@ -16,6 +16,13 @@
 
 package whisk.core.entity
 
+import scala.util.Try
+
+import spray.json.JsString
+import spray.json.JsValue
+import spray.json.RootJsonFormat
+import spray.json.deserializationError
```
markusthoemmes: `import spray.json._`

core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
```diff
@@ -69,8 +73,17 @@ trait Authenticate extends Logging {
...
+                    Try {
+                        Identity(s.subject, EntityName(s.subject()), s.authkey, Privilege.ALL)
+                    } match {
+                        case Success(id) => promise.success(Some(id))
+                        case Failure(t) =>
+                            error(this, s"failed to created identity profile: $t")
+                            promise.failure(t)
+                    }
+                case Success(None) =>
+                    promise.success(None)
```
markusthoemmes: I believe we can get rid of the Promise, will try to provide a patch proposal later.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -214,11 +206,37 @@ protected[core] abstract class EntitlementService(config: WhiskConfig)(
         grant
     }
 
+    // NOTE: explicit grants do not work with package bindings because the current model
+    // for authorization does not allow for a continuation to check that both the binding
+    // and the references package are both either implicitly or explicitly granted; this is
+    // accepted for the time being however because there exists no external mechanism to create
+    // explicit grants
+    protected def check(subject: Subject, right: Privilege, resource: Resource)(
```
markusthoemmes: Note to self: Understand why we need this.

core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
markusthoemmes: Proposed patch.

```
diff --git a/core/controller/src/main/scala/whisk/core/controller/Authenticate.scala b/core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
index 444f049..325ed09 100644
--- a/core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
@@ -56,45 +56,33 @@ trait Authenticate extends Logging {
      * authentication directive.
      */
     def validateCredentials(userpass: Option[UserPass])(implicit transid: TransactionId): Future[Option[Identity]] = {
-        val promise = Promise[Option[Identity]]
-        userpass map { pw =>
+        val future = userpass.map { pw =>
             Future {
-                // inside a future should these fail to validate
                 AuthKey(UUID(pw.user), Secret(pw.pass))
             } flatMap { authkey =>
                 info(this, s"authenticate: ${authkey.uuid}")
                 WhiskAuth.get(authStore, authkey.uuid) map { result =>
                     if (authkey == result.authkey) {
                         info(this, s"authentication valid")
-                        Some(result)
+                        Some(Identity(result.subject, EntityName(result.subject()), result.authkey, Privilege.ALL))
                     } else {
                         info(this, s"authentication not valid")
                         None
                     }
                 }
-            } onComplete {
-                case Success(Some(s)) =>
-                    Try {
-                        Identity(s.subject, EntityName(s.subject()), s.authkey, Privilege.ALL)
-                    } match {
-                        case Success(id) => promise.success(Some(id))
-                        case Failure(t) =>
-                            error(this, s"failed to created identity profile: $t")
-                            promise.failure(t)
-                    }
-                case Success(None) =>
-                    promise.success(None)
-                case Failure(_: NoDocumentException | _: IllegalArgumentException) =>
+            } recover {
+                case _: NoDocumentException | _: IllegalArgumentException =>
                     info(this, s"authentication not valid")
-                    promise.success(None)
-                case Failure(t) =>
-                    info(this, s"authentication error: $t")
-                    promise.failure(t)
+                    None
             }
-        } getOrElse {
+        }.getOrElse {
             info(this, s"credentials are malformed")
-            promise.success(None)
+            Future.successful(None)
+        }
+
+        future.onFailure {
+            case t: Throwable => info(this, s"authentication error: $t")
         }
-        promise.future
+        future
     }
 }
```

common/scala/src/main/scala/whisk/core/entity/AuthKey.scala
rabbah: The explicit imports are a function of the IDE's organize import facility (used for example to import new dependencies or remove unused ones). Is there a way to configure it to use wildcards when the number of imports from a package or class exceeds a threshold? Otherwise this will turn into another edit war.

markusthoemmes: You can ignore these comments if you want. But I'ld strongly recommend to configure your IDE to leave Wildcard imports alone and always use the wildcard ones, at least for spray.json. It makes later editing of these classes and files a whole lot easier. Won't block the PR though, just a thing i noticed.

rabbah: > But I'ld strongly recommend to configure your IDE to leave Wildcard imports alone

We need to automate and standardize such conventions otherwise...

markusthoemmes: Agreed, take the comments as a friendly nudge :)

common/scala/src/main/scala/whisk/core/entity/Subject.scala
```diff
@@ -17,15 +17,19 @@
...
+import whisk.core.entitlement.Privilege.Privilege
+import whisk.core.entitlement.Privilege
 
 protected[core] class Subject private (private val subject: String) extends AnyVal {
     protected[core] def apply() = subject
     protected[core] def namespace = EntityPath(subject)
     protected[entity] def toJson = JsString(subject)
+    protected[core] def toIdentity = Identity(this, EntityName(subject), AuthKey(), Privilege.ALL)
```
markusthoemmes: Not sure about this, as it creates a "dummy" `Identity` (for testing?) which might be misleading in usual coding.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -424,15 +425,22 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
         // merge package parameters with action (action parameters supersede), then merge in payload
         val args = { env map { _ ++ action.parameters } getOrElse action.parameters } merge payload
-        val message = Message(transid, s"/actions/invoke/${action.namespace}/${action.name}/${action.rev}", user, activationId.make(), args)
+        val message = Message(
+            transid,
+            FullyQualifiedEntityName(action.namespace, action.name, Some(action.rev())),
+            user.subject,
+            user.authkey,
+            activationId.make(),
+            user.subject.namespace,
```
markusthoemmes: Should this rather be `user.namespace` (which also currently falls back to `subject` being the namespace anyway)? Would be more in line with where we want to be.

common/scala/src/main/scala/whisk/core/entity/Subject.scala
rabbah: > Not sure about this, as it creates a "dummy" Identity (for testing?) which might be misleading in usual coding.

Negative. I meant to use this in the authentication trait.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
markusthoemmes: Got it, @rabbah want to rename this to `checkPrivilege` or thelike. Overloading `check` seems like both are equivalent in effects, while they aren't.

rabbah:
pg2/145.

core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
```diff
@@ -64,17 +64,18 @@ trait Authenticate extends Logging {
                         info(this, s"authentication not valid")
                         None
                     }
+                } recover {
+                    case _: NoDocumentException | _: IllegalArgumentException =>
+                        info(this, s"authentication not valid")
+                        None
+                    case t: Throwable =>
+                        error(this, s"authentication error: $t")
+                        None
```
markusthoemmes: This is not semantically preserving i believe. In this case you `recover` any error to a `None` whereas before, only `NoDocumentException` and `IllegalArgumentException` have been recovered. Was this change intended?

common/scala/src/main/scala/whisk/core/entity/AuthKey.scala
rabbah: IDE settings updated for future reviewing pleasure.

rabbah:
pg1/458 re-approved.

markusthoemmes:
rebased and merged in 6152dbad30e9df58763c94a06674def61922c5d0

PR #1261, opened 2016-09-14
update vagrant and readmes with utils/echo

csantanapr:


rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/aef7885f8277977ef1a44cb81ea73db816edb007.

PR #1224, opened 2016-09-08
Update CLI Build

dubeejw:
- Place binary at the root of uncompressed folder
- Rename compressed files
- Make tar.gz extensions .tgz
- Compress Mac binary with zip

Fixes https://github.com/openwhisk/openwhisk/issues/1039

PR #1265, opened 2016-09-14
Do Not Cross Compile CLI by Default

dubeejw:
- Use build argument to determine whether to cross-compile the CLI
- Use build argument to determine whether or not to run cross-compiled CLI tests

### DIFF
--- a/tests/src/system/rest/GoCLINginxTests.scala
+++ b/tests/src/system/rest/GoCLINginxTests.scala
@@ -25,12 +25,12 @@ import com.jayway.restassured.RestAssured
 
 import spray.json._
 import DefaultJsonProtocol._
+
 /**
  * Basic tests of the download link for Go CLI binaries
  */
 @RunWith(classOf[JUnitRunner])
 class GoCLINginxTests extends FlatSpec with Matchers with RestUtil {
-
     val ProductName = "OpenWhisk_CLI"
     val DownloadLinkGoCli = "cli/go/download"
     val OperatingSystems = List("mac", "linux", "windows")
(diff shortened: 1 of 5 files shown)


rabbah:
Should we make this the default?

csantanapr:
build only one binary by default
and implement a flag that get's passed to gradle distDocker to build all

csantanapr:
from gradle you can get the current, os, and arch you can try to get or just assume 64bit

csantanapr:
Build all could driven using env variable that way both the build and tests know to do handle build and test all binaries

dubeejw:
From the Travis build output: Building 'go-cli' took 1 minutes, 30.747 seconds.

tools/go-cli/Dockerfile
```diff
@@ -34,33 +34,70 @@ RUN mkdir /src/github.com/go-whisk-cli/build
...
+        export GOOS=$LOCAL_OS && \
+        export ARCHS=$LOCAL_ARCH; \
+      else \
+        unset GOOS && \
+        unset ARCHS; \
+      fi; \
+    else \
+      export GOOS=windows && \
+      export ARCHS="386 amd64"; \
+    fi && \
```
markusthoemmes: A lot of duplication in these, can we have only on if disambiguating this?

```
if [ "$LOCAL_BUILD" = "true" ]; then \
  export OSS=$LOCAL_OS && \
  export ARCHS=$LOCAL_ARCH; \
else \
  export OS="windows darwin linux" && \
  export ARCHS="386 amd64"; \
fi && \


for GOOS in $OSS; do \
  EXEC=wsk \
  if [ "$GOOS" = "windows" ]; then \
    EXEC="${EXEC}.exe" \
  fi \
  for GOARCH in $ARCHS; do \
    cd /src/github.com/go-whisk-cli && \
    go build -ldflags "-X main.CLI_BUILD_TIME=`date -u '+%Y-%m-%dT%H:%M:%S%:z'`" -v -o build/$GOOS/$GOARCH/$EXEC main.go && \
    if [ "$GOOS" = "windows"]; then \
      zip build/$GOOS/$GOARCH/openwhisk-win-$GOARCH.zip build/$GOOS/$GOARCH/$EXEC; \
    else \
      if [ "$GOOS" = "darwin" ]; then; export GOOS=mac; fi \
      tar -cf build/$GOOS/$GOARCH/openwhisk-$GOOS-$GOARCH.tar build/$GOOS/$GOARCH/$EXEC && \
      gzip < build/$GOOS/$GOARCH/openwhisk-$GOOS-$GOARCH.tar > build/$GOOS/$GOARCH/openwhisk-$GOOS-$GOARCH.tar.gz && \
      rm -f build/$GOOS/$GOARCH/openwhisk-$GOOS-$GOARCH.tar; \
    fi \
  done \
done \
```

should be roughly the same as the three different `RUN` statements. Will safe layers and is easier to maintain imho. (Not tested, highly likely there are errors in this)

ddragosd: I think too that it's easier to read in this shape without much repetition. What would the difference be if it were to have a build script invoked from a single RUN command ?
```
RUN build.sh
```

csantanapr: I think that's the idea, the PR is `wip` so I give a chance to @dubeejw to have it ready for review.
here is POC: on how would look like running from a script, and the script can even be smaller
https://github.com/openwhisk/openwhisk/compare/master...csantanapr:gocli_build_script

dubeejw:
I'll probably role this update into https://github.com/openwhisk/openwhisk/pull/1224. Doesn't make sense to rewrite the same file twice.

dubeejw:
@rabbah, I'm not sure that only building one CLI by default is the best option. The need to build only one binary is the result of wanting to speed up the local build process for developers. By building only one binary by default, we would complicate the production build process and testing environment. For instance, a flag to build/test all binaries in such environments would be needed even though we want all of the cross-platform binaries to always to be built in the mentioned environments. I think it makes more sense to require a developer that wants to speed up his, or her local build to set a necessary environment variable or pass a flag to the build.

dubeejw:
Plus we could easily add a command to redo that would allow for only building the local binary.

ddragosd:
IMHO the time-to-value is important for a person who's giving OpenWhisk a try. In my case it became a blocker as the build would timeout so I had to debug the problem and make some changes myself. In such a case the option suggested by @rabbah would be helpful.

dubeejw:
PG2 351.

tools/cli/build.sh
```diff
@@ -98,8 +98,15 @@ create_cli_packages() {
...
-platforms="mac linux windows"
-archs="386 amd64"
+
+if [ "$LOCAL_BUILD" = "true" ]; then \
+  platforms="$LOCAL_OS" && \
+  archs="$LOCAL_ARCH"; \
+else \
+  platforms="mac linux windows" && \
+  archs="386 amd64"; \
+fi && \
```
markusthoemmes: Why make the distinction here? Why don't we just pass `platforms` and `archs` through from the gradle project. `crosscompile` then means: pass this whole list. Thoughts?

tests/src/system/rest/GoCLINginxTests.scala
```diff
@@ -37,7 +43,7 @@ class GoCLINginxTests extends FlatSpec with Matchers with RestUtil {
     val Architectures = List("386", "amd64")
     val ServiceURL = getServiceURL()
 
-    it should s"respond to all files in root directory" in {
+    it should s"respond to all files in root directory" taggedAs(crossCompileCLI) in {
```
dubeejw: Any Scala suggestions on the best way to check the JSON to see if specific keys exist in the JSON index file?

markusthoemmes: You can use Akka-HTTP + built in JSON unmarshallers (check `ConsulClient` for an example of that) to get the JSON in a programmatic, safe way.

dubeejw: @markusthoemmes, having trouble parsing the JSON in Scala as it will be dynamic.

dubeejw:
PG2 475.

dubeejw:
@markusthoemmes, anymore comments on this one?

tests/src/system/rest/GoCLINginxTests.scala
```diff
@@ -42,80 +43,60 @@ class GoCLINginxTests extends FlatSpec with Matchers with RestUtil {
...
+        cli.foreach {
+            case(os, arch) =>
+                arch.foreach {
+                    case (arch, path) =>
+                        path.foreach {
+                            case (path, value) =>
+                                RestAssured.given().config(sslconfig).
+                                  get(s"$ServiceURL/$DownloadLinkGoCli/$value").statusCode should be(200)
+                        }
+                }
```
markusthoemmes: This should be equivalent to:

```
cli.values.flatMap(_.values).flatMap(_.values).foreach { path =>
    RestAssured.given().config(sslconfig).get(s"$athServiceURL/$DownloadLinkGoCli/$path").statusCode should be(200)
}
```

Maybe you find that more readable?

markusthoemmes:
Do you have a PG on the latest refactorings?

dubeejw:
PG2 497.

markusthoemmes:
@dubeejw As a sanity check, I checked build times for the go CLI in this travis build and another (we mainly do this for build time optimization, yes?). I noticed that it takes 2:33min in this build and exactly as long in another build without these changes. Would you investigate?

dubeejw:
@markusthoemmes, As a test, I did a couple of builds. The results are below. I think not cross-compiling the CLI does not save as much time anymore because a lot of changes have gone in to make the build faster. However, it looks like not cross-compiling the CLI does shave off a little bit of time from the build.

Do not cross-compile CLI:
```
./gradlew tools:cli:distDocker
Total time: 49.633 secs
```

Cross-compile CLI:
```
./gradlew tools:cli:distDocker -PcrossCompileCLI=true
Total time: 1 mins 23.203 secs
```

dubeejw:
@ddragosd, can you confirm that this PR helps your build issues?

rabbah:
for local dev without pr:
Total time: 1 mins 34.1 secs

with pr:
Total time: 34.272 secs

On a slow connection, this is even worse. I favor adopting this. If the complexity for the blue deployment is mitigated by flipping the defaults, I'm ok with that too if we can stuff the bits into `wskdev`.

markusthoemmes:
Yep we agreed to adopt it, just a matter of me merging all the dependencies. Will get to it ASAP.

markusthoemmes:
rebased and merged in 79d225546247f87a1d144edee054803be13b83c7

Issue #450, opened 2016-05-17
Provide Whisk SDK for Python


ioana-blue:
I think it would be really nice to have a small SDK for Python that allows a python coder to access whisk rest api from their python action very easily. For example, the SDK should support invoking actions, firing triggers, getting logs, etc. I believe there is such an SDK for javascript.

rabbah:
Repurpose the `wsk cli`?

ioana-blue:
One thing that would be useful part of the SDK is a way to retrieve auth credentials such that it's easier for a python action to invoke a different action.

PR #1269, opened 2016-09-15
Remove Python CLI

dubeejw:
- Remove Python CLI from the build, and deploy process
- Remove Python CLI source code
     - Backup located here: https://github.com/openwhisk/openwhisk-python-cli
- Remove all Python CLI references


ansible/README.md
```diff
@@ -156,7 +156,7 @@ data that include actions and activations are lost.
 Use `ansible-playbook -i environments/<environment> openwhisk.yml` to avoid wiping the data store. This is useful to start OpenWhisk after restarting your Operating System.
 
 ### Verification after Deployment
-After a successful deployment you can use the `wsk` CLI (located in the `bin` folder of the repository) to verify that OpenWhisk is operable.
+After a successful deployment you can use the `wsk` CLI (located in the `bin\go-cli` folder of the repository) to verify that OpenWhisk is operable.
```
rabbah: `bin/go-cli`
actually why not make the new CLI in `bin` we no longer need the `go-cli` distinction.

rabbah:
Having looked through the PR, we should not rename `cli` to `go-cli` and instead move `go-cli` to `cli` instead. The `go-` distinction is no longer necessary once the python cli is removed (and in fact it could be misleading, as in a cli of `go`).

rabbah:
Another point - since we want to use the python version of the CLI as part of the python actions (as an SDK for whisk in python), perhaps an intermediate step is to move the directory into `pythonAction` until we have the plumbing in place to build and package its inclusion. Thoughts @csantanapr?

ioana-blue:
I agree with RR that we should keep the code around. It really needs to be resurrected as a python SDK instead of completely removed. See #450

dubeejw:
@rabbah, I had to leave some of the files in tools/cli in place (wskadmin, wskprop.py, wskutil.py) as other parts of the system that are written in Python rely on them. 

Just to be clear, do you want a "bin/cli" folder, and move "tools/go-cli" to "tools/cli" ?

dubeejw:
Also, just to be clean, I've backed up the old CLI here:  https://github.com/openwhisk/openwhisk-python-cli with commits preserved.

rabbah:
is it plausible to move `wskadmin, wskprops and wskutils` to `bin`?
it makes sense to me to rename `tools/go-cli` to `tools/cli`.

csantanapr:
I don't like having dead stale code in the repo, but I think if we want to maintain the code base up to date as an sdk lib for python, let's keep it in the repo for now. 


Here is my proposal
`sdk/phython/` (wsk client library for python can be use from inside an action or outside the system)
`tools/cli/` (the go based cli, we don't need the string `go` in directory names anymore)
This one I'm not sure:
Option 1
`bin/wsk` (when the build runs this will be the binary specific to the current OS and Arch)
`bin/cli/[linux/,mac/,windows/]` (cli for all binaries OS/Arch)
Option 2
`bin/wsk` (when the build runs this will be the binary specific to the current OS and Arch)
`bin/[linux/,mac/,windows/]` (cli for all binaries OS/Arch)
Option 3
`bin/wsk/wsk` (when the build runs this will be the binary specific to the current OS and Arch)
`bin/wsk/[linux/,mac/,windows/]` (cli for all binaries OS/Arch)

rabbah:
+1 for option 2; for local development there will be only one anyway that's useful - should these even be in bin vs say a build dir?

dubeejw:
A change to openwhisk-catalog will need to happen to accommodate theses changes.

csantanapr:
@dubeejw all repos and tests jobs that reference this locations would need to be updated and synchronize accordantly

csantanapr:
@rabbah 
>+1 for option 2; for local development there will be only one anyway that's useful - should these even be in bin vs say a build dir?

Good point maybe Option 2A ? using `build/`
`bin/wsk` (when the build runs this will be the binary specific to the current OS and Arch)
`build/wsk/[linux/,mac/,windows/]` (cli for all binaries OS/Arch)

ansible/roles/cli/tasks/clean.yml
```diff
@@ -1,8 +1,7 @@
 ---
-# Remove cli_conf_dir.
 
-- name: remove cli config directory
+- name: remove go cli nginx directory
```
rabbah: ~~go~~

dubeejw:
PG2 171

rabbah:
Resolved conflict, fixed one omission and merged as https://github.com/openwhisk/openwhisk/commit/31632bddb8401316fc2378242cb7d06b41dc72ca.

Issue #1253, opened 2016-09-13
Write method for package resolution and intermediate entities in package bindings


ioana-blue:
Write a method that given a package produces the fully resolved package name wrt to package bindings. Write a method that given a package produces all the entities for which entitlement needs to be checked (intermediary packages due to bindings).

Refactor existing code to use these methods.

PR #1270, opened 2016-09-15
Issue #1253 helpers to resolve packages and bindings

ioana-blue:
Add methods for resolving packages and retrieving all intermediary bindings. Refactor code to use these methods. No semantic change intended

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskPackage.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskPackage.scala
@@ -16,6 +16,10 @@
 
 package whisk.core.entity
 
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+
 import scala.language.postfixOps
 import scala.util.Try
 import spray.json.DefaultJsonProtocol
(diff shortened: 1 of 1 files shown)


ioana-blue:
@rabbah after some iterations, only one helper function left as discussed.

common/scala/src/main/scala/whisk/core/entity/WhiskPackage.scala
```diff
@@ -140,6 +148,20 @@ object WhiskPackage
     val bindingFieldName = "binding"
     override val collectionName = "packages"
 
+    /**
+     * an utility function that traverses all potential binding references and returns the resolved package
+     * @param pkg the package for which the resolution to be applied
+     * @return the same package if there is no binding, or the fully resolved package
+     */
+    def resolveBinding(entityStore: EntityStore, pkg: DocId)(
```
rabbah: Isn't there a layer on top of this which given a sequence of strings resolves those into fully qualified names?

ioana-blue: It depends on how it is used. For one action, there will not be such a sequence. For a sequence of actions, you'll have such a layer.

ioana-blue:
I think I will close this and incorporate it in the work for #1274 that actually uses it. It's a small code change and better have it together with its usage.

PR #1271, opened 2016-09-15
updating test actions to use promises

jasonpet:
handles part of issue #932 
the others are handled in PR https://github.com/openwhisk/openwhisk-catalog/pull/138

### DIFF
--- a/tests/dat/actions/wcbin.js
+++ b/tests/dat/actions/wcbin.js
@@ -5,25 +5,19 @@
 function main(params) {
     var str = params.payload;
     console.log("The payload is '" + str + "'");
-    whisk.invoke({
-        name : 'wc',
-        parameters : {
-            payload : str
+
+    return whisk.invoke({
+        name: 'wc',
+        parameters: {
+            payload: str
         },
-        blocking : true,
-        next : function(error, activation) {
-            console.log('activation:', activation);
-            if (!error) {
-                var wordsInDecimal = activation.result.count;
-                var wordsInBinary = wordsInDecimal.toString(2) + ' (base 2)';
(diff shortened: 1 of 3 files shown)


jasonpet:
pg2#153

tests/dat/actions/wcbin.js
```diff
@@ -5,25 +5,26 @@
...
-    whisk.invoke({
-        name : 'wc',
-        parameters : {
-            payload : str
-        },
-        blocking : true,
-        next : function(error, activation) {
+
+    var promise = new Promise(function(resolve, reject) {
+        whisk.invoke({
```
bjustin-ibm: @jasonpet What do you think about this instead? I _believe_ it is functionally the same (minus one `console.log()` in the error condition) and does not require you to explicitly instantiate a promise:

```javascript
function main(params) {
    var str = params.payload;
    console.log("The payload is '" + str + "'");

    return whisk.invoke({
            name: 'wc',
            parameters: { payload: str },
            blocking: true})
      .then(function (activation) {
            console.log('activation:', activation);
            var wordsInDecimal = activation.result.count;
            var wordsInBinary = wordsInDecimal.toString(2) + ' (base 2)';

            return { binaryCount: wordsInBinary };
        });
    });
}
```

jasonpet: I like it and actually started with this exact same syntax:)  Started second guessing myself though and changed it.  Will test it out and go with this if it works.

bjustin-ibm:
Rebased/merged as 8cceedec5e19b7d61a6291d1c7cc4efa2a0994ce

PR #1272, opened 2016-09-15
Upload logs to Box when build completes

psuter:
(Applies to TravisCI logs only.)

Waiting to see if it works in production. This PR should have its logs uploaded.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,3 +26,4 @@ before_install:
 
 script:
   - ./tools/travis/build.sh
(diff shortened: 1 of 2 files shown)


markusthoemmes:
I experimented with this as well, you're aware that only `master` will be able to publish its logs?

psuter:
Yeah, I figured it out eventually :/

Branches in `openwhisk/openwhisk` would work as expected. Maybe we can
enable that just for those? For situations where everything is fine locally
but Travis fails, we could create such a branch..

Alternatively we can setup some public ftp server.

Only unsatisfactory solutions it seems.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/e33ead32224e1a640c906cd0cfb155dcda44ad7a.

PR #1275, opened 2016-09-16
Use variable to represent CLI executable path

jasonpet:
CLI location is changing and this will affect the catalog repository.  Updating the catalog script to allow the CLI path to be passed in as a variable.


jasonpet:
pg2#163

PR #1278, opened 2016-09-16
Move wskadmin to tools/admin.

rabbah:
`wskadmin` does not belong in pythonAction; moved to `tools/admin`.
Removed reference to python whisk cli from eclipse project resources. 
Removed unnecessary scanCode exemptions.

pg2/172 and 174 approved.


dubeejw:
Here are other Python scripts that rely on the files from the Python CLI: tools/health/killComponent, tools/health/kvstore, tools/deploy/deployInvoker. These files might need to be updated to accommodate your purposed changes.

dubeejw:
I would not recommend having a two copies of wskprop.py, and wskutil.py in different locations as they might accidentally diverge over time.

rabbah:
> I would not recommend having a two copies of wskprop.py, and wskutil.py in different locations as they might accidentally diverge over time.

They are already different - I removed the code that is not needed by wskadmin. We could rename the one set of these files but since the code in corePython/cli is transitioning, I view this as unnecessary.

rabbah:
New commits pushed; new pg pending.

rabbah:
pg1/518 approved.

PR #1282, opened 2016-09-18
Allow Namespace to be Set When Setting an API Host

dubeejw:
- Fix bug that prevents a namespace from being set at the same time the API is set
- Add test to ensure namespace can be set with API host and auth


dubeejw:
PG2 198.

csantanapr:
LGTM

PR #1295, opened 2016-09-20
gitignore tests of packages that are checked out during deployment

cbickel:


cbickel:
PG1#498 is running

rabbah:
Per @jasonpet this is temporary and will be corrected from the push-notification test suite. Should we hold off?

cbickel:
Files are not copied anymore.

PR #1298, opened 2016-09-20
Adjust throttle tests to settle correctly should test fail.

rabbah:
@perryibm I think we need these additional fixes to the test.
pg1/514 approved.

### DIFF
--- a/tests/src/limits/ThrottleTests.scala
+++ b/tests/src/limits/ThrottleTests.scala
@@ -29,18 +29,18 @@ import org.scalatest.Matchers
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.junit.JUnitRunner
 
-import spray.json._
-import spray.json.DefaultJsonProtocol._
-
 import common.TestHelpers
 import common.TestUtils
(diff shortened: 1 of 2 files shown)


tests/src/limits/ThrottleTests.scala
```diff
@@ -67,9 +69,6 @@ class ThrottleTests
     println(s"maximumFiringsPerMinute  = $maximumFiringsPerMinute")
     println(s"maximumConcurrentInvokes = $maximumConcurrentInvokes")
 
-    val rateMessage = "Too many requests from user"
-    val concurrencyMessage = "The user has sent too many requests in a given amount of time."
```
markusthoemmes: Don't we want those, to actually ensure we hit the correct throttle?

rabbah:
These are unused references. The error messages are defined in a singleton object shared by the controller and the tests.

tests/src/limits/ThrottleTests.scala
markusthoemmes: Ugh, early morning review, you're right of course.

PR #1301, opened 2016-09-21
Reading old invoker state upon startup

markusthoemmes:
This is intended to be a stop-gap solution to make the Invoker restartable until the new algorithm by @perryibm is in place.


perryibm:
Thanks - consul work still in queue but getting bumped.   For this temp fix, has it been tested with an actual restart?   Regular PG won't test this I think.

markusthoemmes:
Well yeah, i tried locally and it counted from the state where it left of.

perryibm:
227 was not clean - retrying PG2 232

markusthoemmes:
PG1 526 was fine.

perryibm:
rebase/merged

PR #1289, opened 2016-09-20
Remove unnecessary use of DocInfo on DB get where DocId is sufficient.

rabbah:
@starpit FYI.
pg2/201.

PR #1302, opened 2016-09-21
replace InMemoryCache with new impl that avoids read under write race

### LINKED ISSUE #1260: improvements to in-memory-cache
The in-memory cache used in the controller has some deficiencies that we should address:

1. it does not expire entries in the cache pro-actively, hence an inconsistency with the database will cause stale records to be observed until the controller is restarted
2. it is prone to a race between a read and write (delete or update) because the state transition modeling is not complete
3. it does not coalesce requests (no hit under miss)

@starpit FYI.

starpit:
This PR introduces a new caching structure, to replace the existing `InMemoryCache`. The goal is to address the read-under-write and write-under-read race conditions discussed in #1260. 

I'm open to suggestions for better names for the new data structure.

### Summary of Data Structure
The new cache operates by:
  - on an ***uncontested*** read or write, it reserves a slot in the cache (i.e. reservation upon initiation)
  - each slot (cache `Entry`) has a `State` that is used to determine how to proceed with contested operations. States are ReadInProgress, WriteInProgress, Cached, InvalidateInProgress
  - ***contested*** operations obey the state transition diagram discussed in #1260
  - reads under reads are coalesced
  - reads under writes read-around the cache
  - writes under reads write-around the cache

### Extra Notes
  * leverages #1289 to remove the multi-key updates that were previously in place.
  * `InMemoryCache` used split get-then-put. This new cache leverages the putIfAbsent behavior of the underlying LRUCache --- necessarily now, in order to avoid races between the get and the put (now that we are reserving-upon-initiation).

### Testing
  - localhost testing SUCCESS
  - PG1 515 (two false flag failures -- timeouts in blackbox operations)
  - PG1 519 SUCCESS
  - Travis green

closes #1260

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
@@ -144,8 +144,6 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
 
             val key = cacheKeyForUpdate(doc)
 
-            cacheInvalidate(key)
-
             cacheUpdate(doc, key, db.put(doc) map { docinfo =>
                 doc match {
                     // if doc has a revision id, update it with new version
@@ -168,10 +166,13 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
             require(doc != null, "doc undefined")
         } map { _ =>
             implicit val logger: Logging = db
+            implicit val ec = db.executionContext
+
             val key = doc.id.asDocInfo
-            cacheInvalidate(key)
-            val src = StreamConverters.fromInputStream(() => bytes)
-            db.attach(doc, attachmentName, contentType, src)
+            cacheInvalidate(key, {
+                val src = StreamConverters.fromInputStream(() => bytes)
+                db.attach(doc, attachmentName, contentType, src)
+            })
         } match {
             case Success(f) => f
             case Failure(t) => Future.failed(t)
@@ -185,9 +186,10 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
             require(doc != null, "doc undefined")
         } map { _ =>
             implicit val logger: Logging = db
+            implicit val ec = db.executionContext
+
             val key = doc.id.asDocInfo
-            cacheInvalidate(key)
-            db.del(doc)
+            cacheInvalidate(key, db.del(doc))
         } match {
             case Success(f) => f
             case Failure(t) => Future.failed(t)
--- a/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
@@ -169,6 +169,8 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
             implicit val ec = db.executionContext
 
             val key = doc.id.asDocInfo
+            // invalidate the key because attachments update the revision;
+            // do not cache the new attachment (controller does not need it)
             cacheInvalidate(key, {
                 val src = StreamConverters.fromInputStream(() => bytes)
                 db.attach(doc, attachmentName, contentType, src)
--- a/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
@@ -142,7 +142,7 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
             implicit val logger = db: Logging
             implicit val ec = db.executionContext
 
-            var key = cacheKeyForUpdate(doc)
+            val key = cacheKeyForUpdate(doc)
 
             cacheInvalidate(key)
 
@@ -152,7 +152,7 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
                     case w: DocumentRevisionProvider => w.revision[W](docinfo.rev)
                 }
                 docinfo
-            }).future
+            })
 
         } match {
             case Success(f) => f
@@ -168,7 +168,7 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
             require(doc != null, "doc undefined")
         } map { _ =>
             implicit val logger: Logging = db
-            var key = doc.id.asDocInfo
+            val key = doc.id.asDocInfo
             cacheInvalidate(key)
             val src = StreamConverters.fromInputStream(() => bytes)
             db.attach(doc, attachmentName, contentType, src)
@@ -185,7 +185,7 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
             require(doc != null, "doc undefined")
         } map { _ =>
             implicit val logger: Logging = db
-            var key = doc.id.asDocInfo
+            val key = doc.id.asDocInfo
             cacheInvalidate(key)
             db.del(doc)
         } match {
--- a/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
@@ -120,7 +120,7 @@ trait DocumentSerializer {
  * but the get permits a datastore of its super type so that a single datastore client
  * may be used for multiple types (because the types are stored in the same database for example).
  */
-trait DocumentFactory[W] extends InMemoryCache[W] {
+trait DocumentFactory[W] extends MultipleReadersSingleWriterCache/*InMemoryCache*/[W, DocInfo] {
     /**
      * Puts a record of type W in the datastore.
      *
@@ -142,16 +142,18 @@ trait DocumentFactory[W] extends InMemoryCache[W] {
             implicit val logger = db: Logging
             implicit val ec = db.executionContext
 
-            cacheInvalidate(cacheKeys(doc))
-            db.put(doc) map { docinfo =>
+            var key = cacheKeyForUpdate(doc)
+
+            cacheInvalidate(key)
+
+            cacheUpdate(doc, key, db.put(doc) map { docinfo =>
                 doc match {
                     // if doc has a revision id, update it with new version
                     case w: DocumentRevisionProvider => w.revision[W](docinfo.rev)
                 }
-                // cache put result iff put future was successful
-                cacheUpdate(cacheKeys(doc), doc)
                 docinfo
-            }
+            }).future
+
         } match {
             case Success(f) => f
             case Failure(t) => Future.failed(t)
@@ -166,7 +168,8 @@ trait DocumentFactory[W] extends InMemoryCache[W] {
             require(doc != null, "doc undefined")
         } map { _ =>
             implicit val logger: Logging = db
-            cacheInvalidate(Set(doc, doc.id.asDocInfo))
+            var key = doc.id.asDocInfo
+            cacheInvalidate(key)
             val src = StreamConverters.fromInputStream(() => bytes)
             db.attach(doc, attachmentName, contentType, src)
         } match {
@@ -182,8 +185,8 @@ trait DocumentFactory[W] extends InMemoryCache[W] {
             require(doc != null, "doc undefined")
         } map { _ =>
             implicit val logger: Logging = db
-            // invalidate two keys: just the id, and id:rev
-            cacheInvalidate(Set(doc, doc.id.asDocInfo))
+            var key = doc.id.asDocInfo
+            cacheInvalidate(key)
             db.del(doc)
         } match {
             case Success(f) => f
@@ -214,8 +217,9 @@ trait DocumentFactory[W] extends InMemoryCache[W] {
             require(db != null, "db undefined")
         } map {
             implicit val logger = db: Logging
+            implicit val ec = db.executionContext
             val key = doc.asDocInfo(rev)
-            _ => cacheLookup(db, key, db.get[W](key), fromCache)
+            _ => cacheLookup(key, db.get[W](key), fromCache)
         } match {
             case Success(f) => f
             case Failure(t) => Future.failed(t)
--- a/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
@@ -120,7 +120,7 @@ trait DocumentSerializer {
  * but the get permits a datastore of its super type so that a single datastore client
  * may be used for multiple types (because the types are stored in the same database for example).
  */
-trait DocumentFactory[W] extends MultipleReadersSingleWriterCache/*InMemoryCache*/[W, DocInfo] {
+trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
     /**
      * Puts a record of type W in the datastore.
      *
--- a/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
@@ -199,7 +199,7 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
     }
 
     /**
-     * FIXME UPDATE Fetches a raw record of type R from the datastore by its id (and revision if given)
+     * Fetches a raw record of type R from the datastore by its id (and revision if given)
      * and converts it to Success(W) or Failure(Throwable) if there is an error fetching
      * the record or deserializing it.
      *
--- a/common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
+++ b/common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
@@ -169,8 +169,16 @@ trait MultipleReadersSingleWriterCache[W, Winfo] {
                                 invalidateEntryAfter(invalidator, key, actualEntry)
                             }
 
-                        case InvalidateInProgress | InvalidateWhenDone =>
-                            // someone else requested an invalidation already
+                        case InvalidateInProgress =>
+                            if (actualEntry.transid == transid) {
+                                // we own the entry, so we are responsible for cleaning it up
+                                invalidateEntryAfter(invalidator, key, actualEntry)
+                            } else {
+                                // someone else requested an invalidation already
+                                invalidator
+                            }
+                        case InvalidateWhenDone =>
+                            // a pre-existing owner will take care of the invalidation
                             invalidator
                     }
                 }
--- a/common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
+++ b/common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
@@ -1,0 +1,315 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.database
+
+import java.util.concurrent.atomic.AtomicReference
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scala.concurrent.Promise
+import scala.concurrent.duration.DurationInt
+import scala.language.implicitConversions
+import scala.util.Failure
+import scala.util.Success
+
+import spray.caching.Cache
+import spray.caching.LruCache
+import spray.caching.ValueMagnet.fromAny
+import whisk.common.Logging
+import whisk.common.LoggingMarkers
+import whisk.common.TransactionId
+
+/**
+ * A cache that allows multiple readers, bit only a single writer, at
+ * a time. It will make a best effort attempt to coalesce reads, but
+ * does not guarantee that all overlapping reads will be coalesced.
+ *
+ * The cache operates by bracketing all reads and writes. A read
+ * imposes a lightweight read lock, by inserting an entry into the
+ * cache with State.ReadInProgress. A write does the same, with
+ * State.WriteInProgress.
+ *
+ * On read or write completion, the value transitions to
+ * State.Cached.
+ *
+ * State.Initial is represented implicitly by absence from the cache.
+ *
+ * The handshake for cache entry state transition is:
+ * 1. if entry is in an agreeable state, proceed
+ *    where agreeable for reads is Initial, ReadInProgress, or Cached
+ *                    and the read proceeds by ensuring the entry is State.ReadInProgress
+ *      and agreeable for writes is Initial or Cached
+ *                    and the write proceeds by ensuring the entry is State.WriteInProgress
+ *
+ * 2. if entry is not in an agreeable state, fail fast with Future.failed(new ConcurrentUpdateException)
+ *
+ * 3. to swap in the new state to an existing entry, we use an AtomicReference.compareAndSet
+ *
+ * 4. after adding a new entry, we double check that the state of the actually-added entry
+ *    is agreeable with the desired action (as per item 1)
+ *
+ */
+trait MultipleReadersSingleWriterCache[W, Winfo] {
+  val DEBUG = true
+
+  /** Subclasses: Toggle this to enable/disable caching for your entity type. */
+  protected def cacheEnabled = true
+
+  /** Subclasses: tell me what key to use for updates */
+  protected def cacheKeyForUpdate(w: W): Any
+
+  /**
+   * Each entry has a state
+   *
+   */
+  protected object State extends Enumeration {
+    type State = Value
+    val ReadInProgress, WriteInProgress, InvalidateInProgress, Cached = Value
+  }
+  import State._;
+  implicit def state2Atomic(state: State): AtomicReference[State] = {
+    new AtomicReference(state);
+  }
+
+  /**
+   * Failure modes
+   *
+   */
+  private class ConcurrentUpdateException extends Exception {}
+
+  /**
+   * The entries in the cache will be a pair of State and the Future value
+   *
+   */
+  private case class Entry(transid: TransactionId, state: AtomicReference[State], value: Future[W]) {
+    def invalidate {
+      state.set(InvalidateInProgress)
+    }
+
+    def writeDone()(implicit logger: Logging): Boolean = {
+      if (DEBUG) logger.info(this, "Write finished")
+      trySet(WriteInProgress, Cached)
+    }
+
+    def readDone(value: W, promise: Promise[W])(implicit logger: Logging): Unit = {
+      logger.info(this, "Read finished")
+
+      if (trySet(ReadInProgress, Cached)) {
+        promise success value
+      } else {
+        promise failure new ConcurrentUpdateException
+      }
+    }
+
+    def trySet(expectedState: State, desiredState: State): Boolean = {
+      state.compareAndSet(expectedState, desiredState)
+    }
+  }
+
+  protected def cacheInvalidate(key: Any)(
+    implicit transid: TransactionId, logger: Logging): Unit = {
+    if (cacheEnabled) {
+      logger.info(this, s"invalidating $key")
+      cache remove key
+    }
+  }
+
+  protected def cacheLookup[Wsuper >: W](
+    key: Any,
+    future: => Future[W],
+    fromCache: Boolean = cacheEnabled)(
+      implicit ec: ExecutionContext, transid: TransactionId, logger: Logging): Future[W] = {
+
+    if (fromCache) {
+      val promise = Promise[W]
+      val desiredEntry = Entry(transid, ReadInProgress, promise.future)
+
+      //
+      // try inserting our desired entry...
+      //
+      cache(key)(desiredEntry) flatMap {
+        // ... and see what we get back
+        case actualEntry =>
+          // if (DEBUG) logger.info(this, s"Cache Lookup Status: the entry is currently in state ${actualEntry.state.get} ${getSize()}")
+
+          actualEntry.state.get match {
+            case Cached => {
+              if (DEBUG) logger.info(this, s"Cached read $key")
+              makeNoteOfCacheHit(key)
+              actualEntry.value
+            }
+
+            case ReadInProgress => {
+              if (actualEntry.transid == transid) {
+                if (DEBUG) logger.info(this, "Read initiated");
+                makeNoteOfCacheMiss(key)
+                listenForReadDone(key, actualEntry, future, promise)
+                actualEntry.value
+
+              } else {
+                if (DEBUG) logger.info(this, "Coalesced read")
+                makeNoteOfCacheHit(key)
+                actualEntry.value
+              }
+            }
+
+            case WriteInProgress | InvalidateInProgress => {
+              if (DEBUG) logger.info(this, "Reading around a write in progress")
+              makeNoteOfCacheMiss(key)
+              future
+            }
+          }
+      }
+
+    } else {
+      // not caching
+      future
+    }
+  }
+
+  protected def cacheUpdate(doc: W, key: Any, future: => Future[Winfo])(
+    implicit transid: TransactionId, logger: Logging, ec: ExecutionContext): Promise[Winfo] = {
+
+    val valPromise = Promise[W]
+    val promise = Promise[Winfo]
+
+    if (cacheEnabled) {
+      logger.info(this, s"caching $key")
+
+      val desiredEntry = Entry(transid, WriteInProgress, valPromise.future)
+
+      //
+      // try inserting our desired entry...
+      //
+      cache(key)(desiredEntry) map {
+        // ... and see what we get back
+        case actualEntry =>
+          if (transid == actualEntry.transid) {
+            //
+            // then we won the race, and now own the entry
+            //
+            if (DEBUG) logger.info(this, s"Write initiated $key")
+            listenForWriteDone(doc, key, actualEntry, future, valPromise, promise)
+
+          } else {
+            //
+            // then we either lost the race, or there is an operation in progress on this key
+            //
+            if (DEBUG) logger.info(this, s"Write under ${actualEntry.state.get} $key")
+            promise completeWith future
+          }
+      }
+
+    } else {
+      // not caching
+      promise completeWith future
+    }
+
+    promise
+  }
+
+  def getSize() = cache.size
+
+  /**
+   * Log a cache hit
+   *
+   */
+  private def makeNoteOfCacheHit(key: Any)(implicit transid: TransactionId, logger: Logging) {
+    transid.mark(this, LoggingMarkers.DATABASE_CACHE_HIT, s"[GET] serving from cache: $key")(logger)
+  }
+
+  /**
+   * Log a cache miss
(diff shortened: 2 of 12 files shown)


common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,315 @@
...
+ * 2. if entry is not in an agreeable state, fail fast with Future.failed(new ConcurrentUpdateException)
+ *
+ * 3. to swap in the new state to an existing entry, we use an AtomicReference.compareAndSet
+ *
+ * 4. after adding a new entry, we double check that the state of the actually-added entry
+ *    is agreeable with the desired action (as per item 1)
+ *
+ */
+trait MultipleReadersSingleWriterCache[W, Winfo] {
+  val DEBUG = true
```
markusthoemmes: Can this be replaced by using an appropriate loglevel instead? We have `logger.debug`.

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
```diff
@@ -0,0 +1,211 @@
...
+      }
+
+      latch.await(2, TimeUnit.MINUTES)
+      // System.out.println("DONE " + i);
+
+      inhibits
+    }
+  }
+
+  def delayed[W](v: W)(implicit delay: Int): Future[W] = {
```
markusthoemmes: Should rather be `akka.after` instead of a blocking sleep. See http://doc.akka.io/docs/akka/2.4.10/scala/futures.html#After

markusthoemmes:
Please install https://github.com/openwhisk/openwhisk/blob/master/tools/eclipse/scala.properties into your Eclipse or adjust your settings accordingly. Mostly 2 spaces vs. 4 spaces.

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
starpit: i think this would be fine. my only word of caution is the DEBUG stuff is in there really just for development, and is fairly liberally placed. is the scala compiler smart enough to excise logger.debug calls if the log level is set higher?

starpit:
I've made the cosmetic changes.

markusthoemmes:
Only made a basic pass. Will look more thoroughly soon.

tests/src/whisk/core/entity/test/MigrationEntities.scala
```diff
@@ -52,6 +52,7 @@ object OldWhiskRule
```
rabbah: we can nuke this test suite @markusthoemmes?

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+    import State._;
+    implicit def state2Atomic(state: State): AtomicReference[State] = {
+        new AtomicReference(state);
+    }
+
+    /**
+     * Failure modes
+     *
+     */
+    private class ConcurrentUpdateException extends Exception {}
```
markusthoemmes: As this is `private` I believe the resulting Exception type is not matchable outside this class? Make it public? Also providing a message might be helpful

`case class ConcurrentUpdateException(msg: String) extends Exception(msg)`

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+            state.set(InvalidateInProgress)
+        }
+
+        def writeDone()(implicit logger: Logging): Boolean = {
+            logger.debug(this, "Write finished")
+            trySet(WriteInProgress, Cached)
+        }
+
+        def readDone(value: W, promise: Promise[W])(implicit logger: Logging): Unit = {
+            logger.info(this, "Read finished")
```
markusthoemmes: Should be `debug`?

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+     *
+     */
+    private class ConcurrentUpdateException extends Exception {}
+
+    /**
+     * The entries in the cache will be a pair of State and the Future value
+     *
+     */
+    private case class Entry(transid: TransactionId, state: AtomicReference[State], value: Future[W]) {
+        def invalidate {
```
markusthoemmes: `def invalidate = state.set(InvalidateInProgress)`

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.database
+
+import java.util.concurrent.atomic.AtomicReference
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.ExecutionContext.Implicits.global
```
markusthoemmes: This introduces a new `ExecutionContext`, which we probably don't want here, as you're already threading through an ExecutionContext from the outside.  You can remove the line and thread that implicit EC through to the methods in need.

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+     *
+     */
+    protected object State extends Enumeration {
+        type State = Value
+        val ReadInProgress, WriteInProgress, InvalidateInProgress, Cached = Value
+    }
+    import State._;
+    implicit def state2Atomic(state: State): AtomicReference[State] = {
+        new AtomicReference(state);
+    }
```
markusthoemmes: I'ld remove the implicit conversion and rather state the types explicitly in the code. It's needed only in two places anyway and make the intent as clear imho.

If you choose to remove this, also remove the implicitConversions import.

Also `;`

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+
+    /**
+     * Each entry has a state
+     *
+     */
+    protected object State extends Enumeration {
+        type State = Value
+        val ReadInProgress, WriteInProgress, InvalidateInProgress, Cached = Value
+    }
+    import State._;
```
markusthoemmes: nuke the `;`

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+            // not caching
+            future
+        }
+    }
+
+    protected def cacheUpdate(doc: W, key: Any, future: => Future[Winfo])(
+        implicit transid: TransactionId, logger: Logging, ec: ExecutionContext): Promise[Winfo] = {
+
+        val valPromise = Promise[W]
+        val promise = Promise[Winfo]
```
markusthoemmes: I believe we do not need `val promise = Promise[Winfo]`. `listenForWriteDone` should return a `Future[Winfo]` which we could return directly instead. Removes the need of sideeffects here.

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+            }
+
+        } else {
+            // not caching
+            future
+        }
+    }
+
+    protected def cacheUpdate(doc: W, key: Any, future: => Future[Winfo])(
+        implicit transid: TransactionId, logger: Logging, ec: ExecutionContext): Promise[Winfo] = {
```
markusthoemmes: Why not return `Future[Winfo]`? Every caller just takes out `Promise.future` anyway as it seems.

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+        val promise = Promise[Winfo]
+
+        if (cacheEnabled) {
+            logger.info(this, s"caching $key")
+
+            val desiredEntry = Entry(transid, WriteInProgress, valPromise.future)
+
+            //
+            // try inserting our desired entry...
+            //
```
markusthoemmes: I'ld prefer those being single liners.

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+            cache(key)(desiredEntry) map {
+                // ... and see what we get back
+                case actualEntry =>
```
markusthoemmes: Doesn't need to be case, just use `actualEntry =>`

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+            case Failure(t)     => readOops(key, entry, promise, t)
+        }
+    }
+
+    /**
+     * We have initiated a write, now handle its completion
+     *
+     */
+    private def listenForWriteDone(doc: W, key: Any, entry: Entry, future: => Future[Winfo], valPromise: Promise[W], promise: Promise[Winfo])(
+        implicit logger: Logging): Unit = {
```
markusthoemmes: As noted above, this should really return `Future[Winfo]`, so error cases can be handled above and not need to thread multiple `Promise` instances through

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+        entry.invalidate
+        cache remove key
+        promise failure t
+    }
+
+    /**
+     * Write completion resulted in a failure
+     *
+     */
+    private def writeOops(key: Any, entry: Entry, valPromise: Promise[W], promise: Promise[Winfo], t: Throwable) {
```
markusthoemmes: When you do the `Future` transformation outlined above, this method becomes unnecessary and can be fold into `cacheUpdate` to further localize what actually happens.

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+        def trySet(expectedState: State, desiredState: State): Boolean = {
+            state.compareAndSet(expectedState, desiredState)
+        }
+    }
+
+    protected def cacheInvalidate(key: Any)(
+        implicit transid: TransactionId, logger: Logging): Unit = {
+        if (cacheEnabled) {
+            logger.info(this, s"invalidating $key")
+            cache remove key
```
markusthoemmes: No need to mark the entry to be `InvalidateInProgress` here?

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+        if (fromCache) {
+            val promise = Promise[W]
+            val desiredEntry = Entry(transid, ReadInProgress, promise.future)
+
+            //
+            // try inserting our desired entry...
+            //
+            cache(key)(desiredEntry) flatMap {
+                // ... and see what we get back
+                case actualEntry =>
```
markusthoemmes: No need for `case`

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+        def invalidate {
+            state.set(InvalidateInProgress)
+        }
+
+        def writeDone()(implicit logger: Logging): Boolean = {
+            logger.debug(this, "Write finished")
+            trySet(WriteInProgress, Cached)
+        }
+
+        def readDone(value: W, promise: Promise[W])(implicit logger: Logging): Unit = {
```
markusthoemmes: Like `writeDone` this should return a `Boolean` to be able to handle state further above and not needing to pass `Promise` around (mutability)

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+                            logger.debug(this, s"Cached read $key")
+                            makeNoteOfCacheHit(key)
+                            actualEntry.value
+                        }
+
+                        case ReadInProgress => {
+                            if (actualEntry.transid == transid) {
+                                logger.debug(this, "Read initiated");
+                                makeNoteOfCacheMiss(key)
+                                listenForReadDone(key, actualEntry, future, promise)
```
markusthoemmes: Basically a no op now, as you do nothing with the result of `promise`.

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+
+        } else {
+            // not caching
+            promise completeWith future
+        }
+
+        promise
+    }
+
+    def getSize() = cache.size
```
markusthoemmes: As per scala idiom: Use `def size = cache.size`. `()` indicate sideeffects, getters usually omit the `get`.

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+    /** Subclasses: Toggle this to enable/disable caching for your entity type. */
+    protected def cacheEnabled = true
+
+    /** Subclasses: tell me what key to use for updates */
+    protected def cacheKeyForUpdate(w: W): Any
+
+    /**
+     * Each entry has a state
+     *
+     */
```
starpit: this is already explained in the top comment. i'll amend this sub-comment to refer to the top

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
starpit: you'd really prefer to see
```
Entry(transid, new AtomicReference(ReadInProgress), promise.future)
```

to
```
Entry(transid, ReadInProgress, promise.future)
```

the latter seems much clearer to me, and doesn't escape the State implementation to the use points

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
starpit: not true, listenForReadDone will eventually mark the entry as Cached

markusthoemmes: You're right! Missed that case.

starpit:
PG1 552 passed with latest (2dd65c5)

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+            case Failure(t) => {
+                //
+                // oops, the datastore read failed. invalidate the cache entry
+                //
+                // note that this might be a perfectly legitimate failure,
+                // e.g. a lookup for a non-existant key; we need to pass the particular t through
+                //
+                readOops(key, entry, promise, t)
+            }
+        }
```
markusthoemmes: The logic here can be simplified:

```
future.flatMap { value =>
    if (entry.readDone()) {
        Future.successful(value)
    } else {
        Future.failed(ConcurrentOperationUnderRead(entry.state.get))
    }
} onFailure {
    case t =>
        entry.invalidate
        cache remove key
        promise failure t
}
```

With the simplification in place, note that i inlined `readOops` because it'ld be only called once anyway. Also we could then inline `listenForReadDone` above, putting the relevant logic in one place, preventing us from passing a `Promise` through different methods. What do you think?

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+                    logger.debug(this, s"Write all done $key ${entry.transid} ${entry.state.get}")
+                }
+            }
+            case Failure(t) => {
+                //
+                // oops, the datastore write failed. invalidate the cache entry
+                //
+                writeOops(key, entry)
+            }
+        }
```
markusthoemmes: Same as above, this can be simplified to reduce duplicated code:

```
future.flatMap { value =>
    logger.debug(this, "Write backend part done, now marking cache entry as done")
    if (entry.writeDone()) {
        logger.debug(this, s"Write all done $key ${entry.transid} ${entry.state.get}")
        Future.successful(value)
    } else {
        Future.failed(ConcurrentOperationUnderUpdate(entry.state.get))
    }
} onFailure {
    case _ =>
        entry.invalidate
        cache remove key
}
```

WriteOops can be inlined, and the whole method could be inlined aswell, keeping the relevant code local to `cacheUpdate`. Thoughts?

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+            logger.debug(this, "Write finished")
+            trySet(WriteInProgress, Cached)
+        }
+
+        def readDone()(implicit logger: Logging): Boolean = {
+            logger.debug(this, "Read finished")
+            trySet(ReadInProgress, Cached)
+        }
+
+        def trySet(expectedState: State, desiredState: State): Boolean = {
```
markusthoemmes: Is this method worthwile? Would it make sense to just call `state.compareAndSet` above?

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+                    // note that we don't care what that other operation is doing; unlike for lookups,
+                    // we don't need to do anything special (e.g. coalescing, as in the case of lookups)
+                    //
+                    logger.debug(this, s"Write under ${actualEntry.state.get} $key")
+                }
+            }
+
+        } else {
+            // not caching
+        }
```
markusthoemmes: Remove the else case?

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+            //
+            // try inserting our desired entry...
+            //
+            cache(key)(desiredEntry) map { actualEntry =>
+                // ... and see what we get back
+
+                if (transid == actualEntry.transid) {
+                    //
+                    // then we won the race, and now own the entry
+                    //
```
markusthoemmes: Would you remove the "empty" comment lines? I feel like they're stretching the code, fitting less on the screen while reading through.

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,313 @@
...
+                        if (actualEntry.transid == transid) {
+                            logger.debug(this, "Read initiated");
+                            makeNoteOfCacheMiss(key)
+                            listenForReadDone(key, actualEntry, future, promise)
+                            actualEntry.value
+
+                        } else {
+                            logger.debug(this, "Coalesced read")
+                            makeNoteOfCacheHit(key)
+                            actualEntry.value
```
markusthoemmes: `actualEntry.value` can be pulled out of the `if ... else ...`, reducing duplication.

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
```diff
@@ -0,0 +1,215 @@
...
+
+        System.out.println();
+        System.out.println(testName);
+
+        val exec = Executors.newFixedThreadPool(nThreads)
+        val inhibits = Inhibits()
+
+        for (i <- 1 to nThreads) {
+            exec.submit(new Runnable { def run() = { doReadWriteRead(key(i), inhibits, readsFirst).go(delay) } })
+        }
```
markusthoemmes: Do we need to manage our own threadpool/concurrency? Can scala handle that for us?

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
```diff
@@ -0,0 +1,215 @@
...
+                }
+            }
+
+            latch.await(2, TimeUnit.MINUTES)
+
+            inhibits
+        }
+    }
+
+    private def delayed[W](v: W)(implicit delay: Int): Future[W] = {
```
markusthoemmes: Does a `Duration` instead of `Int` make more sense in this context?

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
```diff
@@ -0,0 +1,215 @@
...
+    case class Inhibits(
+        nReadInhibits: AtomicInteger = new AtomicInteger(0),
+        nWriteInhibits: AtomicInteger = new AtomicInteger(0)) {
+
+        def debug(from: AnyRef) {
+            logger.debug(from, "InhibitedReads: " + nReadInhibits);
+            logger.debug(from, "InhibitedWrites: " + nWriteInhibits);
+        }
+
+        def someInhibits: Boolean = { nReadInhibits.get > 0 || nWriteInhibits.get > 0 }
```
markusthoemmes: Does `hasInhibits` make more sense as a method name?

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
```diff
@@ -0,0 +1,215 @@
...
+
+        inhibits.debug(this)
+        inhibits
+    }
+
+    case class Inhibits(
+        nReadInhibits: AtomicInteger = new AtomicInteger(0),
+        nWriteInhibits: AtomicInteger = new AtomicInteger(0)) {
+
+        def debug(from: AnyRef) {
```
markusthoemmes: missing `=`

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
```diff
@@ -0,0 +1,215 @@
...
+        }
+    }
+
+    "the cache" should "support concurrent CRUD to shared keys (zero latency)" in {
+        var someInhibits = false
+        for (i <- 1 to nIters) {
+            someInhibits = doCRUD("concurrent CRUD to shared keys (zero latency)", sharedKeys, 0)
+                .someInhibits
+        }
+        someInhibits should not be (false)
```
markusthoemmes: What if the second `doCrud` returns false, but the third returns `true`? Then this test will falsely assert that everything is fine?

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
```diff
@@ -0,0 +1,215 @@
...
+
+import akka.actor.ActorSystem
+import whisk.common.Logging
+import whisk.common.TransactionId
+import whisk.core.database.MultipleReadersSingleWriterCache
+
+class MultipleReadersSingleWriterCacheTests(nIters: Int = 3) extends FlatSpec
+    with Matchers
+    with MultipleReadersSingleWriterCache[String, String]
+    with Logging {
```
markusthoemmes: Do we need `Logging` here? In tests we used to use just `println`

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
```diff
@@ -0,0 +1,215 @@
...
+
+    private def delayed[W](v: W)(implicit delay: Int): Future[W] = {
+        akka.pattern.after(duration = delay.millis, using = actorSystem.scheduler)(
+            Future.successful { v })
+    }
+
+    /** we are using cache keys, so the update key is just the string itself */
+    override protected def cacheKeyForUpdate(w: String): String = (w)
+
+    implicit private val actorSystem: ActorSystem = ActorSystem("MRSW-cache-test-actor-system")
```
markusthoemmes: Import the WskActorSystem possibly?

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
```diff
@@ -0,0 +1,215 @@
...
+    implicit private val logger = this
```
markusthoemmes: see above

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
```diff
@@ -0,0 +1,215 @@
...
+                case Failure(t) => {
+                    inhibits.nReadInhibits.incrementAndGet();
+                    latch.countDown();
+                }
+            }
+
+            latch.await(2, TimeUnit.MINUTES)
+
+            inhibits
+        }
```
markusthoemmes: Does this testcase assume, that the calls are running in the order they are put? It reads a bit like that.

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
starpit: the cache itself needs an implicit logger. how do you usually deal with that situation?

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
starpit: to me, this helps to preserve the understandability of the code; it also maintains a parallel structure between the lookups and updates.

i realize that we're getting into style, but this is something i feel the rest of the code base sorely lacks. it is too terse, and with very few comments explaining the flow. terseness is fine (and helpful), to a point. but, taken to far, and you produce code that only the core team understands.

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
starpit: but readOops is called twice.

and how is it valid to change `promise success value` with `Future.successful(value)`?

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
starpit: overall, i think this test suite is of low value. i implemented it initially just to help me catch glaring bugs. the reality is that, the entire point of this cache is to catch race conditions. it is impossible to generate an interleaving that, in a fixed amount of time, is guaranteed to trigger (or not trigger) a race. 

so we need to decide what to do with this test. i would suggest that we leave it disabled for normal runs, and only use it when further developing (if we ever do) the cache.

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
starpit: will fix

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
starpit: will fix

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
starpit: sure, will fix

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
starpit: no, the point of the test kernel is to do a read, a write, and a read, but in a random order.

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
starpit: will do

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
rabbah: valid criticism about the rest of the code base - we can do better!

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
markusthoemmes: With the rewritten logic, its only called once. You're right on the Promise though.

```
future.flatMap { value =>
    if (entry.readDone()) {
        Future.successful(value)
    } else {
        Future.failed(ConcurrentOperationUnderRead(entry.state.get))
    }
} onComplete {
    case Success(value) => promise success value
    case Failure(t) =>
        entry.invalidate
        cache remove key
        promise failure t
}
```

starpit: readOops does more than hook the promise, it also removes the entry from the cache. i don't see that in your code. i'm confused as to what problem you're fixing?

rabbah:
I reviewed with @starpit some gaps in the implementation (around "delete") and we also uncovered some corner cases with updates under reads that require revision.

rabbah:
I discussed the latest revision with Nick. There is still a race between delete and get. To address it the plan is to replicate the logic for update so that handling delete is the same (locks the transaction to invalidate and bypass cache on get).

tests/dat/actions/CacheConcurrencyTests.sh
```diff
@@ -0,0 +1,94 @@
...
+echo -n "get after update+get "
+for i in `seq 1 $N`; do
+    $WSK action get testy$i | grep zoomba > /dev/null
+    if [ $? == 0 ]; then
+	echo "FAIL"
+	exit 1
+    fi
+done
+echo "PASS"
+
```
starpit: partly because i wanted to be able to debug the test against YP and localhost (verifying failure on YP, and success on localhost), and partly because doing so required about 12 debug iterations to find the spot that'd give me a valid test -- i didn't have the patience to wait 12*30 minutes that it would take to do this with `tests:test`

markusthoemmes: You can run single tests with either `gradle` and or your IDE.

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,408 @@
...
+    case class SquashedInvalidation(actualState: State) extends Exception(s"Cache invalidation squashed due $actualState.")
+    case class StaleRead(actualState: State) extends Exception(s"Attempted read of invalid entry due to $actualState.")
+}
+
+trait MultipleReadersSingleWriterCache[W, Winfo] {
+    import MultipleReadersSingleWriterCache._
+    import MultipleReadersSingleWriterCache.State._
+
+    /** Subclasses: Toggle this to enable/disable caching for your entity type. */
+    protected lazy val cacheEnabled = true
```
psuter: Why `lazy`?

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
```diff
@@ -0,0 +1,217 @@
...
+
+class MultipleReadersSingleWriterCacheTests(nIters: Int = 3) extends FlatSpec
+    with Matchers
+    with MultipleReadersSingleWriterCache[String, String]
+    with WskActorSystem
+    with Logging {
+
+    "the cache" should "support simple CRUD" in {
+        System.out.println();
+        System.out.println("simple CRUD");
```
psuter: No need to add to our logs

tests/src/whisk/core/database/test/MultipleReadersSingleWriterCacheTests.scala
psuter: I imagine it's to guarantee there is the desired concurrency, which is important to functional correctness here?

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
rabbah: it's used as a default argument in DocumentFactory.
`    def get[Wsuper >: W](db: ArtifactStore[Wsuper], doc: DocId, rev: DocRevision = DocRevision(), fromCache: Boolean = cacheEnabled)`

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,408 @@
...
+    private implicit def toAtomic(state: State): AtomicReference[State] = new AtomicReference(state)
+    private implicit def toAtomic(value: Future[W]): AtomicReference[Future[W]] = new AtomicReference(value)
+    private implicit def toAtomic(transid: TransactionId): AtomicReference[TransactionId] = new AtomicReference(transid)
+
+    /**
+     * The entries in the cache will be a triple of (transid, State, Future[W]?).
+     *
+     * We need the transid in order to detect whether we have won the race to add an entry to the cache.
+     */
+    private class Entry(
```
rabbah: i think we were a bit lax here - the grab writer lock should've been synchronized. i removed the atomic refs in exchange for @volatile rather than synchronizing other methods as well.

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,408 @@
...
+        def grabInvalidationLock() = state.set(InvalidateInProgress)
+    }
+
+    /**
+     * This method posts a delete to the backing store, and either directly invalidates the cache entry
+     * or informs any outstanding transaction that it must invalidate the cache on completion.
+     */
+    protected def cacheInvalidate[R](key: Any, invalidator: => Future[R])(
+        implicit ec: ExecutionContext, transid: TransactionId, logger: Logging): Future[R] = {
+        cacheEnabled match {
```
rabbah: ok i'll change it back; i thought this was clearer here ...

common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
```diff
@@ -214,8 +221,9 @@ trait DocumentFactory[W] extends InMemoryCache[W] {
             require(db != null, "db undefined")
         } map {
             implicit val logger = db: Logging
+            implicit val ec = db.executionContext
```
rabbah: removed - the comment looks correct.

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -0,0 +1,408 @@
...
+                                // then the pre-existing owner will take care of the invalidation
+                                invalidator
+                            } else {
+                                // the pre-existing reader or writer finished and so must
+                                // explicitly invalidate here
+                                invalidateEntryAfter(invalidator, key, actualEntry)
+                            }
+
+                        case InvalidateInProgress =>
+                            if (actualEntry.transid == transid) {
```
rabbah: it will be a bug if they're not propagated correctly through the system.
your observation is right.
if we don't use this transid, we'll have to create or use another id (object ref/hashcode?).

rabbah:
new pg2/500.

rabbah:
@psuter addressed your comments and re'pged. On you to merge when satisfied.

psuter:
Merged as 22a12c9687119d3582b667c9c9bbdc0915134277.

Issue #1093, opened 2016-08-19
WhiskConfig is not typesafe


rabbah:
All properties are assumed to be of type String and there are many `toInt` conversions that can fail if there is a typo. The validation of the configuration should also check the expected value types. This is pervasive to all uses of WhiskConfig.

rabbah:
@markusthoemmes has introduce pureconfig and we're slowly migrating to it. Re-titling accordingly.

Issue #239, opened 2016-04-20
confusing error message on trying to create rule with shared action


csantanapr:
Trying this in the CLI

```
$ wsk rule create ifGitThenEcho myGitTrigger /whisk.system/samples/echo
error: The request content was malformed:
entity name malformed (code 1385580)
```

In `wsk rule create name trigger action` trigger or action can't be private
User needs to copy the action to his space like

```
$ wsk action create echo --copy /whisk.system/samples/echo
ok: created action echo
```

CLI error should be more clear about this and the [docs](https://github.com/openwhisk/openwhisk/blob/master/docs/triggers_rules.md#using-rules-to-associate-triggers-and-actions) should be updated to reflect this

rabbah:
The trigger and action that make a rule must be in the default package, which is in fact *private*. The docs should state this, you are right, and the API/CLI should report a better error message.

We plan to relax the noted constraint so that you can create a rule using an action in a package.

rabbah:
@markusthoemmes this will need to be addressed as part of the activator work.

lzbj:
@rabbah , is the default package means the system package or the package under the user namespace? I searched under the doc seems not mentioned this.
After the relax the constraint, is that user can create rule no matter the package or action is private or not?

kurtismgit:
I'm a new OpenWhisk user (1 week or so) and hit this pretty quickly.  As you know the error wasn't helpful.  Found myself searching through github issues to figure out about this limitation.  Definitely have my vote for fixing it.

sjfink:
not sure why this was closed -- still can't have package actions in rules AFAIK.

csantanapr:
don't think is fix
```
$ wsk rule create r1 t1 /whisk.system/samples/echo
error: Unable to create rule 'r1': Request failure: The request content was malformed:
entity name malformed (code 35142)
```

csantanapr:
current workarounds are 
1. put the action inside a sequence
2. copy the action to the namespace outside of a package
example for 1. :
```
$ wsk action create --sequence s1 /whisk.system/samples/echo,/whisk.system/samples/echo
ok: created action s1
$ wsk rule create r1 t1 s1
ok: created rule r1
```

sjfink:
I was trying for this in my free time, but now I'm scared because I think it's a schema change -- it appears we'll have to store QualfiedNames in Rule entities, and I'm scared to do schema changes by myself.

Assigning to @rabbah since he's mentioned this was "easy".

[...]

PR #1303, opened 2016-09-21
Check static limits for sequences

ioana-blue:
wip: check limits when creating/updating sequences. check for recursion as well.


ioana-blue:
I think this is ready for review. 

`596 tests completed, 1 failed, 19 skipped` - swift action timed-out, unrelated to these changes. 

Once I address your comments, I'll squash the second commit.

ansible/templates/whisk.properties.j2
```diff
@@ -17,7 +17,7 @@ whisk.ssl.challenge=openwhisk
 defaultLimits.actions.invokes.perMinute={{ defaultLimits.actions.invokes.perMinute }}
 defaultLimits.actions.invokes.concurrent={{ defaultLimits.actions.invokes.concurrent }}
 defaultLimits.triggers.fires.perMinute={{ defaultLimits.triggers.fires.perMinute }}
-
+limits.actions.sequence.max={{ limits.actions.sequence.max }}
```
rabbah: maxLength. Is this the max length of primitive actions (not sequence actions)?

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -106,10 +106,10 @@ class WhiskConfig(
     val kafkaDockerEndpoint = this(WhiskConfig.kafkaDockerEndpoint)
     val mainDockerEndpoint = this(WhiskConfig.mainDockerEndpoint)
 
-    val actionInvokePerMinuteLimit = this(WhiskConfig.actionInvokePerMinuteDefaultLimit, WhiskConfig.actionInvokePerMinuteLimit)
```
rabbah: bad merge

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -264,9 +264,9 @@ object WhiskConfig extends Logging {
     val actionInvokePerMinuteDefaultLimit = "defaultLimits.actions.invokes.perMinute"
     val actionInvokeConcurrentDefaultLimit = "defaultLimits.actions.invokes.concurrent"
     val triggerFirePerMinuteDefaultLimit = "defaultLimits.triggers.fires.perMinute"
-
+    val actionSequenceLimit = "limits.actions.sequence.max"
     val actionInvokePerMinuteLimit = "limits.actions.invokes.perMinute"
     val actionInvokeConcurrentLimit = "limits.actions.invokes.concurrent"
     val triggerFirePerMinuteLimit = "limits.triggers.fires.perMinute"
-
+    val actionSequenceMaxLimit = Map(actionSequenceLimit -> null)
```
rabbah: we don't do this for other props; is it necessary?

common/scala/src/main/scala/whisk/core/entity/EntityPath.scala
```diff
@@ -45,9 +45,24 @@ protected[core] class EntityPath private (val path: Seq[String]) extends AnyVal
...
+    def defaultPackage = path.size == 1   // if only one element in the path, then it's the namespace with a default package
     def toJson = JsString(namespace)
     def apply() = namespace
     override def toString = namespace
+
+    /**
+     * helper method to replace the default namespace with a given namespace
+     * no effect if the namespace is not the default one
+     */
+    def replaceDefaultNamespace(newNamespace: String): EntityPath ={
```
rabbah: i'd call this resolveDefaultNamespace instead.

rabbah: should type of newNamespace be EntityName instead of String?

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -93,7 +93,10 @@ protected[core] case class BlackBoxExec(image: String) extends Exec(Exec.BLACKBO
     override val sentinelledLogs = false
 }
 
-protected[core] case class SequenceExec(code: String, components: Vector[String]) extends Exec(Exec.SEQUENCE) {
+/**
+ * add temporary field that holds the "fixed" names for components where the '_' is replaced by the user's namespace
+ */
+protected[core] case class SequenceExec(code: String, components: Vector[String], fixedComponents: Option[Vector[String]] = None) extends Exec(Exec.SEQUENCE) {
```
rabbah: why do we need components and fixedComponents? should only store the resolved components.

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -238,7 +240,26 @@ object WhiskAction
                     Future.successful(action)
             }
         }
+    }
 
+    /** utility function that given a fully qualified name for an action, resolve its possible package bindings and returns
+     *  the fully qualified name of the resolved action
+     */
+    def resolveAction(entityStore: EntityStore, fullyQualifiedName: String)(
```
rabbah: Does this belong here as opposed to the controller?

common/scala/src/main/scala/whisk/core/entity/WhiskPackage.scala
```diff
@@ -140,6 +148,31 @@ object WhiskPackage
...
+        }
+    }
+
+    /**
+     * utility function that extracts the package information from a fully qualified name,
+     * which contains a package and returns the corresponding docid
+     * @param fullyQualifiedName the fully qualified name of the entity
+     * @return the docid for the corresponding package
+     */
+    def packageDocId(fullyQualifiedName: String): DocId = {
```
rabbah: leaking the fully qualified names as strings to these entities feels wrong; you don't know if they'll match requirements which means these could fail and throw illegal argument exceptions.

instead the types should be more specific at this point here and resolve them to the narrower types elsewhere; otherwise, handle the non-conformance.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -210,11 +214,11 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
      * - 409 Conflict
      * - 500 Internal Server Error
      */
-    override def create(namespace: EntityPath, name: EntityName)(implicit transid: TransactionId) = {
+    override def create(user: Identity, namespace: EntityPath, name: EntityName)(implicit transid: TransactionId) = {
```
rabbah: is it redundant to pass user + namespace?

ansible/templates/whisk.properties.j2
ioana-blue: correct. this is the max (atomic) actions allowed in a sequence.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
ioana-blue: good catch. don't understand why no test caught this.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
ioana-blue: I need to add this prop as a required prop. I thought that's how we did it. If not, please advise.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
ioana-blue: Eventually I'll store only the resolved ones. I don't want to touch Perry's code to make sure it still works as designed. On a second thought, you mentioned that it may not make a difference in this case. I'll try to change it to see if Perry's code still works.

common/scala/src/main/scala/whisk/core/entity/WhiskPackage.scala
ioana-blue: I agree. I may need to add a new class that has two parts: an EntityPath and an EntityName. this is similar to the one that SJF added, but without the version. I'll look into it.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
ioana-blue: Yep, I think it is. Fixing.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
ioana-blue: I just quickly check by storing fixedComponents in components that Perry's code fails. Since at some point I'll remove Perry's code, I'll leave this as is and open an issue to keep track of cleanup stuff.

common/scala/src/main/scala/whisk/core/entity/WhiskPackage.scala
rabbah: why not make the semver optional then in the existing type?

common/scala/src/main/scala/whisk/core/entity/Exec.scala
rabbah: how about an intermediate step where fully resolved names works with "pipe.js"? 
it should NOT fail in principle.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
rabbah: Where you require the property, add the code above. It doesn't belong in whisk config IMO.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
rabbah: you just ran with higher limits.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
ioana-blue: No, it's not since the user.namespace  can be different than the namespace in this call. In particular, I think in this case `namespace` is overloaded since it contains the possible package name as well. This is a problematic use of `namespace` throughout our code that we're all aware of, but too costly to fix now...

ioana-blue: Too bad that I figured this out only after I made the change. reversing now.

ioana-blue:
Beautiful Green Travis - @rabbah ready for a second round of comments. I think I fixed everything from the first round. Once I fix everything, I'll squash in one commit.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
ioana-blue: Not sure I understand this comment. We do this for a bunch of other properties in the same class.  Like this:

```
val edgeHost = Map(edgeHostName -> null, edgeHostApiPort -> null)
    val consulServer = Map(consulServerHost -> null, consulPort -> null)
    val invokerHosts = Map(invokerHostsList -> null)
    val kafkaHost = Map(kafkaHostName -> null, kafkaHostPort -> null)
    val controllerHost = Map(controllerHostName -> null, controllerHostPort -> null)
    val loadbalancerHost = Map(loadbalancerHostName -> null, loadbalancerHostPort -> null)
```

rabbah:
I've generally done for pairs of properties for convenience. For a singleton, we have generally referenced it by name in the map for the required properties.

ioana-blue:
ok, I'll move it then.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -501,7 +501,7 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
         val args = { env map { _ ++ action.parameters } getOrElse action.parameters } merge payload
         val message = Message(
             transid,
-            FullyQualifiedEntityName(action.namespace, action.name, action.version),
+            FullyQualifiedEntityName(action.namespace, action.name, Some(action.version)),
```
rabbah: Since FullyQualifiedEntityName has a singleton, you could have added a second apply method which accepts SemVer and boxes it into Some instead.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -118,7 +118,7 @@ class Invoker(
                 }
 
                 info(this, s"activation failed")
-                completeTransactionWithError(actionid.id, msg.action.version, tran, failure.activationResponse)
+                completeTransactionWithError(actionid.id, msg.action.version.get, tran, failure.activationResponse)
```
rabbah: Let's make sure this is defined - add a require version is defined at entry to onMessage.
Squash related changes to into this commit (if possible).

common/scala/src/main/scala/whisk/core/entity/EntityPath.scala
```diff
@@ -55,14 +55,20 @@ protected[core] class EntityPath private (val path: Seq[String]) extends AnyVal
...
      * no effect if the namespace is not the default one
      */
-    def resolveNamespace(newNamespace: String): EntityPath ={
+    def resolveNamespace(newNamespace: EntityName): EntityPath ={
         // check if namespace is default
         if (root == EntityPath.DEFAULT) {
-            val newPath = path.updated(0, newNamespace)
+            val newPath = path.updated(0, newNamespace.name)
             EntityPath(newPath)
         } else
```
rabbah: curlies please around else.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -96,7 +96,7 @@ protected[core] case class BlackBoxExec(image: String) extends Exec(Exec.BLACKBO
 /**
  * add temporary field that holds the "fixed" names for components where the '_' is replaced by the user's namespace
  */
-protected[core] case class SequenceExec(code: String, components: Vector[String], fixedComponents: Option[Vector[String]] = None) extends Exec(Exec.SEQUENCE) {
+protected[core] case class SequenceExec(code: String, components: Vector[String], fixedComponents: Option[Vector[FullyQualifiedEntityName]] = None) extends Exec(Exec.SEQUENCE) {
```
rabbah: Can you explain to me again why we need fixedComponents and components? When you fully resolve the names, I don't see why the existing implementation can't accept those instead. I would view this as a good checkpoint/commit point that should be semantic preserving.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -676,11 +676,11 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
      * @param sequence the sequence to check
      * @param components the components of the sequence
      */
-    private def checkSequenceActionLimits(sequence: String, components: Vector[String]) (
+    private def checkSequenceActionLimits(sequence: FullyQualifiedEntityName, components: Vector[FullyQualifiedEntityName]) (
         implicit transid: TransactionId): Future[Boolean] = {
         // traverse all actions and "inline" all actions that are sequence
         // keep track of all sequences to detect recursion
-        info(this, "Checking limits and recursion for actions in a sequence")
+        info(this, s"Checking limits and recursion for actions in a sequence $sequence $components")
```
rabbah: do we need to print these in the log? they can make for a long log line.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -702,19 +702,19 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
      * @param sequences the sequences encountered during inlining
      * @return a successful future in case the conditions were met or a failed fugure with either too many actions or recursion found
      */
-    private def inlineComponentsAndCountAtomicActions(atomicActionsCnt: Int, actionsToInline: List[String], sequences: List[String])(
+    private def inlineComponentsAndCountAtomicActions(atomicActionsCnt: Int, actionsToInline: List[FullyQualifiedEntityName], sequences: List[FullyQualifiedEntityName])(
         implicit transid: TransactionId): Future[Boolean] = {
         if (atomicActionsCnt > whiskConfig.actionSequenceLimit.toInt)
```
rabbah: do you make sure somewhere at init time that actionSequenceLimit.toInt doesn't throw an exception? WhiskConfig is not typesafe (#1093).

common/scala/src/main/scala/whisk/core/entity/WhiskPackage.scala
```diff
@@ -168,8 +168,8 @@ object WhiskPackage
      * @param fullyQualifiedName the fully qualified name of the entity
      * @return the docid for the corresponding package
      */
-    def packageDocId(fullyQualifiedName: String): DocId = {
-        val wskPackage = EntityPath(fullyQualifiedName).dropLast
+    def packageDocId(fullyQualifiedName: FullyQualifiedEntityName): DocId = {
+        val wskPackage = fullyQualifiedName.path
         DocId(wskPackage.toString)
```
rabbah: can we avoid use to toString here? How about adding a toDocId method on EntityPath?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -702,19 +702,19 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
             Future.failed(new TooManyActionsInSequence())
         else {
             actionsToInline match {
                 case action :: restActions =>
-                    val docid = DocId(action)
+                    val docid = DocId(action.toString)
```
rabbah: Not entirely sure (without digging deeper) what the type of action is here but I think it's better to avoid toString and instead add a toDocId method on action.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -89,7 +89,7 @@ object WhiskActionsApi {
     def requiredProperties = WhiskServices.requiredProperties ++
         WhiskEntityStore.requiredProperties ++
         WhiskActivationStore.requiredProperties ++
-        WhiskConfig.actionSequenceMaxLimit
+        Map(WhiskConfig.actionSequenceLimit -> null)
```
rabbah: See my comment in another commit about checking the type of this value - if it's not of type int, we want to fail fast at startup.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -214,11 +214,11 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
      * - 409 Conflict
      * - 500 Internal Server Error
      */
-    override def create(user: Identity, name: EntityName)(implicit transid: TransactionId) = {
```
rabbah: tip: since this commit reverts the previous, for future reference and to help in reviewing, you can squash it into the previous commit (or cherry pick and drop the previous commit as needed).

rabbah:
Adding @markusthoemmes as a reviewer - these changes are relevant for #239.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
ioana-blue: I don't think they are really needed in the implementation. However, some tests that Perry added for the current implementation verify the names. I could try to modify those tests such that they pass with this change.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
ioana-blue: I don't think so as I didn't see any such examples for other int properties.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
rabbah: I think that's worthwhile.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
rabbah: EntitlementService - will fail during construction - indirectly. I think we list this value to a private val in the Actions API handler so that it too would fail in the same way. The real fix is more involved.

common/scala/src/main/scala/whisk/core/entity/EntityPath.scala
```diff
@@ -45,9 +45,30 @@ protected[core] class EntityPath private (val path: Seq[String]) extends AnyVal
...
+    def resolveNamespace(newNamespace: EntityName): EntityPath ={
+        // check if namespace is default
+        if (root == EntityPath.DEFAULT) {
+            val newPath = path.updated(0, newNamespace.name)
+            EntityPath(newPath)
+        } else
+            this
+    }
+
+    def fullyQualifiedEntityName = {
```
markusthoemmes: Should this be called `toFullyQualifiedEntityName` to make the intent clear?

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -238,7 +240,25 @@ object WhiskAction
                     Future.successful(action)
             }
         }
+    }
 
+    /** utility function that given a fully qualified name for an action, resolve its possible package bindings and returns
+     *  the fully qualified name of the resolved action
+     */
```
markusthoemmes: Please adjust this to conform to the other ScalaDoc comments.

common/scala/src/main/scala/whisk/core/entity/EntityPath.scala
```diff
@@ -45,9 +45,30 @@ protected[core] class EntityPath private (val path: Seq[String]) extends AnyVal
     def addpath(e: EntityName) = EntityPath(path :+ e.name)
     def root = EntityPath(Seq(path(0)))
     def last = EntityName(path.last)
+    def dropLast = EntityPath(path.dropRight(1)) // remove last element from path (e.g., name)
+    def defaultPackage = path.size == 1   // if only one element in the path, then it's the namespace with a default package
```
markusthoemmes: I believe this is dangerous or misleading: We currently abuse this type everywhere to express FQN's instead of more typesafe ones. `myPackage/test` will match this as well while not being the default package.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -336,40 +340,103 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+            limits,
+            content.version getOrElse SemVer(),
+            content.publish getOrElse false,
+            content.annotations getOrElse Parameters())
+    }
+
     /** Updates a WhiskAction from PUT content, merging old action where necessary. */
-    private def update(content: WhiskActionPut)(action: WhiskAction) = Future successful {
+    private def update(user: Identity, content: WhiskActionPut)(action: WhiskAction)(implicit transid: TransactionId) = {
+        if (content.exec.isDefined) {
```
markusthoemmes: Would it make sense to use the more idiomatic `content.exec.map { // do something when exec is Some }.getOrElse { // do something when exec is None }`

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -604,6 +671,77 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+     * @param components the components of the sequence
+     */
+    private def checkSequenceActionLimits(sequence: FullyQualifiedEntityName, components: Vector[FullyQualifiedEntityName]) (
+        implicit transid: TransactionId): Future[Boolean] = {
+        // traverse all actions and "inline" all actions that are sequence
+        // keep track of all sequences to detect recursion
+        info(this, s"Checking limits and recursion for actions in a sequence $sequence $components")
+        // first check that out of the box no more components than necessary
+        if (components.size > whiskConfig.actionSequenceLimit.toInt)
+            Future.failed(new TooManyActionsInSequence())
```
markusthoemmes: What is this supposed to do, don't think this exits early.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -604,6 +671,77 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+                    WhiskAction.get(entityStore, docid) flatMap { act =>
+                        act.exec match {
+                            case seq: SequenceExec =>
+                                // this is actually a sequence, check already traversed sequences
+                                if (sequences.contains(FullyQualifiedEntityName(act.namespace, act.name))) {
+                                   Future.failed(new SequenceWithRecursion())
+                                } else {
+                                    // resolve the components first
+                                    // need to inline each of its components
+                                    val resolvedComponents = seq.fixedComponents.get map { c => WhiskAction.resolveAction(entityStore, c) }
```
markusthoemmes: Is it guarenteed that this doesn't throw an exception? In general `Option.get` should never be used.

ioana-blue:
Thanks for the feedback. 

The first thing I'll try to do is get rid of the components field and use instead the new class we have so we always store FQEN for components. I think this is the way to go, we shouldn't have used Strings to begin with. This involves messing around with @perryibm 's code (for sequences with pipe.js), I hope I'll iron out the differences so I can still have both versions of the code in place.

common/scala/src/main/scala/whisk/core/entity/EntityPath.scala
ioana-blue: @markusthoemmes I completely agree we abuse this type, and in fact, at some point we do need to do a pass and remove any abusive statement since we introduced FQENs. But I'm stuck, I need to figure out whether something needs to resolve a binding or not (i.e., if it's in the default package or not). One option could be to push this code to the only place where it's used to not proliferate this check. But other than that, I don't have a better idea. Suggestion?

ioana-blue:
@rabbah @markusthoemmes I still have a couple of your comments to address + fixing a check that it's missing right now (+ test for it). Let me know if once I'm done, you'd like me to  squash or it's better to progressively see the changes for now.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
ioana-blue: If there are too many components, it should fail the future with that error

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -93,9 +93,12 @@ protected[core] case class BlackBoxExec(image: String) extends Exec(Exec.BLACKBO
...
 }
 
-protected[core] case class SequenceExec(code: String, components: Vector[String]) extends Exec(Exec.SEQUENCE) {
+/**
+ * add temporary field that holds the "fixed" names for components where the '_' is replaced by the user's namespace
+ */
+protected[core] case class SequenceExec(code: String, components: Vector[FullyQualifiedEntityName]) extends Exec(Exec.SEQUENCE) {
     val image = Exec.imagename(Exec.NODEJS)
-    def size = components.map(_ sizeInBytes).reduce(_ + _)
+    def size = ByteSize.fromString("0B") // not used for the hacky implementation and not used for the new implementation either
```
markusthoemmes: Not following the comment? If this is supposed to stay in, use `0.bytes` instead.

common/scala/src/main/scala/whisk/core/entity/WhiskPackage.scala
```diff
@@ -140,6 +148,31 @@ object WhiskPackage
...
+        }
+    }
+
+    /**
+     * utility function that extracts the package information from a fully qualified name,
+     * which contains a package and returns the corresponding docid
+     * @param fullyQualifiedName the fully qualified name of the entity
+     * @return the docid for the corresponding package
+     */
+    def packageDocId(fullyQualifiedName: FullyQualifiedEntityName): DocId = {
```
markusthoemmes: Seems like the method doesn't relate to what the comment says it does? It's essentially `EntityPath -> DocId`?

common/scala/src/main/scala/whisk/core/entity/WhiskPackage.scala
```diff
@@ -140,6 +148,31 @@ object WhiskPackage
...
+     * utility function that traverses all potential binding references and returns the resolved package
+     * @param pkg the package for which the resolution to be applied
+     * @return the same package if there is no binding, or the fully resolved package
+     */
+    def resolveBinding(entityStore: EntityStore, pkg: DocId)(
+        implicit ec: ExecutionContext, transid: TransactionId): Future[WhiskPackage] = {
+        WhiskPackage.get(entityStore, pkg) flatMap {
+            case wp if wp.binding.isEmpty =>
+                Future.successful(wp)
+            case wp => resolveBinding(entityStore, wp.binding.get.docid)
```
markusthoemmes: Should this rather use idiomatic `Option` handling:

```
WhiskPackage.get(entityStore, pkg) flatMap {
    _.binding.map { binding =>
        resolveBinding(entityStore, binding.docid)
    }.getOrElse(Future.successful(pkg))
}
```

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -336,40 +340,103 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
         }
     }
 
+    /**
+     * helper method that fixes the sequence components that are in the default namespace
+     * replace default namespace witht the user namespace
+     */
+    private def fixDefaultPackage(seq: SequenceExec, user: Identity)(implicit transid: TransactionId): SequenceExec = {
```
markusthoemmes: `resolveDefaultPackage`?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -336,40 +340,103 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+        // if components are part of the default namespace, they contain `_`; replace it!
+        val resolvedComponents = seq.components map { c: FullyQualifiedEntityName => FullyQualifiedEntityName(c.path.resolveNamespace(user.namespace), c.name) }
```
markusthoemmes: remove the parameter type.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -336,40 +340,103 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+                val fixedContent = WhiskActionPut(Some(fixedExec), content.parameters, content.limits, content.version, content.publish, content.annotations)
+                checkSequenceActionLimits(FullyQualifiedEntityName(action.namespace, action.name), fixedExec.components)  map { _ =>
+                    updateWhiskAction(fixedContent, action)
+                } recoverWith{
+                    case _: TooManyActionsInSequence =>
+                        Future failed RejectRequest(BadRequest, "too many actions in sequence")
+                    case _: SequenceWithRecursion =>
+                        Future failed RejectRequest(BadRequest, "recursion detected in sequence")
+                    case _: NoDocumentException =>
+                        Future failed RejectRequest(BadRequest, "action not found")
```
markusthoemmes: Fold the catches into `checkSequenceActionLimits`?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
markusthoemmes: I don't think it will? There is no else block, nothing will happen with the failed Future i believe.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -605,6 +674,79 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+            Future.failed(new TooManyActionsInSequence())
+        else {
+            actionsToInline match {
+                case action :: restActions =>
+                    val docid = DocId(action.toString)
+                    WhiskAction.get(entityStore, docid) flatMap { act =>
+                        act.exec match {
+                            case seq: SequenceExec =>
+                                // this is actually a sequence, check already traversed sequences
+                                if (sequences.contains(FullyQualifiedEntityName(act.namespace, act.name))) {
```
markusthoemmes: Why recreate a `FullyQualifiedEntityName`, `action` already is one?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -605,6 +674,79 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+                                   Future.failed(new SequenceWithRecursion())
+                                } else {
+                                    // resolve the components first
+                                    // need to inline each of its components
+                                    val resolvedComponents = seq.components map { c => WhiskAction.resolveAction(entityStore, c) }
+                                    val futureResComponents = Future.sequence(resolvedComponents)
+                                    futureResComponents flatMap { components =>  // these are resolved components
+                                        // check that these components don't overlap with the sequences found so far
+                                        val overlap = components.foldLeft(false){(b, elem) => b || sequences.contains(elem) }
```
markusthoemmes: Is this the same as `val overlap = (components intersect sequences).nonEmpty` ?

markusthoemmes:
Could you please install https://github.com/openwhisk/openwhisk/blob/master/tools/eclipse/scala.properties to avoid formatting mismatches?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
ioana-blue: Good catch. This didn't use to be the case.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
ioana-blue: why?

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -238,7 +240,26 @@ object WhiskAction
...
 
+    /**
+     *  utility function that given a fully qualified name for an action, resolve its possible package bindings and returns
+     *  the fully qualified name of the resolved action
+     */
+    def resolveAction(entityStore: EntityStore, fullyQualifiedName: FullyQualifiedEntityName)(
+        implicit ec: ExecutionContext, transid: TransactionId): Future[FullyQualifiedEntityName] = {
+        // first check that there is a package to be resolved
+        val entityPath = fullyQualifiedName.path
+        if (entityPath.defaultPackage) {
```
ioana-blue: The package will not be found and it will fail the future which is caught later on. There is no way around it, unless you see one. If we really want these things to typecheck we need to have Namespace, Package and EnitityName separately. And then an FQEN has all three. This requires an entire pass over the code which I'm not going to do now.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
markusthoemmes: Not needed to be there and doesn't add more clarity, while adding noise to me as a reader.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
ioana-blue: I changed it. It needs to stay because it's in the interface. It won't be used at all for sequences since there is no "size" associated with it.

ioana-blue: Unless we start limiting the sum of the size of the actions that can be part of a seq.

ioana-blue:
@rabbah @markusthoemmes another look before I rebase and squash? I think I addressed all issues, let me know if I missed anything.

ioana-blue:
Closing this, subsumed by #1337

PR #1304, opened 2016-09-21
Update JSON Index for CLI Download

dubeejw:
- Uses double qoutes instead of single qoutes in content.json
- Add test to ensure content.json exists
- Add test to ensure the JSON file contains the right content
- Add tests to download CLI from paths specified in content.json

### DIFF
--- a/tests/src/system/rest/GoCLINginxTests.scala
+++ b/tests/src/system/rest/GoCLINginxTests.scala
@@ -23,6 +23,8 @@ import org.scalatest.junit.JUnitRunner
 
 import com.jayway.restassured.RestAssured
 
+import spray.json._
+import DefaultJsonProtocol._
 /**
  * Basic tests of the download link for Go CLI binaries
  */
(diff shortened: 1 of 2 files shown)


csantanapr:
@dubeejw I think is useful to add to the cli property some other info
maybe start with version

```
cli: {
  "version": "2016-09-21T07:55:05Z",
}
```

In the future we could have something like this:
```
cli: {
  "version": "1.0.0",
  "build": "whisk-build-531"
  "date": "2016-09-21T07:55:05Z"
}
```

Where version is the semantic version of the CLI, and build could be a git tag, or some CICD build tag

csantanapr:
forgot to add, the reason behind this it's a user feature that the CLI could have the ability to self check if it's outdated and even download a new version.

Also helpful for a download web page UI if they provide the links to the download the binary they can show the "version" and the "date" that was updated.

tools/cli/build.sh
```diff
@@ -70,10 +70,10 @@ compress_binary() {
 };
 
 create_cli_packages() {
-  local dirIndex="{'cli':{"
+  local dirIndex="{\"cli\":{"
```
markusthoemmes: Can you use single quotes on the outside here to prevent the escaping?

csantanapr:
@dubeejw about cli version, this can be handle as a new issue/enhancement doesn't have to be part of this PR to define "cli":"platforms":{..}

dubeejw:
PG2 221.

dubeejw:
Created an issue for the additions to content.json: https://github.com/openwhisk/openwhisk/issues/1309.

dubeejw:
@csantanapr, I opened an issue for your suggestions. Anything else on this one?

csantanapr:
@dubeejw I just tried your PR with Vagrant, the data in content.json seems corrupted
image

dubeejw:
String interpolation does not work when using single quotes in a shell script. That means we will have to use double quotes to create strings, and escape double quotes inside those strings.

csantanapr:
Yes you need to updat build.sh to work by adding the escape for double quotes `\"`

dubeejw:
Should be working now.

tests/src/system/rest/GoCLINginxTests.scala
```diff
@@ -81,6 +82,13 @@ class GoCLINginxTests extends FlatSpec with Matchers with RestUtil {
         }
     }
 
+    it should "respond to {$DownloadLinkGoCli}/content.json" in {
```
rabbah: why wrap with {} on a scalar?

tests/src/system/rest/GoCLINginxTests.scala
```diff
@@ -81,6 +82,13 @@ class GoCLINginxTests extends FlatSpec with Matchers with RestUtil {
...
+        val response = RestAssured.given().config(sslconfig).
+          get(getServiceURL() + DownloadLinkGoCli + "/content.json")
+
+        response.statusCode should be(200)
```
rabbah: check the contents?

rabbah:
Should the CLI show this information on `wsk property get`? Whatever CLI info is shown by the CLI should be consistent with the json file - is it?

dubeejw:
The CLI version, build, and date are not included in the JSON file yet. Those entries will be added in a different PR. When we include that information it should be consistent with what is shown in ```wsk property get```.

tests/src/system/rest/GoCLINginxTests.scala
csantanapr: Now with content.json, maybe the existing tests should be driven by the content of content.json
For example check if you can download mac and the location of the binary you get from content.json

csantanapr:
I can confirm content.json has correct content now, at least by eye inspection :-)

dubeejw:
Awesome, adding a test to make sure it is always correct.

dubeejw:
PG1 555.

csantanapr:
rebase and merged fc51a4f3ff8f9965e873dbb9dd2f3d86617902be

PR #1306, opened 2016-09-21
Fixes two errors in invoke path

### LINKED ISSUE #1149: invoker should yield a clear error message if docker action fails to pull/run
If docker pull fails for blackbox action, invoker tries to run action anyway

```
> wsk action create bogus <something that doesn't exist>
> wsk action invoke bogus
```

The invoker will attempt to pull the docker image, as it should but this should fail:

```
Running command: /usr/bin/docker pull <something that doesn't exist>
> stdout: Using default tag: latest Pulling repository docker.io/library/<something that doesn't exist>
> stderr: Error: image library/<something that doesn't exist> not found 
```

The invoker continues to run the container which again tries to pull the image (run pulls if needed):

```
> stderr: Unable to find image '<something that doesn't exist>:latest' locally Pulling repository docker.io/library/<something that doesn't exist> Error: image library/<something that doesn't exist> not found
```

It might be OK to go ahead and run should pull fails (keeps the logic simpler in the invoker) but here's the activation response:

```
{
    "logs": [
        "Error starting container"
    ],
    "response": {
        "result": {
            "error": "the action did not produce a valid JSON response"
        },
        "status": "action developer error",
        "success": false
    },
}
```

The system is working as designed, the logs are consistent with the error, as is the activation response --- but we should report a more specific error message e.g., “pull from registry failed”. Arguably, we should leave the logs alone (empty) and instead use the `error` property of the activation record to report a more specific explanation.

rabbah:
1. if docker pull fails for black box container, do not attempt to run container
2. if action is a black box, set appropriate bits correctly

FYI @markusthoemmes @psuter.
Fixes #1149.

Note that at the level of the docker pull/run command we can't distinguish (without additional context) if the command fails because it's a black box action or not. This should matter so that we can attribute the error to the appropriate status (internal vs client).

Rather than wrap operations with `Try` types because it is pervasive, I left the `throws` in place and handled the exception at the outermost level after writing them to a more specific type along the way (where the context is available).

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
@@ -50,8 +50,8 @@ import whisk.core.entity.WhiskEntityStore
 import whisk.core.entity.NodeJS6Exec
 import akka.event.Logging.LogLevel
 import akka.event.Logging.InfoLevel
-import whisk.core.entity.BlackBoxExec
 import whisk.core.entity.AuthKey
+import whisk.core.entity.Exec
 
 /**
  * A thread-safe container pool that internalizes container creation/teardown and allows users
@@ -536,10 +536,10 @@ class ContainerPool(
             try {
                 info(this, s"making new container because none available")
                 startingCounter.next()
-                makeGeneralContainer(key, containerName, imageName, limits, action.exec.kind == BlackBoxExec)
+                makeGeneralContainer(key, containerName, imageName, limits, action.exec.kind == Exec.BLACKBOX)
             } finally {
                 val newCount = startingCounter.prev()
-                info(this, s"made container, number of remaining containers to start: $newCount")
+                info(this, s"finished trying to make container, $newCount more containers to start")
             }
         }
         initWhiskContainer(action, con)
@@ -557,15 +557,31 @@ class ContainerPool(
         val cpuShare = ContainerPool.cpuShare(config)
         val policy = config.invokerContainerPolicy
         val env = getContainerEnvironment()
-        // This will start up the container
-        if (pull) runDockerPull {
-            ContainerUtils.pullImage(dockerhost, imageName)
+
+        // distinguishes between a black box container vs a whisk container
+        def disambiguateContainerError[T](op: => T) = {
+            try { op } catch {
+                case e: ContainerError =>
+                    throw if (pull) BlackBoxContainerError(e.msg) else WhiskContainerError(e.msg)
+            }
         }
+
+        if (pull) runDockerPull {
+            disambiguateContainerError {
+                // if pull fails, the transaction is aborted by throwing an exception;
+                // a pull is only done for black box container
+                ContainerUtils.pullImage(dockerhost, imageName)
+            }
+        }
+
+        // This will start up the container
         runDockerOp {
-            // because of the docker lock, by the time the container gets around to be started
-            // there could be a container to reuse (from a previous run of the same action, or
-            // from a stem cell container); should revisit this logic
-            new WhiskContainer(transid, this.dockerhost, key, containerName, imageName, network, cpuShare, policy, env, limits, isBlackbox = pull, logLevel = this.getVerbosity())
+            disambiguateContainerError {
+                // because of the docker lock, by the time the container gets around to be started
+                // there could be a container to reuse (from a previous run of the same action, or
+                // from a stem cell container); should revisit this logic
+                new WhiskContainer(transid, this.dockerhost, key, containerName, imageName, network, cpuShare, policy, env, limits, isBlackbox = pull, logLevel = this.getVerbosity())
+            }
         }
     }
 
--- a/core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
@@ -64,6 +64,7 @@ trait ContainerUtils extends Logging {
     /*
      * TODO: The file handle and process limits should be moved to some global limits config.
      */
+    @throws[ContainerError]
     def makeContainer(name: Option[ContainerName], image: String, network: String, cpuShare: Int, env: Map[String, String], args: Seq[String], limits: ActionLimits, policy: Option[String])(implicit transid: TransactionId): ContainerHash = {
         val nameOption = name.map(n => Array("--name", n.name)).getOrElse(Array.empty[String])
         val cpuArg = Array("-c", cpuShare.toString)
@@ -81,7 +82,7 @@ trait ContainerUtils extends Logging {
         runDockerCmd(cmd: _*).toOption.map { result =>
             ContainerHash.fromString(result)
         } getOrElse {
-            throw new Exception("Container hash or name expected in `makeContainer`.")
+            throw new ContainerError("Failed to start container.")
         }
     }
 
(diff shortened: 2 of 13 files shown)


perryibm:
I guess the testing lesson is that since we do not delete the blackbox image, the pulling aspect can regress without us knowing.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -16,90 +16,35 @@
 
 package whisk.core.invoker
 
-import java.time.Clock
-import java.time.Instant
+import java.time.{ Clock, Instant }
```
rabbah: @markusthoemmes what do you think?

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -536,10 +536,10 @@ class ContainerPool(
             try {
                 info(this, s"making new container because none available")
                 startingCounter.next()
-                makeGeneralContainer(key, containerName, imageName, limits, action.exec.kind == BlackBoxExec)
+                makeGeneralContainer(key, containerName, imageName, limits, action.exec.kind == Exec.BLACKBOX)
```
rabbah: kind is of type string, it will not match the type BlacBoxExec.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -536,10 +536,10 @@ class ContainerPool(
...
             } finally {
                 val newCount = startingCounter.prev()
-                info(this, s"made container, number of remaining containers to start: $newCount")
+                info(this, s"finished making container, number of remaining containers to start: $newCount")
```
rabbah: changed message because at this point the container may have failed already.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -557,15 +557,31 @@ class ContainerPool(
         val cpuShare = ContainerPool.cpuShare(config)
         val policy = config.invokerContainerPolicy
         val env = getContainerEnvironment()
-        // This will start up the container
+
```
rabbah: next few lines can rewrite the low level docker exception.

core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
```diff
@@ -247,9 +249,15 @@ object ContainerUtils extends Logging {
...
      */
+    @throws[ContainerError]
     def pullImage(dockerhost: String, image: String)(implicit transid: TransactionId): DockerOutput = {
         val cmd = Array("pull", image)
-        runDockerCmd(dockerhost, false, cmd)
+        val result = runDockerCmd(dockerhost, false, cmd)
+        if (result != DockerOutput.unavailable) {
+            result
+        } else {
+            throw new ContainerError(s"Failed to pull container image '$image'.")
```
rabbah: this aborts the container making.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -133,66 +78,61 @@ class Invoker(
...
 
-        info(this, s"${action.id} $subject ${msg.activationId}")
+        info(this, s"${actionid.id} $subject ${msg.activationId}")
 
         // caching is enabled since actions have revision id and an updated
         // action will not hit in the cache due to change in the revision id;
         // if the doc revision is missing, then bypass cache
-        WhiskAction.get(entityStore, action.id, action.rev, action.rev != DocRevision()) flatMap {
+        WhiskAction.get(entityStore, actionid.id, actionid.rev, actionid.rev != DocRevision()) recoverWith {
+            case t =>
```
rabbah: this allows us to report a more specific failure: failure to fetch action.

rabbah:
We do have a test for a good black box action - Adding a new test for a bogus one (random string hopefully good enough).

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
perryibm: btw, there is a proposal to get rid of universal ==

perryibm:
I mean to test that "pull" is always working we would need to pre-delete the image.

rabbah:
@perryibm check the added test.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
rabbah: awesome - was surprised the compiler didn't flag that as a fruitless test.

rabbah:
pg2/220 pending.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -557,15 +557,31 @@ class ContainerPool(
...
+        // distinguishes between a black box container vs a whisk container
+        def disambiguateContainerError[T](op: => T) = {
+            try { op } catch {
+                case e: ContainerError => throw if (pull) BlackBoxContainerError(e.msg) else WhiskContainerError(e.msg)
+                case t: Throwable      => throw t
```
markusthoemmes: just don't catch `t`?

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
markusthoemmes: lovely!

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -133,66 +78,61 @@ class Invoker(
...
-            case Failure(t) =>
-                info(this, s"failed to invoke action ${action.id} due to ${t.getMessage}")
-                completeTransactionWithError(action, tran, s"failed to invoke action ${action.id}: ${t.getMessage}")
+            case Failure(ActivationException(e, internal)) =>
+                val response = if (internal) ActivationResponse.whiskError(e) else ActivationResponse.applicationError(e)
+                completeTransactionWithError(actionid.id, msg.action.version, tran, response)
+        } recoverWith {
+            case t: ActivationException =>
+                // for handled exceptions, don't return a failed future
+                Future.successful({})
```
markusthoemmes: Can the `andThen` + `recoverWith` be collapsed?

```
recoverWith {
    case _  @ ActivationException(e, internal)) =>
        val response = if (internal) ActivationResponse.whiskError(e) else ActivationResponse.applicationError(e)
        completeTransactionWithError(actionid.id, msg.action.version, tran, response)
        Future.successful({})
}
```

seems equivalent?

rabbah:
Rethrows a different exception.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
rabbah: note that the thrown exception is different.

rabbah: yup thanks.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
rabbah: adopted - don't need the future.successful since we can flatmap the future from completeTransactionWithError; this is better.

rabbah:
pg2/228 approved.

rabbah:
also pg2/230 and 231.

perryibm:
You must be right that in travis, it's clean.  But wouldn't that mean that for a while we are not pulling at all?  Then how could tests pass?

rabbah:
docker run pulls if needed.

perryibm:
rebased/merged

PR #1312, opened 2016-09-22
quick-start with docker-compose

ddragosd:
I'm looking to get your feedback and I'm sharing my early progress for a simpler setup for Mac using `docker-compose` with Docker `1.10` and above ( I'm currently testing with `1.12.1` )

I'm using a `Makefile` providing 2 commands: `make run` and `make stop`. 
My end-goal would be to create `make quick-start` which could potentially handle the build then invoke `make run`. 

I'm thinking that for people like me that want to get started ASAP and have struggles with ansible and the particular local setup with other dependencies like python, etc, `docker-compose` is a faster way to get started. 

For people using ansible for deployment in other environments, having a local sandbox for ansible is of great help. For this use-case I'm wondering if a VM can be a better choice as it doesn't surface OS X specifics problems, as it could be an Ubuntu/CentOS/others.     

In this way both options could be maintained and the community can choose the one that fits best for a particular setup.

> NOTE: This is WIP. I'll update the PR title once I get something working end to end.


csantanapr:
Hi @ddragosd Are you aware that we have VM approach to setup OpenWhisk locally with no more requirement than having virtual box and vagrant installed?

Why is this not good option for you?

If you don't want to use a VM and want to run locally on your mac or linux you can use ansible, using ansible helps us have a good story around CICD for devops, what contributors are using their local environment matches and very close to what we use to build, deploy in production.

In the past we were using ant and gradle to do the build and deploy, then we re-engineer the whole process to use gradle only for building and ansible for dploying.

We moved into ansible to help us with our CI/CD and be able to maintained a

What are the problems you are facing with ansible?

ddragosd:
thanks for the feedback @csantanapr .
 
I think the first problem is downgrading my docker version which I prefer to avoid if possible. I tried to work around it by using `ansible/environments/local` then I got stuck at a python library to manage docker ( IIRC `docker-py` was having issues )

The second aspect (not a direct issue for me) is enforcing a deployment technology on everyone unless it's really needed. I do appreciate the idea of consistency and I personally like it a lot, but I'm not sure how aligned is with other devops teams . 

A docker-compose approach may  also be used for integration tests between components. I find myself using this quite often in a microservices environment.  I did not dig too much into the code to see how is this handled currently.

markusthoemmes:
Not trying to be defensive at all and open for discussions of any kind of course. Two thoughts though:

1. We've been on docker 1.9 for a reason. It's not because the deployment otherwise doesn't work, but we had issues with docker 1.10 and 1.11 in our production systems, which seem to be resolved with 1.12, which is why we're currently in the progress of migrating to it. Hopefully we'll be able to give green light to update all developer setups as well.
2. We've also looked at docker-compose as a deployment tool (actually, I personally did the prototypes for docker-compose, ansible, chef etc.). I found that ansible is basically a superset of docker-compose (it really is with the newest version of the docker module) with a bit more flexibility when it comes to setting up VMs the way we want it, copying files around etc. All in all: The decision was deliberate.

Feedback though is of course always welcome and I'm happy to give feedback on the docker-compose version of our deployment.

ddragosd:
thanks @markusthoemmes. 

Given that right now I can't make much progress with Docker 1.12 do you happen to have something that would help me progress faster with `docker-compose` instead ? 

I'm not trying to make this issue a debate on ansible. I think that however we do it, if we can provide a 1 liner simple command such as `make quick-start` and *it works* it would be of great help.  

I hope I'm not sending the wrong message; I have nothing against ansible, I'd just love to have a much easier setup for local development.

clehene:
Large organizations tend to standardize deployment technologies if that's Ansible they're in luck, but otherwise, it's unlikely they'll switch to or mix-in Ansible. Likely these will become available as external projects. 

Having multiple options within the project is convenient for users but places a burden on the project maintainers :)

Kubernetes for example chose to keep them internally https://github.com/kubernetes/kubernetes/tree/master/cluster

As for Vagrant vs Docker https://circleci.com/blog/its-the-future/ ;)

rabbah:
> if we can provide a 1 liner simple command such as make quick-start 

We do have `wskdev fresh` where `wskdev` is hides the underlying `gradle` and `ansible` tools used to build and deploy; for most local development, it's sufficient and convenient to work with `wskdev` (see https://github.com/openwhisk/openwhisk/tree/master/tools/build#how-to-use-redo, which is aliased to `wskdev`).

As for whether this "works" - if you're having trouble with gradle and ansible, then using `wskdev` will have the same trouble since it's just a think wrapper around these.

I agree with @clehene's point that "having multiple options within the project is convenient for users but places a burden on the project maintainers" - if we don't test it, it will break (so putting additional options outside the main repo has the advantage of shifting the burden elsewhere).

ddragosd:
@rabbah thanks for clarifying. I'd be happy to give that a try as well and compare the time-to-value for both options.

ddragosd:
I finally got the first version that *runs on my machine* 😄 
@rabbah @psuter @markusthoemmes  @csantanapr thanks for helping.

By reusing a lot of pieces from the existing code-base I managed to create:

```make quick-start```

1. it builds the project and the docker container

  > NOTE: it reuses `gradlew distdocker`. 

2. then it starts all components using docker-compose and sets up the database reusing existing scripts

  >  I've also added an API Gateway based on NGINX and Openresty which for now it only does load-balancing but it can do much more. 

3. it then runs a sample `hello-world` function and prints the result

  > if all is good it prints
   ```{
    "payload": "Hello, World!"
}```

To run the function again:

```make hello-world```

To stop the environment:

```make destroy```

This command stops all docker images, removing them, deleting the temporary folders, and also removing dangling docker volumes.

So now we have support for people who want to develop on docker `1.12+`.

**Comment by psuter** [comment]
This is great! I'll try to reproduce your steps in my environment and
report back.

On Sep 29, 2016 4:18 AM, "ddragosd" <notifications@github.com> wrote:

I finally got the first version that *runs on my machine* 😄
@rabbah <https://github.com/rabbah> @psuter <https://github.com/psuter>
@markusthoemmes <https://github.com/markusthoemmes> @csantanapr
<https://github.com/csantanapr> thanks for helping.

By reusing a lot of pieces from the existing code-base I managed to create:

make quick-start

   1.

   it builds the project and the docker container

   NOTE: it reuses gradlew distdocker.

   2.

   then it starts all components using docker-compose and sets up the
   database reusing existing scripts

   I've also added an API Gateway based on NGINX and Openresty which for
   now it only does load-balancing but it can do much more.

   3.

   it then runs a sample hello-world function and prints the result

   if all is good it prints
   {
   "payload": "Hello, World!"
   }


To run the function again:

make hello-world

To stop the environment:

make destroy

This command stops all docker images, removing them, deleting the temporary
folders, and also removing dangling docker volumes.

So now we have support for people who want to develop on docker 1.12+.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openwhisk/openwhisk/pull/1312#issuecomment-250401167>,
or mute the thread

**Comment by psuter** [comment]
On my first try, building and deployment seemed fine, but it then gets stuck doing:

    waiting for the Whisk controller to come up

and the dots just keep adding up. Not sure whether this is related, but the container for the gateway seems to be listening to `:80` while the `curl` pinging seems to occur on `:8888`? Will look again later.

**Comment by ddragosd** [comment]
@psuter thanks for giving it a try. I'm wondering what docker version was used and if it's `>1.10`. The docker-compose makes ports for the components available on the localhost so that developers can test them directly. This assumes some ports would be available: `5984` (couchdb), `2181` (zk) , `9092` (kafka), `8400`,`8500`,`8600`, `8302` (consul), `8888`(controller), `8081` (invoker), `80` and `443` (gateway). 

curl pinging goes directly to the port of the component but to your point we could make it go through the gateway too.

**Comment by ddragosd** [comment]
I've actually updated my branch with the master and the controller logs show:
```
[ERROR] [??] [Config] required property defaultLimits.actions.invokes.concurrentInSystem still not set
```
It would be nice if such properties would set a default value and continue.

**Comment by rabbah** [comment]
> It would be nice if such properties would set a default value and continue.
> 

We've tried to avoid this by having default system properties defined in the ansible files.

**Comment by ddragosd** [comment]
thanks @selfxp for updating to `bin/wsk`

I've also added a performance benchmark script using Apache Benchmark. It can be invoked by executing:
```
make hello-world-perf-test
```

The baseline I got so far locally is:
```
Concurrency Level:      20
Time taken for tests:   426.259 seconds
Complete requests:      2000
Failed requests:        0
Keep-Alive requests:    2000
Total transferred:      1612000 bytes
HTML transferred:       1084000 bytes
Requests per second:    4.69 [#/sec] (mean)
Time per request:       4262.586 [ms] (mean)
Time per request:       213.129 [ms] (mean, across all concurrent requests)
Transfer rate:          3.69 [Kbytes/sec] received
```

With
```
Percentage of the requests served within a certain time (ms)
  50%   4177
  66%   4265
  75%   4340
  80%   4397
  90%   4645
  95%   5012
  98%   5508
  99%   6133
 100%  10676 (longest request)
```

**Comment by rabbah** [comment]
If I understood your benchmarks you're paying the cost for the docker run on each invoke - the cost of which is going to dominate your measurement.

**Comment by rabbah** [comment]
Scratches my comment - I missed the `jordi/ab` container reference in there.

**Comment by selfxp** [comment]
I had to exclude `swift3Action` and `swiftAction` task from build in order to run docker compose, but other than that, everything went super smooth. As a developer I love when I can build and test a project using only a few commands and using docker for this process totally makes sense. 

I've also ran the tests @ddragosd provided and I got similar results. It's just, `4.7 Requests per seconds` doesn't seem to be a lot. Is this an expected value? 

```
Concurrency Level:      20
Time taken for tests:   425.478 seconds
Complete requests:      2000
Failed requests:        0
Keep-Alive requests:    2000
Total transferred:      1612000 bytes
HTML transferred:       1084000 bytes
Requests per second:    4.70 [#/sec] (mean)
Time per request:       4254.776 [ms] (mean)
Time per request:       212.739 [ms] (mean, across all concurrent requests)
Transfer rate:          3.70 [Kbytes/sec] received
```
```
Percentage of the requests served within a certain time (ms)
  50%   4166
  66%   4295
  75%   4357
  80%   4423
  90%   4702
  95%   5028
  98%   5574
  99%   6070
 100%   9231 (longest request)
```

**Comment by rabbah** [comment]
```
Benchmarking 172.17.0.7 (be patient)
Completed 200 requests
Completed 400 requests
Completed 600 requests
Completed 800 requests
Completed 1000 requests
Completed 1200 requests
Completed 1400 requests
Completed 1600 requests
Completed 1800 requests
Completed 2000 requests
Finished 2000 requests


Server Software:        spray-can/1.3.3
Server Hostname:        172.17.0.7
Server Port:            8080

Document Path:          /api/v1/namespaces/guest/actions/hello-perf?blocking=true
Document Length:        542 bytes

Concurrency Level:      20
Time taken for tests:   28.916 seconds
Complete requests:      2000
Failed requests:        1940
   (Connect: 0, Receive: 0, Length: 1940, Exceptions: 0)
Non-2xx responses:      1940
Keep-Alive requests:    2000
Total transferred:      755518 bytes
HTML transferred:       200358 bytes
Requests per second:    69.17 [#/sec] (mean)
Time per request:       289.161 [ms] (mean)
Time per request:       14.458 [ms] (mean, across all concurrent requests)
Transfer rate:          25.52 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       4
Processing:     8  289 809.2    134    8759
Waiting:        8  289 809.2    134    8759
Total:          8  289 809.2    134    8760

Percentage of the requests served within a certain time (ms)
  50%    134
  66%    203
  75%    255
  80%    292
  90%    360
  95%    464
  98%   2907
  99%   4857
 100%   8760 (longest request)
```

**Comment by ddragosd** [comment]
@rabbah I noticed it says 
```
Failed requests:        1940
```

When this happened to me I checked the limits and they were set to low values. Not sure if this is your case too.

**Comment by rabbah** [comment]
Yup - forgot to bump the limits on master. Rerunning.

**Comment by rabbah** [comment]
The numbers above are low - I did not try to reproduce them via docker compose. Instead, I used `wskdev fresh` to build and deploy from master after lifting the `limits` in the ansible files. And then grabbed the benchmarking Makefile targets out of this pull request to run against my deployment (macbook with docker machine).

```
Server Software:        spray-can/1.3.3
Server Hostname:        172.17.0.7
Server Port:            8080

Document Path:          /api/v1/namespaces/guest/actions/hello-perf?blocking=true
Document Length:        542 bytes

Concurrency Level:      20
Time taken for tests:   238.550 seconds
Complete requests:      2000
Failed requests:        0
Keep-Alive requests:    2000
Total transferred:      1612000 bytes
HTML transferred:       1084000 bytes
Requests per second:    8.38 [#/sec] (mean)
Time per request:       2385.495 [ms] (mean)
Time per request:       119.275 [ms] (mean, across all concurrent requests)
Transfer rate:          6.60 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       3
Processing:  1210 2379 902.8   2188    8847
Waiting:     1210 2379 902.8   2188    8847
Total:       1210 2379 902.8   2188    8847

Percentage of the requests served within a certain time (ms)
  50%   2188
  66%   2573
  75%   2799
  80%   2931
  90%   3322
  95%   3814
  98%   4953
  99%   5623
 100%   8847 (longest request)
```

Cursory inspection of the logs shows that a lot of the time is spent in docker pause/unpause which is not necessary for the pattern of activations in this suite against one invoker. Suppressing those operations improves the results — but I caution against reading too much into these numbers. There are other parameters to tweak and should be done relative to the host machine’s resource capacity.

```
Server Software:        spray-can/1.3.3
Server Hostname:        172.17.0.7
Server Port:            8080

Document Path:          /api/v1/namespaces/guest/actions/hello-perf?blocking=true
Document Length:        542 bytes

Concurrency Level:      20
Time taken for tests:   86.374 seconds
Complete requests:      2000
Failed requests:        0
Keep-Alive requests:    2000
Total transferred:      1612000 bytes
HTML transferred:       1084000 bytes
Requests per second:    23.16 [#/sec] (mean)
Time per request:       863.743 [ms] (mean)
Time per request:       43.187 [ms] (mean, across all concurrent requests)
Transfer rate:          18.23 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       4
Processing:   218  862 1633.2    651   18429
Waiting:      218  862 1633.2    651   18429
Total:        218  862 1633.3    651   18429

Percentage of the requests served within a certain time (ms)
  50%    651
  66%    747
  75%    831
  80%    890
  90%   1113
  95%   1306
  98%   1516
  99%  15382
 100%  18429 (longest request)

```

ddragosd:
I think I'm done with this PR and is ready for your review. 

I've tested it on a Mac with Docker `1.12` and on a CentOS 7.2 with Docker `1.11.2` and everything worked fine end-to-end.  Because the Docker Compose can be used not just on Mac but on other OSs as well, I have placed it under its own folder in `./tools/docker-compose`.  

I've also created a [README](https://github.com/ddragosd/openwhisk/tree/fork/tools/docker-compose) with more details.

For organizations such as mine this proves beneficial as we're not using Ansible, we're going to deploy and run OpenWhisk on top of Apache Mesos, and the way that components such as Kafka and others get installed is different. 

Docker Compose's `YML` file makes it easy to understand the components to be executed and it makes it easy to understand the dependencies between them by looking at hints such as `depends_on` or `links`. Credits go to Docker for making not just this spec simple, but also the tooling around containers easy to use.

ddragosd:
Given that docker-compose is OS independent I've automated it with Travis to make it easy to maintain. 

Comparing to a previous build this one adds about 94 seconds out of which 30 are used currently to wait for the invoker to set up a Kafka topic. 

The build reuses the docker containers built by the existing Travis build and that's why it's pretty fast, but it could also be invoked independently, in which case it builds the containers. When executed independently the total time is approx 15 minutes ( b/c it only builds distDocker )

We could think of invoking it separately maybe by using OpenWhisk itself and write a simple function to be executed on every Git commit :). I don't know otherwise if we could invoke multiple Travis builds on each commit. Another idea is that Travis builds the containers and uploads them to a place like Docker Hub from which another build pulls the latest versions of the containers and runs some tests - this would work well for a performance tests suite.

For the moment I tried doing the simplest thing that works.

animeshsingh:
@ddragosd Have you tested this with a swarm cluster - in theory pointing this compose install to a swarm cluster and having it push Whisk there?

rabbah:
This content of this pr is seeding a new repo openwhisk-devtools. @ddragosd shall we close this?

ddragosd:
@rabbah you're right. I'm closing this PR as the work has been moved under: 
https://github.com/openwhisk/openwhisk-devtools/pull/1

Issue #1268, opened 2016-09-15
API v2:  add "gateway" entities


sjfink:
As discussed with @csantanapr,

In API v2 we plan to add support for "API Gateways" ... that is, a way to declaratively specify a mapping between a set of whisk assets (actions, triggers) and a REST API.

The general architecture would have two features:
     1)  a "gateway descriptor", an artifact which describes the mapping between the REST API and OpenWhisk stuff.   This will be based on OpenAPI/swagger.
     2) an "API Gateway" service, which implements the mapping.   This can be based on , say, nginx and OpenResty.

The OpenWhisk system will manage and persist entities which are "gateway descriptors" ... that is, extended swagger files.

For now, let's call these entities "gateways" ... we can debate and change the name if someone has a better idea.

We will add "gateway" entities to the set of entities managed by the OW API, which currently includes actions, triggers, rules, and packages.

This issue is intended to track the addition to the OW API to track gateway entities.

As an initial proposal, we would add CRUD REST as usual to the paths:
/openwhisk/namespaces/{ns}/gateways/{gatewayName}

So to be consistent with the rest of the API, we'd have the usual REST Verbs for GET and DELETE.    POST would be used to create or update a gateway entity, where the form body is the extended swagger object that is the gateway descriptor.   Let's not discuss PUT vs. POST REST philosophy in this issue, please ? :)

We intended to dogfood the implementation of the gateway management service ... we will introduce OW "system" actions which manage the persistence of these gateway entities, in a couchdb/cloudant database.

Work items include:
- [ ] Building the CRUD OW actions for gateway artifacts
- [ ](optional) introducing v2 of the OW API (technically not necessary since this is a non-breaking addition to the API)
- [ ] hook up routes in the controller to the CRUD actions
- [ ] ensure the gateway actions are only available to a privileged "system" user, not normal users.

sjfink:
Step 2 of the API design:

Each gateway entity will include at least a collection of entities called routes.   A route will specify at least an HTTP Verb, an OW action, and some information about parameter mapping.

The API will support the usual CRUD OW operations on individual routes via:

/openwhisk/namespaces/{ns}/gateways/{gatewayName/routes/{id}

mbehrendt:
thanks for the writeup @sjfink . 

a few thoughts / comments on this:
- naming: instead of gateway', what about calling it '(api-)mapping'? the rationale is...we're not allow crud operations of gateways or descriptors of gateways, but rather of these mappings?
[...]

Issue #199, opened 2016-04-08
invoker mounts /var/lib/docker/containers - prevents docker rm from working correctly


domdom82:
When invoker is launched, it mounts a volume as

`/var/lib/docker/containers:/containers`

which is the folder where docker keeps filesystems of all its containers.

I presume this is used for the invoker to access the filesystems of launched actions.

However, when trying to remove other containers (like registrator, consul etc.) you get the following error messages:

```
Docker API Error: Unable to remove filesystem for 57f62f43c6280934c44be469f33709498ac82dd6c960ea5a2bf02ac924db502e: remove /var/lib/docker/containers/57f62f43c6280934c44be469f33709498ac82dd6c960ea5a2bf02ac924db502e/mqueue: device or resource busy"
```

domdom82:
Recent discussion has brought up the problems with fetching logs from user containers making this a non-trivial issue.

Current ideas for mitigation would be:

- `docker logs --since=nnn containerid` to fetch logs instead. needs to be evaluated on performance impact.

- customizing log driver to write into a different file. no apparent option available at least for the json driver. we could only limit log size with regard to security concerns.

danielfederschmidt:
Some thoughts on this from a discussion with @domdom82.

* We should get rid of mounting `/var/lib/docker/containers/` in order to get user container logs if possible
* We could introduce a daemon which runs on the host pipes user container logs from `/var/lib/docker/containers` to a separate directory which is then mounted as a volume by the invoker. This daemon could potentially leverage docker events.

This should prevent the error observed. As a step forward on this issue, I will try to reproduce the error in a minimal working example as a playground for experimenting with different approaches to make fast log extraction possible without mounting `/var/lib/docker/containers`.

perryibm:
Some thoughts on this:

1.  Why doesn't the error occur for the many other action containers which are routinely  removed from the system?

2. An external agent can overcome the problem but will complicate our container-only principle and make deployment potentially complicated.  The agent will itself need to be restarted and coordinated with the invoker as essentially part of the logic is being moved into another component.

3. If we separate out blackbox containers, it is feasible to avoid logging through docker by having the action container push out its logs by some other means.  This is of course a lot more work but might give us much more control and splitting off blackbox means we are committed to multiple implementation strategies.

danielfederschmidt:
I was able to reproduce the issue using the following sequence of actions:

* Checkout openwhisk's master branch
[...]

PR #1317, opened 2016-09-23
API Gateway - db creation,  package/action ansible infrastructure, intial createRoute task

mdeuser:
api gateway route mgmt
- db creation
- package and task install/uninstall ansible infrastructure
- intial createRoute task (still needs better RAS + API GW integration)

Related to #1268


core/routemgmt/createRoute.js
```diff
@@ -0,0 +1,98 @@
+/**
```
sjfink: This file looks like an exact copy of the cloudant create document action.

https://github.ibm.com/BlueMix-Fabric/bluewhisk/blob/master/blue/catalog/cloudant/database-actions/create-document.js

It would be nice to not clone code.   Did you consider either using the cloudant package directly, or delegating interactions with cloudant to the cloudant package via whisk.invoke or a sequence?

mdeuser: yes, it's a copy from the cloudant package.  it's really just a POC placeholder to get a good feel for cloudant/couch operations.

The cloudant package is a Bluemix only feature and this API GW feature is targeted for openwhisk (last I understood), the cloudant package could not be a dependency.

mdeuser:
@markusthoemmes - would you please take a look at the ansible updates to make sure it all looks correct.  Thanks!

markusthoemmes:
@lehoanganh ^^ would you do me the honor?

ansible/roles/routemgmt/tasks/clean.yml
```diff
@@ -0,0 +1,8 @@
+---
+# Remove API Gateway route management actions.
+
+- name: remove route management actions
+  shell: ./uninstallRouteMgmt.sh {{ catalog_auth_key }} {{ groups['edge'] | first }} {{ catalog_namespace }} {{ cli_path }} chdir="{{ openwhisk_home }}/ansible/roles/routemgmt/scripts"
+  environment:
+    OPENWHISK_HOME: "{{ openwhisk_home }}"
+  ignore_errors: True
```
lehoanganh: Why  `ignore_errors: true`?

lehoanganh:
In general, it looks fine, however I would like to avoid using the external scripts, the scripts read properties again in whisk.properties which are again generated from ansible envs. Here Ansible does not do much magic except calling those scripts. Can we move the cli calls in scripts to ansible task/playbooks. It's easier to manage.

ansible/roles/routemgmt/tasks/clean.yml
mdeuser: @lehoanganh - Thanks for taking a look at the ansible files.  This file was modeled after other openwhisk roles (invoker, controller, consul, etc) and most had `ignore errors: True` within their clean.yml ansible files.

My thinking is that routemgmt errors associated with removing non-existent entities are ignored, but other persistent openwhisk failures should be caught during the deploy.yml.

mdeuser:
@lehoanganh - Good point.  I did not think all all whisk.properties values were exactly available as ansible environment variables.  For example, the db.whisk.actions whisk.properties value has no exact ansible equivalent.  Same for db.whisk.gwapi.  whisk.properties.j2 template file has some hardcoded strings that ansible does not have and I was reluctant to include the same hardcoded string values in a second place.

ansible/roles/routemgmt/tasks/clean.yml
lehoanganh: the other openwhisk roles like invoker, controller, etc have to have 'ignore_errors: True` because they can't remove the docker containers (invoker, controller, etc) due to this issue: https://github.com/openwhisk/openwhisk/issues/199
This only happen in an AIO topology, in a distributed topo it does not happen cause invoker containers are not deployed in the same machine as controller.

So, the router clean step should be able to clean the entities, not ignore the errors.

lehoanganh:
you have rights :).

ansible/roles/routemgmt/tasks/clean.yml
mdeuser: Ah, ok.  Ignoring clean up errors is not the norm.  I've updated the uninstall script and ansible file to respect errors during cleanup.

csantanapr:
This is old code, let's close this one so there is no confusion

PR #1324, opened 2016-09-23
Raised payload size limit in nodejsaction proxy

psuter:
Added a test that ensures "large" source files can be processed as actions.

### DIFF
--- a/core/nodejsActionBase/app.js
+++ b/core/nodejsActionBase/app.js
@@ -33,7 +33,7 @@ var logger  = require('./src/logger').getLogger('logs/nodejsaction.log', 'nodejs
 var service = require('./src/service').getService(config, logger);
 
 app.set('port', config.port);
-app.use(bodyParser.json());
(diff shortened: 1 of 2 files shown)


psuter:
PG1/543 on its way

core/nodejsActionBase/app.js
```diff
@@ -33,7 +33,7 @@ var logger  = require('./src/logger').getLogger('logs/nodejsaction.log', 'nodejs
 var service = require('./src/service').getService(config, logger);
 
 app.set('port', config.port);
-app.use(bodyParser.json());
+app.use(bodyParser.json({ limit: "48mb" }));
```
psuter: FYI: https://github.com/expressjs/body-parser#limit

csantanapr: ha good find @psuter 👍  `Defaults to '100kb'`

csantanapr:
@psuter other runtimes OK?, only nodejs had this problem and missing test

PR #1326, opened 2016-09-24
[proposal ]consume timestamp parameter from kafka records and log in activation …

jeremiaswerner:
…feed

The broker adds a timestamp when a message get's created in kafka. This PR will read the timestamp from the record consumed and print in activation feed along with the pipeline occupancy. It's supposed to give some insights how long the messages live in the queue.

ideally, the timestamp should be printed with the tid. need to figure out how to do. suggestions?

alternative would be to extend the ActivationMessage and set the timestamp in the controller. would prefer the proposed option.

reference: https://cwiki.apache.org/confluence/display/KAFKA/KIP-32+-+Add+timestamps+to+Kafka+message#KIP-32-AddtimestampstoKafkamessage-Brokerconfigurationchange


rabbah:
don't we get the time spent in the queue from the markers you added in a previous pr?

rabbah:
@jeremiaswerner do you intend to pursue this pr still?

PR #1330, opened 2016-09-26
Avoid printing object hashes in log messages related to containers

psuter:
I'd like to request a PG exemption.


core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
```diff
@@ -126,7 +126,7 @@ trait ContainerUtils extends Logging {
             getDockerLogFile(containerId, mounted).length
         } catch {
             case e: Exception =>
```
markusthoemmes: `case _: Exception`? \*duck and run\*

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/f63b5359c12a55d1ff2a890a7b1fba8b6224dc75.

PR #1332, opened 2016-09-27
Make sure nanny thread is protected against exceptions.

perryibm:
Protect each of the 3 subcalls of the nanny thread against exceptions.  PG2-262

1. killStragglers is not related to the data structures and is just a startup/teardown action
2. addWarmContainer can fail to create the container in which case it is not introduced into the data structure
3. teardownContainer only operates on containers in a toBeRemoved list which means they are not logically in the data structure anymore


rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/df619b7ffee567bdc58ecf43a92fe503da49f483.

PR #1333, opened 2016-09-27
Addition of /info route for system metadata

nwspeete-ibm:


nwspeete-ibm:
@csantanapr Rodric and I agreed to take out "authorization_endpoint": and 
"token_endpoint", unless there is a reason to have it. Could also just redirect /api/v1 to /api/v1/info?

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -78,6 +81,18 @@ abstract protected[controller] class RestAPIVersion(
...
             "buildno" -> buildno.toJson)
     }
+	
+	/**
+     * Information which describes details of the system metadata.
+     */
+    def systemInfo = {
+        JsObject(
+            "name" -> "openwhisk".toJson,
+            "support" -> "https://github.com/openwhisk/openwhisk/issues".toJson,
```
markusthoemmes: Do we want to hardcode this? Other deployments might want to have a different URL?

markusthoemmes:
Is it worth having an extra route for this? What's the motivation of not just using `/api/v1`?

csantanapr:
I think is fine to leave it at `api/v1`

I think what we want is to solve is provide more information that can be configured at deployment time, meaning thru ansible.

maybe have a generic way to define a json object in ansible, where an operator can but any data they want to surface thru `/api/v1`

build id
build date
support url
docs url
openwhisk version (i.e. git repo and hash)

rabbah:
@nwspeete-ibm is this still an active branch?

nwspeete-ibm:
@rabbah was told to put this work item as lower priority some time back. I believe the current consenses is to have: /api/v1

build id
build date
support url
docs url
openwhisk version (i.e. git repo and hash)

rabbah:
thanks for the update - if you're still picking this up and intend to finish the feature, it's fine keeping this open. if it's going to remain open for another extended period of time as wip, I'd rather we close it.

rabbah:
This shouldn't take more than a couple of hours with tests - I think you're in a better position to decide how to spend your time :).

PR #1336, opened 2016-09-28
Add new var concurrentInSystem to local mac environment

ioana-blue:
Add new var concurrentInSystem to local mac environment.

The new var was left out in one of the recent commits and it breaks local deployment on mac.

See https://github.com/openwhisk/openwhisk/pull/1191/files#r80839287 for original PR and comment.

### DIFF
--- a/ansible/environments/mac/group_vars/all
+++ b/ansible/environments/mac/group_vars/all
@@ -22,6 +22,7 @@ limits:
     invokes:
       perMinute: 60
       concurrent: 30
+      concurrentInSystem: 5000
   triggers:
     fires:
       perMinute: 60


ioana-blue:
There is still something missing. I get an error that the default limit is not set although I can see it in the props file.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/dee761c07449c9636c0b0999c1b352f1e4dfd87e. You need to restart consul to pick up the changes.

markusthoemmes:
Argh. Sorry I even discarded this change manually not knowing why it was still there :/

PR #1339, opened 2016-09-28
3 improvements to invoker

perryibm:
Do not save container log by default.
Run stem cell container on a scheduled Future (but after killStragglers has finished).
Include toBeRemoved in slack calculation.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
@@ -25,11 +25,13 @@ import java.util.concurrent.ConcurrentLinkedQueue
 import scala.collection.concurrent.TrieMap
 import scala.collection.mutable.ListBuffer
 import scala.concurrent.duration._
+import scala.concurrent.Future
 import scala.annotation.tailrec
 import scala.util.{Try, Success, Failure}
 import akka.actor.ActorSystem
 import whisk.common.Counter
 import whisk.common.Logging
+import whisk.common.Scheduler
 import whisk.common.TimingUtil
 import whisk.common.TransactionId
 import whisk.core.WhiskConfig
@@ -67,8 +69,11 @@ class ContainerPool(
     config: WhiskConfig,
     invokerInstance: Integer = 0,
     verbosity: LogLevel = InfoLevel,
-    standalone: Boolean = false)(implicit actorSystem: ActorSystem)
+    standalone: Boolean = false,
+    saveContainerLog: Boolean = false)(implicit actorSystem: ActorSystem)
     extends ContainerUtils {
+
+    implicit val executionContext = actorSystem.dispatcher
 
     // These must be defined before verbosity is set
     private val datastore = WhiskEntityStore.datastore(config)
@@ -175,6 +180,8 @@ class ContainerPool(
                           | myPos = $myPos
                           | completed = ${completedPosition.cur}
                           | slack = ${slack()}
+                          | activeCount = ${activeCount()}
+                          | toBeRemoved = ${toBeRemoved.size}
                           | startingCounter = ${startingCounter.cur}""".stripMargin)
             val conResult = Try(getContainer(1, myPos, key, () => makeWhiskContainer(action, auth)))
             completedPosition.next()
@@ -245,7 +252,7 @@ class ContainerPool(
      * How many containers can we start?  Someone could have fully started a container so we must include startingCounter.
      * The use of a method rather than a getter is meant to signify the synchronization in the implementation.
      */
-    private def slack() = _maxActive - (activeCount() + startingCounter.cur)
+    private def slack() = _maxActive - (activeCount() + startingCounter.cur + toBeRemoved.size)
 
     /*
      * Try to get or create a container, returning None if there are too many
@@ -394,6 +401,7 @@ class ContainerPool(
     private val warmNodejsKey = WarmNodeJsActionContainerId
     private val nodejsExec = NodeJS6Exec("", None)
     private val WARM_NODEJS_CONTAINERS = 2
+    private val RM_THRESHOLD = 4
 
     private def keyMapToString(): String = {
         keyMap.map(p => s"[${p._1.stringRepr} -> ${p._2}]").mkString("  ")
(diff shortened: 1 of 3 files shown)


perryibm:
PG almost passed but failed in standalone test mode.  A fix has been applied and now retesting under PG2-275.

perryibm:
PG2-277 passed.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -422,25 +430,31 @@ class ContainerPool(
...
+                    implicit val tid = TransactionId.invokerWarmup
+                    if (getNumberOfIdleContainers(warmNodejsKey) < WARM_NODEJS_CONTAINERS && slack() > 0 && toBeRemoved.size < RM_THRESHOLD) {
+                        addWarmNodejsContainer()(tid)
+                    } else {
+                        Future.successful(())
+                    }
                 }
+            }
+            while (true) {
+                Thread.sleep(100) // serves to prevent busy looping
```
markusthoemmes: Can this be put in a Scheduler aswell?

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -502,7 +516,7 @@ class ContainerPool(
      * This method will introduce a stem cell container into the system.
      * If container creation fails, the container will not be entered into the pool.
      */
-    private def addWarmNodejsContainer()(implicit transid: TransactionId) = try {
+    private def addWarmNodejsContainer()(implicit transid: TransactionId) = Future { Try {
```
markusthoemmes: `Future[Try[...]]` really necessary? `Future[...]` should be enough (already is a `Try`)

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
markusthoemmes: That being said, do we need the `nannyThread` at all? Shouldn't these be

1. Scheduler that handles warmContainers (should be an `Actor` reacting to messages at some point, but one step at a time)
2. Scheduler to garbage collect containers
3. Off the mainThread one time execution of `killStragglers` (just use a `Future` and/or a `scheduleOnce`?) Thoughts?

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -394,6 +401,7 @@ class ContainerPool(
     private val warmNodejsKey = WarmNodeJsActionContainerId
     private val nodejsExec = NodeJS6Exec("", None)
     private val WARM_NODEJS_CONTAINERS = 2
+    private val RM_THRESHOLD = 4
```
domdom82: shouldn't this be based on maxActive? i.e. system size?

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -717,13 +731,15 @@ class ContainerPool(
      * Actually deletes the containers.
      */
     private def teardownContainer(container: Container)(implicit transid: TransactionId) = try {
-        val size = this.getLogSize(container, !standalone)
-        val rawLogBytes = container.synchronized {
-            this.getDockerLogContent(container.containerId, 0, size, !standalone)
+        if (saveContainerLog) {
```
domdom82: I assume we only record activation logs to db and never to disk in production? So will only be true during tests? This should increase performance quite a bit.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
jeremiaswerner: see line 449. I think this is an arbitrary threshold that avoids creating new warm container if the toBeRemoved queue is >= this threshold.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -422,25 +430,31 @@ class ContainerPool(
...
                 // start this operation while there is slack and end up waiting on the docker lock later
-                if (!standalone && getNumberOfIdleContainers(warmNodejsKey) < WARM_NODEJS_CONTAINERS && slack() > 0) {
-                    addWarmNodejsContainer()(tid)
+                val warmupInterval = 100.milliseconds
+                Scheduler.scheduleWaitAtLeast(warmupInterval) { () =>
+                    implicit val tid = TransactionId.invokerWarmup
+                    if (getNumberOfIdleContainers(warmNodejsKey) < WARM_NODEJS_CONTAINERS && slack() > 0 && toBeRemoved.size < RM_THRESHOLD) {
+                        addWarmNodejsContainer()(tid)
+                    } else {
+                        Future.successful(())
```
jeremiaswerner: should we add a log here so that we can see when we run into this?

domdom82: We should add more logs in general to the critical decision paths. 

Albeit debug-level only.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
perryibm: Jeremias is right, just an arbitrary threshold as an additional safeguard to prioritize the rm ahead of stem cell container.  The accounting with toBeRemoved already protects against the main bug.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
perryibm: 1.  Possibly but there would only be one message so it seems more like a plain periodic task to me.
2.  For now, I want the removal operation to be prioritized as high as possible and not be part of the Future scheduling so it is guaranteed dedicated access.   We can revisit this once the current issue has been handled.
3. It cannot be off the main thread (during construction) because killStragglers can be slow and we need the service to start up.  However, it could be moved to a future of its own though this would still have to sequentially precede the others as killStragglers must run to completion first.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
perryibm: agreed

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
perryibm: Yes we need this only for standalone unit test as in ContainerPoolTests.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
perryibm: We could but the `else` branch is not an unusual or error condition.  In fact, it is common and corresponds to when no one is using the warm containers and so any logging here would represent a ton of verbose logging for normal idle condition.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
perryibm: I shouldn't say it's entirely arbitrary but it's not dependent on the number of core but rather on the tolerable backlog of work based on the performance of `docker rm`.

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
markusthoemmes: So it made it to the codebase as is?

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -512,9 +526,9 @@ class ContainerPool(
             introduceContainer(warmNodejsKey, con)
         }
         info(this, s"Started warm nodejs container $con.id: $con.containerId")
-    } catch {
-        case t: Throwable => warn(this, s"addWarmNodejsContainer encountered an exception: ${t.getMessage}")
-    }
+    }.recover {
+        case t => warn(this, s"addWarmNodejsContainer encountered an exception: ${t.getMessage}")
```
markusthoemmes: Note that this will recover the `Failure` to a `Success` which might (will) be unexpected?

`andThen` would be a better solution to add a log here.

Issue #1334, opened 2016-09-27
Replace docker inspect with looking at config.json


perryibm:
`docker inspect` like `docker ps` empirically grabs a lock and is slow when the number of active containers is large even when inspecting a single container.  Speculating that docker does not use a multi-grain locking scheme.   Even when the number of containers is slow, this is a slowish call.  There is consensus to replace this by inspecting the config.json file directly in the way we obtain logs in an efficient manner by looking in `..../containers/HASH/`.

However, the configuration file has moved from config.json to config.v2.json when going from docker 9 to docker 12 so we will defer working on this until we have completed our transition.

domdom82:
@rabbah @sjfink 
for less disruption: can we do a `docker version` at invoker startup to collect the server version and branch off into the config.json reader if version >= 1.12.x and just keep the old code otherwise? this would be like a feature-toggle and @perryibm would not have to wait until all systems are on 1.12.

we will probably need to check for kernel version, too. some security features only work on certain kernel levels.

perryibm:
We are now fully on docker 1.12 and I'll start looking into this again.  The main issue is to note that there may need to be a retry for the config file (or a particular field) to appear as `docker run` is not instant and docker makes no promise about how started a container is (or daemon state) when the call returns.

perryibm:
The code for this has been pushed and we are undergoing stress testing.  Because it's a fundamental change, I will keep this open until we see evidence that this is working as designed when under load.

markusthoemmes:
Adopted.

PR #1342, opened 2016-09-29
use docker port instead of docker inspect for action containers

csantanapr:
This will avoid the need for special networking route to private network on docker containers for local development on Mac.
With potential to allow the use of docker for mac.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
@@ -75,8 +75,9 @@ trait ContainerUtils extends Logging {
         val processLimit = Array("--ulimit", "nproc=512:512")
         val securityOpts = policy map { p => Array("--security-opt", s"apparmor:${p}") } getOrElse (Array.empty[String])
         val containerNetwork = Array("--net", network)
+        val portOption = Array("-p",":8080")
 
-        val cmd = Seq("run") ++ makeEnvVars(env) ++ consulServiceIgnore ++ nameOption ++ cpuArg ++ memoryArg ++
+        val cmd = Seq("run") ++ makeEnvVars(env) ++ consulServiceIgnore ++ portOption ++nameOption ++ cpuArg ++ memoryArg ++
             capabilityArg ++ fileHandleLimit ++ processLimit ++ securityOpts ++ containerNetwork ++ Seq("-d", image) ++ args
 
         runDockerCmd(cmd: _*).toOption.map { result =>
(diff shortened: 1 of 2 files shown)


rabbah:
Note that per this issue https://github.com/openwhisk/openwhisk/issues/1334, the invoker will eschew calls to `docker inspect`.

csantanapr:
@rabbah This PR is work in progress not ready.

I was already about replacing docker inspect for getting the info directly from config.json from under docker.

This PR is WIP, I wanted to investigate the option of having only for Mac Dev Environment  the ability of using `docker port`

This way we have a easier for people to setup their local dev environment, by avoiding the special network route tweak, and the ability to use docker for mac instead of docker-machine if they like

I was planning to make it conditional, if Mac then, if not then use the current mode.

But then I was curious if using `docker port` we don't hit the performance problems of using `inspect` 

That's why I'm debating if I should make condition in code with if Mac, or maybe control it thru wiskprops.

By doing with wiskprops property we than can control which method to use from the outside, and we can run the benchmark side by side more easier.

I think I would go in phases, solve one problem at a time and just solve it for Mac Dev, by detecting Mac.

markusthoemmes:
Is the route tweak even necessary for plain deployment of OpenWhisk. I was under the impression its only needed for the actionContainer tests? Everything else is handled inside the VM anyway?

csantanapr:
@markusthoemmes I think I know see what you mean.
Making myself more familiar with invoker, you are correct when the invoker executes docker commands does them from it's container using the docker cli from the invoker container out to the host where docker engine is located.
For the run and init also you are correct going from invoker container to action container they seat on the same subnet so goes thru internal network

But I think I found some tests that call run and init from docker host, when tests runs in addition of actionContainer tests.
Take a look at [tests/src/whisk/core/container/test/ContainerPoolTests.scala)](https://github.com/openwhisk/openwhisk/blob/master/tests/src/whisk/core/container/test/ContainerPoolTests.scala)

Is my assumption correct that when this tests run the run and init against the docker action happens from the machine running the test, not the invoker.

rabbah:
You're right Carlos. These are the actionContainer tests that Markus alluded to.

csantanapr:
what I don't get about this test is the use of localhost
@rr any ideas
https://github.com/openwhisk/openwhisk/blob/master/tests/src/whisk/core/container/test/ContainerPoolTests.scala#L62
```
    val config = new WhiskConfig(
        Map(selfDockerEndpoint -> "localhost",
            dockerEndpoint -> null,
            edgeHostName -> "localhost",
            invokerSerializeDockerOp -> "true",
            invokerSerializeDockerPull -> "true")
            ++ WhiskEntityStore.requiredProperties
            ++ WhiskAuthStore.requiredProperties)
```

rabbah:
> was planning to make it conditional, if Mac then, if not then use the current mode.

This is dangerous in that we don't have Mac specific tests and makes the invoker platform aware where it isn't otherwise. A general solution is preferred.

csantanapr:
@rabbah I think I understand now better. I agree from the invoker there is no need to do port mapping.

We already have a case where we detect Mac but only from test system 


For the test cases I was I aware of 1. actionContainers like @markusthoemmes commented
https://github.com/openwhisk/openwhisk/tree/master/tests/src/actionContainers

I also found 2. 
ContainerPoolTests.scala
https://github.com/openwhisk/openwhisk/blob/master/tests/src/whisk/core/container/test/ContainerPoolTests.scala#L62

Are there any other test that do docker run and then init/run on port 8080 with the internal ip from the test system?

csantanapr:
revisiting this in a later time.

PR #1347, opened 2016-09-29
Directly access loadbalancer from ActivationThrottler so we can skip consul

perryibm:
We stop the loadbalancer from publishing to consul counts of issued invocations.  Since only the controller is consuming this count and they are now in the same process, the data can be communicated locally.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Backend.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Backend.scala
@@ -24,7 +24,6 @@ import akka.actor.ActorSystem
 import akka.event.Logging.InfoLevel
 import akka.util.Timeout
 import akka.util.Timeout.durationToTimeout
-import spray.json.JsObject
 import whisk.common.TransactionId
 import whisk.core.connector.ActivationMessage
 import whisk.core.WhiskConfig
@@ -47,14 +46,14 @@ object WhiskServices {
     /**
      * Creates instance of an entitlement service.
      */
-    def entitlementService(config: WhiskConfig, timeout: FiniteDuration = 5 seconds)(
+    def entitlementService(config: WhiskConfig, loadBalancer: LoadBalancerService, timeout: FiniteDuration = 5 seconds)(
         implicit as: ActorSystem) = {
         // remote entitlement service requires a host:port definition. If not given,
         // i.e., the value equals ":" or ":xxxx", use a local entitlement flow.
         if (config.entitlementHost.startsWith(":")) {
-            new LocalEntitlementService(config)
+            new LocalEntitlementService(config, loadBalancer)
         } else {
-            new RemoteEntitlementService(config, timeout)
+            new RemoteEntitlementService(config, loadBalancer, timeout)
         }
     }
 
@@ -67,10 +66,9 @@ object WhiskServices {
      * and returns the HTTP response from the load balancer as a future
      */
     def makeLoadBalancerComponent(config: WhiskConfig, timeout: Timeout = 10 seconds)(
-        implicit as: ActorSystem): (LoadBalancerReq => Future[Unit], () => JsObject, (ActivationId, FiniteDuration, TransactionId) => Future[WhiskActivation]) = {
+        implicit as: ActorSystem): LoadBalancerService = {
         val loadBalancer = new LoadBalancerService(config, InfoLevel)
-        val requestTaker = (lbr: LoadBalancerReq) => { loadBalancer.publish(lbr._1)(lbr._2) }
-        (requestTaker, loadBalancer.getInvokerHealth, loadBalancer.queryActivationResponse)
+        loadBalancer
     }
 
 }
--- a/core/controller/src/main/scala/whisk/core/controller/Backend.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Backend.scala
@@ -16,14 +16,12 @@
 
 package whisk.core.controller
 
-import scala.concurrent.Future
 import scala.concurrent.duration.DurationInt
 import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.Future
 
 import akka.actor.ActorSystem
 import akka.event.Logging.InfoLevel
-import akka.util.Timeout
-import akka.util.Timeout.durationToTimeout
 import whisk.common.TransactionId
 import whisk.core.connector.ActivationMessage
 import whisk.core.WhiskConfig
@@ -36,8 +34,6 @@ import scala.language.postfixOps
 import whisk.core.entity.ActivationId.ActivationIdGenerator
 
 object WhiskServices {
-
-    type LoadBalancerReq = (ActivationMessage, TransactionId)
 
     def requiredProperties = WhiskConfig.loadbalancerHost ++ WhiskConfig.consulServer ++ EntitlementService.requiredProperties
 
@@ -62,14 +58,10 @@ object WhiskServices {
      *
      * @param config the configuration with loadbalancerHost defined
      * @param timeout the duration before timing out the HTTP request
-     * @return function that accepts a LoadBalancerReq, posts request to load balancer
-     * and returns the HTTP response from the load balancer as a future
+     * @return a load balancer component
      */
-    def makeLoadBalancerComponent(config: WhiskConfig, timeout: Timeout = 10 seconds)(
-        implicit as: ActorSystem): LoadBalancerService = {
-        val loadBalancer = new LoadBalancerService(config, InfoLevel)
-        loadBalancer
-    }
+    def makeLoadBalancerComponent(config: WhiskConfig)(
+        implicit as: ActorSystem): LoadBalancerService = new LoadBalancerService(config, InfoLevel)
 
 }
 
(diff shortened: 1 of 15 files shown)


perryibm:
PG1-579 passed.  Travis failure due to swift image.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerToKafka.scala
```diff
@@ -102,21 +99,9 @@ trait LoadBalancerToKafka extends Logging {
...
-    protected def getUserActivationCounts(): Map[String, JsObject] = {
-        userActivationCounter.toMap mapValues {
-            _.cur
-        } groupBy {
-            case (key, _) => LoadBalancerKeys.userActivationCountKey + "/" + key.substring(0, 1)
-        } mapValues { map =>
-            map.toJson.asJsObject
-        }
-    }
+    def getUserActivationCounts(): Map[String, Long] = userActivationCounter.toMap mapValues { _.cur.toLong }
```
markusthoemmes: Can we omit the `()` as this operation is no longer "side-effecting"?

Also: Why `Long` over `Int`? `Counter` only emits an `Int` (but in general I agree that we should use `Long` for these, but then everywhere, starting at `Counter`)

core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
```diff
@@ -113,8 +100,8 @@ class ActivationThrottler(consulServer: String, concurrencyLimit: Int, systemOve
     }
 
     Scheduler.scheduleWaitAtLeast(healthCheckInterval) { () =>
+        val loadbalancerActivationCount = loadBalancer.getUserActivationCounts()
         for {
-            loadbalancerActivationCount <- getLoadBalancerActivationCount()
```
markusthoemmes: Either write

```
for {
    loadbalancerActivationCount = loadBalancer.getUserActivationCounts()
    invokerActivationCount <- getInvokerActivationCount()
}
...
```

or we should get rid of the for comprehension in favor of just a "map". What do you think?

core/controller/src/main/scala/whisk/core/controller/Backend.scala
```diff
@@ -67,10 +66,9 @@ object WhiskServices {
      * and returns the HTTP response from the load balancer as a future
      */
     def makeLoadBalancerComponent(config: WhiskConfig, timeout: Timeout = 10 seconds)(
-        implicit as: ActorSystem): (LoadBalancerReq => Future[Unit], () => JsObject, (ActivationId, FiniteDuration, TransactionId) => Future[WhiskActivation]) = {
+        implicit as: ActorSystem): LoadBalancerService = {
         val loadBalancer = new LoadBalancerService(config, InfoLevel)
-        val requestTaker = (lbr: LoadBalancerReq) => { loadBalancer.publish(lbr._1)(lbr._2) }
-        (requestTaker, loadBalancer.getInvokerHealth, loadBalancer.queryActivationResponse)
+        loadBalancer
```
markusthoemmes: I think this can be just

```
def makeLoadBalancerComponent(config: WhiskConfig)(implicit as: ActorSystem) =
        new LoadBalancerService(config, InfoLevel)
```

core/controller/src/main/scala/whisk/core/controller/Backend.scala
```diff
@@ -67,10 +66,9 @@ object WhiskServices {
      * and returns the HTTP response from the load balancer as a future
      */
```
markusthoemmes: ScalaDoc does no longer apply, please adjust to the changes.

core/controller/src/main/scala/whisk/core/controller/Backend.scala
```diff
@@ -67,10 +66,9 @@ object WhiskServices {
...
     def makeLoadBalancerComponent(config: WhiskConfig, timeout: Timeout = 10 seconds)(
```
markusthoemmes: `timeout` is unused, can just go away.

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -163,8 +163,11 @@ protected[controller] class RestAPIVersion_v1(
 
     // initialize backend services
     protected implicit val consulServer = WhiskServices.consulServer(config)
-    protected implicit val entitlementService = WhiskServices.entitlementService(config)
-    protected implicit val (performLoadBalancerRequest, getInvokerHealth, queryActivationResponse) = WhiskServices.makeLoadBalancerComponent(config)
+    protected implicit val loadBalancer = WhiskServices.makeLoadBalancerComponent(config)
+    protected implicit val performLoadBalancerRequest = loadBalancer.acceptRequest _
+    protected implicit val getInvokerHealth = loadBalancer.getInvokerHealth _
+    protected implicit val queryActiavtionResponse = loadBalancer.queryActivationResponse _
```
markusthoemmes: `queryActivationResponse` (typo)

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -163,8 +163,11 @@ protected[controller] class RestAPIVersion_v1(
 
     // initialize backend services
     protected implicit val consulServer = WhiskServices.consulServer(config)
-    protected implicit val entitlementService = WhiskServices.entitlementService(config)
-    protected implicit val (performLoadBalancerRequest, getInvokerHealth, queryActivationResponse) = WhiskServices.makeLoadBalancerComponent(config)
+    protected implicit val loadBalancer = WhiskServices.makeLoadBalancerComponent(config)
+    protected implicit val performLoadBalancerRequest = loadBalancer.acceptRequest _
+    protected implicit val getInvokerHealth = loadBalancer.getInvokerHealth _
```
markusthoemmes: Do we need this? I think this can be inlined as `loadBalancer.getInvokerHealth()` in `internalInvokerHealth` at the bottom of this file.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -95,11 +96,15 @@ class LoadBalancerService(config: WhiskConfig, verbosity: LogLevel)(
...
             }
-            Map(LoadBalancerKeys.invokerHealth -> getInvokerHealth()) ++
-                getUserActivationCounts()
+            Map(LoadBalancerKeys.invokerHealth -> getInvokerHealth())
         })
 
     /**
+     * A bridge method to impedance match against RestAPI.
+     */
+    def acceptRequest(lbr: LoadBalancerReq) = publish(lbr._1)(lbr._2)
```
markusthoemmes: Is this really needed? I think we should get rid of `LoadBalancerReq` and this `acceptRequest` method.

Depending on which level we want to have the abstraction:
1. (less intrusive): remove the types + method as mentioned and replace them with explicit calls. The line  167 in RestAPI.scala becomes `protected implicit val performLoadBalancerRequest = loadBalancer.publish(_: ActivationMessage)(_: TransactionId)` for example.
2. (more intrusive, but not really): just remove the abstraction layer and pass the LoadBalancerService through, as we do for the EntitlementService anyway.

Thoughts?

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerToKafka.scala
perryibm: Yes.
Because most other places are using Long and so actually we should switch to Long.

core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
perryibm: Get rid of that comprehension - I never liked it.

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
perryibm: sure

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
markusthoemmes: Yep I agree. I can contribute a commit to this PR if you want, but I feel like we should get rid of these bridge methods, if they are of no real help. And we're leaking `loadBalancer` into Entitlement anyway

perryibm: Now retesting without this last bit (as 1st step)

perryibm: It's a no go to get rid of the bridge methods.  They are in fact needed due to unit testing of the controller by itself (ControllerTestCommon) when Kafka and DB are not even up.   I am reverting aspects of the last commit so that it's slightly cleaner but without getting rid of these two methods.

perryibm:
Basically this is working now.  The builds (travis and PG) are currently failing due to swift issues.

perryibm:
After rebase, PG1-596 now passed.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
markusthoemmes: Good one. I think it's okay now, much cleaner already!

core/controller/src/main/scala/whisk/core/controller/Backend.scala
```diff
@@ -63,15 +58,10 @@ object WhiskServices {
      *
      * @param config the configuration with loadbalancerHost defined
      * @param timeout the duration before timing out the HTTP request
```
markusthoemmes: param went away. Will fix in merge.

perryibm: thx

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/411864ce295a02a49497c9352e76a0283424f3f0.

PR #1348, opened 2016-09-29
Use Promises within nodeJsAction proxy code

psuter:
Trying to dial down the callback hell in the proxy code.

The underlying reason is that with attachments, `doInit()` is becoming asynchronous.

FYI @rabbah

### DIFF
--- a/core/nodejsActionBase/runner.js
+++ b/core/nodejsActionBase/runner.js
@@ -19,63 +19,94 @@
  * an action.
  *
  * This file (runner.js) must currently live in root directory for nodeJsAction.
- * See issue #102 for a discussion.
  */
-var util  = require('util');
+var util = require('util');
 
-function NodeActionRunner(message, whisk) {
-    this.userScriptName = getname(message.name);
+function NodeActionRunner(whisk) {
+    // Use this ref inside lambdas etc.
+    var thisRunner = this;
+
+    this.userScriptName = undefined;
     this.userScriptMain = undefined;
-    this.initError = undefined;
 
-    try {
-        eval(message.code);
-        this.userScriptMain = eval(message.main);
-    } catch (e) {
-        this.initError = e;
-        return;
+    // This structure is reset for every action invocation. It contains two fields:
+    //   - completed; indicating whether the action has already signaled completion
+    //   - next; a callback to be invoked with the result of the action.
+    // Note that { error: ... } results are still results.
+    var callback = {
+        completed : undefined,
+        next      : function (result) { return; }
+    };
+
+    this.whisk = modWhisk(whisk, callback);
+
+    this.init = function(message) {
+        // Determining a sensible name for the action.
+        var name = typeof message.name === 'string' ? message.name.trim() : '';
+        name = name !== '' ? name : 'Anonymous User Action';
+
+        this.userScriptName = name;
+
+        // Loading the user code.
+        return new Promise(
+            function (resolve, reject) {
+                try {
+                    eval(message.code);
+                    thisRunner.userScriptMain = eval(message.main);
+
+                    if (typeof thisRunner.userScriptMain === 'function') {
+                        // The value 'true' has no special meaning here;
+                        // the successful state is fully reflected in the
+                        // successful resolution of the promise.
+                        resolve(true);
+                    } else {
+                        throw "Action entrypoint '" + message.main + "' is not a function.";
+                    }
+                } catch (e) {
+                    reject(e);
+                }
+            }
+        );
     }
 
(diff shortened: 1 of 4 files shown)


psuter:
Will be PG1 577.

core/nodejsActionBase/src/service.js
```diff
@@ -44,68 +52,78 @@ function NodeActionService(config, logger) {
...
+                    console.error('Result must be of type object but has type "' + typeof result + '":', result);
+                    return errorMessage(502, "The action did not return a dictionary.");
                 } else {
-                    res.status(500).json({ error: 'An error has occurred: ' + errMsg });
+                    return responseMessage(200, result);
                 }
+            }).catch(function (error) {
+                setStatus(Status.ready);
+
+                return Promise.reject(errorMessage(500, "An error has occurred: " + error));
```
bjustin-ibm: TIL you can return promises from within a `catch` callback.

psuter:
@bjustin-ibm thanks for the review, I agree with your suggestion and will make `doInit` async now.

psuter:
Re-submitting for your reviewing pleasure. I continued the callback witch-hunt deeper into `service.js` and `runner.js`. I removed a bunch of deadcode and simplified some logic in the process.

PG2 303 is running.

core/nodejsActionBase/src/service.js
```diff
@@ -44,112 +65,96 @@ function NodeActionService(config, logger) {
...
+
+        return userCodeRunner.run(args).then(function(response) {
             writeMarkers();
-            next(false, e);
-        }
+            return response;
+        }).catch(function (error) {
+            console.error(error);
+            writeMarkers();
+            reject(e);
```
bjustin-ibm: There is no `reject` function available here. Seems like perhaps a leftover from your previous implementation? Probably meant to be `return Promise.reject(e)`

core/nodejsActionBase/src/service.js
```diff
@@ -44,112 +65,96 @@ function NodeActionService(config, logger) {
...
     }
 
     function doInit(message) {
         var context = newWhiskContext(config, logger);
-        userScript = new NodeActionRunner(message, context);
-        if (typeof userScript.userScriptMain === 'function') {
+
+        userCodeRunner = new NodeActionRunner(context);
+
+        return userCodeRunner.init(message).then(function (result) {
```
bjustin-ibm: Same comment here about `result`. It isn't used here and the value coming from `NodeActionRunner.init()` isn't really meaningful anyway.

core/nodejsActionBase/runner.js
```diff
@@ -19,63 +19,91 @@
...
+
+        // Loading the user code.
+        return new Promise(
+            function (resolve, reject) {
+                try {
+                    eval(message.code);
+                    thisRunner.userScriptMain = eval(message.main);
+
+                    if (typeof thisRunner.userScriptMain === 'function') {
+                        resolve(true);
```
bjustin-ibm: Resolving on `true` is not meaningful to the caller (since this is the only value that is ever resolved from this function) and can be confusing to the reader who may be looking for what circumstances `false` is ever used.

psuter:
Good catch on the invalid `reject(e)`. I updated the tests so that they would have caught it.

I disagree on the other changes, though. Removing the `true`s and the names of unused params seems like a recipe to run into `undefined`s later on. If JS had a `Unit` type I'd use that, but it doesn't, so I used the simplest thing that's not falsy. I added comments to help the future reader, though.

bjustin-ibm:
@psuter You are welcome to prefer the existing pattern, but your comment seems to indicate that you may be under the impression that promises _must_ be resolved on some value, which is not true.  In order to make my suggestion completely clear (which you are still welcome to disagree with), I have coded it up in this commit: c9918fb0f396383e74ff34656e1fb6de7ece6e70

The overall idea is that if your function doesn't produce a meaningful result that the caller is interested in, then it is reasonable to indicate success by resolving on nothing. This is a very common pattern with promises and is well documented:
http://www.2ality.com/2014/10/es6-promises-api.html
http://exploringjs.com/es6/ch_promises.html
http://www.datchley.name/es6-promises/

psuter:
Thanks for the pointers. I'm still unconvinced that it's a good idea to send `undefined` values around, whatever the accepted standard is. Feel free to remove/edit and squash the changes in if you feel strongly enough about this, I won't object.

In other news, PG1 646 is blue-ish.

bjustin-ibm:
@psuter I left my changes out as this is just a matter of readability preference.

PR #1354, opened 2016-09-30
Issue #1329 Add manual compilation to action

### LINKED ISSUE #1329: Remove unnecessary checks from swift action code initialization in Swift 3.0 runtime
We can reduce the initialization time for a Swift 3.0 action by removing some unnecessary steps performed by the swift package manager.

paulcastro:
This PR changes Swift 3 GA runtime to use swiftc rather than SPM to speedup up init times. It pins the dependencies in Package.swift. It also bumps the Swift version to Preview-1. Updates associated container tests for Swift 2 and Swift 3. Fixes #1329. PG 302.


core/swift3Action/Dockerfile
```diff
@@ -53,9 +53,12 @@ ADD epilogue.swift /swift3Action
 ADD swift3runner.py /swift3Action
 ADD spm-build /swift3Action/spm-build
```
psuter: so this poor guy gets written then immediately overwritten?

tests/src/actionContainers/Swift3ActionContainerTests.scala
```diff
@@ -41,6 +41,16 @@ class Swift3ActionContainerTests extends SwiftActionContainerTests {
          |     return ["auth": auth, "edge": edge]
          |}
          """.stripMargin
+    override lazy val errorCode = """
```
psuter: Why do you need this here? `Swift3Action[...]` extends `SwiftAction[...]`?

paulcastro: Its getting less necessary.  Originally, Swift 3 tests were a superset of Swift 2 tests.  Now some of the common tests have diverged because of the changes in the language.

core/swift3Action/Dockerfile
paulcastro: Can change ADD to COPY and remove the ADD spm-build

psuter:
After being helped in understanding the changes, all my comments can be ignored :)

tests/src/actionContainers/Swift3ActionContainerTests.scala
psuter: I'm apparently a terrible "spot-the-difference" player.

psuter:
Merged as 736fbb76ee3762ae5f14efd804c777dbcf106a1a

PR #1357, opened 2016-09-30
Zipped NodeJS modules as actions

psuter:
This PR introduces support for Node actions written as zipped modules.

Builds on #1348 which should be merged first.

### DIFF
--- a/core/nodejsActionBase/runner.js
+++ b/core/nodejsActionBase/runner.js
@@ -21,6 +21,9 @@
  * This file (runner.js) must currently live in root directory for nodeJsAction.
  */
 var util = require('util');
+var child_process = require('child_process');
+var fs = require('fs');
+var path = require('path');
 
 function NodeActionRunner(whisk) {
     // Use this ref inside lambdas etc.
@@ -47,26 +50,41 @@ function NodeActionRunner(whisk) {
 
         this.userScriptName = name;
 
-        // Loading the user code.
-        return new Promise(
-            function (resolve, reject) {
-                try {
-                    eval(message.code);
-                    thisRunner.userScriptMain = eval(message.main);
-
-                    if (typeof thisRunner.userScriptMain === 'function') {
-                        // The value 'true' has no special meaning here;
-                        // the successful state is fully reflected in the
-                        // successful resolution of the promise.
-                        resolve(true);
-                    } else {
-                        throw "Action entrypoint '" + message.main + "' is not a function.";
-                    }
-                } catch (e) {
-                    reject(e);
-                }
+        function assertMainIsFunction() {
+            if (typeof thisRunner.userScriptMain !== 'function') {
+                throw "Action entrypoint '" + message.main + "' is not a function.";
             }
-        );
+        }
+
+        // Loading the user code.
+        if(message.binary) {
+            // The code is a base64-encoded zip file.
+            return unzipInTmpDir(message.code).then(function (moduleDir) {
+                try {
+                    thisRunner.userScriptMain = eval('require("' + moduleDir + '").' + message.main);
+                    assertMainIsFunction();
+                    // The value 'true' has no special meaning here;
+                    // the successful state is fully reflected in the
+                    // successful resolution of the promise.
+                    return true;
+                } catch (e) {
+                    return Promise.reject(e);
+                }
+            }).catch(function (error) {
+                return Promise.reject(error);
+            });
+        } else {
+            // The code is a plain old JS file.
+            try {
+                eval(message.code);
+                thisRunner.userScriptMain = eval(message.main);
+                assertMainIsFunction()
+                // See comment above about 'true'; it has no specific meaning.
+                return Promise.resolve(true);
+            } catch (e) {
+                return Promise.reject(e);
+            }
+        }
     }
 
     // Returns a Promise with the result of the user code invocation.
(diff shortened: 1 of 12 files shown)


psuter:
PG2 381 is blue-ish

bjustin-ibm:
It might be nice to supply a [Yeoman generator](http://yeoman.io/learning/index.html) to help generate OpenWhisk actions.

bjustin-ibm:
Thinking longer term, it might be advisable to support .tgz files produced with the `npm pack` command. `npm pack` is smart enough to include "bundled" dependencies and exclude files that are not needed for production use (automated tests, build files, README, etc.).

This would require that `npm install` is run on the .tgz each time, however, if all the required dependencies are bundled, this command is essentially equivalent to just unpacking the .tgz. However, if all the required dependencies are not bundled, the installation will still work (albeit slower due to the overhead of fetching dependencies). At this point, using `bundledDependencies` is an optimization we can strongly encourage to the user through documentation, the aforementioned Yeoman generator, and even log messages when running actions that did not include all the required deps.

core/nodejsActionBase/runner.js
```diff
@@ -108,6 +127,55 @@ function NodeActionRunner(whisk) {
...
+    // decompress it into temporary directory, and return the name of that directory.
+    // Note that this makes heavy use of shell commands because:
+    //   1) Node 0.12 doesn't have many of the useful fs functions.
+    //   2) We know in which environment we're running.
+    function unzipInTmpDir(base64) {
+        var mkTempCmd = "mktemp -d XXXXXXXX";
+        return exec(mkTempCmd).then(function (tmpDir1) {
+            return new Promise(
+                function (resolve, reject) {
+                    var zipFile = tmpDir1 + "/action.zip";
```
bjustin-ibm: Using the `path` module to join paths is ever-so-slightly nicer:
```javascript
var zipFile = path.join(tmpDir1, 'action.zip')
```

core/nodejsActionBase/runner.js
```diff
@@ -47,26 +50,42 @@ function NodeActionRunner(whisk) {
...
+                return Promise.reject("Could not uncompress action archive.");
+            });
+        } else {
+            // The code is a plain old JS file.
+            try {
+                eval(message.code);
+                thisRunner.userScriptMain = eval(message.main);
+                assertMainIsFunction()
+                // See comment above about 'true'; it has no specific meaning.
+                return Promise.resolve(true);
```
bjustin-ibm: 😒

core/nodejsActionBase/runner.js
```diff
@@ -47,26 +50,42 @@ function NodeActionRunner(whisk) {
...
+            } catch (e) {
+                return Promise.reject(e);
```
psuter: I did not know that! :smiley: Thanks for the education.

core/nodejsActionBase/runner.js
```diff
@@ -108,6 +127,55 @@ function NodeActionRunner(whisk) {
...
+                });
+            }
+        );
+    }
+
+    // Not exactly correct as it doesn't check the number of ='s at the end.
+    var base64Regex = /^([0-9a-zA-Z+/])*=*$/;
+    function isBase64(str) {
+        return base64Regex.test(str);
+    }
```
psuter: Sounds good.

core/nodejsActionBase/runner.js
```diff
@@ -21,6 +21,9 @@
  * This file (runner.js) must currently live in root directory for nodeJsAction.
  */
 var util = require('util');
+var child_process = require('child_process');
+var fs = require('fs');
+var path = require('path');
```
psuter: Out of curiosity, why is that better? Is it a loading time issue?

bjustin-ibm: Exactly. `require` immediately goes out to disk reads the file and sticks it in memory. Lazy loading modules saves both IO time and memory usage in paths where you don't call the `require`.

psuter: It's a delicate balance then.. Assuming a warm js container, `require`ing at the top saves some latency when initializing a zip-based action. Assuming a cold container, it wastes some time for non-zip actions?

tests/src/actionContainers/NodeJsActionContainerTests.scala
```diff
@@ -434,4 +436,60 @@ class NodeJsActionContainerTests extends BasicActionRunnerTests with WskActorSys
...
+            c.init(initPayload(code))._1 should not be (200)
+        }
+
+        // Somewhere, the logs should mention the connection to the archive.
+        checkStreams(out, err, {
+            case (o, e) =>
+                (o + e).toLowerCase should include("error")
+                (o + e).toLowerCase should include("archive")
+        })
+    }
```
psuter: Yes

psuter:
The way I see it, supporting `.tgz` could be part of this PR, but supporting `npm install` seems beyond its scope.

core/nodejsActionBase/runner.js
bjustin-ibm: I believe what you are saying is correct. My gut feeling is that in all cases, we should delay this loading until we know we need it. If that initialization happens to get re-used by a warm container, then that is even better.

bjustin-ibm:
> The way I see it, supporting .tgz could be part of this PR, but supporting npm install seems beyond its scope.

I agree, and I even think that supporting `.tgz` without `npm install` may be more confusing than valuable. In this way, we may be able to say `.zip` is for completely self-contained archives, and `.tgz` is for using with `npm install`.

psuter:
Addressed the `binary` marker (#1391 needs to be merged first).

Improved the error messages for the various failure modes of uploading a broken/incomplete zip file.

Added tests for the above.

psuter:
PG2 415

rabbah:
Merged as https://github.com/openwhisk/openwhisk/compare/cb02a8850a4e...b36fd6faafbe.

PR #1364, opened 2016-10-05
update documentation of throttling limits

sjfink:


csantanapr:
Fixes #1365

rabbah:
As noted in the issue, the limits documented here are specific to bluemix and not local development.

csantanapr:
So it means this PR is invalid on this repo then

csantanapr:
I see we removed the hour limit, is this true only for bluemix but not for openwhisk?

rabbah:
no more hour limit anywhere.

csantanapr:
Can we make the limits the same?
I know we have throttle tests, but they are unreliable anyway they failed a lot

markusthoemmes:
The limits are configurable for anyone deploying OpenWhisk. Those tests would 1. take muuuuuch longer and 2. be way more unreliable with the same limits configured everywhere. It's not sustainable to test the same limits in dev envs. as in highly scalable prod. envs.

csantanapr:
Maybe the tests need to be disable by default and only to be use in a system integration test with were there could be more control over the environment.

In a dev environment functional tests are the ones that are most important, testing limits come into play when deploying into a test or production environment, not when doing local development I should not have any limits when I run OpenWhisk in my laptop.

We could set the limit only for travis and keep running the tests only in travis CI.

rabbah:
These tests are the equivalent of a mini load test - they can exhibit some pathologies even with reduced limits that are important to test when working on the invoker. We should not exclude them for local development.

csantanapr:
this one was superseded by #1376

Issue #1322, opened 2016-09-23
CLI omission of empty but expected fields from activation is confusing


rabbah:
```
> wsk action invoke /whisk.system/utils/echo -b
ok: invoked /whisk.system/utils/echo with id ...
{
    "namespace": "guest",
    "name": "echo",
    "version": "0.0.1",
    "subject": "guest",
    "activationId": "...",
    "start": 1474659304974,
    "end": 1474659304995,
    "response": {
        "status": "success",
        "success": true
    },
    "annotations": [
        {
            "key": "limits",
            "value": {
                "timeout": 60000,
                "memory": 256,
                "logs": 10
            }
        },
        {
            "key": "path",
            "value": "whisk.system/utils/echo"
        }
    ]
}
```

missing are the `logs` and the `result` where the former is `[]` and the latter is `{}` (and should be present to show that the result of the action was in fact an empty object).

PR #1373, opened 2016-10-06
Do Not Omit Fields from Activation Records

dubeejw:
- Always display fields in an activation record even if their values are empty
- Fixes https://github.com/openwhisk/openwhisk/issues/1322


dubeejw:
PG2 383.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/395597e3e9c7b93e005a7c51137e3f90d5c21594.

PR #1375, opened 2016-10-06
Consistently omit logs for blocking calls.

perryibm:
Note all active acks unconditionally in kafka handler.  PG2-353


markusthoemmes:
Merging this will hide a lot of failures we had recently in testcases which are due to active ack timeout. Shall we adjust these tests to timeout appropriatly, so that information is not lost and the problem hidden?

lehoanganh:
I think for UX it's a good thing. Either we provide logs or no logs at all. Not sometimes with sometimes without. It's a strange experience. If user needs logs, `activation get` will do the job.

rabbah:
@perryibm will add explicit tests for active ack - the test failures we had were doing it indirectly.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/fdf5a7f2cb67cdba8d2a9fca28b7c43d5664b84a.

PR #1384, opened 2016-10-09
Remove kafka monitor

markusthoemmes:
On our way to a more standardized Kafka image.

- Remove slf4j logger from kafka
- Remove kafka from isAlive and PingTest
- Precreate all topics needed, don't rely on auto-create


rabbah:
auto-create is still on?

markusthoemmes:
PG1 #683 is blueish.
Yes auto-create still on for the moment. You want it completely off for this PR?

rabbah:
No. Just confirming.

ansible/roles/kafka/tasks/deploy.yml
```diff
@@ -28,19 +27,25 @@
...
+  shell: docker logs kafka
   register: result
-  until: (result.status == 200) and (result.json.msg == 'yes')
-  retries: 36
-  delay: 5
+  until: ('[Kafka Server 0], started' in result.stdout)
+  retries: 10
+  delay: 2
+
+- name: create the active-ack topic
```
rabbah: in the words of george pig: dinosaur!

tools/health/isAlive
```diff
@@ -85,13 +85,6 @@ def isElkAlive() :
     okString = '"cluster_name"'
     return monitorUtil.outputChecker(cmd, okString, 1)
 
-def isKafkaAlive() :
```
rabbah: i've discussed this with @markusthoemmes - while it appears we don't use this method anymore, it does make `isAlive` inconsistent in that all the other components are known to this script.

markusthoemmes: I believe this is the start of dumping it. As stated in our discussion I believe `consul` should hold the state of all components consistently, which won't need to rely on `isAlive`. I'll continue the dump of it.

services/kafka/Dockerfile
```diff
@@ -32,27 +25,20 @@ ENV KAFKA_RELEASE_ARCHIVE kafka_${KAFKA_SCALA_VERSION}-${KAFKA_RELEASE}.tgz
...
-RUN cd /kafka/libs/ && wget --no-verbose https://jcenter.bintray.com/org/slf4j/slf4j-log4j12/1.7.6/slf4j-log4j12-1.7.6.jar
 ADD config /kafka/config
 ADD start.sh /start.sh
 
 # Set up a user to run Kafka
 RUN chmod uog+rx /start.sh && \
   groupadd kafka && \
   useradd -d /kafka -g kafka -s /bin/false kafka && \
-  chown -R kafka:kafka /kafka /data /logs /monitor
+  chown -R kafka:kafka /kafka /data /logs
```
markusthoemmes: Nope, this `/logs` is used by Kafka itself. See https://hub.docker.com/r/ches/kafka/~/dockerfile/ (which is the Dockerfile which i believe was used as the base for ours and is the one I replaced ours in #1395 )

tools/health/isAlive
perryibm: So what's the plan to get the new kafka's component into consul?  A separate kafka checker component?

markusthoemmes: For the record: Consul didn't even use `isAlive` before.

Still a bit undecided on how to get consul to properly check for Kafka's health, but I think that's a separate issue even? Maybe `controller` and `invoker` expose their Kafka connectivity as part of their `/ping`?

What I'm saying is: This is probably no worse than before in terms of visibility of Kafka's health.

perryibm:
LGTM.

1. I agree that log checking is better than what we had before and topic creation should fail if it's not really up.
2. However, we should turn off auto-create in the next Kafka PR so that we can be sure it's really doing what we expect.  Test topic creation could be moved to the testing task or by rewriting the unit test.

markusthoemmes:
PG2 421 reapproved.

markusthoemmes:
PG2 439 reapproved after a rebase.

perryibm:
rebased/merged

PR #675, opened 2016-06-16
Go CLI binary file compression

mdeuser:
Compress Go CLi binaries for faster downloads.  See #577

For all architectures:
Windows: wsk.exe -> wsk.exe.zip
Mac:     wsk     -> wsk.tar.gz
Linux:   wsk     -> wsk.tar.gz

PR #1388, opened 2016-10-12
Use docker_container instead of docker (deprecated)

markusthoemmes:
This depends on #1378. Due to issues with ansible features @domdom82 might want to block this for a bit. Leaving for him to decide how to proceed.


domdom82:
We will have to hop on to docker_container anyway, even though it does not yet support all features. Hopefully they will fix https://github.com/ansible/ansible-modules-core/issues/5054 at some point soon. The old docker module is a dead-end anyway.

markusthoemmes:
PG1 #675 greeny beeny.

PR #1391, opened 2016-10-12
Added a `binary` field to action records

psuter:
Only applies to actions that hold code (i.e. not blackbox or sequence).

Will be useful for clients that need to determine, e.g., whether the code can be displayed to the user.

Stepping stone for a requested change in #1357.

Additional refactorings to how `Exec`s are represented.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/Attachments.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Attachments.scala
@@ -23,6 +23,8 @@ import akka.http.scaladsl.model.ContentType
 import spray.json._
 import spray.json.DefaultJsonProtocol._
 
+import whisk.core.entity.size._
+
 object Attachments {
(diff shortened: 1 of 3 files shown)


psuter:
PG2 409

dubeejw:
@psuter, will changes to the CLI be required for these changes?

psuter:
No. The idea is to signal the binary-ness, but not require clients to know what they're sending. In #1357 I'll detect the only non-obvious case (JS actions) at creation time.

psuter:
(When we, eventually, move to API v2, we will also be able to use the `Content-type` header to determine the nature of actions being created. The CLI will need support for that then.)

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/cb02a8850a4ec38ed9070dbcb8667c719c668308.

PR #1395, opened 2016-10-13
Dockerhub kafka

markusthoemmes:
I believe our own Kafka image was based on this anyway.

- Removed `redo` target
- Removed Kafka from settings.gradle
- Removed the `services` folder :tada:

Builds on #1384 which should be merged first. Split the PR's to not conflate any errors that *might* occur with two changes to Kafka at once.


tools/health/isAlive
```diff
@@ -85,13 +85,6 @@ def isElkAlive() :
     okString = '"cluster_name"'
     return monitorUtil.outputChecker(cmd, okString, 1)
 
-def isKafkaAlive() :
```
rabbah: if we remove this we have no way to check if kafka is alive and responding. 
can we replace this with a producer and consumer pair on a dedicate health topic?

markusthoemmes: Why do we need a way to check that? I do believe our testsuite is making sure enough that Kafka is fine?

ansible/roles/kafka/tasks/deploy.yml
```diff
@@ -25,22 +24,28 @@
...
+  retries: 10
+  delay: 2
+
+- name: create the active-ack topic
+  shell: "docker exec kafka bash -c 'unset JMX_PORT; kafka-topics.sh --create --topic {{ item }} --replication-factor 1 --partitions 1 --zookeeper {{ inventory_hostname }}:{{ zookeeper.port }}'"
+  with_items:
+    - completed
+
+- name: create the invoker topics
+  shell: "docker exec kafka bash -c 'unset JMX_PORT; kafka-topics.sh --create --topic invoke{{ item.0 }} --replication-factor 1 --partitions 1 --zookeeper {{ inventory_hostname }}:{{ zookeeper.port }}'"
```
ddragosd: this is interesting. IIRC Kafka [creates a topic automatically](http://kafka.apache.org/documentation.html#basic_ops_add_topic) when data is first published to a non-existing topic, using default settings for replication factor and partitions.

markusthoemmes: Yes that's correct. We experienced problems with that before and I personally always prefer explicit topic creation over "magic".

Did you see any 5 minute "lag" problems in your docker-compose PR, where the invoker only gets messages after Kafka has rebalanced? I've got those big times and the only real solution was to create the topics explicitly in advance. Any idea where the culprit might be?

rabbah:
pg?

rabbah:
what are the default logger settings for the image?

markusthoemmes:
PG1 736

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/2b5653944ff7b58be3e59bf324755b917d4e760e.

PR #433, opened 2016-05-13
Fixes issues #413 and #423

rabbah:
Improve couch docs. Fixes issues #413 and #423 by documenting the requirements.

Issue #412, opened 2016-05-13
Update user documentation for Swift3 Actions


csantanapr:
Update user level docs in [docs/actions.md](https://github.com/openwhisk/openwhisk/blob/master/docs/actions.md) for swift3 actions including any new CLI flags to choose between swift@2 or swift@3

csantanapr:
@paulcastro can you drive this completion bringing other folks for help as you need feet, including testing the docs steps

PR #1397, opened 2016-10-13
Issue #1350 Make Swift 3 default

### LINKED ISSUE #1350: Make Swift 3 default swift runtime container
At a minimum files with ".swift" should use the swift:3 kind by default in the go CLI.

There is a node:default kind that provides the notion of a default set by OpenWhisk.  We can add a swift:default kind, however this may not provide any value to the developer because of the significant differences between Swift 2 and Swift 3.

paulcastro:
This PR fixes #1350 and #1400 

- adds "swift:default" as a kind and sets the default to "swift:3", update CLI to default to swift:3
- adds test for above
- restores Swift3 Whisk Object Tests

PG #412 is blue.  PG #433 which includes #1400 fix is blue.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/DefaultRuntimeVersions.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/DefaultRuntimeVersions.scala
@@ -20,6 +20,7 @@ trait DefaultRuntimeVersions {
     def resolveDefaultRuntime(kind: String): String = {
         kind match {
             case "nodejs:default" => "nodejs:6"
+            case "swift:default" => "swift:3"
             case anyOther => anyOther
         }
     }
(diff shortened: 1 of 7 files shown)


dubeejw:
@paulcastro, is Swift 3 the only version that we support?

paulcastro:
We support Swift 2 and 3.  Swift 2 is going to be deprecated in the near future.

dubeejw:
With your updates, users will no longer be able to access Swift 2. Perhaps adding a ```swift:2``` tag as well would be ideal since we are deprecating it instead of removing the version.

paulcastro:
True.  They can still set --kind swift though, where "swift" maps to swift 2.

dubeejw:
Ah, I see now. Looks good.

paulcastro:
@dubeejw the Swift 3 Whisk Object fails on Travis, but succeeds in all PG and local builds.  I'll disable the test and open an issue to investigate why it is only failing on Travis.

PR #1403, opened 2016-10-17
Have the CLI base64-encode zipped NodeJS actions

psuter:
This PR makes the CLI treat artifacts ending in `.zip` as `nodejs:6` actions, and properly encodes the archive in base64 (builds on #1357).


tools/cli/go-whisk-cli/commands/action.go
```diff
@@ -544,6 +544,9 @@ func parseAction(cmd *cobra.Command, args []string) (*whisk.Action, bool, error)
       action.Exec.Kind = "swift"
     } else if ext == ".js" {
       action.Exec.Kind = "nodejs:6"
+    } else if ext == ".zip" {
```
psuter: I would be grateful if @dubeejw or @mdeuser could chime in on this change.

mdeuser: The .zip file extension is pretty general purpose.  I think having a hard association with .zip and nodejs:6 is to confining.  Perhaps if .zip is the extension, the `kind` flag must be specified.

rabbah:
I'd like to use this for the dockerskeleton pr.

psuter:
Will be PG2 437

psuter:
PG2 437 is blue

csantanapr:
How do specify in the cli that .zip it's only supported for nodejs kind?

csantanapr:
Will `wsk action create -h` and `wsk action update -h` specify this?

tools/cli/go-whisk-cli/commands/action.go
```diff
@@ -556,14 +556,28 @@ func parseAction(cmd *cobra.Command, args []string) (*whisk.Action, bool, error)
         return nil, sharedSet, err
       }
     } else {
-      whisk.Debug(whisk.DbgError, "Action runtime extension '%s' is not supported\n", ext)
-      errMsg := fmt.Sprintf(
-        wski18n.T("'{{.name}}' is not a supported action runtime",
-          map[string]interface{}{"name": ext}))
+      errMsg := ""
```
mdeuser: Minor comment.
```
    } else {
      errMsg := ""
      if ext == ".zip" {
```
No need for the `errMsg := ""` line.  Just add the `:=` when the errMsg is being created.  Then the `if` statement can be directly joined to the `else` following the existing `else if` flow.

tests/src/system/basic/WskBasicNodeTests.scala
```diff
@@ -62,10 +62,27 @@ class WskBasicNodeTests
             }
     }
 
+    it should "Ensure that zipped actions are encoded and uploaded as NodeJS actions" in withAssetCleaner(wskprops) {
```
dubeejw: Perhaps we could add a negative test that ensure the proper error message is displayed if a file with a .zip extension is passed without the --kind flag.

csantanapr:
Also I don't see a test for when user uses .zip and not nodejs `wsk create action --kind swift foo.zip`

csantanapr:
That's should error if user pass .zip and not nodejs

tests/src/system/basic/WskBasicNodeTests.scala
psuter: Added that.

psuter:
@csantanapr I understand your concern, but note that the behavior is consistent; currently when you specify the `kind` flag you can pass anything you want (e.g. a `.swift` file as a NodeJS action, etc.).

If we want to change that we should probably open a different issue.

Similarly, `wsk action create -h` currently doesn't indicate any connection between extensions and kinds (our online docs do that).

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/84e4ce64a9f7a93dc7499165acffb7e161922b97.

PR #1404, opened 2016-10-18
Use tid -102 for unpausing the container before removing it.

cbickel:
The unpause of the warm container still uses the tid of the activation.


cbickel:
PG1#698 is started

perryibm:
The way we adjust transaction id is by modifying the active transid of the container itself.  We should continue doing this rather than doing a specific override for one of the methods.   Line 278 of ContainerPool shows how this is done (binding a transsaction id to a container).

jeremiaswerner:
review adopted in https://github.com/openwhisk/openwhisk/pull/1417

Issue #1394, opened 2016-10-12
Confusing error in CLI when a blocking invoke fails


markusthoemmes:
Blocking activations fail with a very confusing error.

## Environment details:

any environment

## Steps to reproduce the issue:

1. launch an action that fails with `blocking=true`

## Provide the expected results and outputs:

something like:

```
error: Invoked the action 'myNamespace/myAction', but it errored:
{
...
}
```

## Provide the actual results and outputs:

```
stderr: error: Unable to invoke action 'myNamespace/myAction': The connection failed, or timed out. (HTTP status code 502)
```

mdeuser:
Possibly duplicate of #1353 ?

markusthoemmes:
Partly agree, though #1353 seems to be more geared towards the internals? I think this is just a bug in the response handling in the CLI.

mdeuser:
Which invoker runtime were you using when this issue was encountered?

I'm pretty sure the CLI encountered a 502 HTTP response code (similar to #1353).  Would you please run the same command with the `-d` option and include the CLI output.

markusthoemmes:
Yes the CLI encountered a 502. That's what the controller returns for a blocking invocation that fails with an `application developer error`. It's specified like that in swagger.

mdeuser:
In addition to the `-d` CLI dump, would you also include the action code you used when encountering this issue.  If the code is not nodejs based and is not using a rejected Promise to return error results, then this is a separate issue.

mdeuser:
@dubeejw - I was able to recreate this failure with the following action.  The CLI is apparently not handling the error parsing well when the error response is an object.  The HTTP response successfully returns the error object, but the CLI has a parsing bug that results in the error `error: Unable to invoke action 'whiskrejectobj': The connection failed, or timed out. (HTTP status code 502)`

whiskrejrectobj.js
```
function main(message) {
    return Promise.reject({
        name: 'myname',
        aflag: true,
        reason: 'failure cause'
    });
}
```

markusthoemmes:
Thanks for doing my job and reproducing the error @mdeuser :+1:

mdeuser:
Fixed by #1405

PR #1405, opened 2016-10-18
Display Proper Error Message from Error Object

dubeejw:
- Parse and display error objects properly
- Closes https://github.com/openwhisk/openwhisk/issues/1394

### DIFF
--- a/tests/dat/actions/asyncError.js
+++ b/tests/dat/actions/asyncError.js
@@ -1,0 +1,13 @@
+function main(params) {
+    var str = params.payload;
+
+    return whisk.invoke({
+        name: '!',    // Invalid action name
+        parameters: {
+            payload: str
+        },
+        blocking: true
+    }).then(function (activation) {
+        return {};
+    });
+}
--- a/tests/src/system/basic/WskBasicTests.scala
+++ b/tests/src/system/basic/WskBasicTests.scala
@@ -295,6 +295,17 @@ class WskBasicTests
                 .stderr should include regex (""""error": "This error thrown on purpose by the action."""")
     }
 
+    it should "create and invoke a blocking action resulting in an error response object" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+            val name = "errorResponseObject"
+            assetHelper.withCleaner(wsk.action, name) {
+                (action, _) => action.create(name, Some(TestUtils.getTestActionFilename("asyncError.js")))
+            }
+
+            wsk.action.invoke(name, blocking = true, expectedExitCode = 246)
+              .stderr should include regex (""""error": "name '!' contains illegal characters \(.+\)"""")
+    }
+
     it should "invoke a blocking action and get only the result" in withAssetCleaner(wskprops) {
         (wp, assetHelper) =>
             val name = "basicInvoke"
--- a/tools/cli/go-whisk/whisk/client.go
+++ b/tools/cli/go-whisk/whisk/client.go
@@ -307,10 +307,10 @@ func parseWhiskErrorResponse(resp *http.Response, data []byte, v interface{}) (*
     err := json.Unmarshal(data, whiskErrorResponse)
 
     // Determine if a whisk.error() response was received. Otherwise, the body contents are unknown (#6)
-    if err == nil && whiskErrorResponse.Response.Result.Error != nil {
-        Debug(DbgInfo, "Detected that a whisk.error(\"%s\") was returned\n", *whiskErrorResponse.Response.Result.Error)
+    if err == nil && whiskErrorResponse.Response.Status != nil {
+        Debug(DbgInfo, "Detected that a whisk.error(\"%s\") was returned\n", *whiskErrorResponse.Response.Status)
         errMsg := wski18n.T("The following application error was received: {{.err}}",
-            map[string]interface{}{"err": *whiskErrorResponse.Response.Result.Error})
+            map[string]interface{}{"err": *whiskErrorResponse.Response.Status})
         whiskErr := MakeWskError(errors.New(errMsg), resp.StatusCode - 256, NO_DISPLAY_MSG, NO_DISPLAY_USAGE,
             NO_MSG_DISPLAYED, APPLICATION_ERR)
         return parseSuccessResponse(resp, data, v), whiskErr
@@ -361,13 +361,7 @@ type WhiskErrorResponse struct {
 }
 
 type WhiskErrorResult struct {
-    Status  string              `json:"status"`
-    Success bool                `json:"success"`
-    Result  WhiskErrorMessage   `json:"result"`
-}
-
-type WhiskErrorMessage struct {
-    Error *string `json:"error"`
+    Status  *string              `json:"status"`
 }
 
 func (r ErrorResponse) Error() string {


tests/dat/actions/errorObject.js
```diff
@@ -0,0 +1,17 @@
+function main(params) {
```
rabbah: why is this called errorObject.js (I don't see the connection).

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -295,6 +295,17 @@ class WskBasicTests
...
     }
 
+    it should "create and invoke a blocking action resulting in an error response object" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+            val name = "errorResponseObject"
+            assetHelper.withCleaner(wsk.action, name) {
+                (action, _) => action.create(name, Some(TestUtils.getTestActionFilename("errorObject.js")))
+            }
+
+            wsk.action.invoke(name, blocking = true, expectedExitCode = 246)
```
rabbah: is it the case that for a blocking invoke where the action returns an error response, the CLI was not interpreting the response correctly? if so, the use of whisk.invoke is a red herring. I would simplify the test to use an action that just returns a failed promise.

dubeejw:
@rabbah, the issue from the CLI perspective is that it is not expecting an error object containing and error string. Instead the CLI is looking for the error string in the response object.

Here's the response from an async promise that fails:
```
"response": {
        "statusCode": 0,
        "success": false,
        "result": {
            "error": {
                "error": "The requested resource does not exist. (undefined)"
            }
        }
    }
```

rabbah:
```
> cat t.js 
function main(params) {
   return Promise.reject({error: 'failed activation'});
}
> wsk action update t.js t.js
>  wsk action invoke t.js --blocking
error: Unable to invoke action 't.js': The connection failed, or timed out. (HTTP status code 502)
```

rabbah:
i think this fix is also not correct - i rebased this pr and for the example above:

```
the response is what's expected:

> wsk action invoke t.js --blocking
ok: invoked /_/t.js with id 8906e48bcbe64e64ba45ae85114acd4a
{
    "namespace": "guest",
    "name": "t.js",
    "version": "0.0.9",
    "publish": false,
    "subject": "guest",
    "activationId": "8906e48bcbe64e64ba45ae85114acd4a",
    "start": 1476815251830,
    "end": 1476815251885,
    "response": {
        "statusCode": 0,
        "success": false,
        "result": {
            "error": "failed activation"
        }
    },
    "logs": [],
    "annotations": [
        {
            "key": "limits",
            "value": {
                "logs": 10,
                "memory": 256,
                "timeout": 60000
            }
        },
        {
            "key": "path",
            "value": "guest/t.js"
        }
    ]
}
```

but now for this case:

```
> cat t.js 
function main(params) {
   return Promise.reject({msg: 'failed activation'});
}
```

the result is broken again:

```
> wsk action update t.js t.js
> wsk action invoke t.js --blocking
error: Unable to invoke action 't.js': The connection failed, or timed out. (HTTP status code 502)
```

tools/cli/go-whisk/whisk/client.go
```diff
@@ -315,11 +315,24 @@ func parseWhiskErrorResponse(resp *http.Response, data []byte, v interface{}) (*
...
     } else {
-        Debug(DbgError, "HTTP response with unexpected body failed due to contents parsing error: '%v'\n", err)
-        errMsg := wski18n.T("The connection failed, or timed out. (HTTP status code {{.code}})",
-            map[string]interface{}{"code": resp.StatusCode})
-        whiskErr := MakeWskError(errors.New(errMsg), resp.StatusCode - 256, DISPLAY_MSG, NO_DISPLAY_USAGE)
-        return resp, whiskErr
+        errorResponse := &RequestErrorResponse{}
+        err := json.Unmarshal(data, errorResponse)
+
+        if err == nil && errorResponse.Response.Result.Error.Error != nil {
```
rabbah: the CLI should not be looking inside of `Error` at all. It should treat `Error` as a JSON object that is opaque.

dubeejw: Currently, I'm looking at the Error object to determine if the 502 HTTP error code is real, or some application error occurred.

rabbah: You need to only check if `Response.Result.Error` is defined (this is the invariant that the application failed) - in which case the activation failed. The nested `Error` is not relevant and could be any application specific value. Alternatively you can check the status of the activation, either is it success: true or else it failed.

dubeejw:
PG2 438

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -295,6 +295,17 @@ class WskBasicTests
...
     }
 
+    it should "create and invoke a blocking action resulting in an error response object" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+            val name = "errorResponseObject"
+            assetHelper.withCleaner(wsk.action, name) {
+                (action, _) => action.create(name, Some(TestUtils.getTestActionFilename("asyncError.js")))
+            }
+
+            wsk.action.invoke(name, blocking = true, expectedExitCode = 246)
```
rabbah: should change 246 to use one of the consts defined in TestUtils (i'll make this change as I merge).

tools/cli/go-whisk/whisk/client.go
```diff
@@ -307,10 +307,10 @@ func parseWhiskErrorResponse(resp *http.Response, data []byte, v interface{}) (*
     err := json.Unmarshal(data, whiskErrorResponse)
 
     // Determine if a whisk.error() response was received. Otherwise, the body contents are unknown (#6)
-    if err == nil && whiskErrorResponse.Response.Result.Error != nil {
-        Debug(DbgInfo, "Detected that a whisk.error(\"%s\") was returned\n", *whiskErrorResponse.Response.Result.Error)
+    if err == nil && whiskErrorResponse.Response.Status != nil {
```
rabbah: can you confirm that this will only run when the HTTP status code for the invoke is not 200?

tests/dat/actions/asyncError.js
```diff
@@ -0,0 +1,13 @@
+function main(params) {
+    var str = params.payload;
+
+    return whisk.invoke({
+        name: 'wc',
```
rabbah: can this be a random string? :)

rabbah: since the point is for this to fail, change `'wc'` to an invalid action name: `'#'`.

rabbah: i still think it's unnecessary to test with whisk.invoke because the issue was general to any failed promise but will acquiesce.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -295,6 +295,17 @@ class WskBasicTests
...
+              .stderr should include regex (""""error": "The requested resource does not exist. \(undefined\)"""")
```
rabbah: note that the `undefined` here is a bug (see #1421). The test will need to allow for `\d+` instead (to be fixed when 1421 is fixed).

dubeejw:
@rabbah, any other comments on this one?

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/37e5857002a54741ccd43dba618bc1ba1dd088ef.

PR #1407, opened 2016-10-18
Extend load balancer to track activations.  Switch activation throttler from consul to this.

perryibm:
The first commit extends data structures in the load balancer to track activations in different ways.  We may twiddle these data structures in the future as we enhance load balancing strategy.  The important thing is that we consistently add/remove ActivationEntry into/out of these data structures.  Note that though there are multiple maps, they share the same underlying ActivationEntry.  Because we limit the global number of activations accepted by the controller, these structures are limited.

The second commit exposes an endpoint from the load balancer for Activation Throttle to use so that it no longer has to consult consul for per-invoker and per-subject activation counts which is a big dictionary that is hurting our performance since we must collect it even for inactive subjects.

Note:  I am deliberately deferring the following from this PR for a followup:  1) Remove the unused endpoints from the load balancer 2) Stop invoker from putting these values to consul.

PG2-419 is green.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
@@ -61,6 +61,13 @@ trait LoadBalancer {
     def getIssuedUserActivationCounts: Map[String, Long]
 
     /**
+     * Retrieves a per subject map of counts representing in-flight activations as seen by the load balancer
+     *
+     * @return a map where the key is the subject and the long is total issued activations by that user
+     */
+    def getActiveUserActivationCounts: Map[String, Long]
+
+    /**
      * Publishes activation message on internal bus for the invoker to pick up.
      *
      * @param msg the activation message to publish on an invoker topic
@@ -106,6 +113,13 @@ class LoadBalancerService(config: WhiskConfig, verbosity: LogLevel)(
     override def getIssuedUserActivationCounts: Map[String, Long] = userActivationCounter.toMap mapValues { _.cur }
 
     /**
+     * Retrieves a per subject map of counts representing in-flight activations as seen by the load balancer
+     *
+     * @return a map where the key is the subject and the long is total issued activations by that user
+     */
+    override def getActiveUserActivationCounts: Map[String, Long] = activationBySubject.toMap mapValues { _.size.toLong }
+
+    /**
      * Publishes message on kafka bus for the invoker to pick up.
      *
      * @param msg the activation message to publish on an invoker topic
@@ -143,10 +157,10 @@ class LoadBalancerService(config: WhiskConfig, verbosity: LogLevel)(
      * The promise value represents the obligation of writing the answer back.
      */
     case class ActivationEntry(id: ActivationId, subject: String, invokerIndex: Int, created: Instant, promise: Promise[WhiskActivation])
-    type ActivationSet = TrieMap[ActivationEntry, Unit]
+    type TrieSet[T] = TrieMap[T, Unit]
     private val activationById = new TrieMap[ActivationId, ActivationEntry]
-    private val activationByInvoker = new TrieMap[Int, ActivationSet]
-    private val activationBySubject = new TrieMap[String, ActivationSet]
+    private val activationByInvoker = new TrieMap[Int, TrieSet[ActivationEntry]]
+    private val activationBySubject = new TrieMap[String, TrieSet[ActivationEntry]]
 
     private val kv = new ConsulClient(config.consulServer)
     private val reporter = new ConsulKVReporter(kv, 3 seconds, 2 seconds,
@@ -198,8 +212,8 @@ class LoadBalancerService(config: WhiskConfig, verbosity: LogLevel)(
         val response = msg.response
         activationById.remove(aid) match {
             case Some(entry @ ActivationEntry(_, subject, invokerIndex, _, p)) =>
-                activationByInvoker.getOrElseUpdate(invokerIndex, new ActivationSet).remove(entry)
-                activationBySubject.getOrElseUpdate(subject, new ActivationSet).remove(entry)
+                activationByInvoker.getOrElseUpdate(invokerIndex, new TrieSet[ActivationEntry]).remove(entry)
+                activationBySubject.getOrElseUpdate(subject, new TrieSet[ActivationEntry]).remove(entry)
                 p.trySuccess(response)
                 info(this, s"processed active response for '$aid'")
             case None =>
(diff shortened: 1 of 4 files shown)


core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -203,36 +222,49 @@ class LoadBalancerService(config: WhiskConfig, verbosity: LogLevel)(
...
-                    if (failedit) info(this, "active response timed out")
+                activationById.get(activationId).foreach { _ =>
+                    if (promise.tryFailure(new ActiveAckTimeout(activationId))) {
+                        info(this, "active response timed out")
+                    }
                 }
             }
-
-            ActivationEntry(Instant.now(Clock.systemUTC()), promise)
+            ActivationEntry(activationId, subject, invokerIndex, Instant.now(Clock.systemUTC()), promise)
```
markusthoemmes: `Clock.systemUTC()` seems unnecessary, as https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#now-- already states its using systemUTC by default.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -142,8 +156,11 @@ class LoadBalancerService(config: WhiskConfig, verbosity: LogLevel)(
      * A map storing current activations based on ActivationId.
      * The promise value represents the obligation of writing the answer back.
      */
-    case class ActivationEntry(created: Instant, promise: Promise[WhiskActivation])
-    private val activationMap = new TrieMap[ActivationId, ActivationEntry]
+    case class ActivationEntry(id: ActivationId, subject: String, invokerIndex: Int, created: Instant, promise: Promise[WhiskActivation])
```
markusthoemmes: Would it make sense to convert all the `String` subjects to proper `Subject`s?

perryibm: yes - agreed

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
rabbah: We use `Instant.now(Clock.systemUTC())` consistently/throughout the core.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
rabbah: should be `Identity` vs `Subject`? is this limit intended for subject or namespace is the operative question.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
markusthoemmes: Then nevermind :)

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -142,8 +156,11 @@ class LoadBalancerService(config: WhiskConfig, verbosity: LogLevel)(
...
+    type ActivationSet = TrieMap[ActivationEntry, Unit]
```
markusthoemmes: As this is pretending to be a `Set`, would it help readability of we implement a tiny stub around this?

Something like:

```
class TrieSet[A] {
	private val backing = new TrieMap[A, Unit]
	
	def add(elem: A) = backing.put(elem, ())
	def remove(elem: A) = backing.remove(elem)
}
```

Even if we don't do that, I'd propose to change this line to:

`type TrieSet[A] = TrieMap[A, Unit]`

Glancing over the code, when reading `ActivationSet` my first thought was that this is a `case class`.

perryibm: I'll update the type def and rename to ActivationEntrySet.

perryibm:
Minor tweaks made.   Additional deferred item is the Identity vs Subject issue which we can do after the consul removal step (this is to avoid additional semantic change in case we need to rollback).

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -203,36 +222,49 @@ class LoadBalancerService(config: WhiskConfig, verbosity: LogLevel)(
...
+      * When invoker health detects a new invoker has come up, this callback is called.
+      */
+    private def invokerChangeCallback(invokerIndices: Array[Int]) = {
+        invokerIndices.foreach { index =>
+            invokerActivationCounter(index) = new Counter()
+            val actSet = activationByInvoker.getOrElseUpdate(index, new TrieSet[ActivationEntry])
+            actSet.keySet map { case ActivationEntry(activationId, subject, invokerIndex, _, promise) =>
+                promise.tryFailure(new LoadBalancerException(s"Invoker $invokerIndex restarted"))
+                activationById.remove(activationId)
+                activationBySubject.remove(subject)
```
markusthoemmes: Is this supposed to kill all the activations from the subject? Shouldn't this just remove the current `ActivationEntry` from the subject's set?

perryibm: Yes, you are right.  Good catch!

perryibm:
Re-running PG2-423.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -103,7 +110,14 @@ class LoadBalancerService(config: WhiskConfig, verbosity: LogLevel)(
...
      * @return a map where the key is the subject and the long is total issued activations by that user
      */
-    override def getUserActivationCounts: Map[String, Long] = userActivationCounter.toMap mapValues { _.cur }
+    override def getIssuedUserActivationCounts: Map[String, Long] = userActivationCounter.toMap mapValues { _.cur }
+
+    /**
+     * Retrieves a per subject map of counts representing in-flight activations as seen by the load balancer
+     *
+     * @return a map where the key is the subject and the long is total issued activations by that user
+     */
```
markusthoemmes: Duplicated ScalaDoc necessary?

perryibm:
Passed PG1-739 (unrelated DB failure)

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
perryibm: Those method will be eliminated or redone on the next PR (dead code and remove consul from invoker side).

PR #1412, opened 2016-10-19
up socket timeout of python action to 120 seconds

domdom82:
python test url is now openwhisk.ng.bluemix.net/api/v1

swift test url is now openwhisk.ng.bluemix.net/api/v1


domdom82:
PG2 430

PR #1415, opened 2016-10-19
Documented JS package actions

psuter:
Documentation only. Should be merged after #1403.

FYI @sjfink

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -281,8 +281,78 @@ This example invokes a Yahoo Weather service to get the current conditions at a 
       "msg": "It is 28 degrees in Brooklyn, NY and Cloudy"
   }
   ```
+
+### Packaging an action as a Node.js module
+
+As an alternative to writing all your action code in a single JavaScript source file, you can write an action as a `npm` package. Consider as an example a directory with the following files:
+
+First, `package.json`:
+
+  ```
+  {
+    "name": "my-action",
+    "version": "1.0.0",
+    "main": "index.js",
+    "dependencies" : {
+      "left-pad" : "1.1.3"
+    }
+  }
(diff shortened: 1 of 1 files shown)


docs/actions.md
```diff
@@ -281,8 +281,76 @@ This example invokes a Yahoo Weather service to get the current conditions at a
...
+  $ zip -r action.zip *
+  ```
+
+3. Create the action:
+
+  ```
+  $ wsk action create packageAction --kind nodejs:6 action.zip
+  ```
+
+  Note that when creating an action from a `.zip` archive using the CLI tool, you must explicitly provide a value for the `--kind` flag.
```
sjfink: Suggest adding a note along the following lines:  

Note: While most npm packages install javascript sources on ```npm install```, some npm packages install and compile binary artifacts.   The zip file upload feature currently does not support binaries uploaded in the zip dependencies, but rather only javascript dependencies.

sjfink:
Added one comment -- otherwise LGTM

docs/actions.md
rabbah: As you know, if such binaries are compiled for a compatible image, they will work in the backend. This could be documented as an advanced feature later (e.g., can do that when I rebase #1166).

psuter:
Added a comment in line with @sjfink's suggestion, hedging it a little to account for @rabbah's scenario.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/eb41402d7dc51cda16744751fc8b343ed79f8916. @csantanapr bluemix doc sync required?

csantanapr:
@rabbah yes, will sync once it's life in production

PR #1424, opened 2016-10-21
upgrade invoker docker client to 1.12.0

domdom82:


domdom82:
PG1 730

core/invoker/Dockerfile
```diff
@@ -1,8 +1,10 @@
 FROM scala
 
 # Uncomment to fetch latest version of docker instead: RUN wget -qO- https://get.docker.com | sh
-# Install docker 1.9.0
-RUN wget --no-verbose -O /usr/bin/docker "https://get.docker.com/builds/Linux/x86_64/docker-1.9.0" && \
+# Install docker 1.12.0 Client
+RUN wget --no-verbose https://get.docker.com/builds/Linux/x86_64/docker-1.12.0.tgz && \
+tar --strip-components 1 -xvzf docker-1.12.0.tgz -C /usr/bin docker/docker && \
+rm -f docker-1.12.0.tgz && \
```
markusthoemmes: Is there a chance that these stay installable like that? So we could have an ENV variable in here

`DOCKER_VERSION=1.12.0` and use it throughout the URLs for an easier upgrade in the future? What do you think?

domdom82: Well I suppose so. They just changed from single file to tar.gz but that was due to the split into multiple binaries. I think it should at least stay as tar.gz for the next couple of releases. I will update the PR accordingly.

markusthoemmes:
LGTM. Leaving for @perryibm for final verification.

perryibm:
PG2-454 had a ton of failures.  Re-running as PG2-456.

domdom82:
finally got a green PG2-470

markusthoemmes:
rebased and merged in 9c6e7d2f77909cf84aa8a0c3f3189791af76b3a0

PR #1425, opened 2016-10-21
Correct timeout doc

ind1go:
This timeout doesn't match the description at the moment.


rabbah:
@csantana I recall changing this to 2s? In any case I think lower timeout is better for the tutorial.

ind1go:
Cool, I can change both places down to 2000ms if we're happier with that.

(sorry for the close/open/close/... history of this pull request - Travis is not happy when there's a worldwide DNS DDOS attack!)

rabbah:
Thanks - i think 2secs is enough.

ind1go:
Reverted to use 2s in the code, and changed the doc to match.

csantanapr:
@ind1go LGTM, thanks for the PR

PR #1430, opened 2016-10-24
Remove invoker tracking/publishing per-subject activation counts.

perryibm:
Switch to information already available in load balancer's new activation maps.  This is a followup to the activation maps we recently begin maintaining.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
@@ -406,7 +406,7 @@ class Invoker(
             InvokerKeys.hostname(instance),
             InvokerKeys.start(instance),
             InvokerKeys.status(instance),
-            { index => Map(InvokerKeys.activationCount(instance) -> activationCounter.cur.toJson) })
+            { _ => Map() })
 
     setVerbosity(verbosity)
 }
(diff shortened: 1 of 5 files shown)


core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -418,39 +398,15 @@ class Invoker(
...
+    new ConsulKVReporter(consul, 3 seconds, 2 seconds,
             InvokerKeys.hostname(instance),
             InvokerKeys.start(instance),
             InvokerKeys.status(instance),
-            { index =>
-                (if (index % 5 == 0) getUserActivationCounts() else Map[String, JsValue]()) ++
-                    Map(InvokerKeys.activationCount(instance) -> activationCounter.cur.toJson)
-            })
-    }
+            { index => Map(InvokerKeys.activationCount(instance) -> activationCounter.cur.toJson) })
```
markusthoemmes: What do we need this for?

perryibm: So it's used in global limits for the whole system.  We could either choose to do this in this PR as well (or not).

perryibm:
PG2-478 true green.

Issue #1421, opened 2016-10-20
transaction code missing when using whisk.invoke in nodejs actions


rabbah:
```
> cat t.js 
function main(params) {
    var str = params.payload;

    return whisk.invoke({
        name: '!',
        parameters: {
            payload: str
        },
        blocking: true
    }).then(function (activation) {
        return {};
    });
}
```

```
> wsk action invoke t.js -b
...
"result": {
            "error": {
                "error": "name '!' contains illegal characters (undefined)"
            }
        }
...
```

Note the missing code in the error message.

rabbah:
probably true for whisk.trigger also (did not confirm).

Issue #1353, opened 2016-09-30
whisk.invoke() does not correctly handle errors returned via promise reject()


mdeuser:
## Environment details:

* local deployment, vagrant, ubuntu VM; also Bluemix

## Steps to reproduce the issue:

Create two actions.  First create two js files -  a callee.js and a caller.js.  Be sure to replace the NAMSPACE related values in caller.js

callee.js
```
function main (message) {
    return Promise.reject('Callee returns an error');
}
```
caller.js
```
function main(message) {
    return whisk.invoke({
        name: '/YOUR_NAMESPACE/callee',
        blocking: true,
        apiKey: 'YOUR_NAMESPACE_APIKEY'
    });
}
```
2.   `wsk action create callee callee.js`
3.   `wsk action create caller caller.js`
4.   `wsk action invoke caller --blocking`
5.   `wsk action invoke caller`    ### Non-blocking, returns activation id
6.   `wsk activation get ACTIVATIONID`

## Provide the expected results and outputs:

```
...
    "response": {
        "status": "application error",
        "success": false,
        "result": {
            "error": "Callee returns an error"
        }
    },
...
```

## Provide the actual results and outputs:
Blocking call
```
error: Unable to invoke action 'caller': The connection failed, or timed out. (HTTP status code {{.code}})
```
Non-blocking activation results
```
...
    "response": {
        "status": "application error",
        "success": false,
        "result": {
            "error": {
                "error": "undefined (undefined)"
            }
        }
    },
...
```

## Additional information you deem important:
* Possibly related to #1351

PR #1435, opened 2016-10-25
'undefined' is displayed when using whisk.invoke

jasonpet:
'undefined' is being shown for two different scenarios.  It is currently being shown as the transaction code when errors are thrown attempting to invoke another action.
It is also being shown as the error value when the invoked action returns an error via a rejected promise or whisk.error().

Resolves issue #1421 
Resolves issue #1353

### DIFF
--- a/core/nodejsActionBase/src/whisk.js
+++ b/core/nodejsActionBase/src/whisk.js
@@ -264,6 +264,7 @@ function post(packet, logger, acceptStatusCode) {
           logger && logger.info('[whisk]', 'response body', body);
 
           var activation = undefined;
+          var result = (body.response || {}).result;
 
           error = error || undefined;
           response = response || {};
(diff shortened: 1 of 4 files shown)


jasonpet:
pg2#481

core/nodejsActionBase/src/whisk.js
```diff
@@ -272,11 +273,16 @@ function post(packet, logger, acceptStatusCode) {
...
               activation = {
                  activationId: body.activationId, // id always present
-                 result: (body.response || {}).result // may not exist
+                 result: result// may not exist
               };
           } else if (!error) {
               // activation failed, set error to API host error response.
-              error = body.error + ' (' + body.errorCode + ')';
+              if (!body.error) {
+                  error = result.error;
```
dubeejw: Is this error message guaranteed to not be an empty string?

PR #1437, opened 2016-10-25
Add slack in ContainerPoolTest.  Change warning messages in ContainerPool.

perryibm:
Fix ContainerPoolTest to be more lenient and give GC a chance to remove container.  Will run this on many PG to see if bug still shows up as this is a low-frequency heisenbug.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
@@ -222,8 +222,12 @@ class ContainerPool(
     final def getContainer(tryCount: Int, position: Long, key: ActionContainerId, conMaker: () => WhiskContainer)(implicit transid: TransactionId): ContainerResult = {
         val positionInLine = position - completedPosition.cur // Indicates queue position.  1 means front of the line
         val available = slack()
-        if (tryCount % 100 == 0) {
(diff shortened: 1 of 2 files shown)


core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -222,8 +222,10 @@ class ContainerPool(
     final def getContainer(tryCount: Int, position: Long, key: ActionContainerId, conMaker: () => WhiskContainer)(implicit transid: TransactionId): ContainerResult = {
         val positionInLine = position - completedPosition.cur // Indicates queue position.  1 means front of the line
         val available = slack()
-        if (tryCount % 100 == 0) {
-            warn(this, s"""getImpl possibly stuck because still in line:
+        val warnAtCount = 200
+        val waitDur = 50.millis
+        if (tryCount == warnAtCount) {
```
perryibm: Maybe not but a constant stream of messages is not useful either.  If the queue is long, you are supposed to wait a while so that's not necessarily an error.

perryibm:
PG1-756

markusthoemmes:
rebased and merged in ad96dc4f2212344c8e8959f717c98f4dc5b89fc5

PR #1438, opened 2016-10-25
Add flush statements to Swift 3 runner

### LINKED ISSUE #1434: Add flush to stdout that is not addressed by "python -u" for Swift 3 runtime
Flushing output in Python proxies is addressed in #518, #1038, however its possible that the solution does not affect the behavior of subprocess.Popen(), which may still buffer its output streams.  We have noticed that sometimes there are no logs even though the compiler is printing to stderr.  As a possible solution we can add flush statements

paulcastro:
This PR fixes #1434 - add flush statements to the swift 3 runner to force output to stdout and stderr. PG2 476

### DIFF
--- a/core/swift3Action/swift3runner.py
+++ b/core/swift3Action/swift3runner.py
@@ -43,9 +43,11 @@ class Swift3Runner(ActionRunner):
 
         if o is not None:
             sys.stdout.write(o)
+            sys.stdout.flush()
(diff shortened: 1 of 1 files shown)


bjustin-ibm:
PG2 488

paulcastro:
PG2 492 blue

PR #1440, opened 2016-10-26
WIP - Controller routes for API Gateway

nwspeete-ibm:
@rabbah, @mdeuser. Three points for discussion in this work item:
1. What logging should be added to these new controller pieces
2. How to check if the action exists from the controller
3. What the unit tests should actually test for the controller pieces added in this PR (Mark and I are planning on end to end testing for API Gateway in a separate place)

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/GatewayRoutesAPI.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/GatewayRoutesAPI.scala
@@ -1,0 +1,199 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.controller
+
+import spray.json._
+import DefaultJsonProtocol._
+import scala.concurrent._
+import spray.http._
+import akka.actor.ActorSystem
+import spray.http.StatusCodes.OK
+import spray.http.StatusCodes.OK
+import spray.httpx.SprayJsonSupport._
+import spray.routing.Directive.pimpApply
+import spray.routing._, Directives._
+import spray.httpx.SprayJsonSupport._
+import whisk.common.TransactionId
+import whisk.core.WhiskConfig
+import akka.event.Logging.LogLevel
+import spray.httpx.SprayJsonSupport.sprayJsonMarshaller
+import spray.httpx.marshalling.ToResponseMarshallable.isMarshallable
+import spray.http.StatusCodes._
+import whisk.core.entity._
+import whisk.common.{  Logging }
+import scala.util.Success
+import whisk.core.entity.EntityPath
+import whisk.core.entity.Identity
+import scala.concurrent._
+import scala.concurrent.{ Future , Promise }
+import scala.concurrent.duration._
+import scala.util.{Failure,Success}
+import whisk.core.entity.WhiskAuth
+import whisk.core.entity.Subject
+import spray.http._
+import spray.client.pipelining._
+import akka.actor.ActorSystem
+import scala.concurrent.Future
+import spray.http.BasicHttpCredentials
+import scala.concurrent.ExecutionContext.Implicits.global
+import spray.http.Uri
+import spray.http.Uri.Path
+import whisk.common.TransactionId
+import whisk.core.WhiskConfig
+import whisk.core.entity._
+import whisk.core.entity.{WhiskAuthStore}
+import akka.event.Logging.{ LogLevel }
+
+class gatewayRoutesApi(
+        val apipath: String,
+        val apiversion: String,
+        val verbosity: LogLevel,
+        val config: WhiskConfig,
+        implicit val actorSystem: ActorSystem)
+        extends Logging {
+        protected implicit val authStore = WhiskAuthStore.datastore(config)
+        private val url = Uri(s"http://localhost:${config.servicePort}")
+        private val key = Await.result(WhiskAuth.get(authStore, Subject("whisk.system"),false)(TransactionId.unknown), 5.seconds)
+        private val validCredentials = BasicHttpCredentials(key.authkey.uuid(), key.authkey.key())
+        /**
+        * Handles invoking create action
+        *
+        * @return Future[JsObject] from action result
+        */
+        def invokeActionCreate(requestBody: JsObject ) : Future[JsObject]  = {
+            val actionPath = Path("/api/"+apiversion) / "namespaces" / "whisk.system" / "actions" / "routemgmt" / "createRoute"
+            val pipeline: HttpRequest => Future[HttpResponse] = (
+                addCredentials(validCredentials)
+                   ~> sendReceive
+            )
+            val p = Promise[JsObject]()
+            val response: Future[HttpResponse] = pipeline(Post(url.withPath(actionPath).toString.concat("?blocking=true"),  requestBody)  )
+            response.onComplete({
+                case Success(result: HttpResponse) => {
+                  p trySuccess result.entity.asString.parseJson.asJsObject
+                }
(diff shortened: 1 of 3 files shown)


core/controller/src/main/scala/whisk/core/controller/GatewayRoutesAPI.scala
```diff
@@ -0,0 +1,197 @@
...
+        val config: WhiskConfig,
+        implicit val actorSystem: ActorSystem){
+        protected implicit val authStore = WhiskAuthStore.datastore(config)
+        /**
+        * Handles invoking create action
+        *
+        * @return Future[String] from action result
+        */
+        def invokeActionCreate(requestBody: String ) : Future[String]  = {
+            val url = Uri(s"http://localhost:${config.servicePort}")
```
rabbah: factor our the url construction into a method so you can reuse it in the other methods.

core/controller/src/main/scala/whisk/core/controller/GatewayRoutesAPI.scala
```diff
@@ -0,0 +1,197 @@
...
+            val actionPath = Path("/api/v1") / "namespaces" / "whisk.system" / "actions" / "routemgmt" / "createRoute"
```
rabbah: factor our the route construction as well - parameterize by api version, and the system namespace.

core/controller/src/main/scala/whisk/core/controller/GatewayRoutesAPI.scala
```diff
@@ -0,0 +1,197 @@
...
+            val key = Await.result(WhiskAuth.get(authStore, Subject("whisk.system"),false)(TransactionId.unknown), 5.seconds)
```
rabbah: unless `key` is defined once at initialization, the use of Await.result is strongly discouraged. Should this fail the method will throw an exception. Instead use future chaining.

core/controller/src/main/scala/whisk/core/controller/GatewayRoutesAPI.scala
```diff
@@ -0,0 +1,197 @@
...
+            val validCredentials = BasicHttpCredentials(key.authkey.uuid.toString, key.authkey.key.toString)
```
rabbah: shouldn't these be lifted to the class constructor?
don't use `toString` directly instead of use `uuid()` and `key()`.

core/controller/src/main/scala/whisk/core/controller/GatewayRoutesAPI.scala
```diff
@@ -0,0 +1,197 @@
...
+            val pipeline: HttpRequest => Future[HttpResponse] = (
+                addCredentials(validCredentials)
+                   ~> sendReceive
+            )
+            val p = Promise[String]()
```
rabbah: why using `String` instead of at least `JsObject`? In fact the response on success will be `WhiskActivation`.

core/controller/src/main/scala/whisk/core/controller/GatewayRoutesAPI.scala
```diff
@@ -0,0 +1,197 @@
...
+             })
+             p.future
+       }
+       /**
+        ** Handles invoking get actions
+        **
+        * @return Future[String] from action result
+        */
+        def invokeActionGet(requestBody: String ) : Future[String]  = {
+            val url = Uri(s"http://localhost:${config.servicePort}")
```
rabbah: suggest you avoid all this clone and own code since it's avoidable ;)

core/controller/src/main/scala/whisk/core/controller/GatewayRoutesAPI.scala
```diff
@@ -0,0 +1,197 @@
...
+      /**
+      * Handles the controller's api gateway routes
+      */
+      def routes(user: Identity)(implicit transid: TransactionId) = {
+       lazy val collectionPrefix = pathPrefix( (EntityPath.DEFAULT.toString.r | Segment) / "routes")
+       get {
+         collectionPrefix { segment => {
+           entity(as[String]) { pattern =>
+             parameters("basepath", "operation"?"","relpath"?"") { (basepath, operation,relpath) =>
+               var postBody = "{" + "\"namespace\":" +"\"" + segment + "\"," + "\"basepath\":" +"\"" + basepath + "\"," + "\"operation\":" +"\"" + operation + "\"," + "\"relpath\":" +"\"" + relpath + "\"}";
```
rabbah: don't construct the content as a string - this is error prone. You can use a type that you serialize to Json or use a Map that you can serialize to a JsObject. See the spray.json package.

core/controller/src/main/scala/whisk/core/controller/GatewayRoutesAPI.scala
```diff
@@ -0,0 +1,197 @@
...
+               }
+               complete (OK, invokeActionGet(postBody))
+              }
+            }
+          }
+        }
+       } ~
+       post {
+         collectionPrefix { segment => {
+             entity(as[String]) { pattern =>
```
rabbah: can this be a more specific type?

mdeuser:
As I was writing some tests, the following comment came to mind
- Have the controller validate that the specified API action (not the routemgmt action, but the user action) actually exists and namespace/auth has access to it before calling the routemgmt create action.

core/controller/src/main/scala/whisk/core/controller/GatewayRoutesAPI.scala
```diff
@@ -0,0 +1,191 @@
...
+        **
+        * @return Future[String] from action result
+        */
+        def invokeActionGet(requestBody: String ) : Future[String]  = {
+            val actionPath = Path("/api/"+apiversion) / "namespaces" / "whisk.system" / "actions" / "routemgmt" / "getApi"
+            val pipeline: HttpRequest => Future[HttpResponse] = (
+                addCredentials(validCredentials)
+                    ~> sendReceive
+            )
+            val p = Promise[String]()
```
rabbah: the promises through out are not necessary - use a `map` on the future instead.

core/controller/src/main/scala/whisk/core/controller/GatewayRoutesAPI.scala
```diff
@@ -0,0 +1,191 @@
...
+      /**
+      * Handles the controller's api gateway routes
+      */
+      def routes(user: Identity)(implicit transid: TransactionId) = {
+       lazy val collectionPrefix = pathPrefix( (EntityPath.DEFAULT.toString.r | Segment) / "routes")
+       get {
+         collectionPrefix { segment => {
+           entity(as[String]) { pattern =>
+             parameters("basepath"?"", "operation"?"","relpath"?"") { (basepath, operation,relpath) =>
+               var postBody = "{" + "\"namespace\":" +"\"" + segment + "\"," + "\"basepath\":" +"\"" + basepath + "\"," + "\"operation\":" +"\"" + operation + "\"," + "\"relpath\":" +"\"" + relpath + "\"}";
```
rabbah: use JsObject and construct a proper JsValue instead - this is error prone.

markusthoemmes: Also: drop the `var`. There are several occurences of checks against empty `String`s. From how I read the code, these should be modeled as `Option[String]` instead. Would that make more sense?

core/controller/src/main/scala/whisk/core/controller/GatewayRoutesAPI.scala
rabbah: I suggest doing this change sooner rather than later so that you can avoid accidentally leaking a promise or forgetting to complete it as the code evolves.

rabbah:
@mdeuser what does this mean, can you clarify?

```
    Have the controller validate that the specified API action (not the routemgmt action, but the user action) actually exists and namespace/auth has access to it before calling the routemgmt create action.
```

rabbah:
@nwspeete-ibm can you squash all these commits and then rebase to head?

nwspeete-ibm:
@rabbah squashed the commits, added logging and json objects

mdeuser:
@rabbah When creating the API endpoint, a target action is a required argument.  If would be nice if that action's existence and accessibility by the user credentials invoking the CLI (or REST) were validated prior to configuring the API GW with this action.

nwspeete-ibm:
Note: Will have to clean up unresolved conflicts and files. Delivered WIP unit tests for viewing to tests/src/whisk/core/controller/test/TriggersApiTests.scala

markusthoemmes:
Why all the chmod changes?

rabbah:
Rebased the primary commit and superseding ins #1600.

PR #1441, opened 2016-10-27
Thread poll limit through correctly

markusthoemmes:
This did fallback to the standards set in in `pollFor` which are quite a bit lower than what we have for the outer helpers.


markusthoemmes:
PG2 495

perryibm:
rebased/merged

PR #1442, opened 2016-10-27
Reorder testexecution to properly clean up, wait longer than check time in consul check

markusthoemmes:
- The tests didn't clean up properly, leading to false traces in test debugging.
- Waiting for the correct status should be longer than the status check takes itself. Probably a race condition.


markusthoemmes:
PG2 494 greeny beeny

PR #1444, opened 2016-10-27
Using ansible's local variables to compute db_prefix

markusthoemmes:
`ansible_user_id` correctly yields `docker` in a docker-machine environment, but all other databases are created with my local username due to the host where the specific playbooks are executed. This aims to make everything even.


markusthoemmes:
Doesn't work as expected.

PR #1446, opened 2016-10-27
Run container pool putBack concurrently with acivation completion.

perryibm:
This will lower latency by not stopping the ack based in case putback (docker pause) is slow.

Passes PG1-769.  Awaiting travis.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
@@ -203,10 +203,12 @@ class Invoker(
 
                     val contents = getContainerLogs(con, action.exec.sentinelledLogs, action.limits.logs)
 
-                    // Force delete the container instead of just pausing it iff the initialization failed or the container
-                    // failed otherwise. An example of a ContainerError is the timeout of an action in which case the
-                    // container is to be removed to prevent leaking
-                    pool.putBack(con, failedInit)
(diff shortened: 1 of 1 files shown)


core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -203,10 +203,12 @@ class Invoker(
...
-                    // Force delete the container instead of just pausing it iff the initialization failed or the container
-                    // failed otherwise. An example of a ContainerError is the timeout of an action in which case the
-                    // container is to be removed to prevent leaking
-                    pool.putBack(con, failedInit)
+                    /* Force delete the container instead of just pausing it iff the initialization failed or the container
+                     * failed otherwise. An example of a ContainerError is the timeout of an action in which case the
+                     * container is to be removed to prevent leaking.  Since putting back the container involves pausing,
+                     * we run this in a Future so as not to block transaction completion but also return resources promptly.
+                     */
+                    Future { pool.putBack(con, failedInit) }
```
rabbah: this is OK because the invoker is running with an infinite thread pool - yes?

perryibm: yes - infinite because we throttle the # of live transactions already so we want maximum concurrency beyond that point

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/c12a2b9c78ac0b4d2a14c4ad7ea11fe240919e89.

PR #1453, opened 2016-10-28
Update self-signed certificate.

### LINKED ISSUE #1419: cant import self-sign certificate because it's expired
Trying to use PostMan to invoke an action locally but having problems with SSL.
PostMan points that certificate can be imported into keychain for chrome to pick it up
http://blog.getpostman.com/2014/01/28/using-self-signed-certificates-with-postman/

I was able to get the certificate to import, but it will not import because of the date expiration is back to March 16 2016
image

I think the fix is to generate new certificates with newer date
I think we can re-use the [openwhisk-key.pem](https://github.com/openwhisk/openwhisk/blob/master/ansible/roles/nginx/files/openwhisk-key.pem)
To generate a new [openwhisk-cert.pem](https://github.com/openwhisk/openwhisk/blob/master/ansible/roles/nginx/files/openwhisk-cert.pem)

@cbickel I think your work on something like this recently, can you help generate new certificate?

rabbah:
Fixes #1419.
pg2/772 approved.


csantanapr:
@rabbah let me give it a try

csantanapr:
@rabbah Did you intended to make the certificate to expire next month

> "Expires: Sunday, November 27, 2016 at 1:05:36 PM Eastern" ?

It's what I see on the certificate you added

csantanapr:
I was not able to use it with postman after deploying using local mac.

I think is because the "hostname" doesn't match the CN name "www.github.com\/openwhisk" that it was generated for.

I will try to generate with CN set to the docker-engine host for mac it would be "192.168.99.100"

ansible/roles/nginx/files/genssl.sh
```diff
@@ -0,0 +1,23 @@
...
+## generates a (self-signed) certificate
+
+## uncomment to regenerate the key
+#openssl genrsa -out openwhisk-key.pem 2048
+
+echo generating certificate request
+openssl req -new \
+    -key openwhisk-key.pem \
+    -nodes \
+    -days 365 \
```
csantanapr: The flag `-days 365` belongs in the `x509` command

csantanapr:
Once I generated using CN=192.168.99.100 then I was able to import certificate into keychain, and got Postman working yay 🎉 

image

csantanapr:
@rabbah It's kind of tricky if the CN Name needs to match the hostname being used.

I think we can use `192.168.99.100` as default, then for Vagrant I can adjust the Vagrant script to generate a new one for the vagrant host `192.168.33.13`

rabbah:
The change here will require openssl is installed on the host. 
Looking at docs that need updating.

markusthoemmes:
Wouldn't it also be viable to use `http` in a local deployment to bypass all of this?

rabbah:
yes but that would mean not testing the https path in open, no?

markusthoemmes:
Do we trust nginx enough to be able to do SSL termination?

csantanapr:
rebase and merge c785663ccfaa70741c6b7601b5d0a4deb67593e6

PR #1455, opened 2016-10-28
Enable trigger caching

rabbah:
now that cache properly transacts reads and writes, enable trigger caching again.

pg2/507 and pg1/773 approved.


starpit:
reviewed

Issue #1458, opened 2016-10-31
Swift code does not compile with the latest CLI


trevorbodz:
I was on version 2016-09-XXXXXXXX of the CLI and this worked, upon upgrading to 2016-10-24T11:55:02+00:00 this now does not work.

wsk sdk install iOS

unzip OpenWhiskIOSStarterApp.zip

pod install

wsk package bind /whisk.system/weather myWeather13 -p username XXXXXXX -p password YYYYYY

cd OpenWhiskStarterApp

wsk action create filter13 FilterForecast.swift

wsk action create --sequence myForecast13 myWeather13/forecast,filter13

wsk action invoke -b myForecast13 -p latitude 43 -p longitude -79

wsk activation list filter13

wsk activation get c592ab7864ee4ae9830608759327533e

ok: got activation c592ab7864ee4ae9830608759327533e
{
    "namespace": "whisktest@hotmail.com",
    "name": "filter13",
    "version": "0.0.1",
    "publish": false,
    "subject": "whisktest@hotmail.com",
    "activationId": "c592ab7864ee4ae9830608759327533e",
    "start": 1477931713204,
    "end": 1477931719438,
    "response": {
        "status": "action developer error",
        "statusCode": 0,
        "success": false,
        "result": {
            "error": "The action failed to generate or locate a binary. See logs for details."
        }
    },
    "logs": [
        "2016-10-31T16:35:14.972006904Z stdout: Compiling",
        "2016-10-31T16:35:19.432291828Z stderr: /swift3Action/spm-build/main.swift:101:26: error: incorrect argument label in call (have 'args:', expected 'params:')",
        "2016-10-31T16:35:19.432338463Z stderr: let result = main(args:parsed)",
        "2016-10-31T16:35:19.432347455Z stderr: ^~~~~",
        "2016-10-31T16:35:19.432354969Z stderr: params",
        "2016-10-31T16:35:19.434989076Z stdout: swiftc status is 1",
        "2016-10-31T16:35:19.436192021Z stdout: Action did not compile"
    ],
    "annotations": [
        {
            "key": "limits",
            "value": {
                "logs": 10,
                "memory": 256,
                "timeout": 60000
            }
        },
        {
            "key": "path",
            "value": "whisktest_dev/filter13"
        }
    ]
}

dubeejw:
@trevorbodz, thanks for the information. Looks like the iOS SDK needs to be updated.

paulcastro:
[...]

Issue #1459, opened 2016-11-01
Deprecate Swift 2 Runtime


paulcastro:
Deprecate Swift 2 container tasks:

- remove Swift Action runtime
- update tests
- provide error message if someone tries to create or invoke an action of "swift" kind

This supercedes tasks identified in issue #1133

paulcastro:
I have a build that will refuse to create an action of kind swift (representing swift:2), and will not invoke any existing action of swift.  This is all done at the invoker level, I can look into moving this up to the controller if it makes more sense.

paulcastro:
Redesigned solution so it now works at the Controller.

paulcastro:
merged

PR #1462, opened 2016-11-02
Issue #1458, update iOS starterApp FilterForecast action to work with…

paulcastro:
… Swift 3 runtime.  Fixes Issue #1458 

PG 781 blue-ish, PG 784.

### DIFF
--- a/mobile/iOS/starterapp/OpenWhiskStarterApp/FilterForecast.swift
+++ b/mobile/iOS/starterapp/OpenWhiskStarterApp/FilterForecast.swift
@@ -17,11 +17,11 @@
 
 import Foundation
 
-func main(params:[String:Any]) -> [String:Any] {
+func main(args:[String:Any]) -> [String:Any] {
     let NumDays = 2
     var filteredForecasts = [[String:Any]]()
     #if os(Linux)
-        if let forecasts = params["forecasts"] as? [Any] {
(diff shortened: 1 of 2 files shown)


bjustin-ibm:
@paulcastro This concerns me a bit. If I understand correctly, this code compiles and runs well on the Swift 2 runtime, but not on Swift 3. If that is true, then why does the Swift 3 runtime expect a different label for the `main` function argument? Wouldn't this same change be required for _all_ actions that are being transitioned from the Swift 2 runtime?

paulcastro:
Swift 3 started enforcing argument labels so `main(args: [String:Any])` has to be called using `main(args: ["name":"Joe"])`.  In Swift 2 you didn't have to specify the argument.  We call main in our epilogue.swift so we have to decide on the name parameter.

I recall a proposal for Swift 3 that dropped the argument labels entirely when calling a function but I don't see that in the current release.

bjustin-ibm:
ah... I see now. So this code was using `params` as the label (and possibly getting away with it) when we have been suggesting the `args` label all along. My only concern was if there were lots of existing Swift 2 actions that were using `params`, but that does not actually seem to be the case.

Thanks for clarifying.

paulcastro:
Yes, we will deprecate Swift 2 (issue #1459 )

PR #1473, opened 2016-11-05
Sequence execution

ioana-blue:
The bulk of the work is done. Submitting for review to get some feedback.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Actions.scala
@@ -23,6 +23,7 @@ import scala.concurrent.TimeoutException
 import scala.concurrent.duration._
 import scala.util.Failure
 import scala.util.Success
+import scala.util.Try
 import scala.language.postfixOps
 
 import akka.actor.ActorSystem
@@ -920,6 +921,14 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
         val logs = ActivationLogs(wskActivations map { activation => activation.activationId.toString })
         // compute duration
         val duration = (wskActivations map { activation => java.time.Duration.between(activation.start, activation.end).toMillis }).sum
+        // compute max memory
+        val maxMemoryOption = Try {
+            val memoryLimits = wskActivations map { activation =>
+                activation.annotations("limits").get.asJsObject.getFields("memory")(0).convertTo[Long]
+            }
+            memoryLimits.max
+        } toOption
+        val maxMemory = maxMemoryOption getOrElse 0L
         // create the whisk activation
         val activation = WhiskActivation(
                 namespace = user.namespace.toPath,  // TODO: double-check on this
@@ -936,7 +945,8 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
                 // TODO: annotation on max memory?
                 annotations = Parameters("topmost", JsBoolean(topmost)) ++
                               Parameters("kind", "sequence") ++
-                              Parameters("duration", JsNumber(duration)))
+                              Parameters("duration", JsNumber(duration)) ++
+                              Parameters("limits", JsObject("memory" -> JsNumber(maxMemory))))
         activation
     }
 
--- a/core/controller/src/main/scala/whisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Actions.scala
@@ -803,7 +803,7 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
      * @param payload the dynamic arguments for the activation
      * @param blocking true iff this is a blocking invoke
      * @param components the actions in the sequence
-     * @param
+     * @param cause the id of the activation that caused this sequence (defined only for inner sequences and None for topmost sequences)
      * @param transid a transaction id for logging
      * @return a future of type (ActivationId, Some(WhiskActivation)) if blocking; else (ActivationId, None)
      */
@@ -844,6 +844,10 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
                         // all activations were successful, return the result of the last one
                         Future.successful((seqActivationId, Some(wskActivations.last)))
                     // TODO distinguish between the types of failure (e.g., account for interrupted sequences)
+                    case Failure(t: SequenceInterruptedErrorException) =>
+                        // the execution of the sequence was interrupted, but pass along the result to the enclosing sequences
+                        Future.successful((seqActivationId,
+                                Some(makeUnsuccessfulSeqActivation(user, action, seqActivationId, futureWskActivations, topmost, t.activationResponse, cause))))
                     case Failure(t) => Future.failed(t)
                 }
             }
@@ -855,7 +859,7 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
                 // the response of the sequence is the response of the very last activation
                 val activation = makeSeqActivation(user, action, seqActivationId, wskActivations, topmost, wskActivations.last.response, cause)
                 storeSequenceActivation(activation)
-            case Failure(t: SequenceRetrieveActivationTimeout) =>
+            case Failure(t: SequenceException) =>  // can be either timeout exception or sequence error interrupt
                 val activation = makeUnsuccessfulSeqActivation(user, action, seqActivationId, futureWskActivations, topmost, t.activationResponse, cause)
                 storeSequenceActivation(activation)
             case Failure(t: Throwable) =>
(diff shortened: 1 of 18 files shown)


core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -722,6 +735,251 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+                        val seqActivation = makeSeqActivation(user, action, seqActivationId, wskActivations, topmost, wskActivations.last.response, cause)
+                        (seqActivationId, Some(seqActivation))
+                    }
+                } else {
+                    // non-blocking sequence execution, return activation id
+                    Future.successful((seqActivationId, None))
+                }
+            } else {
+                // not topmost, blocking for sure, wait for the "aggregate" future to finish (no timeout)
+                val seqResult = Await.ready(futureSeqResult, Duration.Inf).value.get
```
rabbah: is there a way to avoid using Await? this blocks the thread and is not guaranteed to be deadlock free. also the Duration.Inf is worrisome.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -722,6 +735,251 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+            cause: Option[ActivationId])(implicit transid: TransactionId): WhiskActivation = {
+        // at this point all futures are finished, filter out the successful ones
+        val failedIdx = futureWskActivations.indexWhere {
+            _.value match {
+                case Some(Failure(_)) => true
+                case _ => false
+            }
+        }
+        // all futures are done, this should return immediately
+        val successfulActivations = Await.result(Future.sequence(futureWskActivations.slice(0, failedIdx)), Duration.Inf)
```
rabbah: See comment about await earlier - your comment suggests all futures are done so why use await?

rabbah:
did you consider moving all the sequence scheduling code to its own trait? (which then you can test for all the corner cases perhaps).

re this: >  remove old sequence code (and fix tests that check for pipe.js specific things)
can we have a feature toggle so that we can back out the switch if needed and then when we’re sure cleanup in a new pr.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -781,6 +796,318 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+                val activation = seqActivation getOrElse {
+                    val wskActivationEithers = wskActivationTuples.map(_._1)
+                    val end = Instant.now(Clock.systemUTC())
+                    // the response of the sequence is the response of the very last activation
+                    makeSequenceActivation(user, action, seqActivationId, wskActivationEithers, topmost, cause, start, end)
+                }
+                storeSequenceActivation(activation)
+            case Failure(t: Throwable) =>
+                // consider this whisk error
+                error(this, s"Sequence activation 'seqActivationId' failed: ${t.getMessage}")
```
psuter: missing `$` before `seqActivationId`

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -781,6 +796,318 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+            user: Identity,
+            action: WhiskAction,
+            activationId: ActivationId,
+            wskActivationEithers: Vector[Either[ActivationResponse, WhiskActivation]],
+            topmost: Boolean,
+            cause: Option[ActivationId],
+            start: Instant,
+            end: Instant) : WhiskActivation = {
+        // extract all successful activations from the vector of activation eithers
+        val leftIndex = wskActivationEithers.indexWhere {
```
psuter: `....indexWhere(_.isLeft)`

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -781,6 +796,318 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+            blocking: Boolean,
+            topmost: Boolean,
+            components: Vector[FullyQualifiedEntityName],
+            cause: Option[ActivationId],
+            atomicActionsCount: Int)(implicit transid: TransactionId): Future[(ActivationId, Option[WhiskActivation], Int)] = {
+        // create new activation id that corresponds to the sequence
+        val seqActivationId = activationIdFactory.make()
+        info(this, s"Invoking sequence $action topmost $topmost activationid '$seqActivationId'")
+        val start = Instant.now(Clock.systemUTC())
+        var seqActivation: Option[WhiskActivation] = None
```
rabbah: this `var` is potentially dangerous as it's racy wrt the future below that assigns it.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -781,6 +796,318 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+            if (topmost) { // need to deal with blocking and closing connection
+                if (blocking) {
+                    val timeout = maxWaitForBlockingActivation + blockingInvokeGrace
+                    val futureSeqResultTimeout = futureSeqResult withTimeout(timeout, new BlockingInvokeTimeout(seqActivationId))
+                    // if the future fails with a timeout, the failure is dealt with at the caller level
+                    futureSeqResultTimeout map { wskActivationTuples =>
+                        val wskActivationEithers = wskActivationTuples.map(_._1)
+                        // the execution of the sequence was successful, return the result
+                        val end = Instant.now(Clock.systemUTC())
+                        seqActivation = Some(makeSequenceActivation(user, action, seqActivationId, wskActivationEithers, topmost, cause, start, end))
```
rabbah: this flow for creating the seqActivation is repeated three times - the only difference is the `end` marker. 
since the `end` marker is not that important for the sequence, is it really worth the code duplication?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -781,6 +796,318 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+        // the execution can reach here without a seqActivation due to non-blocking activations OR blocking activations that reach the blocking invoke timeout
+        // futureSeqResult should always be successful, if failed, there is an error
+        futureSeqResult onComplete {
+            case Success(wskActivationTuples) =>
+                // all activations were successful
+                val activation = seqActivation getOrElse {
+                    val wskActivationEithers = wskActivationTuples.map(_._1)
+                    val end = Instant.now(Clock.systemUTC())
+                    // the response of the sequence is the response of the very last activation
+                    makeSequenceActivation(user, action, seqActivationId, wskActivationEithers, topmost, cause, start, end)
```
rabbah: this can throw an exception in which case no activation will be created - either this should be wrapped with a future itself or a try to allow for recovery.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -781,6 +796,318 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
+                    // non-blocking sequence execution, return activation id
+                    Future.successful((seqActivationId, None, 0))
+                }
+            } else {
+                // not topmost, no need to worry about terminating incoming request
+                futureSeqResult map { wskActivationTuples =>
+                     val wskActivationEithers = wskActivationTuples.map(_._1)
+                    // all activations are successful, the result of the sequence is the result of the last activation
+                    val end = Instant.now(Clock.systemUTC())
+                    seqActivation = Some(makeSequenceActivation(user, action, seqActivationId, wskActivationEithers, topmost, cause, start, end))
```
rabbah: makeSequenceActivation could throw an exception and this will cause the future to complete with an error and not create an activation.

tests/src/system/basic/WskSequenceTests.scala
```diff
@@ -68,23 +79,359 @@ class WskSequenceTests
...
+                    checkSequenceLogsAndAnnotations(activation, 3) // 3 activations in this sequence
+                    activation.cause shouldBe None   // topmost sequence
+                    val result = activation.response.result.get
+                    result.fields.get("payload") shouldBe defined
+                    result.fields.get("length") should not be defined
+                    result.fields.get("lines") shouldBe Some(JsArray(Vector(now.toJson)))
+            }
+    }
+
+    it should "create and run a sequence in a package with parameters" in withAssetCleaner(wskprops) {
```
rabbah: need a test where package is private/owned by other namespace.

rabbah:
closing for #1533.

PR #1481, opened 2016-11-08
Migrate Controller's authentication to the new schema

markusthoemmes:
1. Add the new view
2. Make the Controller read from the view
3. Thread the namespace extracted from the view through the API

The general "_" case now resolves to the namespace infered from the key vs. defaulting to the subject's name. For migration of the API, the "old" keys' namespace default to the subject name which preserves previous behaviour.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/Identity.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Identity.scala
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.entity
+
+import scala.concurrent.Future
+
+import spray.json._
+import types.AuthStore
+import whisk.common.Logging
+import whisk.common.TransactionId
+import whisk.core.database.MultipleReadersSingleWriterCache
+import whisk.core.database.NoDocumentException
+import whisk.core.entitlement.Privilege
+import whisk.core.entitlement.Privilege.Privilege
+
+protected[core] case class Identity(subject: Subject, namespace: EntityName, authkey: AuthKey, rights: Set[Privilege])
+
+object Identities extends MultipleReadersSingleWriterCache[Identity, DocInfo] {
+
+    private val viewName = "subjects/identities"
+
+    override val cacheEnabled = true
+    override def cacheKeyForUpdate(i: Identity) = i.authkey
+
+    def get(datastore: AuthStore, authkey: AuthKey)(
+        implicit transid: TransactionId): Future[Identity] = {
+        implicit val logger: Logging = datastore
+        implicit val ec = datastore.executionContext
+
+        cacheLookup(authkey, {
+            list(datastore, authkey) map { list =>
+                list.length match {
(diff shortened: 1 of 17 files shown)


markusthoemmes:
PG2 554 approved.

common/scala/src/main/scala/whisk/core/entity/Identity.scala
```diff
@@ -0,0 +1,82 @@
...
+    override val cacheEnabled = true
+    override def cacheKeyForUpdate(i: Identity) = i.authkey
+
+    def get(datastore: AuthStore, authkey: AuthKey)(
+        implicit transid: TransactionId): Future[Identity] = {
+        implicit val logger: Logging = datastore
+        implicit val ec = datastore.executionContext
+
+        cacheLookup(authkey, {
+            implicit val ec = datastore.executionContext
```
rabbah: redundant with line 43.

common/scala/src/main/scala/whisk/core/entity/Identity.scala
```diff
@@ -0,0 +1,82 @@
...
+                        val subject = Subject(obj.fields("id").convertTo[String])
+
+                        val value = obj.fields("value").convertTo[Map[String, String]]
+                        val namespace = EntityName(value("namespace"))
+                        val uuid = UUID(value("uuid"))
+                        val key = Secret(value("key"))
+
+                        Identity(subject, namespace, AuthKey(uuid, key), Privilege.ALL)
+                    case 0 =>
+                        logger.info(this, s"$viewName[$authkey] does not exist")
```
rabbah: don't log the authkey.

common/scala/src/main/scala/whisk/core/entity/Identity.scala
```diff
@@ -0,0 +1,82 @@
...
+                        throw new NoDocumentException("uuid does not exist")
+                    case _ =>
+                        logger.error(this, s"$viewName[$authkey] is not unique")
```
rabbah: remove key from log.

common/scala/src/main/scala/whisk/core/entity/WhiskAuthV2.scala
```diff
@@ -0,0 +1,66 @@
...
+            "name" -> w.name.toJson,
+            "uuid" -> w.authkey.uuid.toJson,
+            "key" -> w.authkey.key.toJson)
+
+        def read(value: JsValue) = Try {
+            value.asJsObject.getFields("name", "uuid", "key") match {
+                case Seq(JsString(n), JsString(u), JsString(k)) =>
+                    WhiskNamespace(EntityName(n), AuthKey(UUID(u), Secret(k)))
+            }
+        } getOrElse deserializationError("auth record malformed")
```
rabbah: namespace record malformed?

common/scala/src/main/scala/whisk/core/entity/WhiskAuthV2.scala
```diff
@@ -0,0 +1,66 @@
...
+    }
+}
+
+/**
+ * Represents the new version of entries in the subjects database. No
+ * top-level authkey is given but each subject rather has a set of
```
rabbah: strike `rather`.

common/scala/src/main/scala/whisk/core/entity/WhiskAuthV2.scala
```diff
@@ -0,0 +1,66 @@
...
+    subject: Subject,
+    namespaces: Set[WhiskNamespace])
+    extends WhiskDocument {
+
+    override def docid = DocId(subject())
+
+    def toJson = JsObject(
+        "subject" -> subject.toJson,
+        "namespaces" -> namespaces.toJson)
+}
```
rabbah: we could define a conversion from `WhiskAuthV2(subject, namespace)` to `WhiskAuth(subject, authkey)` for the case where `namespaces.size == 1` and `namespaces(0).name == subject` (and vice versa). This will allow us to work with either (and convert `WhiskAuth to WhiskAuthV2`).

markusthoemmes: I agree but I'd define those methods as soon as needed to avoid dead code. What do you think?

PR #576, opened 2016-06-06
WIP - Addition of install for push notifications package.

nwspeete-ibm:

Issue #550, opened 2016-06-02
Users can't tell if their action is terminated because of action time limit


csantanapr:
I have seeing two users report this, as they try to figure out why their actions are failing.

I'm not sure if we already print to stderr logs, so they can easily tell from the logs that action was terminated.
Also from activation id information might be another way to inform them that activation was terminated because it reach limit.

cc @rabbah @markusthoemmes

rabbah:
The activation start and end times in the activation record will allow a user to know: if `end - start >= action limit` then it was a time out vs an application error. We could adjust the error message accordingly.

ryanjbaxter:
I think having an error message/code that indicates a timeout error as opposed to the current error which just indicates that the action did not return valid json, would be a better way of handling the situation

csantanapr:
having user do calculation on end - start, it's too low level, I agree error message is more usable to users.

mbehrendt:
+1

on top of that, should there be a corresponding error code, eg a 504?

csantanapr:
I think having the 3 are very useful 
1. return code
2. return string
3. stderr message in user logs (user will see it in `wsk activation poll` or UI console)

rabbah:
I favor doing _only_ 2 from the list above. 

We can achieve `3` by projecting the error from the activation during an activation poll. 

We return `BadGateway` already for application/developer errors including actions that timeout and don't produce a response. Consider this nodejs action as an example `function main() { return whisk.async(); }`. This is a developer error and the action will timeout.

csantanapr:
I'm OK not writing to stderr 3 above, if we can achieve goal of letting user know what happens when his running `wsk activation poll`

rabbah:
The error message will now indicate the action timed out. `3` from above is not yet done.

markusthoemmes:
As @rabbah mentioned, warnings are in place. @csantanapr close?

rabbah:
The status code returned by an invoke is covered by #645.
To generate a message in the wsk activation poll, if the result of the activation is an error, it could be printed to the console. This would be the last remaining bit from the items listed in this issue.

PR #1487, opened 2016-11-09
User namespaces issue 898

xin-cai:
docker user namespaces issue 898

### DIFF
--- a/ansible/roles/invoker/tasks/deploy.yml
+++ b/ansible/roles/invoker/tasks/deploy.yml
@@ -14,30 +14,39 @@
     - '{{ docker_image_prefix }}/javaaction'
   when: docker_registry != ""
 
-- name: (re)start invoker
-  docker_container:
-    name: "invoker{{play_hosts.index(inventory_hostname)}}"
-    image: "{{docker_registry}}{{ docker_image_prefix }}/invoker:{{docker_image_tag}}"
-    state: started
-    recreate: true
-    restart_policy: "{{docker.restart.policy}}"
-    hostname: "invoker{{play_hosts.index(inventory_hostname)}}"
-    env:
-      "COMPONENT_NAME": "invoker{{play_hosts.index(inventory_hostname)}}"
-      "CONSULSERVER_HOST": "{{groups['consul_servers'] | first}}"
-      "CONSUL_HOST_PORT4": "{{consul.port.http}}"
-      "SELF_DOCKER_ENDPOINT": "{{inventory_hostname}}:{{docker.port}}"
-      "SERVICE_NAME": "invoker{{play_hosts.index(inventory_hostname)}}"
-      "SERVICE_CHECK_HTTP": "/ping"
-      "SERVICE_CHECK_TIMEOUT": "2s"
-      "SERVICE_CHECK_INTERVAL": "15s"
-      "PORT": 8080 
-    volumes:
-      - "{{whisk_logs_dir}}/invoker{{play_hosts.index(inventory_hostname)}}:/logs"
-      - "/var/lib/docker/containers/:/containers"
-    ports:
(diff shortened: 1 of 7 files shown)


sjfink:
why did we revert to aufs?

xin-cai:
probably @domdom82 can answer why revert to aufs ...

xin-cai:
Got green playground build #550

markusthoemmes:
@sjfink there is no revert to `aufs`, the commit message is misleading. It seems like the vagrant env was never elevated to `overlay`, probably due to kernel version in the 14.04 image.

tools/macos/tweak-dockermachine.sh
```diff
@@ -9,6 +9,7 @@ MACHINE_NAME=${1:-whisk}
 # Disable TLS.
 docker-machine ssh $MACHINE_NAME "echo DOCKER_TLS=no |sudo tee -a /var/lib/boot2docker/profile > /dev/null"
 docker-machine ssh $MACHINE_NAME "echo DOCKER_HOST=\'-H tcp://0.0.0.0:4243\' |sudo tee -a /var/lib/boot2docker/profile > /dev/null"
+docker-machine ssh $MACHINE_NAME "echo EXTRA_ARGS=\'--userns-remap=default\' |sudo tee -a /var/lib/boot2docker/profile > /dev/null"
```
domdom82: when docker-machine launches the daemon it reads the boot2docker/profile and appends EXTRA_ARGS to the dockerd parameters. there is no special env var for the userns flag. check `/usr/local/etc/init.d/docker` on a docker-machine VM.

domdom82:
@sjfink @markusthoemmes re aufs, that is correct. the commit message was "revert to auf" because I tried to move to overlay but then realized we can't use it with the 3.13 kernel in vagrant at it was only available from 3.18. I also tried to upgrade to ubuntu 16 but the vagrant image has lots of bugs so I stayed with 14.04 for now.
see https://docs.docker.com/engine/userguide/storagedriver/overlayfs-driver/ for details.

ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -14,30 +14,32 @@
...
-      "PORT": 8080 
-    volumes:
-      - "{{whisk_logs_dir}}/invoker{{play_hosts.index(inventory_hostname)}}:/logs"
-      - "/var/lib/docker/containers/:/containers"
-    ports:
-      - "{{invoker.port + play_hosts.index(inventory_hostname)}}:8080"
-    command: /bin/sh -c "/invoker/bin/invoker {{play_hosts.index(inventory_hostname)}} >> /logs/invoker{{play_hosts.index(inventory_hostname)}}_logs.log 2>&1"
+- name: determine docker root dir
+  shell: "docker info 2>/dev/null | grep 'Docker Root Dir' | awk -F ': ' '{print $2}'"
+  register: docker_root_dir
```
domdom82: good idea. we can omit the version tho to get the latest. we have already pinned the api version implicitly by pinning the docker version.

```
curl host:port/info
```
should yield the same results. (calls /1.24/info on docker 1.12)

xin-cai:
green build without user namespace enabled in Travis docker daemon

domdom82:
@perryibm pls review as this is going to affect the invoker. should not do any harm though.

domdom82:
@xin-cai squash and rebase after implementing feedback. thanks

domdom82:
the following steps are planned to get this in:
- push this to master with only travis having user namespaces enabled
- existing systems (ci, dev envs, prod) do not use this feature yet and will gracefully ignore it (docker ignores the --userns=host flag if no namespaces are configured)
- push blue parts to master (can be done independently)
- once the code is in place, update docker daemons (vagrant, docker-machine, ci, prod). this should automatically pick up the feature.

xin-cai:
Get green PG2 Build #576

PR #1493, opened 2016-11-10
Fix optional config override when not defined (with test).

rabbah:


jasonpet:
LGTM.  I tested this out and the fix worked.

jasonpet:
rebased and merged.

PR #1497, opened 2016-11-11
Add exec kind to annotation for an action and show it in action list.

### LINKED ISSUE #427: add ability to distinguish plain actions from sequences at the "list" level
From @starpit : 
right now, the `wsk action list` output doesn't not allow me to distinguish sequences from actions, at least for sequences created on the CLI

for sequences created in the UI, i have been adding an annotation `{ key: "isSequence", value: true }`

it'd be great if we could update the CLI to adopt this behavior. otherwise, i cannot know how to render an action, without doing a "get" on it.

rabbah:
Closes #427.

### DIFF
--- a/tools/cli/go-whisk-cli/commands/util.go
+++ b/tools/cli/go-whisk-cli/commands/util.go
@@ -263,7 +263,8 @@ func printActionList(actions []whisk.Action) {
         if action.Publish {
             publishState = wski18n.T("shared")
         }
-        fmt.Printf("%-70s %s\n", fmt.Sprintf("/%s/%s", action.Namespace, action.Name), publishState)
+        kind := getValueString(action.Annotations, "exec")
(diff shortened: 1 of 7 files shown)


tools/cli/go-whisk-cli/commands/util.go
```diff
@@ -263,7 +263,8 @@ func printActionList(actions []whisk.Action) {
         if action.Publish {
             publishState = wski18n.T("shared")
         }
-        fmt.Printf("%-70s %s\n", fmt.Sprintf("/%s/%s", action.Namespace, action.Name), publishState)
```
rabbah: @dubeejw if getValueString doesn't find the `exec` property, what happens?

dubeejw: Yes, everything should be fine if exec isn't present.

rabbah:
pg2/560 approved.

psuter:
Squashed and merged as df8d99a177e06070e74ed4e0524ee5083c1a80ae

PR #1501, opened 2016-11-11
Normalize property reading from file and consul.

rabbah:
Reading properties from a file did not discriminate on the required+optional properties listed and read all properties in the file. This was different from reading from the environment or consul. Normalized so that file based reading of values is the same. Added additional tests and fixed a bug that caused values to be dropped when in the optional set.


rabbah:
pg1/820 approved.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -143,8 +144,8 @@ object WhiskConfig extends Logging {
     def readPropertiesFromConsul(properties: scala.collection.mutable.Map[String, String])(implicit system: ActorSystem) = {
```
rabbah: it doesn't escape - updating in place is desirable?

rabbah:
pg2/594 approved.

sjfink:
rebased and merged 48a0362..3af6035

PR #1490, opened 2016-11-10
Referenced entities

rabbah:
Main features of this pr:
1. entitlement for sequences (will need more testing)
2. entitlement for rules with actions in packages (will need more tests, fixes #239)
3. fixes some gaps in the package entitlement (added several new tests)

The first 5 commits are rebased from other prs for completeness.

PR #1502, opened 2016-11-14
Compute referenced entities in common trait

rabbah:
Additional refactoring on top of #1490.
0. Replace Binding type with FullyQualifiedEntityName (Binding type was Binding(namespace: EntityPath, name: EntityName) which is the same as FullyQualifiedEntityName with no Version defined).
1. Introduce a common trait for computing referenced entities.
2. Use this trait for packages, rules and actions.
3. Add more tests related to entitlement.


rabbah:
I'm going to rebase and restage these commits into smaller prs.

Issue #1494, opened 2016-11-11
Add `request-promise` to nodejs runtime


csantanapr:
Someone said that `request-promise` would be a good npm package to have installed by default.
I did quick look and it looks like a very popular package, well maintained, compatible license
https://www.npmjs.com/package/request-promise

csantanapr:
merged 83bfc64e69bd4148eefeb5be273344c71afbd552

PR #1506, opened 2016-11-14
API GW database and actions deployment

mdeuser:
API GW database creation and wipe
API GW backend route management action and package creation
backend route management action negative parameter tests

### DIFF
--- a/ansible/environments/local/group_vars/all
+++ b/ansible/environments/local/group_vars/all
@@ -26,3 +26,9 @@ limits:
   triggers:
     fires:
       perMinute: 60
+
+# API GW connection configuration
+apigw_auth_user: ""
+apigw_auth_pwd: ""
+apigw_host: "http://{{ groups['apigateway']|first }}:{{apigateway.port.api}}/v1"
+
--- a/ansible/roles/routemgmt/files/gwapis_views.json
+++ b/ansible/roles/routemgmt/files/gwapis_views.json
@@ -1,0 +1,13 @@
+{
+  "_id": "_design/gwapis",
+  "views": {
+    "routes-by-api-name": {
+      "map": "function (doc) {\n  var PATHSEP = \"/\";\n\n  if (doc.namespace && doc.apidoc && doc.apidoc.info && doc.apidoc.info.title) {\n    if (! (doc._id.indexOf(\"STAGED_\") == 0) ) {\n      emit([doc.namespace, doc.apidoc.info.title], doc);\n    }\n  }\n}"
+    },
+    "routes-by-namespace": {
+      "map": "function (doc) {\n  var PATHSEP = \"/\";\n\n  if (doc.namespace && doc.apidoc && doc.apidoc.info && doc.apidoc.info.title) {\n    if (! (doc._id.indexOf(\"STAGED_\") == 0) ) {\n        emit(doc.namespace, doc);\n    }\n  }\n}"
+    }
+  },
+  "language": "javascript",
+  "indexes": {}
+}
--- a/ansible/roles/routemgmt/tasks/db.yml
+++ b/ansible/roles/routemgmt/tasks/db.yml
@@ -1,0 +1,36 @@
+---
+# This creates the immortal api gateway configuration (gwapis) database, only if it doesn't exist yet
+
+- name: check if API Gateway database already exists
+  uri:
+    url: "{{ db_protocol }}://{{ db_host }}:{{ db_port }}/{{ db_apigw }}"
+    method: GET
+    status_code: 200,404
+    user: "{{ db_username }}"
+    password: "{{ db_password }}"
+    force_basic_auth: yes
+  register: result
+
+- name: create immortal API Gateway database
+  uri:
+    url: "{{ db_protocol }}://{{ db_host }}:{{ db_port }}/{{ db_apigw }}"
+    method: PUT
+    status_code: 200,201
+    user: "{{ db_username }}"
+    password: "{{ db_password }}"
+    force_basic_auth: yes
+  when: result.status == 404
+
+- name: create the API Gateway query views for the API Gateway database
+  vars:
+    gwapis_views: "{{ lookup('file', 'gwapis_views.json') }}"
+  uri:
+    url: "{{ db_protocol }}://{{ db_host }}:{{ db_port }}/{{ db_apigw }}"
+    method: POST
+    status_code: 200,201
+    body_format: json
+    body: "{{ gwapis_views }}"
+    user: "{{ db_username }}"
+    password: "{{ db_password }}"
+    force_basic_auth: yes
+  when: result.status == 404
--- a/ansible/templates/whisk.properties.j2
+++ b/ansible/templates/whisk.properties.j2
@@ -86,3 +86,8 @@ db.password={{ db_password }}
 db.prefix={{ db_prefix }}
 db.whisk.actions={{ db_prefix }}whisks
 db.whisk.auths={{ db_auth }}
+db.whisk.apigw={{ db_apigw }}
+
+apigw.auth.user={{apigw_auth_user}}
+apigw.auth.pwd={{apigw_auth_pwd}}
+apigw.host={{apigw_host}}
--- a/core/routemgmt/activateApi.js
+++ b/core/routemgmt/activateApi.js
@@ -1,0 +1,387 @@
+/**
+ *
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Create an API Gateway to action mapping document in database:
+ * https://docs.cloudant.com/document.html#documentCreate
+ *
+ * Parameters (all as fields in the message JSON object)
+ *   host       Required. The database dns host name
+ *   port       Required. The database port number
+ *   protocol   Required. The database protocol (i.e. http, https)
+ *   dbname     Required. The name of the database
+ *   username   Required. The database user name used to access the database
+ *   password   Required. The database user password
+ *   gwUrl      Required. The API Gateway base path (i.e. http://gw.com)
+ *   gwUser     Optional. The API Gateway user ID
+ *   gwAuth     Optional. The API Gateway user ID's password
+ *   namespace  Required. The namespace of the API to be activated
+ *   basepath   Required. The basepath of the API to be activated
+ *   docid      Optional. If specified, use this docid to retrieve API doc from DB
+ *
+ * NOTE: The package containing this action will be bound to the following values:
+ *         host, port, protocol, dbname, username, password, gwUrl, gwUser, gwPwd
+ *       As such, the caller to this action should normally avoid explicitly setting
+ *       these values
+ **/
+ var request = require('request');
+ var url = require('url');
+
+function main(message) {
+  var badArgMsg = '';
+  if (badArgMsg = validateArgs(message)) {
+    return whisk.error(badArgMsg);
+  }
+  var dbname = message.dbname;
+
+  var gwInfo = {
+    gwUrl: message.gwUrl,
+  };
+  if (message.gwUser && message.gwPwd) {
+    gwInfo.gwAuth = Buffer.from(message.gwUser+':'+message.gwPwd,'ascii').toString('base64');
+  }
+
+  // Log parameter values
+  console.log('DB host    : '+message.host);
+  console.log('DB port    : '+message.port);
+  console.log('DB protocol: '+message.protocol);
+  console.log('DB username: '+confidentialPrint(message.username));
+  console.log('DB password: '+confidentialPrint(message.password));
+  console.log('DB database: '+message.dbname);
+  console.log('GW URL     : '+message.gwUrl);
+  console.log('GW User    : '+confidentialPrint(message.gwUser));
+  console.log('GW Pwd     : '+confidentialPrint(message.gwPwd));
+  console.log('namespace  : '+message.namespace);
+  console.log('basepath   : '+message.basepath);
+  console.log('docid      : '+message.docid);
+
+  var cloudantOrError = getCloudantAccount(message);
+  if (typeof cloudantOrError !== 'object') {
+    console.error('CloudantAccount returned an unexpected object type: '+(typeof cloudantOrError));
+    return whisk.error('Internal error.  An unexpected object type was obtained.');
+  }
+  var cloudant = cloudantOrError;
+  var cloudantDb = cloudant.use(dbname);
+
+  // Activating a specified API
+  // 1. Pull the API's configuration document from the DB
+  // 2. Create the request document to send to the API Gateway
+  // 3. Send the request to the API Gateway to configure and activate the API
+  // 4. If successful, update the API configuration in the DB to indicate the activation status
+  var tenantGuid;
+  var tenantAdded = false;
+  var gwApiActivated = false;
+  var dbUpdated = false;
+  var gwApiDoc;
+  var dbApiDoc;
+
+  return getApiDoc(message.namespace, message.basepath, message.docid)
+  .then(function(dbdoc) {
+    dbApiDoc = dbdoc;
+    gwApiDoc = makeGwApiDoc(dbdoc);
+    console.log("Gateway doc: "+JSON.stringify(gwApiDoc));
+    return addTenantToGateway(gwInfo, message.namespace);
+  })
+  .then(function(tenant){
+      if (tenant && tenant.id ) {
+        console.log('Tenant guid: '+tenant.id);
+        tenantGuid = tenant.id;
+        gwApiDoc.tenantId = tenant.id;
+      } else {
+        console.error('No tenant guid returned');
+        return Promise.reject('Internal error. API gateway did not return a tenant guid.')
+      }
+      tenantAdded = true;
+      return addRouteToGateway(gwInfo, gwApiDoc);
+  })
+  .then(function(gwApiResponse) {
+      if (gwApiResponse) {
+        console.log('API Gateway response: '+JSON.stringify(gwApiResponse));
+      } else {
+        console.error('No configured API route returned');
+        return Promise.reject('Internal error.  API gateway did not return a configured API route.')
+      }
+      gwApiActivated = true;
+      dbApiDoc.gwApiUrl = gwApiResponse.managedUrl;
+      var gwUrl = url.parse(gwApiResponse.managedUrl);
+      dbApiDoc.apidoc.host = gwUrl.host;
+      dbApiDoc.gwApiGuid = gwApiResponse.id;
+      dbApiDoc.tenantId = tenantGuid;
+      dbApiDoc.gwApiActivated = true;
+      console.log('Updating DB API doc: '+JSON.stringify(dbApiDoc));
+      return updateApiDocInDb(cloudantDb, dbApiDoc); })
+  .then(function(insertResp) {
+      dbUpdated = true;
+      return Promise.resolve(dbApiDoc);
+  })
+  .catch(function(reason) {
+      if (!gwApiActivated) {
+        console.error('API activation failure: '+reason);
+        return Promise.reject('API activation failure: '+reason);
+      }
+      console.error('API was activated; however, the database could not be udpated: '+reason);
+      return Promise.reject('API was activated; however, an internal error occurred after activation: '+reason);
+  });
+}
+
+function getCloudantAccount(message) {
+  // full cloudant URL - Cloudant NPM package has issues creating valid URLs
+  // when the username contains dashes (common in Bluemix scenarios)
+  var cloudantUrl;
+
+  if (message.url) {
+    // use bluemix binding
+    cloudantUrl = message.url;
+  } else {
+    if (!message.host) {
+      whisk.error('cloudant account host is required.');
+      return;
+    }
+    if (!message.username) {
+      whisk.error('cloudant account username is required.');
+      return;
+    }
+    if (!message.password) {
+      whisk.error('cloudant account password is required.');
+      return;
+    }
+    if (!message.port) {
+      whisk.error('cloudant account port is required.');
+      return;
+    }
+    if (!message.protocol) {
+      whisk.error('cloudant account protocol is required.');
+      return;
+    }
+
+    cloudantUrl = message.protocol + "://" + message.username + ":" + message.password + "@" + message.host + ":" + message.port;
+  }
+
+  return require('cloudant')({
+    url: cloudantUrl
+  });
+}
+
+// Retrieve the specific API document from the DB
+// This is a wrapper around an action invocation (getApi)
+function getApiDoc(namespace, basepath, docid) {
+  var actionName = '/whisk.system/routemgmt/getApi';
+  var params = {
+    'namespace': namespace,
+    'basepath': basepath
+  }
+  if (docid) params.docid = docid;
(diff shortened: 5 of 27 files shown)


rabbah:
a quick comment following a cursory scan - there is quite a bit of clone and own logic for accessing the db, did you consider pulling that out into a single file and requiring it in the resultant smaller "actions" and use zip files to package the dependencies?

ansible/roles/routemgmt/files/gwapis_views.json
```diff
@@ -0,0 +1,16 @@
+{
+  "_id": "_design/gwapis",
+  "views": {
+    "routes-by-path": {
+      "map": "function (doc) {\n  var PATHSEP = \"/\";\n\n  if (doc.namespace && doc.gatewayPath) {\n    emit([doc.namespace, doc.gatewayPath], {\"gatewayMethod\":doc.gatewayMethod, \"gatewayPath\":doc.gatewayPath, \"action\":doc.action, \"route-collection\":doc[\"route-collection\"], \"apidoc\":doc.apidoc});\n  }\n}"
```
markusthoemmes: Will there be any documents in the database that are not routes?

mdeuser: @markusthoemmes - at the moment, no.  But for performance, I can imagine views that return a subset of data from the docs

mdeuser:
@rabbah - Good point.  Yes, the packaging is not currently optimal.  With the node packaging support (.zip) I think this can be repackaged a bit more cleanly.  I'll open an issue for that

mdeuser:
PG1 578 (API GW tests all passed)

mdeuser:
@markusthoemmes - Do you have time to complete the review this PR?  I'll squash/rebase when all looks good.

One question I have is how to resolve the following actor failures that occur during test (and not just in this PR's new tests).  Any ideas?
```
[2016-11-15T21:00:49.244Z] [ERROR] Outgoing request stream error [akka.actor.ActorSystemImpl:71] 
    akka.stream.AbruptTerminationException: Processor actor [Actor[akka://default/user/StreamSupervisor-521/flow-5-0-unknown-operation#-363075754]] terminated abruptly
```

tools/db/wipeTransientDBs.sh
```diff
@@ -55,6 +55,20 @@ do
...
+# drop the database
+CMD="$CURL_ADMIN -X DELETE $URL_BASE/$DB_APIGW"
+RES=$($CMD)
+if [[ "$RES" == '{"ok":true}' ]]; then
+    echo DELETED
+else
+    # table may not exist
+    echo WARNING: $RES
+fi
+
```
markusthoemmes: I feel like we should not conflate the standard whisks db with the API gateway db. Any reason we need to delete the DB in this script?

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
```diff
@@ -0,0 +1,203 @@
...
+              Seq("-p", "apidoc", "{\"namespace\":\"_\",\"gatewayBasePath\":\"/ApiGwRoutemgmtActionTests_bp\",\"gatewayPath\":\"ApiGwRoutemgmtActionTests_rp\",\"gatewayMethod\":\"get\",\"action\":{\"backendMethod\":\"post\",\"backendUrl\":\"URL\",\"namespace\":\"_\"}}")),
+            ("/whisk.system/routemgmt/createRoute", ANY_ERROR_EXIT, "action is missing the authkey field",
+              Seq("-p", "apidoc", "{\"namespace\":\"_\",\"gatewayBasePath\":\"/ApiGwRoutemgmtActionTests_bp\",\"gatewayPath\":\"ApiGwRoutemgmtActionTests_rp\",\"gatewayMethod\":\"get\",\"action\":{\"backendMethod\":\"post\",\"backendUrl\":\"URL\",\"namespace\":\"_\",\"name\":\"N\"}}")),
+            ("/whisk.system/routemgmt/createRoute", ANY_ERROR_EXIT, "swagger and gatewayBasePath are mutually exclusive and cannot be specified together",
+              Seq("-p", "apidoc", "{\"namespace\":\"_\",\"gatewayBasePath\":\"/ApiGwRoutemgmtActionTests_bp\",\"gatewayPath\":\"ApiGwRoutemgmtActionTests_rp\",\"gatewayMethod\":\"get\",\"action\":{\"backendMethod\":\"post\",\"backendUrl\":\"URL\",\"namespace\":\"_\",\"name\":\"N\",\"authkey\":\"XXXX\"},\"swagger\":{}}")),
```
markusthoemmes: Use triple-quote notation to surpass the need to escape the JSON strings.

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
```diff
@@ -0,0 +1,203 @@
...
+
+        invalidArgs foreach {
+            case (action:String, exitcode:Int, errmsg:String, params:Seq[String]) =>
+                var cmd: Seq[String] = Seq("action",
+                    "invoke",
+                    action,
+                    "-i", "-b", "-r",
+                    "--apihost", wskprops.apihost,
+                    "--auth", whisksystemAuthId+":"+whisksystemAuthPwd
+                ) ++ params
```
markusthoemmes: Why write your own version of invoke? Is the usual `wsk.action.invoke` not enough?

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
```diff
@@ -0,0 +1,203 @@
...
+        authstore.shutdown()
+        super.afterAll()
+    }
+
+    behavior of "API Gateway routemgmt action parameter validation"
+
+    it should "reject routemgmt actions that are invoked with not enough parameters" in {
+        val invalidArgs = Seq(
+            //getApi
+            ("/whisk.system/routemgmt/getApi", ANY_ERROR_EXIT, "namespace is required", Seq()),
```
markusthoemmes: All of them use `ANY_ERROR_EXIT`, is that even needed then?

ansible/roles/routemgmt/tasks/db.yml
```diff
@@ -0,0 +1,52 @@
...
+    method: POST
+    status_code: 200,201
+    body_format: json
+    body: "{{ gwapis_views }}"
+    user: "{{ db_username }}"
+    password: "{{ db_password }}"
+    force_basic_auth: yes
+  when: result.status == 404
+
+- name: create the API Gateway index (Cloudant only)
```
markusthoemmes: Why is this needed?

core/routemgmt/activateApi.js
```diff
@@ -0,0 +1,383 @@
...
+  // 2. Create the request document to send to the API Gateway
+  // 3. Send the request to the API Gateway to configure and activate the API
+  // 4. If successful, update the API configuration in the DB to indicate the activation status
+  var retApi;
+  var tenantGuid;
+  var tenantAdded = false;
+  var gwApiActivated = false;
+  var dbUpdated = false;
+  var gwApiDoc;
+  var dbApiDoc;
```
markusthoemmes: I understand why these definitions are here, but this feels like quite brittle to me as you mutate all those variables in Promises down the road. Is there a way to overcome this `undefined` hell?

core/routemgmt/activateApi.js
```diff
@@ -0,0 +1,383 @@
...
+  console.log('GW URL     : '+message.gwUrl);
+  console.log('GW User    : '+message.gwUser);
+  console.log('GW Pwd     : '+message.gwPwd);
+  console.log('GW Auth    : '+gwInfo.gwAuth);
+  console.log('namespace  : '+message.namespace);
+  console.log('basepath   : '+message.basepath);
+  console.log('docid      : '+message.docid);
+
+  var cloudantOrError = getCloudantAccount(message);
+  if (typeof cloudantOrError !== 'object') {
```
markusthoemmes: Why not return `false` if something went wrong in `getCloudantAccount` instead? Or `null`.

core/routemgmt/activateApi.js
```diff
@@ -0,0 +1,383 @@
...
+  console.log('getApiDoc() for: ', params);
+    return whisk.invoke({
+      name: actionName,
+      blocking: true,
+      parameters: params
+    })
+    .then(function (activation) {
+      console.log('whisk.invoke('+actionName+', '+params.namespace+', '+params.basepath+') ok');
+      console.log('Results: '+JSON.stringify(activation));
+    if (activation && activation.result && activation.result.apis &&
```
markusthoemmes: check the identation

core/routemgmt/activateApi.js
```diff
@@ -0,0 +1,383 @@
...
+    .catch(function (error) {
+      console.error('whisk.invoke('+actionName+', '+params.namespace+', '+params.basepath+') error: '+JSON.stringify(error));
+      return Promise.reject(error);
+    });
+}
+
+function addTenantToGateway(gwInfo, namespace) {
+  var options = {
+    url: gwInfo.gwUrl+'/tenants',
+    agentOptions: {rejectUnauthorized: false},
```
markusthoemmes: This effectively deactivates SSL? Is that what's wanted? Sounds like it should not be in production?

core/routemgmt/activateApi.js
```diff
@@ -0,0 +1,383 @@
...
+    headers: {
+      'Content-Type': 'application/json',
+      'Accept': 'application/json'
```
markusthoemmes: Is this necessary when you already set the `json` attribute?

core/routemgmt/activateApi.js
```diff
@@ -0,0 +1,383 @@
...
+        if (body && body.id) {  // body has format like:  { id: GUID, namespace: NAMESPACE, instance: 'openwhisk' }
+          console.log('addTenantToGateway: got a single tenant resposne');
+          resolve(body);
+        } else {
+          console.error('addTenantToGateway: failure: No tenant guid provided')
+          reject('Unable to configure the API Gateway: Invalid response from API Gateway');
+        }
+      }
+    });
+  });
```
markusthoemmes: Should we rely on #1494 for this? Small change, big impact on readability.

core/routemgmt/deactivateApi.js
```diff
@@ -0,0 +1,265 @@
...
+  console.log('DB port    : '+message.port);
+  console.log('DB protocol: '+message.protocol);
+  console.log('DB username: '+message.username);
+  console.log('DB database: '+message.dbname);
+  console.log('GW URL     : '+message.gwUrl);
+  console.log('GW Auth API: '+message.gwAuth);
+  console.log('namespace  : '+message.namespace);
+  console.log('basepath   : '+message.basepath);
+
+  var cloudantOrError = getCloudantAccount(message);
```
markusthoemmes: Is this really just possible on Cloudant? Not CouchDB?

core/routemgmt/deleteApi.js
```diff
@@ -0,0 +1,448 @@
...
+    name: actionName,
+    blocking: true,
+    parameters: params
+  })
+  .then(function (activation) {
+    console.log('whisk.invoke('+actionName+', '+params.namespace+', '+params.basepath+') ok');
+    console.log('Results: '+JSON.stringify(activation));
+    if (activation && activation.result && activation.result.apis &&
+        activation.result.apis.length == 1 && activation.result.apis[0].value &&
+        activation.result.apis[0].value._rev) {
```
markusthoemmes: I find that defensive coding quite annoying to read, while i realize its necessary.

Will open an issue to add lodash to our runtimes. You could use https://lodash.com/docs/4.16.4#has then, which would make that way more readable.

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
mdeuser: Nice tip.  Wish I'd known that before typing all of those escapes :-)

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
mdeuser: I included the exit code just in case we need more granular exit code validation.

core/routemgmt/deactivateApi.js
mdeuser: In theory, it should work with CouchDB as well; although, I've not tried that out yet.  Issue #1515

core/routemgmt/activateApi.js
mdeuser: I agree.  Use of the request-promise package would add a bit more readability.  Alas, that issue is not currently resolved; but I'm thinking that it should not block the initial API GW code delivery.

ansible/roles/routemgmt/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+apigw_auth_user: ""
+apigw_auth_pwd: ""
+apigw_host: "http://{{ groups['edge']|first }}:3100/v1"
```
mdeuser: This defaults file can now be deleted.  Based on http://docs.ansible.com/ansible/playbooks_variables.html#id31 it seemed to provide default values that could be overridden by other playbook variable settings.  But as builds progressed, these default values were not as pervasive as I had assumed.

core/routemgmt/activateApi.js
mdeuser: The function is actually reused from the Cloudant package.  I left the interface as-is thinking this function might be leveraged directly from the package when the entire set of API GW actions were re-factored to use single-source common functions..

core/routemgmt/activateApi.js
```diff
@@ -0,0 +1,383 @@
...
+  // Log parameter values
+  console.log('DB host    : '+message.host);
+  console.log('DB port    : '+message.port);
+  console.log('DB protocol: '+message.protocol);
+  console.log('DB username: '+message.username);
+  console.log('DB database: '+message.dbname);
+  console.log('GW URL     : '+message.gwUrl);
+  console.log('GW User    : '+message.gwUser);
+  console.log('GW Pwd     : '+message.gwPwd);
+  console.log('GW Auth    : '+gwInfo.gwAuth);
```
markusthoemmes: Logging credentials seems like a bad idea in a production environment?

core/routemgmt/activateApi.js
mdeuser: I struggled in several places with the need to use promises, but also with the need to maintain some state across promise processing functions.  At the time I had implemented something even uglier and @bjustin-ibm suggested the using variables that were scoped outside of the promise chain coupled with a single catch.

The retApi variable can be removed; but that does not address your overall point.

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
```diff
@@ -0,0 +1,203 @@
...
+                    "invoke",
+                    action,
+                    "-i", "-b", "-r",
+                    "--apihost", wskprops.apihost,
+                    "--auth", whisksystemAuthId+":"+whisksystemAuthPwd
+                ) ++ params
+                val rr = wsk.cli(cmd, expectedExitCode = exitcode)
+                rr.stderr should include regex (errmsg)
+        }
+    }
```
markusthoemmes: Any tests to verify that the actions actually do the right job?

core/routemgmt/activateApi.js
bjustin-ibm: A couple thoughts here... 

Some of these variables are not actually read, and can be removed.

It may be possible to handle failures more granularly by adding error callbacks as the second argument to `then()`. This will let you handle specific errors as they happen, rather than trying to handle all errors in a single `catch()` at the end.

If you are incrementally building up a result, perhaps it is reasonable to define a single `result = {};` object which is modified as asynchronous steps complete. @markusthoemmes would this address your concern about `undefined` variables?

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
mdeuser: Yes, those will be forthcoming when the API GW is delivered.

core/routemgmt/activateApi.js
mdeuser: Good observation.  I just checked with the API GW folks and they now use a CA signed cert, so this limitation can be removed.

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
mdeuser: I believe the `wsk.action.invoke` hardwired the use of the whisk.properties auth key (auth.guest).  When invoking these actions, I needed to use the /whisk.system's auth key.

mdeuser:
PG2 599 - passed

mdeuser:
@markusthoemmes - now that a local API GW has been merged, I've added some positive test cases to add coverage for actual interaction with the API GW.

mdeuser:
The recent travis build runs the API GW tests (positive and negative) against a local API GW container running on travis.

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
```diff
@@ -0,0 +1,470 @@
...
+               swagger: String = "",
+               expectedExitCode: Int = SUCCESS_EXIT): RunResult ={
+        val parms = Map[String,JsValue]() ++
+            Map("namespace" -> "_".toJson) ++
+            { if (basepath.length() > 0)  Map("basepath" -> basepath.toJson) else Map[String,JsValue]() } ++
+            { if (relpath.length() > 0)   Map("relpath" -> relpath.toJson) else Map[String,JsValue]() } ++
+            { if (operation.length() > 0) Map("operation" -> operation.toJson) else Map[String,JsValue]() } ++
+            { if (apiname.length() > 0) Map("apiname" -> apiname.toJson) else Map[String,JsValue]() } ++
+            { if (action != null) Map("action" -> action.toJson) else Map[String,JsValue]() } ++
+            { if (swagger.length() > 0) Map("swagger" -> swagger.toJson) else Map[String,JsValue]() }
```
markusthoemmes: Seems like you want all of those to be `Option[String]` instead of checking the length of the String?

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
```diff
@@ -0,0 +1,470 @@
...
+            expectedExitCode = SUCCESS_EXIT)(wskprops)
+        var apiJsArray: JsArray =
+            try {
+                var apisobj = rr.stdout.parseJson.asJsObject.fields("apis")
+                apisobj.convertTo[JsArray]
+            } catch {
+                case e:Exception =>
+                    JsArray.empty
+            }
+        return apiJsArray
```
markusthoemmes: One does not simply use `return` in Scala :).

Above code is equivalent to the more idiomatic: `Try(rr.stdout.parseJson.asJsObject.fields("apis").convertTo[JsArray]).getOrElse(JsArray.empty)`

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
markusthoemmes: Using the standard  `wsk.action.invoke` you can also greatly simplify the parameter declaration in your testdata.

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
mdeuser: I have played a bit with Scala's implicit `return` feature, but as a personal preference I tend to lean toward an explicit coding style, so I will usually retain the `return` for clarity - and to keep me from accidentally changing return behavior :-)

Re: use of `JsArray`.  Yes, this method could have returned `Vector[JsValue]`, which is one step closer to the desired results.

markusthoemmes: Does the `JsValue` follow a specific format?

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
mdeuser: After reading up on the Scala `Option` type, yes, using `Option[String]` would be helpful in handling the `None` case.  I found some `getOrElse` usages in other test code to mirror.

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
mdeuser: From what I understand, it's generic class that can be used for dynamic JSON navigation (i.e. the value might be a JSON string, array, number, object, etc)

markusthoemmes: Right, but in this specific case, can we narrow it down to a more typed approach and thus easier to understand and read

tests/src/whisk/core/apigw/actions/test/ApiGwRoutemgmtActionTests.scala
mdeuser: Good tips.  wsk.action.invoke is used in all of the the new positive test cases (with a properties file using the whisk.system credentials).

ansible/templates/whisk.properties.j2
```diff
@@ -86,3 +86,8 @@ db.password={{ db_password }}
 db.prefix={{ db_prefix }}
 db.whisk.actions={{ db_prefix }}whisks
 db.whisk.auths={{ db_auth }}
+db.whisk.apigw={{ db_apigw }}
+
+apigw.auth.user={{apigw_auth_user}}
+apigw.auth.pwd={{apigw_auth_pwd}}
+apigw.host={{apigw_host}}
```
domdom82: why not change this to `http://{{ groups['apigateway']|first }}:{{apigateway.port.api}}/v1`
and remove from environment group vars? much cleaner this way and less overhead.

ansible/environments/local/group_vars/all
```diff
@@ -26,3 +26,9 @@ limits:
   triggers:
     fires:
       perMinute: 60
+
+# API GW connection configuration
+apigw_auth_user: ""
+apigw_auth_pwd: ""
+apigw_host: "http://{{ groups['apigateway']|first }}:{{apigateway.port.api}}/v1"
```
mdeuser: @domdom82 - Yes, this will vary by environment.  In the OW environments it will always refer to a local API GW, in BW other fixed API GWs will be available.

mdeuser:
@domdom82 - Recent commit has the ansible changes we talked about.  Thanks for the input!

rabbah:
Keep in mind that `whisk.invoke` is now deprecated as of https://github.com/openwhisk/openwhisk/commit/85070a20f560e77ea064ebfc57f2c1eb4608f420.

core/routemgmt/activateApi.js
```diff
@@ -0,0 +1,387 @@
...
+  return '';
+}
+
+function confidentialPrint(str) {
+    var printStr;
+    if (str) {
+        printStr = 'XXXXXXXXXX';
+    }
+    return printStr;
+}
```
markusthoemmes: What is this for? Why not simply omit the print of credentials at all?

rabbah: Let's please get all the credential logging out.

mdeuser: if the credential is not supplied (indicates a corrupt package), the log will show an empty string for the credential.  non-empty credentials are not logged.

mdeuser:
PG3 27 passed

mdeuser:
@markusthoemmes - any other comments?  If you agree, any outstanding nice-to-have comments will have a separate issue opened up for them.

mdeuser:
Travis CI build success.  PG1 972 and PG2 692 are both in progress

csantanapr:
merged 06e8a075ca216fe37e21b3b33c50f7271e550455

Issue #898, opened 2016-07-15
User containers should run as non-root user


domdom82:
### This is a design-proposal for security hardening of the invoker.

I have been testing with ways to achieve better security against forkbomb attacks using AppArmor. One of the main problem with applying the `nproc` limit is that it does not apply to root, no matter whether you apply the limit via ulimit or AppArmor (which uses ulimit under the hood anyway).

The second problem is that the `nproc` limit applies only per user, not per container. It is, after all, a kernel feature and way older than lxc, docker or any other container tech. So to make this work for docker < 1.10 (i.e. without user namespaces) you have to use a new user id for every container you run. This makes sure that each process gets a fresh `nproc` limit.

If you use the same user id for all containers, you will run into the problems described [here](https://docs.docker.com/engine/reference/commandline/run/#/for-nproc-usage) as they will all share the same limit.

**My proposal is this:**

When running a user container, the invoker runs the container using a high UID and passes it using the `docker run -u <uid>` option. The user does not have to exist on the host for this to work. Inside the container, the user will still be member of the root group, however no longer be root itself, thus all ulimits like **nproc** apply. This will effectively protect against forkbomb attacks.

This is how it would work:
1. when `makeContainer` gets called, the invoker first calls `getFreeUID(containerName)`
2. getFreeUID returns an unused user id and blocks this user id for other containers.
3. `makeContainer` then starts the action container like so
   
   ```
   docker run -d -u <freeUID> --security-stuff whisk/nodeJsAction
   ```
4. `rmContainer` finally unblocks the user id to make it available again.

There are two drawbacks I see for this approach:
1. We have to keep track of the user ids. So the invoker needs to make sure every new container gets a unique uid to prevent nproc limit collision. This is basically doing user namespaces manually.
2. The author of a blackbox container can no longer make the "I am root" assumption. Their stuff needs to be able to run as non-root users.

domdom82:
With --pids-limit coming in Docker 1.11 and Kernel 4.3 we no longer need nproc ulimit. This makes this problem much easier to handle as we don't need different user ids for every container. Instead I suggest we use an unprivileged user id like "nobody" to run the containers and see where we land.

domdom82:
Team meeting decision from Oct 4:

- we will go forward with user namespaces
- we will not mount containers read-only
- we will limit storage using overlay2 + xfs using docker 1.13 (modulo stemcell being able to mount root fs as xfs)
[...]

PR #1509, opened 2016-11-15
replace nproc limit with pids-limit to limit number of forks

domdom82:
nproc ulimit does not work with docker containers. with docker 1.12 we can leverage the pid limit on cgroups using the --pids-limit flag.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
@@ -70,7 +70,7 @@ trait ContainerUtils extends Logging {
         val capabilityArg = Array("--cap-drop", "NET_RAW", "--cap-drop", "NET_ADMIN")
         val consulServiceIgnore = Array("-e", "SERVICE_IGNORE=true")
         val fileHandleLimit = Array("--ulimit", "nofile=64:64")
-        val processLimit = Array("--ulimit", "nproc=512:512")
+        val processLimit = Array("--pids-limit", "64")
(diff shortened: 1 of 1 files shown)


domdom82:
related to #898

perryibm:
Originally the value 512 was set high because the limits were global across containers.  If this new `--pids-limit` option is specific to the container, we should choose a much lower value.

domdom82:
@perryibm yes it is per container. what would be a reasonable value? assuming we have about 16 actions per invoker on average would be 512 / 16 = 32. we could also calculate dynamically based on coreshare * numcore.

perryibm:
LGTM - run test?

domdom82:
@xin-cai please merge this with your new fb test

xin-cai:
We are currently encountering the travis build problem, the couchdb image from travis doesn't exist any more:

```
TASK [couchdb : (re)start CouchDB] *********************************************

Wednesday 30 November 2016  13:44:18 +0000 (0:00:00.077)       0:00:01.288 **** 

fatal: [172.17.0.1]: FAILED! => {"changed": false, "failed": true, "msg": "Error pulling couchdb - code: None message: Get http://: http: no Host in request URL"}
```

xin-cai:
blue PG2 644

domdom82:
couch startup fixed by https://github.com/openwhisk/openwhisk/pull/1571

domdom82:
@perryibm we have to move the test to blue because the kernel in travis is too old to support the --pids-limit flag. need at least 4.3 for the flag to work which is currently unsupported in travis.

domdom82:
PG3 21

xin-cai:
PG1 961 approved

PR #1512, opened 2016-11-15
update docs for alarm example on using steps

### LINKED ISSUE #1169: Add example to alarm for doing a step like every 2 hours
Based on some user feedback on [#openwhisk](https://dwopen.slack.com/messages/openwhisk) Slack channel, it will be good to have an example on specifying steps in the docs using cron syntax.

Based on [crontab.org](http://crontab.org/)

```
Step  values can be used in conjunction with ranges.  Fol-
             lowing a range with ``/<number>'' specifies skips  of  the
             number's value through the range.  For example, ``0-23/2''
             can be used in the hours field to specify  command  execu-
             tion  every other hour (the alternative in the V7 standard
             is ``0,2,4,6,8,10,12,14,16,18,20,22'').   Steps  are  also
             permitted after an asterisk, so if you want to say ``every
             two hours'', just use ``*/2''.
```

Example to add
- `* * */2 * * *`: every 2 hours.

csantanapr:
Fixes #1169 

Fixed the top of the hour example, added example for step `*/2`

### DIFF
--- a/docs/catalog.md
+++ b/docs/catalog.md
@@ -225,7 +225,8 @@ The `/whisk.system/alarms/alarm` feed configures the Alarm service to fire a tri
 
   - `* * * * * *`: every second.
   - `0 * * * * *`: top of every minute.
-  - `* 0 * * * *`: top of every hour.
+  - `0 0 * * * *`: top of every hour.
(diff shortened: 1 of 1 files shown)


csantanapr:
replace by #1551

PR #1518, opened 2016-11-16
Allow entitlement check to operate on a set of resources in bulk.

rabbah:
pg1/846.


sjfink:
rebased merged and pushed 
0c112a4..7662140

PR #1526, opened 2016-11-17
Entitlement check on sequence components.

rabbah:
pg2/601.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Actions.scala
@@ -34,6 +34,7 @@ import spray.http.HttpMethods.PUT
 import spray.http.StatusCodes.BadGateway
 import spray.http.StatusCodes.BadRequest
 import spray.http.StatusCodes.InternalServerError
+import spray.http.StatusCodes.NotFound
 import spray.http.StatusCodes.OK
 import spray.http.StatusCodes.Accepted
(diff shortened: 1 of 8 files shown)


core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -218,7 +219,24 @@ trait WhiskActionsApi extends WhiskCollectionAPI {
...
                 val docid = DocId(WhiskEntity.qualifiedName(namespace, name))
-                putEntity(WhiskAction, entityStore, docid, overwrite, update(user, content)_, () => { make(user, namespace, content, name) })
+                def doput = {
+                    putEntity(WhiskAction, entityStore, docid, overwrite,
+                        update(user, content)_, () => { make(user, namespace, content, name) })
+                }
+
+                content.exec match {
+                    case Some(seq @ SequenceExec(_, components)) =>
+                        info(this, "checking if sequence components are accessible")
```
psuter: This may succeed at creation time and the access rights could still be different at runtime, right? So it's a "don't shoot yourself in the foot" kind of check?

rabbah: right, since we're not tracking referenced entities.

psuter:
Merged as 884c913c1bddfd4d2cc5b1409cb2b8e3c46d8977

PR #1530, opened 2016-11-18
Drop support for Swift 2

paulcastro:
This PR fixes Issue #1459. Submitted for review to see if solution is acceptable.  

Changes:
- Add deprecation check to controller for create/update/invoke actions on runtimes
- add error message if deprecated runtime is required.  
- Update Swift 3 tests, update run script test that can  timeout
- remove tests for Swift 2.

Swift 2 container is still in the code base (for now).

Blue PG 867, 871


rabbah:
Is the goal to prevent any new action from being created with swift 2 as the exec kind *and* stop any from being executed?

paulcastro:
Yes. The requirement is to prevent any new actions from being created (or updated), and prevent any execution of already existing Swift 2 actions.

csantanapr:
@paulcastro if you stop folks from creating new actions, or run existing actions, this should be called "deprecated" it should be call "drop support for swift2". "deprecated means no new enhancements, but everything continues to execute"

@rabbah should we do the same for NodeJS 0.12? Drop support

common/scala/src/main/scala/whisk/core/entity/DeprecatedRuntimeVersions.scala
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.entity
+
+
+trait DeprecatedRuntimeVersions {
+    def isDeprecated(exec: AnyRef): Boolean = exec match {
```
markusthoemmes: Can `exec` be of type `Exec`?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -233,6 +233,14 @@ trait WhiskActionsApi extends WhiskCollectionAPI with SequenceActions with Refer
                                         case SequenceExec(_, components) =>
                                             val futureSeqTuple = invokeSequence(user, action, payload, blocking, topmost = true, components, cause = None, 0)
                                             futureSeqTuple map { case (activationId, wskActivation, _) => (activationId, wskActivation) }
+                                        case exec: Exec =>
```
rabbah: I will try to do this.

rabbah: won't work - deserializer doesn't know the context.

rabbah:
@paulcastro I've rebased this PR and made some changes + added tests:

https://github.com/rabbah/openwhisk/commit/2f6ae26cf6377f9a3b98c3bb5dbb67320f731c34

rabbah:
Closing in favor of new pr linked above.

PR #1531, opened 2016-11-18
Allow rules to reference actions in packages.

### LINKED ISSUE #239: confusing error message on trying to create rule with shared action

rabbah:
Fixes #239.
This required some refactoring of the entitlement path which allowed for consolidation of entitlement checks on referenced entities.

PG2/613 and PG1/882.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Rules.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Rules.scala
@@ -45,7 +45,7 @@ object WhiskRulesApi {
 }
 
 /** A trait implementing the rules API */
-trait WhiskRulesApi extends WhiskCollectionAPI {
+trait WhiskRulesApi extends WhiskCollectionAPI with ReferencedEntities {
     services: WhiskServices =>
 
     protected override val collection = Collection(Collection.RULES)
@@ -98,16 +98,6 @@ trait WhiskRulesApi extends WhiskCollectionAPI {
                 }
             }
         }
-    }
-
-    private def referencedEntities(content: WhiskRulePut) = {
-        val triggerResource = content.trigger.map {
-            t => Resource(t.path, Collection(Collection.TRIGGERS), Some(t.name()))
-        }
-        val actionResource = content.action map {
-            a => Resource(a.path, Collection(Collection.ACTIONS), Some(a.name()))
-        }
-        Set(triggerResource, actionResource).flatten
     }
 
     /**
--- a/core/controller/src/main/scala/whisk/core/controller/Rules.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Rules.scala
@@ -21,37 +21,17 @@ import scala.util.Failure
 import scala.util.Success
 
 import akka.actor.ActorSystem
-import spray.http.StatusCodes.BadRequest
-import spray.http.StatusCodes.Conflict
-import spray.http.StatusCodes.InternalServerError
-import spray.http.StatusCodes.OK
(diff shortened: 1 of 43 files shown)


tests/src/whisk/core/controller/test/RulesApiTests.scala
```diff
@@ -247,74 +236,66 @@ class RulesApiTests extends ControllerTestCommon with WhiskRulesApi {
...
             status should be(OK)
-            t.rules.get(ruleNameQualified) shouldBe ReducedRule(actionNameQualified, Status.ACTIVE)
             val response = responseAs[WhiskRuleResponse]
             response should be(rule.withStatus(Status.ACTIVE))
+            t.rules.get(rule.fullyQualifiedName(false)) shouldBe ReducedRule(action.fullyQualifiedName(false), Status.ACTIVE)
         }
     }
 
-    ignore should "create rule with an action in a package" in {
+    it should "create rule with an action in a package" in {
```
psuter: :+1: :+1: :+1:

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -178,7 +177,7 @@ protected[core] abstract class EntitlementService(config: WhiskConfig, loadBalan
         }
 
         entitlementCheck andThen {
-            case Success(r) =>
+            case Success(r) if resources.nonEmpty =>
```
rabbah: it's `andThen` simply for logging; it skips the success case where resource is empty.

core/controller/src/main/scala/whisk/core/controller/Rules.scala
```diff
@@ -396,13 +370,26 @@ trait WhiskRulesApi extends WhiskCollectionAPI {
...
+                    new NoDocumentException(s"trigger ${trigger.qualifiedNameWithLeadingSlash} does not exist")
+                }
+                case _: DeserializationException => Future.failed {
+                    new DeserializationException(s"trigger ${trigger.qualifiedNameWithLeadingSlash} is corrupted")
+                }
             }
-            actionExists <- WhiskAction.get(entityStore, action) recoverWith {
-                case _: NoDocumentException => Future.failed(new NoDocumentException(s"action $action does not exist"))
+
+            actionExists <- WhiskAction.get(entityStore, action.toDocId) recoverWith {
```
rabbah: since it's a CRUD operation, I concluded it's OK and avoid fetching a doc if the first doesn't exist.

rabbah: short circuits the second fetch if it's not necessary - it's a CRUD, and on the success path, cache is prewarmed with these entities.

psuter:
how squashed do you want this?

psuter:
Merged as 12ae1dbaf9bb4626086dfd4b2476de01b70fdf47...2c7174eb063a436d7f8447911cf3b547c2819f29

PR #1533, opened 2016-11-18
Internalize sequence combinator to controller.

### LINKED ISSUE #992: Provide sequence ID in addition to activation IDs
To track the behavior of a sequence (e.g. when debugging going through logs) comprised of multiple activations the same way one would use a transaction or session ID on other products customers suggested the introduction of a sequence ID being associated with each action invoked as part of a particular sequence.

Thoughts?

### LINKED ISSUE #995: Discussion: Sequence abort convention
@perryibm @rabbah and I discussed sequences. I'm summarizing the conclusions of the discussion.

User-visible aspects:
- A sequence will have an activation id
- The logs of a sequence are constructed by the OW runtime as follows: each line contains an activation id and the result (success, failure) for the activation id
- Entitlement check is performed at the moment of invocation for all actions (i.e., to invoke a sequence, the user has to have the rights to invoke each action in the sequence). There won't be any partial execution of sequences (e.g., first 2 actions in the sequence are executed as the user has rights, while the remaining ones are not if the user doesn't have the right to invoke the third action in the sequence).

Implementation internals:
- The current implementation with `pipe.js` in the invoker will stay in place but guarded by a flag for experimental/comparison reasons
- The logic of executing a sequence are moved to the controller. The controller starts an actor each time it needs to execute a sequence. The actor executes in blocking fashion each action by posting directly to Kafka.
- The logs for the sequence are constructed as explained above and stored in the activation record for the sequence.
- The annotation for the activation will contain a field kind that indicates this activation was a sequence

More comments from @rabbah 
- entitlement check for the sequence component should be done both at create/update and invoke time.
- note: the actor doesn't perform an entitlement check before invoking - it's the route handler that does; the actor simply executes the invokes, waits for responses, and then writes the activation record.
- the result of the sequence should be the result of the last action in the sequence.

### LINKED ISSUE #1253: Write method for package resolution and intermediate entities in package bindings

rabbah:
Squashed all of #1473 and fixing some defects.
Fixes #995.
Fixes #992.
Fixes #1253.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/SequenceActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/SequenceActions.scala
@@ -1,0 +1,400 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.controller
+
+import java.time.Clock
+import java.time.Instant
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.Promise
+import scala.concurrent.duration._
+import scala.util.Failure
+import scala.util.Success
+import scala.util.Try
+
+import WhiskActionsApi._
+import akka.actor.ActorSystem
+import spray.json._
+import whisk.common.Logging
+import whisk.common.TransactionId
+import whisk.core.entity._
+import whisk.core.entity.size.SizeInt
+import whisk.core.entity.types._
+import whisk.http.Messages._
+import whisk.utils.ExecutionContextFactory.FutureExtensions
+
+trait SequenceActions extends Logging {
+    services: WhiskServices =>
+
+    protected implicit val executionContext: ExecutionContext
+
+    /** An actor system for timed based futures. */
+    protected implicit val actorSystem: ActorSystem
+
+    /** Database service to CRUD actions. */
+    protected val entityStore: EntityStore
+
+    /** Database service to get activations. */
+    protected val activationStore: ActivationStore
+
+    protected def invokeSingleAction(
+        user: Identity,
+        action: WhiskAction,
+        env: Option[Parameters],
+        payload: Option[JsObject],
+        timeout: FiniteDuration,
+        blocking: Boolean,
+        cause: Option[ActivationId] = None)(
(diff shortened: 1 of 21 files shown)


core/controller/src/main/scala/whisk/core/controller/SequenceActions.scala
```diff
@@ -0,0 +1,399 @@
...
+        action: WhiskAction,
+        activationId: ActivationId,
+        wskActivationEithers: Vector[Either[ActivationResponse, WhiskActivation]],
+        topmost: Boolean,
+        cause: Option[ActivationId],
+        start: Instant,
+        end: Instant): WhiskActivation = {
+
+        // extract all successful activations from the vector of activation eithers
+        val (right, left) = wskActivationEithers.span(_.isRight)
```
markusthoemmes: Are you sure `span` is the right choice vs. `partition`?

From the scala doc:

> Splits this list into a prefix/suffix pair according to a predicate.

As I read that, `right` in your case is not guaranteed to only contain rights, which will result in an exception in the next line.

core/controller/src/main/scala/whisk/core/controller/SequenceActions.scala
```diff
@@ -0,0 +1,400 @@
...
+                    // if the future fails with a timeout, the failure is dealt with at the caller level
+                    futureSeqResultTimeout map { wskActivationTuples =>
+                        val wskActivationEithers = wskActivationTuples.map(_._1)
+                        // the execution of the sequence was successful, return the result
+                        val end = Instant.now(Clock.systemUTC())
+                        val seqActivation = Some(makeSequenceActivation(user, action, seqActivationId, wskActivationEithers, topmost, cause, start, end))
+                        val atomicActionCnt = wskActivationTuples.last._2
+                        (seqActivationId, seqActivation, atomicActionCnt)
+                    } andThen {
+                        case Success((_, id, _)) => seqActivationPromise.success(id)
```
ioana-blue: `id` is really an activation, so `activation` would be less confusing.

core/controller/src/main/scala/whisk/core/controller/SequenceActions.scala
```diff
@@ -0,0 +1,400 @@
...
+                    val end = Instant.now(Clock.systemUTC())
+                    // the response of the sequence is the response of the very last activation
+                    makeSequenceActivation(user, action, seqActivationId, wskActivationEithers, topmost, cause, start, end)
+                }
+                storeSequenceActivation(activation)
+            case Failure(t: Throwable) =>
+                // consider this whisk error
+                // TODO shall we attempt storing the activation if it exists or even inspect the futures?
+                // this should be a pretty serious whisk errror if it gets here
+                error(this, s"Sequence activation 'seqActivationId' failed: ${t.getMessage}")
```
ioana-blue: missing '$'

rabbah:
pg2/610.

psuter:
Merged as ca15c68d348a2a02cf9da54475e96b43d48a3dac

rabbah:
@ioana-blue up to you to close #992, #995 and #1253.

PR #1436, opened 2016-10-25
.spawn instead of .exec to avoid buffering output

psuter:
In the NodeJS container, buffering the full output causes issues with archives that contain many files.

PR #1540, opened 2016-11-21
Keep `unzip` quiet to avoid buffering issues.

psuter:
Alternative fix to the same issue #1436 was addressing.


psuter:
Merged as b36e785e408cfcbddfa6b0e670d8b04a49be9c65

Issue #47, opened 2016-03-06
discussion: relax convention on main() entrypoint in javascript


sjfink:
One user has suggested that forcing the action method to be main() is too constraining and we should allow the user to specify the main method.

Originally we did this via module.exports.    As we move to support more languages, it may be better to define a language-neutral attribute on actions.   

Alternatively, we could decide not to support this.

markusthoemmes:
Any details on what is constraining about it?

mbehrendt:
imho, it feels unnatural to prescribe a 'main' function to exist in js code

markusthoemmes:
@sjfink Can you elaborate what you meant with "it may be better to define a language-neutral attribute on actions".

The way I understand it is: We would have a CLI parameter which determines the entry point for the script like we do in Java but in a standardized way for all languages?

Additionally, several languages could make use of their own ecosystem specificities (like `module.exports` in Node) to lift hard-coded `main` requirement.

csantanapr:
I think we should think of the UX as a developer focusing in a specific language, and then later see what's common across.
For example me a NodeJS fanboy :-) , I have the same complaint that forcing me to create a function and name it `main` is not natural, also not natural not allowing me to export my action thru as the module using `module.exports`

We should start with good and simple way, to allow to export the module as the action function like:
```
//es6
module.exports = (params)=> params;
//es5
module.exports = function (params) {return params};
```
I like that zip nodejs actions today use `module.exports` but I don't like this `module.exports.main`
Why the extra `main` ? Not need it I think. I prefer the anonymous (just a preference :-)) 
I know that lambda and other have this extra level like `modules.exports.handler`, I think having the extra handler is OK but not necessary as a base level to get started with simple actions.

For nodejs we can have customization with nice fall back defaults allowing for backwards compatibility.

Let's say we added a property to the action metadata call `Handler: string`
If Handler is defined (let's say `Handler: 'handler'`) then we assume user did this:
```
module.exports.handler = (params)=> params;
``` 
Then if Handler is not defined we can check for `main` to see if user did this (for backwards compatibility, or deprecated)
```
module.exports.main = (params)=> params;
```
Then if Handler not defined, and no function in `module.exports.main`, then we can check if the user exported the action as the module like
```
module.exports = (params)=> params;
```

[...]

PR #1544, opened 2016-11-21
Introduce --main flag

### LINKED ISSUE #1018: CLI spends a long time scanning JAR file to find main for Java actions
If one creates a Java action from a JAR containing a large number of classes, the time it takes the CLI to "create" the action starts to be come unacceptably long. The CLI runs `javap` for each class to find `main` and each of these can take a second or so. So it adds up fast. The CLI could give up once it find 1 match (rather than look for all), or allow/require the user to specify their `package.class.method` for `"main"`.

Related to #47.
FYI @psuter @mdeuser @dubeejw.

### LINKED ISSUE #1026: Unable to use the same jar in multiple Java actions
I would like to use the same jar for two Java actions, each with a slightly different 'main' class/method (but sharing the rest of the jar).
Today (with wsk CLI) this does not seem possible, because I can pass only one JAR (which must include all the dependencies etc), and the CLI 'guesses' the main class (by analyzing the jar and picking the first class with proper 'main' method signature).

Related to #1018

psuter:
Introduces a `--main` flag in the CLI for action creation and update.

For Java, the flag is required; the CLI no longer attempts to determine the action entry point by analyzing the JAR file, which was potentially slow and error prone.

Users can now also create multiple Java actions using the same JAR file but a different entry point.

Fixes #1018.
Fixes #1026.

For NodeJS, Swift, Python, the flag lets the user optionally chose an alternative entry point than `main`.

Fixes (most of) #47.


rabbah:
:+1: 

We should ensure that the invoker will always include the "main" metadata on `init` and shift the "main" defined check there (although defensively coding against it in the action runner/proxy is ok).

psuter:
I've reached the set of features I wanted. Opening up for (rest of) review.

psuter:
Rebased. FYI `d96fe1d` has become 116684478e8f03a552de762b3519b15b6fc7b5a4.

psuter:
PG3 37 is blue.

tests/src/system/basic/CLIJavaTests.scala
```diff
@@ -46,11 +47,13 @@ class CLIJavaTests
     /**
      * Test the Java "hello world" demo sequence
      */
-    it should "invoke a java action" in withAssetCleaner(wskprops) {
+    it should "Invoke a java action" in withAssetCleaner(wskprops) {
```
rabbah: ah man - we'll have to rename all our tests now :)

psuter: I like to focus on the essential.

psuter:
Rebased again, and triggered builds for the "critical" commits (the ones with rebase conflicts). When all is green, I suggest we squash the first two commits into one, and the last 3 into one. This will nicely separate the work by language and components (containers first, CLI+entities then).

rabbah:
Merged as https://github.com/openwhisk/openwhisk/compare/da7ac1328e92...d98188d7eb74.

PR #1546, opened 2016-11-22
Allow PUT request for package to have optional binding specified as empty JSON object

### LINKED ISSUE #1368: [REST API] Empty "binding" property will cause malformed error when creating/updating package
## Environment details:
- REST API
1.   Send a PUT request to create or update a package
2.   In the JSON object, specify "binding" with "{}".

```
Package is successfully created or updated.
```

```
Error message is returned:
The request content was malformed: Object is missing required member 'namespace'
```
- It's quite common that JSON library will create empty object when the property is null. At least we should accept an empty object.
- This error seems related to creation of a bind. A bind action re-uses the same REST API.

rabbah:
Fixes  #1368.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskPackage.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskPackage.scala
@@ -180,22 +180,14 @@ object WhiskPackage
     }
 
     override implicit val serdes = {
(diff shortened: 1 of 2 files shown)


rabbah:
pg1/935 approved.

psuter:
Merged as c906bd9c8f83055ce9a5b1f6eee6f411ed161460

PR #1547, opened 2016-11-22
System tests for seq with package visibility changes.

rabbah:
System tests where entitlements change on referenced actions after creation.

### DIFF
--- a/tests/src/whisk/core/cli/test/WskEntitlementTests.scala
+++ b/tests/src/whisk/core/cli/test/WskEntitlementTests.scala
@@ -86,7 +86,13 @@ class WskEntitlementTests
                 (wp, assetHelper) =>
                     assetHelper.withCleaner(wsk.action, fullyQualifiedActionName, confirmDelete = false) {
                         (action, name) =>
-                            val rr = action.create(name, None, shared = Some(true), update = true, expectedExitCode = FORBIDDEN)(wp)
+                            val rr = action.create(name, None, update = true, expectedExitCode = FORBIDDEN)(wp)
+                            rr.stderr should include("not authorized")
+                            rr
+                    }
+                    assetHelper.withCleaner(wsk.action, "unauthorized sequence", confirmDelete = false) {
+                        (action, name) =>
+                            val rr = action.create(name, Some(fullyQualifiedActionName), kind = Some("sequence"), update = true, expectedExitCode = FORBIDDEN)(wp)
                             rr.stderr should include("not authorized")
                             rr
                     }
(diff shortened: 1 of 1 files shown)


tests/src/whisk/core/cli/test/WskEntitlementTests.scala
```diff
@@ -86,7 +86,13 @@ class WskEntitlementTests
                 (wp, assetHelper) =>
                     assetHelper.withCleaner(wsk.action, fullyQualifiedActionName, confirmDelete = false) {
                         (action, name) =>
-                            val rr = action.create(name, None, shared = Some(true), update = true, expectedExitCode = FORBIDDEN)(wp)
+                            val rr = action.create(name, None, update = true, expectedExitCode = FORBIDDEN)(wp)
```
ioana-blue: could you please use var names that are more intuitive/less ambiguous? I'm guessing rr means "run result" and not "rodric rabbah". how about using "run" or "result"?

tests/src/whisk/core/cli/test/WskEntitlementTests.scala
```diff
@@ -259,6 +265,63 @@ class WskEntitlementTests
             })(defaultWskProps)
     }
 
+    it should "invoke an action sequence from package" in withAssetCleaner(guestWskProps) {
```
ioana-blue: Not immediately obvious whether we need this test. We have this one: https://github.com/openwhisk/openwhisk/blob/master/tests/src/system/basic/WskSequenceTests.scala#L228

which tests a sequence that points to an action in a package (like the test under review); in addition, the existing test performs checks around parameter passing. Do we need this test you're adding?

tests/src/whisk/core/cli/test/WskEntitlementTests.scala
```diff
@@ -259,6 +265,63 @@ class WskEntitlementTests
...
+
+                    val run = wsk.action.invoke("sequence")(defaultWskProps)
+                    withActivation(wsk.activation, run)({
+                        _.response.success shouldBe true
+                    })(defaultWskProps)
+            }
+    }
+
+    it should "not allow invoke an action sequence with more than one component from package after entitlement change" in withAssetCleaner(guestWskProps) {
+        (gwp, assetHelper) =>
```
ioana-blue: gwp = global wealth predictor? :) see comment above about rr.

tests/src/whisk/core/cli/test/WskEntitlementTests.scala
```diff
@@ -259,6 +265,63 @@ class WskEntitlementTests
...
+            assetHelper.withCleaner(wsk.pkg, samplePackage) {
+                (pkg, _) =>
+                    pkg.create(samplePackage, parameters = Map("a" -> "A".toJson), shared = Some(true))(gwp)
+                    pkg.create(samplePackage + "prv", parameters = Map("a" -> "A".toJson), shared = Some(true))(gwp)
```
ioana-blue: factor out in a val `samplePackage + "prv"` ?

tests/src/whisk/core/cli/test/WskEntitlementTests.scala
rabbah: we use `rr` in a few places in the test, it means run result, yes.

rabbah:
pg1/936 approved.

ioana-blue:
Rebased and squashed. Committed as 7ad4590

PR #1558, opened 2016-11-24
Warn user when consul configuration fails.

pauldub:
It may happen that we cannot read configuration from consul, in such
cases (server is down / unreachable / etc.) log a warning to inform the
user of the issue.

This happened to me on my first deployment, and I was startled at why
the controller was unable to pull its configuration from consul.

I have sent a signed CLA as is asked int CONTRIBUTING.md


rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/8efa1f96876af05f25f7e7ea2a5dace5bcac3054. 

Thank you for the fix and contribution.

PR #1560, opened 2016-11-24
give zookeeper a second to respond before closing connection during ruok query

domdom82:
the -q option does only exist in bsd versions of netcat. -i exists in both gnu and bsd versions. so using it in all cases for simplicity.


domdom82:
PG1 910

markusthoemmes:
rebased and merged in cb584ee3316d0321a7b7a80522e439becc9022b7

PR #1566, opened 2016-11-29
Handle Exceeded Blocking Invocation Time Limit

dubeejw:
- Display the activation record and an appropriate message if the maximum time limit for an blocking action is exceeded

### DIFF
--- a/tools/cli/go-whisk-cli/wski18n/resources/en_US.all.json
+++ b/tools/cli/go-whisk-cli/wski18n/resources/en_US.all.json
@@ -1150,5 +1150,13 @@
   {
     "id": "{{.ok}} got trigger {{.name}}, displaying field {{.field}}\n",
     "translation": "{{.ok}} got trigger {{.name}}, displaying field {{.field}}\n"
+  },
+  {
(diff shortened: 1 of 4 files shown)


tools/cli/go-whisk-cli/wski18n/resources/en_US.all.json
```diff
@@ -1150,5 +1150,13 @@
   {
     "id": "{{.ok}} got trigger {{.name}}, displaying field {{.field}}\n",
     "translation": "{{.ok}} got trigger {{.name}}, displaying field {{.field}}\n"
+  },
+  {
+    "id": "{{.ok}} invoked /{{.namespace}}/{{.name}} with id {{.id}}, blocking action exceeded maximum time limit.\n",
```
rabbah: "maximum time limit" might be confusing - since one can interpret this to mean the action's time limit.
perhaps this should state instead that the action exceeded the maximum allowed time for a blocking action to complete and produce a response; the action may still be running, use the activation id to check for a result later.

too long, but a start.

rabbah:
@dubeejw is this an active branch?

PR #1567, opened 2016-11-29
bump up CouchDB queue size to 16K

sjfink:


common/scala/src/main/scala/whisk/core/database/CouchDbRestClient.scala
```diff
@@ -68,7 +68,8 @@ class CouchDbRestClient(protocol: String, host: String, port: Int, username: Str
     // Additional queue in case all connections are busy. Should hardly ever be
     // filled in practice but can be useful, e.g., in tests starting many
     // asynchronous requests in a very short period of time.
-    private val requestQueue = Source.queue(128, OverflowStrategy.dropNew)
+    private val QUEUE_SIZE = 16* 1024;
```
rabbah: should we consider making this an ansible knob?

sjfink: seems like overkill ... really going to customize this value per deployment?

rabbah: I suppose there's a few other similar values that if we were to pull out we should do consistently. So considered and rejected at this point. I can't think of a good reason one would change this alone at deployment.

psuter:
Merged as a48202f07b5940387ab64abee4f442679f79f862.

PR #1573, opened 2016-12-01
Fix SchedulerTests: Wait for the actor to be terminated

markusthoemmes:
This should fix race conditions with the Actor receiving messages before the PoisonPill.


psuter:
Merged as aaa20ea674193c74b69a18a799a2927b440170d8

Issue #1577, opened 2016-12-01
API now requiring namespace on rule creation (breaking change)


krook:
## Environment details:
Running the Mac CLI against the Bluemix API:

```
whisk CLI version	2016-10-11T11:00:00+00:00
whisk API build		2016-12-01T16:15:42Z
whisk API build number	whisk-build-2174-accelerate1
```

## Steps to reproduce the issue:
Yesterday I was able to create a new rule linking an action to a trigger by running the following commands. This no longer works.

1.  `wsk action create handler handler.js`
2.  `wsk trigger create every-20-seconds --feed /whisk.system/alarms/alarm --param cron "*/20 * * * * *"`
3.  `wsk rule create invoke-periodically every-20-seconds handler`

## Provide the expected results and outputs:

```
ok: created rule invoke-periodically
```

## Provide the actual results and outputs:

```
error: Unable to create rule 'invoke-periodically': The request content was malformed:
requirement failed: The fully qualified name of the entity must contain at least the namespace and the name of the entity. (code 11978)
```

## Additional information you deem important:
* This worked with the whisk API build before 12/1

krook:
This was solved by downloading the latest CLI. 

Perhaps a better error message to prompt the user to upgrade?

rabbah:
Hmm. This should not be a breaking change.

rabbah:
This is an inadvertent breaking change in that my intention was to preserve the previous behavior but I did not have the right test in place. I've addressed the issue with this proposed change (and the accompanying test): https://github.com/openwhisk/openwhisk/commit/cffa53bdaf679cd7fe825a06a5d587d680bfee88#diff-c7a6ca91e16420b15c39ab67b4584401R289).

However, after further consideration, allowing this for rules only is asymmetric - in that for other API operations that permit references to entities, we do require a fully qualified name. This is true for a sequence components and for a package binding. Hence for consistency, I am inclined to close as won't fix in the API and instead resolve by documenting that a CLI upgrade is required instead.

@jthomas does npm openwhisk provide fully qualified names on rule creation? If not, the library is also affected by this change.

jthomas:
@rabbah The module relies on the user to provide the correct name format for triggers and actions.
https://github.com/openwhisk/openwhisk-client-js/blob/master/lib/rules.js#L35

I think the API change above is more consistent.

I would change the library to still allow for unqualified names and default to providing a fully-qualified name in the request using the namespace parameter. 

[...]

PR #1579, opened 2016-12-02
Issue 1577.2

rabbah:
Adds rules API handler test for use of `_` in action and trigger name. Reject missing namespace from either.

The second commit is a proposed fix for #1577 that permits unqualified action/trigger names in rule API operations which I think we should reject. The last commit adds additional tests to assert that a namespace must be specified (`_` is permitted as with the other referenced entities in sequences and bindings).

### DIFF
--- a/tests/src/whisk/core/entity/test/SchemaTests.scala
+++ b/tests/src/whisk/core/entity/test/SchemaTests.scala
@@ -100,6 +100,9 @@ class SchemaTests extends FlatSpec with BeforeAndAfter with Matchers {
         val spaces = paths.zip(expected).foreach {
             p => EntityPath(p._1).namespace shouldBe p._2
         }
+
+        EntityPath.DEFAULT.addpath(EntityPath("a")).toString shouldBe "_/a"
+        EntityPath.DEFAULT.addpath(EntityPath("a/b")).toString shouldBe "_/a/b"
     }
 
     it should "reject malformed paths" in {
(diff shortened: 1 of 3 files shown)


rabbah:
pg2/664 approved.

rabbah:
Dropped second commit: API requires fully qualified names (note `_` is allowed for namespace in the path).

tests/src/whisk/core/entity/test/SchemaTests.scala
```diff
@@ -145,15 +133,24 @@ class SchemaTests extends FlatSpec with BeforeAndAfter with Matchers {
             JsObject("path" -> "a".toJson, "name" -> "b".toJson),
             JsObject("path" -> "a".toJson, "name" -> "b".toJson, "version" -> "0.0.1".toJson),
             JsString("a/b"),
-            JsObject("namespace" -> "a".toJson, "name" -> "b".toJson))
+            JsString("n/a/b"),
+            JsString("/a/b"),
+            JsString("/n/a/b"),
+            JsString("b")) //JsObject("namespace" -> "a".toJson, "name" -> "b".toJson))
```
cbickel: Is this entry used anywhere?

rabbah:
Not now it isn't.

markusthoemmes:
rebased, squashed, rephrased, merged, pushed and smiled in c95d7541f60d88716ef9dd2d5d975ac6df279110

PR #1581, opened 2016-12-02
Sequence test with package bindings and parameters

ioana-blue:


tests/src/system/basic/WskSequenceTests.scala
```diff
@@ -250,22 +252,53 @@ class WskSequenceTests
             }
             val run = wsk.action.invoke(sName)
             // action params trump package params
-            checkLogsAtomicAction(0, run, actionStr)
+            checkLogsAtomicAction(0, run, new Regex(actionStr))
```
psuter: Scala has `.r` to turn a `String` into a `Regex`, FWIW

psuter:
PG?

ioana-blue:
PG3 35

Issue #1584, opened 2016-12-02
Possible bug with picking up namespace from .wskprops for component in sequence


ioana-blue:
While working on test in #1581 I noticed some behavior that looked buggy. Initially, I didn't specify any namespace for any of the entities created. `guest` was used by default. However, for the action component, the value in `.wskprops` from my home dir seemed to have been used. I need to investigate if there is a bug and where the bug is. 

## Environment details:

* local deployment: Mac OS
* unit testing

## Steps to reproduce the issue:

Related to test in #1581

ioana-blue:
I think it's a cli bug. See output below (edited out auth and path):

```
bluewhisk/open/ansible/../bin/wsk --verbose -i --apihost 192.168.99.100 --apiversion v1 action create --auth *****  /_/sequenceWithBindingParams package2/print --sequence
exit code = 147
stdout: [/go-whisk/whisk.NewClient]:070:[Inf] Disabling certificate checking.
[mmands.parseQualifiedName]:112:[Inf] Qualified entityName: sequenceWithBindingParams
[mmands.parseQualifiedName]:113:[Inf] Qaulified namespace: _
[-cli/commands.parseAction]:477:[Inf] Parsing parameters: []string(nil)
[.getKeyValueFormattedJSON]:197:[Inf] Created key/value format '[]' from 'map[]'
[-cli/commands.parseAction]:489:[Inf] Parsing annotations: []string(nil)
[.getKeyValueFormattedJSON]:197:[Inf] Created key/value format '[]' from 'map[]'
[-cli/commands.parseAction]:624:[Inf] Parsed action struct: &whisk.Action{Namespace:"_", Name:"sequenceWithBindingParams", Version:"", Publish:false, Exec:(*whisk.Exec)(0x4208aa300), Annotations:whisk.KeyValueArr(nil), Parameters:whisk.KeyValueArr(nil), Limits:(*whisk.Limits)(nil)}
[k.(*ActionService).Insert]:153:[Inf] Action insert route: actions/sequenceWithBindingParams?overwrite=false
REQUEST:
[PUT]	https://192.168.99.100/api/v1/namespaces/_/actions/sequenceWithBindingParams?overwrite=false
Req Headers
{
  "Authorization": [
    "Basic ***********"
  ],
  "Content-Type": [
    "application/json"
  ]
}Req Body
{{"exec":{"kind":"sequence","code":"","components":["/ioana@us.ibm.com_dev/package2/print"]}}
}
[...]

PR #1585, opened 2016-12-03
fix missing fqns and overwrite wsk config file

ioana-blue:
Issue #1584

### DIFF
--- a/tests/src/common/Wsk.scala
+++ b/tests/src/common/Wsk.scala
@@ -803,7 +803,7 @@ sealed trait RunWskCmd extends Matchers {
     def cli(params: Seq[String],
             expectedExitCode: Int = SUCCESS_EXIT,
             verbose: Boolean = false,
-            env: Map[String, String] = Map[String, String](),
+            env: Map[String, String] = Map("WSK_CONFIG_FILE" -> ""),
             workingDir: File = new File("."),
             showCmd: Boolean = false): RunResult = {
         val args = baseCommand
--- a/tests/src/system/basic/WskSequenceTests.scala
+++ b/tests/src/system/basic/WskSequenceTests.scala
@@ -270,8 +270,8 @@ class WskSequenceTests
 
     it should "run a sequence with an action in a package binding with parameters" in withAssetCleaner(wskprops) {
         (wp, assetHelper) =>
-            val packageName = s"/$defaultNamespace/package1"
-            val bindName = s"/$defaultNamespace/package2"
+            val packageName = "package1"
+            val bindName = "package2"
             val actionName = "print"
             val packageActionName = packageName + "/" + actionName
             val bindActionName = bindName + "/" + actionName


ioana-blue:
PG2 676

tests/src/common/Wsk.scala
```diff
@@ -276,7 +276,13 @@ class WskAction()
...
             implicit wp: WskProps): RunResult = {
         val params = Seq(noun, if (!update) "create" else "update", "--auth", wp.authKey, fqn(name)) ++
-            { artifact map { Seq(_) } getOrElse Seq() } ++
+            // split artifact, apply name resolution and reconstruct back the string
+            {
+                kind map {
+                    case "sequence" => artifact map { art => Seq(art.split(",").map(fqn(_)).mkString(",")) } getOrElse Seq()
+                    case _ => artifact map { Seq(_) } getOrElse Seq()
+                } getOrElse { artifact map { Seq(_) } getOrElse Seq() }
+            } ++
```
markusthoemmes: You got quite a few repetitions of `getOrElse Seq()` and so on in there. You can abuse `Option`s `Seq`-like behaviour (either one or none element) to to write the whole thing a bit more concise and understandable.

**Note:** The old code is equivalent to just `{ artifact }`. There is no need to convert the `Option` to a `Seq` (for which a `toSeq` would've been enough as well). You can always write `Seq() ++ someOption` because `Option` is "iterable" itself. `Option` can be read as a List of either 1 or 0 entries. Thus: `Seq() ++ someOption` will produce an empty list, if the `Option` is `None`, or a list of exactly 1 entry... you get the idea.

That being said, the code above can be refactored to:

```
kind match {
    case Some("sequence") => artifact.map(_.split(",").map(fqn(_)).mkString(","))
    case _ => artifact
}

rabbah:
I know I suggested the `fqn` on rules and sequences but note that this means we're losing test coverage (the ability to write a simple name instead of a fully qualified name); the need for the latter is replaced by the `env` override. We will need to test the CLI with the simple names as well.

ioana-blue:
@rabbah Now that I understand a bit more, I think the `fqn` on rules and sequences are not really necessary. In fact, I think some of the sequence tests use both fqn's and relative names, which test both usage patterns. As you pointed out, if we do add the `fqn`s everywhere, we're missing on some type of tests. I think it's better to leave them out and just submit the "real" fix (the env override).

ioana-blue:
Overwrite wsk config file in cli and took out the default namespace explicitly used in names in one sequence test (that's how I discovered this bug). Tested locally for a couple of tests and they work as expected. I'll start a PG shortly.

ioana-blue:
PG3 43

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/48191242ba28b770acc8d7f9d6ca7cdbbb08e4df.

PR #1587, opened 2016-12-03
start consul as server to fix restartability

jeremiaswerner:
ready for review

### DIFF
--- a/ansible/roles/consul/tasks/deploy.yml
+++ b/ansible/roles/consul/tasks/deploy.yml
@@ -34,6 +34,7 @@
     state: started
     recreate: true
     restart_policy: "{{ docker.restart.policy }}"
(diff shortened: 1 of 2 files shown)


domdom82:
good finding. with the change from our own consul image to public docker hub image we did not put a custom CMD anymore, thus resulting in default CMD which is "dev" mode. see https://github.com/hashicorp/docker-consul/blob/2c2873f9d619220d1eef0bc46ec78443f55a10b5/0.X/Dockerfile

Using a custom CMD should do the trick. We wanted to use `--net=host` anyway for stability which is a long-lasting open issue.

jeremiaswerner:
PG1 #968 greeny beeny

domdom82:
LGTM, we will handle net=host in another PR (has implications with user ns)

PR #1591, opened 2016-12-05
issue #1590 disable swapping for user containers.

### LINKED ISSUE #1590: user actions are using swap. no good.
## Environment details:

* Mac OS   
* Docker 1.12

1.   Start nodejs action with 256M limit
2.   `var b = Buffer.alloc(256*1024*1024, 'a')`

```
Action is killed
```

```
Action is not killed. 
```

* Action gets actually killed if you allocate at least twice the RAM (e.g. 600M)

This is the root cause:

https://docs.docker.com/engine/reference/run/#user-memory-constraints

We are currently omitting the `--memory-swap` flag which implies a swap of 2x memory.

Quoting from Docker Docs:

```
$ docker run -it -m 300M ubuntu:14.04 /bin/bash
We set memory limit only, this means the processes in the container can use 300M memory and 300M swap memory, by default, the total virtual memory size (–memory-swap) will be set as double of memory, in this case, memory + swap would be 2*300M, so processes can use 300M swap memory as well.
```

I have verified that the OOM killer actually works if we set `--memory-swap` to the same amount. The doc is a bit misleading as `--memory-swap` actually means "total memory, ram + swap" not just "swap".

I will open a PR to fix this.

domdom82:
fixes #1590


domdom82:
PG3 32

jeremiaswerner:
rebased and merged in 74deffd4fc372d6f43a3652747b2885a225e2eac

PR #1592, opened 2016-12-05
Add the method to get the credentials as a json object

houshengbo:
Since we are going to add the credentials for message hub in the
vcap file, we expect the property "kafka_brokers_sasl" as a list.
That is the reason why we need to get the credentials as a json
object instead of a hashmap.

### DIFF
--- a/tests/src/common/TestUtils.java
+++ b/tests/src/common/TestUtils.java
@@ -122,9 +122,7 @@ public class TestUtils {
      */
     public static Map<String, String> getVCAPcredentials(String vcapService) {
         try {
-            JsonArray vcapArray = getVCAPServices().get(vcapService).getAsJsonArray();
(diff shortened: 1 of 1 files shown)


tests/src/common/TestUtils.java
```diff
@@ -138,6 +136,18 @@ public static JsonObject getVCAPServices() {
     }
 
     /**
+     * Gets a VCAP_SERVICES credentials as the json objects.
+     *
+     * @return VCAP credentials as a json object.
+     */
+    public static JsonObject getCredentials(String vcapService) {
+    	JsonArray vcapArray = getVCAPServices().get(vcapService).getAsJsonArray();
```
markusthoemmes: please indent correctly.

Are we really willing to keep GSON in our repository? Why not spray.json?

bjustin-ibm:
PG1 982

Issue #1601, opened 2016-12-07
router.host always equal to cli.api.host?


rabbah:
I don't remember why we define both a `route.host` and a `cli.api.host` in our ansible hosts and whisk.properties files. @sjfink posits that `router.host == cli.api.host` now and I concluded the same... and we can't recall if that was always the case or why there needs to be a distinction. I think we should have just the `api.host` and the `edge.host`.

@jeremiaswerner @markusthoemmes input please.

sjfink:
I vaguely recall now.   It originated in blue-green testing of the CLI.

When swapping from blue to green, we would first run regression tests on the green cluster, and then swap if tests all passed.   For the initial set of tests in green, the "cli api host" would be the green edge.    Of course the CLI used by customers would always point to router host.

I think at some point we changed the tests to use a different flag with cli set property api.host, but the ansible variable is probably legacy from the days when the CLI_API_HOST was a build-time property of the CLI.    This is old stuff -- python CLI and pre-ansible.

domdom82:
The cli.api.host is the dns name per environment (e.g. "openwhisk.ng.bluemix.net")
The router.host is the actual host, e.g. "123.123.123.123"

Which makes sense in scenarios where they are not the same, e.g. loadbalancers, proxies etc.

Also `cli_api_host` is a property that is independent from the inventory, while `router.host` is always generated from the inventory of hosts for that particular environment.

rabbah:
@domdom82 thanks for the clarification. In open, the router is not used/deployed and there is only the edge. The cli.api.host is an ip address, not a name. Should we make it clean in open as well then and define api.host to a name or leave it be? The complication with a name is for mac e.g., where the localhost isn't good enough and one needs a `dockerhost` aliased.

rabbah:
Proposed renaming cli_api_host to whisk_api_host for clarity in #1602.

PR #1602, opened 2016-12-07
Actions should receive API_HOST not EDGE_HOST.

rabbah:
Also remove legacy router.host and rename cli.api.host to api.host.

Related to #1601. Pending comments and clarification on noted issue.

### DIFF
--- a/core/nodejsActionBase/src/whisk.js
+++ b/core/nodejsActionBase/src/whisk.js
@@ -28,14 +28,6 @@ var request = require('request');
  */
 function Whisk(apihost, logger) {
 
-    // export __OW_API_HOST to environment
-    {
-        apihost = apihost || '';
-        var edgeHostParts = apihost.split(':');
-        var protocol = (edgeHostParts.length >= 2  &&  edgeHostParts[1] == '443') ? 'https' : 'http';
-        apihost = protocol + '://' + apihost;
-    }
-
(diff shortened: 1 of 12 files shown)


core/nodejsActionBase/src/whisk.js
```diff
@@ -28,14 +28,7 @@ var request = require('request');
...
 
-    // export __OW_API_HOST to environment
-    {
-        apihost = apihost || '';
-        var edgeHostParts = apihost.split(':');
-        var protocol = (edgeHostParts.length >= 2  &&  edgeHostParts[1] == '443') ? 'https' : 'http';
-        apihost = protocol + '://' + apihost;
-    }
-
+    apihost = 'https://' + apihost;
```
csantanapr: This new code is not equivalent to the cold being replace when the input parameter `apihost` is falsy like `undefined`, `null`. 
In the old code it will default to type string `http://`
and in the new code it will default to `https://undefined` type string

If some other code tries to parse this string they will see the `undefined` in the uri.

rabbah:
pg2/779.

rabbah:
@domdom82 can you give this a look?

bjustin-ibm:
Rebased/squashed/merged as 68fccef1d12b952ba492d435fbfe00138a4fdc78

PR #1603, opened 2016-12-07
Tighten container reclamation.

### LINKED ISSUE #1554: Invoker should dispense with container if it times out on "run"
The invoker will currently attempt to reuse a container even if a previous run of an action inside that same container timed out - this should be treated as a failed activation and the container reclaimed instead.

rabbah:
If the result of running an action in the container is not successful, delete the container. Makes this is true for both failed init and run.

Fixes #1554 but need to engineer a proper test.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala
@@ -19,13 +19,9 @@ package whisk.core.entity
 import scala.util.Try
 
 import spray.json.DefaultJsonProtocol
-import spray.json.JsBoolean
-import spray.json.JsObject
-import spray.json.JsString
-import spray.json.JsValue
-import spray.json.pimpString
+import spray.json._
 import whisk.common.Logging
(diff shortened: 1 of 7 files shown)


rabbah:
test added.
pg1/1035.

perryibm:
LGTM.  One question, is the precise text of the timed out case part of the contract?

rabbah:
Yes - admittedly I was lazy this time and didn't move this to the `Messages` singleton where I've incrementally accumulated the error responses. At your insistence, I'll do that.

perryibm:
Either that or inline a comment in the test case so that if there is a discrepancy (we'll eventually change wording or internationalize), whoever gets the bug will have a clue.

rabbah:
thanks for the nudge; addressed in latest commit.

perryibm:
Can you fix the conflict in ErrorResponse?

rabbah:
rebased.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/64531db21363a709f73f97b8bbd6a0f9e55425f0.

PR #1604, opened 2016-12-07
Remove --shared on wsk action cmds.

rabbah:


rabbah:
This fixes a breakage from merging the CLI updates that removed `--shared`.

PR #1605, opened 2016-12-07
Posible undefined error for the node whisk.invoke

jasonpet:
Current code does not handle situation where result is undefined when it checks result.error

### DIFF
--- a/core/nodejsActionBase/src/whisk.js
+++ b/core/nodejsActionBase/src/whisk.js
@@ -298,7 +298,12 @@ function post(packet, logger, acceptStatusCode) {
           } else if (!error) {
               // activation failed, set error to API host error response.
               if (!body.error) {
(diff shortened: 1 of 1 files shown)


jasonpet:
running pg3#52

bjustin-ibm:
LGTM

FYI @rabbah

rabbah:
whisk.[invoke|trigger] deprecated ;)

rabbah:
Shall we close this and remove whisk.js entirely instead?

jasonpet:
I assumed that was the plan since @bjustin-ibm never merged it.  I think we are waiting on @mdeuser to remove whisk object references from apigateway before removing whisk.js.

mdeuser:
@jasonpet - the whisk object references are removed in PR #1730.. it's in review state now

Issue #1151, opened 2016-08-29
Clarify wsk CLI update help


mdeuser:
The REST endpoint used by the wsk CLI `update` commands, will create the entity if it does not exist.  The help text for the associated CLI `update` commands should be change to reflect this behavior.

For example,  change the current `wsk action update -h` help from `update an existing action` to something like `update an existing action; if action does not exist, create it.`

For some discussion leading to this issue, refer to issue #38

PR #1609, opened 2016-12-07
Display Proper Message for Update Command Usage

dubeejw:
- Inform users that an entity will be created if it does not exist when an update command is executed

Closes https://github.com/openwhisk/openwhisk/issues/1151

### DIFF
--- a/tools/cli/go-whisk-cli/commands/action.go
+++ b/tools/cli/go-whisk-cli/commands/action.go
@@ -88,7 +88,7 @@ var actionCreateCmd = &cobra.Command{
 
 var actionUpdateCmd = &cobra.Command{
   Use:           "update ACTION_NAME [ACTION]",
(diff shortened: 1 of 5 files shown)


tools/cli/go-whisk-cli/commands/api.go
```diff
@@ -137,7 +137,7 @@ var apiCreateCmd = &cobra.Command{
 
 var apiUpdateCmd = &cobra.Command{
     Use:           "update API_PATH API_VERB ACTION",
-    Short:         wski18n.T("update an existing API"),
+    Short:         wski18n.T("update an existing API, or create a new API if it does not exist"),
```
mdeuser: The `wsk api update` command is not yet implemented; however, the current plan is for it to only update, not create an API.

PR #1611, opened 2016-12-07
Proper handling of Promise.reject().

### LINKED ISSUE #1562: Calling reject() with no reason results in a successful activation result
Consider this action:

```
function main(args) {
  return new Promise((resolve, reject) => {
    reject();
  });
}
```

When invoked, it results in a successful activation:

```
    "response": {
        "status": "success",
        "statusCode": 0,
        "success": true,
        "result": {}
    },
```

Although I would have expected it to fail as I called `reject`.

If I change `reject()` to `reject("with a reason")`, now it fails as expected:

```
    "response": {
        "statusCode": 0,
        "success": false,
        "result": {
            "error": "with a reason"
        }
    },
```

It would be great if `reject()` would actually result in a failed execution.

psuter:
Ensure that Promises rejected with no message result in failed activations.

Fixes #1562.

### DIFF
--- a/core/nodejsActionBase/runner.js
+++ b/core/nodejsActionBase/runner.js
@@ -115,7 +115,13 @@ function NodeActionRunner(whisk) {
                     }).catch(function (error) {
                         // A rejected Promise from the user code maps into a
                         // successful promise wrapping a whisk-encoded error.
-                        resolve({ error: error });
+
+                        // Special case if the user just called `reject()`.
+                        if (!error) {
+                            resolve({ error: {}})
+                        } else {
+                            resolve({ error: error });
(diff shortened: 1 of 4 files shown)


psuter:
PG3/61 has one unrelated failure. Running a new one anyway.

psuter:
Actually, ignore that last message. It's on me to determine whether it's actually unrelated.

psuter:
PG3/65 has the same (only) failure.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/0e99db90eabb9eb894c079e59d6ca9e567b7ca49.

PR #1618, opened 2016-12-08
Bug fix: resolve action before entitlement service in activate path

ioana-blue:
This is a special case for sequences that may contain _ as default namespace in their components. Add test with sequence components that contain _ as default namespace.


ioana-blue:
PG3 67

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/c2aa0bc2a823faa91c92998ce1558d32eafd276b.

PR #1619, opened 2016-12-08
Sequence cleanup

ioana-blue:
Remove old pipe.js code.


common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -122,9 +122,9 @@ protected[core] case class BlackBoxExec(override val image: String, code: Option
     override val pull = image != Exec.BLACKBOX_SKELETON
 }
 
-protected[core] case class SequenceExec(code: String, components: Vector[FullyQualifiedEntityName]) extends Exec(Exec.SEQUENCE) {
-    val image = Exec.imagename(Exec.NODEJS)
-    def size = components.map(c => c.size).reduce(_ + _)
+protected[core] case class SequenceExec(components: Vector[FullyQualifiedEntityName]) extends Exec(Exec.SEQUENCE) {
+    override val image = ""  // image does not have a meaning for sequences anymore
```
rabbah: should we change the type of image then to be an option in the abstract type and make this none?
or make this throw an exception?

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -104,7 +104,7 @@ case class WhiskAction(
     version: SemVer = SemVer(),
     publish: Boolean = false,
     annotations: Parameters = Parameters())
-    extends WhiskEntity(name) {
+    extends WhiskEntity(name) /* with Logging */ {
```
rabbah: remove with logging comment.

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -163,6 +162,9 @@ case class WhiskAction(
                 code map {
                     c => JsObject("code" -> c.toJson, "binary" -> JsBoolean(b.binary))
                 } getOrElse JsObject()
+            case SequenceExec(components) =>
+                //error(this, "Container initializer not supported for sequences")
```
rabbah: case _ => throw exception instead?

rabbah:
"Ioana sequence cleanup" :)  is this personal?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -367,21 +367,12 @@ trait WhiskActionsApi extends WhiskCollectionAPI with SequenceActions with Refer
                 l.memory getOrElse MemoryLimit(),
                 l.logs getOrElse LogLimit())
         } getOrElse ActionLimits()
-        // This is temporary while we are making sequencing directly supported in the controller.
```
rabbah: I suggest we don't yet do this - instead we continue to ignore sequence parameters until schema migration is resolved. Once we give up the invariant we wont be able to disambiguate real vs system injected parameters.

ioana-blue: Agree, as per discussion yesterday. I will revert the two commits that deal with parameters.

ioana-blue:
I applied your suggestions (left the commits in place without squashing for you to see the logical changes). 

re: test. I added a test that checks the first part of what you suggest above. the second part, I'm not sure I understand. `code` is not a member of `SequenceExec` anymore, so how would it make it to the db?

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -177,11 +177,11 @@ protected[core] object Exec
...
                 main.map(m => JsObject(base + ("main" -> JsString(m)))).getOrElse(JsObject(base))
 
-            case SequenceExec(code, comp) =>
-                JsObject("kind" -> JsString(Exec.SEQUENCE), "code" -> JsString(code), "components" -> comp.map(_.qualifiedNameWithLeadingSlash).toJson)
+            case SequenceExec(comp) =>
+                JsObject("kind" -> JsString(Exec.SEQUENCE), "components" -> comp.map(_.qualifiedNameWithLeadingSlash).toJson)
 
             case b: BlackBoxExec =>
-                val base = Map("kind" -> JsString(Exec.BLACKBOX), "image" -> JsString(b.image), "binary" -> JsBoolean(b.binary))
+                val base = Map("kind" -> JsString(Exec.BLACKBOX), "image" -> JsString(b.image.get), "binary" -> JsBoolean(b.binary))
```
markusthoemmes: Please add a comment here and above in `BlackBoxExec` when using `image.get` stating why its safe to do so and that it'll never throw an exception.

ioana-blue: image is of type `Some[String]` , so safe to do the `get`. I can add this comment, sure. 

`image` used to be defined for all Execs. With the new implementation for Sequences, `image` does not make sense for SequenceExec anymore. Using `Option[String]` for image was a suggestion from @rabbah.

rabbah:
Here's an alternative removal of `code` without making image an option.
https://github.com/openwhisk/openwhisk/compare/master...rabbah:alt-seq-cleanup?expand=1

tests/src/common/Wsk.scala
```diff
@@ -153,7 +153,7 @@ trait ListOrGetFromCollection extends FullyQualifiedNames {
         val params = Seq(noun, "get", "--auth", wp.authKey) ++
             Seq(fqn(name)) ++
             { if (summary) Seq("--summary") else Seq() } ++
-            { fieldFilter map { f => Seq(f) } getOrElse Seq() }
+            fieldFilter
```
rabbah: there are several more instances of this in Wsk.scala. Defer to a later commit.

PR #1624, opened 2016-12-08
Add openwhisk home as the location to search for whisk property

houshengbo:
If this function is called by the scala code under the directory or parent directory, we can look up the whisk.property file even recursively until we find it. However, if other repo directories like openwhisk-catalog and openwhisk-package-kafka, are saved in the same level as openwhisk, we will fail to find the whisk.property file under openwhisk, but the test cases located under openwhisk-catalog and openwhisk-package-kafka, still need to read from whisk.property. Currently, we have to load this file by explicitly loading the path of OPENWHISK_HOME in openwhisk-catalog or openwhisk-package-kafka. If we can provide the option to read from OPENWHISK_HOME in WhiskConfig.scala, we do not need to get the OPENWHISK_HOME variable each time it is in need of the test cases under other repo directories.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
+++ b/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
@@ -129,12 +129,16 @@ object WhiskConfig extends Logging {
                 else null
             } else null
 
-        val dir = sys.props.get("user.dir")
-        if (dir.isDefined) {
-            propfile(dir.get, true)
-        } else {
-            null
+        def propfileFromDir(dir: Option[String]): File = {
+            dir.map { propfile(_, true) } getOrElse null
         }
+
+        var dir = sys.props.get("user.dir")
+        val file = propfileFromDir(dir)
+        if (file == null) {
+            dir = sys.env.get("OPENWHISK_HOME")
+            propfileFromDir(dir)
+        } else file
     }
 
     /**


bjustin-ibm:
I'm not familiar enough with this class to properly review this. Deferring to @markusthoemmes

rabbah:
Clarify why we need this? To run openwhisk tests from outside the repo?

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -129,11 +129,23 @@ object WhiskConfig extends Logging {
                 else null
             } else null
 
+        def propfileOpenwhiskHome(): File = {
+            val dir = sys.env.get("OPENWHISK_HOME")
```
rabbah: For your reference: You can rewrite the if-else below as: `dir.map { propfile(_, true) } getOrElse null`.

markusthoemmes:
Also curious why we would need this. Please don't leave the PR description empty.

houshengbo:
I have added the description of this commit.

rabbah:
pg?

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -129,12 +129,16 @@ object WhiskConfig extends Logging {
                 else null
             } else null
 
-        val dir = sys.props.get("user.dir")
```
rabbah: are the changes below accomplished more simply with just this one line change?
`val dir = sys.props.get("user.dir") orElse sys.env.get("OPENWHISK_HOME")`

rabbah: and should the precedence be given to the env var vs the user dir?

jasonpet:
not sure why any of this is needed.  can't you get the properties by using WskProps.
it in turn calls WhiskProperties that already uses OPENWHISK_HOME to find whisk.properties.

jasonpet:
I agree that this update is good for users of WhiskConfig.  just not sure you need it @houshengbo

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
jasonpet: yes, precedence should be given to env var.

markusthoemmes:
`WhiskConfig` shouldn't be used in "blackboxy" tests btw. Always prefer to use the more explicit `WhiskProperties` in those cases.

rabbah:
@jasonpet @houshengbo status?

houshengbo:
@rabbah, after I discussed with @jasonpet, I found out that I can directly import common.WhiskProperties to get the whisk properties as defined, for example, getKafkaHost.

I used to WhiskConfig to get the kafka host without knowing the WhiskProperties, and I thought openwhisk home needed to be put in WhiskConfig as a path to search for the property.

However I can now use WhiskProperties.getKafkaHost instead, so this PR is not that necessary any more.

PR #1626, opened 2016-12-09
Updates to the routemgmt package for the meta API

rabbah:
This pr updates the routemgmt package with changes required to work with the meta-api from #1600.


csantanapr:
PG number?

Issue #1616, opened 2016-12-08
Deleting a package triggers a memory error in the CLI


psuter:
## Environment details:

MacOS

```
$ wsk property get --all
whisk auth              ...
whisk API host          openwhisk.ng.bluemix.net
whisk API version       v1
whisk namespace         _
whisk CLI version       2016-12-07T15:29:55+00:00
whisk API build         2016-12-06T13:02:10Z
whisk API build number  whisk-build-2253
```

## Steps to reproduce the issue:

```
$ wsk package create P
ok: created package P
$ wsk package delete P
runtime error: invalid memory address or nil pointer dereference
Application exited unexpectedly
```

Note that the CLI does get a response:

```
{
  "name": "P",
  "binding": {

  },
  "publish": false,
  "annotations": [],
  "version": "0.0.1",
  "parameters": [],
  "namespace": "psuter@us.ibm.com"
}
```
...and the package is actually deleted.

## Provide the expected results and outputs:

CLI shouldn't crash.
FYI @dubeejw

l2fprod:
got it to with

```
whisk CLI version	2016-12-12T12:10:55+00:00
whisk API build		2016-12-12T12:20:34Z
whisk API build number	whisk-build-2331
```

PR #1634, opened 2016-12-12
Fix Package Delete

dubeejw:
- Ensure that package delete does not throw an invalid memory address or nil pointer deference error

Closes https://github.com/openwhisk/openwhisk/issues/1616#issuecomment-266459365

### DIFF
--- a/tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
+++ b/tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
@@ -515,6 +515,12 @@ class WskBasicUsageTests
 
     behavior of "Wsk packages"
 
+    it should "create, and delete a package" in {
+        val name = "createDeletePackage"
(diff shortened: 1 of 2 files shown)


tests/src/system/basic/WskBasicTests.scala
```diff
@@ -95,6 +95,12 @@ class WskBasicTests
 
     behavior of "Wsk Package CLI"
 
+    it should "create, and delete a package" in {
```
rabbah: WskBasicUsageTests is a better place for this (needs to run once only)?

dubeejw:
PG2 771.

dubeejw:
@rabbah, would be nice to have this PR in before GA.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/7fcce72c0c5f92d68b7dc3450ec4274b810fe1a9.

PR #1636, opened 2016-12-12
Fix API GW CLI issues

mdeuser:
- Handle scenarios when base path ends in `/`
- Fix CLI tests cases so that the test APIs are entirely deleted


mdeuser:
PG3 84

csantanapr:
merged 11d2cd56645747c5fc75b0e23f4e85688d0033df

PR #1641, opened 2016-12-13
Dusting

rabbah:
Adding some missing tests for package and action entitlement as well as tests which cover the action/package implicit right inference.

Remove `apply()` idiom which converted some of our types to `String` in favor of an explicit `asString` method (not using `toString` since that may format types for printing instead of type conversion).

This is further dusting ahead of some changes to the object access model.

@nwspeete-ibm FYI.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/DocInfo.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/DocInfo.scala
@@ -67,9 +67,7 @@ protected[core] class DocRevision private (val rev: String) extends AnyVal {
  * @param rev the document revision, optional; this is an opaque value determined by the datastore
  */
 protected[core] case class DocInfo protected[entity] (id: DocId, rev: DocRevision = DocRevision()) {
-    override def toString =
-        s"""|id: $id
-            |rev: $rev""".stripMargin.replace("\n", ", ")
+    override def toString = s"id: $id, rev: $rev"
 
     override def hashCode = {
         if (rev.empty) {
--- a/common/scala/src/main/scala/whisk/core/entity/DocInfo.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/DocInfo.scala
@@ -34,7 +34,7 @@ import spray.json.deserializationError
  * @param id the document id, required not null
  */
 protected[core] class DocId private (val id: String) extends AnyVal {
-    def apply() = id
+    def asString = id // to make explicit that this is a string conversion
     protected[core] def asDocInfo = DocInfo(this)
     protected[core] def asDocInfo(rev: DocRevision) = DocInfo(this, rev)
     protected[entity] def toJson = JsString(id)
@@ -52,7 +52,7 @@ protected[core] class DocId private (val id: String) extends AnyVal {
  * @param rev the document revision, optional
  */
 protected[core] class DocRevision private (val rev: String) extends AnyVal {
-    def apply() = rev
+    def asString = rev // to make explicit that this is a string conversion
     def empty = rev == null
     override def toString = rev
 }
@@ -67,17 +67,15 @@ protected[core] class DocRevision private (val rev: String) extends AnyVal {
  * @param rev the document revision, optional; this is an opaque value determined by the datastore
  */
 protected[core] case class DocInfo protected[entity] (id: DocId, rev: DocRevision = DocRevision()) {
-    def apply() = id()
-
     override def toString =
-        s"""|id: ${id()}
-            |rev: ${rev()}""".stripMargin.replace("\n", ", ")
+        s"""|id: $id
+            |rev: $rev""".stripMargin.replace("\n", ", ")
 
     override def hashCode = {
         if (rev.empty) {
             id.hashCode
         } else {
-            "$id.$rev".hashCode
+            s"$id.$rev".hashCode
         }
     }
 }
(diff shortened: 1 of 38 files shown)


rabbah:
pg2/777 :airplane: has landed.

tests/src/whisk/core/controller/test/EntitlementProviderTests.scala
```diff
@@ -180,4 +184,240 @@ class EntitlementProviderTests extends ControllerTestCommon {
...
+            (ACTIVATE, someUser, Right(false)),
+            (REJECT, someUser, Right(false)),
+
+            (READ, guestUser, Right(true)),
+            (PUT, guestUser, Right(false)),
+            (DELETE, guestUser, Right(false)),
+            (ACTIVATE, guestUser, Right(false)),
+            (REJECT, guestUser, Right(false)))
+
+        paths.map {
```
markusthoemmes: `foreach` for sideeffects please

tests/src/whisk/core/controller/test/EntitlementProviderTests.scala
```diff
@@ -42,7 +44,9 @@ import whisk.core.entity._
  * "using Specs2RouteTest DSL to chain HTTP requests for unit testing, as in ~>"
  */
 @RunWith(classOf[JUnitRunner])
-class EntitlementProviderTests extends ControllerTestCommon {
+class EntitlementProviderTests
+    extends ControllerTestCommon
+    with ScalaFutures {
```
markusthoemmes: What are you using this trait for? Note that it gives you even simpler convenience methods like

`check.futureValue shouldBe expected`

rabbah: futureValue is for the successful case - some of these tests check that the future failed. this would require using "an exception thrown" matcher; i found the eitherValue to be more convenient for specifying the collection of vectors per test.

tests/src/whisk/core/controller/test/EntitlementProviderTests.scala
psuter: I vote for a `for` loop to spare the `case` :)

tests/src/whisk/core/controller/test/EntitlementProviderTests.scala
```diff
@@ -180,4 +184,240 @@ class EntitlementProviderTests extends ControllerTestCommon {
...
+        paths.map {
+            case (priv, who, expected) =>
+                val check = new PackageCollection(entityStore).implicitRights(
+                    who,
+                    Set(who.namespace()),
+                    priv,
+                    // any user can list any namespace packages
+                    // (because this performs a db view lookup which is later filtered)
+                    Resource(someUser.namespace.toPath, PACKAGES, None))
+                Await.ready(check, requestTimeout).eitherValue.get shouldBe expected
```
psuter: How crazy would it be to check them all concurrently?

common/scala/src/main/scala/whisk/core/entity/WhiskPackage.scala
```diff
@@ -101,7 +101,7 @@ case class WhiskPackage(
     /**
      * Gets binding for package iff this is not already a package reference.
      */
-    def bind = binding map { _ => None } getOrElse Some { Binding(namespace, name) }
+    def bind = binding map { _ => None } getOrElse Some { Binding(namespace.root, name) }
```
psuter: The logic in this line makes my head hurt. How about:

```
if(binding.isDefined) {
    None
} else {
    Some(Binding(namespace.root, name))
}
```

common/scala/src/main/scala/whisk/core/entity/DocInfo.scala
```diff
@@ -67,17 +67,15 @@ protected[core] class DocRevision private (val rev: String) extends AnyVal {
  * @param rev the document revision, optional; this is an opaque value determined by the datastore
  */
 protected[core] case class DocInfo protected[entity] (id: DocId, rev: DocRevision = DocRevision()) {
-    def apply() = id()
-
     override def toString =
-        s"""|id: ${id()}
-            |rev: ${rev()}""".stripMargin.replace("\n", ", ")
+        s"""|id: $id
+            |rev: $rev""".stripMargin.replace("\n", ", ")
```
psuter: Why not `s"id: $id, rev: $rev"`?

common/scala/src/main/scala/whisk/core/entity/DocInfo.scala
```diff
@@ -67,17 +67,15 @@ protected[core] class DocRevision private (val rev: String) extends AnyVal {
...
 
     override def hashCode = {
         if (rev.empty) {
             id.hashCode
         } else {
-            "$id.$rev".hashCode
+            s"$id.$rev".hashCode
```
psuter: Would it be sensible to give a `hashCode` to `DocRevision` and leave this one be the default for `case class`es?

tests/src/whisk/core/controller/test/EntitlementProviderTests.scala
rabbah: this whole suite runs in 3 seconds - eh...
I do intend to replace the couchdb backing store with the mock store coming up in a new pr.

common/scala/src/main/scala/whisk/core/entity/DocInfo.scala
rabbah: DocRevision is a value class and may not define hashCode.

rabbah:
applied companion label. pg2/792.

rabbah:
pg2/803 blue as the sky.

rabbah:
Rebased and resolve conflict.

psuter:
Merged as 424fc75387763a4c5831e9764027fc2d172551ff

PR #1645, opened 2016-12-14
wsk rules don't need explicitly enabling any more so update docs

andreas-ibm:
Tried following the docs to "Creating and using trigger feeds" and the wsk cli barfed at the --enable flag. I was told this is no longer needed, so I've updated the docs with the new command and the new output.


rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/c92bdbdd97aaeb80274f9d8a15fe6fa5478d951a.

PR #1657, opened 2016-12-16
add actionname to marker

RSulzmann:


core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -226,7 +226,7 @@ trait WhiskActionsApi
                         val action = act.resolve(user.namespace)
                         onComplete(entitleReferencedEntities(user, Privilege.ACTIVATE, Some(action.exec))) {
                             case Success(true) =>
-                                transid.started(this, if (blocking) LoggingMarkers.CONTROLLER_ACTIVATION_BLOCKING else LoggingMarkers.CONTROLLER_ACTIVATION)
+                                transid.started(this, if (blocking) LoggingMarkers.CONTROLLER_ACTIVATION_BLOCKING else LoggingMarkers.CONTROLLER_ACTIVATION, s"${action.name}")
```
rabbah: suggestion instead of using string interpolation here: `action.name.toString`.

the action name is available in other log messages for this transid - do we need to log the action name here explicitly?

RSulzmann: Yes, we have the action name in marker invoker_activationRun_start. 
To handle also sequence actions in the perftool, I need to filter a marker by action name at the beginning of the chain in the controller (sequence names do not go to the invoker marker).

I will change to action.name.toString

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -226,7 +226,7 @@ trait WhiskActionsApi
                         val action = act.resolve(user.namespace)
                         onComplete(entitleReferencedEntities(user, Privilege.ACTIVATE, Some(action.exec))) {
                             case Success(true) =>
-                                transid.started(this, if (blocking) LoggingMarkers.CONTROLLER_ACTIVATION_BLOCKING else LoggingMarkers.CONTROLLER_ACTIVATION)
+                                transid.started(this, if (blocking) LoggingMarkers.CONTROLLER_ACTIVATION_BLOCKING else LoggingMarkers.CONTROLLER_ACTIVATION, action.name.toString)
```
rabbah: rather that .toString use .asString.

rabbah:
Closing as abandoned - rebase and reopen when relevant again.

Issue #1610, opened 2016-12-07
CLI Should Include Binary Field


dubeejw:
```'touch x.zip
wsk action create z x.zip --kind nodejs

Response body received:
{
  "name": "z",
  "publish": false,
  "annotations": [],
  "version": "0.0.1",
  "exec": {
    "kind": "nodejs",
    "code": "",
    "binary": true
  },
  "parameters": [],
  "limits": {
    "timeout": 60000,
    "memory": 256,
    "logs": 10
  },
  "namespace": "guest"
}

wsk action get z -v
ok: got action z
{
    "namespace": "guest",
    "name": "z",
    "version": "0.0.1",
    "publish": false,
    "exec": {
        "kind": "nodejs",
        "code": ""
    },
    "limits": {
        "timeout": 60000,
        "memory": 256,
        "logs": 10
    }
}
```
The binary property is defined for some but not all action types (docker and node being two of them).

dubeejw:
@csantanapr, this can be closed as a dupe of https://github.com/openwhisk/openwhisk/issues/1489.

PR #1660, opened 2016-12-16
Add Binary Field to Action Exec

dubeejw:
- A binary field is displayed for Docker and Zip actions

Closes https://github.com/openwhisk/openwhisk/issues/1610


rabbah:
@dubeejw can we do this by going the other way and starting to rip out all the type encodings in the cli?

Issue #418, opened 2016-05-13
wsk action copy only copies exec not parameters or annotations


csantanapr:
From @rabbah "i think it be convenient to copy parameters, and also the annotations.
it is less useful to copy the exec and miss out on all the rest."

dubeejw:
@csantanapr, can you explain this one more? I'm not sure what `wsk action copy` is?

rabbah:
`wsk action create X --copy Y` is a way to create action `X` by copying the `exec` from action `Y`. It does not copy parameters or annotations from `Y` to `X`. The suggestion above was to make the copy more complete.

mdeuser:
@csantanapr - Just confirming...  If this action were part of a package, the above proposal would NOT include the package parameters/annotation values, but only the parameter/annotation values directly associated with the action.

At some point, would there ever be a need to allow the action developer to prohibit a specific action parameter/annotation from being copied?  I suppose this would also affect `wsk action get` as well.

csantanapr:
@mdeuser 
>@csantanapr - Just confirming... If this action were part of a package, the above proposal would NOT include the package parameters/annotation values, but only the parameter/annotation values directly associated with the action.

Correct, I would not copy parameters annotations from package, since this parameter values come into play at invocation time, also user could be copying the action into a new package, which could have a different set of default parameters set.

>At some point, would there ever be a need to allow the action developer to prohibit a specific action parameter/annotation from being copied? I suppose this would also affect wsk action get as well.

Correct, if the subject is using a key that allows him visibility to read that information, then there is no harm to allow them to copy that information.

dubeejw:
Do we still want this functionality? If not, please close. If so, please merge the corresponding PR.

csantanapr:
I think it would be good to have it in and merge #1662, lets continue conversation there, about the concern that @rabbah express.

PR #1662, opened 2016-12-16
Include Parameters and Annotations when Copying an Action

dubeejw:
- When copying an action ensure that parameters and annotations are also copied

Closes https://github.com/openwhisk/openwhisk/issues/418

### DIFF
--- a/tests/src/system/basic/WskActionTests.scala
+++ b/tests/src/system/basic/WskActionTests.scala
@@ -134,6 +134,31 @@ class WskActionTests
             }
     }
 
+    it should "copy an action and ensure exec, parameters, and annotations copied" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+            val origActionName = "orignAction"
+            val copiedActionName = "copiedAction"
+            val params = Map("a" -> "A".toJson)
+            val annots = Map("b" -> "B".toJson)
+
+            assetHelper.withCleaner(wsk.action, origActionName) {
+                val file = Some(TestUtils.getTestActionFilename("wc.js"))
+                (action, _) => action.create(origActionName, file, parameters = params, annotations = annots)
+            }
+
+            assetHelper.withCleaner(wsk.action, copiedActionName) {
+                (action, _) => action.create(copiedActionName, Some(origActionName), Some("copy"))
+            }
+
+            val copiedAction = getJSONFromCLIResponse(wsk.action.get(copiedActionName).stdout)
+            val origAction = getJSONFromCLIResponse(wsk.action.get(copiedActionName).stdout)
+
+            copiedAction.fields("annotations") shouldBe origAction.fields("annotations")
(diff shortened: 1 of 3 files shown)


dubeejw:
PG2 8​24

dubeejw:
@rabbah, any comments on this one?

rabbah:
I'm having second thoughts about this - may be a good one to take on the dev list.
The reason: we don't make it possible to delete a parameter from the CLI.

csantanapr:
@rabbah I think this is good thing to add to the CLI, be able to copy all the metadata.

The reason: we don't make it possible to delete a parameter from the CLI.
What's the concern here about deleting a parameter?

Merging this PR doesn't block us from offering a way to delete a single parameter in future update of CLI.

tools/cli/go-whisk-cli/commands/action.go
```diff
@@ -557,7 +559,9 @@ func parseAction(cmd *cobra.Command, args []string) (*whisk.Action, error) {
 
   action.Name = qName.entityName
   action.Namespace = qName.namespace
-  action.Annotations = annotations.(whisk.KeyValueArr)
+  if !flags.action.copy {
```
rabbah: What does this guard?

tests/src/system/basic/WskActionTests.scala
```diff
@@ -134,6 +134,34 @@ class WskActionTests
...
+            }
+
+            val originActionAnnots = wsk.action.get(origActionName, fieldFilter = Some("annotations")).stdout
+            val copiedActionAnnots = wsk.action.get(copiedActionName, fieldFilter = Some("annotations")).stdout
+            val originActionParams = wsk.action.get(origActionName, fieldFilter = Some("parameters")).stdout
+            val copiedActionParams = wsk.action.get(copiedActionName, fieldFilter = Some("parameters")).stdout
+            val originActionExec = wsk.action.get(origActionName, fieldFilter = Some("exec")).stdout
+            val copiedActionExec = wsk.action.get(copiedActionName, fieldFilter = Some("exec")).stdout
+
+            copiedActionAnnots.substring(copiedActionAnnots.indexOf("\n")) shouldBe originActionAnnots.substring(originActionAnnots.indexOf("\n"))
```
rabbah: Should check that the semver of the copy is 0.0.1 too?

tests/src/system/basic/WskActionTests.scala
```diff
@@ -134,6 +134,34 @@ class WskActionTests
...
+            copiedActionParams.substring(copiedActionParams.indexOf("\n")) shouldBe originActionParams.substring(originActionParams.indexOf("\n"))
```
rabbah: Is the substring necessary?

dubeejw: I updated the test to make it more readable, and added some methods that will make writing tests that require examination of non-activation record responses from the CLI easier in the future.

dubeejw:
PG2 1029

PR #1667, opened 2016-12-19
Print exception that may have occurred during test execution

ioana-blue:
Added missing catch to not lose info on potential exceptions that happened during test and are not logged elsewhere.

### DIFF
--- a/tests/src/common/WskTestHelpers.scala
+++ b/tests/src/common/WskTestHelpers.scala
@@ -80,6 +80,11 @@ trait WskTestHelpers extends Matchers {
 
         try {
             test(wskprops, new AssetCleaner(assetsToDeleteAfterTest, wskprops))
+        } catch {
+            case t: Throwable =>
+                // log the exception that occurred in the test
+                println("Exception occurred during test execution:")
+                println(t)
         } finally {
             // delete assets in reverse order so that was created last is deleted first
             val deletedAll = assetsToDeleteAfterTest.reverse map {


ioana-blue:
Due to the nature of the change, I request a PG-exemption. But let me know if you think a PG is needed.

markusthoemmes:
What are we gaining with this? Shouldn't the exception go uncaught anyway and thus be logged by gradle/JUnit whomever?

rabbah:
Yes except there could be two exceptions and the last masks the first. Here was the thought behind this maybe it leads to something better. Suppose a create failed within the asset helper - this would throw an exception which is caught by the try. But then the cleaner throws an exception also because the delete will fail as well. So you don't see the exception from the create.

tests/src/common/WskTestHelpers.scala
```diff
@@ -80,6 +80,11 @@ trait WskTestHelpers extends Matchers {
 
         try {
             test(wskprops, new AssetCleaner(assetsToDeleteAfterTest, wskprops))
+        } catch {
+            case t: Throwable =>
+                // log the exception that occurred in the test
+                println("Exception occurred during test execution:")
+                println(t)
```
markusthoemmes: Stacktrace not needed/helpful?

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/4f24aad8985c3f8370c4aa1ece8920c810500c1a.

PR #1670, opened 2016-12-20
Fix logging of container ids

markusthoemmes:
Will try to hunt more of those with narrower compiler settings.


rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/4a9fc926f91d483b7e6ad61e90b0e4f8eef03116.

Issue #1673, opened 2016-12-21
go cli bug: wsk action create --sequence with package actions fails with relpath


starpit:
the go cli fails to create a sequence that includes in-package actions, where the actions are not specified with fully qualified names. i will post a test script in a separate comment

here is tail of the buggy output. note the "//" in the components. the Python CLI does not have this bug.

```
}Req Body
{{"namespace":"_","name":"foos","exec":{"kind":"sequence","code":"","components":["//foop/foo","//foop/foo"]}}
}
RESPONSE:Got response with code 403
Response body size is 106 bytes
Response body received:
{
  "error": "The supplied authentication is not authorized to access this resource.",
  "code": 2228744
}
error: Unable to create action: The supplied authentication is not authorized to access this resource. (code 2228744)
```

starpit:
```
#!/usr/bin/env bash

# make sure you're using the go cli
export PATH=~/bin:$PATH

echo "function main(params) {    return {        name: "Step1 " + params.name    };}" > foo.js

# clean up (don't care about errors here)
wsk action delete foos >& /dev/null
wsk action delete foop/foo >& /dev/null
wsk package delete foop >& /dev/null

which wsk | grep -vi python && \
    wsk package create foop && \
    wsk action create foop/foo foo.js && \
    wsk -v action create --sequence foos foop/foo,foop/foo
```

rabbah:
i suspect this will probably affect rules that name actions in packages too.

dubeejw:
Seems to be working fine for me. What version of the CLI are you using? Show me the output of `wsk property get` without your auth key.

dubeejw:
The only way the above can happen is if your namespace is set to nothing in ~/.wskprops.

dubeejw:
Run `wsk property unset --namespace` then try again.

starpit:
i have `NAMESPACE=`, i.e. with nothing but a newline after the equals

dubeejw:
It should be `NAMESPACE=_`.

starpit:
ok, thanks for the help! this still seems like a bug. `NAMESPACE=`` in wskprops should not result in failures?

rabbah:
Closed via https://github.com/openwhisk/openwhisk/commit/2fc75d2cae8bc12b14cd81bba5560798f789d1e5.

PR #1674, opened 2016-12-21
Use Default Namespace if the Property is Blank

dubeejw:
- Use _ as the namespace if the NAMESPACE property in .wskprops is set to blank

Fixes https://github.com/openwhisk/openwhisk/issues/1673


dubeejw:
PG3 116

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/2fc75d2cae8bc12b14cd81bba5560798f789d1e5.

PR #1680, opened 2017-01-02
Change action names of echo actions for different tests

cbickel:
PG3#126 is running

### DIFF
--- a/tests/src/system/basic/WskSequenceTests.scala
+++ b/tests/src/system/basic/WskSequenceTests.scala
@@ -134,10 +134,10 @@ class WskSequenceTests
         (wp, assetHelper) =>
             val xName = "xSequence"
             val sName = "sSequence"
-            val echo = "echo"
+            val echo = "echoForSequences"
 
             // create echo action
-            val file = TestUtils.getTestActionFilename(s"$echo.js")
+            val file = TestUtils.getTestActionFilename("echo.js")
             assetHelper.withCleaner(wsk.action, echo) { (action, actionName) =>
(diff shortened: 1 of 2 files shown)


tests/dat/actions/echoForSequences.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * Returns params, or an empty string if no parameter values are provided
+ */
+function main(params) {
+    return params || {};
+}
```
markusthoemmes: Why is the extra file needed?

cbickel: In some tests, a Sequence with some actionnames is created.
This name is used for the actionname and the filename.
I could change this, if you want, but I think it is more clear, if this test structure is kept.

tests/src/system/basic/WskSequenceTests.scala
```diff
@@ -431,7 +431,7 @@ class WskSequenceTests
                 (trigger, name) => trigger.create(name, parameters = triggerPayload)
             }
             // action
-            val file = TestUtils.getTestActionFilename(s"$actionName.js")
+            val file = TestUtils.getTestActionFilename("echo.js")
```
rabbah: are changes needed to `echo.js`?

rabbah: you only need to rename one (the api gw test)?

cbickel: No, there are no changes to echo.
I'd prefer to change the name in both tests. Otherwise, we have some new tests some day, and they are also called echo. Then we have the same clache again with these tests.

Issue #1107, opened 2016-08-23
Authentication Keys bound to specific Namespaces


markusthoemmes:
# Namespace Keys
## Problem statement

Currently, we issue one username/password pair per subject, which is used via HTTP Basic Authentication to authenticate users using OpenWhisk.

The **Entitlement Service** component of the Controller correlates every single subject in our database with N namespaces, potentially coming from a third party Entitlement services (see **Remote Entitlement Service**.

The 1:N mapping of subject (read: auth-key) to namespaces induces problems in **Activation correlation**: When an action is invoked, it is not clear in which namespace the activation record should live. If User X invokes the action _/namespace1/action1_ intuitively that activation should go in _/namespace1_. But if that user invokes the **shared** action _/whisk.system/util/date_, where does that activation live? It shouldn't live in _/whisk.system_ for sure, as that's a shared entity which shouldn't know of any details provided by the caller of the action. Currently, this is worked-around by having all activation of a subject live in the _default namespace_ (equals to the name of the subject). This is very unintuitive and ultimately has an impact on causality in your system, as you cannot correlate which namespace issued which invocations. Depending on your namespace source, that might or might not be awkward.
## Proposed solution

To solve the issue, we resolve the 1:N mapping of auth-key to namespaces by introducing one auth-key per namespace. To do that, we introduce 1 key per namespace per subject.

The status quo of the subject record is something like:

```
{
  "id": "subjectName",
  "uuid": "something-totally-random",
  "key: "somerandomthing"
}
```

Which then evolves to:

```
{
  "id": "subjectName",
  "namespaces": [
    {
      "name": "A human readable name",
      "uuid": "some-guid-maybe",
      "key": "somerandomthing"
    }
  ]
}
```

(`namespaceName` could equal `subjectName` to reflect current behaviour)

In the `namespaces` map, we then have a Namespace -> auth-key mapping, which can be of arbitrary size. Different users being part of the same namespace don't necessarily need to have the same key for said namespace, leading for a safer system with less key-sharing involved.

API authentication continues to be HTTP Basic Authentication.

With this change in place, every invocation can be tied the namespace associated with the key the invocation was authenticated with. The activation records will then live in that namespace and will only be accessible through `wsk activation list` using a valid key for that namespace.

[...]

PR #1681, opened 2017-01-02
Rework codebase to reflect v2 authentication

markusthoemmes:
This pull-request intends to reflect the new way of authentication with namespace bound keys throughout the codebase.


ddragosd:
@markusthoemmes I'm trying to understand what is the advantage of coupling `identity` (represented by the AuthKey) with the `namespaces`.  I thought the AuthKey is used to authenticate a user/developer and that whether a user is authorized to manage a namespace or not is another concern. Can you share more details to help me understand ?

markusthoemmes:
@ddragosd in a nutshell: Using the old keys we had a hard time associating a namespace to the current usage, as in: To which namespace should the activation of the current invocation go? The namespace of the action? Doesn't work for `whisk.system` actions for example.

We decided the best solution would be to always identify as a user + a namespace, hence each key is associated with a namespace now.

For posterity: Here's the initial proposal: #1107

ddragosd:
thanks @markusthoemmes.

Am I reading correctly that the assumption is that the developer creating the action is also the one invoking the action ? How about actions that are exposed as REST endpoints via the API Gateway which have no OW AuthKey, but have only an OAuth 2 Token from a 3rd party identity provider ? How would we connect these OAuth tokens with some OW credentials used to deploy the action so that we can invoke it ? I think we have to solve this problem if we go down this path. WDYT ? 

This is connected to another aspect that I struggle to understand in OW and that is the separation between the _management_ aspect ( the developers creating and managing actions ) vs the _execution_ aspect (the end-users of an application that invokes actions). In my mind these 2 categories : developers vs end-users have different concerns. They authenticate differently, end-users might not care about namespaces, logs, package names, actions names.

Thinking at the access to the activations and logs of an action, I imagine it's something that concerns more the team that is maintaining that action. If an application that uses an action gets an error back, even though it can access the activations and logs, but if it can't fix the action anyway, what's the utility other than debugging that action and maybe billing ?  

I feel like we're missing another entity that would represent "apps". "Apps" are part of the _execution_ concern b/c they invoke the actions. Apps would subscribe to actions to be able to invoke them. Application info is part of the OAuth token ( the `client_id` that has generated the token ). When an OAuth token is not present we can ask the application to identity itself using an API-KEY for instance.  This way we could show to multiple personas information about activations, _who_ has invoked actions, and _what_ actions. In this way we don't have to solve the problem of matching an OAuth token with an OW Auth Key anymore. We're solving a different problem instead and we can also give access to multiple people to see the analytics for an application. Based on the usage data we can also do billing or charge-backs.

rabbah:
> Am I reading correctly that the assumption is that the developer creating the action is also the one invoking the action ?

No of course not. Any authorized subject can invoke their actions (in their namespace which may be a shared space with other subjects) or actions in another namespace if the action is in a public package.

rabbah:
It's worth noting this change is semantic preserving for openwhisk.

ansible/templates/whisk.properties.j2
```diff
@@ -14,6 +14,8 @@ whisk.ssl.cert={{ openwhisk_home }}/ansible/roles/nginx/files/openwhisk-cert.pem
 whisk.ssl.key={{ openwhisk_home }}/ansible/roles/nginx/files/openwhisk-key.pem
 whisk.ssl.challenge=openwhisk
 whisk.api.host={{ whisk_api_host | default('https://' + (groups['edge'] | first)) }}
+whisk.system.id={{ whisk.system.id }}
+whisk.system.key={{ whisk.system.key }}
```
rabbah: we need only the key, could look up the second, no?

tests/src/whisk/core/controller/test/AuthenticateTests.scala
```diff
@@ -164,16 +164,16 @@ class AuthenticatedRouteTests
         }
     }
 
-    it should "report service unavailable when db lookup fails" in {
+    ignore should "report service unavailable when db lookup fails" in {
```
rabbah: ignored?

markusthoemmes: This is partly WIP. The condition that forced the internal error no longer holds water. I'll need to find a different way to trigger this.

ansible/templates/whisk.properties.j2
markusthoemmes: Depends, will double-check if the lookup is only needed in tests, i'd say: yes.

tests/src/whisk/core/entity/test/DatastoreTests.scala
```diff
@@ -77,16 +77,6 @@ class DatastoreTests extends FlatSpec
         putGetCheck(authstore, auth, WhiskAuth)
     }
 
-    it should "lookup key by uuid" in {
```
rabbah: do we need a replacement test?

tests/src/whisk/core/controller/test/AuthenticateTests.scala
rabbah: maybe I can cherry pick my mock data store :)

tests/src/whisk/core/entity/test/DatastoreTests.scala
markusthoemmes: I don't think so, functionality is gone.

We need testcoverage for the `Identity` object/class though.

markusthoemmes:
PG2 889

ddragosd:
I'm referencing the other PR #1715 which solves the problem with invoking actions without having to provide whisk credentials in case someone else stumbles about the same problem I did.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/compare/189968b19df1...ddee7fdcc820.

Issue #1396, opened 2016-10-13
CLI is not formatting path strings correctly when certain chars are present


testLime:
The failure of a recent test shows that the CLI output of certain paths is not being formatted correctly if that path has certain character present. 

The invalid output appeared as:
`Post https://fakehost/api/v1/namespaces/asdf/actions/asdf%!!(MISSING)!(MISSING)a(MISSING)sdf?blocking=false:`

@dubeejw mentioned that this is a known issue from the CLI's use of a color package.

PR #1683, opened 2017-01-03
Properly Format Strings Containing Precent Characters

dubeejw:
- Remove unnecessary sprintf() calls that cause strings containing '%' characters to be malformed

Closes: https://github.com/openwhisk/openwhisk/issues/1396


dubeejw:
PG3 128

dubeejw:
PG3 129.

PR #1702, opened 2017-01-10
Update copyright-headers

cbickel:
This PR updates all copyright headers to 2017.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ interested in:
 
 ### License
 
-Copyright 2015-2016 IBM Corporation
+Copyright 2015-2017 IBM Corporation
 
 Licensed under the [Apache License, Version 2.0 (the "License")](http://www.apache.org/licenses/LICENSE-2.0.html).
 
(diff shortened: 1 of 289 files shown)


rabbah:
should we check on dev@openwhisk.apache first on what the practice is (unless it's already documented somewhere)?

mrutkows:
@rabbah @cbickel It would be good to ask th epolicy for Apache; I would assume you would leave the earliest (year) copyright "as is" and only update to 2017 as you go/add new function.

csantanapr:
@rabbah @markusthoemmes Here is the Apache information on dealing with headers: https://www.apache.org/legal/src-headers.html

The net is there is standard header for Apache, and doesn't include dates.

The bigger question is do we need to change all our headers to Apache copyright, or just new files, or only for new files once graduated from Incubation.

I think is worth asking in the legal discuss mailing list.

csantanapr:
As a side note, Apache provides an auditing tool [RAT](https://creadur.apache.org/rat/) that helps audit licenses, including parsing the headers.

csantanapr:
FYI: We use RAT in cordova release automation tools https://github.com/apache/cordova-coho/blob/master/src/audit-license-headers.js

PR #1704, opened 2017-01-10
Remove pipecode and code from sequence exec (replaces #1619)

rabbah:
Replaces #1619.


common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -38,24 +38,19 @@ import whisk.core.entity.size.SizeOptionString
...
- *
- * exec: { kind  : one of supported language runtimes
- *         code  : code to execute if kind is supported
- *         image : container name when kind is "blackbox",
- *         main  : a fully-qualified class name when kind is "java" }
+ * An example exec looks like this:
+ * { kind  : one of supported language runtimes
+ *   code  : code to execute if kind is supported
+ *   image : container name when kind is "blackbox",
+ *   main  : a fully-qualified class name when kind is "java" }
```
psuter: That's not accurate anymore (it's on me, though). It should say something like: "The name of the entrypoint function, when using a non-default value. For Java, the name of the main class."

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -559,7 +560,7 @@ class ContainerPool(
             try {
                 info(this, s"making new container because none available")
                 startingCounter.next()
-                makeGeneralContainer(key, containerName, imageName, limits, action.exec.pull)
+                makeGeneralContainer(key, containerName, imageName, limits, action.exec.asInstanceOf[CodeExec[_]].pull)
```
psuter: That seems bold? `makeWhiskContainer` doesn't get called for non-JS actions? If so can we add a `require` at the top? (or at least a comment similar to the other new ones)

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -111,6 +111,9 @@ class Invoker(
         // if the doc revision is missing, then bypass cache
         WhiskAction.get(entityStore, actionid.id, actionid.rev, fromCache = actionid.rev != DocRevision()) onComplete {
             case Success(action) =>
+                // only Exec instances that are subtypes of CodeExec reach the invoker
+                assume(action.exec.isInstanceOf[CodeExec[_]])
```
psuter: all these type checks are making me think we should type-parametrize `WhiskAction[ExecType]` or something, and have methods defined as `def foo[T <: CodeExec](a: WhiskAction[T])`

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
rabbah: `CodeExec` is the root for all actions other than sequences which includes non-js actions and blackbox.

rabbah: Note the `assume(action.exec.isInstanceOf[CodeExec[_]])` in Invoker.scala asserts this cast is ok.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
rabbah: I agree in principle but that's a bigger change to stage separately IMO.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
rabbah: eh - I read it and decided to leave it. So it's on both of us.
I'll amend.

psuter:
What's the PG id?

rabbah:
oops - sorry. pg 2/857.

csantanapr:
@rabbah This means that we can remove [pipe.js](https://github.com/openwhisk/openwhisk-catalog/blob/master/packages/system/pipe.js) and the system package from the catalog?

cc @houshengbo @jasonpet

rabbah:
This has been true for a while - pipe.js has not been used for sequences in a long time.

psuter:
Sqashed and merged as 7159ef9bed6bc446c534251820be241c1256a27d

PR #1706, opened 2017-01-11
Preparation for HA of the controller

cbickel:
This PR is to prepare HA for the controller. The goal is not to make the controller able to be HA.

This PR treats only:
- controller-container has an index
- the port for each controller is unique
- active ack will be sent to the right controller
- Tids, generated by controller will be unique (All tids of controllerX are `tid % numberOfInvokers = x`
- Edit nginx.conf to loadbalance between controllers (if more than one controller is deployed)

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -23,6 +23,7 @@ defaultLimits:
 # port means outer port
 controller:
   port: 10001
+  instances: "{{ groups['controllers'] | length }}"
   blackboxFraction: 0.10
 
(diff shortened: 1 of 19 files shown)


rabbah:
@cbickel we need to open an issue for this pr and discuss the strategy (for example: you can't enable this without disabling the cache. Tracking system load will need to be global across controllers. Should we separate CRUD vs Activate?).

Also, need to reference the issue from the Apache dev list.

PR #1710, opened 2017-01-11
allow parameters on actions to be treated as final

rabbah:
This pull request allows actions to declare their parameters "final" and may not be overridden by invoke-time parameters.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/Parameter.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Parameter.scala
@@ -53,11 +53,11 @@ protected[core] class Parameters protected[entity] (
             name.size + value.size
     }.foldLeft(0 B)(_ + _)
 
-    protected[entity] def ++(p: (ParameterName, ParameterValue)) = {
+    protected[entity] def +(p: (ParameterName, ParameterValue)) = {
         Option(p) map { p => new Parameters(params + (p._1 -> p._2)) } getOrElse this
     }
 
-    protected[entity] def ++(p: ParameterName, v: ParameterValue) = {
+    protected[entity] def +(p: ParameterName, v: ParameterValue) = {
         new Parameters(params + (p -> v))
     }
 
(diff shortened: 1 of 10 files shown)


common/scala/src/main/scala/whisk/core/entity/Parameter.scala
```diff
@@ -53,26 +53,30 @@ protected[core] class Parameters protected[entity] (
             name.size + value.size
     }.foldLeft(0 B)(_ + _)
 
-    protected[entity] def ++(p: (ParameterName, ParameterValue)) = {
+    protected[entity] def +(p: (ParameterName, ParameterValue)) = {
```
markusthoemmes: FWIW regarding scala patterns: The `apply` method for `Parameters` returns an `Option` vs. with `Map` `apply` is the dangerous version which will throw an exception. Having a `get` method instead could be desirable? (I stumbled over this at least once)

rabbah: Fair. I'll nuke it as I've removed almost all of them.

rabbah: https://github.com/openwhisk/openwhisk/pull/1710/commits/12255b8a9fba9250227835f5b0c372c59e234c37.

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -109,6 +109,14 @@ case class WhiskAction(
     require(exec != null, "exec undefined")
     require(limits != null, "limits undefined")
 
+    /** @return true iff action has appropriate annotation. */
+    def hasFinalParamsAnnotation = {
```
markusthoemmes: Want to fold `asBool` https://github.com/openwhisk/openwhisk/pull/1715/files#diff-85adc8c36c3cfcd17be782427443d12eR87 into this pull-request?

tests/src/whisk/core/controller/test/MetaApiTests.scala
```diff
@@ -205,15 +212,20 @@ class MetaApiTests extends ControllerTestCommon with WhiskMetaApi with BeforeAnd
         implicit val tid = transid()
         val methods = Seq(Get, Post, Delete)
 
-        methods.map { m =>
+        methods.foreach { m =>
```
markusthoemmes: ❤️

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -105,15 +143,23 @@ trait WhiskMetaApi extends Directives with PostActionActivation {
...
-                    "__ow_meta_verb" -> method.value.toLowerCase.toJson,
-                    "__ow_meta_path" -> restofPath.toJson,
-                    "__ow_meta_namespace" -> user.namespace.toJson)
-                invokeAction(_, action, Some(JsObject(content)), blocking = true, waitOverride = true)
+            systemKey flatMap { identity =>
+                val invokeParams = if (action.hasFinalParamsAnnotation) {
+                    requestParams -- action.parameters.immutableParameters
+                } else requestParams // in the absence of immutable annotations, return the request untouched
+
+                if (invokeParams.size == requestParams.size) {
```
markusthoemmes: If I understand this correctly, `invokeParams` is redundant. Can this be rewritten as:

```
// only allow execution iff request does not contain parameters declared as final
if(requestParams.keySet.intersect(action.parameters.immutableParameters).nonEmpty) ...
```

I think its more self-explanatory than comparing the sizes? What do you think?

markusthoemmes:
Don't we want this on **any** action? Currently it's only enforced in the Meta API?

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
rabbah: No. 😀

rabbah:
We could make apply this to all parameters but it's arguable whether we want to consider all defined params immutable (vs invoke time prohibition only as in this or) so as not to allow overriding say between binding and package or action and package/binding. For that it's "experimental".  Adopting immutable (and losing default params with overrides in the process) consistently is appealing but also not backward comprible. Invoke time only (as in this pr) can be generalized to all actions though, as you noted.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
rabbah: Yes but intersect could be more expensive so I opted for a size check instead against the original map.

rabbah:
sorry about that - pgs 3/139 and 3/143.

psuter:
Merged as 0eaee0463103e416d12ef07fbe736a9f344ea5e9

PR #1715, opened 2017-01-12
Allow actions to be accessible from a web browser

### LINKED ISSUE #1222: API v2: initial support for blocking calls via GET
We currently support invoking actions via POST on the entity that represents the action.

It has been suggested that it would be useful to support invoking an action via a GET, so that one can simply copy-paste a URL into a browser.

The initial version of this will have limitations (ugly URLs, URLs that expose credentials) but follow on work can address these individually.

So for an initial implementation I propose we do the following:
- [ ] in API v2, in our REST API, add a sub-entity to each action called 'invoke'.  To invoke an action via POST, one would POST to https://username:pw@openwhisk/api/v2/ns/package/action/invoke
- [ ] support a simple transformation to turn query parameters into a JSON object which holds action parameters.  
- [ ] support GET on action/invoke as a blocking invocation 

follow on work could implement a mapping to hide credentials and make shorter URIs by a 1-way hash function, similar in spirit to bit.ly  .    Will leave that feature for a separate issue.

### LINKED ISSUE #1449: Discussion: invoke-only auth keys
I believe (?) our current design for "new" auth keys has a single master key per namespace, which allows all CRUD ops in that namespace, as well as permissions to invoke actions and fire triggers.
Please correct me if this has changed.

Suppose I want to expose actions as REST endpoints which are unauthenticated: I want to allow anyone in the world to hit my REST API, and I'll pay for it.

We could implement this via API Gateway.

However, maybe API Gateway is overkill.   This issue proposes to issue each namespace an auth key with permission to invoke and fire, but not CRUD.

Of course we could go even more fine-grained and associate auth keys with individual entities.  However, doing it only for namespaces is sufficient for some simple use cases.

This came up today as a potential requirement for a mobile app.

rabbah:
Adds route to allow web based activations. Actions invoked this way are anonymous in that the
caller is not authenticated. The intended action must be named in the path as a fully qualified
name as in `/experimental/web/some-namespace/some-package/some-action`. The package is optional in that the action may be in the default package. In which case, the string "default" must be used.

If the action doesn't exist (or the namespace is not valid) a `BadRequest` is generated.
Optionally, the result form the action may be projected based on a named property. As in
`/experimental/web/some-namespace/some-package/some-action/some-property`. If the property
does not exist in the result then a `BadRequest` is generated. By convention, the "html" property will attempt to respond with media type "text/html".

Actions may be exposed to this web proxy by adding an annotation ("export" -> true).
Demo video https://ibm.box.com/s/5c6ignvejihbai3f59uvqcxee9etf0lf.

Note the first commit rebases #1740. 
Fixes #1222.
Fixes #1449.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/EntityPath.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/EntityPath.scala
@@ -170,6 +170,7 @@ protected[core] class EntityName private (val name: String) extends AnyVal {
     def toJson = JsString(name)
     def toPath: EntityPath = EntityPath(name)
     def addPath(e: EntityName): EntityPath = toPath.addPath(e)
+    def addPath(e: Option[EntityName]): EntityPath = e map { toPath.addPath(_) } getOrElse toPath
     override def toString = name
 }
 
--- a/common/scala/src/main/scala/whisk/core/entity/Identity.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Identity.scala
@@ -37,27 +37,43 @@ object Identity extends MultipleReadersSingleWriterCache[Identity, DocInfo] with
     override def cacheKeyForUpdate(i: Identity) = i.authkey
     implicit val serdes = jsonFormat4(Identity.apply)
 
+    /**
+     * Retrieves a key for namespace.
+     * There may be more than one key for the namespace, in which case,
+     * one is picked arbitrarily.
+     */
+    def get(datastore: AuthStore, namespace: EntityName)(
+        implicit transid: TransactionId): Future[Identity] = {
+        implicit val logger: Logging = datastore
+        implicit val ec = datastore.executionContext
+        val ns = namespace.asString
+
+        cacheLookup(ns, {
+            list(datastore, List(ns), limit = 1) map { list =>
+                list.length match {
+                    case 1 =>
+                        rowToIdentity(list.head, ns)
+                    case 0 =>
+                        logger.info(this, s"$viewName[$namespace] does not exist")
+                        throw new NoDocumentException("namespace does not exist")
+                    case _ =>
+                        logger.error(this, s"$viewName[$namespace] is not unique")
+                        throw new IllegalStateException("namespace is not unique")
+                }
+            }
+        })
+    }
+
     def get(datastore: AuthStore, authkey: AuthKey)(
         implicit transid: TransactionId): Future[Identity] = {
         implicit val logger: Logging = datastore
         implicit val ec = datastore.executionContext
 
         cacheLookup(authkey, {
-            list(datastore, authkey) map { list =>
+            list(datastore, List(authkey.uuid.asString, authkey.key.asString)) map { list =>
                 list.length match {
                     case 1 =>
-                        val row = list.head
-                        row.getFields("id", "value") match {
-                            case Seq(JsString(id), JsObject(value)) =>
-                                val subject = Subject(id)
-                                val JsString(uuid) = value("uuid")
-                                val JsString(secret) = value("key")
-                                val JsString(namespace) = value("namespace")
-                                Identity(subject, EntityName(namespace), AuthKey(UUID(uuid), Secret(secret)), Privilege.ALL)
-                            case _ =>
-                                logger.error(this, s"$viewName[${authkey.uuid}] has malformed view '${row.compactPrint}'")
-                                throw new IllegalStateException("identities view malformed")
-                        }
+                        rowToIdentity(list.head, authkey.uuid.asString)
                     case 0 =>
                         logger.info(this, s"$viewName[${authkey.uuid}] does not exist")
                         throw new NoDocumentException("uuid does not exist")
@@ -69,16 +85,30 @@ object Identity extends MultipleReadersSingleWriterCache[Identity, DocInfo] with
         })
     }
 
-    def list(datastore: AuthStore, authkey: AuthKey)(
+    def list(datastore: AuthStore, key: List[Any], limit: Int = 2)(
         implicit transid: TransactionId): Future[List[JsObject]] = {
-        val key = List(authkey.uuid.toString, authkey.key.toString)
         datastore.query(viewName,
             startKey = key,
             endKey = key,
             skip = 0,
-            limit = 2,
+            limit = limit,
             includeDocs = false,
             descending = true,
             reduce = false)
+    }
+
+    private def rowToIdentity(row: JsObject, key: String)(
+        implicit transid: TransactionId, logger: Logging) = {
+        row.getFields("id", "value") match {
+            case Seq(JsString(id), JsObject(value)) =>
+                val subject = Subject(id)
+                val JsString(uuid) = value("uuid")
+                val JsString(secret) = value("key")
+                val JsString(namespace) = value("namespace")
+                Identity(subject, EntityName(namespace), AuthKey(UUID(uuid), Secret(secret)), Privilege.ALL)
+            case _ =>
+                logger.error(this, s"$viewName[$key] has malformed view '${row.compactPrint}'")
+                throw new IllegalStateException("identities view malformed")
+        }
     }
 }
--- a/core/controller/src/main/scala/whisk/core/controller/Meta.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Meta.scala
@@ -19,14 +19,19 @@ package whisk.core.controller
 import scala.concurrent.Future
 import scala.util.Failure
 import scala.util.Success
+import scala.util.Try
 
 import spray.http._
+import spray.http.HttpHeaders.RawHeader
+import spray.http.MediaTypes._
 import spray.http.StatusCodes._
 import spray.httpx.SprayJsonSupport._
 import spray.json._
 import spray.json.DefaultJsonProtocol._
 import spray.routing.Directives
+import spray.routing.RequestContext
 import whisk.common.TransactionId
+import whisk.core.controller.actions.BlockingInvokeTimeout
 import whisk.core.controller.actions.PostActionActivation
 import whisk.core.database._
 import whisk.core.entity._
@@ -34,13 +39,109 @@ import whisk.core.entity.types._
 import whisk.http.ErrorResponse.terminate
 import whisk.http.Messages
 import whisk.core.WhiskConfig
+import whisk.utils.JsHelpers._
 
-/**
- * A singleton object which defines the properties that must be present in a configuration
- * in order to implement the Meta API.
- */
-object WhiskMetaApi {
+private case class Context(
+    method: HttpMethod,
+    headers: List[HttpHeader],
+    path: String,
+    query: Map[String, String],
+    body: Option[JsObject] = None) {
+    val requestParams = query.toJson.asJsObject.fields ++ { body.map(_.fields) getOrElse Map() }
+    val overrides = WhiskMetaApi.reservedProperties.intersect(requestParams.keySet)
+    def withBody(b: Option[JsObject]) = Context(method, headers, path, query, b)
+    def metadata(user: Option[Identity]): Map[String, JsValue] = {
+        Map("__ow_meta_verb" -> method.value.toLowerCase.toJson,
+            "__ow_meta_headers" -> headers.map(h => h.lowercaseName -> h.value).toMap.toJson,
+            "__ow_meta_path" -> path.toJson) ++
+            user.map(u => "__ow_meta_namespace" -> u.namespace.asString.toJson)
+    }
+}
+
+protected[core] object WhiskMetaApi extends Directives {
+    /**
+     * Defines the properties that must be present in a configuration
+     * in order to implement the Meta API.
+     */
     def requiredProperties = Map(WhiskConfig.systemKey -> null)
+
+    /** Reserved parameters that requests may no defined. */
+    val reservedProperties = Set(
+        "__ow_meta_verb",
+        "__ow_meta_headers",
+        "__ow_meta_path",
+        "__ow_meta_namespace")
+
+    /** Supported media types for action response. */
+    val mediaTranscoders = Map(
+        "html" -> resultAsHtml _,
+        "text" -> resultAsText _,
+        "json" -> resultAsJson _,
+        "http" -> resultAsHttp _)
+
+    /** Default media extensions for action projection. */
+    val defaultMediaProjection: Map[String, Option[List[String]]] = Map(
+        "html" -> Some(List("html")),
+        "text" -> Some(List("text")),
+        "json" -> None,
+        "http" -> None)
+
+    val supportedMediaTypes = mediaTranscoders.keySet
+
+    protected def resultAsHtml(result: JsValue): RequestContext => Unit = result match {
+        case JsString(html) => respondWithMediaType(`text/html`) { complete(OK, html) }
+        case _              => complete(BadRequest, Messages.invalidMedia(`text/html`))
+    }
+
+    protected def resultAsText(result: JsValue): RequestContext => Unit = {
+        result match {
+            case r: JsObject  => complete(OK, r.prettyPrint)
+            case r: JsArray   => complete(OK, r.prettyPrint)
+            case JsString(s)  => complete(OK, s)
+            case JsBoolean(b) => complete(OK, b.toString)
+            case JsNumber(n)  => complete(OK, n.toString)
+            case JsNull       => complete(OK)
+        }
+    }
+
+    protected def resultAsJson(result: JsValue): RequestContext => Unit = {
+        result match {
+            case r: JsObject => complete(OK, r)
+            case r: JsArray  => complete(OK, r)
+            case _           => complete(BadRequest, Messages.invalidMedia(`application/json`))
+        }
+    }
+
+    protected def resultAsHttp(result: JsValue): RequestContext => Unit = {
+        result match {
+            case JsObject(fields) => Try {
+                val headers = fields.get("headers").map {
+                    case JsObject(hs) => hs.map {
+                        case (k, JsString(v))  => RawHeader(k, v)
+                        case (k, JsBoolean(v)) => RawHeader(k, v.toString)
+                        case (k, JsNumber(v))  => RawHeader(k, v.toString)
+                        case _                 => throw new Throwable("Invalid header")
+                    }.toList
+
+                    case _ => throw new Throwable("Invalid header")
+                } getOrElse List()
+
+                val code = fields.get("code").map {
+                    case JsNumber(c) =>
+                        // the following throws an exception if the code is
+                        // not a whole number or a valid code
+                        StatusCode.int2StatusCode(c.toIntExact)
+                    case _ => throw new Throwable("Illegal code")
+                } getOrElse (OK)
+
+                respondWithHeaders(headers) {
+                    complete(code, JsObject())
+                }
+            } getOrElse complete(BadRequest, Messages.invalidMedia(`message/http`))
+
+            case _ => complete(BadRequest, Messages.invalidMedia(`message/http`))
+        }
+    }
 }
 
 trait WhiskMetaApi extends Directives with PostActionActivation {
@@ -53,26 +154,34 @@ trait WhiskMetaApi extends Directives with PostActionActivation {
     /** Store for identities. */
     protected val authStore: AuthStore
 
-    /** The route prefix e.g., /experimental/package-name. */
-    protected val routePrefix = pathPrefix("experimental")
+    /** The route prefix e.g., /experimental. */
+    protected val routePath = "experimental"
+
+    /** The prefix for anonymous invokes e.g., /experimental/web. */
+    protected val anonymousInvokePath = "web"
 
     /** The name and apikey of the system namespace. */
     protected lazy val systemKey = AuthKey(whiskConfig.systemKey)
     protected lazy val systemIdentity = Identity.get(authStore, systemKey)(TransactionId.controller)
 
-    /** Reserved parameters that requests may no defined. */
-    protected lazy val reservedProperties = Set("__ow_meta_verb", "__ow_meta_path", "__ow_meta_namespace")
+    private val routePrefix = pathPrefix(routePath)
+    private val anonymousInvokePrefix = pathPrefix(anonymousInvokePath)
 
     /** Allowed verbs. */
     private lazy val allowedOperations = get | delete | post
 
-    /** Extracts the HTTP method, query params and unmatched (remaining) path. */
+    private lazy val validNameSegment = pathPrefix(EntityName.REGEX.r)
+    private lazy val packagePrefix = pathPrefix("default".r | EntityName.REGEX.r)
+    private lazy val extensionSplitter = EntityName.REGEX.dropRight(2).concat("""\.([a-zA-Z0-9.+-]+)\z""").r
+
+    /** Extracts the HTTP method, headers, query params and unmatched (remaining) path. */
     private val requestMethodParamsAndPath = {
         extract { ctx =>
             val method = ctx.request.method
             val params = ctx.request.message.uri.query.toMap
             val path = ctx.unmatchedPath.toString
-            (method, params, path)
+            val headers = ctx.request.headers
+            Context(method, headers, path, params)
         }
     }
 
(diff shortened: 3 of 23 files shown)


rabbah:
Suggestions from the dev list (credit: Felix Meschberger):

- As for the status codes, 404/Not Found for non-existing actions rather than 400/Bad Request.
- And 404 as well for the case of the action not being available anonymously.
- Missing support for the requested content type could be 406/Not Acceptable (or also 404/Not Found).

rabbah:
note to reviewers: I removed two commits which will come by way of #1720 so next travis will break pending a rebase.

common/scala/src/main/scala/whisk/core/entity/EntityPath.scala
```diff
@@ -170,6 +170,7 @@ protected[core] class EntityName private (val name: String) extends AnyVal {
     def toJson = JsString(name)
     def toPath: EntityPath = EntityPath(name)
     def addPath(e: EntityName): EntityPath = toPath.addPath(e)
+    def addPath(e: Option[EntityName]): EntityPath = e map { toPath.addPath(_) } getOrElse toPath
```
markusthoemmes: Is this used that often, that it's worth hiding the `Option` handling?

common/scala/src/main/scala/whisk/core/entity/Identity.scala
```diff
@@ -37,27 +37,38 @@ object Identity extends MultipleReadersSingleWriterCache[Identity, DocInfo] with
     override def cacheKeyForUpdate(i: Identity) = i.authkey
     implicit val serdes = jsonFormat4(Identity.apply)
 
+    def get(datastore: AuthStore, namespace: EntityName)(
+        implicit transid: TransactionId): Future[Identity] = {
+        implicit val logger: Logging = datastore
+        implicit val ec = datastore.executionContext
+        val ns = namespace.asString
+
+        cacheLookup(ns, {
```
markusthoemmes: As noted in a private chat: Can this cause problems with the `cacheKeyForUpdate` above?

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -16,23 +16,135 @@
...
+            "__ow_meta_headers" -> headers.toJson,
+            "__ow_meta_path" -> path.toJson) ++
+            namespace.map(ns => "__ow_meta_namespace" -> ns.toJson)
+    }
+
+    protected def resultAsHtml(result: JsValue): RequestContext => Unit = {
+        Try {
+            val JsString(str) = result
+            val input = new ByteArrayInputStream(str.getBytes)
+            scala.xml.XML.load(input)
```
markusthoemmes: Is it necessary to parse this? Note that html5 is not necessary valid XML and thus this might fail. Also this puts unnecessary load on the controller i'd say.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -16,23 +16,135 @@
...
+    }
+
+    protected def resultAsJson(result: JsValue): RequestContext => Unit = {
+        result match {
+            case r: JsObject  => complete(OK, r)
+            case r: JsArray   => complete(OK, r)
+            case JsString(s)  => complete(OK, s)
+            case JsBoolean(b) => complete(OK, b.toString)
+            case JsNumber(n)  => complete(OK, n.toString)
+            case JsNull       => complete(OK)
```
markusthoemmes: Are those even valid JSON responses? JSON only allows object or array as the top-level type.

rabbah: A proper JSON response is indeed either an array or object, not a primitive type.
So the marshaler could treat these as error cases instead and if the caller desired to serialize primitive values then they must request `text` as the extension.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -16,23 +16,135 @@
...
+
+                val code = fields.get("code").map {
+                    case JsNumber(c) =>
+                        val status = c.toIntExact // throws exception if not whole
+                        if (status >= 100 && status < 600) {
+                            status
+                        } else throw new Throwable("Illegal code")
+
+                    case _ => throw new Throwable("Illegal code")
+                } getOrElse 200
```
markusthoemmes: You might want to check out https://github.com/spray/spray/blob/master/spray-http/src/main/scala/spray/http/StatusCode.scala#L29-L35 which also throws an `Exception` on an illegal code and does all the standard handling for you.

common/scala/src/main/scala/whisk/core/entity/Identity.scala
rabbah: No because `cacheKeyForUpdate` is for _updates_ and this is a read which provides an explicit key to lookup.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
rabbah: ooops - left over from early debugging. will nuke.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -16,23 +16,135 @@
...
+    /** Supported media types for action response. */
+    val mediaTranscoders = Map(
+        "html" -> resultAsHtml _,
+        "text" -> resultAsText _,
+        "js" -> resultAsJson _,
+        "http" -> resultAsHttp _)
+
+    val supportedMediaTypes = mediaTranscoders.keySet
+
+    def metadata(method: String, headers: String, path: String, namespace: Option[String]) = {
```
markusthoemmes: Why is `headers` a `String` vs. a `List[String]`?

rabbah: Fair question. I was avoid two mkstrings. You'll notice on like 54 there's a `headers.mkString("\n")` which will become `headers.map(_.toString)` and then on line 77 the code changes to `headers.mkString("\n").toJson`. I'm not opposed to changing it, it's probably cleaner. The reason `metadata` is a public method is to expose it to the test, so the added typing is arguably less important.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -44,27 +156,35 @@ trait WhiskMetaApi extends Directives with PostActionActivation {
...
-    protected lazy val reservedProperties = Set("__ow_meta_verb", "__ow_meta_path", "__ow_meta_namespace")
+    private val routePrefix = pathPrefix(routePath)
+    private val anonymousInvokePrefix = pathPrefix(anonymousInvokePath)
 
     /** Allowed verbs. */
     private lazy val allowedOperations = get | delete | post
 
-    /** Extracts the HTTP method, query params and unmatched (remaining) path. */
+    private lazy val validNameSegment = pathPrefix(EntityName.REGEX.r)
+    private lazy val packagePrefix = pathPrefix("""\Adefault\z""".r | EntityName.REGEX.r)
```
markusthoemmes: I believe `"default".r` is sufficient and equivalent to `"""\Adefault\z""".r` as regexes are anchored in Scala by default.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
markusthoemmes: Yes, the `headers.mkString` feels quite ugly though. I'd +1 for changing it.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -16,23 +16,135 @@
...
+        "__ow_meta_verb",
+        "__ow_meta_headers",
+        "__ow_meta_path",
+        "__ow_meta_namespace")
+
+    /** Supported media types for action response. */
+    val mediaTranscoders = Map(
+        "html" -> resultAsHtml _,
+        "text" -> resultAsText _,
+        "js" -> resultAsJson _,
```
markusthoemmes: Why use a non-standard `js` vs `json`? Note that they are *not* the same (JSON is way more strict than pure JavaScript notation). I'd change this to json.

rabbah: hence the relaxed marshaling of json values.

common/scala/src/main/scala/whisk/core/entity/EntityPath.scala
rabbah: No but it's going to exist somewhere (e.g. Meta.api), might as well put it here for convenience.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
markusthoemmes: Don't know how to feel about both to be honest. Feels a bit unclean and non-standard. Do I read it correctly though, that `string`, `boolean` and `number` will be sent as `text/plain` vs objects and arrays being sent as `application/json`?

rabbah: That's the correct interpretation of the code as it is.  @sjfink want to further justify the `js` choice vs `json`?

core/controller/src/main/scala/whisk/core/controller/Meta.scala
rabbah: OK - then might as well pass the headers as an array of strings to the action.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -19,20 +19,127 @@ package whisk.core.controller
...
+                val code = fields.get("code").map {
+                    case JsNumber(c) =>
+                        // the following throws an exception if the code is
+                        // not a whole number or a valid code
+                        StatusCode.int2StatusCode(c.toIntExact)
+                    case _ => throw new Throwable("Illegal code")
+                } getOrElse (OK)
+
+                respondWithHeaders(headers) {
+                    complete(code, JsObject())
```
markusthoemmes: Why not also return a `body` field of the JSON result as a `String` here? The user could then set `Content-Type` via the headers, effectively giving him full control over the HTTP response.

rabbah: :) Yes, I know. I got lazy.

rabbah: :) Yes, I know. I got lazy. You shall have it and it will be beautiful.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -44,27 +151,35 @@ trait WhiskMetaApi extends Directives with PostActionActivation {
...
+    private val routePrefix = pathPrefix(routePath)
+    private val anonymousInvokePrefix = pathPrefix(anonymousInvokePath)
 
     /** Allowed verbs. */
     private lazy val allowedOperations = get | delete | post
 
-    /** Extracts the HTTP method, query params and unmatched (remaining) path. */
+    private lazy val validNameSegment = pathPrefix(EntityName.REGEX.r)
+    private lazy val packagePrefix = pathPrefix("default".r | EntityName.REGEX.r)
+    private lazy val extensionSplitter = EntityName.REGEX.dropRight(2).concat("""\.([a-zA-Z0-9.+-]+)\z""").r
```
markusthoemmes: As the `anchored` comment from above seems to be true, should we alter `EntityName.REGEX` to be `"""([\w]|[\w][\w@ .-]*[\w@.-]+)"""` thus removing the nasty `dropRight(2)` and end up with: s"""${EntityName.REGEX\.([a-zA-Z0-9.+-]+)""".r`?

rabbah: Rather not mess with it because I'll have to think about the implications of that (i.e., where are all the places we use that regex match). I think being explicit does not hurt (i.e., specifying the anchors and make it clear this is a full match). I should add a comment to explain the drop right though in that case.

rabbah: As expected - removing the anchors from EntityName.REGEX fails all the unit tests...

rabbah: OK now I remember - If look though the implementation of `pathPrefix` matching for a regex. It uses `findPrefixMatchOf` which will not match the entire segment.

rabbah: The regex is `([\w]|[\w][\w@ .-]*[\w@.-]+)` but perhaps should be `([\w]+|[\w][\w@ .-]*[\w@.-]+)` instead.

markusthoemmes: that means my comment on `"""default""".r` was wrong as well I suppose.

rabbah: No - it means we can change the EntityName.REGEX. Since "default" is a complete match it's OK.

rabbah: And your observations shows there's a bug in the meta route matcher.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -19,20 +19,127 @@ package whisk.core.controller
...
+        "html" -> resultAsHtml _,
+        "text" -> resultAsText _,
+        "js" -> resultAsJson _,
+        "http" -> resultAsHttp _)
+
+    val supportedMediaTypes = mediaTranscoders.keySet
+
+    def metadata(method: String, headers: List[String], path: String, namespace: Option[String]) = {
+        Map("__ow_meta_verb" -> method.toLowerCase.toJson,
+            "__ow_meta_headers" -> headers.toJson,
```
markusthoemmes: I'm still wondering: What will this array look like now? Something like

```
["Authentication: Basic Foo", "Custom-Header": "bar"]
```
?

Would make sense to even pass the headers as a `Map` to make header handling closer to what *express* and other frameworks do in node.js for example?

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -19,20 +19,127 @@ package whisk.core.controller
...
+    /** Supported media types for action response. */
+    val mediaTranscoders = Map(
+        "html" -> resultAsHtml _,
+        "text" -> resultAsText _,
+        "js" -> resultAsJson _,
+        "http" -> resultAsHttp _)
+
+    val supportedMediaTypes = mediaTranscoders.keySet
+
+    def metadata(method: String, headers: List[String], path: String, namespace: Option[String]) = {
```
markusthoemmes: I might well be missing something, but would it make sense to have all those types be the strongest possible and  do all the `toString` magic in this method? I'd rather push working with `String` as far out as possible.

psuter:
Some high-level comments from a first, superficial, pass:

* `"export": true` is a very generic-sounding annotation. Maybe `web_export`? (do we CamelCase?)
* I'm a concerned with the monumental work that comes with trying to encode an HTTP request into JSON fields (passed to the action), and interpreting a JSON object as an HTTP response. I'm afraid that conventions can only ever get us the easy cases, and that soon enough we'll be maintaining and documenting dozens of them. Could we instead resort to some externally documented schema? E.g. HAR has subsections for [requests](http://www.softwareishard.com/blog/har-12-spec/#request) and [responses](http://www.softwareishard.com/blog/har-12-spec/#response), but they may be other formats. Within actions, I'd expect that we will rapidly support translating these encoded objects into user-friendly libraries (e.g. `request`).
* With the comment above in mind, I would be in favor of simplifying the implementation and removing path projection (the path would be one of the components passed to the action, which can decide how to handle it).

rabbah:
It's fair to change the annotation. When this is moved out of experimental the annotation should then be promoted to a first class property which might require a schema change.

Note that the http bits are the last commit. Much of the work in place is to enable the web based activation. For that and for making standard actions accessible without buying into http request/response, the projection is still useful. That said as you noticed the action does receive the unmatched parts and can decide how to respond anyway. This I posit makes for a gentler migration for actions. For moving express apps the counter is likely true and your points are well taken. 

For the http bits what I supported is header setting and nothing else. This was motivated by a use case around oauth flow where having redirects is helpful.

Your suggestion of not reinventing the wheel and using some accepted schema based approach is all goodness. I'm for it.

rabbah:
@csantanapr any thoughts on this pr?
@markusthoemmes @psuter I think I addressed the actionable issues from your first pass. I would like to defer the http schema handling to a future enhancement.

rabbah:
PG3/176 approved.

rabbah:
PG3/180 approved.

csantanapr:
@rabbah What would be the problem to use `.json` instead of `.js` in the web api url `.../action.json` ?
I think using `.json` is more clear to talk about `application/json` instead of `.js` or maybe I'm missing something.

rabbah:
Great comments @csantanapr. We can add CORS support - it's in the controller already, just have to wrap these new routes with it -- I think at cursory glance. Yes images could be supported. I opted not to include it here but did try it (with an svg). I agree on POST but adding it without supporting form-urlencoded input will be broken, so we'll have to add both at once. Lastly, it's .js not .json deliberately. This is raised by Markus also. It would make projecting a JSON value (string) for example an error (one must use a text extension - similarly for the other value types). So it's a convenience so that if one is projecting top level fields or nested fields they're not constantly changing both the extension and the projection at the same time. I can be swayed toward stricter form though. I'd like to consider all but the last as additive features to address as future enhancements.

csantanapr:
@rabbah My comment was just feedback for additive future enhancements to discuss, and possible to break into different issues/PRs .

csantanapr:
No magic 🌮 🌮

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -448,6 +449,10 @@ trait WhiskMetaApi extends Directives with PostActionActivation {
                 // blocking invoke which got queued instead
                 complete(Accepted, JsObject("code" -> transid.id.toJson))
 
+            case Failure(t: BlockingInvokeTimeout) =>
```
csantanapr: This will give us the tid for api-experimental meta calls that go beyond the limit timeout. 👍

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -34,13 +38,110 @@ import whisk.core.entity.types._
...
+    headers: List[HttpHeader],
+    path: String,
+    query: Map[String, String],
+    body: Option[JsObject] = None) {
+    val requestParams = query.toJson.asJsObject.fields ++ { body.map(_.fields) getOrElse Map() }
+    val overrides = WhiskMetaApi.reservedProperties.intersect(requestParams.keySet)
+    def withBody(b: Option[JsObject]) = Context(method, headers, path, query, b)
+    def metadata(user: Option[Identity]): Map[String, JsValue] = {
+        Map("__ow_meta_verb" -> method.value.toLowerCase.toJson,
+            "__ow_meta_headers" -> headers.map(h => h.name -> h.value).toMap.toJson,
```
markusthoemmes: Should we normalize the `name` to be lowercased as well? There is `lowercasedName` on `HttpHeader`

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -34,13 +38,110 @@ import whisk.core.entity.types._
...
+    val supportedMediaTypes = mediaTranscoders.keySet
+
+    protected def resultAsHtml(result: JsValue): RequestContext => Unit = {
+        Try {
+            val JsString(str) = result
+            str
+        } match {
+            case Success(xml) => respondWithMediaType(`text/html`) { complete(OK, xml) }
+            case Failure(t)   => complete(BadRequest, Messages.invalidMedia(`text/html`))
+        }
```
markusthoemmes: Might as well be just a `match`?

```
result match {
    case JsString(str) => respondWithMediaType(`text/html`) { complete(OK, str) }
    case _ => complete(BadRequest, Messages.invalidMedia(`text/html`))
}
```

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -190,77 +448,108 @@ trait WhiskMetaApi extends Directives with PostActionActivation {
...
-                Future.failed(RejectRequest(InternalServerError))
+        actionLookup flatMap { action =>
+            action.annotations.asBool("web-export").filter(b => b).map {
+                _ =>
+                    info(this, s"${action.fullyQualifiedName(true)} is exported")
+                    Future.successful(action)
+            } getOrElse {
+                info(this, s"${action.fullyQualifiedName(true)} not exported")
+                Future.failed(RejectRequest(NotFound))
+            }
```
markusthoemmes: Minor scalaism nit: You can convert an `Option` to a `Boolean` by using `exists`:

```
if(action.annotations.asBool("web-export").exists(identity)) {
    info(this, s"${action.fullyQualifiedName(true)} is exported")
    Future.successful(action)
} else {
    info(this, s"${action.fullyQualifiedName(true)} not exported")
    Future.failed(RejectRequest(NotFound))
}
```

You decide what you like more.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -190,77 +448,108 @@ trait WhiskMetaApi extends Directives with PostActionActivation {
...
+                case Some(actionName) =>
+                    info(this, s"'${pkg.name}' maps '${method.name}' to action '${actionName}'")
+                    Future.successful(EntityName(actionName), pkg.parameters)
+
+                case _ =>
+                    info(this, s"'${pkg.name}' is missing 'meta' annotation or action name for '${method.name.toLowerCase}'")
+                    isMetaPackage match {
+                        case Some(true) => Future.failed(RejectRequest(MethodNotAllowed))
+                        case _          => Future.failed(RejectRequest(NotFound))
+                    }
```
markusthoemmes: Same as below on converting `Option[Boolean]` to `Boolean`, handling `None` as `false`:

```
val isMetaPackage = pkg.annotations.asBool("meta").exists(identity)
if(isMetaPackage) {
    pkg.annotations.asString(method.name.toLowerCase).map { actionName => 
        info(this, s"'${pkg.name}' maps '${method.name}' to action '${actionName}'")
        Future.successful(EntityName(actionName), pkg.parameters)
    }.getOrElse {
        info(this, s"'${pkg.name}' is missing action name for '${method.name.toLowerCase}'")
        Future.failed(RejectRequest(MethodNotAllowed))
    }
} else {
    info(this, s"'${pkg.name}' is missing 'meta' annotation")
    Future.failed(RejectRequest(NotFound))
}
```

To me that is easier to follow what exactly is happening and as a bonus, we get a disambiguated log message.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -115,67 +224,216 @@ trait WhiskMetaApi extends Directives with PostActionActivation {
...
+
+    private def activate(identity: Identity, action: WhiskAction, context: Context, behalfOf: Option[Identity] = None)(
+        implicit transid: TransactionId): Future[(ActivationId, Option[WhiskActivation])] = {
+        // precedence order for parameters:
+        // package.params -> action.params -> query.params -> request.entity (body) -> augment arguments (namespace, path)
+        val invokeParams = if (action.hasFinalParamsAnnotation) {
+            context.requestParams -- action.parameters.immutableParameters
+        } else context.requestParams // in the absence of immutable annotations, return the request untouched
+
+        if (invokeParams.size == context.requestParams.size) {
```
markusthoemmes: Is there a reason you used the `intersect` method proposed by me above, but not in this case?

rabbah: as we discussed - the reason was tied to where the immutable parameters are checked against the incoming parameters... the original thought was not to leak this to WhiskAction but that was already happening with the presence of the annotation check there anyway. Thanks for the patch that improves it.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
rabbah: left over from debugging/parsing xml.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
rabbah: thank you for the suggestion.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -527,22 +522,20 @@ trait WhiskMetaApi extends Directives with PostActionActivation {
...
-                case Some(actionName) =>
+            if (isMetaPackage) {
+                pkg.annotations.asString(method.name.toLowerCase).map { actionName =>
+                    // if action name is defined as a string, accept it, else fail request
                     info(this, s"'${pkg.name}' maps '${method.name}' to action '${actionName}'")
                     Future.successful(EntityName(actionName), pkg.parameters)
-
-                case _ =>
+                } getOrElse {
                     info(this, s"'${pkg.name}' is missing 'meta' annotation or action name for '${method.name.toLowerCase}'")
```
markusthoemmes: That log line should be 

```
"'${pkg.name}' is missing action name for '${method.name.toLowerCase}'"
```

rabbah: amended.

rabbah:
Rebased and squashed.

rabbah:
Rebased again to pick up the library bumps. Also added a CLI test. Removing blocked tag because this PR can be merged as is, or if staged with #1740, rebasing there after will remove the over lapping commits. The last commit in this pr should be squashed into the one that precedes it but is left on its own for the reviewer's benefit. PG1/1106 reapproved with latest test.

csantanapr:
#1740 is ready to be merge or your waiting on more reviews?

rabbah:
It's ready pending new input from @markusthoemmes.

sjfink:
pushed in b5be59f..09471ee

Issue #1705, opened 2017-01-10
Whisk.invoke object is not working in Swift3 container


paulcastro:
The Whisk.invoke object is not working correctly in the Swift 3 container.  If you try to use the object to invoke an action, the invoking action will always return:

```
"result": {
            "error": "The action did not return a dictionary."
        }

```

paulcastro:
There appears to be two necessary changes:

1. the open whisk environment variable that has the hostname now seems to include the protocol.  We'll need to parse that
2. We should leave the dispatch group after the callback, not in the callback.  I'm not quite sure how this worked before, but its broken now.

paulcastro:
I'll add a function that uses URLSession in case we want to move to that in the future.  The URLCredential in Linux doesn't have a complete implementation yet so we'd be limited to invoking actions on OpenWhisk installations that have valid certificates if we switch to it now.  This may be the preferred behavior but  not ideal for development environments.

paulcastro:
Re-activating test cases for Whisk object.

paulcastro:
Fixed, has passed tests.  Will submit PR

paulcastro:
Merged PR #1718

PR #1718, opened 2017-01-13
Whisk object in Swift 3 container

paulcastro:
Fixes Issue #1705 where Whisk object was failing to invoke actions and triggers in the Swift 3 container. Blue PG 871.  This fix:

- now parses for the protocol prefix in the __OW_API_HOST environment variable
- refactors how we wait for the async call to return.  We leave the dispatch group at the end of the post function call instead of in the callback passed to the post function call.  

I also added a URLSession based function that can replace the KituraNet implementation.  Its more limited right now (can't handle self-signed certificates because URLCredential is not completely implemented in the Linux-based Foundation) so this function is unused right now.

### DIFF
--- a/tests/src/system/basic/Swift3WhiskObjectTests.scala
+++ b/tests/src/system/basic/Swift3WhiskObjectTests.scala
@@ -42,7 +42,7 @@ class Swift3WhiskObjectTests
 
     behavior of "Swift 3 Whisk backend API"
 
-    ignore should "allow Swift actions to invoke other actions" in withAssetCleaner(wskprops) {
+    it should "allow Swift actions to invoke other actions" in withAssetCleaner(wskprops) {
         (wp, assetHelper) =>
             // use CLI to create action from dat/actions/invokeAction.swift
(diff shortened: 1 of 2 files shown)


paulcastro:
GCD change may just be stylistic - root cause is the protocol addition.  Will update code to use Swift 3 defer structure to provide single exit point for async function calls instead of exiting group in the callback.

paulcastro:
Updated code to use defer.  PG Blue #884

tests/src/system/basic/Swift3WhiskObjectTests.scala
```diff
@@ -42,7 +42,7 @@ class Swift3WhiskObjectTests
 
     behavior of "Swift 3 Whisk backend API"
 
-    ignore should "allow Swift actions to invoke other actions" in withAssetCleaner(wskprops) {
+    it should "allow Swift actions to invoke other actions" in withAssetCleaner(wskprops) {
```
csantanapr: happy to see enable test cases

PR #1720, opened 2017-01-14
Minor fixes

rabbah:
Fix couple of bugs in JsHelpers where there was an assumption that nested values are instances of JsObject. Move helper to whisk.common and make trait a singleton instead.

Also prevent logging of "malformed credentials" when there are none presented.
PG2/881.

### DIFF
--- a/common/scala/src/main/scala/whisk/utils/JsHelpers.scala
+++ b/common/scala/src/main/scala/whisk/utils/JsHelpers.scala
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.utils
+
+import spray.json.JsObject
+import spray.json.JsValue
+
+object JsHelpers {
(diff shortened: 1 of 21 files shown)


core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
```diff
@@ -72,7 +72,7 @@ trait Authenticate extends Logging {
                 future
             }.toOption
         } getOrElse {
-            info(this, s"credentials are malformed")
+            userpass.map(_ => info(this, s"credentials are malformed"))
```
psuter: surely you meant to write `foreach`

common/scala/src/main/scala/whisk/utils/JsHelpers.scala
```diff
@@ -0,0 +1,43 @@
...
+import spray.json.JsValue
+
+object JsHelpers {
+    implicit class JsObjectHelper(js: JsObject) {
+        def getFieldPath(path: List[String]): Option[JsValue] = {
+            path match {
+                case Nil       => Option(js)
+                case p :: Nil  => js.fields.get(p)
+                case p :: tail => js.fields.get(p) match {
+                    case Some(o: JsObject) => o.getFieldPath(tail)
```
psuter: Probably not terribly efficient given that every recursive call involves creating a new instance of the `implicit` class

rabbah: https://github.com/openwhisk/openwhisk/pull/1720/commits/c1db4e5fb67ede51ded62b89910c6de12f2b73d8

core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
rabbah: thanks for the nudge.

rabbah:
rebased to master to resolve conflicts.

psuter:
Merged as 121d63149cc6d6f7c9cf4f85e5150cde537ab413...ad9b444268068a9d60fcaf109fd00ac8cee8dfc5

Issue #1569, opened 2016-11-30
CLI should sort packages/actions/triggers/rules alphabetically


l2fprod:
Here is the output of `wsk list` in my namespace:

```
Entities in namespace: f@fr.ibm.com_scratch
packages
/f@fr.ibm.com_scratch/logistics-wizard-recommendation   private
/f@fr.ibm.com_scratch/dev-logistics-wizard-recommendation private
/f@fr.ibm.com_scratch/Post to Slack                     private
/f@fr.ibm.com_scratch/lwr                               private
actions
/f@fr.ibm.com_scratch/dev-logistics-wizard-recommendation/prepare-for-slack private
/f@fr.ibm.com_scratch/dev-logistics-wizard-recommendation/observations private
/f@fr.ibm.com_scratch/dev-logistics-wizard-recommendation/acknowledge private
/f@fr.ibm.com_scratch/dev-logistics-wizard-recommendation/retrieve private
/f@fr.ibm.com_scratch/logistics-wizard-recommendation/acknowledge private
/f@fr.ibm.com_scratch/logistics-wizard-recommendation/retrieve private
/f@fr.ibm.com_scratch/dev-logistics-wizard-recommendation/recommend private
/f@fr.ibm.com_scratch/logistics-wizard-recommendation/recommend private
/f@fr.ibm.com_scratch/lwr/prepare-for-slack             private
/f@fr.ibm.com_scratch/lwr/observations                  private
/f@fr.ibm.com_scratch/lwr/acknowledge                   private
/f@fr.ibm.com_scratch/lwr/retrieve                      private
/f@fr.ibm.com_scratch/lwr/recommend                     private
/f@fr.ibm.com_scratch/Send recommendations to Slack     private
triggers
/f@fr.ibm.com_scratch/recommendations-trigger           private
rules
/f@fr.ibm.com_scratch/Trigger Slack recommendation message private
```

using `whisk CLI version	2016-11-17T09:38:33+00:00`

would be great to sort the entries by name.

dubeejw:
As part of the work done on this task (https://github.com/apache/incubator-openwhisk/pull/2326) we've determined that the CLI should not be responsible for sorting entity names. Instead we should try to retrieve the entity names from the database using a view that sorts for us. The type of sort should be controllable via a flag passed to the REST interface. 

[...]

PR #1729, opened 2017-01-17
Sort packages, actions, triggers, rules by name alphabetically.

lzbj:
Address issue #1569.

### DIFF
--- a/tools/cli/go-whisk-cli/commands/namespace.go
+++ b/tools/cli/go-whisk-cli/commands/namespace.go
@@ -119,8 +119,10 @@ var listCmd = &cobra.Command{
 }
 
 func init() {
+    listCmd.Flags().BoolVarP(&flags.namespace.date, "sort", "t", false, wski18n.T("sort by date"))
     namespaceCmd.AddCommand(
         namespaceListCmd,
         namespaceGetCmd,
     )
+
 }
--- a/tools/cli/go-whisk-cli/commands/util.go
+++ b/tools/cli/go-whisk-cli/commands/util.go
@@ -285,8 +285,26 @@ func printSummary(collection interface{}) {
     }
 }
 
+func Len(collection interface{}) func() int {
+    switch collection := collection.(type) {
+    case []whisk.Action:
+    case []whisk.Trigger:
+    case []whisk.Package:
+    case []whisk.Rule:
+        return func() int{
+            return len(collection)
+        }
+    }
+    return 0
+}
+
+
 func printActionList(actions []whisk.Action) {
     fmt.Fprintf(color.Output, "%s\n", boldString("actions"))
+    if (!flags.namespace.date) {
+        sort.Sort(whisk.ActionArray(actions))
+    }
+
     for _, action := range actions {
         publishState := wski18n.T("private")
         kind := getValueString(action.Annotations, "exec")
@@ -296,6 +314,10 @@ func printActionList(actions []whisk.Action) {
 
 func printTriggerList(triggers []whisk.Trigger) {
     fmt.Fprintf(color.Output, "%s\n", boldString("triggers"))
+    if (!flags.namespace.date) {
+        sort.Sort(whisk.TriggerArray(triggers))
+    }
+
     for _, trigger := range triggers {
         publishState := wski18n.T("private")
         fmt.Printf("%-70s %s\n", fmt.Sprintf("/%s/%s", trigger.Namespace, trigger.Name), publishState)
@@ -304,6 +326,10 @@ func printTriggerList(triggers []whisk.Trigger) {
 
 func printPackageList(packages []whisk.Package) {
     fmt.Fprintf(color.Output, "%s\n", boldString("packages"))
+    if (!flags.namespace.date) {
+        sort.Sort(whisk.PackageArray(packages))
+    }
+
     for _, xPackage := range packages {
         publishState := wski18n.T("private")
         if xPackage.Publish != nil && *xPackage.Publish {
@@ -315,6 +341,10 @@ func printPackageList(packages []whisk.Package) {
 
 func printRuleList(rules []whisk.Rule) {
     fmt.Fprintf(color.Output, "%s\n", boldString("rules"))
+    if (!flags.namespace.date) {
+        sort.Sort((whisk.RuleArray(rules)))
+    }
+
     for _, rule := range rules {
         publishState := wski18n.T("private")
         fmt.Printf("%-70s %s\n", fmt.Sprintf("/%s/%s", rule.Namespace, rule.Name), publishState)
--- a/tools/cli/go-whisk/whisk/trigger.go
+++ b/tools/cli/go-whisk/whisk/trigger.go
@@ -46,6 +46,20 @@ type TriggerListOptions struct {
     Docs            bool            `url:"docs,omitempty"`
 }
 
+type TriggerArray []Trigger
+
+func (ta TriggerArray) Len() int {
+    return len(ta)
+}
+
+func (ta TriggerArray) Less(i int, j int) bool {
+    return ta[i].Name < ta[j].Name
+}
+
+func (ta TriggerArray) Swap(i int, j int) {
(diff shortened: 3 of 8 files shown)


rabbah:
I would make this a feature toggled by a flag: sort by date or alphabetelically.

csantanapr:
@rabbah 😕  date? I'm not aware that entities like rules, packages, etc.. have a date associated to it.

rabbah:
The current order is by date (last updated eg). All entities when stored in the database have a generated date property (it's hidden from the user) if the entity doesn't have an obvious property to use as date (which is the case for activations for example).

csantanapr:
ah so your saying that the REST API already support a `sort` parameter that CLI can pass to it to sort by date or alphabetically?

csantanapr:
I tried `wsk list` and `wsk rule get` with `-v` and was not able to see a date property in the return data. I was just trying to understand where is this sorting that your proposing happens client or server side.

rabbah:
No I'm saying records in the database have a date field. It is not visible to you as the end user.

csantanapr:
OK How are you proposing we implement this sorting featured flag?

It looks like it can't be done ONLY with CLI changes.

Something along the lines to add support in the backend to provide a parameter that get's used to return the sorted list, or return the list with the date information for the client to do the sorting.

rabbah:
We already have sort by date. This is the current ordering. I'm simply proposing that you choose a default (name vs date) and guard the sort in the cli by such a flag.

csantanapr:
Ah see, make sense now. I think it's being a long day for me 🤣  going home now.

@lzbj I assume you understood the propose change, very simple to handle. 

I propose flag name `sort` possible values `date` or `name` 

My OCD tells me that default when flag not specify should be `name`
But practical person says `date`

I'm OK flipping a coin, or if someone has a strong opinion

rabbah:
Thought: sort by name by default and add `-t` to sort by time (matching `ls -t` for example).
For activations however should we support sorting lexically? One thought is to sort lexically by the corresponding action/trigger/rule name and within each group sort by time. And the flag `-t` could then flatten the grouping... Seems confusing though.

lzbj:
@rabbah ,@csantanapr, I'll try to figure out and and a flag. Thanks for review.

rabbah:
@lzbj I added `wip` flag - remove it when ready and apply `review` label.

rabbah:
Looks good. A test should be added. I suggest WskBasicUsageTests class.

lzbj:
Thanks, I'll add a basic test cases for it.

dubeejw:
Needs a PG.

lzbj:
@dubeejw , could you please help run a PG? thank you.

tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -1242,4 +1242,43 @@ class WskBasicUsageTests
...
+    behavior of "Wsk cli list"
+
+    it should "wsk cli list entities by name alphabetically by default" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+
+            val tmpwskprops = File.createTempFile("wskprops", ".tmp")
+            try {
+                val namespace = wsk.namespace.list().stdout.trim.split("\n").last
+                val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
+                wsk.cli(Seq("property", "set", "-i", "--apihost", wskprops.apihost, "--auth", wskprops.authKey,
```
rabbah: using a config file seems like overkill - this is to avoid adding `sort` to the `Wsk.list` method?

tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -1242,4 +1242,43 @@ class WskBasicUsageTests
...
+
+                val name2 = "echoTest"
+                val file2 = Some(TestUtils.getTestActionFilename("echo.js"))
+
+                assetHelper.withCleaner(wsk.action, name2) {
+                    (action, _) =>
+                        action.create(name2, file2, annotations = getValidJSONTestArgInput,
+                            parameters = getValidJSONTestArgInput)
+                }
+                val stdout = wsk.cli(Seq("list", "-i", "--apihost", wskprops.apihost, "--auth", wskprops.authKey), env = env).stdout
```
rabbah: I think a better approach is to change `list` to accept `sort`.

tools/cli/go-whisk-cli/commands/namespace.go
```diff
@@ -119,8 +119,10 @@ var listCmd = &cobra.Command{
 }
 
 func init() {
+    listCmd.Flags().BoolVarP(&flags.namespace.date, "sort", "t", false, wski18n.T("sort by date"))
```
dubeejw: Looks like this flag will work for `wsk list`, but not for individual list commands like `wsk action list`.

tools/cli/go-whisk/whisk/trigger.go
```diff
@@ -46,6 +46,20 @@ type TriggerListOptions struct {
     Docs            bool            `url:"docs,omitempty"`
 }
 
+type TriggerArray []Trigger
```
dubeejw: I think you might be able to create a functions that expect an generic `interface {}` type. Then cast to the appropriate type for `Name` comparison. This way there isn't specific sort functions need for for triggers, actions, rules, and packages. See https://golang.org/doc/effective_go.html#type_switch.

csantanapr:
@rabbah assigning to you since you are already reviewing this PR

lzbj:
@rabbah @dubeejw @csantanapr ,thank you, I'll try to update accordingly.

tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
lzbj: Correct, I'll remove it.

tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
lzbj: ok, add it.

tools/cli/go-whisk-cli/commands/namespace.go
lzbj: Yes, just to address issue #1569, ok, I'll try to change a better one for your review.

tools/cli/go-whisk/whisk/trigger.go
lzbj: Ok, I'll try this approach, my initial though was to let entities like action, trigger implements some interface which is convenient for container manipulates etc, but anyway, if you think it's not necessary, let's change it.

cfjedimaster:
As just an FYI - and sorry if I missed this in the thread - but I'd argue for sorting by alpha to be the *default*. I can see supporting date sorting too as an option, but I'd imagine it being more normal to have an alpha sort.

rabbah:
@lzbj are you still working on this? It would be nice to get this in.

lzbj:
@rabbah , not recently, I'll work on this when available.

rabbah:
@RSulzmann can you take this over? it's fairly close and I think @lzbj isn't actively working on this. David, can you provide an update if you are still working on this?

lzbj:
@rabbah , yes, I'll provide an update.

tools/cli/go-whisk-cli/commands/util.go
```diff
@@ -285,8 +285,26 @@ func printSummary(collection interface{}) {
     }
 }
 
+func Len(collection interface{}) func() int {
```
lzbj: @dubeejw , is your comments means this kind of function? I thought perhaps the action array should implement the sort interfaces, right?

lzbj:
@rabbah , I just added a simple function to discuss with dubeejw about his comments, anyhow, feel free to reassign other people if this high priority.

rabbah:
Closing in favor of https://github.com/apache/incubator-openwhisk/pull/2326.

PR #1731, opened 2017-01-17
Run invoker as privileged container

perryibm:
The invoker container needs access to privileged operations that are not available through the docker daemon.  In particular, access to the underlying runc binary is needed to improve performance.

### DIFF
--- a/ansible/roles/invoker/tasks/deploy.yml
+++ b/ansible/roles/invoker/tasks/deploy.yml
@@ -32,7 +32,7 @@
         --userns=host
         --privileged
         -v /sys/fs/cgroup:/sys/fs/cgroup
-        -v /usr/bin/runc:/usr/bin/runc
+        -v /usr/bin/docker-runc:/usr/bin/docker-runc
         -v /run/runc:/run/runc
         --name invoker{{play_hosts.index(inventory_hostname)}}
         --hostname invoker{{play_hosts.index(inventory_hostname)}}
--- a/ansible/roles/invoker/tasks/deploy.yml
+++ b/ansible/roles/invoker/tasks/deploy.yml
@@ -32,8 +32,9 @@
         --userns=host
         --privileged
         -v /sys/fs/cgroup:/sys/fs/cgroup
-        -v /usr/bin/docker-runc:/usr/bin/docker-runc
         -v /run/runc:/run/runc
+        -v /usr/bin/docker-runc:/usr/bin/docker-runc
+        -v /usr/lib/x86_64-linux-gnu/libapparmor.so.1:/usr/lib/x86_64-linux-gnu/libapparmor.so.1
         --name invoker{{play_hosts.index(inventory_hostname)}}
         --hostname invoker{{play_hosts.index(inventory_hostname)}}
         --restart {{ docker.restart.policy }}
@@ -51,6 +52,7 @@
         --userns host
         {{docker_registry}}{{ docker_image_prefix }}/invoker:{{docker_image_tag}}
         /bin/sh -c "/invoker/bin/invoker {{play_hosts.index(inventory_hostname)}} >> /logs/invoker{{play_hosts.index(inventory_hostname)}}_logs.log 2>&1"
+  register: invokerOutput
 
 # todo: re-enable docker_container module once https://github.com/ansible/ansible-modules-core/issues/5054 is resolved
 
--- a/ansible/roles/invoker/tasks/deploy.yml
+++ b/ansible/roles/invoker/tasks/deploy.yml
@@ -28,6 +28,12 @@
 - name: start invoker using docker cli
   shell: >
         docker run -d
+        --pid=host
+        --userns=host
+        --privileged
+        -v /sys/fs/cgroup:/sys/fs/cgroup
+        -v /usr/bin/runc:/usr/bin/runc
+        -v /run/runc:/run/runc
         --name invoker{{play_hosts.index(inventory_hostname)}}
         --hostname invoker{{play_hosts.index(inventory_hostname)}}
         --restart {{ docker.restart.policy }}
--- a/ansible/roles/invoker/tasks/deploy.yml
+++ b/ansible/roles/invoker/tasks/deploy.yml
@@ -25,16 +25,18 @@
     mode: 0777
   become: true
 
+- name: define options when deploying invoker on Ubuntu
+  set_fact:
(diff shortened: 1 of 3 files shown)


perryibm:
1st commit passed with PG3-169

perryibm:
PG2-948  Commits should be squashed.

ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -45,6 +52,7 @@
         --userns host
         {{docker_registry}}{{ docker_image_prefix }}/invoker:{{docker_image_tag}}
         /bin/sh -c "/invoker/bin/invoker {{play_hosts.index(inventory_hostname)}} >> /logs/invoker{{play_hosts.index(inventory_hostname)}}_logs.log 2>&1"
+  register: invokerOutput
```
markusthoemmes: Why's that line needed?

ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -28,6 +28,13 @@
 - name: start invoker using docker cli
   shell: >
         docker run -d
+        --pid=host
+        --userns=host
+        --privileged
+        -v /sys/fs/cgroup:/sys/fs/cgroup
+        -v /run/runc:/run/runc
+        -v /usr/bin/docker-runc:/usr/bin/docker-runc
+        -v /usr/lib/x86_64-linux-gnu/libapparmor.so.1:/usr/lib/x86_64-linux-gnu/libapparmor.so.1
```
markusthoemmes: Will this work on *any* system, even if apparmor is not installed (like docker-machine)?

ansible/roles/invoker/tasks/deploy.yml
perryibm: I can remove.  It was part of systematic debugging.

ansible/roles/invoker/tasks/deploy.yml
domdom82: most likely not. apparmor is ubuntu-specific and not a given on all *nix systems.
this is also part of the reason why apparmor is not yet merged. my suggestion would be to provide something like a "feature discovery", maybe as a singleton like whiskconfig. it should be able to discover available docker features based on the `docker info` command (or corresponding rest api call)

core/invoker/src/main/scala/whisk/core/container/RuncUtils.scala
```diff
@@ -0,0 +1,68 @@
...
+object RuncUtils extends Logging {
+
+    private implicit val emitter: PrintStreamEmitter = this
+
+    def list()(implicit transid: TransactionId): String = {
+        runRuncCmd(false, Seq("list"))
+    }
+
+    /**
+     * Synchronously runs the given docker command returning stdout if successful.
```
domdom82: "runs the given runc command..."

ansible/roles/invoker/tasks/deploy.yml
perryibm: The current idea is to conditionalize the behavior.  Not checking the mounting is ok though not clean because the invoker can internally check.   So to do this cleanly, do we have a ansible flag for mac vs linux-ness?

core/invoker/src/main/scala/whisk/core/container/RuncUtils.scala
perryibm: fixed

ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -25,9 +25,16 @@
     mode: 0777
   become: true
 
+- name: define options when deploying invoker on Linux
+  set_fact:
+    linuxOptions: "--pid=host --privileged -v /sys/fs/cgroup:/sys/fs/cgroup -v /run/runc:/run/runc -v /usr/bin/docker-runc:/usr/bin/docker-runc -v /usr/lib/x86_64-linux-gnu/libapparmor.so.1:/usr/lib/x86_64-linux-gnu/libapparmor.so.1"
+  when: whisk_version_name != "mac"
```
markusthoemmes: I think this approach is not sufficient. `mac` will be true for `docker-machine` environments, which we currently support. `run-c` is available there, although in `/usr/bin/local`.

perryibm:
PG2-959

ansible/roles/invoker/tasks/deploy.yml
domdom82: is it possible to use runc w/o apparmor? then you could only put that into the linux_options

rabbah:
I had to modify the deploy playbook for mac as follows:

```
diff --git a/ansible/roles/invoker/tasks/deploy.yml b/ansible/roles/invoker/tasks/deploy.yml
index 9bd10e4..63d585d 100644
--- a/ansible/roles/invoker/tasks/deploy.yml
+++ b/ansible/roles/invoker/tasks/deploy.yml
@@ -30,15 +30,11 @@
     linuxOptions: "--pid=host --privileged -v /sys/fs/cgroup:/sys/fs/cgroup -v /run/runc:/run/runc -v /usr/bin/docker-runc:/usr/bin/docker-runc -v /usr/lib/x86_64-linux-gnu/libapparmor.so.1:/usr/lib/x86_64-linux-gnu/libapparmor.so.1"
   when: whisk_version_name != "mac"
 
-- set_fact:
-    linuxOptions: ""
-  when: not linuxOptions
-
 - name: start invoker using docker cli
   shell: >
         docker run -d
         --userns=host
-        {{ linuxOptions }}
+        {{ linuxOptions | default('') }}
         --name invoker{{play_hosts.index(inventory_hostname)}}
         --hostname invoker{{play_hosts.index(inventory_hostname)}}
         --restart {{ docker.restart.policy }}
```

ansible/roles/invoker/tasks/deploy.yml
rabbah: where does `whisk_version_name` get defined?

perryibm:
Thanks for the work and the fix looks nicer and I've adjusted the commit as such.   However, do you know why the use of  `not` failed?  It does not seem like an uncommon pattern.  Re-testing.

ansible/roles/invoker/tasks/deploy.yml
perryibm: AFAIK, not possible as runc assumes the presence of libapparmor.

rabbah:
Squashed and merged as https://github.com/openwhisk/openwhisk/commit/41a20fd67a6e9ba7ce814357b23e117539894d37.

PR #1740, opened 2017-01-19
Add namespace CRUD operations for wskadmin.

### LINKED ISSUE #1132: Update wskadmin with namespace related commands
`wskadmin` has user creation and deletion commands. With #1107 in place, we should also support namespace creation and deletion commands.

### LINKED ISSUE #1748: break up initdb to separate view updates from key initialization
the ansible playbook for initdb recreates the subject table (drop/create), inserts predefined keys (guest, system), and installs the views.

arguably there are separate tasks that can be done independent of each other, but at the very least updating the view should be doable on its own,

rabbah:
Fixes #1132.
Fixes #1748.
PG1/1084, PG2/913, PG3/185 all approved.
PG1/1085, PG2/915, PG3/186 re-approved.

### DIFF
--- a/ansible/tasks/initdb.yml
+++ b/ansible/tasks/initdb.yml
@@ -10,7 +10,7 @@
     user: "{{ db_username }}"
     password: "{{ db_password }}"
     force_basic_auth: yes
-  register: result
+  register: dbexists
 
 # create only the missing db_auth
 - name: create immortal {{ db_auth }} db with {{ db_provider }}
@@ -21,21 +21,51 @@
     user: "{{ db_username }}"
     password: "{{ db_password }}"
     force_basic_auth: yes
-  when: result.status == 404
+  when: dbexists is defined and dbexists.status == 404
 
-- name: recreate the "full" index on the "auth" database
+# fetches the revision of previous view (to update it) if it exists
+- name: check for previous view in "auth" database
+  vars:
+    auth_index: "{{ lookup('file', '{{ openwhisk_home }}/ansible/files/auth_index.json') }}"
+  uri:
+    url: "{{ db_protocol }}://{{ db_host }}:{{ db_port }}/{{ db_auth }}/{{ auth_index['_id'] }}"
+    return_content: yes
+    method: GET
+    status_code: 200, 404
+    user: "{{ db_username }}"
+    password: "{{ db_password }}"
+    force_basic_auth: yes
+  register: previousView
+  when: dbexists is defined and dbexists.status != 404 #and mode=="updateview"
+
+- name: extract revision from previous view
+  vars:
+    previousContent: "{{ previousView['content']|from_json }}"
+    revision: "{{ previousContent['_rev'] }}"
+    auth_index: "{{ lookup('file', '{{ openwhisk_home }}/ansible/files/auth_index.json') }}"
+  set_fact:
+    previousContent: "{{ previousContent }}"
+    updateWithRevision: "{{ auth_index | combine({'_rev': revision}) }}"
+  when: previousView is defined and previousView.status != 404
+
+- name: check if a view update is required
+  set_fact:
+    updateView: "{{ updateWithRevision }}"
+  when: previousContent is defined and previousContent != updateWithRevision
+
+- name: recreate or update the index on the "auth" database
   vars:
     auth_index: "{{ lookup('file', '{{ openwhisk_home }}/ansible/files/auth_index.json') }}"
   uri:
     url: "{{ db_protocol }}://{{ db_host }}:{{ db_port }}/{{ db_auth }}"
     method: POST
-    status_code: 200,201
+    status_code: 200, 201
     body_format: json
-    body: "{{ auth_index }}"
+    body: "{{ updateView | default(auth_index) }}"
     user: "{{ db_username }}"
     password: "{{ db_password }}"
     force_basic_auth: yes
-  when: result.status == 404
+  when: (dbexists is defined and dbexists.status == 404) or (updateView is defined)
 
 - name: recreate necessary "auth" keys
   vars:
@@ -61,4 +91,4 @@
     password: "{{ db_password }}"
     force_basic_auth: yes
   with_items: "{{ db.authkeys }}"
-  when: result.status == 404
+  when: dbexists is defined and dbexists.status == 404
--- a/tests/src/whisk/core/admin/WskAdminTests.scala
+++ b/tests/src/whisk/core/admin/WskAdminTests.scala
@@ -16,6 +16,8 @@
 
 package whisk.core.admin
 
+import scala.concurrent.duration.DurationInt
+
 import org.junit.runner.RunWith
 import org.scalatest.Matchers
 import org.scalatest.junit.JUnitRunner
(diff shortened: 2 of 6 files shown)


tests/src/whisk/core/admin/WskAdminTests.scala
```diff
@@ -42,23 +42,37 @@ class WskAdminTests
...
+            wskadmin.cli(Seq("user", "delete", subject)).stdout should include("Subject deleted")
 
-        wskadmin.cli(Seq("user", "whois", authkey)).stdout.trim should be(Seq(s"subject: $subject", s"namespace: $subject").mkString("\n"))
-        wskadmin.cli(Seq("user", "delete", subject)).stdout should include("Subject deleted")
-
-        val recreate = wskadmin.cli(Seq("user", "create", subject, "-u", auth.authkey.compact))
-        wskadmin.cli(Seq("user", "get", subject)).stdout.trim should be(auth.authkey.compact)
-        wskadmin.cli(Seq("user", "delete", subject)).stdout should include("Subject deleted")
+            // recreate with explicit
+            val newspace = s"${subject}.myspace"
```
psuter: that gave me a flashback to 2005

ansible/tasks/initdb.yml
```diff
@@ -22,7 +22,30 @@
...
+  uri:
+    url: "{{ db_protocol }}://{{ db_host }}:{{ db_port }}/{{ db_auth }}/{{ auth_index['_id'] }}"
+    return_content: yes
+    method: GET
+    status_code: 200, 404
+    user: "{{ db_username }}"
+    password: "{{ db_password }}"
+    force_basic_auth: yes
+  register: previousView
+  when: dbexists is defined and dbexists.status != 404 # and mode=="updateview"
```
rabbah: i opted for default behavior is always update view _hence_ the ignored `mode` predicate.

rabbah:
@domdom82 please lend your expertise to the last three commits.

ansible/tasks/initdb.yml
```diff
@@ -22,7 +22,30 @@
...
+
+- name: check for changes in view
+  vars:
+    revision: "{{ (previousView['content']|from_json)['_rev'] }}"
+    auth_index: "{{ lookup('file', '{{ openwhisk_home }}/ansible/files/auth_index.json') }}"
+  set_fact:
+    updateWithRevision: "{{ auth_index | combine({'_rev': revision}) }}"
+  when: previousView is defined and previousView.status != 404
```
markusthoemmes: This doesn't check for changes but always updates the view, or am I reading it wrong?

rabbah: That's right.

ansible/tasks/initdb.yml
markusthoemmes: I'm unsure **always** updating is a good idea. For a small database that might be viable, for a large database though a recomputation of the view can take hours.

rabbah: Are you concerned about accidentally running this against a large database in a production environment? My thought was that you would not run `initdb` again in production, other than to update the view.

There are two points:
1. always updating when these tasks run (i.e., without needing an `updateview` mode for example)
2. not updating if the new design doc is the same as the previous design doc modulo the revision.

You might be thinking of doing 2 as well? I'm good with that.

rabbah: See https://github.com/openwhisk/openwhisk/pull/1740/commits/8affb628b782760d13f4976ff8c51a325e2bcfd3.

ansible/tasks/initdb.yml
```diff
@@ -21,21 +22,51 @@
...
+  register: previousView
+  when: dbexists is defined and dbexists.status != 404 #and mode=="updateview"
+
+- name: extract revision from previous view
+  vars:
+    previousContent: "{{ previousView['content']|from_json }}"
+    revision: "{{ previousContent['_rev'] }}"
+    auth_index: "{{ lookup('file', '{{ openwhisk_home }}/ansible/files/auth_index.json') }}"
+  set_fact:
+    previousContent: "{{ previousContent }}"
```
domdom82: is that really necessary? shouldn't `previousContent` be set already?

ansible/tasks/initdb.yml
```diff
@@ -21,21 +22,51 @@
...
+    updateWithRevision: "{{ auth_index | combine({'_rev': revision}) }}"
+  when: previousView is defined and previousView.status != 404
+
+- name: check if a view update is required
+  set_fact:
+    updateView: "{{ updateWithRevision }}"
+  when: previousContent is defined and previousContent != updateWithRevision
 
 - name: recreate the "full" index on the "auth" database
```
domdom82: is this description still correct with the new changes?

ansible/tasks/initdb.yml
```diff
@@ -61,4 +92,17 @@
     password: "{{ db_password }}"
     force_basic_auth: yes
   with_items: "{{ db.authkeys }}"
-  when: result.status == 404
+  when: dbexists is defined and dbexists.status == 404
+
+- name: secure db against anonymous access
```
domdom82: hmm is there a way to put this in the couchdb role? this feels like it is in the wrong place here.

rabbah: This is additive. The role is currently creating the db unsecured. I wanted the option to toggle secure vs insecure mode. That strikes me more as a task than a deployment role. Additionally, to put this in the role and toggle it we would need a different flag - since mode is used for up/down of the container.

ansible/tasks/initdb.yml
rabbah: I'll change to recreate or update. Thanks.

ansible/tasks/initdb.yml
rabbah: do vars leak across tasks?

ansible/tasks/initdb.yml
domdom82: but you are now conflating the initialization of the db with the configuration of the db provider. 
the flag you are setting applies only to couchdb, right? not subject db. also is there a reason to not use "secure" mode? even if you want this to be an additional task, it should be logically grouped within the couchdb role. after all, it does tweak the couch server.

rabbah: enabling secure mode by default is fine. In that case putting it in the deployment role is ok. But then why would I run a deployment playbook to change a configuration and toggle it on off when I'm not bringing down or tearing down the container? Isn't that more appropriate for a task?

rabbah: I'll need further guidance on how to do this as part of the couchdb role in a way that allows for separate toggle on/off - since this is orthogonal to the main change in this pr, I will remove it and deal with it separately.

ansible/tasks/initdb.yml
markusthoemmes: Yes, I think this should not be necessary.

ansible/tasks/initdb.yml
```diff
@@ -21,21 +21,51 @@
     user: "{{ db_username }}"
     password: "{{ db_password }}"
     force_basic_auth: yes
-  when: result.status == 404
+  when: dbexists is defined and dbexists.status == 404
```
markusthoemmes: Are those additional guards necessary? Encountered problems with the old versions of just asking for the status code?

tools/admin/wskadmin
```diff
@@ -252,43 +288,94 @@ def getSubjectFromDb(args, props):
...
+    password = props[DB_PASSWORD]
+    database = props[DB_WHISK_AUTHS]
+
+    if args.subject.strip() == '':
+        print 'Subject must not be empty'
+        return 2
+
+    if args.namespace and args.namespace.strip() == '':
+        print 'Namespace must not be empty'
+        return 2
```
markusthoemmes: FWIW: Those sanity checks could go to the argument parser directly, making for a more concise reading experience.

For example:

```
def non_empty_string(str):
    if str.strip() == '': raise argparse.ArgumentTypeError("The provided option must not be empty.")
    return str

subcmd.add_argument('subject', help='the subject to get key for', type=non_empty_string)
```

Not blocking the PR for it, but a pattern I generally liked.

ansible/tasks/initdb.yml
rabbah: Defensive coding doesn't hurt?

tools/vagrant/README.md
```diff
@@ -170,14 +170,22 @@ wskadmin user create <subject>
 
 This command will create a new *subject* with the authorization key shown on the console once you run `wskadmin`. This key is required when making API calls to OpenWhisk, or when using the command line interface (CLI). The namespace is the same as the `<subject>` name used to create the key.
 
-The same tool may be used to delete a subject.
+A namespace allows two or more subjects to share resources. Each subject will have their own authorization key to work with resources in a namespace, but
+will have equal rights to the namespace.
```
markusthoemmes: Weird newline

PR #1742, opened 2017-01-19
Allow blocking requests to specify a desired timeout.

rabbah:
For blocking requests, rather than impose a system-defined timeout, let the API caller decide (up to the current limit). This allows for example API gateway to specify a lower value if desirable.
PG1/1082 approved.


rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/3b2f7bc6005d99b88bc24b88a149e97ce51f613b.

PR #1747, opened 2017-01-20
Remove throwable's message from HTTP response on 500 codes.

rabbah:
PG1/1081 approved.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
+++ b/common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
@@ -25,15 +25,14 @@ import akka.event.Logging.ErrorLevel
 import akka.http.scaladsl.model._
 import akka.stream.scaladsl._
 import akka.util.ByteString
-
 import spray.json._
-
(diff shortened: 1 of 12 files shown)


common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -169,6 +168,8 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](
                     transid.finished(this, start, s"[GET] '$dbName' failed to get document: '${doc}'; http status: '${code}'")
                     throw new Exception("Unexpected http response code: " + code)
             }
+        } recoverWith {
```
psuter: That's for _all_ other errors? Didn't want to match on the applicable exceptions and throw a generic "unknown internal server error" otherwise?

rabbah: don't we throw this exception from all the json serializers for the document types?
that's why i went for the top level error.

PR #1753, opened 2017-01-24
Allow delete/update of rule even if in active state.

### LINKED ISSUE #1561: Support deleting "active" rules
Rules must be disabled before being deleted. 
If I write tools or scripts to manage my openwhisk projects, using a client library, I end up having to always write an extra call to disable every rule I want to remove. 
Would it be possible to modify the API to relax this constraint?

rabbah:
Fixes #1561.


rabbah:
PG2/953 approved.

sjfink:
rebased, merged in 53395f4..c5da36f

PR #1756, opened 2017-01-24
Utf 8 db

rabbah:
Going to use these tests to avoid a similar slip in the future.
I will add the patch on top.


rabbah:
Picking it up in #1757.

PR #1757, opened 2017-01-24
Fix encoding of payloads to respect UTF8 for actions

### LINKED ISSUE #252: properly handle unicode characters
The decoding is localized to reading records from the database (on the client side) which is normalizing to the HTTP default charset ISO-8859-1 instead of UTF-8. 

As a sanity check, I confirmed that for a sample code:

```
function main() {
  var frenchMessage = "Le périphérique est connecté";
  var chineseMesage = "装置被连接";
  return ({"fr": frenchMessage, "zh-Hans": chineseMesage});
}
```

The record is stored correctly in the database. If I were to encode the string using base64 encoding and logging the result, the activation result and logs are correctly stored in the database:

```
function main() {
  var msg = 'eyJmciI6IkxlIHDDqXJpcGjDqXJpcXVlIGVzdCBjb25uZWN0w6kiLCJ6aC1IYW5zIjoi6KOF572u6KKr6L+e5o6lIn0=';
  var dec = new Buffer(msg, 'base64').toString('utf-8');
  console.log(dec);
  var res = JSON.parse(dec);
  return res;
}
```

results in this activation record:

```
"response": {
    "statusCode": 0,
    "result": {
      "fr": "Le périphérique est connecté",
      "zh-Hans": "装置被连接"
    }
  },
  "logs": [
    "2016-04-23T13:36:16.030199912Z stdout: {\"fr\":\"Le périphérique est connecté\",\"zh-Hans\":\"装置被连接\"}"
  ],
```

@psuter FYI this is true both using the Cloudant SDK as well as the Spray client.

### LINKED ISSUE #362: Swift action parsing doesn't tolerate special characters
For example, if I try to run an action with the following code:

``` swift
//  Copyright © 2016 Dirk Beanberger. All rights reserved.
  return ["message": "This should just work..."]
}
```

I always get an error of `the action did not produce a valid JSON response`

However, if I simply remove the `©` everything works as expected.

rabbah:
Fixes invoker-side defect affecting actions and enables proper language support for node.js and java, python and swift actions.

Fixes #252.
Fixes #362.


tests/src/whisk/core/entity/test/DatastoreTests.scala
```diff
@@ -141,7 +141,7 @@ class DatastoreTests extends FlatSpec
         }
     }
 
-    ignore should "CRD activation with utf8 characters" in {
+    it should "CRD activation with utf8 characters" in {
```
csantanapr: This is unrelated to the fix applied in the invoker http call right?
But if it works now let's enable the test I agree.

rabbah:
Fixed Java actions as well in latest commit.

rabbah:
This pr has passed pg2/952 and pg3/200 but is failing in Travis for the python and swift tests. I will suppress those two tests and investigate them separately.

sjfink:
rebased, merged as 61a8ed2..53395f4

Issue #1749, opened 2017-01-21
post deploy to install catalog ignores/overwrites local checkout


rabbah:
if one has a fork of openwhisk-catalog locally and runs this playbook
https://github.com/openwhisk/openwhisk/blob/master/ansible/tasks/installOpenwhiskCatalog.yml

it modifies one's git remote origin pointer from one's fork to point to the upstream repo; it's good at least that it uses https vs git or i would have accidentally pushed to master.

```
> git remote -v
origin	git@github.com:rabbah/openwhisk-catalog.git (fetch)
origin	git@github.com:rabbah/openwhisk-catalog.git (push)
```

and after the playbook:
```
> git remote -v
origin	https://github.com/openwhisk/openwhisk-catalog.git (fetch)
origin	https://github.com/openwhisk/openwhisk-catalog.git (push)
```

PR #1758, opened 2017-01-24
Do not retrieve new git revisions by default for openwhisk-catalog

jasonpet:
Resolves issue #1749


rabbah:
Confirmed to work.

rabbah:
The invariant for jenkins jobs is that it always starts with a clean workspace - otherwise, one needs to override the default.

PR #1762, opened 2017-01-25
Update trigger default and limits in documentation for Cloudant and Alarms

jasonpet:
Documentation still says 1000 is the default and 10,000 is the max.


csantanapr:
merged 4db361d9554abdc97b08309431bb3f0acca999ef

PR #1763, opened 2017-01-25
Bump akka and spray dependencies

markusthoemmes:
Getting the latest bugfixes is goodness 👍

### DIFF
--- a/common/scala/build.gradle
+++ b/common/scala/build.gradle
@@ -11,18 +11,18 @@ repositories {
 dependencies {
     compile "org.scala-lang:scala-library:${gradle.scala.version}"
 
-    compile 'io.spray:spray-caching_2.11:1.3.3'
-    compile 'io.spray:spray-json_2.11:1.3.2'
-    compile 'io.spray:spray-can_2.11:1.3.3'
-    compile 'io.spray:spray-client_2.11:1.3.1'
-    compile 'io.spray:spray-httpx_2.11:1.3.3'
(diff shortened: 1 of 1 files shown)


common/scala/build.gradle
```diff
@@ -11,18 +11,18 @@ repositories {
...
+    compile 'io.spray:spray-routing_2.11:1.3.4'
 
-    compile 'com.typesafe.akka:akka-actor_2.11:2.4.10'
-    compile 'com.typesafe.akka:akka-slf4j_2.11:2.4.10'
-    compile 'com.typesafe.akka:akka-http-core_2.11:2.4.10'
-    compile 'com.typesafe.akka:akka-http-spray-json-experimental_2.11:2.4.10'
+    compile 'com.typesafe.akka:akka-actor_2.11:2.4.16'
+    compile 'com.typesafe.akka:akka-slf4j_2.11:2.4.16'
+    compile 'com.typesafe.akka:akka-http-core_2.11:10.0.2'
+    compile 'com.typesafe.akka:akka-http-spray-json_2.11:10.0.2'
```
markusthoemmes: Note this is not as much a bump as you'd expect:

> Please note that while the version number changed significantly, the actual surface APIs did not by much. The akka-http-core module was already stable back in Akka 2.4, so it remained binary compatible, and the sources for akka-http are almost completely compatible with those that were shipped as Akka 2.4.11 – you should not have any trouble upgrading.

PR #1771, opened 2017-01-27
Support formdata in web actions.

rabbah:
Return a transaction id with failed web activations and meta activations.
Support formdata.

### DIFF
--- a/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
+++ b/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
@@ -90,6 +90,7 @@ object Messages {
     def invalidMedia(m: MediaType) = s"Response is not valid '${m.value}'."
     val contentTypeNotSupported = "Content type not supported."
     val contentTypeRequired = "Content type is missing."
+    val responseNotReady = "Response not yet ready."
 
     def invalidInitResponse(actualResponse: String) = {
         "The action failed during initialization" + {
--- a/core/controller/src/main/scala/whisk/core/controller/Meta.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Meta.scala
@@ -135,7 +135,7 @@ protected[core] object WhiskMetaApi extends Directives {
                 } getOrElse (OK)
 
                 respondWithHeaders(headers) {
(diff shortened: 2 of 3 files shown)


mdeuser:
@rabbah - Can you provide a sample output containing the tid.

rabbah:
See the tests.

mdeuser:
@rabbah - yep, took a look at the tests.. if you have a concrete example of the output that would help confirm my understanding

csantanapr:
What about in addition passing the tid in a header and also passing the tid on success ?

This way the http client doesn't matter success or error gets back a tid for meta invocation and potentially any type of invocation i.e. Regular invocation thru apigateway then there is potential to correlated logs from OpenWhisk and apigateway Server

rabbah:
Attaching the `tid` on a failed activation is consistent with the current error model: the response is an error, here's a ticket number should you want/need to investigate. The API defines the contract for error responses and so we can define that it shall contains an `error` field and a `code`.

When you're doing a blocking invoke, the response of the API is just the result of the action, on success. So where would you put the `tid`? In a custom header? Further, the `tid` is really to investigate system issues. Why should the system surface its own code when running user actions?

The meta actions are not special. They should not be treated as such. The action can log its activation ID and return a meaningful response to the user/caller (which could be its own activation id for example).

rabbah:
Proposed patch for supporting binary data: https://github.com/rabbah/openwhisk/commit/9b275c9b254d76c5f1ac5243cbf516b194c6d9bc

csantanapr:
Yeah I was talking returning via custom header. 

I think my comment is outside he scope of this this PR.

My comment was more the use case  it meta API but a call hitting the apigateway the. Doing an invoke, since to the outside world apigateway and core look one single system, was trying to see if there is a way to surface analytics on tracing the API call on going thru apigateway and the. Going thru different components of core.  But maybe a use case that doesn't hold water.  

Yes I agree that meta API is not special and having one way on tracking is proper approach for example apigw mgmt tasks/calls.

rabbah:
PG 1/1112 approved.

rabbah:
Closing in favor of #1782.

Issue #1777, opened 2017-01-28
Add URL Encoding to Whisk Object in Swift 3


paulcastro:
Add URL encoding to the uri path of the Whisk object in the Swift 3 container to avoid possible errors.

Issue #1752, opened 2017-01-23
Remove log statement that prints communication details in Swift 3 Whisk object.


paulcastro:
Remove errant log statement that prints environment variables in log statement when Whisk object is used.

PR #1778, opened 2017-01-28
Add url encoding to Whisk Object in Swift 3 container

paulcastro:
This fixes Issues #1777, #1752. 

- add url encoding to the uri path of the Whisk object's REST calls to avoid issues
- remove errant log statement in Whisk object that printed out communication details


paulcastro:
PG Blue 970

paulcastro:
Yes, it will encode the whole uri path.

PR #1779, opened 2017-01-28
Reject activation if entity size exceeds limit

### LINKED ISSUE #1059: Short-circuit invokes/fires with a payload being too big
#1044 and the work on #1046 surfaced that the controller is currently not checking the payload size for action invokes or trigger fires. In fact Kafka doesn't take messages bigger than 1MB.

When trying to invoke an action with a bigger payload than 1MB the Kafka Producer throws a `RecordTooLargeException` which is propagated to the user. #1046 fixes the propagation and instead returns the proper HTTP error for it.

We should still implement a short-circuit on the API itself, checking the payload size and returning before even trying to put the message into Kafka.

rabbah:
Action and trigger activations are subject to limits on the content payload. This pull request enforces the limit earlier in the pipeline.

PG2//971 approved.
Fixes #1059.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/ActivationEntityLimit.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/ActivationEntityLimit.scala
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.entity
+
+import whisk.core.entity.size.SizeInt
+
+/**
+ * ActivationEntityLimit defines the limits on the input/output payloads for actions
+ * and triggers. This refers to the invoke-time parameters for actions or the trigger-time
+ * parameters for triggers.
+ */
+protected[core] object ActivationEntityLimit {
+    protected[core] val MAX_ACTIVATION_ENTITY_LIMIT = 1.MB
+}
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskEntity.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskEntity.scala
@@ -20,12 +20,10 @@ import java.time.Clock
 import java.time.Instant
 
 import scala.Stream
-import scala.language.postfixOps
 import scala.util.Try
 
 import spray.json._
 import whisk.core.database.DocumentUnreadable
-import whisk.core.entity.size.SizeInt
 import whisk.http.Messages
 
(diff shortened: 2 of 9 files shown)


common/scala/src/main/scala/whisk/core/entity/ActivationEntityLimit.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import whisk.core.entity.size.SizeInt
+
+/**
+ * ActivationEntityLimit defines the limits on the input/output payloads for actions
+ * and triggers.
+ */
+
+protected[core] object ActivationEntityLimit {
+    protected[core] val MAX_ACTIVATION_ENTITY_LIMIT = 1.MB
```
markusthoemmes: Shall we add a comment to clarify that this also correlates with parameters in Kafka?

core/controller/src/main/scala/whisk/core/controller/Entities.scala
```diff
@@ -60,13 +55,15 @@ trait WhiskCollectionAPI
...
     with AuthorizedRouteProvider
-    with ValidateEntitySize
+    with ValidateRequestSize
     with ReadOps
     with WriteOps {
 
     /** The core collections require backend services to be injected in this trait. */
     services: WhiskServices =>
 
+    protected val allowedActivationEntitySize: Long = ActivationEntityLimit.MAX_ACTIVATION_ENTITY_LIMIT.toBytes
```
markusthoemmes: Keep it as `ByteSize` here and use `toBytes` where necessary?

common/scala/src/main/scala/whisk/core/entity/WhiskEntity.scala
```diff
@@ -158,9 +156,16 @@ case class LimitedWhiskEntityPut(
...
+        exec.flatMap { e =>
+            val is = e.size
+            if (is <= Exec.sizeLimit) None else Some(is, Exec.sizeLimit)
+        } orElse parameters.flatMap { p =>
+            val is = p.size
+            if (is <= Parameters.sizeLimit) None else Some(is, Parameters.sizeLimit)
+        } orElse annotations.flatMap { a =>
+            val is = a.size
+            if (is <= Parameters.sizeLimit) None else Some(is, Parameters.sizeLimit)
+        }
```
markusthoemmes: The return value of this could be confusing with regards to the error message returned to the user. If I paste a 40MB file and 2MB parameters I will get: Something like: You sent 2Megabytes but only 1 is allowed. Maybe we should add the `field` as well and put that into the error message?

core/controller/src/main/scala/whisk/core/controller/Entities.scala
rabbah: We have to do this conversion on every API call. I don't think it's worth it.

common/scala/src/main/scala/whisk/core/entity/ActivationEntityLimit.scala
rabbah: Not yet.

common/scala/src/main/scala/whisk/core/entity/WhiskEntity.scala
rabbah: https://github.com/openwhisk/openwhisk/pull/1779/commits/eb5e6112d46887aed7732d64d4e284b90df9bf82.

rabbah:
Addressed feedback from @markusthoemmes. Thank you!

rabbah:
PG2/986 re-approved.

PR #1781, opened 2017-01-29
Limit response size.

### LINKED ISSUE #566: Need to place upper limit on activation response object
As with other resources, this needs to be controlled and documented.  The idea would be that if a user action returns a result that is bigger than a threshold, then an error is flagged similar to exhausting other limits such as CPU limits.   @rabbah and I considered a 1MB limit as a reasonable starting point.

rabbah:
Moved `HttpUtil` to `container` package where it is actually used (it is no longer used elsewhere) and removed a lot of now dead code from it. Also removed connection pool since there is at most one connection to a container at any given time. Added mock HTTP testing for the container connection as well.

Then I changed container response from Option(int, array of bytes) to Either(container error, container result) with more useful typing.

The invoker detects larger than allowed response from container and truncate it properly and wrap with container error. The limit is set in the `core.entity` package following the model for other limits. At some point we should move all of these limits to ansible configs.

PG3/215.
Fixes #566.


common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala
```diff
@@ -88,29 +88,50 @@ protected[core] object ActivationResponse extends DefaultJsonProtocol {
...
-        } getOrElse {
-            // This indicates a terminal failure in the container (it exited prematurely).
-            containerError(abnormalInitialization)
+    protected[core] def processInitResponseContent(response: Either[ContainerConnectionError, ContainerResponse], logger: Logging): ActivationResponse = {
+        require(response.isLeft || !response.right.get.ok, s"should not interpret init response when status code is OK")
+        response match {
+            case Right(ContainerResponse(code, str, truncated)) => truncated match {
+                case None =>
+                    Try { str.parseJson.asJsObject } match {
+                        case scala.util.Success(result @ JsObject(fields)) =>
```
markusthoemmes: `import scala.util.Success`

core/invoker/src/main/scala/whisk/core/container/package.scala
```diff
@@ -57,8 +58,10 @@ package object container {
      * Start time, End time, Some(response) from container consisting of status code and payload
      * If there is no response or an exception, then None.
      */
-    case class RunResult(interval: Interval, response: Option[(Int, String)]) {
+    case class RunResult(interval: Interval, response: Either[ContainerConnectionError, ContainerResponse]) {
         def duration = interval.duration
+        def ok = response.isRight && response.right.get.ok
```
markusthoemmes: equivalent to:

`def ok = response.right.exists(_.ok)`. Yours to decide what is prettier.

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
```diff
@@ -0,0 +1,135 @@
...
+        var response: CloseableHttpResponse = null
+        try {
+            response = connection.execute(request)
+            Try(response.getEntity).filter(_ != null).map { entity =>
+                val statusCode = response.getStatusLine.getStatusCode
+                entity.getContentLength match {
+                    case length if length != -1 =>
+                        if (length <= maxResponseBytes) {
+                            val str = EntityUtils.toString(entity, StandardCharsets.UTF_8)
+                            Right(ContainerResponse(statusCode, str))
```
markusthoemmes: As this is tightly coupled to spray anyway, we might as well use `StatusCode` instead of an `Int` here for some extra typedness and easier interchangeability later?

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    private def execute(request: HttpRequestBase, timeoutMsec: Integer): Either[ContainerConnectionError, ContainerResponse] = {
+        var response: CloseableHttpResponse = null
+        try {
+            response = connection.execute(request)
+            Try(response.getEntity).filter(_ != null).map { entity =>
+                val statusCode = response.getStatusLine.getStatusCode
+                entity.getContentLength match {
+                    case length if length != -1 =>
```
markusthoemmes: Want to collapse the `if` and the `case`?

```
case length if length > 0 && length <= maxResponseBytes =>
    ...
case length if length > maxResponseBytes =>
    // truncate 
    ...
```

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
```diff
@@ -0,0 +1,135 @@
...
+        request.setEntity(entity)
+
+        execute(request, timeout.toMillis.toInt)
+    }
+
+    private def execute(request: HttpRequestBase, timeoutMsec: Integer): Either[ContainerConnectionError, ContainerResponse] = {
+        var response: CloseableHttpResponse = null
+        try {
+            response = connection.execute(request)
+            Try(response.getEntity).filter(_ != null).map { entity =>
```
markusthoemmes: Why is this wrapped in a `Try`? Should rather be `Option(response.getEntity).map { entity => ... }`

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
```diff
@@ -0,0 +1,135 @@
...
+                            Right(ContainerResponse(statusCode, str))
+                        } else {
+                            // truncate the response
+                            val bytesToRead = Math.min(length, maxResponseBytes)
+                            val bytes = IOUtils.toByteArray(entity.getContent, bytesToRead)
+                            val str = new String(bytes, StandardCharsets.UTF_8)
+                            Right(ContainerResponse(statusCode, str, Some(length.B, maxResponse)))
+                        }
+
+                    case _ => Left(NoResponseReceived())
```
markusthoemmes: This is just the `-1` case? I think this is too weak in general:

```
/**
     * Tells the length of the content, if known.
     *
     * @return  the number of bytes of the content, or
     *          a negative number if unknown. If the content length is known
     *          but exceeds {@link java.lang.Long#MAX_VALUE Long.MAX_VALUE},
     *          a negative number is returned.
     */
```

Why don't we do:
```
val contentLength = entity.getContentLength
if(contentLength > 0 && contentLength <= maxResponseBytes) {
    // good case
} else if(contentLength > maxResponseBytes) {
    // truncated case
} else {
    Left(NoResponseReceived())
}
```

I think that handles the degenerated cases better, as its not specified to return `-1` on every degenerated case? Thoughts?

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * This HTTP client is used only in the invoker to communicate with the action container.
+ * It allows to POST a JSON object and receive JSON object back; that is the
+ * content type and the accept headers are both 'application/json.
+ * The reason we still use this class for the action container is a mysterious hang
+ * in the Akka http client where a future fails to properly timeout and we have not
+ * determined why that is.
+ *
+ * @param hostname the host name
+ * @param timeoutMsec the timeout in msecs to wait for a response
```
markusthoemmes: Comment on `maxResponse` missing.

core/invoker/src/main/scala/whisk/core/container/WhiskContainer.scala
```diff
@@ -67,6 +64,9 @@ class WhiskContainer(
     var lastLogSize = 0L
     private implicit val emitter: PrintStreamEmitter = this
 
+    /** HTTP connection to container. Initialized on /init. */
+    private var connection: HttpUtils = null
```
markusthoemmes: 😢 

Shall we make this `private var connection: Option[HttpUtils] = None` to deal with the different cases in a safer way?

core/invoker/src/main/scala/whisk/core/container/WhiskContainer.scala
```diff
@@ -195,29 +197,19 @@ class WhiskContainer(
...
-                    warn(this, s"Exception while posting to action container ${t.getMessage}")
-                    None
-                }
+
+        val result = containerHostAndPort map { hp =>
+            if (connection == null) {
+                // initialize http connection to container
+                val hostWithPort = s"${hp.host}:${hp.port}"
+                connection = new HttpUtils(hostWithPort, timeout, ActivationEntityLimit.MAX_ACTIVATION_ENTITY_LIMIT)
             }
```
markusthoemmes: With the recommended change that'd be:

```
val result = containerHostAndPort.map { hp =>
    connection = connection.getOrElse(new HttpUtils(s"${hp.host}:${hp.port}", timeout, ActivationEntityLimit.MAX_ACTIVATION_ENTITY_LIMIT))
    connection.post(endpoint, msg)
}

File: core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala [hunk 10]
```diff
@@ -0,0 +1,135 @@
...
+
+        val request = new HttpPost(baseUri.setPath(endpoint).build)
+        request.addHeader(HttpHeaders.ACCEPT, "application/json")
+        request.setEntity(entity)
+
+        execute(request, timeout.toMillis.toInt)
+    }
+
+    private def execute(request: HttpRequestBase, timeoutMsec: Integer): Either[ContainerConnectionError, ContainerResponse] = {
+        var response: CloseableHttpResponse = null
```
markusthoemmes: Same as the other: Should this be `Option[CloseableHttpResponse]`?

common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala
rabbah: Success clashes.

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
rabbah: This helper uses Apache http (you might recall why).

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
rabbah: Good catch - that's what it was before the refactoring but I dont see this method throwing an exception. Will change.

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
rabbah: Thanks for the inspiration.

rabbah:
PG2/986 re-approved.

common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala
```diff
@@ -88,29 +88,50 @@ protected[core] object ActivationResponse extends DefaultJsonProtocol {
...
-                        val errorContent = errorOpt getOrElse invalidInitResponse(contents).toJson
-                        containerError(errorContent)
-                    case _ =>
-                        containerError(invalidInitResponse(contents))
-                }
-        } getOrElse {
-            // This indicates a terminal failure in the container (it exited prematurely).
-            containerError(abnormalInitialization)
+    protected[core] def processInitResponseContent(response: Either[ContainerConnectionError, ContainerResponse], logger: Logging): ActivationResponse = {
+        require(response.isLeft || !response.right.get.ok, s"should not interpret init response when status code is OK")
```
markusthoemmes: same as below, could be

`require(response.isLeft || !response.right.exists(_.ok), ...)`

I'm aware that it's safe to call `get` here. I'm just avoiding it in general.

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
```diff
@@ -0,0 +1,139 @@
...
+                if (contentLength >= 0 && contentLength <= maxResponseBytes) {
+                    // the good case
+                    val str = EntityUtils.toString(entity, StandardCharsets.UTF_8)
+                    Right(ContainerResponse(statusCode, str))
+                } else if (contentLength > maxResponseBytes) {
+                    // truncate the response
+                    val bytesToRead = Math.min(contentLength, maxResponseBytes)
+                    val bytes = IOUtils.toByteArray(entity.getContent, bytesToRead)
+                    val str = new String(bytes, StandardCharsets.UTF_8)
+                    Right(ContainerResponse(statusCode, str, Some(contentLength.B, maxResponse)))
```
markusthoemmes: I wonder if we could/should collapse the good case and the truncate case?

```
if(contentLength >= 0) {
    // reads at most maxResponseBytes
    val bytes = IOUtils.toByteArray(entity.getContent, contentLength min maxResponseBytes)
    val str = new String(bytes, StandardCharsets.UTF_8)
    // generate a notice iff the stream was truncated
    val truncationInfo = if(contentLength > maxResponseBytes) Some(contentLength.B, maxResponse) else None
    Right(ContainerResponse(statusCode, str, truncationInfo))
}
```

rabbah:
@markusthoemmes I removed "200", there's no reason really to keep it around anymore.

rabbah:
rebased and squashed.

perryibm:
merged

PR #1782, opened 2017-01-29
Allow for binary data response for http actions.

### LINKED ISSUE #567: Discuss support for different type of payload format for fire (trigger) and invoke (action)
Related to different types of Feeds, can produce events with payload, see #559 for background info.

Today `application/json` is the only format supported.

We can support other use cases by supporting other type of formats.

Please discuss here use cases and formats to support.

cc @sjfink @mbehrendt

rabbah:
The http web action must respond with either a plain text body or a base64 encoded body.
The header content-type must be set to interpret the reponse if base64 encoded body is given. (This applies to a JSON response as well.) Content-type must be one of known Spray types - no custom headers permitted.

This PR can subsume #1771 - the last commit adds binary support. PG1/1113.
Fixes #567.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Meta.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Meta.scala
@@ -16,21 +16,25 @@
 
 package whisk.core.controller
 
+import java.util.Base64
+
 import scala.concurrent.Future
 import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
 
 import spray.http._
-import spray.http.HttpHeaders.RawHeader
+import spray.http.HttpHeaders._
 import spray.http.MediaTypes._
 import spray.http.StatusCodes._
+import spray.http.parser.HttpParser
 import spray.httpx.SprayJsonSupport._
 import spray.json._
 import spray.json.DefaultJsonProtocol._
 import spray.routing.Directives
 import spray.routing.RequestContext
 import whisk.common.TransactionId
+import whisk.core.WhiskConfig
 import whisk.core.controller.actions.BlockingInvokeTimeout
 import whisk.core.controller.actions.PostActionActivation
 import whisk.core.database._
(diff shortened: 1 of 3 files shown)


core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -134,14 +136,54 @@ protected[core] object WhiskMetaApi extends Directives {
...
+            // header must be one of known headers, else reject it;
+            // if not specified, assume text/html.
+            headers find {
+                _.lowercaseName == HttpHeaders.`Content-Type`.lowercaseName
+            } map { t =>
+                // per spec, main/sub type and parameters are separated by ';'
+                // if this fails, it's not a valid header
+                val parts = t.value.split(';')
+                val mainAndSubtype = parts(0).split('/')
+                val contentType = mainAndSubtype(0).toLowerCase -> mainAndSubtype(1).toLowerCase
```
markusthoemmes: Might want to use destructuring, to get rid of the index getters:

```
val Array(type, _) = t.value.split(';')
val Array(mainType, subType) = type.split('/')
```

improves readability imho.

rabbah:
Addressed feedback from @markusthoemmes.

csantanapr:
I'm guessing that the list of content types is finite and can be found here:
https://github.com/spray/spray/blob/master/spray-http/src/main/scala/spray/http/MediaType.scala#L282

But for users. I think it would be convenient to spell it out some where.

In terms of binary data, I think decoding in the controller will be of extra work to service an action invoke compare on the load that does today, but it should not be that big of a deal if the 1MB result limit is in place.

Which we should spell out for users, that they should not send large videos/pngs about the limit.

Should I assume from this code that the 1MB is for the whole result and also after the user does the base64 encoding.

Meaning that the user needs to be aware that sum of the headers, code, and the body after being encoded (base64) can't be larger than 1MB.

I don't have any other comments than as always excellent job on adding this feature, I want to support form-url-encoded to be able to implement a Slack command, and closing the circle on doing an entire web page including images.

rabbah:
PG2/986 re-approved

PR #1783, opened 2017-01-29
Documenting meta api and web actions.

rabbah:


rabbah:
Excellent!

rabbah:
@csantanapr I reorganized the docs taking your input into account.

michaelmarth:
Kudos - this is an excellent improvement which will greatly extend OW's possible use cases.

The one thing I somewhat stumbled over is that the current proposal's "web-export" annotations seem to mix orthogonal concerns, in particular billing vs functional concerns (headers etc). Could we take the "anon access" out of "web-export" and make it an explicit annotation?

btw: the way I read it, a package can be _both_ "meta" *and* "web-export" (so that shared libraries producing HTML are possible). Right?

rabbah:
> Could we take the "anon access" out of "web-export" and make it an explicit annotation?

Great observation - Yes indeed. I considered this to be part of moving this code out of "experimental" - these features can be made available to existing mainline activations indeed. In the future enhancement section of the doc, I added one way to do that.

> a package can be both "meta" and "web-export"

Yes it's possible to fix - although I think by extending the web actions annotations, it's questionable if we need the meta approach (they can be subsumed by web actions). It's just that historically I added meta first, and then generalized for web.

Great thought provoking feedback - thank you!

michaelmarth:
Rodric, I thought a bit more about this - especially in the light of [this comment](https://github.com/openwhisk/openwhisk/issues/1793#issuecomment-277006734). IIUC the current proposal then developers would choose at _design_ time if an action is a "web-export" or not. From an _internal OW_ perspective that means choosing if the features described [here](https://github.com/openwhisk/openwhisk/pull/1783/files#diff-444445a530f5b36d76f58f15f9220f49R39) are enabled. But look at it from a _developer_'s perspective: it means choosing which HTTP clients to support - either browsers or non-browsers (IoT devices, curl, etc).
One could argue that developers know the intended usage and can make that decision at design time. But at least in the scenario in above mentioned comment things become tricky: would action "D" the one catching errors be "web-export"'ed or not? Would it depend on the annotations of the whole composite?
Rather than trying to come up with a proposal/solution for the complexity above, I have a more generic proposal (or at least question): can we make *all* actions "web-export"? What would happen if we did that?
The big advantage I see with this approach is that developers could consistently use the features you have [described](https://github.com/openwhisk/openwhisk/pull/1783/files#diff-444445a530f5b36d76f58f15f9220f49R39) in _any_ action they write - i.e. they would not have to make design time choices about the user agents that use the action.
Potential downside I see: adding semantic meaning to fields that are potentially used already might break legacy code (but I guess that could be managed through deprecation etc). Mabe there are other disadvantages I am not aware of.
WDYT?

rabbah:
The reason I put all these features under a separate API route is precisely not to worry about legacy, or how to resolve tension with the main line API at first. The web action features are orthogonal to the try-catch solutions and should be considered separately.

I would plan to promote the web actions related changes to the main line action activation in stages. For example, while we can introduce path projection for action invokation with a small change to the path matcher, supporting extensions will be a breaking change (if we require it on all activations, it's not backward compatible, and if we treat it as an extension when present, then it's also a breaking change since it might have been part of the action name). Allowing form data and query parameters is doable on the v1 API.

One approach I considered is to disallow extensions on the main API path but offer all the other features. Extensions would remain in effect only for web-exported actions.

What you're pointing out is that web-export currently implies a few semantic changes that are conflated. They are:

1. anonymous vs authenticated activation
2. extensions
3. result path projection
4. query and form data
5. http caller context
6. error handling

I don't disagree with your insights that we should promote these features to the main API for action activation. It's just a matter of doing so in a deliberate order and considering impact on existing code. A v2 of the API could give us greater latitude and there are other reasons to forge ahead with a second rev of the API.

michaelmarth:
Hi Rodric, thank you! However, I believe I was not very clear with my question/proposal (or I don't understand you answer :) )
What I meant was: IIUC the current proposal then a developer has to explicitly set the web-export annotation on actions in order to use web-export's features. I wondered if we could remove this explicit web-export annotation and give all actions the features that are currently associated with web-export-annotated actions only.
This question is basically independent from the migration plan - it is more a question where we want to land.

For completeness, let me also be a bit more clear on _why_ I ask this question:
Consider an app developer who writes an action that (depending on some application logic) produces either HTML or an HTTP redirect. Pretty simple. He can deploy that action with the web-export.
But the way it is currently specified, _only_ actions can be web-export-annotated. So, how should a composite like a sequence behave - which might contain a mixture of web-export-annotated actions and not-annotated actions? Would path projection for the whole sequence work?
I could think of different ways to answer/resolve this complexity, but the simplest one I could think of was to not have the web-export annotation at all - and provide the same feature-set to all actions.

rabbah:
> I wondered if we could ... give all actions the features that are currently associated with web-export-annotated actions only.

Yes we can - I suppose I didn't make that explicit (that we can and we should). I have been thinking about the "how" - taking it for granted that many of these features are useful for all actions.

rabbah:
> So, how should a composite like a sequence behave - which might contain a mixture of web-export-annotated actions and not-annotated actions? Would path projection for the whole sequence work?

In the current implementations, it's a matter of where the annotation is attached and how the action is activated. Really you should think of `web-export` as a feature toggle to make all the feature accessible to the API nested under "experimental".

rabbah:
> I could think of was to not have the web-export annotation at all - and provide the same feature-set to all actions.

I think this is problematic - I don't know how to talk about this without talking about the how. Do we introduce a new API to invoke actions with these features? Do we fold it into the current API? etc.

mbehrendt:
along the lines of the chat we had yesterday -- could we make the http support available without requiring an annotation, while anonymous invocation requires it?

rabbah:
OK to be very clear: Yes we can. And we will.

mbehrendt:
cool, thank you. Referring back to your earlier comment:

> I think this is problematic

what specifically is problematic here (I *think* @michaelmarth essentially suggested the same thing as I did). Is it that he was suggesting to not have an annotation at all vs I suggested to still have it, but with a reduced scope?

docs/webactions.md
```diff
@@ -0,0 +1,141 @@
...
+3. `__ow_meta_path`: the unmatched path of the request (matching stops after consuming the action extension).
+
+The request may not override any of the named `__ow_` parameters above; doing so will result in a failed request with status equal to 400 Bad Request.
+Action parameters may also be protected and treated as immutable. To finalize parameters, and to make an action web accessible, two [annotations](annotations.md) must be attached to the action: `final` and `web-export` either of which must be set to `true` to have affect. Revisiting the action deployment earlier, we add the annotations as follows:
+
+```
+$ wsk action create /guest/demo/hello hello.js \
+      --parameter name Jane \
+      --annotation final true \
+      --annotation web-export true
```
markusthoemmes: Had to scroll way too far to find the example command. Maybe put it straight above below the initial example?

rabbah:
Thanks @csantanapr for the modifications - I've accepted them and made two small passes.

rabbah:
Made one more change to lift the HTTP bits higher.

rabbah:
note the error handling is blocked by #1801.

rabbah:
@michaelmarth @mbehrendt I think I addressed your comments with some reorg and the futures section.

mbehrendt:
thx a lot @rabbah .

Just a few comments:
* I still think it would be preferrable to start with the raw http section ("Handling HTTP requests with actions"), followed by an explanation of the html shorthand.
* For completeness, should the description of the `error` field be included in the enumeration of potential json content in the "Handling HTTP requests with actions" section?

rabbah:
@mbehrendt the intro is all about HTTP actions - the four examples use `body` and `headers` and `code` plus `.http`. I think you can't start with the second section without the intro plus the examples are intuitive, aren't they?

I also find it hard to lift the error section up without all the other context around it.

mbehrendt:
ah, thx for the clarification, I had misread, my mistake

docs/webactions.md
```diff
@@ -0,0 +1,133 @@
...
+3. `__ow_meta_path`: the unmatched path of the request (matching stops after consuming the action extension).
+
+The request may not override any of the named `__ow_` parameters above; doing so will result in a failed request with status equal to 400 Bad Request.
+
+## Additional features
+
+Web actions bring some additional features that include:
+
+1. `Content extensions`: the request must specify its desired content type as one of `.json`, `.html`, `.text` or `.http`. This is done by adding an extension to the action name in the URI, so that an action `/guest/demo/hello` is referenced as `/guest/demo/hello.http` for example to receive an HTTP response back.
+2. `Projecting fields from the result`: the path that follows the action name is used to project out one or more levels of the response. For example,
```
rabbah: well it is true - if you follow the contract.
enhancing/changing this is subject of the issue you opened.

docs/webactions.md
```diff
@@ -0,0 +1,133 @@
...
+Or returns an `image/png`:
+```
+function main() {
+    let png = <base 64 encoded string>
+    return { headers: { 'Content-Type': 'image/png' },
+             code: 200,
+             body: png };
+}
+```
+
```
rabbah: what do you mean?

docs/webactions.md
```diff
@@ -0,0 +1,133 @@
...
+2. `Projecting fields from the result`: the path that follows the action name is used to project out one or more levels of the response. For example, 
+`/guest/demo/hello.html/body`. This allows an action which returns a dictionary `{body: "..." }` to project the `body` property and directly return its string value instead. The projected path follows an absolute path model (as in XPath).
+3. `Query and body parameters as input`: the action receives query parameters as well as parameters in the request body. The precedence order for merging parameters is: package parameters, action parameters, query parameter, body parameters with each of these overriding any previous values in case of overlap . As an example `/guest/demo/hello.http?name=Jane` will pass the argument `{name: "Jane"}` to the action.
+4. `Form data`: in addition to the standard `application/json`, web actions may receive URL encoded from data `application/x-www-form-urlencoded data` as input.
+5. `Activation via multiple HTTP verbs`: a web action may be invoked via one of four HTTP methods: `GET`, `POST`, `PUT` or `DELETE`.
+
```
rabbah: you mean in addition to this example?
this is not a tutorial on how to use these new features but rather documenting what the new features are. the examples should be small and self contained. there's no magic here - they're just actions is the point. i don't think showing every possible combination really helps.
i tend to believe it's the opposite - makes the exposition confusing.

learn by trying, these are just actions.

csantanapr: >you mean in addition to this example?

No,  that's the example that's it nothing more.

docs/webactions.md
csantanapr: This are outdated comments, see below the example of `application/json` it's already in the PR

csantanapr:
merged ee656247142c05cf62af96379813416179703a1a

rabbah:
Thanks for all the good feedback.

PR #1786, opened 2017-01-31
Include node-rdkafka@0.6.4

bjustin-ibm:
This can be used to write actions which communicate with Kafka instances, a key integration point.
Also including SSL and SASL libraries to allow communication with Kafka instances using SASL authentication (e.g. Message Hub)

### DIFF
--- a/core/nodejs6Action/Dockerfile
+++ b/core/nodejs6Action/Dockerfile
@@ -7,6 +7,17 @@ RUN curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-
   && rm "node-v$NODE_VERSION-linux-x64.tar.gz"
 
 
+# install SASL libs
+RUN apt-get install -y \
+    g++ \
+    make \
+    python \
+    openssl \
+    libssl-dev \
+    zlib1g-dev \
+    libsasl2-dev \
+    libsasl2-modules
+
 # workaround for this: https://github.com/npm/npm/issues/9863
 RUN cd $(npm root -g)/npm \
  && npm install fs-extra \
@@ -43,6 +54,7 @@ moment@2.17.0 \
 mongodb@2.2.11 \
 mustache@2.3.0 \
 nano@6.2.0 \
+node-rdkafka@0.6.4 \
 node-uuid@1.4.7 \
 nodemailer@2.6.4 \
 oauth2-server@2.4.1 \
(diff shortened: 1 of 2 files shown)


core/nodejs6Action/Dockerfile
```diff
@@ -7,6 +7,17 @@ RUN curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-
   && rm "node-v$NODE_VERSION-linux-x64.tar.gz"
 
 
+# install SASL libs
+RUN apt-get install -y \
+    g++ \
+    make \
+    python \
```
rabbah: 😳

bjustin-ibm: Python is needed by node-gyp to compile the underlying C library (librdkafka)

rabbah: i'm curious what is the image size before/after the pr?

bjustin-ibm: @rabbah Before, the image is 470M and after it is about 692M. I can get it down to about 623M (or perhaps a _bit_ smaller) but that requires a fairly drastic change like this: cdfd63d2a6767bc984b63247c50d3609e951abe2

The trouble is that g++, make, and python are only needed during the npm install step. While I can remove these later in the Dockerfile, there is no space savings in the entire image because the adds and deletes happen in separate RUN instructions and, therefore, separate layers. However, putting everything in a shell script and executing that in a single RUN instruction allows the deletes to effect the size of the resulting image.

mbehrendt:
@bjustin-ibm this should also include an update to the docs
- add this to the list of preinstalled pkgs
- include a code example for how to use this.

is that already captured elsewhere?

bjustin-ibm:
@mbehrendt I have updated the list of available packages with b6a7b57a6a290fa83313318583bb6b98a1a0a442. However, we typically rely on individual npm package providers to supply their own examples, rather than writing our own examples for each included package. Do you think this should be sufficient in this case as well?

mbehrendt:
@bjustin-ibm i agree we shouldn't provide code samples for any random npm package we provide ootb. However, providing direct doc guidance to users on how to programmatically interact with one of our key event providers only has upsides in terms of user experience, in my mind.

rabbah:
@bjustin-ibm close per https://github.com/openwhisk/openwhisk/pull/2188#issuecomment-298059147?

Issue #1431, opened 2016-10-24
Document How to Setup OpenWhisk Using Docker for Mac.


dubeejw:
Currently, the documentation for setting up Docker on a Mac illustrates how to use OpenWhisk with Docker Machine. The documentation should be updated to demonstrate how to setup OpenWhisk on Docker for Mac.

style95:
@dubeejw Is there anything in progress about this issue?
If not, I want to contribute on this issue.

It looks many things should be discussed such as "ansible environment in mac", "how to consider xhyve vm as local environment" and so on..

dubeejw:
@style95, feel free to help out. I don't think any documentation has been created yet for this issue.

style95:
While proceeding, I will describe the situation and workaround here.
Then it will be better for us to have same view and someone might give some advice or opinion.

style95:
I faced two issues.

### 1. Ip address for local environment

In Docker for Mac environment, there is no docker network interface.
That means there is no 172.17.0.0/32 network.

This is because Docker for Mac takes advantage of xhyve VM.
In xhyve VM, there might be above network.
But in OS X side, there is no such network, it just do the port-forwarding to xhyve VM.

So current local setup inventory file does not work for "Docker for Mac".
I changed ip address of local inventory file as follow.

```
; db group is only used if db_provider is CouchDB
[db]
10.64.131.142 ansible_connection=local
```


### 2. Upper case in db name

While proceeding "wipe.yml" step, I noticed that if there any upper case in "ansible_user_id" or "ansible_hostanme", the step fails.

It can be easily resolved by changing them manually to lower case.
However, it will be better to make them to lower case during the ansible deployment steps because such case can be frequently occurred.

While I am doing this task, I manually changed my user name and host name to lower case.

style95:
### Enabling Docker remote api
I faced an another issue:

```
TASK [invoker : determine docker root dir] *************************************
Friday 13 January 2017  11:50:22 +0900 (0:00:00.113)       0:03:50.436 ********
fatal: [10.64.131.142]: FAILED! => {"changed": false, "content": "", "failed": true, "msg": "Status code was not [200]: Request failed: <urlopen error [Errno 61] Connection refused>", "redirected": false, "status": -1, "url": "http://10.64.131.142:4243/info"}
```

With "Docker for Mac", docker remote api is not available by default.
It is listed as known issue in docker.com.
[...]

PR #1790, opened 2017-02-01
Make OW run with docker for mac

style95:
Draft version of documentation to setup OW on a Mac host is added.

I have no idea on documentation convention, PG(?), rebase rule.
I want someone to guide me about these community rules.

I know this branch is conflicting with master branch.
Once I get the review, I will amend them and rebase it.

I tried to minimize the changes and maximize to utilize existing contents.


rabbah:
Thanks Dominic for this pull request!

The git flow we follow is described here https://github.com/openwhisk/openwhisk/wiki/Contributing:-Git-guidelines; as you said we can deal with that later. For reference PG refers to our Jenkins CI. For documentation it is not necessary usually. Lastly on protocol, have you submitted an Apache CLA?

style95:
Thank you for sharing, Rodric.
I will look into it

Yes I have submitted an Apache CLA, got the response and this would clse #1431.

ansible/README.md
```diff
@@ -17,7 +17,35 @@ Nothing to be done, Ansible is already installed during vagrant provisioning.
 You can skip setup and prereq steps as those have been done by vagrant for you.  
 You may jump directly to [Deploying Using CouchDB](#deploying-using-couchdb)
 
-#### Mac users
+#### Docker for Mac users
+Docker for Mac does not provide any official ways to fulfill some requirements for Openwhisk.
```
rabbah: here and elsewhere: Openwhisk -> OpenWhisk

ansible/README.md
```diff
@@ -17,7 +17,35 @@ Nothing to be done, Ansible is already installed during vagrant provisioning.
...
+You need to depends on the workarounds until Docker provides official methods.
+
+##### Enable docker remote API
+
+During the deployment steps, each component communicates with docker via docker remote API.
+Currently however, Docker for Mac dose not support such a feature.
```
rabbah: dose -> does

tools/macos/docker-for-mac/README.md
```diff
@@ -0,0 +1,58 @@
...
+# Prerequisites
+
+The following are required to build and deploy OpenWhisk from a Mac host:
+
+- [Docker 1.12.0](https://docs.docker.com/docker-for-mac/)
+- [Java 8](http://www.oracle.com/technetwork/java/javase/downloads/index.html)
+- [Scala 2.11](http://scala-lang.org/download/)
+- [Ansible 2.1.2.0](http://docs.ansible.com/ansible/intro_installation.html)
+
+**Tip** Versions of Docker and Ansible are lower than the latest released versions, the versions used in OpenWhisk are pinned to have stability during continues integration and deployment.
```
rabbah: continues -> continuous

rabbah:
I'll give this a try - we don't have a CI for docker on mac but perhaps we should use this opportunity to set it up; @csantanapr what do you think?

csantanapr:
Thanks @style95 for the PR
Thanks @rabbah for jumping so quickly on the review. 

@style95 where you able to run all test cases ? I recall there some tests that don't  work if a special network route it's not created, we run a script currently when using docker-machine

I have to give this a try and see if it works for me on my system. 

I would say for now supporting both methods are OK but wonder if in the future we drop docker-machine and stick with docker for Mac. 

@rabbah I agree that would be good to have CI for mac, without the need for running a Docker host VM, now the CI job can run on a mac VM

Need to look into getting a couple of VMs from Apache INFRA to start building our Jenkins jobs, wonder if they also have macOS envs.

style95:
@csantanapr Actually, I just ran `gradlew distDocker` on my local macbook.

I will rebase my code with the latest master branch and apply the feedback.
After that, I will try to run tests and see what happens.

As of now, if I run `gradlew test`, I get the following error.

```
:tests:createKeystore FAILED

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/Naver/git/openwhisk/tests/build.gradle' line: 78

* What went wrong:
> /Users/Naver/git/openwhisk/whisk.properties (No such file or directory)

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 56.276 secs
```

csantanapr:
Hi @style95 you need to deploy the system before you can run tests. 
The tests run against a running system. 

Follow the instruction how to deploy with ansible first, you can verify first you can invoke echo action. 
The whisk.properties gets created as part of the deployment. 

The run the tests against the running system.

style95:
@csantanapr, I rebased the code to the latest master and ran the tests.

From ActionProxyContainerTests, all the tests are being failed.

It looks each test is creating a container and requesting to the container with the container IP address(172.17.0.0/32).
But as you know that Ip address is not accessible from a host(Mac).

We may have few options.
1. Update all the tests to use main docker endpoint ip instead of container IP.
In this case port should be different with existing components.

2. Make a Mac host be accessible to container IP.
As of now, I have not detail ideas, but we may create overlay network with network configuration tool such as flannel.

However, I am not sure these ideas are feasible.

```
.
.
.
actionContainers.ActionProxyContainerTests > openwhisk/dockerskeleton should run sample without init FAILED
    java.util.concurrent.TimeoutException: Request to //172.17.0.12:8080 could not be completed in time.

actionContainers.ActionProxyContainerTests > openwhisk/dockerskeleton should run sample with init that does nothing FAILED
    java.util.concurrent.TimeoutException: Request to //172.17.0.12:8080 could not be completed in time.

actionContainers.ActionProxyContainerTests > openwhisk/dockerskeleton should respond with 404 for bad run argument FAILED
    java.util.concurrent.TimeoutException: Request to //172.17.0.12:8080 could not be completed in time.

actionContainers.ActionProxyContainerTests > openwhisk/dockerskeleton should fail to run a bad script FAILED
    java.util.concurrent.TimeoutException: Request to //172.17.0.12:8080 could not be completed in time.

actionContainers.ActionProxyContainerTests > openwhisk/dockerskeleton should extract and run a compatible zip exec STANDARD_OUT
    JUnit version is: 4.11

actionContainers.ActionProxyContainerTests > openwhisk/dockerskeleton should extract and run a compatible zip exec FAILED
    java.util.concurrent.TimeoutException: Request to //172.17.0.12:8080 could not be completed in time.

actionContainers.ActionProxyContainerTests > openwhisk/dockerskeleton should run and report an error for script not returning a json object FAILED
    java.util.concurrent.TimeoutException: Request to //172.17.0.12:8080 could not be completed in time.

.
.
.
```

csantanapr:
@style95 thanks for checking on tests and reporting results and options.

I suspected that you were going to hit that problem, this is why on using docker-machine we have a script we run to enable the network connection.
https://github.com/openwhisk/openwhisk/blob/master/tools/macos/tweak-dockerhost.sh
```
MACHINE_NAME=whisk
MACHINE_VM_IP=$(docker-machine ip $MACHINE_NAME)

sudo route -n delete 172.17.0.0/16
sudo route -n -q add 172.17.0.0/16 $MACHINE_VM_IP
netstat -nr |grep 172\.17
```

Running this script would not fix the problem?

Option 1, is not good as the test needs to remain as is since this is really how it works in reality when the system is deployed on ubuntu production.
I wen this path in the past, as I was not a fan of always running this script, but discover that the test actually is design in this way on purpose, using the private network from invoker container to runtime container.

Not familiar with Option 2, but it sounds that making the network work would be the desire solution.

@rabbah @domdom82 thoughts?

style95:
@csantanapr Thank you for the opinion.

Regarding `tweak-dockerhost.sh` script, unfortunately it does not work on docker for mac environment.

That script adds route rule for 172.17.0.0/16 network.
All the packets sent to some containers will be forwarded to docker machine.

In the docker machine, which is linux most of the cases, there is docker0 network, 
and it will properly forward the packets to a target container.

In docker-machine environment, docker-machine itself also belongs to container network.
It is a gateway(172.17.0.1) for container network.
So it can forward the packets to container, but in docker for mac environment, it is not possible..

I am looking for workaround.

markusthoemmes:
@style95 Yes, tests will try to talk to components directly rather than through the API.

style95:
@markusthoemmes Thank you for let me know.
Let me clarify my query.

In both local/distributed environment, user can CRUD/invoke of actions.
In local environment, endpoint IP is 172.17.0.1 and distributed environment, it is host IP.
So if there is no case user should directly access to containers with container IP such as 172.17.0.X, it might work.

I am just wondering in which cases user/components are communicating with containers with container IPs directly.
As @csantanapr mentioned above, if this is really how it works in reality, it will be better to find workaround if I know the cases.

As per my understanding, user/components communicate each other according to the IPs in `whisk.properties` file.
So I just thought that if test cases also can communicate with the IP in `whisk.properties` file rather than container IP which is retrieved by `docker inspect`, it might work.

csantanapr:
Hi @style95 the network path is required because all the unit tests in 
https://github.com/openwhisk/openwhisk/blob/master/tests/src/actionContainers/

The user will never call directly to the containers for /init and /run on port 8080, this will be the invoker.
The tests are there to emulate/mock the invoker calling out /init and /run to just test the different runtime containers and that they behave the expected way
https://github.com/openwhisk/openwhisk/blob/master/tests/src/actionContainers/ActionContainer.scala#L139
```
        val mock = new ActionContainer {
            def init(value: JsValue) = syncPost(ip, 8080, "/init", value)
            def run(value: JsValue) = syncPost(ip, 8080, "/run", value)
        }
```

I really would like to offer users and contributors to be able to run OpenWhisk on their macs with Docker for Mac.

I proposed that we create a new test suite in gradle for testing on docker for mac that exclude this test cases that are not possible to run because of the network constraints.
Something similar on what we have today for Travis https://github.com/openwhisk/openwhisk/blob/master/tests/build.gradle#L30
```
task testOnDockerForMac(type: Test) {
    exclude '**/actionContainers*'
}
```
Contributors can use this gradle tasks to run all tests excepts those. `./gradlew :tests:testOnDockerForMac`
If they submit a PR Travis will run the tests for them (They always have the option in addition to run them locally with a ubuntu vm if they are working in that area of the code and to work locally)

And for other Contributors using docker-machine they can continue running the whole suite locally with `./gradlew :tests:test`

csantanapr:
@rabbah @markusthoemmes What you think about proposal on defining a gradle task for docker for mac users?

markusthoemmes:
@csantanapr not a fan at all. Is it maybe possible to achieve this in a different way?

csantanapr:
>@csantanapr not a fan at all. Is it maybe possible to achieve this in a different way?

I ran out of ideas, that was my last resource.

What do you propose?

style95:
Actually, I could manually send the /run request to dockerSkeleton container from host.

```
macbook15:openwhisk style95$ curl -X POST http://172.17.0.19:8080/run
{
  "error": "This is a stub action. Replace it with custom logic."
}
```

The workaround is as follow.
I run squid http proxy container.

```
docker run --name squid -d --publish 3128:3128 sameersbn/squid:3.3.8-23
```
And configured http_proxy as that endpoint.

```
http_proxy=localhost:3128
```

This proxy container also belongs to docker0 network in xyhve VM and it can reach to all other containers as well.
So if we send any packet via this container, it will be properly forwarded via docker0 network(172.17.0.0/24)

The problem is however, tests are still failing.
It looks that traffic from tests code does not pass through the proxy.
I am looking for the way to configure proxy in tests code(akka.http.scaladsl.model.HttpRequest).

But if anyone have any idea, kindly share it.
Thanks in advance.

csantanapr:
That was what I was thinking to be able to run those tests, the tests will not be able to reach the internal network of the containers. So the solution will be some how use another container as trampoline/proxy to route the /int and /run to the internal ip and port 8080

csantanapr:
We could also create a custom mock-invoker container that only role is to route the test http calls, maybe this gets the test closer to reality, since this is how it actually works.

style95:
@csantanapr I am inclined to your first option.

If we find the way to route the traffic to proxy container, it will be much simpler.

Currently, test code is checking the OS type, and if it is mac, we can just enable the proxy configuration during the tests.
No changes on existing code are required and it will transparently work on other platform.

But I couldn`t find the way to configure proxy on test code yet.

csantanapr:
You mean how to change the scala test code to configure the http client with a proxy configuration?

Maybe @rabbah or @markusthoemmes might know how to do this.

style95:
Hi @rabbah 
Do you have any updates on this issue?

rabbah:
@style95 you mean on the proxy issue?

style95:
@rabbah I just want to know whether you have any idea or opinion.
You look very busy, so I don`t want to urge you.

rabbah:
no worries.

i think we should strive for little divergence, so not being able to run the container tests should be really off the table. since the proxy approach seems to provide a work around, i'd pursue that. this is not unlike what the tweak-dockermachine script is doing (it's bridging the routes from the host to the containers in the vm).

fwiw, i run the container tests often, and even use the approach to develop/debug regularly.

csantanapr:
@style95 
Let's try to get proxy approach by getting the test code leverage it.
Can you share the specifics on what have you tried ?

>I am looking for the way to configure proxy in tests code(akka.http.scaladsl.model.HttpRequest).

How you attempted so far to configure the http client?

style95:
To run the code seamlessly in different OS, it will be better for user to run the code without any code changes.
I tried to find the way to use environment variable or configuration rather than change the code.

Please refer to what I tried.

**1. Socks proxy configuration.**
I setup socks proxy with ssh to a container, and configure it in osx system configuration.

 * System Preferences -> Network -> Proxy  -> SOCKS Proxy 
 * `networksetup -setsocksfirewallproxy "Wi-Fi" localhost 8888`

Both the ways did not work.

I could configure the proxy and traffic go through the proxy during web browsing.
But test code traffic didn`t pass through the proxy.

**2. http_proxy environment variable.**

I setup squid http proxy as I mentioned above.
I set environment variable http_proxy and https_proxy

I could send any http request to container using curl.
But test code traffic didn`t pass through the proxy.

**3. gradle http proxy**

I tried to configure proxy in gradle side.

gradle.properties
```
systemProp.http.proxyHost=localhost
systemProp.http.proxyPort=3128
```

* Refercence: https://docs.gradle.org/current/userguide/build_environment.html#sec:accessing_the_web_via_a_proxy

It did not work.

**4. IntelliJ proxy setup**
I tried to run test code on IntelliJ and configure the proxy in intellij configuration.

Preferences -> Appearance & Behavior -> System Settings -> HTTP Proxy -> Host name: localhost, Port number: 3128

Unfortunately, it did not work also..

At this point, I thought system-wide configuration might not work for the test code.
I started looking into the code itself.

I thought `akka.http.scaladsl.model.HttpRequest` and `AkkaHttpUtils.singleRequest` are the keys.

I look into the following documents.
http://doc.akka.io/docs/akka/2.4.3/scala/http/common/http-model.html
http://doc.akka.io/docs/akka/2.4.7/scala/http/configuration.html#akka-http-core

I found this information, and was trying to find out how to configure proxy in client side as currently Http().outgoingConnection is being used.

```
# IMPORTANT:
    # Please note that this section mirrors `akka.http.client` however is used only for pool-based APIs,
    # such as `Http().superPool` or `Http().singleRequest`.
    client = {
      # The default value of the `User-Agent` header to produce if no
      # explicit `User-Agent`-header was included in a request.
      # If this value is the empty string and no header was included in
      # the request, no `User-Agent` header will be rendered at all.
      user-agent-header = akka-http/${akka.version}
 
      # The time period within which the TCP connecting process must be completed.
      connecting-timeout = 10s
 
      # The time after which an idle connection will be automatically closed.
      # Set to `infinite` to completely disable idle timeouts.
      idle-timeout = 60 s
 
      # The initial size of the buffer to render the request headers in.
      # Can be used for fine-tuning request rendering performance but probably
      # doesn't have to be fiddled with in most applications.
      request-header-size-hint = 512
 
      # The proxy configurations to be used for requests with the specified
      # scheme.
      proxy {
        # Proxy settings for unencrypted HTTP requests
        # Set to 'none' to always connect directly, 'default' to use the system
        # settings as described in http://docs.oracle.com/javase/6/docs/technotes/guides/net/proxies.html
        # or specify the proxy host, port and non proxy hosts as demonstrated
        # in the following example:
        # http {
        #   host = myproxy.com
        #   port = 8080
        #   non-proxy-hosts = ["*.direct-access.net"]
        # }
        http = default
 
        # Proxy settings for HTTPS requests (currently unsupported)
        https = default
      }
```

Any idea?

rabbah:
Setting the proxy on the akka-http client should work - it should be possible to do that without the configuration file.

style95:
Where is akka-http client being used?

rabbah:
https://github.com/openwhisk/openwhisk/blob/master/tests/src/actionContainers/ActionContainer.scala#L173

style95:
It would be the best if user configured the system proxy, then tests also use the same proxy settings no matter which OS user uses.

Setting up `http_proxy` and `https_proxy` environment variables are very common way.
There are many systems which recognize those environment variables.

style95:
I am not sure whether akka-http support proxy configuration.
https://github.com/akka/akka-http/issues/115

style95:
I could send the request to the container using container IP with [scalaj-http](https://github.com/scalaj/scalaj-http) which explicitly supports proxy configuration.

```
import scalaj.http._
.
.
.
it should "run sample without init" in {
    println(Http("http://172.17.0.3:5984").proxy("localhost", 3128).asString)
  }
```

ansible/README.md
```diff
@@ -17,7 +17,37 @@ Nothing to be done, Ansible is already installed during vagrant provisioning.
 You can skip setup and prereq steps as those have been done by vagrant for you.  
 You may jump directly to [Deploying Using CouchDB](#deploying-using-couchdb)
 
-#### Mac users
+#### Docker for Mac users
+Docker for Mac does not provide any official ways to fulfill some requirements for OpenWhisk.
+You need to depends on the workarounds until Docker provides official methods.
```
ddragosd: Did you also consider the `docker-compose` method - https://github.com/openwhisk/openwhisk-devtools/tree/master/docker-compose ?

style95: Do you think in which case docker-compose can handle the limitation of Docker for Mac?

style95:
I added commits for proxy configuration.

Actually, it was working with all the tests in `testLean`.
But once I rebase the code to the latest master, it does not work.

I could fix that also, but before that I pushed the commits.
I am not quite sure the way I developed, so I want to have the same view with other people first.

Once we reach a consensus about how to resolve this issue, I will resolve all other test failures also.
I expect lots of comments.

core/invoker/build.gradle
```diff
@@ -12,6 +12,7 @@ repositories {
 
 dependencies {
     compile "org.scala-lang:scala-library:${gradle.scala.version}"
+    compile 'com.typesafe.play:play-ws_2.11:2.5.11'
```
rabbah: this seems unnecessary to include the invoker build. (do we really need it?)

core/invoker/src/main/scala/whisk/core/container/AkkaHttpUtils.scala
```diff
@@ -24,24 +24,38 @@ import scala.concurrent.duration.DurationInt
 import scala.concurrent.duration.FiniteDuration
 import scala.util.Try
 import akka.actor.ActorSystem
-import akka.http.scaladsl._
 import akka.http.scaladsl.model._
 import akka.stream.ActorMaterializer
-import akka.stream.scaladsl._
 import java.util.concurrent.TimeoutException
 
+import play.api.http.{ContentTypeOf, Writeable}
```
rabbah: we're not using this utility in the invoker - instead we use the apache http client.
@markusthoemmes is this likely to change with new invoker work?
perhaps we should move this utility to the test project.

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
```diff
@@ -85,6 +87,9 @@ protected[container] class HttpUtils(
             val containerResponse = Option(response.getEntity).map { entity =>
                 val statusCode = response.getStatusLine.getStatusCode
                 val contentLength = entity.getContentLength
+                if(statusCode == 503 && isProxied) {
```
rabbah: is this necessary? I had understood the issue was related to the unit tests for the container runtimes - not the invoker to container communication. If so, then these changes are not necessary since this class is not used for those unit tests.

style95: I found this unit tests are using this utility.
`whisk.core.container.test.ContainerConnectionTests`

core/invoker/src/main/scala/whisk/core/container/AkkaHttpUtils.scala
style95: If we can move this utility to the test project, it will be the best.
Changes will be limited to the test project.

core/invoker/build.gradle
style95: If we can move `AkkaHttpUtils` to test project, it is not required.

style95:
With this changes, CI for `docker for mac` is highly required.
When I rebase the code to the latest master, I noticed that newly added tests are designed in a way that proxy is not taken account.

If we keep supporting `docker for mac`, all the new tests should be designed with proxy configuration.
(I think, mostly this could be done by using AkkaHttpUtils in this pr)

rabbah:
Makes sense to me to have one supported mac platform (docker for mac).

style95:
One more proposal is, how about hiding `docker-machine` guide from `ansible/README.md`?
After merge of this pr, multiple guides for `docker for mac` and `docker-machine` will make users confused.

Since `docker for mac` is more natural way to use docker on OSX, I think it will be better to expose `docker for mac` guide only in `ansible/README.md`, though we keep the guide for `docker-machine` also.

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
style95: Once it is confirmed along with above AkkaHttpUtils issue, I will update them accordingly.

style95:
I will move AkkaHttpUtils to test project and make following tests to use AkkaHttpUtils to segregate HttpUtils from changes.
`whisk.core.container.test.ContainerConnectionTests`, `whisk.core.container.test.ContainerPoolTests`

Then `compile 'com.typesafe.play:play-ws_2.11:2.5.11'` may be only included in test project.
I hope this is the end of long journey..

core/invoker/src/main/scala/whisk/core/container/AkkaHttpUtils.scala
markusthoemmes: It is subject to investigation, yes. But I don't think it'll happen in the next weeks. I can just refactor back if I need it.

tools/macos/docker-for-mac/README.md
```diff
@@ -0,0 +1,58 @@
...
+- [Scala 2.11](http://scala-lang.org/download/)
+- [Ansible 2.1.2.0](http://docs.ansible.com/ansible/intro_installation.html)
+
+**Tip** Versions of Docker and Ansible are lower than the latest released versions, the versions used in OpenWhisk are pinned to have stability during continuous integration and deployment.
+
+
+[Homebrew](http://brew.sh/) is an easy way to install all of these and prepare your Mac to build and deploy OpenWhik. The following shell command is provided for your convenience to install `brew` with [Cask](https://github.com/caskroom/homebrew-cask) and bootstraps these to complete the setup. Copy the entire section below and paste it into your terminal to run it.
+
+```
+echo '
```
cclauss: Opens a quote but does not close it!

tools/macos/docker-for-mac/README.md
```diff
@@ -0,0 +1,58 @@
...
+# install homebrew
+/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+# install cask
+brew tap caskroom/cask
+# install java 8
+brew cask install java
+# install scala
+brew install scala
+# install pip
+sudo easy_install pip
```
cclauss: `brew install --upgrade python`  # would be safer / better
* it avoids relying on the Apple system Python that is know to have issues
* it installs its own (relatively current) pip
* that pip no longer requires the use of `sudo`  (The entire brew project is anti sudo for very good reasons)

tools/macos/docker-for-mac/README.md
```diff
@@ -0,0 +1,58 @@
...
+# install ansible
+sudo -H pip install ansible==2.1.2.0
```
cclauss: ~sudo -H~ pip install ansible==2.1.2.0  # using brew's pip

tools/macos/docker-for-mac/README.md
```diff
@@ -0,0 +1,58 @@
...
+# the following is required for running tests
+sudo -H pip install jsonschema
```
cclauss: ~sudo -H~ pip install jsonschema  # using brew's pip

tools/macos/docker-for-mac/README.md
```diff
@@ -0,0 +1,58 @@
...
+# install docker-py
+sudo -H pip install docker-py' | bash
```
cclauss: ~sudo -H~ pip install docker-py' | bash  # using brew's pip

tools/macos/docker-machine/README.md
```diff
@@ -0,0 +1,124 @@
...
+# install docker 1.12.0
+brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/33301827c3d770bfd49f0e50d84e0b125b06b0b7/Formula/docker.rb
+# install docker-machine
+brew install docker-machine
+# install java 8
+brew cask install java
+# install scala
+brew install scala
+# install pip
+sudo easy_install pip
```
cclauss: See comments above about using: `brew install --upgrade python` and about removing `sudo` from pip installs.

tests/build.gradle
```diff
@@ -46,6 +48,7 @@ dependencies {
     compile 'io.spray:spray-testkit_2.11:1.3.3'
     compile 'com.google.code.gson:gson:2.3.1'
     compile 'org.scalamock:scalamock-scalatest-support_2.11:3.4.2'
+    testCompile 'com.typesafe.play:play-ws_2.11:2.5.11'
```
markusthoemmes: Make it `compile`. I chose to have those as compile-time dependencies, because we have an anomalie having a seperate test-project anyway. Gradle doesn't import testCompile dependencies if i reference the test project (which catalog etc. do for example).

style95:
Finally all tests have passed on OSX!!

@rabbah @markusthoemmes kindly review this.

Changes are as follow:

1. Change `mac` environment for `docker for mac`
2. Add new environment `docker-machine` for `mac with docker-machine`
3. Make `HttpUtils` respect proxy.
4. Make `HttpUtils` retry on 503 service unavailable error.

style95:
Are there any updates on this?
Is this pending until the completion of new implementation of ContainerPool?

rabbah:
No it isn't. We were trying to find infrastructure to setup automatic testing for a mac based deployment.

rabbah:
Also since this replaces the previous mac environment, all mac developers will need to update to the latest docker machine support for mac.

style95:
Yes. At first time, I tried to use `local` whick_version_name as well to make user deploy OW transparently no matter of OS.

Only this part is the issue.

**ansible/roles/invoker/tasks/deploy.yml**
```
- name: define options when deploying invoker on Ubuntu
  set_fact:
    linuxOptions: "-v /usr/lib/x86_64-linux-gnu/libapparmor.so.1:/usr/lib/x86_64-linux-gnu/libapparmor.so.1"
  when: whisk_version_name != "mac"
```
I could fix this `when` statement, but thought that if we keep `mac` environment for `docker-machine`, user may think ow can`t be deployed on mac natively(with `local`).

So I made one more environment, docker-machine.

But as of now, I think it would be better to remove `mac` environment and keep `docker-machine` environment only.

Then user can deploy OW transparently with `local` as I intended at first.
Only someone wants to deploy OW with Docker-machine, they will use `docker-machine` environment.

In either cases, those who use `mac` currently, should change their environment.

How do you think?

csantanapr:
I agree this gives transition time
env mac will mean future using docker for Mac
env docker-machine means depreacted env and later removed to give time for folks to updgrade to docker for Mac

style95:
@rabbah @markusthoemmes @csantanapr 
Since there was no related task in progress, I took over the task to check the feasibility of mac travis build environment.

But other than that, don`t we need to discuss the way this pr implemented further?
I want to make sure there are no other issue to merge this pr.

AFAIR, @rabbah and @markusthoemmes had concern about changes on `HttpUtils` and the behavior in case of `503 service unavailable`.

Since now many things have been changed, I want to discuss it more.

Is travis mac build environment only requirement to merge this pr?
(Surely, I will rebase it)

rabbah:
iirc: 1) make native docker for mac a new supported platform with testing, 2) do not disrupt current users of docker machine, 3) avoid semantic changes to container protocol.

style95:
Regarding 3) do you have any idea?
I am looking for options to make squid proxy send intact response in case of connection failure.

style95:
Since squid proxy is returning `403 service unavailable` for all problematic requests, we might consider using ssh proxy instead of squid.

ansible/environments/docker-machine/group_vars/all
```diff
@@ -0,0 +1,35 @@
+whisk_version_name: mac
+nginx_conf_dir: /Users/Shared/nginx
+consul_conf_dir: /Users/Shared/consul
+whisk_logs_dir: /Users/Shared/wsklogs
+cli_nginx_dir: "{{ nginx_conf_dir }}/cli/go/download"
+docker_registry: ""
+docker_dns: ""
+
+db_prefix: "{{ ansible_user_id }}_{{ ansible_hostname|lower }}_"
```
mrutkows: Bot seeing the update to db_prefix to assure ansible_user_id is lower cased:
https://github.com/apache/incubator-openwhisk/pull/2217

style95: Thanks.
Surely I need to rebase it from the latest master

style95:
Rebased to the latest master.
TestLean has passed on my local setup.
Travis build is failing due to timeout.

style95:
@rabbah @markusthoemmes @csantanapr 
I updated the logic.
Now there is no retry logic on 503 service unavailable.

Let me explain previous problem first.

It was a kind of timing issue. When test codes send `syncPost(ip, 8080, "/init", value)`, there is a possibility that target action container is not ready yet. In this case, proxy server will get connection failed(No route to host), and accordingly return `503 service unavailable`.

I found the way to configure squid to retry when connection attempts fails with "No route to host".
It will retry for some time(it is configurable) in case of connection failure until the action container is ready.

Now every tests worked as expected except only once test case.

```
...
actionContainers.DockerExampleContainerTests > bad containers should timeout bad proxy with exception FAILED
...
```

Difference on this test is as follow:

This test run `badproxy` container and try to send `/init` request to it.
Since `badproxy` does not expose any port, squid proxy server get "Connection refused" instead of "No route to host" And in this case, squid proxy returns with `503 service unavailable` immediately.

I think, squid proxy is supposed to retry on both the cases.
I opened an issue at the squid repo: https://github.com/squid-cache/squid/issues/23

If this is resolved, it will finally make it.
In the meantime, please review the code again.

style95:
One question is.. do we really need this test?

```
...
actionContainers.DockerExampleContainerTests > bad containers should timeout bad proxy with exception FAILED
...
```

rabbah:
The test is to assert the invariant that a docker action which doesn't implement the right interface or protocol fails with a timeout exception.

ddragosd:
@style95 folks like me wanting to use docker for mac are thankful for the hard work you put in this PR over the last months. Kudos for this! 

I had a q on squid: how is squid different than an NGINX proxy which is deployed anyway for OW ?

ansible/environments/mac/group_vars/all
```diff
@@ -1,7 +1,7 @@
 whisk_version_name: mac
-nginx_conf_dir: /Users/Shared/nginx
-consul_conf_dir: /Users/Shared/consul
-whisk_logs_dir: /Users/Shared/wsklogs
+nginx_conf_dir: /tmp/nginx
```
style95: I think this is because I copied environment file from `local` environment at first time.
I will update it.

tests/src/test/scala/common/AkkaHttpUtils.scala
```diff
@@ -53,50 +64,41 @@ object AkkaHttpUtils {
...
-        retryInterval: FiniteDuration = 100.milliseconds)
-        (implicit system: ActorSystem) : Future[HttpResponse] = {
-
+         uri: String,
+         content: JsValue,
+         timeout: FiniteDuration,
+         retryOnTCPErrors: Boolean = false,
+         retryOn4xxErrors: Boolean = false,
+         retryOn5xxErrors: Boolean = false,
+                       retryInterval: FiniteDuration = 100.milliseconds)
```
style95: Will be updated accordingly

style95:
@ddragosd Thanks.
I also hope many people can develop ow with docker for mac.

The main Issues being handled in this pr are as follow.

1. Docker0 network does not exist on OSX. (No direct connection to container is possible)
2. Squid proxy returns `503 service unavailable` in case of connection failure(`connection refused`)
3. Because of 2, few tests are failing.

If you use docker-compose with docker for mac, surely you will face similar issues.
And same workaround will work for you.

Last time, I tried to solve 2 by changing the semantic of core modules.
But many people had concerns on it.
Now I am trying to change the behavior of proxy.


And regarding question on squid, I am using squid as a forward proxy.
In OW world, nginx is being used as a reverse proxy.

While proceeding this issue, I focused on minimizing manual user work and changes on existing module as much as possible.

Nginx also can be used as a forward proxy.
But we need many configuration changes on nginx.
(Even I am not sure nginx can work as forward and reverse proxy at the same time)

With squid, user can just `docker pull`/`docker run` it from public docker hub.
I have no preference on squid. I just used it because it is a mature solution.

I have tested few other forward proxies such as polipo, but did not work.
Even I was thinking of implementing new proxy by myself and publicize it on docker hub.
Since it requires too much efforts, I am looking for the other way.

style95:
I have a clue.
If I modify retry interval and recompile squid-proxy.
It shall work.

style95:
Finally, I could run above test also with manually built squid-proxy container.

```
actionContainers.DockerExampleContainerTests > openwhisk/example should run sample without init PASSED

actionContainers.DockerExampleContainerTests > openwhisk/example should run sample with init that does nothing PASSED

actionContainers.DockerExampleContainerTests > openwhisk/example should run sample with argument PASSED

actionContainers.DockerExampleContainerTests > bad containers should timeout init with exception PASSED

actionContainers.DockerExampleContainerTests > bad containers should abort init with empty response PASSED

actionContainers.DockerExampleContainerTests > bad containers should timeout run with exception PASSED

actionContainers.DockerExampleContainerTests > bad containers should abort run with empty response PASSED

actionContainers.DockerExampleContainerTests > bad containers should timeout bad proxy with exception PASSED
```

I will rebase the code and test it again.

style95:
All the tests have passed with `docker-for-mac` : )

```
BUILD SUCCESSFUL

Total time: 28 mins 51.595 secs
```

I will update the guide accordingly.

style95:
@rabbah @markusthoemmes @csantanapr 
Please review this.

To recall, let me recap the changes.

1. Socat container.
This is to enable docker remote api.

2. Docker-machine and Mac environment in ansible.
With this pr, there will be one more environment, `docker-machine`.
This environment is kept for those who use OW with docker-machine.
In the future, this environment will be deprecated.

So, `mac` environment means Docker for mac environment.


3. Convert `AkkaHttpUtils` to `Play-ws client`
Since `AkkaHttpUtils` does not support proxy configuration, http client is changed to `play-ws`.
There is no change in logic.

4. Squid proxy container
This is only to run unit tests with Docker for mac.
It will forward all the request to a given target container.

Since there was no retry delay on connection error in upstream squid proxy, following tests failed.

```
actionContainers.DockerExampleContainerTests > bad containers should timeout bad proxy with exception FAILED
```

So I added retry delay and manually compiled it and built docker container.
Now it works as expected.


One remaining task is update [this](https://github.com/apache/incubator-openwhisk/pull/1790/files#diff-5a5d069dcad99a9f8109ac77d1b9860cR60) to my manually built squid container.

I think it would be the best to publish that squid container to [openwhisk namespace](https://hub.docker.com/u/openwhisk/) in docker hub.
But if it`s not possible, I will publish it in my namesapce.(ex: `style95/squid:3.5`)

style95:
FYI, this script is what I am using to deploy OW with docker for mac.
It`s not that complicated.

```
export OPENWHISK_HOME=/Users/Naver/git/openwhisk
docker run -d -v /var/run/docker.sock:/var/run/docker.sock -p 4243:2375 bobrik/socat TCP4-LISTEN:2375,fork,reuseaddr UNIX-CONNECT:/var/run/docker.sock
docker run --name squid -d -v /Users/Naver/software/squid/squid.conf:/etc/squid3/squid.conf --publish 3128:3128 sameersbn/squid:3.3.8-23 ## this will be changed.

DOCKER0=`ifconfig | grep 172.17.0.1`

if [ -z "$DOCKER0" ]
then
    sudo ifconfig lo0 alias 172.17.0.1/24
fi

ansible-playbook -i environments/mac setup.yml
ansible-playbook -i environments/mac couchdb.yml
ansible-playbook -i environments/mac initdb.yml
ansible-playbook -i environments/mac wipe.yml
ansible-playbook -i environments/mac apigateway.yml
ansible-playbook -i environments/mac openwhisk.yml
```

style95:
I published custom version of squid proxy in my repository.

* Docker hub: https://hub.docker.com/r/style95/squid/
* Github: https://github.com/style95/squid-custom

And document is also updated accordingly.

Now everything is ready.

markusthoemmes:
I only looked at the code, the changes look solid to me. I'm just wondering: Should we just switch to use `HttpUtils` in tests as well to work around needing another dependency here which we don't use in the production system anyway?

My only anxious point is, that we might/should swap to use akka-http vs. Apache HTTP client at some point in the future, since we're bought into the akka ecosystem.

style95:
In this pr, it keeps current logic but just changed its client; keeps `HttpUtils` for production, and uses another http client(AkkaHttpUtils -> play-ws) for tests.

If it is desirable to switch to `HttpUtils` for tests as well, I could do this.
But since this pr is too long and not that readable, I will open another pr.

Regarding  migration from `Apache HTTP` to `Akka-Http`, there is [on-going PR](https://github.com/akka/akka-http/issues/115) to support proxy in `Akka-Http`.

Once it is merged, we can gladly take advantage of it.
(If required, I will contribute to it as well)

markusthoemmes:
@style95 thank you so much for your endurance, I greatly appreciate the time you put into researching these bits :)

Keeping `AkkaHttpUtils` as is but temporaly using `HttpUtils` for the tests without introducing a new dependency would be my preferred way to go for this. It seems like HTTP_PROXY support has landed in akka but we don't need to wait for it to propagate to a version. Going back to akka in a later PR is very fine for me.

rabbah:
Great work on this! I will confirm that it doesn't break docker-machine users (like me :) today.

style95:
@markusthoemmes ok I will try that approach

rabbah:
I am trying this PR - Since "mac" is default and that's not docker-machine, first observation is that docker-machine users will need to use `wskdev -t docker-machine` for the previous environment. Should we leave docker-machine as the default until we have more users with docker for mac?

Then I attempted to run a fresh deployment:
```
> redo fresh -t docker-machine -n
system setup
ansible-playbook -i environments/docker-machine setup.yml
install requisites
ansible-playbook -i environments/docker-machine prereq.yml
deploy couchdb
ansible-playbook -i environments/docker-machine couchdb.yml -e mode=clean
ansible-playbook -i environments/docker-machine couchdb.yml
initialize db with guest/system keys
ansible-playbook -i environments/docker-machine initdb.yml
recreate main db for entities
ansible-playbook -i environments/docker-machine wipe.yml
build/deploy system
/Users/rabbah/projects/openwhisk/tools/build/../..//gradlew distDocker  --parallel -PdockerHost=192.168.99.100:4243
ansible-playbook -i environments/docker-machine openwhisk.yml -e mode=clean
ansible-playbook -i environments/docker-machine openwhisk.yml
install catalog
ansible-playbook -i environments/docker-machine postdeploy.yml
```

This fails in setup:
```
TASK [gen untrusted certificate for host] *******************************************************************************************************************************************************************
Wednesday 21 June 2017  21:27:23 -0400 (0:00:00.048)       0:00:02.026 ******** 
fatal: [localhost]: FAILED! => {"failed": true, "msg": "The conditional check 'nginx.ssl.cert == \"openwhisk-cert.pem\"' failed. The error was: error while evaluating conditional (nginx.ssl.cert == \"openwhisk-cert.pem\"): {u'confdir': u'{{ config_root_dir }}/nginx', u'ssl': {u'path': u'{{ openwhisk_home }}/ansible/roles/nginx/files', u'cert': u'openwhisk-cert.pem', u'password_enabled': False, u'key': u'openwhisk-key.pem', u'password_file': u'ssl.pass'}, u'version': 1.11, u'port': {u'adminportal': 8443, u'api': 443, u'http': 80}}: 'config_root_dir' is undefined\n\nThe error appears to have been in '/Users/rabbah/projects/openwhisk/ansible/setup.yml': line 33, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n  - name: gen untrusted certificate for host\n    ^ here\n"}
```

Investigating further.

style95:
@rabbah 
I changed nothing but the name from `mac` to `docker-machine`.
I will also check on it.

Regarding default env for mac, I want to make clear what is the goal.
I believe it is desirable to make docker for mac as default.
Because docker-machine will be deprecated. (at least I hope that)
And it would be not easy to figure out how many people use `docker-machine` and `docker for mac` in the future.

I know to use word, `docker-machine` is inconvenient.
So that many people will move to docker for mac environment and be able to use OW in more convenient way.

If required, we can change `docker-machine` to some other word.
But I argue that `mac` should mean docker for mac with this PR.

One other option is, to use `local` environment for docker for mac as well.
I am also using it in that way in my local.

But in this case also, I believe it would be better to rename `mac` environment not to confuse users.

rabbah:
I'm suggesting something far simpler: think of this as a feature toggle. Keep all the renaming but we modify wskdev to initially continue to detect mac and map it to docker-machine. Then as we have more people migrating, we can flip the default setting and eventually drop docker-machine.

I'm not sure why I'm getting error on setup with docker-machine. On master, redo fresh is working for comparison.

style95:
ok good idea.
I will take it into account.

style95:
@markusthoemmes 
I noticed that scope of `HttpUtils` is `protected`.

```
protected[core] class HttpUtils(
    hostname: String,
    timeout: FiniteDuration,
    maxResponse: ByteSize) {
```

It`s not accessible from the class other than `core` package class.

I think it would be better to move `tests/src/tests/scala/actionContainers/ActionContainer.scala` to move `core` package in test module.
How do you think?

Other option is, to change the scope `HttpUtills`.

rabbah:
OK to change the scope. I protected is because it has semantics that are peculiar to the invoker but there's a comment to that effect in the class.

style95:
@rabbah I updated the redo command also.
I have tested with docker-machine also.
It worked as expected.

Kindly check on it.

style95:
I hope this is reviewed before it becomes staled again..

rabbah:
off the latest head of this pr, `wskdev fresh` deployed on mac with no intervention on my part.
i am now running tests locally.

ansible/README.md
```diff
@@ -17,40 +17,55 @@ Nothing to be done, Ansible is already installed during vagrant provisioning.
...
+sudo pip install ansible==2.3.0.0
 ```
-brew install python
-pip install ansible==2.3.0.0
+Docker for Mac does not provide any official ways to meet some requirements for OpenWhisk.
+You need to depends on the workarounds until Docker provides official methods.
 
-cd ansible
-ansible-playbook -i environments/mac setup.yml [-e docker_machine_name=whisk]
+If you prefer [Docker-machine](https://docs.docker.com/machine/) to [Docker for mac](https://docs.docker.com/docker-for-mac/), you can follow instructions in [docker-machine/README.md](../tools/macos/docker-machine/README.md)
```
rabbah: add period at the end of the sentence.

ansible/README.md
```diff
@@ -17,40 +17,55 @@ Nothing to be done, Ansible is already installed during vagrant provisioning.
...
+
+##### Enable docker remote API
+
+During the deployment steps, each component communicates with docker via docker remote API.
```
rabbah: s/docker/Docker/ everywhere except in actual commands.

ansible/README.md
```diff
@@ -17,40 +17,55 @@ Nothing to be done, Ansible is already installed during vagrant provisioning.
...
+Currently however, Docker for Mac does not support such a feature.
+
+There are many workarounds for this.
+One way is to use `socat` command to setup proxy for the unix socket.
```
rabbah: s/unix/UNIX/

ansible/README.md
```diff
@@ -17,40 +17,55 @@ Nothing to be done, Ansible is already installed during vagrant provisioning.
...
 ```
 
-**Hint:** If you omit the optional `-e docker_machine_name` parameter, it will default to "whisk".  
-If your docker-machine VM has a different name you may pass it via the `-e docker_machine_name` parameter.
+If you want to deploy OpenWhisk in distributed environment, you are required to enable docker remote API in all Mac hosts.
 
-After this there should be a `hosts` file in the `ansible/environments/mac` directory.
+##### Activate docker0 network
+This is an optional step for local deployment.
+OpenWhisk Ansible script uses docker0 network interface to deploy OpenWhisk and it does not exist on Docker for Mac environment.
```
rabbah: The OpenWhisk deployment via Ansible uses the `docker0` network interface ...

ansible/README.md
```diff
@@ -17,40 +17,55 @@ Nothing to be done, Ansible is already installed during vagrant provisioning.
...
 
-To verify the hosts file you can do a quick ping to the docker machine:
+An expedient workaround is to add alias for docker0 network to loopback interface.
```
rabbah: `docker0`.

ansible/README.md
```diff
@@ -17,40 +17,55 @@ Nothing to be done, Ansible is already installed during vagrant provisioning.
...
-cd ansible
-ansible all -i environments/mac -m ping
+sudo ifconfig lo0 alias 172.17.0.1/24
 ```
 
-Should result in something like:
+##### Setup proxy container to run unit tests (optional)
+
+This step is only required to run tests with Docker for Mac.
+If you don't run tests, you can just skip this step.
```
rabbah: gasp! perhaps change this to: if you do not run tests _locally_, ...

ansible/README.md
```diff
@@ -244,6 +261,12 @@ An expedient workaround is to create a link to the expected location:
 ln -s $(which python) /usr/local/bin/python
 ```
 
+Alternatively, you can also configure the location of python interpreter in `environments/<environment>/group_vars`.
```
rabbah: s/python/Python/

ansible/environments/docker-machine/group_vars/all
```diff
@@ -0,0 +1,40 @@
+whisk_version_name: mac
```
rabbah: is this the mac file moves to docker-machine (ie just renamed)? I'm surprised git didn't pick it up as a move.
edit: ok i see why.

ansible/environments/mac/group_vars/all
```diff
@@ -4,11 +4,6 @@ whisk_logs_dir: /Users/Shared/wsklogs
 docker_registry: ""
 docker_dns: ""
 
-# The whisk_api_localhost_name is used to configure nginx to permit vanity URLs for web actions.
-# It is also used for the SSL certificate generation. For a local deployment, this is typically
-# a hostname that is resolved on the client, via /etc/hosts for example.
-whisk_api_localhost_name: "openwhisk"
```
rabbah: why is this removed?

ansible/environments/mac/hosts
```diff
@@ -0,0 +1,33 @@
+; the first parameter in a host is the inventory_hostname which has to be
+; either an ip
+; or a resolvable hostname
+
+; used for local actions only
```
rabbah: could we use the "local" (ie Ubuntu) hosts?

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
```diff
@@ -132,5 +132,6 @@ protected[core] class HttpUtils(
     private val connection = HttpClientBuilder
         .create
         .setDefaultRequestConfig(httpconfig)
+        .useSystemProperties()
```
rabbah: why is this needed?

tools/macos/README.md
```diff
@@ -1,18 +1,18 @@
-# How to setup a Mac host for OpenWhisk
+# Setting up OpenWhisk with Docker for Mac
 
-One way to develop or deploy OpenWhisk on a Mac is to use [docker-machine](https://docs.docker.com/machine/install-machine/) which runs the Docker daemon inside a virtual machine accessible from the Mac host.
+OpenWhisk will natively be running on a Mac host with Docker for Mac.
```
rabbah: > OpenWhisk will natively be running on a Mac host with Docker for Mac.

OpenWhisk can run on a Mac host with [Docker for Mac](link to docker for mac).

tools/macos/docker-machine/README.md
```diff
@@ -0,0 +1,156 @@
+# Setting up OpenWhisk with Docker-machine
+
+OpenWhisk will be running on virtual machine in which Docker daemon is running.
```
rabbah: > OpenWhisk will be running on virtual machine in which Docker daemon is running.

OpenWhisk can on a Mac using a virtual machine in which Docker daemon is running.

ansible/environments/mac/group_vars/all
style95: With this pr, `mac` environment is for Docker for Mac, and Ansible can deploy OW on Mac with configurations in `local` environment.
So I copied `group_vars/all` from local environment.

AFAIK, Ansible can not deploy OW without that option in `docker-machine` env, but in `mac` env, it does not.

ansible/environments/mac/hosts
style95: Does `local` mean Ubuntu hosts?
With Docker for Mac, Ansible will deploy OW without any remote connection.
So I thought `ansible_connection` should be `local`.

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
style95: There are few test cases which depends on this utils and it requires proxy configuration.

markusthoemmes:
I thought we agreed on not switching to the play client but rather using the Apache HTTP based client for tests as well to not include another dependency?

style95:
@markusthoemmes
First, sorry for the confusion.
I should talk to you first.

As you proposed, I am trying to use `HttpUtils` instead of `AkkaHttpUtils` in tests.
I found `AkkaHttpUtils` is being used in `ActionContainer.scala`.
In current version, it returns `(status_code, Option[value])`, but in `HttpUtils` it returns `Either[ContainerConnectionError, ContainerResponse]`.
Both of responses does not include status code.

I could update `HttpUtils` to return status code as well.
But I am afraid it is not accepted, because it will affect the core logic.
And if I modify response type to `Either[ContainerConnectionError, ContainerResponse]`, it may break semantic in test code.

For example,

```
it should "abort init with empty response" in {
        val (out, err) = withContainer("badaction") { c =>
            val (code, out) = c.init(initPayload("exit"))
            code shouldBe 500
            out shouldBe empty
        }

        out should include("exit")
        // err stream may not be empty if the proxy did not get a chance to
        // drain the action's out/err streams; skip check on err stream
    }
```

This code, it wants to check whether it returns 500 response in case of `badaciton` container.
In some cases, response should be 502.
There are more cases.

```
it should "report error if zip-encoded action does not include required file" in {
        val srcs = Seq(
            Seq("echo.py") -> """
                |def echo(args):
                |  return { "echo": args }
            """.stripMargin)

        val code = ZipBuilder.mkBase64Zip(srcs)

        val (out, err) = withActionContainer() { c =>
            val (initCode, initRes) = c.init(initPayload(code, main = "echo"))
            initCode should be(502)
        }

        checkStreams(out, err, {
            case (o, e) =>
                o shouldBe empty
                e should include("Zip file does not include")
        })
    }
```

So if I change the response type, and omit status code, it may change semantic of many tests.

I discussed with @rabbah on Slack.
Since expected code changes are not that simple and small, and this thread is too long already, it would be better to open a separate pr iff you approve it.

style95:
It is rebased from the latest master again, and all the conflicts have been resolved.

markusthoemmes:
Sounds good to me, thanks for clarification :+1:

ansible/environments/mac/hosts
rabbah: Right - this file is the same as https://github.com/apache/incubator-openwhisk/blob/master/ansible/environments/local/hosts

rabbah:
We will try to further improve this in the future to use `local` for docker for mac, and further strip the play dependence. Thanks for being patient and working this out!

PR #1792, opened 2017-02-01
Fix revision preserving behaviour of WhiskAction

markusthoemmes:
Constructing new objects results in a loss of the DocumentRevision and thus breaks caching in the invoker.

### DIFF
--- a/tests/src/whisk/core/entity/test/WhiskEntityTests.scala
+++ b/tests/src/whisk/core/entity/test/WhiskEntityTests.scala
@@ -1,0 +1,162 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.entity.test
+
+import org.junit.runner.RunWith
+import org.scalatest.FlatSpec
+import org.scalatest.Matchers
+import org.scalatest.junit.JUnitRunner
+import whisk.core.entity.EntityPath
+import whisk.core.entity.EntityName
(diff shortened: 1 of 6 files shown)


tests/src/whisk/core/entity/test/WhiskActionTests.scala
```diff
@@ -0,0 +1,54 @@
...
+
+    behavior of "WhiskAction"
+
+    it should "preserve its revision through inheritance of parameters" in {
+        val action = WhiskAction(
+            EntityPath("testspace"),
+            EntityName("testname"),
+            Exec.js("js1")).revision[WhiskAction](DocRevision("test"))
+        val inherited = action.inherit(Parameters())
+        action.rev shouldBe inherited.rev
```
rabbah: `action shouldBe inherited` in this case since inheriting empty parameter set.

rabbah:
should fix inherit in `WhiskPackage` also.

tests/src/whisk/core/entity/test/WhiskActionTests.scala
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    it should "preserve its revision through resolving namespaces" in {
+        val sequenceAction = FullyQualifiedEntityName(EntityPath("_"), EntityName("testaction"))
+        val action = WhiskAction(
+            EntityPath("testspace"),
+            EntityName("testname"),
+            Exec.sequence(Vector(sequenceAction))).revision[WhiskAction](DocRevision("test"))
+        val resolved = action.resolve(EntityName("testuser"))
```
rabbah: since these two tests are effectively unit tests for the resolve and inherit method, let's make the test tighter (_ should be replaced with testuser), rest of the action the same.

tests/src/whisk/core/entity/test/WhiskEntityTests.scala
```diff
@@ -0,0 +1,73 @@
...
+            EntityPath("testspace"),
+            EntityName("testname"),
+            Exec.sequence(Vector(sequenceAction))).revision[WhiskAction](DocRevision("test"))
+
+        val resolved = action.resolve(EntityName(user))
+        resolved.exec.asInstanceOf[SequenceExec].components.head shouldBe sequenceAction.copy(path = EntityPath(user))
+        action.rev shouldBe resolved.rev
+    }
+
+    behavior of "WhiskPackage"
```
rabbah: ??

markusthoemmes: Should've added wip label.

markusthoemmes:
PG1 1130

PR #1795, opened 2017-02-01
Increase timeout value for entitlement tests

jasonpet:
Various entitlement tests often throw java.util.concurrent.TimeoutException: Futures timed out after [1 second].  Increasing the timeout value.


jasonpet:
running pg2#1001

rabbah:
make it 10.seconds. no pg needed again.

jasonpet:
@rabbah - sorry just noticed your above comment.  Updated to 10 seconds.

PR #1799, opened 2017-02-01
remove dead code

### LINKED ISSUE #1796: Crypt not implementing proper IV
`Crypt` is using AES in CBC mode with a fixed IV. CBC mode requires a 128-bit **random** IV for every use of the cipher. This ensures that the first block of the cipher is properly secured and changes the outcome of every subsequent block so that encrypting the same plaintext twice with the same key will result in different ciphertexts. 

[Code Link](https://github.com/openwhisk/openwhisk/blob/master/common/scala/src/main/scala/whisk/common/Crypt.scala#L43)

Additionally, proper key lengths of 128, 192, & 256-bit should probably be enforced.

rabbah:
Closes #1796.
PG1/1128.


csantanapr:
merged 32b8421e2e2a38157e0b303fb3385dd53024a898

Issue #1793, opened 2017-02-01
feature request: "catch", needed for 401/403 handling in web-exported actions


starpit:
Say I have a web-exported sequence `S: A=>B=>C` and action `B` is responsible for some kind of authentication or authorization. Thus, `B` may encounter a 403 or 401 situation, and, if so, needs to short-circuit the execution of the sequence.

Unfortunately, short-circuiting of sequences can only be done by `reject()`'ing in action `B`. Today, this results in the web-exporter communicating a 400 to the client: no opportunity for clearing server-side session cookies, or redirecting the client to a login page.

More broadly, I suggest that `B` should not know about web-export. It *should* simply reject the request. It is the entire sequence `S` that has been exported, so the notion of web-oriented error handling should be encapsulated at the level of `S`, not `B`.

We could accomplish this by supporting a `catch` operator. For example: 
```
wsk package bind Http BluemixHttp -p redirect_uri https://bluemix.net/login
wsk action update S A,B,C --annotation catch BluemixHttp/403
```

i.e. when the sequence short-circuits, invoke the action `Http/403`. Here, we have the opportunity to provide a general package `Http` that implements common catch patterns.

starpit:
an open question here is at what level is the error code matching done? within the catch action? or by the controller? there is something attractive about having separate 403, 500, etc. catch handler actions.

markusthoemmes:
Alternative proposal: The Controller tries to evaluate an `error` response as well, much like we do for actual responses today, to alter the statusCode returned?

starpit:
it's still the case that there is custom logic that needs to be invoked. e.g. which cookies to clear? where to redirect the user?

i agree that ew can either put the "match" in the controller on in the catch action. but, in either case, some custom logic needs to be invoked?

rabbah:
> The Controller tries to evaluate an error response as well, much like we do for actual responses today, to alter the statusCode returned?

That will break separation of concerns - web extensions should not permeate actions which are not web-exported and hence do not understand the extensions.

sjfink:
alternative proposal:

high-level sketch here, low-level details to be discussed.

We create a built-in system feed in the whisk.system package, called 'exception'.   Define the set of abnormal exit conditions to be, for example, E := { 'error', 'abort' }.     You can instantiate the exception feed for action A and exit condition C, which causes the system feed to fire a trigger when A exits with condition C.

So to install an exception handler the user creates a rule to handle the exception feed.

[...]

PR #1801, opened 2017-02-02
Handle application errors in web actions.

rabbah:
If the activaiton result is an activation error, treat it like a successful
activation with one distinction: rather than projecting the result per the
path specified in the URL, project only the error field. The resulting JSON
value must be of the expected type otherwise an appropriate error is reported.

This change will allow an http response to set it own status code on error
for example, but also means that the action producing the error response must
be aware of the extension type.

The extension type is not passed to the action - perhaps it should be. Even so,
in the case of a sequence, one would need to forward this type to individual
components to handle the error case (or result) correctly.

@starpit FYI. This adopt's @markusthoemmes' suggestion but does not solve the bigger issue discussed in #1793.

PG2/1006.

### DIFF
--- a/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
+++ b/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
@@ -52,10 +52,10 @@ object Messages {
     val resourceDoesNotExist = "The requested resource does not exist."
 
     /** Standard message for too many activation requests within a rolling time window. */
-    val tooManyRequests = "Too many requests from user in a given amount of time."
+    val tooManyRequests = "Too many requests in a given amount of time for namespace."
 
     /** Standard message for too many concurrent activation requests within a time window. */
-    val tooManyConcurrentRequests = "The user has sent too many concurrent requests."
+    val tooManyConcurrentRequests = "Too many concurrent requests in flight for namespace."
 
     /** Standard message when supplied authkey is not authorized for an operation. */
     val notAuthorizedtoOperateOnResource = "The supplied authentication is not authorized to access this resource."
@@ -112,6 +112,7 @@ object Messages {
     val responseNotReady = "Response not yet ready."
     val httpUnknownContentType = "Response did not specify a known content-type."
     val httpContentTypeError = "Response type in header did not match generated content type."
+    val errorProcessingRequest = "There was an error processing your request."
 
     def invalidInitResponse(actualResponse: String) = {
         "The action failed during initialization" + {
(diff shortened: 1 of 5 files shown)


rabbah:
PG2/1006 🔵.

common/scala/src/main/scala/whisk/http/ErrorResponse.scala
```diff
@@ -52,10 +52,10 @@ object Messages {
     val resourceDoesNotExist = "The requested resource does not exist."
 
     /** Standard message for too many activation requests within a rolling time window. */
-    val tooManyRequests = "Too many requests from user in a given amount of time."
+    val tooManyRequests = "Too many requests in a given amount of time."
 
     /** Standard message for too many concurrent activation requests within a time window. */
-    val tooManyConcurrentRequests = "The user has sent too many concurrent requests."
+    val tooManyConcurrentRequests = "Too many concurrent requests in flight."
```
rabbah:
```
    val tooManyRequests = "Too many requests in a given amount of time for resource."
    val tooManyConcurrentRequests = "Too many concurrent requests in flight for resource."
```
?

rabbah: reworded- thoughts?

common/scala/src/main/scala/whisk/http/ErrorResponse.scala
```diff
@@ -52,10 +52,10 @@ object Messages {
     val resourceDoesNotExist = "The requested resource does not exist."
 
     /** Standard message for too many activation requests within a rolling time window. */
-    val tooManyRequests = "Too many requests from user in a given amount of time."
+    val tooManyRequests = "Too many requests in a given amount of time for namespace."
 
     /** Standard message for too many concurrent activation requests within a time window. */
-    val tooManyConcurrentRequests = "The user has sent too many concurrent requests."
+    val tooManyConcurrentRequests = "Too many concurrent requests in flight for namespace."
```
cbickel: 👍

common/scala/src/main/scala/whisk/http/ErrorResponse.scala
```diff
@@ -106,8 +106,8 @@ object Messages {
     /** Error for meta api. */
     val propertyNotFound = "Response does not include requested property."
     def invalidMedia(m: MediaType) = s"Response is not valid '${m.value}'."
-    val contentTypeNotSupported = "Content type not supported."
-    val contentTypeRequired = "Content type is missing."
+    val contentTypeNotSupported = """Content type must be specified and one of ["json", "html", "http", "text"]."""
```
markusthoemmes: Should we parametrize this and pass `mediaTranscoders.keySet`?

rabbah: This is a packaging issue - the messages are in the common project but the set is in the controller project. As you suggested, we should split these messages and localize them to their respective services (controller vs invoker). Leaving as is until the messages are refactored.

rabbah:
Rebased to pick up logger changes.

Issue #1489, opened 2016-11-10
CLI should not encode API response schema


rabbah:
I've experience several instances where a schema change in the controller must be reflected in the CLI to properly render the expected output on the command line because the current CLI implementation encodes the types too strictly and *unnecessarily* - in every case I've encountered so far where I've made a schema change to add or remove a field, the CLI would have been just fine if it treated the response from the controller opaquely.

We should refactor the API respond handling in the CLI so that it uses more generic JSON object interface rather than strict types that just get in the way.

I've also found several instances where the CLI is not in sync with the controller (Evidenced by several defects I've opened for these in recent days) and the CLI injects properties that should not exist (and hence confusing).

ioana-blue:
I was bit by this as well and it's not fun. In fact, if I hadn't checked with RR, I would have wasted a lot of time on chasing a "bug" in my code, while it was really a schema change for activations which is not reflected in the cli.

dubeejw:
@ioana-blue, see https://github.com/openwhisk/openwhisk/pull/1566 for the PR that will address your issue.

csantanapr:
Issue moved to [apache/incubator-openwhisk-cli #148](https://github.com/apache/incubator-openwhisk-cli/issues/148) via [**ZenHub**](https://www.zenhub.com/)

PR #1802, opened 2017-02-02
Remove Decoding of Scheme from Action Invocations

dubeejw:
- No longer decode action invocation response into structs

Closes https://github.com/openwhisk/openwhisk/issues/1489

### DIFF
--- a/tests/src/common/WskTestHelpers.scala
+++ b/tests/src/common/WskTestHelpers.scala
@@ -123,8 +123,8 @@ trait WskTestHelpers extends Matchers {
         activationId: String,
         logs: Option[List[String]],
         response: CliActivationResponse,
-        start: Long,
-        end: Long,
+        start: Instant,
+        end: Instant,
         duration: Long,
         cause: Option[String],
         annotations: Option[List[JsObject]]) {
@@ -141,6 +141,17 @@ trait WskTestHelpers extends Matchers {
     }
 
     object CliActivation extends DefaultJsonProtocol {
+        private implicit val instantSerdes = new RootJsonFormat[Instant] {
+            def write(t: Instant) = t.toEpochMilli.toJson
+
+            def read(value: JsValue) = Try {
+                value match {
+                    case JsNumber(i) => Instant.ofEpochMilli(i.bigDecimal.longValue)
+                    case _           => deserializationError("timetsamp malformed")
+                }
+            } getOrElse deserializationError("timetsamp malformed 2")
+        }
+
         implicit val serdes = jsonFormat8(CliActivation.apply)
     }
 
--- a/tests/src/system/basic/WskBasicTests.scala
+++ b/tests/src/system/basic/WskBasicTests.scala
@@ -487,16 +487,16 @@ class WskBasicTests
             withActivation(wsk.activation, run) {
                 activation =>
                     activation.response.result shouldBe Some(dynamicParams.toJson)
-                    activation.duration shouldBe 0L // shouldn't exist but CLI generates it
-                    activation.end shouldBe Instant.EPOCH.toEpochMilli // shouldn't exist but CLI generates it
+                    activation.duration shouldBe 0L         // shouldn't exist but CLI generates it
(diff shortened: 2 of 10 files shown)


dubeejw:
PG3 255

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -477,15 +477,15 @@ class WskBasicTests
                 activation =>
                     activation.response.result shouldBe Some(dynamicParams.toJson)
                     activation.duration shouldBe 0L // shouldn't exist but CLI generates it
-                    activation.end shouldBe Instant.EPOCH.toEpochMilli // shouldn't exist but CLI generates it
+                    activation.end shouldBe Instant.ofEpochMilli(0L) // shouldn't exist but CLI generates it
```
rabbah: why dos the comment say shouldn't exist but CLI generates it - should the CLI stop generating it when this schema stripping is done?

dubeejw: Yes, the CLI will stop automatically producing these fields when the schema is removed. Removal of the schema decoding will have to be done for each command and subcommand. For instance, this PR currently removes the decoding for action invocations only. In order to remove mentioned generated fields, trigger fire would have to be updated.

markusthoemmes: Also: Shouldn't these two lines produce an identical result?

tests/src/common/WskTestHelpers.scala
```diff
@@ -141,6 +141,18 @@ trait WskTestHelpers extends Matchers {
     }
 
     object CliActivation extends DefaultJsonProtocol {
+        private implicit val instantSerdes = new RootJsonFormat[Instant] {
+            def write(t: Instant) = t.toEpochMilli.toJson
+
+            def read(value: JsValue) = Try {
+                value match {
+                    case JsString(t) => Instant.parse(t)
```
markusthoemmes: While this is nice to have, our code should never return a `JsString` here, so we should drop the serdes there?

tests/src/system/basic/WskBasicTests.scala
dubeejw: Instant.EPOCH.toEpochMilli == 0
Instant.ofEpochMilli(0L) == 1970-01-01T00:00:00Z

markusthoemmes: Okay, but `Instant.EPOCH` should be equal to `ofEpochMilli(0L)`?

dubeejw: That is correct! I'll use `Instant.EPOCH` instead.

dubeejw:
PG1 1165.

dubeejw:
@rabbah, every sub-command will have to have the same updates that `action invoke` have here. I'd suggest to merge each updated sub-command individual as it will take some time to update all of them.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -487,16 +487,16 @@ class WskBasicTests
             withActivation(wsk.activation, run) {
                 activation =>
                     activation.response.result shouldBe Some(dynamicParams.toJson)
-                    activation.duration shouldBe 0L // shouldn't exist but CLI generates it
-                    activation.end shouldBe Instant.EPOCH.toEpochMilli // shouldn't exist but CLI generates it
+                    activation.duration shouldBe 0L         // shouldn't exist but CLI generates it
+                    activation.end shouldBe Instant.EPOCH   // shouldn't exist but CLI generates it
```
rabbah: did i ask this already? the comment says "shouldn't exist" but CLI generates it - is this not being removed/addressed by this pr?

rabbah: i did ask already and ou answered.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/a600110aa1511f08871618f45745e5b5c0cdcea4.

PR #1806, opened 2017-02-03
update webhook for travis whiskbot

csantanapr:


csantanapr:
@rabbah took care of this with another PR

PR #1809, opened 2017-02-03
Make LoadBalancer's round-robin parameter configurable

RSulzmann:
The parameter in the load-balancer that decides when to dispatch to the next invoker is hard-coded. We need to make the parameter configurable per environment by adding a property to WhiskConfig.

Property Name: loadbalancer_activation_count_before_next_invoker
Default Value: 10

### DIFF
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
@@ -47,7 +47,7 @@ import whisk.common.TransactionId
 import whisk.connector.kafka.KafkaConsumerConnector
 import whisk.connector.kafka.KafkaProducerConnector
 import whisk.core.WhiskConfig
-import whisk.core.WhiskConfig.{ consulServer, kafkaHost }
+import whisk.core.WhiskConfig.{ consulServer, kafkaHost, loadbalancerActivationCountBeforeNextInvoker }
 import whisk.core.connector.{ ActivationMessage, CompletionMessage }
 import whisk.core.entity.{ ActivationId, CodeExec, WhiskAction, WhiskActivation }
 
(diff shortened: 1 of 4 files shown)


core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -95,7 +95,8 @@ class LoadBalancerService(
     info(this, s"blackboxFraction = $blackboxFraction")
 
     /** We run this often on an invoker before going onto the next. */
-    private val activationCountBeforeNextInvoker = 10
+    private val activationCountBeforeNextInvoker = Math.max(1, config.loadbalancerActivationCountBeforeNextInvoker.toInt)
```
rabbah: @markusthoemmes @perryibm we've done this in a few places because the WhiskConfig is not type safe - I'm wondering though why we don't move `toInt` into the getter as a first step?

(@RSulzmann you may ignore my comment for this pr)

rabbah:
@RSulzmann please update the commit message.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
RSulzmann: @rabbah 
I can extract the property as a numeric value, the same way it is done for the property controllerBlackboxFraction :

val controllerBlackboxFraction = this.getAsDouble(WhiskConfig.controllerBlackboxFraction, 0.10)

I've already tested this way on my local machine.

The good thing with this approach is, that parsing errors come up at startup time, and not when the property is used at some point in runtime.

rabbah: yup that was my thinking as well - there are some more of these `toInts` that we should get rid of too elsewhere (noting here for the record: rate throttler and sequence limit come to mind).

RSulzmann:
change: parse the property as Int in WhiskConfig.

There are two options to handle parsing errors:
- use a default             getAsInt(key: String, defaultValue: Int)
- raise an exception    getAsInt(key: String)

I decided to raise an exception on parsing errors (e.g. invalid number format)

(We have a default value set in whisk.properties.j2 template, when the parameter is not specified in one environment)

jeremiaswerner:
PG1#1157
PG1#1158

jeremiaswerner:
rebased and merged in 47edf410a14b3fdda2a3612fad0ce0b33c30ff79

PR #1733, opened 2017-01-17
Additional Tests for api gw, large swagger file, and invalid swagger file

nwspeete-ibm:
- large swagger file
   - multiple path/endpoints
- invalid swagger file

PR #1818, opened 2017-02-07
API GW additional tests

mdeuser:
- large swagger file
  - multiple path/endpoints
- invalid swagger file

Supersedes PR #1733

### DIFF
--- a/tests/src/common/WhiskProperties.java
+++ b/tests/src/common/WhiskProperties.java
@@ -236,7 +236,11 @@ public class WhiskProperties {
     }
 
     public static int getMaxActionInvokesPerMinute() {
-        return Integer.parseInt(whiskProperties.getProperty("limits.actions.invokes.perMinute"));
+        String valStr = whiskProperties.getProperty("limits.actions.invokes.perMinute");
+        if (null == valStr) {
+            valStr = whiskProperties.getProperty("defaultLimits.actions.invokes.perMinute");
+        }
+        return Integer.parseInt(valStr);
     }
 
     /**
--- a/tests/src/common/WhiskProperties.java
+++ b/tests/src/common/WhiskProperties.java
@@ -235,6 +235,10 @@ public class WhiskProperties {
         return Integer.parseInt(whiskProperties.getProperty("controller.host.port"));
     }
 
+    public static int getMaxActionInvokesPerMinute() {
+        return Integer.parseInt(whiskProperties.getProperty("limits.actions.invokes.perMinute"));
+    }
+
     /**
      * read the contents of auth key file and return as a Pair
      * <username,password>
--- a/tests/src/whisk/core/cli/test/ApiGwTests.scala
+++ b/tests/src/whisk/core/cli/test/ApiGwTests.scala
@@ -80,6 +80,7 @@ class ApiGwTests
       }
       finally {
         val deleteresult = wsk.api.delete(basepathOrApiName = testbasepath)
+        Thread.sleep(5000)  // Test suite will overrun the local dev throttling limit
       }
     }
 
@@ -105,6 +106,7 @@ class ApiGwTests
         }
         finally {
             val deleteresult = wsk.api.delete(basepathOrApiName = testbasepath, expectedExitCode = DONTCARE_EXIT)
+            Thread.sleep(5000)  // Test suite will overrun the local dev throttling limit
         }
     }
 
@@ -125,6 +127,7 @@ class ApiGwTests
         }
         finally {
             val deleteresult = wsk.api.delete(basepathOrApiName = testbasepath, expectedExitCode = DONTCARE_EXIT)
+            Thread.sleep(5000)  // Test suite will overrun the local dev throttling limit
         }
     }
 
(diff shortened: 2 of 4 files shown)


mdeuser:
PG2 1023

mdeuser:
Need to close/reopen pr to re-trigger travis build - directly rebuilding from travis is not working at the moment.

mdeuser:
PG3 246

tests/src/whisk/core/cli/test/ApiGwTests.scala
```diff
@@ -80,6 +80,7 @@ class ApiGwTests
       }
       finally {
         val deleteresult = wsk.api.delete(basepathOrApiName = testbasepath)
+        Thread.sleep(5000)  // Test suite will overrun the local dev throttling limit
```
csantanapr: @rabbah @markusthoemmes What you think about this approach to sleep, to avoid hitting the 60 api calls per minute limit in tests?

markusthoemmes: It will break eventually. I think it's too brittle of an approach. Should we rather exempt `whisk.system` from throttles?

csantanapr: Yes that's another alternative to excempt whisk.system during testing, or your saying I general ?
I think having limits on whisk.system is also a protection we have in place, maybe increasing the limit on some priviledge namespaces would be a way to do if you are referring in general and not test run. 

There is no way to eliminate or set higher  limits and then in the specific throttle tests set the limits before running the tests dynamically and then when test is finish remove the limits?  

This way no other tests are affected by this lower limit set by default

rabbah:
👎 To adding exemption. This tests creates so many assets/invokes 100 actions?
If you need to settle the throttle do it outside the suite as we do with other tests.

csantanapr:
@rabbah Yep, I think someone mentioned verbally that the we already do that on adjusting the limits for the test suite to allow more requests per minute. 

Can you point to a file line number example so we can follow the same?

rabbah:
Looks in the Jenkins scripts for "settle throttle". But I'm still wondering why you're running into this limit with this suite. Are you running a 100 tests per minute?

csantanapr:
>But I'm still wondering why you're running into this limit with this suite. Are you running a 100 tests per minute?

Yes the limit is 60 per minute, so when it gets to 61 requests within the minute it get's the 429 status code

markusthoemmes:
You can also take a look at the ThrottleTests, they settle the limits quite "intelligently".

mdeuser:
@markusthoemmes - In the ThrottleTests test suite, the actual throttle limit is not dynamically increased; but that may not be what you are referring to.

mdeuser:
@csantanapr - test suite now implements a more intelligent throttle avoidance.  PG3 268

mdeuser:
Travis not responding to build restart, so closing/reopening PR to restart travis

mdeuser:
PG3 269

mdeuser:
@csantanapr - ready for re-review, squash and merge.

csantanapr:
@markusthoemmes @rabbah  is this good now to merge?

tests/src/common/WhiskProperties.java
```diff
@@ -235,6 +235,10 @@ public static int getControllerPort() {
         return Integer.parseInt(whiskProperties.getProperty("controller.host.port"));
     }
 
+    public static int getMaxActionInvokesPerMinute() {
+        return Integer.parseInt(whiskProperties.getProperty("limits.actions.invokes.perMinute"));
```
rabbah: you can get this from `WhiskConfig` also - making this method unnecessary.

rabbah: This may not work for all deployments.
See: https://github.com/openwhisk/openwhisk/blob/master/tests/src/limits/ThrottleTests.scala#L64

rabbah:
I approved but added a comment - I think the throttle/limit may not be correct for all deployments.

tests/src/common/WhiskProperties.java
csantanapr: @mdeuser could address concern about required key, and optional key.
The key `limits.actions.invokes.perMinute` might not be present, it's optional

```
val maximumInvokesPerMinute = getLimit("defaultLimits.actions.invokes.perMinute", "limits.actions.invokes.perMinute")


    println(s"maximumInvokesPerMinute  = $maximumInvokesPerMinute")

    /*
     * Retrieve a numeric limit for the key from the property set.  If the overrideKey is present, use that.
     */
    def getLimit(key: String, overrideKey: String) = Try {
        WhiskProperties.getProperty(overrideKey).toInt
    } getOrElse {
        WhiskProperties.getProperty(key).toInt
    }
```

mdeuser:
ah.. I see.  Added code to account for both values - a primary value and a default/fallback value.
PG3 283

csantanapr:
@rabbah any comments before I merge this in?

rabbah:
merge away.

csantanapr:
merged a51bbe17fe2b34d37332381d4365201e7d93c836

Issue #1822, opened 2017-02-08
sequence invoked with "error" parameter automatically fails


starpit:
when i pass an "error" parameter to a sequence, it fails with that parameter as a return value. note that the sequence fails, without even invoking any of the components. also note that this only affects sequences, not individual actions.

this breaks slack oauth flows: when the user clicks Cancel in an oauth flow, slack calls my redirect_uri with an `error` parameter. but the issue seems to be unrelated to web-export. the undesirable behavior occurs even with conventional invocations.

here is an example:

```
echo 'function main(params) { return { code: 302, headers: { location: "https://google.com" }}}' > /tmp/foo.js

wsk action update foo /tmp/foo.js
wsk action update --sequence fooseq foo,foo

wsk action invoke fooseq -p error weird -br
{
    "error": "weird"
}
wsk action invoke fooseq -p error weird -b 2>&1 | grep status
        "status": "application error",
        "statusCode": 0,


wsk action invoke fooseq -p error_nope weird -br
{
    "code": 302,
    "headers": {
        "location": "https://google.com"
    }
}```

sjfink:
So we currently have a convention, obviously, that any action that returns a result with an "error" attribute is considered an error.

Seems clear that reserving the keyword "error" is a mistake.   Easy fix: use something like __OPENWHISK_ABNORMAL_EXIT instead.

starpit:
the problem here is that none of my actions were even invoked... the sequence just prematurely fails without even invoking any of the components

rabbah:
Not to mention the action isn't returning an error property.

rabbah:
Can I bet of the fold?

rabbah:
@ioana-blue assigned to you speculatively.

sjfink:
ok ... so the check to abort a sequence based on "error" arguably should not apply to the first method in the outermost sequence .... although I could argue the other way and claim this is working as designed.

I still propose changing the keyword as a better fix.

rabbah:
Should not apply to the incoming invoke parameters - they are not the result of an action. Why do we need to change the convention to abort the sequence? This strikes me as completely orthogonal.

sjfink:
it's orthogonal, ok -- but the current convention sucks so let's fix it as long as we're touching that code?

rabbah:
Well if we change it it seems quite a change. Now actions might return error or some other property if they are stand alone vs composed in a sequence?

rabbah:
[...]

PR #1831, opened 2017-02-08
Treat error field differently if present in the initial payload to a sequence

ioana-blue:
Update test accordingly.

This is a fix for #1822 

Note that I left the scanLeft unchanged. I'd rather do the "optimization" refactoring in a different pass since it's  a lot more involved than current change. I'm guessing we'd like to promote this change soonish.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
@@ -300,41 +300,46 @@ protected[actions] trait SequenceActions {
         val fakeEnd = Instant.now()
         val fakeResponse = ActivationResponse.payloadPlaceholder(payload)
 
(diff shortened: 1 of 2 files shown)


ioana-blue:
PG2 1033

starpit:
i still think its dangerous ever to look at inputs. these are out of control of the action for several reasons. you are protecting against one, now -- data from the outside world.

the other potential source of inputs is package bindings. wouldn't it be odd behavior if, on day N, the team adds a package parameter named `error`, and their app suddenly stops working?

ioana-blue:
@starpit not sure I understand your comment above. IIRC package bindings are merged in the input params of the actions. After this is PR is merged in, the input parameters are *not* filtered for error, so nothing would break.

rabbah:
I'll accept this patch but this change highlights how the sequence implementation should be done with a state machine/akka actor. This will have the added benefit of reducing the state held in flight to just the "last" activation result.

ioana-blue:
PG3 419

PR #1835, opened 2017-02-09
Move markers to more appropriate places

markusthoemmes:
This should get rid of marker confusion when running sequences. @RSulzmann please confirm.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
@@ -22,14 +22,11 @@ import scala.concurrent.Promise
 import scala.concurrent.TimeoutException
 import scala.concurrent.duration._
 import scala.util.Failure
-import scala.util.Success
 
 import akka.actor.ActorSystem
-import spray.http.StatusCodes._
 import spray.json._
 import whisk.common.Logging
 import whisk.common.LoggingMarkers
-import whisk.common.StartMarker
 import whisk.common.TransactionId
 import whisk.core.connector.ActivationMessage
 import whisk.core.controller.WhiskServices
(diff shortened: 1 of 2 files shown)


rabbah:
what's the goal for this marker? measure total time for an activation (so for a sequence, the totality of all activations within)?

RSulzmann:
Calculating performance data for sequence actions has two problems:

1. The start and finish of markers CONTROLLER_ACTIVATION (CONTROLLER_ACTIVATION_BLOCKING) were not in sync

The above changes should fix this.

2. Because all activations inside a sequence have the same Tid, the perftool cannot relate the different marker entries to a single activation and calculate correct durations.

do we plan to adress this also ?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -152,19 +150,12 @@ protected[actions] trait PrimitiveActions {
...
-            case Failure(t) =>
-                // short circuits db polling
-                promise.tryFailure(t)
-        }
-
-        response // will either complete with activation or fail with timeout
+        // Will either complete with activation or timeout
+        promise.future
+            .withTimeout(totalWaitTime, new BlockingInvokeTimeout(activationId))
+            .andThen {
```
rabbah: this could just be `recover` no?

markusthoemmes: Nope. `tryFailure` returns a `Boolean` which'll alter the return-type. `andThen` just passes the original future through and only applies the sideeffect.

markusthoemmes:
@rabbah any more concerns?

rabbah:
I approved the changes.

Issue #1833, opened 2017-02-09
wsk activation list does not align action names nicely


rabbah:
```
> wsk activation list
activations
12313131312                 short
23442342344 foobar2baz.xyz.blahblah.longname
```
it would be nice if the action names are all left aligned in the second column.

rabbah:
Fixed by https://github.com/openwhisk/openwhisk/commit/d127c8f5153c602d9c1e069682e513f677959be5.

Issue #1570, opened 2016-11-30
store untrusted certificate setting for CLI in wskprops


rabbah:
To use `wsk` with a local deployment that does not have a properly signed certificate requires using `-i` on every command. So one workaround is to alias `wsk` to `wsk -i`. This setting however should be stored in the wskrpops file so that one can use multiple files and not give up SSL sanity when using a deployment with a valid SSL.

rabbah:
@csantanapr this is the issue for the pr #1838 - let's have the discussion here instead of the pr?

csantanapr:
> @csantanapr this is the issue for the pr #1838 - let's have the discussion here instead of the pr?

Sure

How is the PR #1838 addressing the problem, and handling concerns with user forgetting if it modifies the ~/.wskprops file to set insecure, and the exposure risk?

rabbah:
We currently provide a global `-i` flag for the CLI. We use this flag for all tests and local development because we don't have a trusted certificate installed for local deployment. We can entertain this idea with `letsencrypt` but I'm not sure it's worth the effort (plus they require renewal every 90 days, so more processes).

So that means for local development either every CLI command you enter must have `-i` or you have to create an alias `wsk=wsk -i`.

What the PR does is allow up to configure the CLI and store the setting in `.wskprops`.

Your concern is that one might use the insecure ssl setting with a trusted domain and that's undesirable. But it's no worse that the alias.

This is primarily for local developers and those who have more than one deployment that they work with. I tend to switch between deployments via `WSK_CONFIG_FILE` changes which point to different `.wskprops`. So each would have it's own setting, including insecure ssl (which is off by default).

So what to do about the case where someone sets `-i` in their property file and forgets and now is using the same property file with a trusted domain? Generating a warning on every CLI command is a non-starter. So if we wanted to be cautious about it, one thought is for the CLI to issue a warning when changing `apihost` or `authkey`. I tend to think it's not necessary  - as I said this is for local development - and you have to opt into it.

csantanapr:
why not entertain something with less effort and just expose http on port 80  in open for local dev from nginx?

Would look something like
```
wsk property set --apihost http://192.168.33.13
```
The cli will connect over http for dev
Then if user wants to use a sfure apihost then setting api host with https
```
wsk property set --apihost https://openwhisk.ng.bluemix.net
```

In this use case no `-i` flag to use, no config file to configure.

csantanapr:
[...]

PR #1838, opened 2017-02-09
Store insecure opt in wskprops

rabbah:


rabbah:
@dubeejw in the second commit, i fixed and enabled an ignored test. 1) do we need that test? and 2) do you recall why it was ignored?

csantanapr:
@rabbah you aware of the risk with this change, opens the door for an user to forget that it configured the ~/.wskprops to allow insecure connections, and then use the file when it switches to interact with a production system with secure properly configured.?

Just want to hear if you did something to help the user in the case, like printing annoying message when using insecure, or other means.

We can merge this feature with the risk in place, just want to be sure you are OK with the risk.

rabbah:
What's the alternative? An alias to `wsk -i`? That's worse. I have several config files. I switch between them regularly using `WSK_CONFIG_FILE`. So for each I'd make sure they have appropriate settings. It's easier to switch configurations this way then to have one file and change keys etc. in my experience.

Printing a warning will be annoying - you're going to print one per command? No thanks :)

I'm open to alternatives - but the problem is that `wsk -i` is the default otherwise. And there are other tools like `npm wab` and `npm ownen` that needs to know secure vs insecure and having one work around per tool is silly when they use `.wskprops` otherwise.

csantanapr:
I guess I don't have alternatives, I was asking what does the PR #1833 does ?
And that there are no other possible alternatives and just the PR and a call it a day.

rabbah:
Adding blocked tagged pending resolution (see #1570 for discussion).

tests/src/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -92,16 +92,17 @@ class WskBasicUsageTests
...
             val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
-            val stdout = wsk.cli(Seq("property", "set", "-i", "--apihost", wskprops.apihost, "--auth", wskprops.authKey,
-                "--namespace", namespace), env = env).stdout
+            val stdout = wsk.cli(Seq("property", "set",
+                "--insecure", "true",
+                "--apihost", wskprops.apihost,
+                "--auth", wskprops.authKey), env = env).stdout
             stdout should include(s"ok: whisk auth set to ${wskprops.authKey}")
             stdout should include(s"ok: whisk API host set to ${wskprops.apihost}")
-            stdout should include(s"ok: whisk namespace set to ${namespace}")
```
dubeejw: Need to test that stdout has a message informing the user that the insecure property has been set.

tools/cli/go-whisk-cli/commands/property.go
```diff
@@ -87,6 +90,14 @@ var propertySetCmd = &cobra.Command{
                     map[string]interface{}{"ok": color.GreenString("ok:"), "auth": boldString(auth)}))
         }
 
+        if insecure := flags.property.insecureSet; len(insecure) > 0 {
+            client.Config.Insecure = strings.ToLower(insecure) == "yes" || strings.ToLower(insecure) == "true"
```
dubeejw: Do we always want to set the insecure property to false if values of `yes` or `true` are not passed?

tools/cli/go-whisk-cli/commands/util.go
```diff
@@ -330,7 +330,7 @@ func printNamespaceList(namespaces []whisk.Namespace) {
 func printActivationList(activations []whisk.Activation) {
     fmt.Fprintf(color.Output, "%s\n", boldString("activations"))
     for _, activation := range activations {
-        fmt.Printf("%s %20s\n", activation.ActivationID, activation.Name)
+        fmt.Printf("%s %-20s\n", activation.ActivationID, activation.Name)
```
dubeejw: Is this related to a different issue?

csantanapr:
@rabbah removing blocked label, 
My questions about the implementation details were answered and I don't have any more comments/concerns.
Let's get this merge if code review looks good by @dubeejw

rabbah:
i should say that i've switched to using the http api host as you suggested (locally) and it is masking my issues with the insecure ssl and switching between configurations. i think it's worth reconsidering this along the points you made in the issue.

PR #1841, opened 2017-02-10
fix download. can't use apihost because not deployed yet.

domdom82:
apihost is usually the router which is not deployed at that point in time, therefore using edge only.


domdom82:
PG1 1151

PR #1846, opened 2017-02-10
please decline this PR - see comments

romankhar:
cc: @csantanapr  - please take a look at the updates for Message Hub integration.


romankhar:
@csantanapr -   you can decline this PR as I submitted changes to the Kafka package repo instead of here.

PR #1852, opened 2017-02-14
Actually use runc pause/resume

perryibm:
Actually use flag to use runc when available and configured.

### DIFF
--- a/ansible/roles/invoker/tasks/clean.yml
+++ b/ansible/roles/invoker/tasks/clean.yml
@@ -9,6 +9,11 @@
   ignore_errors: True
 
 # Unless invoker cleans up its own containers we have to do this here.
+#  shell: "(docker ps -aq --no-trunc --filter name=wsk | sudo xargs docker-runc resume); sleep 5"
+- name: pause at runc-level to restore docker consistency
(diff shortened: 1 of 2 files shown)


perryibm:
PG2-1050 passed.   More testing ahead.

ansible/roles/invoker/tasks/clean.yml
```diff
@@ -9,6 +9,11 @@
   ignore_errors: True
 
 # Unless invoker cleans up its own containers we have to do this here.
+#  shell: "(docker ps -aq --no-trunc --filter name=wsk | sudo xargs docker-runc resume); sleep 5"
+- name: pause at runc-level to restore docker consistency
```
domdom82: I think now is the time to make that change in the invoker. I recall having opened an issue "invoker should clean up on terminate" or something way back in time. We agreed to dismiss it at the time and have ansible clean up manually. 

@perryibm how much effort would it be to unpause and remove all action containers when the invoker receives a SIGTERM?

jeremiaswerner: We discussed to address this in a different PR.

jeremiaswerner:
rebased and merged in 85b82a7025bcaa9e1d32c8796a14081f6b852b14

PR #1857, opened 2017-02-15
fix check in actionLimitTest

RSulzmann:
testcase sometimes failes parsing the log string( because of the spaces after timestamp):
"2017-02-15T10:13:58.82285493Z  stdout: ERROR: opened files =  1011"

### DIFF
--- a/tests/src/whisk/core/limits/ActionLimitsTests.scala
+++ b/tests/src/whisk/core/limits/ActionLimitsTests.scala
@@ -229,8 +229,7 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers {
                     n.toInt should be >= minExpectedOpenFiles
 
                     activation.logs.getOrElse(List()).filter {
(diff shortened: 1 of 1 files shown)


rabbah:
hmm do you have an example?

markusthoemmes:
Argh, the pasted line swallowed the extra space after the date.

rabbah:
Markus showed me the extra space. Two thoughts. Split on regex with space+. Or find out where there's an extra space there at all - there should not be.

rabbah:
@RSulzmann explained the extra space is from the fixed-space padding.
I suggest we change the split to a regex, eg. ` raw"\s+".r.split(...)...`

markusthoemmes:
@rabbah I think the `contains` check is enough. Nailing the logmessage down so narrowly doesn't really help?

rabbah:
alright - overly paranoid.

PR #1860, opened 2017-02-15
tweak text describing per minute system limit to avoid term "user"

sjfink:


rabbah:
How about May be submitted ...

sjfink:
> How about May be submitted ...

The previous row uses "are allowed".   So I think both rows should use the same convention -- don't care which one as long as its clear.

csantanapr:
duplicate of #1866

PR #1867, opened 2017-02-16
Handle bytes to string conversion in actionproxy.

jtyberg:
[actionproxy.py](https://github.com/openwhisk/openwhisk/blob/master/core/actionProxy/actionproxy.py#L130) assumes that subprocess output is a string, which is true in Python 2.  However,
Python 3 default behavior switched to returning bytes instead of strings.  (This [SO answer](http://stackoverflow.com/questions/21689365/python-3-typeerror-must-be-str-not-bytes-with-sys-stdout-write) provides a good explanation).  

If a user creates a Docker action that uses python 3 in their container, for example, invoking the action will fail with an internal error:

```
{
  "error": "Internal error."
}
```

After some local debugging in my own Docker container (using python 3), the issue is:

```
Traceback (most recent call last):
  File "actionproxy.py", line 207, in run
    (code, result) = runner.run(args, runner.env(message if message else {}))
  File "actionproxy.py", line 132, in run
    process_output_lines = o.strip().split('\n')
TypeError: a bytes-like object is required, not 'str'
```

or 

```
Traceback (most recent call last):
  File "actionproxy.py", line 207, in run
    (code, result) = runner.run(args, runner.env(message if message else {}))
  File "actionproxy.py", line 140, in run
    sys.stderr.write(e)
TypeError: write() argument must be str, not bytes
```

These changes work in both Python 2 and Python 3 environments.

Note that other spots like `sys.stdout.write('%s\n' % line)` work because of the implicit conversion to `str`.  In other words, we're getting lucky.

### DIFF
--- a/core/actionProxy/actionproxy.py
+++ b/core/actionProxy/actionproxy.py
@@ -134,17 +134,23 @@ class ActionRunner:
             return error(e)
 
         # run the process and wait until it completes.
+        # stdout/stderr will always be set because we passed PIPEs to Popen
         (o, e) = p.communicate()
 
-        if o is not None:
-            process_output_lines = o.strip().split('\n')
-            last_line = process_output_lines[-1]
-            for line in process_output_lines[:-1]:
-                sys.stdout.write('%s\n' % line)
-        else:
-            last_line = '{}'
+        # stdout/stderr may be either text or bytes, depending on Python
+        # version.   In the latter case, decode to text.
+        if isinstance(o, bytes):
+            o = o.decode('utf-8')
+        if isinstance(e, bytes):
+            e = e.decode('utf-8')
 
-        if e is not None:
+        # get the last line of stdout, even if empty
+        process_output_lines = o.strip().split('\n')
+        last_line = process_output_lines[-1]
+        for line in process_output_lines[:-1]:
+            sys.stdout.write('%s\n' % line)
+
+        if e:
             sys.stderr.write(e)
 
         try:
--- a/core/swift3Action/swift3runner.py
+++ b/core/swift3Action/swift3runner.py
@@ -61,14 +61,28 @@ class Swift3Runner(ActionRunner):
             os.chmod(self.binary, 0o555)
             return
 
-        p = subprocess.Popen(BUILD_PROCESS, cwd=DEST_SCRIPT_DIR)
+        p = subprocess.Popen(
+            BUILD_PROCESS,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            cwd=DEST_SCRIPT_DIR)
+
+        # run the process and wait until it completes.
+        # stdout/stderr will not be None because we passed PIPEs to Popen
         (o, e) = p.communicate()
 
-        if o is not None:
+        # stdout/stderr may be either text or bytes, depending on Python
+        # version.   In the latter case, decode to text.
+        if isinstance(o, bytes):
+            o = o.decode('utf-8')
+        if isinstance(e, bytes):
+            e = e.decode('utf-8')
+
+        if o:
             sys.stdout.write(o)
             sys.stdout.flush()
 
-        if e is not None:
+        if e:
             sys.stderr.write(e)
             sys.stderr.flush()
 


rabbah:
pg3/276.

rabbah:
rebased and running as pg3/278.

rabbah:
These two tests failed twice in jenkins

```
    actionContainers.Swift3ActionContainerTests.swift3action should return some error on action error
    actionContainers.SwiftActionContainerTests.swiftaction should return some error on action error
```

with status code 500 not equal to the expected 502; will have to investigate it.

jtyberg:
I'll try to replicate the error locally.

rabbah:
pg2/1127 failed again
```
rg.scalatest.exceptions.TestFailedException: 500 was not equal to 502

Stacktrace

org.scalatest.exceptions.TestFailedException: 500 was not equal to 502
	at org.scalatest.MatchersHelper$.newTestFailedException(MatchersHelper.scala:160)
	at org.scalatest.Matchers$ShouldMethodHelper$.shouldMatcher(Matchers.scala:6231)
	at org.scalatest.Matchers$AnyShouldWrapper.should(Matchers.scala:6265)
	at actionContainers.SwiftActionContainerTests$$anonfun$2$$anonfun$7.apply(SwiftActionContainerTests.scala:143)
	at actionContainers.SwiftActionContainerTests$$anonfun$2$$anonfun$7.apply(SwiftActionContainerTests.scala:136)
```

@jtyberg i'll have a quick look, see if i can save you the trouble.

rabbah:
I can reproduce the failure locally with the response being `{"error":"Internal error."}`

jtyberg:
Yeah, trying to deal with all string types in Python 2 and 3 is hard.  After some further reading, I think the best solution might be to decode the subprocess output to text, if necessary, and always deal with text.  

> Making the distinction easier to handle can be accomplished by encoding and decoding between binary data and text at the edge of your code. This means that when you receive text in binary data, you should immediately decode it. And if your code needs to send text as binary data then encode it as late as possible. This allows your code to work with only text internally and thus eliminates having to keep track of what type of data you are working with.

https://docs.python.org/3/howto/pyporting.html#text-versus-binary-data

cclauss:
@jtyberg could you please [try changing](https://github.com/openwhisk/openwhisk/blob/master/core/actionProxy/actionproxy.py#L132):

```python
process_output_lines = o.strip().split('\n')
# to
process_output_lines = str(o).strip().split('\n')
# or
process_output_lines = o.decode('utf-8').strip().split('\n')
```

jtyberg:
@cclauss The current PR already has a line that decodes stdout _and_ stderr **IF** they are bytes (Python 3 only).  Does this suffice?

```
# stdout/stderr may be either text or bytes, depending on Python
# version.   In the latter case, decode to text.
o, e = map(lambda x: x.decode() if isinstance(x, bytes) else x, (o, e))
```

In Python 2, stdout and stderr would be text, which means `o.decode()` would not work because strings don't have a `decode` method.

```
Python 3.6.0 (v3.6.0:41df79263a11, Dec 22 2016, 17:23:13) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> o = u'foo'
>>> type(o)
<class 'str'>
>>> o.decode('utf-8')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'str' object has no attribute 'decode'
```

cclauss:
Thanks for the explanation.  As you say, trying to deal with all string types in Python 2 and 3 is hard.
- [ ] So does your solution above work in both Python 2 and Python 3?
- [ ] Does it pass the [two tests mentioned above](https://github.com/openwhisk/openwhisk/pull/1867#issuecomment-280817200)?
- [ ] Is this PR ready for review?
- [ ] Are there [other places](https://github.com/openwhisk/openwhisk/search?utf8=%E2%9C%93&q=p.communicate%28%29&type=Code) where we could see similar bytes vs. strings issues?

jtyberg:
@cclauss 

> - [x] So does your solution above work in both Python 2 and Python 3?

Yes.

> - [ ] Does it pass the two tests mentioned above?

Only one of the tests actually exists anymore.  I do not see any failures in Jenkins.

> - [x] Is this PR ready for review?

Maybe? (see below)

> - [x] Are there other places where we could see similar bytes vs. strings issues?

If I had to give an answer, then I'd say we're sure to hit one or two issues down the line ;-).  Since `actionproxy.py` is going to straddle the Python2/3 world, we should probably follow [recommended practices](https://docs.python.org/3/howto/pyporting.html#text-versus-binary-data), including

> 1. Decide which of your APIs take text and which take binary data
> 2. Make sure that your code that works with text also works with unicode and code for binary data works with bytes in Python 2 (see the table above for what methods you cannot use for each type)
> 3. Mark all binary literals with a 'b' prefix, textual literals with a 'u' prefix
> 4. Decode binary data to text as soon as possible, encode text as binary data as late as possible
> 5. Open files using `io.open()` and make sure to specify the b mode when appropriate

Right now, this PR handles (4).  I can include additional commits to this PR to address (3) and (5) rather easily.

core/actionProxy/actionproxy.py
```diff
@@ -136,13 +136,17 @@ def error(msg):
         # run the process and wait until it completes.
         (o, e) = p.communicate()
 
+        # stdout/stderr may be either text or bytes, depending on Python
+        # version.   In the latter case, decode to text.
+        o, e = map(lambda x: x.decode() if isinstance(x, bytes) else x, (o, e))
```
cclauss: The map(lambda) makes my head spin.  Please consider changing this to the longer, but more understandable:
```python
if isinstance(o, bytes):
        o = o.decode('utf-8')
if isinstance(e, bytes):
        e = e.decode('utf-8')
```

Also line 143 (if o is not None) will _always_ be True.  o might be '' but it will _never_ be None.  Same for e on line 151.

You might consider changing these to `if o:` and `if e:` which more closely aligns with the original intention.

Finally, I not recommend all of the [u'' changes](http://python-future.org/unicode_literals.html#unicode-literals) unless you see an absolute necessity.

jtyberg: > The map(lambda) makes my head spin.

Fair enough.

> Also line 143 (if o is not None) will always be True. o might be '' but it will never be None. Same for e on line 151.

True.  The check was already there in existing code.  I did not remove it originally because it might be required if someone were to change the parameters passed to `Popen`.  It felt safer to leave it.

> Finally, I would not recommend all of the u'' changes unless you see an absolute necessity.

Removed. Not necessary, just the best practice mentioned in the above comment.

core/actionProxy/actionproxy.py
```diff
@@ -136,15 +136,19 @@ def error(msg):
...
-
-        if e is not None:
+        # stdout/stderr may be either text or bytes, depending on Python
+        # version.   In the latter case, decode to text.
+        if isinstance(o, bytes):
+            o = o.decode('utf-8')
+        if isinstance(e, bytes):
+            e = e.decode('utf-8')
+
+        process_output_lines = o.strip().split('\n')
```
cclauss: I am sorry.  Perhaps my writing was unclear.  I think we still need the if/then/else.  The change that I propose was just to change `if o is not None:` --> `if o:`.

jtyberg: My fault.  Sorry for being dense.

cclauss: Perfect.  Thanks much.

cclauss:
This fix is VERY important for Python 3 compatibility.  It allows us to pass 11 test cases that fail when we move from Python 2 to Python 3.

rabbah:
pg3/403 🚀.

rabbah:
Failed same swift test:

```
actionContainers.SwiftActionContainerTests.swift3action should return some error on action error
org.scalatest.exceptions.TestFailedException: 200 was not equal to 502
```

cclauss:
@jtyberg A similar change could be made at https://github.com/openwhisk/openwhisk/blob/master/core/swift3Action/swift3runner.py#L65

jtyberg:
Yes, `swift3runner.py` should also be updated.  As it stands, it will never get anything back from the `swiftbuildandlink.sh` script in the subprocess because it's not passing stdout/stderr into the Popen constructor.  So `o` and `e` will always be `None`.

```
p = subprocess.Popen(BUILD_PROCESS, cwd=DEST_SCRIPT_DIR)
(o, e) = p.communicate()
```

I'm not sure how this affects the test cases though, as the subprocess is used for the build step.  It should probably print the output of the script, regardless.

Let me get the tests running locally, then I'll update this PR.

jtyberg:
I modified the commit yet again.  As it turns out, switching from 

```
if o is not None:
    process_output_lines = o.strip().split('\n')
    last_line = process_output_lines[-1]
    for line in process_output_lines[:-1]:
        sys.stdout.write('%s\n' % line)
else:
    last_line = '{}'
```
to
```
if o:
    ...
```
changes the behavior.  This is because stdout/stderr (`o` and `e`) would NEVER be `None`, and `last_line` would ALWAYS be set, even if just to an empty string.  The `else` block NEVER got called.

In the failing swift3action test [above](https://github.com/openwhisk/openwhisk/pull/1867#issuecomment-286529111), `o` would be empty, and thus, `last_line = '{}'`, but `e` would have the stderr output from the failed action.  

The logic for checking if the action failed depends on whether `last_line` is valid JSON, NOT on the content of stdout/stderr, or on the return code from the subprocess.  In the future, I wonder if this should change, but for now, I'm simply removing the conditional check `if o...` to get back to the original behavior.

rabbah:
pg2/1240 🔵

rabbah:
@bjustin-ibm I reviewed the changes with @jtyberg and consider the pr ready. Defer to you at this point if you have outstanding concerns.

PR #1868, opened 2017-02-17
Remove meta packages, normalize on web actions.

### LINKED ISSUE #1850: web actions should not project return when using extension .http
When using content extension `.http` with a path like `demo/hello.http/pets/123` it will automatically try to project the field `pets.123` from the return object.

This is not the use case I want, I want to be able to return a `body` and be able to access the path `pets/123` (`__ow_meta_path`) to parse it and use it a path parameter in my web action.

rabbah:
Adds (whisk) authenticated web actions.
Removes result projection for http actions.
Fixes #1850.


csantanapr:
This is very good to have this cleanup now. 

Question I see that now the action can be "protected" meaning any valid subject with a valid keycan invoke. 

How do you envision in the future on how to make the action "private" meaning only the owner subject can invoke it. 
1. It could be by setting an annotation with a "invoke-key" that owner generates a value and set
2. It can be by reusing the anotation require-whisk-auth with another posible value "self", then as long an owkey owned by the owner of the action can invoke

rabbah:
> How do you envision in the future on how to make the action "private" meaning only the owner subject can invoke it.


don't web export it - further the action will get the namespace of the authenticated user and can implement its own ACL.

csantanapr:
But I want the benefit of being able to invoke it with http verbs in a private way from the apigateway or other backend, and not to have to incur into the cost of an invocation just to return auth error from inside the action, if I could avoid it by declaring that only the owner is allowed to invoke it then controller can do the auth for me and not waiste work to do the invocation of the action

rabbah:
This is adding authentication - not authorization. It's nearly semantic preserving with removing meta package support for web actions.

I don't dispute adding authorization would be useful - but outside of the scope of this pr, no?
Further you're touching on feature which are related to moving web action enhancements to the core API.

csantanapr:
Yes all my comments are outside this PR
This PR as it is is well defined and complete. 

I was just taking opportunity to have a comment about the new annotation require-whisk-auth 

I don't have any further comments. 

Good job 👍

rabbah:
Thanks @csantanapr - shall I assign to you pending reviews?

rabbah:
pg2/1073 approved.

tests/src/common/Wsk.scala
```diff
@@ -946,8 +946,23 @@ object WskAdmin {
         Buffer(WhiskProperties.python, new File(binDir, binaryName).toString)
     }
 
+    def getKey(subject: String): String = {
```
cbickel: This method is used nowhere.
Is it planned to use it anywhere?

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -62,11 +62,7 @@ private case class Context(
 }
 
 protected[core] object WhiskMetaApi extends Directives {
-    /**
-     * Defines the properties that must be present in a configuration
-     * in order to implement the Meta API.
-     */
-    def requiredProperties = Map(WhiskConfig.systemKey -> null)
+    def requiredProperties = Map()
```
cbickel: Do we really need this function, if it is empty anyway?

tests/src/common/Wsk.scala
rabbah: good catch - will remove it.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
rabbah: debatable - i removed it then changed my mind.
i cant articulate my reasons strongly, so i'll remove it again.

rabbah:
pg2/1079 re-approved.

csantanapr:
@cbickel @markusthoemmes could you review this PR, anything else to comment before merging this in?

cbickel:
@csantanapr  Looks good now 👍

csantanapr:
merged 4550ac1316d4e1690b7d5e461c26977d7abca34a

PR #1874, opened 2017-02-20
Update to allow for multiple License during scans (i.e., Apache-IBM, ASF)

mrutkows:
This change will allow the scanCode.py script to still accept the current Apache/IBM LICENSE/Copyright file header, as well as supporting the new Apache Software Foundation (ASF) header.  This approach will allow us to change, over time, the header for all source files as is required of us as part of being an Apache project.  

The approach is to provide an array of header text (strings) and pass the test if either appear at the top of an appropriate file.

In addition, I have assured that this script is now "Pythonic" and it now will pass all Flake8 and Pep8 tests.  These changes include:

- Provide proper Python header (string literal); which includes the new ASF LICENSE header.
- Fix all indentation throughout file
- Follow proper line continuation practices
- Assure code does not extend past 80 columns
- Assure all functions have proper docstrings (instead of general comment blocks)
- Convert lambda functions to "def" (full functions) if they are being assigned to variable (will persist)
- Define (formatted) strings used for stdout at top of file as constants (which also better allows them to be translated in the future).


mrutkows:
If you would like to see the files I used to test with they are in my personal repo. here: https://github.com/mrutkows/codescan

csantanapr:
PG1 1177

@mrutkows Its recomendable to create a feature branch instead of using `master` when you submit a PR.
See https://github.com/openwhisk/openwhisk/wiki/Contributing:-Git-guidelines#preparing-pull-requests for more information.

tools/build/scanCode.py
```diff
@@ -1,20 +1,31 @@
...
+"""Executable Python script for scanning source code for compliance.
+
+   This script checks some (simple) conventions:
+   - no symlinks
+   - no tabs
+   - no trailing whitespaces
+   - files end with EOL
+   - valid license headers in source files (where applicable)
+
+/*
```
rabbah: comment lines do not align 
```
/*
 * ...
 */
```

vs
```
/*
* ...
*/
```

tools/build/scanCode.py
```diff
@@ -25,85 +36,134 @@
...
-         * Unless required by applicable law or agreed to in writing, software
-         * distributed under the License is distributed on an "AS IS" BASIS,
-         * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-         * See the License for the specific language governing permissions and
-         * limitations under the License.
-         */
-        """)
+"""Declare approved softare license headers as strings."""
+
+LICENSE_APACHE_SOFTWARE_FOUNDATION = """\
```
csantanapr: My understanding was that once the scancode tool is in, we can go ahead and move forward and replace all copyright headers with the Apache one.
Since the CREDITS file now contains the attribution to IBM.

Once we know there is no more IBM Copyright across all repos, then we can modify scancode again to only check for Apache header. Apache header doesn't have year

rabbah: Great.

csantanapr: The only catch now is that this script is use for internal repos, and we update the copyright for the year we need to adjust. maybe we can update the tool to some how ignore the copyright year line.

mrutkows: The plan is to remove the IBM header once we are confident that all ASF headers are in place. This header SHOULD be the only one we use then going forward.  I also have plans (started work) on adding more supported source file types that have different grammars with different comment block designators (so the array approach will still be useful).  I am starting work to support Python (.py) and account for she-bang statements, etc..

tools/build/scanCode.py
```diff
@@ -117,96 +177,133 @@ def run_line_checks(file_path):
         return errors
     return run_line_checks
 
-# Runs a series of file-by-file checks.
+
 def run_file_checks(file_path, checks):
+    """Run a series of file-by-file checks."""
     errors = []
-
+    # if VERBOSE (True) then print filename neing checked
```
mrutkows: Will fix typos and resubmit.

tools/build/scanCode.py
```diff
@@ -117,96 +177,133 @@ def run_line_checks(file_path):
...
-        ("*.scala",    [ has_block_copyright, line_checks([ no_tabs, no_trailing_spaces, eol_at_eof ]) ]),
-        ("*.py",       [ line_checks([ no_tabs, no_trailing_spaces, eol_at_eof ]) ]),
-        ("*.java",     [ has_block_copyright, line_checks([ no_tabs, no_trailing_spaces, eol_at_eof ]) ]),
-        ("*.js",       [ line_checks([ no_tabs, no_trailing_spaces, eol_at_eof ]) ]),
-        ("build.xml",  [ line_checks([ no_tabs, no_trailing_spaces, eol_at_eof ]) ]),
-        ("deploy.xml", [ line_checks([ no_tabs, no_trailing_spaces, eol_at_eof ]) ]),
-        ("*.gradle",   [ line_checks([ no_tabs, no_trailing_spaces, eol_at_eof ]) ]),
-        ("*.md",       [ line_checks([ no_tabs, eol_at_eof ]) ]),
-        ("*.go",       [ has_block_copyright, line_checks([ no_tabs, no_trailing_spaces, eol_at_eof ]) ])
+        ("*", [is_not_symlink]),
```
mrutkows: However, it is does not pass Pep8 checks which all Python should.

tools/build/scanCode.py
mrutkows: danke

tools/build/scanCode.py
rabbah: "should"? too strong.

there are other coding conventions not just Pep8. i think when editing pre-existing code, it's better to follow the conventions it already applies when there are no enforcement in place.  
we can discuss this elsewhere as this could get into the weeds.

mrutkows: not too strong IMO, Pep8 is expected within Python communities (they may turn off/on specific numeric requirements here and there(, but general spacing, line-continuation is pretty much a should... or you will be called "non-Pythonic", even to the extent Python developers will not look at your code.  I have found this out the hard way...

rabbah: weeds ahead ;)

mrutkows:
committed updated addressing comments and ran a full spell check as well.

PR #1876, opened 2017-02-20
Rework replicator tests to use environment-specific prefix

cbickel:
Use the dbPrefix of the environment, where the test runs as part of the dbPrefix for all replicator tests.
Also delete databases in replicator only, if the database is a backup of a db with the current prefix.

PG2#1075 started.


cbickel:
PG2#1075 is green

PR #1878, opened 2017-02-20
Remove whisk object from node.js

jasonpet:
The global object whisk had been deprecated and will now be removed.

migrate your nodejs actions to use altenative methods. For the functions `whisk.invoke()` and `whisk.trigger()` use the already installed client library [openwhisk](https://www.npmjs.com/package/openwhisk). For the `whisk.getAuthKey()` you can get the API key value from the environment variable `__OW_API_KEY`. For the `whisk.error()` you can return a rejected Promise (i.e. Promise.reject).


jasonpet:
pg2#1076

bjustin-ibm:
pg3 282

tests/src/system/basic/WskBasicNodeTests.scala
```diff
@@ -147,181 +145,6 @@ class WskBasicNodeTests
...
-            val asyncAction = Some(TestUtils.getTestActionFilename("threeSecondRule.js"))
-
-            assetHelper.withCleaner(wsk.action, asyncName) {
-                (action, _) =>
-                    action.create(asyncName, asyncAction)
-            }
-
-            // this action supplies a 'next' callback to whisk.invoke()
-            val invokeActionName = "invokeAction"
-            val invokeAction = Some(TestUtils.getTestActionFilename("invokeCallback.js"))
```
rabbah: also removed invokeCallback.js?

rabbah: and threeSecondRule.js?

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -339,24 +339,6 @@ class WskBasicTests
             }
     }
 
-    /**
-     * Tests creating an nodejs action that throws a whisk.error() response. The error message thrown by the
-     * whisk.error() should be returned.
-     */
-    it should "create and invoke a blocking action resulting in a whisk.error response" in withAssetCleaner(wskprops) {
```
rabbah: this test should still be valid, no?

tests/src/whisk/core/cli/test/SequenceMigrationTests.scala
```diff
@@ -82,11 +82,6 @@ class SequenceMigrationTests
         }
     }
 
-    it should "invoke an old-style sequence (original pipe.js) and get the result" in {
```
rabbah: @ioana-blue since whisk.invoke is now gone - any other pipe.js cleanup we can do?

tests/src/system/basic/WskBasicTests.scala
jasonpet: whisk.error is gone

tests/src/system/basic/WskBasicNodeTests.scala
jasonpet: removed

tests/src/system/basic/WskBasicTests.scala
rabbah: The tests are to confirm returning {error:...} work as expected with the cli for example.

rabbah: I don't think we need two versions of this tests - remove ...2.js and rename ...1.js accordingly perhaps.

jasonpet: there are actually three versions now for testing that returning {error: ....} works as expected.  I had left the test that uses the applicationError3.js file and had removed the test that used the other two since they are now essentially all testing the same thing.  We only need one of these now.  I will update the test I had left to say applicationError1.js and remove the other files that are no longer necessary.

rabbah: 3: 😳 thanks!

jasonpet:
pg2#1080

tests/src/system/basic/WskBasicNodeTests.scala
```diff
@@ -147,181 +145,6 @@ class WskBasicNodeTests
...
-                    duration should be >= expectedDuration.toMillis
-            }
-    }
-
-    // TODO: remove this tests and its assets when "whisk.js" is removed entirely as it is no longer necessary
-    it should "Ensure that whisk.trigger() still uses a callback when provided one" in withAssetCleaner(wskprops) {
-        (wp, assetHelper) =>
-            // this action supplies a 'next' callback to whisk.trigger()
-            val nameOfActionThatTriggers = "triggerAction"
-            val actionThatTriggers = Some(TestUtils.getTestActionFilename("triggerCallback.js"))
```
bjustin-ibm: @jasonpet Looks like we can also remove `triggerCallback.js` and `triggerPromise.js`

jasonpet: done

PR #1884, opened 2017-02-21
Access HTTP Body from Web Actions

dubeejw:
- Provide __ow_meta_body argument that allows access to the content of the HTTP body passed to a Web Action

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Meta.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Meta.scala
@@ -56,7 +56,8 @@ private case class Context(
     def metadata(user: Option[Identity]): Map[String, JsValue] = {
         Map("__ow_meta_verb" -> method.value.toLowerCase.toJson,
             "__ow_meta_headers" -> headers.map(h => h.lowercaseName -> h.value).toMap.toJson,
-            "__ow_meta_path" -> path.toJson) ++
+            "__ow_meta_path" -> path.toJson,
+            "__ow_meta_body" -> body.toJson) ++
             user.map(u => "__ow_meta_namespace" -> u.namespace.asString.toJson)
     }
 }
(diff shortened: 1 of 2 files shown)


dubeejw:
Related to #1868.

dubeejw:
PG2 284.

dubeejw:
Existing tests that apply to this PR:
`Meta API should invoke action and merge query and body parameters`
`Meta API should invoke action and merge body parameters`

dubeejw:
@rabbah, PG is approved, ready for review.

rabbah:
We shouldn't send the body as the meta object always. Since for form data, and json it's already the args this would effectively cut the allowed payload size in half. Further sending anything other than formdata or json object in the body will fail to match in the routes and will be rejected. 

So what is the goal here? To accept body other than formdata and json object? If so, 1. You will need a test for that, and 2. You'll need to also consider how to deal with the entity extractors which don't currently permit anything other  that afformenrioned types.

rabbah:
> @rabbah, PG is approved, ready for review. See notes about tests above.

I think you need a new test that sends a content type other than formdata or json object - and once you add one, you'll see why.

See my previous comment.

dubeejw:
@rabbah, I noticed the content restriction as I was manually test. @csantanapr, @mdeuser, see above comments.

dubeejw:
I believe the goal is to accept payloads other than JSON or Form Data. Users would be responsible for processing input in other forms themselves to make Whisk more flexible in the future.

rabbah:
I'm curious if there is a motivating use case or requirement underlying this change --- or is this done just to be general? Form data, json object, and query parameters should cover a large set of use cases.

mdeuser:
@rabbah - Perhaps there's a technical reason/cost for restricting media-type form data and json, but IMHO it's beneficial to support as many as possible so an API developer is free choose whatever content type is desired.

rabbah:
There's added complexity to the code. Before we support the kitchen sink we should at least make sure we support the common cases first and do them well. So I'm curious if there's an actual need for this vs let's do it to be general.

mbehrendt:
i agree with @mdeuser and @bjustin-ibm . providing access to the req body is the most generic approach, that provides most flexibility. content-type specific specializations can be layered on top.

mdeuser:
Not sure this would help reduce the complexity... the json/formdata media-types can be considered "first class" content (i.e. content is parsed for parameters), while all other media-types are "second class" in that no parameter parsing is provided - the action gets the entire payload as-is in a single json field.  Although, I don't know if the as-is will involve encode/decode of the payload

rabbah:
See #1965 for plan for record for making "body" available to a web action.

PR #1888, opened 2017-02-21
Do not retry /run on action containers

bjustin-ibm:
It should be assumed that /run only happens after successful /init which should itself ensure that the container is running and ready. With that in mind, it does not make sense to attempt to retry posting to the /run endpoint on connection failures. Such a failure is an indication that something catastrophic has happened to the container, and no attempt at recovery should be made.

This is accomplished by adding a "retry" flag to the appropriate HTTP utilities to control whether or not a retry should be attempted. ~This new flag defaults to `true` (to allow existing call sites to remain unchanged) and is explicitly set to `false` when posting to /run~

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
@@ -66,9 +66,10 @@ protected[container] class HttpUtils(
      *
      * @param endpoint the path the api call relative to hostname
      * @param body the json object to post
+     * @param retry whether or not to retry on connection failure
      * @return Left(Error Message) or Right(Status Code, Response as UTF-8 String)
      */
-    def post(endpoint: String, body: JsObject): Either[ContainerConnectionError, ContainerResponse] = {
+    def post(endpoint: String, body: JsObject, retry: Boolean): Either[ContainerConnectionError, ContainerResponse] = {
         val entity = new StringEntity(body.compactPrint, StandardCharsets.UTF_8)
         entity.setContentType("application/json")
 
@@ -76,10 +77,10 @@ protected[container] class HttpUtils(
         request.addHeader(HttpHeaders.ACCEPT, "application/json")
         request.setEntity(entity)
 
-        execute(request, timeout.toMillis.toInt)
+        execute(request, timeout.toMillis.toInt, retry)
     }
 
-    private def execute(request: HttpRequestBase, timeoutMsec: Integer): Either[ContainerConnectionError, ContainerResponse] = {
+    private def execute(request: HttpRequestBase, timeoutMsec: Integer, retry: Boolean): Either[ContainerConnectionError, ContainerResponse] = {
         Try(connection.execute(request)).map { response =>
             val containerResponse = Option(response.getEntity).map { entity =>
                 val statusCode = response.getStatusLine.getStatusCode
@@ -103,11 +104,11 @@ protected[container] class HttpUtils(
             containerResponse
         } match {
             case Success(r) => r
-            case Failure(t: HttpHostConnectException) =>
+            case Failure(t: HttpHostConnectException) if retry =>
                 if (timeoutMsec > 0) {
                     Thread sleep 100
                     val newTimeout = timeoutMsec - 100
-                    execute(request, newTimeout)
+                    execute(request, newTimeout, retry)
                 } else {
                     Left(Timeout())
                 }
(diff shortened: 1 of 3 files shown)


core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
```diff
@@ -66,20 +66,21 @@ protected[container] class HttpUtils(
      *
      * @param endpoint the path the api call relative to hostname
      * @param body the json object to post
+     * @param retry whether or not to retry on connection failure
      * @return Left(Error Message) or Right(Status Code, Response as UTF-8 String)
      */
-    def post(endpoint: String, body: JsObject): Either[ContainerConnectionError, ContainerResponse] = {
+    def post(endpoint: String, body: JsObject, retry: Boolean = true): Either[ContainerConnectionError, ContainerResponse] = {
```
rabbah: i'd remove the default - particularly since you're threading the value through everywhere.

core/invoker/src/main/scala/whisk/core/container/WhiskContainer.scala
```diff
@@ -134,10 +134,11 @@ class WhiskContainer(
      * Posts a message to the container.
      *
      * @param msg the message to post
+     * @param retry whether or not to retry on connection failure
      * @return response from container if any as array of byte
      */
-    private def sendPayload(endpoint: String, msg: JsObject, timeout: FiniteDuration)(implicit system: ActorSystem): RunResult = {
-        sendPayloadApache(endpoint, msg, timeout)
+    private def sendPayload(endpoint: String, msg: JsObject, timeout: FiniteDuration, retry: Boolean = true)(implicit system: ActorSystem): RunResult = {
```
rabbah: and also remove default value here.

core/invoker/src/main/scala/whisk/core/container/WhiskContainer.scala
```diff
@@ -194,7 +195,7 @@ class WhiskContainer(
         }
     }
 
-    private def sendPayloadApache(endpoint: String, msg: JsObject, timeout: FiniteDuration): RunResult = {
+    private def sendPayloadApache(endpoint: String, msg: JsObject, timeout: FiniteDuration, retry: Boolean = true): RunResult = {
```
rabbah: and here.

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
bjustin-ibm: I understand the other two since they are private, but for this one won't all existing callers need to be found and updated to explicitly include the new parameter?

rabbah: iirc: There is only one call site in the invoker (from WhiskContainer) and from one test suite ContainerConnectionTests. Not too bad.

rabbah: The rationale is that retry should be explicitly opted into (given the reason we're making this change).

bjustin-ibm: Can confirm: only one other place to update. I guess I assumed HTTPUtils was more widely used.

bjustin-ibm:
PG1 1182

domdom82:
A couple of thoughts on this PR:

- I agree on not retrying on /run which makes little sense after /init worked we can assume the container is healthy.
- However this PR does not yet address the root cause of prematurely exiting containers:
When an action runs and that action somehow manages to kill the container (i.e. container is stopped or dead afterwards), a subsequent pause will fail, followed by a failed unpause which currently is ignored and a call to /init follows.

@perryibm keep me honest here but I think with the change to runc for pause / unpause we can tighten the checks here again and assume the following:

1. A cold-started container is running
2. After an action has run, the container is still running
2.1. If it is not running (i.e. pause fails), the container is deemed broken and gets gc'ed
3. After pause the container is paused
4. After unpause the container is running
4.1. If it is not running (i.e. unpause fails), the container is deemed broken and gets gc'ed


So IMHO we still need another PR for fixing the invoker.

bjustin-ibm:
> So IMHO we still need another PR for fixing the invoker.

This is correct and understood. We can roll that additional change into this PR as well, but I'm taking this on as a sort of Invoker training exercise, so it may not be as expedient as we all might like 😉

rabbah:
@domdom82 there were two actions to take - this is the first of them. So yes, we need to address the second issue as well but that can be done as a separate PR.

bjustin-ibm:
PG1 1187

perryibm:
LGTM.  Retry should be explicit and without a default as is done.  Can I have one more PG please?

We do need to address the underlying issue of state inconsistency but I agree it should be a separate PR as it's orthogonal.   The fundamental problem is potentially even more difficult because the inconsistency can arise without any operation (pause/resume) to witness the container going away so we might want to think about remedies there as well.

bjustin-ibm:
PG1 1188

bjustin-ibm:
Merged as 6a95fd7adaf5b6319535beb33e850d89ac614782

PR #1889, opened 2017-02-22
Change runc marker to existing format

cbickel:
FYI @RSulzmann 

This PR changes the runc log markers to the format we use with all other markers.

PG1#1186 is running.


jeremiaswerner:
rebased and merged in 3817675bc677fd52cb275cb1f96fe09edae0507d

PR #1897, opened 2017-02-22
Created new server_or_less.md file

romankhar:
If you have any edits or comments - please edit the file directly here: https://docs.google.com/document/d/1MJzUSQA_ckRmS872C58-oFQJLE35PNLBM58RBzmNRsE/edit?usp=sharing


csantanapr:
@romankhar you are including the same file changes across multiple PRs.
Please either have a PR have only the changes corresponding file changes to the PR, Or just one single PR with all change files.

#1897 
#1896 
#1895 
#1894 

PR are suppose to be independent change sets compared to what's in master.

If you wan to work with multiple PRs one per file, then only include one file per PR.
The way to do this is to
```
git checkout master
git checkout -b new_file1
modify docs/file1.md
#only git add this file and nothing else
git add docs/file1.md
git push origin new_file1
#go to github to submit PR
```
Do the same for the file2.md take into account that when you do `git status` you will see file1 being untracked or modifed, don't worry just ignore it and just add file2 for PR2 `git add docs/file2.md`

PR #1898, opened 2017-02-23
ContainerPool should gracefully handle a failed unpause

bjustin-ibm:
In certain scenarios, a container can crash which will result in a failed pause and potentially later unpause. However, the ContainerPool does not recognize these failures and proceeds as normal, even though (most importantly) when the unpause fails, the container is not present to handle an action invocation.

This PR is to detect the failed unpause and to treat this the same as a cache miss. The result is that a new container will be created instead of attempting to reuse the crashed container.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/container/Container.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/Container.scala
@@ -59,18 +59,18 @@ class Container(
         s"container [$name] [$id] [$ip]"
     }
 
-    def pause(): Unit =
+    def pause(): Boolean =
         if (useRunc) {
-            RuncUtils.pause(containerId)
+            RuncUtils.isSuccessful(RuncUtils.pause(containerId))
         } else {
-            pauseContainer(containerId)
+            DockerOutput.isSuccessful(pauseContainer(containerId))
         }
 
-    def unpause(): Unit =
+    def unpause(): Boolean =
         if (useRunc) {
-            RuncUtils.resume(containerId)
+            RuncUtils.isSuccessful(RuncUtils.resume(containerId))
         } else {
-            unpauseContainer(containerId)
+            DockerOutput.isSuccessful(unpauseContainer(containerId))
         }
 
     /**
--- a/core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
@@ -291,8 +291,15 @@ class ContainerPool(
             }
             case CacheHit(con) =>
                 con.transid = transid
-                runDockerOp { con.unpause() }
-                Some(Warm(con))
+                runDockerOp {
+                    if (con.unpause()) {
+                        Some(Warm(con))
+                    } else {
+                        // resume failed, gc the container
+                        putBack(con, delete=true)
+                        None
+                    }
+                }
             case CacheBusy => None
         }
     }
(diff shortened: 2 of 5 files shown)


core/invoker/src/main/scala/whisk/core/container/Container.scala
```diff
@@ -59,18 +59,18 @@ class Container(
...
         }
 
-    def unpause(): Unit =
+    def unpause(): Boolean =
         if (useRunc) {
-            RuncUtils.resume(containerId)
+            Container.isRuncSuccessfull(RuncUtils.resume(containerId))
         } else {
-            unpauseContainer(containerId)
+            Container.isDockerSuccessful(unpauseContainer(containerId))
```
rabbah: did you consider moving this to `DockerOutput` as in `DockerOutput.isSuccessful` and provide the same in `RuncUtils`?

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -291,8 +292,17 @@ class ContainerPool(
             }
             case CacheHit(con) =>
                 con.transid = transid
-                runDockerOp { con.unpause() }
-                Some(Warm(con))
+                runDockerOp {
+                    if(con.unpause()) {
```
rabbah: space after `if`.

bjustin-ibm:
PG2 1091

bjustin-ibm:
I added extra logging statements in f2fc12b8e25a64255eb17cd62a61f49b9a29d47c for debugging purposes and didn't originally intend them to get merged but... maybe they are generally useful?

core/invoker/src/main/scala/whisk/core/container/Container.scala
```diff
@@ -114,4 +114,16 @@ object Container {
     def requiredProperties = Map(selfDockerEndpoint -> null, invokerContainerNetwork -> "bridge")
     private val idCounter = new Counter()
     private val removeContainerRetryCount = 2
+
+    private def isRuncSuccessfull(result : (Int, String)) : Boolean =
```
bjustin-ibm: spelling... 😊

core/invoker/src/main/scala/whisk/core/container/Container.scala
bjustin-ibm: You are thinking `RuncUtils.isSuccessful(output : (Int,String)) : Boolean`?

rabbah:
I think the logging is redundant with other log messages that already exist - you can inspect a transaction output to verify.

core/invoker/src/main/scala/whisk/core/container/Container.scala
rabbah: Yes - refactoring the two methods you added into their respective types.

bjustin-ibm: No, I had not considered that :)

bjustin-ibm:
PG2 1097

bjustin-ibm:
@perryibm I believe this is ready unless you have any other concerns or suggestions.

bjustin-ibm:
I removed the additional (redundant) logging in f2fc12b8e25a64255eb17cd62a61f49b9a29d47c and then squashed everything into one commit.

bjustin-ibm:
Merged as 45d2c5615b545c9eb6dda3f01247f395765e9f15

Issue #1899, opened 2017-02-23
Why use the legacy version of Python?


cclauss:
Python 3 is more than 3,000 days old and more than 50% of the top 5,000 most used Python packages are Python 3 compatible, and Python 3.6 has great support of asyncio which is key to the Open Whisk approach.

100% of the packages in the Python execution environment have been ported to Python 3 with the exception of `enum34` and `ipaddress` which are built into the Python 3 standard library.

rabbah:
Bandwidth. We're happy to accept a pull request to add python 3 support.

rabbah:
Duplicate of https://github.com/openwhisk/openwhisk/issues/1184.

rabbah:
Closing as duplicate of above referenced issue.

cclauss:
Reopened because of https://github.com/openwhisk/openwhisk/pull/2094

csantanapr:
@cclauss both python 2 and 3 are enable, any errors or pain using python 3?

What's the actionable item here?

rabbah:
@csantanapr it might be worth explaining and documenting why the python:default mode is python:2 not python:3 - I suspect that's @cclauss' concern.

csantanapr:
oh, yes
If initial python support was done with python 3 instead of python 2, then having now default 3 would be fine.

But now we have users that written python 2 actions and have them in the system and using kind default, switching to python 3 as the default will cause them pain as there is high chance their code braking running on python 3 if they re-create/teardown-redeploy those actions.

So for now leaving python:2 as default and making clear in docs and other means to tell folks that now python 3 is available, that is recommended by the python community that python 3 is  a better choice and that we are evaluating the option of making python 3 the default, so best practice and to protect your code be explicit which python you want when you create or update your action use python:2 and python:3 not default.

rabbah:
for posterity, let me add:

`--kind python` continues to mean `python:2`.
we added explicit kind options for `python:2` and `python:3`.
not that we did not have `python:default` until last week in master, so i don't think this is an assume (Changing it meaning to be 2 vs 3).

instead, the real wrinkle is/was with respect to what kind does the CLI assume when creating python actions and no default was specified. in #1903, this was changed to `python: default`.
I favored the more focused change of making that `python:2` and leaving `python:default` to mean `python:3`.

rabbah:
It will be useful to document the plan to change to bump the default from 2 to 3, and get it rolling.

cclauss:
[...]

Issue #1184, opened 2016-09-01
Python action snippets' Dockerfile  to use Python 3


iainhouston:
## This is a feature / upgrade request

It would be preferable to change the [Openwhisk python snippet `Dockerfile`](https://github.com/openwhisk/openwhisk/blob/master/core/actionProxy/Dockerfile) to  `FROM python:3.5.2-alpine` rather than `FROM python:2.7.12-alpine`

For a number of reasons including the trivial example issue below.
## Rationale

Whilst it may make a lot of sense to Apple, for example,  to have OS X default to  python 2.7 because OS X / MacOS has a big investment in python 2 system programs, it makes better sense really to have (new) OpenWhisk Python snippets written in python 3.
## Environment details:
- Bluemix, Openwhisk `wsk action create pyAction pyProgram.py` 
- (Docker 1.12)
## Steps to reproduce the issue:
1.   Use the python `with` statement (available since 2.6 - thus available to Openwhisk python snippets in python 2.7)  
   The advantages being that opening and closing of the connection (or file or whatever) is done by the `with` statement and need not be managed by the programmer
2. Use it with python 2.7 libraries (like `smtplib`) that have not been upgraded to python 3 and thus have no attribute `__exit__`
## Provide the expected results and outputs:

If it were python 3 this would work without error

``` python
from smtplib import SMTP
with SMTP("domain.org") as smtp:     
    smtp.noop()
```
## Provide the actual results and outputs:

In python 2.7 this is generated:

```
AttributeError: SMTP instance has no attribute '__exit__'
```

rabbah:
We can do this the way we support multiple node or swift version, namely, as python (default to latest), python:2.7 and python:3.5.

iainhouston:
Would it be possible for `wsk action` to look at a `version.txt` file for `python 3.5` or `python 2.7` if present?

csantanapr:
@rabbah Is the request to support python 3 via the action `kind` property?

This enhancement will be in in the context of `kind` for python versions
`wsk action create --kind python:3 myaction mysnippet.py`
or
`wsk action create --kind python:2 myaction mysnippet.py2`

sjfink:
I'd prefer to simply drop python 2.7 rather than support both.

csantanapr:
>I'd prefer to simply drop python 2.7 rather than support both.

Works for me, I was not sure why the runtime got implemented with 2 instead of 3

User can always choose any python version, and any packages they need via docker actions.

[...]

PR #1903, opened 2017-02-24
Step up to Python 3.6.1 and update pip packages

cclauss:
Modernization should help with https://github.com/openwhisk/openwhisk/issues/1184, https://github.com/openwhisk/openwhisk/issues/1899, etc.

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -13,6 +13,7 @@ runtimesManifest:
   - kind: "nodejs"
     deprecated: true
   - kind: "nodejs:6"
+    image: "nodejs6action"
     default: true
   python:
   - kind: "python"
@@ -26,6 +27,7 @@ runtimesManifest:
   - kind: "swift"
     deprecated: true
   - kind: "swift:3"
+    image: "swift3action"
     default: true
   java:
   - kind: "java"
@@ -34,6 +36,7 @@ runtimesManifest:
       attachmentType: "application/java-archive"
     sentinelledLogs: false
     requireMain: true
+    image: "java8action"
     default: true
 
 defaultLimits:
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -20,7 +20,7 @@ runtimesManifest:
   - kind: "python:2"
     image: "python2action"
   - kind: "python:3"
-    image: "pythonaction"
+    image: "python3action"
     default: true
   swift:
   - kind: "swift"
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -16,6 +16,11 @@ runtimesManifest:
     default: true
   python:
   - kind: "python"
+    image: "python2action"
+  - kind: "python:2"
+    image: "python2action"
+  - kind: "python:3"
+    image: "pythonaction"
     default: true
   swift:
   - kind: "swift"
--- a/core/actionProxy/Dockerfile
+++ b/core/actionProxy/Dockerfile
@@ -1,13 +1,14 @@
-# Dockerfile for docker skeleton (useful for running blackbox binaries, scripts, or python actions).
-FROM python:2.7.12-alpine
+# Dockerfile for docker skeleton (useful for running blackbox binaries, scripts, or Python 3 actions) .
+FROM python:3.6.1-alpine
 
-# Upgrade and install basic Python dependencies
+# Upgrade and install basic Python dependencies.
 RUN apk add --no-cache bash \
  && apk add --no-cache --virtual .build-deps \
         bzip2-dev \
         gcc \
         libc-dev \
-  && pip install --no-cache-dir gevent==1.1.2 flask==0.11.1 \
+  && pip install --upgrade pip setuptools \
+  && pip install --no-cache-dir gevent==1.2.1 flask==0.12 \
   && apk del .build-deps
 
 ENV FLASK_PROXY_PORT 8080
--- a/core/actionProxy/actionproxy.py
+++ b/core/actionProxy/actionproxy.py
@@ -235,12 +235,11 @@ def run():
 
     if runner.verify():
         try:
-            (code, result) = runner.run(args,
-                                        runner.env(message if message else {}))
+            code, result = runner.run(args, runner.env(message or {}))
             response = flask.jsonify(result)
             response.status_code = code
         except Exception as e:
-            response = flask.jsonify({'error': 'Internal error.'})
+            response = flask.jsonify({'error': 'Internal error. {}'.format(e)})
             response.status_code = 500
     else:
         response = flask.jsonify({'error': 'The action failed to locate '
--- a/core/python2Action/Dockerfile
+++ b/core/python2Action/Dockerfile
@@ -1,0 +1,36 @@
+# Dockerfile for Python 2 actions, similar to the Python 3-based core/pythonAction
+FROM python:2.7.12-alpine
+
+# Upgrade and install basic Python dependencies
+RUN apk add --no-cache \
+        bash \
+        bzip2-dev \
+        gcc \
+        libc-dev \
+        libxslt-dev \
+        libxml2-dev \
+        libffi-dev \
+        openssl-dev \
+        python-dev
+
+# Install common modules for python
+RUN pip install --no-cache-dir --upgrade pip setuptools \
+ && pip install --no-cache-dir \
+        gevent==1.1.2 \
+        flask==0.11.1 \
+        beautifulsoup4==4.5.1 \
+        httplib2==0.9.2 \
+        kafka_python==1.3.1 \
+        lxml==3.6.4 \
+        python-dateutil==2.5.3 \
+        requests==2.11.1 \
+        scrapy==1.1.2 \
+        simplejson==3.8.2 \
+        virtualenv==15.1.0 \
+        twisted==16.4.0
+
+ENV FLASK_PROXY_PORT 8080
+
+RUN mkdir -p /pythonAction
+ADD actionproxy.py /pythonAction/
+CMD ["python"]
--- a/core/python2Action/Dockerfile
+++ b/core/python2Action/Dockerfile
@@ -33,4 +33,6 @@ ENV FLASK_PROXY_PORT 8080
 
 RUN mkdir -p /pythonAction
 ADD actionproxy.py /pythonAction/
-CMD ["python"]
+ADD pythonrunner.py /pythonAction/
+
+CMD ["/bin/bash", "-c", "cd pythonAction && python -u pythonrunner.py"]
--- a/core/python2Action/build.gradle
+++ b/core/python2Action/build.gradle
@@ -1,17 +1,19 @@
 ext.dockerImageName = 'python2action'
 apply from: '../../gradle/docker.gradle'
 distDocker.dependsOn 'copyFiles'
-distDocker.finalizedBy('rmFiles')
+distDocker.finalizedBy 'rmFiles'
+
+def runners = files(
+    new File(project(':core:actionProxy').projectDir, 'actionproxy.py'),
+    new File(project(':core:pythonAction').projectDir, 'pythonrunner.py')
+)
 
 task copyFiles(type: Copy) {
-    from('../actionProxy/actionproxy.py') {
-    }
-    from('../pythonAction/pythonrunner.py') {
-    }
+    from runners
     into '.'
 }
 
 task rmFiles(type: Delete) {
-    delete 'actionproxy.py', 'pythonrunner.py'
+    delete runners.collect { it.getName() }
 }
 
--- a/core/python2Action/build.gradle
+++ b/core/python2Action/build.gradle
@@ -1,0 +1,17 @@
+ext.dockerImageName = 'python2action'
+apply from: '../../gradle/docker.gradle'
+distDocker.dependsOn 'copyFiles'
+distDocker.finalizedBy('rmFiles')
+
+task copyFiles(type: Copy) {
+    from('../actionProxy/actionproxy.py') {
+    }
+    from('../pythonAction/pythonrunner.py') {
+    }
+    into '.'
+}
+
+task rmFiles(type: Delete) {
+    delete 'actionproxy.py', 'pythonrunner.py'
+}
+
--- a/core/pythonAction/Dockerfile
+++ b/core/pythonAction/Dockerfile
@@ -8,20 +8,20 @@ RUN apk add --no-cache \
         libxslt-dev \
         libxml2-dev \
         libffi-dev \
-        openssl-dev \
-        python-dev
+        openssl-dev
 
 # Install common modules for python
 RUN pip install \
-    beautifulsoup4==4.5.1 \
-    httplib2==0.9.2 \
-    kafka_python==1.3.1 \
-    lxml==3.6.4 \
-    python-dateutil==2.5.3 \
-    requests==2.11.1 \
-    scrapy==1.1.2 \
-    simplejson==3.8.2 \
-    twisted==16.4.0
+    beautifulsoup4==4.5.3 \
+    httplib2==0.10.3 \
+    kafka_python==1.3.2 \
+    lxml==3.7.3 \
+    python-dateutil==2.6.0 \
+    requests==2.13.0 \
+    scrapy==1.3.3 \
+    simplejson==3.10.0 \
+    virtualenv==15.1.0 \
+    twisted==17.1.0
 
 ENV FLASK_PROXY_PORT 8080
 
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -260,48 +260,103 @@ The following packages are available to be used in the Node.js 6.9.1 environment
 
 ## Python actions
 
-Python actions are executed by default using Python 2.7.12.
-In addition to the standard Python library, the following packages are also available for use by Python actions.
+Python actions are executed by default using Python 3.6.1.
+In addition to the Python 3 standard library, the following packages are also available for use by Python actions.
 
-- attrs v16.1.0
+- aiohttp v1.3.3
+- appdirs v1.4.3
+- asn1crypto v0.21.1
+- async-timeout v1.2.0
+- attrs v16.3.0
 - beautifulsoup4 v4.5.1
-- cffi v1.7.0
-- click v6.6
-- cryptography v1.5
-- cssselect v0.9.2
-- enum34 v1.1.6
-- flask v0.11.1
-- gevent v1.1.2
-- greenlet v0.4.10
+- cffi v1.9.1
+- chardet v2.3.0
+- click v6.7
+- cryptography v1.8.1
+- cssselect v1.0.1
+- Flask v0.12
+- gevent v1.2.1
+- greenlet v0.4.12
 - httplib2 v0.9.2
-- idna v2.1
-- ipaddress v1.0.16
+- idna v2.5
 - itsdangerous v0.24
-- jinja2 v2.8
+- Jinja2 v2.9.5
+- kafka-python v1.3.1
 - lxml v3.6.4
-- kafka_python v1.3.1
-- markupsafe v0.23
-- parsel v1.0.3
-- pyasn1 v0.1.9
+- MarkupSafe v1.0
+- multidict v2.1.4
+- packaging v16.8
+- parsel v1.1.0
+- pyasn1 v0.2.3
 - pyasn1-modules v0.0.8
-- pycparser v2.14
-- pydispatcher v2.0.5
-- pyopenssl v16.1.0
+- pycparser v2.17
+- PyDispatcher v2.0.5
+- pyOpenSSL v16.2.0
+- pyparsing v2.2.0
 - python-dateutil v2.5.3
 - queuelib v1.4.2
 - requests v2.11.1
-- scrapy v1.1.2
+- Scrapy v1.1.2
 - service-identity v16.0.0
 - simplejson v3.8.2
 - six v1.10.0
-- twisted v16.4.0
-- w3lib v1.15.0
-- werkzeug v0.11.10
-- zope.interface v4.3.1
+- Twisted v16.4.0
+- w3lib v1.17.0
+- Werkzeug v0.12
+- yarl v0.9.8
+- zope.interface v4.3.3
+
+## Python 2 actions
+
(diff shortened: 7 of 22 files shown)


rabbah:
@cclauss thanks for the contribution!
Do you have an Apache iCLA; All contributors must sign and submit an Apache CLA (Contributor License Agreement)?

Instructions on how to do this can be found here: http://www.apache.org/licenses/#clas.
Once submitted, you will receive a confirmation email from the Apache Software Foundation (ASF) and be added to the following list: http://people.apache.org/unlistedclas.html.

Project committers will use this list to verify pull requests (PRs) come from contributors that have signed a CLA.

More info on CONTRIBUTING.md file on the root of the project

rabbah:
Why did you revert in the last commit? Travis failure?

cclauss:
Hi @rabbah Yes...  I think the other changes made are correct and useful but that last change broke the build and I could not untangle the Scala runs Python error messages in the Travis report: https://travis-ci.org/openwhisk/openwhisk/builds/204927740

rabbah:
Likely the travis failure was an indicator of other failures to come:

The following failed in Jenkins (note that Travis doesn't run all the tests because we've run into timeout issues and took a blunt approach to getting under the cap):
```
    actionContainers.PythonActionContainerTests.pythonaction should run and report an error for script not returning a json object
    actionContainers.PythonActionContainerTests.pythonaction should run a python script
    actionContainers.PythonActionContainerTests.pythonaction should run a python script and confirm expected environment variables
    actionContainers.PythonActionContainerTests.pythonaction should support actions using non-default entry points
    actionContainers.PythonActionContainerTests.pythonaction should support zip-encoded action using non-default entry points
    actionContainers.PythonActionContainerTests.pythonaction should report error if zip-encoded action does not include required file
    actionContainers.PythonActionContainerTests.pythonaction should handle unicode in source, input params, logs, and result
    actionContainers.PythonActionContainerTests.pythonaction should return on action error when action fails
    actionContainers.PythonActionContainerTests.pythonaction should log compilation errors
    actionContainers.PythonActionContainerTests.pythonaction should support application errors
    actionContainers.PythonActionContainerTests.pythonaction should error when importing a not-supported package
    actionContainers.PythonActionContainerTests.pythonaction should be able to import additional packages as installed in the image
    system.basic.CLIPythonTests.Native Python Action should invoke an action and get the result
    system.basic.CLIPythonTests.Native Python Action should invoke an action with a non-default entry point
    system.basic.CLIPythonTests.Native Python Action should invoke an action from a zip file with a non-default entry point
    system.basic.CLIPythonTests.Native Python Action should invoke an action and confirm expected environment is defined
    system.basic.CLIPythonTests.Native Python Action should invoke an invalid action and get error back
    whisk.core.limits.MaxActionDurationTests.node-, python, and java-action should run up to the max allowed duration
```

cclauss:
> The following failed in Jenkins

Is that for the commit that I backed out or is that for this pull request as it currently stands?  I was able to build the container locally and it starts OK.  My sense is that there is some Python code that is not Python 3 compatible and an exception is being thrown which is preventing a dict from being returned.

rabbah:
All the commits (https://github.com/openwhisk/openwhisk/compare/master...rabbah:pr-1903?expand=1).

rabbah:
@cclauss try this? https://medium.com/openwhisk/advanced-debugging-of-openwhisk-actions-518414636932#.2022ukn6a

cclauss:
@rabbah please see https://github.com/openwhisk/openwhisk/pull/1944  We won't get to Python3 if invoke.py is not yet Python 3 compatible.

rabbah:
Ok will have a look - this script is a debugging aid (it is not used by the system fwiw).

core/pythonAction/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Dockerfile for python actions, overrides and extends ActionRunner from actionProxy
-FROM dockerskeleton
+FROM python:3-alpine
```
rabbah: this should remain as dockerskeleton (it allows us to have one base image for several different runtimes which we can pre-warm for example); can we just change the image in actionProxy instead?

rabbah: https://github.com/openwhisk/openwhisk/blob/master/core/actionProxy/Dockerfile#L2

core/pythonAction/Dockerfile
```diff
@@ -8,24 +8,23 @@ RUN apk add --no-cache \
...
 RUN pip install \
-    beautifulsoup4==4.5.1 \
-    httplib2==0.9.2 \
-    kafka_python==1.3.1 \
-    lxml==3.6.4 \
-    python-dateutil==2.5.3 \
-    requests==2.11.1 \
-    scrapy==1.1.2 \
-    simplejson==3.8.2 \
-    twisted==16.4.0
```
rabbah: need to update these versions https://github.com/openwhisk/openwhisk/blob/master/docs/reference.md#python-actions

rabbah:
can we combine this and #1957?

cclauss:
Sure.  Does "we" mean me or does "we" mean you?   ;-)

I will work on the reference.md#python-actions ...

rabbah:
hehe I meant you or me - since it's one file and related to this pr it made sense to me. Do you mind doing it?

core/pythonAction/Dockerfile
cclauss: Done.

rabbah: Thank you!

core/actionProxy/Dockerfile
```diff
@@ -1,13 +1,13 @@
...
+FROM python:2.7.13-alpine
 
 # Upgrade and install basic Python dependencies
 RUN apk add --no-cache bash \
  && apk add --no-cache --virtual .build-deps \
         bzip2-dev \
         gcc \
         libc-dev \
-  && pip install --no-cache-dir gevent==1.1.2 flask==0.11.1 \
+  && pip install --no-cache-dir gevent==1.2.1 flask==0.12 \
```
rabbah: at line 22, should use python3 also?

rabbah:
PG in progress for this consolidated commit: https://github.com/rabbah/openwhisk/commit/1726b458aeebef01e728cebfbd175942e41aa039 (based on 
	aa8c7e3).

core/actionProxy/Dockerfile
cclauss: The official Python 2 docker image does not contain Python 3 and vice versa so I had to create a [mind-meld between the two](https://github.com/docker-library/python/issues/178).  So`dockerskeleton`is now a single Docker image with both Python 2.7.13 and Python 3.6.0 available.  We need to be explicit about using python2 vs. python3 as well as pip2 vs. pip3.

cclauss:
Layering`python:2.7.13-alpine`on top of`python:3.6.0-alpine`takes 5 minutes (and adds lots of chatter) to the Travis process and makes openwhisk:dockerskeleton more than twice as big as`python:2.7.13-alpine`.  This might work on a temporary basis but is not optimal for the long term.

```sh
$ docker images
python-2-and-3      146.0 MB
3.6.0-alpine               88.6 MB
2.7.13-alpine              71.5 MB
```

Perhaps we should pull the building of openwhisk:dockerskeleton out of the Travis process or at least put that into the[`before_install`or`before_script`sections](https://github.com/openwhisk/openwhisk/blob/master/.travis.yml#L31-L35) to keep the chatter hidden.  Is there any way to use [Travis build matrix](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix) to speed up the builds?

core/actionProxy/Dockerfile
cclauss: mind-meld undone.

cclauss:
Ready to review/merge.

core/python2Action/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+  && apk del .build-deps
+
+RUN apk add --no-cache \
+        bzip2-dev \
+        gcc \
+        libc-dev \
+        libxslt-dev \
+        libxml2-dev \
+        libffi-dev \
+        openssl-dev
```
markusthoemmes: This looks redundant to the command above.

core/python2Action/Dockerfile
```diff
@@ -0,0 +1,41 @@
+# Dockerfile for Python 2 actions, mirrors the Python 3-based pythonAction
+
+FROM python:2.7.13-alpine
```
markusthoemmes: Should build on dockerskeleton.

core/python2Action/pythonrunner.py
```diff
@@ -0,0 +1,87 @@
+"""Executable Python script for running Python actions.
```
rabbah: can we get away with just one python runner and two different docker builds?

cclauss: Fixed in https://github.com/openwhisk/openwhisk/pull/1903/files#diff-01c3377ed05fef923a1cc0d1b2257e1dR39

core/python2Action/Dockerfile
cclauss: core/python2Action is meant to [temporary](https://github.com/openwhisk/openwhisk/issues/1184#issuecomment-245334998) replacement core/actionProxy (aka dockerskeleton) and core/pythonAction as we transition them from from Python 2.7 to Python 3.6.  We  can optimize the calls to `apk add` in new core/python2Proxy as you suggest but that will increase the cognitive load and chance of errors for those who need to keep core/python2Proxy in sync with changes made to core/actionProxy and core/pythonProxy.

core/python2Action/Dockerfile
cclauss: The [idea](https://github.com/openwhisk/openwhisk/issues/1184#issuecomment-285323326) here is to make dockerskeleton into a **Python 3** container.  The plan is to migrate both the docker images core/actionProxy (aka dockerskeleton) and core/pythonAction from Python 2.7 to Python 3.6.  The new core/python2Action image is meant to be our _temporary_ method of continuing to run Python 2 jobs after that transition to Python 3.  The alternative was to create a single core/actionProxy that contains _both_ Python 2 and Python 3 in a single image.  That [was tried -- it works -- but was rejected](https://github.com/openwhisk/openwhisk/pull/1903#issuecomment-284717687) because the resulting container is too big for daily use.

core/actionProxy/actionproxy.py
```diff
@@ -219,9 +219,9 @@ def init():
 
 @proxy.route('/run', methods=['POST'])
 def run():
-    def error():
+    def error(msg=''):
```
rabbah: Is msg used?

core/actionProxy/Dockerfile
```diff
@@ -1,13 +1,14 @@
...
+FROM python:3.6.0-alpine
 
 # Upgrade and install basic Python dependencies
 RUN apk add --no-cache bash \
  && apk add --no-cache --virtual .build-deps \
         bzip2-dev \
         gcc \
         libc-dev \
-  && pip install --no-cache-dir gevent==1.1.2 flask==0.11.1 \
+  && pip install --upgrade pip setuptools \
```
rabbah: Should setup tools remain part of python dockerfile? What do we gain by putting it here?

docs/reference.md
```diff
@@ -260,48 +260,103 @@ The following packages are available to be used in the Node.js 6.9.1 environment
...
+
+## Python 2 actions
+
+For a **limited time** support will continue for Python 2 actions executed using Python 2.7.12.
+In addition to the Python 2.7 standard library, the following packages are also available for use by Python actions.
+
+**Note**: Python 2 is deprecated in Open Whisk, please migrate all your Python actions to use Python 3 version 3.6+.
 
-- attrs v16.1.0
+- appdirs v1.4.3
```
rabbah: Should we leave python2 as is wrt installed packages?

core/actionProxy/Dockerfile
cclauss: There are [frequent ongoing changes](https://github.com/pypa/setuptools/releases) in both pip and in setuptools to make sure that Python packaging is ready for the shift for [PyPI](https://pypi.python.org/pypi) to its more modern replacement, [warehouse](https://pypi.org/).  We want to ensure that dockerskeleton can properly install both legacy packages like twisted and more modern packages that are built for warehouse.  This will become even more important as features like PR 1940 will allow action developers to layer their own packages (or package versions) on top of the runtime.

docs/reference.md
cclauss: In core/python2Action/Dockerfile we pin the same packages to the same versions as the **old version** of core/pythonAction/Dockerfile.  Each of those packages may in turn import its own dependencies to form the full dependency graph.  Sometimes their setup.py or requirements.txt files specifies an exact version, sometimes a version range, and sometimes does not specify a version at all.  When the same module is requested by multiple parents, tools like pip and setuptools try attempt to do dependency negotiation to reach a compromise on what version is imported.
All of this means that not all of the packages in the dependency graph are pinned.  This in turn means that the dockerskeleton that I built one month ago might not have the same version of all packages as the dockerskeleton that I build today even if they were both built using the exact same Dockerfile.  We could chose to pin 100% of all packages in the dependency graph but that has several disadvantages.  Pip is not longer able to do dependency negotiation.  An upgrade to Jinja2 might require an upgrade to [its depandancy markupsafe](https://github.com/pallets/jinja/blob/master/setup.py#L72) but if we pin all dependencies then we would need to catch these incompatibilities manually.  Also, if Jinja dropped a dependency altogether, we might keep that pinned dependency in the build even though it is never used.

All of the above means that the package version numbers in `docs/reference.md` are just a point-in-time statement, not a guarantee.  Allowing pip & setuptools to dependency negotiation has great advantages.

core/actionProxy/actionproxy.py
```diff
@@ -235,12 +235,11 @@ def error():
...
         try:
-            (code, result) = runner.run(args,
-                                        runner.env(message if message else {}))
+            code, result = runner.run(args, runner.env(message or {}))
             response = flask.jsonify(result)
             response.status_code = code
         except Exception as e:
-            response = flask.jsonify({'error': 'Internal error.'})
+            response = flask.jsonify({'error': 'Internal error. {}'.format(e)})
             response.status_code = 500
```
rabbah: is the format at the end correct?

core/python2Action/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+        requests==2.11.1 \
+        scrapy==1.1.2 \
+        simplejson==3.8.2 \
+        virtualenv==15.1.0 \
+        twisted==16.4.0
+
+ENV FLASK_PROXY_PORT 8080
+
+RUN mkdir -p /pythonAction
+ADD ../pythonAction/pythonrunner.py /pythonAction/
```
rabbah: This doesn't work (relative paths must be forward paths) in docker. You'll have to copy the file in the gradle build script then delete it. See swift action for an example.

tests/src/test/scala/actionContainers/ActionProxyContainerTests.scala
```diff
@@ -56,9 +56,10 @@ class ActionProxyContainerTests extends BasicActionRunnerTests with WskActorSyst
 
         val python = """
                 |#!/usr/bin/env python
+                |from __future__ import print_function
                 |import sys
-                |print 'hello stdout'
-                |print >> sys.stderr, 'hello stderr'
+                |print('hello stdout')
```
rabbah: When you rebase this change should go away.

cclauss: Rebase done.

core/python2Action/pythonrunner.py
cclauss: Docker does not allow you to get local resources that are not in the current context (i.e. the build dir).  https://travis-ci.org/openwhisk/openwhisk/builds/213370715#L3568  Reduplicating pythonrunner.py

ansible/group_vars/all
```diff
@@ -16,6 +16,7 @@ runtimesManifest:
     default: true
   python:
   - kind: "python"
```
rabbah: kind `python` will resolve to the image `pythonaction`
we could make the image mapping explicit here - or fix it in the ExecManifest. 
making the action to image name mapping explicit here is a good idea and previously suggested by others (@ddragosd @glikson iirc).

rabbah: general comment (not specific to this pr):
we might consider adding python:2 as well, make it explicit. 
we should adopt it as a policy, a naked family name is same as :default or disallow altogether.
without versions, we're asking for more headaches in the future.

fyi @bjustin-ibm @jasonpet

core/python2Action/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+        requests==2.11.1 \
+        scrapy==1.1.2 \
+        simplejson==3.8.2 \
+        virtualenv==15.1.0 \
+        twisted==16.4.0
+
+ENV FLASK_PROXY_PORT 8080
+
+RUN mkdir -p /pythonAction
+ADD pythonrunner.py /pythonAction/
```
rabbah: don't you also need to add `actionproxy.py`?

core/python2Action/build.gradle
```diff
@@ -0,0 +1,25 @@
+ext.dockerImageName = 'python2action'
+apply from: '../../gradle/docker.gradle'
+distDocker.dependsOn 'copyRunner'
+distDocker.finalizedBy('rmRunner')
+
+task copyProxy(type: Copy) {
+    from '../actionProxy/actionproxy.py'
```
rabbah: how about just one copy task and one clean up task?

core/python2Action/Dockerfile
rabbah: actually - you need to copy action proxy, not python runner.

core/python2Action/build.gradle
rabbah: see comment above.

tests/src/test/scala/actionContainers/Python2ActionContainerTests.scala
```diff
@@ -0,0 +1,37 @@
...
+class Python2ActionContainerTests extends PythonActionContainerTests with WskActorSystem {
+
+    /** indicates if strings in python are unicode by default (i.e., python3 -> true, python2.7 -> false) */
+    override val pythonStringAsUnicode = false
+
+    override def withActionContainer(env: Map[String, String] = Map.empty)(code: ActionContainer => Unit) = {
+        withContainer("python2action", env)(code)
+    }
+
+    behavior of "python2action"
```
rabbah: you can remove this - it wont have the effect you think.

tests/src/test/scala/actionContainers/PythonActionContainerTests.scala
```diff
@@ -137,13 +140,23 @@ class PythonActionContainerTests extends BasicActionRunnerTests with WskActorSys
...
     it should "handle unicode in source, input params, logs, and result" in {
         val (out, err) = withActionContainer() { c =>
-            val code = """
-                |def main(dict):
-                |    sep = dict['delimiter']
-                |    str = sep + " ☃ ".decode('utf-8') + sep
-                |    print(str.encode('utf-8'))
-                |    return {"winter" : str }
-            """.stripMargin
+            val code = if (pythonStringAsUnicode) {
```
rabbah: are you sure you rebased to master?

ansible/group_vars/all
rabbah: For reference: https://github.com/openwhisk/openwhisk/blob/master/common/scala/src/main/scala/whisk/core/entity/Exec.scala#L167

Hence we need a test that confirms a kind `python` action will run as python:2 and otherwise `python:3` as python:3.

The manifest allows you to name the image - I'd have to look further to confirm but the the images that are prefixed with `openwhisk/` are recognized by the system and are rewritten to use an internal registry.

core/python2Action/Dockerfile
cclauss: On the first point, you are correct.  pythonrunner.py imports  actionproxy so we need actionproxy.  I will ADD.

On the second point, I think we need both but will experiment to make sure.

rabbah: Ah - right - you do need to copy python runner too. Forgot that it's the same as p3.

ansible/group_vars/all
cclauss: Just to clarify, our objective is that `kind: python:2` is the ONLY way to get python2Action:
```
kind: python,         image_name: pythonAction   # i.e. Python 3
kind: python:default, image_name: pythonAction   # i.e. Python 3
kind: python:2,       image_name: python2Action  # i.e. Python 2
kind: python:3,       image_name: pythonAction   # i.e. Python 3
```

rabbah: `kind:python` must be python2action because all pre-existing python actions are of this kind and hence may not work with python3. Hence to preserve the behavior, we have to map `python` to 2, not 3.

so the changes are:

```
kind: python,         image_name: python2Action
kind: python:2,       image_name: python2Action
kind: python:3,       image_name: python3Action, default: true
```

cclauss:
Ready for review.

rabbah:
pg3/452 🚀

rabbah:
pg3/452 approved.

rabbah:
pg2/1306 for the latest changes 🚀

rabbah:
@bjustin-ibm @markusthoemmes your 👀 are appreciated.

core/python2Action/build.gradle
```diff
@@ -0,0 +1,17 @@
...
+    from('../actionProxy/actionproxy.py') {
+    }
+    from('../pythonAction/pythonrunner.py') {
+    }
+    into '.'
+}
+
+task rmFiles(type: Delete) {
+    delete 'actionproxy.py', 'pythonrunner.py'
+}
```
markusthoemmes: As per an earlier comment, this should be solved more programmatically with information we already have:

```
def runners = files(
    new File(project(':core:actionProxy').projectDir, 'actionproxy.py'),
    new File(project(':core:pythonAction').projectDir, 'pythonrunner.py')
)
task copyFiles(type: Copy) {
    from runners
    into '.'
}

task rmFiles(type: Delete) {
    delete runners.collect { it.getName() }
}
```

The *finalizedBy* bit should leave out the parentheses for consistency

rabbah:
pg2/1306 🔵

rabbah:
Picking up in #2070.

PR #1911, opened 2017-02-26
Flatten Exec hiearchy

rabbah:
This PR removes some unnecessary distinctions in the `exec` hierarchy - making the treatments of node, python, and swift actions the same.

PG3/303.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/Exec.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Exec.scala
@@ -74,42 +74,35 @@ sealed abstract class CodeExec[T <% SizeConversion](kind: String) extends Exec(k
     val pull = false
 
     // Whether the code is stored in a text-readable or binary format.
-    def binary: Boolean = false
+    // the binary bit may be read from the database but currently it is always computed
+    // when the "code" is moved to an attachment this may get changed to avoid recomputing
+    // the binary property
+    lazy val binary = {
+        code match {
+            case s: String => Exec.isBinaryCode(s)
+            case _         => false
+        }
+    }
 
     override def size = code.sizeInBytes
 }
 
-sealed abstract class NodeJSAbstractExec(kind: String) extends CodeExec[String](kind) {
-    val main: Option[String]
+protected[core] abstract class CodeExecAsString(kind: String) extends CodeExec[String](kind)
 
-    override val entryPoint = main
-    // the binary bit may be read from the database but currently it is always computed
-    // when the "code" is moved to an attachment this may get changed to avoid recomputing
-    // the binary property
-    override lazy val binary = Exec.isBinaryCode(code)
-}
+protected[core] case class NodeJSExec(code: String, override val entryPoint: Option[String]) extends CodeExecAsString(Exec.NODEJS)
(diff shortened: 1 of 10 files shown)


common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -74,44 +74,37 @@ sealed abstract class CodeExec[T <% SizeConversion](kind: String) extends Exec(k
...
+            case _         => false
+        }
+    }
+
+    override def size = code.sizeInBytes
 }
 
-protected[core] case class NodeJSExec(code: String, main: Option[String]) extends NodeJSAbstractExec(Exec.NODEJS)
-protected[core] case class NodeJS6Exec(code: String, main: Option[String]) extends NodeJSAbstractExec(Exec.NODEJS6)
+protected[core] abstract class CodeExecAsString(kind: String) extends CodeExec[String](kind)
```
markusthoemmes: Is the extra type necessary? `CodeExecAsString` and `CodeExec[String]` read pretty much the same I think.

rabbah: See WhiskAction for the reason - wanted to avoid a typeable or typecase when matching on the exec type since it's generic.

markusthoemmes: Got it, you're right!

tools/cli/go-whisk-cli/commands/action.go
```diff
@@ -523,10 +522,10 @@ func parseAction(cmd *cobra.Command, args []string) (*whisk.Action, error) {
       }
     }
 
-    // For zip-encoded NodeJS action, the code needs to be base64-encoded.
+    // For zip-encoded actions, the code needs to be base64-encoded.
     // We reach this point if the kind has already be determined. Since the extension is not js,
     // this means the kind was specified explicitly.
-    if ext == ".zip" && (strings.HasPrefix(action.Exec.Kind, "nodejs") || action.Exec.Kind == "blackbox") {
+    if ext == ".zip" {
```
rabbah: @dubeejw are you OK with this change?

dubeejw: See the else block starting on line 497. It seems like we wouldn't need a requirement of specifying a kind flag with .zip extensions based on your changes.

rabbah: I don't follow. On 497 the block asserts kind is required with zip. The extra conjunctions I removed eliminate the restriction that zip is allowed only for nodejs and docker.

dubeejw: Does the backend still need to know the kind for zip files? I'm fine with the line you changed. Just curious.

rabbah: Yes - the API requires that `exec` includes `kind`.

```
   val kindField = obj.getFields("kind") match {
                case Seq(JsString(k)) => k.trim.toLowerCase
                case _                => throw new DeserializationException("'kind' must be a string defined in 'exec'")
            }
```

PR #1913, opened 2017-02-27
docker-machine  has no bash, we actually don't need it here. 

domdom82:
also change ignore_errors to failed_when false to beautify logs.


domdom82:
PG2 1106

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/efa04b9cb12460ed7179b1920e49dc18025e4f70.

PR #1914, opened 2017-02-27
Removing Remote Entitlement

markusthoemmes:
This removes remote entitlement in its entirety and also removes the questionable "feature" of doing things cross-namespaces. Namespace switching is hereby explicit by setting the correct API key when authenticating with the OpenWhisk backend.

TBD:

- [ ] repurpose /namespaces to a "whoami" and tighten respective tests


core/controller/src/main/scala/whisk/core/controller/Namespaces.scala
```diff
@@ -129,18 +113,8 @@ trait WhiskNamespacesApi
...
-                complete(OK, namespaces)
-            case Failure(r: RejectRequest) =>
-                logging.info(this, s"[GET] namespaces failed: ${r.message}")
-                terminate(r.code, r.message)
-            case Failure(t) =>
-                logging.error(this, s"[GET] namespaces failed: ${t.getMessage}")
-                terminate(InternalServerError)
-        }
+    private def getNamespaces(user: Identity)(implicit transid: TransactionId) = {
+        complete(OK, List(user.namespace))
```
rabbah: should this remain a list?

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -84,12 +83,12 @@ abstract protected[controller] class RestAPIVersion(
 protected[controller] object RestAPIVersion_v1 {
     def requiredProperties =
         WhiskConfig.whiskVersion ++
-            WhiskServices.requiredProperties ++
+            WhiskAuthStore.requiredProperties ++
```
rabbah: consolidating the required properties is ok - since the api unit tests use the common controller test trait which unifies the required properties also.

core/controller/src/main/scala/whisk/core/controller/Namespaces.scala
markusthoemmes: In the foreseeable future probably not. Didn't want too much of a semantic change in here though. That way, behaviour is at least theoretically preserved, as in: You can access every namespace in that list with that key.

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
markusthoemmes: I also thought it makes sense to have the properties required local to where the corresponding objects are created, which is in RestAPIs

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -688,9 +688,9 @@ class WskBasicTests
 
     behavior of "Wsk Namespace CLI"
 
-    it should "list namespaces" in {
+    it should "return a list of exactly one namespace" in {
         wsk.namespace.list().
-            stdout should include regex ("@|guest")
+            stdout.lines should have size 2 // headline + namespace
```
rabbah: Can't we use wsk admin to retrieve what this value should be?

markusthoemmes: No, `wskadmin` is not to be used in a system level test. Pressure on the `whoami` endpoint rises.

rabbah:
Good catch.

markusthoemmes:
PG3 311

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -221,23 +220,8 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
                 resource.collection.implicitRights(user, defaultNamespaces, right, resource) flatMap {
                     case true => Future.successful(true)
```
rabbah: is this better as an if/else?

tests/src/system/basic/WskBasicTests.scala
rabbah: fair point.

tests/src/system/basic/WskBasicTests.scala
```diff
@@ -728,7 +728,7 @@ class WskBasicTests
                     wsk.activation.get(activation.activationId, fieldFilter = Some("name")).stdout should include(s"""$successMsg name\n"$name"""")
                     wsk.activation.get(activation.activationId, fieldFilter = Some("version")).stdout should include(s"""$successMsg version\n"0.0.1"""")
                     wsk.activation.get(activation.activationId, fieldFilter = Some("publish")).stdout should include(s"""$successMsg publish\nfalse""")
-                    wsk.activation.get(activation.activationId, fieldFilter = Some("subject")).stdout should include regex (s"""(?i)$successMsg subject\n$ns_regex_list""")
```
rabbah: why is this removed? `$ns_regex_lis`

markusthoemmes: Because it's not true. The subject is not necessarily equal to the namespace used.

junoyoon:
@markusthoemmes 

Is it planned to externalize the entitlement in the other way? 
We have been working on integrate openwhisk with [Keystone](https://docs.openstack.org/developer/keystone/)  using this feature.

rabbah:
If we need to restore the entitlement host through whisk config to enable remote entitlement that's reasonable. Really what this cleanup was about was to remove some entanglement between the authorization and namespacing that belongs in a proper authorizarion layer and not the core. 

Once you have an integration with keystone, I suggest we look at what changes you have/are needed in the core and we can review that via a pr. Does that make sense?

markusthoemmes:
It's also worth mentioning that our very own integration with Bluemix works completely asynchronous now. It only cares for the entries in the subjects database and is not entangled with the controller whatsoever.

Maybe you can achieve something similar with your Keystone integration?

junoyoon:
@markusthoemmes @rabbah  
Thanks for explanation.  If I really need some changes on the core, I'll propose PR. :-)

csantanapr:
Related to #1107

PR #1919, opened 2017-02-28
remove intermediate shell to make shutdown hook great again.

domdom82:
@perryibm the shutdown hook did not work because SIGTERM got sent to the shell that runs the invoker JVM instead of the JVM directly. @sven-lange-last and I replaced it with exec to make the JVM the top level process in the invoker container.

In case we need similar behavior in the controller, it can also be applied to the CMD stanza of a Dockerfile.


domdom82:
PG2 1110

domdom82:
PG1 1225

perryibm:
PG1-1225 and 1226 both failed.  Retrying at PG1-1228.

markusthoemmes:
You'll need to rebase this.

domdom82:
PG1 1228 green

Issue #1042, opened 2016-08-12
implement internal kafka roundtrip latency test within controller


rabbah:
The controller should measure and monitor its observed kafka round trip latency (by posting to a dedicated topic and reading from it) and adjusting active ack/split phase monitoring accordingly. This allows the controller to back off active ack if there's a kafka anomaly detected.

markusthoemmes:
This is already done with the newest health protocol, isn't it?

rabbah:
The health protocol informs the controller of invoker health - which might as well be kafka health if there are no messages received. I intended in this issue to also tie that to the active ack short circuiting and fall over to the datastore for activations. You're pointing out that the period for which this would be an issue with the new health protocol is much shorter than the previous 30 seconds. I agree.

PR #1922, opened 2017-02-28
Don't use Invoker if it is UnHealthy

cbickel:
This PR adds the state "UnHealthy" to InvokerHealth. 
If an Invoker is Unhealthy, the loadbalancer won't dispatch user actions to it.
It becomes Unhealthy, if too many activations failed with the status > 2 (WhiskError).
When it is Healthy, some test actions (`whisk.system/utils/date`) will be sent to the invoker.
If enough of these test messages pass again, the Invoker turns to Healthy again and user activations are dispatched to it.

In addition this PR moves the Handling of Kafka from the LoadBalancerService to the InvokerSupervision.

### DIFF
--- a/common/scala/build.gradle
+++ b/common/scala/build.gradle
@@ -28,6 +28,7 @@ dependencies {
     compile 'commons-codec:commons-codec:1.9'
     compile 'commons-io:commons-io:2.4'
     compile 'commons-logging:commons-logging:1.2'
+    compile 'commons-collections:commons-collections:3.2.2'
     compile 'org.apache.zookeeper:zookeeper:3.4.6'
     compile 'org.apache.kafka:kafka-clients:0.10.0.0'
     compile 'org.apache.httpcomponents:httpclient:4.4.1'
--- a/common/scala/src/main/scala/whisk/common/RingBuffer.scala
+++ b/common/scala/src/main/scala/whisk/common/RingBuffer.scala
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.common
+
+import org.apache.commons.collections.buffer.CircularFifoBuffer
+
+object RingBuffer {
+    def apply[T](size: Int) = new RingBuffer[T](size)
+}
+
+class RingBuffer[T](size: Int) {
+    private val inner = new CircularFifoBuffer(size)
+
+    def add(el: T) = inner.add(el)
+
+    def toList() = inner.toArray().asInstanceOf[Array[T]].toList
+}
--- a/common/scala/src/main/scala/whisk/core/connector/Message.scala
+++ b/common/scala/src/main/scala/whisk/core/connector/Message.scala
@@ -88,7 +88,8 @@ object ActivationMessage extends DefaultJsonProtocol {
  */
 case class CompletionMessage(
     override val transid: TransactionId,
-    response: WhiskActivation)
+    response: WhiskActivation,
+    invoker: String)
     extends Message {
 
     override def serialize = CompletionMessage.serdes.write(this).compactPrint
@@ -100,7 +101,7 @@ case class CompletionMessage(
 
 object CompletionMessage extends DefaultJsonProtocol {
     def parse(msg: String) = Try(serdes.read(msg.parseJson))
-    implicit val serdes = jsonFormat2(CompletionMessage.apply)
+    implicit val serdes = jsonFormat3(CompletionMessage.apply)
 }
 
 case class PingMessage(name: String) extends Message {
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
@@ -16,9 +16,16 @@
 
 package whisk.core.loadBalancer
 
+import java.nio.charset.StandardCharsets
+
 import scala.collection.mutable
+import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration._
+import scala.util.Failure
+import scala.util.Success
+
+import org.apache.kafka.clients.producer.RecordMetadata
 
 import akka.actor.Actor
 import akka.actor.ActorRef
@@ -28,25 +35,54 @@ import akka.actor.FSM.CurrentState
 import akka.actor.FSM.SubscribeTransitionCallBack
 import akka.actor.FSM.Transition
 import akka.actor.Props
+import akka.pattern.pipe
 import akka.util.Timeout
 import spray.json._
 import spray.json.DefaultJsonProtocol._
 import whisk.common.AkkaLogging
 import whisk.common.ConsulKV.LoadBalancerKeys
 import whisk.common.KeyValueStore
+import whisk.common.Logging
+import whisk.common.LoggingMarkers
+import whisk.common.RingBuffer
 import whisk.common.TransactionId
+import whisk.core.connector.ActivationMessage
+import whisk.core.connector.CompletionMessage
+import whisk.core.connector.MessageConsumer
+import whisk.core.connector.MessageProducer
 import whisk.core.connector.PingMessage
+import whisk.core.database.NoDocumentException
+import whisk.core.entitlement.Privilege.Privilege
+import whisk.core.entity.ActivationId.ActivationIdGenerator
+import whisk.core.entity.AuthKey
+import whisk.core.entity.CodeExecAsString
+import whisk.core.entity.DocRevision
+import whisk.core.entity.EntityName
+import whisk.core.entity.EntityPath
+import whisk.core.entity.ExecManifest
+import whisk.core.entity.Identity
+import whisk.core.entity.Secret
+import whisk.core.entity.Subject
+import whisk.core.entity.UUID
+import whisk.core.entity.WhiskAction
+import whisk.core.entity.types.EntityStore
 
 // Received events
 case object GetStatus
+
+case object Tick
 
 // States an Invoker can be in
 sealed trait InvokerState { val asString: String }
 case object Offline extends InvokerState { val asString = "down" }
 case object Healthy extends InvokerState { val asString = "up" }
+case object UnHealthy extends InvokerState { val asString = "unhealthy" }
+
+case class ActivationRequest(msg: ActivationMessage, invoker: String)
+case class InvocationFinishedMessage(name: String, successful: Boolean)
 
 // Data stored in the Invoker
-final case class InvokerInfo()
+final case class InvokerInfo(buffer: RingBuffer[Boolean])
 
 /**
  * Actor representing a pool of invokers
@@ -62,12 +98,16 @@ final case class InvokerInfo()
 class InvokerPool(
     childFactory: (ActorRefFactory, String) => ActorRef,
     kv: KeyValueStore,
-    invokerDownCallback: String => Unit) extends Actor {
+    invokerDownCallback: String => Unit,
+    producer: MessageProducer,
+    activeAckConsumer: MessageConsumer,
+    activeAckCallback: CompletionMessage => Unit,
+    pingConsumer: MessageConsumer) extends Actor {
 
     implicit val transid = TransactionId.invokerHealth
+    implicit val logging = new AkkaLogging(context.system.log)
     implicit val timeout = Timeout(5.seconds)
     implicit val ec = context.dispatcher
-    val logging = new AkkaLogging(context.system.log)
 
     // State of the actor. It's important not to close over these
     // references directly, so they don't escape the Actor.
@@ -85,6 +125,21 @@ class InvokerPool(
             invoker.forward(p)
 
         case GetStatus => sender() ! invokerStatus.toMap
+
+        case msg: InvocationFinishedMessage => {
+            // Forward message to invoker, if InvokerActor exists
+            invokers.get(msg.name).map(_.forward(msg))
+        }
+
+        case msg: ActivationRequest => {
+            implicit val transid = msg.msg.transid
+            val start = transid.started(this, LoggingMarkers.CONTROLLER_KAFKA, s"posting topic '${msg.invoker}' with activation id '${msg.msg.activationId}'")
+
+            producer.send(msg.invoker, msg.msg).andThen {
+                case Success(status) => transid.finished(this, start, s"Posted to ${status.topic()}[${status.partition()}][${status.offset()}]")
+                case Failure(e)      => transid.failed(this, start, s"Error on posting to topic ${msg.invoker}")
+            }.pipeTo(sender())
+        }
 
         case CurrentState(invoker, currentState: InvokerState) =>
             invokerStatus.update(invoker.path.name, currentState)
@@ -108,13 +163,63 @@ class InvokerPool(
         }.map { case (name, state) => s"$name: $state" }
         logging.info(this, s"invoker status changed to ${pretty.mkString(", ")}")
     }
+
+    /** Receive activeAcks from Invoker. */
+    activeAckConsumer.onMessage((topic, _, _, bytes) => {
+        val raw = new String(bytes, StandardCharsets.UTF_8)
+        CompletionMessage.parse(raw) match {
+            case Success(m: CompletionMessage) => {
+                self ! InvocationFinishedMessage(m.invoker, m.response.response.statusCode <= 2)
+                activeAckCallback(m)
+            }
+            case Failure(t) => logging.error(this, s"failed processing message: $raw with $t")
+        }
+    })
+
+    /** Receive Ping messages from invokers. */
+    pingConsumer.onMessage((topic, _, _, bytes) => {
+        val raw = new String(bytes, StandardCharsets.UTF_8)
+        PingMessage.parse(raw) match {
+            case Success(p: PingMessage) => self ! p
+            case Failure(t)              => logging.error(this, s"failed processing message: $raw with $t")
+        }
+    })
 }
 
 object InvokerPool {
     def props(
         f: (ActorRefFactory, String) => ActorRef,
         kv: KeyValueStore,
-        cb: String => Unit) = Props(new InvokerPool(f, kv, cb))
+        cb: String => Unit,
+        p: MessageProducer,
+        ackC: MessageConsumer,
+        acb: CompletionMessage => Unit,
+        pc: MessageConsumer) = Props(new InvokerPool(f, kv, cb, p, ackC, acb, pc))
+
+    val testAction = ExecManifest.runtimesManifest.resolveDefaultRuntime("nodejs:6").map { manifest =>
+        new WhiskAction(
+            namespace = EntityPath("whisk.system"),
+            name = EntityName("invokerHealthTestAction"),
+            exec = new CodeExecAsString(manifest, """function main(params) { return params || {}; }""", None))
+    }
+
+    /** Create an test action on startup if it is not there. */
+    def createTestActionForInvokerHealth(db: EntityStore)(implicit logging: Logging, ec: ExecutionContext): Future[WhiskAction] = {
+        implicit val tid = TransactionId.loadbalancer
+        testAction match {
+            case Some(action) =>
+                WhiskAction.get(db, action.docid).recover {
+                    case _: NoDocumentException => WhiskAction.put(db, action)
+                }.map(_ => action).andThen {
+                    case Success(_) => logging.info(this, "Testaction for invokerHealth exists now.")
+                    case Failure(e) => logging.error(this, s"Error on creating testaction for invokerHealth: $e")
+                }
+            case None => {
+                logging.error(this, "Error on creating testaction for invokerHealth: Probably there were problems with the manifest.")
+                Future.failed(new Exception("Error on creating testaction for invokerHealth: Probably there were problems with the manifest."))
+            }
+        }
+    }
 }
 
 /**
@@ -125,19 +230,38 @@ object InvokerPool {
  */
 class InvokerActor extends FSM[InvokerState, InvokerInfo] {
     implicit val transid = TransactionId.invokerHealth
-    val logging = new AkkaLogging(context.system.log)
+    implicit val logging = new AkkaLogging(context.system.log)
     def name = self.path.name
 
     val healthyTimeout = 10.seconds
 
-    startWith(Healthy, InvokerInfo())
+    // This is done at this point to not intermingle with the state-machine
+    // especially their timeouts.
+    def customReceive: Receive = {
+        case _: RecordMetadata => // The response of putting testactions to the MessageProducer. We don't have to do anything with them.
+    }
+    override def receive = customReceive.orElse(super.receive)
+
+    startWith(UnHealthy, InvokerInfo(new RingBuffer[Boolean](InvokerActor.bufferSize)))
 
     /**
      * An Offline invoker represents an existing but broken
-     * invoker.
+     * invoker. This means, that it does not send pings anymore.
      */
     when(Offline) {
-        case Event(_: PingMessage, _) => goto(Healthy)
+        case Event(_: PingMessage, _) => goto(UnHealthy)
+    }
+
+    /**
+     * An UnHealthy invoker represents an invoker, that was not able to handle actions successfully.
+     */
+    when(UnHealthy, stateTimeout = healthyTimeout) {
+        case Event(_: PingMessage, _) => stay
+        case Event(StateTimeout, _)   => goto(Offline)
+        case Event(Tick, info) => {
+            invokeTestActions()
+            stay
+        }
     }
 
     /**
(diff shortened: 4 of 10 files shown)


common/scala/src/main/scala/whisk/common/RingBuffer.scala
```diff
@@ -0,0 +1,16 @@
...
+    def apply[T](size: Int) = new RingBuffer[T](size)
+}
+
+class RingBuffer[T](size: Int) {
+    private val inner = new CircularFifoBuffer(size)
+
+    def add(el: T) = inner.add(el)
+    def count(el: T) = {
+        inner.toArray().asInstanceOf[Array[T]].count(_ == el)
+    }
```
markusthoemmes: My initial implementation had just `add` and `toList` which makes it easier to apply whatever functionality i need to the values in the buffer outside of the buffer's business code. This might be okay, if we just need `count`.

common/scala/src/main/scala/whisk/common/RingBuffer.scala
```diff
@@ -0,0 +1,16 @@
+package whisk.common
```
cbickel: License header is missing.

common/scala/src/main/scala/whisk/core/connector/Message.scala
```diff
@@ -53,7 +53,8 @@ case class ActivationMessage(
     activationId: ActivationId,
     activationNamespace: EntityPath,
     content: Option[JsObject],
-    cause: Option[ActivationId] = None)
+    cause: Option[ActivationId] = None,
+    invoker: Option[String] = None)
```
markusthoemmes: Why is this an `Option`? Don't we always need an invoker here?

common/scala/src/main/scala/whisk/core/connector/Message.scala
```diff
@@ -88,7 +89,8 @@ object ActivationMessage extends DefaultJsonProtocol {
  */
 case class CompletionMessage(
     override val transid: TransactionId,
-    response: WhiskActivation)
+    response: WhiskActivation,
+    invoker: Option[String])
```
markusthoemmes: See above.

common/scala/src/main/scala/whisk/common/RingBuffer.scala
rabbah: this seems like an expensive way to count - how often do you count?

common/scala/src/main/scala/whisk/core/connector/Message.scala
rabbah: why is the invoker in the message?

ansible/group_vars/all
```diff
@@ -25,6 +25,9 @@ controller:
   port: 10001
   blackboxFraction: 0.10
 
+invokerHealth:
+  testAction: whisk.system/utils/date
```
rabbah: do we need this test action? the reason is that the catalog isn't usually installed by default (ie wasn't a requirement) but this suggests it must be installed.

the invoker can just have a builtin action string (ping/date) that it can run - do we need this to be an actual action?

there are two things you're potentially testing by making the health action a real action:
1. the invoker has a healthy connection to cloudant
2. the invoker can execute an action

i think these are being conflated and can be tested separately.
did you consider an independent cloudant ping and a precanned action string to invoke?

markusthoemmes: Reasoning is, that having "real" payload is preferable here but without loosing user activations, hence this simulates real load as close as possible which I think is desirable to catch whatever might be wrong with the invoker.

rabbah:
Can we avoid dependence on whisk.system catalog then and have the controller create a ping action? (For example - I hardly ever run the post deploy playbook; now if I don't I can develop... is that right)?

common/scala/src/main/scala/whisk/common/RingBuffer.scala
cbickel: The `count` call was always when an ActiveAck came back to the invoker.
Now I changed the code, that the count is not called when the Invoker is already Healthy and the state of the activation was successful: https://github.com/openwhisk/openwhisk/pull/1922/commits/dd9db127c2d5973e2c0287123ca220ba4d829358#diff-d2e5bc035bf23b1b663aceb699f5f6e0L288

common/scala/src/main/scala/whisk/core/connector/Message.scala
cbickel: I put the invoker to the Message, because I moved connectivities (Kafka) from the LoadBalancerService to the InvokerPool.
To transfer the invokername, that the LoadBalancer has chosen to the invokerPool, I use this field.

ansible/group_vars/all
cbickel: At the moment it is not possible to transfer the action itself on Kafka.
We'd need to implement this with
- a new message
- the existing ActivationMessage

In both cases this implementation would not represent the real Health status between putting the activation on Kafka and getting the active ack.
The reason is, that we cannot put user actions on Kafka, because they might be too large.
So we cannot be sure, if the invoker (and the path to it) is really healthy.

In addition: Each invoker starts with the state `Healthy`. This action is only needed, if an invoker turns `UnHealthy` or `Offline`.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -104,6 +104,8 @@ protected[actions] trait PrimitiveActions {
             activationIdFactory.make(), // activation id created here
             activationNamespace = user.namespace.toPath,
             args,
+            // Flag will be set in LoadBalancer
+            invoker = "",
```
markusthoemmes: This is not better than an option. At best its worse.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -230,7 +225,7 @@ class InvokerActor(testAction: String) extends FSM[InvokerState, InvokerInfo] {
     when(UnHealthy, stateTimeout = healthyTimeout) {
         case Event(_: PingMessage, _)                   => stay
         case Event(StateTimeout, _)                     => goto(Offline)
-        case Event(cm: InvocationFinishedMessage, info) => handleCompletionMessage(cm.successful, info.buffer, UnHealthy)
+        case Event(cm: InvocationFinishedMessage, info) => handleCompletionMessage(cm.successful, info.buffer)
```
markusthoemmes: I think we should *always* handle those. Could move the whole method to `onUnhandled` then.

rabbah:
@cbickel does this pr close https://github.com/openwhisk/openwhisk/issues/1042?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -191,7 +192,7 @@ object InvokerPool {
     val testAction = new WhiskAction(
         namespace = EntityPath("whisk.system"),
         name = EntityName("invokerHealthTestAction"),
-        exec = NodeJSExec("""function main(params) { return params || {}; }""", None))
+        exec = new CodeExecAsString(RuntimeManifest("nodejs:6"), """function main(params) { return params || {}; }""", None))
```
cbickel: @rabbah Is there any other possibility to get a RuntimeManifest?

rabbah:
Can you mix this in https://github.com/openwhisk/openwhisk/blob/master/tests/src/test/scala/whisk/core/entity/test/ExecHelpers.scala

rabbah:
Never mind I thought that was for a test. In the invoker and controller it is loaded at startup. You can get it from the singleton.

rabbah:
https://github.com/openwhisk/openwhisk/blob/master/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -184,20 +173,35 @@ class LoadBalancerService(config: WhiskConfig)(implicit val actorSystem: ActorSy
...
+                    case _: NoDocumentException => WhiskAction.put(db, action)
+                }.andThen {
+                    case Success(_) => logging.info(this, "Testaction for invokerHealth exists now.")
+                    case Failure(e) => logging.error(this, s"Error on creating testaction for invokerHealth: $e")
+                }
+            case None => logging.error(this, "Error on creating testaction for invokerHealth: Probably there were problems with the manifest.")
         }
-    })
+    }
+    createTestActionForInvokerHealth()
```
markusthoemmes: Do we need to return a `Future` here or `Await` it? Otherwise this will race with the InvokerPool?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -105,10 +161,42 @@ class InvokerPool(kv: KeyValueStore, invokerDownCallback: String => Unit) extend
...
+              ackC: MessageConsumer,
+              acb: CompletionMessage => Unit,
+              pc: MessageConsumer) = Props(new InvokerPool(kv, cb, mp, ackC, acb, pc))
+
+    val testAction = ExecManifest.runtimesManifest.resolveDefaultRuntime("nodejs").map { manifest =>
+        new WhiskAction(
+            namespace = EntityPath("whisk.system"),
+            name = EntityName("invokerHealthTestAction"),
+            exec = new CodeExecAsString(manifest, """function main(params) { return params || {}; }""", None))
+    }
```
markusthoemmes: Should we move action creation here?

`def testAction(implicit entityStore: EntityStore): Future[DocInfo]`

?

rabbah: The heartbeat action runs on every invoker?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -105,10 +161,42 @@ class InvokerPool(kv: KeyValueStore, invokerDownCallback: String => Unit) extend
                 case _       =>
             }
     }
+
+    /** Receive activeAcks from Invoker. */
+    activeAckConsumer.onMessage((topic, _, _, bytes) => {
+        val raw = new String(bytes, "utf-8")
```
markusthoemmes: import `java.nio.charset.StandardCharsets` and use `StandardCharsets.UTF_8`

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -86,6 +124,24 @@ class InvokerPool(kv: KeyValueStore, invokerDownCallback: String => Unit) extend
...
+            implicit val transid = msg.msg.transid
+            val start = transid.started(this, LoggingMarkers.CONTROLLER_KAFKA, s"posting topic '${msg.invoker}' with activation id '${msg.msg.activationId}'")(logging)
+            val post = producer.send(msg.invoker, msg.msg).map { status =>
+                transid.finished(this, start, s"Posted to ${status.topic()}[${status.partition()}][${status.offset()}]")(logging)
+                status
+            }.andThen {
+                case Failure(e) => transid.failed(this, start, s"Error on posting to topic ${msg.invoker}")(logging)
+            }
+
+            pipe(post) to sender()
```
markusthoemmes: Can be rewritten to

```
producer.send(msg.invoker, msg.msg).andThen {
    case Success(status) => transid.finished(this, start, s"Posted to ${status.topic()}[${status.partition()}][${status.offset()}]")(logging)
    case Failure(e) => transid.failed(this, start, s"Error on posting to topic ${msg.invoker}")(logging)
}.pipeTo(sender())
```

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -58,7 +89,14 @@ final case class InvokerInfo()
...
+class InvokerPool(kv: KeyValueStore,
+                  invokerDownCallback: String => Unit,
+                  producer: MessageProducer,
+                  activeAckConsumer: MessageConsumer,
+                  activeAckCallback: CompletionMessage => Unit,
+                  pingConsumer: MessageConsumer)
+    extends Actor {
+
     implicit val transid = TransactionId.invokerHealth
     val logging = new AkkaLogging(context.system.log)
```
markusthoemmes: should we make this implicit and remove the explicit passing below?

common/scala/build.gradle
```diff
@@ -28,6 +28,8 @@ dependencies {
     compile 'commons-codec:commons-codec:1.9'
     compile 'commons-io:commons-io:2.4'
     compile 'commons-logging:commons-logging:1.2'
+    compile 'commons-codec:commons-codec:1.9'
+    compile 'commons-collections:commons-collections:3.2.2'
```
markusthoemmes: Do we need both?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -105,10 +161,42 @@ class InvokerPool(kv: KeyValueStore, invokerDownCallback: String => Unit) extend
...
+                self ! InvocationFinishedMessage(m.invoker, m.response.response.statusCode <= 2)
+                activeAckCallback(m)
+            }
+            case Failure(t) => logging.error(this, s"failed processing message: $raw with $t")
+        }
+    })
+
+    /** Receive Ping messages from invokers. */
+    pingConsumer.onMessage((topic, _, _, bytes) => {
+        val raw = new String(bytes, "utf-8")
```
markusthoemmes: See above.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -123,20 +211,36 @@ class InvokerActor extends FSM[InvokerState, InvokerInfo] {
     def name = self.path.name
 
     val healthyTimeout = 10.seconds
+    val namespace = EntityPath("whisk.system")
```
markusthoemmes: What do we need this for in here?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
markusthoemmes: In theory yes. At steady state, the Controller will always have a cache-hit though.

cbickel: I moved it here now, but I left it as variable.
The reason is as discussed: Don't start if it is not possible to create the action.

common/scala/src/main/scala/whisk/core/connector/Message.scala
```diff
@@ -100,7 +101,15 @@ case class CompletionMessage(
...
-    implicit val serdes = jsonFormat2(CompletionMessage.apply)
+    implicit val serdes = jsonFormat3(CompletionMessage.apply)
+}
+
+case class InvocationFinishedMessage(name: String, successful: Boolean) extends Message {
+    override def serialize = InvocationFinishedMessage.serdes.write(this).compactPrint
+}
+
+object InvocationFinishedMessage extends DefaultJsonProtocol {
+    implicit val serdes = jsonFormat(InvocationFinishedMessage.apply _, "name", "successful")
```
markusthoemmes: Why is this defined in `Messages`? After all it seems like only being used by the `InvokerPool` and thus should go there?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -105,10 +162,66 @@ class InvokerPool(kv: KeyValueStore, invokerDownCallback: String => Unit) extend
...
+    }
+
+    /** Create an test action on startup if it is not there. */
+    def createTestActionForInvokerHealth(db: EntityStore)(implicit logging: Logging, ec: ExecutionContext): Future[WhiskAction] = {
+        implicit val tid = TransactionId.loadbalancer
+        testAction match {
+            case Some(action) =>
+                WhiskAction.get(db, action.docid).recover {
+                    case _: NoDocumentException => WhiskAction.put(db, action)
+                }.flatMap {
```
markusthoemmes: Attention: This will ignore errors on `WhiskAction.get`. What you're after is

```
WhiskAction.get(...).recoverWith {
    case case _: NoDocumentException => WhiskAction.put(db, action)
}.flatMap(_ => action).andThen {
    case Success(_) => logging.info(this, "Testaction for invokerHealth exists now.")
    case Failure(e) => logging.error(this, s"Error on creating testaction for invokerHealth: $e")
}

**Comment by cbickel** [comment]
PG1#1349 is green

**Comment by cbickel** [comment]
I implemented one little change:
If an invoker registers at the controller, it starts with UnHealthy instead of Healthy.
The reason is to recognise a potential failure of the test action, that brings up the invoker again, as soon as possible.

File: core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala [hunk 19]
```diff
@@ -63,6 +63,7 @@ class Invoker(
     with ActionLogDriver {
 
     private implicit val executionContext: ExecutionContext = actorSystem.dispatcher
+    val invokerName = s"invoker$instance"
```
rabbah: in the constructor, the Invoker extends `MessageHandler(s"invoker$instance")` which provides `this.name`.

cbickel: good point. Already fixed with a new commit.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -86,6 +126,21 @@ class InvokerPool(
 
         case GetStatus => sender() ! invokerStatus.toMap
 
+        case msg: InvocationFinishedMessage => {
+            // Forward message to invoker, if InvokerActor exists
+            invokers.get(msg.name).map(_.forward(msg))
+        }
+
+        case msg: ActivationRequest => {
```
rabbah: all activation requests are now serialized through this actor - do we really want to do this? since this actor processes different messages, including health pings, it seems like we can end up delaying a producer send unnecessarily. this changes the role of the actor from a supervisor to a gatekeeper for all invoker operations.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -108,13 +163,63 @@ class InvokerPool(
         }.map { case (name, state) => s"$name: $state" }
         logging.info(this, s"invoker status changed to ${pretty.mkString(", ")}")
     }
+
+    /** Receive activeAcks from Invoker. */
+    activeAckConsumer.onMessage((topic, _, _, bytes) => {
+        val raw = new String(bytes, StandardCharsets.UTF_8)
+        CompletionMessage.parse(raw) match {
+            case Success(m: CompletionMessage) => {
+                self ! InvocationFinishedMessage(m.invoker, m.response.response.statusCode <= 2)
```
rabbah: similarly here, now active acks are queued along with other messages this actor will receive.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -108,13 +163,63 @@ class InvokerPool(
...
+        p: MessageProducer,
+        ackC: MessageConsumer,
+        acb: CompletionMessage => Unit,
+        pc: MessageConsumer) = Props(new InvokerPool(f, kv, cb, p, ackC, acb, pc))
+
+    val testAction = ExecManifest.runtimesManifest.resolveDefaultRuntime("nodejs:6").map { manifest =>
+        new WhiskAction(
+            namespace = EntityPath("whisk.system"),
+            name = EntityName("invokerHealthTestAction"),
+            exec = new CodeExecAsString(manifest, """function main(params) { return params || {}; }""", None))
```
rabbah: `return params || {}` is not necessary, the action is guaranteed to receive a json object.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -108,13 +163,63 @@ class InvokerPool(
...
+    }
+
+    /** Create an test action on startup if it is not there. */
+    def createTestActionForInvokerHealth(db: EntityStore)(implicit logging: Logging, ec: ExecutionContext): Future[WhiskAction] = {
+        implicit val tid = TransactionId.loadbalancer
+        testAction match {
+            case Some(action) =>
+                WhiskAction.get(db, action.docid).recover {
+                    case _: NoDocumentException => WhiskAction.put(db, action)
```
rabbah: what if the put fails (because of a conflict)?
should this action be created once at startup?
is it ever deleted?
it is shared by all invokers?
it might make the management cleaner to install it once at startup and fail to come up if there's a problem creating this actin.

rabbah: ok i see this is already the case later (in LoadBalancerService line 181) - makes me wonder why is method is here.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -108,13 +163,63 @@ class InvokerPool(
...
+                }.map(_ => action).andThen {
```
rabbah: are you deliberately not storing the document revision to force the invoker to fetch this action from the datastore vs its cache?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -108,13 +163,63 @@ class InvokerPool(
...
+                    case Success(_) => logging.info(this, "Testaction for invokerHealth exists now.")
+                    case Failure(e) => logging.error(this, s"Error on creating testaction for invokerHealth: $e")
+                }
+            case None => {
```
rabbah: since you really want this to fail at startup/init, should you just force test action to be Some?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -125,19 +230,38 @@ object InvokerPool {
...
-     * invoker.
+     * invoker. This means, that it does not send pings anymore.
      */
     when(Offline) {
-        case Event(_: PingMessage, _) => goto(Healthy)
+        case Event(_: PingMessage, _) => goto(UnHealthy)
+    }
+
+    /**
+     * An UnHealthy invoker represents an invoker, that was not able to handle actions successfully.
```
rabbah: an invoker, that ... remove comma

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -150,14 +274,95 @@ class InvokerActor extends FSM[InvokerState, InvokerInfo] {
...
+                // Use the sid of the InvokerSupervisor as tid
+                transid = transid,
+                action = action.fullyQualifiedName(true),
+                // Use empty DocRevision to force the invoker to pull the action from db all the time
+                revision = DocRevision(),
+                // Authentication is not needed anymore at this point.
+                user = Identity(Subject("unhealthyInvokerCheck"), EntityName("unhealthyInvokerCheck"), AuthKey(UUID(), Secret()), Set[Privilege]()),
+                // Create a new Activation ID for this activation
+                activationId = new ActivationIdGenerator {}.make(),
+                activationNamespace = EntityPath("whisk.system"),
```
rabbah: doesn't the action have the namespace set? can we avoid more than one occurrence of `whisk.system`.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -150,14 +274,95 @@ class InvokerActor extends FSM[InvokerState, InvokerInfo] {
...
+    private def invokeTestActions() = {
+        InvokerPool.testAction.map { action =>
+            val activationMessage = ActivationMessage(
+                // Use the sid of the InvokerSupervisor as tid
+                transid = transid,
+                action = action.fullyQualifiedName(true),
+                // Use empty DocRevision to force the invoker to pull the action from db all the time
+                revision = DocRevision(),
+                // Authentication is not needed anymore at this point.
+                user = Identity(Subject("unhealthyInvokerCheck"), EntityName("unhealthyInvokerCheck"), AuthKey(UUID(), Secret()), Set[Privilege]()),
```
rabbah: if the Identity is not semantically important, should we have just one created for all requests?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -150,14 +274,95 @@ class InvokerActor extends FSM[InvokerState, InvokerInfo] {
...
+     * Creates an Activationmessage with the given action and sends it to the InvokerPool.
+     * The InvokerPool redirects it to the invoker which is represented by this InvokerActor.
+     */
+    private def invokeTestActions() = {
+        InvokerPool.testAction.map { action =>
+            val activationMessage = ActivationMessage(
+                // Use the sid of the InvokerSupervisor as tid
+                transid = transid,
+                action = action.fullyQualifiedName(true),
+                // Use empty DocRevision to force the invoker to pull the action from db all the time
```
rabbah: ok comment answers my question above.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -150,14 +274,95 @@ class InvokerActor extends FSM[InvokerState, InvokerInfo] {
...
+                gotoIfNotThere(UnHealthy)
+            }
+        }
+    }
+
+    /**
+     * Creates an Activationmessage with the given action and sends it to the InvokerPool.
+     * The InvokerPool redirects it to the invoker which is represented by this InvokerActor.
+     */
+    private def invokeTestActions() = {
```
rabbah: why is this plural? As far as I see, it's invoke one action.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -185,23 +171,21 @@ class LoadBalancerService(config: WhiskConfig)(implicit val actorSystem: ActorSy
...
-    })
-
     private val consul = new ConsulClient(config.consulServer)
     private val invokerFactory = (f: ActorRefFactory, name: String) => f.actorOf(InvokerActor.props, name)
-    private val invokerPool = actorSystem.actorOf(InvokerPool.props(invokerFactory, consul.kv, invoker => {
-        clearInvokerState(invoker)
-        logging.info(this, s"cleared loadbalancer state of $invoker")(TransactionId.invokerHealth)
-    }))
+    // Do not create the invokerpool, if it is not possible to create the test action to recover the invokers.
+    private val invokerPool = InvokerPool.createTestActionForInvokerHealth(entityStore).map { _ =>
```
rabbah: should this construction be more deliberate and explicit here and just fail via a `require` or `assert`?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
cbickel: I did a little code change:
the action is now created/updated on every startup of the controller.
It will be deleted never.

Now the startup of the controller fails, if the creation f the action is not possible.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
cbickel: I talked to Markus, because he has worked a bit more with actors.
He agreed, that there could be an impact, but he is not sure.

My idea was to handle the complete communication with the Invokers in `InvokerSupervision`, to implement in the Load Balancer only the algorithm for the dispatching of the actions to their invokerInstances.

markusthoemmes: My 2 cents: I cannot quantify the impact. It is true, that the messages are serialized through the actor, but the kafka sending runs parallelized because the actor is not blocked by it. So the serialization could (should even?) be sufficiently small to be neglectable.

rabbah: I think under load this could become a bottleneck. This change makes me nervous  without data to show otherwise, since both the request and active acks are funneled through the same actor.

cbickel:
PG1#1365 is green.

common/scala/src/main/scala/whisk/core/connector/Message.scala
```diff
@@ -88,7 +88,8 @@ object ActivationMessage extends DefaultJsonProtocol {
  */
 case class CompletionMessage(
     override val transid: TransactionId,
-    response: WhiskActivation)
+    response: WhiskActivation,
+    invoker: String)
```
rabbah: we do know when the request is posted which invoker it was assigned to.
it may be necessary anyway to keep this field in the active ack since in the future another invoker may steal a request.

cbickel: We only know it in the LoadBalancer itself, but this information is not passed to the Actor, that handles the health of the invoker.
That was the initial idea of this field. But your point was right as well.

cbickel:
PG1#1382 is green. (After rebase)

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -179,23 +165,12 @@ class InvokerPool(
 
 object InvokerPool {
     def props(
-        f: (ActorRefFactory, String) => ActorRef,
```
cbickel: We have to pass the InvokerFactory into the supervisor from outside. Main reason is testability.

cbickel:
PG1#1428 is green

cbickel:
PG4#154 is green

rabbah:
Great. I expect i can get this in tomorrow.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -181,27 +176,70 @@ class LoadBalancerService(config: WhiskConfig)(implicit val actorSystem: ActorSy
...
+    /** Subscribes to active acks (completion messages from the invokers). */
+    private val activeAckConsumer = new KafkaConsumerConnector(config.kafkaHost, "completions", "completed")
+
+    /** Registers a handler for received active acks from invokers. */
+    activeAckConsumer.onMessage((topic, _, _, bytes) => {
+        val raw = new String(bytes, StandardCharsets.UTF_8)
+        CompletionMessage.parse(raw) match {
+            case Success(m: CompletionMessage) =>
+                processCompletion(m)
+                invokerPool ! InvocationFinishedMessage(m.invoker, m.response.response.statusCode <= 2)
```
rabbah: should the status code check here be explicit whisk error vs <= 2?

rabbah: you can do this by just calling `.isWhiskError` on the response.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -108,13 +147,40 @@ class InvokerPool(
         }.map { case (name, state) => s"$name: $state" }
         logging.info(this, s"invoker status changed to ${pretty.mkString(", ")}")
     }
+
+    /** Receive Ping messages from invokers. */
+    pingConsumer.onMessage((topic, _, _, bytes) => {
+        val raw = new String(bytes, StandardCharsets.UTF_8)
+        PingMessage.parse(raw) match {
+            case Success(p: PingMessage) => self ! p
+            case Failure(t)              => logging.error(this, s"failed processing message: $raw with $t")
```
rabbah: do we need a tid here?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -150,14 +238,99 @@ class InvokerActor extends FSM[InvokerState, InvokerInfo] {
...
+
+    /**
+     * Handling for active acks. This method saves the result (successful or unsuccessful)
+     * into an RingBuffer and checks, if the InvokerActor has to be changed to UnHealthy.
+     *
+     * @param wasActivationSuccessful: result of Activation
+     * @param buffer to be used
+     */
+    private def handleCompletionMessage(wasActivationSuccessful: Boolean, buffer: RingBuffer[Boolean]) = {
+        buffer.add(wasActivationSuccessful)
```
rabbah: seems to me you can avoid the ring buffer and just use a saturating counter (reason: you add and count) - i won't hold up the pr for this, but would like to understand if the ring buffer is really necessary.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -150,14 +238,99 @@ class InvokerActor extends FSM[InvokerState, InvokerInfo] {
...
+
+        // If the current state is UnHealthy, this means, that the activeAck is from an testaction.
+        // If this is successful it seems like the Invoker is Healthy again. So we execute immediately
+        // a new testaction to remove the errors out of the RingBuffer as fast as possible.
+        if (wasActivationSuccessful && stateName == UnHealthy) {
+            invokeTestAction()
```
rabbah: i'm wondering if it's necessary to have this `if` clause (279-281) since the `else` clause below will check if the number of activations is above a threshold and the actor will remain in the unhealthy state until the threshold is exceeded.

the comment suggests that the this is proactively sending another test action (one more more) - vs letting the timer (1 minute intervals) fire a next one... is this violating the actor contract (namely, should there be a slow pole (1 min) and then a faster reactive one? In essence you can end up sending some additional test actions here that pile up with the ones from the timer.

This may be acceptable - but if that's deliberate, we should expand the comments further.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
rabbah: I addressed this one in the merge.

rabbah:
Small teaks, squashed and merged as https://github.com/openwhisk/openwhisk/commit/d2aef8c58cce802d7f190b954753bf5dce31647f.

PR #1924, opened 2017-03-01
Improve action log driver to truncate more tightly.

rabbah:
When truncating, read all bytes up to the limit.
Separate log processors into its own trait - add unit tests.
Fix/tighten pre-existing test.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.invoker
+
+import java.nio.charset.StandardCharsets
+
+import scala.Vector
+import scala.util.{ Failure, Success, Try }
+
+import spray.json._
+import spray.json.DefaultJsonProtocol
+import whisk.common.{ Logging, TransactionId }
+import whisk.core.entity._
+import whisk.core.entity.size.{ SizeInt, SizeString }
+import scala.collection.mutable.Buffer
+
+/**
+ * Represents a single log line as read from a docker log
+ */
+protected[invoker] case class LogLine(time: String, stream: String, log: String) {
+    def toFormattedString = f"$time%-30s $stream: ${log.trim}"
+    def dropRight(maxBytes: ByteSize) = {
+        val bytes = log.getBytes(StandardCharsets.UTF_8).dropRight(maxBytes.toBytes.toInt)
+        LogLine(time, stream, new String(bytes, StandardCharsets.UTF_8))
+    }
+}
+
+protected[invoker] object LogLine extends DefaultJsonProtocol {
+    implicit val serdes = jsonFormat3(LogLine.apply)
+}
+
+protected[invoker] trait ActionLogDriver {
+
+    // The action proxies inserts this line in the logs at the end of each activation for stdout/stderr
+    protected val LOG_ACTIVATION_SENTINEL = "XXX_THE_END_OF_A_WHISK_ACTIVATION_XXX"
+
+    /**
+     * Given the JSON driver's raw output of a docker container, convert it into our own
+     * JSON format. If asked, check for sentinel markers (which are not included in the output).
+     *
+     * Only parses and returns so much logs to fit into the LogLimit passed.
+     *
+     * @param logMsgs raw String read from a JSON log-driver written file
+     * @param requireSentinel determines if the processor should wait for a sentinel to appear
+     * @param limit the limit to apply to the log size
+     *
+     * @return Tuple containing (isComplete, isTruncated, logs)
+     */
+    protected def processJsonDriverLogContents(logMsgs: String, requireSentinel: Boolean, limit: ByteSize)(
+        implicit transid: TransactionId, logging: Logging): (Boolean, Boolean, Vector[LogLine]) = {
+
+        var hasOut = false
+        var hasErr = false
+        var truncated = false
+        var bytesSoFar = 0.B
+        val logLines = Buffer[LogLine]()
+        val lines = logMsgs.lines
+
+        // read whiles bytesSoFar <= limit when requireSentinel to try and grab sentinel if they exist to indicate completion
+        while (lines.hasNext && ((requireSentinel && bytesSoFar <= limit) || bytesSoFar < limit)) {
+            // if line does not parse, there's an error in the container log driver
+            Try(lines.next().parseJson.convertTo[LogLine]) match {
+                case Success(t) =>
+                    // if sentinels are expected, do not account for their size, otherwise, all bytes are accounted for
+                    if (requireSentinel && t.log.trim != LOG_ACTIVATION_SENTINEL || !requireSentinel) {
+                        // ignore empty log lines
+                        if (t.log.nonEmpty) {
+                            bytesSoFar += t.log.sizeInBytes
+                            if (bytesSoFar <= limit) {
+                                logLines.append(t)
+                            } else {
+                                // chop off the right most bytes that overflow
+                                val chopped = t.dropRight(bytesSoFar - limit)
+                                if (chopped.log.nonEmpty) {
+                                    logLines.append(chopped)
+                                }
+                                truncated = true
+                            }
+                        }
+                    } else if (requireSentinel) {
+                        // there may be more than one sentinel in stdout/stderr (as logs may leak across activations
+                        // if for example there log limit was exceeded in one activation and the container was reused
+                        // to run the same action again (this is considered a feature - otherwise, must drain the logs
+                        // or destroy the container as if it errored)
+                        if (t.stream == "stdout") {
+                            hasOut = true
+                        } else if (t.stream == "stderr") {
+                            hasErr = true
+                        }
+                    }
+
+                case Failure(t) =>
+                    // Drop lines that did not parse to JSON objects.
+                    // However, should not happen since we are using the json log driver.
+                    logging.error(this, s"log line skipped/did not parse: $t")
+            }
+        }
+
+        ((hasOut && hasErr) || !requireSentinel, truncated || lines.hasNext, logLines.toVector)
+    }
+}
--- a/common/scala/src/main/scala/whisk/core/entity/LogLimit.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/LogLimit.scala
@@ -41,8 +41,6 @@ import whisk.core.entity.size.SizeInt
  */
 protected[core] class LogLimit private (val megabytes: Int) extends AnyVal {
     protected[core] def asMegaBytes: ByteSize = megabytes.megabytes
-    protected[core] def truncatedLogMessage = s"Logs were truncated because they exceeded the limit of $megabytes megabytes."
-
 }
 
 protected[core] object LogLimit extends ArgNormalizer[LogLimit] {
--- a/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
+++ b/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
@@ -111,6 +111,10 @@ object Messages {
         s"${error.field} larger than allowed: ${error.is.toBytes} > ${error.allowed.toBytes} bytes."
     }
 
+    def truncateLogs(limit: ByteSize) = {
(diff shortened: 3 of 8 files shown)


core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
```diff
@@ -0,0 +1,119 @@
...
+     * @param requireSentinel determines if the processor should wait for a sentinel to appear
+     * @param limit the limit to apply to the log size
+     *
+     * @return Tuple containing (isComplete, isTruncated, logs)
+     */
+    protected def processJsonDriverLogContents(logMsgs: String, requireSentinel: Boolean, limit: ByteSize)(
+        implicit transid: TransactionId, logging: Logging): (Boolean, Boolean, Vector[LogLine]) = {
+
+        if (logMsgs.nonEmpty) {
+            val records = logMsgs.split("\n").toStream flatMap {
```
markusthoemmes: `logMsgs.lines` instead of `split("\n")`

Can we get away with just using `Iterator`?

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
```diff
@@ -0,0 +1,119 @@
...
+
+    private def appendTruncateLastLine(resultBuffer: Stream[(LogLine, ByteSize)], allRecords: Stream[LogLine], nextLineIndex: Int, limit: ByteSize, truncate: Boolean) = {
+        if (truncate) {
+            val bytesRead = if (resultBuffer.nonEmpty) resultBuffer.last._2 else 0.B
+            val readPartial = limit - bytesRead
+            if (readPartial > 0.B) {
+                val truncatedLine = allRecords(nextLineIndex).truncateTo(limit - bytesRead)
+                resultBuffer.map(_._1).toVector :+ truncatedLine
+            } else resultBuffer.map(_._1).toVector
+        } else resultBuffer.map(_._1).toVector
```
markusthoemmes: extract `resultBuffer.map(_._1).toVector` to a value up top, i.e. `val logsRead = resultBuffer.map(_._1).toVector`

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
```diff
@@ -0,0 +1,119 @@
...
+            if (requireSentinel) {
+                // there may be more than one sentinel in stdout/stderr (as logs may leak across activations
+                // if for example there log limit was exceeded in one activation and the container was reused
+                // to run the same action again (this is considered a feature - otherwise, must drain the logs
+                // or destroy the container as if it errored)
+                val (sentinels, regulars) = logBuffer.partition(_._1.log.trim == LOG_ACTIVATION_SENTINEL)
+                val hasOut = sentinels.exists(_._1.stream == "stdout")
+                val hasErr = sentinels.exists(_._1.stream == "stderr")
+
+                val bytesRead = if (regulars.nonEmpty) regulars.last._2 else 0.B
```
markusthoemmes: `val bytesRead = regulars.lastOption.map(_._2).getOrElse(0.B)`?

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
```diff
@@ -0,0 +1,119 @@
...
+            }
+        } else {
+            logging.warn(this, s"log message is empty")
+            (!requireSentinel, false, Vector())
+        }
+    }
+
+    private def appendTruncateLastLine(resultBuffer: Stream[(LogLine, ByteSize)], allRecords: Stream[LogLine], nextLineIndex: Int, limit: ByteSize, truncate: Boolean) = {
+        if (truncate) {
+            val bytesRead = if (resultBuffer.nonEmpty) resultBuffer.last._2 else 0.B
```
markusthoemmes: See above.

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
```diff
@@ -0,0 +1,119 @@
...
+
+            val cumulativeSize = records.scanLeft(0.bytes) { (acc, current) =>
+                acc + {
+                    // do not count the sentinels
+                    if (current.log != LOG_ACTIVATION_SENTINEL) current.log.sizeInBytes else 0.B
+                }
+            }.tail
+
+            val logBuffer = records.zip(cumulativeSize).takeWhile(_._2 <= limit)
+            val isTruncated = logBuffer.size < records.size
```
markusthoemmes: cache `logBuffer.size` as its used below again?

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
```diff
@@ -0,0 +1,119 @@
...
+
+            if (requireSentinel) {
+                // there may be more than one sentinel in stdout/stderr (as logs may leak across activations
+                // if for example there log limit was exceeded in one activation and the container was reused
+                // to run the same action again (this is considered a feature - otherwise, must drain the logs
+                // or destroy the container as if it errored)
+                val (sentinels, regulars) = logBuffer.partition(_._1.log.trim == LOG_ACTIVATION_SENTINEL)
```
markusthoemmes: Shall we do the `trim` in the jsonFormat of `LogLine`? It's used again in the format of the log and it'd be neat if you can just rely on it being trimmed instead of doing it just in case?

tests/src/whisk/core/invoker/test/ActionLogDriverTests.scala
```diff
@@ -0,0 +1,135 @@
...
+        val msgs = if (addSentinel) {
+            lines.map((stream, _)) :+
+                ("stdout", LOG_ACTIVATION_SENTINEL) :+
+                ("stderr", LOG_ACTIVATION_SENTINEL)
+        } else {
+            lines.map((stream, _))
+        }
+
+        msgs.map(p => LogLine("", p._1, p._2))
+            .map(LogLine.serdes.write(_).compactPrint)
```
markusthoemmes: `.map(_.toJson.compactPrint)` not possible?

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
rabbah: I thought about that but then you will both trim every time and not account for the new line character. Should it be free?

rabbah: every line (possibly more than needed. Might be silly optimization but the accounting point stands).

rabbah: Hmm I take back the every line point - maybe the sentinel should include a newcomer instead?

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
rabbah: i prefer the explicit if/else for clarity.

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
rabbah: ok although once evaluated, it's not a big hit.

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
rabbah: switched although lines seems to do more that's unnecessary.

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
rabbah: actually the trim is not necessary!

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
```diff
@@ -0,0 +1,120 @@
...
+        if (logMsgs.nonEmpty) {
+            var hasOut = false
+            var hasErr = false
+            var truncated = false
+            var bytesSoFar = 0.B
+            val logLines = Buffer[LogLine]()
+            val lines = logMsgs.lines
+
+            // read whiles bytesSoFar <= limit when requireSentinel to try and grab sentinel if they exist to indicate completion
+            while (lines.hasNext && ((requireSentinel && bytesSoFar <= limit) || bytesSoFar < limit)) {
```
markusthoemmes: Could we have `val limit = if(requireSentinel) limit + sentinel.size * 2 else limit` to simplify logic below? (pseudocode)

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
```diff
@@ -0,0 +1,120 @@
...
+                        // Drop lines that did not parse to JSON objects.
+                        // However, should not happen since we are using the json log driver.
+                        logging.error(this, s"log line skipped/did not parse: $t")
+                }
+            }
+
+            ((hasOut && hasErr) || !requireSentinel, truncated || lines.hasNext, logLines.toVector)
+        } else {
+            logging.warn(this, s"log message is empty")
+            (!requireSentinel, false, Vector())
```
markusthoemmes: Is this case only needed for logging? The above "return" line should evaluate to the same response, so we could add logging above it if lines was empty to flatten the branch hierarchy a bit.

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
```diff
@@ -0,0 +1,120 @@
...
+                        if (requireSentinel && t.log.trim != LOG_ACTIVATION_SENTINEL || !requireSentinel) {
+                            // ignore empty log lines
+                            if (t.log.nonEmpty) {
+                                bytesSoFar += t.log.sizeInBytes
+                                if (bytesSoFar <= limit) {
+                                    logLines.append(t)
+                                } else {
+                                    // chop off the right most bytes that overflow
+                                    val chopped = t.dropRight(bytesSoFar - limit)
+                                    if (chopped.log.length > 0) {
```
markusthoemmes: `if(chopped.log.nonEmpty)`

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
```diff
@@ -0,0 +1,120 @@
...
+
+            // read whiles bytesSoFar <= limit when requireSentinel to try and grab sentinel if they exist to indicate completion
+            while (lines.hasNext && ((requireSentinel && bytesSoFar <= limit) || bytesSoFar < limit)) {
+                // if line does not parse, there's an error in the container log driver
+                Try(lines.next().parseJson.convertTo[LogLine]) match {
+                    case Success(t) =>
+                        // if sentinels are expected, do not account for their size, otherwise, all bytes are accounted for
+                        if (requireSentinel && t.log.trim != LOG_ACTIVATION_SENTINEL || !requireSentinel) {
+                            // ignore empty log lines
+                            if (t.log.nonEmpty) {
```
markusthoemmes: Should we do this? Why do we ignore empty log lines?

rabbah: You can't print an empty log line, can you?

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
rabbah: we can remove the else entirely I suppose. This log message is of little value.

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
markusthoemmes: If you can't, why do we need this if? Defensive coding?

rabbah: ✅

rabbah:
pg3/326 🔵.

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * Only parses and returns so much logs to fit into the LogLimit passed.
+     *
+     * @param logMsgs raw String read from a JSON log-driver written file
+     * @param requireSentinel determines if the processor should wait for a sentinel to appear
+     * @param limit the limit to apply to the log size
+     *
+     * @return Tuple containing (isComplete, isTruncated, logs)
+     */
+    protected def processJsonDriverLogContents(logMsgs: String, requireSentinel: Boolean, limit: ByteSize)(
```
markusthoemmes: Not necessarily to be done in this PR: Should this get an `Iterator[String]` instead of all messages as a whole? That way we could read the file with `Source.fromFile(logFile).getLines` and thereby not load the whole file into memory?

rabbah: Maybe - note though we need to maintain cursors since the file has activations from than one instance. We could change this method as you suggest now, lifting the iterator out of this method. Having incremental updates is also another improvement vs the full ingest.

markusthoemmes: Yep its fine doing it afterwards. This is okay to merge as-is.

rabbah:
rebased and squashed.

rabbah:
Rebased/test moved to `tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala`.

Issue #171, opened 2016-04-04
publish container images to dockerhub


seriousme:
Hi,

I managed to get openwhisk running using vagrant. After some fiddling I got the examples to work.
This made me wonder: if the openwhisk specific containers were available on docker hub it would be much easier to give it a spin (e.g. [docker-compose](https://docs.docker.com/compose/) up would be enough).
It would have also saved me a truckload of downloads :-)

I also noticed that the images are fairly large, e.g.
-  a simple node.js container can be had for about [13 MB](https://hub.docker.com/r/mhart/alpine-node/tags/) instead of the current 400+ MB
- Consul can be had for [18 Mb](https://hub.docker.com/r/gliderlabs/consul/tags/) instead of the current 1347 MB
  etc.

I know that images share layers, but still smaller images are easier to juggle around ;-)

Cheers,
Hans

domdom82:
Hi @seriousme thanks for your excellent input.

I think it is a great idea to have the stack as compose topology on dockerhub. 

Before we do that we should first reduce the image sizes so ppl won't have to download huge chunks.

Most images are based on whisk/scala which itself is pretty large. That itself can be reduced imho.

As for the consul part, we are currently working on moving to ansible for our deployment. During that process we are also looking into ways to reduce overhead on common tools such as consul which we currently build ourselves. So chances are we got rid of that image soon.

seriousme:
@domdom82 Thanks for the positive feedback :-)

I found a relatively small scala package [frolvlad](https://hub.docker.com/r/frolvlad/alpine-scala/) based on Alpine Linux. Now I'm not a Scala expert, so I might be way off here, but alpine-scala is build on alpine-java (oracle jdk) which in turn is build on Alpine.  

However if you want to stay main stream there is also a [docker java base image](https://hub.docker.com/_/java/) which might be used as a starter to a Scala image.

Ideally there would imho be a few docker compose files: 
- one that uses cloudant as its DB
- one that uses a couchdb container as its DB (preferably with a separate data volume)

I noticed my repository on the VM also contains a CLI container, I would include that as well in the compose files.

Last point: the couchdb container is the only one started without a predefined name in [start-couchdb-box.sh] (https://github.com/openwhisk/openwhisk/blob/master/tools/db/couchdb/start-couchdb-box.sh#L30), adding a --name parameter would make it easier to reference it, and it just looks more clean as all the other containers have nice names :-)

Cheers,
Hans

[...]

PR #1928, opened 2017-03-01
Couchdb Docker Container for OpenWhisk

ddragosd:
This PR creates a CouchDB Docker Container for OpenWhisk. This container can automatically bootstrap the databases used by OpenWhisk making it easier to spin new environments without the need to run a post-install script. This idea came up from [this discussion](https://lists.apache.org/thread.html/76e64c8c658d2c0f201ee1b9412d4658e44acc7d471d066669f25abb@%3Cdev.openwhisk.apache.org%3E) on the Openwhisk DL.

The build script for this container reuses existing Ansible playbooks:
* setup.yml
* initdb.yml
* wipe.yml

Once the DB is initialized the build exports the data to JSON files which are eventually used to populate a fresh new DB when the container starts.  

Currently the gradle build is not part of the overall build so that we can discuss if we wanna keep it separate or include it with the main build.


ddragosd:
I was able to successfully test this container with [docker-compose](https://github.com/openwhisk/openwhisk-devtools/pull/10) as well as Mesos and DCOS. As soon  as we close this PR the plan is to publish DCOS packages for installing Openwhisk components in DCOS in an easy fashion.

tools/db/docker/.gitignore
```diff
@@ -0,0 +1 @@
+import
```
rabbah: we've tried to have just one gitignore file at the top level of the repo.
ok to add there?

ddragosd: 👍

ddragosd: done.

ddragosd:
I've included the CouchDB docker build into the main build. I took special care in the build script to avoid re-building the container every time, but only when something changes in `./ansible` or `./tools/db` folders. I could have been more restrictive on _which ansible scripts to watch for changes_ but I wasn't sure it's needed. If you have suggestions I'm happy to include them.

ddragosd:
@rabbah @markusthoemmes I'm done with the changes I wanted to make in this PR.

markusthoemmes:
I'm honestly unsure about this. Why is this needed? Is it the docker way to create so highly specialized images instead of having reusable pieces at hand to configure?

ddragosd:
@markusthoemmes this would be needed to ease the deployment in Mesos and Kube clusters.

rabbah:
The criteria for merging into the main repo is that new features are continuously tested. If they're not, they will rot. For example, the multicloud bits are surely rotted and broken - and I would like to move them to a different repo but have not gotten around to it. In the last week or so, we jettisoned several features, code, and the like because it's not tested, including the mobile starter app. And more is planned. 

So while I think there is consistent agreement that we should provide images for the container on dockerhub (issue https://github.com/openwhisk/openwhisk/issues/171), adding these builds steps for a deployment that is closer tied to the [developer tools repository](https://github.com/openwhisk/openwhisk-devtools/) suggests that is the better home for this change.

ddragosd:
>  new features are continuously tested

ddragosd:
> Couldn't this be part of a seperate repository containing bits and pieces for different cloud providers?

@markusthoemmes we could; I'm not sure what's *cloud specific* in this couchdb image built reusing the ansible playbooks in this repo.

> adding these builds steps for a deployment that is closer tied to the developer tools repository suggests that is the better home for this change.

`docker-compose` flow in `openwhisk-devtools` repo it's currently building the images b/c #171 is not complete. The plan is to reuse those images from docker hub and skip the build. See: https://github.com/openwhisk/openwhisk-devtools/issues/3 
Then we'll need to find another home for building this container. 

To me it looked more natural to have it part of the main build which today builds all the other containers and also hosts the ansible scripts.

>  new features are continuously tested

good point. I wasn't sure how to integrate this so I left it off thinking we could have other flows (i.e. `travis` based) to do end-to-end testing including various deployments ( Ansible, Mesos, Kube, etc ).  I think for this container we could add a test showing that when it starts it initializes the Database as per the ansible config.

If regardless of the automated testing aspect we feel that architecturally it makes sense to keep it separate from this repo that builds all the other docker images, I'm happy to move down that path in order to make progress. We could have a special location on `openwhisk-devtools` repo, maybe a folder called `openwhisk-couchdb` and have this CouchDB docker image built there. We would also need to push it to Docker Hub from there as well. 

WDYT ?

markusthoemmes:
> good point. I wasn't sure how to integrate this so I left it off thinking we could have other flows (i.e. travis based) to do end-to-end testing including various deployments ( Ansible, Mesos, Kube, etc ). I think for this container we could add a test showing that when it starts it initializes the Database as per the ansible config.

That's where I think we could have one repo per supported provider? I.e: The main repo is Ansible, as-is. We can have `openwhisk-kubernetes` which includes bits and pieces needed there. Each of the repos have their respective travis builds. If you ask me, that's the cleanest solution to the problem and it will make it far easier to advance in the smaller repos. Thoughts?

ddragosd:
> If you ask me, that's the cleanest solution to the problem and it will make it far easier to advance in the smaller repos.

ddragosd:
> If you ask me, that's the cleanest solution to the problem and it will make it far easier to advance in the smaller repos. 

👍 💯  The common thing between `openwhisk-kubernetes`, `openwhisk-mesos-dcos`, `docker-compose`, etc are the docker images needed to be deployed. 

So if we feel we should build all containers but the couchDB one in this repo, then build `openwhisk-couchdb` as part of `openwhisk-devtools` I'm happy to go down this path as well.  Maybe in the future we'll build the other components in a standalone fashion, having the `controller` and `invoker` being built separately as well; looking from such an angle it makes sense to isolate the DB container given that it's not a *direct* dependency to the code, but to the deployment.

markusthoemmes:
Yes you are absolutely right. We'd love to fan out all the repositories at one point, just haven't gotten to the point thus far.

Sounds like a good direction to take 👍

markusthoemmes:
@ddragosd is this still an issue with our images published to Dockerhub already?

ddragosd:
@markusthoemmes there is still an issue with the `couchdb` container. See: https://github.com/openwhisk/openwhisk/issues/2117

rabbah:
would it make more sense to actually piggy back on the `docker commit` flow that @csantanapr created? In particular, before taking the snapshot
 
```
docker exec -it couchdb bash -c "apt-get install curl -y && curl 'localhost:5984/{local_subjects,local_whisks}/_all_docs?include_docs=true&attachments=true' -o db_#1.json"
```

then inject your `initdb` script and box the default docker entry point.

ddragosd:
@rabbah 👍  sounds like a simpler approach than the one I've taken. @csantanapr WDYT ?

csantanapr:
Hum @ddragosd I don't follow I was going to update the Jenkins script to do a full deployment including post deploy of the catalog actions this way the whisk.system packages like utils github Slack etc are present in db and then take the snapshot and push to docker hubbthe couchdb-snapshot

If I do this, do I still need to export this db_#1.json and then use the file how?

csantanapr:
Similar for edge I was going to build all CLI binaries and after deployment take a snapshot of edge and push to dockerhub and edge image

rabbah:
We need to pick this up and drive it through since it helps several downstream projects. @ddragosd or @csantanapr agree?

ningyougang:
@ddragosd ,currently does couchdb supports cluster deployment?

ningyougang:
You refered `Couch DB is a replicated only DB, not distributed. At Least at Adobe we're looking to replace it with another distributed DB later on`

Answer:regarding couchdb, i have found that it is very power for its replication feature, support `master-master` and `synchronize data by two directions(nodeA-->nodeB, nodeB->nodeA)`
So why do we find another distributed DB replace it, have any progress for it?

another point, if we use another `distributed DB`, has any influences on currently codes?

ddragosd:
Based on the conversations in [Slack](https://openwhisk-team.slack.com/messages/C3TPCAQG1/convo/C3TPCAQG1-1505839352.000618/?thread_ts=1505839352.000618) the direction is to create an Ansible playbook containing the same logic with the gradle build.

ningyougang:
@ddragosd ,as far as i know, currently, couchdb version is already changed from 1.6 to 2.0.
couchdb 2.0 supports `distributed` feature. So my question as below
1. Since couchdb 2.0 supports  `distributed`, but our `ansible playpook` seems don't support installing mutliple couchdb nodes.

ningyougang:
What's mean for `create an Ansible playbook containing the same logic with the gradle build.`?

ddragosd:
@ningyougang you're bringing very good points, which may not be covered as part of this PR. This PR simply intends to build a couchDB image, pre-initialized for OpenWhisk, so that we don't have to run ansible to configure it.

dgrove-oss:
Does the openwhisk/kube-couchdb image we are now publishing on dockerhub mean that this PR can be closed as obsolete?  

https://hub.docker.com/r/openwhisk/kube-couchdb/

rabbah:
I think so. Closing as no longer needed. Reopen if you disagree @ddragosd.

PR #1934, opened 2017-03-02
Update webactions.md

cfjedimaster:
Fix response in demo

### DIFF
--- a/docs/webactions.md
+++ b/docs/webactions.md
@@ -120,13 +120,6 @@ and invoking the action with a query parameter `name`, using the `.json` extensi
 $ curl https://${APIHOST}/api/v1/experimental/web/guest/demo/hello.json/response?name=Jane
 {
     "name": "Jane"
-    "__ow_meta_verb": "get",
-    "__ow_meta_headers": {
-        "accept": "*/*",
-        "user-agent": "curl/7.51.0",
-        ...
-    },
-    "__ow_meta_path": "/response"
 }
 ``` 
 


rabbah:
@csantanapr was the intention to show the input arguments to the action?
the way it's written now - thanks @cfjedimaster for pointing this out - it suggests that's the response from the action which is of course wrong.

should we show both? input + output explicitly (two different sections)?

rabbah:
I read this more carefully. The response was correct. But I suggest we clarify this text as suggested in this inline patch:

```
diff --git a/docs/webactions.md b/docs/webactions.md
index fbffd30..e8a51ff 100644
--- a/docs/webactions.md
+++ b/docs/webactions.md
@@ -109,19 +109,55 @@ Web actions bring some additional features that include:
 5. `Activation via multiple HTTP verbs`: a web action may be invoked via one of four HTTP methods: `GET`, `POST`, `PUT` or `DELETE`.
 
 
-The example below briefly sketches how you might use these features in a web action. Given an action `/guest/demo/hello` with the following body:
+The example below briefly sketches how you might use these features in a web action. Consider an action `/guest/demo/hello` with the following body:
 ```javascript
 function main(params) { 
-    return { 'response': params};
+    return { response: params };
 }
 ```
-and invoking the action with a query parameter `name`, using the `.json` extension to indicate a JSON response, and projecting the field `/response` will yield the following HTTP response:
+
+When this action is invoked as a web action, you can alter the response of the web action by projecting different paths from the result.
+For example, to return the entire object, and see what arguments the action receives:
+
 ```bash
-$ curl https://${APIHOST}/api/v1/experimental/web/guest/demo/hello.json/response?name=Jane
+$ curl https://${APIHOST}/api/v1/experimental/web/guest/demo/hello.json
 {
-    "name": "Jane"
+  "response": {
+    "__ow_meta_verb": "get",
+    "__ow_meta_headers": {
+      "host": "172.17.0.1",
+      "connection": "close",
+      "user-agent": "curl/7.43.0",
+      "accept": "*/*"
+    },
+    "__ow_meta_path": ""
+  }
 }
-``` 
+```
+
+and with a query parameter:
+```bash
+$ curl https://${APIHOST}/api/v1/experimental/web/guest/demo/hello.json?name=Jane
+{
+  "response": {
+    "name": "Jane",
+    "__ow_meta_verb": "get",
+    "__ow_meta_headers": {
+      "host": "172.17.0.1",
+      "connection": "close",
+      "user-agent": "curl/7.43.0",
+      "accept": "*/*"
+    },
+    "__ow_meta_path": ""
+  }
+}
+```
+
+and to project just the name (as text):
+```bash
+$ curl https://${APIHOST}/api/v1/experimental/web/guest/demo/hello.text/response/name?name=Jane
+Jane
+```
```

cfjedimaster:
Ah ok. So - do I do anything now? Seems like your change should just take over.

rabbah:
K i'll open a pr.

rabbah:
Superseded by #1934.

csantanapr:
I was confuse on what was @cfjedimaster looking to clarify. 
I did the example in a way to cover many concepts in one , I agree siting into multiples look like good approach by #1966

cfjedimaster:
@csantanapr the original text had the wrong result for the call that was made - that was the confusion.

csantanapr:
It has the correct result give it a try, that's what @rabbah also pointed out
The examples returns all parameters regardless of the query parameter you pass down

cfjedimaster:
Ok - not sure what to say here then. My original PR did fix a bug in the docs, @rabbah agreed. The result was NOT right, I stand by that, but @rabbah has changed the docs to be more appropriate I think and my mod no longer applies. If you really want me to explain why it was wrong, _at the time_, I can, but it's a moot point.

cfjedimaster:
Looking again - maybe I'm wrong too. But again - *moot* point, @rabbah has made it better, so let's just go with his stuff. :)

csantanapr:
Yes I agree moot point, let's not discuss how I'm right 😜

PR #1935, opened 2017-03-02
Reduce time to execute tests by 10 minutes

cbickel:
Do we really need this test?
If yes, what about paralleling it?

### DIFF
--- a/tests/src/whisk/core/limits/MaxActionDurationTests.scala
+++ b/tests/src/whisk/core/limits/MaxActionDurationTests.scala
@@ -43,15 +43,18 @@ class MaxActionDurationTests
     implicit val wskprops = WskProps()
     val wsk = new Wsk
 
-    Map("node" -> "helloDeadline.js", "python" -> "timedout.py", "java" -> "timedout.jar").foreach {
-        case (k, name) =>
-            s"$k action" should "run up to the max allowed duration" in withAssetCleaner(wskprops) {
-                (wp, assetHelper) =>
+    // swift is not tested, because it uses the same proxy like python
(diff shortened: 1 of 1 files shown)


rabbah:
yes we need it because we had a bug where actions were not running for the full 5 minutes.

rabbah:
note it doesn't run on travis, only internal jenkins.

markusthoemmes:
Why don't we do this for swift?

rabbah:
we could - iirc i explained this away as python proxy is the same one used for swift, so it was low risk because the bug/issues was from the node proxy.

markusthoemmes:
Expected that answer.

Change is still LGTM. Parallelising this is goodness.

tests/src/whisk/core/limits/MaxActionDurationTests.scala
```diff
@@ -43,15 +43,17 @@ class MaxActionDurationTests
     implicit val wskprops = WskProps()
     val wsk = new Wsk
 
-    Map("node" -> "helloDeadline.js", "python" -> "timedout.py", "java" -> "timedout.jar").foreach {
-        case (k, name) =>
-            s"$k action" should "run up to the max allowed duration" in withAssetCleaner(wskprops) {
-                (wp, assetHelper) =>
+    s"node-, python, and java-action" should "run up to the max allowed duration" in withAssetCleaner(wskprops) {
```
rabbah: nit: drop `s` no string interpolation here `s"node-, python, and java-action" `

rabbah:
already pg'ed?

Issue #1886, opened 2017-02-21
rejections from controller are not specific enough


rabbah:
The rejection handler in the controller favors the first rejection vs a potentially more specific one later in the route matching. This is attributed to this method in `BasicHttpService`:

```
/** Rejection handler to terminate connection on a bad request. Delegates to Spray handler. */
    protected def customRejectionHandler(implicit transid: TransactionId) = RejectionHandler {
        case rejections =>
            logging.info(this, s"[REJECT] $rejections")
            BasicHttpService.customRejectionHandler.apply(rejections)
    }
```

The `rejections` list should be scanned and prioritized to take either the `last` rejection from the list, or a `CustomRejection` if present.

starpit:
another odd variant of this:

if i nodejs request with `json: true, body: data` to a web-exported route `/my-action.text`, then i get back the "requires authentication" error. whereas if i use `headers: {content-type: 'application/json'}, body: JSON.stringify(data)` then things are happy.

this is i assume due to `json:true` also setting up the `accept` headers? i'm not sure, but this one was pretty opaque.

dubeejw:
@starpit, this will happen as the content type must be JSON or Form Data.

starpit:
A JSON:true option will set the content type to JSON. The puzzling thing is that whisk still rejects. I am guessing that this due to a mismatch of Accept. Because we are invoking with a .Text web exported suffix.

dubeejw:
Can you post here the HTTP header being sent when you do a request with `json: true`?

rabbah:
@starpit that's consistent with this issue.

starpit:
@dubeejw i just verified: the difference seems to boil down to an `Accept: application/json' header, when invoking as `.text`

rabbah:
@l2fprod, you're not alone :)

rabbah:
@dubeejw want to take this one?

starpit:
@rabbah the weird thing to me is that text *is* valid JSON. so an Accept: application/json should not be incompatible with a .text invocation?

rabbah:
there is another related issue: when the `entity` is empty, this will run twice:
```
entity(as[Option[JsObject]]) {
    body => process(body)
} ~ entity(as[FormData]) {
     form => process(Some(form.fields.toMap.toJson.asJsObject))
}
```

dubeejw:
@starpit, it'd be very helpful if you posted the exact options you are using to send the request. Below are some options I experimented with and the result.

[...]

PR #1939, opened 2017-03-02
Return Appropriate Error Message when a Web Action is Invoked with an Unsupported Content-type

dubeejw:
- Inform user's when an unsupported content-type is passed to a Web Action

Closes: https://github.com/openwhisk/openwhisk/issues/1886

### DIFF
--- a/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
+++ b/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
@@ -117,6 +117,7 @@ object Messages {
     def invalidMedia(m: MediaType) = s"Response is not valid '${m.value}'."
     val contentTypeExtentionNotSupported = """Extension must be specified and one of [".json", ".html", ".http", ".text"]."""
     val contentTypeNotSupported = s"Content-type must be ${MediaTypes.`application/json`} or ${MediaTypes.`application/x-www-form-urlencoded`}."
+    def invalidAcceptType(m: MediaType) = s"Response of ${m.value} does not correspond with accept header."
 
     val responseNotReady = "Response not yet ready."
     val httpUnknownContentType = "Response did not specify a known content-type."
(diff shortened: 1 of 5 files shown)


common/scala/src/main/scala/whisk/http/ErrorResponse.scala
```diff
@@ -114,8 +114,8 @@ object Messages {
     /** Error for meta api. */
     val propertyNotFound = "Response does not include requested property."
     def invalidMedia(m: MediaType) = s"Response is not valid '${m.value}'."
-    val contentTypeNotSupported = """Content type must be specified and one of ["json", "html", "http", "text"]."""
-
+    val contentTypeExtentionNotSupported = """Content type must be specified and one of ["json", "html", "http", "text"]."""
```
rabbah: while you're at it - mind changing ` contentTypeExtentionNotSupported = """Content type must` to ` contentTypeExtentionNotSupported = """Extension must`?

maybe also prefix each of the extensions with a `.` (dot).

common/scala/src/main/scala/whisk/http/ErrorResponse.scala
```diff
@@ -114,8 +114,8 @@ object Messages {
...
+    val contentTypeNotSupported = "Content type must be appication/json or application/x-www-form-urlencoded."
```
rabbah: typo - maybe better to use "`application/json`" from `MediaTypes` instead.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -330,7 +330,7 @@ trait WhiskMetaApi
                     body => process(body)
                 } ~ entity(as[FormData]) {
                     form => process(Some(form.fields.toMap.toJson.asJsObject))
-                }
+                } ~ terminate(BadRequest, Messages.contentTypeNotSupported)
```
rabbah: please add a test.

dubeejw:
PG2 1120, approved.

rabbah:
Nice. Checking the issue again I think we covered the ingress but not an error on egress: https://github.com/openwhisk/openwhisk/issues/1886#issuecomment-282353511 we should add one more test where accept header is set to an incompatible type (json and return text).

rabbah:
superseded by #2004.

PR #1943, opened 2017-03-04
retry in test cleanup routine if package deletion fails with conflict

jeremiaswerner:


jeremiaswerner:
PG2/1144 🔵

tests/src/common/WskTestHelpers.scala
```diff
@@ -89,10 +75,19 @@ trait WskTestHelpers extends Matchers {
...
-                        cli.sanitize(n)(wskprops)
+                    cli match {
+                        case _: WskPackage =>
+                            val rr = if (delete) cli.delete(n)(wskprops) else cli.sanitize(n)(wskprops)
+                            rr.exitCode match {
+                                case CONFLICT =>
+                                    // retry sanitization on a package since there may be a list (view)
+                                    // operation that requires a retry for eventual consistency
+                                    whisk.utils.retry({
+                                        if (delete) cli.delete(n)(wskprops) else cli.sanitize(n)(wskprops)
```
rabbah: comment says retry sanitize on a package but this will not retry since sanitize does not fail with an exception.
Should the `if (delete)` predicate guard the retry at the outermost level instead?

jeremiaswerner:
@rabbah adopted your suggestion

tests/src/common/WskTestHelpers.scala
```diff
@@ -89,10 +75,24 @@ trait WskTestHelpers extends Matchers {
...
+                                case CONFLICT =>
+                                    // retry deletion on a package since there may be a list (view)
+                                    // operation that requires a retry for eventual consistency
+                                    // don't retry sanitize because it does not throw an exception and silently ignores any failures
+                                    if (delete) {
+                                        whisk.utils.retry({
+                                            cli.delete(n)(wskprops)
+                                        }, 5, Some(1 second))
+                                    } else {
+                                        cli.sanitize(n)(wskprops)
```
rabbah: this does retry sanitize once more - intended?

tests/src/test/scala/common/WskTestHelpers.scala
```diff
@@ -89,10 +75,20 @@ trait WskTestHelpers extends Matchers {
             // delete assets in reverse order so that was created last is deleted first
             val deletedAll = assetsToDeleteAfterTest.reverse map {
                 case ((cli, n, delete)) => n -> Try {
-                    if (delete) {
-                        cli.delete(n)(wskprops)
-                    } else {
-                        cli.sanitize(n)(wskprops)
+                    cli match {
+                        case _: WskPackage =>
+                            val rr = if (delete) cli.delete(n)(wskprops) else cli.sanitize(n)(wskprops)
```
markusthoemmes: You can move this even further up, no?

```
cli match {
    case _: WhiskPackage if delete =>
        val rr = cli.delete(n)(wskprops)
        rr.exitCode => match {
            case CONFLICT => whisk.utils.retry(cli.delete(n)(wskprops), 5, Some(1.second))
            case _ => rr
        }
    case _ => if (delete) cli.delete(n)(wskprops) else cli.sanitize(n)(wskprops)
}
```

jeremiaswerner:
PG2-1227

PR #1946, opened 2017-03-05
Better handling for zip actions with incorrect structure

### LINKED ISSUE #1945: doc should mention that Windows Explorer zip will make invalid zipped actions
To create a zipped action on Windows, the most natural thing to do is to select the files in Explorer, right-click, and choose the archive files option. Unfortunately, in this case Windows creates a top-level folder inside the zip which, at least in the case of Node.js actions, will not be usable by OW as the root of the zip is assumed to contain `package.json`.

This appears to be a common enough issue that is probably warrants explicit documentation.

bjustin-ibm:
resolves #1945 

In addition to updating the doc to specifically warn against using the Windows Explorer zip action, update the Node.js runtime to fail on `/init` with a specific message telling the user what is wrong with their zip action.

### DIFF
--- a/core/nodejsActionBase/runner.js
+++ b/core/nodejsActionBase/runner.js
@@ -51,6 +51,10 @@ function NodeActionRunner() {
         if (message.binary) {
             // The code is a base64-encoded zip file.
             return unzipInTmpDir(message.code).then(function (moduleDir) {
+                if(!fs.existsSync(path.join(moduleDir, 'package.json'))) {
+                    return Promise.reject('package.json must be located at the root of a zipped action.')
+                }
+
                 try {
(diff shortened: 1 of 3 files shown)


bjustin-ibm:
PG2 1135

bjustin-ibm:
@rabbah turns out there was an existing test for this case. I updated it to expect the new error message.

rabbah:
great - is  tests/dat/actions/nestedFolderZipAction.zip  needed then?

bjustin-ibm:
> great - is tests/dat/actions/nestedFolderZipAction.zip needed then?

Nope... removed.

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/69b855e3e3c38a3d98e109607fa5d85be2af2b43.

PR #1948, opened 2017-03-05
Utilities to read Exec runtimes from manifest.

rabbah:
Add runtimes manifest to ansible group vars.
Write out manifest to whisk.properties and make available through WhiskConfig.


rabbah:
PG3/350 🔵

perryibm:
merged

PR #1952, opened 2017-03-06
Allow pre-compiled Swift binaries.

rabbah:
This allows zip files for swift actions which contain a precompiled binary.
PG2/1152 🔵.


tests/src/actionContainers/SwiftActionContainerTests.scala
```diff
@@ -196,7 +198,29 @@ class SwiftActionContainerTests extends BasicActionRunnerTests with WskActorSyst
         })
     }
 
-        it should "properly use KituraNet and Dispatch" in {
+    it should "support pre-compiled binary in a zip file" in {
```
bjustin-ibm: It would be nice to have a test that has a zip that uses libraries that aren't already included in the Swift runtime.

rabbah: good idea - want to take the pr over?

rabbah: @bjustin-ibm do you have an example?
the build script doesn't support this either, right @paulcastro?
if so, we'd need some one off zip file that does this, ... thoughts?

akrabat: Given how Swift isn't ABI stable, there's a risk that if you included a one-off zip file, it would have to be re-created whenever the Swift compiler version is updated. 

This could be a bit brittle for a test of this feature?

rabbah: My rationale: 1. Yes the test will break and so we'd have to regenerate the zip file. But this might require tweaking the build steps anyway for the new swift version. 2. I expect we all automate the binary building in a systematic way for native actions in the near future.

akrabat: Fair point.

bjustin-ibm:
Ah ok... looking at the doc update, this change is not what I thought it was :) I was expecting/assuming that the user could create their own `Package.swift` with 3rd party deps listed and end up with a zip that contains that whole project pre-compiled and ready to run as an OW action. This change gets us pretty close to that, but I think the end goal would be to add a little more automation to the swift3action container such that the user could run something like:

```bash
$ docker run "$(pwd):/owexec" openwhisk/swift3action createZip
```

Where `$(pwd)` is expected to contain `Package.swift` and all user-written Swift code. Upon completion, the user would find the precompiled action.zip sitting in their cwd.

So... two questions: 
- Is this a reasonable end goal?
- If so, should this PR be merged as-is, or should we wait until the above is supported?

@paulcastro @rabbah @csantanapr @akrabat , thoughts?

rabbah:
>    Is this a reasonable end goal?

yes, i consider that the same as " I expect we all automate the binary building in a systematic way for native actions in the near future." from my earlier comment.

>    If so, should this PR be merged as-is, or should we wait until the above is supported?

as is - because it is already useful.

csantanapr:
> Is this a reasonable end goal?

same as https://github.com/openwhisk/openwhisk/pull/1952#issuecomment-284728565

yes using docker run will be the lowest denominator, but then adding on top, like helper actions that automate this, in conjuction with a `build` cli plugin `$wsk build ..` that produces the pre-compiled zip. To the point when we have uber `wsk deploy` that can do the build from a git url 

>If so, should this PR be merged as-is, or should we wait until the above is supported?

As is give us the ability to lower the cold latency on swift, that is seating around the 3 seconds.
Anything else builds on top

PR #1962, opened 2017-03-06
Fix bug in cache.

rabbah:
Fixed bug with @starpit where read listener, under certain conditions: 
1. did not complete the promise at all (if readDone failed to transition cache state)
2. completes promise with a failure when it shouldn't (write finished before read)


rabbah:
note: temporarily bumped niters and threads for pgs.

rabbah:
pg1/1229 pg2/1153 pg3/353 🔵.

rabbah:
pg3/356 and pg2/1156 are :large_blue_circle:

Issue #197, opened 2016-04-08
finalize ansible playbook/roles


lehoanganh:
This issue covers the following tasks
- [x] introduce `mode` for each role, standard is **deploy**, overwrite to **clean** to undeploy the component
- [x] merge controller and loadbalancer roles
- [x] write whisk.properties file from template

PR #1971, opened 2017-03-07
Replace sh-script on wiping db with ansible

cbickel:


ansible/tasks/db/actionsDb.yml
```diff
@@ -0,0 +1,33 @@
...
+- name: create {{ db.whisk.actions }} with {{ db_provider }}
+  uri:
+    url: "{{ db_protocol }}://{{ db_host }}:{{ db_port }}/{{ db.whisk.actions }}"
+    method: PUT
+    status_code: 200,201
+    user: "{{ db_username }}"
+    password: "{{ db_password }}"
+    force_basic_auth: yes
+
+- name: Create whisks-Design document in {{ db.whisk.actions }} with {{ db_provider }}
```
markusthoemmes: consistent casing please: all lowercase

rabbah:
Removed @domdom82 as assignee (he is not an Apache committer for this poddling 😢). Assigned to @markusthoemmes instead.

tools/db/loadTransientDBViews.sh
```diff
@@ -1,105 +0,0 @@
...
-# drop and recreate the transient whisk cloudant views.
-# NOTE: before editing this file, review the notes in
-# whisk.core.entity.WhiskStore as any changes here to the views
-# may require changes to the supporting query methods.
-#
-
-SCRIPTDIR="$(cd $(dirname "$0")/ && pwd)"
-source "$SCRIPTDIR/common.sh"
-
-function addRevision() {
```
rabbah: are we retaining the ability to reload the views without wiping the db (as this script allowed)?

ansible/tasks/db/actionsDb.yml
```diff
@@ -0,0 +1,33 @@
...
+- name: create {{ db.whisk.actions }} with {{ db_provider }}
+  uri:
+    url: "{{ db_protocol }}://{{ db_host }}:{{ db_port }}/{{ db.whisk.actions }}"
+    method: PUT
+    status_code: 200,201
+    user: "{{ db_username }}"
+    password: "{{ db_password }}"
+    force_basic_auth: yes
+
+- name: create whisks-Design document in {{ db.whisk.actions }} with {{ db_provider }}
```
rabbah: this creates the design doc for a fresh db - I don't see a task to update the design doc without wiping the db.

cbickel:
PG2#1174 with the requested changes is running.

ansible/tasks/wipeDatabase.yml
```diff
@@ -5,3 +5,5 @@
 - include: writeWhiskProperties.yml
```
markusthoemmes: Can we take this out (see the comment)

ansible/group_vars/all
```diff
@@ -85,6 +85,8 @@ db:
   authkeys:
   - guest
   - whisk.system
+  whisk:
+    actions: "{{ db_prefix }}whisks"
```
markusthoemmes: If this does in fact work (I expect it not too): Can we move all database name definitions here?

ansible/tasks/recreateViews.yml
```diff
@@ -0,0 +1,4 @@
+---
+# Recreates all views in all databases.
+
+- include: db/actionsDbViews.yml
```
markusthoemmes: Why is this "duplicated" to here?

cbickel: This is the script, which creates all views in all databases.
Currently we create only views in one Database.
After the seperation of the actions- and activations database, this single script will recreate all views in all transient dbs.

ansible/tasks/db/actionsDb.yml
```diff
@@ -0,0 +1,20 @@
+---
+# Delete and recreate the actions-Db for whisk with all DesignDocs.
+
+- name: Delete the {{ db.whisk.actions }} with {{ db_provider }}
```
markusthoemmes: lowercase please

ansible/tasks/wipeDatabase.yml
```diff
@@ -2,7 +2,6 @@
 # Wipe transient databases. You should know what you are doing here.
 
 # make sure we have whisk.properties because wipeTransientDBs.sh needs it
```
markusthoemmes: comment is obsolete

rabbah:
@ddragosd do these changes affect your deployment tools with docker compose?

ansible/templates/whisk.properties.j2
```diff
@@ -80,8 +80,8 @@ db.port={{ db_port }}
 db.username={{ db_username }}
 db.password={{ db_password }}
 db.prefix={{ db_prefix }}
-db.whisk.actions={{ db_prefix }}whisks
-db.whisk.auths={{ db_auth }}
+db.whisk.actions={{ db.whisk.actions }}
```
ddragosd: I'm seeing that this property keeps changing.  See https://github.com/openwhisk/openwhisk/issues/197 .

Since it has a potential breaking impact in the ENV VARs, that would have to change too, it would be good to avoid changing it unless there's critical functionality that depends on it.

ansible/group_vars/all
```diff
@@ -81,10 +81,12 @@ nginx:
     adminportal: 8443
 
 db:
-  immortal: subjects
   authkeys:
   - guest
   - whisk.system
+  whisk:
```
markusthoemmes: Why the extra nesting underneath `whisk`? should rather be something like `db.names.actions`?

ansible/templates/whisk.properties.j2
markusthoemmes: How does this break ENV vars? The names of the databases stay exactly the same.

ddragosd:
Yes it does in the current `master` branch : https://github.com/openwhisk/openwhisk-devtools/blob/0326d39c7b76207815a5fb551846423d124d5d89/docker-compose/Makefile#L122 

But we have a PR to use the `ansible` script to setup a DB in this PR: https://github.com/openwhisk/openwhisk-devtools/pull/10

Actually the more important PR is #1928 . If we could review this we're solving more problems.

ansible/templates/whisk.properties.j2
ddragosd: > The names of the databases stay exactly the same.

If it's stays the same it's good. I missed this. I was rather referring on not introducing a breaking change.

cbickel:
PG3#386 is green.

ansible/group_vars/all
cbickel: Of cause, this is working. Subjets DB is added as well now.

ansible/tasks/initdb.yml
```diff
@@ -2,20 +2,20 @@
 # This task will initialize the immortal DBs in the database account.
 # This step is usually done only once per account.
 
-- name: check if the immortal {{ db_auth }} db with {{ db_provider }} exists?
+- name: check if the immortal {{ db.whisk.auth }} db with {{ db_provider }} exists?
```
domdom82: @cbickel have you checked that the vars are replaced? I can recall having to put ".." around the test to have them replaced by ansible.

ansible/tasks/initdb.yml
```diff
@@ -2,20 +2,20 @@
...
   uri:
-    url: "{{ db_protocol }}://{{ db_host }}:{{ db_port }}/{{ db_auth }}"
+    url: "{{ db_protocol }}://{{ db_host }}:{{ db_port }}/{{ db.whisk.auth }}"
```
domdom82: see above comment. usually you have to put quotes to make ansible expand the vars

ansible/tasks/initdb.yml
cbickel: In our playground it has been processed correctly:
```
TASK [check if the immortal whiskjenkins_playground3_subjects db with Cloudant exists?] ***
Friday 10 March 2017  05:34:28 -0500 (0:00:00.088)       0:00:00.890 ********** 
ok: [ansible]
```

ansible/tasks/initdb.yml
markusthoemmes: There are quotes around it.

domdom82: @markusthoemmes I know. I was referring to this one as an example for quoting. seems like ansible treats task titles different then.

Issue #1072, opened 2016-08-16
RAS - Rework usage of consul health protocol and kv store


rabbah:
(update from jeremiaswerner)

This epic will cover the implementation required to rework our usage of the consul health checks and the kv store for user throttles.

health check:
- we will remove the registrator component from the deployment since our environment is not that dynamic and we know when we add a new management container
- the service registry in consul will be managed during deployment time by ansible playbooks.
- we will introduce the notion of service (invoker) and service instance (invoker0, invoker1,...) 
- the health checks will be used as today to check the availability of the service instances
- the load balancer will rely on the health check in consul rather than using the own protocol in the kvstore

kv store for throttling:
- controller will maintain the state of the number of invocations (total and per user)
- controller will increment when activtion is dropped to kafka
- controller will decrement when active ack is retrieved
- invoker will only require to do active ack
- optional:
  - controller will persist from time to time in redis to share the knowledge with other controllers (preparing for HA/3)
  - we need to introduce redis as a distributed cache

rabbah:
@jeremiaswerner @domdom82 @markusthoemmes some of the issues/prs in flight should perhaps be linked to this epic, no?

rabbah:
Many of these things are already done. HA/3 is in the works.

PR #1976, opened 2017-03-07
Remove registrator and use explicit consul registration instead

markusthoemmes:
This gets rid of using registator to "dynamically" register services in consul and uses explicit registration instead.
Reduces general deployment overhead and an exceptional container.

Completes #1072 

TBD:

- [ ] Multiple components per "service-name" (like multiple invokers)
- [ ] Tests


ansible/roles/invoker/tasks/clean.yml
```diff
@@ -38,3 +38,9 @@
     path: "{{ whisk_logs_dir }}/invoker{{play_hosts.index(inventory_hostname)}}"
     state: absent
   become: true
+
+- name: unregister invoker with consul
+  consul:
+    service_name: invoker
+    state: absent
+  ignore_errors: True
```
markusthoemmes: memo to self: needs to be a specific invoker

markusthoemmes:
Going a different route with getting rid of consul alltogether.

Issue #428, opened 2016-05-13
Go CLI Receives a Certificate Error from the Local Server


csantanapr:
From @dubeejw : 

When running Go CLI commands that communicate with the local server a certificate error occurs. This can be solved by enabling the InsecureSkipVerify property for TLS in go-whisk/client.go NewClient(). However, this may not be ideal for security reasons.

Error: 
Get https://172.17.0.1/api/v1/namespaces/whisk.system/packages: x509: certificate has expired or is not yet valid

csantanapr:
From @rabbah : 
deally we just add -i, --insecure option to not verify SSL certificate of API endpoint.

rabbah:
ideally we just add -i, --insecure option to not verify SSL certificate of API endpoint.

csantanapr:
-i is implemented in Go CLI Now closing

rabbah:
What is the default 
Not sure what the GO default is - making insecure false for python CLI is wrong because all of the devs will need to use -i. This is probably something we can put in the config file (for the GO cli). I certainly don't want to type -i all the time.

csantanapr:
Yeah I agree
As part lf a local build we can set the property
We can added to the propeties like
wsk property set --i true

mdeuser:
Just thinking... Do we really want to put a security override in an external properties file?  If that file is hacked or accidentally carried forward, the intended security check can be compromised.

If we leave it as a CLI option only, developers can use `alias wsk='wsk -i'`

rabbah:
i was thinking this is saved in the user's wskprops file. `wsp property set --insecure true`.

csantanapr:
I agree with @rabbah on https://github.com/openwhisk/openwhisk/issues/428#issuecomment-229198377
I can add `wsk property set --insecure true` and we can add to the docs, this way it also works for windows users.

I don't think windows has "alias", it's being a while since I used "windows", and we should not assume they will have `bash` or `cygwin`

rabbah:
done although #1570 is left.

g12megha:
When i try to curl to list the triggers under a namespace by providing the openwhisk-self signed cert along the parameter --cacert, it throws the error of unmatching OPENWHISK_API_HOST. When i try to search this variable, i cannot find it .Where is it defined?  
![wskerror](https://cloud.githubusercontent.com/assets/3980709/23638359/25672158-0296-11e7-82bb-fcf7637b6aa7.JPG)

rabbah:
[...]

PR #1977, opened 2017-03-07
Fix CN host name when generating certificate.

rabbah:
If the API host name is not defined, default to "localhost". Fix lookup of API host from the environment.


csantanapr:
PG2 1166 q

csantanapr:
merged aa9cfcf10187ea08ece2daa2c7e80ad9353f5d05

csantanapr:
Closes #428

Issue #1908, opened 2017-02-25
Get https://172.17.0.1/api/v1/namespaces: x509: cannot validate certificate for 172.17.0.1 because it doesn't contain any IP SANs


rhdotone:
Running :
./bin/wsk property set --namespace guest 

error: Authenticated user does not have namespace 'guest '; set command failed: Get https://172.17.0.1/api/v1/namespaces: x509: cannot validate certificate for 172.17.0.1 because it doesn't contain any IP SANs

However if I run in insecure mode, 

./bin/wsk -i property set --namespace guest

It sets fine.

Due to this when I try to set "Service Node Config" in Node-RED UI, I also get 

[Error: Error encountered calling OpenWhisk: Error: self signed certificate]

I am assuming it is because I cannot setup insecure from Node-RED

Is there anyway I can fix this certificate issue in openwhisk?

rabbah:
You can use certbot to generate a signed certificate https://certbot.eff.org/.

rabbah:
Did you try to disable TLS for nodered? :

Try that by exporting NODE_TLS_REJECT_UNAUTHORIZED = "0" in your environment.

tardieu:
@rabbah I encountered the same issue. The workaround does the trick. But is the certificate error the expected behavior? If it is, the doc should discuss it.

rabbah:
They do but not in the best place: https://github.com/openwhisk/openwhisk/tree/master/tools/vagrant#ssl-certificate-configuration-optional.

See this discussion for a longer debate about this: https://github.com/openwhisk/openwhisk/issues/1570 (and related pr https://github.com/openwhisk/openwhisk/pull/1838).

rabbah:
Closed by #2000.

PR #1989, opened 2017-03-09
Document how to use the non-SSL apihost for Vagrant

akrabat:
When working locally, it’s easier to use the non-SSL `apihost` for the `wsk` tool so that you don’t have to keep remembering to use the `-I` flag.

### DIFF
--- a/tools/vagrant/README.md
+++ b/tools/vagrant/README.md
@@ -76,7 +76,7 @@ wsk -i action invoke /whisk.system/utils/echo -p message hello --blocking --resu
 }
 ```
 
-**Tip:** You need to use the `-i` switch as the default SSL certificate used by the Vagrant installation is self-signed. Alternatively, you can configure you __apihost__ to use the non-SSL interface:
+**Tip:** You need to use the `-i` switch as the default SSL certificate used by the Vagrant installation is self-signed. Alternatively, you can configure your __apihost__ to use the non-SSL interface:
 
 ```
 wsk -i property set --apihost http://192.168.33.13:10001 --auth `vagrant ssh -- cat openwhisk/ansible/files/auth.guest
--- a/tools/vagrant/README.md
+++ b/tools/vagrant/README.md
@@ -75,7 +75,17 @@ wsk -i action invoke /whisk.system/utils/echo -p message hello --blocking --resu
     "message": "hello"
 }
 ```
-**Tip:** To connect to a different host API (i.e. bluemix.net) with the CLI, you will need to 
+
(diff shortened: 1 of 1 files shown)


rabbah:
I think we should put that here https://github.com/openwhisk/openwhisk/blob/master/docs/cli.md and cross link. What do you think?

rabbah:
Closes https://github.com/openwhisk/openwhisk/issues/1908 perhaps?

tools/vagrant/README.md
```diff
@@ -75,7 +75,17 @@ wsk -i action invoke /whisk.system/utils/echo -p message hello --blocking --resu
     "message": "hello"
 }
 ```
-**Tip:** To connect to a different host API (i.e. bluemix.net) with the CLI, you will need to 
+
+**Tip:** You need to use the `-i` switch as the default SSL certificate used by the Vagrant installation is self-signed. Alternatively, you can configure you __apihost__ to use the non-SSL interface:
```
rabbah: Configure "your"

akrabat:
> I think we can put much of that here https://github.com/openwhisk/openwhisk/blob/master/docs/cli.md and cross link. What do you think?

My gut feeling is that non-SSL installations shouldn't happen in any case other than development of OpenWhisk itself. In any other deployment, the assumption should be that you've installed a proper SSL cert.

As such, it seems Vagrant specific and also, I wanted the full line there so I could copy/paste it next time I need it!

rabbah:
Fair although local deployments include vagrant, docker machine, or ununtu vms directly. But what I think you're pointing out is that we sort of buried all the info here anyway atm.

akrabat:
Yeah - there's a strong argument for refactoring the development documentation in general.

akrabat:
(Rebased against latest master)

csantanapr:
merged ef589554e6de96bf2747af69dda8e64ee040ccf6

tools/vagrant/README.md
```diff
@@ -75,7 +75,17 @@ wsk -i action invoke /whisk.system/utils/echo -p message hello --blocking --resu
     "message": "hello"
 }
 ```
-**Tip:** To connect to a different host API (i.e. bluemix.net) with the CLI, you will need to 
+
+**Tip:** You need to use the `-i` switch as the default SSL certificate used by the Vagrant installation is self-signed. Alternatively, you can configure your __apihost__ to use the non-SSL interface:
+
+```
+wsk -i property set --apihost http://192.168.33.13:10001 --auth `vagrant ssh -- cat openwhisk/ansible/files/auth.guest
```
akrabat: > Also isn't the whole point of this command to NOT use the `-i` and it's the line.

That's a good question. I assumed it was required as we haven't told it to use a non-secure host. However, thinking about it, I'm guessing that `wsk property set` works locally when setting the apihost so the `-i` is immaterial?

PR #1992, opened 2017-03-09
Include zip in swift3Action

### LINKED ISSUE #1991: Add zip to openwhisk/swift3action Docker image
The most reliable way to create a Swift binary to run on OpenWhisk is to use the same Docker container that OpenWhisk itself uses (`openwhisk/swift3action`).

The steps to compile are documented at [Packaging an action as a Swift executable](https://github.com/openwhisk/openwhisk/blob/master/docs/actions.md#packaging-an-action-as-a-swift-executable) and the first step is to install zip:

```
apt-get install -y zip
```

This takes a little time as it has to pull it from the network, and you have to do this every time you compile every action.

It would therefore be convenient if `openwhisk/swift3action` had zip installed by default.

akrabat:
This makes it easier to compile Swift actions locally for upload to OpenWhisk as zip files containing the Action binary. 

In terms of space, this makes very little difference as the image is based on Ubuntu 14.04. Looking at `docker images` locally, the current image is 1.45GB and one with zip included is 1.46GB.

Closes #1991

### DIFF
--- a/core/swift3Action/Dockerfile
+++ b/core/swift3Action/Dockerfile
@@ -12,6 +12,9 @@ RUN apt-get -y purge \
 # Upgrade and install Swift dependencies
  && apt-get -y install --fix-missing build-essential curl wget libicu-dev \
 \
+# Install zip for compiling Swift actions
+ && apt-get -y install zip \
+\
 # Clean up
  && apt-get clean
 


rabbah:
Travis failure unrelated. Confirmed works as expected.

Issue #1997, opened 2017-03-10
Can not deploy OpenWhisk on Cent OS


style95:
## Environment details:

* CentOS 7.3
* Docker 1.12.6
* Ansible 2.1.2

## Steps to reproduce the issue:

1.   gradlew distDocker
2.  ansible-playbook apigateway.yml 
3.   ansible-playbook openwhisk.yml
   

## Provide the expected results and outputs:

```
$ echo PING | nc -w 3 -i 1 localhost 6379
+PONG

$ echo ruok | nc -w 3 -i 1 localhost 2181
imok
```

## Provide the actual results and outputs:

```
$ echo PING | nc -w 3 -i 1 localhost 6379
Ncat: Idle timeout expired (1000 ms).

$ echo ruok | nc -w 3 -i 1 localhost 2181
Ncat: Idle timeout expired (1000 ms).
```

## Additional information you deem important:
* Current health-check logic for zookeeper and redis does not work on CentOS.

PR #1998, opened 2017-03-10
Fix health check logic for zookeeper and redis to make it work on CentOS

style95:
This is to make health check logic for zookeeper and redis work on Cent OS.

It has been tested on following OSs:
* CentOS 7.3
* OSX 10.11
* Ubuntu 16.04

Ansible deployment worked as expected.

Closes: #1997

### DIFF
--- a/ansible/roles/kafka/tasks/deploy.yml
+++ b/ansible/roles/kafka/tasks/deploy.yml
@@ -15,7 +15,7 @@
       - "{{ zookeeper.port }}:2181"
 
 - name: wait until the Zookeeper in this host is up and running
-  action: shell echo ruok | nc -w 3 -i 1 {{ inventory_hostname }} {{ zookeeper.port }}
+  action: shell (echo ruok; sleep 1) | nc {{ inventory_hostname }} {{ zookeeper.port }}
   register: result
   until: (result.rc == 0) and (result.stdout == 'imok')
   retries: 36
--- a/tools/health/monitorUtil.py
+++ b/tools/health/monitorUtil.py
@@ -44,15 +44,17 @@ def checkLoop(name, checker, delay):
 # Runs the given command with optional input.
 # The command is run in a blocking fashion and the return code
 # and output (both stdout/stderr combined) returned as a pair.
-def run(cmd, inData=""):
+def run(cmd, inData="", isShell=False):
     try:
         p = subprocess.Popen(cmd, stdout=subprocess.PIPE,
+                             shell=isShell,
                              stdin=subprocess.PIPE,
                              stderr=subprocess.STDOUT)
         if (inData != ""):
             p.stdin.write(inData)
-        p.wait()
-        output = p.stdout.read()
+        (output, err) = p.communicate()
+        if (err != None):
+            return (-1, str(err))
         rc = p.returncode
     except Exception as e:
         print("exec: died with exception ", end="")
--- a/ansible/roles/redis/tasks/deploy.yml
+++ b/ansible/roles/redis/tasks/deploy.yml
@@ -15,7 +15,7 @@
       - "{{ redis.port }}:6379"
 
 - name: wait until redis is up and running
-  action: shell echo PING | nc -w 3 -i 1 {{ inventory_hostname }} {{ redis.port }}
+  action: shell (echo PING; sleep 1) | nc {{ inventory_hostname }} {{ redis.port }}
   register: result
   until: (result.rc == 0) and (result.stdout == '+PONG')
   retries: 12
--- a/tools/health/isAlive
+++ b/tools/health/isAlive
@@ -98,9 +98,9 @@ def isConsulAlive() :
 
 def isZookeeperAlive() :
     host_port = getHostAndPort('zookeeper').split(':')
-    cmd = ['nc', '-w', '3', host_port[0], host_port[1]]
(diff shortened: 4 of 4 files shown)


ansible/roles/kafka/tasks/deploy.yml
```diff
@@ -15,7 +15,7 @@
       - "{{ zookeeper.port }}:2181"
 
 - name: wait until the Zookeeper in this host is up and running
-  action: shell echo ruok | nc -w 3 -i 1 {{ inventory_hostname }} {{ zookeeper.port }}
+  action: shell (echo ruok; sleep 1) | nc {{ inventory_hostname }} {{ zookeeper.port }}
```
domdom82: this seems very sketchy to me. the problem that was fixed by adding `-w 3 -i  1` is that netcat did close the connection too early for zookeeper to respond in some cases. the flags make netcat wait 3 seconds before closing the connection and read only one line per second.
how would a local sleep help here?

style95: On CentOS, if idle timeout happens according to `-i 1` option, it displays the timeout log.
`Ncat: Idle timeout expired (1000 ms).`

I am not sure about other version of CentOS, but it did not work on CentOS 7

domdom82: great idea! should remove the bsd vs gnu issue of nc. not all flags are supported everywhere.

tools/health/monitorUtil.py
```diff
@@ -46,8 +46,9 @@ def run(cmd, inData="") :
         p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stdin=subprocess.PIPE, stderr=subprocess.STDOUT)
         if (inData != ""):
             p.stdin.write(inData)
-        p.wait()
```
style95: I included one more change.
As per the python reference, it is recommended to use `communicate()` instead of `wait()` due to deadlock.
https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait

And the deadlock occurred on Cent OS

markusthoemmes: I think we can just remove this file. A quick search through the repo didn't find any reference to it any more. Can you confirm and remove?

rabbah: not used in open - but used in blue.
we'd have to move either the test from blue here or the scripts there.

style95: This file is being used by `tools/health/isAlive`.
And `tools/health/isAlive` is being used by `services.PingTests`

rabbah:
pg1/1292 🚀

rabbah:
```
Starting test pingZookeeper at 2017-03-22 10:00:17.149
expected exit code = 0
exit code = 254
stdout: isZooKeeperAlive: Unexpected output string to imok query 
isZooKeeperAlive: Unexpected output string to imok query 
isZooKeeperAlive: Unexpected output string to imok query 
isZooKeeperAlive: Unexpected output string to imok query 
isZooKeeperAlive: Unexpected output string to imok query 
isZooKeeperAlive: Unexpected output string to imok query 
isZooKeeperAlive: Unexpected output string to imok query 
isZooKeeperAlive: Unexpected output string to imok query 
isZooKeeperAlive: Unexpected output string to imok query 
isZooKeeperAlive: Unexpected output string to imok query 
isZooKeeperAlive: Unexpected output string to imok query 
zookeeper is not responding
```

ansible/roles/kafka/tasks/deploy.yml
rabbah: I think you need a similar change here https://github.com/openwhisk/openwhisk/blob/master/tools/health/isAlive#L101.

style95: Actually, to use pipe in python.subprocess, we need `shell=True`.
But `shell=True` options is strongly discouraged due to security reason.
https://docs.python.org/2/library/subprocess.html#frequently-used-arguments

Could you share the test environment detail?
How can I reproduce that issue?

rabbah: There's no issue here in terms of user input that needs to be sanitized, is there? Hence the security concerns don't apply; do you disagree?

style95: Fair point, cmd input is fixed, there`s no user input.
Could you share PG environment?
Then it will be useful for double check.

rabbah: it's a ubuntu vm - let me find some details i can share.

style95:
Could you share the PG jenkins details?
I`ve successfully tested this on Ubuntu 16.04

style95:
All Travis checks have passed.

ansible/roles/kafka/tasks/deploy.yml
style95: @rabbah I updated this line: https://github.com/openwhisk/openwhisk/blob/master/tools/health/isAlive#L101 as well

style95:
@rabbah kindly review this.

markusthoemmes:
PG1 1542 🔵

rabbah:
looks good.

markusthoemmes:
Verified locally to work on Mac. Merging. Thanks for the contribution @style95 .

PR #1999, opened 2017-03-10
Adding documentation on gradle usage

markusthoemmes:


rabbah:
@csantanapr good idea - but is that a separate pr? as you point out there are many of these docs are dev only scatttered through out.

csantanapr:
>@csantanapr good idea - but is that a separate pr?

Good point lets keep moving forward 🚆

csantanapr:
I can take care of the index and location in later PR

csantanapr:
merged 601da30ebcd5afc4563275ab5fc9f7fe8df9297b

PR #2000, opened 2017-03-10
Document WSK_CONFIG_FILE setting. Closes #225.

### LINKED ISSUE #225: document the environment variable WSK_CONFIG_FILE for wsk CLI
Document `WSK_CONFIG_FILE` for CLI

if env variable doesn't exist $HOME/.wskprops is used

### LINKED ISSUE #1908: Get https://172.17.0.1/api/v1/namespaces: x509: cannot validate certificate for 172.17.0.1 because it doesn't contain any IP SANs

markusthoemmes:
Has been around for a long time but nobody knows.
Closes #225.
Closes #1908.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -23,6 +23,8 @@ deployment of OpenWhisk. Run the following command from your `openwhisk` directo
 ./bin/wsk property set --apihost <openwhisk_baseurl>
 ```
 
+**Tip:** If you are using a local OpenWhisk deployment with a self-signed SSL certificate, you can use `--insecure` to bypass certificate validation.
+
 If you know your authorization key, you can configure the CLI to use it. Otherwise, you will need to obtain an
 authorization key for most CLI operations. A _guest_ account is available in local installations with an authorization
 key located in [ansible/files/auth.guest](../ansible/files/auth.guest). To configure the CLI to use the guest account,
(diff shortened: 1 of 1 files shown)


docs/cli.md
```diff
@@ -32,6 +32,8 @@ you can run the following command from your `openwhisk` directory:
 ./bin/wsk property set --auth `cat ansible/files/auth.guest`
 ```
 
+**Tip:** The OpenWhisk CLI stores the properties set in `~/.wskprops` by default. The location of this file can be altered by setting the `WSK_CONFIG_FILE` environment variable.
+
```
rabbah: Do you mind adding one more line to close https://github.com/openwhisk/openwhisk/issues/1908 as well?

If you are using an local OpenWhisk deployment with a self-signed SSL certificate, you can use `--insecure` to bypass certificate validation.

csantanapr:
merged af6b00306f5060cf6abfca2a09ab91c482e8cc1a

PR #2002, opened 2017-03-10
Doc Fix: Highlight use of default package identifier

jthomas:
I got caught out trying to work out what the package identifier should be for actions without an explicit package. 

This will be a common usecase for people trying out this feature. 

There's a note about using `default` much later in this file that people might not read at first. I've copied this reference and added it to the initial example to highlight this with users.


csantanapr:
As always thank you @jthomas

Issue #1947, opened 2017-03-05
test content type header for html actions


rabbah:
on [slack today](https://openwhisk-team.slack.com/archives/random/p1488725937000478) there was a question of whether returning an `text/xml` with a web action using an `.html` extension returns a content type of `text/xml` or `text/html`... when in doubt, write a test:

```
        it should s"handle html web action with text/xml response (auth? ${creds.isDefined})" in {
            implicit val tid = transid()

            Seq(s"$systemId/proxy/export_c.html").
                foreach { path =>
                    val html = """<html><body>test</body></html>"""
                    val xml = """<?xml version="1.0" encoding="UTF-8"?><note><from>test</from></note>"""

                    actionResult = Some(JsObject("html" -> xml.toJson))
                    Get(s"$testRoutePath/$path") ~> addHeader("Accept", MediaTypes.`text/xml`.value) ~> sealRoute(routes(creds)) ~> check {
                        status should be(NotAcceptable)
                    }

                    Seq((html, MediaTypes.`text/html`), (xml, MediaTypes.`text/html`)).
                        foreach {
                            case (res, expectedMediaType) =>
                                actionResult = Some(JsObject("html" -> res.toJson))

                                Get(s"$testRoutePath/$path") ~> addHeader("Accept", expectedMediaType.value) ~> sealRoute(routes(creds)) ~> check {
                                    status should be(OK)
                                    responseAs[String] shouldBe res
                                    mediaType shouldBe expectedMediaType
                                }
                        }
                }
        }
```

this should probably we added to the `MetaApiTests` suite.

Issue #1965, opened 2017-03-07
promote web actions to API v2


rabbah:
This is an epic of proposed changes to promote web actions from experimental to a formal API. It will be part of version 2 of the API.

1. [x] In line with normalization of the v2 path in this https://github.com/openwhisk/openwhisk/issues/219#issuecomment-272951007, the resource paths will adopt the fully qualified name format (`/namespace/package/action`, where `"default"` is used for the default package). The API will separate into CRUD operations (thus allowing us to reclaim the verb `POST`), an "activation" path for what is today action invocation (trigger firing) via `POST`, and "web" activations. For example (subject to revision):
    * `/api/v2/actions/namespace/package/name` for CRUD on actions with fully qualified name
    * `/api/v2/web/namespace/package/name` for web invoke
    * `/api/v2/activate/namespace/package/name` for what we currently call POST invoke (could use this for both trigger or action which might be a nice feature although requires type disambiguation in the backend). 
     * For this issue, we are only concerned with a path for web actions e.g., /api/v1/web (or /api/v2/web).

2. [x] make `.http` optional for web invoke and it is the default extension (when none is specified); preserve `.json`, `.html`, and `.text` as syntactic sugar for `.http` and possibly add `.svg`. The caveat when `.http` is omitted is that there is ambiguity when an action name might include a period.

3. [x] add `__ow_body` for a request body that is neither JSON or form data (otherwise there is no such property and the contents are first class as done today).

4. [x] remove `meta` from the `__ow_*` context variables (`__meta_ow_headers` -> `__ow_headers` and `__meta_ow_path` -> `__ow_path`).

5. [x] rename `code` field from HTTP response to `statusCode`.

6. [x] add `X-OW-RAW-HTTP` for web activations which always pass the request body via `__ow_body` instead of promoting the values to first class parameters. The query parameters are passed as part of the path.

7. [x] add a new parameter/directive to expose web action instead of using a `web-export` annotation. `open`, `public`, and `anonymous` were considered, but no winner yet.

8. [ ]  the CLI shall provide the web action URL on demand.

From discussion with @mbehrendt @sjfink @nickm and @csantanapr. CC @dubeejw @mdeuser

rabbah:
A comment on the "api-experimental" and API gateway. The plan is to separate that into an entirely self container package and CLI plugin. The distinction between the API gateway and web actions reduces to:

1. custom/control over mapping of paths (dont have to follow the openwhisk fully qualified naming)
[...]

PR #2010, opened 2017-03-13
Promote Web Actions from Experimental

dubeejw:
- Promote `api/v1/experimental/web` to `api/v1/web`
- Allow CORS header to be set from Web Actions in `v1/web`
- Make extensions optional defaulting to `.http`
- Added .svg extension
- Inline body when it is not JSON
- Includes `raw-http` annotation support allowing Web Actions to receive query parameters and raw body content in `api/v1/web`
- Prevent invocation of an action multiple times from one request
- Display proper error when accept header type does not match actual return type

Closes: https://github.com/openwhisk/openwhisk/issues/1947
Closes: https://github.com/openwhisk/openwhisk/issues/1886
Related: https://github.com/openwhisk/openwhisk/issues/1965


dubeejw:
PG2 1218, PG4 59.

dubeejw:
PG4 50 approved.

dubeejw:
PG1 1262, PG4 60.

rabbah:
@markusthoemmes @cbickel do either of you have bandwidth to review these changes?

starpit:
will /api/v1/experimental routes continue to work? we have apps that depend on this, and will need some grace period to port over to the non-experimental route

dubeejw:
@starpit, the experimental route will remain intact for now.

dubeejw:
PG4 65, approved.

core/controller/src/main/scala/whisk/core/controller/Meta.scala
```diff
@@ -98,14 +190,21 @@ protected[core] object WhiskMetaApi extends Directives {
         extension: String,
         defaultProjection: Option[List[String]],
         projectionAllowed: Boolean,
-        transcoder: (JsValue, TransactionId) => RequestContext => Unit)
+        transcoder: (JsValue, TransactionId, WebApiDirectives) => RequestContext => Unit) {
+        val extensionLength = extension.length
+    }
 
-    private def resultAsHtml(result: JsValue, transid: TransactionId): RequestContext => Unit = result match {
+    private def resultAsHtml(result: JsValue, transid: TransactionId, rp: WebApiDirectives): RequestContext => Unit = result match {
```
cbickel: Why is `rp` needed here?

rabbah: fair question - and i tried to find a convenient workaround. we need to access the status code directives inside the response as http handler to determine the value of the "status code" field. the signature for all of these methods is the same, so for that reason, the directive is provided to all the methods. alternatively, one could create another closure with the status code field captured which seemed more complex and unnecessary.

cbickel: Okay, that's a fair point.

PR #2020, opened 2017-03-13
Minimum change to test Python 3

cclauss:
A minimal changes version of https://github.com/openwhisk/openwhisk/pull/1903

If we can get this one working then we have a chance of getting 1903 working.

See https://github.com/openwhisk/openwhisk/issues/1184#issuecomment-285323326

We have removed all the Python 3 syntax errors from the codebase but the Scala tests are still failing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 # A Travis CI configuration file.
 
-dist: trusty
 sudo: required
 
 language: scala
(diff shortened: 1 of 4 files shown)


cclauss:
#1867 is VERY important for Python 3 compatibility. It allows us to pass 11 test cases that fail when we move from Python 2 to Python 3.  Now we have just two failing tests remaining...

**actionContainers.ActionProxyContainerTests > openwhisk/dockerskeleton should run a python script FAILED**
    org.scalatest.exceptions.TestFailedException: Some({}) was not equal to Some({"string":"hello"})

**actionContainers.ActionProxyContainerTests > openwhisk/dockerskeleton should run a python script and confirm expected environment variables FAILED**
    java.util.NoSuchElementException: key not found: api_host

PR #2022, opened 2017-03-14
drop support for nodejs 0.12.x

jasonpet:
Only get and delete will be allowed for nodejs 0.12.x actions (everything else has been removed).


docs/reference.md
```diff
@@ -258,55 +258,14 @@ The following packages are available to be used in the Node.js 6.9.1 environment
...
-- yauzl v2.3.1
-
 ## Python actions
 
 Python actions are executed by default using Python 2.7.12.
 In addition to the standard Python library, the following packages are also available for use by Python actions.
 
 - attrs v16.1.0
 - beautifulsoup4 v4.5.1
+- beautifulsoup4 v4.5.1
```
csantanapr: what is this duplicate line?

ansible/group_vars/all
```diff
@@ -11,6 +11,7 @@ whisk:
 runtimesManifest:
   nodejs:
   - kind: "nodejs"
+    deprecated: true
```
csantanapr: I ❤️ this manifest

jasonpet:
pg2#1229

jasonpet:
anyone understand why pg and travis results are so different?

rabbah:
Sounds like an issue with pg build - open in blue?

jasonpet:
pg3#405

tests/src/test/scala/system/basic/WskBasicTests.scala
```diff
@@ -192,23 +192,6 @@ class WskBasicTests
...
-            }
-
-            // create action as nodejs (v0.12)
-            wsk.action.get(name).stdout should include regex (""""kind": "nodejs"""")
-
-            // update to nodejs:6
-            wsk.action.create(name, file, kind = Some("nodejs:6"), update = true)
-            wsk.action.get(name).stdout should include regex (""""kind": "nodejs:6"""")
-    }
-
```
markusthoemmes: Update this test rather than removing it?

tests/src/test/scala/whisk/core/controller/test/RulesApiTests.scala
```diff
@@ -898,7 +898,7 @@ class RulesApiTests extends ControllerTestCommon with WhiskRulesApi {
         val aentity = BadEntity(namespace, aname())
         val rule = WhiskRule(namespace, aname(), afullname(namespace, aname().name), afullname(namespace, aname().name))
         val trigger = WhiskTrigger(namespace, rule.trigger.name)
-        val action = WhiskAction(namespace, rule.action.name, js("??"))
+        val action = WhiskAction(namespace, rule.action.name, js6("??"))
```
markusthoemmes: How about defining a `testKind` in `ControllerTestCommon` which is mapped to `js6` for now, so we can update faster in the future?

rabbah: Or use jsdefault() in the exec helper. The kind in most of these tests doesn't matter. ?

tests/src/test/scala/whisk/core/cli/test/WskEntitlementTests.scala
```diff
@@ -186,7 +186,7 @@ class WskEntitlementTests
 
             assetHelper.withCleaner(wsk.action, fullSampleActionName) {
                 val file = Some(TestUtils.getTestActionFilename("empty.js"))
-                (action, _) => action.create(fullSampleActionName, file, kind = Some("nodejs"))(wp)
+                (action, _) => action.create(fullSampleActionName, file, kind = Some("nodejs:6"))(wp)
```
rabbah: Could be nodejs default actual kind not important.

tests/src/test/scala/whisk/core/controller/test/RulesApiTests.scala
rabbah: Since the kind doesn't matter in these tests actually how about a new exec: empty :)

tests/src/test/scala/system/basic/WskBasicTests.scala
jasonpet: The purpose of this test was for updating kind to go from one version of a language to another.  With swift2 and nodejs 0.12.x removed we only have one version of each language now.  Makes sense to add this type of test back once we get new python/nodejs/swift versions.

rabbah: Do we have a cli test that updates a node action with a new kind that is say python?

tests/src/test/scala/whisk/core/controller/test/RulesApiTests.scala
jasonpet: Added a new jsDefault that maps to js6.  Will make it easy to update in the future as Markus suggested.

As for the empty exec that would not work everywhere I had called js6, but only in cases where it passes "" or "??".

jasonpet: Will handle empty exec in a new PR.

PR #2024, opened 2017-03-15
Some scala tweaks 

rabbah:
Bumping active ack to 60 seconds (match blocking invoke timeout); with tighter invoker health check, active ack is short circuited when invoker health changes.


rabbah:
pg2/1235 🔵

PR #2025, opened 2017-03-15
Split actions and activations DB

cbickel:
This PR splits the Database into an activations and an actions (and other entities) DB.


ansible/tasks/db/activationsDb.yml
```diff
@@ -0,0 +1,20 @@
+---
+# Delete and recreate the activationsDb for whisk.
+
+- name: "delete the activationsDb {{ db.whisk.activations }} with {{ db_provider }}"
```
rabbah: any way to share these tasks? I think we now have 3 copies?

ansible/tasks/db/activationsDbViews.yml
```diff
@@ -0,0 +1,35 @@
+---
+# Recreates the views in activationsDb.
+
+- name: get rev of existing whisks-Design document in {{ db.whisk.activations }} with {{ db_provider }}, if it exists
```
rabbah: same comment: updating view, we have 3 copies of this too?

ansible/tasks/db/activationsDb.yml
cbickel: I'll address this.

cbickel:
PG2#1363 is green.

ansible/files/whisks_design_document_for_activations_db.json
```diff
@@ -0,0 +1,13 @@
+{
```
rabbah: note/question to self: how will this work if _split_ is false since both design docs have the same name?

ansible/recreateDesignDocs.yml
```diff
@@ -0,0 +1,7 @@
+---
+# This playbook wipes the database. Pls be sure you know what you are doing now!
```
rabbah: how about a more ominous comment: `WARNING: This playbook wipes the database. This action is not reversible. Be very careful and know what you are doing.`

ansible/recreateDesignDocs.yml
```diff
@@ -0,0 +1,7 @@
...
+
+- hosts: ansible
+  tasks:
+    - include: tasks/recreateViews.yml
```
rabbah: does this actual wipe the db?

ansible/tasks/initdb.yml
```diff
@@ -2,95 +2,32 @@
 # This task will initialize the immortal DBs in the database account.
 # This step is usually done only once per account.
```
rabbah: `  # This step is usually done only once per account. ` per account?  how about per deployment.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -100,7 +100,8 @@ class WhiskConfig(
     val dbPort = this(WhiskConfig.dbPort)
     val dbWhisk = this(WhiskConfig.dbWhisk)
     val dbAuths = this(WhiskConfig.dbAuths)
-    val dbActivations = this(WhiskConfig.dbActivations)
+    lazy val dbSplitActionsAndActivations = this(WhiskConfig.dbSplitActionsAndActivations).toBoolean
```
rabbah: i'm having a hard time seeing how this is used - or can work if the value is false.
convince me.

ansible/files/whisks_design_document_for_activations_db.json
cbickel: When you use wipedb, there is no check if `split` is false or true.
The reason is, that I want to remove this flag in one of my next commits to maintain only the version with the splitted database.

ansible/recreateDesignDocs.yml
cbickel: Copy paster error.
But I used this message for wipeDb.

ansible/recreateDesignDocs.yml
cbickel: Sorry, the comment was wrong.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
cbickel: If `dbSplitActionsAndActivations`  is false, the `dbWhisk` is used for activations.
That's the same behaviour like we have today.

```
lazy val dbSplitActionsAndActivations = this(WhiskConfig.dbSplitActionsAndActivations).toBoolean
    lazy val dbActivations = if (dbSplitActionsAndActivations) this(WhiskConfig.dbActivations) else dbWhisk
```

cbickel:
PG1#1370 is green

jeremiaswerner:
rebased and merged in 7b5a64481aedfd060f3ccaab72a2b6197de47df2

PR #2029, opened 2017-03-15
Only take backups of core databases

cbickel:
It is planned to shrink the size of the snapshots, that are taken.
Currently this is only possible for the core components.

This PR takes out all other databases from the backup process, to avoid any breaking changes of databases, that are not part of the core of OpenWhisk.

### DIFF
--- a/tools/db/replicateDbs.py
+++ b/tools/db/replicateDbs.py
@@ -25,6 +25,11 @@ import time
 import re
 import couchdb.client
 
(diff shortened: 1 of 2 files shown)


tools/db/replicateDbs.py
```diff
@@ -25,6 +25,11 @@
 import re
 import couchdb.client
 
+databasesToBeReplicated = ["namespaces", "subjects", "whisks"]
```
markusthoemmes: Can we just pass this via a parameter vs. hardcoding it?

PR #2030, opened 2017-03-15
Update scanCode.py so that it is configuration file driven.

mrutkows:
This version of scancode is fully configurable (using an external .cfg file).  A default scanCode.cfg is provided with all the defaults that reflect all the settings openwhisk uses today.  In addition, the text for the 2 supported licenses are separated out (and loaded by scanCode) after being listed in the .cfg file.  The config file supports the following sections/configurations:

[Licenses] = list of license files to load/scan for
[Inclusions] = filters (wildcard supported) e.g., *.scal, *.py, etc.) with lists of scans to run for those file filters.
[Exclusions] = list of "paths" (dirs) to exclude from tests.  These are now defaulted to all paths 
[Options] = ony 1 option for now...
    > LICENSE_SLACK_LENGTH=500

Used when code files (e.g., Python) need to have licenses to appear beyond the very first character of the file. This option tells the scan to allow licenses to appear starting
within first 'x' characters of each code file (as provided by this option's value).  This value is large enough (default) to handle all OW files (actually, 270ihs might work) but Python files can allow for longer module descriptions before appearance of the LICENSE text.

I have tested this tool with ALL OW .org repos. and reflect in the cfg file which repos. do not yet pass scancode and I will subsequently work to open issues to have them support Travis/scanCode.  For those repos. that support Travis/scancode already i fixed all their Python files via PRs and all have been merged at this time.  

Please also see this repo. with all my personal tests...
https://github.com/mrutkows/codescan

As a closing thought, perhaps this tool should be in its own repo. so as not to pre-req openwhisk repo. for subprojects that do not require an complete OW platform installed (perhaps already using bmix?).

### DIFF
--- a/tools/build/scanCode.cfg
+++ b/tools/build/scanCode.cfg
@@ -1,0 +1,54 @@
+# scanCode.py configuration file
+
+# List of filenames containing the text of valid license (headers)
+# These files SHOULD be in the same directory path where scanCode.py
+# resides.
+[Licenses]
+ASFLicenseHeader.txt
+ApacheIBMLicenseHeader.txt
+
+# Filters (path/filename) with wildcards and associated scan checks
+# that are to be run against them.  The checks are actual valid
+# function names found in scanCode.py.
+[Includes]
+*=is_not_symlink
+*.scala=has_block_license, no_tabs, no_trailing_spaces, eol_at_eof
+*.py=no_tabs, no_trailing_spaces, eol_at_eof
+*.java=has_block_license, no_tabs, no_trailing_spaces, eol_at_eof
+*.js=no_tabs, no_trailing_spaces, eol_at_eof
+*.gradle=no_tabs, no_trailing_spaces, eol_at_eof
+*.md=no_tabs, eol_at_eof
+*.go=has_block_license, no_tabs, no_trailing_spaces, eol_at_eof
+build.xml=no_tabs, no_trailing_spaces, eol_at_eof
+deploy.xml=no_tabs, no_trailing_spaces, eol_at_eof
+
+# List of paths (inclusive of subdirectories) to exlude from code scanning
+[Excludes]
+# General exclusions
+.tox
+.git
+# OpenWhisk exclusions
+bin
+tests/build/reports
+tests/dat
+# openwhisk-catalog exclusions (Python samples)
+packages/samples/hello
+# The following repos. have so far been identified as having scanning errors
+# and will be excluded until their owners have completed updates to add Travis
+# support, run scancode and become compliant.
+openwhisk-debugger
+openwhisk-client-python
+openwhisk-client-go
+openwhisk-client-js
+openwhisk-vscode
+openwhisk-devtools
+openwhisk-package-jira
+openwhisk-sample-matos
+openwhisk-GitHubSlackBot
+
+[Options]
+# Not all code files allow licenses to appear starting at the first character
+# of the file. This option tells the scan to allow licenses to appear starting
+# within first 'x' characters of each code file (as provided by this option's
+# value).
+LICENSE_SLACK_LENGTH=500
--- a/tools/build/scanCode.py
+++ b/tools/build/scanCode.py
@@ -24,10 +24,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 """
-
+import argparse
 import collections
+import ConfigParser
 import fnmatch
 import itertools
 import os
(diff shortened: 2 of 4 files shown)


mrutkows:
@csantanapr scanCode.py gets:
./tools/travis/build.sh: line 10: tools/build/scanCode.py: Permission denied
Not allowed to read a .cfg file or .txt files in same directory?  How do I debug Travis remotely?

csantanapr:
@mrutkows there is no way to login into Travis

You can modify the tools/travis/build.sh
and put a `ls -l tools/travis` before running the scan script, this will let you know what are the permissions of the scrip of the cfg file you are trying to read.

mrutkows:
Fixed... chmod +x scanCode.py... forgot that i copied the whole file from another directory and it was missing the correct flags to execute (which Mac OSX seems to be forgiving of).

csantanapr:
PG2 1251 🏃

tools/build/scanCode.cfg
```diff
@@ -0,0 +1,54 @@
...
+.tox
+.git
+# OpenWhisk exclusions
+bin
+tests/build/reports
+tests/dat
+# openwhisk-catalog exclusions (Python samples)
+packages/samples/hello
+# The following repos. have so far been identified as having scanning errors
+# and will be excluded until their owners have completed updates to add Travis
```
rabbah: can these exclusions be part of their respective repos?

rabbah:
We should add at lease one positive and one negative test.

tools/build/scanCode.cfg
csantanapr: I like the idea of having exclusion file in another repo  
But exclusion should be for build artifacts and such not main source code. 
The catalog already have Travis and runs scan code 

For this particular case it's only the hello sample let's fix the hello sample should be super easy and remove this exclusion before we merge this. Let's take the time for this take there is no rush 

And again Matt please know that I really really thank you on your contributions on this making OpenWhisk great!!

mrutkows: @rabbah  Yes, but it require that the build scripts create their own .cfg file and pass it in as a parameter e.g., --config openwhisk-catalog.cfg (with its own exclusions).   I thought that once this is in place we would encourage that.  For now I added all possible exclusions.

rabbah: you're excluding an entire repo, right? then why bother run scan code in those repos?

mrutkows: @rabbah as part of my testing this helped me identify which repos are not currently running scans at all and reporting, in some cases, 100s of errors each.  It was my hope that we open issues for each to support scanCode.py in their Travis using the method the 5 repos that now do scan by pulling in openwhisk do (by reference) (note that most do not even have Travis) then open other issues to fix their scanCode errors and then uncomment them in .cfg in openwhisk.  It was a planned progression to get all repos. on board using the same config.  Although each could supply their own config, it would be desirable IMO if all would use a project-wide set of checks (or we might see repo. owners start removing checks.  The initial goal is to assure all code, in all project repos., have proper ASF license headers.  Just my approach.

mrutkows:
@rabbah what approach should be taken for negative tests within the openwhisk repo.?  It speaks to my statement above asking if the tool itself should perhaps be in its own repo.  with its own set of tests much like I have in my private (referenced) repo.  Currently, it seems scanCode (and may of our tools) are assumed to work and not aware any have negative tests.  If there is a path/example to follow I am happy to do so, but my hope is to provide equivalent function, but make it config. file driven to allow flex. and prevent the need for code changes (.py).

rabbah:
we've had broken scan code in the past - where it would not catch the case of a missing header, or trailing white space. So I would add a test that confirms scan code catches trailing spaces and header at the least.

tools/build/scanCode.cfg
rabbah: OK.

rabbah:
>  if the tool itself should perhaps be in its own repo

plausible - is it more generally useful for other (not our) projects?

cclauss:
[scanCode.py#L191](https://github.com/mrutkows/openwhisk/blob/96984a75db1fa3f18fa2a92ab4f0ca5e91506a8f/tools/build/scanCode.py#L191) and [scanCode.py#L238](https://github.com/mrutkows/openwhisk/blob/96984a75db1fa3f18fa2a92ab4f0ca5e91506a8f/tools/build/scanCode.py#L238) are Python 3 syntax errors.  Please consider changing `except Exception, e:` to `except Exception as e:` in these two locations to be compatible with both Python 2 and Python 3.  http://python-future.org/compatible_idioms.html#catching-exceptions  Once #2047 lands, these issues would halt the build on a pull request but before that happens, we need to be on the lookout.

It would also be advisable to avoid bare exceptions (#L149, L151, L191, L238, and L322) where possible to avoid the pitfalls discussed in https://realpython.com/blog/python/the-most-diabolical-python-antipattern/

mrutkows:
Claus I have my Python 3 updates ready to go for a separate PR, as we use 2.7 now for all repo. builds. @cclauss  The existing scanCode.py (historical) is 2.7.

mrutkows:
@cclauss copied my P3 from my personal repo.

mrutkows:
Forgive, but gonna /bump this @rabbah @csantanapr as we have addressed all issues raised, including bringing in P3 updates I was going to bring in on anther PR.

tools/build/scanCode.py
```diff
@@ -162,32 +300,65 @@ def is_not_symlink(path):
...
+        file_check_fxs = []
+        line_check_fxs = []
+        for fname in function_names:
+            try:
+                fx = globals()[fname]
+                if fname in FILE_CHECK_FUNCTIONS:
+                    file_check_fxs.append(fx)
+                elif fname in LINE_CHECK_FUNCTIONS:
+                    line_check_fxs.append(fx)
+            except Exception:
```
cclauss: Would it be possible to be specific about what Exception we are looking for?  Is it a KeyError?

mrutkows: Depending on the data structure I was getting Exceptions here when adding functions when authoring the code.  As for generalized Exception handling,  it should be fine for check for non-granular exceptions types for graceful exit.  The code works, if you wish to fine tune each line, feel free to after getting the actual function in, but this is starting to feel "over the top".

csantanapr:
PG2 1​33​5 is 🏃

csantanapr:
@mrutkows created #2074

PR #2033, opened 2017-03-16
Minimum change to test Python 3

cclauss:


rabbah:
see #2044.

core/actionProxy/actionproxy.py
```diff
@@ -251,6 +250,7 @@ def error():
 
 def complete(response):
     # Add sentinel to stdout/stderr
+    print(response)
```
rabbah: Extra logging may cause tests to fail (which might be asserting empty logs ie no message from proxy appearing in action logs).

cclauss: Got it.  Thanks.

cclauss:
Closing as no longer needed.  #2044 contained the solution.  Results will be bundled into #1903.

PR #2034, opened 2017-03-16
Update apigw docs with web actions comment

csantanapr:


csantanapr:
@rabbah this is ready, quick 👀

PR #2035, opened 2017-03-16
Add controller and invoker0 as default components for syslog get

rabbah:
finding myself having to do get tid twice, once for controller, once for invoker: this makes those default, hence can grab tid logs from both at once.


bjustin-ibm:
PG1 1280 🔵

cclauss:
This file contains Python 3 syntax errors.

PR #2037, opened 2017-03-16
Adding a slack badge

markusthoemmes:
Makes the invite a bit more prominent.


csantanapr:
Just a reminder for my self, we should customize Travis to check if the build for PR AND the diff/patch it's only docs then only do the scan code and not do the whole 40 minute build.

PR #2039, opened 2017-03-17
Refactor InvokerPool for better decoupling

markusthoemmes:
According to akka-standards, synchronous testing is discouraged. Refactored the whole pool to reflect common patterns and make tests much better.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
@@ -39,7 +39,6 @@ import whisk.core.connector.PingMessage
 
 // Received events
 case object GetStatus
-case object StatusChange
 
 // States an Invoker can be in
 sealed trait InvokerState { val asString: String }
@@ -66,10 +65,12 @@ class InvokerPool(
     invokerDownCallback: String => Unit) extends Actor {
 
     implicit val transid = TransactionId.invokerHealth
-    val logging = new AkkaLogging(context.system.log)
     implicit val timeout = Timeout(5.seconds)
     implicit val ec = context.dispatcher
+    val logging = new AkkaLogging(context.system.log)
 
+    // State of the actor. It's important not to close over these
+    // references directly, so they don't escape the Actor.
     val invokers = mutable.HashMap.empty[String, ActorRef]
     val invokerStatus = mutable.HashMap.empty[String, InvokerState]
 
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
@@ -16,24 +16,26 @@
 
 package whisk.core.loadBalancer
 
-import scala.concurrent.duration._
+import scala.collection.mutable
 import scala.concurrent.Future
+import scala.concurrent.duration._
 
 import akka.actor.Actor
+import akka.actor.ActorRef
+import akka.actor.ActorRefFactory
 import akka.actor.FSM
-import akka.actor.Props
-import akka.pattern.ask
-import akka.pattern.pipe
-import akka.util.Timeout
+import akka.actor.FSM.CurrentState
 import akka.actor.FSM.SubscribeTransitionCallBack
 import akka.actor.FSM.Transition
-import whisk.common.AkkaLogging
-import whisk.common.TransactionId
-import whisk.core.connector.PingMessage
-import whisk.common.KeyValueStore
+import akka.actor.Props
+import akka.util.Timeout
 import spray.json._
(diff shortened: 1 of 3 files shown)


markusthoemmes:
PG2 1281 🔵

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -58,57 +60,60 @@ final case class InvokerInfo()
...
+            invokerStatus.update(invoker.path.name, newState)
             newState match {
                 case Offline => Future(invokerDownCallback(invoker.path.name))
                 case _       =>
             }
+            publishStatus()
+    }
+
+    def publishStatus() = {
+        val json = invokerStatus.toMap.mapValues(_.asString).toJson
```
rabbah: `toMap.mapValues` is unnecessary, `.mapValues` skips building an immutable version of the map that is thrown away.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -58,57 +60,60 @@ final case class InvokerInfo()
...
+        kv.put(LoadBalancerKeys.invokerHealth, json.compactPrint)
```
rabbah: is it necessary to build the immutable map (``) when the value is serialized to a string right after (given that this is an actor).

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -58,57 +60,60 @@ final case class InvokerInfo()
...
+        case Transition(invoker, oldState: InvokerState, newState: InvokerState) =>
+            invokerStatus.update(invoker.path.name, newState)
             newState match {
                 case Offline => Future(invokerDownCallback(invoker.path.name))
                 case _       =>
             }
+            publishStatus()
+    }
+
+    def publishStatus() = {
```
rabbah: why not make this method private?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -58,57 +60,60 @@ final case class InvokerInfo()
...
-    /** Collects the status of all invokers */
-    def collectStatus = {
-        val list = context.children.map { invoker =>
-            val name = invoker.path.name
-            invoker.ask(GetStatus).mapTo[InvokerState].map(name -> _)
-        }
-        Future.sequence(list).map(_.toMap)
-    }
+    val invokers = mutable.HashMap.empty[String, ActorRef]
+    val invokerStatus = mutable.HashMap.empty[String, InvokerState]
```
rabbah: rationale for two maps instead of one?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -16,24 +16,26 @@
...
 import spray.json._
 import spray.json.DefaultJsonProtocol._
+import whisk.common.AkkaLogging
 import whisk.common.ConsulKV.LoadBalancerKeys
+import whisk.common.KeyValueStore
+import whisk.common.TransactionId
+import whisk.core.connector.PingMessage
 
 // Received events
 case object GetStatus
```
rabbah: is `GetStatus` used outside of the test class?
`StatusChange` appears unused.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
rabbah: given that these are mutable maps, should they be private?

markusthoemmes: Making fields in Actors private is somewhat redundant as there is no way to get a reference to the object itself. You only talk to the Actor via `ActorRef`

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
markusthoemmes: Yes, it's used by the invoker-choosing algorithm and the `/invokers` route.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
markusthoemmes: `toMap` is needed for the `toJson` call. I could of course generate the `JsObject` manually. Think it's worth it?

jeremiaswerner:
rebased and merged in 4d4e9fdf0e9dd4493fbcb32dd50af2f9e0afb65a

PR #2040, opened 2017-03-17
Add explicit defaults for all exec kinds

bjustin-ibm:
This allows the `:default` designation to work for all kinds, as well as reinforces the idea that all kinds should explicitly state which is the default.

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -16,6 +16,7 @@ runtimesManifest:
     default: true
   python:
   - kind: "python"
+    default: true
   swift:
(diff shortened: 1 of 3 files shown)


rabbah:
thinking out loud - to avoid the out of sync issue that just hit us couple of day ago - should we have one (1) manifest in open (in a files directory) and just ingest it in the ansible files like we do for other things?

the manifest could be a .j2 file even.

thoughts?

@domdom82 ?

bjustin-ibm:
PG3 420

Issue #2038, opened 2017-03-16
document how to decode raw web action body for json


rabbah:
```
> cat t.js 
function main(args) { return {body: new Buffer(args.__ow_body, 'base64').toString('utf-8')} }

> wsk action update raw t.js --web raw
ok: updated action raw

> curl -k https://guest.dockerhost/default/raw  \
  -H "Content-Type: application/json" \
  -X POST -d '{"username":"xyz","password":"xyz"}'
{"username":"xyz","password":"xyz"}

```

This was helpful to at least one user. It should be in the docs.

@dubeejw FYI.

dubeejw:
This needs to be done for other languages as well like Python, Swift, ect...

dubeejw:
@rabbah, this issue can be closed.

PR #2050, opened 2017-03-20
Update Web Action Doc

dubeejw:
- Demonstrate how to decode body in Node, Python, and Swift
- Use CLI to enable/disable web actions and raw HTTP web actions

Closes: https://github.com/openwhisk/openwhisk/issues/2038

### DIFF
--- a/docs/webactions.md
+++ b/docs/webactions.md
@@ -22,7 +22,10 @@ $ wsk package create demo
 $ wsk action create /guest/demo/hello hello.js --web true
 ```
 
-The `--web` flag with a value of `true` or `yes` allows the action to be accessible as a web action via a new REST interface. The URL that is structured as follows: `https://{APIHOST}/api/v1/web/{QUALIFIED ACTION NAME}.{EXT}`. The fully qualified name of an action consists of three parts: the namespace, the package name, and the action name.
+Using the `--web` flag with a value of `true` or `yes` allows an action to be accessible via REST interface without the
+need for credentials. A web action can be invoked using a URL that is structured as follows:
+`https://{APIHOST}/api/v1/web/{QUALIFIED ACTION NAME}.{EXT}`. The fully qualified name of an action consists of three
+parts: the namespace, the package name, and the action name.
 
 *The fully qualified name of the action must include its package name, which is `default` if the action is not in a named package.*
 
@@ -250,7 +253,7 @@ The result of these changes is that the `name` is bound to `Jane` and may not be
 
 ## Disabling web actions
 
-To disable a web action from being invoked via the new API (`https://APIHOST/api/v1/web/`), update the action using the CLI, setting the `--web` flag to a value of `false` or `no`.
+To disable a web action from being invoked via web API (`https://APIHOST/api/v1/web/`), pass a value of `false` or `no` to the `--web` flag while updating an action with the CLI.
 
 ```bash
 $ wsk action update /guest/demo/hello hello.js --web false
@@ -292,7 +295,7 @@ $ wsk action create /guest/demo/hello hello.js --web raw
 
 ### Disabling raw HTTP handling
 
-Disabling raw HTTP can be accomplished by passing a value of `false` or `no` to the `--web`.
+Disabling raw HTTP can be accomplished by passing a value of `false` or `no` to the `--web` flag.
 
 ```bash
 $ wsk update create /guest/demo/hello hello.js --web false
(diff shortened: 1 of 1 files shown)


dubeejw:
@csantanapr, @rabbah, needs review.

docs/webactions.md
```diff
@@ -238,24 +238,22 @@ A content extension is generally required when invoking a web action; the absenc
...
-      --annotation web-export true
+      --web true
 ```
 
 The result of these changes is that the `name` is bound to `Jane` and may not be overridden by query or body parameters because of the final annotation. This secures the action against query or body parameters that try to change this value whether by accident or intentionally. 
 
 ## Disabling web actions
 
-To disable a web action from being invoked via the new API (`https://APIHOST/api/v1/web/`), it’s enough to remove the annotation or set it to `false`.
+To disable a web action from being invoked via the new API (`https://APIHOST/api/v1/web/`), update the action using the CLI, setting the `--web` flag to a value of `false` or `no`.
```
rabbah: new API -> web API (and elsewhere).

docs/webactions.md
```diff
@@ -286,29 +284,99 @@ Notice in this case the JSON content is base64 encoded because it is treated as
...
 
 ```bash
-$ wsk update create /guest/demo/hello hello.js \
-      --annotation web-export true
-      --annotation raw-http false
+$ wsk action create nodeDecode decode.js --web raw
+ok: created action nodeDecode
+$ curl -k -H "content-type: application" -X POST -d "Decoded body" https://${APIHOST}/api/v1/web/guest/default/nodeDecode.json
+{
+  "body": "Decided body"
```
rabbah: typo.

docs/webactions.md
```diff
@@ -286,29 +284,99 @@ Notice in this case the JSON content is base64 encoded because it is treated as
...
-Disabling raw HTTP is accomplished by setting the `raw-http` [annotation](annotations.md) value to `false`.
+### Decoding binary body content from Base64
+
+When using raw HTTP handling, the `__ow_body` content will be encoded in Base64 when the request content-type is binary.
+Below are examples demonstrating how to decode the body content in Node, Python, and Swift.
+
+#### Node
+
+```javascript
+$ cat decode.js
```
rabbah: drop cat ...

rabbah: can add a line below, save the function above to a file called ...

docs/webactions.md
```diff
@@ -286,29 +284,99 @@ Notice in this case the JSON content is base64 encoded because it is treated as
...
+  "body": "Decided body"
+}
 ```
 
-**Note:**  All annotations for a single action must be set at the same time, either when creating or updating the action. This is due to a current limitation on the API and CLI. Failure to do so will result is removal of any previously attached annotations.
 
+#### Python
+
+```python
+$ cat decode.py
```
rabbah: drop cat, same as above.

docs/webactions.md
```diff
@@ -286,29 +284,99 @@ Notice in this case the JSON content is base64 encoded because it is treated as
...
+def main(dict):
```
rabbah: use args instead of dict as parameter name

docs/webactions.md
```diff
@@ -286,29 +284,99 @@ Notice in this case the JSON content is base64 encoded because it is treated as
...
+    try:
+        decoded = dict['__ow_body'].decode('base64').strip()
```
rabbah: check if this code works with python3 (since we're about to enable that as the default).

docs/webactions.md
```diff
@@ -286,29 +284,99 @@ Notice in this case the JSON content is base64 encoded because it is treated as
...
+        return {"body": decoded}
+    except:
+        return {"body": "Could not decode body from Base64."}
+```
+
+```bash
+$ wsk action create pythonDecode decode.py --web raw
+ok: created action pythonDecode
+$ curl -k -H "content-type: application" -X POST -d "Decoded body" https://${APIHOST}/api/v1/web/guest/default/pythonDecode.json
```
rabbah: since the response is going to be same in all cases, and since since the curl command is also the same... consider instead just showing the three functions and then just one example of how to curl it.

docs/webactions.md
```diff
@@ -286,29 +284,99 @@ Notice in this case the JSON content is base64 encoded because it is treated as
...
+{
+  "body": "Decoded body"
+}
+```
+
+#### Swift
+
+```swift
+$ cat decode.swift
```
rabbah: @bjustin-ibm @paulcastro thoughts on below?

paulcastro: Swift looks fine to me,

dubeejw:
Review comments addressed.

csantanapr:
@rr ^^ re-review when you have a chance

docs/webactions.md
```diff
@@ -286,29 +287,79 @@ Notice in this case the JSON content is base64 encoded because it is treated as
...
+        if let decoded = body.base64Decode() {
+            return [ "body" : decoded ]
+        }
+    }
+
+    return ["body": "Could not decode body from Base64."]
+}
+```
 
+As an example, save the NodeJS function as `decode.js` and execute the following commands:
```
rabbah: Here NodeJS and above Node.

csantanapr: Always bet on nodejs 😸

csantanapr: It looks like in their docs they refer to "Node" and other official places as "Node.js" if I have to pick then "Node"

csantanapr:
merged 83dfb9637ce39505c59763afd5ad69211bf98369

Issue #2019, opened 2017-03-13
always print the requested resource that does not exist


sjfink:
currently done correctly for trigger and rule, but not action and package:

`error: Unable to delete action: The requested resource does not exist. (code 2579841)
error: Unable to delete trigger 'T': The requested resource does not exist. (code 2579889)
error: Unable to delete rule 'R': The requested resource does not exist. (code 2579895)
error: Package delete failed: The requested resource does not exist. (code 2579898)
`

PR #2054, opened 2017-03-21
Make Error Messages Consistent in CLI

dubeejw:
- Display entity names for actions and packages when deleting
- Add tests to ensure proper failure when deleting entities that do not exist

Closes: https://github.com/openwhisk/openwhisk/issues/2019

### DIFF
--- a/tests/src/test/scala/common/Wsk.scala
+++ b/tests/src/test/scala/common/Wsk.scala
@@ -426,9 +426,7 @@ class WskRule()
         val params = Seq(noun, if (!update) "create" else "update", "--auth", wp.authKey, fqn(name), (trigger), (action)) ++
             { annotations flatMap { p => Seq("-a", p._1, p._2.compactPrint) } } ++
             { shared map { s => Seq("--shared", if (s) "yes" else "no") } getOrElse Seq() }
-        val result = cli(wp.overrides ++ params, expectedExitCode)
-        if (expectedExitCode == SUCCESS_EXIT) assert(result.stdout.contains("ok:"), result)
-        result
+        cli(wp.overrides ++ params, expectedExitCode)
     }
 
     /**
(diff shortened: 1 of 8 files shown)


dubeejw:
PG3 445 approved. @mdeuser, needs review.

mdeuser:
@dubeejw - changes LGTM.

One comment on a related set of commands that display the nonexistent entity error message incorrectly.  It's the `wsk ENTITY list NAMESPACE` command, and this very well can be a separate issue..  The behavior of the command across the various entities varies.  Most ignore the invalid NAMESPACE and display no error, but the `action list NS` command displays an error with the CLI-configured namespace as the namespace in the error message.

dubeejw:
@mdeuser, the scenario you described seems to run deeper than just changing error messages. I'll open a separate issue to address that defect.

@csantanapr, looks like this is good to merge.

tests/src/test/scala/common/Wsk.scala
```diff
@@ -408,9 +408,7 @@ class WskRule()
         val params = Seq(noun, if (!update) "create" else "update", "--auth", wp.authKey, fqn(name), (trigger), (action)) ++
             { annotations flatMap { p => Seq("-a", p._1, p._2.compactPrint) } } ++
             { shared map { s => Seq("--shared", if (s) "yes" else "no") } getOrElse Seq() }
-        val result = cli(wp.overrides ++ params, expectedExitCode)
-        if (expectedExitCode == SUCCESS_EXIT) assert(result.stdout.contains("ok:"), result)
```
rabbah: general comment about removing these assertions: is it because these messages will be removed in the future?

dubeejw: I just modified the rules class here to make it match the other classes in `Wsk.scala`. Meaning the other classes (action, trigger, etc) do not make assertions. Instead we rely on the actual tests to do so. And yes, the messages may change over time.

tests/src/test/scala/system/basic/WskBasicTests.scala
```diff
@@ -720,4 +853,22 @@ class WskBasicTests
                     wsk.activation.get(activation.activationId, fieldFilter = Some("annotations")).stdout should include(s"""$successMsg annotations\n[]""")
             }
     }
+
+    it should "reject get of activation that does not exist" in {
+        val name = "00000000000000000000000000000000"
```
rabbah: or `"0"*32`

rabbah: you can also do `ActivationId()` it should not match any valid id.

dubeejw:
@rabbah, updated the test.

Issue #1963, opened 2017-03-06
Improve the Python code base to achieve PEP8 compliance


cclauss:
https://github.com/openwhisk/openwhisk/pull/1954 adds [flake8](http://flake8.pycqa.org/en/latest/) testing of all Python files for each pull request to this repo.  We ask Python contributors to review the associated errors and help us to eventually achieve no flake8 errors.  Currently the maximum-line-length is set to 127 (the width of the GitHub editor) but eventually it should be reduced to 79 characters so that our Python code achieves full [PEP8](https://www.python.org/dev/peps/pep-0008/) compliance.

rabbah:
I appreciate that @mrutkows did this and applied pep8 in #1987. But I am not a fan of the 80 character lines - I find it silly (it's 2017, we don't print code to read, we have high res monitors, ...) and think longer lines are actually easier to read in _some_ cases.

This project is not a python project - we have a tiny amount of python code.
Can we at least relax the line limit to 127 characters as in the linter?

I'm not going to make a bigger deal about it; I'll relunctlantly relent and move on regardless.

cclauss:
Agreed on all points...  With the recent migration of code over to [openwhisk-client-python](https://github.com/openwhisk/openwhisk-client-python), Python is now just 1.5% of this repo's code.  For me 127 chars per line is a good limit.  When I write new code, I keep it to 79 characters but lots of the argparse code makes that conversion difficult.

It will be interesting to see how we rationalize https://github.com/openwhisk/openwhisk/pull/1987 and https://github.com/openwhisk/openwhisk/pull/1973 as they touch many of the same files for similar but not identical purposes.  My sense is that once it passes review, you should merge @mrutkows 's 1987 and then I can go back through 1973 manually to see if there is anything left that should still come across.

Please also review https://github.com/openwhisk/openwhisk/pull/1975 which concerns just one file (PythonActionContainerTests.scala) which currently contains Python 2-only code.

Let's get @mrutkows to weigh in before we close this issue.

cclauss:
Where do we want to go with this one?

[...]

PR #2056, opened 2017-03-21
Suppress flake8's nitpicky formatting issues

cclauss:
Issue #1963 says that flake8 is too chatty so we use flake8's `--ignore=E` option to suppress its most nitpicky formatting complaints.  This lets us focus on more serious issues like the issues that it found in the five other commits of this PR.


cclauss:
Ready for review.

PR #2059, opened 2017-03-22
Simplify Print Statements to Make Code More Readable in CLI

dubeejw:
- Extract parameters from externalized strings automatically
- Generate map of parameters/arguments using a print/format method

Current way to print in the CLI:
```
fmt.Fprintf(
  color.Output,
  wski18n.T(
    "{{.ok}} created action {{.name}}\n",
    map[string]interface{}{
      "ok": color.GreenString("ok:"),
      "name": boldString(entityName),
    }))
```

Printing the same message with this update:
```
print(
  "{{.ok}} created action {{.name}}\n",
  color.GreenString("ok:"),
  boldString(action.Name))
```

### DIFF
--- a/tools/cli/go-whisk-cli/commands/action.go
+++ b/tools/cli/go-whisk-cli/commands/action.go
@@ -21,7 +21,6 @@ import (
     "errors"
     "fmt"
     "path/filepath"
-    "io"
     "strings"
 
     "../../go-whisk/whisk"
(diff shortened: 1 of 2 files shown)


dubeejw:
@mdeuser, thoughts on printing this way? 

We can also update WskError to be a linked list of errors so we do not have to manually concatenate error messages together.

mdeuser:
@dubeejw - during the pluggable-cli research, we may find a "best practice" way of printing out messages to the console.

Issue #2026, opened 2017-03-15
Allow concurrent activation execution, in some cases


tysonnorris:
When creating an action, it would be nice to be able to annotate it as -a concurrent <max concurrent  activations> to indicate that concurrent activations are allowed, up to some maximum. This would, for example, allow web actions to be more scalable, instead of limited to the number of concurrently running containers.  This would only be allowed for actions/users that are repeatedly invoked - parameters may be unique per activation, but the action+user would have to be the same for concurrent execution.

This would complicate:
- log collection (log would include interleaved messages from multiple activations, which may be ok)
- pause/unpause after activation completion (don't pause till there are no more activations running)

rabbah:
This would be a radical departure from the current model and comes with some significant baggage. I'll add to the list of complications:

* what does guaranteeing a CPU share mean if multiple instances of an action are running concurrently in the same container?
* similarly how do any of the limits now apply with respect to memory, io, network, diskspace?
* how do you save the user from themselves when their code is racy, or uses the file system incorrectly, what does it mean for debugging, or performance isolation?

If someone were to try - it's certainly doable (the very first whisk prototypes did this) - I can't yet imagine how it would be deployed in a production environment. I can imagine how with some program analysis, it's possible to identify actions for which intra-container concurrency may be plausible. The annotation is the early escape clause to "release" for the platform from undefined, or unintended behavior. 

The serveless promise is that the platform scales to handle any given load. Before going down this road to increase density, might it be more prudent to explore leaner isolation technologies instead?

tysonnorris:
@rabbah I'm not sure this baggage is different than what a developer would deal with if, instead of use a serverless approach, was to package there code into a node container, and run it that way? That is, when you run a container as a traditional web service, you have to gauge cpu/memory requirements in line with traffic and usage characteristics.  Instead of analysis to infer concurrent activation support, I would expect making this more explicit, like based on a different function name (main() vs mainConcurrent()) or annotation or ? would be better, but maybe there are cases where inferred behavior would also be useful. 

I look at it as one additional knob for tuning resource utilization, where you can gain some benefits of serverless (scaling is done, but now based on reaching max allowed concurrency), and you also gain some benefits of increasing resource efficiency (in some cases functions may be just as happy serving 200 (or 2000+) concurrent activations as a single activation).  Tuning cpu shares for  concurrent requests support may be too coarse grained of an adjustment for many situations.

[...]

PR #2021, opened 2017-03-14
Implement a new ContainerPool based on reactive patterns

markusthoemmes:
This implements an entirely new ContainerPool (and eventually Invoker) which is based entirely on an event- and message-driven architecture. It does not use any busy-loops nor manually tweaked Thread.sleeps to distribute its work. As a consequence, this pool does not need locking or synchronized access to any data-structures.

Containers are modeled as a finite-state-machine to reflect the lifecycle of a container in the code itself as close as possible.

Signed-off-by: Sven Lange-Last <slange@de.ibm.com>

## Todos

- [x] Feature toggle (to activate new or old pool)
- [x] Blackbox support (`docker pull`)
- [x] Container eviction strategy (start with LRU first)
- [x] Pre-Warmed containers
- [x] Error handling (Remove containers in certain states)
- [x] Start/Shutdown Cleanup
- [x] Container parameter passing (environment variables)
- [x] Container startup parameters
- [x] Container Naming
- [x] Docker file inspect
- [x] Proper log reading according to limits
- [x] Select containers properly (not based on action but based on action + user)
- [x] Testcoverage

PR #2060, opened 2017-03-22
allow concurrent activation processing

tysonnorris:
allows concurrent processing of activations in a single container IFF 
- action created with "concurrent-max <long>" annotation
- only allow up to specified concurrent actions in-flight per container at any given time
 to address https://github.com/openwhisk/openwhisk/issues/2026

Should be refactored/replaced after #2021 (which will make this logic simpler I think), but providing this PR as basis for conversation in the meantime before that PR is merged.


markusthoemmes:
@tysonnorris the code this is building upon got removed entirely. How do you want to proceed with this PR?

tysonnorris:
closing this one, will open another related soon that is based on SPI for loadbalancer etc

PR #2069, opened 2017-03-25
Enable python and swift unicode tests (and fix regression)

rabbah:


core/actionProxy/actionproxy.py
```diff
@@ -139,9 +142,9 @@ def error(msg):
 
         # stdout/stderr may be either text or bytes, depending on Python
         # version.   In the latter case, decode to text.
-        if isinstance(o, bytes):
+        if not isinstance(o, str):
```
rabbah: @jtyberg the changes made pr# caused a regression in swift actions (we had a test disabled and didn't catch the regression). This pr fixes the regression and enables the test. But I'm not sure that we have this right for the two different python versions we're supporting. Should we explicitly just check the sys.version_info.major and predicate the decoding?

rabbah: @jtyberg the changes made pr #1867 caused a regression in swift actions (we had a test disabled and didn't catch the regression). This pr fixes the regression and enables the test. But I'm not sure that we have this right for the two different python versions we're supporting. Should we explicitly just check the sys.version_info.major and predicate the decoding?

FYI @cclauss

jtyberg: Does this change work in both Python 2 and 3 environments?  (This is really a side question: Does OpenWhisk run the Docker action tests using both Python 2 and Python 3?)

In the end, we don't care about the version of Python (and [should avoid using version detection](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection)); we want `actionProxy.py` to handle the output from actions robustly.  

#1867 was originally opened to handle `Popen` returning text in Python 2 and bytes in Python 3.  I'm assuming the tests you mention are failing because of a unicode error?  In other words, this is not a bytes vs. text issue, but `actionProxy.py` needs to handle output that contains unicode characters?

I keep seeing the same advice when dealing with text in Python 2/3, which is to convert text at the edge. https://pythonhosted.org/kitchen/unicode-frustrations.html#a-few-solutions

> If you get some piece of text from a library, read from a file, etc, turn it into a unicode string immediately. 

This fix probably works in most cases, but there's a lot of implicit logic in it:

```
if not isinstance(o, str): 
    o = o.decode('utf-8')
```

The original `if isinstance(o, bytes):` was really intended to say, "are you bytes from Python 3? if so, do the extra step to convert to unicode str; otherwise, skip this step".

But it seems we need to decode regardless; that is, convert `str` to `unicode` (Python 2), or `bytes` to `str`, which is really unicode (Python 3).

To recap: 

Python 3: as written, subprocess output is always `bytes`, use `decode` to convert to unicode `str`

```
Python 3.5.2
>>> import sys, subprocess
>>> p = subprocess.Popen(['echo', '{"foo":"и"}'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
>>> o, e = p.communicate()
>>> type(o)
<class 'bytes'>
>>> o = o.decode('utf-8')
>>> type(o)
<class 'str'>
>>> process_output_lines = o.strip().split('\n')
>>> sys.stdout.write(o)
{"foo":"и"}
```

Python 2: as written, subprocess output is always `str`, use `decode` to convert to `unicode` (same as `str` in Python 3)

```
Python 2.7.13 
>>> import sys, subprocess
>>> p = subprocess.Popen(['echo', '{"foo":"и"}'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
>>> o, e = p.communicate()
>>> type(o)
<type 'str'>
>>> o = o.decode('utf-8')
>>> type(o)
<type 'unicode'>
>>> process_output_lines = o.strip().split('\n')
>>> sys.stdout.write(o)
{"foo":"и"}
``` 

So maybe the correct way to handle is to `decode` always?  If so, I wonder if we should skip the manual step and instead force `Popen` to give us text always.  The `encoding` parameter was added in Python 3, but it does not exist in Python 2.

https://docs.python.org/3/library/subprocess.html#frequently-used-arguments

> If encoding or errors are specified, or universal_newlines is true, the file objects stdin, stdout and stderr will be opened in text mode using the encoding and errors specified in the call or the defaults for io.TextIOWrapper.

The only way I see to force text output in both Python 2 and 3 is to use `universal_newlines=True`, which seems hacky.

In any case, maybe more comments in the code will help folks understand what's going on.

tests/src/test/scala/actionContainers/PythonActionContainerTests.scala
```diff
@@ -43,13 +43,23 @@ class PythonActionContainerTests extends BasicActionRunnerTests with WskActorSys
...
+         |def main(args):
+         |    print('hello stdout')
+         |    print('hello stderr', file=sys.stderr)
+         |    return args
+         """.stripMargin)
+    })
+
+    testUnicode(Seq {
+        ("python", """
+         |def main(dict):
```
cclauss: This clobbers the builtin dict datatype.  Consider using args instead of dict.

cclauss:
> Does OpenWhisk run the Docker action tests using both Python 2 and Python 3?

The goal of #1903 is that Python 3 would be used for all actions with the exception of Python 2 actions which we would want to deprecate as quickly as possible.  Given the Python 3 has a far more rational/modern approach, we should focus on making other languages like Swift work with Python 3 and stop attempting to create Python 2 solutions to this issue.

cclauss:
I 100% agree with the statement that it is better to use feature detection instead of version detection.

rabbah:
There's no disagreement about getting on to python 3 but the regression is worse. I have a patch which rebase 1903 and this pr and fixes the regression. It's in my branch p3 which is currently running in Jenkins.

To get swift to build with python 3 is surely doable but it currently builds off of trusty which doesn't have the required packages installed and it was more work than I was willing to take on just to fix the regression. I suggest we open a separate issue to update the swift dockerfile.

rabbah:
> Does this change work in both Python 2 and 3 environments? (This is really a side question: Does OpenWhisk run the Docker action tests using both Python 2 and Python 3?)

Yes my change works for both and there are tests now for both versions of python including the Unicode tests. 

In python 2, isinstance(string , binary) returned true. So it's not correct. Always decoding also doesn't work and threw an exception although I did not investigate too deeply.

rabbah:
Picked up in #2070.

jtyberg:
> In python 2, isinstance(string , binary) returned true. 

Yeah, in Python 2 `bytes` is a subset of `str`, so `isinstance(string, bytes)` is `True` in Python 2 and would introduce a `decode()` after #1867.  

So you're fix

```
if not isinstance(o, str):
    o = o.decode('utf-8')
```

would skip decoding in Python 2 altogether 

```
Python 2.7.13
>>> p = subprocess.Popen(['echo', '{"foo":"и"}'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
>>> o, e = p.communicate()
>>> type(o)
<type 'str'>
>>> isinstance(o, bytes)
True
>>> isinstance(o, str)
True
```

and always decode in Python 3

```
Python 3.5.2
>>> p = subprocess.Popen(['echo', '{"foo":"и"}'], stdout=subprocess.PIPE, stderr=subprocess.PIPE,)
>>> o, e = p.communicate()
>>> type(o)
<class 'bytes'>
>>> isinstance(o, bytes)
True
>>> isinstance(o, str)
False
>>> x = o.decode('utf-8')
>>> type(x)
<class 'str'>
```

So the only decoding being done is Python 3 `bytes` to `str` (unicode).  I guess we're "handling" unicode strings by not doing anything to them?  I'd be interested to see what the errors are once the regression is fixed.

rabbah:
> I guess we're "handling" unicode strings by not doing anything to them?

Exactly. The proxy treats them opaquely.

cclauss:
Instead of using `isinstance()`, would it work to ask `hasattr("❄ ☃ ❄", 'decode')`?

jtyberg:
@cclauss Ordinarily, I'd agree with you about using `hasattr`, but I think @rabbah stated the only fix that passed the current tests (and fixed the regression) was if we only decode bytes to unicode in Python 3.  In Python 2, it didn't work to do any decoding, regardless of the type.   

The `if not isinstance(o, str)` check is the logic that produces this result, but it is definitely opaque to anyone reading the code.

It's a mess, but I'd still like to know what the errors were trying to decode a `str` in Python 2, because the "winter" case seems to work fine for me in straight Python:

```
Python 2.7.13
>>> o = "❄ ☃ ❄"
>>> type(o)
<type 'str'>
>>> x = "❄ ☃ ❄".decode('utf-8')
>>> x
u'\u2744 \u2603 \u2744'
>>> type(x)
<type 'unicode'>
>>> print(x)
❄ ☃ ❄
```

Here's a _theory_: the test case(s) are busted?  Perhaps embedding a unicode character in a multi-line text in Scala ([like this](https://github.com/openwhisk/openwhisk/pull/2070/files#diff-3126a23e4a4e7b5798e46a23458e983cR77)) is doing something unexpected? Like using unicode escape characters or something that Python does not handle by default?

rabbah:
I highly doubt this has anything to do with the scala unit tests since the CLI integration tests also fail. If I had to take a guess, it's probably the lack of a codec defined when running using python:2. If I get a chance, I can confirm this fairly quickly although I opted for the lack of decoding when not in binary since there's no need to decode in general. This isn't arbitrary user code: the proxy is a controller environment.

rabbah:
OK here's a quick confirmation, using my [guide on how to run against the container directly](https://medium.com/openwhisk/advanced-debugging-of-openwhisk-actions-518414636932#.qzpzfqhcl) and using the `swift3action` image from my pr with the changes to restore the behavior on master:

```
-        if not isinstance(o, str):
+        if isinstance(o, bytes):
             o = o.decode('utf-8')
-        if not isinstance(e, str):
+        if isinstance(e, bytes):
             e = e.decode('utf-8')
```

```
> ./invoke.py init unicode.swift 
OK

> ./invoke.py run '{"delimiter":"❄"}'
{
  "error": "Internal error. 'ascii' codec can't encode character u'\\u2744' in position 0: ordinal not in range(128)"
}
```

Editing the action proxy to emit the stack trace shows the failure occurs on this line:
```
            sys.stdout.write('%s\n' % line)
```
which for Swift actions will require an encoding to `utf-8` since it's running with python 2.

So why bother decoding, then re-encoding, rather than treat the string opaquely since we know it's already utf-8.

The only scenario to guard against is when the result is bytes not a string as reported initially for python 3 in #1867.

rabbah:
I rewrote the output splitting into logs and a result string in the commit: https://github.com/openwhisk/openwhisk/pull/2070/commits/0e9302fd9e36cfa73b1809d319d46e4a2a0161e4. This avoid emmiting log lines one at time.

jtyberg:
@rabbah Thanks for the additional debugging.  So the Python 2 case is failing when encoding the unicode to ascii when writing to stdout.  That makes sense.

I think skipping the response decoding in Python 2 is fine.  For determining between bytes and text, we could use the `six` library, which keeps actionProxy from having to check the Python versions.

```
import six
if isinstance(obj, six.binary_type): obj.decode('utf-8')
```

https://pythonhosted.org/six/#six.binary_type

rabbah:
Tests passed with commits above.

jtyberg:
`six.string_types` should work to detect whether subprocess response is text or binary:

```
Python 2.7.13
>>> import sys, subprocess, six
>>> p = subprocess.Popen(['echo', '{"foo":"и"}'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
>>> o, e = p.communicate()
>>> type(o)
<type 'str'>
>>> isinstance(o, six.string_types)
True
```

```
Python 3.5.2
>>> import sys, subprocess, six
>>> p = subprocess.Popen(['echo', '{"foo":"и"}'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
>>> o, e = p.communicate()
>>> type(o)
<class 'bytes'>
>>> isinstance(o, six.string_types)
False
```

PR #2070, opened 2017-03-26
fix unicode handling for swift and python 2 actions

### LINKED ISSUE #362: Swift action parsing doesn't tolerate special characters

rabbah:
This replaces #2069 (was rebased on  pr #1903).
Fixes #362.
PG2/1327 🔵

### DIFF
--- a/core/actionProxy/actionproxy.py
+++ b/core/actionProxy/actionproxy.py
@@ -52,6 +52,9 @@ class ActionRunner:
         self.source = source if source else defaultBinary
         self.binary = binary if binary else defaultBinary
 
+    def preinit(self):
+        return
+
     # extracts from the JSON object message a 'code' property and
     # writes it to the <source> path. The source code may have an
     # an optional <epilogue>. The source code is subsequently built
@@ -60,6 +63,7 @@ class ActionRunner:
     # @return True iff binary exists and is executable
     def init(self, message):
         def prep():
+            self.preinit()
             if 'code' in message and message['code'] is not None:
                 binary = message['binary'] if 'binary' in message else False
                 if not binary:
@@ -105,8 +109,7 @@ class ActionRunner:
     def env(self, message):
         # make sure to include all the env vars passed in by the invoker
         env = os.environ
-        for p in ['api_key', 'namespace', 'action_name',
-                  'activation_id', 'deadline']:
+        for p in ['api_key', 'namespace', 'action_name', 'activation_id', 'deadline']:
             if p in message:
                 env['__OW_%s' % p.upper()] = message[p]
         return env
@@ -138,29 +141,35 @@ class ActionRunner:
         (o, e) = p.communicate()
 
         # stdout/stderr may be either text or bytes, depending on Python
-        # version.   In the latter case, decode to text.
-        if isinstance(o, bytes):
+        # version, so if bytes, decode to text. Note that in Python 2
+        # a string will match both types; so also skip decoding in that case
+        if isinstance(o, bytes) and not isinstance(o, str):
             o = o.decode('utf-8')
-        if isinstance(e, bytes):
+        if isinstance(e, bytes) and not isinstance(e, str):
             e = e.decode('utf-8')
 
         # get the last line of stdout, even if empty
-        process_output_lines = o.strip().split('\n')
-        last_line = process_output_lines[-1]
-        for line in process_output_lines[:-1]:
-            sys.stdout.write('%s\n' % line)
+        lastNewLine = o.rfind('\n', 0, len(o)-1)
+        if lastNewLine != -1:
+            # this is the result string to JSON parse
+            lastLine = o[lastNewLine+1:].strip()
+            # emit the rest as logs to stdout (including last new line)
+            sys.stdout.write(o[:lastNewLine+1])
+        else:
+            # either o is empty or it is the result string
+            lastLine = o.strip()
 
         if e:
             sys.stderr.write(e)
 
         try:
-            json_output = json.loads(last_line)
+            json_output = json.loads(lastLine)
             if isinstance(json_output, dict):
                 return (200, json_output)
             else:
-                return error(last_line)
+                return error(lastLine)
         except Exception:
-            return error(last_line)
+            return error(lastLine)
 
     # initialize code from inlined string
     def initCodeFromString(self, message):
@@ -210,9 +219,7 @@ def init():
     if status is True:
         return ('OK', 200)
     else:
-        response = flask.jsonify({'error':
-                                  'The action failed to generate or '
-                                  'locate a binary. See logs for details.'})
+        response = flask.jsonify({'error': 'The action failed to generate or locate a binary. See logs for details.'})
         response.status_code = 502
         return complete(response)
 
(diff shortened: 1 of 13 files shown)


tools/cli/go-whisk-cli/commands/action.go
```diff
@@ -412,6 +412,11 @@ func getExec(artifact string, kind string, isDocker bool, mainEntry string) (*wh
         code, err = readFile(artifact)
         exec.Code = &code
 
+        if (ext == ".jar") {
```
rabbah: @dubeejw this fixes a bug in the CLI where --kind java doesn't encode the jar file.

core/actionProxy/actionproxy.py
```diff
@@ -138,29 +141,35 @@ def error(msg):
...
-        if isinstance(o, bytes):
+        # version, so if bytes, decode to text. Note that in Python 2
+        # a string will match both types; so also skip decoding in that case
+        if isinstance(o, bytes) and not isinstance(o, str):
             o = o.decode('utf-8')
-        if isinstance(e, bytes):
+        if isinstance(e, bytes) and not isinstance(e, str):
             e = e.decode('utf-8')
 
         # get the last line of stdout, even if empty
```
rabbah: the change below is worth extra scrutiny.

csantanapr:
@markusthoemmes could you help review this PR?

rabbah:
@cbickel also a good reviewer for this if he has bandwidth.

rabbah:
I suggest reviewing with https://github.com/openwhisk/openwhisk/pull/2070/files?w=1 to focus on actual changes which are small.

tools/travis/build.sh
```diff
@@ -21,7 +21,7 @@ $ANSIBLE_CMD apigateway.yml
 
 cd $ROOTDIR
 
-./gradlew distDocker -x :core:swift3Action:distDocker -x :core:pythonAction:distDocker -PdockerImagePrefix=testing
```
csantanapr: Why you remove this?
This excludes is to make the Travis build end below the 50 minutes max.
This related to the gradle task testLean, where swift3 and python are excluded 
https://github.com/openwhisk/openwhisk/blob/master/tests/build.gradle#L18

csantanapr:
@rabbah 
>@cbickel also a good reviewer for this if he has bandwidth.

I added @cbickel to reviewers list

tools/travis/build.sh
rabbah: I looked at the build times and building the containers did not add enough time to get close to the limit and it was in the noise relative to other builds. We will now run sniff tests against all the runtimes in Travis which is better than contributors not knowing they've broken something systemic for the previously ignored runtimes.

We can further cut the swift build time by putting up a base image in dockerhub. We should we do as part of your work on all images published to docker hub.

csantanapr: @rabbah if you are going to build the images, then it make sense to enable the tests in the build.gradle task testLean used in travis

The current tasks exclude swift and python tests.
```
task testLean(type: Test) {
    exclude '**/*Swift*'
    exclude '**/*Python*'
    exclude '**/*ThrottleTests*'
    exclude '**/MaxActionDurationTests*'
}
```

rabbah: No change is needed here. The tests I enabled run as part of lean now.

csantanapr: I don't understand what you mean, but OK
 you didn't modified gradle.build and travis still calls `./gradlew :tests:testLean`

rabbah: Inspect this line: https://github.com/openwhisk/openwhisk/pull/2070/files#diff-31a3b6664356184ce7bd3d6d5d6ef4d0R39

This test suite is already running (it's included in `testLean` already).

tests/src/test/scala/system/basic/WskUnicodeTests.scala
```diff
@@ -36,16 +36,19 @@ class WskUnicodeTests
...
+                    val name = s"unicodeGalore.${k.replace(":", "")}"
 
                     assetHelper.withCleaner(wsk.action, name) {
                         (action, _) =>
-                            action.create(name, Some(TestUtils.getTestActionFilename(file)), main = if (k == "java") Some("Unicode") else None)
+                            action.create(
+                                name,
+                                Some(TestUtils.getTestActionFilename(file)),
+                                main = if (k == "java") Some("Unicode") else None,
+                                kind = Some(k))
```
rabbah: @dubeejw here's the test that surfaced the bug mentioned below.

rabbah:
thanks @cbickel for having a look.

tools/cli/go-whisk-cli/commands/action.go
dubeejw: The CLI change LGTM.

dubeejw: I think we might be able to Base64 encode here for .jar and .zip files. And remove the encoding for .zip files that is at the end of the function.

```
    if ext == “.jar” || ext == “.zip”  {
        code = base64.StdEncoding.EncodeToString([]byte(code))
        exec.Code = &code
    }

```

rabbah: good call. https://github.com/openwhisk/openwhisk/pull/2070/commits/0e0cc206a499719794d7f0e6024a1b19a6bdf4c9

tools/cli/go-whisk-cli/commands/action.go
```diff
@@ -410,11 +410,18 @@ func getExec(artifact string, kind string, isDocker bool, mainEntry string) (*wh
...
-        if (ext == ".jar") {
-            exec.Jar = base64.StdEncoding.EncodeToString([]byte(code))
-            exec.Code = nil
+        if ext == ".zip" || ext == ".jar" {
+            // Base64 encode the file
+            code = base64.StdEncoding.EncodeToString([]byte(code))
+            if (ext == ".zip") {
+                exec.Code = &code
+            } else {
+                exec.Code = nil
```
dubeejw: Is the assignment of nil necessary here?

rabbah: https://github.com/openwhisk/openwhisk/pull/2070/files#diff-41c5935cb68718d73fa6ef611c927cc4L439

rabbah: It seems now that since exec.Code is not already assigned, we can remove this.

dubeejw: The way to tell would be to create a action with a .jar file and make sure the `code` field is not present in the package sent from the CLI.

rabbah: I removed `Jar` entirely.

rabbah:
@bjustin-ibm this is now rebased to master and fully squashed.

bjustin-ibm:
PG2 1376 🔵 (for good measure)

PR #2072, opened 2017-03-27
CacheConcurrencyTests: tie number of threads to number of available CPUs instead of hard-coded value.

domdom82:
this should reduce the number of intermittent failures of this test on smaller systems.


domdom82:
PG4 104

rabbah:
@markusthoemmes what shall we do about this pr?

markusthoemmes:
@rabbah same question to you 😁 . Haven't seen that test fail in ages, seems like a bad idea to risk make it less stable?

rabbah:
leave well enough alone.

Issue #2073, opened 2017-03-27
Entity List Commands Should Error if an Entity Name is Given


dubeejw:
Supplying an entity name for a list command is an invalid operation even when a namespace is provided. When such a scenario occurs, an error should be generated. Shown below is the existing behavior of each list command given an entity name instead of namespace. 

```
$ wsk action list entityName
error: Unable to obtain the list of actions for namespace 'default': The requested resource does not exist. (code 2827)
```

```
$ wsk package list entityName
packages...
```

```
$ wsk trigger list entityName
triggers...
```

```
 wsk rule list entityName
rules...
```

```
$ wsk activation list entityName
activations...
```

```
$ wsk list entityName
Entities in namespace: default
packages...
actions...
triggers...
rules...
```

PR #2075, opened 2017-03-27
Error when Entity Names are Provided to List Commands

dubeejw:
- When entity names are passed to list commands display an appropriate error
- Share qualified name parsing error messages between commands
- Move `wsk list` command to `wsk.go`

Closes: https://github.com/openwhisk/openwhisk/issues/2073


dubeejw:
PG1 1673, approved.

Issue #2079, opened 2017-03-28
Upgrade Swift runtime to Swift 3.1


paulcastro:
Upgrade to Swift 3.1

paulcastro:
Pulling common files from swift3Action now.

PR #2085, opened 2017-03-30
swift3 docker build speed boost base on image

csantanapr:
PG1 1347 🏃

### DIFF
--- a/core/swift3Action/Dockerfile
+++ b/core/swift3Action/Dockerfile
@@ -9,7 +9,7 @@ WORKDIR /
 
 # Upgrade and install basic Python dependencies
 RUN apt-get -y update \
- && apt-get -y install --fix-missing python-gevent python-flask
+ && apt-get -y install --fix-missing python-gevent python-flask \
 \
 # Install zip and curl for compiling Swift actions
  && apt-get -y install curl zip \
--- a/core/swift3Action/Dockerfile
+++ b/core/swift3Action/Dockerfile
@@ -1,51 +1,15 @@
 # Dockerfile for swift actions, overrides and extends ActionRunner from actionProxy
-# This Dockerfile is partially based on: https://github.com/swiftdocker/docker-swift/
-FROM buildpack-deps:trusty
+# This Dockerfile is partially based on: https://github.com/IBM-Swift/swift-ubuntu-docker/blob/master/swift-development/Dockerfile
+FROM ibmcom/swift-ubuntu:3.0.2
 
-ENV DEBIAN_FRONTEND noninteractive
+ENV SWIFT_LOCATION /root/swift-3.0.2-RELEASE-ubuntu14.04
+
+# Set WORKDIR
+WORKDIR /
 
 # Upgrade and install basic Python dependencies
-RUN apt-get -y purge \
- && apt-get -y update \
- && apt-get -y install --fix-missing python2.7 python-gevent python-flask \
-\
-# Upgrade and install Swift dependencies
- && apt-get -y install --fix-missing build-essential curl wget libicu-dev \
-\
-# Install zip for compiling Swift actions
- && apt-get -y install zip \
-\
-# Clean up
- && apt-get clean
-
-# Install clang manually, since SPM wants at least Clang 3-6
-RUN cd / &&\
-(curl -L -k http://llvm.org/releases/3.6.2/clang+llvm-3.6.2-x86_64-linux-gnu-ubuntu-14.04.tar.xz | tar xJ) &&\
-cp -r /clang+llvm-3.6.2-x86_64-linux-gnu-ubuntu-14.04/* /usr/ &&\
-rm -rf /clang+llvm-3.6.2-x86_64-linux-gnu-ubuntu-14.04
-
-RUN update-alternatives --install /usr/bin/g++ g++ /usr/bin/clang++ 20 
-RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/clang 20 
-
-# Install Swift keys
-RUN wget --no-verbose -O - https://swift.org/keys/all-keys.asc | gpg --import - && \
-    gpg --keyserver hkp://pool.sks-keyservers.net --refresh-keys Swift
-
-# Install Swift Ubuntu 14.04 Snapshot
-#https://swift.org/builds/swift-3.0.1-release/ubuntu1404/swift-3.0.1-RELEASE/swift-3.0.1-RELEASE-ubuntu14.04.tar.gz
-
-ENV SWIFT_VERSION 3.0.2
-ENV SWIFT_RELEASE_TYPE RELEASE
-ENV SWIFT_PLATFORM ubuntu14.04
-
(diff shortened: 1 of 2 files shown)


rabbah:
@paulcastro using the ibm image, reliable?

rabbah:
Carlos, one bit to keep in mind is we want to run the proxy with python 3 soon so we might need to independently build and publish our own swift image. In any case having the action here built from a different base is the right idea in terms of getting the separation we want. 👍

paulcastro:
Look good to me - 1404, etc matches us.  We had an issue with an older clang in the base image.  Upping the version was part of our slow-down.

paulcastro:
One other comment, this container is Swift 3.0.1. Per issue #2079 I'd like to upgrade to Swift 3.1 and also move to the new implementation of the  associated Watson SDKs.  A PG build indicates this may exacerbate the existing startup cost for the first swift3action container tests.

csantanapr:
@paulcastro I want to do the change gradually, I purposely didn't update to swift 3.1 and updated the WatsonSDK and JSON lib to latest. 
I think that should be done in a separate PR

Also @rabbah I think we need to deprecate the use of a single swift file for swift actions.
I think we should restrict to only accepting swift zip precompiled actions. 
This will avoid users to hit a very large cold start by mistake, and also reduce the docker image size were we can use `ibmcon/swift-ubuntu-runtime` and cut the size of the image.

rabbah:
we can debate that separately: if we add the _init_ actions, then users can get that for free.

csantanapr:
That was my thinking also, but from the perspective of the runtime it only takes zip
Even it means the unit actions are helpers to get the zip created/refreshed

csantanapr:
@paulcastro 
When you have a chance can you try this PR.
The image is not being built correctly
This step:
```
 python /swift3Action/buildandrecord.py
Building action
action built. Decoding compile and link commands
Cannot generate build script: compile or link command not found
```

paulcastro:
I'm testing it now - good catch.  I noticed that the home directory gets set to the swift runtime instead of "/"

csantanapr:
fetch again I made an update to make it `/`

the ibmcon/swift-ubuntu image sets the workdir to /root

csantanapr:
@paulcastro https://github.com/openwhisk/openwhisk/pull/2085/commits/bb9d21037e7c7d21f67f83ff3c5b0e265b3f9a10

paulcastro:
`swiftc` is in a different location so lines 26 and 27 of buildandrecord.py should be:

```
COMPILE_PREFIX = "/root/swift-3.0.2-RELEASE-ubuntu14.04/usr/bin/swiftc -module-name Action "
LINKER_PREFIX = "'/root/swift-3.0.2-RELEASE-ubuntu14.04/usr/bin/swiftc' -Xlinker '-rpath=$ORIGIN' '-L/swift3Action/spm-build/.build/release' -o '/swift3Action/spm-build/.build/release/Action'"
```

I have an example in my 3.1 [branch](https://github.com/paulcastro/openwhisk/blob/issue_2079b-x/core/swift3Action/buildandrecord.py)

csantanapr:
Thanks @paulcastro that did the trick https://github.com/openwhisk/openwhisk/pull/2085/commits/33ef2959b69bea6d8873924eb99b89578b38d5a5

rabbah:
I will review - but going to declare this is depend on 2070 to make sure we're not again regressing tests I just fixed.

core/swift3Action/Dockerfile
```diff
@@ -1,51 +1,15 @@
...
-    SWIFT_URL=https://swift.org/builds/swift-$SWIFT_VERSION-$(echo "$SWIFT_RELEASE_TYPE" | tr '[:upper:]' '[:lower:]')/$(echo "$SWIFT_PLATFORM" | tr -d .)/swift-$SWIFT_VERSION-$SWIFT_RELEASE_TYPE/$SWIFT_ARCHIVE_NAME.tar.gz && \
-    echo $SWIFT_URL && \
-    wget --no-verbose $SWIFT_URL && \
-    wget --no-verbose $SWIFT_URL.sig && \
-    gpg --verify $SWIFT_ARCHIVE_NAME.tar.gz.sig && \
-    tar -xzf $SWIFT_ARCHIVE_NAME.tar.gz --directory / --strip-components=1 && \
-    rm -rf $SWIFT_ARCHIVE_NAME* /tmp/* /var/tmp/*
+# Upgrade and install basic Python dependencies
+RUN apt-get -y update \
+ && apt-get -y install --fix-missing python2.7 python-gevent python-flask zip
```
rabbah: > python2.7 is already the newest version.
> python2.7 set to manually installed.

can skip python2.7 it's already in the image.

core/swift3Action/buildandrecord.py
```diff
@@ -23,10 +23,8 @@
 from subprocess import check_output
 
 # Settings
-COMPILE_PREFIX = "/usr/bin/swiftc -module-name Action "
-LINKER_PREFIX = "/usr/bin/swiftc -Xlinker '-rpath=$ORIGIN' " \
-                "'-L/swift3Action/spm-build/.build/release' " \
-                "-o '/swift3Action/spm-build/.build/release/Action'"
+COMPILE_PREFIX = os.environ["SWIFT_LOCATION"]+"/usr/bin/swiftc -module-name Action "
+LINKER_PREFIX = "'"+os.environ["SWIFT_LOCATION"]+"/usr/bin/swiftc' -Xlinker '-rpath=$ORIGIN' '-L/swift3Action/spm-build/.build/release' -o '/swift3Action/spm-build/.build/release/Action'"
```
markusthoemmes: Want to use string interpolation to make this easier to grasp?

```
LINKER_PREFIX = "'%s/usr/bin/swiftc' -Xlinker '-rpath=$ORIGIN' '-L/swift3Action/spm-build/.build/release' -o '/swift3Action/spm-build/.build/release/Action'" % os.environ["SWIFT_LOCATION"]

**Comment by csantanapr** [comment]
@paulcastro going to take care of this when he updates to swift 3.1

**Comment by rabbah** [comment]
FYI I rebased this and fixed a missing slash:

```
- && apt-get -y install --fix-missing python-gevent python-flask
+ && apt-get -y install --fix-missing python-gevent python-flask \
```

one test fails
screen shot 2017-04-03 at 2 58 30 pm

it's because of this error:
```
"/swift3Action/spm-build/.build/release/Action: error while loading shared libraries: libswiftCore.so: cannot open shared object file: No such file or directory" was not empty (SwiftActionContainerTests.scala:235)
```

@paulcastro FYI if it's helpful.

csantanapr:
Yeah I rebase and added a slash on branch @paulcastro https://github.com/csantanapr/openwhisk/commit/91555aaaa3c3b416af56c2415d9fc984c2890ebb

paulcastro:
Updated helloSwift.zip file in `tests/dat/actions` should fix the test.

markusthoemmes:
@csantanapr can we revive this seeing the latest build breakages due to swift downloads directly from swift.org.

csantanapr:
@markusthoemmes this PR is not good as I'm afraid it will break pre-compiled zip actions. 
I think you already took care by pinning the publish image

PR #2092, opened 2017-03-31
Provide reactive Docker and Runc client

### LINKED ISSUE #2087: Implement a reactive Docker client
For the reactive ContainerPool we need a reactive DockerClient. It needs to support all docker operations we need.

- run
- pause
- unpause
- rm
- inspect

### LINKED ISSUE #2088: Implement a reactive Runc client
We need a reactive RunC client, that supports all the operations we need:

- pause
- resume

sven-lange-last:
Provide reactive Docker and Runc client in preparation for reactive container pool implementation.

Closes #2087 
Closes #2088

Signed-off-by: Markus Thoemmes <markus.thoemmes@de.ibm.com>

### DIFF
--- a/common/scala/src/main/scala/whisk/common/Logging.scala
+++ b/common/scala/src/main/scala/whisk/common/Logging.scala
@@ -142,13 +142,10 @@ object LogMarker {
     val keyword = "marker"
 
     /** Convenience method for parsing log markers in unit tests. */
-    def unapply(s: String) = {
+    def parse(s: String) = {
         val logmarker = raw"\[${keyword}:([^\s:]+):(\d+)(?::(\d+))?\]".r.unanchored
-        s match {
-            case logmarker(token, deltaToTransactionStart, deltaToMarkerStart) =>
-                Some(LogMarkerToken.parse(token), deltaToTransactionStart.toLong, Option(deltaToMarkerStart).map(_.toLong))
-            case _ => None
-        }
+        val logmarker(token, deltaToTransactionStart, deltaToMarkerStart) = s
+        LogMarker(LogMarkerToken.parse(token), deltaToTransactionStart.toLong, Option(deltaToMarkerStart).map(_.toLong))
     }
 }
 
--- a/common/scala/src/main/scala/whisk/common/Logging.scala
+++ b/common/scala/src/main/scala/whisk/common/Logging.scala
@@ -178,6 +178,8 @@ case class LogMarkerToken(component: String, action: String, state: String) {
 
 object LogMarkerToken {
     def parse(s: String) = {
+        // Per convention the components are guaranteed to not contain '_'
+        // thus it's safe to split at '_' to get the components
         val Array(component, action, state) = s.split("_")
         LogMarkerToken(component, action, state)
     }
--- a/common/scala/src/main/scala/whisk/common/Logging.scala
+++ b/common/scala/src/main/scala/whisk/common/Logging.scala
@@ -132,8 +132,23 @@ class PrintStreamLogging(outputStream: PrintStream = Console.out) extends Loggin
  */
 case class LogMarker(token: LogMarkerToken, deltaToTransactionStart: Long, deltaToMarkerStart: Option[Long] = None) {
     override def toString() = {
-        val parts = Seq("marker", token.toString, deltaToTransactionStart) ++ deltaToMarkerStart
+        val parts = Seq(LogMarker.keyword, token.toString, deltaToTransactionStart) ++ deltaToMarkerStart
         "[" + parts.mkString(":") + "]"
+    }
+}
+
+object LogMarker {
+
+    val keyword = "marker"
+
+    /** Convenience method for parsing log markers in unit tests. */
+    def unapply(s: String) = {
+        val logmarker = raw"\[${keyword}:([^\s:]+):(\d+)(?::(\d+))?\]".r.unanchored
+        s match {
+            case logmarker(token, deltaToTransactionStart, deltaToMarkerStart) =>
+                Some(LogMarkerToken.parse(token), deltaToTransactionStart.toLong, Option(deltaToMarkerStart).map(_.toLong))
+            case _ => None
+        }
     }
 }
 
@@ -156,6 +171,16 @@ private object Emitter {
 
 case class LogMarkerToken(component: String, action: String, state: String) {
     override def toString() = component + "_" + action + "_" + state
+
+    def asFinish = copy(state = LoggingMarkers.finish)
+    def asError = copy(state = LoggingMarkers.error)
+}
+
+object LogMarkerToken {
+    def parse(s: String) = {
+        val Array(component, action, state) = s.split("_")
+        LogMarkerToken(component, action, state)
+    }
 }
 
 object LoggingMarkers {
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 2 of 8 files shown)


core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
```diff
@@ -0,0 +1,160 @@
...
+case class ContainerIp(val asString: String) {
+    require(asString.nonEmpty, "ContainerIp must not be empty")
+}
+
+trait DockerApi {
+    /**
+     * Spawns a container in detached mode.
+     *
+     * @param image the image to start the container with
+     * @param args arguments for the docker run command
```
rabbah: Docs out of sync.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
```diff
@@ -0,0 +1,160 @@
...
+     *
+     * @param id the id of the container to remove
+     * @return a Future completing according to the command's exit-code
+     */
+    def rm(id: ContainerId)(implicit transid: TransactionId): Future[Unit]
+
+    /**
+     * Returns a list of ContainerIds in the system.
+     *
+     * @param all Whether or not to return stopped containers as well
```
rabbah: Here also.

core/invoker/src/main/scala/whisk/core/containerpool/docker/ProcessRunner.scala
```diff
@@ -0,0 +1,52 @@
...
+     *
+     * @param args command to be run including arguments
+     * @return a future completing according to the command's exit code
+     */
+    protected def executeProcess(args: String*)(implicit ec: ExecutionContext) =
+        Future(blocking {
+            val out = new mutable.ListBuffer[String]
+            val err = new mutable.ListBuffer[String]
+            val exitCode = args ! ProcessLogger(o => out += o, e => err += e)
+
```
rabbah: Are we losing automatic logging of cmds on start of process and marker at end?

core/invoker/src/main/scala/whisk/core/containerpool/docker/ProcessRunner.scala
```diff
@@ -0,0 +1,52 @@
...
+import scala.collection.mutable
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.blocking
+import scala.sys.process._
+
+trait ProcessRunner {
+
+    /**
```
rabbah: Run to Runs

core/invoker/src/main/scala/whisk/core/containerpool/docker/RuncClient.scala
```diff
@@ -0,0 +1,73 @@
...
+ * Be cautious with the ExecutionContext passed to this, as the
+ * calls to the CLI are blocking.
+ *
+ * You only need one instance (and you shouldn't get more).
+ */
+class RuncClient(executionContext: ExecutionContext)(implicit log: Logging) extends RuncApi with ProcessRunner {
+    implicit private val ec = executionContext
+
+    // Determines how to run docker. Failure to find a Docker binary implies
+    // a failure to initialize this instance of DockerClient.
```
rabbah: Are there alternative paths like with docker?

core/invoker/src/main/scala/whisk/core/containerpool/docker/RuncClient.scala
```diff
@@ -0,0 +1,73 @@
...
+    protected val runcCmd: Seq[String] = Seq("/usr/bin/docker-runc")
+
+    def pause(id: ContainerId)(implicit transid: TransactionId): Future[Unit] = runCmd("pause", id.asString).map(_ => ())
+
+    def resume(id: ContainerId)(implicit transid: TransactionId): Future[Unit] = runCmd("resume", id.asString).map(_ => ())
+
+    private def runCmd(args: String*)(implicit transid: TransactionId): Future[String] = {
+        val cmd = runcCmd ++ args
```
rabbah: Ok so logging is now explicit on process calls for docker. (Checking for myself if this is the way it already is)

markusthoemmes: It's not. But I think it's more appropriate and less voluminous the way it is now vs. logging a layer further down.

tests/src/test/scala/common/StreamLogging.scala
```diff
@@ -32,4 +33,6 @@ trait StreamLogging {
     val stream = new ByteArrayOutputStream
     val printstream = new PrintStream(stream)
```
rabbah: Can we make these lazy for more convenient override? I've been bit by this.

core/invoker/src/main/scala/whisk/core/containerpool/docker/RuncClient.scala
markusthoemmes: AFAIK no. In fact the `runc` binary is only available in conjunction with the docker daemon. On mac for instance, you don't even have that binary at hand.

tests/src/test/scala/common/StreamLogging.scala
markusthoemmes: Why would you need to override those? They are test convenience already? If you need to override them, you might as well write your own helper?

rabbah:
Nice first scala pr @sven-lange-last!

core/invoker/src/main/scala/whisk/core/containerpool/docker/ProcessRunner.scala
markusthoemmes: Yes we do. See below for explanation.

tests/src/test/scala/common/StreamLogging.scala
rabbah: Because it would be useful sometimes to log to console readily.

sven-lange-last:
Thanks @rabbah for your feedback and review.

rabbah:
pg?

markusthoemmes:
PG1 1348 🔵

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -156,6 +168,18 @@ private object Emitter {
...
 case class LogMarkerToken(component: String, action: String, state: String) {
     override def toString() = component + "_" + action + "_" + state
+
+    def asFinish = copy(state = LoggingMarkers.finish)
+    def asError = copy(state = LoggingMarkers.error)
+}
+
+object LogMarkerToken {
+    def parse(s: String) = {
+        // Per convention the components are guaranteed to not contain '_'
```
cbickel: Is there also a check in the code to avoid using `_` in one of the three parts?

markusthoemmes: There isn't one deliberately. A runtime check will likely introduce quite some overhead for the markers we generate on the fly.

cbickel: Makes sense to me.

rabbah: for posterity; one option discussed on slack with @markusthoemmes is to lift two of the necessary `require` statements out to constructors for `component` and `state` values (by making those types other than String). Since these two are at least a closed set, one can enforce the invariant at construction time when the invoker is started.

core/invoker/src/main/scala/whisk/core/containerpool/docker/ProcessRunner.scala
rabbah: Any other places `executeProcess` was used for which the logging needs to be restored at the call site?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        val host = dockerHost.map(host => Seq("--host", s"tcp://$host")).getOrElse(Seq.empty[String])
+        Seq(dockerBin) ++ host
+    }
+
+    def run(image: String, args: Seq[String] = Seq.empty[String])(implicit transid: TransactionId): Future[ContainerId] =
+        runCmd((Seq("run", "-d") ++ args ++ Seq(image)): _*).map(ContainerId.apply)
+
+    def inspectIPAddress(id: ContainerId)(implicit transid: TransactionId): Future[ContainerIp] =
```
rabbah: isn't it the case that `docker inspect` is not used in the invoker anymore?
so while this is provided for completeness, does the container pool actually use it?

markusthoemmes: Correct, it's only here for completeness.

sven-lange-last: The upcoming DockerContainer code will use `docker inspect` command as fallback if it fails to determine the container's IP address from the file system.

PR #2107, opened 2017-04-05
Implement a proxy for a container to implement concurrent behavior.

### LINKED ISSUE #2091: Model and Implement a Container in its different states
This aims to implement a statemachine on top of the container abstractions we already have. It is key to keep concurrency to a minimum, to have a clear model of what the containers are doing when without the need of having synchronized datastructures to keep their state.

markusthoemmes:
A proxy that wraps a Container. It is used to keep track of the lifecycle
of a container and to guarantee a contract between the client of the container
and the container itself.

The contract is as follows:
1. Only one job is to be sent to the ContainerProxy at one time. ContainerProxy
   will delay all further jobs until the first job is finished for defensiveness
   reasons.
2. The next job can be sent to the ContainerProxy after it indicated capacity by
   sending NeedWork to its parent.
3. A Remove message can be sent at any point in time. Like multiple jobs though,
   it will be delayed until the currently running job has finished.

Closes #2091

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
@@ -150,6 +150,54 @@ case class WhiskAction(
     }
 
     def toJson = WhiskAction.serdes.write(this).asJsObject
+
+    def toExecutableWhiskAction = exec match {
+        case codeExec: CodeExec[_] => Some(ExecutableWhiskAction(namespace, name, codeExec, limits, version, rev))
+        case _                     => None
+    }
+}
+
+/**
+ * Variant of WhiskAction which only includes information necessary to be
+ * executed by an Invoker.
+ *
+ * exec is typed to CodeExec to guarantee executability by an Invoker.
+ *
+ * rev is stored as part of the case-class to make action-matching as
+ * narrow as possible. version is not enough, because a user might delete
+ * an action and recreate it later, effectively resetting the version
+ * counter and thus producing "duplicates".
+ *
+ * @param namespace the namespace for the action
+ * @param name the name of the action
+ * @param exec the action executable details
+ * @param limits the limits to impose on the action
+ * @param version the semantic version
+ * @param rev the revision of the document
+ */
+case class ExecutableWhiskAction(
+    namespace: EntityPath,
+    name: EntityName,
+    exec: CodeExec[_],
+    limits: ActionLimits = ActionLimits(),
+    version: SemVer = SemVer(),
+    rev: DocRevision = DocRevision()) {
+
+    /**
+     * Gets initializer for action. This typically includes the code to execute,
+     * or a zip file containing the executable artifacts.
+     */
+    def containerInitializer: JsObject = {
+        val code = Option(exec.codeAsJson).filter(_ != JsNull).map("code" -> _)
+        val base = Map("name" -> name.toJson, "binary" -> exec.binary.toJson, "main" -> exec.entryPoint.getOrElse("main").toJson)
+        JsObject(base ++ code)
+    }
+
+    /**
+     * The name of the entity qualified with its namespace and version for
+     * creating unique keys in backend services.
+     */
+    final def fullyQualifiedName(withVersion: Boolean) = FullyQualifiedEntityName(namespace, name, if (withVersion) Some(version) else None)
 }
 
 object WhiskAction
--- a/core/invoker/src/main/scala/whisk/core/container/package.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/package.scala
@@ -53,6 +53,14 @@ package object container {
         def duration = Duration.create(end.toEpochMilli() - start.toEpochMilli(), MILLISECONDS)
     }
 
+    object Interval {
+        /** An interval starting now with zero duration. */
+        def zero = {
+            val now = Instant.now
+            Interval(now, now)
+        }
+    }
+
     /**
      * Represents the result of accessing an endpoint in a container:
      * Start time, End time, Some(response) from container consisting of status code and payload
@@ -128,7 +136,7 @@ package object container {
         def apply(content: String) = new DockerOutput(Some(content))
         def unavailable = new DockerOutput(None)
 
-        def isSuccessful(output : DockerOutput) : Boolean =
+        def isSuccessful(output: DockerOutput): Boolean =
             output match {
                 case output if output == DockerOutput.unavailable => false
                 case _ => true
--- a/core/invoker/src/main/scala/whisk/core/containerpool/Container.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/Container.scala
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.containerpool
+
+import scala.concurrent.Future
+import scala.concurrent.duration.FiniteDuration
+
+import spray.json.JsObject
+import whisk.common.TransactionId
+import whisk.core.container.Interval
+import whisk.core.entity.ActivationResponse
+import whisk.core.entity.ByteSize
+
+/**
+ * An OpenWhisk biased container abstraction. This is **not only** an abstraction
+ * for different container providers, but the implementation also needs to include
+ * OpenWhisk specific behavior, especially for initialize and run.
+ */
+trait Container {
+    /** Stops the container from consuming CPU cycles. */
+    def suspend()(implicit transid: TransactionId): Future[Unit]
+
+    /** Dual of halt. */
+    def resume()(implicit transid: TransactionId): Future[Unit]
+
+    /** Completely destroys this instance of the container. */
+    def destroy()(implicit transid: TransactionId): Future[Unit]
+
+    /** Initializes code in the container. */
+    def initialize(initializer: JsObject, timeout: FiniteDuration)(implicit transid: TransactionId): Future[Interval]
+
+    /** Runs code in the container. */
+    def run(parameters: JsObject, environment: JsObject, timeout: FiniteDuration)(implicit transid: TransactionId): Future[(Interval, ActivationResponse)]
+
+    /** Obtains logs up to a given threshold from the container. Optionally waits for a sentinel to appear. */
+    def logs(limit: ByteSize, waitForSentinel: Boolean)(implicit transid: TransactionId): Future[List[String]]
+}
+
+/** Indicates a general error with the container */
+sealed abstract class ContainerError(msg: String) extends Exception(msg)
+
+/** Indicates an error while starting a container */
+sealed abstract class ContainerStartupError(msg: String) extends ContainerError(msg)
+
+/** Indicates an error while starting a container of a managed runtime */
+case class WhiskContainerStartupError(msg: String) extends ContainerStartupError(msg)
+
+/** Indicates an error while starting a blackbox container */
+case class BlackboxStartupError(msg: String) extends ContainerStartupError(msg)
+
+/** Indicates an error while initializing a container */
+case class InitializationError(interval: Interval, response: ActivationResponse) extends Exception(response.toString)
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
@@ -1,0 +1,390 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.containerpool
+
+import java.time.Instant
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.util.Success
+import scala.util.Failure
+
+import akka.actor.FSM
+import akka.actor.Props
+import akka.actor.Stash
+import akka.actor.Status.{ Failure => FailureMessage }
+import akka.pattern.pipe
+import spray.json._
+import spray.json.DefaultJsonProtocol._
+import whisk.common.TransactionId
+import whisk.core.connector.ActivationMessage
+import whisk.core.container.Interval
+import whisk.core.entity._
+import whisk.core.entity.size._
+import whisk.common.Counter
+import whisk.core.entity.ExecManifest.ImageName
+
+// States
+sealed trait ContainerState
+case object Uninitialized extends ContainerState
+case object Starting extends ContainerState
+case object Started extends ContainerState
+case object Running extends ContainerState
+case object Ready extends ContainerState
+case object Pausing extends ContainerState
+case object Paused extends ContainerState
+case object Removing extends ContainerState
+
+// Data
+sealed abstract class ContainerData(val lastUsed: Instant)
+case class NoData() extends ContainerData(Instant.EPOCH)
+case class PreWarmedData(container: Container, kind: String) extends ContainerData(Instant.EPOCH)
+case class WarmedData(container: Container, namespace: EntityName, action: ExecutableWhiskAction, override val lastUsed: Instant) extends ContainerData(lastUsed)
+
+// Events received by the actor
+case class Start(exec: CodeExec[_], memoryLimit: ByteSize)
+case class Run(action: ExecutableWhiskAction, msg: ActivationMessage)
+case object Remove
+
+// Events sent by the actor
+case class NeedWork(data: ContainerData)
+case object ContainerPaused
+case object ContainerRemoved
+
+/**
+ * A proxy that wraps a Container. It is used to keep track of the lifecycle
+ * of a container and to guarantee a contract between the client of the container
+ * and the container itself.
+ *
+ * The contract is as follows:
+ * 1. Only one job is to be sent to the ContainerProxy at one time. ContainerProxy
+ *    will delay all further jobs until a previous job has finished.
+ * 2. The next job can be sent to the ContainerProxy after it indicates available
+ *    capacity by sending NeedWork to its parent.
+ * 3. A Remove message can be sent at any point in time. Like multiple jobs though,
+ *    it will be delayed until the currently running job finishes.
+ *
+ * @constructor
+ * @param factory a function generating a Container
+ * @param sendActiveAck a function sending the activation via active ack
+ * @param storeActivation a function storing the activation in a persistent store
+ */
+class ContainerProxy(
+    factory: (TransactionId, String, ImageName, Boolean, ByteSize) => Future[Container],
+    sendActiveAck: (TransactionId, WhiskActivation) => Future[Any],
+    storeActivation: (TransactionId, WhiskActivation) => Future[Any]) extends FSM[ContainerState, ContainerData] with Stash {
+    implicit val ec = context.system.dispatcher
+
+    // The container is destroyed after this period of time
+    val unusedTimeout = 30.seconds
+
+    // The container is not paused for this period of time
+    // after an activation has finished successfully
+    val pauseGrace = 1.second
+
+    startWith(Uninitialized, NoData())
+
+    when(Uninitialized) {
+        // pre warm a container
+        case Event(job: Start, _) =>
+            factory(
+                TransactionId.invokerWarmup,
+                ContainerProxy.containerName("prewarm", job.exec.kind),
+                job.exec.image,
+                job.exec.pull,
+                job.memoryLimit)
+                .map(container => PreWarmedData(container, job.exec.kind))
+                .pipeTo(self)
+
+            goto(Starting)
+
+        // cold start
+        case Event(job: Run, _) =>
+            implicit val transid = job.msg.transid
+            factory(
+                job.msg.transid,
+                ContainerProxy.containerName(job.msg.user.namespace.name, job.action.name.name),
+                job.action.exec.image,
+                job.action.exec.pull,
+                job.action.limits.memory.megabytes.MB)
+                .andThen {
+                    case Success(container) => self ! PreWarmedData(container, job.action.exec.kind)
+                    case Failure(t) =>
+                        val response = t match {
+                            case WhiskContainerStartupError(msg) => ActivationResponse.whiskError(msg)
+                            case BlackboxStartupError(msg)       => ActivationResponse.applicationError(msg)
+                            case _                               => ActivationResponse.whiskError(t.getMessage)
+                        }
+                        val activation = ContainerProxy.constructWhiskActivation(job, Interval.zero, response)
+                        sendActiveAck(transid, activation)
+                        storeActivation(transid, activation)
+                }
+                .flatMap {
+                    container =>
+                        initializeAndRun(container, job)
+                            .map(_ => WarmedData(container, job.msg.user.namespace, job.action, Instant.now))
+                }.pipeTo(self)
+
+            goto(Running)
+    }
+
+    when(Starting) {
+        // container was successfully obtained
+        case Event(data: PreWarmedData, _) =>
+            context.parent ! NeedWork(data)
+            goto(Started) using data
+
+        // container creation failed
+        case Event(_: FailureMessage, _) =>
+            context.parent ! ContainerRemoved
+            stop()
+
+        case _ => delay
+    }
+
+    when(Started) {
+        case Event(job: Run, data: PreWarmedData) =>
+            implicit val transid = job.msg.transid
+            initializeAndRun(data.container, job)
+                .map(_ => WarmedData(data.container, job.msg.user.namespace, job.action, Instant.now))
+                .pipeTo(self)
+
+            goto(Running)
+
+        case Event(Remove, data: PreWarmedData) => destroyContainer(data.container)
+    }
+
+    when(Running) {
+        // Intermediate state, we were able to start a container
+        // and we keep it in case we need to destroy it.
+        case Event(data: PreWarmedData, _) => stay using data
+
+        // Run was successful
+        case Event(data: WarmedData, _) =>
+            context.parent ! NeedWork(data)
+            goto(Ready) using data
+
+        // Failed after /init (the first run failed)
+        case Event(_: FailureMessage, data: PreWarmedData) => destroyContainer(data.container)
+
+        // Failed for a subsequent /run
+        case Event(_: FailureMessage, data: WarmedData)    => destroyContainer(data.container)
+
+        // Failed at getting a container for a cold-start run
+        case Event(_: FailureMessage, _) =>
+            context.parent ! ContainerRemoved
+            stop()
+
+        case _ => delay
+    }
+
+    when(Ready, stateTimeout = pauseGrace) {
+        case Event(job: Run, data: WarmedData) =>
+            implicit val transid = job.msg.transid
+            initializeAndRun(data.container, job)
+                .map(_ => WarmedData(data.container, job.msg.user.namespace, job.action, Instant.now))
+                .pipeTo(self)
+
+            goto(Running)
+
+        // pause grace timed out
+        case Event(StateTimeout, data: WarmedData) =>
+            data.container.suspend()(TransactionId.invokerNanny).map(_ => ContainerPaused).pipeTo(self)
+            goto(Pausing)
+
+        case Event(Remove, data: WarmedData) => destroyContainer(data.container)
+    }
+
+    when(Pausing) {
+        case Event(ContainerPaused, data: WarmedData) =>
+            context.parent ! NeedWork(data)
+            goto(Paused)
+
+        case Event(_: FailureMessage, data: WarmedData) => destroyContainer(data.container)
+        case _ => delay
+    }
+
+    when(Paused, stateTimeout = unusedTimeout) {
+        case Event(job: Run, data: WarmedData) =>
+            implicit val transid = job.msg.transid
+            data.container.resume().andThen {
+                // Sending the message to self on a failure will cause the message
+                // to ultimately be sent back to the parent (which will retry it)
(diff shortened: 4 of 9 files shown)


core/invoker/src/main/scala/whisk/core/container/package.scala
```diff
@@ -53,6 +53,14 @@ package object container {
         def duration = Duration.create(end.toEpochMilli() - start.toEpochMilli(), MILLISECONDS)
     }
 
+    object Interval {
+        /** An interval starting now with zero duration. */
+        def zero = {
```
rabbah: can't you just make this a value? (does the value of `now` actually matter?)

markusthoemmes: It matters, yes. Imagine a socket exception on `/init`. The activation should have the correct start time, even if it is a zero interval.

core/invoker/src/main/scala/whisk/core/containerpool/Container.scala
```diff
@@ -0,0 +1,57 @@
...
+import whisk.core.entity.ByteSize
+
+/**
+ * An OpenWhisk biased container abstraction. This is **not only** an abstraction
+ * for different container providers, but the implementation also needs to include
+ * OpenWhisk specific behavior, especially for initialize and run.
+ */
+trait Container {
+    /** Stops the container from consuming CPU cycles. */
+    def halt()(implicit transid: TransactionId): Future[Unit]
```
rabbah: curious why you called this `halt` and not `pause`?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+case object Removing extends ContainerState
+
+// Data
+sealed abstract class ContainerData(val lastUsed: Instant)
+case class NoData() extends ContainerData(Instant.EPOCH)
+case class PreWarmedData(container: Container, kind: String) extends ContainerData(Instant.EPOCH)
+case class WarmedData(container: Container, namespace: EntityName, action: WhiskAction, override val lastUsed: Instant) extends ContainerData(lastUsed)
+
+// Events received by the actor
+case class Start(exec: Exec)
```
rabbah: should this be `CodeExec`?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+case object ContainerPaused
+case object ContainerRemoved
+
+/**
+ * A proxy that wraps a Container. It is used to keep track of the lifecycle
+ * of a container and to guarantee a contract between the client of the container
+ * and the container itself.
+ *
+ * The contract is as follows:
+ * 1. Only one job is to be sent to the ContainerProxy at one time. ContainerProxy
```
rabbah: Only one job is to be sent to the ContainerProxy at one time. ->  Only one job is sent to the ContainerProxy at a time?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+ *    will delay all further jobs until the first job is finished for defensiveness
```
rabbah: the first job -> a previous job?

rabbah: why is this for "Defensiveness reasons"? isn't this a correctness issue as well?

core/invoker/src/main/scala/whisk/core/containerpool/Container.scala
markusthoemmes: I chose a different naming than docker commands to force myself not to hardwire this with docker and to make the distinction more "clear".

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+ *    reasons.
+ * 2. The next job can be sent to the ContainerProxy after it indicated capacity by
+ *    sending NeedWork to its parent.
```
rabbah: indicated capacity by sending NeedWork to its parent. -> indicates available capacity by ...

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+ * 3. A Remove message can be sent at any point in time. Like multiple jobs though,
+ *    it will be delayed until the currently running job has finished.
```
rabbah: has finished -> finishes?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
markusthoemmes: Thought the first sounded more like a contract

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+ *
+ * @constructor
+ * @param factory a function generating a Container
+ * @param sendActiveAck a function sending the activation via active ack
+ * @param storeActivation a function storing the activation in a persistent store
+ */
+class ContainerProxy(
+    factory: (TransactionId, String, Exec, ByteSize) => Future[Container],
+    sendActiveAck: (TransactionId, WhiskActivation) => Future[Any],
```
rabbah: Any? is this for legacy reasons?

markusthoemmes: No, the caller just doesn't care about the return type. Using `Future[Unit]` requires a cast at the defining function (like `producer.send(...).map(_ => ())`), which i disliked.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+    storeActivation: (TransactionId, WhiskActivation) => Future[Any]) extends FSM[ContainerState, ContainerData] with Stash {
```
rabbah: one slight tweak on this is to have a factory and a postprocess, which encapsulates active ack, storing the activation, and whatever else we might do in the future.
it's ok for now since there are only 2 post processing steps, but should these grow in the future (eg log drain), will we need to modify this signature further, or can those be encapsulated in the existing hooks.

just a thought.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+    implicit val ec = context.system.dispatcher
+
+    val unusedTimeout = 30.seconds
```
rabbah: comment for these two values

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
markusthoemmes: Don't know. Will have to reassert once we need to do more.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+
+    startWith(Uninitialized, NoData())
+
+    when(Uninitialized) {
+        // pre warm a container
+        case Event(job: Start, _) =>
+            factory(
+                TransactionId.invokerWarmup,
+                ContainerProxy.containerName("prewarm", job.exec.kind),
+                job.exec,
```
rabbah: it feels a bit off to prewarm a container while also passing it an `[job.]exec`.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+                256.MB)
+                .map(container => PreWarmedData(container, job.exec.kind))
```
rabbah: it feels a bit off to prewarm a container while also passing it an `[job.]exec`.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+        // cold start
+        case Event(job: Run, _) =>
+            implicit val transid = job.msg.transid
+            factory(
+                job.msg.transid,
+                ContainerProxy.containerName(job.msg.user.namespace.name, job.action.name.name),
+                job.action.exec,
+                job.action.limits.memory.megabytes.MB)
+                .andThen {
+                    case Success(container) => self ! PreWarmedData(container, job.action.exec.kind)
```
rabbah: using `PreWarmedData` for the cold start, which is also used in the above case, of a stem cell container - should use different type? does that add anything if so other than clarity?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+                            case _                               => ActivationResponse.whiskError(t.getMessage)
+                        }
+                        val activation = ContainerProxy.constructWhiskActivation(job, Interval.zero, response)
+                        sendActiveAck(transid, activation)
+                        storeActivation(transid, activation)
+                }
+                .flatMap {
+                    container =>
+                        run(container, job)
+                            .map(_ => WarmedData(container, job.msg.user.namespace, job.action, Instant.now))
```
rabbah: is holding a reference to the `[job.]action` here necessary? (Seems unused.)

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+    startWith(Uninitialized, NoData())
+
+    when(Uninitialized) {
+        // pre warm a container
+        case Event(job: Start, _) =>
+            factory(
+                TransactionId.invokerWarmup,
+                ContainerProxy.containerName("prewarm", job.exec.kind),
+                job.exec,
+                256.MB)
```
rabbah: should the limit be part of the job as well?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+                    case Success(container) => self ! PreWarmedData(container, job.action.exec.kind)
+                    case Failure(t) =>
+                        val response = t match {
+                            case WhiskContainerStartupError(msg) => ActivationResponse.whiskError(msg)
+                            case BlackboxStartupError(msg)       => ActivationResponse.applicationError(msg)
+                            case _                               => ActivationResponse.whiskError(t.getMessage)
+                        }
+                        val activation = ContainerProxy.constructWhiskActivation(job, Interval.zero, response)
+                        sendActiveAck(transid, activation)
+                        storeActivation(transid, activation)
```
rabbah: this stores the activation without any logs - isn't it possible the container did generate logs but crashed and hence was not responsive? (Without an implementation of `factory` I can't judge completely if that's guaranteed not to be the case.)

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+        // container creation failed
+        case Event(_: FailureMessage, _) =>
+            context.parent ! ContainerRemoved
+            stop()
+
+        case _ => delay
+    }
+
+    when(Started) {
+        case Event(job: Run, data: PreWarmedData) =>
```
rabbah: should `run` be called `initialize[IfNecessary]AndRun`? or maybe a comment here since it looks like you're running a stem cell without initialization of the code (which happens inside `run`).

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+            implicit val transid = job.msg.transid
+            run(data.container, job)
+                .map(_ => WarmedData(data.container, job.msg.user.namespace, job.action, Instant.now))
+                .pipeTo(self)
+
+            goto(Running)
+
+        // pause grace timed out
+        case Event(StateTimeout, data: WarmedData) =>
+            data.container.halt()(TransactionId.invokerNanny).map(_ => ContainerPaused).pipeTo(self)
```
rabbah: fwiw finding a bit of mental friction with `halt` (which i think of as no only ceasing the CPU but also powering off a traditional system).

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+
+        // Only initialize iff we haven't yet warmed the container
+        val initialize = stateData match {
+            case data: WarmedData => Future.successful(Interval.zero)
+            case _                => container.initialize(job.action.containerInitializer, actionTimeout)
+        }
+
+        val activation: Future[WhiskActivation] = initialize.flatMap { initInterval =>
+            val passedParameters = job.msg.content getOrElse JsObject()
+            val boundParameters = job.action.parameters.toJsObject
```
rabbah: I know this is preserving behavior with existing implementation... but this is not necessary because the parameters have already been merged in the controller, so we should be ignoring the action parameters here. Alternatively, the controller should not merge the action parameter (but only merge the package/bindings with the incoming values). I prefer the no remerging here with the invariant that the passed parameters are already fully merged.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+                "namespace" -> job.msg.user.namespace.toJson,
+                "action_name" -> job.msg.action.qualifiedNameWithLeadingSlash.toJson,
+                "activation_id" -> job.msg.activationId.toString.toJson,
+                // compute deadline on invoker side avoids discrepancies inside container
+                // but potentially under-estimates actual deadline
+                "deadline" -> (Instant.now.toEpochMilli + actionTimeout.toMillis).toString.toJson)
+
+            container.run(parameters, environment, actionTimeout)(job.msg.transid).map {
+                case (runInterval, response) =>
+                    val initRunInterval = Interval(runInterval.start.minusMillis(initInterval.duration.toMillis), runInterval.end)
```
rabbah: questioning again why `Interval.zero` actually need to know what instant of time it is "now". since this is only used to subtract from the run start time the duration of the initialization, having a duration of 0 regardless of now should be sufficient.

perhaps it's unavoidable with the current types but I don't see how the start time for a zero interval will matter (it will not reach the recover below, for example).

there is the case above where a "cold start" initialization failure also uses `zero` - but perhaps there is a way to use that interval constructor differently then in this case... and it could be that i'm just paying too much attention to a very small detail.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+            container.logs(job.action.limits.logs.asMegaBytes, exec.sentinelledLogs).map { logs =>
+                activation.withLogs(ActivationLogs(logs.toVector))
+            }
+        }.andThen {
+            case Success(activation) => storeActivation(tid, activation)
+        }.flatMap { activation =>
+            // Fail the future iff the activation was unsuccessful to facilitate
+            // better cleanup logic.
+            if (activation.response.isSuccess) Future.successful(activation)
+            else Future.failed(new Exception())
```
rabbah: why a generic failure/exception with no message - not needed later?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,388 @@
...
+}
+
+object ContainerProxy {
+    def props(factory: (TransactionId, String, Exec, ByteSize) => Future[Container],
+              ack: (TransactionId, WhiskActivation) => Future[Any],
+              store: (TransactionId, WhiskActivation) => Future[Any]) = Props(new ContainerProxy(factory, ack, store))
+
+    // Needs to be threadsafe as it's used by multiple proxys concurrently.
+    private val count = new AtomicInteger(0)
+    private def containerNumber() = count.incrementAndGet()
```
rabbah: don't we have a counter class that is basically this (`whisk.common.Counter`)?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
markusthoemmes: It's used in scheduling later (the old implementation constructed a unique string, while the new one "just" matches that the actions are identical + the invocation namespace is the same)

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
markusthoemmes: is `freeze` a better term maybe?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
markusthoemmes: It's certainly possible, although the scheme I intended to implement was: The factory either returns a Container or it fails.

That case here will only happen though, if `docker run` and/or inspecting the ip address failed (for docker at least). Even for a broken image, that shouldn't fail even if the container crashes immediatly? That'd mean we'd rush through to /init, which will fail but collect logs? Does that make sense?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
markusthoemmes: That's because the implemented docker factory will need to construct the imagename.

`def containerImageName(registry: String, prefix: String, exec: CodeExec[_], tag: String): String`

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
markusthoemmes: Testing this in the big branch first.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
markusthoemmes: I think there are cases (as mentioned above) that will in fact use the Instants to define start and end of the activation.

markusthoemmes:
@rabbah review comments are in, see commits from 8b62c5b on

markusthoemmes:
Rebased to `master`.

When we squash this, please seize all commit messages other than the first (fixup all commits)

markusthoemmes:
PG1 1441 🔵

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -150,6 +150,51 @@ case class WhiskAction(
     }
 
     def toJson = WhiskAction.serdes.write(this).asJsObject
+
+    def toExecutableWhiskAction = ExecutableWhiskAction(namespace, name, exec.asInstanceOf[CodeExec[_]], limits, version, rev)
```
rabbah: Should this return a try or option and defer to the caller to handle the case of a failing cast?

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -0,0 +1,458 @@
...
+
+        // We don't know of this timeout, so we schedule a run.
+        machine ! Run(action, message)
+
+        // State-machine shuts down nonetheless.
+        expectMsg(ContainerRemoved)
+        expectMsg(Transition(machine, Paused, Removing))
+
+        // Pool gets the message again.
+        expectMsg(Run(action, message))
```
rabbah: wondering if we'll want a prioritized message queue in the future - so that message that were ready to execute aren't delayed at the back of the line through no fault of their own.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,383 @@
...
+
+    startWith(Uninitialized, NoData())
+
+    when(Uninitialized) {
+        // pre warm a container
+        case Event(job: Start, _) =>
+            factory(
+                TransactionId.invokerWarmup,
+                ContainerProxy.containerName("prewarm", job.exec.kind),
+                job.exec,
```
rabbah: it would be nice if `job.exec` was avoidable for a stem cell (since it doesn't actually initialize the container).

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,383 @@
...
+                ContainerProxy.containerName(job.msg.user.namespace.name, job.action.name.name),
+                job.action.exec,
+                job.action.limits.memory.megabytes.MB)
+                .andThen {
+                    case Success(container) => self ! PreWarmedData(container, job.action.exec.kind)
+                    case Failure(t) =>
+                        val response = t match {
+                            case WhiskContainerStartupError(msg) => ActivationResponse.whiskError(msg)
+                            case BlackboxStartupError(msg)       => ActivationResponse.applicationError(msg)
+                            case _                               => ActivationResponse.whiskError(t.getMessage)
```
rabbah: should the proxy log the exception message and return a more generic error message here?

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -0,0 +1,458 @@
...
+     */
+    class TestContainer extends Container {
+        var haltCount = 0
+        var resumeCount = 0
+        var destroyCount = 0
+        var initializeCount = 0
+        var runCount = 0
+        var logsCount = 0
+
+        def halt()(implicit transid: TransactionId): Future[Unit] = {
```
rabbah: i think there are missing tests for failure to halt or resume - yes?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,383 @@
...
+object ContainerProxy {
+    def props(factory: (TransactionId, String, CodeExec[_], ByteSize) => Future[Container],
+              ack: (TransactionId, WhiskActivation) => Future[Any],
+              store: (TransactionId, WhiskActivation) => Future[Any]) = Props(new ContainerProxy(factory, ack, store))
+
+    // Needs to be threadsafe as it's used by multiple proxys concurrently.
+    private val containerCount = new Counter
+
+    /**
+     * Generates a unique containername
```
rabbah: containername -> container name

rabbah: (do you mind adding a period at the end of the comment 👍 )

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,383 @@
...
+        }
+    }
+}
+
+object ContainerProxy {
+    def props(factory: (TransactionId, String, CodeExec[_], ByteSize) => Future[Container],
+              ack: (TransactionId, WhiskActivation) => Future[Any],
+              store: (TransactionId, WhiskActivation) => Future[Any]) = Props(new ContainerProxy(factory, ack, store))
+
+    // Needs to be threadsafe as it's used by multiple proxys concurrently.
```
rabbah: threadsafe -> thread-safe

rabbah: proxys -> proxies

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,383 @@
...
+        }.recover {
+            case InitializationError(interval, response) =>
+                ContainerProxy.constructWhiskActivation(job, interval, response)
+        }
+
+        // Sending active ack and storing the activation are concurrent side-effects
+        // and do not block further execution of the future. They are completely
+        // asynchronous.
+        activation.andThen {
+            case Success(activation) => sendActiveAck(tid, activation)
```
rabbah: is the recover broad enough to ensure active ack response on all failures?
(This is related to your past complaints that init/run complete the future with success invariably - or at least that's the idea.)

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,383 @@
...
+        }.flatMap { activation =>
+            val exec = job.action.exec.asInstanceOf[CodeExec[_]]
```
rabbah: no longer need to cast to `CodeExec`.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,383 @@
...
+            response = response,
+            annotations = {
+                Parameters("limits", job.action.limits.toJson) ++
+                    Parameters("path", job.action.fullyQualifiedName(false).toString.toJson) ++ causedBy
+            })
+    }
+}
+
+/** Indicates an activation with a non-successful response */
+case class ActivationUnsuccessfulError(activation: WhiskActivation) extends Exception(activation.toString)
```
rabbah: should this use a more generic exception message rather than serialize the whole record to string?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
rabbah: seems this case wont be reachable anyway https://github.com/openwhisk/openwhisk/pull/2177/commits/157fc6b07897b036032bdc14f32b8c1035193175#diff-b22922b8a6c5f4355fd5f2b343a106e9R103.

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -150,6 +150,51 @@ case class WhiskAction(
...
+    exec: CodeExec[_],
+    limits: ActionLimits = ActionLimits(),
+    version: SemVer = SemVer(),
+    rev: DocRevision = DocRevision()) {
+
+    /**
+     * Gets initializer for action. This typically includes the code to execute,
+     * or a zip file containing the executable artifacts.
+     */
+    def containerInitializer: JsObject = {
```
markusthoemmes: Only if you also change the existing Invoker impl, which I didn't.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
markusthoemmes: I think that's due to the fact that we currently build our image names using a whole exec. I agree that kind would be enough here, but needs a more throrough change to our image methods?

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
markusthoemmes: Note that this only puts the message on the end of the "internal" queue, that is: in the 16th position.

It is also assumed that the scheduling algorithm takes very little time, so internal queue messages are recomputed almost immediatly (at least that was the assumption I took when doing this)

Hence I claim that we won't have prioritized message queues on this level (not speaking of the other places though)

rabbah: Not convinced since being q'ed behind some small N actions still means some delay proportional to the expected action execution in the best case; if that's small (msecs) all good if that's (secs), eh. Anyway that's for the future.

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
rabbah: Fair. Repeal and replace!

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
rabbah: Maybe my outstanding manifest pr fixes this. We'll see

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
markusthoemmes: It wouldn't depend on the execution time, scheduling is decoupled of that.

rabbah: I don't understand what that means. The time a request sits in the queue is dependent on how soon a container becomes available to dispatch the request to. Is that not true?

markusthoemmes: That is true, but in this case i think it doesn't matter. I'll explain further f2f, will be easier.

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -0,0 +1,458 @@
...
+            }
+        }
+        def factory(tid: TransactionId, name: String, exec: Exec, memoryLimit: ByteSize) = Future.successful(container)
+
+        val machine = childActorOf(ContainerProxy.props(factory, ack, store))
+        registerCallback(machine)
+        machine ! Run(action, message)
+        expectMsg(Transition(machine, Uninitialized, Running))
+        expectMsg(ContainerRemoved)
+        expectMsg(Transition(machine, Running, Removing))
```
markusthoemmes: The `ContainerRemoved` message is sent long before the container is really removed. It's sent in the moment the container decides to remove itself, and thus tells its parent it's gone already (as the parent should treat it like its gone).

You're right, from this perspective it's confusing. I'll have a look.

markusthoemmes:
@rabbah thanks for the thorough review, see the fixes applied from https://github.com/openwhisk/openwhisk/pull/2107/commits/152da172197759981bc2a87da5841aa1968c7669 on.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
rabbah: this is nicer now - can we comment the factory signature (boolean i imagine is `pull`?)

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -318,6 +318,59 @@ class ContainerProxyTests extends TestKit(ActorSystem("ContainerProxys"))
...
+        run(machine, Uninitialized)
+        timeout(machine) // times out Ready state so container suspends
+        expectMsg(Transition(machine, Ready, Pausing))
+        expectMsg(ContainerRemoved)
+        expectMsg(Transition(machine, Pausing, Removing))
+
+        awaitAssert {
+            container.suspendCount shouldBe 1
+            container.destroyCount shouldBe 1
+        }
```
rabbah: Thanks for adding these. If docker gets sick and removing itself fails is there anything else that can be done? I'm thinking of some of the cases we've encountered in the past iirc.

Issue #2011, opened 2017-03-13
support specifying both image and code in wsk CLI


glikson:
In some cases it is useful to augment the standard runtime image (e.g., openwhisk/nodejs6action) with additional binaries (e.g., installation of a certain library, such as `opencv`). In order to do so and still take advantage of 'standard' lifecycle management of the action code itself, it is possible today to use the REST API, specifying `kind=blackbox` with custom `image` as well as `code` (plain text or 'binary' -- base64-encoded zip).
It would be great to enable this option in the CLI too, e.g.:
```
$ wsk action create --image mydockerhubaccount/mynodejsactionimage myaction.zip
```
(Note: can consider reusing the "--docker" flag)

related to #1927

rabbah:
Closed by https://github.com/apache/incubator-openwhisk/commit/9c1f77fdb1e4a7339634434dc90c45026d6a43d2.

PR #2112, opened 2017-04-06
Allow docker actions with zip files. 

rabbah:
Add native action for binary injection.
Proposed fix for #2011.

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -827,7 +827,7 @@ For the instructions that follow, assume that the Docker user ID is `janesmith` 
 
 
   ```
-  wsk action create --docker example janesmith/blackboxdemo
+  wsk action create example --docker janesmith/blackboxdemo
   ```
 
   Notice the use of `--docker` when creating an action. Currently all Docker images are assumed to be hosted on Docker Hub.
@@ -853,10 +853,29 @@ For the instructions that follow, assume that the Docker user ID is `janesmith` 
   ./buildAndPush.sh janesmith/blackboxdemo
   ```
   ```
-  wsk action update --docker example janesmith/blackboxdemo
+  wsk action update example --docker janesmith/blackboxdemo
   ```
 
   You can find more information about creating Docker actions in the [References](./reference.md#docker-actions) section.
+
+  *Note:* Previous version of the CLI supported `--docker` without a parameter and the image name was a positional argument.
+  In order to allow Docker actions to accept initialization data via a (zip) file, similar to other actions kinds, we have
+  normalized the user experience for Docker actions so that a positional argument if present must be a file (e.g., a zip file)
+  instead. The image name must be specified following the `--docker` option. Furthermore, due to user feedback, we have added
+  `--native` as shorthand for `--docker openwhisk/dockerskeleton` so that executables that run inside the standard Docker action
+  SDK are more convenient to create and deploy.
+
+  For example, the tutorial above created a binary executable inside the container locate at `/action/exec`. If you copy this file
+  to your local file system and zip it into `exec.zip` then you can use the following commands to create a docker action which receives
+  the executable as initialization data.
+
+  ```bash
+  wsk action create example exec.zip --native
+  ```
+  which is equivalent to the following command.
+  ```bash
+  wsk action create example exec.zip --docker openwhisk/dockerskeleton
+  ```
 
 ## Watching action output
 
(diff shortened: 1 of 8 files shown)


csantanapr:
@rabbah want to close this one as superseded by #2331 ?

csantanapr:
Closes #2011
Closes #2331

rabbah:
@dubeejw @mdeuser may i get your critique on these changes?
The idea is to allow `wsk action create action-name some.zip --docker image/name`.

mdeuser:
This looks to be a breaking change to existing CLI behavior.  Currently, the `--docker` flag is a boolean; with this PR the `--docker flag` will take an argument.  With this change will existing scripts that leverage the `--docker` flag break?

rabbah:
If one used `--docker` as `wsk action create example --docker janesmith/blackboxdemo` then there is no effect with the proposed change. I'll call this the `conforming` case.

If on the other had the `--docker` option was used any where else, then yes such commands may no longer work.

I have no evidence to suggest the common case is the  `conforming` case, but I think it is. It's too bad we don't have semantic versioning for the CLI.

What do you suggest @mdeuser? Using a different flag (`--image`?) or accepting that this is a breaking change?

mdeuser:
I prefer avoiding breaking CLI changes.  Is the intent of this new option to allow for 1 or more files to be "copied" somewhere into the action runtime folder - regardless of the runtime?

rabbah:
Yes that is the intent. `wsk action create foo --kind nodjes:6 my.zip` and `wsk action create bar --docker imagename my.zip` become the same API call.

I don't think I want to introduce another option (what would we call it `--image`?) - maybe I can preserve the existing behavior where `--docker` accepts an optional argument and in the absence of one, the artifact name is used as today. This seems like unnecessary complexity though. What do you suggest?

rabbah:
with --image and --docker being redundant, I'd want to deprecate the latter we shouldn't have two way of doing the same thing.

rabbah:
Summarizing discussion on Slack with @mdeuser and @dubeejw:

1. there is a strong preference to preserve `--docker` and not break behavior
2. there is preference to add second position argument which if present is the zip file for the docker action (`wsk action create name image [zipfile] --docker`)
3. add new option `--image` which accepts an image name and positional argument is the zip file (`wsk action create name [zipfile] --image image`); this preserves `--docker` behavior.
4. add `--docker` with `--image` to specify the image name and positional argument is the zip file (`wsk action create name [zipfile] --docker --image image`)
5. add `--kind docker` with `--image` to specify the image name, and positional argument is the zip file (`wsk action create name [zipfile] --kind docker --image image`)
6. make a breaking change so that the docker flag takes an artifact which is this PR in its current state (`wsk action create name [zipfile] --docker image`)

csantanapr:
I think the two options I will be OK will be with a modify version of 4 or 5

Propose modify versions where --image is optional, if not specify use the default value `openwhisk/dockerskeleton`
Option 4b
```
wsk action create name [zipfile] --docker [--image image]
```
Option 5b
```
wsk action create name [zipfile] --kind docker [--image image]
```

csantanapr:
OK option 4b is invalid
```
wsk action create name [zipfile] --docker [--image image]
```
Because user can use `wsk action create name path/file --docker` and we don't know if `path/file` is a path to the zip/archive or an image name

csantanapr:
I don't like option 3 too vague
```
wsk action create name [zipfile] --image image
```

csantanapr:
If I have to pick again multiple options option 2 or option 5b.
option 5b is being more consistency with other type of actions.
option 2 is being a short way of creating actions.

I'm starting to like option 2 😸

rabbah:
we’ve made breaking changes to the cli before - i feel like we’re obsessing over this small detail which i think will break very few users and is actually the best of all the options (option 6). I tried to preserve the old behavior and avoid adding a new parameter - it can’t be done easily and the refactoring was staring to pile up. I then implemented option 2 but don’t like it, i feel it’s prone to error and inconsistent with the other kinds. So since we don't have auto cli upgrades, those that don’t upgrade wont actually be broken at all. I therefore posit we should adopt the best option going forward and suggest that's option 6. I updated the documentation accordingly to include examples and a note on the new behavior.

csantanapr:
OK I think the return on investment is too high, I'm OK with option 6
Low risk of corner case, but let's doc the notice and show the benefits

rabbah:
PG2/1672 🔵

Issue #2028, opened 2017-03-15
wsk property set --auth should not echo back the API key that was set


krook:
When setting the `wsk` auth property the response message back prints the key that was given. This can lead to inadvertent logging of the key in public logs.

It would be great if the message back masked this property. If the user needed the info, they could use `wsk property get --auth` explicitly.

For example, current behavior:
```bash
wsk property set --auth $KEY_FROM_ENV_VARIABLE
ok: whisk auth set to [Explicit auth key displayed]
```

Ideally it just masks this while confirming it is set on the reply instead:
```bash
wsk property set --auth $KEY_FROM_ENV_VARIABLE
ok: whisk auth set. Run "wsk property get --auth" to see the new value.
```

PR #517, opened 2016-05-26
Fix issue #388.

lzbj:
Add the delete, list webhook operation.
Changed parameter names and descriptions.

PR #2115, opened 2017-04-06
issue 2028 do not print auth key on console

RSulzmann:
Fixes https://github.com/openwhisk/openwhisk/issues/2028
wsk property set --auth <xxxxxxx>    should not echo the auth key to the console

changed output string, NLS translation id and text, and the related unit test case

### DIFF
--- a/tools/cli/go-whisk-cli/commands/property.go
+++ b/tools/cli/go-whisk-cli/commands/property.go
@@ -186,15 +186,8 @@ var propertyUnsetCmd = &cobra.Command{
         if flags.property.auth {
             delete(props, "AUTH")
             okMsg += fmt.Sprintf(
-                wski18n.T("{{.ok}} whisk auth unset",
+                wski18n.T("{{.ok}} whisk auth unset.\n",
                     map[string]interface{}{"ok": color.GreenString("ok:")}))
-            if len(DefaultAuth) > 0 {
-                okMsg += fmt.Sprintf(
-                    wski18n.T("; the default value will be used. Run 'wsk property get --auth' to see the new value.\n"))
-            } else {
-                okMsg += fmt.Sprint(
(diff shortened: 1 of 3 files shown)


rabbah:
@RSulzmann if you post an approving PG, I'll merge.

RSulzmann:
PG3   #517 ok

tools/cli/go-whisk-cli/commands/property.go
```diff
@@ -190,8 +190,7 @@ var propertyUnsetCmd = &cobra.Command{
                     map[string]interface{}{"ok": color.GreenString("ok:")}))
             if len(DefaultAuth) > 0 {
                 okMsg += fmt.Sprintf(
-                    wski18n.T("; the default value of {{.default}} will be used.\n",
-                        map[string]interface{}{"default": boldString(DefaultAuth)}))
+                    wski18n.T("; the default value will be used. Run 'wsk property get --auth' to see the new value.\n"))
```
rabbah: should the error message remove the reference to "default value" here since there is no default value or rather the default value is undefined?

mdeuser: Since the `DefaultAuth` const value is currently an empty string, this code will actually never be hit in practice.

rabbah: Should this be asserted as an error then instead?

RSulzmann: The unset property function does remove the current property setting, write and reload the property file.
During reload of the property file, defaults are set.
For `DefaultAuth` and `DefaultAPIHost` the default value is "".

If these default values never changes, the discussed info message can be removed for the two properties.
Updating code.

markusthoemmes:
@rabbah any more concerns?

PR #494, opened 2016-05-24
Add the explanation on how to add a Python action

houshengbo:
Closes-Bug: #376

Issue #487, opened 2016-05-22
move log markers to the end of log messages


rabbah:
The presence of log markers makes reading raw log messages harder difficult - since they are not fixed width and not present on all messages. I suggest moving them to the end of the log message.

@ChrBi

rabbah:
@markusthoemmes FYI.

Issue #474, opened 2016-05-19
Avoid sequences to fail silently in case of compile errors within actions


csantanapr:
From @nauerz-ibm 
Sequences fail silently if there are compile errors within an action of the sequence.

@sjfink : This behavior was at least observed before InterConnect; close if the behavior has meanwhile be improved.

From @rabbah : 
we need unit tests for these kinds of bugs - this is a good way of starting to fix them.

ioana-blue:
I think with the new implementation the sequence will not fail silently. I think the sequence will fail with the error returned by the "faulty" action, but if any "good" actions precede the faulty one, they will be executed. We need a test for this.

rabbah:
A lot has changed with sequences.

PR #2116, opened 2017-04-06
Bump Kafka Connector Test timeout to avoid build breaks

mdeuser:
- current limit of 10 seconds is too small on PG3 - causing build failures
- other PG3 connector test times have recently been observed at 8/9+ seconds

### DIFF
--- a/tests/src/test/scala/services/KafkaConnectorTests.scala
+++ b/tests/src/test/scala/services/KafkaConnectorTests.scala
@@ -73,7 +73,7 @@ class KafkaConnectorTests
         for (i <- 0 until 5) {
             val message = new Message { override val serialize = Calendar.getInstance().getTime().toString }
             val start = java.lang.System.currentTimeMillis
-            val sent = Await.result(producer.send(topic, message), 10 seconds)
+            val sent = Await.result(producer.send(topic, message), 20 seconds)
             val received = consumer.peek(10 seconds).map { case (_, _, _, msg) => new String(msg, "utf-8") }
             val end = java.lang.System.currentTimeMillis
             val elapsed = end - start


mdeuser:
@jeremiaswerner - Would you kindly review/merge?

mdeuser:
PG2 1407

rabbah:
@mdeuser I suspect there's a race in these tests actually - not convinced the bump will help.

mdeuser:
@rabbah - Would the race condition result in a timeout or a message conflict?  PG3 builds #500, #494, and #487 all passed the test, but with elapsed times very close to exceeding the current timeout value.  I did see another failure in PG3 #474 in which all tests failed... that could be due to messages spilling over into the follow-on kafka tests.  Perhaps both issues are at play?

mdeuser:
@rabbah - I'm still seeing various build failures due to these Kafka connector tests.  Were you able to find a root cause for the longer-than-expectect initial ```producer.send(topic, message)``` call?

rabbah:
I was wrong about the race - wast thinking of a different kafka test suite. I don't see a race here. Inspecting one of the failed jenkins builds (pg2/1448), I see:

```
    [2017-04-16T20:01:07.292Z] [DEBUG] [??] [KafkaProducerConnector] sending to topic 'Dinosaurs' msg 'Sun Apr 16 16:01:07 EDT 2017'
    [2017-04-16T20:01:17.301Z] [DEBUG] [??] [KafkaProducerConnector] sending to topic 'Dinosaurs' msg 'Sun Apr 16 16:01:17 EDT 2017'
    [2017-04-16T20:01:18.305Z] [DEBUG] [??] [KafkaProducerConnector] sending to topic 'Dinosaurs' msg 'Sun Apr 16 16:01:18 EDT 2017'
    [2017-04-16T20:01:23.598Z] [DEBUG] [??] [KafkaProducerConnector] sent message: Dinosaurs[0][1]
    [2017-04-16T20:01:23.598Z] [DEBUG] [??] [KafkaProducerConnector] sent message: Dinosaurs[0][0]
    [2017-04-16T20:01:23.598Z] [DEBUG] [??] [KafkaProducerConnector] sent message: Dinosaurs[0][2]
```

The first three send requests are 10 seconds apart - each of these correspond to one send attempt per test. You'll see from the `sent` message that there was a delay sending these and the publisher didn't actually complete until 16 seconds after the very first send request.

Reasons TBD.

mdeuser:
The other Kafka Connector test failure I've encountered is
```
services.KafkaConnectorTests > Kafka connector should send and receive a kafka message which sets up the topic FAILED
    java.util.concurrent.TimeoutException: Futures timed out after [10 seconds]
```
When the test has been successful, the wait times for the initial post are close to the 10 second limit; so if this is normal Kafka behavior for producers publishing a message to a new topic, then perhaps the timeout needs increasing.
```
services.KafkaConnectorTests > Kafka connector should send and receive a kafka message which sets up the topic STANDARD_OUT
    (0) Received 1. Took 8029 msec: List(Thu Apr 06 09:34:02 EDT 2017)

    (1) Received 1. Took 2 msec: List(Thu Apr 06 09:34:10 EDT 2017)

    (2) Received 1. Took 1 msec: List(Thu Apr 06 09:34:10 EDT 2017)

    (3) Received 1. Took 1 msec: List(Thu Apr 06 09:34:10 EDT 2017)

    (4) Received 1. Took 1 msec: List(Thu Apr 06 09:34:10 EDT 2017)


services.KafkaConnectorTests > Kafka connector should send and receive a kafka message which sets up the topic PASSED
```

mdeuser:
pg3/543 hit the kafka connector test failure.
@bjustin-ibm - Is this kafka delay expected when initially publishing a msg to a new topic?

bjustin-ibm:
Just taking a quick look at this, I don't think the topic is "new" in that it is created in the test - from what I can tell the test assumes the topic already exists. 

Regardless if the topic is newly-created or not, you _might_ expect that the consumer could be delayed in consuming the posted message, or may miss it entirely, if it had not yet been assigned the partition the message was produced to. This assignment is done only after some initial handshaking between the consumer and the broker(s) and, in my experience, can take numerous seconds for this assignment to happen.

That _may_ be what is happening here to explain the longer time to consume the first message posted. That being said, the timeout increased in this PR is on the producer side, which would not address this problem.

@mdeuser Are you sure the timeout was on the produce side, or was it on the consume side?

I'm not terribly familiar with the Java/Scala API for Kafka, but the correct fix may involve instead waiting until such time that the consumer actually gets partition assignments from the broker(s) before producing the messages. If that isn't possible with the API, then I suppose increasing the consumer timeout may help.

Or... the timeout really did happen on the producer, and I have no explanation for that 😉

rabbah:
@bjustin-ibm note that from the logs I posted - the issue is that the post (producer) was delayed (by 16 seconds - before the producer succeeded).

bjustin-ibm:
Ah ok. I was focused on this `(0) Received 1. Took 8029 msec: List(Thu Apr 06 09:34:02 EDT 2017)`

In this case, I don't know what might cause the delay to produce a message, other than a laggy network, or slow broker(s) handling the initial producer handshake. In this case, I can't think of anything better than simply increasing the timeout value.

rabbah:
shall we go to 30 seconds then? i'm with you - don't have a good explanation.

mdeuser:
I'm not sure this is even possible, but are there kakfa metrics we can glean that might show where this delay is coming from?
https://www.datadoghq.com/blog/monitoring-kafka-performance-metrics/

bjustin-ibm:
@mdeuser I'm not familiar with this client, but the one I use (Python) allows you to set some debug flags which cause it to log lots of interesting information. Perhaps your client has something similar.

csantanapr:
@mdeuser told me you would be able to review and merge this one, have fun 🌮  !

PR #2122, opened 2017-04-10
Exit components on OutOfMemory exceptions

markusthoemmes:
Scala's idiomatic error handling makes it "hardish" to exit the JVM on fatal errors (for example if a Future fails with a fatal exception). The added flags just ignore error handling and exit the JVM, which we can recover by a docker auto-restart for example.

I bumped the Java versions because the flag was added in 8u92. Figured we might as well go to latest. It got released in January, so it should be stable.

### DIFF
--- a/core/controller/build.gradle
+++ b/core/controller/build.gradle
@@ -20,4 +20,4 @@ tasks.withType(ScalaCompile) {
 }
 
 mainClassName = "whisk.core.controller.Controller"
-applicationDefaultJvmArgs = ["-Xmx2g", "-XX:+ExitOnOutOfMemoryError", "-XX:+PrintGCDetails"]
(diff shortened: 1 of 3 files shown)


markusthoemmes:
Unfortunately, yes. Other oracle JDK dockerhub images do the same.

core/controller/build.gradle
```diff
@@ -20,4 +20,4 @@ tasks.withType(ScalaCompile) {
 }
 
 mainClassName = "whisk.core.controller.Controller"
-applicationDefaultJvmArgs = ["-Xmx2g"]
+applicationDefaultJvmArgs = ["-Xmx2g", "-XX:+ExitOnOutOfMemoryError", "-XX:+PrintGCDetails"]
```
rabbah: @perryibm ok?

PR #2128, opened 2017-04-11
change file name

KrishMunot:
Standardising licence file name to markdown


rabbah:
Thanks for the suggestion.
I expect that as this repository is moved to the Apache organization (which is an ongoing process), this file will become simply LICENSE (see for example https://github.com/apache/kafka/blob/trunk/LICENSE).

rabbah:
@csantanapr @mrutkows is there a standard practice for how to name the license file? i see other apache repos use `LICENSE`.

daisy-ycguo:
I think both LICENSE and LICENSE.txt are fine. I see some projects uses LICESE.txt and some project uses LICENSE.

rabbah:
@KrishMunot  do you have an Apache CLI on file?

rabbah:
shall we rename CREDITS.txt as well?

rabbah:
@KrishMunot thanks for this suggestion. If I look at the two files:

https://github.com/KrishMunot/openwhisk/blob/c1b836d685466b799583e671f3bf5a7c03edec88/LICENSE.md

and
https://github.com/apache/incubator-openwhisk/blob/master/LICENSE.txt

Notice in the second:
screen shot 2017-09-07 at 9 50 17 am

https://help.github.com/articles/licensing-a-repository/
either is accepted - i think we'll keep it as text though since the license isn't really using markdown.

PR #2130, opened 2017-04-11
Reduce unnecessary typecasts in log reading + turn off activation caching

markusthoemmes:
These bits are rather sensitive and can have quite high volumes of data passing through. Therefore, casting through different collections is to be avoided.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
@@ -113,6 +113,6 @@ object WhiskActivation
     override val collectionName = "activations"
     override implicit val serdes = jsonFormat13(WhiskActivation.apply)
 
-    override val cacheEnabled = true
+    override val cacheEnabled = false
     override def cacheKeyForUpdate(w: WhiskActivation) = w.docid.asDocInfo
 }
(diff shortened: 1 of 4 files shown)


common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
```diff
@@ -113,6 +113,6 @@ object WhiskActivation
     override val collectionName = "activations"
     override implicit val serdes = jsonFormat13(WhiskActivation.apply)
 
-    override val cacheEnabled = true
+    override val cacheEnabled = false
```
rabbah: add a comment perhaps that caching activations is not of much use in the common case?

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
```diff
@@ -110,6 +111,9 @@ protected[invoker] trait ActionLogDriver {
             }
         }
 
+        if (lines.hasNext) truncated = true
+        if (truncated) logLines.append(Messages.truncateLogs(limit))
+
         ((hasOut && hasErr) || !requireSentinel, truncated || lines.hasNext, logLines.toVector)
```
rabbah: can remove this now: `truncated || lines.hasNext`

markusthoemmes:
@rabbah changes done.

Issue #2101, opened 2017-04-03
Deployment of Go CLI from the separate repo openwhisk-client-go


houshengbo:
Since the Go CLI will be moved under a separate repo openwhisk-client-go, openwhisk needs to change the current deployment for Go CLI to install it from openwhisk-client-go repo.

After the movement is done for Go CLI, the Go CLI code under openwhisk will be removed.

pritidesai:
I have the PR at https://github.com/openwhisk/openwhisk-cli/pull/4

This PR can generate wsk CLI under openwhisk-cli/bin.

dubeejw:
Please create a repo that preserves the Git history of the existing CLI, before continuing this work.

houshengbo:
@pritidesai Could you start to work on the script for openwhisk to innstall Go CLI in the new repo?
I think the build part is ready in Go CLI repo.

csantanapr:
I'm happy to close this now 🎉

PR #2131, opened 2017-04-11
WIP: Building WSK CLI

pritidesai:
building wsk cli off of openwhisk/openwhisk-cli instead
of from local openwhisk/openwhisk repo

These changes are still under testing and not completely working on my local system.


markusthoemmes:
What's the intended flow in the end for our CLI story? How will we release new versions of the CLI?

dubeejw:
Please add me as a reviewer, @pritidesai.

dubeejw:
@pritidesai, please do not build the CLI using Ansible.

houshengbo:
@dubeejw Do we plan to download the binaries directly from the release? I thought we should build them from the latest code. At least we should have an option to either build from code or download directly.

dubeejw:
@houshengbo, every time new code is pushed to the CLI repo new binaries will be generated and posted to the git release page. Therefore, OpenWhisk will always download the latest CLI.

houshengbo:
@markusthoemmes The go cli is starting its journey to become a separate repo: https://github.com/openwhisk/openwhisk-cli/. This repo has already hosted the latest cli and able to build automatically if we add a new tag to the master branch.

mdeuser:
@pritidesai - Same basic comment as above.  Will CLI builds from the new repo be responsible for publishing the CLI binaries ?

dubeejw:
@mdeuser, @csantanapr

houshengbo:
@dubeejw If we need to build the binary for every commit, how do we deal with the old binary built from old commits? Keep them all or overwrite them? How do we manage the major version?

houshengbo:
@mdeuser @dubeejw Currently the new cli repo https://github.com/openwhisk/openwhisk-cli/ is able to build the binaries in an auto way, each time we add a tag to the master branch.

mdeuser:
@houshengbo - Is the tag set and updated automatically or manually?

houshengbo:
@mdeuser Tag is done manually.

pritidesai:
@dubeejw what should we use to build wsk cli if not ansible? 

I guess it depends on if we go in the same route as openwhisk-client-swift where the builds are published under GitHub/releases and have openwhisk point to the hosted binary.

https://github.com/openwhisk/openwhisk-client-swift/releases/download/0.2.3/starterapp-0.2.3.zip

Or we keep this PR as is to have openwhisk build latest binaries from openwhisk-cli.

houshengbo:
From my perspective, downloading the latest code of go cli makes sense in ansible, since we need provide an option to build from the latest code of go cli. This is something similar we did for openwhisk-catalog in ansible.

Well, if the user likes to directly download the binary, we can also add that part into the ansible scripts. What do you think? @dubeejw @csantanapr @mdeuser

dubeejw:
If creating a new git release automatically is not possible, let's publish the latest CLI binaries in DockerHub automatically and have OpenWhisk pull the latest every time.

dubeejw:
We use Ansible for deploying containers that are already built.

csantanapr:
This repo can produce a CLI binary and also push docker images to docker hub as "latest" and also as "tag" version 1.x.y

dubeejw:
Part of the benefit of removing the CLI from the OpenWhisk repo is that OpenWhisk will take less time to build as it does not have to build the CLI.

dubeejw:
This document describes how to publish to DockerHub using Git: https://docs.docker.com/docker-hub/builds/.

pritidesai:
@dubeejw docker hub needs an account, what is the best practice for creating such account?

mdeuser:
What triggers the build and on what system(s) will the CLI be built?  That is, will docker hub build a docker container with the CLI?  Will another system also build the CLI?

dubeejw:
We have a Docker account here: https://hub.docker.com/r/openwhisk/. I believe @csantanapr has the credentials.

pritidesai:
@csantanapr is out until 4/20 :( anyone else who would have those credentials to https://hub.docker.com/r/openwhisk/?

csantanapr:
@rr had the credentials if needed also has admin for Travis

dubeejw:
@pritidesai, @houshengbo, we don't want to use the latest CLI code that means we don't need to use DockerHub. Since the CLI is being versioned, OpenWhisk should download the latest versioned binaries from the Git releases page. If we want OpenWhisk to use a newer version of the CLI, we will have to manually tag and publish new binaries to the Git releases page.

houshengbo:
@dubeejw @pritidesai 
I have already enabled the travis ci to build the latest binaries for go cli, and they are able to publish on the releases page automatically.

We can always access the latest version via the following links:
https://github.com/openwhisk/openwhisk-cli/releases/download/latest/OpenWhisk_CLI-latest-linux-amd64.zip;
https://github.com/openwhisk/openwhisk-cli/releases/download/latest/OpenWhisk_CLI-latest-darwin-amd64.zip;
https://github.com/openwhisk/openwhisk-cli/releases/download/latest/OpenWhisk_CLI-latest-windows-amd64.zip

We have three binaries at this moment. They can be used in this PR to get the latest binaries.

We also have one major release available, v1.0:
https://github.com/openwhisk/openwhisk-cli/releases/download/v1.0/OpenWhisk_CLI-v1.0-darwin-amd64.zip
https://github.com/openwhisk/openwhisk-cli/releases/download/v1.0/OpenWhisk_CLI-v1.0-linux-amd64.zip
https://github.com/openwhisk/openwhisk-cli/releases/download/v1.0/OpenWhisk_CLI-v1.0-windows-amd64.zip


The naming convention for the download link is https://github.com/openwhisk/openwhisk-cli/releases/download/<tag>/OpenWhisk_CLI-<tag>-<OS>-amd64.zip.

mdeuser:
@houshengbo - prior to publishing the CLI binary, have all CLI tests passed successfully?

dubeejw:
Travis in the CLI repo would have to pull down the latest OpenWhisk, build/deploy it then run just the CLI tests against it.

houshengbo:
@mdeuser @dubeejw Do we currently have test cases available for CLI, in order to guarantee the CLI's and CLI binaries' quality?

There are no test code within the go-whisk and go-whisk-cli folders. There are tests written in scala in openwhisk, using the CLI binaries, but they are not the tests for CLI.

mdeuser:
@houshengbo - Actually, the predominant CLI tests are all currently written in scala - both positive and negative tests.

pritidesai:
Are they all located [here](https://github.com/openwhisk/openwhisk/tree/master/tests/src/test/scala/whisk/core/cli/test)?

mdeuser:
The bulk of the CLI-centric tests are in WskBasicUsageTests.scala.  I think a few other CLI tests are outside this file, but not as many

pritidesai:
I want to now update [nginx.conf](https://github.com/openwhisk/openwhisk/blob/master/ansible/roles/nginx/templates/nginx.conf.j2#L63) file so that it pulls CLI from openwhisk-cli [releases](https://github.com/openwhisk/openwhisk-cli/releases/tag/latest) page. 

How do we maintain following directory structure and decompress `wsk` cli, all this being part of nginx? do we still need role for [CLI](https://github.com/openwhisk/openwhisk/tree/master/ansible/roles/cli/tasks)?

```
ls /etc/nginx/cli/go/download/linux/amd64/
OpenWhisk_CLI-linux.tgz  wsk
```

rabbah:
You could just have nginx serve a json object with the links for each platform.
I never liked the way we rendered the directory anyway.

csantanapr:
Yes content.json

pritidesai:
I have the changes for ready to review now. I have tested these changes on mac.

Here are the list of changes:

WSK CLI is downloaded from openwhisk/openwhisk-cli repo.
CLI is located at releases page on github in a zip format [here](https://github.com/openwhisk/openwhisk-cli/releases).
These changes downloads zip file and unarchives it to
produce wsk executable file depending on the env.

```
ls -R /Users/Shared/nginx/cli/go/download/
mac/

/Users/Shared/nginx/cli/go/download//mac:
amd64/

/Users/Shared/nginx/cli/go/download//mac/amd64:
OpenWhisk_CLI-mac.zip  wsk*
```

We are loosing content.json file with these changes as its not available with cli built on GitHub. Do we need that file at `/Users/Shared/nginx/cli/go/download/`?

Also, would like to parameterize this url (https://github.com/openwhisk/openwhisk-cli/releases/download/latest/OpenWhisk_CLI-latest-linux-amd64.zip) instead of hardcoding it to always get latest. Don't know the exact syntax and parameter name for that. Please help.

In the end, please provide your feedback/comments on ansible code structure, I have no expertise in Ansible or Enginx. 

Closes #2101

ansible/roles/cli/tasks/download_cli_mac.yml
```diff
@@ -0,0 +1,28 @@
...
+  file:
+    path: "{{ cli_nginx_dir }}/mac/amd64"
+    state: directory
+
+- name: "download cli into ngnix"
+  local_action: >
+    get_url
+    url="https://github.com/openwhisk/openwhisk-cli/releases/download/latest/OpenWhisk_CLI-latest-darwin-amd64.zip"
+    dest="{{ cli_nginx_dir }}/mac/amd64/OpenWhisk_CLI-mac.zip"
+    validate_certs=False
```
markusthoemmes: Why not validate certs? github certs should be perfectly fine?

ansible/roles/cli/tasks/download_cli_linux.yml
```diff
@@ -0,0 +1,28 @@
...
+    creates=wsk
+    mode=0755
+
+- name: "download cli (linux) to openwhisk home at {{ openwhisk_home }}"
+  local_action: >
+    get_url
+    url="https://{{ groups['edge'] | first }}/cli/go/download/linux/amd64/wsk"
+    dest="{{ openwhisk_home }}/bin/wsk"
+    mode=0755
+    validate_certs=False
```
markusthoemmes: That file and the mac download file are close to exact copies. Can we parametrize? You can have a look at the db scripts for inspiration.

pritidesai:
@rabbah windows version of cli is available on openwhisk-cli release page but wasn't part of edge router before I started working on this. Do we want to include windows version as well? Also, to address local build of cli and bundling it with edge router, I don't know if its possible or not. OpenWhisk-CLI does build a CLI image just like before, may be we can use this image if it exists and pull the CLI out from the image instead of downloading it from releases page, thoughts?

ansible/roles/cli/tasks/download_cli_mac.yml
pritidesai: It was not done before and hence following the same workflow. We can validate them if needed since now CLI is part of separate repo. Thoughts?

ansible/roles/cli/tasks/download_cli_linux.yml
pritidesai: Yup would love to parameterize, db script as in this [deploy.yml](https://github.com/openwhisk/openwhisk/blob/master/ansible/roles/couchdb/tasks/deploy.yml)?

rabbah:
> @rabbah windows version of cli ...  wasn't part of edge router

really? i don't think that's true - is it because when you compile and deploy locally you only get one CLI (for your environment) - I suspect so. You have to opt into compiling all three versions.

pritidesai:
No no, I was referring to the code in [download_cli.yml](https://github.com/openwhisk/openwhisk/blob/master/ansible/roles/cli/tasks/download_cli.yml) which doesn't address windows version.

---

# Download the cli to openwhisk_home

```
- name: "download cli (linux) to openwhisk home at {{ openwhisk_home }}"
  local_action: >
    get_url
    url="https://{{ groups['edge'] | first }}/cli/go/download/linux/amd64/wsk"
    ...
  when: "'environments/mac' not in inventory_dir"

- name: "download cli (mac) to openwhisk home at {{ openwhisk_home }}"
  local_action: >
    get_url
    url="https://{{ groups['edge'] | first }}/cli/go/download/mac/amd64/wsk"
    ...
  when: "'environments/mac' in inventory_dir"
```

houshengbo:
@rabbah @pritidesai The tricky thing is that we have well documented and supported the deployment of openwhisk under ubuntu and mac. Our ansible script has both of them covered. However, we have no support of openwhisk installation under windows. When ansible is deploying openwhisk under either mac or ubuntu, the CLI is built based on the OS openwhisk is running on, so either we build CLI locally for mac or ubuntu, which means there is sort no chance to build the windows CLI. This is basically how the current ansible for cli build works.

Anyhow, the deployment of openwhisk can be under mac or ubuntu, but the cli can be installed under any system, when they are separate. However, our deployment of openwhisk put the openwhisk and cli on the same machine. That is why we do not have the chance to reach windows cli.

One approach to improve this can be the host machine, where cli will be installed, checks the OS, and pull the cli binary accordingly.

rabbah:
You can build the cli for hosts regardless if the host environment. @dubeejw can help.

houshengbo:
We are able to build the CLIs for all the operating systems, and they are publish on this page https://github.com/openwhisk/openwhisk-cli/releases. What this PR needs to do is to pick up the one, that fits the hosting OS.

rabbah:
OK - we're communication across purposes. Let me try with a blank slate.

When a developer is building and deploying openwhisk locally, by _default_ and to cut down the build time of the CLI, we build exactly 1 version of the CLI. This version is bundled with the edge router.

A developer change override this behavior and build all 3 binaries of the CLI regardless of the host environment. In this case, the edge router should bundle all the binaries.

In essence, since the building of the binary is now _not_ part of this core repo, and since the cost is therefore roughly free, the edge router should bundle _all_ the binaries by default.

In the CLI repo, this is where I would allow a developer to build only 1 binary as opposed to all of the versions.

So what I would like to see are these features:
1. by default, the edge router bundles into the edge all available versions of the CLI binary from a well defined source.
2. with an optional override, the edge router bundles from a local source the binaries that may be available.

The second feature is to allow a developer to build the CLI locally and bundle it with the edge router, and hence decouples the bundling, optionally, from the source we publish.

houshengbo:
@pritidesai 
Hi Priti. Based on the comments above, we have two features to be implemented in this PR. 

The first one is to simply pull all the binaries available for mac, ubuntu and windows on the releases page.

The second one is to build openwhisk-cli locally. If we run the command "./gradlew distDocker -PcrossCompileCLI=true", all the binaries will be built. If we run "./gradlew distDocker", only one binary will be built to meet the current edge host. We should have a flag to specify the option.

houshengbo:
@pritidesai @rabbah 
I submitted a new PR to resolve this issue:
https://github.com/openwhisk/openwhisk/pull/2190.

PR #2134, opened 2017-04-12
Set default db for activations to activations-db

cbickel:
This PR changes the default setting of all environments to use the activations-db for activations, instead of the whisks-db.
It also requires the flag `db_split_actions_and_activations` to be set. It still can be false, but if it is false, you should care about the migration of your activations to the activations-db.

If you don't matter throwing away all your data, just execute the ansible playbook `wipe.yml`, after `db_split_actions_and_activations` is set to true.
If you want to keep all your data, follow the instructions on the following page:
https://gist.github.com/cbickel/37e651965781b27de245eac0ce831a53

PG1#1437


cbickel:
PG1#1452 is green.

PR #2137, opened 2017-04-13
Make action container log reading more robust

sven-lange-last:
This PR fixes a small defect in action container log file reading in the invoker and addresses a premature file end when reading the log file.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
@@ -158,7 +158,7 @@ trait ContainerUtils {
      * Reads the contents of the file at the given position.
      * It is assumed that the contents does exist and that region is not changing concurrently.
      */
-    def getDockerLogContent(containerHash: ContainerHash, start: Long, end: Long, mounted: Boolean)(implicit transid: TransactionId): Array[Byte] = {
+    def getDockerLogContent(containerHash: ContainerHash, start: Long, end: Long, mounted: Boolean)(implicit transid: TransactionId): java.nio.ByteBuffer = {
         var fis: java.io.FileInputStream = null
         try {
             val file = getDockerLogFile(containerHash, mounted)
@@ -168,14 +168,19 @@ trait ContainerUtils {
             val buffer = java.nio.ByteBuffer.allocate(remain)
             while (remain > 0) {
                 val read = channel.read(buffer)
-                if (read > 0)
-                    remain = read - read.toInt
+                if (read >= 0) {
+                    remain -= read
+                } else {
+                    // read < 0 means premature end of file => terminate loop
+                    remain = 0
+                }
             }
-            buffer.array
+            buffer
         } catch {
             case e: Exception =>
                 logging.error(this, s"getDockerLogContent failed on ${containerHash.hash}: ${e.getClass}: ${e.getMessage}")
-                Array()
+                // Return an empty buffer
+                java.nio.ByteBuffer.wrap(Array())
         } finally {
             if (fis != null) fis.close()
         }
--- a/core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
@@ -17,7 +17,8 @@
 
 package whisk.core.container
 
-import java.io.{ File, FileNotFoundException }
+import java.io.{ File, FileNotFoundException, FileInputStream }
+import java.nio.ByteBuffer
 
 import scala.language.postfixOps
 import scala.util.Try
@@ -158,14 +159,14 @@ trait ContainerUtils {
      * Reads the contents of the file at the given position.
      * It is assumed that the contents does exist and that region is not changing concurrently.
      */
-    def getDockerLogContent(containerHash: ContainerHash, start: Long, end: Long, mounted: Boolean)(implicit transid: TransactionId): java.nio.ByteBuffer = {
-        var fis: java.io.FileInputStream = null
+    def getDockerLogContent(containerHash: ContainerHash, start: Long, end: Long, mounted: Boolean)(implicit transid: TransactionId): ByteBuffer = {
+        var fis: FileInputStream = null
         try {
             val file = getDockerLogFile(containerHash, mounted)
-            fis = new java.io.FileInputStream(file)
+            fis = new FileInputStream(file)
             val channel = fis.getChannel().position(start)
             var remain = (end - start).toInt
-            val buffer = java.nio.ByteBuffer.allocate(remain)
+            val buffer = ByteBuffer.allocate(remain)
             while (remain > 0) {
                 val read = channel.read(buffer)
                 if (read >= 0) {
@@ -180,7 +181,7 @@ trait ContainerUtils {
             case e: Exception =>
                 logging.error(this, s"getDockerLogContent failed on ${containerHash.hash}: ${e.getClass}: ${e.getMessage}")
                 // Return an empty buffer
-                java.nio.ByteBuffer.wrap(Array())
+                ByteBuffer.wrap(Array())
         } finally {
             if (fis != null) fis.close()
         }
@@ -239,7 +240,7 @@ trait ContainerUtils {
     /**
      * Gets the filename of the docker logs of other containers that is mapped back into the invoker.
      */
-    private def getDockerLogFile(containerId: ContainerHash, mounted: Boolean) = {
+    protected def getDockerLogFile(containerId: ContainerHash, mounted: Boolean) = {
         new java.io.File(s"""${dockerContainerDir(mounted, containerId)}/${containerId.hash}-json.log""").getCanonicalFile()
     }
 
--- a/tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
+++ b/tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
@@ -1,0 +1,147 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.container.test
+
+import org.junit.runner.RunWith
+import org.scalatest.fixture.FlatSpec
+import org.scalatest.junit.JUnitRunner
+import java.io.File
+
+import common.StreamLogging
+
+import whisk.common.TransactionId
+import whisk.core.container.ContainerUtils
+import whisk.common.Logging
+import whisk.core.container.ContainerHash
+import org.scalatest.BeforeAndAfter
+import java.io.FileWriter
+import java.nio.charset.StandardCharsets
+import org.scalatest.Matchers
+import org.scalatest.concurrent.TimeLimitedTests
+import org.scalatest.time.SpanSugar._
+
+/**
+ * Unit tests for ContainerPool and, by association, Container and WhiskContainer.
+ */
+@RunWith(classOf[JUnitRunner])
+class ContainerUtilsTests extends FlatSpec
+    with BeforeAndAfter
+    with Matchers
+    with StreamLogging
+    with TimeLimitedTests {
+
+    val timeLimit = 100 millis
+
+    before {
+        stream.reset()
+    }
+
+    implicit val transid = TransactionId.testing
+
+    behavior of "getDockerLogContent"
+
+    case class FixtureParam(file: File, writer: FileWriter, cu: ContainerUtilsTester)
+
+    def withFixture(test: OneArgTest) = {
+        val file = File.createTempFile(this.getClass.getName, test.name.replaceAll("[^a-zA-Z0-9.-]", "_"))
+        val writer = new FileWriter(file)
+        val cu = new ContainerUtilsTester(file)
+
+        val fixture = FixtureParam(file, writer, cu)
+
+        try {
+            withFixture(test.toNoArgTest(fixture))
+        } finally {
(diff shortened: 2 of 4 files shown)


core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
```diff
@@ -154,7 +154,7 @@ trait ContainerUtils {
      * Reads the contents of the file at the given position.
      * It is assumed that the contents does exist and that region is not changing concurrently.
      */
-    def getDockerLogContent(containerHash: ContainerHash, start: Long, end: Long, mounted: Boolean)(implicit transid: TransactionId): Array[Byte] = {
+    def getDockerLogContent(containerHash: ContainerHash, start: Long, end: Long, mounted: Boolean)(implicit transid: TransactionId): java.nio.ByteBuffer = {
```
markusthoemmes: `import java.nio.ByteBuffer` and remove the qualifications please.

core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
```diff
@@ -164,14 +164,19 @@ trait ContainerUtils {
...
             while (remain > 0) {
                 val read = channel.read(buffer)
-                if (read > 0)
-                    remain = read - read.toInt
+                if (read >= 0) {
+                    remain -= read
+                } else {
+                    // read < 0 means premature end of file => terminate loop
+                    remain = 0
+                }
```
markusthoemmes: Per personal discussion: We'll need a test that checks for this condition (and shows "wrong" behavior before the change)

core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
sven-lange-last: Changed as advised.

core/invoker/src/main/scala/whisk/core/container/ContainerUtils.scala
rabbah: Hmm. I'm not seeing my comments here. I had asked for the same. I also explained there is a retry at the outer level; let me check where my comment went before rewriting.

rabbah: I don't know what happened to my comment. So here goes again: This log reader runs before tearing down a container to save the logs from an action container to a file for post inspection. Should we just nuke this code? There are also no tests for this code in general. This is not the log reader we use for activation log extraction.

rabbah: I posted and deleted two comments this morning as I try to remember what my concern was previously when I reviewed this PR. I can neither find that comment or remember exactly what it was but there are two high level bits. One, same as Markus' - test for the corner case? Second, looking back at the code, `read()` should not return -1 since we are reading a pre-determined number of bytes `pool.getDockerLogContent(con.containerId, con.lastLogSize, size, runningInContainer)`.

sven-lange-last: We are only talking about edge cases here.

Current line 168 is wrong and would set `remain`to zero whenever `channel.read()` was able to read something - in other words: almost always. This bug actually delegates all problems with reading to the caller of `getDockerLogContent()` eating up the retry count.

Two different cases are possible:

1. It _could_ happen that `channel.read()` is not able to fill the whole buffer in the first attempt, i.e. `0 < read < remain`. The bug in line 168 causes the loop to terminate. The calling function cannot find the sentinel marker, will drop the read buffer and will retry - invoking `getDockerLogContent()` again. This condition may arise if the log file is large and a lot of I/O is going on.

2. It _could_ happen that Docker truncates the log files after the its size has been determined by the caller of `getDockerLogContent()`. In case of full truncation, the loop is repeated until `channel.read()` returns a positive value - potentially blocking forever.

Both cases are expected to rarely happen in the wild - but if they happen, they can degrade performance considerably. And the fix is easy.

My recent commit `1e1494b847f4e0548d66db9d2108b3b514b5e082` adds unit tests.

rabbah: read - read 😳Thanks for pointing that out.

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
```diff
@@ -0,0 +1,147 @@
...
+        fixture.writer.write(content)
+        fixture.writer.flush()
+        fixture.file.length
+    }
+
+    it should "tolerate an empty log file" in { fixture =>
+        val logText = ""
+        val size = writeLogFile(fixture, logText)
+
+        val buffer = fixture.cu.getDockerLogContent(containerHash = ContainerHash.fromString("0123"), start = 0, end = size, mounted = false)
```
markusthoemmes: externalize the default ContainerHash. We don't care about it anyways.

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
```diff
@@ -0,0 +1,147 @@
...
+        val logContent = new String(buffer.array, buffer.arrayOffset, buffer.position, StandardCharsets.UTF_8)
+
+        logContent shouldBe logText
+        stream.size() shouldBe 0
```
markusthoemmes: `stream should have size 0`

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
```diff
@@ -0,0 +1,147 @@
...
+
+    it should "read a log file portion" in { fixture =>
+        val count = 4
+        val logText = (1 to count).map(i => s"Line ${i}\n").mkString("")
+        val size = writeLogFile(fixture, logText)
+        val quarter = count / 4
+        val qlen = logText.length / 4 // length of a quarter
+        val from = qlen // start reading from second quarter
+        val to = from + 2 * qlen // read second and third quarter
+        val expectedText = logText.substring(from, to)
```
markusthoemmes: Why so complex? Why not just:

```
val logText = "hello scala, this is quite a nice thing we got running here, eh?"
val from  = 4
val to = 10

...

logContent shouldBe logText.substring(from, to)
```

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
```diff
@@ -0,0 +1,147 @@
...
+
+        val fixture = FixtureParam(file, writer, cu)
+
+        try {
+            withFixture(test.toNoArgTest(fixture))
+        } finally {
+            writer.close()
+            file.delete()
+        }
+    }
```
markusthoemmes: This looks a bit overcomplicated + magic happens (methodnames need to match). Why not take a simpler approach like:

```
def createFile(content: String): File = //create your file and write the string to it
```

and then for each test you just initialize the class under test with the respective file:

```
it should .... in {
    val file = createFile("text")
    val utils = new TestContainerUtils(file)
}
```

I'd say that's more readable and involves far less magic. thoughts?

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
```diff
@@ -0,0 +1,147 @@
...
+        fixture.cu.logFile = new File("/nonsense")
+
+        val buffer = fixture.cu.getDockerLogContent(containerHash = ContainerHash.fromString("0123"), start = 0, end = 1, mounted = false)
+
+        buffer.capacity() shouldBe 0
+        logLines.head should include("getDockerLogContent failed")
+    }
+}
+
+class ContainerUtilsTester(var logFile: File)(implicit val logging: Logging) extends ContainerUtils {
```
markusthoemmes: If you apply the changes suggested above, this can become a `val` or even easier: a private member.

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
```diff
@@ -0,0 +1,147 @@
...
+
+        val buffer = fixture.cu.getDockerLogContent(containerHash = ContainerHash.fromString("0123"), start = from, end = to, mounted = false)
+        val logContent = new String(buffer.array, buffer.arrayOffset, buffer.position, StandardCharsets.UTF_8)
+
+        logContent shouldBe expectedText
+        stream.size() shouldBe 0
+    }
+
+    it should "tolerate premature end of log file" in { fixture =>
+        val logText = (1 to 2).map(i => s"Line ${i}\n").mkString("")
```
markusthoemmes: If you want to write lines, why not do

```
(1 to 2).map(i => s"Line ${i}").mkString("\n")
```

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Unit tests for ContainerPool and, by association, Container and WhiskContainer.
+ */
+@RunWith(classOf[JUnitRunner])
+class ContainerUtilsTests extends FlatSpec
+    with BeforeAndAfter
+    with Matchers
+    with StreamLogging
+    with TimeLimitedTests {
```
markusthoemmes: Why is this needed?

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
```diff
@@ -0,0 +1,147 @@
...
+
+    val timeLimit = 100 millis
```
markusthoemmes: Unused value.

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
sven-lange-last: Time limitation of tests is not strictly needed at the moment. The tests perform real I/O instead of mocked I/O so that a time limitation may prevent too long runtimes when running this test suite as part of a full build.

In addition, my initial assumption was to prepare unit tests for the original code. With the bug fixed by this PR, at least one unit test would have looped such that a time limitation had been necessary to prevent excessive test runtime.

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
sven-lange-last: See http://doc.scalatest.org/1.8/org/scalatest/concurrent/TimeLimitedTests.html.

`timeLimit` overrides a `def` from trait `org.scalatest.concurrent.TimeLimitedTests` activating the specified time limit for each single test.

sven-lange-last: I added a comment to this line to point this out.

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
sven-lange-last: Just discussed with @markusthoemmes - removing the time limit because it isn't required at the moment.

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
sven-lange-last: I was using fixtures because they provide a central place for setting and - particularly - cleaning up. The fixture makes sure that temporary files are closed and removed after each test.

All this is also possible with the approach you suggest above - but requires a `try ... finally`bracket around each single test.

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
sven-lange-last: Makes sense.

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
sven-lange-last: Good proposal - thanks. Changed.

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
sven-lange-last: Your proposal does only add `\n` between the strings included in the range. In other words: with your proposal, the last line won't be followed by a `\n`.

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
sven-lange-last: I understand... When keeping the fixture approach, test `provide an empty result on failure` will be invoked with a fixture that isn't really used...

At the same time, said test overrides the `logFile` to a file that doesn't exist to provoke a failure.

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
sven-lange-last: Done

tests/src/test/scala/whisk/core/container/test/ContainerUtilsTests.scala
rabbah: There's a mkstring variant with separators at start, middle and end.

sven-lange-last:
PG2 - 1488 running

rabbah:
@markusthoemmes rebased this to master; pg4/552 running since I couldn't verify the above noted pg.

rabbah:
pg 4/552 🔵

andrewnet:
Hi, can anyone merge this fix into master? I am still facing the problem and would like to give it a try once this is pulled in. I would do it if I have write access.

rabbah:
We've actually switched to a new implementation of the invoker which doesn't have this bug. Are you seeing the issue still in bluemix?

rabbah:
This is the code that's now active https://github.com/apache/incubator-openwhisk/blob/master/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala. The code affected by this pr is effectively deadcode we plan on removing.

andrewnet:
Yes, I am still seeing logs got truncated on YP-QA environment.

andrewnet:
I have opened a defect https://github.com/apache/incubator-openwhisk/issues/2548 to report the issue separately since this defect is out dated as @rabbah.

markusthoemmes:
@sven-lange-last want to port this to the new folderstructure?

rabbah:
@sven-lange-last @markusthoemmes is this still relevant or now out of date and should be closed?

markusthoemmes:
I think we can safely close this, the code is gone anyway.

Issue #957, opened 2016-07-27
show "start" value in activation in a human-readable format


mbehrendt:

RSulzmann:
`wsk activation get <activationId> --summary` already shows the end time in a human readable format:
```
$ wsk activation get 7a3c8cf312f04acdba395a74208ed574 --summary
activation result for /lime@us.ibm.com/robi01 (success at 2017-04-11 14:42:36 +0200 CEST)
{
    "name": "hallo"
}
```

`wsk activation get <activationId>` 
can show the start time the same way in the first line :
```
$ wsk activation get 7a3c8cf312f04acdba395a74208ed574
ok: got activation 7a3c8cf312f04acdba395a74208ed574 (started at 2017-04-11 14:42:36 +0200 CEST)
{
    "namespace": "lime@us.ibm.com",
    "name": "robi01",
    "version": "0.0.1",
    "subject": "lime@us.ibm.com",
    "activationId": "7a3c8cf312f04acdba395a74208ed574",
    "start": 1491914556836,
    "end": 1491914556891,
    "duration": 55,
    "response": {
        "status": "success",
        "statusCode": 0,
        "success": true,
        "result": {
            "name": "hallo"
        }
    },
    "logs": [],
    "annotations": [
        {
            "key": "limits",
            "value": {
                "logs": 10,
                "memory": 256,
                "timeout": 60000
            }
        },
        {
            "key": "path",
            "value": "lime@us.ibm.com/robi01"
        }
    ],
    "publish": false
}
```

mdeuser:
Perhaps adding some of this summary information to the `activation list` output would be useful (i.e. response status, execution start/duration)

RSulzmann:
`$ wsk activation list` 
can also show a human readable start time for the listed activations :
```
wskl activation list
activations
2017-04-12 14:49:09 +0200 CEST 84fff57d8fc44bbb917b74ac1bcbe56f action-invoke-nodejsdefault-1492001349145
2017-04-12 14:49:09 +0200 CEST 7c54d0a83968404fbb4e87d7f18585ac action-invoke-nodejsdefault-1492001349145
2017-04-12 14:48:09 +0200 CEST 55a8beb21ee64489a127a94b8be3e32f action-invoke-nodejsdefault-1492001289125
2017-04-12 14:48:09 +0200 CEST fcd08d2dab0744ad86a67dc1af33e07f action-invoke-nodejsdefault-1492001289125
[...]

Issue #20, opened 2016-02-25
"wsk activation list" should show more information about activations


l2fprod:
If I'm looking at a recent activation to debug it and I don't know its ID, it would be useful to have its start/end time, its status (success, running, failed).

Currently other than confirming there are activations, it is not very useful:

```
$ wsk activation list
activations
123fb4243552491b94b6fcf6a94be745             vision-analysis                         
21526c4dd7564090a9095941fa1d546b             vision-analysis                         
0fa2dd0bb55f4df3ba68a571da1a6d34             vision-analysis                         
3b33904f79f448beb0c913856363a088             vision-analysis                         
e2e22077ad3945b48b612436fe0bb334             vision-analysis                         
48581ba45fac4119a5970250e3f0b0d7             vision-analysis                         
dec75b9e241942e6a9ddd0fc9b8dd68b             vision-analysis                         
3a0e3cbf45734ce1b13220414d6c10af             vision-analysis
```

fxulusoy:
I have discussed the issue with @rabbah. He proposed more general solution.

activation entity records can be divided into two parts: small constant size and unbounded size. logs, result object are examples of the latter. The other fields of an activation are finite and bounded
Default activation list should get and show a bounded part of each activation. activation list with a flag (include_docs) should get everything that belongs to activation.

I will do the changes in this direction. Any comments?

l2fprod:
@fxulusoy: makes sense. eager to see more info on activation list. thanks!

csantanapr:
@mdeuser I see go-cli has a `--full` flag
Should we close this issue?

@l2fprod you can use `--since` `--upto` and `--full` to get a list of the activations and be able to find the one you are looking within the time frame you scoped.


```
$ wsk activation list -h
list activations
Usage:
  wsk activation list [NAMESPACE] [flags]

Flags:
  -f, --full          include full activation description
  -l, --limit LIMIT   only return LIMIT number of activations from the collection (default 30)
      --since SINCE   return activations with timestamps later than SINCE; measured in milliseconds since Th, 01, Jan 1970
  -s, --skip SKIP     exclude the first SKIP number of activations from the result
      --upto UPTO     return activations with timestamps earlier than UPTO; measured in milliseconds since Th, 01, Jan 1970

```

rabbah:
Fred's request is to show more information in the `list` as in 

```
$ wsk activation list
activations
[...]

PR #2138, opened 2017-04-13
i957 adding fields to activation list

RSulzmann:
Related to #20 and #957.

added timestamp to `wsk activation get <activationID>`
added new fields to `wsk activation list`


RSulzmann:
Because the used database view is changed, this has impact on deployment time.
To be checked before merge:
- change the code to use a new view instead of changing the current one.

ansible/files/whisks_design_document_for_activations_db.json
```diff
@@ -5,7 +5,7 @@
...
+      "map": "function (doc) {\n  var PATHSEP = \"/\";\n\n  try {\n    var date = new Date(doc.start || doc.updated);\n    emit([doc.namespace, date.getTime(), doc.name], {namespace: doc.namespace, name: doc.name, version: doc.version, publish: doc.publish, annotations: doc.annotations, activationId: doc.activationId, start: doc.start, duration: doc.duration, response: {statusCode: doc.response.statusCode}});\n    emit([doc.namespace+PATHSEP+doc.name, date.getTime()], {namespace: doc.namespace, name: doc.name, version: doc.version, publish: doc.publish, annotations: doc.annotations, activationId: doc.activationId, start: doc.start, duration: doc.duration, response: {statusCode: doc.response.statusCode}});\n  } catch (e) {}\n}\n"
```
rabbah: you should not use `getTime` here - this will be non-deterministic and affect the results of the view is what will appear to be surprising ways.

What is the goal? Isn't start time already in the record?

rabbah:
this is exposing virtually all the information except logs - since we want to split logs (and results) out anyway into attachment, I would approach this differently.

rabbah:
Closing as abandoned. Reopen if you disagree.

PR #2142, opened 2017-04-13
API GW V2 swagger generation change (remove auth, content headers)

mdeuser:
API GW swagger specification/implementation change for V2
- No longer use the `set-variable` object with the V2 API GW


mdeuser:
@rabbah - would you mind review/merge these changes.  Travis passes and PG passes the associated tests (minus a persistent test failure at the moment)

PR #2148, opened 2017-04-15
Remove 'OK' from HTTP response when no 'body' is present.

### LINKED ISSUE #2132: Web action with http extension defining "OK" when body is missing from response
Create a web action which doesn't return a "body" top level property and curl it as a web action with .http as the extension. It responds with "OK" which comes from the lower spray layer that binds a message to the status code 200.

rabbah:
Fixes #2132.

### DIFF
--- a/docs/webactions.md
+++ b/docs/webactions.md
@@ -88,7 +88,7 @@ An OpenWhisk action that is not a web action requires both authentication and mu
 
 1. `headers`: a JSON object where the keys are header-names and the values are string values for those headers (default is no headers).
 2. `statusCode`: a valid HTTP status code (default is 200 OK).
-3. `body`: a string which is either plain text or a base64 encoded string (for binary data).
+3. `body`: a string which is either plain text or a base64 encoded string for binary data (default is empty response).
 
 The controller will pass along the action-specified headers, if any, to the HTTP client when terminating the request/response. Similarly the controller will respond with the given status code when present. Lastly, the body is passed along as the body of the response. Unless a `content-type header` is declared in the action result’s `headers`, the body is passed along as is if it’s a string (or results in an error otherwise). When the `content-type` is defined, the controller will determine if the response is binary data or plain text and decode the string using a base64 decoder as needed. Should the body fail to decoded correctly, an error is returned to the caller.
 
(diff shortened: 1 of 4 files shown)


csantanapr:
PG?

rabbah:
Not needed but will launch if you like.

dubeejw:
@rabbah, there is some discussion around not returning an HTTP code of 200 when the response is empty.

rabbah:
I reject that as non-nonsensical but willing to debate it if someone wants to take the opposing view.
 
Per https://github.com/openwhisk/openwhisk/issues/2132#issuecomment-293562946, the implementation follows the defaults and I updated the missing default in the docs for body.

rabbah:
pg1/1411 approved.

docs/webactions.md
```diff
@@ -88,7 +88,7 @@ An OpenWhisk action that is not a web action requires both authentication and mu
 
 1. `headers`: a JSON object where the keys are header-names and the values are string values for those headers (default is no headers).
 2. `statusCode`: a valid HTTP status code (default is 200 OK).
-3. `body`: a string which is either plain text or a base64 encoded string (for binary data).
+3. `body`: a string which is either plain text or a base64 encoded string for binary data (default is empty response).
```
akrabat: Is this true?

if I `return ["statusCode": 405]`, I'll get a `text/plain` response with the body set to `A request was made of a resource using a request method not supported by that resource;`.

i.e. default is an empty response only in the case of a `200` status code.

rabbah: With this patch? I will add an explicit test for this but a quick sniff test:

```
> curl -vk https://guest.openwhisk/default/t
*   Trying 192.168.99.100...
* Connected to guest.openwhisk (192.168.99.100) port 443 (#0)
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_RC4_128_SHA
* Server certificate: *.openwhisk
> GET /default/t HTTP/1.1
> Host: guest.openwhisk
> User-Agent: curl/7.43.0
> Accept: */*
> 
< HTTP/1.1 405 Method Not Allowed
< Server: nginx/1.11.13
< Date: Mon, 17 Apr 2017 23:54:07 GMT
< Content-Length: 0
< Connection: keep-alive
< 
* Connection #0 to host guest.openwhisk left intact
```

for this action:
```
function main() { return { statusCode: 405 } }
```

akrabat: You are correct - I was on the wrong branch when I tested…

PR #2151, opened 2017-04-18
Implement a containerpool controlling container proxies

### LINKED ISSUE #2090: Implement a reactive ContainerPool
This aims to implement a ContainerPool which is completely independent of the underlying container driver to facilitate testing and extendability. The pool shall have clear seperation of concerns and its only concern is scheduling work to the Containers in its pool.

markusthoemmes:
This pool fulfills the other half of the ContainerProxy contract. Only
one job (either Start or Run) is sent to a child-actor at any given
time. The pool then waits for a response of that container, indicating
the container is done with the job. Only then will the pool send another
request to that container.

Upon actor creation, the pool will start to prewarm containers according
to the provided prewarmConfig, iff set. Those containers will **not** be
part of the poolsize calculation, which is capped by the poolSize parameter.
Prewarm containers are only used, if they have matching arguments
(kind, memory) and there is space in the pool.

Closes #2090

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
@@ -1,0 +1,219 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.containerpool
+
+import scala.collection.mutable
+
+import akka.actor.Actor
+import akka.actor.ActorRef
+import akka.actor.ActorRefFactory
+import akka.actor.Props
+import whisk.common.AkkaLogging
+import whisk.core.dispatcher.ActivationFeed.ContainerReleased
+import whisk.core.entity.ByteSize
+import whisk.core.entity.CodeExec
+import whisk.core.entity.EntityName
+import whisk.core.entity.ExecutableWhiskAction
+import whisk.core.entity.size._
+
+sealed trait WorkerState
+case object Busy extends WorkerState
+case object Free extends WorkerState
+
+case class WorkerData(data: ContainerData, state: WorkerState)
+
+/**
+ * A pool managing containers to run actions on.
+ *
+ * This pool fulfills the other half of the ContainerProxy contract. Only
+ * one job (either Start or Run) is sent to a child-actor at any given
+ * time. The pool then waits for a response of that container, indicating
+ * the container is done with the job. Only then will the pool send another
+ * request to that container.
+ *
+ * Upon actor creation, the pool will start to prewarm containers according
+ * to the provided prewarmConfig, iff set. Those containers will **not** be
+ * part of the poolsize calculation, which is capped by the poolSize parameter.
+ * Prewarm containers are only used, if they have matching arguments
+ * (kind, memory) and there is space in the pool.
+ *
+ * @param childFactory method to create new containers
+ * @param poolSize maximum size of containers allowed in the pool
+ * @param feed actor to request more work from
+ * @param prewarmConfig optional settings for container prewarming
+ */
+class ContainerPool(
+    childFactory: ActorRefFactory => ActorRef,
+    poolSize: Int,
+    feed: ActorRef,
+    prewarmConfig: Option[PrewarmingConfig] = None) extends Actor {
+    val logging = new AkkaLogging(context.system.log)
+
+    val pool = new mutable.HashMap[ActorRef, WorkerData]
+    val prewarmedPool = new mutable.HashMap[ActorRef, WorkerData]
+
+    prewarmConfig.foreach { config =>
+        logging.info(this, s"pre-warming ${config.count} ${config.exec.kind} containers")
+        (1 to config.count).foreach { _ =>
+            prewarmContainer()
+        }
+    }
+
+    def receive: Receive = {
+        // A job to run on a container
+        case r: Run =>
+            // Schedule a job to a warm container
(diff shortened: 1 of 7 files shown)


bbrowning:
This looks like it adds ContainerPool and Container abstractions but doesn't yet use them in anything but tests? It's laying the groundwork for creating a DockerContainer and other Container implementations? I'm interested in a KubernetesContainer implementation and want to make sure I'm following the proposed changes correctly.

rabbah:
Indeed - this is part of the plan to incrementally introduce an entirely new reactive container pool and load balancing algorithm.

markusthoemmes:
Rebased to `master`.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -0,0 +1,219 @@
...
+    def prewarmContainer() = prewarmConfig.foreach(config => childFactory(context) ! Start(config.exec, config.memoryLimit))
+
+    /**
+     * Takes a prewarm container out of the prewarmed pool
+     * iff a container with a matching kind is found.
+     *
+     * @param kind the kind you want to invoke
+     * @return the container iff found
+     */
+    def takePrewarmContainer(action: ExecutableWhiskAction) = prewarmConfig.flatMap { config =>
```
cbickel: why do we need the config (map of prewarmConfig) here? Isn't the prewarmedPool empty, if there are no prewarmed containers?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -0,0 +1,383 @@
...
+    startWith(Uninitialized, NoData())
+
+    when(Uninitialized) {
+        // pre warm a container
+        case Event(job: Start, _) =>
+            factory(
+                TransactionId.invokerWarmup,
+                ContainerProxy.containerName("prewarm", job.exec.kind),
+                job.exec,
+                job.memoryLimit)
```
sven-lange-last: The `factory()` returns a `Future` which means it needs an `ExecutionContext`. Apparently, the `factory` passed in the class constructor has no `ExecutionContext` in its parameter list. Which `ExecutionContext` will it use?

I would expect that it uses the same `ExecutionContext` as the `ContainerProxy` which is an actor - but `ExecutionContext` is wired in `factory`.

rabbah: the closure will have specified the EC. Are you stating that it should be the same EC as the proxy?
in any case all the ECs should be the same in the invoker.

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
```diff
@@ -0,0 +1,384 @@
...
+        val (containers, factory) = testContainers(2)
+        val feed = TestProbe()
+        val pool = system.actorOf(ContainerPool.props(factory, 2, feed.ref))
+
+        pool ! runMessage
+        containers(0).expectMsg(runMessage)
+        containers(0).send(pool, NeedWork(warmedData()))
+
+        pool ! runMessage
+        containers(0).expectMsg(runMessage)
```
rabbah: can you also assert no messages for `containers(1)`?

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
```diff
@@ -0,0 +1,384 @@
...
+        // Note that the container doesn't respond, thus it's not free to take work
+        pool ! runMessage
+        containers(1).expectMsg(runMessage)
+    }
+
+    it should "remove a container to make space in the pool if it is already full" in within(timeout) {
+        val (containers, factory) = testContainers(2)
+        val feed = TestProbe()
+
+        // a pool with only 1 slot
```
rabbah: might be useful to repeat this test with 1 test container as well.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -0,0 +1,219 @@
...
+     * Determines which namespace consumes most resources in the current pool and
+     * takes away one of their containers iff the namespace placing the new job is
+     * not already the most consuming one.
+     *
+     * @param namespace the namespace that wants to get a container
+     * @param pool a map of all containers in the pool
+     * @return a container to be removed iff found
+     */
+    def remove[A](namespace: EntityName, pool: Map[A, WorkerData]): Option[A] = {
+        val grouped = pool.collect {
```
markusthoemmes: I agree, this is specifically scoped to the rather small pools we're having now. When we expand the idea we'll have to revisit those.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -0,0 +1,219 @@
...
+    prewarmConfig: Option[PrewarmingConfig] = None) extends Actor {
+    val logging = new AkkaLogging(context.system.log)
+
+    val pool = new mutable.HashMap[ActorRef, WorkerData]
+    val prewarmedPool = new mutable.HashMap[ActorRef, WorkerData]
+
+    prewarmConfig.foreach { config =>
+        logging.info(this, s"pre-warming ${config.count} ${config.exec.kind} containers")
+        (1 to config.count).foreach { _ =>
+            prewarmContainer()
```
markusthoemmes: The `foreach` in `prewarmContainer` is for the Option. Christian already commented on some weirdness here, i'll resolve.

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
markusthoemmes: I can, but that'll have the test run until the timeout in every case. Maybe I can mitigate and only wait for 200ms for noMessages.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
markusthoemmes: That's right, i need access to the "empty" exec of the config though, as constructing such an Exec is as verbose as just doing a flatMap on the config here anyway.

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
markusthoemmes: Not sure why? Can you elaborate?

PR #2153, opened 2017-04-19
relax the ssl certificate validation for test cases

jeremiaswerner:
This change relaxes the https certificate validation in test cases and will unblock integration testing of API gateway. Today we only accept certificates that are added to the keystore. An alternative solution is to add the certificate of the API Gateway to that keystore. However, to avoid the same issue to happen in the future and since this is only for testing purposes I'am proposing to relax the ssl handling.

### DIFF
--- a/tests/src/test/scala/system/rest/RestUtil.scala
+++ b/tests/src/test/scala/system/rest/RestUtil.scala
@@ -37,7 +37,7 @@ trait RestUtil {
     // force RestAssured to allow all hosts in SSL certificates
     protected val sslconfig = {
         new RestAssuredConfig().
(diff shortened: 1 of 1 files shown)


jeremiaswerner:
PG2/1460

mdeuser:
I recommend only relaxing the API GW tests; I have a branch testing out a change to do just that.

domdom82:
@jeremiaswerner oops, sorry. didn't think I had any rights to update your branch. pls remove me as a collaborator.

jeremiaswerner:
this PR is resolved by another PR form mdeuser

PR #2154, opened 2017-04-19
Rework replicator to abort replication when it is still running on de…

cbickel:
…leting old snapshot

PG1#1433 is green

### DIFF
--- a/tests/src/test/scala/whisk/core/database/test/ReplicatorTests.scala
+++ b/tests/src/test/scala/whisk/core/database/test/ReplicatorTests.scala
@@ -104,6 +104,14 @@ class ReplicatorTests extends FlatSpec
         db
     }
 
+    /** Removes a document from the _replicator-database */
+    def removeReplicationDoc(id: String) = {
+        println(s"Removing replication doc: $id")
+        val response = replicatorClient.getDoc(id).futureValue
+        val rev = if (response.isRight) Some(response.right.get.fields("_rev").convertTo[String]) else None
+        rev.map(replicatorClient.deleteDoc(id, _).futureValue)
+    }
+
     /** Runs the replicator script to replicate databases */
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/whisk/core/database/test/ReplicatorTests.scala
```diff
@@ -104,6 +104,14 @@ class ReplicatorTests extends FlatSpec
         db
     }
 
+    /** Removes a document from the _replicator-database */
+    def removeReplicationDoc(id: String) = {
+        println(s"Removing replication doc: $id")
+        val response = replicatorClient.getDoc(id).futureValue
+        val rev = if (response.isRight) Some(response.right.get.fields("_rev").convertTo[String]) else None
```
markusthoemmes: Not that it would really matter that much, but for posterity:

```
val rev = response.right.toOption.map(_.fields("_rev").convertTo[String]
```

tools/db/replicateDbs.py
```diff
@@ -66,6 +68,12 @@ def extractTimestamp(dbName):
     def isExpired(timestamp):
         return now - args.expires > timestamp
 
+    # Delete all documents in the _replicator-database of old backups to avoid that they continue after they are deprecated
+    print("----- Delete backup-documents older than %d seconds -----" % args.expires)
+    for doc in filter(lambda doc: isBackupDb(doc.id) and isExpired(extractTimestamp(doc.id)), replicator.view('_all_docs', include_docs=True)):
+        print("deleting backup document: %s" % doc.id)
+        replicator.delete(doc.doc)
```
markusthoemmes: Shouldn't we merge this with the filter below? As in:

For each database we're going to delete I look up the document in the replicator db and remove it?

cbickel: As discussed yesterday: I did it in a seperate loop, to also delete old replication documents.

PR #2159, opened 2017-04-20
Make CleanupTest more robust

cbickel:


xin-cai:
LGTM, is there a green playground?

Issue #2086, opened 2017-03-30
Implement a new ContainerPool based on reactive patterns


markusthoemmes:
Implement an entirely new ContainerPool (and eventually Invoker) which is based entirely on an event- and message-driven architecture. It does not use any busy-loops nor manually tweaked Thread.sleeps to distribute its work. As a consequence, this pool does not need locking or synchronized access to any data-structures.

Containers are modeled as a finite-state-machine to reflect the lifecycle of a container in the code itself as close as possible.

The new pool can factor in all the bits and pieces we learned over the development of the current pool. It needs to have testability in mind as a first-class-citizen to facilitate Invoker simulation and therefore simplify playing around with different scheduling algorithms.

markusthoemmes:
All pull-requests relative to this epic are open and ready for review. For convenience, here's how they add up:

* #2161 (DockerClientWithFileAccess)
* #2107 (ContainerProxy)

Those two can be reviewed and merged independently. They should be merged before:

* #2151 (ContainerPool)
* #2177 (DockerContainer)

Those two again should be independently review-/and mergeable. After the first two have been merged, the latter should be rebased to incooperate any changes that might apply.

After those 4 requests are in, I'll assemble a final PR which will hook up the ContainerPool + DockerContainer with the invoker as seen in #2021. Note that #2021 is only the "big picture" PR. It does not need to be reviewed nor merged. It only lives there to keep continuous track of our testcoverage and that we didn't introduce regressions.

@rabbah @jeremiaswerner FYI

rabbah:
@markusthoemmes done?

markusthoemmes:
Indeed.

PR #2161, opened 2017-04-20
Factor out Docker config and log file handling to DockerClientFileAccess

sven-lange-last:
Prepare new container-pool implementation - see #2086.

Factor out all file-based operations from DockerClient such that the DockerClient can also work with a remote docker host. Two operations are supported:

- Obtain a container's IP address: Read from container's config file in the first place and fall back to using `docker inspect` if file reading doesn't work. The file reading variant is preferable because it's faster than `docker inspect`.
- Obtain a container's stdout / stderr output: Read from container's log output written by the JSON log driver. At the moment, there is no falling back to `docker logs`.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
@@ -1,0 +1,188 @@
+/*
+ * Copyright 2015-2016 IBM Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.containerpool.docker
+
+import java.io.File
+import java.io.FileInputStream
+import java.io.IOException
+import java.nio.ByteBuffer
+import java.nio.file.Paths
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.blocking
+
+import spray.json.DefaultJsonProtocol._
+import spray.json._
+import whisk.common.Logging
+import whisk.common.TransactionId
+
+class DockerClientWithFileAccess(
+    dockerHost: Option[String] = None,
+    containersDirectory: File = Paths.get("containers").toFile)(executionContext: ExecutionContext)(implicit log: Logging)
+    extends DockerClient(dockerHost)(executionContext)(log) with DockerApiWithFileAccess {
+
+    implicit private val ec = executionContext
+
+    /**
+     * Provides the home directory of the specified Docker container.
+     *
+     * Assumes that property "containersDirectory" holds the location of the
+     * home directory of all Docker containers. Default: directory "containers"
+     * in the current working directory.
+     *
+     * Does not verify that the returned directory actually exists.
+     *
+     * @param containerId Id of the desired Docker container
+     * @return canonical location of the container's home directory
+     */
+    protected def containerDirectory(containerId: ContainerId) =
+        new File(containersDirectory, containerId.asString).getCanonicalFile()
+
+    /**
+     * Provides the configuration file of the specified Docker container.
+     *
+     * Assumes that the file has the well-known location and name.
+     *
+     * Does not verify that the returned file actually exists.
+     *
+     * @param containerId Id of the desired Docker container
+     * @return canonical location of the container's configuration file
+     */
+    protected def containerConfigFile(containerId: ContainerId) = {
+        new File(containerDirectory(containerId), "config.v2.json").getCanonicalFile()
+    }
+
+    /**
+     * Provides the log file of the specified Docker container written by
+     * Docker's JSON log driver.
+     *
+     * Assumes that the file has the well-known location and name.
+     *
+     * Does not verify that the returned file actually exists.
+     *
+     * @param containerId Id of the desired Docker container
+     * @return canonical location of the container's log file
+     */
+    protected def containerLogFile(containerId: ContainerId) =
+        new File(containerDirectory(containerId), s"${containerId.asString}-json.log").getCanonicalFile()
+
+    /**
+     * Provides the contents of the specified Docker container's configuration
+     * file as JSON object.
+     *
+     * @param configFile the container's configuration file in JSON format
+     * @return contents of configuration file as JSON object
+     */
+    protected def configFileContents(configFile: File): Future[JsObject] = Future {
+        blocking { // Needed due to synchronous file operations
+            val source = scala.io.Source.fromFile(configFile)
+            val config = try source.mkString finally source.close()
+            config.parseJson.asJsObject
+        }
+    }
+
+    /**
+     * Extracts the IP of the container from the local config file of the docker daemon.
+     *
+     * A container may have more than one network. The container has an
+     * IP address in each of these networks such that the network name
+     * is needed.
+     *
+     * @param id the id of the container to get the IP address from
+     * @param network name of the network to get the IP address from
+     * @return the ip address of the container
+     */
+    protected def ipAddressFromFile(id: ContainerId, network: String): Future[ContainerIp] = {
+        configFileContents(containerConfigFile(id)).map { json =>
+            val networks = json.fields("NetworkSettings").asJsObject.fields("Networks").asJsObject
+            val specifiedNetwork = networks.fields(network).asJsObject
+            val ipAddr = specifiedNetwork.fields("IPAddress")
+            ContainerIp(ipAddr.convertTo[String])
+        }
+    }
+
+    // See implemented trait for description
+    override def inspectIPAddress(id: ContainerId, network: String)(implicit transid: TransactionId): Future[ContainerIp] = {
+        ipAddressFromFile(id, network).recoverWith {
+            case _ => super.inspectIPAddress(id, network)
+        }
+    }
+
+    // See implemented trait for description
+    def rawContainerLogs(containerId: ContainerId, fromPos: Long): Future[ByteBuffer] = Future {
+        blocking { // Needed due to synchronous file operations
+            var fis: FileInputStream = null
+            try {
+                val file = containerLogFile(containerId)
+                val size = file.length
+
+                fis = new FileInputStream(file)
+                val channel = fis.getChannel().position(fromPos)
+
+                // Buffer allocation may fail if the log file is too large to hold in memory or
+                // too few space is left on the heap, respectively.
+                var remainingBytes = (size - fromPos).toInt
+                val readBuffer = ByteBuffer.allocate(remainingBytes)
+
+                while (remainingBytes > 0) {
+                    val readBytes = channel.read(readBuffer)
+                    if (readBytes > 0) {
+                        remainingBytes -= readBytes
+                    } else if (readBytes < 0) {
+                        remainingBytes = 0
+                    }
+                }
+
+                readBuffer
+            } catch {
+                case e: Exception =>
+                    throw new IOException(s"rawContainerLogs failed on ${containerId}", e)
+
+            } finally {
+                if (fis != null) fis.close()
+            }
+        }
+    }
+}
+
+trait DockerApiWithFileAccess extends DockerApi {
+
+    /**
+     * Obtains the container's stdout and stderr by reading the internal docker log file
+     * for the container. Said file is written by docker's JSON log driver and has
+     * a "well-known" location and name.
+     *
+     * Reads the log file from the specified position to its end. The returned ByteBuffer
+     * indicates how many bytes were actually read from the file.
+     *
+     * For warm containers, the container log file already holds output from
+     * previous activations that have to be skipped. For this reason, a starting position can be specified.
+     *
+     * Attention: a ByteBuffer is allocated to keep the file from the specified position to its end
+     * fully in memory. At the moment, there is no size limit checking which can lead to
+     * out-of-memory exceptions for very large files.
+     *
+     * Deals with incomplete reads and premature end of file situations. Behavior is undefined
+     * if the log file is changed or truncated while reading.
+     *
+     * @param containerId the container for which to provide logs
+     * @param fromPos position where to start reading the container's log file
+     * @return a ByteBuffer holding the read log file contents
+     */
+    def rawContainerLogs(containerId: ContainerId, fromPos: Long): Future[ByteBuffer]
+}
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
@@ -51,8 +51,9 @@ class DockerClientWithFileAccess(
      * @param containerId Id of the desired Docker container
      * @return canonical location of the container's home directory
      */
-    protected def containerDirectory(containerId: ContainerId) =
+    protected def containerDirectory(containerId: ContainerId) = {
         new File(containersDirectory, containerId.asString).getCanonicalFile()
+    }
 
     /**
      * Provides the configuration file of the specified Docker container.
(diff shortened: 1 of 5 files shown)


core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
```diff
@@ -0,0 +1,188 @@
...
+    /**
+     * Obtains the container's stdout and stderr by reading the internal docker log file
+     * for the container. Said file is written by docker's JSON log driver and has
+     * a "well-known" location and name.
+     *
+     * Reads the log file from the specified position to its end. The returned ByteBuffer
+     * indicates how many bytes were actually read from the file.
+     *
+     * For warm containers, the container log file already holds output from
+     * previous activations that have to be skipped. For this reason, a starting position can be specified.
```
markusthoemmes: Drop that line. Separation of concerns-wise, this is not relevant for the client here.

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerClientWithFileAccessTests.scala
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    def await[A](f: Future[A], timeout: FiniteDuration = 500.milliseconds) = Await.result(f, timeout)
+
+    case class FixtureParam(file: File, writer: FileWriter, dc: DockerClientWithFileAccess)
+
+    def withFixture(test: OneArgTest) = {
+        val file = File.createTempFile(this.getClass.getName, test.name.replaceAll("[^a-zA-Z0-9.-]", "_"))
+        val writer = new FileWriter(file)
+        val dc = dockerClient(file)
```
markusthoemmes: Can we name this `docker`? I have a hard time reading the tests below because it feels like some magic variable to me.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
```diff
@@ -0,0 +1,188 @@
...
+    /**
+     * Provides the contents of the specified Docker container's configuration
+     * file as JSON object.
+     *
+     * @param configFile the container's configuration file in JSON format
+     * @return contents of configuration file as JSON object
+     */
+    protected def configFileContents(configFile: File): Future[JsObject] = Future {
+        blocking { // Needed due to synchronous file operations
+            val source = scala.io.Source.fromFile(configFile)
```
markusthoemmes: `import scala.io.Source`

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerClientWithFileAccessTests.scala
```diff
@@ -0,0 +1,196 @@
...
+    /** Returns a DockerClient with mocked results */
+    def dockerClient(logFile: File) = new DockerClientWithFileAccess()(global) {
+        override def containerLogFile(containerId: ContainerId) = logFile
+    }
+
+    def await[A](f: Future[A], timeout: FiniteDuration = 500.milliseconds) = Await.result(f, timeout)
+
+    case class FixtureParam(file: File, writer: FileWriter, dc: DockerClientWithFileAccess)
+
+    def withFixture(test: OneArgTest) = {
```
markusthoemmes: Please put a short comment on the fixture usage and that the method + case class need to be named exactly as they are.

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerClientWithFileAccessTests.scala
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    val containerId = ContainerId("Id")
+
+    it should "tolerate an empty log file" in { fixture =>
+        val logText = ""
+        writeLogFile(fixture, logText)
+
+        val buffer = await(fixture.dc.rawContainerLogs(containerId, fromPos = 0))
+
```
markusthoemmes: drop the empty line.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
sven-lange-last: I agree. This information is only important when viewing from the OpenWhisk perspective.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
```diff
@@ -0,0 +1,188 @@
...
+    protected def ipAddressFromFile(id: ContainerId, network: String): Future[ContainerIp] = {
+        configFileContents(containerConfigFile(id)).map { json =>
+            val networks = json.fields("NetworkSettings").asJsObject.fields("Networks").asJsObject
+            val specifiedNetwork = networks.fields(network).asJsObject
+            val ipAddr = specifiedNetwork.fields("IPAddress")
+            ContainerIp(ipAddr.convertTo[String])
+        }
+    }
+
+    // See implemented trait for description
```
cbickel: extended? instead of implemented?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
```diff
@@ -0,0 +1,188 @@
...
+    override def inspectIPAddress(id: ContainerId, network: String)(implicit transid: TransactionId): Future[ContainerIp] = {
+        ipAddressFromFile(id, network).recoverWith {
+            case _ => super.inspectIPAddress(id, network)
+        }
+    }
+
+    // See implemented trait for description
```
cbickel: as above

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
```diff
@@ -0,0 +1,188 @@
...
+    def rawContainerLogs(containerId: ContainerId, fromPos: Long): Future[ByteBuffer] = Future {
```
cbickel: `inspectIPAddress` has the keyword `override` and this method not. Can we make this consistent?

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerClientWithFileAccessTests.scala
sven-lange-last: Added comments with some background info on why I use fixtures and marked `FixtureParam` and `withFixture` as hard-wired names.

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerClientWithFileAccessTests.scala
sven-lange-last: Done

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerClientWithFileAccessTests.scala
sven-lange-last: Done

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
```diff
@@ -0,0 +1,186 @@
...
+        blocking { // Needed due to synchronous file operations
+            var fis: FileInputStream = null
```
cbickel: can we move the definition of fis up here to use vals instead of these vars?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
```diff
@@ -0,0 +1,186 @@
...
+            try {
+                val file = containerLogFile(containerId)
+                val size = file.length
+
+                fis = new FileInputStream(file)
+                val channel = fis.getChannel().position(fromPos)
+
+                // Buffer allocation may fail if the log file is too large to hold in memory or
+                // too few space is left on the heap, respectively.
+                var remainingBytes = (size - fromPos).toInt
```
cbickel: Can we use a recursion instead of the while loop to avoid the vars?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
```diff
@@ -0,0 +1,186 @@
...
+                    } else if (readBytes < 0) {
+                        remainingBytes = 0
+                    }
+                }
+
+                readBuffer
+            } catch {
+                case e: Exception =>
+                    throw new IOException(s"rawContainerLogs failed on ${containerId}", e)
+
```
cbickel: Empty line

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
markusthoemmes: The reasoning is, that `inspectIPAddress` truly is an override, while `rawContainerLogs` implements the trait only. We don't use `override` for trait implementations, if they don't override anything.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
markusthoemmes: Note that the whole code here is copied from the old invoker code. It's true, that there may be value in refactoring it. In this specific case (log-file reading) I want to be very cautious with that though. Log files have quite a history and i'd rather not conflate the new containerpool with fixes on that front.

That's not to say we shouldn't try though. Just a disclaimer for caution.

sven-lange-last: Created the variant below and reviewed with @cbickel and @markusthoemmes. Jointly came to the conclusion that we want to keep the current implementation to avoid the risk of changing the current behaviour by mistake. I'm going to create an issue for refactoring log file reading according to the proposal below.

```
    protected def readFile(fileChannel: FileChannel, readBuffer: ByteBuffer, remainingBytes: Int): ByteBuffer = {
        println(s"readBuffer=${readBuffer}")
        if (remainingBytes <= 0) {
            readBuffer
        } else {
            val readBytes = fileChannel.read(readBuffer)
            if (readBytes < 0) {
                readBuffer
            } else {
                readFile(fileChannel, readBuffer, remainingBytes - readBytes)
            }
        }
    }

    protected def withFileAtPosition(file: File, position: Long)(reader: FileChannel => ByteBuffer): ByteBuffer = {
        val fis = new FileInputStream(file)
        try {
            val channel = fis.getChannel().position(position)
            reader(channel)
        } finally {
            fis.close()
        }
    }

    // See implemented trait for description
    def rawContainerLogs(containerId: ContainerId, fromPos: Long): Future[ByteBuffer] = Future {
        blocking { // Needed due to synchronous file operations
            try {
                val file = containerLogFile(containerId)
                val size = file.length

                withFileAtPosition(file, fromPos) { channel =>
                    // Buffer allocation may fail if the log file is too large to hold in memory or
                    // too few space is left on the heap, respectively.
                    val remainingBytes = (size - fromPos).toInt
                    val readBuffer = ByteBuffer.allocate(remainingBytes)

                    readFile(channel, readBuffer, remainingBytes)
                }
            } catch {
                case e: Exception =>
                    throw new IOException(s"rawContainerLogs failed on ${containerId}", e)
            }
        }
    }
```

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
sven-lange-last: Thanks for proposal - keeping the code as is according to discussion [below for line 140](https://github.com/openwhisk/openwhisk/pull/2161/files#r113189580).

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
sven-lange-last: Changed to _extended_ as proposed. Or would _originating_ be more comprehensible?

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
```diff
@@ -32,19 +32,19 @@ import whisk.http.Messages
 /**
  * Represents a single log line as read from a docker log
  */
-protected[invoker] case class LogLine(time: String, stream: String, log: String) {
+protected[core] case class LogLine(time: String, stream: String, log: String) {
```
rabbah: reason for widening scope is... tests?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
```diff
@@ -0,0 +1,185 @@
...
+     * Assumes that property "containersDirectory" holds the location of the
+     * home directory of all Docker containers. Default: directory "containers"
+     * in the current working directory.
+     *
+     * Does not verify that the returned directory actually exists.
+     *
+     * @param containerId Id of the desired Docker container
+     * @return canonical location of the container's home directory
+     */
+    protected def containerDirectory(containerId: ContainerId) =
```
rabbah: please use `{}`.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * Obtains the container's stdout and stderr by reading the internal docker log file
+     * for the container. Said file is written by docker's JSON log driver and has
+     * a "well-known" location and name.
+     *
+     * Reads the log file from the specified position to its end. The returned ByteBuffer
+     * indicates how many bytes were actually read from the file.
+     *
+     * Attention: a ByteBuffer is allocated to keep the file from the specified position to its end
+     * fully in memory. At the moment, there is no size limit checking which can lead to
```
rabbah: is this replacing the action log driver?

markusthoemmes: Wouldn' that be a change in the behavior vs. our behavior today? I'd rather do that in a subsequent PR after the pool is in the codebase.

core/invoker/src/main/scala/whisk/core/invoker/ActionLogDriver.scala
markusthoemmes: Yep.

sven-lange-last: I was able to revert the scope back to `invoker`. The wider scope must have been an intermediate work result. Good finding.

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerClientWithFileAccessTests.scala
```diff
@@ -0,0 +1,206 @@
...
+    val dockerCommand = "docker"
+    val networkInConfigFile = "networkConfig"
+    val networkInDockerInspect = "networkInspect"
+    val ipInConfigFile = ContainerIp("10.0.0.1")
+    val ipInDockerInspect = ContainerIp("10.0.0.2")
+    val dockerConfig =
+        JsObject("NetworkSettings" ->
+            JsObject("Networks" ->
+                JsObject(s"${networkInConfigFile}" ->
```
markusthoemmes: `JsObject(networkInConfigFile)`

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerClientWithFileAccessTests.scala
```diff
@@ -0,0 +1,206 @@
...
+                    JsObject("IPAddress" -> JsString(s"${ipInConfigFile.asString}")))))
```
markusthoemmes: `JsString(ipInConfigFile.asString)`

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerClientWithFileAccessTests.scala
```diff
@@ -0,0 +1,206 @@
...
+        } finally {
+            writer.close()
+            file.delete()
+        }
+    }
+
+    def writeLogFile(fixture: FixtureParam, content: String) = {
+        fixture.writer.write(content)
+        fixture.writer.flush()
+    }
```
markusthoemmes: Couldn't you add this function to the `FixtureParam` class?

sven-lange-last:
PG1 - 1510 running

PR #2165, opened 2017-04-21
Avoid clobbering Python's built-in dict datatype

cclauss:
It considered bad practice to name a variable in such a way that it clobbers access to Python's built-in [dict](https://docs.python.org/3/library/stdtypes.html#mapping-types-dict) datatype.

### DIFF
--- a/tests/dat/actions/hello.py
+++ b/tests/dat/actions/hello.py
@@ -19,12 +19,9 @@
 """
 
 
-def main(dict):
+def main(args_dict):
     """Main."""
-    if 'name' in dict:
-        name = dict['name']
-    else:
-        name = "stranger"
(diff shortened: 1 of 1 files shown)


tests/dat/actions/hello.py
```diff
@@ -19,12 +19,9 @@
 """
 
 
-def main(dict):
+def main(args_dict):
```
akrabat: In Swift actions the parameter name is "`args`" and I think Node actions use "`params`". Maybe pick one of those two?

rabbah:
Merged as https://github.com/openwhisk/openwhisk/commit/0dad5f7b8a4715feb7291f33775b7c06214137c4.

PR #2168, opened 2017-04-21
update watson-developer-cloud npm to 2.29.0

sjfink:
This is necessary to pick up recent changes to watson services.

:large_blue_circle:  PG2 1480


rabbah:
does the version bump introduce any breaking changes for their api?

sjfink:
@rabbah no breaking changes I know of and it passes our watson package regression tests

PR #2172, opened 2017-04-23
`wsk api delete` command may not completely delete endpoint

mdeuser:
Fix is to delete all artifacts in the swagger for the specified endpoint


mdeuser:
PG1/1436
@csantanapr - for review/merge after pg/travis

PR #2177, opened 2017-04-25
Provide DockerContainer implementation

### LINKED ISSUE #2089: Implement a Container abstraction
This aims to implement a container abstraction. That is, for now we need `DockerContainer` but the abstraction should facilitate plugin in whatever container driver we may want to use. This will facilitate work on other container providers as well as make Docker-less testing and simulation possible.

sven-lange-last:
This implementation is an abstraction layer for using Docker containers as OpenWhisk action containers - in preparation of a new container pool implementation (#2086). The `DockerContainer` implementation provides a high-level abstraction for the operations required by OpenWhisk:

- Create, destroy, halt, resume a container
- Initialize and run an action inside a container
- Obtain container logs

Closes #2089

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
@@ -50,7 +50,7 @@ import whisk.core.entity.size.SizeLong
  * @param timeoutMsec the timeout in msecs to wait for a response
  * @param maxResponse the maximum size in bytes the connection will accept
  */
-class HttpUtils(
+protected[core] class HttpUtils(
     hostname: String,
     timeout: FiniteDuration,
     maxResponse: ByteSize) {
--- a/core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
@@ -50,7 +50,7 @@ import whisk.core.entity.size.SizeLong
  * @param timeoutMsec the timeout in msecs to wait for a response
  * @param maxResponse the maximum size in bytes the connection will accept
  */
-protected[container] class HttpUtils(
+class HttpUtils(
     hostname: String,
     timeout: FiniteDuration,
     maxResponse: ByteSize) {
--- a/core/invoker/src/main/scala/whisk/core/containerpool/Container.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/Container.scala
@@ -47,7 +47,7 @@ trait Container {
     def run(parameters: JsObject, environment: JsObject, timeout: FiniteDuration)(implicit transid: TransactionId): Future[(Interval, ActivationResponse)]
 
     /** Obtains logs up to a given threshold from the container. Optionally waits for a sentinel to appear. */
-    def logs(limit: ByteSize, waitForSentinel: Boolean)(implicit transid: TransactionId): Future[List[String]]
+    def logs(limit: ByteSize, waitForSentinel: Boolean)(implicit transid: TransactionId): Future[Vector[String]]
 }
 
 /** Indicates a general error with the container */
--- a/core/invoker/src/main/scala/whisk/core/containerpool/Container.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/Container.scala
@@ -56,10 +56,10 @@ sealed abstract class ContainerError(msg: String) extends Exception(msg)
 /** Indicates an error while starting a container */
 sealed abstract class ContainerStartupError(msg: String) extends ContainerError(msg)
 
-/** Indicates an error while starting a container of a managed runtime */
+/** Indicates any error while starting a container either of a managed runtime or a non-application-specific blackbox container */
 case class WhiskContainerStartupError(msg: String) extends ContainerStartupError(msg)
 
-/** Indicates an error while starting a blackbox container */
+/** Indicates an application-specific error while starting a blackbox container */
 case class BlackboxStartupError(msg: String) extends ContainerStartupError(msg)
 
 /** Indicates an error while initializing a container */
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
@@ -21,8 +21,7 @@ import java.time.Instant
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
-import scala.concurrent.duration.DurationInt
-import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration._
 import scala.util.Failure
 import scala.util.Success
 
@@ -57,7 +56,7 @@ object DockerContainer {
      * @param environment environment variables to set on the container
      * @param network network to launch the container in
      * @param name optional name for the container
-     * @return a container
+     * @return a Future which either completes with a DockerContainer or one of two specific failures
      */
     def create(transid: TransactionId,
                image: String,
@@ -67,7 +66,7 @@ object DockerContainer {
                environment: Map[String, String] = Map(),
                network: String = "bridge",
                name: Option[String] = None)(
-                   implicit docker: DockerApiWithFileAccess, runc: RuncApi, ec: ExecutionContext, log: Logging) = {
+                   implicit docker: DockerApiWithFileAccess, runc: RuncApi, ec: ExecutionContext, log: Logging): Future[DockerContainer] = {
         implicit val tid = transid
 
         val environmentArgs = (environment + ("SERVICE_IGNORE" -> true.toString)).map {
@@ -78,7 +77,7 @@ object DockerContainer {
             "--cap-drop", "NET_RAW",
             "--cap-drop", "NET_ADMIN",
             "--ulimit", "nofile=1024:1024",
-            "--pids-limit", "1024", // OW PR 2119
+            "--pids-limit", "1024",
             "--cpu-shares", cpuShares.toString,
             "--memory", s"${memory.toMB}m",
             "--memory-swap", s"${memory.toMB}m",
@@ -123,7 +122,7 @@ object DockerContainer {
  * @param ip the ip of the container
  */
 class DockerContainer(id: ContainerId, ip: ContainerIp)(
-    implicit docker: DockerApiWithFileAccess, runc: RuncApi, ec: ExecutionContext, log: Logging) extends Container with ActionLogDriver {
+    implicit docker: DockerApiWithFileAccess, runc: RuncApi, ec: ExecutionContext, logger: Logging) extends Container with ActionLogDriver {
 
     /** The last read-position in the log file */
     private var logFileOffset = 0L
@@ -153,7 +152,7 @@ class DockerContainer(id: ContainerId, ip: ContainerIp)(
             } else if (result.interval.duration >= timeout) {
                 Future.failed(InitializationError(ActivationResponse.applicationError(Messages.timedoutActivation(timeout, true)), result.interval))
             } else {
-                Future.failed(InitializationError(ActivationResponse.processInitResponseContent(result.response, log), result.interval))
+                Future.failed(InitializationError(ActivationResponse.processInitResponseContent(result.response, logger), result.interval))
             }
         }
     }
@@ -168,12 +167,12 @@ class DockerContainer(id: ContainerId, ip: ContainerIp)(
             case Success(r: RunResult) =>
                 transid.finished(this, start.copy(start = r.interval.start), s"running result: ${r.toBriefString}", endTime = r.interval.end)
             case Failure(t) =>
-                transid.failed(this, start, s"initializiation failed with $t")
+                transid.failed(this, start, s"run failed with $t")
         }.map { result =>
             val response = if (result.interval.duration >= timeout) {
                 ActivationResponse.applicationError(Messages.timedoutActivation(timeout, false))
             } else {
-                ActivationResponse.processRunResponseContent(result.response, log)
+                ActivationResponse.processRunResponseContent(result.response, logger)
             }
 
             (result.interval, response)
@@ -210,7 +209,7 @@ class DockerContainer(id: ContainerId, ip: ContainerIp)(
                 val (isComplete, isTruncated, formattedLogs) = processJsonDriverLogContents(rawLog, waitForSentinel, limit)
 
                 if (retries > 0 && !isComplete && !isTruncated) {
-                    log.info(this, s"log cursor advanced but missing sentinel, trying $retries more times")
+                    logger.info(this, s"log cursor advanced but missing sentinel, trying $retries more times")
                     Thread.sleep(logsRetryWait.toMillis)
                     readLogs(retries - 1)
                 } else {
@@ -219,7 +218,7 @@ class DockerContainer(id: ContainerId, ip: ContainerIp)(
                 }
             }.andThen {
                 case Failure(e) =>
-                    log.error(this, s"Failed to obtain logs of ${id.asString}: ${e.getClass} - ${e.getMessage}")
+                    logger.error(this, s"Failed to obtain logs of ${id.asString}: ${e.getClass} - ${e.getMessage}")
             }
         }
 
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
@@ -133,7 +133,7 @@ class DockerContainer(id: ContainerId, ip: ContainerIp)(
     /** HTTP connection to the container, will be lazily established by callContainer */
     private var httpConnection: Option[HttpUtils] = None
 
-    def halt()(implicit transid: TransactionId): Future[Unit] = runc.pause(id)
+    def suspend()(implicit transid: TransactionId): Future[Unit] = runc.pause(id)
     def resume()(implicit transid: TransactionId): Future[Unit] = runc.resume(id)
     def destroy()(implicit transid: TransactionId): Future[Unit] = docker.rm(id)
 
@@ -150,9 +150,9 @@ class DockerContainer(id: ContainerId, ip: ContainerIp)(
             if (result.ok) {
                 Future.successful(result.interval)
             } else if (result.interval.duration >= timeout) {
-                Future.failed(InitializationError(ActivationResponse.applicationError(Messages.timedoutActivation(timeout, true)), result.interval))
+                Future.failed(InitializationError(result.interval, ActivationResponse.applicationError(Messages.timedoutActivation(timeout, true))))
             } else {
-                Future.failed(InitializationError(ActivationResponse.processInitResponseContent(result.response, logger), result.interval))
+                Future.failed(InitializationError(result.interval, ActivationResponse.processInitResponseContent(result.response, logger)))
             }
         }
     }
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
@@ -87,27 +87,23 @@ object DockerContainer {
 
         val pulled = if (userProvidedImage) {
             docker.pull(image).recoverWith {
-                case _ => Future.failed(BlackboxStartupError(s"Failed to pull container image '$image'."))
+                case _ => Future.failed(BlackboxStartupError(s"Failed to pull container image '${image}'."))
             }
         } else Future.successful(())
 
-        val container = for {
+        for {
             _ <- pulled
-            id <- docker.run(image, args)
-            ip <- docker.inspectIPAddress(id, network).andThen {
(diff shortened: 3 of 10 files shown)


core/invoker/src/main/scala/whisk/core/containerpool/Container.scala
```diff
@@ -0,0 +1,60 @@
...
+     * Initializes code in the container.
+     */
+    def initialize(initializer: JsObject, timeout: FiniteDuration)(implicit transid: TransactionId): Future[Interval]
+
+    /**
+     * Runs code in the container.
+     */
+    def run(parameters: JsObject, environment: JsObject, timeout: FiniteDuration)(implicit transid: TransactionId): Future[(Interval, ActivationResponse)]
+
+    def logs(limit: ByteSize, waitForSentinel: Boolean)(implicit transid: TransactionId): Future[Vector[String]]
```
markusthoemmes: Scaladoc please.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -0,0 +1,251 @@
...
+import whisk.common.LoggingMarkers
+import whisk.core.entity.ByteSize
+import java.nio.charset.StandardCharsets
+import whisk.http.Messages
+import whisk.core.entity.ActivationResponse
+import whisk.core.containerpool.BlackboxStartupError
+import whisk.core.containerpool.WhiskContainerStartupError
+import whisk.core.containerpool.InitializationError
+import whisk.core.containerpool.Container
+import whisk.core.invoker.ActionLogDriver
```
markusthoemmes: care to reorg imports?

sven-lange-last: Done - also taken care of imports in other files...

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
```diff
@@ -50,7 +50,7 @@ import whisk.core.entity.size.SizeLong
  * @param timeoutMsec the timeout in msecs to wait for a response
  * @param maxResponse the maximum size in bytes the connection will accept
  */
-protected[container] class HttpUtils(
+class HttpUtils(
```
rabbah: why making this generally visible?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -0,0 +1,251 @@
...
+        val actionName = environment.fields.get("action_name").map(_.convertTo[String]).getOrElse("")
+        val start = transid.started(this, LoggingMarkers.INVOKER_ACTIVATION_RUN, s"sending arguments to $actionName at $id $ip")
+
+        val parameterWrapper = JsObject("value" -> parameters)
+        val body = JsObject(parameterWrapper.fields ++ environment.fields)
+        callContainer("/run", body, timeout, retry = false).andThen { // never fails
+            case Success(r: RunResult) =>
+                transid.finished(this, start.copy(start = r.interval.start), s"running result: ${r.toBriefString}", endTime = r.interval.end)
+            case Failure(t) =>
+                transid.failed(this, start, s"initializiation failed with $t")
```
rabbah: initializiation -> should be run

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -0,0 +1,251 @@
...
+
+        val environmentArgs = (environment + ("SERVICE_IGNORE" -> true.toString)).map {
+            case (key, value) => Seq("-e", s"$key=$value")
+        }.flatten
+
+        val args = Seq(
+            "--cap-drop", "NET_RAW",
+            "--cap-drop", "NET_ADMIN",
+            "--ulimit", "nofile=1024:1024",
+            "--pids-limit", "1024", // OW PR 2119
```
rabbah: 2119 already merged; remove comment.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -0,0 +1,251 @@
...
+            "--cpu-shares", cpuShares.toString,
+            "--memory", s"${memory.toMB}m",
+            "--memory-swap", s"${memory.toMB}m",
+            "--network", network) ++
+            environmentArgs ++
+            name.map(n => Seq("--name", n)).getOrElse(Seq.empty)
+
+        val pulled = if (userProvidedImage) {
+            docker.pull(image).recoverWith {
```
rabbah: we should put a limit on the docker pull operation (per previous discussion with @perryibm this was not done in the past so as penalize for connectivity to dockerhub - but the absence of any time limit is also generally bad).

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -0,0 +1,251 @@
...
+     * @param userProvidedImage whether the image is provided by the user
+     *     or is an OpenWhisk provided image
+     * @param memory memorylimit of the container
+     * @param cpuShares sharefactor for the container
+     * @param environment environment variables to set on the container
+     * @param network network to launch the container in
+     * @param name optional name for the container
+     * @return a container
+     */
+    def create(transid: TransactionId,
```
rabbah: add return type to the definition.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -0,0 +1,251 @@
...
+     * @param transid transaction creating the container
+     * @param image image to create the container from
+     * @param userProvidedImage whether the image is provided by the user
+     *     or is an OpenWhisk provided image
+     * @param memory memorylimit of the container
+     * @param cpuShares sharefactor for the container
+     * @param environment environment variables to set on the container
+     * @param network network to launch the container in
+     * @param name optional name for the container
+     * @return a container
```
rabbah: returns a future which either completes with a container or one of two specific failures.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -0,0 +1,251 @@
...
+ *
+ * This class contains OpenWhisk specific behavior and as such does not necessarily
+ * use docker commands to achieve the effects needed.
+ *
+ * @constructor
+ * @param id the id of the container
+ * @param ip the ip of the container
+ */
+class DockerContainer(id: ContainerId, ip: ContainerIp)(
+    implicit docker: DockerApiWithFileAccess, runc: RuncApi, ec: ExecutionContext, log: Logging) extends Container with ActionLogDriver {
```
rabbah: i suggest renaming `log` to `logger` or something that more obviously connotes this is the system logger.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -0,0 +1,251 @@
...
+                Future.failed(InitializationError(ActivationResponse.applicationError(Messages.timedoutActivation(timeout, true)), result.interval))
+            } else {
+                Future.failed(InitializationError(ActivationResponse.processInitResponseContent(result.response, log), result.interval))
+            }
+        }
+    }
+
+    def run(parameters: JsObject, environment: JsObject, timeout: FiniteDuration)(implicit transid: TransactionId): Future[(Interval, ActivationResponse)] = {
+        val actionName = environment.fields.get("action_name").map(_.convertTo[String]).getOrElse("")
+        val start = transid.started(this, LoggingMarkers.INVOKER_ACTIVATION_RUN, s"sending arguments to $actionName at $id $ip")
```
rabbah: is the action name logging important here?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -0,0 +1,251 @@
...
+        val parameterWrapper = JsObject("value" -> parameters)
+        val body = JsObject(parameterWrapper.fields ++ environment.fields)
+        callContainer("/run", body, timeout, retry = false).andThen { // never fails
+            case Success(r: RunResult) =>
+                transid.finished(this, start.copy(start = r.interval.start), s"running result: ${r.toBriefString}", endTime = r.interval.end)
+            case Failure(t) =>
+                transid.failed(this, start, s"initializiation failed with $t")
+        }.map { result =>
+            val response = if (result.interval.duration >= timeout) {
+                ActivationResponse.applicationError(Messages.timedoutActivation(timeout, false))
```
rabbah: this is a corner case here where duration = timeout and result.response is valid.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -0,0 +1,251 @@
...
+            val conn = new HttpUtils(s"${ip.asString}:8080", timeout, 1.MB)
+            httpConnection = Some(conn)
+            conn
+        }
+        Future {
+            http.post(path, body, retry)
+        }.map { response =>
+            val finished = Instant.now()
+            RunResult(Interval(started, finished), response)
+        }
```
rabbah: how about a recover for defense sake and log an error; or assert a violation.

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerContainerTests.scala
```diff
@@ -0,0 +1,640 @@
...
+        docker.inspects should have size 1
+        docker.rms should have size 0
+
+        val (testImage, args) = docker.runs.head
+        testImage shouldBe "image"
+
+        // Assert fixed values are passed as well
+        args should contain allOf ("--cap-drop", "NET_RAW", "NET_ADMIN")
+        args should contain inOrder ("--ulimit", "nofile=1024:1024")
+        args should contain inOrder ("--pids-limit", "1024") // OW PR 2119
```
rabbah: remove 2119 comment

rabbah: could consolidate all these into a `Seq(x, y).foreach args should contain inOrder (x, y)`

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerContainerTests.scala
```diff
@@ -0,0 +1,640 @@
...
+
+        docker.pulls should have size 0
+        docker.runs should have size 1
+        docker.inspects should have size 1
+        docker.rms should have size 1
+    }
+
+    it should "disambiguate errors if user images are provided" in {
+        implicit val docker = new TestDockerClient {
+            override def inspectIPAddress(id: ContainerId, network: String)(implicit transid: TransactionId): Future[ContainerIp] = {
```
rabbah: this test seems wrong in that unlike the next test that follows, failure on inspect is a system error not a black box error, is it not?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
markusthoemmes: It is for some performance related tooling. This keeps this log as-is.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
markusthoemmes: You're saying it should be `duration > timeout`? I think this is semantically preserving the current code, but worth double-checking.

tests/src/test/scala/whisk/core/invoker/ActionLogDriverTests.scala
```diff
@@ -77,8 +77,8 @@ class ActionLogDriverTests
             raw"""|{"time":"","stream":"stdout","log":"a"}
                   |{"time":"","stream":"stdout","log":"b"}
                   |{"time":"","stream":"stdout","log":"c"}
-                  |{"time":"","stream":"stdout","log":"$LOG_ACTIVATION_SENTINEL\n"}
-                  |{"time":"","stream":"stderr","log":"$LOG_ACTIVATION_SENTINEL\n"}""".stripMargin('|')
+                  |{"time":"","stream":"stdout","log":"$ActionLogDriver.LOG_ACTIVATION_SENTINEL\n"}
+                  |{"time":"","stream":"stderr","log":"$ActionLogDriver.LOG_ACTIVATION_SENTINEL\n"}""".stripMargin('|')
```
markusthoemmes: How do those still pass? Doesn't this need to be `${ActionLogDriver.LOG_ACTIVATION_SENTINEL}`

core/invoker/src/main/scala/whisk/core/container/HttpUtils.scala
sven-lange-last: `whisk.core.containerpool.docker.DockerContainer` uses `HttpUtils` so it cannot be `protected[container]`. Changed to `protected[core]`.

tests/src/test/scala/whisk/core/invoker/ActionLogDriverTests.scala
sven-lange-last: Good finding - changed to `${ActionLogDriver.LOG_ACTIVATION_SENTINEL}`. Also removed the explicit line endings `\n` because further processing removes them anyway.

The problem with `${ActionLogDriver.LOG_ACTIVATION_SENTINEL}` wasn't detected by unit tests because those don't cover sentinel processing.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
sven-lange-last: I'm not sure what you have in mind. At the moment, we have two consumers of `callContainer()`, i.e. `initialize()` and `run()`. Those two consumers deal with a failed `Future`.

If we add a `recover` here and return a successful `Future[RunResult]` in any case, the caller's error handling isn't used any more.

So should we keep the error handling as is - or move it down to `callContainer()`?

markusthoemmes: The clients don't have recovers, so they don't deal with a failing future.

rabbah: the contract from the http client is that it should never fail the future - so my comment was to defensively code against an incompatible change there (or, to make sure we have a test to assert the call always returns a successful future).

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
rabbah: likely preserving behavior - i trust you'll check. but if so the behavior is wrong for the border case anyway. either fix or open issue to fix later.

tests/src/test/scala/whisk/core/invoker/ActionLogDriverTests.scala
rabbah: so is the test wrong? i'm not sure i understood the last comment.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
sven-lange-last: I checked - it's in `getActivationResponse()` in `Invoker.scala`:

```
        if (interval.duration >= timeout) {
            ActivationResponse.applicationError(Messages.timedoutActivation(timeout, failedInit))
        } else ...
```

I opt for keeping the current semantics even if we have the impression that the corner case `duration == timeout` may need special handling. I think that we should keep behaviour changes with the new container pool at a minimum so that we can compare test results. We can open an issue and follow up once the new container pool is in place.

tests/src/test/scala/whisk/core/invoker/ActionLogDriverTests.scala
sven-lange-last: The `ActionLogDriver` unit test isn't wrong - but doesn't fully cover the area of sentinel processing. Our goal was to keep the `ActionLogDriver` and the existing unit tests as is. So we aligned our new log processing code around `ActionLogDriver` - but externalized the `LOG_ACTIVATION_SENTINEL` such that we can use it in our new unit tests for `DockerClientWithFileAccess`.

In summary, no urgent need for changes in `ActionLogDriver` and corresponding unit tests.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
sven-lange-last: Besides the access scope change, we left `HttpUtils` as is. It still returns a `Left()` if unsuccessful or a `Right()` if successful.

Both callers of the code under discussion here - namely, `initialize()` and `run()` - have the following error handling code:

```
        callContainer("/init", body, timeout, retry = true).andThen { // never fails
            case Success(r: RunResult) =>
                transid.finished(this, start.copy(start = r.interval.start), s"initialization result: ${r.toBriefString}", endTime = r.interval.end)
            case Failure(t) =>
                transid.failed(this, start, s"initializiation failed with $t")
        } ...
```

So we are going to log a failed `Future` returned by `callContainer()` and return the failed `Future` to the caller of `initialize()` and `run()`. Looking at https://github.com/openwhisk/openwhisk/pull/2021/files, I do NOT think that the caller of `initialize()` and `run()` fully addresses a failed `Future`.

@rabbah do you suggest to `recover` here at line 251 and transform the `Failure(t)` into a `RunResult(..., Left(t))`? This transformation would bypass the code snippet from above with `transid.failed()`. I guess we need a decision how hard the penalty for a failed `Future` is: a log statement, a failed transaction marker, a failed init / run, or even an assertion stopping the invoker process. What do you have in mind?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
sven-lange-last: I agree - `docker pull` takes potentially very long and may block a single Docker container for a long time. It's reasonable to let the operation fail if pulling an image takes too long.

- What would be a reasonable timeout duration for `docker pull`? Should we re-use the same timeout we apply to initialize / run? Or do we need a general `docker pull` timeout?
- At the moment, I'm building a prototype with a timeout mechanism based on `akka.pattern.after`. Unfortunately, said `after()` needs a scheduler - most likely implying that we need the proper actor system at this point. So instead of passing the `ExecutionContext`, we need the whole `ActorSystem` and use `ActorSystem.scheduler` and `ActorSystem.dispatcher`.
- Said approach with `akka.pattern.after` used in `DockerContainer.create()` will let the creation fail with a timeout if it takes too long - but won't actually abort the running `docker pull` command process. Our current command execution approach does not support aborting a running process. We may think about providing a process executor that observes timeouts and abort execution upon timeout.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
markusthoemmes: As we return a Future[Either] we might aswell leave it as-is. The documentation I think states the contract already. If we want to leave HttpUtils as-is I'd leave that wrinkle in.

Alternatively we could flatten the Future[Either] where for the Left case we fail the Future. Note though that we leave type-information in that case.

rabbah: @sven-lange-last are you able to restore the scope of the `HttpUtils` to `container`? If so, we could do nothing beyond the comment. I am a little concerned that earlier `protected[container]` became `public` where we might end up making an incompatible change that we don't catch.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
rabbah: we have a timeout feature we use in the controller that may be reusable here see `ExecutionContextFactory` for the future and promise extensions that offer `after`.

rabbah: We should not use the init and run timeout - unless we also treat this as a billable operation. Perhaps we should.

sven-lange-last: We created a variant with timeout in `ProcessRunner` that will be provided in a separate PR.

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerContainerTests.scala
sven-lange-last: You are right. We inspect the Docker config to obtain the IPv4 address iff `docker run` has been successful. This inspection is expected to work properly even if the container has exited meanwhile.
For this reason, an inspection failure should not be a `BlackboxStartupError`which is mapped to an application error.

sven-lange-last:
PG3 - 594 🔵

PR #2189, opened 2017-04-26

houshengbo:
Closes-Bug: #2101


bin/wskadmin
```diff
@@ -1 +0,0 @@
-../tools/admin/wskadmin
```
rabbah: why are these being removed?

PR #2193, opened 2017-04-27
Implement a cache backed by Caffeine

markusthoemmes:
Caffeine is the successor of ConcurrentLinkedHashMap, which is no longer under active development. spray.caching hasn't upgraded yet, so this is our own implementation of spray.caching using Caffeine.

Benefits are:
- Better performance.
- Usage of SoftReferences to correspond to memory pressure.

For more details, see: https://github.com/ben-manes/caffeine/

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
+++ b/common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
@@ -17,20 +17,21 @@
 package whisk.core.database
 
 import java.util.concurrent.atomic.AtomicReference
+import java.util.concurrent.ConcurrentMap
+import java.util.concurrent.TimeUnit
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.Promise
-import scala.concurrent.duration.DurationInt
 import scala.util.Failure
 import scala.util.Success
 
 import spray.caching.Cache
-import spray.caching.LruCache
 import spray.caching.ValueMagnet.fromAny
 import whisk.common.Logging
 import whisk.common.LoggingMarkers
 import whisk.common.TransactionId
+import com.github.benmanes.caffeine.cache.Caffeine
 
 /**
  * A cache that allows multiple readers, but only a single writer, at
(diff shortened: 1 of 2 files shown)


common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -409,5 +411,50 @@ trait MultipleReadersSingleWriterCache[W, Winfo] {
...
+ * and reimplemented using the slightly differing API of Caffeine.
+ *
+ * The unimplemented methods are not needed for our internal implementation.
+ */
+private class CaffeineBackedCache[V](store: CaffeineCache[Any, Future[V]]) extends Cache[V] {
+
+    def get(key: Any) = Option(store.getIfPresent(key))
+
+    def apply(key: Any, genValue: () ⇒ Future[V])(implicit ec: ExecutionContext): Future[V] = {
+        val promise = Promise[V]()
```
rabbah: Unused promise?

markusthoemmes: Yep, copy&paste garbage.

markusthoemmes:
PG2 1502 🔵

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -409,5 +410,43 @@ trait MultipleReadersSingleWriterCache[W, Winfo] {
...
+ * The implementation is mainly copied from [[spray.caching.LruCache]]
+ * and reimplemented using the slightly differing API of Caffeine.
+ *
+ * The unimplemented methods are not needed for our internal implementation.
+ */
+private class CaffeineBackedCache[V](store: CaffeineCache[Any, Future[V]]) extends Cache[V] {
+
+    def get(key: Any) = Option(store.getIfPresent(key))
+
+    def apply(key: Any, genValue: () ⇒ Future[V])(implicit ec: ExecutionContext): Future[V] = get(key).getOrElse {
```
starpit: is there a race condition here? what we're missing (from SprayCache) is the `putIfAbsent` method, which atomically inserts the entry (but only when the key does not exist). in contrast this line has `get(key).getOrElse` followed by (on line 437) a `store.put`

consider a racing put and remove. the put gets in first, and does a `get(key)`. the remove then comes along and invalidates the key. but then the put actor wakes up and, at line 437, restores the entry.

markusthoemmes: Yep, that's actually the one thing I thought our augmentation with SingleWriterMultipleReaders would help us.

Apparently there's an equivalent to `putIfAbsent`, updating.

starpit:
in order to restore the required `putIfAbsent` atomicity, there are a few options:

1) code directly to ConcurrentHashMap, which offers `putIfAbsent`. this won't be a huge amount of code, but it is new code

2) make a PR against CaffeineCache to expose `putIfAbsent` -- after all, CaffeineCache uses ConcurrentHashMap as its impl. we can't currently do this with a subclass, because the map impl is private. so a PR would either (or both) make the private map protected, and/or add the putIfAbsent method

is it notable that the comments for BasicCache (which is the interface for CaffeineCache) states "A facade for benchmark implementations"? and that Caffeine also offers a more complete BoundedLocalCache impl? which leads to option 3...

3) use BoundedLocalCache (BLC) as our impl, instead of CaffeineCache. BLC offers putIfAbsent. Also, it is possible that we can use BLC for much of our MultipleReadersSingleWriter (MRSW) impl. Not sure yet how much, but initial inspection leads to the guess that BLC does much of what MRSW does?

ben-manes:
You might be a interested in the [akka-http cache](https://github.com/akka/akka-http/pull/986) that uses Caffeine and is intended to help migrations from Spray. You can also use the asMap() view to get access to methods like putIfAbsent and computations

markusthoemmes:
@ben-manes thank you **so much** for pointing me to `asMap()` 🤕 . Was about to blame Java/Scala interop 😀

ben-manes:
My pleasure. It looks like you might also want to use policy() for the ordered keys method. The [migration notes](https://github.com/ben-manes/caffeine/wiki/ConcurrentLinkedHashMap) might be helpful.

rabbah:
Thanks @ben-manes for the tip!

PR #2194, opened 2017-04-27
Make JVM options available at deployment time

markusthoemmes:
Currently, JVM settings are hardcoded into our images. The only value we really want to have on those applications is the restart on OOM behavior though. The other settings are highly dependent on the deployment infrastructure.

- `controller|invoker_heap` are used to control the heap of both components.
- `controller|invoker_arguments` are used for free-form JVM arguments.


markusthoemmes:
PG1 1453 🔵

PR #2196, opened 2017-04-27
Remove flag to configure split of whisks and activations db.

cbickel:
With this PR, the ability to save activations in the whisks database will go away.
All activations will be saved in the activations database. If you have activations in your whisks database, they will not be returned by an OpenWhisk call (e.g. wsk activation list). Only the activations in the activations database are returned by the controller.
If you have not migrated your databases yet, there is an manual at https://gist.github.com/cbickel/37e651965781b27de245eac0ce831a53.
If you don't need old data, the `wipe.yml` will update your databases by recreating the activations and actions db. This will wipe all your data.


ansible/roles/controller/tasks/deploy.yml
```diff
@@ -38,6 +38,7 @@
       - "{{ whisk_logs_dir }}/controller:/logs"
     ports:
       - "{{ controller.port }}:8080"
+    command: /bin/sh -c "controller/bin/controller {{play_hosts.index(inventory_hostname)}} >> /logs/controller{{play_hosts.index(inventory_hostname)}}_logs.log 2>&1"
```
markusthoemmes: Bad push? Why is this needed in here?

core/controller/Dockerfile
```diff
@@ -5,14 +5,13 @@ ENV DEBIAN_FRONTEND noninteractive
...
-    tar zxf v2.1.4.tar.gz -C /swagger-ui --strip-components=2 swagger-ui-2.1.4/dist && \ 
+    tar zxf v2.1.4.tar.gz -C /swagger-ui --strip-components=2 swagger-ui-2.1.4/dist && \
     rm v2.1.4.tar.gz && \
     perl -pi -w -e  's{http://petstore.swagger.io/v2/swagger.json}{/api/v1/api-docs}g;' /swagger-ui/index.html
 
 #
 # Copy app jars
 COPY build/distributions/controller.tar ./
 RUN tar xf controller.tar
-CMD controller/bin/controller >> /logs/${COMPONENT_NAME}_logs.log 2>&1
```
markusthoemmes: Superfluous Controller changes.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -150,8 +150,8 @@ object Controller {
         val config = new WhiskConfig(requiredProperties, optionalProperties)
 
         // if deploying multiple instances (scale out), must pass the instance number as the
-        // second argument.  (TODO .. seems fragile)
-        val instance = if (args.length > 0) args(1).toInt else 0
+        require(args.length >= 1, "controller instance required")
+        val instance = args(0).toInt
```
markusthoemmes: Not needed in this PR?

cbickel:
As suggested in the mail:
I think this PR can be merged by the end of the week.

cbickel:
PG3#591 is green.

PR #2197, opened 2017-04-27
Add a fudge factor to handle clock skew in the invokers

### LINKED ISSUE #2195: WskRuleTests tends to break with invoker clock skew
This test creates a rule, fires the trigger and expects to see that the rule has run. It does this by looking for activations that start at the same time (or later) as the activation obtained from firing the trigger.

The problem is that the invoker which ran the action's clock is behind the container that fired the trigger. In this sense, the logs indicate that the action ran before the trigger was fired, and the test fails to find the activation for the action.

WskRuleTests seem especially susceptible to failing as a result. Perhaps a fudge-factor could be introduced into the since parameter when looking for action activations - perhaps 500 or even a 1000 ms?

bjustin-ibm:
Due to clock skew among invokers, it is possible for an action invoked by a rule _appears_ to start before the trigger which activated the rule was even fired. As such, these tests tend to fail because it is looking for action activations that start strictly after the time the trigger activation started.

This change introduces a "fudge factor" to allow the test to find the expected action activations that appear to start at most 500ms before the trigger activation start time.

Resolves #2195

### DIFF
--- a/tests/src/test/scala/system/basic/WskRuleTests.scala
+++ b/tests/src/test/scala/system/basic/WskRuleTests.scala
@@ -82,13 +82,18 @@ class WskRuleTests
         }
     }
 
+    /**
+     * Append the current timestamp in ms
+     */
+    def withTimestamp(text: String) = s"${text}-${System.currentTimeMillis}"
+
     behavior of "Whisk rules"
 
     it should "invoke the action attached on trigger fire, creating an activation for each entity including the cause" in withAssetCleaner(wskprops) {
         (wp, assetHelper) =>
-            val ruleName = "r1to1"
-            val triggerName = "t1to1"
-            val actionName = "a1 to 1" // spaces in name intended for greater test coverage
+            val ruleName = withTimestamp("r1to1")
+            val triggerName = withTimestamp("t1to1")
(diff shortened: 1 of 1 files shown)


rabbah:
Is there an issue for the address the clock skew in the containers?

markusthoemmes:
The clock skew needs to be adressed on the underlying machines, not the containers themselves @rabbah.

markusthoemmes:
@bjustin-ibm I think we should make sure that all entity names are different. Otherwise I could see those tests failing because a later test grabs an activation from an earlier test because of the fudge factor.

bjustin-ibm:
> I think we should make sure that all entity names are different

@markusthoemmes I added timestamps to all the entity names to make them unique.

bjustin-ibm:
> The clock skew needs to be addressed on the underlying machines

In order to solve this particular problem with better clock syncing, we would have to ensure that clocks never skew by even 1ms, as activation times are recorded in ms. I don't know enough about clock syncing to say for sure but, to me, this doesn't seem practically feasible.

bjustin-ibm:
PG2 1508

rabbah:
spanner does it :) my point was only to make sure we're also looking at addressing the underlying issue (to a reasonable extent).

PR #2205, opened 2017-04-28
Add ability to deploy a "hot-standby" controller

cbickel:
This PR adds the ability to deploy a second controller, which will be a hot-standby controller.
The plan is, that this controller is only used in case the first controller crashes.
All work, that needs to be done, to run two controllers in parallel (like cache invalidation) will not be part of this PR.

### DIFF
--- a/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/ansible/roles/nginx/templates/nginx.conf.j2
@@ -22,10 +22,10 @@ http {
         # Otherwise, nginx would dispatch requests when the container is up, but the backend in the container not.
         # From the docs:
         # "normally, requests with a non-idempotent method (POST, LOCK, PATCH) are not passed to the next server if a request has been sent to an upstream server"
-        server {{ groups['controllers'] | first }}:{{ controller.port }} fail_timeout=60s;
+        server {{ groups['controllers'] | first }}:{{ controller.basePort }} fail_timeout=60s;
 {% for ip in groups['controllers'] %}
     {% if groups['controllers'].index(ip) > 0 %}
-        server {{ ip }}:{{ controller.port + groups['controllers'].index(ip) }} backup;
+        server {{ ip }}:{{ controller.basePort + groups['controllers'].index(ip) }} backup;
     {% endif %}
 {% endfor %}
     }
--- a/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/ansible/roles/nginx/templates/nginx.conf.j2
@@ -2,7 +2,7 @@
 
 events {
 {# default: 1024 #}
-    worker_connections  4096;  
+    worker_connections  4096;
 }
 
 http {
@@ -15,6 +15,15 @@ http {
         '$request $status $body_bytes_sent '
         '$http_referer $http_user_agent $upstream_addr';
     access_log /logs/nginx_access.log combined-upstream;
+
+    upstream controllers {
+        server {{ groups['controllers'] | first }}:{{ controller.port }};
+{% for ip in groups['controllers'] %}
+    {% if groups['controllers'].index(ip) > 0 %}
+        server {{ ip }}:{{ controller.port + groups['controllers'].index(ip) }} backup;
+    {% endif %}
+{% endfor %}
+    }
 
     server {
         listen 443 default ssl;
@@ -42,7 +51,7 @@ http {
             if ($namespace) {
               rewrite    /(.*) /api/v1/web/${namespace}/$1 break;
             }
-            proxy_pass http://{{ groups['controllers']|first }}:{{ controller.port }};
+            proxy_pass http://controllers;
             proxy_read_timeout 70s; # 60+10 additional seconds to allow controller to terminate request
         }
 
@@ -51,7 +60,7 @@ http {
             if ($namespace) {
               rewrite    ^ /api/v1/web/${namespace}/public/index.html break;
             }
-            proxy_pass http://{{ groups['controllers']|first }}:{{ controller.port }};
+            proxy_pass http://controllers;
             proxy_read_timeout 70s; # 60+10 additional seconds to allow controller to terminate request
         }
 
--- a/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/ansible/roles/nginx/templates/nginx.conf.j2
@@ -17,7 +17,12 @@ http {
     access_log /logs/nginx_access.log combined-upstream;
 
     upstream controllers {
-        server {{ groups['controllers'] | first }}:{{ controller.port }};
+        # fail_timeout: period of time the server will be considered unavailable
+        # Mark the controller as unavailable for at least 60 seconds, to not get any requests during restart.
+        # Otherwise, nginx would dispatch requests when the container is up, but the backend in the container not.
+        # From the docs:
+        # "normally, requests with a non-idempotent method (POST, LOCK, PATCH) are not passed to the next server if a request has been sent to an upstream server"
+        server {{ groups['controllers'] | first }}:{{ controller.port }} fail_timeout=60s;
 {% for ip in groups['controllers'] %}
     {% if groups['controllers'].index(ip) > 0 %}
         server {{ ip }}:{{ controller.port + groups['controllers'].index(ip) }} backup;
--- a/ansible/templates/whisk.properties.j2
+++ b/ansible/templates/whisk.properties.j2
@@ -47,10 +47,8 @@ limits.triggers.fires.perMinute={{ limits.triggers.fires.perMinute }}
 {% endif %}
 
 consulserver.host={{ groups["consul_servers"]|first }}
-controller.host={{ groups["controllers"]|first }}
 edge.host={{ groups["edge"]|first }}
 kafka.host={{ groups["kafka"]|first }}
-loadbalancer.host={{ groups["controllers"]|first }}
 router.host={{ groups["edge"]|first }}
 zookeeper.host={{ groups["kafka"]|first }}
 invoker.hosts={{ groups["invokers"] | join(",") }}
@@ -59,12 +57,12 @@ edge.host.apiport=443
 zookeeper.host.port={{ zookeeper.port }}
 kafka.host.port={{ kafka.port }}
 kafkaras.host.port={{ kafka.ras.port }}
-controller.host.port={{ controller.port }}
-loadbalancer.host.port={{ controller.port }}
 consul.host.port4={{ consul.port.http }}
 consul.host.port5={{ consul.port.server }}
 invoker.hosts.baseport={{ invoker.port }}
 
+controller.hosts={{ groups["controllers"] | join(",") }}
+controller.host.basePort={{ controller.basePort }}
 controller.instances={{ controller.instances }}
 
 invoker.container.network=bridge
--- a/ansible/templates/whisk.properties.j2
+++ b/ansible/templates/whisk.properties.j2
@@ -65,6 +65,8 @@ consul.host.port4={{ consul.port.http }}
 consul.host.port5={{ consul.port.server }}
 invoker.hosts.baseport={{ invoker.port }}
 
+controller.instances={{ controller.instances }}
+
 invoker.container.network=bridge
 invoker.container.policy={{ invoker_container_policy_name | default()}}
 invoker.container.dns={{ invoker_container_network_dns_servers | default()}}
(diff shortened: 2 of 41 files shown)


rabbah:
s/redo/wskadmin.

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -174,9 +174,12 @@ object TransactionId {
...
+    val numberOfInstances: Int
+    val instance: Int
+
     def transid(): TransactionId = {
-        TransactionId(cnt.incrementAndGet())
+        TransactionId(cnt.addAndGet(numberOfInstances))
     }
 
-    private val cnt = new AtomicInteger(1)
+    private val cnt = new AtomicInteger(numberOfInstances + instance)
```
cbickel: For example if you have N controllers:
The tid of invoker  0 will start with N, the tid of invoker 1 with N+1 and so on...
After the change in line 181, the tid is not incremented by one, but by N.
So the tid of controller0 will always be x*N, the tid of controller 1 will be x*N+1 and the tid of controller(N-1) will be x*N+(N-1).

So if you have two controllers, controller 0 will have all eval tids and invoker 1 all odd tids. So the tid is unique across controllers.

rabbah: This is just setting the initial value however isn't it? Don't you mean to use this as a delta on the increment; or am I missing something?

rabbah: Yes I am. You did increment by the delta the line above. Ok shame on me.

ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -16,6 +16,15 @@ http {
         '$http_referer $http_user_agent $upstream_addr';
     access_log /logs/nginx_access.log combined-upstream;
 
+    upstream controllers {
+        server {{ groups['controllers'] | first }}:{{ controller.port }};
+{% for ip in groups['controllers'] %}
+    {% if groups['controllers'].index(ip) > 0 %}
+        server {{ ip }}:{{ controller.port + groups['controllers'].index(ip) }} backup;
```
rabbah: Will you address cache invalidation between primary and backup instances in this pr?

cbickel: Cache invalidation is not part of this PR.
The plan of this PR is to deploy a backup controller.
It will be there in case the first controller crashes (for the time to recover and for the case, that the first controller does not recover).

cbickel: So the plan is, that the second controller is not used (yet). So I think that the cache does not have to be invalidated.

rabbah: Can we be sure the backup is only used when primary crashed? Failure for nginx is just failure to establish a connection upstream which doesn't necessarily say primary is down - just unresponsive. So it's possible the caches will slip and eventually have to become consistent again.

cbickel:
With the nginx-config in this PR, the behaviour will be the following:
- as long as the first controller is working, all requests will be redirected to the first controller
- when the first controller crashes, all new requests will be redirected to the second controller
  If the current activations in the controller have not reached Kafka, they will be lost.
  If the activation has already reached Kafka, it will be executed.
  If the connection to the client is still open, but the activation already reached Kafka, he will receive a 502, but the activation will be processed. It will be processed only once. Nginx does not trigger the same action on the second controller again.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -189,7 +189,7 @@ object InvokerPool {
  * This finite state-machine represents an Invoker in its possible
  * states "Healthy" and "Offline".
  */
-class InvokerActor extends FSM[InvokerState, InvokerInfo] {
+class InvokerActor(instance: Int) extends FSM[InvokerState, InvokerInfo] {
```
markusthoemmes: Want to rename this to `controllerInstance`?

ansible/templates/whisk.properties.j2
```diff
@@ -59,12 +57,13 @@ edge.host.apiport=443
...
 kafkaras.host.port={{ kafka.ras.port }}
-controller.host.port={{ controller.port }}
-loadbalancer.host.port={{ controller.port }}
 consul.host.port4={{ consul.port.http }}
 consul.host.port5={{ consul.port.server }}
 invoker.hosts.baseport={{ invoker.port }}
 
+controller.hosts={{ groups["controllers"] | join(",") }}
+controller.host.basePort={{ controller.basePort }}
+controller.docker.port=8080
```
markusthoemmes: What's this needed for?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
cbickel: Addressed in 51f45abb96297b9c6dd2885f0c00f94c7b060156

cbickel:
PG2#1578 is green.

cbickel:
PG3#617 is green

cbickel:
PG3#652 is green

rabbah:
Is the hot standby mandatory? For local dev will there be one or two controllers?

rabbah:
@ddragosd heads up this will affect devtools.

core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
```diff
@@ -73,7 +73,7 @@ class ActivationThrottler(consulServer: String, loadBalancer: LoadBalancer, conc
         // Any sort of partitioning will be ok for monitoring
         Future.sequence(userActivationCounter.groupBy(_._1.take(1)).map {
             case (prefix, items) =>
-                val key = ControllerKeys.userActivationCountKey + "/" + prefix
+                val key = ControllerKeys.userActivationCountKey(controllerInstance) + "/" + prefix
```
rabbah: Should this (the throttle) aggregate over all instances - a comment is warranted in this class that explains the accounting is local to an instance (and hence clears on restarts).

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -77,16 +76,14 @@ protected[core] object EntitlementProvider {
  * A trait that implements entitlements to resources. It performs checks for CRUD and Acivation requests.
  * This is where enforcement of activation quotas takes place, in additional to basic authorization.
  */
-protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBalancer: LoadBalancer)(
+protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBalancer: LoadBalancer, controllerInstance: Int)(
```
rabbah: Related to my comment above - the leaking of the instance into the entitlement strikes me as wrong. I understand why it's happening but suggests we're not quite done on this front. Redis perhaps will allow us to address the inconsistency.

cbickel:
To your question: `Is the hot standby mandatory? For local dev will there be one or two controllers?`

This PR just enables the ability to deploy a second controller on another machine. For all deployments on only one machine, there will be no hot-standby controller.
If you add a second host into the controller group into the hosts-file from ansible, only the first controller will be active. All other controllers will be hot-standby controllers.

core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
```diff
@@ -53,34 +50,27 @@ class ActivationThrottler(consulServer: String, loadBalancer: LoadBalancer, conc
...
+    def isOverloaded()(implicit tid: TransactionId): Boolean = {
+        val concurrentActivations = userActivationCounter.values.sum
+        logging.info(this, s"concurrent activations in system = $concurrentActivations, below limit = $systemOverloadLimit")
+        concurrentActivations > systemOverloadLimit
     }
 
     Scheduler.scheduleWaitAtLeast(healthCheckInterval) { () =>
         userActivationCounter = loadBalancer.getActiveUserActivationCounts
-        publishUserConcurrentActivation()
+        Future.successful(true)
```
markusthoemmes: `Future.successful(())` (`Unit`) should suffice here.

cbickel:
PG3#694 is green.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -115,6 +117,7 @@ class Controller(
     Collection.initialize(entityStore)
 
     /** The REST APIs. */
+    implicit val controllerInstance = instance
```
markusthoemmes: Uh oh. Is this an implicit `Int`? You don't really want to do that, because `Int` is such a generic type that by looking at the type annotations you have no clue as to what is actually in the implicit scope. That's quite dangerous and can lead to ambiguity really fast.

Instead you can define an enclosing value class, which won't get instantiated on runtime but serves as a wrapper explicitly identifying what you're talking about in the implicit scope.

```
class InstanceId(val instance: Int) extends AnyVal
```

The implicit will now be of type `InstanceId` which transports the meaning that you want.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -177,13 +181,13 @@ object Controller {
...
-        // second argument.  (TODO .. seems fragile)
-        val instance = if (args.length > 0) args(1).toInt else 0
+        require(args.length >= 1, "controller instance required")
+        val instance = args(0).toInt
 
         // initialize the runtimes manifest
         if (config.isValid && ExecManifest.initialize(config)) {
             val port = config.servicePort.toInt
-            BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, instance, logger))
+            BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, instance, logger), 30.seconds)
```
rabbah: this will delay the server bringup by 30 seconds - and creates a phase ordering with the invoker deployment.
if the invokers are not yet up, waiting 30 seconds is of no use and only delays deployment since it will take 30s for the controller to respond to the health ping.

so this is highly nuanced for the case of a restart in the presence of a hot standby on an existing deployed system.

cbickel:
PG2#1705 is green.

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -45,8 +44,7 @@ class KafkaConsumerConnector(
     sessionTimeout: FiniteDuration = 30 seconds,
     autoCommitInterval: FiniteDuration = 10 seconds)(
         implicit logging: Logging)
-    extends MessageConsumer
-    with TransactionCounter {
```
rabbah: wow that's really old!

core/controller/src/main/scala/whisk/core/controller/Controller.scala
rabbah: why is this implicit anyway?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -48,6 +48,9 @@ protected[actions] trait PrimitiveActions {
 
     protected implicit val logging: Logging
 
+    /** The index of the active ack topic, this controller is listening for. Typically this is also the instance number of the controller */
```
rabbah: nit: too long for a single line. Add period at the end of the second line.

```
/**
 * ...
  * ..
   */
``

File: tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala [hunk 12]
```diff
@@ -151,7 +151,7 @@ class WskBasicUsageTests
         val tmpwskprops = File.createTempFile("wskprops", ".tmp")
         try {
             val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
-            val apihost = s"http://${WhiskProperties.getControllerHost}:${WhiskProperties.getControllerPort}"
+            val apihost = s"http://${WhiskProperties.getControllerHosts.split(",")(0)}:${WhiskProperties.getControllerBasePort}"
```
rabbah: should whisk properties just have a method to return the base controller?

common/scala/src/main/scala/whisk/core/entity/InstanceId.scala
```diff
@@ -0,0 +1,27 @@
+/*
```
rabbah: this should be the new apache license.

rabbah: i'll fix it in #2384.

PR #2207, opened 2017-04-30
API GW CLI SSO support

mdeuser:
Base CLI http changes to allow for custom error response body parsing

### DIFF
--- a/tools/cli/go-whisk-cli/commands/api.go
+++ b/tools/cli/go-whisk-cli/commands/api.go
@@ -820,12 +820,8 @@ var apiCreateCmdV2 = &cobra.Command{
         apiCreateReq.ApiDoc = api
 
         apiCreateReqOptions := new(whisk.ApiCreateRequestOptions)
-        props, _ := readProps(Properties.PropsFile)
-        apiCreateReqOptions.SpaceGuid = strings.Split(props["AUTH"], ":")[0]
-        apiCreateReqOptions.AccessToken = "DUMMY_TOKEN"
-        if len(props["APIGW_ACCESS_TOKEN"]) > 0 {
-            apiCreateReqOptions.AccessToken = props["APIGW_ACCESS_TOKEN"]
-        }
+        apiCreateReqOptions.SpaceGuid, _ = getUserContextId()
+        apiCreateReqOptions.AccessToken, _ = getAccessToken()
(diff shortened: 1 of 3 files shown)


mdeuser:
PG3/559

mdeuser:
@dubeejw - Would you please review.

cc: @csantanapr

mdeuser:
PG1/1471

tools/cli/go-whisk-cli/commands/api.go
```diff
@@ -1050,35 +1038,22 @@ var apiListCmdV2 = &cobra.Command{
...
-        apiListReqOptions.Skip = flags.common.skip
-        apiListReqOptions.AccessToken = accesstoken
-        apiListReqOptions.SpaceGuid = spaceguid
-
         if (len(args) == 0) {
+            // List API request query parameters
+            apiListReqOptions := new(whisk.ApiListRequestOptions)
+            apiListReqOptions.Limit = flags.common.limit
+            apiListReqOptions.Skip = flags.common.skip
+            apiListReqOptions.SpaceGuid, _ = getUserContextId()
```
dubeejw: What happens if an error occurs for getUserContextId() or AccessToken()?

houshengbo:
@mdeuser Be aware that we also have a separate repo openwhisk-cli to host the source code of cli. When this is merged, could you change it in the new repo accordingly?

mdeuser:
@houshengbo - when will there be a single repo?

houshengbo:
@mdeuser There has already been a repo for openwhisk-cli: https://github.com/openwhisk/openwhisk-cli. I submitted a PR to install the binaries from the new repo.

tools/cli/go-whisk-cli/commands/api.go
mdeuser: @dubeejw - When an error (only if problem reading the configuration file) occurs, a default value is returned for each of these value.  Since the properties file is read earlier in the command flow, it's unlikely, but possible, the additional properties file read will fail when the earlier one does not.  The dummy values are:
- "DUMMY TOKEN" for the API GW access token, which will work fine with the open API GW
- "" for the Space GUID.  Since this is pulled from the auth key, having an error here but not elsewhere is unlikely; but just in case the logic flow changes; i'll add a quick check for the failure

mdeuser:
pg4/206 - single unrelated failure

PR #2208, opened 2017-05-01
Define `long` integer for Python 3

cclauss:
* https://docs.python.org/3.3/whatsnew/3.0.html#integers
* https://www.python.org/dev/peps/pep-0237

### DIFF
--- a/tools/travis/flake8.sh
+++ b/tools/travis/flake8.sh
@@ -14,7 +14,7 @@ declare -a PYTHON_FILES=("."
 echo 'Flake8: first round (fast fail) stops the build if there are any Python 3 syntax errors...'
 for i in "${PYTHON_FILES[@]}"
 do
-    flake8 "$i" --select=E999 --statistics
+    flake8 "$i" --select=E999,F821 --statistics
     RETURN_CODE=$?
     if [ $RETURN_CODE != 0 ]; then
         echo 'Flake8 found Python 3 syntax errors above.  See: https://docs.python.org/3/howto/pyporting.html'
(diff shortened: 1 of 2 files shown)


markusthoemmes:
Do we have a documented setup for our python checking so each developer individually can adjust her editors to warn on stuff like this? Have we agreed on a set of checks?

cclauss:
[Travis runs flake8](https://travis-ci.org/openwhisk/openwhisk/builds/227574950#L325) to find these issues.   I use the Atom editor and the [linter-flake8 package](https://atom.io/packages/linter-flake8) which highlights these issues in realtime as you type code.  Other editors will have similar hooks into [flake8](http://flake8.pycqa.org).

markusthoemmes:
Can we make travis fail on those failures?

tools/travis/flake8.sh
```diff
@@ -14,7 +14,7 @@ declare -a PYTHON_FILES=("."
 echo 'Flake8: first round (fast fail) stops the build if there are any Python 3 syntax errors...'
 for i in "${PYTHON_FILES[@]}"
 do
-    flake8 "$i" --select=E999 --statistics
+    flake8 "$i" --select=E999,F821 --statistics
```
markusthoemmes: Should we rather do this the other way round: Fail on everything and only exclude the rules we don't like?

cclauss: There are a [ton of rules in flake8](https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes) and as was discussed in https://github.com/openwhisk/openwhisk/pull/2056 we do not want to get too nitpicky in this project where Python is just 1.7% of the codebase.  I think the two pass compromise that we settled on is the right one.  Code issues that halt the runtime should also halt PR builds but those code issues that do not halt the runtime are recorded for those that care to review them but do not halt progress.

PR #2214, opened 2017-05-04
increase nginx worker connections

xin-cai:
increase nginx worker connections and change worker processes to auto

### DIFF
--- a/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/ansible/roles/nginx/templates/nginx.conf.j2
@@ -1,8 +1,10 @@
 {# this template is used to generate a nginx.conf for booting a nginx server based on the given environment inventory #}
 
+worker_processes auto;
+
 events {
 {# default: 1024 #}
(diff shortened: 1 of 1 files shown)


xin-cai:
PG4/202 is running

markusthoemmes:
Can you elaborate on this change? What are we expecting to change since you bumped the connection by quite little.

xin-cai:
reset `worker_connections` back to `4096` to minimize the impact,  the new configuration of `worker_processes auto` can already increase the nginx capacity


```
worker_processes – The number of NGINX worker processes (the default is 1). In most cases, running one worker process per CPU core works well, and we recommend setting this directive to auto to achieve that. There are times when you may want to increase this number, such as when the worker processes have to do a lot of disk I/O.
```

xin-cai:
PG4/203 is green

xin-cai:
dup with https://github.com/openwhisk/openwhisk/pull/2233, close this one

Issue #2097, opened 2017-04-01
update CLI help activation list command to specify max is 200


csantanapr:
`wsk activation list -h` doesn't specify that max is 200 for `limit` only default is 30

```
.... from the collection (default 30)
```

maybe update to 

```
.... from the collection (default 30, max 200)
```

```
wsk activation list -h
list activations
Usage:
  wsk activation list [NAMESPACE or NAME] [flags]

Flags:
  -f, --full          include full activation description
  -l, --limit LIMIT   only return LIMIT number of activations from the collection (default 30)
      --since SINCE   return activations with timestamps later than SINCE; measured in milliseconds since Th, 01, Jan 1970
  -s, --skip SKIP     exclude the first SKIP number of activations from the result
      --upto UPTO     return activations with timestamps earlier than UPTO; measured in milliseconds since Th, 01, Jan 1970

Global Flags:
      --apihost HOST         whisk API HOST
      --apiversion VERSION   whisk API VERSION
  -u, --auth KEY             authorization KEY
  -d, --debug                debug level output
  -i, --insecure             bypass certificate checking
  -v, --verbose              verbose output
```

cc @rabbah

dubeejw:
@csantanapr, the minimum and maximums should be added to all command (action, trigger, etc...) flags that take numeric input.  The only issue is that we can't include the min and max in parentheses with the default as that text is generated by a package we utilize to parse arguments. We'd have to do something that would look like this `only return `LIMIT` number of activations from the collection (min 1, max 200) (default 30)`

rabbah:
it would nice if this is done to avoid hard coding any of these values in the cli directly - to avoid going out of sync with the api/backend.

some limits and available will soon be available as part of pr #1980; we could consider adding more here, or reading the info from the swagger where appropriate.

i would like to avoid a repeat of the schema getting out of sync with the backend with bit us repeatedly until some of it was removed.

csantanapr:
+1 drive thru configuration, do not hard code numbers if possible.

dubeejw:
Hmm, sounds like we will have to query the backend for limits when the help commands are invoked.

rabbah:
I've seen this several times where users try to use wsk activation list with limit greater than 200 and don't see any more than 200 activations. The help says default is 30 but doesn't state a max. In the interest of less confusion doing the simple thing first works for me (hardcoding 200 in the cli help).

dubeejw:
[...]

PR #1980, opened 2017-03-08
Add info routes to provide deployment configuration

rabbah:
Provide deployment configuration: routes to API paths, available runtime information, and action/trigger limits.

See #335 for details.

PR #2215, opened 2017-05-04
Updates for Max Activation List Limit

dubeejw:
- Inform users that 200 activations is the maximum limit for the list CLI command
- Make controller return an error if activation limit is greater than 200
- Closes https://github.com/openwhisk/openwhisk/issues/2097

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Activations.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Activations.scala
@@ -43,6 +43,10 @@ import whisk.core.entity.types.ActivationStore
 import whisk.http.Messages
 import whisk.http.ErrorResponse.terminate
 
+object WhiskActivationsApi {
+    protected[core] val maxActivationLimit = 200
+}
+
 /** A trait implementing the activations API. */
 trait WhiskActivationsApi
     extends Directives
@@ -64,8 +68,6 @@ trait WhiskActivationsApi
 
     /** Only GET is supported in this API. */
     protected override lazy val entityOps = get
-
-    protected val maxActivationLimit = 200
 
     /** Validated entity name as an ActivationId from the matched path segment. */
     protected override def entityname(n: String) = {
@@ -114,10 +116,10 @@ trait WhiskActivationsApi
     private def list(namespace: EntityPath)(implicit transid: TransactionId) = {
         parameter('skip ? 0, 'limit ? collection.listLimit, 'count ? false, 'docs ? false, 'name.as[EntityName]?, 'since.as[Instant]?, 'upto.as[Instant]?) {
             (skip, limit, count, docs, name, since, upto) =>
-                val cappedLimit = if (limit == 0) maxActivationLimit else limit
+                val cappedLimit = if (limit == 0) WhiskActivationsApi.maxActivationLimit else limit
 
                 // regardless of limit, cap at maxActivationLimit (200) records, client must paginate
-                if (cappedLimit <= maxActivationLimit) {
+                if (cappedLimit <= WhiskActivationsApi.maxActivationLimit) {
                     val activations = name match {
                         case Some(action) =>
                             WhiskActivation.listCollectionByName(activationStore, namespace, action, skip, cappedLimit, docs, since, upto)
@@ -133,7 +135,7 @@ trait WhiskActivationsApi
                         }
                     }
                 } else {
-                    terminate(BadRequest, Messages.maxActivationLimitExceeded(limit, maxActivationLimit))
+                    terminate(BadRequest, Messages.maxActivationLimitExceeded(limit, WhiskActivationsApi.maxActivationLimit))
                 }
         }
     }
--- a/core/controller/src/main/scala/whisk/core/controller/Activations.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Activations.scala
@@ -31,6 +31,7 @@ import spray.httpx.unmarshalling.MalformedContent
 import spray.routing.Directives
 import spray.json.DefaultJsonProtocol.RootJsObjectFormat
 import spray.json._
+import spray.http.StatusCodes.BadRequest
 
 import whisk.common.TransactionId
 import whisk.core.entitlement.Collection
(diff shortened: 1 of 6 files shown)


dubeejw:
@mdeuser, please review.

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -1337,4 +1337,13 @@ class WskBasicUsageTests
             }
             tmpProps.delete()
     }
+
+    behavior of "Wsk activation"
+
+    it should "reject activation list with a limit that exceeds the maximum limit" in {
+        val maxLimit = 200
```
mdeuser: Is a configurable openwhisk limit - that is, it's available in whisk.properties or other configuration file?

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -31,12 +31,9 @@ import (
     "github.com/spf13/cobra"
 )
 
-const (
-    PollInterval = time.Second * 2
-    Delay        = time.Second * 5
-)
+const MAX_ACTIVATION_LIMIT = 200
```
mdeuser: Is this a CLI limit or a backend limit?  If it's a backend limit, augmenting the existing backend /api/v1 api or add a new one to provide various backend configuration properties/limits would reduce possibility of CLI needing to change in the future if/when this limit is increased.  I suppose this could also be baked into the CLI at compile time, but there's a small exposure of a mismatched CLI/backend.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -340,7 +347,7 @@ var activationPollCmd = &cobra.Command{
 
 func init() {
     activationListCmd.Flags().IntVarP(&flags.common.skip, "skip", "s", 0, wski18n.T("exclude the first `SKIP` number of activations from the result"))
-    activationListCmd.Flags().IntVarP(&flags.common.limit, "limit", "l", 30, wski18n.T("only return `LIMIT` number of activations from the collection"))
+    activationListCmd.Flags().IntVarP(&flags.common.limit, "limit", "l", DEFAULT_ACTIVATION_LIMIT, wski18n.T("only return `LIMIT` number of activations from the collection with a maximum LIMIT of 200 activations"))
```
mdeuser: Make the `200` value a message parameter instead of part of the static string

tools/cli/go-whisk-cli/commands/activation.go
dubeejw: I agree, we wanted a quick fix right now. See the issue for the convo on dynamically providing max/min values.

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
dubeejw: The max limit is hardcoded in open/core/controller/src/main/scala/whisk/core/controller/Activations.scala. `val cappedLimit = if (limit == 0 || limit > 200) 200 else limit`

dubeejw: Making controller throw an error.

dubeejw:
@mdeuser, changes made base on review comments. @rabbah, probably need your review on controller changes.

dubeejw:
PG4 207

tools/cli/go-whisk-cli/commands/activation.go
mdeuser: It looks like merged pr #1980 has the basic structure to include this limit now..  Is there another issue to pull this value from that backend interface?

dubeejw: Yeah, this is a quick fix for now. It'll take a little longer to implement what we actually want.

mdeuser: Open an issue so it's not forgotten :-)

dubeejw: https://github.com/openwhisk/openwhisk/issues/2219

csantanapr:
@rabbah go ahead and merge when you review it

dubeejw:
@rabbah, I think functionality would be more unified across consumers of the controller by modifying the backend activation code. Otherwise, we'd have different behavior in CLI, the Node JS client, etc. In any event, the first commit only modifies the CLI. Second commit modifies the controller.

dubeejw:
@rabbah, @csantanapr, I'm not sure what is needed to do to this PR since there isn't any feedback comments. Am open to whatever changes are best.

dubeejw:
First commit only changes the CLI code. An error is originated in the CLI if activation list limit is > 200. Alternatively, the second commit generates an error in the controller if the activation list limit is > 200.

rabbah:
Thanks @dubeejw I'll queue it up for review again on my end.

mdeuser:
@dubeejw - Since this is a backend limit and not a CLI limit, it's best to have the backend enforce the limit.  As you mentioned, this will also support all other OW clients.  This value will need documenting.  If the value can be dynamic (i.e. controlled by OW config), then the doc should include a way to retrieve that value - perhaps from the existing (or new) OW info REST API.

core/controller/src/main/scala/whisk/core/controller/Activations.scala
```diff
@@ -62,6 +64,8 @@ trait WhiskActivationsApi
     /** Only GET is supported in this API. */
     protected override lazy val entityOps = get
 
+    protected val maxActivationLimit = 200
```
rabbah: is this better located in a singleton object?

tools/cli/go-whisk-cli/commands/activation.go
rabbah: do you intend to keep these limits in the cli if the change is in the backend?
that is, will you hardcode the 200 in this pr or defer this change to a time when you can get the limit from the api info?

dubeejw:
@rabbah, for now I'll hardcode activation limit into the CLI for the usage information. I have another PR open for making the max activation limit configurable with Ansible and displaying the value through the controller info route. After that change is finished, we can update the CLI to read the limits dynamically from the controller and remove the hardcoded values.

PR #2218, opened 2017-05-05
Port Controller from Spray to Akka

dubeejw:
- Refactor controller to use Akka as Spray is no longer supported


dubeejw:
Need to consider if we want to replace the existing controller with the port to Akka, or have tons of duplicate code hanging around for version 2 of the controller.

dubeejw:
Still need to port over triggers, activations, rules, and packages.

core/controller/src/main/scala/whisk/core/controller/v2/Actions.scala
```diff
@@ -0,0 +1,658 @@
...
+     */
+    private def execAnnotation(exec: Exec): Parameters = {
+        Parameters(WhiskAction.execFieldName, exec.kind)
+    }
+
+    /** Max atomic action count allowed for sequences */
+    private lazy val actionSequenceLimit = whiskConfig.actionSequenceLimit.toInt
+
+    /** Custom deserializer for timeout query parameter. */
+    private implicit val stringToTimeoutDeserializer = new FromStringDeserializer[FiniteDuration] {
```
dubeejw: Need to fix this to deserialize FiniteDuration with Akka.

rabbah:
> Need to consider if we want to replace the existing controller with the port to Akka, or have tons of duplicate code hanging around for version 2 of the controller.

I imagined - where necessary - to refactor the handlers from v1 so they are free of routing dependencies and so v1 vs v2 just become routing specific, with serializers for the the corresponding http response types.

Definitely do _not_ want to duplicate/clone-and-own all of v1 as v2.

markusthoemmes:
Can we port v1 as is, without going to v2? Benefits include: 

- Move away from an unsupported spray.
- Move away from an unsupported scala version (spray makes us stay on 2.11, while 2.12 is out)

markusthoemmes:
Quick q on the procedure here: Do we want to copy all the code or do we want to migrate v1 to akka-http in-place? Asking specifically because I'd like to keep the git history of these files intact.

rabbah:
+1 I do think there's a way to do the migration to Akka-http without all this copying.

dubeejw:
POST requests currently require a payload. Otherwise, an error response is generated. Not sure if this is new in Akka, @rabbah. Will have to figure it out.

Another defect is that generated error responses are not currently JSON formatted.

dubeejw:
For the port, I plan on overwriting the existing controller so that there isn't a lot of duplicate code. For now, I'm just working with the duplicate code until I get everything converted over. Just need web actions now.

dubeejw:
Side note, keep getting a consul test failure in Travis when running two controllers.

dubeejw:
Currently I have replaced the Spray controller with the Akka controller. However, controller unit tests are failing as they are still using Spray, so these tests will have to be updated. Also the BasicHttpService will have to be updated to use Akka since the controller and invoker extend this class.

dubeejw:
Currently I have replaced the Spray controller with the Akka controller. However, controller unit tests are failing as they are still using Spray, so these tests will have to be updated. Also the BasicHttpService will have to be updated to use Akka since the controller and invoker extend it.

dubeejw:
I'm going to squash all the commits into one commit due to the nightmare that just resulted after rebasing against upstream.

core/controller/src/main/scala/whisk/core/controller/API.scala
```diff
@@ -77,16 +77,16 @@ protected[controller] class SwaggerDocs(apipath: Uri.Path, doc: String)(implicit
...
-        WhiskConfig.whiskVersion ++
-                WhiskAuthStore.requiredProperties ++
-                WhiskEntityStore.requiredProperties ++
-                WhiskActivationStore.requiredProperties ++
-                WhiskConfig.consulServer ++
-                EntitlementProvider.requiredProperties ++
-                WhiskActionsApi.requiredProperties ++
-                Authenticate.requiredProperties ++
-                Collection.requiredProperties
+    def requiredProperties = Map(WhiskConfig.servicePort -> 8080.toString) ++
```
dubeejw: Service port should be configurable in Ansible.

dubeejw:
TODO
- Fix test failures
- Add back auth tests
- Transaction IDs not work
- Figure out custom error handling

dubeejw:
PG4 516, approved.

dubeejw:
@rabbah, please review. All tests are passing.

dubeejw:
BasicHttpService.scala is a little different since HTTPService is not offered by Akka. Might be able to move `bindAndHandle` calls into there and remove the Actor extension.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -173,16 +174,21 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
                                             Path.SingleSlash + rule.action.name.asString
                                         }
                                     }.toString
-
                                     val actionUrl = Path("/api/v1") / "namespaces" / actionNamespace / "actions"
+                                    val request = HttpRequest(
+                                        method = POST,
+                                        uri = url.withPath(actionUrl + actionPath),
+                                        headers = List(Authorization(BasicHttpCredentials(user.authkey.uuid.toString, user.authkey.key.toString))),
+                                        entity = HttpEntity(MediaTypes.`application/json`, args.get.compactPrint)
```
markusthoemmes: That `get` I believe is unsafe.

rabbah: Two lines up toString should be asString (in other places to - all our types have asString converters; toString used for printing primarily).

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
```diff
@@ -19,56 +19,64 @@ package whisk.http
...
     implicit def logging: Logging
+    implicit val mat = ActorMaterializer()
+
+    /** Rejection handler to terminate connection on a bad request. Delegates to Akka handler. */
+    implicit def customRejectionHandler(implicit transid: TransactionId) =
+        RejectionHandler.default.mapRejectionResponse {
+                case resp @ HttpResponse(_, _, ent: HttpEntity.Strict, _) =>
+                    val timeout = 1.second
+                    val marshalledError = Marshal(ErrorResponse(ent.data.utf8String, transid)).to[ResponseEntity]
+                    val marshalledRes = Await.result(marshalledError.flatMap(_.toStrict(timeout)), timeout)
```
markusthoemmes: Blocking here could lead us to terrible performance, why is it needed?

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
```diff
@@ -19,56 +19,64 @@ package whisk.http
 
 import scala.concurrent.duration.DurationInt
 import scala.language.postfixOps
+import scala.concurrent.Await
+import scala.collection.immutable.Seq
+import scala.concurrent.ExecutionContext.Implicits.global
```
markusthoemmes: What's the default execution context used for?

rabbah: As a general rule neither the implicit context should be used or an explicit Await. @dubeejw can you scrub all occurrences that might have been introduced in the code (tests are different).

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
```diff
@@ -86,39 +94,58 @@ trait BasicHttpService extends HttpService with TransactionCounter {
...
 
+    /** Rejection handler to terminate connection on a bad request. Delegates to Akka handler. */
+    val prioritizeRejections = recoverRejections { rejections =>
+        val priorityRejection = rejections.find(_.isInstanceOf[UnacceptedResponseContentTypeRejection]).getOrElse(null)
+
+        if (priorityRejection != null) {
+            Rejected(Seq(priorityRejection))
+        } else {
+            Rejected(rejections)
+        }
```
markusthoemmes: Idiomatically these would be:

```scala
val priorityRejection = rejections.find { case _: UnacceptedResponseContentTypeRejection => true }
priorityRejection.map(rejection => Rejected(Seq(rejection))).getOrElse(Rejected(rejections))
```

`getOrElse(null)` should be avoided at any costs and is in fact not useful in this case. Alternatively you could've omitted that and called `priorityRejection.isDefined` on the option. Using map though is the safest way here.

common/scala/build.gradle
```diff
@@ -21,7 +21,7 @@ dependencies {
```
markusthoemmes: All of those should be removable to complete the pull-request, modulo spray-json.

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
```diff
@@ -86,39 +94,58 @@ trait BasicHttpService extends HttpService with TransactionCounter {
...
+                    }
                 }
             }
-        })
+        }
+    }
+
+    def receive = {
+        case _ =>
+    }
```
markusthoemmes: why is this needed? Is an Actor required to implement it?

core/controller/src/main/scala/whisk/core/controller/Activations.scala
```diff
@@ -205,12 +206,37 @@ trait WhiskActivationsApi
     }
 
     /** Custom deserializer for query parameters "since" and "upto" into a valid Instant. */
-    private implicit val stringToInstantDeserializer = new FromStringDeserializer[Instant] {
+    private implicit val stringToInstantDeserializer = new FromStringUnmarshaller[Instant] {
```
markusthoemmes: Are these needed anymore? In general: Try to scrub all the spray dependencies and then remove everything that's related to that.

tests/build.gradle
```diff
@@ -57,6 +57,8 @@ dependencies {
     compile project(':common:scala')
     compile project(':core:controller')
     compile project(':core:invoker')
+
+    testCompile 'com.typesafe.akka:akka-http-testkit_2.11:10.0.9'
```
markusthoemmes: Above deps are under "compile" for a reason: gradle has its problems sharing testDeps throughout projects, having them under compile makes them available to projects depending on tests.

common/scala/build.gradle
dubeejw: From the above, spray-caching hasn't been ported to Akka yet. There is an issue open for it here: https://github.com/akka/akka-http/issues/213. Will look at replacing spray-client.

dubeejw: Looks like spray-client is used by HttpClient.scala, but nothing uses the HttpClient code, so I deleted it.

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
dubeejw: Yes, the Actor requires receive to be override. Tried not inheriting from the Actor, but it looks like we need access to the Actor context in controller.scala

dubeejw:
PG3 873. :large_blue_circle:

core/controller/src/main/scala/whisk/core/controller/API.scala
```diff
@@ -0,0 +1,277 @@
...
+    implicit val activeAckTopicIndex: InstanceId,
+    implicit val actorSystem: ActorSystem,
+    implicit val logging: Logging,
+    implicit val entityStore: EntityStore,
+    implicit val entitlementProvider: EntitlementProvider,
+    implicit val activationIdFactory: ActivationIdGenerator,
+    implicit val loadBalancer: LoadBalancerService,
+    implicit val activationStore: ActivationStore,
+    implicit val consulServer: String,
+    implicit val whiskConfig: WhiskConfig)
```
markusthoemmes: Passing WhiskConfig and consulServer seems redundant. Can we just pass WhiskConfig? Also lifting a String in the implicit scope is dangerous.

The same goes for ActorSystem and ExecutionContext. While not entirely redundant, you can usually use the actorSystem's executor by using `system.dispatcher`. (Note: Only do this in entirely non-blocking scenarios)

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -96,7 +105,7 @@ private case class Context(
 
     // returns true iff the attached query and body parameters contain a property
     // that conflicts with the given reserved parameters
-    def overrides(reservedParams: Set[String]): Set[String] = {
+    def overrides(reservedParams: scala.collection.Set[String]): scala.collection.Set[String] = {
```
markusthoemmes: Why not use the immutable variant?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -251,35 +260,33 @@ protected[core] object WhiskWebActionsApi extends Directives {
...
             }
         } match {
-            case Success((mediaType, data)) =>
+            case Success((mediaType, data: String)) =>
                 respondWithHeaders(headers) {
-                    respondWithMediaType(mediaType) {
-                        complete(code, data)
+                    mediaType.toString.split("/") match {
+                        case Array(mainType, subType) =>
+                            complete(code, HttpEntity(ContentType(MediaType.customWithFixedCharset(mainType, subType, HttpCharsets.`UTF-8`)), data))
```
markusthoemmes: I'm a bit uneasy with the changes to this method.

Do we need to manually split Strings here vs. parsing the data out using akka methods?

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
```diff
@@ -19,56 +19,57 @@ package whisk.http
...
-    implicit def actorRefFactory: ActorContext
-
-    /**
-     * Gets the logging
-     */
+trait BasicHttpService extends Directives with Actor with TransactionCounter {
     implicit def logging: Logging
+    implicit val materializer = ActorMaterializer()
+    implicit val actorSystem = context.system
+    implicit val executionContext: ExecutionContext
```
markusthoemmes: Same as the other comment: you can just go for `implicit val executionContext = actorSystem.dispatcher`

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -124,10 +108,11 @@ class Controller(
 
     // register collections
     Collection.initialize(entityStore)
+    whisk.core.entitlement.Collection.initialize(entityStore)
```
markusthoemmes: Where is this coming from?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
dubeejw: @rabbah, @markusthoemmes, any suggestions on this one? The change is a result of `respondWithMediaType()` being removed from Akka as it was deemed an anti-pattern. 

I've tried using `ContentType.parse` to get the content-type from the headers and passing it into an `HttpEntity`. Ie, `complete(code, HttpEntity(contentType, ByteString.fromString(data)))`. However, test failures result.

core/controller/src/main/scala/whisk/core/controller/API.scala
dubeejw: Was able to remove the redundant values, but got test failures. Will look at further. These redundancies exist in the current implementation of the controller, however.

dubeejw: Since this is the way the current controller implementation is set up, I'd rather open an issue and remove the redundancies as part of a separate PR. Looks like there will have to be test updates  for both scenarios you described here.

rabbah: can we do that first?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
dubeejw: @markusthoemmes, I was able to refactor so that using `complete(code, HttpEntity(contentType, ByteString.fromString(data)))`.

core/controller/src/main/scala/whisk/core/controller/API.scala
dubeejw: Depends if this can be fixed or not:
```
ActionsApiTests.scala:72: ambiguous implicit values:
 both method executor in trait RouteTest of type => scala.concurrent.ExecutionContextExecutor
 and value executionContext in trait SequenceActions of type => scala.concurrent.ExecutionContextExecutor
 match expected type scala.concurrent.ExecutionContext
        waitOnView(entityStore, WhiskAction, namespace, 2)
```

dubeejw: Created an issue here: https://github.com/apache/incubator-openwhisk/issues/2500.

dubeejw:
@rabbah, @markusthoemmes, anymore comments on this one?

dubeejw:
PG4 582 🔵 .

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
```diff
@@ -86,39 +86,57 @@ trait BasicHttpService extends HttpService with TransactionCounter {
...
+            case rejection: UnacceptedResponseContentTypeRejection => true
+            case _ => false
+        }
+
+        priorityRejection.map(rejection => Rejected(Seq(rejection))).getOrElse(Rejected(rejections))
+    }
+
+    val toStrict = mapInnerRoute { innerRoute =>
+        extractRequest { req =>
+            onSuccess(req.toStrict(1.second)) { strictReq =>
```
markusthoemmes: What is the effect of this hardcoded value `1.second`?

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
```diff
@@ -86,39 +86,57 @@ trait BasicHttpService extends HttpService with TransactionCounter {
...
+    val prioritizeRejections = recoverRejections { rejections =>
+        val priorityRejection = rejections.find {
+            case rejection: UnacceptedResponseContentTypeRejection => true
+            case _ => false
+        }
+
+        priorityRejection.map(rejection => Rejected(Seq(rejection))).getOrElse(Rejected(rejections))
+    }
+
+    val toStrict = mapInnerRoute { innerRoute =>
```
markusthoemmes: In general what is the background of this conversion? What's the difference between this and `toStrictEntity`?

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -619,18 +626,33 @@ trait WhiskActionsApi
...
+            Try {
+                value.toInt
+            } match {
+                case Success(i) if i > 0 && i <= max => i.milliseconds
+                case _ => throw new IllegalArgumentException(Messages.invalidTimeout(WhiskActionsApi.maxWaitForBlockingActivation))
             }
         }
     }
+
+    implicit val entityToJsObject: Unmarshaller[HttpEntity, JsObject] = {
```
markusthoemmes: Shouldn't this be solved via: http://doc.akka.io/docs/akka-http/10.0.7/scala/http/common/json-support.html#spray-json-support

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -321,7 +326,21 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
      * @param rule the rule to send
      * @param status the status to include in the response
      */
-    private def completeAsTriggerResponse(trigger: WhiskTrigger): RequestContext => Unit = {
+    private def completeAsTriggerResponse(trigger: WhiskTrigger): RequestContext => Future[RouteResult] = {
         complete(OK, trigger.withoutRules)
     }
+
+    implicit val entityToJsObject: Unmarshaller[HttpEntity, JsObject] = {
```
markusthoemmes: Same as above on standard unmarshallers.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -251,36 +258,31 @@ protected[core] object WhiskWebActionsApi extends Directives {
...
+                    case _ => Failure(RejectRequest(BadRequest, Messages.httpUnknownContentType)(transid))
                 }
             case None => Success(`text/html`)
         }
 
         parsedHeader.flatMap { mediaType =>
-            if (mediaType.binary) {
-                Try(HttpData(Base64.getDecoder().decode(str))).map((mediaType, _))
+            if (mediaType.binary || mediaType == `application/json`) {
+                Try(new String(Base64.getDecoder().decode(str), "UTF-8")).map((mediaType, _))
```
markusthoemmes: drop the hardcoded string in favor of `StandardCharsets.UTF_8`

tests/build.gradle
```diff
@@ -48,11 +48,9 @@ dependencies {
     compile 'junit:junit:4.11'
     compile 'com.jayway.restassured:rest-assured:2.6.0'
     compile 'org.scalatest:scalatest_2.11:3.0.1'
-    compile 'io.spray:spray-testkit_2.11:1.3.3'
-    compile 'com.typesafe.akka:akka-testkit_2.11:2.4.16'
```
markusthoemmes: Why remove the explicit akka testkit? it might be included in akka http but ultimately, they have different release cycles, no?

dubeejw: I removed the old one and bumped the version to `'com.typesafe.akka:akka-http-testkit_2.11:10.0.9'`.

markusthoemmes: You removed `akka-testkit` and added `akka-http-testkit`. Is the second supposed to subsume the first?

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
rabbah: same question for me - also should add a comment.

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
rabbah: can you point out where that is?

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
```diff
@@ -86,39 +86,57 @@ trait BasicHttpService extends HttpService with TransactionCounter {
...
-            }
-        }
-    }
-
     /** Generates log entry for every request. */
     protected def logRequestInfo(req: HttpRequest)(implicit tid: TransactionId): LogEntry = {
         val m = req.method.toString
         val p = req.uri.path.toString
-        val q = req.uri.query.toString
+        val q = req.uri.query().toString
```
rabbah: is calling query side effecting?

common/scala/src/main/scala/whisk/http/ErrorResponse.scala
```diff
@@ -168,7 +168,7 @@ case class ErrorResponse(error: String, code: TransactionId)
 /** Custom rejection, wraps status code for response and a cause. */
 case class CustomRejection private (status: StatusCode, cause: String) extends Rejection
 
-object ErrorResponse extends Directives {
+object ErrorResponse extends Directives with DefaultJsonProtocol {
```
rabbah: do you know why the json serdes is required here?

core/controller/src/main/scala/whisk/core/controller/API.scala
```diff
@@ -0,0 +1,275 @@
+/*
```
rabbah: is this file the equivalent of RestAPI?

rabbah: i suspect so - wonder if we should leave the file name unchanged for now to make a diff easier.

core/controller/src/main/scala/whisk/core/controller/ApiUtils.scala
```diff
@@ -105,7 +107,7 @@ protected[controller] object FilterEntityList {
  * on an operation and terminate the HTTP request.
  */
 package object PostProcess {
-    type PostProcessEntity[A] = A => RequestContext => Unit
+    type PostProcessEntity[A] = A => RequestContext => Future[RouteResult]
```
rabbah: i see these in several places - can you explain the change for the record? thanks!

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -208,8 +191,7 @@ object Controller {
 
         ExecManifest.initialize(config) match {
             case Success(_) =>
-                val port = config.servicePort.toInt
-                BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, InstanceId(instance), logger))
+                BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, InstanceId(instance), logger, port))
```
rabbah: hmm passing port twice - looks odd; is it avoidable?

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -1,292 +0,0 @@
...
-            implicit override val entityStore: EntityStore,
-            override val entitlementProvider: EntitlementProvider,
-            override val activationIdFactory: ActivationIdGenerator,
-            override val loadBalancer: LoadBalancerService,
-            override val consulServer: String,
-            override val executionContext: ExecutionContext,
-            override val logging: Logging,
-            override val whiskConfig: WhiskConfig)
-        extends WhiskPackagesApi with WhiskServices
-}
```
rabbah: as noted above - let's leave the renaming to a separate pr.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -118,7 +125,7 @@ private case class Context(
 
     def toActionArgument(user: Option[Identity], boxQueryAndBody: Boolean): Map[String, JsValue] = {
         val queryParams = if (boxQueryAndBody) {
-            Map(propertyMap.query -> JsString(query.render(new StringRendering, StandardCharsets.UTF_8).get))
+            Map(propertyMap.query -> JsString(query.toString))
```
rabbah: hmm using toString seems dangerous here - is this the best thing to do?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
rabbah: is json considered binary still per the akka types (vs spray) and could we perhaps allow json body in http requests to skip base64 encoding in the future?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -497,25 +499,26 @@ trait WhiskWebActionsApi
                 case Empty =>
                     process(None, isRawHttpAction)
 
-                case NonEmpty(ContentType(`application/json`, _), json) if !isRawHttpAction =>
+                case HttpEntity.Strict(ContentTypes.`application/json`, data) if !isRawHttpAction =>
```
rabbah: data not used - required?

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -503,8 +503,9 @@ object Invoker {
         })
 
         val port = config.servicePort.toInt
+
         BasicHttpService.startService(actorSystem, "invoker", "0.0.0.0", port, new Creator[InvokerServer] {
-            def create = new InvokerServer(invokerInstance, config.invokerInstances.toInt)
+            def create = new InvokerServer(invokerInstance, config.invokerInstances.toInt, port)
```
rabbah: port passed twice (i know to different constructors) - would be nice to avoid if doable.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
dubeejw: I had to add a custom unmarshaller since the functionality in the default one differed from Spray. The issue was that when an empty payload was received, the default unmarshaller would through an error since an empty payload isn't valid JSON.

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
```diff
@@ -72,7 +75,7 @@ class ActivationsApiTests extends ControllerTestCommon with WhiskActivationsApi
         activations foreach { put(activationStore, _) }
         waitOnView(activationStore, namespace, 2)
         whisk.utils.retry {
-            Get(s"$collectionPath") ~> sealRoute(routes(creds)) ~> check {
+            Get(s"$collectionPath") ~> Route.seal(routes(creds)) ~> check {
```
rabbah: maybe to avoid all these changes in this file and others, can you import Route.seal as sealRoute?

tests/src/test/scala/whisk/core/controller/test/TriggersApiTests.scala
```diff
@@ -199,10 +203,10 @@ class TriggersApiTests extends ControllerTestCommon with WhiskTriggersApi {
         implicit val tid = transid()
         val code = "a" * (allowedActivationEntitySize.toInt + 1)
         val content = s"""{"a":"$code"}""".stripMargin
-        Post(s"$collectionPath/${aname}", content.parseJson.asJsObject) ~> sealRoute(routes(creds)) ~> check {
+        Post(s"$collectionPath/${aname}", content.parseJson.asJsObject) ~> Route.seal(routes(creds)) ~> check {
             status should be(RequestEntityTooLarge)
             responseAs[String] should include {
-                Messages.entityTooBig(SizeError(fieldDescriptionForSizeError, (content.length + 5).B, allowedActivationEntitySize.B))
+                Messages.entityTooBig(SizeError(fieldDescriptionForSizeError, (content.length).B, allowedActivationEntitySize.B))
```
rabbah: what made the difference here?

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
rabbah: i'm ok with the change - just have to carefully check all the changes to make sure it's not changing the tests (found 1 test difference on a cursory scan).

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
dubeejw: The `application/json` content-type is no longer considered a binary type in Akka, so we can probably remove the base64 encoding at some point.

rabbah: woot!

tests/build.gradle
dubeejw: Assuming akka-testkit is included in akka-http-testkit since the tests still compile and run. I'll add it back though.

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
dubeejw: Seems that the code won't compile without the parentheses.

```
missing argument list for method query in class Uri
Unapplied methods are only converted to functions when a function type is expected.
You can make this conversion explicit by writing `query _` or `query(_,_)` instead of `query`.
            val query = ctx.request.uri.query

```

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
dubeejw: The query render method has been removed in Akka.

rabbah: can you look at the toString implementation? can we bottom out into where it ends up?

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
dubeejw: Can't seem to comment on your above comment for `def receive = {`. Anyway that is needed for the actor that BasicHttpService extends.

core/controller/src/main/scala/whisk/core/controller/ApiUtils.scala
dubeejw: From the Spray to Akka migration doc (http://doc.akka.io/docs/akka-http/10.0.8/scala/http/migration-guide/migration-from-spray.html): ```
Changes in Route type
Route type has changed from Route = RequestContext ⇒ Unit to Route = RequestContext ⇒ Future[RouteResult]. Which means that now we must complete the Request inside the controller and we can’t simply pass the request to another Actor and complete it there. This has been done intentionally, because in Spray it was easy to forget to complete requests but the code would still compile.
```

File: tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala (hunk 31 above)
Comment (dubeejw) [inline tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala]: What was the test that differed?

File: core/controller/src/main/scala/whisk/core/controller/WebActions.scala (hunk 28 above)
Comment (dubeejw) [inline core/controller/src/main/scala/whisk/core/controller/WebActions.scala]: Looked at the code, the render method wasn't removed, it was made private. Here's what the toString method does: ```
    override def toString = UriRendering.QueryRenderer.render(new StringRendering, this).get
```

tests/build.gradle
dubeejw: Tests failed when I added akka-testkit. It looks like the tests are using `ScalatestRouteTest`, which was moved from akka-testkit to akka-http-testkit, so the former isn't needed in the build.gradle file.

common/scala/src/main/scala/whisk/http/ErrorResponse.scala
dubeejw: Previously we marshaled the ErrorResponse case class ourselves with `                            resp.withEntity(marshalling.marshalUnsafe(ErrorResponse(msg.asString, transid)))`. This `marshalUnsafe` method no longer seems to exist in Akka. Now I am calling `.toJson` on the `ErrorResponse`. 

Ex: 
```
val error = ErrorResponse(ent.data.utf8String, transid).toJson
```

tests/build.gradle
markusthoemmes: It is needed for other tests, we aren't only using akka-http but we're also using plain akka in many places. We need `akka-testkit` as well and depending on it only transiently is dangerous.

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
dubeejw:
```
open/common/scala/src/main/scala/whisk/http/BasicHttpService.scala:58: implicit ActorRefFactory required: if outside of an Actor you need an implicit ActorSystem, inside of an actor this should be the implicit ActorContext
    implicit val materializer = ActorMaterializer()
                                                 ^
open/common/scala/src/main/scala/whisk/http/BasicHttpService.scala:59: not found: value context
    implicit val actorSystem = context.system

```

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
dubeejw: Looks like `toStrictEntity` will work. As for as the timeout goes, the directive will fail if the stream doesn't complete in that about alotted time.

Ran into an error when creating an action with a zip file greater than 1 mb. This is the error message `Substream Source cannot be materialized more than once` There seems to be a bug in akka-http causing the problem, but `toStrictEntity` is being used as a work around right now. Found this info about the error: https://github.com/akka/akka-http/issues/73.

https://github.com/akka/akka-http/issues/73

tests/build.gradle
dubeejw: Got it. Added akka-testkit back to the build.gradle file.

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
dubeejw: @rabbah, @markusthoemmes, I actually think the substream error is due to large payloads being chunked so the `toStrictEntity` will disable chunking. We have a plan to address this in the future anyway.

dubeejw: Here's some supporting evidence: https://github.com/akka/akka-http/issues/750.

dubeejw:
PG3 938 🔵

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -619,18 +626,33 @@ trait WhiskActionsApi
...
+        Unmarshaller.byteStringUnmarshaller.forContentTypes(`application/json`).mapWithCharset { (data, charset) =>
+            val decoded = data.decodeString(charset.nioCharset.name)
+
+            Try {
```
rabbah: perhaps this should check `decoded.length > 0` and decode else return JsObject and skip the try/catch in that case.

markusthoemmes: FWIW, here is the unmarshaller I wrote while debugging performance issues:

```scala
implicit val entityToJsObject: FromEntityUnmarshaller[JsObject] =
        Unmarshaller.byteStringUnmarshaller.forContentTypes(`application/json`).mapWithCharset { (data, charset) =>
            if (data.size == 0) {
                JsObject()
            } else {
                val input =
                    if (charset == HttpCharsets.`UTF-8`) ParserInput(data.toArray)
                    else ParserInput(data.decodeString(charset.nioCharset))
                JsonParser(input).asJsObject
            }
        }
```

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -256,36 +264,31 @@ protected[core] object WhiskWebActionsApi extends Directives {
...
+                    case _ => Failure(RejectRequest(BadRequest, Messages.httpUnknownContentType)(transid))
                 }
             case None => Success(`text/html`)
         }
 
         parsedHeader.flatMap { mediaType =>
-            if (mediaType.binary) {
-                Try(HttpData(Base64.getDecoder().decode(str))).map((mediaType, _))
+            if (mediaType.binary || mediaType == `application/json`) {
+                Try(new String(Base64.getDecoder().decode(str), StandardCharsets.UTF_8)).map((mediaType, _))
```
rabbah: once this is merged - let's open an issue to support json without base64 serdes.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -502,25 +505,26 @@ trait WhiskWebActionsApi
...
                     entity(as[FormData]) { form =>
                         val body = form.fields.toMap.toJson.asJsObject
                         process(Some(body), isRawHttpAction)
                     }
 
-                case NonEmpty(contentType, data) =>
-                    if (contentType.mediaType.binary) {
-                        Try(JsString(Base64.getEncoder.encodeToString(data.toByteArray))) match {
+                case HttpEntity.Strict(contentType, data) =>
+                    // application/json is not a binary type in Akka, but is binary in Spray
```
rabbah: this will make a lot of users happy 🎉

tests/src/test/scala/whisk/core/controller/test/AuthenticateTests.scala
```diff
@@ -113,61 +108,22 @@ class AuthenticateTests extends ControllerTestCommon with Authenticate {
...
-        handleRejections(BasicHttpService.customRejectionHandler) {
-            path("secured") {
-                authenticate(basicauth) {
-                    user => complete("ok")
-                }
-            }
-
-        }
-    }
-
```
rabbah: what are we losing by removing these tests?

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -932,17 +940,17 @@ trait WebActionsApiTests extends ControllerTestCommon with BeforeAndAfterEach wi
                     val largeEntity = "a" * (allowedActivationEntitySize.toInt + 1)
 
                     val content = s"""{"a":"$largeEntity"}"""
-                    Post(s"$testRoutePath/$path", content.parseJson.asJsObject) ~> sealRoute(routes(creds)) ~> check {
+                    Post(s"$testRoutePath/$path", content.parseJson.asJsObject) ~> Route.seal(routes(creds)) ~> check {
                         status should be(RequestEntityTooLarge)
                         val expectedErrorMsg = Messages.entityTooBig(SizeError(
                             fieldDescriptionForSizeError,
-                            (largeEntity.length + 13).B,
+                            (largeEntity.length + 8).B,
```
rabbah: +5B again - guessing quotes are the difference.

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
```diff
@@ -86,39 +86,47 @@ trait BasicHttpService extends HttpService with TransactionCounter {
...
+    }
 
     /** Assigns transaction id to every request. */
     protected val assignId = extract(_ => transid())
 
-    /** Rejection handler to terminate connection on a bad request. Delegates to Spray handler. */
-
-    protected def customRejectionHandler(implicit transid: TransactionId) = RejectionHandler {
-        case rejections => {
-            logging.info(this, s"[REJECT] $rejections")
```
rabbah: i'm ok dropping this - it's of nominal value.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
rabbah: https://github.com/apache/incubator-openwhisk/issues/2606

core/controller/src/main/scala/whisk/core/controller/AuthenticatedRoute.scala
```diff
@@ -38,19 +37,17 @@ trait AuthenticatedRoute {
...
     /** Creates HTTP BasicAuth handler */
-    protected def basicauth(implicit transid: TransactionId) = {
-        new BasicHttpAuthenticator[Identity](realm = "whisk rest service", validateCredentials _) {
-            override def apply(ctx: RequestContext) = {
-                super.apply(ctx) recover {
-                    case t: IllegalStateException => Left(CustomRejection(InternalServerError))
-                    case t                        => Left(CustomRejection(ServiceUnavailable))
-                }
+    def basicAuth[A](verify: Option[BasicHttpCredentials] => Future[Option[A]]) = {
+        authenticateOrRejectWithChallenge[BasicHttpCredentials, A] { creds =>
```
rabbah: the new logic does report an internal server error if the database connection is severed.

rabbah:
@dubeejw great job on getting all the issues sorted out in this migration!
a couple of small nits to address and I'll merge.

core/controller/src/main/scala/whisk/core/controller/Activations.scala
```diff
@@ -184,33 +184,30 @@ trait WhiskActivationsApi
...
-    private implicit val stringToEntityName = new FromStringDeserializer[EntityName] {
-        def apply(value: String): Either[DeserializationError, EntityName] = {
+    /** Custom unmarshaller for query parameters "name" into valid entity name. */
+    private implicit val stringToEntityName: Unmarshaller[String, EntityName] =
+        Unmarshaller.strict[String, EntityName] { value =>
             Try { EntityName(value) } match {
-                case Success(e) => Right(e)
-                case Failure(t) => Left(MalformedContent(t.getMessage))
+                case Success(e) => e
+                case Failure(t) => throw new IllegalArgumentException(s"Parameter is not a valid value for a entity name: $value")
```
rabbah: pushed a commit to do this

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -321,7 +328,19 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
      * @param rule the rule to send
      * @param status the status to include in the response
      */
-    private def completeAsTriggerResponse(trigger: WhiskTrigger): RequestContext => Unit = {
+    private def completeAsTriggerResponse(trigger: WhiskTrigger): RequestContext => Future[RouteResult] = {
         complete(OK, trigger.withoutRules)
     }
+
+    implicit val entityToJsObject: FromEntityUnmarshaller[JsObject] =
```
rabbah: pushed a commit to address this.

core/controller/src/main/scala/whisk/core/controller/AuthenticatedRoute.scala
rabbah: opened https://github.com/apache/incubator-openwhisk/issues/2607

tests/src/test/scala/whisk/core/controller/test/TriggersApiTests.scala
```diff
@@ -214,7 +218,7 @@ class TriggersApiTests extends ControllerTestCommon with WhiskTriggersApi {
             Parameters(key.toString, "a" * 10)
         } reduce (_ ++ _)
         val content = s"""{"parameters":$parameters}""".parseJson.asJsObject
-        Put(s"$collectionPath/${aname}", content) ~> sealRoute(routes(creds)) ~> check {
+        Put(s"$collectionPath/${aname}", content) ~> Route.seal(routes(creds)) ~> check {
```
satybald: just wondering why would want to use curly braces syntax instead of omitting them?

rabbah: Around aname? Historically this may have been something else then copy paste. Should be removed.

satybald: sorry, yep around aname variable.

rabbah: 👍 if you want to remove them and send a pr

satybald: yup, I'll do that ;)

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -173,16 +173,20 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
+                                        headers = List(Authorization(BasicHttpCredentials(user.authkey.uuid.asString, user.authkey.key.asString))),
+                                        entity = HttpEntity(MediaTypes.`application/json`, args.getOrElse(JsObject()).compactPrint))
 
-                                    pipeline(Post(url.withPath(actionUrl + actionPath), args)) onComplete {
-                                        case Success(o) =>
-                                            logging.info(this, s"successfully invoked ${rule.action} -> ${o.fields("activationId")}")
-                                        case Failure(usr: UnsuccessfulResponseException) if usr.response.status == StatusCodes.NotFound =>
+                                    Http().singleRequest(request).map {
+                                        case HttpResponse(StatusCodes.OK, headers, entity, _) =>
+                                            logging.info(this, s"successfully invoked ${rule.action} -> ")
```
rabbah: looks like we missed this - was previously logging the activation id.

rabbah: also looks like something is not quite right here - even after firing the trigger correctly, inspecting the controller logs shows:

```
[2017-08-13T13:48:32.233Z] [WARN] [#tid_7] [TriggersApi] action guest/a could not be invoked due to Source(SourceShape(StatefulMapConcat.out), CompositeModule [3c60a25e] Name: iterableSource Modules: (singleSource) GraphStage(SingleSource(List(ByteString(123, 34, 97, 99, 116, 105, 118, 97, 116, 105, 111, 110, 73, 100, 34, 58, 34, 98, 97, 53, 99, 57, 51, 55, 102, 51, 97, 57, 101, 52, 48, 56, 52, 56, 98, 49, 102, 100, 54, 99, 53, 54, 54, 52, 55, 48, 100, 50, 50, 34, 125)))) [41c5bf50] (unnamed) [703a0c97] copy of GraphStage(StatefulMapConcat) [50915aa5] Downstreams: single.out -> StatefulMapConcat.in Upstreams: StatefulMapConcat.in -> single.out MatValue: Atomic(singleSource[41c5bf50])) 
```

PR #2228, opened 2017-05-10
Hook up reactive pool into invoker startup

markusthoemmes:
**Note for reviewers:** We can (should?) take the flags out before merging this. It's merely there to test the new pool in Travis for now. Alternatively, we can state that nobody should use local/mac in production and thus it's safeish to let the flag in and immediately switch to the new pool. For production environments though, we need more experience with it.

Makes the new InvokerReactive accessible behind a feature flag `invoker_use_reactive_pool`.

This concludes the work on the new containerpool.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
@@ -99,10 +99,10 @@ class Invoker(
         // caching is enabled since actions have revision id and an updated
         // action will not hit in the cache due to change in the revision id;
         // if the doc revision is missing, then bypass cache
-        if (actionid.rev == DocRevision()) {
+        if (actionid.rev == DocRevision.empty) {
             logging.error(this, s"revision was not provided for ${actionid.id}")
         }
-        WhiskAction.get(entityStore, actionid.id, actionid.rev, fromCache = actionid.rev != DocRevision()) onComplete {
+        WhiskAction.get(entityStore, actionid.id, actionid.rev, fromCache = actionid.rev != DocRevision.empty) onComplete {
             case Success(action) =>
                 // only Exec instances that are subtypes of CodeExec reach the invoker
                 assume(action.exec.isInstanceOf[CodeExec[_]])
--- a/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
@@ -34,7 +34,7 @@ import whisk.common.{ Counter, Logging, LoggingMarkers, TransactionId }
 import whisk.common.AkkaLogging
 import whisk.connector.kafka.{ KafkaConsumerConnector, KafkaProducerConnector }
 import whisk.core.WhiskConfig
-import whisk.core.WhiskConfig.{ consulServer, dockerImagePrefix, dockerRegistry, kafkaHost, logsDir, servicePort, whiskVersion }
+import whisk.core.WhiskConfig.{ consulServer, dockerImagePrefix, dockerRegistry, kafkaHost, logsDir, servicePort, whiskVersion, invokerUseReactivePool }
 import whisk.core.connector.{ ActivationMessage, CompletionMessage }
 import whisk.core.container._
 import whisk.core.dispatcher.{ Dispatcher, MessageHandler }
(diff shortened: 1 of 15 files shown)


markusthoemmes:
PG3 595 🔵 (pool disabled)

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -45,6 +45,8 @@ import whisk.http.Messages
 import whisk.utils.ExecutionContextFactory
 import whisk.common.Scheduler
 import whisk.core.connector.PingMessage
+import scala.util.Try
+import whisk.core.connector.MessageProducer
```
rabbah: interesting this wasn't needed before.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -442,12 +443,25 @@ object Invoker {
...
+            } else {
+                new Invoker(config, instance, dispatcher.activationFeed, producer)
+            }
+            logger.info(this, s"using $invoker")
+
             dispatcher.addHandler(invoker, true)
             dispatcher.start()
 
+            Scheduler.scheduleWaitAtMost(1.seconds)(() => {
+                producer.send("health", PingMessage(s"invoker$instance")).andThen {
```
rabbah: should you remove the health pings from old `Invoker` for good measure to avoid double pings if the flag is false?

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -0,0 +1,181 @@
...
+            }
+            Future.sequence(removals)
+        }
+
+        Await.ready(cleaning, 30.seconds)
+    }
+    cleanup()
+    sys.addShutdownHook(cleanup())
+
+    val containerFactory = (tid: TransactionId, name: String, actionImage: ImageName, userProvidedImage: Boolean, memory: ByteSize) => {
```
rabbah: why val not def? (same below)

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -0,0 +1,181 @@
...
+        val actionid = FullyQualifiedEntityName(namespace, name).toDocId.asDocInfo(msg.revision)
+        val tran = Transaction(msg)
+        val subject = msg.user.subject
+
+        logging.info(this, s"${actionid.id} $subject ${msg.activationId}")
+
+        // caching is enabled since actions have revision id and an updated
+        // action will not hit in the cache due to change in the revision id;
+        // if the doc revision is missing, then bypass cache
+        if (actionid.rev == DocRevision()) {
```
rabbah: should probably define a singleton DocRevision.empty

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -0,0 +1,181 @@
...
+            logging.error(this, s"revision was not provided for ${actionid.id}")
+        }
+        WhiskAction.get(entityStore, actionid.id, actionid.rev, fromCache = actionid.rev != DocRevision()).flatMap { action =>
```
rabbah: for future: could push the executable whisk action deserialization into the get.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
markusthoemmes: Note, that `Invoker.scala` also includes the `main` method, hence a couple of changes in that file which seem "strange". The producer is part of the main method and thus is valid for both Invokers.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
markusthoemmes: Same comment on `Invoker.scala`

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
markusthoemmes: Easier to pass the method along. With a `def` you need to explicitly cast it via `containerFactory _`

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
markusthoemmes: Hmm, we could. Not sure if it would be too beneficial though. I imagine the `Future` would then fail with different Exceptions, indicating a missing action or a failed deserialization? As we only need it in so little places, I'd leave it as is, at least for now.

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
markusthoemmes: Done.

PR #2230, opened 2017-05-10
Improve ExecutableWhiskAction coverage and general typesafety

markusthoemmes:
- Improves the implementation of ExecutableWhiskAction to conform closer to existing implementations.
- Use ExecutableWhiskAction as early as possible to statically check only the right types reach the Invoker.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
@@ -72,6 +72,32 @@ case class WhiskActionPut(
     }
 }
 
+abstract class WhiskActionLike(override val name: EntityName) extends WhiskEntity(name) {
+    def exec: Exec
+    def parameters: Parameters
+    def limits: ActionLimits
+
+    /** @return true iff action has appropriate annotation. */
+    def hasFinalParamsAnnotation = {
+        annotations.asBool(WhiskAction.finalParamsAnnotationName) getOrElse false
+    }
+
+    /** @return a Set of immutable parameternames */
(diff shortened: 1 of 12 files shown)


common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -71,6 +71,32 @@ case class WhiskActionPut(
     }
 }
 
+abstract class WhiskActionLike(override val name: EntityName) extends WhiskEntity(name) {
```
rabbah: why did you need to introduce this intermediate type?

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -163,25 +163,30 @@ case class WhiskAction(
...
-    name: EntityName,
+    override val name: EntityName,
     exec: CodeExec[_],
+    parameters: Parameters = Parameters(),
     limits: ActionLimits = ActionLimits(),
     version: SemVer = SemVer(),
-    rev: DocRevision = DocRevision()) {
+    publish: Boolean = false,
+    annotations: Parameters = Parameters())
+    extends WhiskActionLike(name) {
```
rabbah: i see. :/

markusthoemmes: I felt uncomfortable with the difference between `WhiskAction` and `ExecutableWhiskAction` so I opted for a slightly more verbose but satisfying approach.

rabbah:
I'm deferring this until some related PRs that refactor related code are resolved. I can address resultant conflicts at that point.

PR #2232, opened 2017-05-11
make dns server configurable for user containers

jeremiaswerner:


jeremiaswerner:
travis with default, i.e. no DNS server
https://travis-ci.org/openwhisk/openwhisk/builds/231312555?utm_source=github_status&utm_medium=notification

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -73,6 +75,8 @@ object DockerContainer {
             case (key, value) => Seq("-e", s"$key=$value")
         }.flatten
 
+        val dnsArgs = dnsServer.foldLeft(Array.empty[String]) { (a,d) => a ++ Array("--dns", d) }
```
markusthoemmes: Should be equivalent to:

```scala
dnsServers.map(Seq("--dns", _)).flatten
```

(I'd always prefer the more abstract `map` over a relatively lowlevel `foldLeft`).

Also note that `Array` is a mutable/indexed datastructure, thus it's usage should be chosen carefully.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -65,6 +66,7 @@ object DockerContainer {
                cpuShares: Int = 0,
                environment: Map[String, String] = Map(),
                network: String = "bridge",
+               dnsServer: Seq[String] = Seq(),
```
markusthoemmes: Should we make this plural, as its a `Seq`?

markusthoemmes:
Thanks for also updating the new containerpool bits :+1:

jeremiaswerner:
PG1-1537 🔵

PR #2244, opened 2017-05-15
Bump connection-pool limit to 128

markusthoemmes:
This will allow 128 concurrent CouchDB requests in a single connection-pool.


rabbah:
should perhaps open an issue to do request aggregation in the future, using the couchdb bulk document API.

PR #2246, opened 2017-05-15
To resolve #2245 - optimize container removal

### LINKED ISSUE #2245: new ContainerPool aggressively restarts containers under load
## Environment details:

* local deployment, Mac OS
* docker 17.03 (server 17.05 on docker-machine) 

1.   use [https://github.com/markusthoemmes/openwhisk-performance] to run throughput test (tests/throughput.sh)
2.   run throughput test: 
```./throughput.sh https://192.168.99.100  "$(cat ~/git/forks/tysonnorris/openwhisk/ansible/files/auth.guest)" 10 1000```
3.   in a separate terminal check for nodesjsaction containers - notice they are constantly recreated
```docker ps | grep nodejs```

Expected: Don't restart containers when scheduling activation for an action that already has max warm containers that are all busy. Instead, requeue the activation for later.

1. Per above, notice the containers are restarting constantly, even though there is only a single action being exercised

2. Notice throughput latency is way up, compared to 4 concurrent users. 
```
[Mon May 15 2017 20:24:41 GMT+0000 (UTC)] INFO Requests: 0 (0%), requests per second: 0, mean latency: 0 ms
[Mon May 15 2017 20:24:46 GMT+0000 (UTC)] INFO Requests: 10 (1%), requests per second: 2, mean latency: 2904.3 ms
[Mon May 15 2017 20:24:51 GMT+0000 (UTC)] INFO Requests: 20 (2%), requests per second: 2, mean latency: 4631.8 ms
[Mon May 15 2017 20:24:56 GMT+0000 (UTC)] INFO Requests: 28 (3%), requests per second: 2, mean latency: 4242.9 ms
[Mon May 15 2017 20:25:01 GMT+0000 (UTC)] INFO Requests: 36 (4%), requests per second: 2, mean latency: 5887.2 ms
[Mon May 15 2017 20:25:06 GMT+0000 (UTC)] INFO Requests: 50 (5%), requests per second: 3, mean latency: 5270.4 ms
[Mon May 15 2017 20:25:11 GMT+0000 (UTC)] INFO Requests: 60 (6%), requests per second: 2, mean latency: 4091.2 ms
[Mon May 15 2017 20:25:16 GMT+0000 (UTC)] INFO Requests: 70 (7%), requests per second: 2, mean latency: 4556.8 ms
[Mon May 15 2017 20:25:21 GMT+0000 (UTC)] INFO Requests: 84 (8%), requests per second: 3, mean latency: 4444.2 ms
[Mon May 15 2017 20:25:26 GMT+0000 (UTC)] INFO Requests: 94 (9%), requests per second: 2, mean latency: 4195.4 ms
```

n/a

### PR DESCRIPTION
**Pull request by tysonnorris**
Resolves #2245
Only remove a running container to make room for activation in a max'ed out ContainerPool IFF a) the running container is in "Free" state and b) the running container is a DIFFERENT action than the incoming activation

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
@@ -86,7 +86,7 @@ class ContainerPool(
                 } else None
             }.orElse {
                 // Remove a container and create a new one for the given job
-                ContainerPool.remove(r.msg.user.namespace, pool.toMap).map { toDelete =>
+                ContainerPool.remove(r.action, pool.toMap).map { toDelete =>
                     removeContainer(toDelete)
                     createContainer()
                 }
@@ -174,7 +174,7 @@ object ContainerPool {
      * Does not consider pre-warmed containers.
      *
      * @param action the action to run
-     * @param namespace the namespace, that wants to run the action
+     * @param invocationNamespace the namespace, that wants to run the action
      * @param idles a map of idle containers, awaiting work
      * @return a container if one found
      */
@@ -192,13 +192,14 @@ object ContainerPool {
      * takes away one of their containers iff the namespace placing the new job is
      * not already the most consuming one.
      *
-     * @param namespace the namespace that wants to get a container
+     * @param action the action that wants to get a container
      * @param pool a map of all containers in the pool
      * @return a container to be removed iff found
      */
-    def remove[A](invocationNamespace: EntityName, pool: Map[A, WorkerData]): Option[A] = {
+    def remove[A](action: ExecutableWhiskAction, pool: Map[A, WorkerData]): Option[A] = {
+        //try to find a Free container that is initialized with any OTHER action
         val grouped = pool.collect {
-            case (ref, WorkerData(w: WarmedData, _)) => ref -> w
+            case (ref, WorkerData(w: WarmedData, Free)) if (w.action != action) => ref -> w
         }.groupBy {
             case (ref, data) => data.invocationNamespace
         }
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
@@ -200,7 +200,7 @@ object ContainerPool {
     def remove[A](action: ExecutableWhiskAction, invocationNamespace: EntityName, pool: Map[A, WorkerData]): Option[A] = {
         //try to find a Free container that is initialized with any OTHER action
         val grouped = pool.collect {
-            case (ref, WorkerData(w: WarmedData, Free)) if (w.action != action || w.invocationNamespace != invocationNamespace) => ref -> w
+            case (ref, WorkerData(w: WarmedData, _)) if (w.action != action || w.invocationNamespace != invocationNamespace) => ref -> w
         }.groupBy {
             case (ref, data) => data.invocationNamespace
         }
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
@@ -86,7 +86,7 @@ class ContainerPool(
                 } else None
             }.orElse {
                 // Remove a container and create a new one for the given job
-                ContainerPool.remove(r.action, pool.toMap).map { toDelete =>
+                ContainerPool.remove(r.action, r.msg.user.namespace, pool.toMap).map { toDelete =>
                     removeContainer(toDelete)
                     createContainer()
                 }
@@ -193,13 +193,14 @@ object ContainerPool {
      * not already the most consuming one.
      *
      * @param action the action that wants to get a container
+     * @param invocationNamespace the namespace, that wants to run the action
      * @param pool a map of all containers in the pool
      * @return a container to be removed iff found
      */
-    def remove[A](action: ExecutableWhiskAction, pool: Map[A, WorkerData]): Option[A] = {
+    def remove[A](action: ExecutableWhiskAction, invocationNamespace: EntityName, pool: Map[A, WorkerData]): Option[A] = {
         //try to find a Free container that is initialized with any OTHER action
         val grouped = pool.collect {
-            case (ref, WorkerData(w: WarmedData, Free)) if (w.action != action) => ref -> w
+            case (ref, WorkerData(w: WarmedData, Free)) if (w.action != action || w.invocationNamespace != invocationNamespace) => ref -> w
         }.groupBy {
             case (ref, data) => data.invocationNamespace
         }
--- a/tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
+++ b/tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
@@ -281,13 +281,15 @@ class ContainerPoolTests extends TestKit(ActorSystem("ContainerPool"))
 class ContainerPoolObjectTests extends FlatSpec with Matchers with MockFactory {
 
     val actionExec = CodeExecAsString(RuntimeManifest("actionKind", ImageName("testImage")), "testCode", None)
+    val standardNamespace = EntityName("standardNamespace")
+    val differentNamespace = EntityName("differentNamespace")
 
     /** Helper to create a new action from String representations */
     def createAction(namespace: String = "actionNS", name: String = "actionName") =
         ExecutableWhiskAction(EntityPath(namespace), EntityName(name), actionExec)
 
     /** Helper to create WarmedData with sensible defaults */
-    def warmedData(action: ExecutableWhiskAction = createAction(), namespace: String = "anyNamespace", lastUsed: Instant = Instant.now) =
+    def warmedData(action: ExecutableWhiskAction = createAction(), namespace: String = standardNamespace.asString, lastUsed: Instant = Instant.now) =
         WarmedData(stub[Container], EntityName(namespace), action, lastUsed)
 
     /** Helper to create PreWarmedData with sensible defaults */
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by tysonnorris** [comment]
Need to fix some tests...

File: core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala [hunk 1]
```diff
@@ -192,13 +192,14 @@ object ContainerPool {
      * takes away one of their containers iff the namespace placing the new job is
      * not already the most consuming one.
      *
-     * @param namespace the namespace that wants to get a container
+     * @param action the action that wants to get a container
      * @param pool a map of all containers in the pool
      * @return a container to be removed iff found
      */
-    def remove[A](invocationNamespace: EntityName, pool: Map[A, WorkerData]): Option[A] = {
+    def remove[A](action: ExecutableWhiskAction, pool: Map[A, WorkerData]): Option[A] = {
```
rabbah: shouldn't this be predicated on both the subject and the action name?

tysonnorris: I thought the container should be reusable across same actions regardless of subject?

rabbah: containers are specific to a subject+action; we don't reuse a container to run an action for more than one subject since we assume the container is tainted (ok to self taint but shall not taint another user or action).

tysonnorris: I see, will take a look after I fix the tests

tysonnorris: @rabbah I don't see this (isolation by subject) implemented the new ContainerPool currently, I expect it would need to be added to WarmedData - should I add that as part of this PR, or separate?

rabbah: See ContainerPool.schedule.

tysonnorris: oh...sorry, was mixing up user namespace and action namespace... will update to address this

rabbah: heh, I had suggested retyping that to `Identity` so it's more obvious.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -192,13 +192,15 @@ object ContainerPool {
...
+     * @param invocationNamespace the namespace, that wants to run the action
      * @param pool a map of all containers in the pool
      * @return a container to be removed iff found
      */
-    def remove[A](invocationNamespace: EntityName, pool: Map[A, WorkerData]): Option[A] = {
+    def remove[A](action: ExecutableWhiskAction, invocationNamespace: EntityName, pool: Map[A, WorkerData]): Option[A] = {
+        //try to find a Free container that is initialized with any OTHER action
         val grouped = pool.collect {
-            case (ref, WorkerData(w: WarmedData, _)) => ref -> w
+            case (ref, WorkerData(w: WarmedData, Free)) if (w.action != action || w.invocationNamespace != invocationNamespace) => ref -> w
```
markusthoemmes: Do we want to only collect `Free` containers? I'd claim no, since the calculation below searches for the most consuming namespace. If that namespace is inducing very high load, it can still monopolize the system if we are not able to claim containers back because that namespace having only `Busy` containers all the time.

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
```diff
@@ -148,11 +169,43 @@ class ContainerPoolTests extends TestKit(ActorSystem("ContainerPool"))
...
+      containers(0).expectMsg(runMessage)
+      containers(0).send(pool, NeedWork(warmedData()))
+      feed.expectMsg(ContainerReleased)
+      pool ! runMessage1b
+      containers(0).expectMsg(Remove)
+      containers(1).expectMsg(runMessage1b)
+    }
+
+
+    it should "not remove a container to make space in the pool if it is already full and the same action + same invocation namespace arrives" in within(timeout) {
```
markusthoemmes: This test is essentially a duplicate of the below, no? I think we shouldn't test the internals of `schedule` and `remove` too much in the more general `ContainerPoolTests`. Thoughts?

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
```diff
@@ -78,8 +81,26 @@ class ContainerPoolTests extends TestKit(ActorSystem("ContainerPool"))
         ActivationId(),
         invocationNamespace.toPath,
         None)
+    val messageb = ActivationMessage(
```
markusthoemmes: Can we give those more meaningful names? Like `differentNamespaceMessage` and `differentMessage`?

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
```diff
@@ -68,7 +69,9 @@ class ContainerPoolTests extends TestKit(ActorSystem("ContainerPool"))
     val memoryLimit = 256.MB
 
     val invocationNamespace = EntityName("invocationSpace")
+    val invocationNamespace2 = EntityName("invocationSpace2")
```
markusthoemmes: Would a more speaking name make sense? Like `differentInvocatioNamespace`?

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
```diff
@@ -68,7 +69,9 @@ class ContainerPoolTests extends TestKit(ActorSystem("ContainerPool"))
...
     val action = ExecutableWhiskAction(EntityPath("actionSpace"), EntityName("actionName"), exec)
+    val action2 = ExecutableWhiskAction(EntityPath("actionSpace"), EntityName("actionName2"), exec)
```
markusthoemmes: Same comment on naming. You can also write:

```
val differentAction = action.copy(name = EntityName("actionName2"))
```

Makes it a bit more apparent that you just change the name.

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
```diff
@@ -78,8 +81,26 @@ class ContainerPoolTests extends TestKit(ActorSystem("ContainerPool"))
...
+      action2.fullyQualifiedName(true),
+      action2.rev,
+      Identity(Subject(), invocationNamespace2, AuthKey(), Set()),
+      ActivationId(),
+      invocationNamespace2.toPath,
+      None)
 
     val runMessage = Run(action, message)
+    val runMessage1b = Run(action, messageb)
+    val runMessage2 = Run(action2, message2)
```
markusthoemmes: Same comment on naming.

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
```diff
@@ -365,7 +430,7 @@ class ContainerPoolObjectTests extends FlatSpec with Matchers with MockFactory {
             'second -> freeWorker(second),
             'oldest -> freeWorker(oldest))
 
-        ContainerPool.remove(EntityName(commonNamespace), pool) shouldBe Some('oldest)
+        ContainerPool.remove(first.action, EntityName("anyNamespace"), pool) shouldBe Some('oldest)
```
markusthoemmes: Should use `createAction()`. The relationship to the first warmed data is not necessary here.

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
```diff
@@ -381,6 +446,6 @@ class ContainerPoolObjectTests extends FlatSpec with Matchers with MockFactory {
             'largeYoung -> freeWorker(warmedData(namespace = largeNamespace, lastUsed = Instant.ofEpochMilli(3))),
             'largeOld -> freeWorker(warmedData(namespace = largeNamespace, lastUsed = Instant.ofEpochMilli(2))))
 
-        ContainerPool.remove(myData.invocationNamespace, pool) shouldBe Some('largeOld)
+        ContainerPool.remove(myData.action, EntityName("anyNamespace"), pool) shouldBe Some('largeOld)
```
markusthoemmes: Same comment as above. Could use `createAction()` as there is no direct relationship to the data.

markusthoemmes:
Thank you so much for finding and fixing this. Great job! 🎉 
Would you please install our formatter for scala (located here: https://github.com/apache/incubator-openwhisk/tree/master/tools/eclipse) so we have consistent spacing in place.

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
```diff
@@ -20,6 +20,7 @@ import java.time.Instant
 
 import scala.collection.mutable
 import scala.concurrent.duration._
+import scala.language.postfixOps
```
markusthoemmes: You'll come to hate `postfixOps` very soon because it induces strange compilation behavior. Personally I avoid it alltogether.

markusthoemmes:
PG1 1540 🔵

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
tysonnorris: Thanks for the updates, and the tips!

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
tysonnorris: below test ends with exectNoMsg() instead of expectMsg() - testing cases where removal result is expected to be None vs Some based on whether this is a different action/invocation namespace

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
tysonnorris: I see, yes Free and Busy should be collected, BUT only Free should be eligible for removal - it looks like current impl would allow a Busy to be removed, although unlikely since the lastUsed should be recent.

markusthoemmes: No, actually Busy should be removed as well.

PR #2251, opened 2017-05-17
Correctly recover from errors when fetching an action

markusthoemmes:
If the invoker fails to fetch an action from the database it needs to inform the ActivationFeed that it still has resources (it didn't consume a container after all).

Reporting of those errors needs some disambiguation, as an error on fetching the action could also be caused by the user (for example by concurrently deleting the action while invoking it). The InvokerHealth protocol would shutdown the invoker, iff the activation was reported as a WHISK_ERROR. All errors but DocumentNotFound are considered WHISK_ERRORs though.

Also in this commit:
- Loglevel change from error to warn for a missing revision. The InvokerHealth protocol produces ERROR otherwise.
- Some documentation and restructuring of the ContainerPool's setup.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
@@ -46,6 +46,10 @@ import whisk.core.dispatcher.MessageHandler
 import whisk.core.entity._
 import whisk.core.entity.ExecManifest.ImageName
 import whisk.core.entity.size._
+import whisk.core.dispatcher.ActivationFeed.ContainerReleased
+import whisk.core.containerpool.ContainerPool
+import whisk.core.database.NoDocumentException
+import whisk.http.Messages
 
 class InvokerReactive(
     config: WhiskConfig,
@@ -67,6 +71,7 @@ class InvokerReactive(
         val cleaning = docker.ps(Seq("name" -> "wsk_"))(TransactionId.invokerNanny).flatMap { containers =>
             val removals = containers.map { id =>
                 runc.resume(id)(TransactionId.invokerNanny).recoverWith {
+                    // Ignore resume failures and try to remove anyway
                     case _ => Future.successful(())
                 }.flatMap {
                     _ => docker.rm(id)(TransactionId.invokerNanny)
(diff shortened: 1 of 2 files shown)


core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -144,7 +154,7 @@ class InvokerReactive(
         // action will not hit in the cache due to change in the revision id;
         // if the doc revision is missing, then bypass cache
         if (actionid.rev == DocRevision.empty) {
-            logging.error(this, s"revision was not provided for ${actionid.id}")
+            logging.warn(this, s"revision was not provided for ${actionid.id}")
```
rabbah: It is an error to receive a request to run an action without a revision. Under which circumstances would it happen and it's not an error in the current architecture?

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -156,7 +166,14 @@ class InvokerReactive(
...
             }
         }.recover {
-            case _ =>
+            case t =>
+                // If the action cannot be found, the user has concurrently deleted it,
+                // making this an application error. All other errors are considered system
+                // errors and should cause the invoker to be considered unhealthy.
+                val response = t match {
+                    case _: NoDocumentException => ActivationResponse.applicationError("action could not be found")
+                    case _                      => ActivationResponse.whiskError("action could not be found")
```
rabbah: Since these are user visible messages, maybe put them in the Messages singleton? I started some time ago to consolidate all messages there which in the future we might want to internationalize. (Is there a test of this case?)

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
markusthoemmes: See description. The InvokerHealth protocol deliberately sends messages without revision to force db access, which produces error logs.

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
markusthoemmes: :+1: on putting it in Messages.

That part of the Invoker is not testable currently.

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
rabbah: ack.

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
rabbah: make invoker great again markus :)

markusthoemmes:
PG1 1546 🔵

Issue #2254, opened 2017-05-17
Reevaluate the need of Consul in our deployment


markusthoemmes:
Consul is supposed to help with service-discovery, service-monitoring, DNS and thelike. We use none of it.

Today we store our configuration in Consul. The Controller also writes a couple of statistics to Consul which can be used to monitor the internals of the system. The fact that we store configuration in Consul has two major drawbacks:

1. The deployment tool (ansible) cannot know, that configuration for Controller/Invoker have changed. It therefore cannot automatically determine they need a redeployment.
2. Other tools like Kubernetes or Docker Compose have a hard time keeping up with changes in the main repository.

I therefore propose to remove Consul from our deployment completely. We should pass configuration via environment. Monitoring can be done via an internal API on the Controller.

Steps to do:

- [ ] Configure Controller/Invoker through environment variables. That is possible today, even with `WhiskConfig`.
- [ ] Create a `/queue` route in the Controller, just like `/invokers` to monitor the Controller's view of the queue.

rabbah:
@markusthoemmes consider a consolidated /admin route prefix where we might add other details in the future or perform administrative tasks as well; this could serve a single authenticated route for operators.

tysonnorris:
Can registrator go away with consul?

rabbah:
indeed. Two for one (pull request).

ddragosd:
how about `whisk.properties` vs ENV VARS ?

markusthoemmes:
`whisk.properties` have no benefit ansible wise! We need something ansible is aware of, which is only env vars if you ask me.

markusthoemmes:
Done.

imesh:
@markusthoemmes As I see still the architecture diagram on [about.md](https://github.com/apache/incubator-openwhisk/blob/master/docs/about.md) has Consul. Should we remove it from that as well?

rabbah:
yes we should but we need the sources; related to #3208.

PR #2256, opened 2017-05-17
surface errors in runtimes.manifest configuration in the controller logs

ddragosd:
This PR ensures than when `runtimes.manifest` property is incorrectly configured that the controller prints what the error is in the logs. 

Without this PR we would only see in the logs:
```
[2017-05-17T21:21:53.457Z] [ERROR] [??] [Controller] Bad configuration, cannot start. 
```

With this PR we can see:
```
Exception in thread "main" java.lang.IllegalStateException: Runtimes manifest is set but it's invalid: spray.json.DeserializationException: Object is missing required member 'image'
	at whisk.core.entity.ExecManifest$.initialize(ExecManifest.scala:55)
	at whisk.core.controller.Controller$.main(Controller.scala:185)
	at whisk.core.controller.Controller.main(Controller.scala)
```

This is very helpful when debugging issues with the configuration such as the one mentioned at https://github.com/apache/incubator-openwhisk-devtools/issues/29

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
@@ -16,9 +16,7 @@
 
 package whisk.core.entity
 
-import scala.util.Try
-import scala.util.Failure
-
+import scala.util.{Failure, Success, Try}
 import spray.json._
 import spray.json.DefaultJsonProtocol._
 import whisk.core.WhiskConfig
@@ -47,10 +45,15 @@ protected[core] object ExecManifest {
      */
     protected[core] def initialize(config: WhiskConfig, reinit: Boolean = false): Boolean = {
         if (manifest.isEmpty || reinit) {
-            Try(config.runtimesManifest.parseJson.asJsObject)
+            val result = Try(config.runtimesManifest.parseJson.asJsObject)
                 .flatMap(runtimes(_))
                 .map(m => manifest = Some(m))
-                .isSuccess
+            result match {
+                case Success(res) =>
+                    true
+                case Failure(e) =>
+                    throw new IllegalStateException("Runtimes manifest is set but it's invalid: " + e.toString)
+            }
         } else true
     }
 
--- a/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
@@ -43,12 +43,12 @@ protected[core] object ExecManifest {
      * @param reinit re-initialize singleton iff true
      * @return the manifest if initialized successfully, or if previously initialized
      */
-    protected[core] def initialize(config: WhiskConfig, reinit: Boolean = false): Try[_] = {
+    protected[core] def initialize(config: WhiskConfig, reinit: Boolean = false): Try[Runtimes] = {
         if (manifest.isEmpty || reinit) {
-            Try(config.runtimesManifest.parseJson.asJsObject)
-                .flatMap(runtimes(_))
-                .map(m => manifest = Some(m))
-        } else Success(manifest)
+            val mf = Try(config.runtimesManifest.parseJson.asJsObject).flatMap(runtimes(_))
+            mf.foreach(m => manifest = Some(m))
+            mf
+        } else Success(manifest.get)
     }
 
     /**
--- a/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
@@ -41,20 +41,14 @@ protected[core] object ExecManifest {
      *
      * @param config a valid configuration
      * @param reinit re-initialize singleton iff true
-     * @return true if initialized successfully, or if previously initialized
+     * @return the manifest if initialized successfully, or if previously initialized
      */
-    protected[core] def initialize(config: WhiskConfig, reinit: Boolean = false): Boolean = {
+    protected[core] def initialize(config: WhiskConfig, reinit: Boolean = false): Try[_] = {
         if (manifest.isEmpty || reinit) {
-            val result = Try(config.runtimesManifest.parseJson.asJsObject)
+            Try(config.runtimesManifest.parseJson.asJsObject)
                 .flatMap(runtimes(_))
                 .map(m => manifest = Some(m))
-            result match {
-                case Success(res) =>
-                    true
-                case Failure(e) =>
-                    throw new IllegalStateException("Runtimes manifest is set but it's invalid: " + e.toString)
-            }
-        } else true
+        } else Success(manifest)
     }
 
     /**
--- a/core/controller/src/main/scala/whisk/core/controller/Controller.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Controller.scala
@@ -182,12 +182,14 @@ object Controller {
 
         // initialize the runtimes manifest
         val execManifest = ExecManifest.initialize(config)
+        if (execManifest.isFailure) {
+            logger.error(this, "Required property runtimes.manifest is invalid: " + execManifest.failed.get.toString)
+        }
         if (config.isValid && execManifest.isSuccess) {
             val port = config.servicePort.toInt
             BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, instance, logger))
         } else {
-            logger.error(this, "Bad configuration, cannot start." +
-              (if (execManifest.isFailure) " Invalid runtimes.manifest: " + execManifest.failed.get.toString else ""))
+            logger.error(this, "Bad configuration, cannot start.")
             actorSystem.terminate()
             Await.result(actorSystem.whenTerminated, 30.seconds)
         }
--- a/core/controller/src/main/scala/whisk/core/controller/Controller.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Controller.scala
@@ -183,11 +183,11 @@ object Controller {
             logger.error(this, "Bad configuration, cannot start.")
             actorSystem.terminate()
             Await.result(actorSystem.whenTerminated, 30.seconds)
+            sys.exit(1)
         }
 
         if (!config.isValid) {
             abort()
-            return
         }
 
         ExecManifest.initialize(config) match {
--- a/core/controller/src/main/scala/whisk/core/controller/Controller.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Controller.scala
@@ -181,11 +181,13 @@ object Controller {
         val instance = if (args.length > 0) args(1).toInt else 0
 
         // initialize the runtimes manifest
-        if (config.isValid && ExecManifest.initialize(config)) {
+        val execManifest = ExecManifest.initialize(config)
+        if (config.isValid && execManifest.isSuccess) {
             val port = config.servicePort.toInt
             BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, instance, logger))
         } else {
-            logger.error(this, "Bad configuration, cannot start.")
+            logger.error(this, "Bad configuration, cannot start." +
+              (if (execManifest.isFailure) " Invalid runtimes.manifest: " + execManifest.failed.get.toString else ""))
             actorSystem.terminate()
             Await.result(actorSystem.whenTerminated, 30.seconds)
         }
--- a/core/controller/src/main/scala/whisk/core/controller/Controller.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Controller.scala
@@ -180,18 +180,24 @@ object Controller {
         // second argument.  (TODO .. seems fragile)
         val instance = if (args.length > 0) args(1).toInt else 0
 
-        // initialize the runtimes manifest
-        val execManifest = ExecManifest.initialize(config)
-        if (execManifest.isFailure) {
-            logger.error(this, "Required property runtimes.manifest is invalid: " + execManifest.failed.get.toString)
-        }
-        if (config.isValid && execManifest.isSuccess) {
-            val port = config.servicePort.toInt
-            BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, instance, logger))
-        } else {
+        def abort() = {
             logger.error(this, "Bad configuration, cannot start.")
             actorSystem.terminate()
             Await.result(actorSystem.whenTerminated, 30.seconds)
         }
+
+        if (!config.isValid) {
+            abort()
+            return
+        }
+
+        val execManifest = ExecManifest.initialize(config)
+        if (execManifest.isFailure) {
+            logger.error(this, s"Invalid runtimes manifest: ${execManifest.failed.get}")
+            abort()
+            return
+        }
+        val port = config.servicePort.toInt
+        BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, instance, logger))
     }
 }
--- a/core/controller/src/main/scala/whisk/core/controller/Controller.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Controller.scala
@@ -18,12 +18,10 @@ package whisk.core.controller
 
 import scala.concurrent.Await
 import scala.concurrent.duration.DurationInt
-
 import akka.actor.Actor
 import akka.actor.ActorContext
 import akka.actor.ActorSystem
 import akka.japi.Creator
-
 import spray.http.StatusCodes._
 import spray.http.Uri
 import spray.httpx.SprayJsonSupport._
@@ -31,7 +29,6 @@ import spray.json._
 import spray.json.DefaultJsonProtocol._
 import spray.routing.Directive.pimpApply
 import spray.routing.Route
-
 import whisk.common.AkkaLogging
 import whisk.common.Logging
 import whisk.common.TransactionId
@@ -45,6 +42,8 @@ import whisk.core.loadBalancer.LoadBalancerService
 import whisk.http.BasicHttpService
 import whisk.http.BasicRasService
 import whisk.common.LoggingMarkers
+
+import scala.util.{Failure, Success}
 
 /**
  * The Controller is the service that provides the REST API for OpenWhisk.
@@ -191,13 +190,14 @@ object Controller {
             return
         }
 
-        val execManifest = ExecManifest.initialize(config)
-        if (execManifest.isFailure) {
-            logger.error(this, s"Invalid runtimes manifest: ${execManifest.failed.get}")
-            abort()
-            return
+        ExecManifest.initialize(config) match {
+            case Success(_) =>
+                val port = config.servicePort.toInt
+                BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, instance, logger))
+
+            case Failure(t) =>
+                logger.error(this, s"Invalid runtimes manifest: $t")
+                abort()
         }
-        val port = config.servicePort.toInt
-        BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, instance, logger))
     }
 }
--- a/tests/src/test/scala/whisk/core/entity/test/ExecHelpers.scala
+++ b/tests/src/test/scala/whisk/core/entity/test/ExecHelpers.scala
@@ -33,7 +33,7 @@ trait ExecHelpers
     self: Suite =>
 
     private val config = new WhiskConfig(ExecManifest.requiredProperties)
-    ExecManifest.initialize(config) shouldBe true
+    ExecManifest.initialize(config).isSuccess shouldBe true
 
     protected val NODEJS = "nodejs"
     protected val NODEJS6 = "nodejs:6"
--- a/tests/src/test/scala/whisk/core/entity/test/ExecHelpers.scala
+++ b/tests/src/test/scala/whisk/core/entity/test/ExecHelpers.scala
@@ -33,7 +33,7 @@ trait ExecHelpers
     self: Suite =>
 
     private val config = new WhiskConfig(ExecManifest.requiredProperties)
-    ExecManifest.initialize(config).isSuccess shouldBe true
+    ExecManifest.initialize(config) should be a 'success
 
     protected val NODEJS = "nodejs"
     protected val NODEJS6 = "nodejs:6"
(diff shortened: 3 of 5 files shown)


rabbah:
You can change the method to return the try and then throw the exception in the controller's main.

ddragosd:
> You can change the method to return the try and then throw the exception in the controller's main.

`ExecManifest.initialize` is called by both: controller and invoker. would you suggest we impl in both of them ?

Open to your suggestion, but before making the change I want to make sure we're ok to duplicate the code.

> There's no point to using a try if this method is going to throw an exception.

 I also thought of that but I changed my mind seeing we have 3 operations to `Try`: `parse`, `flatMap`, and `map`, and the intent is to throw an exception only on error.

Let me know what you think.

rabbah:
The operation that will throw is the json parse; hence removing the try and documenting that the method will throw is all the same. I slightly prefer the return of a Try so it's explicitly known through the types this method can fail and the caller should address failure; yes you'd have to throw in both controller/invoker main but really it would be achieved by just doing a .get in the result for example hence not much duplication at all. But it's fine to just throw here and document it. You will need to adjust some tests iirc.

ddragosd:
>  I slightly prefer the return of a Try so it's explicitly known through the types this method can fail and the caller should address failure

👍 thanks @rabbah . I'll make the changes to be more explicit.

ddragosd:
@rabbah I'm done with the changes based on your review. I've also added a unit test with an invalid `runtimes.manifest` value.

common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
```diff
@@ -41,20 +41,14 @@ protected[core] object ExecManifest {
...
                 .flatMap(runtimes(_))
                 .map(m => manifest = Some(m))
-            result match {
-                case Success(res) =>
-                    true
-                case Failure(e) =>
-                    throw new IllegalStateException("Runtimes manifest is set but it's invalid: " + e.toString)
-            }
-        } else true
+        } else Success(manifest)
```
rabbah: how about this alternate:

```git
diff --git a/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala b/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
index 36bc262..71e3971 100644
--- a/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
@@ -16,7 +16,7 @@
 
 package whisk.core.entity
 
-import scala.util.{Failure, Success, Try}
+import scala.util.{ Failure, Success, Try }
 import spray.json._
 import whisk.core.WhiskConfig
@@ -43,12 +43,12 @@ protected[core] object ExecManifest {
      * @param reinit re-initialize singleton iff true
      * @return the manifest if initialized successfully, or if previously initialized
      */
-    protected[core] def initialize(config: WhiskConfig, reinit: Boolean = false): Try[_] = {
+    protected[core] def initialize(config: WhiskConfig, reinit: Boolean = false): Try[Runtimes] = {
         if (manifest.isEmpty || reinit) {
-            Try(config.runtimesManifest.parseJson.asJsObject)
-                .flatMap(runtimes(_))
-                .map(m => manifest = Some(m))
-        } else Success(manifest)
+            val mf = Try(config.runtimesManifest.parseJson.asJsObject).flatMap(runtimes(_))
+            mf.foreach(m => manifest = Some(m))
+            mf
+        } else Success(manifest.get)
```

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -181,11 +181,13 @@ object Controller {
...
         // initialize the runtimes manifest
-        if (config.isValid && ExecManifest.initialize(config)) {
+        val execManifest = ExecManifest.initialize(config)
+        if (config.isValid && execManifest.isSuccess) {
             val port = config.servicePort.toInt
             BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, instance, logger))
         } else {
-            logger.error(this, "Bad configuration, cannot start.")
+            logger.error(this, "Bad configuration, cannot start." +
+              (if (execManifest.isFailure) " Invalid runtimes.manifest: " + execManifest.failed.get.toString else ""))
```
rabbah: how about making this a separate log message?

```
-        // initialize the runtimes manifest
-        val execManifest = ExecManifest.initialize(config)
-        if (config.isValid && execManifest.isSuccess) {
-            val port = config.servicePort.toInt
-            BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, instance, logger))
-        } else {
-            logger.error(this, "Bad configuration, cannot start." +
-              (if (execManifest.isFailure) " Invalid runtimes.manifest: " + execManifest.failed.get.toString else ""))
+        def abort() = {
+            logger.error(this, "Bad configuration, cannot start.")
             actorSystem.terminate()
             Await.result(actorSystem.whenTerminated, 30.seconds)
         }
+
+        // initialize the runtimes manifest
+        if (config.isValid) {
+            val execManifest = ExecManifest.initialize(config)
+            if (execManifest.isSuccess) {
+                val port = config.servicePort.toInt
+                BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, instance, logger))
+            } else {
+                logger.error(this, s"Invalid runtimes manifest: ${execManifest.failed.get}")
+                abort()
+            }
+        } else abort()
```

Note that pulling the exec manifest initialization to precede the configuration check changes the behavior on which error is caught/reported first.

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
```diff
@@ -170,4 +176,22 @@ class ExecManifestTests
...
+    it should "throw an error when configured manifest is invalid" in {
+        val config_manifest = "{\"nodejs\":[{\"kind\":\"nodejs\",\"image\":{\"name\":\"nodejsaction\"},\"deprecated\":true},{\"kind\":\"nodejs:6\",\"default\":true,\"image\":{\"name\":\"nodejs6action\"}}]}"
+        val file = File.createTempFile("cxt", ".txt")
+        file.deleteOnExit()
+
+        val bw = new BufferedWriter(new FileWriter(file))
+        bw.write("runtimes.manifest=" + config_manifest + "\n")
+        bw.close()
+
+        var result:Try[_] = ExecManifest.initialize(new WhiskConfig(Map("runtimes.manifest" -> null), Set(), file), true)
```
rabbah: a concern with this test is that WhiskConfig will attempt to read from consul (and will succeed if it's deployed and initialized). So will this test work given that the value read last will supersede?

ddragosd: Good catch. I could change this to prevent consul, but at the same time I have another PR in the works to remove consul from WhiskConfig. In this light should we worry about this in the test ?

core/controller/src/main/scala/whisk/core/controller/Controller.scala
ddragosd: Yes, I could make it a separate log message to appear before the line saying `Bad configuration, cannot start.` . The other failures in config with required properties appear before too.

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
rabbah: I was hoping you'd ask that question :) I can live without this test. Or once consul is removed, adding this test back is OK.

common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
```diff
@@ -43,15 +41,14 @@ protected[core] object ExecManifest {
...
      */
-    protected[core] def initialize(config: WhiskConfig, reinit: Boolean = false): Boolean = {
+    protected[core] def initialize(config: WhiskConfig, reinit: Boolean = false): Try[_] = {
         if (manifest.isEmpty || reinit) {
             Try(config.runtimesManifest.parseJson.asJsObject)
                 .flatMap(runtimes(_))
                 .map(m => manifest = Some(m))
-                .isSuccess
-        } else true
+        } else Success(manifest)
```
rabbah: Consider this alternate implementation - it types the Try more strictly:

```
diff --git a/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala b/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
index 36bc262..71e3971 100644
--- a/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
@@ -16,7 +16,7 @@
 
 package whisk.core.entity
 
-import scala.util.{Failure, Success, Try}
+import scala.util.{ Failure, Success, Try }
 import spray.json._
 import whisk.core.WhiskConfig
@@ -43,12 +43,12 @@ protected[core] object ExecManifest {
      * @param reinit re-initialize singleton iff true
      * @return the manifest if initialized successfully, or if previously initialized
      */
-    protected[core] def initialize(config: WhiskConfig, reinit: Boolean = false): Try[_] = {
+    protected[core] def initialize(config: WhiskConfig, reinit: Boolean = false): Try[Runtimes] = {
         if (manifest.isEmpty || reinit) {
-            Try(config.runtimesManifest.parseJson.asJsObject)
-                .flatMap(runtimes(_))
-                .map(m => manifest = Some(m))
-        } else Success(manifest)
+            val mf = Try(config.runtimesManifest.parseJson.asJsObject).flatMap(runtimes(_))
+            mf.foreach(m => manifest = Some(m))
+            mf
+        } else Success(manifest.get)
```

core/controller/src/main/scala/whisk/core/controller/Controller.scala
ddragosd: Done.

> Note that pulling the exec manifest initialization to precede the configuration check changes the behavior on which error is caught/reported first. 

Right. Now we see the errors like this:
```
[2017-05-18T20:13:57.743Z] [ERROR] [??] [Controller] Required property runtimes.manifest is invalid: java.util.NoSuchElementException: key not found: runtimes 
[2017-05-18T20:13:57.743Z] [ERROR] [??] [Config] required property whisk.version.buildno still not set
```

If there's a preference on the order, let me know.

ddragosd: I just realized I missed the diff. I'll apply it, please mind my prev commit.

ddragosd:
@rabbah thank you very much for the suggestions. I've applied them all.
Only a small diff on `else` (I personally try to live without it to save indentations, and make the code easier to follow).

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
ddragosd: > I can live without this test

I will 👍 this b/c removing Consul it's WIP for me, coming up before EOD.

ddragosd:
@rabbah when you get the chance let me know if there's any other change you'd like before merging.

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
rabbah: are you going to remove this test in this pr?

rabbah:
Are you going to remove or ignore the new test or wait until your consul PR is in?

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -180,14 +180,24 @@ object Controller {
...
             logger.error(this, "Bad configuration, cannot start.")
             actorSystem.terminate()
             Await.result(actorSystem.whenTerminated, 30.seconds)
         }
+
+        if (!config.isValid) {
+            abort()
+            return
+        }
+
```
rabbah: you could rewrite the rest as 

```
ExecManifest.initialize(config) match {
            case Success(_) =>
                BasicHttpService.startService(actorSystem, "controller", "0.0.0.0", port, new ServiceBuilder(config, instance, logger))

            case Failure(t) =>
                logger.error(this, s"Invalid runtimes manifest: ${execManifest.failed.get}")
                abort()
        }
```

i can live with the `return`s since the scope is small and limited here.

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
ddragosd: I'm planning to deal with consul as part of #2254, not in this PR.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
ddragosd: 👍 I've updated the controller. Let me know if you want something similar for the invoker.

rabbah:
Ok I thought the test would fail since whisk config reads from consul. I'm surprised it passed. Do you know why?

I can make any small changes for merging. I think the main refactoring is already in place. Thanks for m/taking the suggestions.

ddragosd:
> Ok I thought the test would fail since whisk config reads from consul. I'm surprised it passed. Do you know why?

Yes, b/c the consul properties are not set when the test runs, `WhiskConfig` doesn't read them from consul ( https://github.com/apache/incubator-openwhisk/blob/master/common/scala/src/main/scala/whisk/core/WhiskConfig.scala#L167 )

rabbah:
Hmm that's only true if consul isn't actually up.

ddragosd:
> Hmm that's only true if consul isn't actually up.

Are you referring to the test added in this PR or in general ? B/c the test in this PR doesn't configure any consul properties.

rabbah:
Got it - because you're writing the property file directly you suppress the consul path. Thanks for clarifying.

ddragosd:
I've merged with the latest changes from master to include #2278 that fixes the build error caused by java install.

ddragosd:
good catch @rabbah. I'm making this change then I suppose we're good to merge here, having your 👍  ?

core/controller/src/main/scala/whisk/core/controller/Controller.scala
markusthoemmes: I dunno if I can live with the `return` though 🤔 . Should we just wrap the below inside the `if` for the sake of not using `return` in our codebase?

tests/src/test/scala/whisk/core/entity/test/ExecHelpers.scala
```diff
@@ -33,7 +33,7 @@ trait ExecHelpers
     self: Suite =>
 
     private val config = new WhiskConfig(ExecManifest.requiredProperties)
-    ExecManifest.initialize(config) shouldBe true
+    ExecManifest.initialize(config).isSuccess shouldBe true
```
markusthoemmes: Should rewrite to: `ExecManifest.initialize(config) should be a 'success` to print a non-conforming value on failure (and for fancyness 😁 )

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
```diff
@@ -170,4 +176,22 @@ class ExecManifestTests
...
     }
+
+    it should "throw an error when configured manifest is invalid" in {
+        val config_manifest = "{\"nodejs\":[{\"kind\":\"nodejs\",\"image\":{\"name\":\"nodejsaction\"},\"deprecated\":true},{\"kind\":\"nodejs:6\",\"default\":true,\"image\":{\"name\":\"nodejs6action\"}}]}"
+        val file = File.createTempFile("cxt", ".txt")
+        file.deleteOnExit()
+
+        val bw = new BufferedWriter(new FileWriter(file))
+        bw.write("runtimes.manifest=" + config_manifest + "\n")
+        bw.close()
```
markusthoemmes: Do we need to do I/O here? Couldn't we initialize `WhiskConfig` with the `String` immediatly like:

`new WhiskConfig(Map("runtimes.manifest" -> config_manifest), Set())`

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
```diff
@@ -170,4 +176,22 @@ class ExecManifestTests
...
+
+        val result = ExecManifest.initialize(new WhiskConfig(Map("runtimes.manifest" -> null), Set(), file), true)
+
+        result.isSuccess should be (false)
```
markusthoemmes: Same as above, should be: `result should be a 'failure`

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
rabbah: Dont think so - need to defeat looking for the prop file.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
rabbah: Exit works too. I accepted the return since this is "main" and in the invoker it is easier to read; and there is more than one "return" in that flow. Good suggestion with sys.exit as an alternative.

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
```diff
@@ -170,4 +176,22 @@ class ExecManifestTests
                     image.localImageName("r", "p", Some("tag")) shouldBe s"r/p/$name:tag"
             }
     }
+
+    it should "throw an error when configured manifest is invalid" in {
+        val config_manifest = "{\"nodejs\":[{\"kind\":\"nodejs\",\"image\":{\"name\":\"nodejsaction\"},\"deprecated\":true},{\"kind\":\"nodejs:6\",\"default\":true,\"image\":{\"name\":\"nodejs6action\"}}]}"
```
markusthoemmes: You can use triple-quotes to prevent escaping in the inner JSON. You could also explore constructing the JSON via scala objects to make this more readable:

```
JsObject(
    "nodejs" -> List(
        JsObject(
            "kind" -> "nodejs".toJson,
            "image" -> JsObject("name" -> "nodejsaction".toJson)
            "deprecated" -> true.toJson
        ),
        JsObject(
            "kind" -> "nodejs:6".toJson,
            "image" -> JsObject("name" -> "nodejs6action".toJson)
            "default" -> true.toJson
        )
    ).toJson
)
```

You can then generate a string out of that using `obj.compactPrint` and write that to the file.

On the same note, we could use a more minimalistic config-manifest as the error is not in the manifest itself, is it?

tests/src/test/scala/whisk/core/entity/test/ExecHelpers.scala
ddragosd: 👍

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
ddragosd: 👍

core/controller/src/main/scala/whisk/core/controller/Controller.scala
ddragosd: I can do `sys.exit(-1)` so that the program signals correctly an error.

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
ddragosd: Thanks for the suggestion. The reason I went with string initially is so that I can create either an invalid JSON, or an invalid manifest even if that JSON is valid, the goal being to simulate errors in config.  Indeed a `"""` would have been 👍

markusthoemmes: A comment, that this JSON is supposed to be valid but missing a key would help readability grasping what's going on in that case. I'm fine with a string.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
markusthoemmes: Why `-1`? Everything not `0` should be an error signal to the caller, no?

ddragosd: > Everything not 0 should be an error

That's right. IIRC the default status code is `0`; we have the option to be explicit, by providing anything else but `0` so that the java program doesn't give the impression it has terminated normally. I believe we could also return an error message so that we don't need to look inside the logs to see which property was not correct.

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
ddragosd: I've updated the test description

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
ddragosd: done

tests/src/test/scala/whisk/core/entity/test/ExecHelpers.scala
ddragosd: done

core/controller/src/main/scala/whisk/core/controller/Controller.scala
ddragosd: done. I went with `sys.exit(1)`

ddragosd:
@rabbah @markusthoemmes thanks for your reviews. I'm done with the changes. Provided the build succeeds, feel free to merge.

tests/src/test/scala/whisk/core/entity/test/ExecManifestTests.scala
```diff
@@ -16,21 +16,27 @@
 
 package whisk.core.entity.test
 
-import scala.util.Success
+import java.io.{BufferedWriter, File, FileWriter}
+import java.util.NoSuchElementException
 
+import scala.util.{Success}
```
rabbah: `{}` not needed here.

rabbah:
PG2/1652 🏃

rabbah:
PG 2/1666 finally 🚀

rabbah:
This is failing repeatedly in jenkins and I don't understand why yet. I can't see how it's related to these changes but I tried several and they all fail in the same way. I will spend some time to investigate. @csantanapr what's the status of an apache jenkins so that then it's easier to share logs or jenkins failures? It would help in this case for example.

rabbah:
Looking through the logs, I think this is just out of date. I'll run a jenkins build off a rebased version.

csantanapr:
@rabbah 
The ubuntu nodes in Apache Jenkins are shared between all projects and they don't provide root access and do not allow to change the default configuration. 
The default configuration is not configured for docker engine is not configure with http API enable on any port i.e 4243

We would need to change all the places we refer to the port to not use the http port and instead use the API over socket file device /vat/... the same API that the docker CLI uses

PR #2258, opened 2017-05-18
Limit overcommit of ContainerPool

markusthoemmes:
The ContainerPool currently can overcommit unboundedly, because removing a container and creating a new one are running in parallel. Removal of a container might not happen instantaniously as the container scheduled to delete might still be executing an action.

This bounds the number of containers which are subject to deletion to a constant similar to the slack of the old ContainerPool.

Needs to be verified and tested still.


markusthoemmes:
Won't implement.

PR #2259, opened 2017-05-18
Ensure that databases will be deleted again, even if the replicator t…

cbickel:
…est fails

This test was failing in the past, becasue the cleanup has not been executed if a test has failed. This PR fixes this issue.

@sven-lange-last
Could you also please have a look?

### DIFF
--- a/tests/src/test/scala/whisk/core/database/test/ReplicatorTests.scala
+++ b/tests/src/test/scala/whisk/core/database/test/ReplicatorTests.scala
@@ -52,7 +52,6 @@ class ReplicatorTests extends FlatSpec
     with StreamLogging {
 
     val config = new WhiskConfig(Map(
-        dbProvider -> null,
         dbProtocol -> null,
         dbUsername -> null,
         dbPassword -> null,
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/whisk/core/database/test/ReplicatorTests.scala
```diff
@@ -209,20 +208,23 @@ class ReplicatorTests extends FlatSpec
```
sven-lange-last: Is it possible that the `putDoc` fails with an exception - for example, due to timeout? If yes, the database won't be removed.

tests/src/test/scala/whisk/core/database/test/ReplicatorTests.scala
```diff
@@ -209,20 +208,23 @@ class ReplicatorTests extends FlatSpec
...
+            waitForReplication(backupDbName)
+
+            // Verify the replicated database is equal to the original database
+            compareDatabases(dbName, backupDbName, true)
+        } finally {
+            // Remove all created databases
+            createdBackupDbs.foreach(removeDatabase(_))
+            createdBackupDbs.foreach(removeReplicationDoc(_))
+            removeDatabase(dbName)
+        }
```
sven-lange-last: Are `removeDatabase()` and `removeReplicationDoc()` resilient enough to deal with error situations that could not happen with the previous code flow?

PR #2265, opened 2017-05-19
Rework config-dir variables

cbickel:
Rework dir-variables in ansible to avoid duplication in every environment.


cbickel:
PG3#629 is green.

cbickel:
PG4#306 is green.

PR #2272, opened 2017-05-21
Replace folder name openwhisk to incubator-openwhisk

thakkaryash94:


tools/ubuntu-setup/README.md
```diff
@@ -8,11 +8,11 @@ The commands below should be executed on the host machine for single VM/server d
   # Install git if it is not installed
   sudo apt-get install git -y
 
-  # Clone openwhisk
+  # Clone incubator-openwhisk
   git clone https://github.com/apache/incubator-openwhisk.git
```
csantanapr: I'm going over other repos updating this.
I preferred not to change the folder name and instead add the folder name to do the checkout.
This way when the `incubator` get's removed once we are out of Apache Incubator the default folder would be openwhisk.

Could you change that it adds the folder name to checkout?
```
git clone https://github.com/apache/incubator-openwhisk.git openwhisk
```
Then leave everything else the same.

thakkaryash94: Yeah, sure. I'll do it.

thakkaryash94: Can you check now?

PR #2275, opened 2017-05-22
remove kafka volume when clean up

xin-cai:
remove kafka volume when clean up

### DIFF
--- a/ansible/roles/kafka/tasks/clean.yml
+++ b/ansible/roles/kafka/tasks/clean.yml
@@ -5,6 +5,7 @@
   docker_container:
     name: kafka
     image: "{{ docker_registry }}{{ docker_image_prefix }}/kafka:{{ docker_image_tag }}"
+    keep_volumes: False
     state: absent
   ignore_errors: True
 
@@ -12,5 +13,6 @@
   docker_container:
     name: zookeeper
     image: "{{ docker_registry }}{{ docker_image_prefix }}/zookeeper:{{ docker_image_tag }}"
+    keep_volumes: False
     state: absent
   ignore_errors: True


xin-cai:
PG4/290 is running

xin-cai:
PG4/290 approved

sven-lange-last:
Observations regarding Docker volumes from Kafka container:

- The `Dockerfile` for Kafka creates two unnamed volumes: https://hub.docker.com/r/ches/kafka/~/dockerfile/
- `docker rm` won't remove volumes unless option `-v` is specified: https://docs.docker.com/v1.12/engine/reference/commandline/rm/
- The `clean`script of Ansible `kafka` role only asserts that the `kafka` Docker container is absent: https://github.com/apache/incubator-openwhisk/blob/master/ansible/roles/kafka/tasks/clean.yml
- The Ansible `docker_container` module support volume management via parameter `keep_volumes` which is set to `true` by default: https://docs.ansible.com/ansible/docker_container_module.html. 

In other words: at least for the `kafka` container, we don't clean up volumes when removing containers but retain them such that disk space fills up if volumes are not removed otherwise.

This PR fixes said problem. Looks good to me.

xin-cai:
The travis build ran exceed the maximum allowed time, unfortunately I don't have the right to restart the build... @jeremiaswerner could you help me to re-trigger the travis build? Thanks! 
https://travis-ci.org/apache/incubator-openwhisk/builds/235197294?utm_source=github_status&utm_medium=notification

xin-cai:
PG4/299 approved

xin-cai:
@jeremiaswerner  could you pls mergen? I don't have right to assign either reviewer or assignees ...

PR #2277, opened 2017-05-22
Disconnect WhiskConfig from Consul

ddragosd:
This is part 1 that relates to #2254. 

Given that removing consul from our deployment requires more code changes, I'm thinking to complete the task with smaller and incremental PRs. 

The scope of this PR is to remove the code from `WhiskConfig` that loads any configuration data from Consul.

### DIFF
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -36,6 +36,7 @@
       "SERVICE_CHECK_INTERVAL": "15s"
       "JAVA_OPTS": "-Xmx{{ controller.heap }}"
       "CONTROLLER_OPTS": "{{ controller.arguments }}"
+    env_file: "{{ openwhisk_home }}/whisk.env"
     volumes:
       - "{{ whisk_logs_dir }}/controller:/logs"
     ports:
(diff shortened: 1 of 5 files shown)


markusthoemmes:
Thanks for the contribution :+1:

How is the env-file approach viable for our goals here though? Ansible will not know if a config parameter specific for the controller (for example a limit change) needs to be applied to the deployment, does it?

ddragosd:
@markusthoemmes  good point .... do you have an idea how to approach this ? Should we change the file name with env vars to add something like a Git SHA ? Would ansible catch this ? ...

markusthoemmes:
Could we just pass all the required env-vars explicitly via ansible?

ddragosd:
> pass all the required env-vars explicitly

I'm ok with this option as well. It looks like we have about 25 of these extra properties to be added according to [docker-compose env vars](https://github.com/apache/incubator-openwhisk-devtools/blob/master/docker-compose/docker-whisk-controller.env)

ddragosd:
Another thought: IIRC the plan is to move to Typesafe Config, when I assume that we would create a generic JSON config file with default values in it, and only overwrite values with ENV VARS.
So the solution to be explicit with ENV VARS might be a good option for Typesafe config as well.

rabbah:
Shall I close in favor of https://github.com/apache/incubator-openwhisk/pull/2452?

PR #2281, opened 2017-05-24
add cron npm package

jasonpet:


markusthoemmes:
Why do we add this? A brief description in the PR would be helpful. You'll also need to update documentation.

csantanapr:
@markusthoemmes 

It's related to this https://github.com/apache/incubator-openwhisk-package-alarms/issues/63

i.e use the cron package to validate if the pattern is valid early in the feed action, this is also required for the new redundant work were action is going to insert new trigger into DB.

csantanapr:
@jasonpet update doc here: https://github.com/apache/incubator-openwhisk/blob/master/docs/reference.md#javascript-runtime-environments

rabbah:
Does it make sense to use a zip for the alarm action rather that put it in the runtime?

csantanapr:
yes it could be done with a zip action

jasonpet:
will do it as a zip

PR #2283, opened 2017-05-24
Add tool to delete logs from old activations

cbickel:


tools/db/deleteLogsFromActivations.py
```diff
@@ -0,0 +1,56 @@
...
+    endkey = long(time.time() * 1000) - args.days * DAY
+    while True:
+        activations = db.view("logCleanup/byDateWithLogs", limit=args.docsPerRequest, start_key=0, end_key=endkey, include_docs=True)
+        if activations:
+            def transform(viewResult):
+                doc = viewResult.doc
+                doc["logs"] = []
+                return doc
+
+            activationsWithoutLogs = map(lambda activation: transform(activation), activations)
```
jeremiaswerner: nice! 👍

cbickel:
PG1#1568 is green.

tests/src/test/scala/whisk/core/database/test/RemoveLogsTests.scala
```diff
@@ -0,0 +1,104 @@
...
+
+    /** Runs the clean up script to delete old activations */
+    def removeLogsTool(dbUrl: String, dbName: String, days: Int, docsPerRequest: Int = 20) = {
+        println(s"Running removeLogs tool: $dbUrl, $dbName, $days, $docsPerRequest")
+
+        val cmd = Seq(python, removeLogsToolPath, "--dbUrl", dbUrl, "--dbName", dbName, "--days", days.toString, "--docsPerRequest", docsPerRequest.toString)
+        val rr = TestUtils.runCmd(0, new File("."), cmd: _*)
+    }
+
+    behavior of "Activation Log Cleanup Skript"
```
rabbah: Skript -> Script.

tests/src/test/scala/whisk/core/database/test/RemoveLogsTests.scala
```diff
@@ -0,0 +1,104 @@
...
+                activationId = ActivationId.apply(),
+                start = Instant.now,
+                end = Instant.now,
+                logs = ActivationLogs(Vector("first line", "second line")))
+
+            db.putDoc(oldActivation.docid.asString, oldActivation.toJson).futureValue shouldBe 'right
+            db.putDoc(newActivation.docid.asString, newActivation.toJson).futureValue shouldBe 'right
+
+            // Run the tool to delete logs from activations that are older than 1 day
+            removeLogsTool(dbUrl, dbName, 1)
```
rabbah: do you expect records matching the date to be removed? if so, add a test for the corner case also. (less than, equal to, greater than).

tests/src/test/scala/whisk/core/database/test/RemoveLogsTests.scala
```diff
@@ -0,0 +1,104 @@
...
+            newActivationRequest shouldBe 'right
+            val newActivationFromDb = newActivationRequest.right.get.convertTo[WhiskActivation]
+            // Compare the Json of the Whiskactivations in case the test fails -> better log output
+            newActivationFromDb.toJson shouldBe newActivation.toJson
+
+            val oldActivationRequest = db.getDoc(oldActivation.docid.asString).futureValue
+            oldActivationRequest shouldBe 'right
+            val oldActivationFromDb = oldActivationRequest.right.get.convertTo[WhiskActivation]
+            // Compare the Json of the Whiskactivations in case the test fails -> better log output
+            oldActivationFromDb.toJson shouldBe oldActivation.copy(logs = ActivationLogs()).toJson
```
rabbah: you could use `oldActivation.withoutLogs` which is the same.

tests/src/test/scala/whisk/core/database/test/DatabaseScriptTestUtils.scala
```diff
@@ -0,0 +1,114 @@
...
+
+import org.scalatest.Matchers
+import org.scalatest.concurrent.ScalaFutures
+
+import akka.actor.ActorSystem
+import common.WaitFor
+import spray.json._
+import spray.json.DefaultJsonProtocol._
+import spray.json.JsObject
+import spray.json.pimpString
```
markusthoemmes: Unneeded imports.

tests/src/test/scala/whisk/core/database/test/DatabaseScriptTestUtils.scala
```diff
@@ -0,0 +1,114 @@
...
+import whisk.core.WhiskConfig
+import whisk.core.database.CouchDbRestClient
+import whisk.utils.retry
+import common.WhiskProperties
+
+trait DatabaseScriptTestUtils extends ScalaFutures
+    with Matchers
+    with WaitFor {
+
+    implicit val testConfig = PatienceConfig(10.seconds)
```
markusthoemmes: Is there a benefit over using `IntegrationPatience`?

tests/src/test/scala/whisk/core/database/test/DatabaseScriptTestUtils.scala
```diff
@@ -0,0 +1,114 @@
...
+        removeDatabase(name, true)
+
+        println(s"Creating database: $name")
+        val db = new ExtendedCouchDbRestClient(dbProtocol, dbHost, dbPort.toInt, dbUsername, dbPassword, name)
+        retry({ db.createDb().futureValue shouldBe 'right }, N = 10, waitBeforeRetry = Some(500.milliseconds))
+
+        retry({
+            val list = db.dbs().futureValue.right.get
+            list should contain(name)
+        }, N = 10, waitBeforeRetry = Some(500.milliseconds))
```
markusthoemmes: Would it be beneficial to create a local alias for retry, which is curried with the retry parameters? Like

```
def retry(task: => Any) = whisk.utils.retry(task, N = 10, waitBeforeRetry = Some(500.milliseconds))
```

(Don't know the type of `task` but you get the idea)

tools/db/deleteLogsFromActivations.py
```diff
@@ -0,0 +1,56 @@
...
+#
+# Delete activations
+#
+def deleteLogsFromOldActivations(args):
+    db = couchdb.client.Server(args.dbUrl)[args.dbName]
+    endkey = long(time.time() * 1000) - args.days * DAY
+    while True:
+        activations = db.view("logCleanup/byDateWithLogs", limit=args.docsPerRequest, start_key=0, end_key=endkey, include_docs=True)
+        if activations:
+            def transform(viewResult):
```
markusthoemmes: I'd either inline this as a lambda function or move it outside of the outer `def` and give it a more speaking name. This function will be created on each loop. (I know that performance is not critical here, but as a general rule of thumb: Don't create functions in a loop).

tools/db/deleteLogsFromActivations.py
markusthoemmes: The intermediate lambda is not necessary in the approach you take: `map(transform, activations)` should work here.

tests/src/test/scala/whisk/core/database/test/RemoveLogsTests.scala
```diff
@@ -0,0 +1,104 @@
...
+        val dbName = dbPrefix + "test_log_cleanup_1"
+        val db = createDatabase(dbName, Some(designDocPath))
+
+        try {
+            // Create an old and a new activation
+            val oldActivation = WhiskActivation(
+                namespace = EntityPath("testns1"),
+                name = EntityName("testname1"),
+                subject = Subject("test-sub"),
+                activationId = ActivationId.apply(),
```
markusthoemmes: Why not `ActivationId()`? Same below.

tests/src/test/scala/whisk/core/database/test/RemoveLogsTests.scala
```diff
@@ -0,0 +1,104 @@
...
+                logs = ActivationLogs(Vector("first line", "second line")))
+
+            val newActivation = WhiskActivation(
+                namespace = EntityPath("testns1"),
+                name = EntityName("testname1"),
+                subject = Subject("test-sub"),
+                activationId = ActivationId.apply(),
+                start = Instant.now,
+                end = Instant.now,
+                logs = ActivationLogs(Vector("first line", "second line")))
```
markusthoemmes: Both activations only seem to differ in their start date, right? You can use `oldActivation.copy(start = Instant.now)` to make this obvious in the code. (Note that you use `Instant.now repeatedly, you might want to cache just one version of that value to prevent ambiguity)

tests/src/test/scala/whisk/core/database/test/RemoveLogsTests.scala
```diff
@@ -0,0 +1,104 @@
...
+
+    val designDocPath = WhiskProperties.getFileRelativeToWhiskHome("ansible/files/logCleanup_design_document_for_activations_db.json").getAbsolutePath
+    val removeLogsToolPath = WhiskProperties.getFileRelativeToWhiskHome("tools/db/deleteLogsFromActivations.py").getAbsolutePath
+
+    /** Runs the clean up script to delete old activations */
+    def removeLogsTool(dbUrl: String, dbName: String, days: Int, docsPerRequest: Int = 20) = {
+        println(s"Running removeLogs tool: $dbUrl, $dbName, $days, $docsPerRequest")
+
+        val cmd = Seq(python, removeLogsToolPath, "--dbUrl", dbUrl, "--dbName", dbName, "--days", days.toString, "--docsPerRequest", docsPerRequest.toString)
+        val rr = TestUtils.runCmd(0, new File("."), cmd: _*)
```
markusthoemmes: Wondering if we want to replace `TestUtils` in new tests, it's the old Java stuff I think?

`cmd ! shouldBe 0` should be the same, using the correct `scala.Process` related imports. Won't sit on that change though, but might be worth a try 😁

tests/src/test/scala/whisk/core/database/test/DatabaseScriptTestUtils.scala
```diff
@@ -0,0 +1,114 @@
...
+        documents.right.get
+    }
+
+    /** wait until all documents are processed by the view */
+    def waitForView(db: CouchDbRestClient, designDoc: String, viewName: String, numDocuments: Int) = {
+        waitfor(() => {
+            val view = db.executeView(designDoc, viewName)().futureValue
+            view shouldBe 'right
+            view.right.get.fields("rows").convertTo[List[JsObject]].length == numDocuments
+        }, totalWait = 2.minutes)
```
markusthoemmes: We sometimes use `waitfor` and sometimes `retry`? Any particular reason for that? Could be easier to read if we normalize to `retry` and write a `should` test for each condition to throw the exception.

tests/src/test/scala/whisk/core/database/test/RemoveLogsTests.scala
cbickel: The script does a check, if the activation is older than the given days (in milliseconds).
I'd like to avoid the tests of the edge cases. The reason is, that I think, that they will fail in the pipeline, if the test is a bit slower than expected.

tests/src/test/scala/whisk/core/database/test/DatabaseScriptTestUtils.scala
cbickel: The difference between the use of `waitfor` and `retry` here is, that `retry` is used, when operations should be there immediately. And `waitfor` is used, when we know, that something takes several seconds.

The retry waits in total only 5 seconds and the `waitfor` waits in total 30 seconds.
We could use a `retry` for the `waitfor`s as well, but then it makes no sence to have the local alias for retry.

rabbah: doesn't retry have a `waitBeforeRetry` which you can use to "wait" before retrying?
you can use that with a retry of `1`, iirc.

rabbah: in any case `waitfor` is pre-existing code, not new right?

cbickel: Markus suggested above to create one local alias for retry, to pass only the task parameter. As result all `waitBeforeRetry` parameters are the same in this test.
If we use `retry` for the "longrunning" cloudant tasks as well, we have to increase this parameter for all retries or use the original one.
I decided to keep the `waitfor` for longrunning tasks and `retry` for the requests that should be there immediately.

cbickel:
PG1#1590 is green.

PR #2284, opened 2017-05-24
configure kafka log retension per topic

xin-cai:
configure kafka log retension per topic:

active-ack topic: 1MB
health topic: 1 MB
per invoker topic: 100 MB

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -113,6 +113,11 @@ kafka:
   port: 9092
   ras:
     port: 9093
+  topic:
+    completed_logByteSize: 1000000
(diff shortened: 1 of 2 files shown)


xin-cai:
PG4/307 is running

markusthoemmes:
What is the effect of this? Will invoker topics just drop messages after 100MB?

Same for the active-ack topic. A single message cam already be 1MB in size.

rabbah:
Moreover these will need to be parametrized per deployment so that they are proportional to the invoker/controller ratio and the operational requirements of the platform (sustained throughout, recovery time, disk availability, etc.)

PR #2298, opened 2017-05-25
temporary pinned apigateway to 0.7.0

csantanapr:


csantanapr:
cc @mdeuser @rabbah

ansible/roles/apigateway/tasks/deploy.yml
```diff
@@ -5,10 +5,11 @@
   shell: "docker pull openwhisk/apigateway"
   when: apigateway_local_build is undefined
 
+#temporary pinned to image 0.7.0 until apigateway fixes regressions
 - name: (re)start apigateway
   docker_container:
     name: apigateway
-    image: openwhisk/apigateway
+    image: openwhisk/apigateway:0.7.0
```
mdeuser: Temp fix looks good to me

rabbah:
Perhaps this should stay pinned. We can't keep getting bit by this.

mdeuser:
Perhaps.... unfortunately, pinning in our build does not guarantee the version of the deployed api gw in ys1 or yp.

rabbah:
Sigh. Good point. For the open source communit at least that not a concern and we should not break master and prs. For the deployed hosted version I don't see a compromise: we need to get the full test suite integrated, or decouple things further so that the impact is limited.

PR #2299, opened 2017-05-26
Normalize rate checks to the identity namespace (uuid).

### LINKED ISSUE #2266: limits discrepancy
the docs state the following system [limits](https://github.com/apache/incubator-openwhisk/blob/master/docs/reference.md#system-limits)

concurrent actions: N per namespace
minuteRate: M per user (aka subject)

however what's implemented is a limit check per subject for both: https://github.com/apache/incubator-openwhisk/blob/master/core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala#L129

rabbah:
Fixes #2266.


rabbah:
Will deliver this in a separate PR that fixes other accounting issues with active acks.

PR #2310, opened 2017-05-29
add out-of-the-box support for db2 database client

jeremiaswerner:
Users are asking more and more for support of db2 client. The PR proposes to include the library in the standard runtime next to the other database drivers like nano, cloudant, mongodb.

### DIFF
--- a/core/nodejs6Action/Dockerfile
+++ b/core/nodejs6Action/Dockerfile
@@ -36,6 +36,7 @@ glob@7.1.1 \
 gm@1.23.0 \
 lodash@4.17.2 \
 log4js@0.6.38 \
+ibm_db@2.0.0 \
 iconv-lite@0.4.15 \
 marked@0.3.6 \
 merge@1.2.0 \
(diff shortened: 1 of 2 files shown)


markusthoemmes:
Please update the docs accordingly.

csantanapr:
@jeremiaswerner This will not work take a look #2311

jeremiaswerner:
@csantanapr thanks for pointing me to the issue. Travis was failing due to that reason.

core/nodejs6Action/Dockerfile
```diff
@@ -36,6 +36,7 @@ glob@7.1.1 \
 gm@1.23.0 \
 lodash@4.17.2 \
 log4js@0.6.38 \
+ibm_db@2.0.0 \
```
csantanapr: This will not installed.
In addition, it adds large amount of files that are not need it see #2311

jeremiaswerner:
@csantanapr thanks. I agree. let's focus on fixing #2311 which would resolve this issue. Users could then leverage the .zip file support which would result in much smaller files. closing this PR.

rabbah:
I'd also consider an approach like the one intended for https://github.com/apache/incubator-openwhisk/issues/2300

mrutkows:
@csantanapr agree with adding ibm specific library/service support if it is not being driven by the Apache community alongside other similar efforts.

PR #2315, opened 2017-05-30
Support interactive cli testing via input file

mdeuser:
Update test framework to allow stdin input from a file

### DIFF
--- a/tests/src/test/scala/common/Wsk.scala
+++ b/tests/src/test/scala/common/Wsk.scala
@@ -1003,11 +1003,12 @@ trait RunWskCmd extends Matchers {
             verbose: Boolean = false,
             env: Map[String, String] = Map("WSK_CONFIG_FILE" -> ""),
             workingDir: File = new File("."),
+            stdinFile: Option[File] = None,
             showCmd: Boolean = false): RunResult = {
         val args = baseCommand
         if (verbose) args += "--verbose"
         if (showCmd) println(args.mkString(" ") + " " + params.mkString(" "))
-        val rr = TestUtils.runCmd(DONTCARE_EXIT, workingDir, TestUtils.logger, sys.env ++ env, args ++ params: _*)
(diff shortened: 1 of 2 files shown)


mdeuser:
PG3/672

mdeuser:
@csantanapr - ready for review

/cc @dubeejw @rabbah

tests/src/test/scala/common/Wsk.scala
```diff
@@ -1003,11 +1003,12 @@ trait RunWskCmd extends Matchers {
             verbose: Boolean = false,
             env: Map[String, String] = Map("WSK_CONFIG_FILE" -> ""),
             workingDir: File = new File("."),
+            stdinFile: File = null,
```
markusthoemmes: That should be an `Option[File]` with a default value of `None`.

Then to use it against the Java API you can use `stdinFile.getOrElse(null)`. At least keeps the `null` out of the Scala API.

mdeuser: With all other parameters are not `Option` parameters, would introducing an `Option` parameter be confusing?

mdeuser: perhaps it's just me, but method signatures with a mix of `Option` parameters and named default parameters is a bit confusing.
https://stackoverflow.com/questions/4199393/are-options-and-named-default-arguments-like-oil-and-water-in-a-scala-api

rabbah: It avoids introducing nulls and hence NPEs. Here's it's fairly localized and passed on to the Java client, so it's not a big deal. FWIW we have explored introducing a scala linter which will flag uses of null. So personally I'd change it to an Option type also.

mdeuser: another scala best practice recommendation to remove `null` from all scala code... removing `null` per @markusthoemmes suggestion :+1: 
http://alvinalexander.com/scala/scala-null-values-option-uninitialized-variables

mdeuser:
rebased and squashed.  PG3/674

Issue #2285, opened 2017-05-24
Invoker with reactive pool does no more process activations if pause grace expires too often


sven-lange-last:
In tests on virtual systems with one CPU (=at most 4 containers on the invoker) with light workload (=more than 40 activations per invoker with >> 1 sec pause between activations), activation feed book-keeping breaks and the invoker no more pulls activation requests from Kafka. For this reason, the invoker stops processing activations but still sends health pings to the controller so that the problem is not detected.

Invoker log lines like the following indicate that the problem occurred:
```
[ActivationFeed] waiting for activation pipeline to drain: 5 > 4
```
If the problem occurs, the invoker is still running but no more log entries are appended.

Found that the `ContainerProxy` sends `NeedWork(WarmedData())` twice to the `ContainerPool`:
- when entering state `Running` with `WarmedData`, i.e. after having processed an activation successfully
- when being in state `Ready` and the `pauseGrace` timeout expires, we transition to `Pausing` sending `NeedWork`.
The `ContainerPool` sends `ContainerReleased` to the `ActivationFeed` when receiving `NeedWork(WarmedData())`.

In addition, the new container pool does NOT send `FailedActivation()` to the `ActivationFeed`.

Need to understand the book-keeping and fix it.

markusthoemmes:
Good find! Just a quick note: Failed activation being sent is irrelevant here.

sven-lange-last:
I agree - in the scenario that raised this defect, we had no failed activations. I just wanted to mention that we need a complete picture with regard to `ActivationFeed` book-keeping.

PR #2319, opened 2017-05-31
Fix activation feed book-keeping in reactive pool (#2285)

sven-lange-last:
- The current implementation sends a `ContainerReleased` message to the `ActivationFeed` whenever the `ContainerPool` receives a `NeedWork` message from a warm `ContainerProxy`.
- The current implementation sends `NeedWork` after an activation was completed successfully (correct) but also after the pause grace timeout expired (incorrect). This double message breaks `ActivationFeed` book-keeping.
- The current implementation does not send any `ContainerReleased` or `FailedActivation` message to the `ActivationFeed` if an activation cannot be completed successfully. So even application errors break `ActivationFeed` book-keeping.

Changes:

- This PR decouples `ActivationFeed` notification from the `NeedWork` message sent by the `ContainerProxy` to `ContainerPool`.
- Whenever the `ContainerProxy` sends an active ack and stores the activation, it now sends an `ActivationCompleted` message to the `ContainerPool`.
- Whenever the `ContainerPool` receives an `ActivationCompleted` message, it relays a `ContainerReleased` or `FailedActivation` message to the `ActivationFeed`.
- Added more log details / new log messages to the `ActivationFeed` to make debugging of book-keeping easier in future.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
@@ -66,6 +66,11 @@ case object Remove
 case class NeedWork(data: ContainerData)
 case object ContainerPaused
 case object ContainerRemoved
+/**
+ * Indicates the container resource is now free to receive a new request.
+ * This message is sent to the parent which in turn notifies the feed that a
+ * resource slot is available.
+ */
 case object ActivationCompleted
 
 /**
@@ -103,7 +108,7 @@ class ContainerProxy(
     startWith(Uninitialized, NoData())
 
     when(Uninitialized) {
-        // pre warm a container
+        // pre warm a container (creates a stem cell container)
         case Event(job: Start, _) =>
             factory(
                 TransactionId.invokerWarmup,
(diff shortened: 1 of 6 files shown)


rabbah:
From the description it reads as though this is not addressed in this PR: "So even application errors break ActivationFeed book-keeping." Is that accurate? (I have not yet reviewed the proposed patch. Just wanted to level set expectation.)

markusthoemmes:
@rabbah, that seems to be a miscommunication. It is fixed.

sven-lange-last:
@markusthoemmes thanks for making sure messages are now sent in a fixed sequence such that corresponding unit tests pass with any thread execution context and any environment.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -336,7 +362,8 @@ class ContainerProxy(
         // and do not block further execution of the future. They are completely
         // asynchronous.
         activation.andThen {
-            case Success(activation) => sendActiveAck(tid, activation)
+            // the activation future will always complete with Success
+            case Success(ack) => sendActiveAck(tid, ack)
```
rabbah: I was tempted here to assert the invariant as:

`case result => val Success(ack)` forcing an error should the contract be vioated

PR #2321, opened 2017-06-01
clean up formatting docs for actions

csantanapr:


docs/actions.md
```diff
@@ -308,13 +310,12 @@ JavaScript functions that run asynchronously may need to return the activation r
...
   ```
-  $ wsk activation get b066ca51e68c4d3382df2d8033265db0
+  wsk activation get b066ca51e68c4d3382df2d8033265db0
   ```
- ```
+ ```json
   {
       "start": 1455881628103,
-      "end":   1455881648126,
-      ...
```
bjustin-ibm: I think these dots should stay. Without them, it appears as if the activation record only contains the fields `start` and `end`.

docs/actions.md
```diff
@@ -632,48 +634,61 @@ When you create an OpenWhisk Swift action with a Swift source file, it has to be
...
+  ```
+  cat /swift3Action/epilogue.swift >> /swift3Action/spm-build/main.swift
+  ```
 
-            $ exit
+  ```
+  echo '_run_main(mainFunction:main)' >> /swift3Action/spm-build/main.swift
+  ```
 
+- zBuild and link
```
bjustin-ibm: Is this leading "z" a typo?

docs/actions.md
```diff
@@ -44,7 +44,7 @@ Review the following steps and examples to create your first JavaScript action.
 2. Create an action from the following JavaScript function. For this example, the action is called 'hello'.
 
   ```
```
bjustin-ibm: There is also a "shell" syntax highlighting available. I wonder if you should go through and mark all the shell command sections with it?

csantanapr: I already played adding this "shell" and also "bash" it didn't do anything to our syntax, and also cause some problems with highlighting that I can't remember now.

docs/actions.md
csantanapr: Will double check if the syntax highlighting still looks ok because the ... make it a none JSON string.

docs/actions.md
csantanapr: yeah it looks like a typo, don't know how it got there 😮

PR #2327, opened 2017-06-02
Update OPTIONS Respones for Web Actions (Review)

dubeejw:
Closes https://github.com/apache/incubator-openwhisk/issues/2234#issuecomment-302120162


core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -406,6 +407,15 @@ trait WhiskWebActionsApi
                         requestMethodParamsAndPath { context =>
                             provide(fullyQualifiedActionName(actionName)) { fullActionName =>
                                 onComplete(verifyWebAction(fullActionName, onBehalfOf.isDefined)) {
+                                    case Success((actionOwnerIdentity, action)) if !action.annotations.asBool("web-custom-options").exists(identity) && context.method.value.toLowerCase == "options" =>
+                                        val allowOrigin = `Access-Control-Allow-Origin`(AllOrigins)
```
dubeejw: CORS values need to be put in general util file.

dubeejw:
Needs tests.

dubeejw:
PG1 1598

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -173,8 +173,6 @@ protected[controller] class RestAPIVersion(apipath: String, apiversion: String)(
                     webexp.routes()
                 } ~ {
                     swaggerRoutes
-                } ~ options {
```
rabbah: if you're not restoring this somewhere you are breaking "experimental" - plus options on the main routes (which we agreed is a separate issue to address/fix), no?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -291,10 +292,16 @@ protected[core] object WhiskWebActionsApi extends Directives {
     }
 }
 
+protected[controller] trait RespondWithMethods extends RespondWithHeaders {
+    val allowMethods = `Access-Control-Allow-Methods`(OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH)
+    override val sendCorsHeaders = respondWithHeaders(allowOrigin, allowHeaders, allowMethods)
```
rabbah: Aren't we dropping `allowHeaders` for Web actions?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -406,6 +413,10 @@ trait WhiskWebActionsApi
                         requestMethodParamsAndPath { context =>
                             provide(fullyQualifiedActionName(actionName)) { fullActionName =>
                                 onComplete(verifyWebAction(fullActionName, onBehalfOf.isDefined)) {
+                                    case Success((actionOwnerIdentity, action)) if context.method.value.toLowerCase == "options" && !action.annotations.asBool("web-custom-options").exists(identity) =>
```
rabbah: rather than string compare, I think you can compare to the `OPTIONS` verb.

rabbah: i think the `if` predicate here is too long and less readable - why not nest it below? (if not options ... else ...)

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -328,6 +338,7 @@ trait WebActionsApiTests extends ControllerTestCommon with BeforeAndAfterEach wi
 
         it should s"not match invalid routes (auth? ${creds.isDefined})" in {
             implicit val tid = transid()
+            customOptions = true
```
rabbah: why are there so many tests that set this flag true?

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -1093,19 +1130,35 @@ trait WebActionsApiTests extends ControllerTestCommon with BeforeAndAfterEach wi
                 }
         }
 
-        it should s"invoke action with options verb (auth? ${creds.isDefined})" in {
+        it should s"invoke action with options verb with custom options (auth? ${creds.isDefined})" in {
             implicit val tid = transid()
+            customOptions = true
```
rabbah: isn't this one enough...

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -1093,19 +1130,35 @@ trait WebActionsApiTests extends ControllerTestCommon with BeforeAndAfterEach wi
...
-                        header("allow").get.toString shouldBe "allow: options, head, get, post, put"
+                        header("Access-Control-Allow-Origin") shouldBe None
+                        header("Access-Control-Allow-Headers") shouldBe None
+                        header("Access-Control-Allow-Methods").get.toString shouldBe "Access-Control-Allow-Methods: OPTIONS, GET, PATCH"
+                    }
+                }
+        }
+
+        it should s"invoke action with options verb without custom options (auth? ${creds.isDefined})" in {
+            implicit val tid = transid()
```
rabbah: ...and this one?

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
dubeejw: The majority of the tests invoke a web action with the OPTIONS verb. In order to maintain the same test behavior I am adding the custom options annotation to those tests. We could enable the annotation by default in the tests when declaring the custonOptions value.

rabbah: should options be pulled out and run in its own tests (do we need more than two?)

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
dubeejw: Could add it back, but experimental features don't guaranteed that API breaking changes won't happen. @csantanapr, what do you think about removing the OPTIONS verb from experimental?

rabbah: this also removes it from the main like API (which I agree is silly to preflight since it's authenticated) but that's a breaking change (vs experimental).

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
dubeejw: Only need one if I declare the variable like so: `var customOptions = true`.

rabbah: i dont follow - to clarify "do we need more than two" i was thinking that we only need two tests for options: one without the default annotation and one with.

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
dubeejw: See if you like the tests now.

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
dubeejw: Okay, how do we get the OPTIONS route to match in two different locations?

dubeejw: If we leave it were it is currently, it will never match the OPTIONS route for web actions.

rabbah: right - you can invert the order of route matching so web comes first (it's a stricter match).
alternatively make the option matching in the top route more strict (not `/web` prefix).

dubeejw: Nevermind, I figured it out.

tests/src/test/scala/whisk/core/cli/test/WskWebActionsTests.scala
```diff
@@ -210,27 +210,6 @@ trait WskWebActionsTests
             authorizedResponse.body.asString shouldBe namespace
     }
 
-    if (testRoutePath == "/api/v1/web") {
```
dubeejw: Replaced this integration test with a unit test.

dubeejw:
PG1 1603.

tests/src/test/scala/whisk/core/cli/test/WskWebActionsTests.scala
rabbah: Hmm. This integration test was the test we needed in the first place (and we had a unit test). I don't think we can replace it since the unit tests don't excercise the full route matching.

rabbah: If you want to replace it with a unit test then you'll need to instatiate the top level route in the controller.

dubeejw:
PG4 332

tests/src/test/scala/whisk/core/cli/test/WskWebActionsTests.scala
dubeejw: Added the unit test back without the conditional statement that prevented it from running.

dubeejw:
PG looks good for this one.

dubeejw:
@rabbah, anymore comments on this one, or should we merge it?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -291,10 +292,16 @@ protected[core] object WhiskWebActionsApi extends Directives {
     }
 }
 
+protected[controller] trait RespondWithMethods extends RespondWithHeaders {
```
rabbah: why does this need to be a trait or top level?

rabbah: is there a documentation change to reflect the new behavior?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -406,6 +413,11 @@ trait WhiskWebActionsApi
                         requestMethodParamsAndPath { context =>
                             provide(fullyQualifiedActionName(actionName)) { fullActionName =>
                                 onComplete(verifyWebAction(fullActionName, onBehalfOf.isDefined)) {
+                                    case Success((actionOwnerIdentity, action)) if context.method == OPTIONS &&
+                                            !action.annotations.asBool("web-custom-options").exists(identity) =>
+                                        sendCorsHeaders {
+                                            complete(OK, "")
```
rabbah: instead of `""` use `HttpEntity.Empty`.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -406,6 +413,11 @@ trait WhiskWebActionsApi
                         requestMethodParamsAndPath { context =>
                             provide(fullyQualifiedActionName(actionName)) { fullActionName =>
                                 onComplete(verifyWebAction(fullActionName, onBehalfOf.isDefined)) {
+                                    case Success((actionOwnerIdentity, action)) if context.method == OPTIONS &&
```
rabbah: i think i prefer to move the `if` predicate inside a single case statement for the `Success((actionOwnerIdentity, action))`; the predicate is sufficiently long I don't think we gain any readability or benefit from this.

dubeejw:
@csantanapr, can you review the doc here?

docs/webactions.md
```diff
@@ -360,6 +360,47 @@ $ curl -k -H "content-type: application" -X POST -d "Decoded body" https://${API
...
+`web-custom-options` annotation with a value of `true` to a web action. When this feature is enabled, CORS headers will
+not automatically be added to the request response. Instead, it is the developer's responsibility to append his or her
+desired headers programmatically. Below is an example of creating custom resposnes to OPTIONS requests.
+
+```
+function main(params) {
+  if (params.__ow_method == "options") {
+    return {
+      headers: {
+        'Access-Control-Allow-Methods': 'OPTIONS, GET'
```
csantanapr: Add a second header to the example, maybe overwriting the origin,
Then there is no assumption that origin header will be return with '*" and the rest with custom.
```
Access-Control-Allow-Origin: 'example.com'
```

docs/webactions.md
```diff
@@ -360,6 +360,47 @@ $ curl -k -H "content-type: application" -X POST -d "Decoded body" https://${API
...
+Save the above function to `custom-options.js` and execute the following commands:
+
+```
+$ wsk action create custom-option custom-options.js --web true -a web-custom-options true
+$ curl https://${APIHOST}/api/v1/web/guest/default/custom-options.http -kvX OPTIONS
+< HTTP/1.1 200 OK
+< Server: nginx/1.11.13
+< Content-Length: 0
+< Connection: keep-alive
+< Access-Control-Allow-Methods: OPTIONS, GET
```
csantanapr: same here add a 2nd header of origin
```
Access-Control-Allow-Origin: example.com
```

dubeejw:
@csantanapr, updated the PR with your suggestions.

dubeejw:
PG1 1644, approved.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -406,6 +413,11 @@ trait WhiskWebActionsApi
                         requestMethodParamsAndPath { context =>
                             provide(fullyQualifiedActionName(actionName)) { fullActionName =>
                                 onComplete(verifyWebAction(fullActionName, onBehalfOf.isDefined)) {
+                                    case Success((actionOwnerIdentity, action))
```
rabbah: I meant:
```
case Success((actionOwnerIdentity, action)) =>
   if (context.method == OPTIONS &&
       !action.annotations.asBool("web-custom-options").exists(identity)) {
                                         sendCorsHeaders {
                                           complete(OK, HttpEntity.Empty)
                                        }
     } else {
        ...
     }
```

or
```
case Success((actionOwnerIdentity, action)) =>
  context.method match {
      case OPTIONS if !... => sendCoresHeaders ...
      case _ => extract ...
  }
```

docs/annotations.md
```diff
@@ -47,4 +47,4 @@ We recently extended the core API with new features. To enable packages and acti
 2. `web-export`: Applies only to an action. If present, it makes its corresponding action accessible to REST calls _without_ authentication. We call these [_web actions_](webactions.md) because they allow one to use OpenWhisk actions from a browser for example. It is important to note that the _owner_ of the web action incurs the cost of running them in the system (i.e., the _owner_ of the action also owns the activations record).
 3. `require-whisk-auth`: Applies onto to an action. If an action carries the `web-export` annotation, and this annotation is also `true`, the route is only accessible to an authenticated subject. It is important to note that the _owner_ of the web action incurs the cost of running them in the system (i.e., the _owner_ of the action also owns the activations record).
 4. `raw-http`: Applies only to an action in the presence of a `web-export` annotation. If present, the HTTP request query and body parameters are passed to the action as reserved properties.
-
+5. `web-custom-options`: Applies ony to an action. When used with a value of `true`, this annotation enables the ability for a web action to respond to OPTIONS requests with customized headers.
```
rabbah: can you link here to the options section in webactions.md to explain the defaults when the value is false?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
rabbah: I find this trait confusing - you can add the parent trait to the web actions api and override there.
I don't think we're gaining anything from this.

rabbah: I wouldn't even extend RespondWithHeaders.

docs/webactions.md
```diff
@@ -360,6 +360,48 @@ $ curl -k -H "content-type: application" -X POST -d "Decoded body" https://${API
...
+The headers are shown below:
+
+```
+Access-Control-Allow-Origin: *
+Access-Control-Allow-Methods: OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH
+```
+
+Alternatively, OPTIONS requests can be handled manually by a web action. To enable this option add a
+`web-custom-options` annotation with a value of `true` to a web action. When this feature is enabled, CORS headers will
+not automatically be added to the request response. Instead, it is the developer's responsibility to append his or her
```
rabbah: developer's responsibility to append his or her -> eveloper's responsibility to append their

docs/webactions.md
```diff
@@ -360,6 +360,48 @@ $ curl -k -H "content-type: application" -X POST -d "Decoded body" https://${API
...
+desired headers programmatically. Below is an example of creating custom resposnes to OPTIONS requests.
```
rabbah: typo

docs/annotations.md
rabbah: ony typo

rabbah:
@dubeejw @csantanapr I edited the docs please review before I merge.

dubeejw:
@rabbah, @csantanapr, LGTM.

PR #2329, opened 2017-06-03
Use unique test API Names for unique basepaths

mdeuser:
- New API GW restriction


mdeuser:
@csantanapr - for your review.  API GW V2 now enforces unique swagger titles.

PG3/685

PR #2332, opened 2017-06-06
Fix a small error in /tools/vagrant/README.md

daisy-ycguo:
Fix a small error in vagrant document.


rabbah:
Only way to assure any of the desired invariants is to add a test. Can we get some Apache infra to test the vagrant machine now?

PR #2333, opened 2017-06-06
General ContainerPool refactorings.

markusthoemmes:
A couple of helpful refactorings for the new ContainerPool:

1. Differentiate between maximum active containers and maximum containers in pool. This can come in handy if you want to increase the poolsize to cache more containers than you run. Currently disabled but contains helpful refactorings so I kept it.
2. Retype scheduling logic: Removes a dead-code situation.
3. Adjust pause grace: Adjusts the pause grace to a more aggressive value, which will still catch bursts bu doesn't gift so much time to a potentially malicious user.


markusthoemmes:
PG3 754 🔵

Issue #821, opened 2016-07-01
Add oauth2 client for trigger authentication with event sources


mbehrendt:
today we require username/pw to be passed on to a trigger. we should offer the ability to avoid that by allowing authentication with oauth2.

rabbah:
@starpit FYI.

starpit:
the UI supports this for github, and soon for slack (pending https://github.com/openwhisk/openwhisk/pull/1097).

i'm not sure if this is quite possible from the CLI.  oauth token negotiation requires that the user be directed to the provider's oauth initiation site, so that they may enter their credentials on a trusted site.

nastacio:
Is there any update on this request?

It seems OpenWhisk should prescribe the OAuth integration or even implement some of it under the covers for actions.

An action is likely to be part of an OAuth machine-to-machine interaction pattern (grant_type=client_authorization) . It may simply receive an access token that it has to pass along to the next action in a sequence, or simple accept and validate the token from a caller. It seems there is a need for a build-in mechanism to receive such token and also for retrieving the introspection URL of a provider to validate a token.

rabbah:
@starpit want to take this? :)

starpit:
Hi @nastacio,

We should be able to augment the backend and CLI to support oauth logins. Initially, I am imagining that this won't support any kind of grouping; rather, we could initially support authenticating individual users against known oauth identity providers, such as github and google.

One of the tricky points here, especially viz. an open source project such as OpenWhisk, would be in supporting the secrets:

1. For each identity provider, the maintainer of an OpenWhisk installation would need to allocate a (id,secret) pair. This we can address via documentation.

2. The more complex issue is how we support continuous integration.  In particular, how can we support Travis tests against PRs? Do we allocate a set of test (id,secret) pairs, and store them in the Travis key store? Does this violate the open source nature of OpenWhisk, to have secrets such as this? And what about localhost testing?

Or, do we make a set of (id,secret) pairs public, stating forcefully that one must not depend on these for production installations?

Finally, to test the full login path, from wsk client through the server-side backend (and back), we would *also* need to have a set of user credentials, so that the test client can log on to the identity providers. This raises mostly the same issues as with the (id,secret) pair, perhaps with some additional concerns; e.g. do identity providers offer REST APIs for user account creation?

starpit:
the implementation will have a degree of overlap with #1247  -- both need backend support for negotiating oauth secrets

nastacio:
[...]

PR #445, opened 2016-05-17
Add markers to instrument Database-library

ChrBi:
Signed-off-by: Markus Thoemmes <markus.thoemmes@de.ibm.com>

This PR includes the instrumentation of the Database-calls.

PR #2334, opened 2017-06-06
(Review) Added --last Activation Flag 

bpoole16:
Added flag  "--last" or "-l" in the CLI that gets the most recent activation id for activation's get, logs and result. Closes https://github.com/apache/incubator-openwhisk/issues/1498

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -921,7 +921,9 @@ You can use the OpenWhisk CLI to watch the output of actions as they are invoked
   ```
   wsk activation poll
   ```
-  
+
+This command starts a polling loop that continuously checks for logs from activations.
+
 2. Switch to another window and invoke an action:
 
   ```
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -921,13 +921,7 @@ You can use the OpenWhisk CLI to watch the output of actions as they are invoked
   ```
   wsk activation poll
   ```
-
-  This command starts a polling loop that continuously checks for logs from activations. Alternatively, a poll can be done for a specific activation id once it has been invoked by adding the activation id.
-
-  ```
-  wsk activation poll [ACTIVATION_ID]
-  ```
-
+  
 2. Switch to another window and invoke an action:
 
   ```
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -117,6 +117,8 @@ Review the following steps and examples to create your first JavaScript action.
   }
   ```
 
+  Note that you should not use an activation ID with the flag `--last`.
+
 7. If you forget to record the activation ID, you can get a list of activations ordered from the most recent to the oldest. Run the following command to get a list of your activations:
 
   ```
@@ -621,7 +623,7 @@ follows:
 wsk action create helloSwift hello.swift
 ```
 
-The CLI automatically infers the type of the action from the source file extension. For `.swift` source files, the action runs using a Swift 3.1.1 runtime. You can also create an action that runs with Swift 3.0 by explicitly specifying the parameter `--kind swift:3`. See the Swift [reference](./reference.md#swift-actions) for more information about Swift 3.0 vs. 3.1. 
+The CLI automatically infers the type of the action from the source file extension. For `.swift` source files, the action runs using a Swift 3.1.1 runtime. You can also create an action that runs with Swift 3.0 by explicitly specifying the parameter `--kind swift:3`. See the Swift [reference](./reference.md#swift-actions) for more information about Swift 3.0 vs. 3.1.
 
 **Note:** The actions you created using the kind `swift:3` will continue to work for a short period, however you should begin migrating your deployment scripts and recompiling your swift actions using the new kind `swift:3.1.1`. Support for Swift 3.0 is deprecated and will be removed soon.
 
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -106,7 +106,18 @@ Review the following steps and examples to create your first JavaScript action.
   }
   ```
 
-6. If you forget to record the activation ID, you can get a list of activations ordered from the most recent to the oldest. Run the following command to get a list of your activations:
+6. To access the most recent activation record, activation results or activation logs, use the `--last` or `-l` flag. Run the following command to get your last activation result.
+
+  ```
+  wsk activation result --last
+  ```
+  ```json
+  {
+      "payload": "Hello world"
+  }
+  ```
+
+7. If you forget to record the activation ID, you can get a list of activations ordered from the most recent to the oldest. Run the following command to get a list of your activations:
 
   ```
   wsk activation list
@@ -640,11 +651,11 @@ To avoid the cold-start delay, you can compile your Swift file into a binary and
   ```
   docker run --rm -it -v "$(pwd):/owexec" openwhisk/action-swift-v3.1.1 bash
   ```
-  This puts you in a bash shell within the Docker container. 
+  This puts you in a bash shell within the Docker container.
 
 - Copy the source code and prepare to build it.
   ```
-  cp /owexec/hello.swift /swift3Action/spm-build/main.swift 
+  cp /owexec/hello.swift /swift3Action/spm-build/main.swift
   ```
   ```
   cat /swift3Action/epilogue.swift >> /swift3Action/spm-build/main.swift
@@ -658,7 +669,7 @@ To avoid the cold-start delay, you can compile your Swift file into a binary and
 - (Optional) Create the `Package.swift` file to add dependencies.
   ```swift
   import PackageDescription
-  
+
   let package = Package(
     name: "Action",
         dependencies: [
@@ -699,17 +710,17 @@ and so you should include them in your own `Package.swift`.
   exit
   ```
 
-  This has created hello.zip in the same directory as hello.swift. 
+  This has created hello.zip in the same directory as hello.swift.
 
 - Upload it to OpenWhisk with the action name helloSwifty:
   ```
   wsk action update helloSwiftly hello.zip --kind swift:3.1.1
   ```
 
-- To check how much faster it is, run 
+- To check how much faster it is, run
   ```
   wsk action invoke helloSwiftly --blocking
-  ``` 
+  ```
 
 
 The time it took for the action to run is in the "duration" property and compare to the time it takes to run with a compilation step in the hello action.
@@ -911,7 +922,11 @@ You can use the OpenWhisk CLI to watch the output of actions as they are invoked
   wsk activation poll
   ```
 
-  This command starts a polling loop that continuously checks for logs from activations.
+  This command starts a polling loop that continuously checks for logs from activations. Alternatively, a poll can be done for a specific activation id once it has been invoked by adding the activation id.
+
+  ```
+  wsk activation poll [ACTIVATION_ID]
+  ```
 
 2. Switch to another window and invoke an action:
 
@@ -940,7 +955,7 @@ You can list all the actions that you have created using:
 wsk action list
 ```
 
-As you write more actions, this list gets longer and it can be helpful to group related actions into [packages](./packages.md). To filter your list of actions to just the those within a specific pacakge, you can use: 
+As you write more actions, this list gets longer and it can be helpful to group related actions into [packages](./packages.md). To filter your list of actions to just the those within a specific pacakge, you can use:
 
 ```
 wsk action list [PACKAGE NAME]
--- a/tests/src/test/scala/common/Wsk.scala
+++ b/tests/src/test/scala/common/Wsk.scala
@@ -559,6 +559,7 @@ class WskActivation()
      * @param activationId the activation id
      * @param expectedExitCode (optional) the expected exit code for the command
      * if the code is anything but DONTCARE_EXIT, assert the code is as expected
+     * @param last retrieves latest acitvation
      */
     def get(
         activationId: Option[String] = None,
@@ -579,6 +580,7 @@ class WskActivation()
      * @param activationId the activation id
      * @param expectedExitCode (optional) the expected exit code for the command
      * if the code is anything but DONTCARE_EXIT, assert the code is as expected
+     * @param last retrieves latest acitvation
      */
     def logs(
         activationId: Option[String] = None,
@@ -597,6 +599,7 @@ class WskActivation()
      * @param activationId the activation id
      * @param expectedExitCode (optional) the expected exit code for the command
      * if the code is anything but DONTCARE_EXIT, assert the code is as expected
+     * @param last retrieves latest acitvation 
      */
     def result(
         activationId: Option[String] = None,
--- a/tests/src/test/scala/common/Wsk.scala
+++ b/tests/src/test/scala/common/Wsk.scala
@@ -561,12 +561,16 @@ class WskActivation()
      * if the code is anything but DONTCARE_EXIT, assert the code is as expected
      */
     def get(
-        activationId: String,
+        activationId: Option[String] = None,
         expectedExitCode: Int = SUCCESS_EXIT,
-        fieldFilter: Option[String] = None)(
+        fieldFilter: Option[String] = None,
+        last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
-        val params = { fieldFilter map { f => Seq(f) } getOrElse Seq() }
-        cli(wp.overrides ++ Seq(noun, "get", "--auth", wp.authKey, activationId) ++ params, expectedExitCode)
+        val params =
+          {activationId map {a => Seq(a)} getOrElse Seq() } ++
+          { fieldFilter map { f => Seq(f) } getOrElse Seq() } ++
+          {last map {l => Seq("--last")} getOrElse Seq() }
+        cli(wp.overrides ++ Seq(noun, "get", "--auth", wp.authKey) ++ params, expectedExitCode)
     }
 
     /**
@@ -577,10 +581,14 @@ class WskActivation()
      * if the code is anything but DONTCARE_EXIT, assert the code is as expected
      */
     def logs(
-        activationId: String,
-        expectedExitCode: Int = SUCCESS_EXIT)(
+        activationId: Option[String] = None,
+        expectedExitCode: Int = SUCCESS_EXIT,
+        last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
-        cli(wp.overrides ++ Seq(noun, "logs", activationId, "--auth", wp.authKey), expectedExitCode)
+        val params =
+          {activationId map {a => Seq(a)} getOrElse Seq() } ++
+          {last map {l => Seq("--last")} getOrElse Seq() }
+        cli(wp.overrides ++ Seq(noun, "logs", "--auth", wp.authKey) ++ params, expectedExitCode)
     }
 
     /**
@@ -591,10 +599,14 @@ class WskActivation()
      * if the code is anything but DONTCARE_EXIT, assert the code is as expected
      */
     def result(
-        activationId: String,
-        expectedExitCode: Int = SUCCESS_EXIT)(
+        activationId: Option[String] = None,
+        expectedExitCode: Int = SUCCESS_EXIT,
+        last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
-        cli(wp.overrides ++ Seq(noun, "result", activationId, "--auth", wp.authKey), expectedExitCode)
+        val params =
+          {activationId map {a => Seq(a)} getOrElse Seq() } ++
+          {last map {l => Seq("--last")} getOrElse Seq() }
+        cli(wp.overrides ++ Seq(noun, "result", "--auth", wp.authKey) ++ params, expectedExitCode)
     }
 
     /**
--- a/tests/src/test/scala/common/Wsk.scala
+++ b/tests/src/test/scala/common/Wsk.scala
@@ -568,9 +568,9 @@ class WskActivation()
         last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
         val params =
-          { activationId map {a => Seq(a)} getOrElse Seq() } ++
+          { activationId map { a => Seq(a) } getOrElse Seq() } ++
           { fieldFilter map { f => Seq(f) } getOrElse Seq() } ++
-          { last map {l => Seq("--last")} getOrElse Seq() }
+          { last map { l => Seq("--last") } getOrElse Seq() }
         cli(wp.overrides ++ Seq(noun, "get", "--auth", wp.authKey) ++ params, expectedExitCode)
     }
 
@@ -588,8 +588,8 @@ class WskActivation()
         last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
         val params =
-          { activationId map {a => Seq(a)} getOrElse Seq() } ++
-          { last map {l => Seq("--last")} getOrElse Seq() }
+          { activationId map { a => Seq(a) } getOrElse Seq() } ++
+          { last map { l => Seq("--last") } getOrElse Seq() }
         cli(wp.overrides ++ Seq(noun, "logs", "--auth", wp.authKey) ++ params, expectedExitCode)
     }
 
@@ -607,8 +607,8 @@ class WskActivation()
         last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
         val params =
-          { activationId map {a => Seq(a)} getOrElse Seq() } ++
-          { last map {l => Seq("--last")} getOrElse Seq() }
+          { activationId map { a => Seq(a) } getOrElse Seq() } ++
+          { last map { l => Seq("--last") } getOrElse Seq() }
         cli(wp.overrides ++ Seq(noun, "result", "--auth", wp.authKey) ++ params, expectedExitCode)
     }
 
--- a/tests/src/test/scala/common/Wsk.scala
+++ b/tests/src/test/scala/common/Wsk.scala
@@ -568,9 +568,9 @@ class WskActivation()
         last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
         val params =
-          {activationId map {a => Seq(a)} getOrElse Seq() } ++
+          { activationId map {a => Seq(a)} getOrElse Seq() } ++
           { fieldFilter map { f => Seq(f) } getOrElse Seq() } ++
-          {last map {l => Seq("--last")} getOrElse Seq() }
+          { last map {l => Seq("--last")} getOrElse Seq() }
         cli(wp.overrides ++ Seq(noun, "get", "--auth", wp.authKey) ++ params, expectedExitCode)
     }
 
@@ -588,8 +588,8 @@ class WskActivation()
         last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
         val params =
-          {activationId map {a => Seq(a)} getOrElse Seq() } ++
-          {last map {l => Seq("--last")} getOrElse Seq() }
+          { activationId map {a => Seq(a)} getOrElse Seq() } ++
+          { last map {l => Seq("--last")} getOrElse Seq() }
         cli(wp.overrides ++ Seq(noun, "logs", "--auth", wp.authKey) ++ params, expectedExitCode)
     }
 
@@ -599,7 +599,7 @@ class WskActivation()
      * @param activationId the activation id
      * @param expectedExitCode (optional) the expected exit code for the command
      * if the code is anything but DONTCARE_EXIT, assert the code is as expected
-     * @param last retrieves latest acitvation 
+     * @param last retrieves latest acitvation
      */
     def result(
         activationId: Option[String] = None,
@@ -607,8 +607,8 @@ class WskActivation()
         last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
         val params =
-          {activationId map {a => Seq(a)} getOrElse Seq() } ++
-          {last map {l => Seq("--last")} getOrElse Seq() }
+          { activationId map {a => Seq(a)} getOrElse Seq() } ++
+          { last map {l => Seq("--last")} getOrElse Seq() }
         cli(wp.overrides ++ Seq(noun, "result", "--auth", wp.authKey) ++ params, expectedExitCode)
     }
 
--- a/tests/src/test/scala/system/basic/WskSequenceTests.scala
+++ b/tests/src/test/scala/system/basic/WskSequenceTests.scala
@@ -216,7 +216,7 @@ class WskSequenceTests
                     result.fields.get("error") shouldBe Some(JsString(sequenceIsTooLong))
                     // check that inner sequence had only (limit - 1) activations
                     val innerSeq = activation.logs.get(1) // the id of the inner sequence activation
-                    val getInnerSeq = wsk.activation.get(innerSeq)
+                    val getInnerSeq = wsk.activation.get(Some(innerSeq))
                     withActivation(wsk.activation, getInnerSeq, totalWait = allowedActionDuration) {
                         innerSeqActivation =>
                             innerSeqActivation.logs.get.size shouldBe (limit - 1)
@@ -510,7 +510,7 @@ class WskSequenceTests
         withActivation(wsk.activation, run, totalWait = 2 * allowedActionDuration) { activation =>
             checkSequenceLogsAndAnnotations(activation, 1)
             val componentId = activation.logs.get(atomicActionIdx)
-            val getComponentActivation = wsk.activation.get(componentId)
+            val getComponentActivation = wsk.activation.get(Some(componentId))
             withActivation(wsk.activation, getComponentActivation, totalWait = allowedActionDuration) { componentActivation =>
                 println(componentActivation)
                 componentActivation.logs shouldBe defined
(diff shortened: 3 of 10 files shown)


dubeejw:
@csantanapr, please add me as reviewer here.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -346,6 +385,11 @@ func init() {
     activationListCmd.Flags().Int64Var(&flags.activation.since, "since", 0, wski18n.T("return activations with timestamps later than `SINCE`; measured in milliseconds since Th, 01, Jan 1970"))
 
     activationGetCmd.Flags().BoolVarP(&flags.common.summary, "summary", "s", false, wski18n.T("summarize activation details"))
+    activationGetCmd.Flags().BoolVarP(&flags.activation.last, "last", "l", false, wski18n.T("retrieves last activation"))
```
dubeejw: Need to add the description string to the externalized strings JSON file.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -114,7 +114,11 @@ var activationGetCmd = &cobra.Command{
     PreRunE: setupClientConfig,
     RunE: func(cmd *cobra.Command, args []string) error {
         var field string
+        args,lerr := lastFlag(args)
 
+        if lerr != nil {
```
dubeejw: `if args, err := lastFlag(args); err != nil {`

dubeejw:
Will need doc updates to info users that this new functionality exists. You may need to create an activation doc file as I do not see one in `https://github.com/apache/incubator-openwhisk/blob/master/docs/`.

Also need tests for everything that is added here.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -164,7 +168,6 @@ var activationGetCmd = &cobra.Command{
                 printJSON(activation)
             }
         }
-
```
dubeejw: Leave the break here.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -176,7 +179,11 @@ var activationLogsCmd = &cobra.Command{
     SilenceErrors:  true,
     PreRunE: setupClientConfig,
     RunE: func(cmd *cobra.Command, args []string) error {
+        args, lerr := lastFlag(args)
 
+        if lerr != nil {
```
dubeejw: Use `if args, err = lastFlag(args); err != nil {` for all of these.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -114,7 +114,11 @@ var activationGetCmd = &cobra.Command{
     PreRunE: setupClientConfig,
     RunE: func(cmd *cobra.Command, args []string) error {
         var field string
+        var err error
 
+        if args, err = lastFlag(args); err != nil {
+          return err
```
dubeejw: @mdeuser, do we want a debug message in each of these if statements stating an error occurred?

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -346,6 +386,11 @@ func init() {
     activationListCmd.Flags().Int64Var(&flags.activation.since, "since", 0, wski18n.T("return activations with timestamps later than `SINCE`; measured in milliseconds since Th, 01, Jan 1970"))
 
     activationGetCmd.Flags().BoolVarP(&flags.common.summary, "summary", "s", false, wski18n.T("summarize activation details"))
+    activationGetCmd.Flags().BoolVarP(&flags.activation.last, "last", "l", false, wski18n.T("retrieves the last activation"))
+
+    activationLogsCmd.Flags().BoolVarP(&flags.activation.last, "last", "l", false, wski18n.T("retrieves the last activation"))
+
+    activationResultCmd.Flags().BoolVarP(&flags.activation.last, "last", "l", false, wski18n.T("retrieves the last activation"))
```
mdeuser: There's another issue related to `wsk activation poll` in which a single activation id can be used as the poll target.  The `--last` option would also then apply to the `poll` command

rabbah: That's a nice feature; with poll exiting Once result is ready.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -224,6 +235,35 @@ var activationResultCmd = &cobra.Command{
         return nil
     },
 }
+//lastFlag(args) retrieves the last activation with flag -l or --last
+//Param: Brings in []strings from args
+//Return: Returns an args([]string) with last ID used or the original args with any errors
+func lastFlag(args []string) ([]string, error) {
+  //Checks to see if there is no ID sent with the last flag
+  //If an ID is given with the last flag then the ID will overide the flag
+  if flags.activation.last && len(args) == 0  {
```
mdeuser: What is behavior when activation id and `--last` are both supplied as cli arguments?  This seems conflicting and should be flagged as an usage error.

tools/cli/go-whisk-cli/commands/activation.go
mdeuser: The `lastFlag` function seems to have some decent debug that seems sufficient.  That said, when an error occurs, the final failure message to the user should be that this command failed with the failure reason.

tools/cli/go-whisk-cli/commands/activation.go
bpoole16: Activation ID will be used if both an ID and --last are given.

mdeuser: This might cause some confusion.  Does the user know which one was selected when conflicting id/--last arguments are used on the same command line.  IMHO, this should be flagged as a usage error and not be permitted.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -254,12 +255,17 @@ func lastFlag(args []string) ([]string, error) {
     }
 
     if len(activations) == 0 {
-      whiskErr := whisk.MakeWskError(errors.New("No activations found"), whisk.EXITCODE_ERR_GENERAL, whisk.DISPLAY_MSG, whisk.DISPLAY_USAGE)
+      errStr := fmt.Sprintf("%s%s %s %s", "clinet.Activation.",funcName,"error:", "No activations found")
```
dubeejw: The user should not be informed of the function name directly through an error message. Such information should be displayed in debug messages. Also, the error message is not very clear to someone that doesn't know the internal workings of the CLI.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -236,11 +237,11 @@ var activationResultCmd = &cobra.Command{
     },
 }
 //lastFlag(args) retrieves the last activation with flag -l or --last
-//Param: Brings in []strings from args
+//Param: Brings in []strings from args and the name of the function that is using it (for error feedback)
 //Return: Returns an args([]string) with last ID used or the original args with any errors
-func lastFlag(args []string) ([]string, error) {
+func lastFlag(args []string, funcName string) ([]string, error) {
```
dubeejw: I don't think `funcName` is necessary here. Easier to just put a debug message in each of the if statement for the function calls.

docs/actions.md
```diff
@@ -116,6 +116,17 @@ Review the following steps and examples to create your first JavaScript action.
   44794bd6aab74415b4e42a308d880e5b         hello
   6bf1f670ee614a7eb5af3c9fde813043         hello
   ```
+7. If you just need access your last activation you can use the flag `--last` or `-l` to retrieve your last activation with get, logs and result. Run the following command to get your last activation result.
```
dubeejw: "...need access `to` your..." To many `you` and `yours` in here.

tools/cli/go-whisk-cli/commands/activation.go
bpoole16: Would something like "No activations found. Be sure wsk activation list contains activations" be more clear? This is simply checking to see if they have any activations and if not alerting the user that they don't.

tools/cli/go-whisk-cli/commands/activation.go
bpoole16: I originally tried to put each error/debug in each separate function but I wasn't able to get `"Unable to get result for activation: {{.err}}"` to display with the actual error message (it would only display the literal {{.err}}) .

docs/actions.md
mdeuser: have this step immediately follow after step 5 (both talking about activation results).  suggested language for the text `To access the most recent activation record, activation results or activation logs, use the --last or -l flag.`

mdeuser:
Affected commands' usage syntax need to be changed.  With the introduction of the `--last` option, the `ACTIVATION_ID` argument is now optional.

Also, need a check to handle when ACTIVATION_ID and and --last are provided at the same time.

docs/actions.md
```diff
@@ -106,7 +106,18 @@ Review the following steps and examples to create your first JavaScript action.
   }
   ```
 
-6. If you forget to record the activation ID, you can get a list of activations ordered from the most recent to the oldest. Run the following command to get a list of your activations:
+  6. To access the most recent activation record, activation results or activation logs, use the `--last` or `-l` flag. Run the following command to get your last activation result.
```
dubeejw: Looks like there is an extra space in front of the 6.

bpoole16:
@mdeuser This checks the conflict between ID and --last 
`else if flags.activation.last && len(args) == 1 {  //Checks conflict between ID and --last
      whiskErr := whisk.MakeWskError(errors.New("Can't use given ID with --last"), whisk.EXITCODE_ERR_GENERAL, whisk.DISPLAY_MSG, whisk.DISPLAY_USAGE)
      return args, whiskErr
  }
  retu`

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -224,6 +248,35 @@ var activationResultCmd = &cobra.Command{
...
+      return args, err
+    }
+    if len(activations) == 0 {  //Checks to to see if there are activations
+      whiskErr := whisk.MakeWskError(errors.New("Activation list is empty"), whisk.EXITCODE_ERR_GENERAL, whisk.DISPLAY_MSG, whisk.DISPLAY_USAGE)
+      return args, whiskErr
+    } else {
+      whisk.Debug(whisk.DbgInfo, "Appending most recent activation ID into args\n")
+      args = append(args, activations[0].ActivationID)
+    }
+  } else if flags.activation.last && len(args) == 1 {  //Checks conflict between ID and --last
```
mdeuser: A `wsk activation get --last` command may have a `FILTER_FIELD` value.  In this case, the check for a command conflict will fail this command.

In the error message, qualify the `ID` - either use the usage label `ACTIVATION_ID` or just spell it out.. i.e. "activation ID".  Message update: `When specifying an activation ID, do not use the --last flag`

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -224,6 +248,35 @@ var activationResultCmd = &cobra.Command{
...
+      whiskErr := whisk.MakeWskError(errors.New("Can't use given ID with --last"), whisk.EXITCODE_ERR_GENERAL, whisk.DISPLAY_MSG, whisk.DISPLAY_USAGE)
```
dubeejw: Currently we can pass an activation ID and a field filter to the activation get command. However, your change break the field filter functionality.

Field filtering with activation get:
```
$ wsk activation get fa613f7e33dd45959b8a93943ea5c3e7 publish
ok: got activation fa613f7e33dd45959b8a93943ea5c3e7, displaying field publish
false
```

Attempting to field filter with your changes:
```
 $ wsk activation get --last  publish
error: Unable to get result for activation: Can't use given ID with --last
Run 'wsk --help' for usage.
```

bpoole16:
Changed conflict code with ID and --last to work with filter fields, updated docs/comments, updated Activation_ID syntax, and added polling for ID with integration of --last flag.

bpoole16:
PG 4 #445

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -98,14 +98,22 @@ var activationListCmd = &cobra.Command{
...
     Short: wski18n.T("get activation"),
     SilenceUsage:   true,
     SilenceErrors:  true,
     PreRunE: setupClientConfig,
     RunE: func(cmd *cobra.Command, args []string) error {
         var field string
+        var err error
 
+        if args, err = lastFlag(args); err != nil {  // Checks if any errors occured in lastFlag(args)
+          whisk.Debug(whisk.DbgError, "client.Activation.Get failed: %s\n", err)
```
dubeejw: Method in debug message should match name of the function called, and add the args to the message.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -161,13 +169,21 @@ var activationGetCmd = &cobra.Command{
...
+    Use:   "logs (ACTIVATION_ID | --last)",
     Short: wski18n.T("get the logs of an activation"),
     SilenceUsage:   true,
     SilenceErrors:  true,
     PreRunE: setupClientConfig,
     RunE: func(cmd *cobra.Command, args []string) error {
+        var err error
 
+        if args, err = lastFlag(args); err != nil {  // Checks if any errors occured in lastFlag(args)
+          whisk.Debug(whisk.DbgError, "client.Activation.Logs failed: %s\n", err)
```
dubeejw: Same as comment above.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -189,13 +205,21 @@ var activationLogsCmd = &cobra.Command{
...
+    Use:   "result (ACTIVATION_ID | --last)",
     Short: "get the result of an activation",
     SilenceUsage:   true,
     SilenceErrors:  true,
     PreRunE: setupClientConfig,
     RunE: func(cmd *cobra.Command, args []string) error {
+        var err error
 
+        if args, err = lastFlag(args); err != nil {  // Checks if any errors occured in lastFlag(args)
+          whisk.Debug(whisk.DbgError, "client.Activation.Result failed: %s\n", err)
```
dubeejw: Same as comment above.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -216,6 +240,47 @@ var activationResultCmd = &cobra.Command{
...
+      Skip: 0,
+    }
+    activations,_, err := client.Activations.List(options)
+    if err != nil {   // Checks Activations.List for errors when retrieving latest activaiton
+      whisk.Debug(whisk.DbgError, "client.Activations.List() error for flag --last: %s\n", err)
+      return args, err
+    }
+    if len(activations) == 0 {    // Checks to to see if there are activations available
+      whisk.Debug(whisk.DbgError, "No activations found in activation list\n")
+      errStr := wski18n.T("Activation list contains no activations")
```
dubeejw: Maybe something like: `Activation list does not contain any activations.`.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -216,6 +240,47 @@ var activationResultCmd = &cobra.Command{
...
+      whiskErr := whisk.MakeWskError(errors.New(errStr), whisk.EXITCODE_ERR_GENERAL, whisk.DISPLAY_MSG, whisk.DISPLAY_USAGE)
+      return args, whiskErr
+    } else {
+      if len(args) == 0 {
+        whisk.Debug(whisk.DbgInfo, "Appending most recent activation ID into args\n")
```
dubeejw: Should probably print the arguments you are appending in the debug messages in this function.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -216,6 +240,47 @@ var activationResultCmd = &cobra.Command{
     },
 }
 
+// lastFlag(args) retrieves the last activation with flag -l or --last
+// Param: Brings in []strings from args
+// Return: Returns a []string with the latest ID or the original args and any errors
+func lastFlag(args []string) ([]string, error) {
+  // Checks to see if there is an ID sent with the --last
+  // If an ID is given with --last then an error will be thrown
+  if  flags.activation.last {
```
dubeejw: Remove a space after `if`.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -216,6 +240,47 @@ var activationResultCmd = &cobra.Command{
...
+    options := &whisk.ActivationListOptions {
+      Limit: 1,
+      Skip: 0,
+    }
+    activations,_, err := client.Activations.List(options)
+    if err != nil {   // Checks Activations.List for errors when retrieving latest activaiton
+      whisk.Debug(whisk.DbgError, "client.Activations.List() error for flag --last: %s\n", err)
```
dubeejw: Show arguments passed to the method in debug statements.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -216,6 +240,47 @@ var activationResultCmd = &cobra.Command{
...
+    if len(activations) == 0 {    // Checks to to see if there are activations available
+      whisk.Debug(whisk.DbgError, "No activations found in activation list\n")
+      errStr := wski18n.T("Activation list contains no activations")
+      whiskErr := whisk.MakeWskError(errors.New(errStr), whisk.EXITCODE_ERR_GENERAL, whisk.DISPLAY_MSG, whisk.DISPLAY_USAGE)
+      return args, whiskErr
+    } else {
+      if len(args) == 0 {
+        whisk.Debug(whisk.DbgInfo, "Appending most recent activation ID into args\n")
+        args = append(args, activations[0].ActivationID)
+      } else if _,_,err := client.Activations.Get(args[0]); len(args) == 1 && err != nil{   // Checks to see if filter field is given and not ID
```
dubeejw: Looks like maybe you should check the arg length before making calls to the server.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -216,6 +240,47 @@ var activationResultCmd = &cobra.Command{
...
+          whisk.Debug(whisk.DbgInfo, "Appending most recent activation ID into args\n")
+          args = append(args, activations[0].ActivationID)
+          whisk.Debug(whisk.DbgInfo, "Allocating appended ID to correct position in args\n")
+          args[0], args[1] = args[1], args[0]   // IDs should be located at args[0], if a filter field is given it has to be moved to args[0]
+      } else if _,_,err := client.Activations.Get(args[0]);  len(args) >= 1 && err == nil {    // Checks conflict between ID, field filter and --last
```
dubeejw: You should probably check to see if the last flag and an activation ID exists in the args before you do anything of this processing.

tools/cli/go-whisk-cli/commands/activation.go
bpoole16: I could do a for loop or something checking the length of the arguments to see if they are an ID (length of 32) or filter field (any length).

tools/cli/go-whisk-cli/commands/activation.go
bpoole16: I could do a for loop or something checking the length of the arguments to see if they are an ID (length of 32) or filter field (any length).

tools/cli/go-whisk-cli/commands/activation.go
bpoole16: I could do a for loop or something checking the length of the arguments to see if they are an ID (length of 32) or filter field (any length). The point of the system calls were to check if they were returning errors due to filter field being sent instead of ID.

rabbah:
for all the style related comments, https://github.com/apache/incubator-openwhisk/tree/master/tools/eclipse has an eclipse code formatting style. maybe we need automatic style checks now.

docs/actions.md
```diff
@@ -887,9 +898,7 @@ You can use the OpenWhisk CLI to watch the output of actions as they are invoked
   ```
   wsk activation poll
   ```
-
-  This command starts a polling loop that continuously checks for logs from activations.
```
bpoole16: Looks like it got deleted when I was removed polling + --last doc stuff, will add back.

bpoole16:
PG #821

tests/src/test/scala/common/Wsk.scala
```diff
@@ -557,14 +557,19 @@ class WskActivation()
...
+        activationId: Option[String] = None,
         expectedExitCode: Int = SUCCESS_EXIT,
-        fieldFilter: Option[String] = None)(
+        fieldFilter: Option[String] = None,
+        last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
-        val params = { fieldFilter map { f => Seq(f) } getOrElse Seq() }
-        cli(wp.overrides ++ Seq(noun, "get", "--auth", wp.authKey, activationId) ++ params, expectedExitCode)
+        val params =
+          { activationId map {a => Seq(a)} getOrElse Seq() } ++
```
dubeejw: Space needed after `activationId map {`, and after `Seq(a)`.

tests/src/test/scala/common/Wsk.scala
```diff
@@ -573,12 +578,17 @@ class WskActivation()
...
     def logs(
-        activationId: String,
-        expectedExitCode: Int = SUCCESS_EXIT)(
+        activationId: Option[String] = None,
+        expectedExitCode: Int = SUCCESS_EXIT,
+        last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
-        cli(wp.overrides ++ Seq(noun, "logs", activationId, "--auth", wp.authKey), expectedExitCode)
+        val params =
+          { activationId map {a => Seq(a)} getOrElse Seq() } ++
```
dubeejw: Space needed after `activationId map {`, and after `Seq(a)`.

tests/src/test/scala/common/Wsk.scala
```diff
@@ -587,12 +597,17 @@ class WskActivation()
...
     def result(
-        activationId: String,
-        expectedExitCode: Int = SUCCESS_EXIT)(
+        activationId: Option[String] = None,
+        expectedExitCode: Int = SUCCESS_EXIT,
+        last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
-        cli(wp.overrides ++ Seq(noun, "result", activationId, "--auth", wp.authKey), expectedExitCode)
+        val params =
+          { activationId map {a => Seq(a)} getOrElse Seq() } ++
```
dubeejw: Space needed after `activationId map {`, and after `Seq(a)`.

tests/src/test/scala/common/Wsk.scala
```diff
@@ -587,12 +597,17 @@ class WskActivation()
...
+          { last map {l => Seq("--last")} getOrElse Seq() }
```
dubeejw: Space needed after `last map {`, and after `Seq("--last")`.

tests/src/test/scala/common/Wsk.scala
```diff
@@ -573,12 +578,17 @@ class WskActivation()
...
-        activationId: String,
-        expectedExitCode: Int = SUCCESS_EXIT)(
+        activationId: Option[String] = None,
+        expectedExitCode: Int = SUCCESS_EXIT,
+        last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
-        cli(wp.overrides ++ Seq(noun, "logs", activationId, "--auth", wp.authKey), expectedExitCode)
+        val params =
+          { activationId map {a => Seq(a)} getOrElse Seq() } ++
+          { last map {l => Seq("--last")} getOrElse Seq() }
```
dubeejw: Space needed after `last map {`, and after `Seq("--last")`.

tests/src/test/scala/common/Wsk.scala
```diff
@@ -557,14 +557,19 @@ class WskActivation()
...
-        fieldFilter: Option[String] = None)(
+        fieldFilter: Option[String] = None,
+        last: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
-        val params = { fieldFilter map { f => Seq(f) } getOrElse Seq() }
-        cli(wp.overrides ++ Seq(noun, "get", "--auth", wp.authKey, activationId) ++ params, expectedExitCode)
+        val params =
+          { activationId map {a => Seq(a)} getOrElse Seq() } ++
+          { fieldFilter map { f => Seq(f) } getOrElse Seq() } ++
+          { last map {l => Seq("--last")} getOrElse Seq() }
```
dubeejw: Space needed after `last map {`, and after `Seq("--last")`.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -101,14 +101,22 @@ var activationListCmd = &cobra.Command{
...
     Short: wski18n.T("get activation"),
     SilenceUsage:   true,
     SilenceErrors:  true,
     PreRunE: setupClientConfig,
     RunE: func(cmd *cobra.Command, args []string) error {
         var field string
+        var err error
 
+        if args, err = lastFlag(args); err != nil {  // Checks if any errors occured in lastFlag(args)
+          whisk.Debug(whisk.DbgError, "client.Activation.Get(%#v) failed: %s\n", args, err)
```
dubeejw: Debug message should contain `lastFlag` instead of `client.Activation.Get` since `lastFlag` is the method you are calling, and allows us to trace the code execution path through debug messages. Same applies to the other instances were `lastFlag` is called below.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -219,6 +243,45 @@ var activationResultCmd = &cobra.Command{
...
+        } else {
+            if len(args) == 0 {
+                whisk.Debug(whisk.DbgInfo, "Appending most recent activation ID(%s) into args\n", activations[0].ActivationID)
+                args = append(args, activations[0].ActivationID)
+            } else if len(args) == 1 && len(args[0]) != 32 {    // Checks to see if filter field is given and not ID
+                  whisk.Debug(whisk.DbgInfo, "Appending most recent activation ID(%s) into args\n", activations[0].ActivationID)
+                  args = append(args, activations[0].ActivationID)
+                  whisk.Debug(whisk.DbgInfo, "Allocating appended ID to correct position in args\n")
+                  args[0], args[1] = args[1], args[0]    // IDs should be located at args[0], if a filter field is given it has to be moved to args[0]
+            } else if len(args[0]) == 32 {    // Checks conflict between ID and --last
```
dubeejw: Same as above if the field filter is 32 characters long, this doesn't work.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -219,6 +243,45 @@ var activationResultCmd = &cobra.Command{
...
+        if len(activations) == 0 {    // Checks to to see if there are activations available
+            whisk.Debug(whisk.DbgError, "No activations found in activation list\n")
+            errStr := wski18n.T("Activation list does not contain any activations.")
+            whiskErr := whisk.MakeWskError(errors.New(errStr), whisk.EXITCODE_ERR_GENERAL, whisk.DISPLAY_MSG, whisk.DISPLAY_USAGE)
+            return args, whiskErr
+        } else {
+            if len(args) == 0 {
+                whisk.Debug(whisk.DbgInfo, "Appending most recent activation ID(%s) into args\n", activations[0].ActivationID)
+                args = append(args, activations[0].ActivationID)
+            } else if len(args) == 1 && len(args[0]) != 32 {    // Checks to see if filter field is given and not ID
```
dubeejw: What if there is a field filter that is 32 characters long?

tools/cli/go-whisk-cli/commands/activation.go
dubeejw: Maybe better to prepend the retrieved activation ID to the args always. I think we will only have an issue if a user specified an activation ID with the `--last` flag. Under that scenario they user would get an invalid field filter message. However, the usage says not to include an activation ID with the `--last` flag. @mdeuser, what do you think?

bpoole16: Not sure how to distinguish them then, was originally using `_,_,err := client.Activations.Get(args[0])` to see if it was an ID and if it wasn't an ID it would return with an error.

bpoole16: Found a way to use the fieldExists method in util to check if the args[0] is holding a filter field or not. Seems like a sound way to get around this issue.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -262,24 +262,26 @@ func lastFlag(args []string) ([]string, error) {
...
-                  whiskErr := whisk.MakeWskError(errors.New(errStr), whisk.EXITCODE_ERR_GENERAL, whisk.DISPLAY_MSG, whisk.DISPLAY_USAGE)
-                  return args, whiskErr
-            }
         }
+        if len(args) == 0 {
+            whisk.Debug(whisk.DbgInfo, "Appending most recent activation ID(%s) into args\n", activations[0].ActivationID)
+            args = append(args, activations[0].ActivationID)
+        } else {
+			filterExists := fieldExists(&whisk.Activation{}, args[0])
+			if filterExists == true {    // Checks to see if filter field is given and not ID
```
bpoole16: Not sure why it is doing this, Doesn't look like this in my IDE.

mdeuser: @bpoole16 - be sure your ide translates tabs into spaces.

tools/cli/go-whisk-cli/commands/activation.go
```diff
@@ -219,6 +243,47 @@ var activationResultCmd = &cobra.Command{
...
+        } else {
+            filterExists := fieldExists(&whisk.Activation{}, args[0])
+            if filterExists == true {    // Checks to see if filter field is given and not ID
+                whisk.Debug(whisk.DbgInfo, "Appending most recent activation ID(%s) into args\n", activations[0].ActivationID)
+                args = append(args, activations[0].ActivationID)
+                whisk.Debug(whisk.DbgInfo, "Allocating appended ID to correct position in args\n")
+                args[0], args[1] = args[1], args[0]    // IDs should be located at args[0], if a filter field is given it has to be moved to args[0]
+            } else if filterExists == false {    // Checks conflict between ID and --last
+                whisk.Debug(whisk.DbgError,"Can not use activation ID with --last flag\n")
+                errStr := wski18n.T("When specifying an activation ID, do not use the --last flag.")
```
dubeejw: This isn't necessarily true when an actual field filter is supplied for a filed that doesn't exist. Ex:
```
wsk activation get --last bogusField
```

dubeejw: Can probably just drop the field filtering checks as the actual command should handle this check.

dubeejw: Also `logs` and `result` doesn't support a field filter.

bpoole16: I'm using the filter field to check to see if it is an ID or a filter field, without it when you use `wsk activation get ACTIVATION_ID --last FIELD_FILTER` and it will use the activation_id given over the flag `--last`. If we are just summing that up to user error then I don't need the check otherwise it will need to check and see if an ID or field filter is given.

dubeejw: I am incline to let to say that is a user error and not provide error handling for it as there is no way to tell the difference between an activation ID and a field filter.

bpoole16: Works for me.

dubeejw:
@bpoole16, you may want to open an issue that states that we could move the field filter to a flag. With that functionality we will be able to tell that the field filter is not the activation ID.  See examples below:

```
wsk activation get cc5ea085c7824846959269f4a4c14e6f --fieldFilter response
```


```
wsk activation get--last --fieldFilter response
```

tools/cli/build.gradle
```diff
@@ -38,9 +38,12 @@ task copyCLIShortcut(type: Copy, dependsOn: [distBinary, dumpOSInfo]) {
     String go_osarch     = mapOsArchToGoArch(getOsArch())
     String from_path_wsk = "${projectDir}/../../bin/${go_osname}/${go_osarch}/wsk"
     String to_path_dir   = "${projectDir}/../../bin"
-
-    from from_path_wsk
-    into to_path_dir
+    if (file(from_path_wsk).exists()) {
+      exec {
```
dubeejw: This exec cal doesn't seem to build the CLI on a mac.

dubeejw: @bpoole16, Remove the changes to the build.gradle and I say we can merge this one after Travis passes. Will resolve your build issues in a different PR.

bpoole16: Okay, will do.

dubeejw:
@bjustin-ibm, please merge.

tests/src/test/scala/system/basic/WskSequenceTests.scala
```diff
@@ -216,7 +216,7 @@ class WskSequenceTests
                     result.fields.get("error") shouldBe Some(JsString(sequenceIsTooLong))
                     // check that inner sequence had only (limit - 1) activations
                     val innerSeq = activation.logs.get(1) // the id of the inner sequence activation
-                    val getInnerSeq = wsk.activation.get(innerSeq)
+                    val getInnerSeq = wsk.activation.get(Some(innerSeq))
```
rabbah: hmm seems wrapping all ids with Some is too common - should there be an overloaded `get` instead of changing all the test cases?

rabbah: counted only 10... ok maybe not as bad as I thought.

PR #2338, opened 2017-06-07
Prepare cache to change it by configuration

cbickel:
This PR changes the Cache to be a class instead of a trait.
With this change, it will be easier to change to another cache by configuration.

The configuration or the implementation of another cache is not part of this PR.

PG3#892 is 🔵 
PG4#593 is 🔵


common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -146,138 +142,127 @@ trait MultipleReadersSingleWriterCache[W, Winfo] {
...
+
+            actualEntry.state.get match {
+                case Cached =>
+                    logger.debug(this, "cached read")
+                    makeNoteOfCacheHit(key)
+                    actualEntry.unpack.map { entry =>
+                        if (entry.getClass != mw.runtimeClass) {
+                            throw DocumentTypeMismatchException(s"document type ${entry.getClass} did not match expected type ${mw.runtimeClass}.")
                         }
+                        entry.asInstanceOf[Wsub]
```
markusthoemmes: Is this needed as a runtime check? Can this be checked statically?

cbickel: As the Cache is a cache for all WhiskEntities, there are Rules, Triggers and Actions in one Cache.
On getting an element by id, we should be sure, that the element the cache found has the expected type.

cbickel: It's the same like getting the WhiskEntity out of the Datastore:
https://github.com/apache/incubator-openwhisk/blob/master/common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala#L148

common/scala/src/main/scala/whisk/core/database/CouchDbStoreProvider.scala
```diff
@@ -23,20 +23,25 @@ import whisk.common.Logging
...
         require(config != null && config.isValid, "config is undefined or not valid")
         require(config.dbProvider == "Cloudant" || config.dbProvider == "CouchDB", "Unsupported db.provider: " + config.dbProvider)
         assume(Set(config.dbProtocol, config.dbHost, config.dbPort, config.dbUsername, config.dbPassword, name(config)).forall(_.nonEmpty), "At least one expected property is missing")
 
-        new CouchDbRestStore[D](config.dbProtocol, config.dbHost, config.dbPort.toInt, config.dbUsername, config.dbPassword, name(config))
+        new CouchDbRestStore[D, CacheAbstraction](config.dbProtocol, config.dbHost, config.dbPort.toInt, config.dbUsername, config.dbPassword, name(config), cache)
+    }
+
+    def makeCache[CacheAbstraction](): WhiskCache[CacheAbstraction, DocInfo] = {
+        new MultipleReadersSingleWriterCache[CacheAbstraction, DocInfo]()
```
markusthoemmes: Should we move this implementation to `ArtifactStoreProvider`? After all, it's not relying on implementation details.

cbickel:
Replaced by #2624

PR #2341, opened 2017-06-08
Change the path to be unarchived from local to remote

houshengbo:


daisy-ycguo:
#2342 is better than this one. Closed as requested by author.

PR #2342, opened 2017-06-08
Change the unzip of Linux binary from local to remote

houshengbo:


daisy-ycguo:
Vincent needs this change for a problem he is seeing in certain install/builds.

PR #2345, opened 2017-06-08

houshengbo:
This PR adds the functionalities to either install openwhisk
binaries from the releases page of openwhisk cli or build the
binaries locally.

The key installation_mode of openwhisk_cli dict can be set to
download, meaning to download the binaries, or build, meaning
to build the binaries locally. The default value is download.

Closes-Bug: #2101


rabbah:
How will we test CLI changes when the tests for the CLI are in this repo?

rabbah:
I think this merge is problematic. It created a situation where all the cli tests are in the core repo. So we cannot accept any cli changes because of a circular dependence we can't break easily when cli changes are also required. So for example if there is a cli change and accompanying test changes, the tests in the core repo will be incompatible with the cli version pulled from the cli repo. I think we need to continue building the cli from this local copy until such time when all the accompanying tests are moved to the cli repo.

ansible/group_vars/all
```diff
@@ -1,6 +1,7 @@
 mode: deploy
 prompt_user: true
 openwhisk_home: "{{ lookup('env', 'OPENWHISK_HOME')|default(playbook_dir + '/..', true) }}"
+openwhisk_cli_home: "{{ lookup('env', 'OPENWHISK_CLI')|default(playbook_dir + '/../../incubator-openwhisk-cli', true) }}"
```
rabbah: shouldn't this be relative to this repo?

ansible/group_vars/all
```diff
@@ -228,3 +229,46 @@ catalog_repos:
...
+#
+# The name specifies the package name of the binaries in remote mode.
+#
+# The dest_name specifies the package name of the binaries in Nginx in remote mode.
+#
+# The location specifies the official website where Openwhisk CLI is hosted in
+# remote mode or location to save the binaries of the OpenWhisk CLI in local mode.
+#
+
+openwhisk_cli:
```
rabbah: simply changing the setting to use local vs remote did not work.
it seems that local mode is wrt the path to the binary but the build path is from the cli repo not this repo; and the cli repo will not deposit the binary in this path.

csantanapr:
@rabbah 
>I think this merge is problematic. It created a situation where all the cli tests are in the core repo. So we cannot accept any cli changes because of a circular dependence we can't break easily when cli changes are also required. So for example if there is a cli change and accompanying test changes, the tests in the core repo will be incompatible with the cli version pulled from the cli repo. I think we need to continue building the cli from this local copy until such time when all the accompanying tests are moved to the cli repo.

I agree with further inspecting the other repos.
I think before we start using the two cli repos we need a better testing story.
We have two repos to deal with 
- https://github.com/apache/incubator-openwhisk-client-go
- https://github.com/apache/incubator-openwhisk-cli

For both there should be tests on the repo, and travis integration that takes the source from the same repo and run tests that are located in the sam repo.

As a transition we should duplicate the tests from core to both, and evaluate which set goes where.

Also provide docs for a contributor on how to setup their dev environment and do testing locally

I proposed we revert this change, and what we can do is introduce a mode for the cli ansible role, where we can pass a flag to do the new behavior to allow experimentation and integration into the other cli repos.

houshengbo:
@rabbah @csantanapr How do we define the CLI test cases in openwhisk? Based on my observation, all the test cases use “val wsk = new Wsk” to interact with openwhisk services, for example, whisk.core.database.test.CacheConcurrencyTests. Is this test cases also cli test case and does it need to be moved to cli repo?

rabbah:
There are some tests which are CLI specific as in `WskBasicUsage` and `WskBasicTests` but almost all our tests go through the CLI vs the API. For the sake of the CLI repos, i think you’ll have to iterate to figure out what is actually testing CLI functionality and what isn’t. 

The core repo should perhaps not use the CLI for tests at all since that’s an external dependence, and instead drive all tests through the API; this could be done by injecting a new `Wsk` implementation for example which instead of calling out to the CLI, makes a REST call. I offer no evidence this will be easy to do, but I think is doable and could be one way to approach the separation.

jasonpet:
there are also many external repos that depend on openwhisk to use the cli based test utilities that it currently provides.  they will also need to be updated to use the REST based utils in openwhisk if the cli stuff is all moved out.

PR #2356, opened 2017-06-09
Fix undefined exception in API GW mgmt

mdeuser:


mdeuser:
@csantanapr - a good travis run should be a sufficient test

PR #2357, opened 2017-06-10
Cli should ignore a config file env variable value of empty string

mdeuser:
Also fix test framework to not generate an empty string WSK_CONFIG_FILE env value by default

### DIFF
--- a/tests/src/test/scala/whisk/core/cli/test/WskConfigTests.scala
+++ b/tests/src/test/scala/whisk/core/cli/test/WskConfigTests.scala
@@ -53,6 +53,12 @@ class WskConfigTests extends TestHelpers with WskTestHelpers {
     }
   }
 
+  it should "use default cli configuration when an empty string WSK_CONFIG_FILE is supplied" in {
+    val env = Map("WSK_CONFIG_FILE" -> "")
+    val stderr = wsk.cli(Seq("property", "get", "-i"), env = env, expectedExitCode = ERROR_EXIT).stderr
+    stderr should include("The API host is not valid: An API host must be provided.")
+  }
+
   it should "validate default property values" in {
     val tmpwskprops = File.createTempFile("wskprops", ".tmp")
     val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
--- a/tools/cli/go-whisk-cli/commands/property.go
+++ b/tools/cli/go-whisk-cli/commands/property.go
@@ -402,8 +402,11 @@ func SetDefaultProperties() {
 func GetPropertiesFilePath() (propsFilePath string, werr error) {
     var envExists bool
 
-    // Environment variable overrides the default properties file path
-    if propsFilePath, envExists = os.LookupEnv("WSK_CONFIG_FILE"); envExists == true || propsFilePath != "" {
+    // WSK_CONFIG_FILE environment variable overrides the default properties file path
+    // NOTE: If this variable is set to an empty string or non-existent/unreadable file
+    // - any existing $HOME/.wskprops is ignored
+    // - a default configuration is used
+    if propsFilePath, envExists = os.LookupEnv("WSK_CONFIG_FILE"); envExists {
         whisk.Debug(whisk.DbgInfo, "Using properties file '%s' from WSK_CONFIG_FILE environment variable\n", propsFilePath)
         return propsFilePath, nil
     } else {


mdeuser:
@csantanapr - good travis/pg.  Ready for review/merge.  Thanks!

csantanapr:
@mdeuser did you addressed @rabbah comments ?

csantanapr:
@rabbah please review the latest changes and either approve or request changes

mdeuser:
This PR also includes the change to have the CLI ignore a WSK_CONFIG_FILE value that's an empty string.  I'd like the ability for a given test suite to run correctly by itself with a set of test defaults, but also to run in another context with an externally provided WSK_CONFIG_FILE setting that provides context aware behavior.  Using the CLI environment variable seemed like a good way to do this.

Using an empty WSK_CONFIG_FILE value still works.  The CLI will simply ignore it.  Travis/PGs passed with this change - are the package installation needs included in those test suites?

mdeuser:
I'll add a separate WSK_CONFIG_FILE empty string test to make sure the empty string test continues to work correctly.

rabbah:
I think I did not communicate my concern clearly - let me try again. In the presence of a `~/.wsprops` file, setting `WSK_CONFIG_FILE=` forces (today) the CLI to ignore the property file and hence start with no properties.

For example:
```
> ls ~/.wskprops
/my/home/.wskprops

> WSK_CONFIG_FILE= wsk property get --all --apihost xxx
whisk auth		
whisk API host		
whisk API version	v1
whisk namespace		_
whisk CLI version	2017-06-09T21:18:47+00:00
whisk API build		Unknown
whisk API build number	Unknown
```

This feature is used by the catalogs to override the presence of a local file when installing catalogs.

mdeuser:
Ah.. I now understand.  Currently, there's no test that enforces the above behavior; thus the need for that test.  I'll also need to back out the CLI change to maintain that behavior.  Probably need a doc update to describe this as well.

mdeuser:
@rabbah - test added.  original capability restored.

tests/src/test/scala/common/Wsk.scala
```diff
@@ -869,7 +869,8 @@ class WskApi()
             { apiname map { a => Seq("--apiname", a) } getOrElse Seq() } ++
             { swagger map { s => Seq("--config-file", s) } getOrElse Seq() } ++
             { responsetype map { t => Seq("--response-type", t) } getOrElse Seq() }
-        cli(wp.overrides ++ params, expectedExitCode, showCmd = true, env=Map("WSK_CONFIG_FILE" -> cliCfgFile.getOrElse("")))
+        val envMap = { cliCfgFile map {f => Map("WSK_CONFIG_FILE" -> f) } getOrElse Map() }
```
rabbah: these changes should be backed out also (to prevent collision with local prop files for testings). the overrides are supposed to cover all the relevant variables but this is an extra precaution. Do you think we should not do this?

mdeuser: hmm.. i would still like the ability to drive the tests via an externally set WSK_CONFIG_FILE variable (i.e. no change to the test code).  with the existing, pre-pr code, this is not possible; if the WSK_CONFIG_FILE is not set in the `env` parameter (of the `cli` method), the WSK_CONFIG_FILE is always set to "".

i think another way to solve might be to change the `sys.env ++ env` to `env ++ sys.env` (wsk.scala line 1015).  this would leave the default `env` parameter WSK_CONFIG_FILE value set to "" and still have the environmental override.  what do you think?

mdeuser: after thinking more about the map ++ reversal proposal, don't it since it seems more logical for method env parameters should always be overriding the system.env

rabbah: my concern is that a local props file will override test settings - this has happened which is why this flag override exists (I can look for the original PR if it helps).

rabbah: So the desired invariant is to ignore ~/.wskprops for tests - I don't claim the env var is the best solution but it is the defensive feature in place today. We could address it with an alternative explicit flag in the CLI.

rabbah: Also it seems dangerous to me to drive these tests with an external setting in this way; because again, I can run the tests and forget that I have this value set and have funky behavior. The point of these tests to integrate against a specific deployment via wsk.properties. So what settings are you changing externally that you can't do with whisk.properties?

rabbah:
@mdeuser is this pr abandoned?

rabbah:
@mdeuser the comment and test for `WSK_CONFIG_FILE == ""` are 👍 
but this change is still in and is a breaking change:
` +        val envMap = { cliCfgFile map {f => Map("WSK_CONFIG_FILE" -> f) } getOrElse Map() } `

can we back these out?

mdeuser:
@rabbah - i missed the earlier comment.  sorry.  yeah, i no longer need this WSK_CONFIG_FILE behavior change as i was able to extend some tests with cli property changes set in the superseding test class.  so the only change in this pr that seems relevant is the updated test case and the property.go comments.  that's probably what you are suggesting.

rabbah:
indeed.

mdeuser:
pg2/2071

mdeuser:
@rabbah - travis passed.  related pg tests passed.  ready for final review/merge

PR #2360, opened 2017-06-12
Choose target invoker based on specific invoker load.

markusthoemmes:
Currently, the loadbalancer advances from one Invoker to another after a fixed amount of invocations, which isn't aware of any load in the system causing suboptimal behavior.

This only advances away from the home invoker of an action (determined by hash) if that home invoker is "heavily" loaded. We advance further if the next chosen invoker is busy and so forth. If we arrive at the home invoker again, the system is completely loaded and we force schedule to the home invoker. Step sizes are determined by prime numbers and also chosen by hashing to prevent chasing behavior.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
@@ -20,7 +20,6 @@ package whisk.core.loadBalancer
 import java.nio.charset.StandardCharsets
 
 import java.time.{ Clock, Instant }
-import java.util.concurrent.atomic.AtomicInteger
 
 import scala.collection.concurrent.TrieMap
 import scala.concurrent.Await
@@ -46,13 +45,14 @@ import whisk.common.TransactionId
 import whisk.connector.kafka.KafkaConsumerConnector
 import whisk.connector.kafka.KafkaProducerConnector
 import whisk.core.WhiskConfig
-import whisk.core.WhiskConfig.{ consulServer, kafkaHost, loadbalancerActivationCountBeforeNextInvoker }
+import whisk.core.WhiskConfig._
 import whisk.core.connector.{ ActivationMessage, CompletionMessage }
 import whisk.core.connector.MessageProducer
 import whisk.core.database.NoDocumentException
 import whisk.core.entity.{ ActivationId, CodeExec, WhiskAction, WhiskActivation }
 import whisk.core.entity.WhiskAction
 import whisk.core.entity.types.EntityStore
+import scala.annotation.tailrec
 
 trait LoadBalancer {
 
@@ -85,10 +85,6 @@ class LoadBalancerService(config: WhiskConfig, entityStore: EntityStore)(implici
     /** How many invokers are dedicated to blackbox images.  We range bound to something sensical regardless of configuration. */
     private val blackboxFraction: Double = Math.max(0.0, Math.min(1.0, config.controllerBlackboxFraction))
     logging.info(this, s"blackboxFraction = $blackboxFraction")
-
-    /** We run this often on an invoker before going onto the next. */
-    private val activationCountBeforeNextInvoker = Math.max(1, config.loadbalancerActivationCountBeforeNextInvoker)
-    logging.info(this, s"activationCountBeforeNextInvoker = $activationCountBeforeNextInvoker")
 
     override def getActiveUserActivationCounts: Map[String, Int] = activationBySubject.toMap mapValues { _.size }
 
@@ -278,18 +274,19 @@ class LoadBalancerService(config: WhiskConfig, entityStore: EntityStore)(implici
             case _              => false
         }
         val invokers = if (isBlackbox) blackboxInvokers else managedInvokers
-        val (hash, count) = hashAndCountSubjectAction(msg)
+        val hash = hashSubjectAction(msg)
 
         invokers.flatMap { invokers =>
-            val numInvokers = invokers.length
-            if (numInvokers > 0) {
-                val hashCount = math.abs(hash + count / activationCountBeforeNextInvoker)
-                val invokerIndex = hashCount % numInvokers
-                Future.successful(invokers(invokerIndex))
-            } else {
-                logging.error(this, s"all invokers down")(TransactionId.invokerHealth)
-                Future.failed(new LoadBalancerException("no invokers available"))
-            }
+            LoadBalancerService.schedule(
+                invokers,
+                activationByInvoker.mapValues(_.size),
+                config.loadbalancerInvokerBusyThreshold,
+                hash) match {
+                    case Some(invoker) => Future.successful(invoker)
+                    case None =>
+                        logging.error(this, s"all invokers down")(TransactionId.invokerHealth)
+                        Future.failed(new LoadBalancerException("no invokers available"))
+                }
         }
     }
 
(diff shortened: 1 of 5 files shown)


perryibm:
In going over this Markus, a couple of issues came up and are worth repeating here for review purposes in case we missed our intent.

1.  The computation of co-primes is dynamic because the number of invokers is changed.  Yet it is too slow to compute in the inner loop for each activation.  Memoization overcomes the overhead.
2.  The explicit introduction of `1` in the stepSizes is needed because for N = 2 (for example), we must include it for completeness.  Inclusion must be separate due to the way the sieve works.
3. The `until x` excludes x which is important if x (number of invokers) happens to be prime.

sjfink:
By my reading, I don't think the hashing based on prime numbers is good.    What we want are not a list of prime numbers, but a list of numbers that are relatively prime to the total number of invokers.

For example:  suppose we have 100 invokers.    "5" is a prime number, but it's not a good step, because 5 is not relatively prime to 100.   (stepping by 5 will not hit all 100 invokers -- only 20 of them.)

On the contrary, "21" is not a prime number, but it is relatively prime to 100.   So 21 is a good step, since stepping by 21 will hit all 100 invokers before coming home.

https://en.wikipedia.org/wiki/Coprime_integers

markusthoemmes:
@sjfink is absolutely right. I rolled back to an earlier version with coprime prime numbers.

sjfink:
After chatting with Markus, I think his original idea "all prime numbers that don't divide N" is probably good.    Even though "21" will technically hit all invokers before wrapping,  I agree with Markus that we don't want both "7" and "21" in the list of steps, since they will collide early in the sequence, which will be the common case.

Bottom line:  I'm good with the algorithm "all prime numbers which don't divide N".

perryibm:
We actually started with something like that but got to this state due to a buggy code change.   But there is an algorithmic issue that underlies this.   Do we want co-primeness wrt the number of invokers only or also wrt to the subset of possible strides?   

Concretely, suppose we have 100 invokers, do we want allow both 7 and 21 as possible steps?  Both are correct in terms of striding through all 100 invokers but there is possibly some greater chance of collisions for a different base hash.

style95:
I hope this could be addressed too.
Situation is as follow:

1. I have two running invokers.
2. User sends action invocation request.
3. At the same time, one invoker which is in charge of invoking the request is down.
4. It takes some time(5~10 secs) for controller to catch up the situation(invoker is down).
5. Accordingly, controller tries to send request to the abnormal invoker.
6. Controller can not receive proper response.
7. (At this point, I expect request will be automatically redirected to normal invoker but) Controller respond with failed response to client though there is running invoker.

This is because currently, there are kafka topics for each invokers.
And controller regularly checks the health of invokers with health topic.
So until the next check, controller will recognize the unhealthy invoker as health invoker.

Because of this, even though we deploy multiple invokers, we can not make invokers high available.


I think there are two options.

1. Make invokers to share one kafka topic.
2. Make controller(or loadbalancer) redirect the request to healthy invoker in case of failure.


**1. Make invokers to share one kafka topic.**
If we can make invokers to share one kafka topic, this situation could be handled.
Then each invokers will be grouped as one consumer group.
In case of failure of a target invoker, request will be automatically redirected to healthy invoker at kafka layer.

AFAIK, same request should be sent to same invoker to take advantage of warm container, so there are topics for each invoker.
So if we take this approach, any invokers should be able to handle any warm containers(which is not created by themselves).
Even though different invoker receives the same request, the invoker could activate warm container again.

I am not sure this could be easily achievable with current implementation.


**2. Controller(or loadbalancer) redirect the request to healthy invoker in case of failure.**
This approach is to make controller do what kafka does.
In case of failure, controller should redirect the request to healthy invoker rather than just respond with fail response.

Since many things are already considered in Kafka side, I am inclined to first option.


I want you guys to check on this issue too.

ddragosd:
@markusthoemmes did you give any though on actually saving the state where each warmed container is found in the cluster ? In this way the load balancer could simply route traffic targeting the right nodes.

ddragosd:
@style95 it would be interesting to see how other container management solutions powered by Kube or Mesos, plus a good load balancer such as NGINX, that is aware of where containers are, and can route requests directly to the warmed containers, change the picture towards what you're saying.

markusthoemmes:
@ddragosd yes, what you're saying could be the next evolution of this. You need to be wary of scale-out scenarios etc. etc. I think that will need a broader discussion probably more suited for the dev list?

@style95 Very valid thoughts, it would be great to have this discussion on the dev list as well.

michaelmarth:
+1 to @ddragosd comment. Would be good to discuss how this approach expands into resource management solutions that:
a) have overview over *all* the available resources in the cluster and
b) maybe even the capability to dynamically expand those clusters if needed.

Good starting point, though - I like the stateless approach you have taken, @markusthoemmes

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -299,26 +329,32 @@ class LoadBalancerService(config: WhiskConfig, entityStore: EntityStore)(implici
...
-                activationCountMap.put(key, new AtomicInteger(0))
-                0
-            }
-        }
-        return (hash, count)
+        subject.hashCode() ^ path.hashCode()
+    }
+
+    /** Memoizes the result of `f` for later use. */
+    private def memoize[I, O](f: I => O): I => O = new scala.collection.mutable.HashMap[I, O]() {
```
rabbah: I would also pre-compute these lists for the number of invokers in a deployment rather than on demand where you might up doing computing the same list multiple times in `chooseInvoker` for overlapping first selections. It seems we don't have that value in the whisk config but perhaps we should to facilitate this. You could also fold this into the invoker health protocol so as new invokers are online the lists are computed.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -277,14 +277,44 @@ class LoadBalancerService(config: WhiskConfig, entityStore: EntityStore)(implici
...
+                /*
+                 * Scans through all invokers and searches for an invoker, which's queue is below
+                 * the defined threshold. Iff no "underloaded" container was found it will default
+                 * to the least loaded invoker in the list.
+                 *
+                 * @param targetInvoker index of the invoker-candidate.
+                 * @return name of the target invoker
+                 */
+                @tailrec
+                def search(targetInvoker: Int): String = {
```
rabbah: I would also factor this method out to the singleton as well and avoid closing on the maps or the home invoker, to make it more easily testable.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -277,14 +277,44 @@ class LoadBalancerService(config: WhiskConfig, entityStore: EntityStore)(implici
...
+                    val invokerName = invokers(targetInvoker)
+
+                    if (activationByInvoker.get(invokerName).map(_.size).getOrElse(0) < nextInvokerThreshold) {
```
rabbah: how about some comments

```
+                    // send the request to the target invoker if it has capacity...
                     if (activationByInvoker.get(invokerName).map(_.size).getOrElse(0) < nextInvokerThreshold) {
                         invokerName
                     } else {
+                        // ... otherwise look for a less loaded invoker by stepping through a pre computed
+                        // list of invokers; there are two possible outcomes:
+                        // 1. the search lands on a new invoker that has capacity, choose it
+                        // 2. walked through the entire list and found no better invoker than the "home invoker", choose the least loaded invoker
```

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -277,14 +277,44 @@ class LoadBalancerService(config: WhiskConfig, entityStore: EntityStore)(implici
...
         invokers.flatMap { invokers =>
             val numInvokers = invokers.length
             if (numInvokers > 0) {
-                val hashCount = math.abs(hash + count / activationCountBeforeNextInvoker)
-                val invokerIndex = hashCount % numInvokers
-                Future.successful(invokers(invokerIndex))
+                val nextInvokerThreshold = config.loadbalancerInvokerBusyThreshold
+                val homeInvoker = hash % numInvokers
+
+                val stepSizes = coprimePrimeNumbersUntil(numInvokers)
```
rabbah: should this list of invokers be filtered for healthy invokers otherwise might you assign an invoker that is not healthy?

rabbah:
@michaelmarth re. " b) maybe even the capability to dynamically expand those clusters if needed." this is already supported in that if new "invoker" clusters are brought online, they self-register with the load balancer which then can use those resources in the scheduling. Did you mean for the loadbalancer to initiate a request for new capacity? So we don't hijack this pr with this discussion do you want to open a separate issue to discuss this?

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
markusthoemmes: Agree on the first, disagree on the latter: Don't think we need to be so cautious with this particular function. At steady state the results shouldn't change much. Folding it into the invokerHealth protocol won't guarantee no overlapping recomputation either. We could mitigate this by storing a Promise's future in the map and return that to each caller so we would fold all computations into one.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
markusthoemmes: It already is, `numInvokers` comes from either `managedInvokers` or `blackboxInvokers` which is derived from `availableInvokers`.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
markusthoemmes: I'd rather get the whole `chooseInvoker` out then. There's more benefit in testing the whole decision rather than testing only this little part of it.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
rabbah: mum invokers is the cardinality only nitnthr index of available/healthy invokers in the system, is it not?

rabbah: To put it in other words it's not clear here if the search is over a list of all invokers that are healthy or over a sequence 1..n.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -277,14 +277,44 @@ class LoadBalancerService(config: WhiskConfig, entityStore: EntityStore)(implici
...
+                 * Scans through all invokers and searches for an invoker, which's queue is below
+                 * the defined threshold. Iff no "underloaded" container was found it will default
+                 * to the least loaded invoker in the list.
+                 *
+                 * @param targetInvoker index of the invoker-candidate.
+                 * @return name of the target invoker
+                 */
+                @tailrec
+                def search(targetInvoker: Int): String = {
+                    val invokerName = invokers(targetInvoker)
```
rabbah: Here, the index is mapped to an actual invoker from the set of live ones.

rabbah: A comment will suffice when is the hash is computed.

rabbah:
@style95 while it is true that a request may go to an unhealthy invoker topic, since messages in queue are persisted, a new invoker to replace the unhealthy one will pick up those messages and process them eventually. This can be achieved either by having a new invoker subscribe to the topic as needed. We had considered consumer groups in the past and more recently rethinking an approach around work stealing.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -332,29 +299,84 @@ class LoadBalancerService(config: WhiskConfig, entityStore: EntityStore)(implici
...
             })
     }
-}
 
-object LoadBalancerService {
-    def requiredProperties = kafkaHost ++ consulServer ++
-        Map(loadbalancerActivationCountBeforeNextInvoker -> null,
-            loadbalancerInvokerBusyThreshold -> null)
+    /**
+     * Scans through all invokers and searches for an invoker, that has a  queue is
```
rabbah: that has a  queue is below -> that has a queue length below

tests/src/test/scala/whisk/core/loadBalancer/test/LoadBalancerServiceObjectTests.scala
```diff
@@ -0,0 +1,114 @@
...
+        LoadBalancerService.schedule(invs, Map("invoker0" -> 1, "invoker1" -> 1), 1, hash) shouldBe Some(hashInto(invs, hash + step + step))
+    }
+
+    it should "choose the least loaded invoker if all invokers are overloaded to begin with" in {
+        val invokerCount = 3
+        val invs = invokers(invokerCount)
+        val hash = 1
+
+        val targetInvoker = hashInto(invs, hash)
+        val step = hashInto(LoadBalancerService.coprimePrimeNumbersUntil(invokerCount), hash)
```
rabbah: targetInvoker and step unused.

style95:
@rabbah 
oh ok.
I just wanted to bring such a case into the scene while we have discussion.

I think though action would be invoked eventually, problem still exists.

How about blocking call and web action(with api gateway)?
Once client receives failure response, client will no more expect the proper response.
(even though the action would be invoked)

I have not much idea on the decision history, so my opinion could be naive.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
rabbah: Did you opt to defer the priming of the lists at load time?

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -312,14 +312,13 @@ object LoadBalancerService {
...
-                    primes :+ cur
-                } else primes
-            })
+    /** Eclidean algorithm to determine the greatest-common-divisor */
+    @tailrec
+    def gcd(a: Int, b: Int): Int = if (b == 0) a else gcd(b, a % b)
+
+    /** All coprime numbers until x including 1. Result is memoized. */
+    val coprimeNumbersUntil: Int => IndexedSeq[Int] = LoadBalancerService.memoize {
+        case x => (1 to x).filter(i => gcd(i, x) == 1)
```
rabbah: Note that the previous algorithm computed a set of coprime numbers which are also pairwise coprime; this algorithm does not. I _thought_ that was a desired property because the alternative can lead to overlaps in the search sequence.

markusthoemmes: It didn't actually compute coprime numbers but numbers that don't divide the upperbound, which is wrong for this scenario.

rabbah: Agree there was a bug. But I'm pointing out that we may want sets that are pairwise coprime as well.

markusthoemmes:
PG1 1685 🔵

PR #2291, opened 2017-05-25
Add subject with a namespace.

rabbah:
Closes #2140.

This reverts a change from this commit https://github.com/apache/incubator-openwhisk/commit/7b5a64481aedfd060f3ccaab72a2b6197de47df2#commitcomment-21755818.

PR #2361, opened 2017-06-12
Replace use of wskadmin inside api gw tests

mdeuser:
wskadmin needs direct access to the openwhisk db to acquire the namespace associated with a particular auth key.  this direct access may not be possible in all test environments. Instead, use the `wsk` cli to access the namespace from the cli configured auth key.

### DIFF
--- a/tests/src/test/scala/whisk/core/cli/test/ApiGwTests.scala
+++ b/tests/src/test/scala/whisk/core/cli/test/ApiGwTests.scala
@@ -28,7 +28,6 @@ import common.TestUtils._
 import common.TestUtils
 import common.WhiskProperties
 import common.Wsk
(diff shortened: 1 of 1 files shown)


rabbah:
I have a related change here: https://github.com/apache/incubator-openwhisk/pull/2291/files#diff-16f5f41b10d362adbb1bc2bf9d2ef157R691 as part of #2291.

mdeuser:
@rabbah - I like it.  Depending on the PR timing, perhaps just that Wsk.scala namespace.whois change be merged in a separate PR so I can leverage that in this test case..?

mdeuser:
pg3/742

rabbah:
This is addressed by recent commits I believe.

Issue #2125, opened 2017-04-10
OpenWhisk hello.bat in tools/vagrant failing on Windows 7


amilenovic:
[hello.cmd.txt](https://github.com/openwhisk/openwhisk/files/912786/hello.cmd.txt)

I tried to start up OpenWhisk on Windows 7 laptop and it failed. 

I tested the same procedure on Fedora 25 and it worked.

## Environment details:

* Windows 7   
* Vagrant 1.9.3
* VirtualBox 5.1.8
* Git windows 2.10.2

## Steps to reproduce the issue:

1.  git clone https://github.com/openwhisk/openwhisk
2.  cd tools/vagrant 
3.  hello.cmd 

## Provide the expected results and outputs:

```
==> default: fatal: [172.17.0.1]: FAILED! => {"changed": true, "cmd": "docker pu
ll couchdb:1.6", "delta": "0:00:03.030560", "end": "2017-04-10 13:37:46.112529",
 "failed": true, "rc": 1, "start": "2017-04-10 13:37:43.081969", "stderr": "erro
r pulling image configuration: Get https://dseasb33srnrn.cloudfront.net/registry
-v2/docker/registry/v2/blobs/sha256/b1/b15a0ea0ffc93107a64358eef95c707f5eb0ed738
1445733977f8e375059d567/data?Expires=1491832665&Signature=cOMpRQ9nSDOfybL6veeqEk
ln6bCmDHx8O1eNtT~szH2zSLVfBPebzUkN99UeHWJoACJqP-6T6K97PezMBlPq99y82wspbu9FbqOGNT
7-vbX4fwrYEJBIoGF6lhtRavsja7SQDdV~bAU62BUHGCuL43kFMCplc-IqB0IREQ4EJGk_&Key-Pair-
Id=APKAJECH5M7VWIS5YZ6Q: read tcp 10.0.2.15:47721->52.222.149.243:443: read: con
nection reset by peer", "stdout": "1.6: Pulling from library/couchdb\n6d827a3ef3
58: Pulling fs layer\n19f81ede1f06: Pulling fs layer\nc02a3a6658a4: Pulling fs l
ayer\n948821c9ca7a: Pulling fs layer\n0db1fcee1af2: Pulling fs layer\n6aea6c280c
e0: Pulling fs layer\nb5f957c0d523: Pulling fs layer\n6aea6c280ce0: Waiting\n948
[...]

Issue #1568, opened 2016-11-30
Setup on Ubuntu OS produces warning / errors 


ibmamnt:
* Ubuntu 14.04LTS via Vagrant(1.7.4) /VitualBox(v5.0.2r102096)
* HostOS : Windows 7

## Steps to reproduce the issue:

Setup OS:
1.  mkdir Whisk
2.  cd Whisk
3.  vagrant init ubuntu/trusty
4. vagrant up
5. vagrant ssh
6. (ubutsu)# sudo apt-get update

Setup OpenWhisk
Based on the instruction.
https://github.com/openwhisk/openwhisk/blob/master/tools/ubuntu-setup/README.md

## Provide the actual results and outputs:

```
W: GPG error: https://apt.dockerproject.org ubuntu-trusty InRelease: The following signatures couldn't be verified becau
Downloading/unpacking markupsafe
  Downloading MarkupSafe-0.23.tar.gz
  Running setup.py (path:/tmp/pip_build_root/markupsafe/setup.py) egg_info for package markupsafe

Installing collected packages: markupsafe
  Running setup.py install for markupsafe

    building 'markupsafe._speedups' extension
    x86_64-linux-gnu-gcc -pthread -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fPIC -I/usr/in
    markupsafe/_speedups.c:12:20: fatal error: Python.h: No such file or directory
     #include <Python.h>
                        ^
```

## Additional information you deem important:
* This error seems to happen on Ubuntu image on OpenStack.
* Run the procedure twice.

Suspecting the reading python environment is in sufficient. Also there is difference between Vagrant version and native ubuntu version.

csantanapr:
User @glikson also reporting this problem in vagrant use case.

I think docker-py or ansible depends on `python-dev` being installed

csantanapr:
@glikson you think we also required `libffi-dev` ?

glikson:
If I am not mistaken, libffi-dev is needed to resolve an issue that pops up later, with cryptography package not finding ffi.h

csantanapr:
Yes so validated that both are required thanks

shekhargulati:
[...]

PR #2362, opened 2017-06-12
have ubuntu setup scripts error on first error

csantanapr:
running the scripts sequentially didn't add more than a minute
it will make vagrant fail on first error
printing commands when executing helps for debugging
adding the installation of libssl-dev to fix problem installing python-dev


csantanapr:
Fixes #1568 
Fixes #2125

PR #2369, opened 2017-06-14
document how to add swift dependencies

csantanapr:


csantanapr:
cc @jthomas @akrabat 
please take a look if you have chance

docs/actions.md
```diff
@@ -636,11 +636,13 @@ When you create an OpenWhisk Swift action with a Swift source file, it has to be
...
-```
-   This puts you in a bash shell within the Docker container. Execute the following commands within it:
+  ```
+  docker run --rm -it -v "$(pwd):/owexec" openwhisk/swift3action bash
+  ```
+
+  This puts you in a bash shell within the Docker container. 
 
+  Execute the following commands within it:
 - Install zip for convenience, to package the binary
```
akrabat: While not directly related to this PR, zip is installed by default in the action, so we can remove this bullet point.

docs/actions.md
```diff
@@ -659,6 +661,42 @@ docker run --rm -it -v "$(pwd):/owexec" openwhisk/swift3action bash
...
+  let package = Package(
+    name: "Action",
+        dependencies: [
+            .Package(url: "https://github.com/IBM-Swift/Kitura-net.git", "1.0.1"),
+            .Package(url: "https://github.com/IBM-Swift/SwiftyJSON.git", "14.2.0"),
+            .Package(url: "https://github.com/IBM-Swift/swift-watson-sdk.git", "0.4.1"),
+            .Package(url: "https://github.com/apple/example-package-deckofplayingcards.git", majorVersion: 3)
+        ]
+  )
+  ```
```
akrabat: Indentation should be so that "name" and "dependencies" are the same level.

docs/actions.md
```diff
@@ -659,6 +661,42 @@ docker run --rm -it -v "$(pwd):/owexec" openwhisk/swift3action bash
...
+
+- Copy Package.swift
+  ```
+  cp -f /owexec/Package.swift /swift3Action/spm-build/Package.swift
+  ```
+
+- Remove .build directory
+  ```
+  rm -rf /swift3Action/spm-build/.build/
+  ```
```
akrabat: You shouldn't need to remove the build directory, and if you haven't created a `Package.swift`, you certainly don't want to do.

docs/actions.md
```diff
@@ -659,6 +661,42 @@ docker run --rm -it -v "$(pwd):/owexec" openwhisk/swift3action bash
...
+
+- Generate a new swiftbuildandlink.sh
+  ```
+  python /swift3Action/buildandrecord.py
+  ```
```
akrabat: This line makes the subsequent `swiftbuildandlink.sh` step fail with the error message: `/swift3Action/spm-build/swiftbuildandlink.sh: line 24: syntax error: unexpected end of file`

docs/actions.md
```diff
@@ -659,6 +661,42 @@ docker run --rm -it -v "$(pwd):/owexec" openwhisk/swift3action bash
   echo '_run_main(mainFunction:main)' >> /swift3Action/spm-build/main.swift
   ```
 
+  Copy any additional source files to `/swift3Action/spm-build/`
+
+- Create Package.swift
```
akrabat: I would start this bullet point with the word "Optionally" as it slows the build down so much!

docs/actions.md
akrabat: This command also takes an _age_! On my computer, `time python /swift3Action/buildandrecord.py` takes 1m16s!

(For reference, the entire compilation and creation of the zip file takes 5s when I don't remove `.build` or run `buildandrecord.py`.)

What does `buildandrecord.py` followed by `swiftbuildandlink.sh` do that's different from just running `swift build -c release`?

akrabat:
Same problem happens when providing my own Package.swift: https://gist.github.com/akrabat/ede6cd366ab3bf535680c5c5f71c5d78

docs/actions.md
jthomas: The user can just do `swift build -c release` as @akrabat suggests. `buildandrecord.py` and `swiftbuildandlink` are not necessary here.

docs/actions.md
csantanapr: yeah I agree is optional

docs/actions.md
csantanapr: would take a look

docs/actions.md
csantanapr: removing the .build was the only way to get `buildandrecord.py` to run correctly and capture the info to generate the `swiftbuildandlink.sh`

docs/actions.md
csantanapr: Not sure, maybe is legacy of swift2 ?
If we use `swift` instead of this python to gen a .sh would be better
@paulcastro @bjustin-ibm Do you have any idea of need of `buildandrecord.py`
instead of using swift build directly

csantanapr:
@akrabat I will try again I did a lot iterations, and capture some of the steps I got working.

docs/actions.md
bjustin-ibm: I only vaguely recall that this may have been done for performance reasons (https://github.com/apache/incubator-openwhisk/issues/1371). @paulcastro would know for sure.

paulcastro: You are correct!

`buildandrecord.py` generates the `swiftbuildandlink.sh` in order to skip unnecessary steps when a swift action is initialized. This shaved off a few seconds from swift action cold starts.  

It's definitely not necessary for pre-compilation.

csantanapr:
@paulcastro I updated to use swift311 and swift build -c

.gitignore
```diff
@@ -41,6 +41,7 @@ node_modules
 
 # IntelliJ
 .idea
+*.class
```
akrabat: Arguably, this change has nothing to do with this PR, but I don't suppose it matters too much.

docs/actions.md
```diff
@@ -607,7 +607,7 @@ You can create a OpenWhisk action called `helloSwift` from this function as
 follows:
 
 ```
-wsk action create helloSwift hello.swift
+wsk action create helloSwift hello.swift --kind swift:3.1.1
```
akrabat: I think that the first time we show how to create an action for a given language, we should use the default and not use `--kind`. (i.e. this should be the same as how JS and Python are shown.)

docs/actions.md
```diff
@@ -636,49 +636,69 @@ When you create an OpenWhisk Swift action with a Swift source file, it has to be
...
   cat /swift3Action/epilogue.swift >> /swift3Action/spm-build/main.swift
   ```
-
   ```
   echo '_run_main(mainFunction:main)' >> /swift3Action/spm-build/main.swift
   ```
+  Copy any additional source files to `/swift3Action/spm-build/`
 
-- Build and link
+- (Optional) Create Package.swift
```
akrabat: backticks around `Package.swift` would make it clearer that it's a file.

docs/actions.md
```diff
@@ -636,49 +636,69 @@ When you create an OpenWhisk Swift action with a Swift source file, it has to be
...
+  If you need to add dependencies create a Package.swift file like the following:
```
akrabat: These two lines render as a single line which looks a little odd:

> * (Optional) Create Package.swift If you need to add dependencies create a Package.swift file like the following:

Maybe add an additional blank like to force it to separate sentence?

docs/actions.md
```diff
@@ -636,49 +636,69 @@ When you create an OpenWhisk Swift action with a Swift source file, it has to be
...
+            .Package(url: "https://github.com/apple/example-package-deckofplayingcards.git", majorVersion: 3),
+            .Package(url: "https://github.com/IBM-Swift/CCurl.git", "0.2.3"),
+            .Package(url: "https://github.com/IBM-Swift/Kitura-net.git", "1.7.10"),
+            .Package(url: "https://github.com/IBM-Swift/SwiftyJSON.git", "15.0.1"),
+            .Package(url: "https://github.com/watson-developer-cloud/swift-sdk.git", "0.16.0")
+        ]
+  )
+  ```
+  As you can see this example adds `swift-watson-sdk` and `example-package-deckofplayingcards` dependencies.
+  Notice that `CCurl`, `Kitura-net` and `SwiftyJSON.git` are the minimum required dependencies to include.
```
akrabat: Technically, I'm pretty sure that CCurl and Kitura-net aren't minimum required dependencies, however, they are default and so it's a good idea to keep them in your `Package.swift` as it would be less confusing.

How about rewording to: 

    Notice that `CCurl`, `Kitura-net` and `SwiftyJSON` are provided in the standard Swift action
    and so you should include them in your own `Package.swift`

? 

At a minimum, remove the `.git` from `SwiftyJSON.git` as it's not included for `CCurl` or `Kitura-net`.

.gitignore
csantanapr: 👍  handling in another PR #2448

docs/actions.md
csantanapr: Hum I'm not a fan that we have :default and would prefer to be clear and explicit of the version being.
I think I agree to leave it out to have less noise for the folks getting started, and we can have separate discussion across all languages and the use of default in the docs.

docs/actions.md
csantanapr: I will make it one sentence, to avoid problems with bullets.

docs/actions.md
csantanapr: If they follow the instructions as is it would be required.
The resulting `main.swift` file will have depend on `WhiskJsonUtils.jsonDataToDictionary`
And there are two swift files `_Whisk.swift` and `_WhiskJSONUtils.swift` in the directory that get compiled that depend on those two libraries.

For CCurl, the problem is that we currently need to pinned to that version, so putting explicit version before Kitura-net works around the problem we have.

I will use your suggested text.

```
root@e37cec41703a:/# ls -l /swift3Action/
-rw-r--r-- 1 root root 10842 Jun 29 03:48 _Whisk.swift
-rw-r--r-- 1 root root  5313 Jun 29 03:48 _WhiskJSONUtils.swift
-rw-r--r-- 1 root root     0 Jun 29 03:49 main.swift
```

```
root@9d6c5f81bf9f:/# cat /swift3Action/spm-build/main.swift
#if os(Linux)
import Glibc
srandom(UInt32(clock()))
#endif

import DeckOfPlayingCards




    let numberOfCards = 10

var deck = Deck.standard52CardDeck()
deck.shuffle()

    if let name = args["name"] as? String {
        return [ "greeting" : "Hello \(name)!" ]
    } else {
        return [ "greeting" : "Hello stranger!" ]
    }
}

import Foundation

#if os(Linux)
    import Glibc
#endif

func _whisk_json2dict(txt: String) -> [String:Any]? {
    if let data = txt.data(using: String.Encoding.utf8, allowLossyConversion: true) {
        do {
            return WhiskJsonUtils.jsonDataToDictionary(jsonData: data)
        } catch {
            return nil
        }
    }
    return nil
}


func _run_main(mainFunction: ([String: Any]) -> [String: Any]) -> Void {
    let env = ProcessInfo.processInfo.environment
    let inputStr: String = env["WHISK_INPUT"] ?? "{}"

    if let parsed = _whisk_json2dict(txt: inputStr) {
        let result = mainFunction(parsed)

        if result is [String:Any] {
            do {
                if let respString = WhiskJsonUtils.dictionaryToJsonString(jsonDict: result) {
                    print("\(respString)")
                } else {
                    print("Error converting \(result) to JSON string")
                    #if os(Linux)
                        fputs("Error converting \(result) to JSON string", stderr)
                    #endif
                }
            } catch {
                print("Error serializing response \(error)")
                #if os(Linux)
                    fputs("Error serializing response \(error)", stderr)
                #endif
            }
        } else {
            print("Cannot serialize response: \(result)")
            #if os(Linux)
                fputs("Cannot serialize response: \(result)", stderr)
            #endif
        }
    } else {
        print("Error: couldn't parse JSON input.")
        #if os(Linux)
            fputs("Error: couldn't parse JSON input.", stderr)
        #endif
    }
}
_run_main(mainFunction:main)
```

csantanapr:
@bjustin-ibm I think this is ready now take a look

bjustin-ibm:
@akrabat thanks a bunch for your help on this one!

PR #2371, opened 2017-06-14
Revert "Add the support to install the binaries from the new repo"

csantanapr:
Reverts apache/incubator-openwhisk#2345


csantanapr:
Some context on this revert here:
https://github.com/apache/incubator-openwhisk/pull/2345#issuecomment-308410151

PR #2379, opened 2017-06-14

houshengbo:
This PR adds the functionalities to either install openwhisk
binaries from the releases page of openwhisk cli or copy the
binaries locally.

The key installation_mode of openwhisk_cli dict can be set to
remote or local. The default one is local.

Closes-Bug: #2101


ansible/group_vars/all
```diff
@@ -228,3 +229,45 @@ catalog_repos:
...
     version: "HEAD"
     repo_update: "no"
+
+openwhisk_cli_tag: "{{ lookup('ini', 'git_tag section=openwhisk-cli file={{ openwhisk_home }}/ansible/files/package-versions.ini') }}"
+
+# The openwhisk_cli is used to determine how to install the OpenWhisk CLI. The
+# installation_mode can be specified into two modes: remote and local.
+# The mode remote means to download the available binaries from the releases page
+# of the official openwhisk cli repository. The mode local means to get the binaries
+# from a local directory. The default value for openwhisk is local.
```
rabbah: local mode is more than the location of the binaries - it also means the location from which to build the binaries.

ansible/group_vars/all
```diff
@@ -228,3 +229,45 @@ catalog_repos:
...
+#
+# The name specifies the package name of the binaries in remote mode.
+#
+# The dest_name specifies the package name of the binaries in Nginx in remote mode.
```
rabbah: which of these are required for remote and which for local.
i suggest grouping the parameters accordingly.

houshengbo: Done.

ansible/group_vars/all
houshengbo: Done.

csantanapr:
PG4 456 🔵

PR #2383, opened 2017-06-14
docker pull invoker images runtimes from manifest

### LINKED ISSUE #2382: modify ansible to get list of docker images for invoker from manifest
This will avoid some hardcoded docker pull images we have in invoker role.
This allow to add images to manifest and pull additional images for a given kind

cc @rabbah

csantanapr:
Closes #2382

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -42,17 +42,21 @@ runtimesManifest:
       default: true
       image:
         name: "nodejs6action"
+      deprecated: false
     python:
     - kind: "python"
       image:
         name: "python2action"
+      deprecated: false
     - kind: "python:2"
       default: true
       image:
         name: "python2action"
+      deprecated: false
     - kind: "python:3"
       image:
         name: "python3action"
+      deprecated: false
     swift:
     - kind: "swift"
       image:
@@ -62,19 +66,22 @@ runtimesManifest:
       default: true
       image:
         name: "swift3action"
+      deprecated: false
     - kind: "swift:3.1.1"
       image:
         name: "action-swift-v3.1.1"
+      deprecated: false
     java:
     - kind: "java"
+      default: true
+      image:
+        name: "java8action"
+      deprecated: false
       attached:
         attachmentName: "jarfile"
         attachmentType: "application/java-archive"
       sentinelledLogs: false
       requireMain: true
-      image:
-        name: "java8action"
-      default: true
   blackboxes:
     - name: "dockerskeleton"
 
--- a/ansible/roles/invoker/tasks/deploy.yml
+++ b/ansible/roles/invoker/tasks/deploy.yml
@@ -3,17 +3,23 @@
 
 - include: docker_login.yml
 
-- name: "pull action images with tag {{ docker_image_tag }}"
-  shell: "docker pull {{ docker_registry }}{{ item }}:{{ docker_image_tag }}"
+- name: "pull invoker image with tag {{docker_image_tag}}"
+  shell: "docker pull {{docker_registry}}{{item}}:{{docker_image_tag}}"
   with_items:
     - '{{ docker_image_prefix }}/invoker'
-    - '{{ docker_image_prefix }}/dockerskeleton'
-    - '{{ docker_image_prefix }}/nodejs6action'
-    - '{{ docker_image_prefix }}/python2action'
-    - '{{ docker_image_prefix }}/python3action'
-    - '{{ docker_image_prefix }}/swift3action'
-    - '{{ docker_image_prefix }}/action-swift-v3.1.1'
-    - '{{ docker_image_prefix }}/java8action'
+  when: docker_registry != ""
+  retries: 3
+
+- name: "pull runtime action images with tag {{docker_image_tag}}"
+  shell: "docker pull {{docker_registry}}{{docker_image_prefix}}/{{item}}:{{docker_image_tag}}"
+  with_items: "{{ runtimesManifest.runtimes.values() | sum(start=[]) | selectattr('deprecated', 'equalto',false)  | map(attribute='image.name') | list | unique }}"
+  when: docker_registry != ""
+  retries: 3
+
+- name: "pull blackboxes action images with tag {{docker_image_tag}}"
+  shell: "docker pull {{docker_registry}}{{docker_image_prefix}}/{{item.name}}:{{docker_image_tag}}"
+  with_items:
+    - "{{ runtimesManifest.blackboxes }}"
   when: docker_registry != ""
   retries: 3
 
--- a/tests/src/test/scala/whisk/core/entity/test/ExecHelpers.scala
+++ b/tests/src/test/scala/whisk/core/entity/test/ExecHelpers.scala
@@ -48,7 +48,7 @@ trait ExecHelpers
     }
 
     protected def js6(code: String, main: Option[String] = None): Exec = {
-        CodeExecAsString(RuntimeManifest(NODEJS6, imagename(NODEJS6), default = Some(true)), trim(code), main.map(_.trim))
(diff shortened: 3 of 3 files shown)


ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -3,16 +3,40 @@
...
-    - '{{ docker_image_prefix }}/python2action'
-    - '{{ docker_image_prefix }}/python3action'
-    - '{{ docker_image_prefix }}/swift3action'
-    - '{{ docker_image_prefix }}/java8action'
+  when: docker_registry != ""
+
+- name: "pull nodejs action images with tag {{docker_image_tag}}"
+  shell: "docker pull {{docker_registry}}{{item.image.name}}:{{docker_image_tag}}"
+  with_items:
+    - "{{ runtimesManifest.runtimes.nodejs }}"
```
csantanapr: I worked with @dubeejw to see if there was a way to not duplicate here and be able to itirate over the dict, and then iterate over the array. we tried different combinations and were not successful.

If there is a ansible ninja that would provide how to achieve this, from the docs it gives the impression is doable but we didn't found a good combination.
http://docs.ansible.com/ansible/playbooks_loops.html

csantanapr:
PG1 1647

ansible/group_vars/all
```diff
@@ -34,10 +34,6 @@ runtimesManifest:
   defaultImageTag: "latest"
   runtimes:
     nodejs:
-    - kind: "nodejs"
```
csantanapr: @rabbah There are errors in Manifest tests cases, when I remove this kinds, it means I need to adjust the test cases?
```
 Exec should initialize exec manifest FAILED
    java.util.NoSuchElementException: None.get
```

csantanapr: Hum I'm going to leave manifest alone, and maybe try skipping the docker pull if kind is deprecated

csantanapr:
PG1 1666 passed.

ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -3,16 +3,40 @@
...
+  shell: "docker pull {{docker_registry}}{{docker_image_prefix}}/{{item.image.name}}:{{docker_image_tag}}"
+  with_items:
+    - "{{ runtimesManifest.runtimes.swift }}"
+  when: (docker_registry != "") and ("deprecated" not in item)
+
+- name: "pull java action images with tag {{docker_image_tag}}"
+  shell: "docker pull {{docker_registry}}{{docker_image_prefix}}/{{item.image.name}}:{{docker_image_tag}}"
+  with_items:
+    - "{{ runtimesManifest.runtimes.java }}"
+  when: (docker_registry != "") and ("deprecated" not in item)
```
markusthoemmes: All of the runtimes can be fold into:

```
- name: "pull java action images with tag {{docker_image_tag}}"
  shell: "docker pull {{docker_registry}}{{docker_image_prefix}}/{{item}}:{{docker_image_tag}}"
  with_items: "{{ runtimesManifest | json_query('runtimes.*[?!deprecated].image.name') }}"
  when: (docker_registry != "")
```

csantanapr: This is great 🌮  @markusthoemmes I will give it a try

csantanapr: @markusthoemmes in order to use json_query we need to require a new library to be install first
```
"You need to install \"jmespath\" prior to running json_query filter"}
```
what do you think do we require this new dependency or use what I have now?

csantanapr: my preference would be less dependencies

markusthoemmes: I'd rather not define all the runtimes here manually, that's quite a tedious process.

Try the following:

```
- name: "pull java action images with tag {{docker_image_tag}}"
  shell: "docker pull {{docker_registry}}{{docker_image_prefix}}/{{item}}:{{docker_image_tag}}"
  with_items: "{{ runtimesManifest.runtimes.values() | sum(start=[]) | map(attribute='image.name') | list }}"
  when: (docker_registry != "")
```

csantanapr: yes I agree I would prefer to not hard coded the runtime languages, I would give this try thanks again 👍

csantanapr: @markusthoemmes 
I was able to dedupe using `unique`
```
with_items: "{{ runtimesManifest.runtimes.values() | sum(start=[]) | map(attribute='image.name') | list | unique }}"
```
But I still have a problem with deprecated, I need to reject only if deprecated is present and true.
I try the filters `reject` and `rejectattr` and was note able to get it to work like
```
with_items: "{{ runtimesManifest.runtimes.values() | sum(start=[]) | rejectattr('deprecated')  | map(attribute='image.name') | list | unique }}"
```
But I have problems 
```
"'dict object' has no attribute 'deprecated'"}
```

markusthoemmes: As it stands right now, you could do `selectattr('deprecated', 'undefined')` because we haven't defined `deprecated: False` anywhere.

I'd prefer we just define the key everywhere and set it to False for completeness. I think that's less confusing for everyone WDYT?

csantanapr:
@markusthoemmes I updated the manifest to be explicit to `deprecated: false` and using this as selectattr.
please review

csantanapr:
Having trouble with Travis with my changes.
Can @rabbah or @markusthoemmes give me a clue why a simple the change to manifest will cause this errors?
```
whisk.core.entity.test.DatastoreTests
----------------
Datastore should CRD action js
Datastore should update action with a revision
Datastore should fail with document conflict when trying to write the same action twice without a revision
Datastore should fail with document does not exist when trying to delete the same action twice

whisk.core.controller.test.ActionsApiTests
----------------
Actions API should get action by name in default namespace
Actions API should get action by name in explicit namespace
Actions API should delete action by name
Actions API should put should accept request with missing optional properties
Actions API should put should accept request with parameters property
Actions API should put should accept request with limits property
Actions API should put and then get action from cache
Actions API should update action with a put
Actions API should update action parameters with a put

whisk.core.entity.test.ViewTests
----------------
Datastore View should query whisk and retrieve full documents

whisk.core.controller.test.PackageActionsApiTests
----------------
Package Actions API should put action in package
Package Actions API should delete action in package
Package Actions API should get action in package
Package Actions API should get action in package binding with public package
Package Actions API should get action in package binding with public package with overriding parameters

whisk.core.entity.test.SchemaTests
----------------
Exec should properly deserialize and reserialize JSON
```

rabbah:
look in exec helpers

 ``` protected def js(code: String, main: Option[String] = None): Exec = {
        CodeExecAsString(RuntimeManifest(NODEJS, imagename(NODEJS), deprecated = Some(true)), trim(code), main.map(_.trim))
    }

    protected def js6(code: String, main: Option[String] = None): Exec = {
        CodeExecAsString(RuntimeManifest(NODEJS6, imagename(NODEJS6), default = Some(true)), trim(code), main.map(_.trim))
    }

    protected def jsDefault(code: String, main: Option[String] = None) = {
        js6(code, main)
    }
```

the value of deprecated in the test is None, not Some(false) which it needs to be for the equality to hold; that's my quick guess based on the log.

**Comment by csantanapr** [comment]
So I need to modify the tests, to now compare against false instead of None?

**Comment by rabbah** [comment]
i think you can modify the ExecHelpers to use Some(false) for the deprecated value.

**Comment by csantanapr** [comment]
@rabbah OK

So we don't have to change `deprecated` to be false in `RuntimeManifest` then?

```
    protected[core] case class RuntimeManifest(
        deprecated: Option[Boolean] = None,
```

rabbah:
I would leave that alone. Otherwise if you're going to always enforce it, I'd say change it from option type to boolean and then you're making more changes.

csantanapr:
PG 2 1738 passed

PR #2384, opened 2017-06-15
Send active ack on failed activations

### LINKED ISSUE #1906: sequences silently fail when internal invocations exceed 1MB threshold
in a sequence ABC, if B returns a payload > 1MB, the sequence fails (as it should), but with no indication as to why

### LINKED ISSUE #2118: controller should slow poll the DB for activations
Today there was an issue where an active ack message could not be produced because the result of the activation was too large. Even though the action finished quickly, the invoker still waited the (newly increased) 60s active ack timeout period. In the meantime, the controller converted this activation to non-blocking for having exceeded the 60 second blocking threshold (because polling throws `BlockingInvokeTimeout`). However, even though the action completed rather quickly, and the activation was written to the DB even before the active ack time expired, the user still had to wait 60 seconds just to find out that their activation was converted to non-blocking (when really it was already completed).

One solution is to have the controller simultaneously poll the DB for the activation, even during the initial active ack period. In this way, the controller can fail over to getting the activation from the DB without necessarily waiting the entire active ack expiration period.

Because this is more of a failover procedure (rather than a performance optimization) the polling could be done relatively slowly (500ms - 1000ms) to limit the amount of load placed on the DB.

### LINKED ISSUE #2324: Many "Message length exceeded" from a subject will result in lock out (for that subject)
## Environment details:

* Ubuntu 14.04.5 LTS
* Docker version 1.12.0, build 8eab29e
* ansible 2.1.2.0

1.   Invoke an Action which returns a message that exceeds the 1MB limit. Observe that even if you invoke it as "blocking" and the invocation is short, you will wait until timeout and receive only the Activation ID in the response. Get the Activation, observe that the duration is short in spite of the long wait. 
2.   Repeat step 1 (invoke as non-blocking or you'll be waiting a very long time) until you get the error "Too many requests for namespace". Check those Activations and verify that they have completed with the "message length exceeded maximum" error.
3.   Wait a few minutes (or any arbitrary length of time) and try an Action invocation. Observe that you still get the "Too many requests" error, even though none of the requests you made in step 2 are still in flight.

The result I would expect is that in step 1, a "message length exceeded" error is returned to the caller, and in step 3, "Too many requests" would either not occur (because they're quickly being completed, not still in-flight), or would go away after a short time.

Blocking invocations only return the Activation ID after a timeout, and the OW instance becomes locked with "Too many requests". This seems like a Denial of Service vulnerability as well.

## Code Samples
break-openwhisk.js:
```
exports.main = (args) => {
    let bigstring = "";
    for (let i = 0; i < 5000000; i++) {
        bigstring += "A";
    }
    return { bigstring: bigstring, message: "Should have failed" };
};
```
spam.sh:
```
#!/bin/bash

for i in {1..200}
do
    echo "invocation #$i..."
    wsk action invoke -iv break-openwhisk
    sleep 1
done
```

rabbah:
Fixes an active ack accounting bug in the load balancer for certain failed activations (where the invoker previously did not send an ack for the completed (but failed) activation. 

Fixes #2118.
Fixes #2324.
Fixes #1906.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
@@ -43,7 +43,6 @@ import spray.routing.RequestContext
 import spray.routing.Route
 import spray.http.HttpMethods.{ OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH }
 import whisk.common.TransactionId
-import whisk.core.controller.actions.BlockingInvokeTimeout
 import whisk.core.controller.actions.PostActionActivation
 import whisk.core.database._
 import whisk.core.entity._
@@ -336,6 +335,8 @@ trait WhiskWebActionsApi
     def routes(user: Identity)(implicit transid: TransactionId): Route = routes(Some(user))
     def routes()(implicit transid: TransactionId): Route = routes(None)
 
+    private val maxWaitForWebActionResult = Some(WhiskActionsApi.maxWaitForBlockingActivation)
+
     /**
      * Adds route to web based activations. Actions invoked this way are anonymous in that the
      * caller is not authenticated. The intended action must be named in the path as a fully qualified
@@ -540,8 +541,7 @@ trait WhiskWebActionsApi
             // they will be overwritten
             if (isRawHttpAction || context.overrides(webApiDirectives.reservedProperties ++ action.immutableParameters).isEmpty) {
                 val content = context.toActionArgument(onBehalfOf, isRawHttpAction)
-                val waitOverride = Some(WhiskActionsApi.maxWaitForBlockingActivation)
-                invokeAction(actionOwnerIdentity, action, Some(JsObject(content)), blocking = true, waitOverride)
+                invokeAction(actionOwnerIdentity, action, Some(JsObject(content)), maxWaitForWebActionResult, cause = None)
             } else {
                 Future.failed(RejectRequest(BadRequest, Messages.parametersNotAllowed))
             }
@@ -551,12 +551,12 @@ trait WhiskWebActionsApi
     }
 
     private def completeRequest(
-        queuedActivation: Future[(ActivationId, Option[WhiskActivation])],
+        queuedActivation: Future[Either[ActivationId, WhiskActivation]],
         projectResultField: => List[String],
         responseType: MediaExtension)(
             implicit transid: TransactionId) = {
         onComplete(queuedActivation) {
-            case Success((activationId, Some(activation))) =>
+            case Success(Right(activation)) =>
                 val result = activation.resultAsJson
 
                 if (activation.response.isSuccess || activation.response.isApplicationError) {
@@ -583,14 +583,9 @@ trait WhiskWebActionsApi
                     terminate(BadRequest, Messages.errorProcessingRequest)
                 }
 
-            case Success((activationId, None)) =>
+            case Success(Left(activationId)) =>
                 // blocking invoke which got queued instead
                 // this should not happen, instead it should be a blocking invoke timeout
-                logging.warn(this, "activation returned an id, expecting timeout error instead")
-                terminate(Accepted, Messages.responseNotReady)
-
-            case Failure(t: BlockingInvokeTimeout) =>
-                // blocking invoke which timed out waiting on response
                 logging.info(this, "activation waiting period expired")
                 terminate(Accepted, Messages.responseNotReady)
 
--- a/core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
@@ -20,21 +20,28 @@ package whisk.core.controller.actions
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.Promise
-import scala.concurrent.TimeoutException
 import scala.concurrent.duration._
+import scala.util.Failure
 
+import akka.actor.Actor
 import akka.actor.ActorSystem
+import akka.actor.Props
+
 import spray.json._
+
 import whisk.common.Logging
 import whisk.common.LoggingMarkers
+import whisk.common.Scheduler
 import whisk.common.TransactionId
 import whisk.core.connector.ActivationMessage
-import whisk.core.controller.WhiskActionsApi
 import whisk.core.controller.WhiskServices
 import whisk.core.database.NoDocumentException
 import whisk.core.entity._
 import whisk.core.entity.types.ActivationStore
 import whisk.core.entity.types.EntityStore
+import whisk.utils.ExecutionContextFactory.FutureExtensions
+import scala.collection.mutable.Buffer
+import akka.actor.Cancellable
 
 protected[actions] trait PrimitiveActions {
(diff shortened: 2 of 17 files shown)


core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -551,12 +551,12 @@ trait WhiskWebActionsApi
     }
 
     private def completeRequest(
-        queuedActivation: Future[(ActivationId, Option[WhiskActivation])],
+        queuedActivation: Future[Either[ActivationId, WhiskActivation]],
```
markusthoemmes: Nice one :+1:

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -121,46 +123,36 @@ class LoadBalancerService(config: WhiskConfig, entityStore: EntityStore)(implici
...
-        logging.info(this, s"received active ack for '$aid' (forced = $forced)")(tid)
+    private def processCompletion(aid: ActivationId, tid: TransactionId, response: Option[WhiskActivation]): Unit = {
         activationById.remove(aid) match {
             case Some(entry @ ActivationEntry(_, subject, invokerIndex, _, p)) =>
+                logging.info(this, s"received active ack for '$aid' (forced = ${response.isEmpty})")(tid)
                 activationByInvoker.getOrElseUpdate(invokerIndex, new TrieSet[ActivationEntry]).remove(entry)
                 activationBySubject.getOrElseUpdate(subject, new TrieSet[ActivationEntry]).remove(entry)
-                response.foreach(p.trySuccess(_))
-                logging.info(this, s"processed active response for '$aid'")(tid)
+                p.trySuccess(response.map(Right(_)).getOrElse(Left(aid)))
```
markusthoemmes: There is: `opt.toRight({{leftValue}}`, so in this case: `p.trySuccess(response.toRight(aid))`

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -102,95 +106,101 @@ protected[actions] trait PrimitiveActions {
...
+                            // complete the future, which in turn will poison pill this scheduler
+                            activation => p.trySuccess(Right(activation.withoutLogs)) // Logs always not provided on blocking call
+                        } andThen {
+                            case Failure(e: NoDocumentException) => // do nothing, scheduler will reschedule another poll
+                            case Failure(t: Throwable) => // something went wrong, abort
+                                logging.error(this, s"failed while waiting on result: ${t.getMessage}")
+                                p.tryFailure(t) // complete the future, which in turn will poison pill this scheduler
+                        }
+
+                    case _ => throw new Exception("activation completed") // will halt the scheduler
```
markusthoemmes: Why not send a `PoisonPill` to self? Is this also the reason why logLevel is set to Info?

rabbah: for benefit of others the future doesn't have a reference to containing actor.

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -107,10 +109,20 @@ class InvokerReactive(
...
+            val msg = CompletionMessage(transid, res, this.name)
+            producer.send("completed", msg).andThen {
+                case Success(_) =>
+                    logging.info(this, s"posted ${if (recovery) "recovery" else ""} completion of activation ${activationResult.activationId}")
+            }
+        }
+
+        send(Right(activationResult)).onFailure {
+            case t if t.getCause.isInstanceOf[RecordTooLargeException] =>
+                send(Left(activationResult.activationId), recovery = true)
```
markusthoemmes: Does this preserve InvokerHealth behavior which I think inspects the response-code of the message to determine healthyness?

Should we try to keep the old interface `WhiskActivation` and instead back off by sending it without the result? (but including the statuscode).

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -94,7 +94,7 @@ case object ActivationCompleted
  */
 class ContainerProxy(
     factory: (TransactionId, String, ImageName, Boolean, ByteSize) => Future[Container],
-    sendActiveAck: (TransactionId, WhiskActivation) => Future[Any],
+    sendActiveAck: (TransactionId, WhiskActivation) => Unit,
```
markusthoemmes: Why don't we preserve the interface here? I know it's unused but why change it if not needed?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -102,95 +107,137 @@ protected[actions] trait PrimitiveActions {
...
-                    logging.error(this, s"[POST] action activation failed while waiting on result: ${t.getMessage}")
-                    promise.tryFailure(t)
-            }
-        } else {
-            logging.info(this, s"[POST] action activation timed out, terminated polling for result")
+        promise: Promise[Either[ActivationId, WhiskActivation]])(
+            implicit transid: TransactionId) extends Actor {
+
+        // when the future completes, self-destruct
+        promise.future.andThen { case _ => shutdown }
```
markusthoemmes: please add `()` to `shutdown` to indicate the sideeffect.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -102,95 +107,137 @@ protected[actions] trait PrimitiveActions {
...
-     * Polls for activation record. It is assumed that an activation record is created atomically and never updated.
-     * Fetch the activation record by its id. If it exists, complete the promise. Otherwise recursively poll until
-     * either there is an error in the get, or the promise has completed because it timed out. The promise MUST
-     * complete in the caller to terminate the polling.
-     */
-    private def pollDbForResult(
-        docid: DocId,
+    /** Periodically polls the db to cover missing active acks. */
+    val datastorePollPeriodForActivation = 15.seconds
+    val datastorePreemtivePolling = Seq(1.second, 3.seconds, 5.seconds, 7.seconds)
```
markusthoemmes: Typo (preemPtive)

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
rabbah: Heh, because it's unused. I'll revert if you prefer.

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
rabbah: You'll see this later. I think it is ok in that it's a recovery for a specific failure case not all Kafka errors.

markusthoemmes: Edit: I see, the Controller would complete via db polling if I read it correctly?

rabbah:
pg3/761 approved.

rabbah:
Also pg2/1727 🔵

PR #2387, opened 2017-06-15
Update sequence impl to tune controller memory consumption

### LINKED ISSUE #2386: the sequence impl drags memory
for the duration of a sequence activation, the controller "drags" the memory of the input payload, and *all* of the intermediate responses.

if done properly, the sequence controller should only need to keep around the response of the prior component of the sequence, i.e. the max memory consumption should be the maximum payload size of any component (or the input to the sequence); whereas right now, the max memory consumption of a sequence is the sum of all payloads, including the input to the sequence.

starpit:
Fixes #2386

This PR changes SequenceActions to use foldLeft rather than scanLeft. The change requires the introduction of a new accumulator data structure, to maintain the accumulated state for the reduction. The accumulated state includes an array of handles to logs (for the components), the cumulative duration, and a few other bits needed for creation of the final sequence activation record.

This fix also patches a closely related memory drag, caused by the way timeouts of sequence invocations are handled. See #2294 which patches other parts of the controller for the same kind of timeout drag. This patch needs to be a little different, due to the complex way the SequenceActions impl uses nested Promises.

PG3 625 🔵

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
@@ -222,7 +222,7 @@ protected[actions] trait SequenceActions {
      * @param user the user invoking the sequence
      * @param seqAction the sequence invoked
      * @param seqActivationId the id of the sequence
-     * @param payload the payload passed to the first component in the sequence
+     * @param inputPayload the payload passed to the first component in the sequence
      * @param components the components in the sequence
      * @param cause the activation id of the sequence that lead to invoking this sequence or None if this sequence is topmost
      * @param atomicActionCnt the dynamic atomic action count observed so far since the start of the execution of the topmost sequence
@@ -260,10 +260,17 @@ protected[actions] trait SequenceActions {
             (accountingFuture, futureAction) =>
                 accountingFuture.flatMap { accounting =>
                     if (accounting.atomicActionCnt < actionSequenceLimit) {
-                        invokeNextAction(user, futureAction, accounting, cause)
+                        invokeNextAction(user, futureAction, accounting, cause).flatMap { accounting =>
+                            if (!accounting.shortcircuit) {
+                                Future.successful(accounting)
+                            } else {
+                                // this is to short circuit the fold
+                                Future.failed(FailedSequenceActivation(accounting)) // terminates the fold
+                            }
+                        }
                     } else {
                         val updatedAccount = accounting.fail(ActivationResponse.applicationError(sequenceIsTooLong), None)
-                        Future.failed(FailedSequenceActivation(updatedAccount))
+                        Future.failed(FailedSequenceActivation(updatedAccount)) // terminates the fold
                     }
                 }
         }.recoverWith {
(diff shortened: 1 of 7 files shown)


rabbah:
@tardieu @markusthoemmes @ioana-blue might any of you want to review this?

rabbah:
@starpit as I noted on slack, we should have unit tests for the new/refactored code. It's far more testable now and we can get good coverage. If you don't have bandwidth for this, I will work on that tomorrow.

core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
```diff
@@ -365,55 +289,195 @@ protected[actions] trait SequenceActions {
...
+ */
+protected[actions] object SequenceAccounting {
+
+    def maxMemory(prevMemoryLimit: Option[Int], newMemoryLimit: Option[Int]): Option[Int] = {
+        prevMemoryLimit.map { prevMax =>
+            newMemoryLimit
+                .map(currentMax => Some(Math.max(prevMax, currentMax)))
+                .getOrElse(prevMemoryLimit)
+        }.getOrElse(newMemoryLimit)
+    }
```
markusthoemmes: Is this the same as:

```
(prevMemoryLimit ++ newMemoryLimit).reduceOption(Math.max)
```

core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
```diff
@@ -365,55 +289,195 @@ protected[actions] trait SequenceActions {
...
+ * @param duration the "user" time so far executing the sequence (sum of durations for
+ *        all actions invoked so far which is different from the total time spent executing the sequence)
+ * @param maxMemory the maximum memory annotation observed so far for the
+ *        components (needed to annotate the sequence with GB-s)
+ * @param shortcircuit when true, stops the execution of the next component in the sequence
+ */
+protected[actions] case class SequenceAccounting(
+    atomicActionCnt: Int,
+    previousResponse: AtomicReference[ActivationResponse],
+    logs: mutable.ArrayBuffer[ActivationId],
```
markusthoemmes: Do we rely on this being an **Array**Buffer anywhere, or could it aswell be the more general `Buffer`?

core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
```diff
@@ -365,55 +289,195 @@ protected[actions] trait SequenceActions {
...
+                    limitsAnnotation.asJsObject.getFields("memory") match {
+                        case Seq(JsNumber(memory)) => Some(memory.toInt) // we have a numerical "memory" field in the "limits" annotation
+                    }
+            } getOrElse { None })
+    }
+
+    /** The previous activation failed (this is used when there is no activation record or an internal error. */
+    def fail(failureResponse: ActivationResponse, activationId: Option[ActivationId]) = {
+        require(!failureResponse.isSuccess)
+        activationId.foreach(logs += _)
```
markusthoemmes: `logs.appendAll(activationId)`?

PR #2388, opened 2017-06-15
Create swagger operationId values for code tooling consumption

mdeuser:
Currently, the CLI generates operationId swagger field values that are usable by coding tools.  The PR alters the operationId algorithm to generate strings that more closely match allowable programmatic variable/method names.

Field description from the Open API 2.0 specification

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operationObject 

>Unique string used to identify the operation. The id MUST be unique among all operations described in the API. Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is recommended to follow common programming naming conventions.


mdeuser:
pg4/401

PR #2390, opened 2017-06-16
Improve docker caching for cli builds

rabbah:
Separate dep fetching from rest of build to avoid repeated fetches when none changed.


rabbah:
@dubeejw @mdeuser I think this will avoid repeatedly fetching dependencies when none changed even though the rest of the source did. Thoughts?

rabbah:
@csantanapr this commit is also part of #2112 but cherry picked here in case we continue to debate the breaking change there.

csantanapr:
already covered by #2112

Issue #2394, opened 2017-06-16
Pin version of CCurl used in Package.swift 


paulcastro:
Swift 3 `action should invoke` tests will sometimes fail if CCurl is not using version 0.2.3 due to self-signed certificate issue.  This can affect local builds

We should update Package.swift:

```import PackageDescription

let package = Package(
    name: "Action",
        dependencies: [
            .Package(url: "https://github.com/IBM-Swift/CCurl.git", "0.2.3"),
            .Package(url: "https://github.com/IBM-Swift/Kitura-net.git", "1.0.1"),
            .Package(url: "https://github.com/IBM-Swift/SwiftyJSON.git", "14.2.0"),
            .Package(url: "https://github.com/IBM-Swift/swift-watson-sdk.git", "0.4.1")
        ]
)
```

paulcastro:
Merged, PR #2120  will fix this problem for Swift 3.1.1

PR #2395, opened 2017-06-16
Pin CCurl version in Package.swift

paulcastro:
Fixes issue #2394.  Pins CCurl version to 0.2.3 to fix potential self-signed certificates issue.


paulcastro:
PG 1656

csantanapr:
👍

PR #2402, opened 2017-06-20
Add Configurable Limits to Controller

dubeejw:
Allows for the configuration of the maximum number of activation records to be displayed for activation list requests, and also allows for the configuration of the blocking action timeout value.

In addition, to making the above values configurable, the mentioned values will also be displayed from the controller's info route.

The goal here is to provide default, min, and max parameter values via controller info route so that the UI/CLI can dynamically read these values.

### DIFF
--- a/ansible/environments/local/group_vars/all
+++ b/ansible/environments/local/group_vars/all
@@ -17,9 +17,12 @@ db_port: "{{ lookup('ini', 'db_port section=db_creds file={{ playbook_dir }}/db_
 limits:
   actions:
     invokes:
-      perMinute: 60
       concurrent: 30
       concurrentInSystem: 5000
+      perMinute: 60
+  activations:
+    polls:
+      maxRecords: 200
   triggers:
     fires:
       perMinute: 60
--- a/ansible/environments/local/group_vars/all
+++ b/ansible/environments/local/group_vars/all
@@ -17,6 +17,7 @@ db_port: "{{ lookup('ini', 'db_port section=db_creds file={{ playbook_dir }}/db_
 limits:
   actions:
     invokes:
+      blockingTimeout: 60
       concurrent: 30
       concurrentInSystem: 5000
       perMinute: 60
--- a/ansible/environments/mac/group_vars/all
+++ b/ansible/environments/mac/group_vars/all
@@ -22,9 +22,12 @@ db_port: "{{ lookup('ini', 'db_port section=db_creds file={{ playbook_dir }}/db_
 limits:
   actions:
     invokes:
-      perMinute: 60
       concurrent: 30
       concurrentInSystem: 5000
+      perMinute: 60
+  activations:
+    polls:
+      maxRecords: 200
   triggers:
     fires:
       perMinute: 60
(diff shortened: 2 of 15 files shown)


dubeejw:
Needs tests.

rabbah:
> Needs tests.
tests usually come first :D

ansible/group_vars/all
```diff
@@ -78,11 +78,15 @@ runtimesManifest:
 defaultLimits:
   actions:
     invokes:
-      perMinute: 120
+      blockingTimeout: 60
```
rabbah: are you planning to add memory and timeout limits?

dubeejw:
@rabbah, @csantanapr, currently all of the limits/default values are displayed in the controller info route under the limit key. However, I think it would be more organized if the limits were organized in corresponding sections for actions, triggers, packages, etc. See below for examples.

Current format:

```
"limits": {
    "actions_per_minute": 120,
    "action_blocking_timeout": 60,
    "concurrent_actions": 100,
    "activation_poll_record_limit": 200,
    "triggers_per_minute": 60
  }
```

Purposed change:
```
"limits": {
    "actions":
         "invoke": {
              "per_minute": 120,
              "blocking_timeout": 60,
              "concurrent": 100
         }
    },
    "activations": {
         "poll": {
               record_limit": 200
          }
    },
    "triggers": {
         "fire": {
              "per_minute": 60
         }
    }
}
```

dubeejw:
TODO:
Make `MAX_ACTIVATION_ENTITY_LIMIT` configurable via Ansible: /common/scala/src/main/scala/whisk/core/entity/ActivationEntityLimit.scala

ansible/environments/local/group_vars/all
```diff
@@ -17,9 +17,13 @@ db_port: "{{ lookup('ini', 'db_port section=db_creds file={{ playbook_dir }}/db_
...
   actions:
     invokes:
-      perMinute: 60
+      blockingTimeout: 60
       concurrent: 30
       concurrentInSystem: 5000
+      perMinute: 60
+  activations:
+    polls:
+      maxRecords: 200
```
rabbah: i thought there was a limit on all list operations; it looks like for activations it is 200 but for the others, only a default is provided and no limit... this is an oversight.

ansible/environments/local/group_vars/all
```diff
@@ -17,9 +17,13 @@ db_port: "{{ lookup('ini', 'db_port section=db_creds file={{ playbook_dir }}/db_
 limits:
   actions:
     invokes:
-      perMinute: 60
+      blockingTimeout: 60
```
rabbah: maybe this is a better name of this? `requestResponseTimeout`?

ansible/environments/mac/group_vars/all
```diff
@@ -22,9 +22,13 @@ db_port: "{{ lookup('ini', 'db_port section=db_creds file={{ playbook_dir }}/db_
 limits:
   actions:
     invokes:
-      perMinute: 60
+      blockingTimeout: 60
```
rabbah: since these are the same as the "all" values, i think you can just drop them.

rabbah:
@dubeejw this has rotted - are you still pursuing this?

dubeejw:
@rabbah, still needs to be done. Seems very out of sync with master. Just got side tracked with other work.

PR #2403, opened 2017-06-20
[wip] Draft 1 - Wrap runs within OpenTracing spans

jpkrohling:
This is just a first draft, intended to start a discussion on the OpenTracing integration. This includes also a concrete OpenTracing tracer implementation (Jaeger) but the code itself is only using OpenTracing APIs. 

Things to discuss:
- [ ] How to use other Trace implementations. I'm not sure how the classpath works for Scala applications, but from the OpenTracing's perspective, it will get whatever implementation is on the classpath (ideally, there would be exactly one)
- [ ] Location of the span -- I added the OpenTracing code on the `run()` method on the invoker, but I'm not 100% familiar with OpenWhisk yet. Is this the right place?

This is how it looks like in Jaeger with this PR:

List of traces (only one span per trace for now):
image

Span details:
image

If this PR is merged, I'll work on context propagation, so that a trace would have more spans (example: if `hello` invokes `aloha`, both are shown in the same trace)


jpkrohling:
Closed in favor of #2413

PR #2409, opened 2017-06-22
Drain stdout/stderr in futures to prevent deadlock.

rabbah:
Failing to drain streams properly will deadlock the process launcher.


markusthoemmes:
Should we move away from the Java implementation and instead use `ProcessRunner`, which we also use in the Invoker?

rabbah:
I was so tempted to do that but there are dependencies on test utils from other repos, I didn't want to break those. Also we use the features of setting the env, piping stdin, and work dir there's no analog for these but I admit I spent only a few minutes in this.

rabbah:
pg 1/1694 🔵

PR #2412, opened 2017-06-22
Use hosts to define order of invokers instead of play_hosts

cbickel:
Using play_hosts seems, like the order is not always the same like in the hosts file.
This PR fixes this issue, by using the hosts file to generate the index for the invokers.


cbickel:
PG4#424 is green.

PR #2413, opened 2017-06-22
Integration with OpenTracing api

Jiri-Kremser:
UPDATE: I've enhanced the PR for the sequences.

This is based on https://github.com/apache/incubator-openwhisk/pull/2403 that added the tracing for the invoker. However, opentracing allows to leverage the paren-child hierarchy among the spans so this PR also adds the support for sequences.

When starting a sequence action a new (parent) span is created and the member actions starts sub-spans. This works also if the sequence itself is member of another sequence. Using the open tracing api, it is easy to change the tracer implementation, simply by providing other jar with tracer impl that is compliant and implements correctly the SPI mechanism (`META-INF/services`). By default, the jeager is used.

Hopefully, the code isn't too invasive. I had to change couple of method signatures, because it is needed to propagate the span context from the controller to the invokers so that the parent-child link can be created. In theory, the `cause` or `activationId` could have been used as a correlation id between child and parent, but to make it implementation agnostic, the opentracing API [requires](http://opentracing.io/documentation/pages/api/cross-process-tracing.html#required-injectextract-carrier-formats) to use either text map or byte array as a "carrier".

a pic:
gnome-shell-screenshot-1qmi2y


rabbah:
no null! 👍

core/invoker/src/main/scala/whisk/core/container/WhiskContainer.scala
```diff
@@ -92,17 +96,53 @@ class WhiskContainer(
     }
 
     /**
+     * Starts the tracing.
+     */
+    private def startSpan(msg: ActivationMessage): Option[Span] = {
+        val tracerOption = if (GlobalTracer.isRegistered) Option.empty else Some(TracerResolver.resolveTracer)
+        tracerOption foreach GlobalTracer.register
```
jpkrohling: It might be that `tracerOption` is null at this point:

https://travis-ci.org/apache/incubator-openwhisk/builds/245871954#L6909-L6913
```
    java.lang.NullPointerException: Cannot register GlobalTracer <null>.
        at io.opentracing.util.GlobalTracer.register(GlobalTracer.java:83)
        at whisk.core.container.WhiskContainer$$anonfun$startSpan$1.apply(WhiskContainer.scala:104)
        at whisk.core.container.WhiskContainer$$anonfun$startSpan$1.apply(WhiskContainer.scala:104)
        at scala.Option.foreach(Option.scala:257)
```

core/invoker/src/main/scala/whisk/core/container/WhiskContainer.scala
```diff
@@ -92,17 +96,53 @@ class WhiskContainer(
     }
 
     /**
+     * Starts the tracing.
+     */
+    private def startSpan(msg: ActivationMessage): Option[Span] = {
+        val tracerOption = if (GlobalTracer.isRegistered) Option.empty else Some(TracerResolver.resolveTracer)
```
Jiri-Kremser: actually, this doesn't work because 
```scala
Option(GlobalTracer.isRegistered).exists(identity)
```
returns either `true` or `false` and I can't call map on Boolean

this works:

```scala
Option(GlobalTracer.isRegistered).flatMap(if (_) Option.empty else Option(TracerResolver.resolveTracer))
```

but perhaps it's too much "flatmappy" :)

core/invoker/src/main/scala/whisk/core/container/WhiskContainer.scala
Jiri-Kremser: right, I didn't realize the `TracerResolver.resolveTracer()` can return also null and when using `Some(null)` it actually creates the `Some(null)` what I need here is the `Option(null)` that results in `None`

rabbah:
Option(null) == None.

rabbah:
oops, use filter instead of exists.

core/invoker/src/main/scala/whisk/core/container/WhiskContainer.scala
Jiri-Kremser: Right, `s/exists/filter/` just worked :+1:

jpkrohling:
@Jiri-Kremser , may I close #2403 in favor of this one?

Jiri-Kremser:
> [@Jiri-Kremser](https://github.com/jiri-kremser) , may I close [#2403](https://github.com/apache/incubator-openwhisk/pull/2403 "[wip] Draft 1 - Wrap runs within OpenTracing spans") in favor of this one?

yes

common/scala/src/main/scala/whisk/core/connector/Message.scala
```diff
@@ -56,7 +56,8 @@ case class ActivationMessage(
     activationNamespace: EntityPath,
     rootControllerIndex: InstanceId,
     content: Option[JsObject],
-    cause: Option[ActivationId] = None)
+    cause: Option[ActivationId] = None,
+    tracingMetadata: Option[Map[String, String]] = None)
```
rabbah: how big might this map get?

rabbah: how big might this map get?

in general this is going to be a problem - we have a standing issue that the control and data plane are conflated.
because these message pass over kafka, and carry the user payload, every bit added to the control is one less bit for the user payload; in other words, we have a bug today that a user cannot send a 1MB payload through and taking away more bits makes it worse... hence the question.

for data, I expect that message to flow externally through a redis for example.

core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
```diff
@@ -353,6 +369,24 @@ protected[actions] trait SequenceActions {
 
     /** Max atomic action count allowed for sequences */
     private lazy val actionSequenceLimit = whiskConfig.actionSequenceLimit.toInt
+
+    /** Starts the tracing for the sequence action and creates the tracing metadata hash map into which
+     *  the tracer implementation  inserts its own ids, these ids are then propagated via the ActivationMessage into
+     *  invoker that can add the reference from the primitive action to the parent sequence action
+     */
+    private def startTracingHelper(cause: Option[ActivationId],
```
rabbah: we should consider if this is fully going to subsume the markers generated in the logs if we want to have two way of tracing or just one because the information is going to be duplicated.

common/scala/src/main/scala/whisk/core/connector/Message.scala
Jiri-Kremser: In case of jaeger it's one key-value pair, where key is "jaeger-id" (or similar) and value some short string, zipkin is similar. So it's couple of bytes. What can also help is setting the sampling to some lower value, like 5% so that only random 5% of all actions will be traced (which reminded me that I should probably set the [default value](https://github.com/apache/incubator-openwhisk/pull/2413/files#diff-1048d039011bd0a7de7d2763733749e8R41) to something more sane, like 10% and perhaps make it configurable in the wisk.properties (?))

>for data, I expect that message to flow externally through a redis for example.

not sure I follow you here, are you planning to add also redis in the future?

core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
Jiri-Kremser: I wasn't aware of the markers in the akka logging system. Do I get it right that you [use it](https://github.com/apache/incubator-openwhisk/blob/master/common/scala/src/main/scala/whisk/common/Logging.scala#L125) for calculating how long some execution took?

common/scala/src/main/scala/whisk/core/connector/Message.scala
rabbah: on the last point/question: https://github.com/apache/incubator-openwhisk/issues/254.
as part of replicating the controller, i expect redis will be in the mix and might double multiple duties.

core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
jeremiaswerner: @Jiri-Kremser yes, that's correct. We basically use a `start` and a `finish` marker. These markers are set in the code using a log marker token, i.e. `activationRun`. Setting the finish marker requires the start marker object. The context of this marker pair is the current transaction, i.e. user request.

In the code you basically instrument a log marker as for the activation execution, i.e. `activationRun`:
```
        implicit transid: TransactionId
        ....
        val start = transid.started(this, LoggingMarkers.INVOKER_ACTIVATION_RUN, s"sending arguments to $actionName at $id $ip")
        ....
       transid.finished(this, start.copy(start = r.interval.start), s"running result: ${r.toBriefString}", endTime = r.interval.end)
```
[example here](https://github.com/apache/incubator-openwhisk/blob/master/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala#L162)

For both markers we write a log entry. The log entry for the `start` marker contains the time since transaction start. The log entry for the `finish` marker contains time since transaction start and the time since it's corresponding start marker.

start marker (141 ms since transaction start)
```
... sending arguments to ...[marker:invoker_activationRun_start:141]
```

finish marker (516 ms since transaction start, and time of `activationRun` took 400ms)
```
... running result: ok [marker:invoker_activationRun_finish:516:400] 
```

Based on this it's possible to calculate the times within a transaction between start and finish markers of the same log marker token and also between log marker tokens.

Jiri-Kremser:
@rabbah is there anything else you want me to improve on this PR? Is it ok to have the open tracing code next to the markers in the logs as the first step? The tracing can be turned off or set to 10% using the env properties, so that only the fraction of the calls will be traced.

rabbah:
There is a second tracing pr #2282 - does it make sense to consider consolidating efforts?

jpkrohling:
I don't think so: there has been a thread on the mailing list, and it looks like that one has a different goal.

rabbah:
Closing per discussion on the dev list https://mail-archives.apache.org/mod_mbox/incubator-openwhisk-dev/201707.mbox/browser.

PR #2414, opened 2017-06-22
SPI approach for pluggable implementations

tysonnorris:
This set of changes introduces SPI concepts and example default spi + impls for:
- db (ArtifactStoreProvider + CouchDBStoreProvider)
- messaging (MessagingProvider + KafkaMessagingProvider)

will likely need a bit of cleanup, but should be enough to get the discussion started


common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
```diff
@@ -0,0 +1,37 @@
...
+import whisk.common.Logging
+import whisk.core.WhiskConfig
+import whisk.core.connector.MessageConsumer
+import whisk.core.connector.MessageProducer
+import whisk.core.connector.MessagingProvider
+import whisk.spi.SpiFactoryModule
+/**
+  * Created by tnorris on 6/20/17.
+  */
+class KafkaMessagingProvider(actorSystem:ActorSystem, config:WhiskConfig)(implicit logging:Logging) extends MessagingProvider {
```
clebertsuconic: What I still don't get on OpenWhisk.. is why not call this MessagingPRovider, and Kafka being the implementation?

If you want to make it pluggable, make it pluggable...

tysonnorris: Not sure what you mean - you want to change the name from "KafkaMessagingProvider" to "Kafka"?

There is already a "MessagingProvider" trait defined (which is the one that extends Spi):
https://github.com/apache/incubator-openwhisk/pull/2414/files#diff-8ba49f030237d2d826b76d762a468013R9

clebertsuconic: ah.. ok.. my bad.. I though this was the intended SPI to be extended.

rabbah:
Will this support loading a plugin from a config file?

tysonnorris:
@rabbah if you mean "indicating which spi impl is used based on config file" then yes. Note there is no adjustment of classpath in this PR, so adding additional impls that are not part of the codebase of this repo will require some additions to the classpath, or a separate classloader etc

common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
```diff
@@ -20,22 +20,21 @@ package whisk.connector.kafka
...
 import org.apache.kafka.clients.producer.ProducerRecord
 import org.apache.kafka.clients.producer.RecordMetadata
 import org.apache.kafka.common.serialization.StringSerializer
-
 import whisk.common.Counter
 import whisk.common.Logging
 import whisk.core.connector.Message
 import whisk.core.connector.MessageProducer
 
+import scala.concurrent.ExecutionContext
```
rabbah: this file is unchanged just reorg of imports, drop it?

common/scala/src/main/scala/whisk/core/connector/MessageProducer.scala
```diff
@@ -18,10 +18,9 @@
 package whisk.core.connector
 
 import scala.concurrent.Future
-
 import org.apache.kafka.clients.producer.RecordMetadata
 
-trait MessageProducer {
+trait MessageProducer{
```
rabbah: formatting nit.
we have an eclipse style file that we recommend using.

common/scala/src/main/scala/whisk/core/entity/WhiskStore.scala
```diff
@@ -86,18 +86,18 @@ protected[core] trait WhiskDocument
...
-        actorSystem: ActorSystem,
-        logging: Logging): ArtifactStore[D] = {
-        require(config != null && config.isValid, "config is undefined or not valid")
-        require(config.dbProvider == "Cloudant" || config.dbProvider == "CouchDB", "Unsupported db.provider: " + config.dbProvider)
-        assume(Set(config.dbProtocol, config.dbHost, config.dbPort, config.dbUsername, config.dbPassword, name(config)).forall(_.nonEmpty), "At least one expected property is missing")
-
-        new CouchDbRestStore[D](config.dbProtocol, config.dbHost, config.dbPort.toInt, config.dbUsername, config.dbPassword, name(config))
-    }
-}
+//protected[core] object Util {
```
rabbah: should this be removed since it's moved to the adapter?

common/scala/src/main/scala/whisk/spi/SpiProvider.scala
```diff
@@ -0,0 +1,181 @@
...
+import scaldi.MutableInjector
+import whisk.common.Logging
+import whisk.core.WhiskConfig
+
+import scala.collection.JavaConverters._
+import scala.reflect.ClassTag
+import scala.reflect.runtime.universe._
+
+/**
+  * Created by tnorris on 6/8/17.
```
rabbah: we avoid any such annotations in general; kindly remove.
git blame ftw.

tests/src/test/scala/whisk/core/cli/test/SequenceMigrationTests.scala
```diff
@@ -55,6 +54,9 @@ class SequenceMigrationTests
     val wsk = new Wsk
     val whiskConfig = new WhiskConfig(WhiskEntityStore.requiredProperties)
     // handle on the entity datastore
+
+    SharedModules.initSharedModules(List(new SharedModule(actorSystem, whiskConfig, logging)))
```
rabbah: perhaps this should go into a helper trait that gets weaved in with many of tests?  perhaps `WskActorSystem`?

tests/src/test/scala/whisk/core/controller/SpiTest.scala
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    val producer1 = MessagingProvider(actorSystem)
+    val producer2 = MessagingProvider(actorSystem)
+    val producer3 = MessagingProvider(actorSystem)
+    val producer4 = MessagingProvider(actorSystem)
+    println(s"producer1: ${producer1}")
+    println(s"producer2: ${producer2}")
+    println(s"producer3: ${producer3}")
+    println(s"producer4: ${producer4}")
```
rabbah: is this a WIP test?

tysonnorris: yes - replaced with a proper test

tests/src/test/scala/whisk/core/cli/test/SequenceMigrationTests.scala
tysonnorris: actually I think this is no longer required - previously it was instantiating all Spi instances on load of first instance, now it only instantiates impl on access of first instance of that Spi type; removed these calls for now, but if they are needed, I now have a way to add them as you describe.

common/scala/src/main/scala/whisk/spi/SpiProvider.scala
tysonnorris: done!

common/scala/src/main/scala/whisk/core/entity/WhiskStore.scala
tysonnorris: done!

common/scala/src/main/scala/whisk/core/connector/MessageProducer.scala
tysonnorris: done!

common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
tysonnorris: done!

tysonnorris:
@rabbah when you have time, can you review and/or merge this? Thanks!

markusthoemmes:
@tysonnorris would you adjust the scala formatting to the rest of the code as defined here: https://github.com/apache/incubator-openwhisk/tree/master/tools/eclipse

common/scala/src/main/scala/whisk/spi/SpiProvider.scala
```diff
@@ -0,0 +1,195 @@
...
+  * Spi providers must extend SpiProvider to expose an Extension for accessing the Spi impl
+  *
+  * @param configKey an key in application.conf which indicates which impl will be used at runtime
+  * @param tag
+  * @param classTag
+  * @tparam T
+  */
+abstract class SpiProvider[T <: Spi](configKey: String)(implicit tag: TypeTag[T], classTag:ClassTag[T])extends ExtensionId[T] with ExtensionIdProvider {
+
+  override def apply(system: ActorSystem): T = {
```
markusthoemmes: Why is this dependent on the `ActorSystem`?

tysonnorris:
@markusthoemmes I'm having trouble applying the eclipse style using intellij - intellij only loads the style into java editor, it does not seem to use it in the scala editor...let me know if you have any advice on getting this done with intellij?

common/scala/src/main/scala/whisk/spi/SpiProvider.scala
tysonnorris: Because Extensions uses the ActorSystem as a service registry; see details at http://doc.akka.io/docs/akka/snapshot/scala/extending-akka.html

tysonnorris:
@markusthoemmes I think formatting is proper now - let me know if you have other feedback?

common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
```diff
@@ -0,0 +1,47 @@
...
+import whisk.core.WhiskConfig
+import whisk.core.connector.MessageConsumer
+import whisk.core.connector.MessageProducer
+import whisk.core.connector.MessagingProvider
+import whisk.spi.SpiFactoryModule
+
+/**
+ * A Kafka based implementation of MessagingProvider
+ */
+class KafkaMessagingProvider(actorSystem: ActorSystem, config: WhiskConfig)(implicit logging: Logging) extends MessagingProvider {
```
rabbah: Does it make sense here to not use implicit arguments?

common/scala/src/main/scala/whisk/spi/SpiProvider.scala
```diff
@@ -0,0 +1,195 @@
...
+import scala.collection.mutable.ListBuffer
+import scala.reflect.ClassTag
+import scala.reflect.runtime.universe._
+
+/**
+ * An Spi (Service Provider Interface) is an extension point.
+ * At runtime, multiple Spi implementations may exist in the classpath, but only a single implementation
+ * will be available for consumers to use. The specific implementation available at runtime is governed by the
+ * configKey specified in the SpiProvider
+ */
```
rabbah: Do you intend for the Spi trait to be more than a marker interface in the future?

common/scala/src/main/scala/whisk/core/database/ArtifactStoreProvider.scala
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * An Spi for providing ArtifactStore implementations
+ */
+
+trait ArtifactStoreProvider extends Spi {
+    def makeStore[D <: DocumentSerializer](config: WhiskConfig, name: WhiskConfig => String)(
+        implicit jsonFormat: RootJsonFormat[D],
+        actorSystem: ActorSystem,
+        logging: Logging): ArtifactStore[D]
```
markusthoemmes: For the `ArtifactStoreProvider` you "hardcoded" the dependencies of all implementations here where for the MessagingProvider you inject them using DI.

If we look at the set of dependencies we have, it seems to boil down to:
- Config
- ActorSystem
- Logging

Could we implement the same scheme for the MessagingProvider and get rid of the scaldi dependency that way?

rabbah: @tysonnorris what do you think about the suggestion above?
I think this is likely to evolve further but we can at least start to experiment with the plugins.
It may be worth though consolidating those dependencies for now since it will avoid having to change dependent adaptors.

I can run a jenkins job to execute the full test suite in the meantime to make sure nothing broke.

tysonnorris: I'm on vacation so will look into it more towards the end of this week.

I think it's ok except that it will imply that all spi impls have (only) the same dependencies passed in - so it definitely won't work for everything. I can see if it can work for these 2, and if there are others that seem more problematic will try to surface those sooner than later.

tysonnorris: @rabbah @markusthoemmes I removed the dependency on scaldi. I chatted with Markus briefly and wanted tomention a couple things:
- not using DI implies that the dependencies are pushed down a layer to the SPI method signatures, this is ok, but implies that all impls have the same args, which may or may not be ok; not pushing them down would work, but requires some more work to remove ServiceLoader usage, I'm not sure it is that important.
- Markus suggested an alternative way to pass dependencies in a generic object - with the current impl, this would be an option to add as a signature on a specific SPI, but I haven't done that with the current example impls (MessagingProvider + ArtifactStoreProvider)
- you can see from the current signature in MessagingProvider, it is leaking impl details, so would be a good candidate to move those kafka details (like maxPollInterval) into the SPI impl by way of config (or the dependencies object approach); I mostly left these signatures alone to keep them from drifting too far before this approach is adopted
- I left some commented code in SpiTest to indicate how some things change removing scaldi, where instead of constructing objects as SPI impls, those now need to be built from some factory layer (which is ok, but if any of this sounds confusing, it might be worth a look)

common/scala/src/main/scala/whisk/spi/SpiProvider.scala
markusthoemmes: Can we get rid of the extra curlies here?

docs/spi.md
```diff
@@ -0,0 +1,89 @@
...
+  def getInstance(implicit injector:Injector) = new YourImpl(inject[SomeDependencyType])
+}
+```
+
+Note: To bind dependencies (must be done before first access of the Spi):
+```scala
+//bind config
+SharedModules.bind[WhiskConfig](whiskConfig)
+//bind an Spi instance
+SharedModules.bind[YourSpi](yourSpi)
```
markusthoemmes: Remove the section on ScalDI?

common/scala/src/main/scala/whisk/core/database/CouchDBStoreProvider.scala
```diff
@@ -0,0 +1,44 @@
...
+        require(config.dbProvider == "Cloudant" || config.dbProvider == "CouchDB", "Unsupported db.provider: " + config.dbProvider)
+        assume(Set(config.dbProtocol, config.dbHost, config.dbPort, config.dbUsername, config.dbPassword, name(config)).forall(_.nonEmpty), "At least one expected property is missing")
+
+        new CouchDbRestStore[D](config.dbProtocol, config.dbHost, config.dbPort.toInt, config.dbUsername, config.dbPassword, name(config))
+    }
+}
+
+class CouchDBStoreProviderModule extends SpiModule[ArtifactStoreProvider] {
+    def getInstance = new CouchDBStoreProvider
+}
```
markusthoemmes: The interface seems weird to me. Why do we need the intermediary object if the class has a constructor without parameters? Can we even implement this solely in a factory object?

tysonnorris: Which interface? The `SpiModule[Spi]` interface makes it so that a) SPIs are constructed in a specific way (currently using ServiceLoader), and (via ServiceLoader) all impls are listed in a specific file (META-INF/services/whisk.spi.SpiModule). Otherwise, we have to provide a separate/custom mechanism for discovering impls of a specific interface. ServiceLoader works fine for this IMHO. It also provides an abstraction point to allow either DI or some extended dependency support in the future, without changing the signature of the SPI itself.

docs/spi.md
tysonnorris: yep, done

common/scala/src/main/scala/whisk/core/database/CouchDBStoreProvider.scala
markusthoemmes: Right, but the Factory objects are basically noops? What prevents you from directly instantiating the `CouchDBStoreProvider` in this case vs. having to write the companion + the arbitrary `getInstance` method?

common/scala/src/main/scala/whisk/spi/SpiLoader.scala
```diff
@@ -0,0 +1,94 @@
...
+    override def getClassnameForKey(key: String): String = key
+}
+
+/**
+ * Lookup the classname for the SPI impl based on a key in the provided Config
+ * @param config
+ */
+class TypesafeConfigClassResolver(config: Config) extends SpiClassResolver {
+    override def getClassnameForKey(key: String): String = config.getString(key)
+}
```
markusthoemmes: I find the ClassResolver usage a bit weird. The TypesafeConfig keys look like a fully qualified classname, which makes it confusing to me and I have to lookup which resolver is used.

Shall we make this independent of the underlying impl and just always use the ClassnameResolver (i.e. remove the need to have a Resolver alltogether). We can always take the specific key out of typesafeconfig as a String and pass it down. Seems like that would be less overhead than what we have now.

common/scala/src/main/scala/whisk/spi/SpiLoader.scala
```diff
@@ -0,0 +1,94 @@
...
+
+class Dependencies(deps: Any*) {
+    def get[T](implicit tag: ClassTag[T]) = {
+        //check for interface compatible types
+        deps.find(d => tag.runtimeClass.isAssignableFrom(d.getClass)) match {
+            case Some(d) => d.asInstanceOf[T]
+            case None => throw new IllegalArgumentException(s"missing dependency of type ${tag.runtimeClass.getName}")
+        }
+    }
+}
```
markusthoemmes: Awesome idea just using the dependency list :+1:

This can be written a bit more concisely to get rid of unnecessary casting and the companion object. Also we should use `Manifest` vs `ClassTag`, it's included in the stdlib vs. needing the reflect package.

```scala
case class Dependencies(private val deps: Any*) {
    require(deps.map(_.getClass).distinct.size == deps.size, "A type can only occur once as a dependency")

    def get[T](implicit man: Manifest[T]): T =
        deps.find(d => man.runtimeClass.isAssignableFrom(d.getClass)) match {
            case Some(d: T) => d
            case _          => throw new IllegalArgumentException(s"missing dependency of type ${man.runtimeClass.getName}")
        }
}
```

common/scala/src/main/scala/whisk/spi/SpiLoader.scala
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+trait Spi {
+}
```
markusthoemmes: omit the curlys

common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
```diff
@@ -0,0 +1,43 @@
...
+    def getConsumer(config: WhiskConfig, groupId: String, topic: String, maxPeek: Int, maxPollInterval: FiniteDuration)(implicit logging: Logging): MessageConsumer = {
+        new KafkaConsumerConnector(config.kafkaHost, groupId, topic, maxPeek, maxPollInterval = maxPollInterval)
+    }
+
+    def getProducer(config: WhiskConfig, ec:ExecutionContext)(implicit logging: Logging): MessageProducer = new KafkaProducerConnector(config.kafkaHost, ec)
+}
+
+object KafkaMessagingProvider extends SingletonSpiFactory[MessagingProvider] {
+    override def buildInstance(dependencies: Dependencies): MessagingProvider = new KafkaMessagingProvider
+}
```
markusthoemmes: We could just have `object KafkaMessagingProvider extends MessagingProvider` and implement the `getConsumer`/`getProducer` there. Singletons are guaranteed by the underlying JVM then, making the SingletonSpiFactory (and SpiFactory itself) unnecessary.

The question at hand is: What is the interface for us? To me it was more like I do:

```scala
SpiLoader.instanceOf[MessageProducer]("openwhisk.KafkaProducer")
```

and get an instance of the `KafkaProducer`. Of course that is non-sensical in that Producer and Consumer should be off the same implementation, so I think you're approach is valid. I wonder though if we can generalize then, that each `Spi` needs to be a `Provider`, that is: A factory object providing methods to actually generate the instances.

common/scala/src/main/resources/META-INF/services/whisk.spi.SpiModule
```diff
@@ -0,0 +1,2 @@
+whisk.core.database.CouchDBStoreProviderModule
+whisk.connector.kafka.KafkaMessagingProviderModule
```
markusthoemmes: These should no longer be needed

common/scala/build.gradle
```diff
@@ -10,6 +10,7 @@ repositories {
 
 dependencies {
     compile "org.scala-lang:scala-library:${gradle.scala.version}"
+    compile "org.scala-lang:scala-reflect:${gradle.scala.version}"
```
markusthoemmes: As noted below, we can get around that dependency by using `MAnifest` over `ClassTag`

common/scala/src/main/scala/whisk/spi/SpiLoader.scala
tysonnorris: I find it weird to use a classname when that classname is always abstracted away in a config. So either I have to create a variable to hold the classname, or have an extra tedious signature like `SpiLoader.getInstance[Spi](config.get("this.is.the.key"))` instead of simply `SpiLoader.getInstance[Spi]("this.is.the.key")`. Its already independent of the impl, in a way, since you can always provide a different resolver. In general I would expect to use the same resolver everywhere, although that it not required. 

I also don't think its a good idea to make it too easy to specify the classname directly in the code since this would be a bad idea. (Although I had to do it in some test classes to get them to work)

common/scala/src/main/scala/whisk/spi/SpiLoader.scala
tysonnorris: good tip thanks!

common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
tysonnorris: I ran into a similar issue with db, but in similar ways, it does not make sense for each entity store to be an spi (and use a different db), although it could be done that way. 

yes dependencies are optional currently and not used; I'm not sure of a great case for them but if there is an Spi that does NOT behave as a factory, that would be it. I would leave them for now.

common/scala/src/main/resources/META-INF/services/whisk.spi.SpiModule
tysonnorris: removed

common/scala/build.gradle
tysonnorris: got it, thx

common/scala/src/main/scala/whisk/spi/SpiLoader.scala
markusthoemmes: Actually, the signatures you shared look pretty similar for me and I would always prefer the first over the latter due to composability and separation of concerns. Also I find having to have the resolvers in scope super tedious. Could we at least make the resolvers `import`able? If both were objects, that wouldn't be so much an issue.

However as per separation of concerns, i don't see much value in tying the implementation to a config provider.

tysonnorris: AFIK the resolver that depends on config cannot be an object since it requires the config? Or maybe I'm not understanding - do you have an example or suggestion on how to do this. 

The problem as I see it is that the contract that SpiLoader.getInstance() should enforce is that you should not pass a "string", rather a "config value" that contains the classname, so that if you are using SpiLoader, you are explicitly exposing a configuration value that can be replaced.

markusthoemmes: Right, and in my mind that is too tightly coupled. One could need to compute the configuration value or something.

Config can be loaded globally if i'm not mistaken. Don't think it is dependent on the actor system or anything.

If you really want to go that way, can we even generalize on the config key you want to read? It's essentially defined by the type you want to get for example:

```
whisk.spi {
    ArtifactStoreProvider: whisk.database....
    MessagingProvider: whisk.connector...
}
```

tysonnorris: AFAIK config is not loaded globally for WhiskConfig or Akka, but maybe I'm mistaken? If that needs changing I don't think it is a good idea to include in the PR but a separate one.

For `generalize the config key`, I think that is a good idea, but it would still require a way to expose the config binding to the loader (via resolver or globally or some other approach).

markusthoemmes: Just tried it out to convince myself i'm not talking bs:

```scala
object SpiLoader {
    val config = ConfigFactory.load()

    def instanceOf[A <: Spi](overrideClass: Option[String] = None)(implicit deps: Dependencies, man: Manifest[A]) = {
        val toLoad = overrideClass.orElse(Try(config.getString(s"whisk.spi.${man.runtimeClass.getSimpleName}")).toOption)

        toLoad match {
            case Some(className) =>
                val clazz = Class.forName(className + "$")
                clazz.getField("MODULE$").get(clazz).asInstanceOf[A]
            case None =>
                throw new IllegalStateException(s"No implementation for ${man.runtimeClass.getName} found")
        }
    }
}
```

You're right, WhiskConfig is not loadable globally (although, now with everything available in the env, it is actually). As shown here, typesafeconfig can be loaded into the object quite easily.

tysonnorris: Ah sure, I was propagating the Akka config previously. I'm not sure there is any difference in (re)constructing the config directly via TypesafeConfig, but this will certainly remove the leaky class resolver. I will set it as the default

common/scala/src/main/scala/whisk/spi/SpiLoader.scala
```diff
@@ -65,15 +72,20 @@ object SpiLoader {
...
+    private val config = ConfigFactory.load()
+
     override def getClassnameForKey(key: String): String = config.getString(key)
     override def getKeyForType[T](implicit man: Manifest[T]): String = "whisk.spi." + man.runtimeClass.getSimpleName
 }
 
+/**
+ * Object containing arbitrary objects acting as dependencies.
+ *
+ * This is solely a helper type to cross the border between possibly heterogeneous Spi
```
tysonnorris: Should "heterogeneous Spi interfaces" be "heterogeneous SpiFactoy implementations"?

docs/spi.md
```diff
@@ -0,0 +1,75 @@
...
+
+```scala
+class YourImplFactory extends SpiFactory[YourSpi]{
+  def apply(dependencies: Dependencies): { ...construct the impl...}
+}
+```
+for singleton behavior you can use
+```scala
+class YourImplFactory extends SingletonSpiFactory[YourSpi]{
+  def buildInstance(dependencies: Dependencies): { ...construct the impl...}
```
markusthoemmes: outdated.

docs/spi.md
```diff
@@ -0,0 +1,75 @@
...
+}
+```
+
+## Invoke SpiLoader.instanceOf to acquire an instance of the SPI
```
markusthoemmes: outdated. is `SpiLoader.get` now

docs/spi.md
```diff
@@ -0,0 +1,75 @@
...
+
+SpiLoader accepts a key to use for resolving which impl should be loaded. The mapping of key to classname is implemented in an instance of `whisk.spi.SpiClassResolver`
+Example whisk.spi.SpiClassResolver impls are:
+* whisk.spi.ClassnameResolver - classname is not interpreted by used literally from the value of the key
+* whisk.spi.TypesafeConfigClassResolver - classname is looked up as a String from the provided instance of com.typesafe.config.Config (available via `ActorSystem.settings.config` within an Akka app)
```
markusthoemmes: Outdated. There's only TypesafeConfigClassResolver left, and it's not available via ActorSystem anymore.

docs/spi.md
```diff
@@ -0,0 +1,75 @@
...
+
+Invoke the loader using `SpiLoader.instanceOf[<the SPI interface>](<key to resolve the classname>)(<implicit resolver>)`
+
+The configKey indicates which config (in application.conf) is used to determine which impl is used at runtime.
+```scala
+val messagingProvider = SpiLoader.instanceOf[MessagingProvider]("whisk.spi.messaging.impl")
```
markusthoemmes: Outdated.

docs/spi.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+## Invoke SpiLoader.instanceOf to acquire an instance of the SPI
+
+SpiLoader accepts a key to use for resolving which impl should be loaded. The mapping of key to classname is implemented in an instance of `whisk.spi.SpiClassResolver`
+Example whisk.spi.SpiClassResolver impls are:
+* whisk.spi.ClassnameResolver - classname is not interpreted by used literally from the value of the key
+* whisk.spi.TypesafeConfigClassResolver - classname is looked up as a String from the provided instance of com.typesafe.config.Config (available via `ActorSystem.settings.config` within an Akka app)
+
+Invoke the loader using `SpiLoader.instanceOf[<the SPI interface>](<key to resolve the classname>)(<implicit resolver>)`
```
markusthoemmes: Outdated.

PR #2416, opened 2017-06-23
Action limit test for log truncation produces less log entries

sven-lange-last:
The test now uses the minimum supported log limit (1M) and creates 10% more log output as the limit. In the past, the test had a limit of 2M and produced 3M logs.

We frequently saw this test failing in mainOpenwhisk in the past few days because it took too
long to PUT the activation record to Cloudant, i.e. more than 60 sec. We suspect that
concurrent load tests in YS0 may have caused the long Cloudant response times because the
Cloudant DB cluster is shared. But we don`t have enough diagnostic data to prove this assumption.

At the same time, making the log portion in the activation record smaller with this change
should also make the Cloudant request faster. This change does only affect limits and sizes but
does not change the test's design.


sven-lange-last:
PG2 / 1729 running

sven-lange-last:
@cbickel Please merge this PR - all checks including PG passed successfully.

PR #2418, opened 2017-06-23
Fix licenses on scala and java files.

rabbah:
Some new scala files were added with the old license.


mrutkows:
Thanks @rabbah ... I pulled down the PR code and ran the latest scancode util. against it and it passed all checks (as visual inspection also indicated it would).

PR #2420, opened 2017-06-26
Support client certificate on cli and nginx

ningyougang:
In order to increase the security of auth, it is necessary to add client
certificate on cli and nginx. So user can use wsk -i property set --cert
openwhisk-client-cert.pem --key openwhisk-client-key.pem to pass client
certificate to nginx.


style95:
LGTM.
If you are trying to open a pr in upstraem and get feedbacks, it would be great to specify that this is the first pr of two subsequent prs(server side).

But I am anxious that they will not accept it without server side changes.

ningyougang:
The whole logic is that if add client certificate, there has two authenticate method
* When use `wsk -i property --cert cert.pem --key key.pem............`, the controller will ignore `--auth` verfiy, directly passed.
* keep original logic for authenticate, if execute like this `wsk -i property --auth ${auth}.....`

style95:
But to pass authentication, we need changes in controller, right?

ningyougang:
yes, in order to keep the whole logic (already said above), should make changes in controller,
if controller identities nginx has verified the client certificate, the controller has no need to verify auth key.

ningyougang:
see here:https://github.com/apache/incubator-openwhisk/pull/2427

PR #2421, opened 2017-06-26
Fix order of messages being sent by the ContainerPool.

markusthoemmes:
The ContainerPool indicates to the ActivationFeed that it can perform more work and as a result gets more messages from the ActivationFeed. In order to properly back-pressure and protect the ContainerPool, it only ever sends such a message iff it changed its internal state to an extent that it is guaranteed to be able to handle more work.

The idea is to send a message to the feed once we remove a container from the busy pool, thus freeing up space to take more load.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
@@ -112,7 +112,7 @@ class ContainerPool(
         // Container is free to take more work
         case NeedWork(data: WarmedData) =>
             freePool.update(sender(), data)
-            busyPool.remove(sender())
+            busyPool.remove(sender()).foreach(_ => feed ! MessageFeed.Processed)
 
         // Container is prewarmed and ready to take work
         case NeedWork(data: PreWarmedData) =>
@@ -121,11 +121,7 @@ class ContainerPool(
         // Container got removed
         case ContainerRemoved =>
             freePool.remove(sender())
-            busyPool.remove(sender())
-
-        // Activation completed
-        case ActivationCompleted =>
-            feed ! MessageFeed.Processed
+            busyPool.remove(sender()).foreach(_ => feed ! MessageFeed.Processed)
     }
 
     /** Creates a new container and updates state accordingly. */
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
@@ -106,6 +106,7 @@ class ContainerPool(
                     freePool.remove(actor)
                     actor ! r // forwards the run request to the container
                 case None =>
+                    logging.error(this, "Rescheduling Run message, too many message in the pool")(r.msg.transid)
                     self ! r
             }
 
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
@@ -67,12 +67,6 @@ case object Remove
 case class NeedWork(data: ContainerData)
 case object ContainerPaused
 case object ContainerRemoved
-/**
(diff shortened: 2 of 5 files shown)


markusthoemmes:
PG4 441 🔵

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -106,12 +106,13 @@ class ContainerPool(
                     actor ! r // forwards the run request to the container
                 case None =>
                     self ! r
+                    logging.error(this, "sent message to self. Feed buffering is off.")
```
rabbah: i dont understand this log message.

tests/src/test/scala/whisk/core/dispatcher/test/DispatcherTests.scala
```diff
@@ -113,20 +113,21 @@ class DispatcherTests
...
                 val logs = stream.toString()
-                logs should include regex (s"waiting for activation pipeline to drain: ${half + 1} > $half")
+                logs should include regex (s"waiting for activation pipeline to drain: ${capacity + 1} > $capacity")
             }
 
             // send one message and check later that it remains in the connector
-            // at this point, total messages sent = half + 2
-            sendMessage(connector, half + 2)
+            // at this point, total messages sent = half + 1
+            sendMessage(connector, capacity + 2)
```
rabbah: hmm comment says +1 code says +2.

core/invoker/src/main/scala/whisk/core/dispatcher/ActivationFeed.scala
```diff
@@ -103,14 +110,29 @@ protected class ActivationFeed(
             } else logging.debug(this, "dropping fill request until feed is drained")
 
         case n: ActivationNotification =>
+            downstreamCapacity += 1
```
rabbah: should this saturate to max capacity defensively?

core/invoker/src/main/scala/whisk/core/dispatcher/ActivationFeed.scala
```diff
@@ -103,14 +110,29 @@ protected class ActivationFeed(
...
             }
+            send()
             fill()
     }
 
+    /** Send as many messages as we're allowed to, or as we have depending on what happens first. */
+    @tailrec
+    private def send(): Unit = {
+        if (downstreamCapacity > 0 && outstanding.size > 0) {
+            downstreamCapacity -= 1
```
rabbah: i think this is missing pipeline occupancy update (fill, then send all messages at once).

rabbah: it seems unnecessary to had a pipeline capacity and in the presence of an internal buffer - the size of the buffer is the pipeline capacity.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -120,11 +120,7 @@ class ContainerPool(
         // Container got removed
         case ContainerRemoved =>
             freePool.remove(sender())
-            busyPool.remove(sender())
-
-        // Activation completed
-        case ActivationCompleted =>
-            feed ! ContainerReleased
+            busyPool.remove(sender()).foreach(_ => feed ! ContainerReleased)
```
rabbah: so where previously the slow was released on completion, and a new request may be received when a container isn't available, instead, indicate capacity availability when the container pool can actually readily allocate a new slot - is that correct?

markusthoemmes: Correct. In effect this should be entirely the same as before but with less boilerplate and it actually sends the message **after** we're free again. With correct buffering upfront this completely removes the "self-send" tight-loop, which is desirable.

markusthoemmes:
@rabbah 
PG2 1808 reapproved and proved no [ERROR] message indicating a resent message as outlined in this PR.

note the change to `Invoker.scala`

PR #2425, opened 2017-06-26
Throttle message bus consumption.

rabbah:
Replaces unthrottled Kafka consumer with FSM actor that can throttle the consumption.
Applied for active active acks and invoker health pings.


common/scala/src/main/scala/whisk/core/connector/MessageConsumer.scala
```diff
@@ -38,15 +47,157 @@ trait MessageConsumer {
...
+            Try {
+                // Grab next batch of messages and commit offsets immediately
+                // essentially marking the activation as having satisfied "at most once"
+                // semantics (this is the point at which the activation is considered started).
+                // If the commit fails, then messages peeked are peeked again on the next poll.
+                // While the commit is synchronous and will block until it completes, at steady
+                // state with enough buffering (i.e., maxPipelineDepth > maxPeek), the latency
+                // of the commit should be masked.
+                val records = consumer.peek(longPollDuration)
+                consumer.commit()
```
markusthoemmes: Are these blocking operations? They'll block the actor and prevent it from receiving potentially state-changing messages. Is that expected?

rabbah: No reviewers assigned yet;)

common/scala/src/main/scala/whisk/core/connector/MessageConsumer.scala
```diff
@@ -38,15 +53,188 @@ trait MessageConsumer {
...
+
+    // double-buffer to make up for message bus read overhead
+    val maxPipelineDepth = maximumHandlerCapacity * 2
+    private val pipelineFillThreshold = maxPipelineDepth - consumer.maxPeek
+
+    require(consumer.maxPeek <= maxPipelineDepth, "consumer may not yield more messages per peek than permitted by max depth")
+
+    private val outstandingMessageQueue = new ConcurrentLinkedQueue[(String, Int, Long, Array[Byte])]()
+    private val outstandingMessageCount = new AtomicInteger() // avoid non-contant time size calculation on queue
+    private val handlerCapacity = new AtomicInteger(maximumHandlerCapacity)
```
markusthoemmes: You shouldn't need concurrent datastructures here, if you don't close over inernal state in `Futures` inside of the Actor. You can achieve this by `pipe`ing the `Future` to `self` to receive the messages and handle them then. That also drops the need for the `FillComplete` message.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -151,13 +141,23 @@ class InvokerPool(
...
         val raw = new String(bytes, StandardCharsets.UTF_8)
         PingMessage.parse(raw) match {
-            case Success(p: PingMessage) => self ! p
-            case Failure(t)              => logging.error(this, s"failed processing message: $raw with $t")
+            case Success(p: PingMessage) =>
+                self ! p
+                invokerPingFeed ! MessageFeed.Processed
+
+            case Failure(t) =>
+                invokerPingFeed ! MessageFeed.Processed
```
markusthoemmes: Send this message after the `match`?

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -234,22 +244,33 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
...
+    })
+
+    def processActiveAck(bytes: Array[Byte]): Future[Unit] = Future {
         val raw = new String(bytes, StandardCharsets.UTF_8)
         CompletionMessage.parse(raw) match {
             case Success(m: CompletionMessage) =>
                 processCompletion(m.response, m.transid, false)
                 // treat left as success (as it is the result a the message exceeding the bus limit)
                 val isSuccess = m.response.fold(l => true, r => !r.response.isWhiskError)
+                activationFeed ! MessageFeed.Processed
```
markusthoemmes: Send this message after the match to avoid duplication?

common/scala/src/main/scala/whisk/core/connector/MessageConsumer.scala
```diff
@@ -38,15 +53,188 @@ trait MessageConsumer {
...
+                // state with enough buffering (i.e., maxPipelineDepth > maxPeek), the latency
+                // of the commit should be masked.
+                val records = consumer.peek(longPollDuration)
+                consumer.commit()
+                (records, records.size)
+            }.map {
+                case (records, count) =>
+                    outstandingMessageQueue.addAll(records)
+                    outstandingMessageCount.addAndGet(count)
+                    sendOutstandingMessages(handlerCapacity.get)
```
markusthoemmes: These should happen responding to an even being sent by the Future (the future could send: `Seq[KafkaMessage]`)

common/scala/src/main/scala/whisk/core/connector/MessageConsumer.scala
```diff
@@ -38,15 +53,188 @@ trait MessageConsumer {
...
+            }.recover {
+                case e: CommitFailedException => logging.error(this, s"failed to commit $description consumer offset: $e")
+                case e: Throwable             => logging.error(this, s"exception while pulling new $description records: $e")
+            }
+        } else {
+            logging.error(this, s"dropping fill request until $description feed is drained")
+        }
+    }.onComplete {
+        case _ => self ! FillCompleted
+    }
```
markusthoemmes: after the change above, this could be `pipeTo(self)`

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
rabbah: i duped so that the message goes out before the logger but it's probably inconsequential in the totality.

tests/src/test/scala/whisk/core/connector/test/TestConnector.scala
```diff
@@ -70,13 +70,15 @@ class TestConnector(
     }
 
     private val producer = new MessageProducer {
-        def send(topic: String, msg: Message): Future[RecordMetadata] = {
+        import scala.concurrent.ExecutionContext.Implicits.global
+
+        def send(topic: String, msg: Message): Future[RecordMetadata] = Future {
```
markusthoemmes: Why not use the old impl. and avoid the implicit ExecutionContext?

rabbah:
pg2/1769 🚀

rabbah:
Pg3/813 🔵

PR #2427, opened 2017-06-27
Support client certificate on cli and nginx

ningyougang:
In order to increase the security of auth, it is necessary to add client
certificate on cli and nginx. So user can use wsk -i property set --cert
openwhisk-client-cert.pem --key openwhisk-client-key.pem to pass client
certificate to nginx. If you don't want to use default client certificate
which system provides, you can create your own client certificate instead
of them.


ningyougang:
This commit just increases the security of auth using `wsk -i property set --cert
openwhisk-client-cert.pem --key openwhisk-client-key.pem --auth XXX...`.But for controller side, i don't know whether there has need to ignore --auth parameter verify if user passes client certificate. the whole logic may like this
* When use `wsk -i property --cert cert.pem --key key.pem............`, the controller will ignore `--auth` verfiy, directly passed, because already verify client certificate by nginx, so there has no need to verify auth key.
* keep original logic for authenticate, if execute like this `wsk -i property --auth ${auth}.....`

So, guys, what's your opinion for above controller side's logic for client certificate?

rabbah:
The controller looks up the key to get identity information. It will still need to do that if I understand your proposal.

ningyougang:
@rabbah i agree with you, if the identitiy information needs to use for some other wheres.
Can you review this the codes, if has something need to improve, i will modify it.

ansible/roles/nginx/files/openwhisk-client-ca-key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
rabbah: should we perhaps generate these as part of the setup rather that store them.
i know it's convenient and it does take a few seconds to get a key.
since i generated the previous keys, i'm also asking myself if that was the right thing to do and we should avoid that. that way you can see what values are used for the key, and can change them.

rabbah:
> if the identitiy information needs to use for some other wheres

it is necessary - the entire flow in the backend for a crud operation or an activation is subject to entitlement based on an identity.

ningyougang:
@rabbah ,for your two suggestions
* should we perhaps generate these as part of the setup rather that store them.
i know it's convenient and it does take a few seconds to get a key.

answer: I have add the key generate script, so key/csr/cert generated flow is fully, if users don't want to use the default client certificate which system provides, they can generate client certificate by themselves refer to genssl.sh script.

* can you add a test you can add that actually uses the client cert?

answer: I think there has no need to add the test, because this test case: https://github.com/ningyougang/incubator-openwhisk/blob/support-client-certificate/tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala#L96 can verify the client certificate by nginx whether right. when execute this test case, it will execute
`wsk -i property set --cert XX --key XX --auth XX --namespace XX --apihost XX`, this cmd will add cert, key to request, see here: https://github.com/ningyougang/incubator-openwhisk/blob/support-client-certificate/tools/cli/go-whisk/whisk/client.go#L87 ,after build Authorization in request Header, it will send the quest to https://${apihost}/api/v1/namespaces, then nginx will deal the request, verfiy the
client certificate whether right, if it is right, the test case running result will be PASSED, if the client certificate is not right, the test case running result will be FAILED

ningyougang:
For above's suggestion two. i have tested it my local env, if the client certificate is not right, the test case will be failed (openwhisk-client-key1.pem doesn't exist)
```
whisk.core.cli.test.WskBasicUsageTests > Wsk CLI usage should set apihost, auth, cert, key and namespace FAILED
    org.scalatest.exceptions.TestFailedException: /home/irteam/jenkins/workspace/lambda/bin/wsk property set -i --apihost ${IP} --auth 23bc46b1-71f6-4ed5-8c54-816aa4f8c502:123zO3xZCLrMN6v2BKK1dXYFpXlPkccOFqm12CdAsMgRU4VrNZ9lyGVCGuMDGIwP --cert /home/irteam/jenkins/workspace/lambda/ansible/roles/nginx/files/openwhisk-client-cert1.pem --key /home/irteam/jenkins/workspace/lambda/ansible/roles/nginx/files/openwhisk-client-key1.pem --namespace guest
    ok: client cert set. Run 'wsk property get --cert' to see the new value.
    ok: client key set. Run 'wsk property get --key' to see the new value.
    ok: whisk auth set. Run 'wsk property get --auth' to see the new value.
    ok: whisk API host set to ${IP}

    error: Authenticated user does not have namespace 'guest'; set command failed: Get https://${IP}/api/v1/namespaces: x509: cannot validate certificate for ${IP} because it doesn't contain any IP SANs
    exit code: 1 was not equal to 0
        at org.scalatest.Matchers$AnyShouldWrapper.shouldBe(Matchers.scala:6864)
        at common.RunWskCmd$$anonfun$cli$1.apply(Wsk.scala:1019)
        at org.scalatest.Assertions$class.withClue(Assertions.scala:1221)
```

csantanapr:
Initial set of comments:
I think is a great idea to have an extra level of security from client to apihost.

A few comments for discussion:
## UX
I think to having the user needing to configure the wsk CLI with a cert and a key properties on top of auth key is going to be a pain.

Why not have the cert and key be compiled as a resource into the wsk CLI binary?

This way user only need to set the `auth` property
For dev, I agree using `--insecure | -i` then you can use the any wsk CLI binary with any apihost.

Maybe is a good idea to still allow property set for `cert` and `key` as an easy way to override in cases operators or contributors quick test different of certs/keys saving the time to re-compile
 
### Docs
Hopefully for users no Docs :-), this change would be transparent in production, user will just get the binary from the openwhisk provider/operator by downloading for example for the same apihost edge (i.e. cli/go/download/mac/amd64/OpenWhisk_CLI-mac.zip) where the binary will already have the client cert burned in, and backend can validate that the binary being used is a trusted source.

I think we still needs some documentation for contributors/operators on how to configure, test, change the settings (i.e. key size, algorithm, etc..) etc to make client and backend configured properly.

### Operations/Deployment
I think we need to roll this in in a way that is toggle-featured, to give time operators currently using the ansible config/files as is a chance to adapt and setup their nginx and front doors. Also toggled in the CLI.
This way people can try it out from master without being the default.

### Testing
I don't see negative tests, if the cert or key are wrongly configured, test the expected errors is shown and no access. have multiple of this, creating a cert with correct keys, but wrong expiration date, etc.. different combination making sure if, this could also be be feature toggle via gradle task/config

ningyougang:
@csantanapr , for your suggestions
* Why not have the cert and key be compiled as a resource into the wsk CLI binary?

answer: because different user should have different cert/key, so compiled cert/key as a resource into wsk CLI binary for every user is not a good idea.   User should execute cmd `wsk -i property set --cert XX --key XX...` only one time, the cert/key will store in ~/.wskprops.

* For docs.

answer: i will add the docs on cli.md

* Testing

answer: i will add some negative tests for support this.

jeremiaswerner:
@ningyougang fyi - I'am currently reviewing and trying to understand the details of this PR and running a PG. will come back to you soon.

ningyougang:
hi, @jeremiaswerner , for @csantanapr 's three suggestion
* add Docs for cli  -------->has beed added.
* add negative test for client certificate verify ------->has been added.
* add the cert and key be compiled as a resource into the wsk CLI binary -----> i think we should discuss it deeply ^^.

ningyougang:
a) are there individual client certificates we need to deal with, or is it just the auto-generated one?
 answer: in openwhisk codes, we don't need to deal with individual client certificates, it is best to generated by the company who use openwhisk to deploy their serverless env for all users. The client certificate can be signed by their self-signed root certificate or trusted CA organization sigend root certificate.
    a.1) if there are individual (per user) client certificates how does those get generated so that it is valid with the OW server certificate?
 answer: server certificate has no relation with client certificate, their root certifiate can be different or same, also worked well. the per user client certificate is generated by
 their self-signed root certificate or trusted CA organization sigend root certificate.(above also refered)`
    a.2) if there are individual (per user) client certificates does they contain any namespace related information that can be used upstream, i.e. openwhisk controller, to authorize the request?
 answer: for our case, the client certificate subj includes CN=${emploeeid}, and other fields such as C, ST, L, O may be the depend on company's organization


b) what exact security issue does this solve? does it protect for man-in-the-middle attacks only, or have you other ideas in mind to use that mechanism?
   answer: this feature just increase the security for auth, but it is optional default, if  set ${nginx.ssl.verify_client} values to on, wsk client must pass client certificate, not every user can use wsk tool to
send request to openwhisk server, if want to do that, must get their own client certificate from comany.  Other guess, we may simplify the auth flow, if pass client certificate, in controlller, we can igore the
auth key verify.(but i don't sure it has possible to simplify the auth flow)


c) is the usage of the client certificate mandatory?  
answer: no, it is optional, default configuration for {ssl.nginx.verify_client} value is optional, so user can pass client certificate or not, also worked well, if set  {ssl.nginx.verify_client} value to on, so user must
pass client certifciate.

d) I saw that you’re generating the server and client certificate during `setup.yml`. what would be the steps to generate the corresponding per user client certificate?
answer: openwhisk has no need to care how to generate the client certificate, it just provide the configuration entry point  how to use it. setup.yml's genssh.sh just  a example provides steps how to generate client certificate.

e) how does it work if I don’t have a hostname and only an IP. I’ve tried it with my `docker-machine` environment and get `x509: cannot validate certificate for 192.168.99.101 because it doesn't contain any IP SANs`
answer: i should verity it.

style95:
I think this feature is worth in the aspect of providing different authentication method.
X509 client certificate is very common way to authenticate client(user).
Many cloud provider such as AWS also support it.

If OW supports this feature, user can just issue certificate for him and set it up for subsequent activities.
In certificate, there will be information about users(subject).
Since certificate is validated by root(trusted) CA, once it is validated, no more authentication process is required in server side.

This can coexists with current auth method(`--auth`), and no need other extra component, it would be great to have this optional feature.

jeremiaswerner:
@ningyougang @style95 many thanks for clarifying and answering my questions above. that helps to better understand. It would be great if you could verify on e) because this might be a typical case in local or test environments and we should avoid that test cases are failing (like in openstack, kubernetes, docker-machine, ... environments)

ansible/roles/nginx/files/genssl.sh
```diff
@@ -8,22 +8,71 @@ if [ "$#" -ne 1 ]; then
...
+#openssl genrsa -out "$SCRIPTDIR/openwhisk-server-key.pem" 2048
 
-echo generating certificate request
-openssl req -new \
-    -key "$SCRIPTDIR/openwhisk-key.pem" \
-    -nodes \
+if [ "$TYPE" == "server" ]; then
+    echo generating server certificate request
+    openssl req -new \
+        -key "$SCRIPTDIR/openwhisk-server-key.pem" \
```
jeremiaswerner: Since the certificate and key name are configurable in `{{ nginx.ssl.cert }}` parameters, would it make sense to pass these as in input parameter to the script as well?

style95: If custom `{{ nginx.ssl.cert }}` is configured, certificate generation step will be skipped.

ningyougang: @jeremiaswerner @style95 , i think there has no need to pass `{{ nginx.ssl.cert}}` parameter to genssh.sh script. Becuase for here's server certificate and client certificate, they are all untrusted, they just example. If user wants to use their own trusted certificate, its name can't be equal with 
`openwhisk-server-cert.pem` (for server certificate) and `openwhisk-client-ca-cert.pem` (for client certificate), and copy their own trusted certificate under roles/nginx/file directory before deploy openwhisk. please refer here: https://github.com/ningyougang/incubator-openwhisk/blob/support-client-certificate/ansible/setup.yml#L35 , if doesn't meets the condition, the ansible task will not be executed. So uses just don't keep the name equal with it, users can use their trusted certificate.

tests/src/test/scala/common/Wsk.scala
```diff
@@ -71,6 +71,8 @@ import whisk.utils.retry
 
 case class WskProps(
     authKey: String = WhiskProperties.readAuthKey(WhiskProperties.getAuthFileForTesting),
+    cert: String = WhiskProperties.getProperty("openwhisk.home") + "/ansible/roles/nginx/files/openwhisk-client-cert.pem",
+    key: String = WhiskProperties.getProperty("openwhisk.home") + "/ansible/roles/nginx/files/openwhisk-client-key.pem",
```
markusthoemmes: The first part could be

```
WhiskProperties.whiskHome
```

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -108,6 +108,36 @@ class WskBasicUsageTests
         }
     }
 
+    it should "set valid cert key to get expected success result for client certificate verification" in {
+        val tmpwskprops = File.createTempFile("wskprops", ".tmp")
+        try {
+            val namespace = wsk.namespace.list().stdout.trim.split("\n").last
+            val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
+            val stdout = wsk.cli(Seq("property", "set", "-i", "--apihost", wskprops.apihost, "--auth", wskprops.authKey,
+                "--cert", wskprops.cert, "--key", wskprops.key, "--namespace", namespace), env = env).stdout
```
markusthoemmes: Adding a comment on how `--namespace` forces an API request and thereby a cert. validation would be helpful to understand the testcase later on.

ansible/group_vars/all
```diff
@@ -166,10 +166,12 @@ nginx:
...
   ssl:
     path: "{{ openwhisk_home }}/ansible/roles/nginx/files"
-    cert: "openwhisk-cert.pem"
-    key: "openwhisk-key.pem"
+    cert: "openwhisk-server-cert.pem"
+    key: "openwhisk-server-key.pem"
     password_enabled: false
     password_file: "ssl.pass"
+    client_ca_cert: "openwhisk-client-ca-cert.pem"
+    verify_client: "optional"
```
markusthoemmes: Could we make this configurable per env? Your documentation suggests that that's the intended behavior anyway.

You can use the following scheme to make it configurable:

```
verify_client: {{ nginx_ssl_verify_client | default("off") }}
```

In `ansible/environments/local` for example you can then turn the feature on by adding:

```
nginx_ssl_verify_client: "optional"
```

In general it'd be great to be able to completely turn this feature off, so we don't impact existing deployments of OpenWhisk too much. You might need to make testcases optional with regards to this property then.

What do you think?

tests/src/test/scala/common/Wsk.scala
ningyougang: @markusthoemmes 
`WhiskProperties.whiskHome` can't be accessed from `Wsk.scala`, because it is private `private static final String whiskHome;`

ansible/group_vars/all
ningyougang: @markusthoemmes , i agree with you, make it default value for `off`, if the value is `off`, skip the test case otherwise execute the test case,  but how can i get the `{{ nginx.ssl.verify_cilent }}` value in `WskBasicJavaTests.scala`?

markusthoemmes: @ningyougang You can write it out to `whisk.properties` and read it from there using the `WhiskProperties` singleton.

tests/src/test/scala/common/Wsk.scala
rabbah: There is a method to return a path relative to whisk home which looks like it will suite you.

ningyougang: @rabbah , i don't find the method to return a relative path to whisk home.
i just find blow method may be can use
```
    /**
     * The path to the CLI directory.
     */
    public static String getCLIDir() {
        return whiskHome + "/bin";
    }

    /**
     * The path to the Go CLI executable.
     */
    public static String getCLIPath() {
        return getCLIDir() + "/wsk";
    }
```

ansible/group_vars/all
ningyougang: @markusthoemmes ,thank you, got your idea

tests/src/test/scala/common/Wsk.scala
ningyougang: Can you show the method, if doesn't exist, how about add a new method  in `WhiskProperties` like below
```
 /**
     * The path to the whisk home directory.
     */
    public static String getCLIDir() {
        return whiskHome;
    }

Comment (rabbah) [inline tests/src/test/scala/common/Wsk.scala]: getFileRelativeToWhiskHome

**Comment by ningyougang** [comment]
@rabbah @markusthoemmes @jeremiaswerner 
I have modifid  the codes as your suggestions. please review again.

File: tests/src/test/scala/common/Wsk.scala (hunk 3 above)
Comment (ningyougang) [inline tests/src/test/scala/common/Wsk.scala]: @rabbah ,thank you, i have modified.

File: ansible/group_vars/all (hunk 5 above)
Comment (ningyougang) [inline ansible/group_vars/all]: @markusthoemmes , i have modified.

**Comment by jeremiaswerner** [comment]
🔵 PG4/570

File: tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala [hunk 6]
```diff
@@ -112,6 +112,41 @@ class WskBasicUsageTests
...
+                stdout should include(s"ok: whisk namespace set to ${namespace}")
+            } finally {
+                tmpwskprops.delete()
+            }
+        }
+
+        it should "set invalid cert key to get expected exception result for client certificate verification" in {
+            val tmpwskprops = File.createTempFile("wskprops", ".tmp")
+            try {
+                val namespace = wsk.namespace.list().stdout.trim.split("\n").last
```
rabbah: This should be Whois. Also seems unnecessary to do all this to validate the cert. can't you use it with wsk namespace list in the first place?

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -112,6 +112,41 @@ class WskBasicUsageTests
         }
     }
 
+    // If client certificate verification is off, should ingore run below tests.
+    if (!WhiskProperties.getProperty("whisk.ssl.client.verification").equals("off")){
+        it should "set valid cert key to get expected success result for client certificate verification" in {
+            val tmpwskprops = File.createTempFile("wskprops", ".tmp")
```
rabbah: What is the difference between this test and the ones below which checks property set/unset?

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -112,6 +112,41 @@ class WskBasicUsageTests
         }
     }
 
+    // If client certificate verification is off, should ingore run below tests.
+    if (!WhiskProperties.getProperty("whisk.ssl.client.verification").equals("off")){
+        it should "set valid cert key to get expected success result for client certificate verification" in {
```
rabbah: Is there a positive test using the certificate?

tests/src/test/scala/common/Wsk.scala
```diff
@@ -71,6 +71,8 @@ import whisk.utils.retry
 
 case class WskProps(
     authKey: String = WhiskProperties.readAuthKey(WhiskProperties.getAuthFileForTesting),
+    cert: String = WhiskProperties.getFileRelativeToWhiskHome("ansible/roles/nginx/files/openwhisk-client-cert.pem").getAbsolutePath,
```
rabbah: Are these supposed to be passed to the wsk cmds?

docs/cli.md
```diff
@@ -52,3 +52,11 @@ After you have configured your environment, you can begin using the OpenWhisk CL
 The CLI can be setup to use an HTTPS proxy. To setup an HTTPS proxy, an environment variable called `HTTPS_PROXY` must be created. The variable must be set to the address of the HTTPS proxy, and its port using the following format:
 `{PROXY IP}:{PROXY PORT}`.
 
+## Configure the CLI to use client certificate
```
rabbah: This needs a little bit of editing.

ningyougang: How to modify?

tests/src/test/scala/common/Wsk.scala
ningyougang: yes
```
 val stdout = wsk.cli(Seq("property", "set", "-i", "--apihost", wskprops.apihost, "--auth", wskprops.authKey,
 +                    "--cert", wskprops.cert, "--key", wskprops.key, "--namespace", namespace), env = env).stdout
```

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
ningyougang: * This test case just verify the unvalid certificate(a not exist certificate), it will throws exception, so it is necessay
* i think we can use it for `wsk namespace list`, actully, except `wsk namespace list`, `wsk action list` and others can also use `--cert/--key`, but for test case, `wsk namespace list` using `--cert/--key` is enough

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
ningyougang: * For above, it will send Http request to `wskprops.apihost`.
* For blow set (it should "set multiple property values with single command"), it will send http request to `openwhisk.ng.bluemix.net`, so for here, i think, can move --cert/key, because `openwhisk.ng.bluemix.net` env doesn't support cert/key.
* For below unset ( it should "validate default property values"), just remove --cert/key

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
ningyougang: yes, it is a positive test case using the certificate if "whisk.ssl.client.verification" is `on`

PR #2429, opened 2017-06-27
adding param for install catalog

beemarie:
Update parameter for this issue:
https://github.com/apache/incubator-openwhisk-catalog/issues/224

### DIFF
--- a/ansible/tasks/installOpenwhiskCatalog.yml
+++ b/ansible/tasks/installOpenwhiskCatalog.yml
@@ -4,7 +4,7 @@
 - set_fact:
     catalog_location={{ item.value.location }}
     catalog_repo_url={{ item.value.url }}
-    edge_host={{ groups['edge'] | first }}
+    api_host={{ groups['edge'] | first }}
     whisk_api_host={{ whisk_api_host_name | default(groups['edge'] | first) }}
     version="HEAD"
     repo_update="yes"
@@ -30,6 +30,6 @@
     version: "{{ version }}"
 
 - name: install the catalog from the catalog location
-  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ edge_host }} {{ catalog_namespace }} {{ cli.path }} {{ whisk_api_host }} chdir="{{ catalog_location }}/packages"
+  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ catalog_namespace }} {{ cli.path }} {{ whisk_api_host }} chdir="{{ catalog_location }}/packages"
   environment:
     OPENWHISK_HOME: "{{ openwhisk_home }}"
--- a/ansible/tasks/installOpenwhiskCatalog.yml
+++ b/ansible/tasks/installOpenwhiskCatalog.yml
@@ -5,6 +5,7 @@
     catalog_location={{ item.value.location }}
     catalog_repo_url={{ item.value.url }}
     api_host={{ groups['edge'] | first }}
+    whisk_host_name={{ whisk_api_host_name | default(groups['edge'] | first) }}
     version="HEAD"
     repo_update="yes"
 
@@ -29,6 +30,6 @@
     version: "{{ version }}"
 
 - name: install the catalog from the catalog location
-  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ catalog_namespace }} {{ cli.path }} chdir="{{ catalog_location }}/packages"
+  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ catalog_namespace }} {{ cli.path }} chdir="{{ catalog_location }}/packages" {{ whisk_host_name }}
   environment:
     OPENWHISK_HOME: "{{ openwhisk_home }}"
--- a/ansible/tasks/installOpenwhiskCatalog.yml
+++ b/ansible/tasks/installOpenwhiskCatalog.yml
@@ -5,7 +5,7 @@
     catalog_location={{ item.value.location }}
     catalog_repo_url={{ item.value.url }}
     api_host={{ groups['edge'] | first }}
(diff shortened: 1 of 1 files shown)


beemarie:
PG2 1748

ansible/tasks/installOpenwhiskCatalog.yml
```diff
@@ -5,6 +5,7 @@
     catalog_location={{ item.value.location }}
     catalog_repo_url={{ item.value.url }}
     api_host={{ groups['edge'] | first }}
```
jasonpet: lets change the name of api_host to edge_host

rabbah: why? Edge and API host are not the same.

ansible/tasks/installOpenwhiskCatalog.yml
```diff
@@ -29,6 +30,6 @@
     version: "{{ version }}"
 
 - name: install the catalog from the catalog location
-  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ catalog_namespace }} {{ cli.path }} chdir="{{ catalog_location }}/packages"
+  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ catalog_namespace }} {{ cli.path }} chdir="{{ catalog_location }}/packages" {{ whisk_host_name }}
   environment:
```
jasonpet: I would move {{ whisk_host_name }} after {{ clid.path }} and before chdir

ansible/tasks/installOpenwhiskCatalog.yml
jasonpet: I know.  that is why api_host should be edge_host and since it is referring to edge {{ groups['edge'] | first }}

jasonpet: we are adding a new one called whisk_host_name for the true api_host

rabbah: Will that work when there is a router and hence API host is the intended target not edge host? E.g. Blue green scenario.

rabbah: Ok.

ansible/tasks/installOpenwhiskCatalog.yml
```diff
@@ -5,6 +5,7 @@
     catalog_location={{ item.value.location }}
     catalog_repo_url={{ item.value.url }}
     api_host={{ groups['edge'] | first }}
+    whisk_host_name={{ whisk_api_host_name | default(groups['edge'] | first) }}
     version="HEAD"
```
jasonpet: I would call this api_host or whisk_api_host instead of whisk_home_name

ansible/tasks/installOpenwhiskCatalog.yml
```diff
@@ -5,6 +5,7 @@
     catalog_location={{ item.value.location }}
     catalog_repo_url={{ item.value.url }}
     api_host={{ groups['edge'] | first }}
+    whisk_api_host={{ whisk_api_host_name | groups['edge'] | first }}
```
rabbah: why do you need the host name and api host?

csantanapr: let's not create a new variable, we already have 3 similar ones `whisk_api_host_name` and `whisk_api_local_host_name` and `api_host`

csantanapr: remove this line and just the value in the `installCatalog.sh`

csantanapr: @rabbah we need both, `api_host` is what we use at installing of the action, and `whisk_api_host_name` will be hostname for the callback for github webhook to call back to whisk. this will help fix https://github.com/apache/incubator-openwhisk-catalog/pull/230

ansible/tasks/installOpenwhiskCatalog.yml
```diff
@@ -29,6 +30,6 @@
     version: "{{ version }}"
 
 - name: install the catalog from the catalog location
-  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ catalog_namespace }} {{ cli.path }} chdir="{{ catalog_location }}/packages"
+  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ catalog_namespace }} {{ cli.path }} {{ whisk_api_host }} chdir="{{ catalog_location }}/packages"
```
csantanapr: instead use `whisk_api_host_name` and not create a new one, even if it's local to this file. change would be to:
```
shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ catalog_namespace }} {{ cli.path }} {{ whisk_api_host_name | groups['edge'] | first }} chdir="{{ catalog_location }}/packages"
```

bjustin-ibm: @csantanapr what is the harm in creating a new variable to encapsulate exactly what you have here inline? If there is some harm caused, then we should get rid of all the variables, shouldn't we? (I'm not actually suggesting this, but just trying to get at the root of your reasoning)

I would suppose that the benefit of creating the variable is that it makes this (already rather long line) easier to read and understand. To justify putting the value inline would require that any harm caused by creating the variable should be greater than the benefit of increased readability.

ansible/tasks/installOpenwhiskCatalog.yml
bjustin-ibm: I'm not familiar with `whisk_api_local_host_name` but it seems like perhaps we should use that for both installing the actions and for URL calculation. This would simplify things a bit, so long as it is always correct to use it for those purposes. @rabbah thoughts?

rabbah:
@csantanapr @jasonpet is this pr still valid?

csantanapr:
Not needed anymore we found a different solution using environment variable from the action to find API host for github to callback whisk for webhooks

PR #2434, opened 2017-06-28
Deterministic container removal for test.

markusthoemmes:
Apparently, having `Instant.now` both times resulted in undeterministic behavior of the test.


sven-lange-last:
Looks good to me. With the change, `containers(0)` now has `lastUsed = Instant.EPOCH` which is earlier than `containers(1)` with `lastUsed = Instant.now`. For this reason, the LRU strategy of `ContainerPool.remove()` will always select `containers(0)` for removal.

PR #2435, opened 2017-06-28
Ensure Action Update Creates a Web Action (Review)

dubeejw:
Closes https://github.com/apache/incubator-openwhisk/issues/2428

### DIFF
--- a/tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
+++ b/tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
@@ -514,15 +514,50 @@ class WskBasicUsageTests
         (wp, assetHelper) =>
             val name = "webaction"
             val file = Some(TestUtils.getTestActionFilename("echo.js"))
-            val key = "someKey"
-            val value = JsString("someValue")
-            val annots = Map(key -> value)
+            val origKey = "origKey"
+            val origValue = JsString("origValue")
+            val updatedKey = "updatedKey"
+            val updatedValue = JsString("updateValue")
+            val origAnnots = Map(origKey -> origValue)
+            val updatedAnnots = Map(updatedKey -> updatedValue)
 
             assetHelper.withCleaner(wsk.action, name) {
-                (action, _) => action.create(name, file, annotations = annots)
+                (action, _) => action.create(name, file, annotations = origAnnots)
             }
 
-            wsk.action.create(name, file, web = Some("true"), update = true)
+            wsk.action.create(name, file, web = Some("true"), update = true, annotations = updatedAnnots)
+
+            val stdout = wsk.action.get(name, fieldFilter = Some("annotations")).stdout
+            assert(stdout.startsWith(s"ok: got action $name, displaying field annotations\n"))
+            removeCLIHeader(stdout).parseJson shouldBe JsArray(
+                JsObject(
+                    "key" -> JsString("web-export"),
+                    "value" -> JsBoolean(true)),
+                JsObject(
+                    "key" -> JsString(origKey),
+                    "value" -> origValue),
+                JsObject(
+                    "key" -> JsString("raw-http"),
+                    "value" -> JsBoolean(false)),
+                JsObject(
+                    "key" -> JsString("final"),
+                    "value" -> JsBoolean(true)),
+                JsObject(
+                    "key" -> JsString(updatedKey),
+                    "value" -> updatedValue),
+                JsObject(
+                    "key" -> JsString("exec"),
+                    "value" -> JsString("nodejs:6")))
+    }
+
+    it should "ensure action update creates an action with --web flag" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+            val name = "webaction"
+            val file = Some(TestUtils.getTestActionFilename("echo.js"))
+
+            assetHelper.withCleaner(wsk.action, name) {
+                (action, _) => action.create(name, file, web = Some("true"), update = true)
+            }
 
             val stdout = wsk.action.get(name, fieldFilter = Some("annotations")).stdout
             assert(stdout.startsWith(s"ok: got action $name, displaying field annotations\n"))
@@ -536,9 +571,6 @@ class WskBasicUsageTests
                 JsObject(
                     "key" -> JsString("final"),
                     "value" -> JsBoolean(true)),
-                JsObject(
-                    "key" -> JsString(key),
-                    "value" -> value),
                 JsObject(
                     "key" -> JsString("exec"),
                     "value" -> JsString("nodejs:6")))
--- a/tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
+++ b/tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
@@ -510,26 +510,29 @@ class WskBasicUsageTests
                 }
     }
 
-    it should "ensure --web flag does not remove existing annotations" in withAssetCleaner(wskprops) {
+    it should "ensure action update with --web flag only copies existing annotations when new annotations are not provided" in withAssetCleaner(wskprops) {
         (wp, assetHelper) =>
             val name = "webaction"
             val file = Some(TestUtils.getTestActionFilename("echo.js"))
+            val createKey = "createKey"
+            val createValue = JsString("createValue")
+            val updateKey = "updateKey"
+            val updateValue = JsString("updateValue")
             val origKey = "origKey"
             val origValue = JsString("origValue")
-            val updatedKey = "updatedKey"
-            val updatedValue = JsString("updateValue")
-            val origAnnots = Map(origKey -> origValue)
-            val updatedAnnots = Map(updatedKey -> updatedValue)
+            val overwrittenValue = JsString("overwrittenValue")
+            val createAnnots = Map(createKey -> createValue, origKey -> origValue)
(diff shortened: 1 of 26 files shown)


dubeejw:
@mdeuser, please review.

dubeejw:
PG3 791.

tools/cli/go-whisk-cli/commands/action.go
```diff
@@ -505,16 +505,20 @@ func webActionAnnotations(
...
             if action, _, err = client.Actions.Get(entityName); err != nil {
-                return nil, actionGetError(entityName, err)
+                whiskErr, isWhiskError := err.(*whisk.WskError)
+
+                if (isWhiskError && whiskErr.ExitCode != whisk.EXIT_CODE_NOT_FOUND) || !isWhiskError {
+                    return nil, actionGetError(entityName, err)
+                }
             } else {
-                annotations = webActionAnnotationMethod(action.Annotations)
+                annotations = append(annotations, action.Annotations...)
```
mdeuser: If `annotations` and `action.Annotations` each define the same annotation, I think the returned `annotations` slice will have that annotation defined twice, with the existing annotation value appearing ordinally after the updated value.  If so, will downstream consumers select the first or last annotation value?

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -652,14 +655,43 @@ class WskBasicUsageTests
...
+                    "value" -> JsString("nodejs:6")))
+    }
+
+    it should "ensure action update creates an action with --web flag" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+            val name = "webaction"
+            val file = Some(TestUtils.getTestActionFilename("echo.js"))
+
+            assetHelper.withCleaner(wsk.action, name) {
+                (action, _) => action.create(name, file, web = Some("true"), update = true)
```
mdeuser: To validate handling of existing action annotations, should the action be created first and then updated?

dubeejw:
PG approved.

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
dubeejw: @mdeuser, the test above this one does what you mentioned.

tools/cli/go-whisk-cli/commands/action.go
dubeejw: @mdeuser, given the situation when `annotations` and `action.Annotations` have a duplicate key, the value passed via the command-line with take precedence. That means the duplicate annotation already associated with the action will be overwritten with the new value.

dubeejw:
@mdeuser, good catch. I made an update that should ensure that existing annotations are overwritten if new annotations with matching keys are provided. Modified the test accordingly.

tools/cli/go-whisk-cli/commands/util.go
```diff
@@ -473,6 +473,26 @@ func addKeyValue(key string, value interface{}, keyValueArr whisk.KeyValueArr) (
...
+            return i
+        }
+    }
+
+    return -1
+}
+
+func appendKeyValueArr(origKeyValueArr whisk.KeyValueArr, appKeyValueArr whisk.KeyValueArr) (whisk.KeyValueArr) {
+    for i := 0; i < len(appKeyValueArr); i++ {
+        if findKeyValue(origKeyValueArr, appKeyValueArr[i].Key) == -1 {
```
mdeuser: @dubeejw - if i'm reading this correctly, it seems that updated annotation values won't override the existing one..?  it does add new annotation values; but i'm wondering about the handling of duplicates.

rabbah: i didn't follow all the changes closely - we agree the semantics we want are:

```
if (new web annotation)
   (prev annotations || []).remove(web annotation).add(new web annotation)
```

mdeuser: Any other new/updated annotation should also be handled, which I think occurs earlier in parseAction().  @dubeejw - it might be clearer if the annotation specific processing functions (i.e removal, update, etc) were added as methods to the `Annotations` type

dubeejw: @mdeuser, from the function call in action.go, the first argument is the new annotations to be added while the second argument passed is the existing annotations for a fetched action. Given that order of arguments, existing annotations will be overwritten.

dubeejw: To be sure, the test entitled `ensure --web flag preserves existing annotations and overwrites annotations that have matching keys` overwrites an existing annotation.

dubeejw:
@rabbah, recent a change went in that flipped the update flag to `true` for action updates (https://github.com/apache/incubator-openwhisk/commit/9c1f77fdb1e4a7339634434dc90c45026d6a43d2#diff-41c5935cb68718d73fa6ef611c927cc4R100). This enabled the ability to get action annotations that already exist and append new annotations to that list when updating a web action. However, preserving annotations that already exist is not a feature that is synced across commands (package update, trigger update, etc...). As per a conversation with @mdeuser, we are thinking about flipping the update flag back to `flase` so that the same update functionality is used across commands.

dubeejw:
@mdeuser, @rabbah, the alternative here would be to always pass in a value of `false` into this line: https://github.com/apache/incubator-openwhisk/blob/master/tools/cli/go-whisk-cli/commands/action.go#L402.

rabbah:
@dubeejw i defer to you and mark as the go CLI experts.

csantanapr:
bump
@akrabat hit this again in the field.
how can we move this one forward

dubeejw:
What is here works and will fix the issue. Problem is that the action update command with the web flag specified will behave differently than a normal action update as it keeps all existing annotations that action has.

dubeejw:
With the implementation from this PR action1 will have annotations for key1, key2 and the web annotations when the following commands are executed.
`wsk action create action1 --annotation key1 value1`
`wsk action update action1 --web true --annotation key2 value2`

dubeejw:
Perhaps I should change the implementation to only preserve existing annotations only when the `--web` flag is specified and no addition annotations are provided. 

That way the following would only have the web and key2 annotation.
`wsk action create action1 --annotation key1 value1`
`wsk action update action1 --web true --annotation key2 value2`

That way the following would have the web and key1 annotation.
`wsk action create action1 --annotation key1 value1`
`wsk action update action1 --web true`

dubeejw:
@mdeuser, @rabbah, @csantanapr, thoughts on the above?

csantanapr:
@dubeejw yes I think this is becoming to fix.
Let's implement like you suggested in https://github.com/apache/incubator-openwhisk/pull/2435#issuecomment-320044929

if `--web` present with other annotations being set in cmd line, only those annotations would be set. it would work like the other entity annotations.
if `--web` present and no annotations in cmd, then annotations are not touch only the export-web

this will unblock the bug, which I lot of people are using `action update` to provision

mdeuser:
imho, i worry a bit about the semantics of "update" changing just because a particular command flag is used.
> https://github.com/apache/incubator-openwhisk/pull/2435#issuecomment-320044929

csantanapr:
@mdeuser I think what @dubeejw proposed and I understood was to leave the semantics the same as today.

dubeejw:
PG2 1909 ⏳

dubeejw:
@mdeuser, take another look. Made some updates to keep the current semantics in place.

akrabat:
This is interesting as `--web` is implemented as a set of annotations under-the-hood, but doesn't look like one to the user. It would be decidedly non-intuitive for `wsk update action test test.js --web true` to remove all other annotations as `--web true` doesn't _look_ like you're dealing annotations.

What I would expect to happen:

1 `wsk update action test test.js --web true` : all current annotations are not touched
2 `wsk update action test test.js --web true -a foo bar` : all other annotations other than foo are removed

From my point of view, `--web` doesn't look like `-a` and so should not touch anything added with `-a`. My understanding of the conversation above is that this is what @dubeejw has implemented in this PR.

csantanapr:
That's my understanding Rob and my desire also. 
Can you confirm @dubeejw ?

dubeejw:
@akrabat, @csantanapr, current code in the PR to has the described functionality. Just needs a final review. Feel free to test it out.

mdeuser:
Probably not what this PR was wanting to remedy, but since this came up....

>wsk update action test test.js --web true -a foo bar : all other annotations other than foo are removed
That behavior may change in the future to "only the annotation foo is updated/created; all other annotations are left as-is"

see also https://github.com/apache/incubator-openwhisk/issues/76

dubeejw:
@mdeuser, any review comments?

tools/cli/go-whisk-cli/commands/util.go
mdeuser: This seems like a good function candidate to the whisk.KeyValueArr structure:
```
func (kvr KeyValueArr) AppendKeyValueArr(kvrToAdd KeyValueArr) KeyValueArr {
...
}
```
Minor nit... I suggest renaming the arg/var `fetch` to reflect its higher purpose of "preserve all existing annotations"

dubeejw:
PG3 975. ⏳

dubeejw:
PG3 976 🔵 .

dubeejw:
@mdeuser, anything else on this one?

dubeejw:
Rebased for API experimental removal conflicts.

PR #2436, opened 2017-06-28
API GW V2 updates

### LINKED ISSUE #2318: API management web actions incorrectly assume __ow_user will always be set
When an invalid auth key is used, the `wsk api` commands will return the following error message:
```error: Unable to obtain the API list: __ow_user is required.```

1. A __ow_user value is not mandatory.
2. The message is not clear and incorrect.

### PR DESCRIPTION
**Pull request by mdeuser**
- Do not enforce need for gwUrl parameter in api gw actions
- Enable http redirect forwarding

Fixes #2318

### DIFF
--- a/core/routemgmt/getApi/getApi.js
+++ b/core/routemgmt/getApi/getApi.js
@@ -15,33 +15,33 @@
  * limitations under the License.
  */
 
- /**
-  * Retrieve API configuration from the API Gateway:
-  *
-  * Parameters (all as fields in the message JSON object)
-  *   gwUrlV2              Required when accesstoken is provided. The V2 API Gateway base path (i.e. http://gw.com)
-  *   gwUrl                Required. The API Gateway base path (i.e. http://gw.com)
-  *   gwUser               Optional. The API Gateway authentication
-  *   gwPwd                Optional. The API Gateway authentication
-  *   namespace            Optional. Input value is now overwritten by __ow_user.  Namespace of API author
-  *   __ow_user            Required. Namespace of API author
-  *   tenantInstance       Optional. Instance identifier used when creating the specific API GW Tenant
-  *   accesstoken          Optional. Dynamic API GW auth.  Overrides gwUser/gwPwd
-  *   spaceguid            Optional. Namespace unique id.
-  *   basepath             Optional. Base path or API name of the API.
-  *                                  If not provided, all APIs for the namespace are returned
-  *   relpath              Optional. Must be defined with 'operation'.  Filters API result to path/operation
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by mdeuser** [comment]
pg4/450

**Comment by mdeuser** [comment]
@bjustin-ibm - ready for review/merge please.  good travis/pg

**Comment by bjustin-ibm** [comment]
@mdeuser is there a test that exercises the changed redirect behavior? Also, is there someone else who would be able to more thoroughly review this? Perhaps, @dubeejw?

**Comment by mdeuser** [comment]
@bjustin-ibm - no, there's no explicit redirection test - kind of tricky and i guess i'm trusting the request package to handle that setting per it's docs.  all existing tests pass as expected.  there's a separate environment with an api gw hosted in a domain that redirects to another domain - my plan is to manually test this change in that environment to confirm the desired behavior.

**Comment by mdeuser** [comment]
pg4/488

**Comment by mdeuser** [comment]
@dubeejw - would you please review.  thanks!

File: core/routemgmt/getApi/getApi.js [hunk 1]
```diff
@@ -154,12 +154,12 @@ function validateArgs(message) {
     return 'Internal error. A message parameter was not supplied.';
   }
 
-  if (!message.__ow_user) {
```
dubeejw: Looks like the `validateArgs` method is duplicated in two files.

tools/cli/go-whisk-cli/commands/action.go
```diff
@@ -857,7 +857,10 @@ func isWebAction(client *whisk.Client, qname QualifiedName) error {
     if err != nil {
         whisk.Debug(whisk.DbgError, "client.Actions.Get(%s) error: %s\n", fullActionName, err)
         whisk.Debug(whisk.DbgError, "Unable to obtain action '%s' for web action validation\n", fullActionName)
-        err = errors.New(wski18n.T("API action '{{.name}}' does not exist", map[string]interface{}{"name": fullActionName}))
+        errMsg := wski18n.T("Unable to get action '{{.name}}': {{.err}}",
```
dubeejw: Need to remove reference to `API action` on line 865. Can probably remove those strings from the externalized strings JSON file.

mdeuser: Good one.  I'll update both the strings file and this module

core/routemgmt/getApi/getApi.js
mdeuser: Right.  Each action has it's own validateArgs function and all are slightly different.

mdeuser:
pg4/512

PR #2437, opened 2017-06-28
Update install Catalog script.

DanLavine:
* Removes unused parameter

There is a PR for the OpenWhiskCatalog repo to remove an unused parameter from the `installCatalog.sh` script. Once this [#pr](https://github.com/apache/incubator-openwhisk-catalog/pull/231) has been merged, this PR can be merged.

### DIFF
--- a/ansible/tasks/installOpenwhiskCatalog.yml
+++ b/ansible/tasks/installOpenwhiskCatalog.yml
@@ -29,6 +29,6 @@
     version: "{{ version }}"
 
 - name: install the catalog from the catalog location
-  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ catalog_namespace }} {{ cli.path }} chdir="{{ catalog_location }}/packages"
+  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ cli.path }} chdir="{{ catalog_location }}/packages"
   environment:
     OPENWHISK_HOME: "{{ openwhisk_home }}"


DanLavine:
Not sure why this PR failed. Ran it a number of times before I squished all my PRs before submitting this. Will need to update travis so that if it fails I grab all of the component logs now.

jasonpet:
what makes you think this is unused?  This variable is defined in the global group_vars/all file

jasonpet:
ok, maybe it is unused.  I see that Rodric had removed it.

jasonpet:
I see you updated validateParameter.sh in the openwhisk-catalog based on my previous comment.  However, you will also need to update the install script (installCatalog.sh) file in the pushnotification repo (https://github.com/apache/incubator-openwhisk-package-pushnotifications) to remove the namespace parameter before this update can go in.  The updates to all three of these repos really needed to go in at the same time to avoid breaks.  Right now the catalog install from openwhisk-catalog is broken due to https://github.com/apache/incubator-openwhisk-catalog/pull/231 getting in and this PR not being in yet.

jasonpet:
@DanLavine, @daisy-ycguo  - please fix this situation (see comment above).  You need to update https://github.com/apache/incubator-openwhisk-package-pushnotifications/blob/master/packages/installCatalog.sh to remove the namespace parameter if you are removing it from here just as you already did for the openwhisk-catalog repo.  All three of these needed to go in together.

DanLavine:
@jasonpet I don't see what is blocking this PR from being merged as it fixes the install catalog script appropriately. 

I had no idea that the other repo also relied on the incubator-openwhisk-catalog repo dependency as I was able to standup OpenWhisk with just these 2 PRs. I can fix it as well.

DanLavine:
created a PR to address the second repo [here](https://github.com/apache/incubator-openwhisk-package-pushnotifications/pull/80).

bjustin-ibm:
🏃 PG2 1788

bjustin-ibm:
Resolves #2468

PR #2441, opened 2017-06-29
Make kafka heap size configurable.

markusthoemmes:
Kafka is set to 1 gigabyte by default which might not be suitable for all deployments.


markusthoemmes:
PG1 1732 🔵

PR #2446, opened 2017-06-29
Fix install command in Docker for Mac bash script

zachschultz:
There is a missing `=` which causes an error when trying to install Docker 2.2.1


zachschultz:
My ICLA has been acknowledged by Apache, is that what you're referring to?

PR #2450, opened 2017-06-30
Adjust kafka's maximum poll period.

markusthoemmes:
Kafka allows setting the maximum period between two polls. As we allow for 5 minutes of action execution, in a fully loaded invoker it might take 5 minutes for the consumer to poll again. To prevent rebalances this value is adjusted accordingly.

Bumping Kafka's version is needed because that value is exposed only in 0.10.1+.

ConsumerGroups are removed (or rather: each Invoker/Controller is in it's own ConsumerGroup) to prevent rebalances to impact other consumers. OpenWhisk does not use ConsumerGroup mechanics anyway.

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -120,7 +120,7 @@ registry:
   confdir: "{{ config_root_dir }}/registry"
 
 kafka:
-  version: 0.10.0.1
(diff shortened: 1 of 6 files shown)


markusthoemmes:
PG1 1749 🔵

PR #2454, opened 2017-07-03
Deal with jvm process leak in TestUtils

style95:
This pr adds line to shutdown executor in tests code to prevent jvm process leak.

Reference: https://blog.tier1app.com/2014/09/30/memory-leak-in-java-executor/

### DIFF
--- a/tests/src/test/scala/common/TestUtils.java
+++ b/tests/src/test/scala/common/TestUtils.java
@@ -17,6 +17,7 @@
 
 package common;
 
+import static common.TestUtils.RunResult.executor;
 import static org.junit.Assert.assertTrue;
 
 import java.io.BufferedReader;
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/common/TestUtils.java
```diff
@@ -327,6 +327,7 @@ public String call() throws IOException {
             logger.info("RunResult: " + rr);
         }
         rr.validateExitCode(expectedExitCode);
+        executor.shutdown();
```
rabbah: I'm tempted to suggested making the executor a static instance. What do you think?

style95: Ok I will update it accordingly.

style95:
Now `ExecutorService` is static field in the `TestUtils` class.
Since all threads in the pool will be shared among tests, there is no explicit shutdown.
Once test is over, it will shutdown as well.

rabbah:
Thanks for fixing the leak!

PR #2455, opened 2017-07-03
Consider mac environment along with docker-machine enviornment

style95:
This pr makes Ansible consider `mac` environment also while downloading cli binaries.

### DIFF
--- a/ansible/roles/cli/tasks/download_cli.yml
+++ b/ansible/roles/cli/tasks/download_cli.yml
@@ -10,7 +10,7 @@
     mode=0755
     validate_certs=False
     force=True
-  when: "'environments/docker-machine' not in inventory_dir"
+  when: "'environments/docker-machine' not in inventory_dir and 'environments/mac' not in inventory_dir"
 
 - name: "download cli (docker-machine) to openwhisk home at {{ openwhisk_home }}"
   local_action: >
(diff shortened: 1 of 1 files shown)


ansible/roles/cli/tasks/download_cli.yml
```diff
@@ -20,4 +20,4 @@
     mode=0755
     validate_certs=False
     force=True
-  when: "'environments/docker-machine' in inventory_dir"
+  when: "'environments/docker-machine' and 'environments/mac' in inventory_dir"
```
houshengbo: I think environments/docker-machine' and 'environments/mac should be "or" instead of and.

ansible/roles/cli/tasks/download_cli.yml
```diff
@@ -10,7 +10,7 @@
     mode=0755
     validate_certs=False
     force=True
-  when: "'environments/docker-machine' not in inventory_dir"
+  when: "'environments/docker-machine' and 'environments/mac' not in inventory_dir"
```
houshengbo: I think this line should be
  when: "'environments/docker-machine' not in inventory_dir and 'environments/mac' not in inventory_dir".

style95:
Updated accordingly

rabbah:
@houshengbo looks correct now. Can you verify and merge if you agree?

Issue #2389, opened 2017-06-15
JVM heap size is not set correctly for Java actions


rabbah:
Java actions run with no heap settings, regardless of the container's memory limit: https://github.com/apache/incubator-openwhisk/blob/master/core/javaAction/Dockerfile#L38

This will cause a reused Java action to eventually be killed because the GC may not get a chance to run.

```bash
>  docker run -m 64M -it javaaction bash
# java -XX:+PrintFlagsFinal -version 2>&1 | grep MaxHeapSize
    uintx MaxHeapSize                              := 1035993088                          {product}
```

That's ~1GB even though memory limit for container is 64M.

jsanda:
Is something like [this|https://github.com/fabric8io-images/java/tree/master/images/jboss/openjdk8/jdk] along the lines of what you have in mind? run-java.sh is the default command for the image. Among other things it sets the JVM heap size based on container limits.

rabbah:
Issue moved to [apache/incubator-openwhisk-runtime-java #5](https://github.com/apache/incubator-openwhisk-runtime-java/issues/5) via [**ZenHub**](https://www.zenhub.com/)

PR #2457, opened 2017-07-05
Correct java action heap size issue #2389.

lzbj:
limit the heap size to 56m, leave some space for stacks etc.

### DIFF
--- a/core/javaAction/Dockerfile
+++ b/core/javaAction/Dockerfile
@@ -35,4 +35,4 @@ RUN cd /javaAction; ./gradlew oneJar
 
 RUN rm -rf /javaAction/src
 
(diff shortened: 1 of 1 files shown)


rabbah:
We have a defect for the heap of java actions https://github.com/apache/incubator-openwhisk/issues/2389 I don't think this fix helps in general and could make things worse relative to actions settings. What do you think?

lzbj:
My fault, I misunderstood the bug meaning. Yes, correct, the best solution perhaps would be a more flexible container/warmup memory strategy. For example, I can think of some user could pay a premium openwhisk service, so their actions could consume more than 512m limit size. But maybe we could set the jvm heap Xmx a little less than 512m as a temp workaround.

rabbah:
Going to close this I don't think it's correct. Reopen if you disagree.

Issue #1155, opened 2016-08-29
Inconsistent entity name quoting in CLI messages


mdeuser:
Some messages surround the entity name with single quotes; others do not.  Quoting entity names allows for easier readability of entity names with embedded white space.

PR #1767, opened 2017-01-26
Fix bugs related to zero components in sequence

ioana-blue:
Don't allow creation of sequences with no components (+ test). Guard size calculation just in case.

PR #826, opened 2016-07-01
remove ant from docs

csantanapr:
@bjustin-ibm please merge

PR #2462, opened 2017-07-06
WIP Fix for 1155

jessealva:
Still working on getting the tests passing.


jessealva:
@dubeejw please review for comments so far

jessealva:
Fixes #1155

jessealva:
PG 3. #826

jessealva:
PG1, #1767  for after a merge with master that contained changes i had to merge manually.

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -26,18 +26,11 @@ import scala.language.postfixOps
...
 import scala.concurrent.duration.DurationInt
 import scala.util.Random
-
 import org.apache.commons.io.FileUtils
 import org.junit.runner.RunWith
 import org.scalatest.junit.JUnitRunner
-
-import common.TestHelpers
-import common.TestUtils
+import common._
```
dubeejw: Wouldn't import everything from common unless we actually use everything in that package.

tools/cli/go-whisk-cli/wski18n/resources/en_US.all.json
```diff
@@ -829,7 +829,7 @@
   },
   {
     "id": "{{.ok}} deleted action {{.name}}\n",
-    "translation": "{{.ok}} deleted action {{.name}}\n"
+    "translation": "{{.ok}} deleted action '{{.name}}'\n"
```
dubeejw: I don't think we need to quote entity names that are displayed in bold font. Here is what it looks like:
ok: created action **'actionName'**

@mdeuser, what do you think?

dubeejw: Also looks like the keys (id) will be out of sync with the values (translation).

mdeuser: Agreed.  Any existing emphasis should be sufficient.

jessealva: So I agree that the keys will be out of sync, but I think its weird and wrong they have the translation string as the key to begin with. Thats unnecessary and confusing in my opinion, just didnt want to go up and change them all to some other string which is more easily searchable. 

At the same time, boldness does not show on tools like Jenkins, travis, etc.  Only on the CLI, so there could still be a reason to quote them @mdeuser @dubeejw   

Although if you guys still think no, i can remove the quoting from the bolded strings.

jessealva:
PG 1, 1932.  @dubeejw @mdeuser

rabbah:
is this done @dubeejw @mdeuser? 
we'd need to resolve conflicts to merge it.

mdeuser:
@jessealva - can you rebase and issue another pg.  out of the changes to 10 output strings, only 1 test was impacted...might indicate a need for an additional test or two (separate issue though)

jessealva:
When rebasing, had to fix about 20 different failures across the 5 files.  Think most of these were covered.

jessealva:
@mdeuser PG1, 2069. Looks clean now

jessealva:
ugh, not clean, looking into this again

jessealva:
PG2, 2095 @mdeuser

dubeejw:
PG2 2096 ⏳

PR #2463, opened 2017-07-07
Correct unable to set apihost issue.

### LINKED ISSUE #2411: Unable to set the api host
## Environment details:

* Ubuntu 16.04 
* Docker version 1.12.0, build 8eab29e

1.   try setting apihost like below
2.   ./wsk property set --apihost 10.xxx.11.xx:5009
3.   since the api host is having port no it will throw error 

```
ok: whisk API host set to 10.xxx.11.xx:5009
```

```
runtime error: invalid memory address or nil pointer dereference
Application exited unexpectedly
```

* i have distributed OW setup with    
* i am using HAProxy as LB. 
* i wanted to add LB ip and port as api host

lzbj:
line 864 url.Parse caused failure.
Fixes #2411.

### DIFF
--- a/tests/src/test/scala/whisk/core/cli/test/WskConfigTests.scala
+++ b/tests/src/test/scala/whisk/core/cli/test/WskConfigTests.scala
@@ -259,6 +259,27 @@ class WskConfigTests extends TestHelpers with WskTestHelpers {
     }
   }
 
+  it should "set api host with or without http prefix" in {
+    val tmpwskprops = File.createTempFile("wskprops", ".tmp")
+    try {
+      val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
+      Seq("", "http://", "https://").foreach { prefix =>
+        Seq("10", "10:123", "aaa", "aaa:123").foreach { host =>
+          val apihost = s"$prefix$host"
+          withClue(apihost) {
+            val rr = wsk.cli(Seq("property", "set", "--apihost", apihost), env = env)
+            rr.stdout.trim shouldBe s"ok: whisk API host set to $apihost"
+            rr.stderr shouldBe 'empty
+            val fileContent = FileUtils.readFileToString(tmpwskprops)
+            fileContent should include(s"APIHOST=$apihost")
+          }
+        }
+      }
+    } finally {
+      tmpwskprops.delete()
+    }
+  }
+
   it should "set auth in property file" in {
     val tmpwskprops = File.createTempFile("wskprops", ".tmp")
     val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
--- a/tools/cli/go-whisk-cli/commands/util.go
+++ b/tools/cli/go-whisk-cli/commands/util.go
@@ -818,6 +818,10 @@ func getURLBase(host string, path string) (*url.URL, error)  {
         return nil, whiskErr
     }
 
+    if !strings.HasPrefix(host, "http") {
+        host = "https://" + host
+    }
+
     urlBase := fmt.Sprintf("%s%s", host, path)
     url, err := url.Parse(urlBase)
 


dubeejw:
@lzbj, commented on the issue about port setting as well.

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -151,7 +151,7 @@ class WskBasicUsageTests
         val tmpwskprops = File.createTempFile("wskprops", ".tmp")
         try {
             val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
-            val apihost = s"http://${WhiskProperties.getBaseControllerAddress()}"
+            val apihost = s"${WhiskProperties.getBaseControllerAddress()}"
```
dubeejw: @lzbj, need a test for setting the API host with and without a protocol.

dubeejw: @lzbj, also looks like the test failed because the server is expecting an http request instead of an https request.

dubeejw: I believe the original error comes from the `url.Parse(urlBase)` function.

lzbj: Yes, correct, I'll update according to your comments.

tools/cli/go-whisk-cli/commands/util.go
```diff
@@ -856,6 +856,10 @@ func getURLBase(host string, path string) (*url.URL, error)  {
         return nil, whiskErr
     }
 
+    if !strings.HasPrefix(host, "http") {
+        host = "http://" + host
```
dubeejw: What if the OpenWhisk deployment only supports https?

dubeejw: I believe the error is thrown blow on line 864 `url.Parse`.

lzbj: correct, updated the code. If we can't avoid deployment using https, we can just add one test. do you have any suggestion we can add more tests?

rabbah:
The cli must support both - there's a test for http as well as https which is the nginx path.

lzbj:
Ok, thank you. Then now the test in WskBasicUsageTests is sufficient for the change, right?

rabbah:
@dubeejw @mdeuser thoughts?

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -148,7 +148,7 @@ class WskBasicUsageTests
         }
     }
 
-    it should "show api build using http apihost" in {
+    it should "show api build using http apihost with protocol specified" in {
```
mdeuser: @lzbj - is another successful test needed that provides the api host without the protocol being provided?

mdeuser: this is the area of code i was considering
```
+    if !strings.HasPrefix(host, "http") {
+        host = "https://" + host
+    }
```
so i was thinking an apihost setting with and without a http prefix should work.  will a protocol-less apihost fail?

lzbj: yes, this line "I believe the error is thrown blow on line 864 url.Parse." will fail as @dubeejw comments.

tools/cli/go-whisk-cli/commands/util.go
```diff
@@ -841,6 +841,10 @@ func getURLBase(host string, path string) (*url.URL, error)  {
         return nil, whiskErr
     }
 
+    if !strings.HasPrefix(host, "http") {
```
dubeejw: Hmm, what if the string begins with `https://`, won't `https://` be prepended again?

tools/cli/go-whisk-cli/commands/util.go
```diff
@@ -841,6 +841,10 @@ func getURLBase(host string, path string) (*url.URL, error)  {
...
+        host = "https://" + host
```
dubeejw: Need a test for this.

rabbah:
I don't know if this pr is still relevant but it doesn't fix this issue:

```
> wsk property set --apihost http://192.168.99.100
> wsk action create t echo.js --web true
ok: created action t
> wsk action get t --url
ok: got action t
https://http://192.168.99.100/api/v1/web/guest/default/t
```

rabbah:
@mdeuser @dubeejw @lzbj i pushed a commit which adds tests.

lzbj:
@rabbah , thanks.

tests/src/test/scala/whisk/core/cli/test/WskConfigTests.scala
```diff
@@ -259,6 +259,27 @@ class WskConfigTests extends TestHelpers with WskTestHelpers {
...
+  it should "set api host with or without http prefix" in {
+    val tmpwskprops = File.createTempFile("wskprops", ".tmp")
+    try {
+      val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
+      Seq("", "http://", "https://").foreach { prefix =>
+        Seq("10", "10:123", "aaa", "aaa:123").foreach { host =>
+          val apihost = s"$prefix$host"
+          withClue(apihost) {
+            val rr = wsk.cli(Seq("property", "set", "--apihost", apihost), env = env)
+            rr.stdout should not include regex("Application exited unexpectedly")
```
dubeejw: Errors are printed to `stderr`, so shouldn't need this check.

rabbah:
pg1/2055 🚀

rabbah:
pg is blue.

PR #2475, opened 2017-07-12
Use error string indirection instead of copying.

lzbj:
Try to fix the error message replicate issue, make some change to the MakeWskErrorFromWskError function which could save one copy.

### DIFF
--- a/tools/cli/go-whisk-cli/commands/action.go
+++ b/tools/cli/go-whisk-cli/commands/action.go
@@ -66,11 +66,11 @@ var actionCreateCmd = &cobra.Command{
         }
 
         if action, err = parseAction(cmd, args, false); err != nil {
-            return actionParseError(cmd, args, err)
+            return actionParseError(cmd, args, &err)
         }
 
         if _, _, err = client.Actions.Insert(action, false); err != nil {
-            return actionInsertError(action, err)
+            return actionInsertError(action, &err)
         }
 
         printActionCreated(action.Name)
@@ -99,11 +99,11 @@ var actionUpdateCmd = &cobra.Command{
         }
 
         if action, err = parseAction(cmd, args, true); err != nil {
-            return actionParseError(cmd, args, err)
+            return actionParseError(cmd, args, &err)
         }
 
         if _, _, err = client.Actions.Insert(action, true); err != nil {
-            return actionInsertError(action, err)
+            return actionInsertError(action, &err)
         }
 
         printActionUpdated(action.Name)
@@ -142,7 +142,7 @@ var actionInvokeCmd = &cobra.Command{
 
         if len(paramArgs) > 0 {
             if parameters, err = getJSONFromStrings(paramArgs, false); err != nil {
-                return getJSONFromStringsParamError(paramArgs, false, err)
+                return getJSONFromStringsParamError(paramArgs, false, &err)
             }
         }
         if flags.action.result {flags.common.blocking = true}
@@ -171,7 +171,7 @@ func handleInvocationResponse(
                 color.Output)
         } else {
             if !flags.common.blocking {
-                return handleInvocationError(err, qualifiedName.entityName, parameters)
+                return handleInvocationError(&err, qualifiedName.entityName, parameters)
             } else {
                 if isBlockingTimeout(err) {
                     printBlockingTimeoutMsg(
@@ -186,7 +186,7 @@ func handleInvocationResponse(
                         result,
                         colorable.NewColorableStderr())
                 } else {
-                    return handleInvocationError(err, qualifiedName.entityName, parameters)
+                    return handleInvocationError(&err, qualifiedName.entityName, parameters)
                 }
             }
         }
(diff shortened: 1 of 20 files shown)


rabbah:
I'm not sure I follow. Can you show me an example of how to save one of three message identifiers? 

@dubeejw I assume it's  possible to replace the actual message id with a more generic constant identifier. What do you suggest?

lzbj:
Ah, got it. You mean the I18n string replication. Yes, I think we can reference them as tags you mentioned with pointers.

What I changed is the error message passing, now the cost is better, sorry for my misunderstood.

dubeejw:
@mdeuser

rabbah:
I'm not too concerned about the runtime cost in this case - but the developer overhead.

lzbj:
Sure, updated a little. Define the error message as a const as cli convention, if you and everyone agree, I'll continue the remain ones.

tools/cli/go-whisk/whisk/action.go
```diff
@@ -153,7 +153,7 @@ func (s *ActionService) List(packageName string, options *ActionListOptions) ([]
     return actions, resp, err
 }
 
-func (s *ActionService) Insert(action *Action, overwrite bool) (*Action, *http.Response, error) {
+func (s *ActionService)         Insert(action *Action, overwrite bool) (*Action, *http.Response, error) {
```
mdeuser: is the added whitespace intentional?

tools/cli/go-whisk/whisk/wskerror.go
```diff
@@ -34,6 +34,7 @@ const NO_DISPLAY_PREFIX         bool = false
 const APPLICATION_ERR           bool = true
 const NO_APPLICATION_ERR        bool = false
 const TIMED_OUT                 bool = true
+const UNABLE_CREATE_ACTION_ERR  string = "Unable to create action '{{.name}}': {{.err}}"
```
mdeuser: this change does not seem related to the rest of the commit changes.  this is probably best moved into another pr focusing on using constants as labels for all pii strings

tools/cli/go-whisk-cli/commands/action.go
```diff
@@ -582,50 +582,50 @@ func getLimits(memorySet bool, logSizeSet bool, timeoutSet bool, memory int, log
...
 }
 
-func actionInsertError(action *whisk.Action, err error) (error) {
-    whisk.Debug(whisk.DbgError, "client.Actions.Insert(%#v, false) error: %s\n", action, err)
+func actionInsertError(action *whisk.Action, err *error) (error) {
+    whisk.Debug(whisk.DbgError, "client.Actions.Insert(%#v, false) error: %s\n", action, *err)
 
     errMsg := wski18n.T(
-        "Unable to create action '{{.name}}': {{.err}}",
+        whisk.UNABLE_CREATE_ACTION_ERR,
```
mdeuser: this change does not seem related to the rest of the commit changes.  this is probably best moved into another pr focusing on using constants as labels for all pii strings

mdeuser:
i suggest making the message const changes in another pr

tools/cli/go-whisk-cli/commands/action.go
rabbah: @mdeuser is the change of using pointers for the error string worth it?

tools/cli/go-whisk/whisk/wskerror.go
rabbah: @mdeuser this was the spirit of the intended change/this pr. I think @lzbj initially interpreted the issue differently. So we should decide: is it worth making the change of passing the string around by pointer? If so, then yes decouple the prs. If not, drop that commit and making the intended change for reducing the number of times an error message is used. For the latter, I defer to you and @dubeejw and @lzbj for the best way to do that.

tools/cli/go-whisk-cli/commands/action.go
mdeuser: as it stands, the conversion to pointer of a subset of error structs is a bit confusing.  in general, the go errors package returns the error struct when creating one, not the pointer to the error struct
https://golang.org/pkg/errors/

rabbah: should we drop this pr then, or repurpose it for the actually intended improvement?

tools/cli/go-whisk/whisk/action.go
lzbj: Np, unintentionally, updated.

lzbj:
@mdeuser, I suggest we keep this pr change, since as you see the MakeWskErrorFromWskError was designed to handle err pointers, the error was in essence is string. Nevertheless, It's up to you decide.

Yes I'll create another PR about the const change.
Anyway, thank you guys for help clarify and review. Appreciate it.

mdeuser:
@lzbj - in general, Go seems to treat errors as stucts and not pointers to structs.  https://golang.org/pkg/errors/  

since WskError is just another implementation of the error interface, i recommend consistency in the treatment of errors - not as pointers, but as structs

lzbj:
Ok, np, close this one. Thanks for your feedback.

PR #2476, opened 2017-07-12
Stabilize loadbalancer under on/off conditions

markusthoemmes:
If invokers are coming online/offline frequently today, the list of available invokers changes frequently which causes the hash-based picking algorithm to constantly change home invokers of actions, thus destroying warm-ratios of that action.

This attacks this issue by accepting that invokers are in fact index based and thus they are now managed in an array which's length will not change if an invoker goes offline. For partitioning cases i.e. blackbox containers) the length of the underlying sequence is kept constant by padding them repsectively.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
@@ -54,6 +54,8 @@ import whisk.core.entity.UUID
 import whisk.core.entity.WhiskAction
 import whisk.core.entity.types.EntityStore
 import scala.annotation.tailrec
+import whisk.core.entity.EntityName
+import whisk.core.entity.Identity
 
 trait LoadBalancer {
 
@@ -103,7 +105,7 @@ class LoadBalancerService(
 
     override def publish(action: ExecutableWhiskAction, msg: ActivationMessage)(
         implicit transid: TransactionId): Future[Future[Either[ActivationId, WhiskActivation]]] = {
-        chooseInvoker(action, msg).flatMap { invokerName =>
+        chooseInvoker(msg.user, action).flatMap { invokerName =>
             val entry = setupActivation(action, msg.activationId, msg.user.uuid, invokerName, transid)
             sendActivationToInvoker(messageProducer, msg, invokerName).map { _ =>
                 entry.promise.future
@@ -111,7 +113,10 @@ class LoadBalancerService(
         }
     }
 
-    def invokerHealth: Future[Map[String, InvokerState]] = invokerPool.ask(GetStatus)(Timeout(5.seconds)).mapTo[Map[String, InvokerState]]
+    /** An indexed sequence of all invokers in the current system */
+    def allInvokers: Future[IndexedSeq[(InstanceId, InvokerState)]] = invokerPool
+        .ask(GetStatus)(Timeout(5.seconds))
+        .mapTo[IndexedSeq[(InstanceId, InvokerState)]]
 
     /**
      * Tries to fill in the result slot (i.e., complete the promise) when a completion message arrives.
@@ -138,8 +143,7 @@ class LoadBalancerService(
     /**
      * Creates an activation entry and insert into various maps.
      */
-    private def setupActivation(action: ExecutableWhiskAction, activationId: ActivationId, namespaceId: UUID, invokerName: String, transid: TransactionId): ActivationEntry = {
-
+    private def setupActivation(action: ExecutableWhiskAction, activationId: ActivationId, namespaceId: UUID, invokerName: InstanceId, transid: TransactionId): ActivationEntry = {
         val timeout = action.limits.timeout.duration + activeAckTimeoutGrace
         // Install a timeout handler for the catastrophic case where an active ack is not received at all
         // (because say an invoker is down completely, or the connection to the message bus is disrupted) or when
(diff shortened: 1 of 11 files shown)


markusthoemmes:
PG2 1802 🔵

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -176,18 +179,17 @@ class LoadBalancerService(
...
     private def clearInvokerState(invokerName: String) = {
-        val actSet = activationByInvoker.getOrElseUpdate(invokerName, new TrieSet[ActivationEntry])
-        actSet.keySet map {
-            case actEntry @ ActivationEntry(activationId, namespaceId, invokerIndex, _, promise) =>
-                promise.tryFailure(new LoadBalancerException(s"Invoker $invokerIndex restarted"))
-                activationById.remove(activationId)
-                activationByNamespaceId.get(namespaceId) map { _.remove(actEntry) }
+        activationByInvoker.get(invokerName).foreach { actSet =>
+            actSet.keySet map {
+                case actEntry @ ActivationEntry(activationId, namespaceId, invokerIndex, _, promise) =>
```
rabbah: Is this active ack promise termination necessary? What if an invoker comes back online and the requests were in the queue anyway. The recovery mechanism on the active ack will clean up the active acks if it's really catastrophic but may take longer.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -176,18 +179,17 @@ class LoadBalancerService(
...
+                    promise.tryFailure(new LoadBalancerException(s"Invoker $invokerIndex offline"))
```
rabbah: Is this active ack promise termination necessary? What if an invoker comes back online and the requests were in the queue anyway. The recovery mechanism on the active ack will clean up the active acks if it's really catastrophic but may take longer

markusthoemmes: Fair point. I think we should adopt that and get rid of that code :+1:

markusthoemmes:
PG1 1802 🔵

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -209,13 +195,15 @@ class LoadBalancerService(
     /** Gets a producer which can publish messages to the kafka bus. */
     private val messageProducer = new KafkaProducerConnector(config.kafkaHost, executionContext)
 
-    private def sendActivationToInvoker(producer: MessageProducer, msg: ActivationMessage, invokerName: String): Future[RecordMetadata] = {
+    private def sendActivationToInvoker(producer: MessageProducer, msg: ActivationMessage, invoker: InstanceId): Future[RecordMetadata] = {
         implicit val transid = msg.transid
-        val start = transid.started(this, LoggingMarkers.CONTROLLER_KAFKA, s"posting topic '$invokerName' with activation id '${msg.activationId}'")
 
-        producer.send(invokerName, msg).andThen {
+        val topic = s"invoker${invoker.toInt}"
```
rabbah: should the `Instance` type precompute this once; ie `Instance(prefix, id)`?

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -269,59 +257,47 @@ class LoadBalancerService(
...
-                        Future.failed(new LoadBalancerException("no invokers available"))
-                }
+    private def chooseInvoker(user: Identity, action: ExecutableWhiskAction): Future[InstanceId] = {
+        val hash = generateHash(user.namespace, action)
+
+        allInvokers.flatMap { invokers =>
+            val invokersToUse = if (action.exec.pull) blackboxInvokers(invokers) else managedInvokers(invokers)
+            val invokersWithUsage = invokersToUse.view.map {
+                // Using a view defers the comparably expensive lookup to actual access of the element
+                case (instance, state) => (instance, state, activationByInvoker.get(instance).map(_.size).getOrElse(0))
```
rabbah: i wonder if state + count should be collapsed: if state is not healthy, count doesn't matter... a thought.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -269,59 +257,47 @@ class LoadBalancerService(
...
-     * Invoker is currently using and which is better avoid if/until
-     * these are moved to some common place (like a subclass of Message?)
-     */
-    private def hashSubjectAction(msg: ActivationMessage): Int = {
-        val subject = msg.user.subject.asString
-        val path = msg.action.toString
-        (subject.hashCode() ^ path.hashCode()).abs
+    /** Generates a hash based on the string representation of namespace and action */
+    private def generateHash(namespace: EntityName, action: ExecutableWhiskAction): Int = {
+        (namespace.asString.hashCode() ^ action.fullyQualifiedName(false).asString.hashCode()).abs
```
rabbah: namespace vs subject - we may want to reconsider this.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -351,58 +327,37 @@ object LoadBalancerService {
...
+                .map(i => (homeInvoker + i * step) % numInvokers)
+                .map(invokers)
+                .filter(_._2 == Healthy)
+
+            invokerProgression.find(_._3 < invokerBusyThreshold)
+                .orElse(invokerProgression.find(_._3 < invokerBusyThreshold * 2))
+                .orElse(invokerProgression.find(_._3 < invokerBusyThreshold * 3))
+                .orElse(invokerProgression.headOption)
+                .map(_._1)
+        } else None
```
rabbah: _this is beautiful._

rabbah: the hard part about reading this was figure out how to destructure the tuple... almost tempted for a case class.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
markusthoemmes: The scheduling in the Invoker happens based on the namespace, so it only makes sense to schedule based on namespaces here.

Issue #2481, opened 2017-07-13
invoker doesn't work propertly in case of heavy load


keunseob:
## Environment details:

* Openwhisk on Kubernetes based on AWS EC2
* ubuntu 14.04, Kube 1.5.6

## Steps to reproduce the issue:

1.   setup Environment Openwhisk on K8s
2.   set 3 invokers pod
2.   create a action (similiar to helloword)
3.   try to requst lots of trigger action as **blocking mode**

## Provide the expected results and outputs:
after lots of request consistently (about 3~4000), invoker can't return proper response. 
it returned only activation id, i already knew about it is normal.. but after invokers consumed all of requests, it has to be return proper response.
but It will not be .... even waiting for a long time.
once invoker can't return response propertly, invoker didn't work properly anymore
so, i tried to check invoker log as below

> [2017-07-11T05:54:55.453Z] [WARN] [#tid_113086] [ContainerPool] getContainer has been waiting about 10000 ms: position = 7872 completed = 7869 slack = 1 maxActive = 4 activeCount = 3 startingCounter = 0
> [2017-07-11T05:54:55.453Z] [WARN] [#tid_113086] [ContainerPool] getContainer has been waiting about 10000 ms: position = 7872 completed = 7869 slack = 1 maxActive = 4 activeCount = 3 startingCounter = 0
> [2017-07-11T05:54:55.453Z] [WARN] [#tid_113086] [ContainerPool] getContainer has been waiting about 10000 ms: position = 7872 completed = 7869 slack = 1 maxActive = 4 activeCount = 3 startingCounter = 0

in this case, invoker not able to handle requests.
however controller log shows it is healthy...

it need to be fixed... current healthy checker mechanism is used only for checking invoker creation..

markusthoemmes:
Those logs inform you of long internal queue times. @rabbah has a PR (#2425) which is going in very soon which at least makes up for a proper backpressure mechanism, these logs should go away.

Nonetheless the invoker should eventually handle all the load you hammer at it. Was this not the case? Are you deploying only on a single invoker?

keunseob:
Thanks for your information. I already know about the poll when i received the activationId from OW.
however i'm afraid that it would be not right result user want.
and @rabbah 's PR also looks great. however regarding health checking is not little different approach of me. I give it a try to suggest another PR that i tried and tested recently.

rabbah:
Referencing https://github.com/apache/incubator-openwhisk/issues/2506 as the underlying issue.

PR #2482, opened 2017-07-13
Add container pool health status in invoker

keunseob:
I tried to tons of test using OW on K8s and found some unexpected result like a #2481.
anyway it is one of my trials which gives good result.
current invoker **always** sends healthy message to loadBalancer.
But sometimes, it is meaingless except for invoker creation time.
so, loadBalancer assigns request to unhealthy invoker.
thus, I added container pool status check code and it disturbs sending health message when it is not healthy.
I think, the invoker who knows best about itself is itself.
so healty message should be sent by invoker.

Signed-off-by: keunseob.kim <keunseob.kim@samsung.com>

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
@@ -128,6 +128,8 @@ class ContainerPool(
     def maxIdle_=(value: Int): Unit = _maxIdle = Math.max(0, value)
     def maxActive_=(value: Int): Unit = _maxActive = Math.max(0, value)
 
+    def getCurrentPoolState: Int = _currentPoolState.id
+
     def resetMaxIdle() = _maxIdle = defaultMaxIdle
     def resetMaxActive() = {
         _maxActive = ContainerPool.getDefaultMaxActive(config)
@@ -228,6 +230,7 @@ class ContainerPool(
         val warnAtCount = 10.seconds.toMillis / waitDur.toMillis
         val warnPeriodic = 60.seconds.toMillis / waitDur.toMillis
         if (tryCount == warnAtCount || tryCount % warnPeriodic == 0) {
+            _currentPoolState = poolState.Unhealthy
             logging.warn(this, s"""getContainer has been waiting about ${warnAtCount * waitDur.toMillis} ms:
                           | position = $position
                           | completed = ${completedPosition.cur}
@@ -277,12 +280,14 @@ class ContainerPool(
                 this.synchronized {
                     introduceContainer(key, con).state = State.Active
                 }
+                _currentPoolState = poolState.Healthy
                 Some(Cold(con))
             }
             case CacheHit(con) =>
                 con.transid = transid
                 runDockerOp {
                     if (con.unpause()) {
+                        _currentPoolState = poolState.Healthy
                         Some(Warm(con))
                     } else {
                         // resume failed, gc the container
@@ -388,6 +393,10 @@ class ContainerPool(
 
     object State extends Enumeration {
         val Idle, Active = Value
+    }
+
+    object poolState extends Enumeration {
+        val Healthy, Unhealthy = Value
     }
 
     /**
@@ -682,6 +691,8 @@ class ContainerPool(
     private var gcOn = true
     private val gcSync = new Object()
     resetMaxActive()
+
+    private var _currentPoolState = poolState.Healthy   // Initial health state
 
     private val timer = new Timer()
     private val gcTask = new TimerTask {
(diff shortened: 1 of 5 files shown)


markusthoemmes:
Thanks for the contribution!

If you look at how the loadbalancer works, it already takes the specific load of each invoker into account. Overload is not a matter of healthyness and should not be. The health protocol is solely to catch error cases, not overload scenarios.

rabbah:
Note that this change makes it impossible to queue requests in the system - instead when the capacity of the system is exhausted, all new requests will get rejected (since no invokers will be available).

keunseob:
The reason why I add this logic is that sometimes invoker can not control docker container due to internal caused failure.
In this case, controller is not able to know about it because invoker health message still coming.
Thus controller assign request to unhealthy invoker

Too heavy request are stacked on Kafka queue (in spite of load balanced) , invoker reach to only waiting status with warning log below and can not get container anymore.

After About thousands of logs below are displayed, invoker can be changed to normal state. (But in case of pool infra, it can not be changed to normal state due to docker error)

The worst state unhealthy invoker never recovered to normal state.

So what I want to do is that controller shouldn't know about the unhealthy invoker. In the worst case.


[WARN] [#tid_113086] [ContainerPool] getContainer has been waiting about 10000 ms: position = 7872 completed = 7869 slack = 1 maxActive = 4 activeCount = 3 startingCounter = 0

keunseob:
@rabbah yes i think never been recovered invoker should be excluded. If so another invoker can handle requests intead of bad invoker.
If all invokers are bad, it is equals to no invokers.

markusthoemmes:
The health protocol should very well catch a broken invoker already. If it gets activation responses containing *system errors*, it will mark the invoker as unhealthy and then tries to get it back via a testpayload.

Do you know why the invoker reached a state of unhealthyness in your case?

rabbah:
Does our health protocol detect the unhealthy invoker if there's no response to the request?

keunseob:
@markusthoemmes current containerpool scheduling and sending health scheduling are separated. So i'm curious why you can guarantee that the invoker environment  status can be checked. Of course i knows the controller unhealthy mark process, but it doesn't cover all unexpected error by docker environment on invoker vm(pod or pm)

I tried to many heavy test on k8s environment. 
when i request heavy request similar to ddos as blocking request.
It returns activationId only till several days pass. And never get real result with waiting logs above. It is bad invoker case.

The heavy request test is necessary to confirm various unexpected result and check OW reliability 

So i think that the bad invoker should be founded and excluded honestly.

It would be better policy in terms of users.
Current mechanism doesn't concern about this, so OW operator is very hard to know the each invoker status at a glance.

It would be minor factor of OW reliability.

So this PR is to exclude bad invoker clearly and if it will be changed to fine, it will be included in invoker pool.

rabbah:
Look in InvokerSupervision. It actually tried to invoke an action. If that fails it will make the invoker unhealthy. Yes some requests might get queued on that invoker while the health protocol completes but eventually it should detect a failing invoker. 

    /** Scheduler to send test activations when the invoker is unhealthy. */
    onTransition {
        case _ -> UnHealthy => {
            invokeTestAction()
            setTimer(InvokerActor.timerName, Tick, 1.minute, true)
        }
        case UnHealthy -> _ => cancelTimer(InvokerActor.timerName)
    }

But I now suspect there's a bug in the protocol in the case of a missing active ack. I have to check again in the morning and confirm if a test exists in that case. 

I think your change is actually undesirable. In that the invoker pool can be full because of load - that should not mark the invoker unavailable.

markusthoemmes:
Let me explain:

The health ping is solely to check if the invoker is there. That one has nothing to do with any state inside the invoker as you pointed out.

We do have a second mechanism named active ack. It's used to fastly send back the response, which is then sent to the user. The controller checks those active ack messages for system errors (not application errors). If for a given timeframe one invoker sends too many subsequent system errors the invoker is considered unhealthy and checked again.

If missing active acks are not caught by that, as @rabbah pointed out, we have a bug there.

rabbah:
Which version of docker are you running and which kernel are you on? Did you check if there are kernel panics or whether the docker daemon is hung? Curious what happened to make docker unresponsive.

keunseob:
@rabbah i will post more detail information on next monday.  And how current mechanism can exclude unexpected error invoker? Like a kernel panic or docker demin error?

In this ack mechanism makes # of fast timers for bad invokers
It wouldn't be good for large number invokers environment
Unnecessary fast timers could be created continuously

keunseob:
And i will try to check same bug focusing active ack

rabbah:
Thanks @keunseob! fixing the health protocol so it is engaged when no responses are received and to account for health correctly in that case will achieve the result you want and fix the bug so the behavior is what's intended . 

I didn't understand your comment about fast timers - do you mind rephrasing or claryifing your statement?

rabbah:
I created an issue for the underlying bug https://github.com/apache/incubator-openwhisk/issues/2506.

rabbah:
@keunseob the proposed patch in #2658 i think will address the same goals as your proposed patch here while fixing the defect that missing active acks are ignored.

keunseob:
@rabbah ok good currently I was not able to see that issue more deeply. So after close my issue and merge the proposed the issue, I will try to check it again thanks

PR #2483, opened 2017-07-13
Externalize access to some CLI methods to facilitate access across packages

mdeuser:
- Also move properties file reading/loading to when the command is executed, not at binary load time


mdeuser:
pg4/528

PR #2489, opened 2017-07-14
Remove unneeded WeakReference.

markusthoemmes:
The closure doesn't keep a reference to the entire result anymore and is decoupled by the timeout handler through the underlying state-map. The WeakReference is not needed anymore and can actually cause failures if a garbage collection happens in between creating and getting the reference in this case.


markusthoemmes:
@starpit think this is good now?

starpit:
it does indeed look like the closure no longer captures the Promise, thus obviating any need for weak references, in this case.

it is also worth noting that the original impl likely had a race bug, as it created a Promise object that, for a brief moment, was only weakly referenced.

rabbah:
it is: @starpit https://github.com/apache/incubator-openwhisk/blob/master/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala#L134

Issue #596, opened 2016-06-08
Provide steps to re-deploy after restarting Vagrant VM


lfzCarlosC:
problem description
--------------------------
For Hello case 
./wsk action invoke /whisk.system/samples/echo -p message hello --blocking --result

returns   "error: [Errno 111] Connection refused"

steps to reproduce
--------------------------
1. Clone openwhisk
git clone --depth=1 https://github.com/openwhisk/openwhisk.git

2. Change directory to tools/vagrant/simple
cd openwhisk/tools/vagrant/simple

3. Run script to create vm and run hello action
vagrant up 
./wsk action invoke /whisk.system/samples/echo -p message hello --blocking --result

lfzCarlosC:
I try the quickcase following  the guide         https://github.com/openwhisk/openwhisk

lfzCarlosC:
I run it locally on my laptop

csantanapr:
Hi @lfzCarlosC could you provide more info on your environment to be able to reproduce?

I tried today with latest hash `bf3fe240ff56f772837f7c9ccf909483b1893a8c`

I'm using Mac OSX El Capitan
```
$ uname -a
Darwin carlosmacbookpro.raleigh.ibm.com 15.5.0 Darwin Kernel Version 15.5.0: Tue Apr 19 18:36:36 PDT 2016; root:xnu-3248.50.21~8/RELEASE_X86_64 x86_64
```

```
$ vagrant -v
Vagrant 1.8.1
```

```
Virtual Box Version 5.0.20 r106931
```
I clean ~/.vagrant.d/ just in case
I made sure no whisk VMs are listed or running on Virtual Box
No other VMs are running 
Steps: 

```
$ rm -rf ~/.vagrant.d/
$ git clone --depth=1 https://github.com/openwhisk/openwhisk.git
$ cd openwhisk/tools/vagrant/simple/
$ ./hello

wait for a while, downloading new vagrant ubuntu to cache (10 minutes) then whisk setup (20 minutes)

==> default: ++ bin/wsk action invoke /whisk.system/samples/echo -p message hello --blocking --result
==> default: {
==> default:     "message": "hello"
==> default: }
{
    "message": "hello"
}
```

```
$ git log
commit bf3fe240ff56f772837f7c9ccf909483b1893a8c
Author: Perry Cheng <perry@us.ibm.com>
Date:   Wed Jun 8 21:34:48 2016 -0400
```

lfzCarlosC:
[...]

PR #2491, opened 2017-07-14
Limit length of HTTP body displayed when debugging.

### LINKED ISSUE #928: Limit amount of HTTP body displayed to stdout when -d option is used
## Environment details:
- Any
1.   create a large file (50 MB) called bigfile.js.  The file does not have to be valid JavaScript
2.   `wsk action create bigfile bigfile.js -d`

Display a subset of the HTTP body (i.e. subset of the script file).

Too much output id dumped to the display.  The terminal may also hang
- May also need to consider handling binary vs text files differently as well.

bpoole16:
This aims to limit the length of HTTP body displayed to stdout when -d option is used. Closes #928


bpoole16:
@mdeuser Let me know if this is what you were thinking.

tests/src/test/scala/system/basic/WskBasicTests.scala
```diff
@@ -492,6 +492,21 @@ class WskBasicTests
...
     }
 
+    it should "create an action and limit the length of req body in the debug to 1000 bytes" in {
+            val file = Some(TestUtils.getTestActionFilename("largeFile.js"))
+            val actionName = "largeFileAction"
+            val msg = "Req Body excedes 1000 bytes and will be truncated"
+            val output = "fWbADPnp\"" //   Ending substring of truncated output
+
+            try {
+                val action = wsk.action.create(name = actionName, artifact = file, expectedExitCode = SUCCESS_EXIT, debug = Some(true)).stdout
```
dubeejw: Should be able to use `assetHelper.withCleaner` here.

tests/src/test/scala/system/basic/WskBasicTests.scala
```diff
@@ -492,6 +492,21 @@ class WskBasicTests
...
+                action should include(output)
+                action should include(msg)
+            } finally {
+                val delAction = wsk.action.delete(name = actionName, expectedExitCode = DONTCARE_EXIT)
```
dubeejw: Use the built-in cleaner instead. We would need the exit code to be successful value. Otherwise, some other test may try to create an action with the same name you are using and fail if this action wasn't actually deleted.

tests/dat/actions/largeFile.js
```diff
@@ -0,0 +1,11 @@
+function main (params) {
+    var name = params.payload || "User";
+    var list = "L69lhnfnD1rjFoesRquHOIXHAMR3cYp8pVxfag38WNHn6ijlyEn7r5VsNlM6kvZQHje6q6nTV5Wts3wwG6NqQGVllPm5bH6cFnburN2UwRIZrVVK9lPXTlL0YmZbpKQaTilpiXhnQH2RwRGEtzK2Puos\n" +
```
dubeejw: Can this list be generated programmatically?

tools/cli/go-whisk/whisk/client.go
```diff
@@ -211,14 +211,31 @@ func (c *Client) addAuthHeader(req *http.Request, authRequired bool) error {
     }
     return nil
 }
-
+//Limiter limits the size of 'Req Body (ASCII quoted string)' output for the debugger ONLY.
+//Returns truncated req body, reloaded io reader and errors.
+func limiter(body io.ReadCloser) (string, io.ReadCloser, error) {
```
dubeejw: Maybe use a more descriptive function name (bodyLimiter).

tests/dat/actions/largeFile.js
bpoole16: I need to fill the text file with around 1000 or more bytes that is why it is hard coded.

dubeejw: Could do something like this var list = "a".repeat(1000)` or `var list = Array(1000).join("a")`.

dubeejw: @bpoole16, I see your testing on action create. The above would only work on invoke.

tests/src/test/scala/system/basic/WskBasicTests.scala
bpoole16: I was trying to but wasn't sure how to get stdout of a action.create with asset-cleaner.

tests/src/test/scala/system/basic/WskBasicTests.scala
```diff
@@ -496,14 +496,14 @@ class WskBasicTests
...
 
             try {
                 val action = wsk.action.create(name = actionName, artifact = file, expectedExitCode = SUCCESS_EXIT, debug = Some(true)).stdout
-                action should include(output)
+                action should include regex(output)
                 action should include(msg)
             } finally {
-                val delAction = wsk.action.delete(name = actionName, expectedExitCode = DONTCARE_EXIT)
+                val delAction = wsk.action.delete(name = actionName, expectedExitCode = SUCCESS_EXIT)
             }
```
bpoole16: Wasn't able to get `.stdout` to work with asset-cleaner, if anyone has any ideas please let me know.

tests/src/test/scala/system/basic/WskBasicTests.scala
```diff
@@ -496,14 +496,14 @@ class WskBasicTests
...
             val msg = "Req Body excedes 1000 bytes and will be truncated"
-            val output = "fWbADPnp\"" //   Ending substring of truncated output
+            val output = "[A-Za-z1-9{25}]\"".r  //   Ending substring of truncated output
 
             try {
                 val action = wsk.action.create(name = actionName, artifact = file, expectedExitCode = SUCCESS_EXIT, debug = Some(true)).stdout
-                action should include(output)
+                action should include regex(output)
                 action should include(msg)
             } finally {
```
bpoole16: Wasn't able to get .stdout to work with asset-cleaner, if anyone has any ideas please let me know.

tests/src/test/scala/system/basic/WskBasicTests.scala
```diff
@@ -492,6 +492,21 @@ class WskBasicTests
...
     }
 
+    it should "create an action and limit the length of req body in the debug to 1000 bytes" in {
+            val file = Some(TestUtils.getTestActionFilename("largeFile.js"))
+            val actionName = "largeFileAction"
+            val msg = "Req Body excedes 1000 bytes and will be truncated"
+            val output = "[A-Za-z1-9{25}]\"".r  //   Ending substring of truncated output
+
+            try {
+                val action = wsk.action.create(name = actionName, artifact = file, expectedExitCode = SUCCESS_EXIT, debug = Some(true)).stdout
```
bpoole16: Having trouble getting assest-cleaner to work with action.create(xxx).stdout. Any ideas would be much appreciated.

mdeuser: Is stdout needed here?  I think it will be the wsk.action.invoke stdout that is being tested for truncation..

tests/dat/actions/largeFile.js
rabbah: we have an action that can do this already - see the action limit tests.

tests/src/test/scala/common/Wsk.scala
```diff
@@ -299,7 +299,8 @@ class WskAction()
         shared: Option[Boolean] = None,
         update: Boolean = false,
         web: Option[String] = None,
-        expectedExitCode: Int = SUCCESS_EXIT)(
+        expectedExitCode: Int = SUCCESS_EXIT,
+        debug: Option[Boolean] = None)(
```
rabbah: do we really want to test cli debug here?

tests/src/test/scala/system/basic/WskBasicTests.scala
```diff
@@ -492,6 +492,21 @@ class WskBasicTests
...
+                action should include regex(output)
+                action should include(msg)
+            } finally {
+                val delAction = wsk.action.delete(name = actionName, expectedExitCode = SUCCESS_EXIT)
```
rabbah: why assign a value that's not used.
also do not use create/delete instead use the asset cleaner.

tests/src/test/scala/system/basic/WskBasicTests.scala
rabbah: @bpoole16 can you explain the issue?

mdeuser:
@rabbah - this is a fix for limiting large http body data being dumped out when the --verbose or --debug flag is specified.

tests/src/test/scala/system/basic/WskBasicTests.scala
bpoole16: The goal, from my understanding, is to limit the HTTP body during output of a debug/verbose. The only way I've been able to access this information is through stdout of a create, invoke doesn't seem to supply the correct req or response. This also leads to the question of whether I should be testing something that is in the debug/verbose.

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
```diff
@@ -269,4 +269,29 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers {
...
 
+    it should "limit the length of HTTP req/resp bodies for debug messages to 1000 bytes" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+            val auth: Seq[String] = Seq("--auth", wskprops.authKey)
+            val name = "largeFileAction"
+            val debugMsg = "Body excedes 1000 bytes and will be truncated"
+            val bodyMsg = "ENDUNIQUE"   //  Message that should be truncated
+
+            val actionCode = new File(s"$testActionsDir${File.separator}$name.js")
+            actionCode.createNewFile()
```
dubeejw: Why create the file for each test run instead of having one in `dat/actions`? You can use `echo.js` and send a large string as a parameter to it.

dubeejw:
```
val input = "a" * 1000
wsk.action.invoke(name, parameters = input)
```

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
```diff
@@ -269,4 +269,29 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers {
             }
     }
 
+    it should "limit the length of HTTP req/resp bodies for debug messages to 1000 bytes" in withAssetCleaner(wskprops) {
+        (wp, assetHelper) =>
+            val auth: Seq[String] = Seq("--auth", wskprops.authKey)
+            val name = "largeFileAction"
+            val debugMsg = "Body excedes 1000 bytes and will be truncated"
```
rabbah: so how would one check if the cli encoded a zip correctly? should debug continue to show the full content but not verbose? one or the other...

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
dubeejw: Looks like the issue is for create, but your body limiter should limit all the received input?

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
rabbah: the reason i say debug is we have had situations where the debug response was useful to find problems with the CLI where the HTTP response is correct but the output of the CLI that's formatted for consumption isn't (due to scema mismatches).

I'm still puzzled by this change - why limit debug output at all.
Should there be an option to dump the output to a file at least?

rabbah: exceeds is mispelled.

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
dubeejw: Or if you really want to test create, you can pass a big parameter to any of the existing files in `action/data`.
```
val input = Map("payload" -> JsString("a" * 1000))
assetHelper.withCleaner(wsk.action, name) {
                (action, _) => action.create(name, file, parameters = input)
            }
```

bpoole16: Are we okay with adding `--debug` to invoke or should I take the path of Seq() so I don't have to create a `--debug` variable?

rabbah: the latter - as the cli tests are moved to their own repository, this flag wont be relevant to the REST api implementation of the Wsk interface.

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
bpoole16: From my understanding, maybe @mdeuser can elaborate more on this, we currently print out the req/resp body a total of 4 times (2 of those being debugs) during a `wsk action create xxx -d`. If you take a large file, the amount of output can overwhelm the CLI and even hang it.  Maybe this fixes something that is too rare to matter if there are more frequent disadvantages?

rabbah: so it is a problem that there are too many debug statements - rather than limiting the debug output, should there be deduping of debug messages?

bpoole16: Just to be clear, there is only 1 debug per body. For instance, req body has 1 debug and println and resp body has 1 debug and println - printing a total of 4 times.

rabbah: OK then I'm entirely confused.

Can you show an example before and after with desired effect?

bpoole16: Example of before and after for req body output of `wsk action create xxx -d`
limitexample

bpoole16: Right, you wouldn't be able to check if a request was encoded correctly. Maybe adding a flag to allow for truncation would be correct path to take?

rabbah: Sounds like the debug flag to me - I'm not sure what truncating is buying us. Debug/Verbose is used sparingly.

bpoole16: If there is a large file, let's say 1MB or larger, there is a great deal of data dumping onto the screen. A truncation could help limit this enormous amount of data being displayed to the user/developer. But in smaller cases, like you pointed out, it seems to hamper the results more than it helps. This is why I've been thinking of either creating a separate flag or a raising the cap from 1000 bytes.

It's true, this issue could be prevented by using verbose where only `req body` and `code` are displayed but there is a loss of important debug features here. Also, the original issue called for the implementation of truncation with debug which is why I proceeded down this path. Since debug/verbose is used so sparingly, that brings up a concern; how sparingly would someone use debug AND have a large file?

rabbah:
the cli has two flags: verbose and debug right? how about verbose truncate and debug doesn't.

bpoole16:
Will attempt to proceed in the direction of limiting verbose only and leaving the debug to print fully.

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
```diff
@@ -60,6 +60,26 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers {
 
     behavior of "Action limits"
 
+    it should "limit the length of HTTP req/resp bodies for debug messages to 1000 bytes" in withAssetCleaner(wskprops) {
```
bpoole16: Old test, need updating.

bpoole16:
PG 4 #596 Passsed

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
```diff
@@ -60,6 +60,31 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers {
 
     behavior of "Action limits"
 
+    it should "limit the length of HTTP Req/Resp Body for --verbose to 1000 bytes" in withAssetCleaner(wskprops) {
```
dubeejw: This test is in the wrong file. ActionLimitTests are for action limits related to the RestAPI. Should move to WskBasicUsageTests.scala.

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
```diff
@@ -60,6 +60,31 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers {
...
+            val largeTestFile = new File(s"$testActionsDir${File.separator}$name.js")    // Creates a file to see if "code" field is limited
+
+            largeTestFile.createNewFile()
+            val pw = new PrintWriter(largeTestFile)
+            pw.write("a" * 1000)
+            pw.write(endMsg)
+            pw.close
+
+            assetHelper.withCleaner(wsk.action, name) {
+                (action, _) => wsk.action.create(name, Some(largeTestFile.getAbsolutePath))
```
dubeejw: Did you try using the a large parameters argument instead of creating the big file for every run?

tools/cli/go-whisk/whisk/client.go
```diff
@@ -222,14 +222,76 @@ func (c *Client) addAuthHeader(req *http.Request, authRequired bool) error {
...
+    if err != nil {
+        Verbose("ioutil.ReadAll(req.Body) error: %s\n", err)
+        werr := MakeWskError(err, EXITCODE_ERR_NETWORK, DISPLAY_MSG, NO_DISPLAY_USAGE)
+        return nil, body, werr
+    }
+    reload := ioutil.NopCloser(bytes.NewBuffer(data))
+    if len(data) > limit {
+        Verbose("Req Body excedes %d bytes and will be truncated\n", limit)
+        newData := string(data[:limit]) + buffer    // Must convert to string to add buffer otherwise malformed error will be thrown
+        return []byte(newData), reload, nil
```
dubeejw: Seems like an unneeded cast back to a byte array here since the data is just converted back to a string later.

tools/cli/go-whisk/whisk/client.go
```diff
@@ -222,14 +222,76 @@ func (c *Client) addAuthHeader(req *http.Request, authRequired bool) error {
...
+    data, err := ioutil.ReadAll(body)
+    if err != nil {
+        Verbose("ioutil.ReadAll(req.Body) error: %s\n", err)
+        werr := MakeWskError(err, EXITCODE_ERR_NETWORK, DISPLAY_MSG, NO_DISPLAY_USAGE)
+        return nil, body, werr
+    }
+    reload := ioutil.NopCloser(bytes.NewBuffer(data))
+    if len(data) > limit {
+        Verbose("Req Body excedes %d bytes and will be truncated\n", limit)
+        newData := string(data[:limit]) + buffer    // Must convert to string to add buffer otherwise malformed error will be thrown
```
dubeejw: Not sure I see why a new line is added here when the data is being printing with `fmt.Println()`.

tools/cli/go-whisk/whisk/client.go
```diff
@@ -222,14 +222,76 @@ func (c *Client) addAuthHeader(req *http.Request, authRequired bool) error {
...
+    limit := 1000    // 1000 byte limit, anything over is truncated
+    buffer := "\"\n    "    // Appended to the end of newData to keep correct formating
+    data, err := ioutil.ReadAll(body)
+    if err != nil {
+        Verbose("ioutil.ReadAll(req.Body) error: %s\n", err)
+        werr := MakeWskError(err, EXITCODE_ERR_NETWORK, DISPLAY_MSG, NO_DISPLAY_USAGE)
+        return nil, body, werr
+    }
+    reload := ioutil.NopCloser(bytes.NewBuffer(data))
+    if  bytes.Contains(data, []byte("\"code\":")) && bytes.Contains(data, []byte("\"binary\":")) {
```
dubeejw: Hmm, why only limit binary data?

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
dubeejw: Does this test both respBodyLimiter and reqBodyLimiter?

tools/cli/go-whisk/whisk/client.go
bpoole16: The goal is to limit the fields that are known to run long, this really only being the code field. I'm trying to figure out the best way to extract the code field and limit it and then reconstruct the output with the newly limited code field.

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
bpoole16: Yes it does, the issue is other commands aren't working properly.

tools/cli/go-whisk/whisk/client.go
bpoole16: I agree, was overthinking this part.

tools/cli/go-whisk/whisk/client.go
bpoole16: I was trying to replicate the original output and without that buffer it does add a space between req body and `RESPONSE:Got response with code 200 `

tools/cli/go-whisk/whisk/client.go
bpoole16: Ah, just remembered why I did that. If you just return a string then you would have to wrap data in string. It seems to be a pick your poison, wrap newData in a []byte or wrap data in a string.

bpoole16:
@dubeejw Made some changes per your request, also let me know if I should convert to strings vs []byte.

bpoole16:
To me this issue has gotten too specific. It's now only limiting in verbose for a certain field in resp body for actions only - feels off to me. The only "global" limiting in verbose is for req body.

dubeejw:
What I noticed was that when I ran an action get for an action that has a big zip file attached to it, I still had to scroll through all the normal output before I could seen the truncated code in the verbose output. I think it'd be better to truncate the code in verbose and normal modes. @rabbah

dubeejw:
@bpoole16, how about doing it this way? https://github.com/dubeejw/openwhisk/commit/5b57eb6ddd4b6d55a051f24f6e221b2fba944dd8

bpoole16:
That works, I had something like that similar until I realized it was cutting off a bunch of resp information. Due to this I was just trying to find problematic fields and reduce them. If we are okay with losing resp body information looks fine to me.

docs/rest_api.md
```diff
@@ -70,7 +70,9 @@ The OpenWhisk API supports request-response calls from web clients. OpenWhisk re
 ## Using the CLI verbose mode
 
 The OpenWhisk CLI is an interface to the OpenWhisk REST API.
-You can run the CLI in verbose mode with the flag `-v`, this will print all the information about the HTTP request and response.
+You can run the CLI in verbose mode with the flag `-v`, this will print truncated information about the HTTP request and response. To print all information use the flag `-d` for debug.
```
bpoole16: Made update to docs here.

docs/rest_api.md
```diff
@@ -70,7 +70,9 @@ The OpenWhisk API supports request-response calls from web clients. OpenWhisk re
...
+
+**Note** Only HTTP request and response bodies will be truncated if they exceed 1000 bytes.
```
bpoole16: Made update to docs here.

mdeuser:
@bpoole16 - what will be the output of `wsk action get LARGEACTION` when the exec.code field is quite large?  will the rest of the action output be intact?

bpoole16:
Originally yes, it would just limit the code field and the rest of the resp body would be normal. With James suggestion it cuts off the resp body after 1000 bytes.

bpoole16:
PG 4 632 passed

PR #2493, opened 2017-07-14
API gw test cleanup

rabbah:
- [x] Remove unnecessary vars.
- [x] Refactor check throttle to eliminate proxy methods.

@mdeuser @dubeejw @bpoole16 FYI

### DIFF
--- a/tests/src/test/scala/whisk/core/cli/test/ApiGwTests.scala
+++ b/tests/src/test/scala/whisk/core/cli/test/ApiGwTests.scala
@@ -21,11 +21,16 @@ import java.io.File
 import java.io.BufferedWriter
 import java.io.FileWriter
 import java.time.Instant
+
 import scala.collection.mutable.ArrayBuffer
 import scala.concurrent.duration._
+
 import org.junit.runner.RunWith
+
 import org.scalatest.BeforeAndAfterAll
+import org.scalatest.BeforeAndAfterEach
 import org.scalatest.junit.JUnitRunner
+
 import common.TestHelpers
 import common.TestUtils._
 import common.TestUtils
@@ -41,9 +46,10 @@ import common.WskTestHelpers
 class ApiGwTests
     extends TestHelpers
     with WskTestHelpers
+    with BeforeAndAfterEach
     with BeforeAndAfterAll {
 
-    implicit var wskprops = WskProps()
+    implicit val wskprops = WskProps()
     val wsk = new Wsk
     val clinamespace = wsk.namespace.whois()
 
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/whisk/core/cli/test/ApiGwTests.scala
```diff
@@ -1188,34 +1167,34 @@ class ApiGwTests
...
-        var rr = apiDelete(basepathOrApiName = nonexistentApi, expectedExitCode = ANY_ERROR_EXIT)
-        rr.stderr should include (s"API '${nonexistentApi}' does not exist")
+        val rr = apiDelete(basepathOrApiName = nonexistentApi, expectedExitCode = ANY_ERROR_EXIT)
+        rr.stderr should include(s"API '${nonexistentApi}' does not exist")
     }
 
     it should "successfully list an API whose endpoints are not mapped to actions" in {
         val testName = "CLI_APIGWTEST23"
-        var testapiname = "A descriptive name"
+        val testapiname = "A descriptive name"
```
dubeejw: Are we not enforcing camel casing in Scala?

rabbah: not yet - i think we should consider a standard formatter as part of scan code/Travis.

Note that my change is for gratuitous use of s/var/val/g not changing and reusing asset cleaner to simplify the tests.

rabbah:
When is API experiments being removed?

tests/src/test/scala/whisk/core/cli/test/ApiGwTests.scala
```diff
@@ -630,218 +161,210 @@ class ApiGwTests
...
+        } finally {
+            wsk.action.delete(name = actionName, expectedExitCode = DONTCARE_EXIT)
+            wsk.api.delete(basepathOrApiName = testbasepath)
         }
     }
 
     it should "verify successful creation and deletion of a new API" in {
         val testName = "CLI_APIGWTEST1"
-        val testbasepath = "/"+testName+"_bp"
+        val testbasepath = "/" + testName + "_bp"
```
dubeejw: s"/$testName_bp"

dubeejw: Looks like several such occurrences can be converted to use the "s" string interpolator.

tests/src/test/scala/whisk/core/cli/test/ApiGwTests.scala
```diff
@@ -853,26 +376,26 @@ class ApiGwTests
...
         try {
-            println("cli user: "+cliuser+"; cli namespace: "+clinamespace)
+            println("cli namespace: " + clinamespace)
+
             // Create the action for the API.  It must be a "web-action" action.
             val file = TestUtils.getTestActionFilename(s"echo.js")
             wsk.action.create(name = actionName, artifact = Some(file), expectedExitCode = SUCCESS_EXIT, web = Some("true"))
 
-            var rr = apiCreate(basepath = Some(testbasepath), relpath = Some(testrelpath), operation = Some(testurlop), action = Some(actionName), apiname = Some(testapiname))
+            var rr = wsk.api.create(basepath = Some(testbasepath), relpath = Some(testrelpath), operation = Some(testurlop), action = Some(actionName), apiname = Some(testapiname))
```
dubeejw: Can we make `withCleaner` work for API commands?

tests/src/test/scala/whisk/core/cli/test/ApiGwTests.scala
rabbah: Going to leave these as concat expressions. It would have to be ${...}_bp which makes + easier to read imo.

tests/src/test/scala/whisk/core/cli/test/ApiGwTests.scala
rabbah: Maybe... will have a look when when I refactor for this last change I intended.

mdeuser:
@rabbah - the experimental bits will likely begin phasing out early next month

rabbah:
Thanks Mark - we can hold this pr back until then since I scrubbed experimental except from all but the js actions. Or I can remove these bits from this pr to get the test cleanups in sooner.

mdeuser:
@rabbah - looks like this pr is not only doing some scala cleanup, but is also removing the api-experimental related test code as well as the api-experimental cli/go code...  perhaps these are best done in separate prs?

api-experimental removal will also involve some other changes in ansible, health monitoring, gradle task targets

rabbah:
ok i can back those changes out.

rabbah:
pg1/2003 🚀

rabbah:
@mdeuser can you sanity check the throttle changes, please.

tests/src/test/scala/whisk/core/cli/test/ApiGwTests.scala
```diff
@@ -58,42 +64,46 @@ class ApiGwTests
...
-      * Expected to be called before or after each CLI invocation
-      * If number of CLI invocations in this suite have reached the throttle limit
-      * then pause the test for enough time so that the throttle restriction is gone
-      */
-    def checkThrottle(maxInvocationsBeforeThrottle: Int = maxActionsPerMin) = {
-        var t = Instant.now
-        var tminus60 = t.minusSeconds(60)
-        var invocationsLast60Seconds = invocationTimes.filter(_.isAfter(tminus60)).sorted
-        var invocationCount = invocationsLast60Seconds.length
+     * Expected to be called before each CLI API command since these invoke an action.
```
mdeuser: if i understand correctly, checkthrottle() is no longer called `before each CLI API command` but is now called once per test.

rabbah:
That's correct. Once the throttle is settled none of the tests generate enough traffic to kick it back on.

mdeuser:
nit.  consistent removal of result variables in the `finally` clauses in ApiGwTests.scala
```
             for (i <- 1 to 3) {
                 val deleteApis = apiDelete(basepathOrApiName = s"${baseName}$i", expectedExitCode = DONTCARE_EXIT)
             }
             val deleteAction = wsk.action.delete(name = actionName, expectedExitCode = DONTCARE_EXIT)
```

rabbah:
Thanks for catching that. That wasn't intended.

tests/src/test/scala/whisk/core/cli/test/ApiGwTests.scala
mdeuser: Comment needs correcting then?

PR #2495, opened 2017-07-14
Remove Unsupported Shared Flag from Package List (Review)

dubeejw:
The `--shared` flag does not work with `package list`, so remove the option from the CLI.

Closes https://github.com/apache/incubator-openwhisk/issues/2485

### DIFF
--- a/tools/cli/go-whisk-cli/commands/package.go
+++ b/tools/cli/go-whisk-cli/commands/package.go
@@ -378,7 +378,6 @@ var packageListCmd = &cobra.Command{
   PreRunE:       setupClientConfig,
   RunE: func(cmd *cobra.Command, args []string) error {
     var err error
-    var shared bool
     var qualifiedName QualifiedName
 
     if whiskErr := checkArgs(args, 0, 1, "Package list",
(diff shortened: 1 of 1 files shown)


dubeejw:
@mdeuser, please review.

dubeejw:
PG2 1817.

dubeejw:
PG approved.

mdeuser:
@dubeejw I'm wondering if the `--shared` flags in the `wsk package create` and `wsk package update` commands still still valid if there's nothing to display this state.. ?

tools/cli/go-whisk-cli/commands/package.go
```diff
@@ -528,7 +528,7 @@ func init() {
   packageBindCmd.Flags().StringSliceVarP(&flags.common.param, "param", "p", []string{}, wski18n.T("parameter values in `KEY VALUE` format"))
   packageBindCmd.Flags().StringVarP(&flags.common.paramFile, "param-file", "P", "", wski18n.T("`FILE` containing parameter values in JSON format"))
 
-  packageListCmd.Flags().StringVar(&flags.common.shared, "shared", "", wski18n.T("include publicly shared entities in the result"))
+  //packageListCmd.Flags().StringVar(&flags.common.shared, "shared", "", wski18n.T("include publicly shared entities in the result"))
```
mdeuser: Any reason to leave this line in comment out?

dubeejw:
@mdeuser, the `--shared` flag for package update and create set the `published` filed that is sent to the backend making the package public or not. The `--shared` flag for package list was meant to display only public or private package depending on the argument passed to it.

mdeuser:
ok. sounds good.  looks like none of the other cli `list` commands (action, rule, trigger) currently support filtering based on the scope.  code removal looks good to me

dubeejw:
PG3 887, approved.

PR #2496, opened 2017-07-15
Remove redundancy in entitlement checks.

markusthoemmes:
Checking proper package access rights is done on a deeper level. Disambiguating here is redundant.


markusthoemmes:
PG3 875 🔵

PR #2501, opened 2017-07-18
must add "sudo" otherwise the command will failed with error "#define…

wangqiang1588:
…    EPERM         1    /＊ Operation not permitted ＊/"

### DIFF
--- a/tools/ubuntu-setup/README.md
+++ b/tools/ubuntu-setup/README.md
@@ -27,7 +27,7 @@ Follow instructions [tools/db/README.md](../db/README.md) on how to configure a 
 
   ```
   cd <home_openwhisk>
-  ./gradlew distDocker
+  sudo ./gradlew distDocker
   ```
 
 ## Deploy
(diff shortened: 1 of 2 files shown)


rabbah:
Where are you getting permission error? Sudo should not be required.

wangqiang1588:
ubuntu default user does not have root permission , if you do not getting permission error ,your current user should be root

wangqiang1588:
it is better to add sudo ,even root

tools/ubuntu-setup/README.md
```diff
@@ -27,7 +27,7 @@ Follow instructions [tools/db/README.md](../db/README.md) on how to configure a
 
   ```
   cd <home_openwhisk>
-  ./gradlew distDocker
+  sudo ./gradlew distDocker
```
rabbah: i don't think we should make this change - building and deploying in a local environment should not require privilege escalation.

rabbah:
Going to close this per my comment above.

PR #2502, opened 2017-07-18
Use error message const to avoid replication.

lzbj:
As we discussed before.. If we agree this is correct, I'll continue the remain ones.

### DIFF
--- a/tools/cli/go-whisk/whisk/wskerror.go
+++ b/tools/cli/go-whisk/whisk/wskerror.go
@@ -34,6 +34,12 @@ const NO_DISPLAY_PREFIX         bool = false
 const APPLICATION_ERR           bool = true
 const NO_APPLICATION_ERR        bool = false
 const TIMED_OUT                 bool = true
+const UNABLE_CREATE_ACTION_ERR    string = "UNABLE_CREATE_ACTION_ERR"
+const UNABLE_GET_ACTION_ERR    string = "UNABLE_GET_ACTION_ERR"
+const INVALID_PARAMETER_ARGUMENT_ERR    string = "INVALID_PARAMETER_ARGUMENT_ERR"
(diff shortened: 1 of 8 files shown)


mrutkows:
As @rr (I believe) is indicating... 
https://github.com/apache/incubator-openwhisk/tree/master/tools/cli/go-whisk-cli/wski18n/resources

and specifically the translated map:
```
{
    "id": "Unable to create action '{{.name}}': {{.err}}",
    "translation": "Unable to create action '{{.name}}': {{.err}}"
},
```

markusthoemmes:
Specifically the id in the translation file should be a well known constant, like

```
{
    "id": "ACTION_CREATE_UNABLE",
    "translation": "Unable to create action '{{.name}}': {{.err}}"
},
```

lzbj:
Seems doesn't work like that, @mdeuser , could  you please help take a review?

tools/cli/go-whisk/whisk/wskerror.go
```diff
@@ -34,6 +34,7 @@ const NO_DISPLAY_PREFIX         bool = false
 const APPLICATION_ERR           bool = true
 const NO_APPLICATION_ERR        bool = false
 const TIMED_OUT                 bool = true
+const UNABLE_CREATE_ACTION_ERR  string = "Unable to create action '{{.name}}': {{.err}}"
```
markusthoemmes: This needs to be 

```go
const UNABLE_CREATE_ACTION_ERR  string = "UNABLE_CREATE_ACTION_ERR"
```

to work.

lzbj: Ok, thank you for your help @markusthoemmes , seems I don't t get the real spirit of of the issue yet.

PR #2503, opened 2017-07-18
Increase the timeout for the actions with big log

vvraskin:
Increase the default test timeout for a single test from 30 to 70 seconds.


markusthoemmes:
That's not the test at failure, those are:

- `succeed but truncate logs, if log size exceeds its limit`
- `succeed but truncate result, if result exceeds its limit (blocking: $blocking)`

PR #2505, opened 2017-07-18
Fix actionContainerPrefix value

keunseob:
I found a minor bug. 
actionContainerPrefix has been used to killStragglers but current static string value uses "wsk" only.
so current code can affect anything else on the kubernetes environment.
I didn't include additional code for this regarding invokerReactive, because invokerReative is not verified on K8s environment our side

Signed-off-by: keunseob.kim <keunseob.kim@samsung.com>

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
@@ -746,7 +746,7 @@ class ContainerPool(
 
     // Getter/setter for this are above.
     private var _logDir = "/logs"
-    private val actionContainerPrefix = "wsk"
+    private val actionContainerPrefix = s"wsk${invokerInstance.toInt}_"
 
     /**
      * Actually deletes the containers.


rabbah:
good find - The reactive pool is ready to go online and will be the supported pool going forward. @DanLavine et al. do you have a plan for rolling this out for kube?

DanLavine:
Hey @rabbah I have not tried using InvokerReactive yet, but I could. As far as setting it up, seems like just setting an environment variable is all thats needed.

keunseob:
please pay attention to this ;)

core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala
```diff
@@ -746,7 +746,7 @@ class ContainerPool(
 
     // Getter/setter for this are above.
     private var _logDir = "/logs"
-    private val actionContainerPrefix = "wsk"
+    private val actionContainerPrefix = s"wsk${invokerInstance.toInt}_"
```
markusthoemmes: Does the `_` change the container naming? Will there be `__` (two underscores) afterwards?

keunseob: wsk{instnaceid}_{containercounter} could make name below cases
wsk1_1
wsk11_1
wsk111_1
if we use wsk{invokerInstanceId} only, another container can be effected.

markusthoemmes: I see makes sense.

keunseob: thank you :)

rabbah:
I may have goofed on this pr - it's true the prefix appears to be strictly used by kill stragglers; but it appears this change leads to more stem cell containers generated (2 instead of 4 locally). I'm investigating.

PR #2326, opened 2017-06-01
Alphabetize listings in CLI

underwoodb-sd-ibm:
Alphabetize, coauthored by myself and @bpoole16, is intended to allow the user to sort a given list of entities alphabetically. This applies to actions, triggers, rules, packages, and apis.

By default, entities are listed in order from most to least recently updated:

```
$ wsk action list
actions
/guest/packageB/A                  private nodejs:6
/guest/C                                  private nodejs:6
/guest/A                                  private nodejs:6
/guest/packageA/B                  private nodejs:6
/guest/packageA/A                  private nodejs:6
/guest/B                                  private nodejs:6
```

To see the list of entities sorted alphabetically, use the flag `--name-sort` or `-n`:
```
$ wsk action list --name-sort
actions
/guest/A                                  private nodejs:6
/guest/B                                  private nodejs:6
/guest/C                                  private nodejs:6
/guest/packageA/A                  private nodejs:6
/guest/packageA/B                  private nodejs:6
/guest/packageB/A                  private nodejs:6
```

The entities are sorted by namespace first, then package name (with the default package actions at the top), and finally by entity name.

Side effects of this PR

- We centralized the printing functions within `commands/util.go`
- We created two new interfaces:
1. `Printables`: structures that are printed in lists on a regular basis.
2. `Sortables`: structures that are meant to be sorted.

Potential upgrades from this branch:
 
- Currently, activations are not sorted alphabetically, only by their start time, and therefore `--name-sort` is not applicable to activations; however, there may be another way of sorting activations which would be beneficial. The ability to sort activations is set up, one only needs to set the function Compare() in `whisk/activation.go` and make a custom flag to activate the sorting for it to be implemented.
- Add a flag to sort alphabetically by `Action name -> Base path -> Rel path -> Verb` for APIs

PR #2513, opened 2017-07-19
(WIP) Centralized printing

bpoole16:
Due to Alphabetize no longer being a CLI issue, sorting needs to be done server side, @underwoodb-sd-ibm and I have extracted the centralized printing statements from Alphabetize to allow for easier access to printing methods. Each command will implement an interface `printable` and be converted to a `printable` to be ran through a singular printing function.  Reference: #2326


rabbah:
> Due to Alphabetize no longer being a CLI issue,

can you elaborate this?

PR #2517, opened 2017-07-20
Support client certificate verify on server side

ningyougang:
Currently, openwhisk supports passing client certificate from wsk to
nginx, but for nginx verification, it just verifies the client certificate's
format whether right, in order to complete the whole verification flow,
it is necessary to verify the CN value which is saved in cert's subject
info whether matches the user saved in couchdb.

### DIFF
--- a/ansible/roles/nginx/files/genssl.sh
+++ b/ansible/roles/nginx/files/genssl.sh
@@ -49,30 +49,56 @@ else
     -out "$SCRIPTDIR/openwhisk-client-ca-cert.pem" \
     -days 365 -sha1 -extensions v3_ca
 
-    echo generating client key
-    openssl genrsa -aes256 -passout pass:$PASSWORD -out "$SCRIPTDIR/openwhisk-client-key.pem" 2048
+    echo generating subject:whisk.system\'s client key
+    openssl genrsa -aes256 -passout pass:$PASSWORD -out "$SCRIPTDIR/openwhisk-client-key-whisk-system.pem" 2048
 
-    echo generating client certificate csr file
+    echo generating subject:whisk.system\'s client certificate csr file
     openssl req -new \
-    -key "$SCRIPTDIR/openwhisk-client-key.pem" \
+    -key "$SCRIPTDIR/openwhisk-client-key-whisk-system.pem" \
     -passin pass:$PASSWORD \
-    -subj "/C=US/ST=NY/L=Yorktown/O=OpenWhisk/CN=guest" \
-    -out "$SCRIPTDIR/openwhisk-client-certificate-request.csr"
+    -subj "/C=US/ST=NY/L=Yorktown/O=OpenWhisk/CN=whisk.system" \
+    -out "$SCRIPTDIR/openwhisk-client-certificate-request-whisk-system.csr"
 
-    echo generating self-signed client certificate
+    echo generating subject:whisk.system\'s self-signed client certificate
     echo 01 > $SCRIPTDIR/openwhisk-client-ca-cert.srl
     openssl x509 -req \
-    -in "$SCRIPTDIR/openwhisk-client-certificate-request.csr" \
+    -in "$SCRIPTDIR/openwhisk-client-certificate-request-whisk-system.csr" \
     -CA "$SCRIPTDIR/openwhisk-client-ca-cert.pem" \
     -CAkey "$SCRIPTDIR/openwhisk-client-ca-key.pem" \
     -CAserial "$SCRIPTDIR/openwhisk-client-ca-cert.srl" \
     -passin pass:$PASSWORD \
-    -out "$SCRIPTDIR/openwhisk-client-cert.pem" \
+    -out "$SCRIPTDIR/openwhisk-client-cert-whisk-system.pem" \
     -days 365 -sha1 -extensions v3_req
 
-    echo remove client key\'s password
+    echo remove subject:whisk.system\'s client key\'s password
     openssl rsa \
-    -in "$SCRIPTDIR/openwhisk-client-key.pem" \
+    -in "$SCRIPTDIR/openwhisk-client-key-whisk-system.pem" \
     -passin pass:$PASSWORD \
-    -out "$SCRIPTDIR/openwhisk-client-key.pem"
+    -out "$SCRIPTDIR/openwhisk-client-key-whisk-system.pem"
+
+    echo generating subject:guest\'s client key
+    openssl genrsa -aes256 -passout pass:$PASSWORD -out "$SCRIPTDIR/openwhisk-client-key-guest.pem" 2048
+
+    echo generating subject:guest\'s client certificate csr file
+    openssl req -new \
+    -key "$SCRIPTDIR/openwhisk-client-key-guest.pem" \
+    -passin pass:$PASSWORD \
+    -subj "/C=US/ST=NY/L=Yorktown/O=OpenWhisk/CN=guest" \
+    -out "$SCRIPTDIR/openwhisk-client-certificate-request-guest.csr"
+
+    echo generating subject:guest\'s self-signed client certificate
+    openssl x509 -req \
+    -in "$SCRIPTDIR/openwhisk-client-certificate-request-guest.csr" \
+    -CA "$SCRIPTDIR/openwhisk-client-ca-cert.pem" \
+    -CAkey "$SCRIPTDIR/openwhisk-client-ca-key.pem" \
+    -CAserial "$SCRIPTDIR/openwhisk-client-ca-cert.srl" \
+    -passin pass:$PASSWORD \
+    -out "$SCRIPTDIR/openwhisk-client-cert-guest.pem" \
+    -days 365 -sha1 -extensions v3_req
+
+    echo remove subject:guest\'s client key\'s password
+    openssl rsa \
+    -in "$SCRIPTDIR/openwhisk-client-key-guest.pem" \
+    -passin pass:$PASSWORD \
+    -out "$SCRIPTDIR/openwhisk-client-key-guest.pem"
 fi
--- a/common/scala/src/main/scala/whisk/core/entity/Identity.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Identity.scala
@@ -74,6 +74,53 @@ object Identity extends MultipleReadersSingleWriterCache[Identity, DocInfo] with
         })
     }
 
+    /**
+      * Many subjects can be belonged to only one namespace
+      * Retrieves a key for namespace + subject.
+      *
+      */
+    def get(datastore: AuthStore, namespace: EntityName, subject: Subject)(
+      implicit transid: TransactionId): Future[Identity] = {
+        implicit val logger: Logging = datastore.logging
+        implicit val ec = datastore.executionContext
+        val ns = namespace.asString
+        val sbj = subject.asString
+
+        cacheLookup(ns + sbj, {
+            //Because under one namespace can have more than one subject, so should query all subjects under this namespace.
+            list(datastore, List(ns), -1) map { list =>
+                list.length match {
+                    case 0 =>
+                        logger.info(this, s"$viewName:subject:[$subject] doesn't exist on namespace:[$namespace]")
+                        throw new NoDocumentException(s"$viewName:subject:[$subject] doesn't exist on namespace:[$namespace]")
+                    case _ =>
+                        list.find { obj =>
+                            val Seq(JsString(id)) = obj.getFields("id")
+                            id == sbj
+                        } match {
+                            case Some(row) => {
+                                row.getFields("id", "value") match {
+                                    case Seq(JsString(id), JsObject(value)) =>
+                                        val subject = Subject(id)
+                                        val JsString(uuid) = value("uuid")
+                                        val JsString(secret) = value("key")
+                                        val JsString(namespace) = value("namespace")
+                                        Identity(subject, EntityName(namespace), AuthKey(UUID(uuid), Secret(secret)), Privilege.ALL)
+                                    case _ =>
+                                        logger.error(this, s"$viewName[$ns] has malformed view '${row.compactPrint}'")
+                                        throw new IllegalStateException("identities view malformed")
+                                }
+                            }
+                            case None => {
+                                logger.info(this, s"$viewName:subject:[$subject] doesn't exist on namespace:[$namespace]")
+                                throw new NoDocumentException(s"$viewName:subject:[$subject] doesn't exist on namespace:[$namespace]")
+                            }
+                        }
+                }
+            }
+        })
+    }
+
     def get(datastore: AuthStore, authkey: AuthKey)(
         implicit transid: TransactionId): Future[Identity] = {
         implicit val logger: Logging = datastore.logging
--- a/core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
@@ -23,22 +23,18 @@ import akka.http.scaladsl.model.headers._
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.util.Try
-
 import whisk.common.Logging
 import whisk.common.TransactionId
 import whisk.core.database.NoDocumentException
-import whisk.core.entity.UUID
+import whisk.core.entity._
 import whisk.core.entity.types.AuthStore
-import whisk.core.entity.WhiskAuthStore
-import whisk.core.entity.AuthKey
-import whisk.core.entity.Identity
-import whisk.core.entity.Secret
-import whisk.core.entity.UUID
 
 object Authenticate {
     /** Required properties for this component */
     def requiredProperties = WhiskAuthStore.requiredProperties
 }
+
+case class CertificateInfo(namespace: EntityName, subject: Subject)
 
 /** A trait to validate basic auth credentials */
 trait Authenticate {
@@ -76,6 +72,24 @@ trait Authenticate {
             }.toOption
         } getOrElse {
             credentials.foreach(_ => logging.info(this, s"credentials are malformed"))
+            Future.successful(None)
+        }
+    }
+
+    def validateCertificate(certificateInfo: Option[CertificateInfo])(implicit transid: TransactionId): Future[Option[Identity]] = {
+        certificateInfo flatMap { certificateEntity =>
+            Try {
+                val future = Identity.get(authStore, certificateEntity.namespace, certificateEntity.subject) map { result =>
+                    Some(result)
+                } recover {
+                    case _: NoDocumentException | _: IllegalArgumentException =>
+                        logging.info(this, s"authentication not valid")
+                        None
+                }
+                future onFailure { case t => logging.error(this, s"authentication error: $t") }
+                future
+            }.toOption
+        } getOrElse {
             Future.successful(None)
         }
     }
--- a/core/controller/src/main/scala/whisk/core/controller/AuthenticatedRoute.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/AuthenticatedRoute.scala
@@ -19,22 +19,37 @@ package whisk.core.controller
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
-
-import akka.http.scaladsl.server.Route
+import akka.http.scaladsl.server.{AuthenticationFailedRejection, Directive1, Route}
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.model.headers.BasicHttpCredentials
 import akka.http.scaladsl.server.directives._
 import akka.http.scaladsl.server.directives.AuthenticationResult
 import akka.http.scaladsl.model.headers._
+import akka.http.scaladsl.server.AuthenticationFailedRejection.{CredentialsMissing, CredentialsRejected}
+import akka.http.scaladsl.server.directives.BasicDirectives.provide
+import akka.http.scaladsl.server.directives.FutureDirectives.onSuccess
+import akka.http.scaladsl.server.directives.RouteDirectives.reject
+import whisk.common.{PrintStreamLogging, TransactionId}
+import whisk.core.WhiskConfig
+import whisk.core.entity.{EntityName, Identity, Subject}
 
-import whisk.common.TransactionId
-import whisk.core.entity.Identity
 
 /** A common trait for secured routes */
 trait AuthenticatedRoute {
 
     /** An execution context for futures */
     protected implicit val executionContext: ExecutionContext
+
+    /*** Auth route for basic auth and certificate auth*/
+    def authenticate(implicit transid: TransactionId) = {
+        implicit val logger = new PrintStreamLogging()
+        val config = new WhiskConfig(Map("whisk.ssl.client.verification" -> null))
+        if (config.sslClientVerification == "off"){
+            basicAuth(validateCredentials)
+        }else{
+            certificateAuth(validateCertificate)
+        }
+    }
 
     /** Creates HTTP BasicAuth handler */
     def basicAuth[A](verify: Option[BasicHttpCredentials] => Future[Option[A]]) = {
@@ -46,8 +61,44 @@ trait AuthenticatedRoute {
         }
     }
 
+
+    /** Create certificate auth handler */
+    def certificateAuth[A](verify: Option[CertificateInfo] => Future[Option[A]]):AuthenticationDirective[A]  = {
+        val result = extract { ctx =>
+            val authHeader = ctx.request.headers.find(_.is("x-ssl-subject"))
+            val subjectHeader = authHeader flatMap {
+                case header =>
+                    header.value.split(",").find(_.startsWith("CN="))
+            }
+            val subject: Option[Subject] = subjectHeader map {
+                case subject => Subject(subject.substring("CN=".length, subject.length))
+            }
+            val namespaceHeader = ctx.request.headers.find(_.is("namespace"))
+            val certificateInfo: Option[CertificateInfo] = namespaceHeader map {
+                case header if header.value() == "_" => CertificateInfo(EntityName(subject.get.asString), subject.get)
+                case header if header.value() != "_" => CertificateInfo(EntityName(header.value), subject.get)
+            }
+            val certificateFuture = verify(certificateInfo).map {
+                case Some(t) => AuthenticationResult.success(t)
+                case None    => AuthenticationResult.failWithChallenge(HttpChallenges.basic("OpenWhisk secure realm"))
+            }
+            onSuccess(certificateFuture).flatMap {
+                case Right(user) ⇒ provide(user)
+                case Left(challenge) ⇒
+                    val cause = if (authHeader.isEmpty) CredentialsMissing else CredentialsRejected
+                    reject(AuthenticationFailedRejection(cause, challenge)): Directive1[A]
+            }
+        }
+        result.flatMap { user =>
+            user
+        }
+    }
+
     /** Validates credentials against database of subjects */
     protected def validateCredentials(credentials: Option[BasicHttpCredentials])(implicit transid: TransactionId): Future[Option[Identity]]
+
+    /** Validates client certificate against database of subjects */
+    protected def validateCertificate(certificateInfo: Option[CertificateInfo])(implicit transid: TransactionId): Future[Option[Identity]]
 }
 
 /** A trait for authenticated routes. */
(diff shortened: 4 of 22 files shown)


ansible/files/auth_index.json
```diff
@@ -2,7 +2,7 @@
   "_id":"_design/subjects",
   "views": {
     "identities": {
-      "map": "function (doc) {\n  if(doc.uuid && doc.key) {\n    var v = {namespace: doc.subject, uuid: doc.uuid, key: doc.key};\n    emit([doc.subject], v);\n    emit([doc.uuid, doc.key], v);\n  }\n  if(doc.namespaces) {\n    doc.namespaces.forEach(function(namespace) {\n      var v = {namespace: namespace.name, uuid: namespace.uuid, key: namespace.key};\n      emit([namespace.name], v);\n      emit([namespace.uuid, namespace.key], v);\n    });\n  }\n}"
+      "map": "function (doc) {\n  if(doc.uuid && doc.key) {\n    var v = {namespace: doc.subject, uuid: doc.uuid, key: doc.key};\n    emit([doc.subject], v);\n    emit([doc.uuid, doc.key], v);\n  }\n  if(doc.namespaces) {\n    var subject = doc.subject;\n    doc.namespaces.forEach(function(namespace) {\n      var v = {namespace: namespace.name, uuid: namespace.uuid, key: namespace.key};\n      emit([namespace.name], v);\n      emit([namespace.uuid, namespace.key], v);\n      emit([subject], v);\n    });\n  }\n}"
     }
```
rabbah: Subject is already a primary index why emit it again?

ningyougang: The function defines here 
{
   "identities": {
       "map": "function (doc) {
          if(doc.uuid && doc.key) {
            var v = {namespace: doc.subject, uuid: doc.uuid, key: doc.key};  
            emit([doc.subject], v);    
            emit([doc.uuid, doc.key], v);
          } 
          if(doc.namespaces) { 
            `var subject = doc.subject; `
            var v = {namespace: namespace.name, uuid: namespace.uuid, key: namespace.key};
            emit([namespace.name], v);
            emit([namespace.uuid, namespace.key], v);
            `emit([subject], v);`
            });
             }
            }"
   }
}
So, i just added two lines in `if(doc.namespaces)` condition. Why i add `this two lines`, the reason is that: the client certificate will pass `CN` value to controller(the `CN` value is username, it is `subject concept in openwhisk),  so it can get the `Identity` by `username` from couchdb, so should modify above view's  function

rabbah: You can do a get on the WhiskAuth store directly. The docs are indexed by the subject already. I suppose this change avoids you changing more of the controller.

ningyougang: For `You can do a get on the WhiskAuth store directly. The docs are indexed by the subject already`, How can i do? there has example, i am also finding.

ningyougang: @rabbah ,yes, you are right, after checking, the docs already are indexed by  the subject.
I have modified

ningyougang:
@rabbah , i have added the test case for client certificate auth in AuthentiateTests.scala

ningyougang:
@rabbah , this pr can be merged?

rabbah:
@mdeuser or @dubeejw can you quickly look through cli changes?

core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
```diff
@@ -79,4 +74,25 @@ trait Authenticate {
             Future.successful(None)
         }
     }
+
+  def validateCertificate(name: Option[String])(implicit transid: TransactionId): Future[Option[Identity]] = {
+      name flatMap { entityName =>
+          val name = entityName.substring("CN=".length, entityName.length)
+          Try {
+              //It also supports getting Identity by subject
+              val future = Identity.get(authStore, EntityName(name)) map { result =>
```
rabbah: I think this is only going to work for the case where the subject has a single namespace and both values match.
If the CN name is going to be a namespace - rather than a subject id, this will generally not work. In retrospect, it should have been obvious to me CN name should be a namespace. Do you agree or disagree?

core/controller/src/main/scala/whisk/core/controller/AuthenticatedRoute.scala
```diff
@@ -48,12 +49,43 @@ trait AuthenticatedRoute {
...
+    protected def validateCertificate(entityName: Option[String])(implicit transid: TransactionId): Future[Option[Identity]]
+
 }
 
 /** A trait for authenticated routes. */
 trait AuthenticatedRouteProvider {
     def routes(user: Identity)(implicit transid: TransactionId): Route
 }
+
+class CertificateAuthenticator[U](val userPassAuthenticator: Option[String] ⇒ Future[Option[U]])(implicit val executionContext: ExecutionContext) extends ContextAuthenticator[U] {
```
rabbah: does it make sense to just inline the apply method here into `validateCertificate` above?

tests/src/test/scala/system/basic/WskBasicTests.scala
```diff
@@ -88,7 +88,10 @@ class WskBasicTests
 
     it should "reject unauthenticated access" in {
         implicit val wskprops = WskProps("xxx") // shadow properties
-        val errormsg = "The supplied authentication is invalid"
+        //First it uses auth key verification, if failed, uses client certificate verification,
+        //for client certificate verification, if header doesn't include x-ssl-subject,it will
+        //throw CredentialsMissing Exception
+        val errormsg = "The resource requires authentication, which was not supplied with the request"
```
rabbah: is this a shared string - can we put it in the `Messages` singleton object?
i suspect actually this is a spray default message - in which case nevermind.

tests/src/test/scala/whisk/core/controller/test/AuthenticateTests.scala
```diff
@@ -145,7 +146,7 @@ class AuthenticatedRouteTests
         implicit val tid = transid()
         handleRejections(BasicHttpService.customRejectionHandler) {
             path("secured") {
-                authenticate(basicauth) {
+                (authenticate(basicauth) | authenticate(certificateAuth)) {
```
rabbah: i think it makes sense to test each authentication mechanism separately.

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -160,7 +160,7 @@ protected[controller] class RestAPIVersion(apipath: String, apiversion: String)(
             sendCorsHeaders {
                 (pathEndOrSingleSlash & get) {
                     complete(OK, info)
-                } ~ authenticate(basicauth) {
+                } ~ (authenticate(basicauth) | authenticate(certificateAuth)) {
```
rabbah: since the certificate based authentication is deployment dependent, perhaps this matching should be guarded by a configuration parameter - is it possible to use client authentication without enabling the flag you added in the previous pull request?

rabbah:
I wonder if the documentation for how to set up the client based authentication belongs in the same place our docs talk about ssl certificates in general - so that the CLI doc is strictly about the CLI not a deployment.

core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
ningyougang: May be you are right, i should test it again.

core/controller/src/main/scala/whisk/core/controller/AuthenticatedRoute.scala
ningyougang: You can see for `basicauth`, i just follow its sytle:
```
    protected def basicauth(implicit transid: TransactionId) = {
        new BasicHttpAuthenticator[Identity](realm = "whisk rest service", validateCredentials _) {
            override def apply(ctx: RequestContext) = {
                super.apply(ctx) recover {
                    case t: IllegalStateException => Left(CustomRejection(InternalServerError))
                    case t                        => Left(CustomRejection(ServiceUnavailable))
                }
            }
        }
    }
```

tests/src/test/scala/system/basic/WskBasicTests.scala
ningyougang: Yes, you are right, it is in `Spray-routing_2.11-1.3.4-sources.jar`'s RejectionHandler.scala
```
      val rejectionMessage = cause match {
        case CredentialsMissing  ⇒ "The resource requires authentication, which was not supplied with the request"
        case CredentialsRejected ⇒ "The supplied authentication is invalid"
      }
```
I wil add it to singleton object

rabbah: If they're spray messages I've generally left those out of the singleton object for error messages which are generally the ones our system produces - in case we want to internationalize these in the future. It's ok to leave this Spray message out as a such.

tests/src/test/scala/whisk/core/controller/test/AuthenticateTests.scala
ningyougang: yes, agree with you

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
ningyougang: * since the certificate based authentication is deployment dependent, perhaps this matching should be guarded by a configuration parameter
answer: actully, the certificate desn't depend on the deployment process, the client certificate generated process in deployment is just an examle like server certificate. as below codes
```
  - name: gen untrusted client certificate for host
    local_action: shell "{{ playbook_dir }}/roles/nginx/files/genssl.sh" "*.{{ whisk_api_localhost_name | default(whisk_api_host_name) | default(whisk_api_localhost_name_default) }}" "client"
    when: nginx.ssl.client_ca_cert == "openwhisk-client-ca-cert.pem"
```
you can see, if <nginx.ssl.client_ca_cert> doesn't equal with `openwhisk-client-ca-cert.pem`, the untrusted client certificate genderated ansible task will not be executed. so in users's own env, they just set <nginx.ssl.client_ca_cert> another name, can use their own client certificate.

* is it possible to use client authentication without enabling the flag you added in the previous pull request?
answer: it is better to use the flag <nginx.ssl.verify_client>.

tests/src/test/scala/system/basic/WskBasicTests.scala
ningyougang: So here, should i move it to singleton object? as your reply, seems no need. but not sure, so asked again.

core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
ningyougang: For openwhisk model relation, subject : namespace == 1: n, namespace : auth key = 1: 1

* Use `namesapce value` to `CN` value, for openwhisk itself,  i think it is ok. 
* But for general meaning, the client certificate like `compressed username/password`, but for here, is `namespace`, feel strange.

What do you think about it?

ningyougang:
```  
I wonder if the documentation for how to set up the client based authentication 
belongs in the same place our docs talk about ssl certificates in general 
- so that the CLI doc is strictly about the CLI not a deployment.
```
Answer: openwhisk itself doens't care how to generate client certificate(genssh.sh just an example), it just provide a entry point configuration to how to use it, so openwhisk users(for example:a company) just configure `nginx.ssl.client_ca_cert` to `their own client ca path`,  and create every client certificate(cert/key) for its employees.

So employee can download its own client certificate(cert/key) from somewhere(may be download url),
and use `wsk -i property set --cert <cert> --key <key> XXX` can access openwhisk env.

So i will improve the docs.

mdeuser:
When authenticating with just the cert, can the namespace property value still be "_"?  Possibly related to a discussion thread above..

mdeuser:
What is the expected behavior when both the authkey and cert property values are set?

core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
rabbah: The identity has to be a namespace or the rest of the system invariant all break. So what's the best way to do this?

core/controller/src/main/scala/whisk/core/controller/AuthenticatedRoute.scala
rabbah: Thanks for the explanation.

tests/src/test/scala/system/basic/WskBasicTests.scala
rabbah: You understood correctly (do not move it) - thanks for double checking.

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
rabbah: if the controller always accepts the header (even in deployments that don't have edge verification) then isn't it easier to bypass all authentication this way?

rabbah:
About the documentation in cli.md: the point is that for the CLI point of view, the user needs to know 1 command. The rest of the text is talking about configuring the deployment. I don't think this is the best place for it. There's another doc where there's discussion of certificates.

core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
ningyougang: What's the best way?

Answer: i think first should clear currently model mapping (subject:namespace=1:n, namespace:auth key=1:1), so `through auth key` can get unique Identity.
So as you said below we can also send `namespace` by Httpheader from cli to nginx to controller

```
Current auth key auth: auth key ------------>unique Identity object
client certificate auth : namespace(pass by Header) + subject(saved in `CN`)---->unique Identity object
```

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
ningyougang: yes, i have answered `isn't it easier to bypass all authentication this way?` on above 
```
Current auth key auth: auth key ------------>unique Identity object
client certificate auth : namespace(pass by Header) + subject(saved in `CN`)---->unique Identity object
```
Pass namespace by header, we can get unique Identity.

ningyougang:
When authenticating with just the cert, can the namespace property value still be "_"? Possibly related to a discussion thread above..

Answer: can the namespace property value still be "_"?What's mean for this? i have replied above, may be
exist your answer there.

ningyougang:
What is the expected behavior when both the authkey and cert property values are set?

answer:
if both set, if set <nginx.ssl.verify_client> to `optional` and `off`, it will only use the `--auth key` 
if both set, if set <nginx.ssl.verify_client> to `on`, it will only use the `--cert/key`

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
rabbah: The routing is an "or" route. I can supply a bad key or no key and specify any namespace and get access to it. How do you prevent that?

ningyougang: First it will use `auth key` to verify, then will failed, then use `client certificate` to verify, so `special any namespace` can't be accessed by `client certificate auth`, because it will use `namespace + subject` from Header to get a Identiy.

ningyougang:
@rabbah @mdeuser , i have modified the codes. please review.
The changes logic is that pass namespace by http header from wsk to controller, because namesapce + `CN`(it is subject) can determine a unique Identity, so it works well.

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
ningyougang: @rabbah , i have already fininshed.

common/scala/src/main/scala/whisk/core/entity/Identity.scala
```diff
@@ -74,6 +74,53 @@ object Identity extends MultipleReadersSingleWriterCache[Identity, DocInfo] with
         })
     }
 
+    /**
+      * Many subjects can be belonged to only one namespace
+      * Retrieves a key for namespace + subject.
+      *
+      */
+    def get(datastore: AuthStore, namespace: EntityName, subject: Subject)(
```
rabbah: since you have both the subject and the entity name - does it make more sense to actually look up the doc by subject id (that's a primary key) and then find the namespace in that document?

@markusthoemmes opinion?

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
rabbah: i am still concerned about this route change in the absence of a guard that couple it to the deployment.
if i deploy without the certificate checking in nginx, then i can bypass the basic auth and gain access as any other space by just guessing the subject/namespace.

common/scala/src/main/scala/whisk/core/entity/Identity.scala
ningyougang: Does it make more sense to actually look up the doc by subject id (that's a primary key) 

answer:can lookup the doc by subject?i find it doesn't exist find by subject id.
regarding `_design/subjects`, the function define as followings

{
   "identities": {
       "map": "function (doc) {
          if(doc.uuid && doc.key) {
            var v = {namespace: doc.subject, uuid: doc.uuid, key: doc.key};  
            `emit([doc.subject], v);  # actully, the key is namespace, because the subject value == namespace value. `
            emit([doc.uuid, doc.key], v);
          } 
          if(doc.namespaces) { 
            var v = {namespace: namespace.name, uuid: namespace.uuid, key: namespace.key};
            `emit([namespace.name], v); # The key is namespace`
            `emit([namespace.uuid, namespace.key], v); # The key is auth key`
            });
             }
            }"
   }
}

So you can see above function, it only supports query docs by namespace or auth key.
If can look up the doc by subject id, can you show the gist?

I first find doc query by namespace, then find the subject in that doc. it seems the solution
is similar with your solution(first find doc by subject, then find the namepsace in that doc )

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
ningyougang: @rabbah, i tested, when use api and directly acess to controller, you said above will happen.
```
$ curl -X GET --header 'Accept: application/json' \
> --header 'x-ssl-subject: CN=guest,O=OpenWhisk,L=Yorktown,ST=NY,C=US' \
> --header 'namespace: guest' \
>  'http://<my_local_ip>:10001/api/v1/namespaces/guest/actions'
[]
```
So do you have any idea how to add the guard?

common/scala/src/main/scala/whisk/core/entity/Identity.scala
ningyougang: @rabbah ,as your suggestion `look up the docs by primary key:subject`, it seems that doesn't work,
if want to query data from couchdb database, first should create a design document (for example "_design/XXXX"),then create a view.

please refer to https://stackoverflow.com/questions/28945405/how-to-retrieve-all-the-documents-in-couchdb-by-given-key

As you know, openwhiks's couchdb has already to support it, but its query key only support `[auth key]` and `[namespace]`, regarding my case here, i think it is better to query results by `namesapce`, then match the results by `subject` which is passed from http header.

rabbah: we don't need a view to look up by subject - it is a primary index.

i suspect though i did not try, that you can do get on AuthStore to fetch the subject, and parameterize that on the type of a subject record. i'll create a gist for you.

rabbah: your approach may be just as good. let me see after i gist it.

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
rabbah: curl: (6) Could not resolve host: GET; Name or service not known
did you curl work in the second case?

i’m not suggesting bypassing the controller
send the request to nginx
but configure nginx to not check the certificate
i claim the controller will respond as if you provided an auth key for the subject named in the header

ningyougang: Because nginx.conf.j2 has below configuration
```
 proxy_set_header X-SSL-Subject $ssl_client_s_dn;
```
It will protect passing `x-ssl-subject` http header from nginx to controller which means the controller can't receive the `x-ssl-subject` from nginx.

so if use blow curl cmd, the controller can't receive the `x-ssl-subject` http header
```
curl -v -X  GET \
--insecure \
--header 'Accept: application/json' \
--header 'X-SSL-Subject: CN=guest,O=OpenWhisk,L=Yorktown,ST=NY,C=US' \
--header 'namespace: guest' \
https://<nginx>/api/v1/namespaces/guest/actions
There was an internal server error.
```
So  i think current codes is ok

markusthoemmes: That tightly couples the security of the Controller to having an nginx with that exact configuration running in front of it. I don't think that is desirable. Also, when you deploy a system using cert authentication, you'll always have the overhead of also checking basic auth unnecessarily.

rabbah: Do you think the SPI approach form @tysonnorris will help here? The auth could be made into a plugin that's specific to a deployment?

tysonnorris: This SPI approach would help expose a specific combination of auth mechanisms, but currently the SPI notion only supports "one impl per interface"; It could change to support "multiple impls per interface", but in the meantime, it is probably good enough for this case (unless people want to enable BOTH basic and cert auth). 

I would assume that support "multiple auth impls" would be a low priority for now, but it could be a problem if a deployment wanted to support many different oauth logins. (even then, that could be implemented as a multi-oauth-provider impl of the auth SPI, so may not be a real problem)

rabbah:
@ningyougang can you verify that you have an Apache SLA on file?

ningyougang:
@rabbah What's mean for `do you have an Apache SLA already on file`?
I use `google` to search it, seems mean: `Service-Level Agreement`
The `Apache SLA` has relation with this PR?

rabbah:
Oops. I meant CLA https://github.com/apache/incubator-openwhisk/blob/master/CONTRIBUTING.md

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
ningyougang: @markusthoemmes @rabbah 
* That tightly couples the security of the Controller to having an nginx with that exact configuration running in front of it.
answer: i have modified, if `ssl_verify_client` in nginx.conf is `off`, it will use auth key verification default, if `ssl_verify_client` in nginx.conf is `on`, it will use client certificate verification.
I judge in Controller by passing the configuration to controller container's env.

* Should use SPI approach to support this feature.
I have readed the sourcecode for `SPI`, it is it is suitable for `Plugin configureale` which means the service can be replaced with another service by changing the applicaton.conf if one day found the original service is not well.  but regarding ` auth mechanisms here`, it is just `if/else logic` can solve this question.

ningyougang:
@rabbah @markusthoemmes 
Because controller's web framework has been changed from spray to akka,
So i have also modifed it. please review again  :)

ansible/roles/nginx/files/genssl.sh
```diff
@@ -56,7 +56,7 @@ else
     openssl req -new \
     -key "$SCRIPTDIR/openwhisk-client-key.pem" \
     -passin pass:$PASSWORD \
-    -subj "/C=US/ST=NY/L=Yorktown/O=OpenWhisk/CN=guest" \
+    -subj "/C=US/ST=NY/L=Yorktown/O=OpenWhisk/CN=whisk.system" \
```
rabbah: Why make this change. Is it to install the catalog? Do we need to generate two?

rabbah: Is guest usable without a cert? I think so but checking. If so then can we avoid changing the catalog scripts?

rabbah:
@jasonpet can you sanity check the catalog related changes?

ansible/roles/nginx/files/genssl.sh
ningyougang: * Why make this change. 
answer: the `guest` user has no permission on do the action : `$WSK_CLI -i --apihost "$APIHOST" package update $AUTH  --shared no "$NAMESPACE/apimgmt"`,so i should change the `guest` to `whisk.system`.
* Is it to install catalog?
answer: i added the below configuration to `ansible/group_vars/all`
```
#The catalog_cert_file and catalog_key_file is used to authenticate the openwhisk service when nginx ssl configuration is opened.
catalog_cert_file: "{{ openwhisk_home }}/ansible/roles/nginx/files/openwhisk-client-cert.pem"
catalog_key_file: "{{ openwhisk_home }}/ansible/roles/nginx/files/openwhisk-client-key.pem"
```
I just follow the below style in `ansible/group_vars/all` which will be used in `installRouiteMgmt.sh`
```
# The catalog_auth_key is used to determine the secret key to authenticate the openwhisk service.
# The value for this variable can be set to either the secret key itself or the file, which
# saves the secret key.
# By default, we take the key from {{ openwhisk_home }}/ansible/files/auth.whisk.system.
catalog_auth_key: "{{ openwhisk_home }}/ansible/files/auth.whisk.system"
```
So it is used to install standard actions by pass client certificate when use client ceriticate to auth.

* Do we need to generate two?
generate `CN=whisk.system` is enough.

ningyougang: * Is guest usable without a cert? I think so but checking.
answer: currently, change from `guest` to `whisk.system`, so there will be no exist client certificate on user `guest`

*  If so then can we avoid changing the catalog scripts?
answer: what's mean for that? i just added `client certificate auth` on `installRouteMgmt.sh`, if client certificate is opened, it will use that. if client certificate is `off`, it will use `auth key` to auth.

mdeuser:
regarding the changes to the api gw package/action installation, changes are in place to handle both case (nginx_ssl_verify_client=off, nginx_ssl_verify_client=on).  since the default is nginx_ssl_verify_client=off, i believe the api gw repo tests should work as-is.

@bjustin-ibm @jasonpet - will the catalog installation scripts need a similar change to support cert base authentication when installing the catalogs?

rabbah:
I had a long chat with @ningyougang on slack about the changes in this PR and side effects to other repositories. It will in fact need changes for the catalog installation and providers otherwise this feature can't be turned on. But we also found some inconsistencies.

1. A client certificate must be provided with CN equal to subject 
2. Every subject must have their own cert

So the installation of openwhisk supports two subjects: whisk.system and guest. It is not clear how Travis passed because the guest account is inoperable without a valid client certificate. Similarly some tests which create new subjects will not work without also addressing the certificate issue as well.

As noted this also affects the other repositories and their CI for the change to be consistent throughout a deployment. We would further need changes to the documentation which describes "wskadmin user" etc.

ningyougang:
@rabbah 
After test again in my local
1. catalog installation will be failed when turn on the client certificate.
    this question is easy to solve, but should add the logic of client certificate on `https://github.com/apache/incubator-openwhisk-catalog.git` too.
2. wskadmin
   has on influences on `wskadmin`, execute below command success when turn on the client certificate
```
[irteam@lambda-nyg bin]$ ./wskadmin user create ningyougang -ns whisk.system
eecfa0a4-b95e-4cb0-9fa3-854ae8a1432f:7zSfpkXnqQRgNqKZZ9ZYED7S8Xtw5NcTXX3viPSSPVQbUYBEcv9vd0jxqn5hYl2H
```

ningyougang:
@rabbah
I added a temp commit on this pr to `turn on the client certificate auth` to run all test cases, the travis-ci running result is failed, i will check it.

PS: currently, i turn on the `client certificate auth` using below configuration:
```
nginx.ssl.verify_client: "{{ nginx_ssl_verify_client | default('optional') }}"
```
Why i use `optional` instead of `on` temporarily?
because currently `ansible version:2.3.0.0`'s `get_url` module doesn't support passing client cert/key file.
so if use `on`, the `download CLI ansible task will be failed`: https://github.com/apache/incubator-openwhisk/blob/master/ansible/roles/cli/tasks/download_cli.yml#L5
if use `optional`, it will ignore the client cert/key file verification when cert/file don't exist.

but this has a good news that `ansible version:2.4.0.0` will support it. please refer to: https://github.com/ansible/ansible/pull/18141

if `ansible version:2.4.0.0` , we can use `on` instead of `optional`

rabbah:
@ningyougang is this a pr you're still pursuing?

ningyougang:
@rabbah , actually, i am working on `doing some benchmark test on openwhisk`,
after doing that, i will go on this PR.

rabbah:
ok going to apply wip label for now.

markusthoemmes:
@ningyougang what's the status of this PR. Still something worth having?

ningyougang:
yes, i think worth having, i will go on to finish it recently

markusthoemmes:
Closing this for now. @ningyougang OpenWhisk meanwhile introduced SPIs around authentication and authorization. You might want to take a look at those.

PR #2518, opened 2017-07-20
Move tests of --last tag to basicUsageTests

cbickel:
Move the tests of the --last flag to BasicUsageTests and invoke the action again if the last activation is not the expected one.


cbickel:
I had 5 PGs where this error did not occur again.

Issue #2516, opened 2017-07-20
Deploy kafka cluster for high-availability


devbv:
Currently, we can deploy single kafka node only with ansible script.
So I edited ansible script, whisk configs, and load balancer for deploying kafka cluster with multiple hosts.
If no one is dealing with this issue, can I make a pull-request for it?

jeremiaswerner:
@devbv that awesome and I'am looking forward to see your pull-request. 👍 from my side!

rabbah:
This is now possible. Thanks @devbv.

PR #2519, opened 2017-07-21
Deploy kafka cluster with ansible

devbv:
#2516 
I fixed ansible scripts and WhiskConfigs to deploy kafka cluster in distributed environments.

In this change,
- Each host has a zookeeper container and a kafka container. And all zookeeper containers are connected with each other. But I'm not sure that kafka should connect with all zookeepers or not, so I just make it to connect to zookeeper in same host only with docker bridge network we've used.

- Kafka client can take several hosts as bootstrap.servers argument so that client can deal with failure server.

- Replication factor of kafka can be adjusted by 'kafka.replicationFactor' in group_vars.


devbv:
Sorry, I'll fix the test case for zookeeper on next monday

ningyougang:
Nobody go on progress this PR?

devbv:
I'm gonna PR again from another branch

PR #2520, opened 2017-07-21
Prints bashauto-completion install script to STDOUT

### LINKED ISSUE #1221: Improve UX to install/setup CLI autocomplete
Today to setup autocomplete the command `wsk sdk install bashauto`

```
$ wsk sdk install bashauto
The bash auto-completion script (wsk_cli_bash_completion.sh) is installed in the curent directory.
To enable command line completion of wsk commands, source the auto completion script into your bash environment
```

creates a file that user will need to source/execute.

This doesn't seem like a good UX in IMHO.
To setup using .bashrc or .profile shell configurations it becomes cumbersome like

```
pushd $(mktemp -d) > /dev/null && wsk sdk install bashauto > /dev/null && source wsk_cli_bash_completion.sh && popd > /dev/null
```

A better approach is to print to STDOUT the content of this file
Then the setup in .bashrc or .profile becomes

```
eval "$(wsk sdk install bashauto)"
```

This issue is a requirement to STOP creating the file wsk_cli_bash_completion.sh, and instead the command `wsk sdk install bashauto` should print to STDOUT the content.

AFTER the wsk CLI is updated then a command like `eval "$(wsk sdk install bashauto)"` will work

bpoole16:
This change simply changes the output of the Cobra auto-completion script from a file to STDOUT. Originally, `wsk sdk install autobash` would create the file `wsk_cli_bash_completion.sh`, but now it will print the contents of the Cobra auto-completion script to the terminal. You can now use the command ``eval "`wsk sdk install autobash`"`` to automatically install the script for that instance of the terminal. Adding the eval command to your `.bashrc` or `.profile` will keep you from rerunning this command. Closes #1221

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -2,7 +2,7 @@
 
 OpenWhisk offers a powerful command line interface that allows complete management of all aspects of the system.
 
-## Setting up the OpenWhisk CLI 
+## Setting up the OpenWhisk CLI
 
 - Building OpenWhisk from a cloned repository results in the generation of the command line interface. The generated CLIs are located in `openwhisk/bin/`. The main CLI is located in `openwhisk/bin/wsk` that runs on the operating system, and CPU architecture on which it was built. Executables for other operating system, and CPU architectures are located in the following directories: `openwhisk/bin/mac/`, `openwhisk/bin/linux/`, `openwhisk/bin/windows/`.
 
@@ -37,6 +37,32 @@ you can run the following command from your `openwhisk` directory:
 **Tip:** The OpenWhisk CLI stores the properties set in `~/.wskprops` by default. The location of this file can be altered by setting the `WSK_CONFIG_FILE` environment variable.
 
 To verify your CLI setup, try [creating and running an action](#openwhisk-hello-world-example).
+
+### Configure auto-completion for Openwhisk CLI
+
+For most bash users, running the following command will be sufficient to install auto-completion for Openwhisk CLI:
+
+```
+eval "`wsk sdk install bashauto --bashrc`"
+```
+
+**Note:** `--bashrc` flag will attempt to locate your `.bashrc` and append the previous command, without `--bashrc`, to it.
+
+If this doesn't work, you can run a temporary installation using the following command:
+
+```
+eval "`wsk sdk install bashauto`"
+```
+
+**Note:** Every time a new terminal is opened this command will have to be ran in order to use auto-completion. Alternatively, adding the previous command to the `.bashrc` or `.profile` will prevent this.
+
+For those who wish to do a custom installation, you can simply print to the terminal and manipulate the output by using the following command:
+
+```
+wsk sdk install bashauto
+```
+
+**Note:** This will NOT install anything but simply print the Cobra auto-completion script to the terminal. Adding `--bashrc` will ONLY attempt to append the `.bashrc` and still not run the install script.
 
 ## Using the OpenWhisk CLI
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -38,31 +38,29 @@ you can run the following command from your `openwhisk` directory:
 
 To verify your CLI setup, try [creating and running an action](#openwhisk-hello-world-example).
 
-### Configure auto-completion for Openwhisk CLI
+### Configure command completion for Openwhisk CLI
 
-For most bash users, running the following command will be sufficient to install auto-completion for Openwhisk CLI:
+For bash command completion to work, bash 4.1 or newer is required. The most recent Linux distributions should have the correct version of bash but Mac users will most likely have an older version.
+Mac users can check their bash version and update it by running the following commands:
 
 ```
-eval "`wsk sdk install bashauto --bashrc`"
+bash --version
+brew install bash-completion
 ```
 
-**Note:** `--bashrc` flag will attempt to locate your `.bashrc` and append the previous command, without `--bashrc`, to it.
-
-If this doesn't work, you can run a temporary installation using the following command:
+To install bash command completion run the following command:
 
 ```
 eval "`wsk sdk install bashauto`"
 ```
 
-**Note:** Every time a new terminal is opened this command will have to be ran in order to use auto-completion. Alternatively, adding the previous command to the `.bashrc` or `.profile` will prevent this.
+**Note:** Every time a new terminal is opened this command must run to enable bash command completion. Alternatively, adding the previous command to the `.bashrc` or `.profile` will prevent this.
 
-For those who wish to do a custom installation, you can simply print to the terminal and manipulate the output by using the following command:
-
+For those who wish to do a custom installation, the following commands show how to output the wsk command completion script into a separate shell script file to be run later. You will need to explicitly run the script to install wsk command completion support.
 ```
-wsk sdk install bashauto
+wsk sdk install bashauto > wsk_cli_bash_completion.sh
+chmod +x wsk_cli_bash_completion.sh
 ```
-
-**Note:** This will NOT install anything but simply print the Cobra auto-completion script to the terminal. Adding `--bashrc` will ONLY attempt to append the `.bashrc` and still not run the install script.
 
 ## Using the OpenWhisk CLI
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -48,19 +48,24 @@ bash --version
 brew install bash-completion
 ```
 
-To install bash command completion run the following command:
+To download the bash command completion to your local directory, run the following command:
 
 ```
-eval "`wsk sdk install bashauto`"
+wsk sdk install bashauto
+```
+The command script `wsk_cli_bash_completion.sh` will now be in your current directory. To enable command line completion of wsk commands, source the auto completion script into your bash environment.
+
+```
+source wsk_cli_bash_completion.sh
 ```
 
-**Note:** Every time a new terminal is opened this command must run to enable bash command completion. Alternatively, adding the previous command to the `.bashrc` or `.profile` will prevent this.
+Alternatively, to install bash command completion, run the following command:
 
-For those who wish to do a custom installation, the following commands show how to output the wsk command completion script into a separate shell script file to be run later. You will need to explicitly run the script to install wsk command completion support.
 ```
-wsk sdk install bashauto > wsk_cli_bash_completion.sh
-chmod +x wsk_cli_bash_completion.sh
+eval "`wsk sdk install bashauto --stdout`"
 ```
+
+**Note:** Every time a new terminal is opened, this command must run to enable bash command completion. Alternatively, adding the previous command to the `.bashrc` or `.profile` will prevent this.
 
 ## Using the OpenWhisk CLI
 
--- a/tests/src/test/scala/system/basic/WskSdkTests.scala
+++ b/tests/src/test/scala/system/basic/WskSdkTests.scala
@@ -89,23 +89,14 @@ class WskSdkTests
         FileUtils.deleteDirectory(dir)
     }
 
-    it should "install the bash auto-completion bash script" in {
-        // Use a temp dir for testing to not disturb user's local folder
-        val dir = File.createTempFile("wskinstall", ".tmp")
-        dir.delete()
-        dir.mkdir() should be(true)
 
-        val scriptfilename = "wsk_cli_bash_completion.sh"
-        var scriptfile = new File(dir.getPath(), scriptfilename)
-        try {
-            val stdout = wsk.cli(Seq("sdk", "install", "bashauto"), workingDir = dir, expectedExitCode = SUCCESS_EXIT).stdout
-            stdout should include("is installed in the current directory")
-            val fileContent = FileUtils.readFileToString(scriptfile)
-            fileContent should include("bash completion for wsk")
-        } finally {
-            scriptfile.delete()
-            FileUtils.deleteDirectory(dir)
-        }
+    it should "install the bash auto-completion bash script using --bashrc flag" in {
+        val auth: Seq[String] = Seq("--auth", wskprops.authKey)
+        val msg = "bash completion for wsk"    // Subject to change, dependent on Cobra script
+        // Doesn't actually install, simply checking if the command is printing correctly to STDOUT
+        val stdout = wsk.cli(Seq("sdk", "install", "bashauto", "--bashrc") ++ wskprops.overrides ++ auth, expectedExitCode = SUCCESS_EXIT).stdout
+
+        stdout should include(msg)
     }
 
 }
--- a/tests/src/test/scala/system/basic/WskSdkTests.scala
+++ b/tests/src/test/scala/system/basic/WskSdkTests.scala
@@ -89,10 +89,29 @@ class WskSdkTests
         FileUtils.deleteDirectory(dir)
     }
 
+    it should "install the bash auto-completion bash script" in {
+        // Use a temp dir for testing to not disturb user's local folder
+        val dir = File.createTempFile("wskinstall", ".tmp")
+        dir.delete()
+        dir.mkdir() should be(true)
+
+        val scriptfilename = "wsk_cli_bash_completion.sh"
+        var scriptfile = new File(dir.getPath(), scriptfilename)
+        try {
+            val stdout = wsk.cli(Seq("sdk", "install", "bashauto"), workingDir = dir, expectedExitCode = SUCCESS_EXIT).stdout
+            stdout should include("is installed in the current directory")
+            val fileContent = FileUtils.readFileToString(scriptfile)
+            fileContent should include("bash completion for wsk")
+        } finally {
+            scriptfile.delete()
+            FileUtils.deleteDirectory(dir)
+        }
+    }
+
     it should "print bash command completion script to STDOUT" in {
         val msg = "bash completion for wsk"    // Subject to change, dependent on Cobra script
 
-        val stdout = wsk.cli(Seq("sdk", "install", "bashauto"), expectedExitCode = SUCCESS_EXIT).stdout
+        val stdout = wsk.cli(Seq("sdk", "install", "bashauto", "--stdout")).stdout
         stdout should include(msg)
     }
 }
--- a/tests/src/test/scala/system/basic/WskSdkTests.scala
+++ b/tests/src/test/scala/system/basic/WskSdkTests.scala
@@ -89,14 +89,10 @@ class WskSdkTests
         FileUtils.deleteDirectory(dir)
     }
 
-
-    it should "install the bash auto-completion bash script using --bashrc flag" in {
-        val auth: Seq[String] = Seq("--auth", wskprops.authKey)
+    it should "print bash command completion script to STDOUT" in {
         val msg = "bash completion for wsk"    // Subject to change, dependent on Cobra script
-        // Doesn't actually install, simply checking if the command is printing correctly to STDOUT
-        val stdout = wsk.cli(Seq("sdk", "install", "bashauto", "--bashrc") ++ wskprops.overrides ++ auth, expectedExitCode = SUCCESS_EXIT).stdout
 
+        val stdout = wsk.cli(Seq("sdk", "install", "bashauto"), expectedExitCode = SUCCESS_EXIT).stdout
         stdout should include(msg)
     }
-
 }
--- a/tools/cli/go-whisk-cli/commands/flags.go
+++ b/tools/cli/go-whisk-cli/commands/flags.go
@@ -106,6 +106,11 @@ type Flags struct {
         summary bool
     }
 
+    //sdk
+    sdk struct {
+        stdout bool
+    }
+
     // api
     api struct {
         action     string
--- a/tools/cli/go-whisk-cli/commands/flags.go
+++ b/tools/cli/go-whisk-cli/commands/flags.go
@@ -116,11 +116,6 @@ type Flags struct {
         configfile string
         resptype   string
     }
-
-    // sdk
-    sdk struct {
-        bashrc bool
-    }
 }
 
 
--- a/tools/cli/go-whisk-cli/commands/flags.go
+++ b/tools/cli/go-whisk-cli/commands/flags.go
@@ -116,6 +116,11 @@ type Flags struct {
         configfile string
         resptype   string
     }
+
+    // sdk
+    sdk struct {
+        bashrc bool
+    }
 }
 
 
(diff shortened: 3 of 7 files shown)


tools/ubuntu-setup/bashprofile.sh
```diff
@@ -1,5 +1,4 @@
 # Adds openwhisk bin to bash profile
 echo 'export PATH=$HOME/openwhisk/bin:$PATH' > "$HOME/.bash_profile"
 # Adds tab completion
-echo 'eval "$(register-python-argcomplete wsk)"' >> "$HOME/.bash_profile"
```
bpoole16: I believe this was referenced to be removed in the original issue. #1221

tools/vagrant/Vagrantfile
```diff
@@ -155,7 +155,6 @@ $script_end = <<SCRIPT
   chmod +x ${HOME}/bin/wsk
   chown vagrant:vagrant ${HOME}/bin/wsk
   PATH=${PATH}:${HOME}/bin
-  echo 'pushd "$(mktemp -d)" > /dev/null && wsk sdk install bashauto > /dev/null && source wsk_cli_bash_completion.sh && popd > /dev/null' >> $HOME/.bashrc
```
bpoole16: I believe this was referenced to be removed in the original issue. #1221

tools/cli/go-whisk-cli/commands/sdk.go
```diff
@@ -99,6 +106,58 @@ var sdkInstallCmd = &cobra.Command{
...
+        }
+        whisk.Debug(whisk.DbgInfo, "Successfully appended .bashrc\n")
+    }
+
+    return nil
+}
+
+// lineExists(string, string) checks to see if a string is contained in a specified file
+// Returns a boolean (true meaning the line exists) and any errors that
+func lineExists(fileLoc string , text string) (bool, error) {
```
bpoole16: This function might be useful in other areas - utils.go might be a better place for it.

tools/cli/go-whisk-cli/commands/sdk.go
```diff
@@ -99,6 +106,58 @@ var sdkInstallCmd = &cobra.Command{
...
+        }
+        whisk.Debug(whisk.DbgInfo, "Successfully appended .bashrc\n")
+    }
+
+    return nil
+}
+
+// lineExists(string, string) checks to see if a string is contained in a specified file
+// Returns a boolean (true meaning the line exists) and any errors
+func lineExists(fileLoc string , text string) (bool, error) {
```
bpoole16: This function might be useful in other areas - utils.go might be a better place for it.

bpoole16:
PG #911 Passed

rabbah:
it appears we haven't documented the tab completion and i was surprised several of our users didn't know about it - i suggest adding to the cli docs as part of this fix.

https://github.com/apache/incubator-openwhisk/issues/1086#issuecomment-245281821

bpoole16:
@rabbah @mdeuser @dubeejw I have updated the docs to include auto-completion, please let me know if you have any feedback regarding them.

tools/cli/go-whisk-cli/commands/sdk.go
```diff
@@ -238,6 +297,8 @@ func sdkInstall(componentName string) error {
 }
 
 func init() {
+	sdkInstallCmd.Flags().BoolVarP(&flags.sdk.bashrc, "bashrc", "b", false, wski18n.T("adds bashauto-completion install command to .bashrc"))
```
mdeuser: same comment re: bash command completion naming

tools/cli/go-whisk-cli/wski18n/resources/en_US.all.json
```diff
@@ -1518,5 +1510,17 @@
   {
     "id": "get action url",
     "translation": "get action url"
+  },
+  {
+    "id": "adds bashauto-completion install command to .bashrc",
```
mdeuser: `updates .bashrc with a command to install the bash command completion`

tools/cli/go-whisk-cli/commands/sdk.go
```diff
@@ -74,17 +75,23 @@ var sdkInstallCmd = &cobra.Command{
...
         case "bashauto":
-            err = WskCmd.GenBashCompletionFile(BASH_AUTOCOMPLETE_FILENAME)
-            if (err != nil) {
-                whisk.Debug(whisk.DbgError, "GenBashCompletionFile('%s`) error: \n", BASH_AUTOCOMPLETE_FILENAME, err)
-                errStr := wski18n.T("Unable to generate '{{.name}}': {{.err}}",
-                        map[string]interface{}{"name": BASH_AUTOCOMPLETE_FILENAME, "err": err})
+            whisk.Debug(whisk.DbgInfo, "Running bashauto script\n")
+            if err = WskCmd.GenBashCompletion(os.Stdout); err != nil {
+                whisk.Debug(whisk.DbgError, "GenBashCompletion error: %s\n", err)
+                errStr := wski18n.T("Unable to generate bashauto-completion {{.err}}",
```
mdeuser: trying to maintain consistent wording with docs.... `unable to output...` (use `output` instead of `generate`)

tools/ubuntu-setup/bashprofile.sh
mdeuser: i'm pretty sure this can be removed.. @rabbah

tools/vagrant/Vagrantfile
mdeuser: replace this with the good old `eval wsk sdk install bashauto` command

tests/src/test/scala/system/basic/WskSdkTests.scala
```diff
@@ -89,23 +89,14 @@ class WskSdkTests
...
-            scriptfile.delete()
-            FileUtils.deleteDirectory(dir)
-        }
+    it should "install the bash auto-completion bash script using --bashrc flag" in {
+        val auth: Seq[String] = Seq("--auth", wskprops.authKey)
+        val msg = "bash completion for wsk"    // Subject to change, dependent on Cobra script
+        // Doesn't actually install, simply checking if the command is printing correctly to STDOUT
+        val stdout = wsk.cli(Seq("sdk", "install", "bashauto", "--bashrc") ++ wskprops.overrides ++ auth, expectedExitCode = SUCCESS_EXIT).stdout
+
+        stdout should include(msg)
```
rabbah: Is the test going to modify the users ~/.bashrc?

bpoole16: I believe it will which is why I wasn't sure if I should even be testing it with the flag.

docs/cli.md
```diff
@@ -38,6 +38,41 @@ you can run the following command from your `openwhisk` directory:
...
+```
+eval "`wsk sdk install bashauto --bashrc`"
+```
+
+**Note:** `--bashrc` flag will attempt to locate your `.bashrc` and append the previous command, without `--bashrc`, to it.
+
+If this doesn't work, you can run a temporary installation using the following command:
+
+```
+eval "`wsk sdk install bashauto`"
```
bpoole16: This was the only way I could get eval to work correctly.

tests/src/test/scala/system/basic/WskSdkTests.scala
mdeuser: maybe save/restore ~/.bashrc?  or remove test.. ?

docs/cli.md
mdeuser: ok

tests/src/test/scala/system/basic/WskSdkTests.scala
bpoole16: I could replace the test to just make sure it prints correctly to stdout and to a file if you want.

mdeuser: test to stdout should be added whether or not the .bashrc test stays or goes :-)

bpoole16: Okay, sounds good to me.

tools/cli/go-whisk-cli/commands/sdk.go
```diff
@@ -99,6 +106,58 @@ var sdkInstallCmd = &cobra.Command{
...
+
+    bashCheck := "wsk sdk install bashauto"
+    bashCmd := "eval \"`wsk sdk install bashauto`\""
+
+    if usr, err = user.Current(); err != nil {
+        whisk.Debug(whisk.DbgError, "Couldn't find User Directory: %s\n", err)
+        return err
+    }
+    whisk.Debug(whisk.DbgInfo, "Found current user %s\n", usr.HomeDir)
+    bashLoc := fmt.Sprintf("%s/%s", usr.HomeDir, ".bashrc")
```
bpoole16: Can you provide any insight into this command `homedir.Expand(".bashrc")`. I have only been able to get it to work by changing it to the following `homedir.Expand("~/.bashrc")`.

mdeuser: you have it.  my typing was from memory, which was poor ;-)

bpoole16: Ah okay, just wanted to double check.

bpoole16:
PG 4 625 Passed

tests/src/test/scala/system/basic/WskSdkTests.scala
```diff
@@ -89,23 +90,40 @@ class WskSdkTests
...
+
         try {
-            val stdout = wsk.cli(Seq("sdk", "install", "bashauto"), workingDir = dir, expectedExitCode = SUCCESS_EXIT).stdout
-            stdout should include("is installed in the current directory")
-            val fileContent = FileUtils.readFileToString(scriptfile)
-            fileContent should include("bash completion for wsk")
+            if (bashrc.exists()) {
+                println("Found bashrc")
+                FileUtils.copyFile(bashrc, backUp)
+                val stdout = wsk.cli(Seq("sdk", "install", "bashauto", "--bashrc") ++ wskprops.overrides ++ auth, expectedExitCode = SUCCESS_EXIT).stdout
```
markusthoemmes: Are `auth` and other overrides needed here?

tests/src/test/scala/system/basic/WskSdkTests.scala
```diff
@@ -89,23 +90,40 @@ class WskSdkTests
...
+        val homeDir = System.getProperty("user.home")
+        val bashrc = new File(homeDir, ".bashrc")
         val dir = File.createTempFile("wskinstall", ".tmp")
         dir.delete()
         dir.mkdir() should be(true)
 
-        val scriptfilename = "wsk_cli_bash_completion.sh"
-        var scriptfile = new File(dir.getPath(), scriptfilename)
+        val backUp = File.createTempFile("backup",".tmp", dir)
+        backUp.exists() should be(true)
```
markusthoemmes: Want to move the creation of temporary files inside the `if` clause and have a `try/finally` in there to clean up?

tests/src/test/scala/system/basic/WskSdkTests.scala
```diff
@@ -89,23 +90,40 @@ class WskSdkTests
...
-            val stdout = wsk.cli(Seq("sdk", "install", "bashauto"), workingDir = dir, expectedExitCode = SUCCESS_EXIT).stdout
-            stdout should include("is installed in the current directory")
-            val fileContent = FileUtils.readFileToString(scriptfile)
-            fileContent should include("bash completion for wsk")
+            if (bashrc.exists()) {
+                println("Found bashrc")
+                FileUtils.copyFile(bashrc, backUp)
+                val stdout = wsk.cli(Seq("sdk", "install", "bashauto", "--bashrc") ++ wskprops.overrides ++ auth, expectedExitCode = SUCCESS_EXIT).stdout
+                val fileOutput = FileUtils.readLines(bashrc)
+                stdout should include("bash completion for wsk" )    // Subject to change, dependent on Cobra script
```
mdeuser: hmmm.. i'm wondering why this line would be output when the `--bashrc` flag is used.  i'm thinking cobra won't be involved here since all this command is doing is amending an existing .bashrc file and not actually running the cobra generated command completion script.

tests/src/test/scala/system/basic/WskSdkTests.scala
```diff
@@ -89,23 +90,40 @@ class WskSdkTests
...
+                fileOutput.contains("eval \"`wsk sdk install bashauto`\"") should be(true)
+            } else {
+                println("Didn't find bashrc")
+                bashrc.exists() should be(false)
```
mdeuser: Is this line needed?

rabbah:
the test shouldn't touch the bashrc on the local file system - is it? is there another way to test this?

tests/src/test/scala/system/basic/WskSdkTests.scala
bpoole16: Correct, it is not needed and was left over from something I was testing previously.

mdeuser:
the test is saving/restoring the original .bashrc, but perhaps that's not enough?

one other thought is for this flag, although renamed, to take an argument as the path to some file to amend with the `eval`.  it could be ~/.bashrc, ~/.profile, or even the system .bashrc

tests/src/test/scala/system/basic/WskSdkTests.scala
bpoole16: No it can be removed, I added it to be able to see when it was not finding the bashrc.

rabbah:
i'd rather avoid messing with something the the shell profile - perhaps the flag has a default file to update and this can be overridden by the command line... i dont have a .bashrc, it wouldn't work for me for example anyway and i think these are usually tailored by end users we shouldn't assume where to stuff the code.

tests/src/test/scala/system/basic/WskSdkTests.scala
bpoole16: Flag `--bashrc` appends the `.bashrc` and runs the script. This is so then you don't have to source your bash after.

bpoole16:
From talking to @csantanapr , we decided to add the `--bashrc` flag to append the `.bashrc` knowing it wouldn't work for everyone. The original idea was just to allow people to capture the STDOUT but just to make things easier for people we added `--bashrc` along side it.

rabbah:
How about -o filepath.

bpoole16:
Sounds like a good idea to me.

tests/src/test/scala/system/basic/WskSdkTests.scala
bpoole16: Honestly, I could probably remove the rest "print bash command completion script to STDOUT" seen as the new test checks it too.

mdeuser: thinking..  it seems like this flag is overloaded.  when this flag is not used, the script is sent to stdout and not sourced.  there's no flag to immediately `eval` the output.  when this flag is set, the target file is amended with the `eval` command; so immediate `eval` is somewhat unexpected...

bpoole16: I should clarify, `--bashrc` paired with eval will append the bashrc and run the script. If you just run `wsk sdk install bashauto --bashrc` it will just append the bashrc.

csantanapr:
For this PR let's just print to STDOUT and no flag involved.
Will see what feedback we get.
Add to the docs if people want to write the script to a file they can use `wsk sdk install bashauto > wsk_cli_bash_completion.sh`
This will be a step forward.

bpoole16:
@mdeuser @csantanapr Updated to have no flag.

docs/cli.md
```diff
@@ -38,6 +38,31 @@ you can run the following command from your `openwhisk` directory:
 
 To verify your CLI setup, try [creating and running an action](#openwhisk-hello-world-example).
 
+### Configure bash command completion for Openwhisk CLI
+
+For bash command completio to work, bash 4.1 or newer is required. The most recent Linux distributions should have the correct version of bash but Mac users will most likely have an older version.
```
dubeejw: `command completion`

bpoole16:
PG 4 628 Passed

dubeejw:
Not sure I like or see the point of `wsk sdk install bashauto` printing the auto complete script to the terminal.

dubeejw:
I still do not like that as soon as I close my terminal, the auto-completion will be removed.

rabbah:
lol - ok now i'm going to ask: what precipitated these changes - as in what's wrong with the current approach?

dubeejw:
Let's look how git does it. https://git-scm.com/book/en/v1/Git-Basics-Tips-and-Tricks

dubeejw:
Inclined to say that `wsk sdk install autobash` should work as it currently does. Meaning it saves the auto-complete script in the current directory. Whereas if we want to use eval there should be a special flag `eval "`wsk sdk install autobash --stdout`"`. Not sure why anyone would ever want to do an eval as the auto-completion will vanish as soon as my terminal is closed.

In order to make autocomplete work on mac whenever the terminal is opened, this is where the auto-complete script needs to live in `/usr/local/etc/bash_completion.d/wsk_cli_bash_completion.d`. We can provide a flag that says to save the script in such directories.

dubeejw:
@csantanapr, @mdeuser, see the above comment.

rabbah:
this is a _one_ time setup - while i haven't followed closely +1 to @dubeejw's points - feels like over engineering this.

dubeejw:
If want want to provide the eval option, why not `eval "source wsk_cli_bash_completion.sh"` instead of the entire content of the file?

bpoole16:
If you want to work in every terminal don't you just put it in your .bashrc or .profile? Wouldn't you have to do the same with the script or am I lacking a piece of information? I thought the goal was so you don't have to put this huge line in your .bashrc `pushd $(mktemp -d) > /dev/null && wsk sdk install bashauto > /dev/null && source wsk_cli_bash_completion.sh && popd > /dev/null`.

rabbah:
The bash profile doesn't need to generate the file repreardly. You create the file once. You put it in your bash profiles directly and then you just source it. That's more common in my experience.

bpoole16:
Ah okay.

mdeuser:
i like the idea of keeping the existing behavior with a `--stdout` flag that outputs to stdout.  if the cli implements a self-updating feature, then using `eval 'wsk sdk install bashauto --stdout'` in the .profile or similar would always provide the latest command support.

dubeejw:
Please help me understand the difference between `source wsk_cli_bash_completion.sh` and `pushd $(mktemp -d) > /dev/null && wsk sdk install bashauto > /dev/null && source wsk_cli_bash_completion.sh && popd > /dev/null`. @csantanapr @bpoole16

tests/src/test/scala/system/basic/WskSdkTests.scala
```diff
@@ -108,4 +108,10 @@ class WskSdkTests
         }
     }
 
+    it should "print bash command completion script to STDOUT" in {
+        val msg = "bash completion for wsk"    // Subject to change, dependent on Cobra script
+
+        val stdout = wsk.cli(Seq("sdk", "install", "bashauto", "--stdout"), expectedExitCode = SUCCESS_EXIT).stdout
```
dubeejw: Remove `expectedExitCode = SUCCESS_EXIT`

tools/cli/go-whisk-cli/commands/flags.go
```diff
@@ -105,6 +105,10 @@ type Flags struct {
     trigger struct {
         summary bool
     }
+    //sdk
```
dubeejw: Add new line before `//sdk` consistency.

tools/cli/go-whisk-cli/commands/sdk.go
```diff
@@ -74,17 +74,27 @@ var sdkInstallCmd = &cobra.Command{
...
-                whisk.Debug(whisk.DbgError, "GenBashCompletionFile('%s`) error: \n", BASH_AUTOCOMPLETE_FILENAME, err)
-                errStr := wski18n.T("Unable to generate '{{.name}}': {{.err}}",
-                        map[string]interface{}{"name": BASH_AUTOCOMPLETE_FILENAME, "err": err})
-                werr := whisk.MakeWskError(errors.New(errStr), whisk.EXITCODE_ERR_GENERAL, whisk.DISPLAY_MSG, whisk.NO_DISPLAY_USAGE)
-                return werr
-            }
-            fmt.Printf(
-                wski18n.T("bash_completion_msg",
+            if flags.sdk.stdout {
+                if err = WskCmd.GenBashCompletion(os.Stdout); err != nil {
```
dubeejw: Seems like this can be a function that you pass the output stream/file to instead of duplicating the code twice.

dubeejw: Wouldn't worry about having two different error output messages.

docs/cli.md
```diff
@@ -38,6 +38,41 @@ you can run the following command from your `openwhisk` directory:
...
+Alternatively, to install bash command completion run the following command:
+
+```
+eval "`wsk sdk install bashauto --stdout`"
+```
+
+**Note:** Every time a new terminal is opened this command must run to enable bash command completion. Alternatively, adding the previous command to the `.bashrc` or `.profile` will prevent this.
+
+For those who wish to do a custom installation, the following commands show how to output the wsk command completion script into a separate shell script file to be run later. You will need to explicitly run the script to install wsk command completion support.
+```
```
dubeejw: New line before the three tick marks.

csantanapr:
I'm OK with having a flag to print to stdout and leaving the current behavior of producing a file.
This will allow me to put in my `$HOME/.profile` the following command
```
eval "$(wsk sdk install bashauto --stdout)"
```
This way every time I open a new terminal my autocomplete is up to date, and the script doesn't become stale. Meaning if we add flags or commands which we do very often then I don't have to worry about who a certain command is not working with autocomplete.

This is just my use case, other people might have other ways of dealing on keeping the bash script up to date and that's OK also.

tests/src/test/scala/system/basic/WskSdkTests.scala
dubeejw: `SUCCES_EXIT` is the default expected exit code.

tools/cli/go-whisk-cli/commands/sdk.go
dubeejw: Alternatively set an string that is passed to the `GenBashCompletion` command based on whether the flag is provided or not. And only have one call `GenBashCompletion`.

bpoole16:
PG 4 633 passed

tools/cli/go-whisk-cli/commands/sdk.go
bpoole16: One uses `GenBashCompletion` the other uses `GenBashCompletionFile`.

csantanapr:
pg 4/633 🔵

PR #2522, opened 2017-07-25
Reuse namespace in the Python 3 action runtime

bjustin-ibm:
This allows actions to set/retrieve global variables as an additional optimization during container reuse.


bjustin-ibm:
🔵 PG1 1847

markusthoemmes:
Considered adding a test?

bjustin-ibm:
> looks good, namespace becomes the global object carried across activations (maybe it should be called something else?).

How about... `global_context`?

bjustin-ibm:
🏃 PG1 1853

PR #2452, opened 2017-07-01
Nuke Consul.

rabbah:
A quick attempt to strip consul - mostly because I don't want to understand why it doesn't deploy with docker-for-mac.

Need to adjust all the deployment parameters (maybe cherry picking from https://github.com/apache/incubator-openwhisk/pull/2277 and merging PRs is the way to go.

Related to #2254..

@ddragosd FYI.

PR #2440, opened 2017-06-28
Use docker API via unix-socket in /var/run/docker.sock

csantanapr:
This changes avoids the need to setup proxy for docker remote API.
docker API should always be available via /var/run/docker.sock, unless in docker-machine

PR #2524, opened 2017-07-25
Docker for mac support

lionelvillard:
This change is a continuation of PR #2440 and PR #2452


rabbah:
worked for me!

@csantanapr @style95 FYI.
@ddragosd your wish is closer to becoming true.

rabbah:
Without the nuke consul PR I failed at the consul deployment phase which seems to require sudo - not going to bother debugging this and instead will wait for that PR to be merged.

rabbah:
@lionelvillard do we need to add this to the docs https://github.com/apache/incubator-openwhisk/pull/2440#issuecomment-312435272?

lionelvillard:
@rabbah I added this to the doc: https://github.com/apache/incubator-openwhisk/pull/2524/commits/0ff4400512cda733b0397e5f6f5771cb177ce999#diff-5a5d069dcad99a9f8109ac77d1b9860cR258

I think that's the proper way when using `brew`

ansible/README.md
```diff
@@ -27,19 +27,6 @@ You need to depends on the workarounds until Docker provides official methods.
...
-Currently however, Docker for Mac does not support such a feature.
-
-There are many workarounds for this.
-One way is to use `socat` command to setup proxy for the UNIX socket.
-
-```
-docker run -d -v /var/run/docker.sock:/var/run/docker.sock -p 4243:2375 bobrik/socat TCP4-LISTEN:2375,fork,reuseaddr UNIX-CONNECT:/var/run/docker.sock
-```
-
-If you want to deploy OpenWhisk in distributed environment, you are required to enable Docker remote API in all Mac hosts.
```
csantanapr: Also nuke the section **Setup proxy container to run unit tests (optional)**
No need it anymore

csantanapr:
@rabbah I just tested this PR and it worked 🎉 🎉 🎉 🎉 🎉

ansible/README.md
style95: With these changes, could we communicate with action container using container IP such as `172.17.0.10` ?

rabbah:
@csantanapr there are still references to port 4243 after this pr; is the goal to remove all those?

rabbah:
i cannot not run the container tests yes - no proxy deployed @csantanapr @style95

rabbah:
having a couple of sudo hiccups that need investigating:

```
TASK [prereq : install requests] ******************************************************************************************************************************************************************************************************************************************************************
Thursday 27 July 2017  12:25:01 -0400 (0:00:00.022)       0:00:01.242 ********* 
fatal: [172.17.0.1]: FAILED! => {"changed": false, "failed": true, "module_stderr": "sudo: a password is required\n", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 1}
```

```
TASK [controller : remove controller log directory] ***********************************************************************************************************************************************************************************************************************************************
Thursday 27 July 2017  12:39:10 -0400 (0:00:00.254)       0:00:02.178 ********* 
fatal: [172.17.0.1]: FAILED! => {"changed": false, "failed": true, "module_stderr": "sudo: a password is required\n", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 1}
```

rabbah:
@jeremiaswerner @markusthoemmes why do we have ansible privilege escalation? https://github.com/apache/incubator-openwhisk/search?utf8=%E2%9C%93&q=become+&type=

i'd like to remove it all (if it's truly needed for some deployments, then i can guard it so it's not enable for local development)

rabbah:
git blames for the becomes point to many fingers - i'm wondering if we really need these.
perhaps in some environments it might be required, in which case i'd exempt all local deployments if possible.

ansible/README.md
rabbah: @style95 is correct, the proxy (or some other solution) is still required.

csantanapr:
@rabbah yeah if possible I would eliminate the need to have docker engine expose via tcp on port 4243, this is not the default way docker is install and causing problems with people running locally.

port 4243 would be required for docker-machine only.

For the tests using 4243 we could look at them to see if they can a property from whisk.properties or an environment variable to indicate to use the flag to use port 4243, if not then assume docker-cli uses the unix-socket device.

ansible/README.md
csantanapr: sorry, about that leave the proxy info as required for testing for now :-)
We should look into our tests to see how it can be optional, on docker for mac the docker host is not need it and we should adjust things like this:
https://github.com/apache/incubator-openwhisk/blob/master/tests/src/test/scala/actionContainers/ActionContainer.scala#L89
```
    private lazy val dockerCmd: String = {
        val hostStr = if (WhiskProperties.onMacOSX()) {
            s" --host tcp://${WhiskProperties.getMainDockerEndpoint()} "
        } else {
            " "
        }
        s"$dockerBin $hostStr"
    }
```

rabbah: ideally this goes away but need either the proxy to tap... no free lunch yet.

csantanapr: Also like this ping test, I think this tests are now move to ansible, this was required when we use this tests that system was deployed correctly, now ansible serves that role.
https://github.com/apache/incubator-openwhisk/blob/master/tests/src/test/scala/services/PingTests.scala#L95
```
    it should "check that the main docker endpoint is functioning" in {
        pingDocker("main.docker.endpoint", WhiskProperties.getMainDockerEndpoint)
    }
```

ansible/environments/mac/group_vars/all
```diff
@@ -4,6 +4,8 @@ whisk_logs_dir: /Users/Shared/wsklogs
 docker_registry: ""
 docker_dns: ""
 
+ansible_python_interpreter: "/usr/local/bin/python"
```
rabbah: this environment should be deleted then.

ansible/README.md
rabbah: we should back out this change initially - the 4243 issue is a bit messy but in general this pr makes an improvement; we should try to get it in incrementally.

ansible/README.md
```diff
@@ -72,8 +59,9 @@ systemProp.http.proxyPort=3128
 **Caveat:** All Ansible commands are meant to be executed from the `ansible` directory.
 This is important because that's where `ansible.cfg` is located which contains generic settings that are needed for the remaining steps.
 
-In all instructions, replace `<environment>` with your target environment. e.g. `mac`if you want to deploy using a local mac setup.
+In all instructions, replace `<environment>` with your target environment. e.g. `docker-machine`if you want to deploy using a local docker-machine setup.
 By default, if you omit the `-i` parameter, the `local` environment will be used.
+If using Docker for Mac use the `local` environment.
```
rabbah: there may be other places to clean up docs for mac - worth double checking.

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -14,7 +14,6 @@
     path: "{{ whisk_logs_dir }}/controller{{ groups['controllers'].index(inventory_hostname) }}"
     state: directory
     mode: 0777
-  become: true
```
rabbah: in general there are many of these there are a problem - addressed in #2536 since it seems wrong to begin with.

ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -26,34 +26,71 @@
   retries: "{{ docker.pull.retries }}"
   delay: "{{ docker.pull.delay }}"
 
-- name: determine docker root dir
+- name: "determine docker root dir on docker-machine"
   uri:  url="http://{{ inventory_hostname }}:{{ docker.port }}/info" return_content=yes
-  register: dockerInfo
+  register: dockerInfo_output
+  when: whisk_version_name == "mac"
```
rabbah: are these all wrong now? should they be "local"?

tools/build/redo
```diff
@@ -261,6 +261,10 @@ Components = [
                   'install catalog',
                   yaml = 'postdeploy.yml'),
 
+    makeComponent('apigateway',
```
rabbah: let's pull this out into a separate pr.

ansible/roles/controller/tasks/deploy.yml
csantanapr: this is clean now

ansible/roles/invoker/tasks/deploy.yml
csantanapr: this is still required only for while we have docker-machine support

tools/build/redo
csantanapr: will do

ansible/roles/invoker/tasks/deploy.yml
rabbah: ok so this condition is wrong? instead of == "mac" it should be == "docker-machine" you're saying?

ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -26,9 +26,25 @@
...
+- name: "determine docker root dir"
+  shell: echo -e "GET http:/v1.24/info HTTP/1.0\r\n" | nc -U /var/run/docker.sock | grep "{"
+  args:
+    executable: /bin/bash
+  register: dockerInfo_output
+  when: whisk_version_name != "mac"
+
+- set_fact:
+    dockerInfo: "{{ dockerInfo_output.stdout|from_json }}"
+  when: whisk_version_name != "mac"
```
rabbah: what about this condition? "mac" is gone so is this always done?

ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -39,20 +55,42 @@
...
+    INFO=`echo -e "GET http:/v1.24/containers/json?filters={{ '{"name":[ "invoker" ]}' | urlencode }} HTTP/1.0\r\n" | nc -U /var/run/docker.sock | grep "{"`
+    if [ -z "$INFO" ]; then
+      echo []
+    else
+      echo $INFO
+    fi
+  args:
+    executable: /bin/bash
+  register: invokerInfo_output
+  when: whisk_version_name != "mac"
```
rabbah: and here.

ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -39,20 +55,42 @@
...
+
+- set_fact:
+    invokerInfo: "{{ invokerInfo_output.stdout|from_json }}"
+  when: whisk_version_name != "mac"
```
rabbah: and here.

ansible/roles/invoker/tasks/deploy.yml
csantanapr: whisk_version_name is only set to `mac` in ansible/environment/mac/group_vars/all

ansible/README.md
```diff
@@ -297,7 +297,7 @@ without spaces as there is no current fix available to this problem.
 
 #### Changing limits
 The system throttling limits can be changed by modifying the `group_vars` for your environment. For example,
-mac users will find the limits in this file [./environments/mac/group_vars/all](./environments/mac/group_vars/all):
+Docker for Mac or ubuntu users will find the limits in this file [./environments/local/group_vars/all](./environments/local/group_vars/all):
```
rabbah: Ubuntu (capital U).

ansible/README.md
csantanapr: just found one additional place to update

ansible/logs.yml
```diff
@@ -27,10 +27,10 @@
 - hosts: shared,core,invokers
   tasks:
   - name: get all docker containers
-    local_action: shell docker --host tcp://{{inventory_hostname}}:{{docker.port}} ps -a --format="{% raw %}{{.Names}}{% endraw %}"
```
rabbah: will this work for a distributed deployment?

ansible/README.md
```diff
@@ -283,7 +297,7 @@ without spaces as there is no current fix available to this problem.
...
```
rabbah: Ubuntu (capital U).

ansible/logs.yml
rabbah: i'm not sure we need this change.

csantanapr:
continue work in #2686

PR #2507, opened 2017-07-19
Add Invoker instnaceId in invokerReative

keunseob:
I made an additional PR due to @rabbah mentioned on #2505 
if the reactive pool is ready to go online, this code is also necessary.
this patch adds a few tiny code to use invoker instance id for creating/deleting container in case of reative poll is enabled.

Signed-off-by: keunseob.kim <keunseob.kim@samsung.com>

PR #2527, opened 2017-07-26
Allow multiple controllers and invokers in local environments.

markusthoemmes:
This allows deploying multiple controllers and invokers locally to be able to test their scaleout ability.

**Todo:**
- Add a variable `allow_multiple_instances` to be able to keep the double-invoker check for production environments.

**Depends on:**
- #2452 for getting rid of the registrator.
- #2507 to be able to deploy multiple invokers in one machine.

### DIFF
--- a/ansible/logs.yml
+++ b/ansible/logs.yml
@@ -27,10 +27,10 @@
 - hosts: shared,core,invokers
   tasks:
   - name: get all docker containers
-    local_action: shell docker --host tcp://{{inventory_hostname}}:{{docker.port}} ps -a --format="{% raw %}{{.Names}}{% endraw %}"
+    local_action: shell docker --host tcp://{{ ansible_host }}:{{ docker.port }} ps -a --format="{% raw %}{{.Names}}{% endraw %}"
     register: container_names
   - name: get logs from all containers
-    local_action: shell docker --host tcp://{{inventory_hostname}}:{{docker.port}} logs {{ item }} > "{{ openwhisk_home }}/logs/{{ item }}.log"; exit 0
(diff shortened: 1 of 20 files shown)


rabbah:
nice!

cbickel:
I found another dependency for this PR:
on generating the whisk.properties, two lists are generated. One with all controllers and one with all invokers.
On using these aliases, the list only contains the list of the aliases, but the IPs or hostnames are needed for the tests.

I found the following line in the internet: `controller.hosts={{ groups["controllers"] | map('extract', hostvars, 'ansible_host') | list | join(",") }}`
But on using this one, it does not work when using no aliases.
We should have a deeper look here.

cbickel:
PG2#1973 is 🔵

ansible/logs.yml
```diff
@@ -24,13 +24,14 @@
         - subjects
       when: "'db' not in exclude_logs_from"
 
-- hosts: shared,core,invokers
+- hosts: controllers,invokers
+  serial: 1
```
markusthoemmes: Is this change intentional?

cbickel: In the open source part there are neither `shared`-hosts nor `core`-hosts.
So I removed them, but added the controllers.

markusthoemmes: Any why the serial: 1?

cbickel: Wewere able to deploy several controllers on the same machine, because we changed the inventory_name of the machine.
Here it means, that this skript will be executed on all "hosts", which is only one physical host if we have only one machine.
I did not find a solution to execute this on all hosts, but only once if they have the same ansible_host.
So I did the workaround to execute it one after another to avoid race conditions when the machine copys the same log several times.

PR #2528, opened 2017-07-26
Enable reactive containerpool for all open environments.

markusthoemmes:
This is a stepping stone to make the new pool the default for everything.

Communication will be done via dev list.

### DIFF
--- a/ansible/environments/docker-machine/group_vars/all
+++ b/ansible/environments/docker-machine/group_vars/all
@@ -37,4 +37,4 @@ apigw_host_v2: "http://{{ groups['apigateway']|first }}:{{apigateway.port.api}}/
 
 # RunC enablement
 invoker_use_runc: true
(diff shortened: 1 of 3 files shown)


rabbah:
what about this one: https://github.com/apache/incubator-openwhisk/blob/9401f8b29ae2ddc470d9e543f0f790776ae0b028/ansible/group_vars/all#L163

markusthoemmes:
@rabbah see the mail on the dev list:

http://mail-archives.apache.org/mod_mbox/incubator-openwhisk-dev/201707.mbox/%3C2d9fc7eb-46c0-49b1-b7fc-ba2bde4157c7%40me.com%3E

I'll change that parameter in about a weeks time (it *could* be breaking for existing deployments) and scheduled the whole old pool to be removed mid of August.

PR #2530, opened 2017-07-26
Add a database flag to completely block a given subject.

markusthoemmes:
There was no preservative way of blocking a user from the system temporalily before. This adds the functionality to provide a flag in the user's subject entry to block its keys from further usage.

### DIFF
--- a/tests/src/test/scala/whisk/core/admin/WskAdminTests.scala
+++ b/tests/src/test/scala/whisk/core/admin/WskAdminTests.scala
@@ -30,6 +30,7 @@ import common.WskAdmin
 import common.WskProps
 import whisk.core.entity.AuthKey
 import whisk.core.entity.Subject
+import common.TestUtils
 
 @RunWith(classOf[JUnitRunner])
 class WskAdminTests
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/whisk/core/admin/WskAdminTests.scala
```diff
@@ -91,4 +91,37 @@ class WskAdminTests
...
+                // reverse lookup by namespace
+                wskadmin.cli(Seq("user", "list", "-k", subject)).stdout.trim should be(auth.compact)
+            }, 10, Some(1.second))
+
+            // block the user
+            wskadmin.cli(Seq("user", "block", subject))
+
+            // wait until the user can no longer be found
+            whisk.utils.retry({
+                wskadmin.cli(Seq("user", "list", subject)).stdout.trim should startWith("no identities")
```
rabbah: this behavior might be a little confusing - i understand why it's done this way: requires no schema changes but maybe in the wskadmin utility, listing namespaces for a blocked subject could at least tell you the identities are blocked... this could be helpful if for example an identity is blocked, but without knowing so, someone tries to add new namespaces and end up "startled" in that the addition will work but listing the spaces will return "no identities".

tools/admin/wskadmin
```diff
@@ -105,6 +105,12 @@ def parseArgs():
     subcmd = subparser.add_parser('whois', help='identify user from an authorization key')
     subcmd.add_argument('authkey', help='the credentials to look up')
 
+    subcmd = subparser.add_parser('block', help='block a user')
+    subcmd.add_argument('subject', help='the user to block')
```
rabbah: should this utility allow blocking per namespace? to block a namespace lookup the subject then block the corresponding document?

rabbah: this is non-nonsensical since it would block all subjects that belong to a namespace which is not the desired outcome. Thanks @markusthoemmes for the sanity check.

tests/src/test/scala/whisk/core/admin/WskAdminTests.scala
markusthoemmes: Hmm. The `list` utitility is to lookup namespaces in the first place. I could though alter the test to add a second subject in the same namespace and check that only one of them disappear from the namespace list. Would that make more sense?

We don't have a "list namespaces per subject" command.

tools/admin/wskadmin
rabbah: on second thought - it may be something desirable _in the future_ (to block all subjects belonging to a namespace).

tests/src/test/scala/whisk/core/admin/WskAdminTests.scala
```diff
@@ -91,4 +91,44 @@ class WskAdminTests
...
     }
+
+    it should "block and unblock a user respectively" in {
+        val wskadmin = new RunWskAdminCmd {}
+        val auth = AuthKey()
+        val subject = Subject().asString
+        val other = Subject().asString
+        val commonNamespace = "testspace"
+        try {
+            wskadmin.cli(Seq("user", "create", subject, "-ns", commonNamespace, "-u", auth.compact))
```
rabbah: i would rename `subject` and `other` with `user1` and `user2` (or `subject1` and `subject2`) :)

rabbah:
could we make wskadmin fail if it tries to add a namespace to a blocked subject?

markusthoemmes:
PG1 1856 🔵

PR #2531, opened 2017-07-26
Share bookkeeping data across controllers

vvraskin:
Hello, 

this PR is a WIP to share the state across multiple controllers to scale out the controllers in OpenWhisk. 
It will allow us to share Activations per namespaces and invokers across several controllers using [Akka Distributed framework](http://doc.akka.io/docs/akka/snapshot/scala/distributed-data.html). The TrieMaps were replaced by PNCounterMaps, which are updated locally without waiting for the replication across other nodes in the cluster. 

UPDATED.
I have a couple of things to mention:
* Parsing of the seed-nodes is defined in a separate method due to a well known akka config issue https://github.com/akka/akka/issues/23600. 
* The old LoadBalancerData implementation is still in place, so in case of emergency we can turn off akka replication across Controllers. 

Any other todos and comments are welcome. 

@cbickel @markusthoemmes @rabbah 

regards, 
Vadim.

### DIFF
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -16,6 +16,12 @@
     mode: 0777
   become: "{{ logs.dir.become }}"
 
+- name: create seed nodes list
+  set_fact:
+    seed_nodes_list: "{{ seed_nodes_list | default([]) }} + [ \"{{item.1}}:{{controller.akka.cluster.basePort+item.0}}\" ]"
+  with_indexed_items:
+  - "{{ controller.akka.cluster.seedNodes }}"
+
 - name: (re)start controller
   docker_container:
     name: controller{{ groups['controllers'].index(inventory_hostname) }}
@@ -58,11 +64,18 @@
       "LOADBALANCER_INVOKERBUSYTHRESHOLD": "{{ invoker.busyThreshold }}"
 
       "RUNTIMES_MANIFEST": "{{ runtimesManifest | to_json }}"
+      "CONTROLLER_LOCALBOOKKEEPING": "{{ controller.localBookkeeping }}"
+      "AKKA_CLUSTER_PORT": "{{ controller.akka.cluster.basePort + groups['controllers'].index(inventory_hostname) }}"
+      "AKKA_CLUSTER_HOST": "{{ controller.akka.cluster.host[groups['controllers'].index(inventory_hostname)] }}"
+      "AKKA_CLUSTER_SEED_NODES": "{{seed_nodes_list | join(' ') }}"
+      "AKKA_CLUSTER_BIND_PORT": "{{ controller.akka.cluster.bindPort }}"
+      "AKKA_ACTOR_PROVIDER": "{{ controller.akka.provider }}"
     volumes:
       - "{{ whisk_logs_dir }}/controller{{ groups['controllers'].index(inventory_hostname) }}:/logs"
     ports:
       - "{{ controller.basePort + groups['controllers'].index(inventory_hostname) }}:8080"
-    command: /bin/sh -c "controller/bin/controller {{ groups['controllers'].index(inventory_hostname) }} >> /logs/controller{{ groups['controllers'].index(inventory_hostname) }}_logs.log 2>&1"
+      - "{{ controller.akka.cluster.basePort + groups['controllers'].index(inventory_hostname) }}:{{ controller.akka.cluster.bindPort }}"
+    command: /bin/sh -c "export CONTAINER_IP=$(hostname -I); controller/bin/controller {{ groups['controllers'].index(inventory_hostname) }} >> /logs/controller{{ groups['controllers'].index(inventory_hostname) }}_logs.log 2>&1"
 
 - name: wait until the Controller in this host is up and running
   uri:
--- a/common/scala/build.gradle
+++ b/common/scala/build.gradle
@@ -13,8 +13,9 @@ dependencies {
 
     compile 'io.spray:spray-json_2.11:1.3.3'
 
-    compile 'com.typesafe.akka:akka-actor_2.11:2.4.16'
-    compile 'com.typesafe.akka:akka-slf4j_2.11:2.4.16'
+    compile 'com.typesafe.akka:akka-actor_2.11:2.5.3'
+    compile 'com.typesafe.akka:akka-stream_2.11:2.5.3'
+    compile 'com.typesafe.akka:akka-slf4j_2.11:2.4.19'
     compile 'com.typesafe.akka:akka-http-core_2.11:10.0.10'
     compile 'com.typesafe.akka:akka-http-spray-json_2.11:10.0.10'
 
--- a/common/scala/build.gradle
+++ b/common/scala/build.gradle
@@ -13,8 +13,8 @@ dependencies {
 
     compile 'io.spray:spray-json_2.11:1.3.3'
 
-    compile 'com.typesafe.akka:akka-actor_2.11:2.5.3'
-    compile 'com.typesafe.akka:akka-stream_2.11:2.5.3'
+    compile 'com.typesafe.akka:akka-actor_2.11:2.5.4'
+    compile 'com.typesafe.akka:akka-stream_2.11:2.5.4'
     compile 'com.typesafe.akka:akka-slf4j_2.11:2.5.4'
     compile 'com.typesafe.akka:akka-http-core_2.11:10.0.10'
     compile 'com.typesafe.akka:akka-http-spray-json_2.11:10.0.10'
--- a/common/scala/build.gradle
+++ b/common/scala/build.gradle
@@ -15,7 +15,7 @@ dependencies {
 
     compile 'com.typesafe.akka:akka-actor_2.11:2.5.3'
     compile 'com.typesafe.akka:akka-stream_2.11:2.5.3'
-    compile 'com.typesafe.akka:akka-slf4j_2.11:2.4.19'
+    compile 'com.typesafe.akka:akka-slf4j_2.11:2.5.4'
     compile 'com.typesafe.akka:akka-http-core_2.11:10.0.10'
     compile 'com.typesafe.akka:akka-http-spray-json_2.11:10.0.10'
 
--- a/common/scala/src/main/scala/whisk/common/Config.scala
+++ b/common/scala/src/main/scala/whisk/common/Config.scala
@@ -95,6 +95,16 @@ class Config(requiredProperties: Map[String, String], optionalProperties: Set[St
   }
 
   /**
+   * Returns the value of a given key parsed as a boolean.
+   * If parsing fails, return the default value.
+   *
+   * @param key the property that has to be returned.
+   */
+  def getAsBoolean(key: String, defaultValue: Boolean): Boolean = {
+    Try(getProperty(key).toBoolean).getOrElse(defaultValue)
+  }
+
+  /**
    * Converts the set of property to a string for debugging.
    */
   def mkString: String = settings.mkString("\n")
@@ -134,6 +144,7 @@ object Config {
       val envp = p.replace('.', '_').toUpperCase
       val envv = env.get(envp)
       if (envv.isDefined) {
+        println(s"environment set value for $p")
         properties += p -> envv.get.trim
       }
     }
@@ -148,6 +159,7 @@ object Config {
   def validateProperties(required: Map[String, String], properties: Map[String, String]): Boolean = {
     required.keys.forall { key =>
       val value = properties(key)
+      if (value == null) println(s"required property $key still not set")
       value != null
     }
   }
--- a/common/scala/src/main/scala/whisk/common/Config.scala
+++ b/common/scala/src/main/scala/whisk/common/Config.scala
@@ -43,7 +43,7 @@ import scala.util.Try
  * @param env an optional environment to read from (defaults to sys.env).
  */
 class Config(requiredProperties: Map[String, String], optionalProperties: Set[String] = Set())(
-  env: Map[String, String] = sys.env)(implicit logging: Logging) {
+  env: Map[String, String] = sys.env) {
 
   private val settings = getProperties().toMap.filter {
     case (k, v) =>
@@ -128,13 +128,12 @@ object Config {
    * Reads a Map of key-value pairs from the environment -- store them in the
    * mutable properties object.
    */
-  def readPropertiesFromEnvironment(properties: scala.collection.mutable.Map[String, String], env: Map[String, String])(
-    implicit logging: Logging) = {
+  def readPropertiesFromEnvironment(properties: scala.collection.mutable.Map[String, String],
+                                    env: Map[String, String]) = {
     for (p <- properties.keys) {
       val envp = p.replace('.', '_').toUpperCase
       val envv = env.get(envp)
       if (envv.isDefined) {
-        logging.info(this, s"environment set value for $p")
         properties += p -> envv.get.trim
       }
     }
@@ -146,11 +145,9 @@ object Config {
    * @param required a key-value map where the keys are required properties
    * @param properties a set of properties to check
    */
-  def validateProperties(required: Map[String, String], properties: Map[String, String])(
-    implicit logging: Logging): Boolean = {
+  def validateProperties(required: Map[String, String], properties: Map[String, String]): Boolean = {
     required.keys.forall { key =>
       val value = properties(key)
-      if (value == null) logging.error(this, s"required property $key still not set")
       value != null
     }
   }
--- a/common/scala/src/main/scala/whisk/common/Config.scala
+++ b/common/scala/src/main/scala/whisk/common/Config.scala
@@ -43,7 +43,7 @@ import scala.util.Try
  * @param env an optional environment to read from (defaults to sys.env).
  */
 class Config(requiredProperties: Map[String, String], optionalProperties: Set[String] = Set())(
-  env: Map[String, String] = sys.env) {
+  env: Map[String, String] = sys.env)(implicit logging: Logging) {
 
   private val settings = getProperties().toMap.filter {
     case (k, v) =>
@@ -138,13 +138,13 @@ object Config {
    * Reads a Map of key-value pairs from the environment -- store them in the
    * mutable properties object.
    */
-  def readPropertiesFromEnvironment(properties: scala.collection.mutable.Map[String, String],
-                                    env: Map[String, String]) = {
+  def readPropertiesFromEnvironment(properties: scala.collection.mutable.Map[String, String], env: Map[String, String])(
+    implicit logging: Logging) = {
     for (p <- properties.keys) {
       val envp = p.replace('.', '_').toUpperCase
       val envv = env.get(envp)
       if (envv.isDefined) {
-        println(s"environment set value for $p")
+        logging.info(this, s"environment set value for $p")
         properties += p -> envv.get.trim
       }
     }
@@ -156,10 +156,11 @@ object Config {
    * @param required a key-value map where the keys are required properties
    * @param properties a set of properties to check
    */
-  def validateProperties(required: Map[String, String], properties: Map[String, String]): Boolean = {
+  def validateProperties(required: Map[String, String], properties: Map[String, String])(
+    implicit logging: Logging): Boolean = {
     required.keys.forall { key =>
       val value = properties(key)
-      if (value == null) println(s"required property $key still not set")
+      if (value == null) logging.error(this, s"required property $key still not set")
       value != null
     }
   }
--- a/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
+++ b/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
@@ -22,8 +22,6 @@ import java.io.File
 import scala.io.Source
 import whisk.common.Config
 
-import scala.util.Try
-
 /**
  * A set of properties which might be needed to run a whisk microservice implemented
  * in scala.
@@ -49,10 +47,6 @@ class WhiskConfig(requiredProperties: Map[String, String],
     val properties = super.getProperties()
     WhiskConfig.readPropertiesFromFile(properties, Option(propertiesFile) getOrElse (WhiskConfig.whiskPropertiesFile))
     properties
-  }
-
-  def getAsBoolean(key: String, defaultValaue :Boolean) :Boolean = {
-    Try(getProperty(key).toBoolean).getOrElse(defaultValaue)
   }
 
   val servicePort = this(WhiskConfig.servicePort)
--- a/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
+++ b/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
@@ -20,8 +20,9 @@ package whisk.core
 import java.io.File
 
 import scala.io.Source
-
 import whisk.common.Config
+
+import scala.util.Try
 
 /**
  * A set of properties which might be needed to run a whisk microservice implemented
@@ -48,6 +49,10 @@ class WhiskConfig(requiredProperties: Map[String, String],
     val properties = super.getProperties()
     WhiskConfig.readPropertiesFromFile(properties, Option(propertiesFile) getOrElse (WhiskConfig.whiskPropertiesFile))
     properties
+  }
+
+  def getAsBoolean(key: String, defaultValaue :Boolean) :Boolean = {
+    Try(getProperty(key).toBoolean).getOrElse(defaultValaue)
   }
 
   val servicePort = this(WhiskConfig.servicePort)
@@ -94,14 +99,13 @@ class WhiskConfig(requiredProperties: Map[String, String],
   val mainDockerEndpoint = this(WhiskConfig.mainDockerEndpoint)
 
   val runtimesManifest = this(WhiskConfig.runtimesManifest)
-
   val actionInvokePerMinuteLimit = this(WhiskConfig.actionInvokePerMinuteLimit)
   val actionInvokeConcurrentLimit = this(WhiskConfig.actionInvokeConcurrentLimit)
   val triggerFirePerMinuteLimit = this(WhiskConfig.triggerFirePerMinuteLimit)
   val actionInvokeSystemOverloadLimit = this(WhiskConfig.actionInvokeSystemOverloadLimit)
   val actionSequenceLimit = this(WhiskConfig.actionSequenceMaxLimit)
   val controllerSeedNodes = this(WhiskConfig.controllerSeedNodes)
-  val controllerLocalBookkeeping = this(WhiskConfig.controllerLocalBookkeeping)
+  val controllerLocalBookkeeping = getAsBoolean(WhiskConfig.controllerLocalBookkeeping, false)
 
 }
 
--- a/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
+++ b/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
@@ -22,7 +22,6 @@ import java.io.File
 import scala.io.Source
 
 import whisk.common.Config
-import whisk.common.Logging
 
 /**
  * A set of properties which might be needed to run a whisk microservice implemented
@@ -37,7 +36,7 @@ import whisk.common.Logging
 class WhiskConfig(requiredProperties: Map[String, String],
                   optionalProperties: Set[String] = Set(),
                   propertiesFile: File = null,
-                  env: Map[String, String] = sys.env)(implicit val logging: Logging)
+                  env: Map[String, String] = sys.env)
     extends Config(requiredProperties, optionalProperties)(env) {
 
   /**
@@ -101,6 +100,9 @@ class WhiskConfig(requiredProperties: Map[String, String],
   val triggerFirePerMinuteLimit = this(WhiskConfig.triggerFirePerMinuteLimit)
   val actionInvokeSystemOverloadLimit = this(WhiskConfig.actionInvokeSystemOverloadLimit)
   val actionSequenceLimit = this(WhiskConfig.actionSequenceMaxLimit)
+  val controllerSeedNodes = this(WhiskConfig.controllerSeedNodes)
+  val controllerLocalBookkeeping = this(WhiskConfig.controllerLocalBookkeeping)
+
 }
 
 object WhiskConfig {
@@ -129,10 +131,8 @@ object WhiskConfig {
    * Reads a Map of key-value pairs from the environment (sys.env) -- store them in the
    * mutable properties object.
    */
-  def readPropertiesFromFile(properties: scala.collection.mutable.Map[String, String], file: File)(
-    implicit logging: Logging) = {
+  def readPropertiesFromFile(properties: scala.collection.mutable.Map[String, String], file: File) = {
     if (file != null && file.exists) {
-      logging.info(this, s"reading properties from file $file")
       for (line <- Source.fromFile(file).getLines if line.trim != "") {
         val parts = line.split('=')
         if (parts.length >= 1) {
@@ -140,10 +140,7 @@ object WhiskConfig {
           val v = if (parts.length == 2) parts(1).trim else ""
           if (properties.contains(p)) {
             properties += p -> v
-            logging.debug(this, s"properties file set value for $p")
           }
-        } else {
-          logging.warn(this, s"ignoring properties $line")
         }
       }
     }
@@ -221,4 +218,6 @@ object WhiskConfig {
   val actionInvokeConcurrentLimit = "limits.actions.invokes.concurrent"
   val actionInvokeSystemOverloadLimit = "limits.actions.invokes.concurrentInSystem"
   val triggerFirePerMinuteLimit = "limits.triggers.fires.perMinute"
+  val controllerSeedNodes = "akka.cluster.seed.nodes"
+  val controllerLocalBookkeeping = "controller.localBookkeeping"
 }
--- a/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
+++ b/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
@@ -20,7 +20,7 @@ package whisk.core
 import java.io.File
 
 import scala.io.Source
-import whisk.common.Config
+import whisk.common.{Config, Logging}
 
 /**
  * A set of properties which might be needed to run a whisk microservice implemented
@@ -35,7 +35,7 @@ import whisk.common.Config
 class WhiskConfig(requiredProperties: Map[String, String],
                   optionalProperties: Set[String] = Set(),
                   propertiesFile: File = null,
-                  env: Map[String, String] = sys.env)
+                  env: Map[String, String] = sys.env)(implicit val logging: Logging)
     extends Config(requiredProperties, optionalProperties)(env) {
 
   /**
@@ -129,8 +129,10 @@ object WhiskConfig {
    * Reads a Map of key-value pairs from the environment (sys.env) -- store them in the
    * mutable properties object.
    */
-  def readPropertiesFromFile(properties: scala.collection.mutable.Map[String, String], file: File) = {
+  def readPropertiesFromFile(properties: scala.collection.mutable.Map[String, String], file: File)(
+    implicit logging: Logging) = {
     if (file != null && file.exists) {
+      logging.info(this, s"reading properties from file $file")
       for (line <- Source.fromFile(file).getLines if line.trim != "") {
         val parts = line.split('=')
         if (parts.length >= 1) {
@@ -138,7 +140,10 @@ object WhiskConfig {
           val v = if (parts.length == 2) parts(1).trim else ""
           if (properties.contains(p)) {
             properties += p -> v
+            logging.debug(this, s"properties file set value for $p")
           }
+        } else {
+          logging.warn(this, s"ignoring properties $line")
         }
       }
     }
--- a/core/controller/build.gradle
+++ b/core/controller/build.gradle
@@ -11,7 +11,7 @@ repositories {
 }
 
 dependencies {
-    compile 'com.typesafe.akka:akka-distributed-data_2.11:2.5.3'
+    compile 'com.typesafe.akka:akka-distributed-data_2.11:2.5.4'
     compile "org.scala-lang:scala-library:${gradle.scala.version}"
     compile project(':common:scala')
 }
--- a/core/controller/build.gradle
+++ b/core/controller/build.gradle
@@ -11,6 +11,9 @@ repositories {
 }
 
 dependencies {
+    compile 'com.typesafe.akka:akka-distributed-data_2.11:2.5.3'
+    compile 'com.typesafe.akka:akka-cluster-metrics_2.11:2.5.3'
+    compile 'org.scalatest:scalatest_2.11:3.0.1'
     compile "org.scala-lang:scala-library:${gradle.scala.version}"
     compile project(':common:scala')
 }
--- a/core/controller/build.gradle
+++ b/core/controller/build.gradle
@@ -12,8 +12,6 @@ repositories {
 
 dependencies {
     compile 'com.typesafe.akka:akka-distributed-data_2.11:2.5.3'
-    compile 'com.typesafe.akka:akka-cluster-metrics_2.11:2.5.3'
-    compile 'org.scalatest:scalatest_2.11:3.0.1'
     compile "org.scala-lang:scala-library:${gradle.scala.version}"
     compile project(':common:scala')
 }
--- a/core/controller/src/main/resources/application.conf
+++ b/core/controller/src/main/resources/application.conf
@@ -56,8 +56,8 @@ akka.http {
   }
 }
 
-# Check out all akka-remote-2.5.3 options here:
-# http://doc.akka.io/docs/akka/2.5.3/scala/general/configuration.html#config-akka-remote
+# Check out all akka-remote-2.5.4 options here:
+# http://doc.akka.io/docs/akka/2.5.4/scala/general/configuration.html#config-akka-remote
 akka {
   actor {
     provider = ${?AKKA_ACTOR_PROVIDER}
@@ -77,12 +77,6 @@ akka {
   cluster {
     # Disable legacy metrics in akka-cluster.
     metrics.enabled=off
-
-    # Sigar native library extract location during tests.
-    # Note: use per-jvm-instance folder when running multiple jvm on one host.
-    metrics.native-library-extract-folder=${user.dir}/target/native
-    
-    auto-down-unreachable-after = 10s
 
     #distributed-data.notify-subscribers-interval = 0.01
   }
--- a/core/controller/src/main/resources/application.conf
+++ b/core/controller/src/main/resources/application.conf
@@ -78,6 +78,7 @@ akka {
     # Disable legacy metrics in akka-cluster.
     metrics.enabled=off
 
+    auto-down-unreachable-after = ${?AUTO_DOWN_UNREACHABLE_AFTER}
     #distributed-data.notify-subscribers-interval = 0.01
   }
 }
--- a/core/controller/src/main/resources/application.conf
+++ b/core/controller/src/main/resources/application.conf
@@ -68,7 +68,6 @@ akka {
     log-sent-messages = on
 
     netty.tcp {
-      maximum-frame-size=1280000000b
       hostname = ${?AKKA_CLUSTER_HOST}
       port = ${?AKKA_CLUSTER_PORT}
       bind-port = ${?AKKA_CLUSTER_BIND_PORT}
@@ -84,9 +83,6 @@ akka {
     metrics.native-library-extract-folder=${user.dir}/target/native
     
     auto-down-unreachable-after = 10s
-
-    # Enable metrics extension in akka-cluster-metrics.
-    #akka.extensions=["akka.cluster.metrics.ClusterMetricsExtension"]
 
     #distributed-data.notify-subscribers-interval = 0.01
   }
--- a/core/controller/src/main/resources/application.conf
+++ b/core/controller/src/main/resources/application.conf
@@ -55,3 +55,39 @@ akka.http {
     }
   }
 }
+
+# Check out all akka-remote-2.5.3 options here:
+# http://doc.akka.io/docs/akka/2.5.3/scala/general/configuration.html#config-akka-remote
+akka {
+  actor {
+    provider = ${?AKKA_ACTOR_PROVIDER}
+  }
+  remote {
+    log-remote-lifecycle-events = DEBUG
+    log-received-messages = on
+    log-sent-messages = on
+
+    netty.tcp {
+      maximum-frame-size=1280000000b
+      hostname = ${?AKKA_CLUSTER_HOST}
+      port = ${?AKKA_CLUSTER_PORT}
+      bind-port = ${?AKKA_CLUSTER_BIND_PORT}
+      bind-hostname = ${?CONTAINER_IP}
+    }
+  }
+  cluster {
+    # Disable legacy metrics in akka-cluster.
+    metrics.enabled=off
+
+    # Sigar native library extract location during tests.
+    # Note: use per-jvm-instance folder when running multiple jvm on one host.
+    metrics.native-library-extract-folder=${user.dir}/target/native
+    
+    auto-down-unreachable-after = 10s
+
+    # Enable metrics extension in akka-cluster-metrics.
+    #akka.extensions=["akka.cluster.metrics.ClusterMetricsExtension"]
+
+    #distributed-data.notify-subscribers-interval = 0.01
+  }
+}
--- a/core/controller/src/main/scala/whisk/core/controller/Controller.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Controller.scala
@@ -39,7 +39,7 @@ import whisk.core.entitlement._
 import whisk.core.entity._
 import whisk.core.entity.ActivationId.ActivationIdGenerator
 import whisk.core.entity.ExecManifest.Runtimes
-import whisk.core.loadBalancer.{LoadBalancerService, SharedDataService}
+import whisk.core.loadBalancer.{LoadBalancerService}
 import whisk.http.BasicHttpService
 import whisk.http.BasicRasService
 
@@ -152,8 +152,7 @@ object Controller {
       ExecManifest.requiredProperties ++
       RestApiCommons.requiredProperties ++
       LoadBalancerService.requiredProperties ++
-      EntitlementProvider.requiredProperties ++
-      SharedDataService.requiredProperties
+      EntitlementProvider.requiredProperties
 
   private def info(config: WhiskConfig, runtimes: Runtimes, apis: List[String]) =
     JsObject(
@@ -169,13 +168,11 @@ object Controller {
       "runtimes" -> runtimes.toJson)
 
   def main(args: Array[String]): Unit = {
+    implicit val actorSystem = ActorSystem("controller-actor-system")
+    implicit val logger = new AkkaLogging(akka.event.Logging.getLogger(actorSystem, this))
 
     // extract configuration data from the environment
     val config = new WhiskConfig(requiredProperties)
-
-    implicit val actorSystem = ActorSystem("controller-actor-system", SharedDataService.addAkkaSeedNodesToConf(config))
-    implicit val logger = new AkkaLogging(akka.event.Logging.getLogger(actorSystem, this))
-
     val port = config.servicePort.toInt
 
     // if deploying multiple instances (scale out), must pass the instance number as the
--- a/core/controller/src/main/scala/whisk/core/controller/Controller.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Controller.scala
@@ -20,8 +20,6 @@ package whisk.core.controller
 import scala.concurrent.Await
 import scala.concurrent.duration.DurationInt
 import scala.util.{Failure, Success}
-
-import akka.actor._
 import akka.actor.ActorSystem
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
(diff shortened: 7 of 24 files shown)


markusthoemmes:
In general: Try to bury all implementation details in the interface, for instance: We cast all the BigInts to Int outside of the replication? Then do the cast as part of the interface.

Returning a String and "casting" it to a UUID is another example. Just cast inside the interface.

That keeps the clients clean.

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -37,10 +37,14 @@
...
       "JAVA_OPTS": "-Xmx{{ controller.heap }}"
       "CONTROLLER_OPTS": "{{ controller.arguments }}"
+      "AKKA_CLUSTER_PORT": "{{ controller.akka.cluster.port }}"
+      "AKKA_CLUSTER_HOST": "{{ controller.akka.cluster.host }}"
+      "AKKA_CLUSTER_BIND_PORT": "{{ controller.akka.cluster.bindPort }}"
     volumes:
       - "{{ whisk_logs_dir }}/controller{{ groups['controllers'].index(inventory_hostname) }}:/logs"
     ports:
       - "{{ controller.basePort + groups['controllers'].index(inventory_hostname) }}:8080"
+      - "{{controller.akka.cluster.port + groups['controllers'].index(inventory_hostname) }}:{{ controller.akka.cluster.bindPort }}"
```
ddragosd: Multiple instances of controller will use the same port  on the host ?

core/controller/src/main/resources/application.conf
```diff
@@ -38,3 +38,13 @@ spray.can.server {
     max-content-length = 50m
   }
 }
+
+# Disable legacy metrics in akka-cluster.
```
ddragosd: Can we document the reason we're turning off these metrics ?

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -185,7 +187,49 @@ object Controller {
...
+
+        val conf = s"""
+                      include "logging"
+                      akka {
+                         actor {
+                           provider = cluster
+                         }
+                         remote {
+                           log-remote-lifecycle-events = off
+                           netty.tcp {
```
ddragosd: `hostname` and `port` should be different from `bind-hostname` and `bind-port` when running in Docker. The latter 2 should bind to Docker's private IP Address and port, and the first 2 should bind to host's IP and port.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -185,7 +187,49 @@ object Controller {
...
+                            hostname = $hostname
+                            port = $port
+                            bind-hostname = $bindHost
+                            bind-port = $bindPort
+                           }
+                         }
+                      cluster {
+                      seed-nodes = ["akka.tcp://controller-actor-system@$hostname:8000"]
```
ddragosd: I assume/hope we have in plan an auto-discovery of nodes.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -185,7 +187,49 @@ object Controller {
...
     def main(args: Array[String]): Unit = {
-        implicit val actorSystem = ActorSystem("controller-actor-system")
+
+        // read controller parameters from env variables
+        val port = sys.env("AKKA_CLUSTER_PORT")
+        val bindPort = sys.env("AKKA_CLUSTER_BIND_PORT")
+        val hostname = sys.env("AKKA_CLUSTER_HOST")
+        val bindHost = InetAddress.getLocalHost.getHostAddress
+
+        val conf = s"""
```
ddragosd: any reason we can't load this from a config file ?

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -185,7 +187,49 @@ object Controller {
     }
 
     def main(args: Array[String]): Unit = {
-        implicit val actorSystem = ActorSystem("controller-actor-system")
+
+        // read controller parameters from env variables
+        val port = sys.env("AKKA_CLUSTER_PORT")
```
ddragosd: Let's plan for this initialization logic to differ when deploying with Mesos and Kube.

rabbah:
I labeled this a "WIP" patch per the description. Thanks @ddragosd for raising these early. I expect this PR is going to involve a bit more.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerData.scala
```diff
@@ -31,82 +38,77 @@ case class ActivationEntry(id: ActivationId, namespaceId: UUID, invokerName: Ins
...
+    * @param update block calculating the entry to add.
+    *               Note: This is evaluated iff the entry
+    *               didn't exist before.
+    * @return the entry calculated by the block or iff it did
+    *         exist before the entry from the state
+    */
+  def putActivation(id: ActivationId, update: => ActivationEntry): ActivationEntry = {
+    activationsById.getOrElseUpdate(id, {
+      val entry = update
+      sharedStateNamespaces ! IncreaseCounter(entry.namespaceId.asString, 1)
```
ddragosd: I'm trying to understand the need to track `namespaces` and `invokers` in the LB. Is it related to the tenant isolation ?

markusthoemmes: 1. Loadbalancer tracks concurrent invocation of namespaces (needed for throttling)
2. Loadbalancer tracks load per invoker to balance the load with the least queuing possible.

ddragosd: 👍

ddragosd:
It's shaping up nicely @vvraskin . 
Given it's a distributed state, are you planning to add tests with multiple instances, including test with failure scenarios ?

core/controller/src/main/resources/application.conf
vvraskin: Thats not the final version of the config, going to document it we indeed turn it off.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
vvraskin: Sure, lets discuss it,  can't we read env vars in mesos and kube?

vvraskin:
@ddragosd 
Sure, I'm going to test it locally and on our load test environment. 
The plan is to stress 3 node cluster under a high load, and look how akka behaves under the burst. 
I'm also planning to test whether the cluster is still functioning in case of node failure (-s). 
The next tests would be to measure the maximal number of nodes in the cluster with an acceptable performance.

ddragosd:
@vvraskin mind adding a customizable way to specify the `seed-nodes` in the cluster ? I.e. [take a look at how the seed nodes can be auto-discovered with Mesos](https://github.com/ddragosd/akka-samples/blob/mesos-cluster/akka-sample-mesos-scala/src/main/scala/sample/cluster/mesos/MarathonConfig.scala) and  [how it's used](https://github.com/ddragosd/akka-samples/blob/mesos-cluster/akka-sample-mesos-scala/src/main/scala/sample/cluster/mesos/SimpleMesosClusterApp.scala#L12). We can use the SPI support already in the `master` branch, with a default value that reads the seed-nodes from ansible.

ddragosd:
Or, instead of SPI, we can configure Akka as:
```javascript
cluster {
    name: "my-akka-cluster",
    discovery {
      source: "ansible | mesos | kubernetes", 
    }
   ...
  }
```
And provide an impl for each discovery source. 

This configuration will become more and more important as we bring more clustering capabilities  into the project.

core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
```diff
@@ -22,6 +22,9 @@ import whisk.common.TransactionId
 import whisk.core.entity.Identity
 import whisk.core.loadBalancer.LoadBalancer
 
+import scala.concurrent.ExecutionContext.Implicits.global
```
markusthoemmes: Please do not use this import in our components at all. It introduces an independent execution context which does not necessarily relate to the EC in that component. You'll need to pass through the actor system's EC.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerData.scala
```diff
@@ -17,107 +17,130 @@
...
+import akka.event.Logging
+import whisk.core.entity.{ActivationId, UUID}
 import scala.collection.concurrent.TrieMap
-import scala.concurrent.Promise
-
-import whisk.core.entity.{ActivationId, UUID, WhiskActivation}
-import whisk.core.entity.InstanceId
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext.Implicits.global
```
markusthoemmes: Same as above.

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
```diff
@@ -0,0 +1,131 @@
...
+import akka.cluster.ddata.Replicator._
+import whisk.common.AkkaLogging
+import whisk.core.WhiskConfig
+import com.typesafe.config.{Config, ConfigFactory}
+
+case class IncreaseCounter(key: String, value: Long)
+case class DecreaseCounter(key: String, value: Long)
+case class ReadCounter(key: String)
+case class RemoveCounter(key: String)
+case class GetTheMap()
```
markusthoemmes: Can be `case object GetMap`

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
```diff
@@ -0,0 +1,131 @@
...
+case class MapWithCounters(dataMap: Map[String, BigInt])
+case class InSync()
```
markusthoemmes: Can be `case object InSync`

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerData.scala
```diff
@@ -17,107 +17,130 @@
...
+  def removeActivation(aid: ActivationId): Option[ActivationEntry]
+}
+
+/**
+ * Encapsulates data used for loadbalancer and active-ack bookkeeping.
+ *
+ * Note: The state keeping is backed by distributed akka actors. All CRUDs operations are done on local values, thus
+ * a stale value might be read.
+ */
+class LoadBalancerData(implicit val actorSystem: ActorSystem) extends BookkeepingData {
```
markusthoemmes: Can we encapsulate this in its own file and name it `LoadBalancerDataDistributed`. (Or adopt the more common naming convention "specialization attribute up front" like "DistributedLoadBalancerData")

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerData.scala
```diff
@@ -17,107 +17,130 @@
...
- * Note: The state keeping is backed by concurrent data-structures. As such,
- * concurrent reads can return stale values (especially the counters returned).
- */
-class LoadBalancerData() {
-
-  private val activationByInvoker = TrieMap[InstanceId, AtomicInteger]()
-  private val activationByNamespaceId = TrieMap[UUID, AtomicInteger]()
-  private val activationsById = TrieMap[ActivationId, ActivationEntry]()
-  private val totalActivations = new AtomicInteger(0)
+trait BookkeepingData {
```
markusthoemmes: Should this remain `LoadBalancerData`?

tests/src/test/scala/whisk/core/loadBalancer/test/LoadBalancerDataTests.scala
```diff
@@ -17,29 +17,49 @@
...
+  val port = 2552
+  val config = ConfigFactory
+    .parseString("akka.cluster { seed-nodes = [\"akka.tcp://controller-actor-system@127.0.0" +
+      s".1:$port" + "\"] }")
+    .withValue("akka.remote.netty.tcp.hostname", ConfigValueFactory.fromAnyRef("127.0.0.1"))
+    .withValue("akka.remote.netty.tcp.port", ConfigValueFactory.fromAnyRef(port))
+    .withValue("akka.cluster.auto-down-unreachable-after", ConfigValueFactory.fromAnyRef("10s"))
+    .withValue("akka.actor.provider", ConfigValueFactory.fromAnyRef("cluster"))
+    .withValue("akka.remote.log-remote-lifecycle-events", ConfigValueFactory.fromAnyRef("off"))
+    .withFallback(ConfigFactory.load())
```
markusthoemmes: The static ones can be put into application.conf of the tests to reduce verbosity here.

tests/src/test/scala/whisk/core/loadBalancer/test/LoadBalancerDataTests.scala
```diff
@@ -17,29 +17,49 @@
...
 
   it should "return the number of activations for a namespace" in {
-
     val loadBalancerData = new LoadBalancerData()
     loadBalancerData.putActivation(firstEntry.id, firstEntry)
 
-    loadBalancerData.activationCountOn(firstEntry.namespaceId) shouldBe 1
-    loadBalancerData.activationCountOn(firstEntry.invokerName) shouldBe 1
+    Await.result(loadBalancerData.activationCountOn(firstEntry.namespaceId), 1.second) shouldBe 1
+    Await.result(loadBalancerData.activationCountPerInvoker, 1.second) shouldBe Map(
```
markusthoemmes: A local await helper like we use in many tests might make sense here, to stop repeating the `1.second` timeout:

```scala
def await[A](f: Future[A], timeout: FiniteDuration = 1.second) = Await.result(f, timeout)
```

tests/src/test/scala/whisk/core/loadBalancer/test/SharedDataServiceTests.scala
```diff
@@ -0,0 +1,113 @@
...
+    }
+    "increase the counter" in {
+      sharedDataService ! (IncreaseCounter("Donald", 1))
+      sharedDataService ! ReadCounter("Donald")
+      expectMsg(1)
+    }
+    "decrease the counter" in {
+      sharedDataService ! (IncreaseCounter("Donald", 2))
+      sharedDataService ! (DecreaseCounter("Donald", 2))
+      Thread.sleep(500)
```
markusthoemmes: Those `Thread.sleeps` make me anxious. Is there a way around them?

tests/src/test/scala/whisk/core/loadBalancer/test/SharedDataServiceTests.scala
```diff
@@ -0,0 +1,113 @@
...
+    val port = 2552
+    val config = ConfigFactory
+      .parseString("akka.cluster { seed-nodes = [\"akka.tcp://controller-actor-system@127.0.0" +
+        s".1:$port" + "\"] }")
+      .withValue("akka.remote.netty.tcp.hostname", ConfigValueFactory.fromAnyRef("127.0.0.1"))
+      .withValue("akka.remote.netty.tcp.port", ConfigValueFactory.fromAnyRef(port))
+      .withValue("akka.cluster.auto-down-unreachable-after", ConfigValueFactory.fromAnyRef("10s"))
+      .withValue("akka.actor.provider", ConfigValueFactory.fromAnyRef("cluster"))
+      .withValue("akka.remote.log-remote-lifecycle-events", ConfigValueFactory.fromAnyRef("off"))
+      .withFallback(ConfigFactory.load())
```
markusthoemmes: Same comment as above on externalizing static bits to application.conf.

tests/src/test/scala/whisk/core/loadBalancer/test/SharedDataServiceTests.scala
```diff
@@ -0,0 +1,113 @@
...
+  override def afterAll {
+    TestKit.shutdownActorSystem(system)
+  }
+
+  "A Shared Data Service" must {
+
+    val port = 2552
+    val config = ConfigFactory
+      .parseString("akka.cluster { seed-nodes = [\"akka.tcp://controller-actor-system@127.0.0" +
+        s".1:$port" + "\"] }")
```
markusthoemmes: This is a strange mixture of concatenation and interpolation. Note there is `"""` to be able to use `"` just as you want to:

```scala
.parseString(s"""akka.cluster { seed-nodes = ["akka.tcp://controller-actor-system@127.0.0.1:$port"] }""")
```

Same applies to the other inline config.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerData.scala
```diff
@@ -17,107 +17,130 @@
...
-   * Get the number of activations for a specific invoker.
+   * Get the number of activations for each invoker.
    *
-   * @param invoker The invoker to get the activation count for
    * @return a map (invoker -> number of activations queued for the invoker)
    */
-  def activationCountOn(invoker: InstanceId): Int = {
-    activationByInvoker.get(invoker).map(_.get).getOrElse(0)
-  }
+  def activationCountPerInvoker: Future[Map[String, Int]]
```
markusthoemmes: I'd rather keep the interface strictly typed to `InstanceId`. Can we hide the underlying `String`?

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -129,21 +131,16 @@ object WhiskConfig {
...
         val parts = line.split('=')
         if (parts.length >= 1) {
           val p = parts(0).trim
           val v = if (parts.length == 2) parts(1).trim else ""
           if (properties.contains(p)) {
             properties += p -> v
-            logging.debug(this, s"properties file set value for $p")
           }
-        } else {
-          logging.warn(this, s"ignoring properties $line")
```
markusthoemmes: Do we lose anything by removing this? Like debuggability of failed startups?

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerDataLocal.scala
```diff
@@ -0,0 +1,77 @@
...
+  private val totalActivations = new AtomicInteger(0)
+
+  override def totalActivationCount: Future[Int] = Future.successful(totalActivations.get)
+
+  override def activationCountOn(namespace: UUID): Future[Int] = {
+    Future.successful(activationByNamespaceId.get(namespace).map(_.get).getOrElse(0))
+  }
+
+  override def activationCountPerInvoker: Future[Map[String, Int]] = {
+    Future.successful(activationByInvoker.map(x => (x._1.toString, x._2.get)).toMap)
```
markusthoemmes: If we can switch back to not using `String` we can use the lazy `mapValues` here so we don't update the whole map everytime it's gotten (once per invocation).

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
```diff
@@ -0,0 +1,131 @@
...
+   * @return Updated Config
+   */
+  def addAkkaSeedNodesToConf(whiskConf: WhiskConfig): Config = {
+    val conf = ConfigFactory.load()
+
+    val cluster = conf.getString("akka.actor.provider")
+
+    if (cluster == "cluster") {
+      val seedNodes = whiskConf.controllerSeedNodes
+      val nodes = seedNodes.split(' ').map(x => "\"akka.tcp://controller-actor-system@" + x + "\"")
```
markusthoemmes: Can be 

```scala
val nodes = seedNodes.split(' ').map(x => s""""akka.tcp://controller-actor-system@$x"""")
```

To avoid espacing.

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
```diff
@@ -0,0 +1,131 @@
...
+import whisk.common.AkkaLogging
+import whisk.core.WhiskConfig
+import com.typesafe.config.{Config, ConfigFactory}
+
+case class IncreaseCounter(key: String, value: Long)
+case class DecreaseCounter(key: String, value: Long)
+case class ReadCounter(key: String)
+case class RemoveCounter(key: String)
+case class GetTheMap()
+case class MapWithCounters(dataMap: Map[String, BigInt])
```
markusthoemmes: Do we need this intermediary type?

tests/src/test/scala/whisk/core/loadBalancer/test/LoadBalancerDataTests.scala
vvraskin: Could be good for this case, but I'd like to keep the config closer to the test, once we introduce more actors we could have certain interference between the test/actor config.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
vvraskin: I think once we needs to debug the config, one can lookup the properties in the env vars in the container.

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
vvraskin: Not really, I was experimenting with it to remove the warning `Warning:(73, 17) non-variable type argument String in type pattern` from the SharedDataService test. However, it doesn't cause an issue in the code itself. 
I've removed the case class. As long as we receive Map[String, BigInt] from `SharedStateService` we should be safe.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -51,6 +51,10 @@ class WhiskConfig(requiredProperties: Map[String, String],
     properties
   }
 
+  def getAsBoolean(key: String, defaultValaue :Boolean) :Boolean = {
+    Try(getProperty(key).toBoolean).getOrElse(defaultValaue)
+  }
```
markusthoemmes: Should live in Config.scala (there are a few other of those methods already)

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -83,6 +83,11 @@ trait LoadBalancer {
 
 }
 
+case class ActivationEntry(id: ActivationId,
```
markusthoemmes: This definition should stay in the file of the LoadBalancerData trait imho.

tests/src/test/scala/whisk/core/loadBalancer/test/LoadBalancerDataTests.scala
```diff
@@ -17,70 +17,108 @@
...
+
+
+  def await[A](f: Future[A], timeout: FiniteDuration = 1.second) = Await.result(f, timeout)
+
   behavior of "LoadBalancerData"
 
   it should "return the number of activations for a namespace" in {
-
-    val loadBalancerData = new LoadBalancerData()
+    val loadBalancerData = new DistributedLoadBalancerData()
```
markusthoemmes: Do we lose testcoverage for the local version? Intended?

tests/src/test/scala/whisk/core/loadBalancer/test/SharedDataServiceTests.scala
```diff
@@ -0,0 +1,106 @@
...
+    val s = ActorSystem("controller-actor-system", config)
+
+    val sharedDataService = s.actorOf(SharedDataService.props("Candidates"), name = "busyMan")
+    implicit val timeout = Timeout(5.seconds)
+
+    it should "retrieve an empty map after initialization" in {
+      sharedDataService ! GetMap
+      expectMsgPF() {
+        case x: Map[String, BigInt] if x.size == 0 => true
+      }
```
markusthoemmes: Could do a more narrow check here:

```scala
expectMsg(Map.empty[String, BigInt])
```

tests/src/test/scala/whisk/core/loadBalancer/test/SharedDataServiceTests.scala
```diff
@@ -0,0 +1,106 @@
...
+      sharedDataService ! (RemoveCounter("Fifi"))
+      sharedDataService ! (ReadCounter("Fifi"))
+      expectMsg(None)
+    }
+    it should "receive the map with all counters" in {
+      sharedDataService ! (IncreaseCounter("Hilary", 1))
+      sharedDataService ! GetMap
+      expectMsgPF() {
+        case x: Map[String, BigInt] if x.size == 2 => true
+      }
```
markusthoemmes: Could use a more narrow check I suppose

```scala
expectMsg(Map(...))
```

tests/src/test/scala/whisk/core/loadBalancer/test/SharedDataServiceTests.scala
```diff
@@ -0,0 +1,106 @@
...
+      .withValue("akka.remote.netty.tcp.hostname", ConfigValueFactory.fromAnyRef("127.0.0.1"))
+      .withValue("akka.remote.netty.tcp.port", ConfigValueFactory.fromAnyRef(port))
+      .withValue("akka.cluster.auto-down-unreachable-after", ConfigValueFactory.fromAnyRef("10s"))
+      .withValue("akka.actor.provider", ConfigValueFactory.fromAnyRef("cluster"))
+      .withValue("akka.remote.log-remote-lifecycle-events", ConfigValueFactory.fromAnyRef("off"))
+      .withFallback(ConfigFactory.load())
+
+    val s = ActorSystem("controller-actor-system", config)
+
+    val sharedDataService = s.actorOf(SharedDataService.props("Candidates"), name = "busyMan")
```
markusthoemmes: Shouldn't we create a new actor per test instead of leaking it across?

tests/src/test/scala/whisk/core/loadBalancer/test/SharedDataServiceTests.scala
```diff
@@ -0,0 +1,106 @@
...
+    """
+}
+
+class SharedDataServiceTests()
+    extends TestKit(ActorSystem("ControllerCluster", ConfigFactory.parseString(TestKitConfig.config)))
+    with ImplicitSender
+    with FlatSpecLike
+    with Matchers
+    with BeforeAndAfterAll
+    with BeforeAndAfterEach {
```
markusthoemmes: Unused.

common/scala/build.gradle
```diff
@@ -13,10 +13,12 @@ dependencies {
 
     compile 'io.spray:spray-json_2.11:1.3.3'
 
-    compile 'com.typesafe.akka:akka-actor_2.11:2.4.16'
-    compile 'com.typesafe.akka:akka-slf4j_2.11:2.4.16'
+
+    compile 'com.typesafe.akka:akka-actor_2.11:2.5.3'
+    compile 'com.typesafe.akka:akka-stream_2.11:2.5.3'
```
markusthoemmes: Version mismatch, please update all akka deps to 2.5.4.

common/scala/src/main/scala/whisk/common/Config.scala
```diff
@@ -146,11 +145,9 @@ object Config {
    * @param required a key-value map where the keys are required properties
    * @param properties a set of properties to check
    */
-  def validateProperties(required: Map[String, String], properties: Map[String, String])(
-    implicit logging: Logging): Boolean = {
+  def validateProperties(required: Map[String, String], properties: Map[String, String]): Boolean = {
     required.keys.forall { key =>
       val value = properties(key)
-      if (value == null) logging.error(this, s"required property $key still not set")
```
rabbah: why removing the loggers?
this is particularly useful when a deployment fails and you dont know why.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
rabbah: Agree - seems like base class functionality.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
rabbah: I disagree - there's been many occasions when there were mismatches in the config parameters, deployment fails, and one doesn't know why.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -169,11 +168,13 @@ object Controller {
...
 
   def main(args: Array[String]): Unit = {
-    implicit val actorSystem = ActorSystem("controller-actor-system")
-    implicit val logger = new AkkaLogging(akka.event.Logging.getLogger(actorSystem, this))
 
     // extract configuration data from the environment
     val config = new WhiskConfig(requiredProperties)
+
+    implicit val actorSystem = ActorSystem("controller-actor-system", SharedDataService.addAkkaSeedNodesToConf(config))
+    implicit val logger = new AkkaLogging(akka.event.Logging.getLogger(actorSystem, this))
```
rabbah: Was the logger removed because of a phase ordering here?

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -130,13 +130,17 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
...
     logging.info(this, s"checking user '${user.subject}' has not exceeded activation quota")
 
-    checkSystemOverload(ACTIVATE) orElse {
-      checkThrottleOverload(!invokeRateThrottler.check(user), tooManyRequests)
-    } orElse {
-      checkThrottleOverload(!concurrentInvokeThrottler.check(user), tooManyConcurrentRequests)
-    } map {
-      Future.failed(_)
-    } getOrElse Future.successful({})
+    checkSystemOverload(ACTIVATE).flatMap { systemOverloaded =>
```
rabbah: the changes in this file are out of date.

core/controller/src/main/scala/whisk/core/loadBalancer/DistributedLoadBalancerData.scala
```diff
@@ -0,0 +1,89 @@
...
+  private val sharedStateInvokers = actorSystem.actorOf(
+    SharedDataService.props("Invokers"),
+    name =
+      "SharedDataServiceInvokers" + UUID())
+  private val sharedStateNamespaces = actorSystem.actorOf(
+    SharedDataService.props("Namespaces"),
+    name =
+      "SharedDataServiceNamespaces" + UUID())
+
+  val logging = Logging(actorSystem, sharedStateInvokers)
```
rabbah: is this introducing a logger that's different from the openwhisk system logger?

core/controller/src/main/scala/whisk/core/loadBalancer/DistributedLoadBalancerData.scala
```diff
@@ -0,0 +1,89 @@
...
+      entry
+    })
+  }
+
+  def removeActivation(entry: ActivationEntry): Option[ActivationEntry] = {
+    activationsById.remove(entry.id).map { x =>
+      sharedStateInvokers ! DecreaseCounter(entry.invokerName.toString, 1)
+      sharedStateNamespaces ! DecreaseCounter(entry.namespaceId.asString, 1)
+      logging.debug(s"decreased shared counters")
+      x
```
rabbah: do you mind renaming `x` to a more descriptive name?

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -95,7 +100,12 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
   private val blackboxFraction: Double = Math.max(0.0, Math.min(1.0, config.controllerBlackboxFraction))
   logging.info(this, s"blackboxFraction = $blackboxFraction")(TransactionId.loadbalancer)
 
-  private val loadBalancerData = new LoadBalancerData()
+  /** Feature switch for shared load balancer data **/
+  private val loadBalancerData = {
+    if (config.controllerLocalBookkeeping) {
```
rabbah: great use of a feature toggle. thank you!

core/controller/src/main/scala/whisk/core/loadBalancer/LocalLoadBalancerData.scala
```diff
@@ -0,0 +1,77 @@
...
+ * concurrent reads can return stale values (especially the counters returned).
+ */
+class LocalLoadBalancerData() extends LoadBalancerData {
+
+  private val activationByInvoker = TrieMap[InstanceId, AtomicInteger]()
+  private val activationByNamespaceId = TrieMap[UUID, AtomicInteger]()
+  private val activationsById = TrieMap[ActivationId, ActivationEntry]()
+  private val totalActivations = new AtomicInteger(0)
+
+  override def totalActivationCount: Future[Int] = Future.successful(totalActivations.get)
```
rabbah: should some of these impls move to the base class?
ie this is the default base class implementation (local)?

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
```diff
@@ -0,0 +1,131 @@
...
+    node.subscribe(self, initialStateMode = InitialStateAsEvents, classOf[MemberEvent], classOf[UnreachableMember])
+    replicator ! Update(storage, PNCounterMap.empty[String], writeLocal)(_.remove(node, "0"))
+  }
+  override def postStop(): Unit = node.unsubscribe(self)
+
+  /**
+   * CRUD operations on the counter, process cluster member events for logging
+   * @return
+   */
+  def receive = {
```
rabbah: why have curlies around case matches?

core/controller/src/main/scala/whisk/core/loadBalancer/LocalLoadBalancerData.scala
markusthoemmes: I thought along the same lines but ended up thinking it might be more confusing to deduplicate 1 or 2 methods vs. having them all implemented in one spot. This impl in particular is not the same for local vs. distributed.

tests/build.gradle
```diff
@@ -48,9 +48,9 @@ dependencies {
     compile 'junit:junit:4.11'
     compile 'com.jayway.restassured:rest-assured:2.6.0'
     compile 'org.scalatest:scalatest_2.11:3.0.1'
+    compile 'com.typesafe.akka:akka-testkit_2.11:2.5.3'
```
markusthoemmes: Version mismatch. Should this be 2.5.4?

core/controller/build.gradle
```diff
@@ -11,6 +11,7 @@ repositories {
 }
 
 dependencies {
+    compile 'com.typesafe.akka:akka-distributed-data_2.11:2.5.3'
```
markusthoemmes: Version mismatch.

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
```diff
@@ -0,0 +1,122 @@
...
+   * https://github.com/typesafehub/config/issues/69
+   * @return Updated Config
+   */
+  def addAkkaSeedNodesToConf(whiskConf: WhiskConfig): Config = {
+    val conf = ConfigFactory.load()
+
+    val cluster = conf.getString("akka.actor.provider")
+
+    if (cluster == "cluster") {
+      val seedNodes = whiskConf.controllerSeedNodes
```
tysonnorris: configuring seed nodes statically is very limiting; it would be better to make this extensible, so that various deployment types (kube, mesos) can dictate how "seed nodes" are resolved at runtime, and then init the Akka cluster explicitly AFTER seed nodes are determined (dynamically or from config): 
```
    val seedNodes = SeedConfigSource.getSeedNodes(config)
    Cluster(system).joinSeedNodes(seedNodes.toList)
```
If you make SeedConfigSource either an SPI or a class that allows various approaches, it would enable different deployments the flexibility to register seed nodes as they see fit.

ddragosd: 👍  We know Mesos and Kube are in the picture and it's not a big stretch to provide support to automatically discover seed nodes using the APIs they provide. Mesos could come later as 1 PR and so does Kube, but it would be good to design the code with this in mind.

markusthoemmes: Not too familiar with both Kube or Mesos APIs but does it make a difference if you determine the seed nodes outside of the component via whichever API and then pass it on startup vs. having the component look them up on startup? Does that break Kube/Mesos deployment models?

vvraskin: The idea of this PR was the initial enablement of akka cluster functionality (see feature toggle), we can iterate over it once we put different deployment models into picture. Probably need to take time to familiarize myself with Mesos and Kube.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -130,7 +130,7 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
 
     logging.info(this, s"checking user '${user.subject}' has not exceeded activation quota")
     checkSystemOverload(ACTIVATE)
-      .flatMap(_ => checkThrottleOverload(invokeRateThrottler.check(user)))
+      .flatMap(_ => checkThrottleOverload(Future(invokeRateThrottler.check(user))))
```
markusthoemmes: Should be `Future.succesful(invokerRateThrottler.check(User))`

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -253,9 +254,9 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
     implicit transid: TransactionId): Future[Unit] = {
     if (right == ACTIVATE) {
       if (resources.exists(_.collection.path == Collection.ACTIONS)) {
-        checkThrottleOverload(invokeRateThrottler.check(user))
+        checkThrottleOverload(Future(invokeRateThrottler.check(user)))
```
markusthoemmes: See above on `Future.successful`

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -253,9 +254,9 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
...
       } else if (resources.exists(_.collection.path == Collection.TRIGGERS)) {
-        checkThrottleOverload(triggerRateThrottler.check(user))
+        checkThrottleOverload(Future(triggerRateThrottler.check(user)))
```
markusthoemmes: See above on Future.successful.

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
```diff
@@ -0,0 +1,122 @@
...
+    case (IncreaseCounter(key, increment)) => {
+      replicator ! Update(storage, PNCounterMap.empty[String], writeLocal)(_.increment(key, increment))
+    }
+    case (DecreaseCounter(key, decrement)) => {
+      replicator ! Update(storage, PNCounterMap[String], writeLocal)(_.decrement(key, decrement))
+    }
+
+    case GetMap => {
+      replicator ! Get(storage, readLocal, request = Some((sender())))
+    }
```
markusthoemmes: Please remove the curlies so the whole case block is consistent.

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
tysonnorris: @markusthoemmes @vvraskin the current impl launches the container system with the assumption that the config contains seed nodes, i.e. the actor system initiates clustering automatically via `implicit val actorSystem = ActorSystem("controller-actor-system", SharedDataService.addAkkaSeedNodesToConf(config))`. I'm suggesting that the more flexible approach is to explicitly initiate clustering at some time after the actorsystem is started, using `Cluster(system).joinSeedNodes(seedNodesList)`, which does change the semantics of startup slightly, but allows situations like "actor system and remoting can initialize, before clustering gossip begins", which is good form for doing things like "only allow clustering to proceed if the startup proceeds to the point where remoting can be enabled". Otherwise you are likely to have a noisey (at least) or failure (at worst) during startup in a more dynamic env, or even in a static env, if the seed nodes are not started before the nodes that attempt to join as a cluster (esp in the case where "all the nodes are seed nodes")

vvraskin:
@tysonnorris @ddragosd 
Thanks for your feedback about the seed nodes, I made a small change to join the seed nodes at the startup of the `SharedDataService` service. So later you can extend the trait`SeedNodesProvider` to add kube and mesos support, or create an SPI. Are you ok with this change?

Actually this change has also helped to get rid of a _special_ seed-nodes parsing and get back the logging to the Config classes.

core/controller/src/main/scala/whisk/core/loadBalancer/SeedNodesProvider.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import akka.actor.Address
+
+import scala.collection.immutable.Seq
+
+trait SeedNodesProvider {
+  def getSeedNodes(): Seq[Address]
+}
+
+class StaticSeedNodesProvider(seedNodes: String, actorSystemName: String) extends SeedNodesProvider {
```
tysonnorris: Can you change this signature to be `StaticSeedNodesProvider(config: WhiskConfig, actorSystem: ActorSystem)`? otherwise it will need to either be wrapped or further refactored to make it extensible in the future

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
```diff
@@ -33,35 +32,14 @@ case class RemoveCounter(key: String)
...
-      val configWithSeedNodes = ConfigFactory.parseString(s"akka.cluster.seed-nodes=[${nodes.mkString(",")}]")
-      configWithSeedNodes.withFallback(conf)
-    } else conf
-  }
+  def props(storageName: String, seedNodes: Seq[Address]): Props =
+    Props(new SharedDataService(storageName, seedNodes))
 }
 
-class SharedDataService(storageName: String) extends Actor with ActorLogging {
+class SharedDataService(storageName: String, seedNodes: Seq[Address]) extends Actor with ActorLogging {
```
tysonnorris: `SharedDataService` appears to be core to enabling clustering, so I'm wondering if it should be less specific to data storage, and rather expose `DataService` as a feature; i.e. `SharedDataService` -> `WhiskCluster` and add `WhiskCluster.dataService`. I don't think enabling clustering should be taken lightly, so this would make it more obvious that it would be the one (and only) entry point to initiating cluster support within either controller or invoker; and if there are cases where invoker clustering should be enabled, move this into `/common/scala` instead of controller. WDYT?

core/controller/src/main/scala/whisk/core/loadBalancer/SeedNodesProvider.scala
vvraskin: I think that passing only what is indeed _needed_ to a method is the right approach in this case, e.g. two strings instead of the overhead of the rest of the whiskConfig and the actor system. It makes it easier to test, so no extra mocking is required. If another implementation would be needed, one could create another class extending `SeedNodesProvider` with custom parameters and no interference.

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
vvraskin: Do you have an example of another use case where clustering could be useful for controller/invoker?

markusthoemmes: While I agree with you in general @tysonnorris, I'd say that here we don't yet have evidence of any clustering needed besides the SharedDataService so I'd leave this for now. If we want to make use of clustering in a different, more generalized manner we can always refactor as needed.

core/controller/src/main/scala/whisk/core/loadBalancer/DistributedLoadBalancerData.scala
```diff
@@ -0,0 +1,107 @@
...
+    val nodes = seedNodesProvider.getSeedNodes()
+    if (nodes.size > 0)
+      nodes
+    else {
+      logging.error(
+        this,
+        "Either seed nodes were not provided or not specified in the correct " +
+          "format, e.g. whitespace " +
+          "separated <IP:PORT>")
+      sys.exit(1)
```
cbickel: Can we avoid exiting anywhere in the code?
Can we instead let the creation of `DistributedLoadBalancerData` let fail, which lets the creation of the Loadbalancer fail?

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
tysonnorris: @vvraskin @markusthoemmes one example is downing nodes - since enabling the clustering is done in SharedDataService, will it also handle downing cluster nodes? Are the semantics for that explicitly available, or are you assuming that auto-downing is enabled, or that default behavior is required to be used in all cases? For example, in the same way that seednodes needs to be specified in different ways to support different deployments, downing behavior should also be configurable, and its not clear how that will happen, except for `Cluster(system).down()` at some unrelated location - while this is technically possible, it seems leaky to spread this around. WDYT?

core/controller/src/main/scala/whisk/core/loadBalancer/SeedNodesProvider.scala
tysonnorris: Maybe what isn't clear is that way SPIs work is that the behave as factories, where each impl uses the same interface, so this impl will have to be refactored to become extensible in any way. To prepare for that you would need to have an extra class `StaticSeedNodesProviderFactory` which knows how to construct your `StaticSeedNodesProvider` using that existing constructor - my ask is to create both classes now, so that making it extensible in the future is ONLY a matter of how `StaticSeedNodesProviderFactory` is instantiated.

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
```diff
@@ -75,6 +53,7 @@ class SharedDataService(storageName: String) extends Actor with ActorLogging {
    * Subscribe this node for the changes in the Map, initialize the Map
    */
   override def preStart(): Unit = {
+    node.joinSeedNodes(seedNodes)
```
ddragosd: I'm a little unsure about `joinSeedNodes` to be used from within an actor. IIUC this command has a cluster wide impact, it's not only related to this actor, in which case, wouldn't it make sense to surface it in a more global place ?

bwmcadams:
@vvraskin 

1) I've been giving some thought to clustering usage… but still absorbing the code base. I may have some wisdom to add to that one down the line
2) auto-downing is, as you point out, generally not a good thing. It's practically begging for split-brain issues. Lightbend has a really nice split-brain-resolver plugin that helps with these kinds of things but it is unfortunately a subscription product so not available for OpenWhisk. 
    - I suspect the best approach to handling failure detection and related items is for OpenWhisk to do testing, come up with some base defaults and possibly documented "alternate defaults based on use case", but let the user turn some knobs with big red blinking lights and buttons marked "do not touch unless you are absolutely sure what you are doing"
    - Clustering & remoting are meant to be referentially transparent versus local actors, so not *entirely* sure what you mean about adding complexity, if you could clarify? That is to say, my experience has been that done right, we should be able to implement this where it works well without much code change for both local and remote (remote including clustered).

core/controller/src/main/scala/whisk/core/loadBalancer/SeedNodesProvider.scala
tysonnorris: Pointing out that it is needed makes it not pre-mature, but if you are not comfortable implementing it, someone will create a PR for this in the near future.

tysonnorris:
@vvraskin to your points: 
1 - in controller immediately, no. In controller in the near future, yes, e.g. in a mesos case we will need to use either ClusterSingleton or MemberEvent to designate a leader; we will also need to share some state within the cluster (may be able to leverage SharedDataService, or else directly use DistributedData)
2 - my concern with downing is the case where nodes are down, and not coming back (or not coming back with the same id). In this case, docs indicate that new members cannot become "up", etc, till the unreachable node is downed. So regardless of whether traffic is routed to an unreachable node, if clustering usage is affected by having an unreachable, but NOT downed node, then there will be other types of problems (e.g. can replacement nodes be added? can the cluster size be grown?). Separately from needing the ability to cause nodes to be downed, the semantics of determining down vs *temp* unreachable will be different in a cluster manager like mesos/kube, similar to the semantics of determining seed nodes.

ddragosd:
:+1: @bwmcadams  @tysonnorris

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -95,7 +93,18 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
...
+      new LocalLoadBalancerData()
+    } else {
+
+      /** Specify how seed nodes are generated */
+      val seedNodesProvider = new StaticSeedNodesProvider(config.controllerSeedNodes, actorSystem.name)
+      Cluster(actorSystem).joinSeedNodes(seedNodesProvider.getSeedNodes())
```
ddragosd: 👍  @vvraskin . Right now we're assuming the LB is the show runner, which I think it's true. We'll see more later when we get to separate the DataPlane(LB) from the Control Plane 😄

ddragosd:
I'm also wondering what's the most performant model to distribute billions of records using distributed data with Akka.  Should we have billions of keys, or should we partition them in groups, distributing groups of N number of keys (where 1 update of 1 key in a group, causes the whole group to be updated in the cluster) ? Assuming `ReadLocal`/ `WriteLocal` is fine, and that all keys fit in each JVM's memory w/o any problems, will either one of these methods facilitate a higher performance, or optimal noise in the network ?

@bwmcadams any suggestions from your side ?

bwmcadams:
At the risk of a vapid comment, I need to do a bit of refresher research  on this (also boarding an airplane). Let me give it some thought tonight. Hmm

vvraskin:
@bwmcadams 
2. What I meant on additional complexity is that if we have cluster management outside of the actors, that could imply implementing an additional actor to monitor the health state of another nodes. I don't believe its the right timing to do that, the costs of not having the freshest state are lower then estimated effort to implement it. And I don't believe its the scope of this PR as well. 
Back to your point wrt to the measurements you mentioned. I think its a tricky one, given non-deterministic failure conditions, environment differences, etc, its difficult to guarantee that one size fits all property would work here.

vvraskin:
@tysonnorris 
1. Given that the cluster is now started from within the `LoadBalancerService`, it should be good enough for near future extensions. 
2. I think you mean the case of network partitioning in between two controllers when the traffic could still flow between nginx and that controller. In this case we could probably set the hard auto-downing timeout - 5 mins, mark it as unreachable and the broken node could join the cluster again. (back to the point of @bwmcadams with a measured value, however, this is just a high enough number). But even without this timeout, the node should become green again, once the partitioning is resolved. 
However, if you speak about some other hypothetical cases when the network is fine, but the node is unreachable und not downed. Then restarting the node should automatically unregister it from the cluster and give it a chance to join it again (actor should signal to the cluster upon going down).

I think we are a bit far way from the initial discussion we have started upon the changes requested on this PR. Do you mind approving or you still think some changes are required?

core/controller/src/main/scala/whisk/core/loadBalancer/SeedNodesProvider.scala
vvraskin: It is not about being not comfortable. Pre-mature in this case means that you are asking for the change that won't be used in our production env now, and probably by anyone else yet. The mesos and kube support is not officially supported yet, and I believe that an spi should come with another feature and it is not necessary for this particular PR.

ddragosd:
@vvraskin how are nodes downed ? I think we still need to add logic for downing the nodes. WDYT ?

vvraskin:
@ddragosd at the moment we don't down the nodes, they must unregister themselves, in worse case scenario nodes are operating on their own vision of the number of activations on invokers/namespaces until the network outage is resolved. Do you have an idea on how we can do it better?

tysonnorris:
@vvraskin I think the worst case scenario is that the node crashes and does not unregister itself ever. This is the point I think Dragos is getting at. If the split brain cluster is a tolerable state as you mention, then maybe it would be good for a) unreachable nodes to be downed and b) later extend this so that downing logic can be extensible (where some external systems may be better equipped to determine the distinction between crash and partition). 

However, the risks involved with running in this split mode (e.g. controllers lose contact and do not share state) should be clearly documented (e.g. what is the impact to invoker routing logic? is there any point of recovery possible, etc)

vvraskin:
Let me summarise what we have already discussed with Tyson over slack. 
Given the deployment model when a container receives a new ip address assigned after the restart. In our case if one controller has failed and rejoined the cluster as a new member, there will be some garbage accumulated in the list of cluster members. It is not harmful per se, e.g. the cluster is still running, however healthy cluster nodes will be still gossiping with a non-existing container. If assigning static IP addresses is not an option, in order to avoid this case one could use auto-downing feature in akka cluster, which allows to a cluster leader to mark the failing node as down and remove it from the cluster. To prevent cluster partitioning due to several leaders this property must be set to a relatively high value. The number is not deterministic and could be defined based on the further ops experience. 
One could also think of another monitoring component that is going to watch the cluster state and down the failed nodes on demand. But this solution is rather an expensive one and I think it doesn't belong to the context of this PR.

tysonnorris:
@vvraskin thanks for updating docs and the default! LGTM

PR #2532, opened 2017-07-26
Update wsk test fwk to allow tests to obfuscate sensitive output

mdeuser:


mdeuser:
pg4/598

mdeuser:
@dubeejw - for your review please

tests/src/test/scala/common/Wsk.scala
```diff
@@ -1040,13 +1040,14 @@ trait RunWskCmd extends Matchers {
             env: Map[String, String] = Map("WSK_CONFIG_FILE" -> ""),
             workingDir: File = new File("."),
             stdinFile: Option[File] = None,
-            showCmd: Boolean = false): RunResult = {
+            showCmd: Boolean = false,
+            hideFromOutput: Seq[String] = Seq()): RunResult = {
         val args = baseCommand
         if (verbose) args += "--verbose"
-        if (showCmd) println(args.mkString(" ") + " " + params.mkString(" "))
+        if (showCmd) println(hideStr(args.mkString(" ") + " " + params.mkString(" "), hideFromOutput))
```
rabbah: hmm show cmd to me is used for debugging - if i'm debugging, dont think i want anything hidden.
do we actually turn this flag on in our tests?

mdeuser: Yep, in some places.  IMHO, I kind of like seeing the actual commands used within a test.  When troubleshooting a failure, it's easy to see what's going on from the logs.. and it's possible to copy/paste the command(s) from the log to manually recreate each step in the test.

rabbah:
Ok so let's turn off all command verbosity :) they should not run in CI. I don't see why we need them.

mdeuser:
By default, no hiding of any output is done by this PR.  Only if the caller supplies filtering strings will filtering take place.  Setting the `showCmd` flag on it's own will continue to display the output.  Individual tests can leverage this facility or not.

rabbah:
i'm suggesting that any test that has show cmd = true is likely a mistake for CI and should be turned off; do you see a reason we want to show the cmd under ci?

mdeuser:
i guess i'm suggesting that the discussion of whether or not a test uses the showCmd flag be considered outside this pr.  by default the showCmd flag is false.

rabbah:
ok then i'm confused - if we agree (and i thought we did) that show cmd is useful for debugging then we should not hide strings in those commands (instead hide the strings in the with clue error message only).

so i suggest backing out the change for the `if (showCmd) ...` bit.
if then there's a concern that there are tests which show the cmd and hence printing sensitive info, all of those instances should be turned to off.

if you want a second pr to turn off all instances of show cmd, fine. but i would still remove the change for showCmd.

mdeuser:
ok.  hide is removed from the showCmd output.

pg1/926

PR #2533, opened 2017-07-26
Add the ability to define user-specific throttles.

markusthoemmes:
This adds the ability to override the rate/concurrency related limits of the system on a per-namespace basis to be able to adapt to specific user needs.

Implementation:
User defined throttles are placed in an extra document in couchdb, $namespace/limits. That document is automatically processed by the existing view, thus no extra request is needed. If no such document exist, default limits will apply.

**Todo:**

- [x] Tests
- [x] Add admin functionality to `wskadmin`

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/Identity.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Identity.scala
@@ -27,8 +27,15 @@ import whisk.core.database.MultipleReadersSingleWriterCache
 import whisk.core.database.NoDocumentException
 import whisk.core.entitlement.Privilege
 import whisk.core.entitlement.Privilege.Privilege
+import scala.util.Try
 
-protected[core] case class Identity(subject: Subject, namespace: EntityName, authkey: AuthKey, rights: Set[Privilege]) {
+case class UserLimits(invocationsPerMinute: Option[Int] = None, concurrentInvocations: Option[Int] = None, firesPerMinute: Option[Int] = None)
+
+object UserLimits extends DefaultJsonProtocol {
+    implicit val serdes = jsonFormat3(UserLimits.apply)
+}
+
+protected[core] case class Identity(subject: Subject, namespace: EntityName, authkey: AuthKey, rights: Set[Privilege], limits: UserLimits = UserLimits()) {
     def uuid = authkey.uuid
 }
 
@@ -38,7 +45,7 @@ object Identity extends MultipleReadersSingleWriterCache[Identity, DocInfo] with
 
     override val cacheEnabled = true
     override def cacheKeyForUpdate(i: Identity) = i.authkey
-    implicit val serdes = jsonFormat4(Identity.apply)
+    implicit val serdes = jsonFormat5(Identity.apply)
 
     /**
      * Retrieves a key for namespace.
@@ -95,20 +102,21 @@ object Identity extends MultipleReadersSingleWriterCache[Identity, DocInfo] with
             endKey = key,
             skip = 0,
             limit = limit,
-            includeDocs = false,
+            includeDocs = true,
             descending = true,
             reduce = false)
     }
 
     private def rowToIdentity(row: JsObject, key: String)(
         implicit transid: TransactionId, logger: Logging) = {
-        row.getFields("id", "value") match {
-            case Seq(JsString(id), JsObject(value)) =>
+        row.getFields("id", "value", "doc") match {
+            case Seq(JsString(id), JsObject(value), doc) =>
+                val limits = Try(doc.convertTo[UserLimits]).getOrElse(UserLimits())
                 val subject = Subject(id)
                 val JsString(uuid) = value("uuid")
                 val JsString(secret) = value("key")
                 val JsString(namespace) = value("namespace")
-                Identity(subject, EntityName(namespace), AuthKey(UUID(uuid), Secret(secret)), Privilege.ALL)
+                Identity(subject, EntityName(namespace), AuthKey(UUID(uuid), Secret(secret)), Privilege.ALL, limits)
             case _ =>
                 logger.error(this, s"$viewName[$key] has malformed view '${row.compactPrint}'")
                 throw new IllegalStateException("identities view malformed")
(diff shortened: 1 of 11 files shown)


markusthoemmes:
PG3 925 🔵

common/scala/src/main/scala/whisk/core/entity/Identity.scala
```diff
@@ -27,8 +27,15 @@ import whisk.core.database.MultipleReadersSingleWriterCache
 import whisk.core.database.NoDocumentException
 import whisk.core.entitlement.Privilege
 import whisk.core.entitlement.Privilege.Privilege
+import scala.util.Try
 
-protected[core] case class Identity(subject: Subject, namespace: EntityName, authkey: AuthKey, rights: Set[Privilege]) {
+case class UserLimits(invocationsPerMinute: Option[Int] = None, concurrentInvocations: Option[Int] = None, firesPerMinute: Option[Int] = None)
```
rabbah: should the user limits be itself the option rather than a class of fields that are all options?
i suppose this is more fine grained and gives more control - i wonder though if in general when changing throttles per user if we're not changing all these values at once.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -255,7 +255,9 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
         def userThrottled = {
             val isInvocation = resources.exists(_.collection.path == Collection.ACTIONS)
             val isTrigger = resources.exists(_.collection.path == Collection.TRIGGERS)
-            (isInvocation && !invokeRateThrottler.check(user)) || (isTrigger && !triggerRateThrottler.check(user))
+            (isInvocation &&
```
rabbah: is this formatting intentional?

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -82,8 +82,8 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
 
     private implicit val executionContext = actorSystem.dispatcher
 
-    private val invokeRateThrottler = new RateThrottler("actions per minute", config.actionInvokePerMinuteLimit.toInt)
-    private val triggerRateThrottler = new RateThrottler("triggers per minute", config.triggerFirePerMinuteLimit.toInt)
+    private val invokeRateThrottler = new RateThrottler("actions per minute", config.actionInvokePerMinuteLimit.toInt, _.limits.invocationsPerMinute)
```
rabbah: the _ is a nice convenience but it took me a pause to understand what's happening here (I had to check the signature for the rate throttler - not sure if it's worth being more verbose for clarity now... maybe.

core/controller/src/main/scala/whisk/core/entitlement/RateThrottler.scala
```diff
@@ -67,7 +68,7 @@ private class RateInfo(maxPerMinute: Int) {
      * Increments operation count in the current time window by
      * one and checks if still below allowed max rate.
      */
-    def check(): Boolean = {
+    def check(maxPerMinute: Int): Boolean = {
```
rabbah: perhaps a comment here that the check always accepts the max allowed per minute since throttles may change dynamically. (maybe here or in the caller above.)

tests/src/test/scala/limits/ThrottleTests.scala
```diff
@@ -278,3 +280,89 @@ class ThrottleTests
...
+        disposeAdditionalTestSubject(zeroProps.namespace)
+        disposeAdditionalTestSubject(zeroConcProps.namespace)
+        disposeAdditionalTestSubject(oneProps.namespace)
+    }
+
+    behavior of "Namespace-specific throttles"
+
+    it should "respect overridden rate-throttles of 0" in withAssetCleaner(zeroProps) {
+        (wp, assetHelper) =>
+            implicit val props = wp
```
rabbah: does it make sense to make these just unit tests against a entitlement provider vs integration tests?

tests/src/test/scala/limits/ThrottleTests.scala
```diff
@@ -278,3 +280,89 @@ class ThrottleTests
...
+            assetHelper.withCleaner(wsk.action, actionName) {
+                (action, _) => action.create(actionName, defaultAction)
+            }
+            assetHelper.withCleaner(wsk.trigger, triggerName) {
+                (trigger, _) => trigger.create(triggerName)
+            }
+
+            // One invoke should be allowed, the second one throttled
+            wsk.action.invoke(actionName)
+            wsk.action.invoke(actionName, expectedExitCode = TestUtils.THROTTLED)
```
rabbah: could check the error message to make sure it's the rate per minute throttle and not the concurrent throttle for tighter check.

tests/src/test/scala/whisk/core/controller/test/RateThrottleTests.scala
```diff
@@ -27,6 +27,7 @@ import org.scalatest.junit.JUnitRunner
 import common.StreamLogging
 import whisk.common.TransactionId
 import whisk.core.entitlement._
+import whisk.core.entity.UserLimits
```
rabbah: i see you did that here already.
very well then :D

tests/src/test/scala/limits/ThrottleTests.scala
markusthoemmes: I opted for these very minimal integration tests, because most of the work is threading values through the system. They caught lots of tiny bugs while developing so I figured it might be worth keeping them.

common/scala/src/main/scala/whisk/core/entity/Identity.scala
markusthoemmes: I thought about this as well, but I figured this is more dynamic as in: If we add more namespace-definable limits in the future (and i expect us to) the schema is easily adapted to that vs. making the whole class `Option`al.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
markusthoemmes: not needed anymore, was amn artifact of an earlier version, thx.

markusthoemmes:
#2530 should be merged first, the first commit is taken from there to prevent merge conflicts.

rabbah:
still had to resolve a conflict - will wait for travis.

Issue #2539, opened 2017-07-27
vagrant redeploy instructions fail


boneskull:
I have restarted my Vagrant machine via `vagrant reload` (host is macOS Sierra).

I'm following the instructions [here](https://github.com/apache/incubator-openwhisk/blob/master/tools/vagrant/README.md#safe-re-deploy-after-vm-restart):

```sh
vagrant ssh
cd ${ANSIBLE_HOME}
# teardown all containers expect couchdb container
ansible-playbook -i environments/local openwhisk.yml -e mode=clean
# deploy openwhisk containers
ansible-playbook -i environments/local openwhisk.yml
```

However, the `deploy` task of the `controller` role fails:

```
TASK [controller : pull the latest image of controller] ***********************************************
Thursday 27 July 2017  23:39:56 +0000 (0:00:00.044)       0:00:55.389 *********
skipping: [172.17.0.1]

TASK [controller : ensure controller log directory is created with permissions] ***********************
Thursday 27 July 2017  23:39:56 +0000 (0:00:00.041)       0:00:55.431 *********
ok: [172.17.0.1]

TASK [controller : (re)start controller] **************************************************************
Thursday 27 July 2017  23:39:56 +0000 (0:00:00.220)       0:00:55.651 *********
fatal: [172.17.0.1]: FAILED! => {"changed": false, "failed": true, "msg": "Error pulling whisk/controller - code: None message: Error: image whisk/controller:latest not found"}
```

I don't know what this implies.  Keep in mind this is immediately after `vagrant reload`, not having interacted with the machine in any way prior.

Here's what's running in this state, in case something is missing:

```sh
vagrant@vagrant-ubuntu-trusty-64:~/openwhisk/ansible$ docker ps --format '{{.Image}} {{.Command}} {{.Names}}' --no-trunc
ches/kafka:0.10.2.1 "/start.sh" kafka
zookeeper:3.4 "/docker-entrypoint.sh zkServer.sh start-foreground" zookeeper,kafka/zookeeper
consul:0.7.0 "docker-entrypoint.sh consul agent -config-dir /consul/config" consul
gliderlabs/registrator "/bin/registrator -ip 172.17.0.1 -resync 2 consul://172.17.0.1:8500" registrator
openwhisk/apigateway:0.8.2 "/usr/local/bin/dumb-init -- /etc/init-container.sh" apigateway
redis:3.2 "docker-entrypoint.sh redis-server" redis
[...]

PR #2540, opened 2017-07-27
don't attempt to run registrator if already running

boneskull:
While attempting to redeploy after a reload of my Vagrant machine, I was encountering trouble.

It appears the `openwhisk` playbook cannot be run twice in succession, because if the `registrator` container is running, `docker run ...` will fail due to a name collision.  I found this frustrating while attempting to debug the issue (#2539).

This change checks to see if the `registrator` container is running, and if so, skips the attempt to manually start it.

### DIFF
--- a/ansible/roles/consul/tasks/deploy.yml
+++ b/ansible/roles/consul/tasks/deploy.yml
@@ -79,6 +79,7 @@
 
 - name: start registrator using docker cli
   shell: >
+        docker top registrator ||
         docker run -d
         --name registrator
         --hostname registrator


rabbah:
Thanks for this pr - I suspect this pr #2452 which removes the registrator will eschew the need for this patch. @markusthoemmes thoughts?

markusthoemmes:
Thought the same, we should get the consul one in soon.

rabbah:
@boneskull consul is now gone so you shouldn't have a problem with this again.

PR #2545, opened 2017-07-28
Kafka topics are created via Docker Images

DanLavine:
This PR moves the logic for Kafka topic creation from Ansible into individual component's Docker Images. The same health checks for creating the topics are kept, as well as the ability to configure the finer details like retention bytes and ms.

** NOTE **: There will need to be a new Docker image published for the OpenWhisk release for Kafka. I'm no sure where to edit a CI build to push new images.

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -111,7 +111,8 @@ registry:
   confdir: "{{ config_root_dir }}/registry"
 
 kafka:
-  version: 0.10.2.1
+  replication_factor: "{{ kafka_replication_factor | default(1) }}"
+  partitions: "{{ kafka_partitions | default(1) }}"
   port: 9092
   ras:
     port: 9093
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -57,11 +57,19 @@
       "LOADBALANCER_INVOKERBUSYTHRESHOLD": "{{ invoker.busyThreshold }}"
 
       "RUNTIMES_MANIFEST": "{{ runtimesManifest | to_json }}"
+
+      "REPLICATION_FACTOR": "{{ kafka.replication_factor}}"
+      "PARTITIONS": "{{ kafka.partitions }}"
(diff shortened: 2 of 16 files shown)


rabbah:
Hmm this takes us back in time. Why do we need to do this/is this the best solution; can we avoid installing Kafka 3 times, avoiding version mismatches, etc?

DanLavine:
@rabbah. The motivation for this PR is so that deployments that do not want to rely on Ansible. For example, in Kubernetes you would could have a number Pods deployed for the Invoker (say 5), but you want to increase that number to 10.  The way one would normally do this is running `kubectl scale --replicas=10 -f <PATH TO YAML FILE>` (or some other way, update required yaml and reapply it).

However, the current workflow would require us to run the Kafka topic creation before those components come up. I image that other deployment strategies like Docker Compose could have this issue as well.

Lastly, the issue with versioning, I think I can set environment variables in gradle that can feed into the Docker images, need to test it out. Or since both the Controller and Invoker use the same base image, we could create a new Docker image that has Kafka built into it, then use that image for the Controller and Invoker, but not sure if you want another Docker Image.

DanLavine:
Controller's and Invoker's Kafka version are now pulled from a common place. I don't think we can build Kafka from the same version file since the tag feature for the image to pull is to new. Here is the PR that added it [Docker PR](https://github.com/moby/moby/pull/31352/commits). We would need to update the Docker version to 1.13+

dgrove-oss:
If we upgrade to kafka .0.11.0 or better we can use the new adminclient API to create the topics with the desired parameters from within the Scala code of the invoker & controller.

I suggest we close this PR and instead pursue a kafka version bump + using the new APIs to create the topics as they are needed.  This is also a better fit for dynamically varying pools of invokers (since we can create the topics lazily as they are needed).

rabbah:
I like the proposal.

dgrove-oss:
@rabbah see #2948 for the kafka 0.11 approach.

rabbah:
Closing in favor of new approach.

PR #2565, opened 2017-08-03
Use Java based JsonSchema validator

markusthoemmes:
Having to install the python tool is mildly annoying just for tests.

The library is double licensed under LGPL and Apache. Does anybody know if that's fine for us? @csantanapr?


markusthoemmes:
PG1 1903 🔵

PR #2575, opened 2017-08-04
Use Configuration File with Scan Code and Add Missing License Header

dubeejw:


mrutkows:
@bjustin-ibm @dubeejw The reason the ASF header was not found by scancode was that the default .cfg excludes (via [Excludes section) all files under any "tests" directory.  This is historical and there was initial pushback at changing this behavior earlier this year.  However, Rodric merged the last PR yesterday (https://github.com/apache/incubator-openwhisk-catalog/pull/229) for Catalog that perhaps would allow us to remove this "tests" folder exclusion from the .cfg file.  

I pulled the latest for ALL repos and ran scancode against it without the "tests" exclusion and it seems to pass.   Would there be any impacts at removing this that I have not foreseen?

bjustin-ibm:
I'll defer to @dubeejw

PR #2576, opened 2017-08-04
Configure components via pureconfig.

markusthoemmes:
The goal of this is to eventually provide all configuration through TypesafeConfig files and keep defaults very close to the component itself vs. keeping them in ansible. That will enable different deployment methods much more easily and also facilitate quick local testing. For instance the controller should be startable only given CouchDb and Kafka connection data.

This adds the ability to configure the main components (controller/invoker) via pureconfig case-classes vs. using WhiskConfig.

**Benefits are:**

- Automatic type derivation
- Static loading of config vs. passing one instance around
- Proper scoping of configuration
- It's the standard in Scala (based on TypesafeConfig)

**Implementation:**

TypesafeConfig reads JSON-like configuration files, which we can nest as we please. Those namespaces can be read by each component independently.
While TypesafeConfig supports expansion of environment variables, one has to do so explicitly. Which System properties, one can override specific values directly without the added overhead. To achieve the best of both worlds (for containers we only have environment variables available) there is a small script which translates environment variables to system properties in Java.

### DIFF
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -75,27 +75,33 @@
       "LIMITS_ACTIONS_INVOKES_CONCURRENTINSYSTEM": "{{ limits.concurrentInvocationsSystem }}"
       "LIMITS_TRIGGERS_FIRES_PERMINUTE": "{{ limits.firesPerMinute }}"
       "LIMITS_ACTIONS_SEQUENCE_MAXLENGTH": "{{ limits.sequenceMaxLength }}"
-      "CONTROLLER_BLACKBOXFRACTION": "{{ controller.blackboxFraction }}"
-      "LOADBALANCER_INVOKERBUSYTHRESHOLD": "{{ invoker.busyThreshold }}"
 
       "RUNTIMES_MANIFEST": "{{ runtimesManifest | to_json }}"
       "CONTROLLER_LOCALBOOKKEEPING": "{{ controller.localBookkeeping }}"
-      "AKKA_CLUSTER_PORT": "{{ controller.akka.cluster.basePort + groups['controllers'].index(inventory_hostname) }}"
-      "AKKA_CLUSTER_HOST": "{{ controller.akka.cluster.host[groups['controllers'].index(inventory_hostname)] }}"
       "AKKA_CLUSTER_SEED_NODES": "{{seed_nodes_list | join(' ') }}"
-      "AKKA_CLUSTER_BIND_PORT": "{{ controller.akka.cluster.bindPort }}"
-      "AKKA_ACTOR_PROVIDER": "{{ controller.akka.provider }}"
       "METRICS_KAMON": "{{ metrics.kamon.enabled }}"
       "METRICS_LOG": "{{ metrics.log.enabled }}"
-      "METRICS_KAMON_HOST": "{{ metrics.kamon.host }}"
-      "METRICS_KAMON_PORT": "{{ metrics.kamon.port }}"
       "CONTROLLER_HA": "{{ controller.ha }}"
+
+      "METRICS_KAMON": "{{ metrics.kamon.enabled }}"
+      "METRICS_LOG": "{{ metrics.log.enabled }}"
+
+      "CONFIG_whisk_loadbalancer_invokerBusyThreshold": "{{ invoker.busyThreshold }}"
+      "CONFIG_whisk_loadbalancer_blackboxFraction": "{{ controller.blackboxFraction }}"
+
+      "CONFIG_akka_actor_provider": "{{ controller.akka.provider }}"
+      "CONFIG_akka_remote_netty_tcp_hostname": "{{ controller.akka.cluster.host[groups['controllers'].index(inventory_hostname)] }}"
+      "CONFIG_akka_remote_netty_tcp_port": "{{ controller.akka.cluster.basePort + groups['controllers'].index(inventory_hostname) }}"
+      "CONFIG_akka_remote_netty_tcp_bindPort": "{{ controller.akka.cluster.bindPort }}"
+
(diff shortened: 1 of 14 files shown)


rabbah:
Early feedback: awesome.

ddragosd:
:+1:

markusthoemmes:
PG2 2177 🔵

common/scala/build.gradle
```diff
@@ -13,6 +13,8 @@ dependencies {
 
     compile 'io.spray:spray-json_2.11:1.3.3'
 
+    compile 'com.github.pureconfig:pureconfig_2.11:0.7.2'
```
rabbah: Mozilla Public License 2.0. @csantanapr do we need to do anything for Apache on new licenses we're depending on?

core/controller/src/main/resources/application.conf
```diff
@@ -59,26 +66,15 @@ akka.http {
 # Check out all akka-remote-2.5.4 options here:
 # http://doc.akka.io/docs/akka/2.5.4/scala/general/configuration.html#config-akka-remote
 akka {
-  actor {
```
rabbah: how comes these settings are removed?

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -86,11 +90,13 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
   logging: Logging)
     extends LoadBalancer {
 
+  private val lbConfig = loadConfigOrThrow[LoadbalancerConfig]("whisk.loadbalancer")
```
rabbah: will this abort the controller startup on throw?

common/scala/transformEnvironment.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+prefix="CONFIG_"
+configVariables=$(compgen -v | grep $prefix)
+
+props=()
+
+for var in $configVariables
```
rabbah: can you add a comment that documents the contract now between settings in the conf files and their use in the code? maybe not necessarily in this file, but we should write it down somewhere.

core/controller/src/main/resources/application.conf
```diff
@@ -2,6 +2,13 @@
 include "logging"
 include "akka-http-version"
 
+whisk {
+  loadbalancer {
+    invoker-busy-threshold: 4
```
rabbah: apropos my comment above about documentation - how does one override these values where today that's done in ansible vars.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
markusthoemmes: If you don't catch the exception (which we don't), yep.

core/controller/src/main/resources/application.conf
markusthoemmes: You still override them via ENV variables as today. Ansible is just a way of providing those environment variables.

core/controller/src/main/resources/application.conf
markusthoemmes: Because they are overridden via environment variables.

common/scala/transformEnvironment.sh
rabbah: @markusthoemmes bump ^^

common/scala/build.gradle
rabbah: https://mvnrepository.com/artifact/com.github.pureconfig/pureconfig_2.11/0.7.2

csantanapr: It should be OK MPL 2.0 is category 2
https://www.apache.org/legal/resolved.html#category-b
we don't include the source, we just need the binary

markusthoemmes:
@rabbah Docu added. Yes this will need a change downstream.

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -72,27 +72,33 @@
...
+
+      "CONFIG_whisk_loadbalancer_invokerBusyThreshold": "{{ invoker.busyThreshold }}"
+      "CONFIG_whisk_loadbalancer_blackboxFraction": "{{ controller.blackboxFraction }}"
+
+      "CONFIG_akka_actor_provider": "{{ controller.akka.provider }}"
+      "CONFIG_akka_remote_netty_tcp_hostname": "{{ controller.akka.cluster.host[groups['controllers'].index(inventory_hostname)] }}"
+      "CONFIG_akka_remote_netty_tcp_port": "{{ controller.akka.cluster.basePort + groups['controllers'].index(inventory_hostname) }}"
+      "CONFIG_akka_remote_netty_tcp_bindPort": "{{ controller.akka.cluster.bindPort }}"
+
+      "CONFIG_kamon_statsd_hostname": "{{ metrics.kamon.host }}"
```
mhenke1: Please change the `CONFIG_kamon*` environment variables back to 
```
METRICS_KAMON_HOST": "{{ metrics.kamon.host }}"
METRICS_KAMON_PORT": "{{ metrics.kamon.port }}"
```
In the already delivered implementation they are consumed by `common/scala/build/resources/main/application.conf` via these names direct from the environment (not anymore via  JVM parameters).

markusthoemmes: Right. The purpose of those env variables + the transform script is to avoid that double-declaration and pass values through directly.

markusthoemmes:
PG1 2308 🔵

rabbah:
@dgrove-oss this will affect kube deployment.
@markusthoemmes thanks for rebasing.

rabbah:
Rebased again to resolve conflict.

dgrove-oss:
We'll need some kube updates to match the new naming convention for envars, but should be straightforward.  Once this gets merged, I can do the kube pieces.

PR #2577, opened 2017-08-05
Support multiple response header values in raw web actions

bbrowning:
This change allows multiple response header values to be set in raw
web actions by using a JSON array as the header value. For example:

```
function main() {
  return {
    headers: {
      "Set-Cookie": ["a=b", "c=d"]
    },
    code: 200
  }
}
```

I have signed the Apache CLA as "Benjamin M. Browning".

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
@@ -216,11 +216,8 @@ protected[core] object WhiskWebActionsApi extends Directives {
             val JsObject(fields) = result
             val headers = fields.get("headers").map {
                 case JsObject(hs) => hs.map {
-                    case (k, JsString(v))  => RawHeader(k, v)
-                    case (k, JsBoolean(v)) => RawHeader(k, v.toString)
-                    case (k, JsNumber(v))  => RawHeader(k, v.toString)
-                    case _                 => throw new Throwable("Invalid header")
-                }.toList
+                    case (k, v) => headersFromJson(k, v)
+                }.flatten.toList
 
                 case _ => throw new Throwable("Invalid header")
             } getOrElse List()
@@ -249,6 +246,14 @@ protected[core] object WhiskWebActionsApi extends Directives {
             // response as an http result
             terminate(BadRequest, Messages.invalidMedia(`message/http`))(transid)
         }
+    }
+
+    private def headersFromJson(k: String, v: JsValue) : Seq[RawHeader] = v match {
+        case JsString(v)  => Seq(RawHeader(k, v))
+        case JsBoolean(v) => Seq(RawHeader(k, v.toString))
+        case JsNumber(v)  => Seq(RawHeader(k, v.toString))
+        case JsArray(v)   => v.map(headersFromJson(k, _:JsValue)).flatten
+        case _            => throw new Throwable("Invalid header")
     }
 
     private def interpretHttpResponse(code: StatusCode, headers: List[RawHeader], str: String, transid: TransactionId): RequestContext => Unit = {
--- a/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
@@ -215,9 +215,9 @@ protected[core] object WhiskWebActionsApi extends Directives {
         Try {
             val JsObject(fields) = result
             val headers = fields.get("headers").map {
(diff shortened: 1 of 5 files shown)


core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -216,11 +216,8 @@ protected[core] object WhiskWebActionsApi extends Directives {
             val JsObject(fields) = result
             val headers = fields.get("headers").map {
                 case JsObject(hs) => hs.map {
-                    case (k, JsString(v))  => RawHeader(k, v)
-                    case (k, JsBoolean(v)) => RawHeader(k, v.toString)
-                    case (k, JsNumber(v))  => RawHeader(k, v.toString)
-                    case _                 => throw new Throwable("Invalid header")
-                }.toList
+                    case (k, v) => headersFromJson(k, v)
+                }.flatten.toList
```
markusthoemmes: You should be able to use `flatMap` above to omit the extra `flatten`

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -1127,6 +1127,27 @@ trait WebActionsApiTests extends ControllerTestCommon with BeforeAndAfterEach wi
...
+                        JsObject(
+                            "headers" -> JsObject(
+                                "Set-Cookie" -> JsArray(JsString("a=b"), JsString("c=d; Path = /")))))
+
+                    Options(s"$testRoutePath/$path") ~> sealRoute(routes(creds)) ~> check {
+                        withClue(headers) {
+                            headers.length shouldBe 2
+                            headers(0).toString shouldBe "Set-Cookie: a=b"
+                            headers(1).toString shouldBe "Set-Cookie: c=d; Path = /"
+                        }
```
markusthoemmes: Using the extensive matchers of ScalaTest, we can write this in a more speaking way and even preserve great error messages in case a test fails. We also should preserve typesafety here and not rely on the string representation of the headers:

Assuming you want to keep the order stable (which I think is not necessary here) you could write:

```scala
headers shouldBe List(RawHeader("Set-Cookie", "a=b"), RawHeader("Set-Cookie", "c=d; Path = /"))
```

This code would fail with appropriate error messages if the list doesn't comply to the test, and automatically print both lists to be checked thus removing the need for `withClue` and a specific length check.

As I think we shouldn't check for a specific order here, we could use `contains allOf` like so:

```scala
headers should contain allOf (
    RawHeader("Set-Cookie", "a=b"),
    RawHeader("Set-Cookie", "c=d; Path = /")
)
```

Externalizing the cookie values and the headername (maybe even strongly typing them with `scala.http.HttpHeaders.Set-Cookie` and `scala.http.HttpCookie`) is valuable here as well.

markusthoemmes:
Great contribution, thank you :+1:

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -251,6 +248,14 @@ protected[core] object WhiskWebActionsApi extends Directives {
         }
     }
 
+    private def headersFromJson(k: String, v: JsValue) : Seq[RawHeader] = v match {
+        case JsString(v)  => Seq(RawHeader(k, v))
+        case JsBoolean(v) => Seq(RawHeader(k, v.toString))
+        case JsNumber(v)  => Seq(RawHeader(k, v.toString))
+        case JsArray(v)   => v.map(headersFromJson(k, _:JsValue)).flatten
```
markusthoemmes: Same comment on `flatMap`. In general, nesting a `_` can become tricky. I'd prefer fat-arrow notation in this case, the non inferred type seems confusing.

```scala
case JsArray(v)   => v.flatMap(inner => headersFromJson(k, inner))
```

bbrowning:
Thanks for the pointers! I pushed another commit addressing the issues raised so far. Normally I'd squash/rebase and do a force push, but wasn't sure how the project preferred to handle that during code reviews and before merging.

rabbah:
Thanks for the contribution and for not squashing the revision! Indeed it helps with reviews not to constantly squash.

Usually the committers can rebase, squash and update the commit message as the PR is merged.

markusthoemmes:
PG1 1908 🔵

markusthoemmes:
@bbrowning congratulations to your first contribution and welcome to the OpenWhisk gang 🎉

PR #2578, opened 2017-08-07
Update dependencies and gitignore

cbickel:
This PR updates the zookeeper library.

In addition, it updates the .gitignore to add own environments without poluting local git.


cbickel:
PG4#641 is 🔵

PR #2582, opened 2017-08-07
Use non-blocking techniques when interfacing to Kafka.

markusthoemmes:
Producing a message can be done with a callback upon receiving the message, effectively acting as a non-blocking API.

Consuming a message is a bit trickier. Adding `blocking` instructs the default dispatcher to create new threads if necessary.

### DIFF
--- a/common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
+++ b/common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
@@ -22,9 +22,11 @@ import java.util.UUID
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+import scala.concurrent.Promise
 import scala.util.Failure
 import scala.util.Success
 
+import org.apache.kafka.clients.producer.Callback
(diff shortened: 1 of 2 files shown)


common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -31,6 +31,10 @@ import org.apache.kafka.common.serialization.ByteArrayDeserializer
 
 import whisk.common.Logging
 import whisk.core.connector.MessageConsumer
+import org.apache.kafka.clients.consumer.OffsetCommitCallback
+import org.apache.kafka.clients.consumer.OffsetAndMetadata
+import scala.concurrent.Promise
+import org.apache.kafka.common.TopicPartition
```
rabbah: small nit can we sort these imports

dubeejw:
@markusthoemmes, when I run this locally to invoke an action, I get the same error that is in your Travis log. Also, I see this in Kibana logs ` action activation failed: no invokers available`.

markusthoemmes:
PG1 1920 🔵

PR #2583, opened 2017-08-07
Create Nginx image with localy built CLI, Blackbox SDK.

DanLavine:
* Locally builds and adds all CLI OS + arch builds to Nginx
* builds and adds Blackbox SDK
* Update Ansible scripts to deploy the new nginx image

The main motivation for this feature is so deployment options that don't want to rely on Ansible's post docker creation steps can have an Nginx image with all the extra CLIs and SDKs built into it.

**NOTE** This PR will require a new Nginx image for OpenWhisk. I'm not sure where this would be built and published from.

**NOTE** This PR removes the ability to install the CLI from the remote github repo. However, when the CLI is split out of the main OpenWhisk repo it should be easy to pull the remote CLI. All that would need to change is not building the CLI locally and instead, pull the CLI binary/zip files from github into the Nginx image.


csantanapr:
@markusthoemmes can you take a look at this, what do you think on this approach?

markusthoemmes:
In general, I'm in favor of this approach. It looks sensible to me. I cannot overlook all the sideeffects of how the CLI is built and distributed at this point though.

@houshengbo could you help out here?

core/nginx/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM nginx:1.11
+
+# make cli directories
+RUN mkdir -p /etc/nginx/cli/go/download/linux/386 && \
+    mkdir -p /etc/nginx/cli/go/download/linux/amd64 && \
+    mkdir -p /etc/nginx/cli/go/download/mac/386 && \
+    mkdir -p /etc/nginx/cli/go/download/mac/amd64 && \
+    mkdir -p /etc/nginx/cli/go/download/windows/386 && \
+    mkdir -p /etc/nginx/cli/go/download/windows/amd64
```
markusthoemmes: Is this needed?

core/nginx/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+COPY build/windows/386/wsk.exe /etc/nginx/cli/go/download/windows/386/wsk.exe
+COPY build/windows/amd64/wsk.exe /etc/nginx/cli/go/download/windows/amd64/wsk.exe
+
+# copy all cli zip files
+COPY build/linux/386/OpenWhisk_CLI-linux-32bit.tgz /etc/nginx/cli/go/download/linux/386/OpenWhisk_CLI-linux-32bit.tgz
+COPY build/linux/amd64/OpenWhisk_CLI-linux.tgz /etc/nginx/cli/go/download/linux/amd64/OpenWhisk_CLI-linux.tgz
+COPY build/mac/386/OpenWhisk_CLI-mac-32bit.zip /etc/nginx/cli/go/download/mac/386/OpenWhisk_CLI-mac-32bit.zip
+COPY build/mac/amd64/OpenWhisk_CLI-mac.zip /etc/nginx/cli/go/download/mac/amd64/OpenWhisk_CLI-mac.zip
+COPY build/windows/386/OpenWhisk_CLI-windows-32bit.zip /etc/nginx/cli/go/download/windows/386/OpenWhisk_CLI-windows-32bit.zip
+COPY build/windows/amd64/OpenWhisk_CLI-windows.zip /etc/nginx/cli/go/download/windows/amd64/OpenWhisk_CLI-windows.zip
```
markusthoemmes: Couldn't this all be done with one single `COPY`/`ADD`? Same for above cli binaries?

core/nginx/build.gradle
```diff
@@ -0,0 +1,72 @@
...
+    copy {
+      from "${projectDir}/../../bin/windows/386/OpenWhisk_CLI-windows-32bit.zip"
+      into "${projectDir}/build/windows/386/"
+    }
+    copy {
+      from "${projectDir}/../../bin/windows/amd64/OpenWhisk_CLI-windows.zip"
+      into "${projectDir}/build/windows/amd64/"
+    }
+  }
+}
```
markusthoemmes: We should have a look if this is worth being replaced by two `Exec` tasks just doing a `cp` or using a `Copy` task directly. Seems quite error prone.

DanLavine:
Fixed up those issues. I thought there had been an issue when copying files into Docker when the whole path didn't exist. Maybe that was an older docker version though. But I tested the new commits on both Ubuntu and Mac and they were able to build fine.

rabbah:
The SDK is now built externally and the edge router just points to a release. The same could be done for the CLI. This would simplify this PR significantly. Hence I am inclined to close this PR in its current state.

csantanapr:
Yep that's the idea once the CLI is out of the core repo to its own repo then edge would just point to the built binaries on the github release page

dgrove-oss:
Agreed.  Let's close this one and redo with the simpler setup enabled by pulling the artifacts from the github releases of the other repos.

PR #2587, opened 2017-08-08
Make LoadbalancerData leaner.

markusthoemmes:
LoadbalancerData keeps a lot of data it doesn't need. Upon fetching a copy of this data, it creates copies which can be expensive, the more user are in the system.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerData.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerData.scala
@@ -17,9 +17,12 @@
 
 package whisk.core.loadBalancer
 
-import whisk.core.entity.{ ActivationId, UUID, WhiskActivation }
+import java.util.concurrent.atomic.AtomicInteger
+
 import scala.collection.concurrent.TrieMap
 import scala.concurrent.Promise
+
+import whisk.core.entity.{ ActivationId, UUID, WhiskActivation }
 import whisk.core.entity.InstanceId
 
 /** Encapsulates data relevant for a single activation */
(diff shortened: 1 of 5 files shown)


core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerData.scala
```diff
@@ -33,28 +36,32 @@ case class ActivationEntry(id: ActivationId, namespaceId: UUID, invokerName: Ins
...
-     * Get the number of activations for each invoker.
+     * Get the number of activations for a specific invoker.
      *
+     * @param invoker The invoker to get the activation count for
      * @return a map (invoker -> number of activations queued for the invoker)
      */
-    def activationCountByInvoker: Map[InstanceId, Int] = {
-        activationByInvoker.toMap.mapValues(_.size)
+    def activationCountOn(invoker: InstanceId): Int = {
+        activationByInvoker.get(invoker).map(_.get).getOrElse(0)
```
rabbah: should this count be 0 for an invoker that doesn't exist?
ie should this return Option[Int] instead?

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -56,12 +56,11 @@ trait LoadBalancer {
 
     val activeAckTimeoutGrace = 1.minute
 
-    /**
-     * Retrieves a per namespace id map of counts representing in-flight activations as seen by the load balancer
-     *
-     * @return a map where the key is the namespace id and the long is total issued activations by that namespace
-     */
-    def getActiveNamespaceActivationCounts: Map[UUID, Int]
+    /** Gets the number of in-flight activations for a specific user */
```
rabbah: missing period at end of comment.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerData.scala
rabbah: should this count be 0 for an invoker that doesn't exist?
ie should this return Option[Int] instead?

rabbah: seeing usage below - i suppose 0 is fine.

rabbah:
no pg?

rabbah:
pg2/1947 :kick_scooter:

rabbah:
pg2/1949  🔵

Issue #2430, opened 2017-06-27
Reimplementation of Wsk.scala for test cases


houshengbo:
In order to remove the dependency of openwhisk cli binary, openwhisk needs to re-implement Wsk.scala for all the test cases. Instead of calling binary to access the openwhisk API, we need to call the rest api.

This issue is a part of work item, which departures openwhisk cli into separate repo.

Currently we only have Wsk as for CLI implementation, and Wsk entities based on issuing CLI commands, in order to access openwhisk services for tests. We are targeting to implement WskRest for REST, and Wsk REST entities based on making REST calls to access openwhisk services for tests as well.

This piece of work can be designed to implement in the following steps:

1. Make the final result for all the test cases consistent.
    -- We use RunResult to save the information from CLI, in terms of exitCode, stdout and stderr. We need to have RestResult to save the information coming back from http response. Http status code can be converted into exitCode (by minus 256 if it is greater than 400, otherwise keep the same), http resposne data can be converted into stdout, http error message can be mapped into stderr if there is error.

-- Result of activation CliActivation need to be generalized. We need to make it accessible by class in packages, not just an inner class. In addition, we need to give a general name, like ActivationResult. Activation information from http should also be converted into this object. PR: #2592

2. Create base Wsk class and base entities for Wsk CLI and REST to extend or create. All the test cases are now using Wsk and Wsk entities of CLI by default. We need to switch to use Base Wsk and Base entities, so that test cases can call the implementation of either Wsk CLI or REST based on different needs. Wsk class and base entities are defined as traits with the methods. Wsk and Wsk entities of CLI and REST need to override the methods with different implementation. PR: #2593

3.When the base Wsk class and base entities are in hand, implement Wsk REST and wsk rest entities.
    -- Create WskRest framework
    -- Add WskRestAction
    -- Add WskRestTrigger
    -- Add WskRestNamespace
    -- Add WskRestRule
    -- Add WskRestPackage
    -- Add WskRestApi
    -- Add WskRestActivation
    PR: #2589

4. Create abstract class for test cases, and create different wsk with either cli or rest. Test result verification may remain the same for most of cases, may need slight changes for some verification.
    PR: #2621

rabbah:
please make this feature toggle-able so that one can choose to run the test with a cli or an api.
in other words, don't repeal and replace.

csantanapr:
[...]

PR #2589, opened 2017-08-08
Add the fundamental framework of REST invocation for test cases

houshengbo:
Almost all the test cases in OpenWhisk are currently running based
on the wsk CLI binary. In order to separate the CLI out of OpenWhisk
core repository, we need to call REST API of OpenWhisk to access
OpenWhisk services instead of calling the wsk CLI binary command.

This PR adds the basic REST implementation for all the test cases
to use. One basic principle is to keep the changes to the existing
test cases as few as possible, so we add WskRest.scala as an equivalent
class to Wsk.scala and WskRestTestHelpers.scala as an equivalent file
to WskTestHelpers.scala for REST.

All the replacement of binary with REST will happen in an increamental fashion.
This PR only changes one existing test case in WskActionTests.scala, and
reimplements it in WskRestActionTests.scala. All the other test cases can follow
the same way to change the test cases one by one. New changes may be necessary
to the basic REST implementation as well in future to accommodate the test cases.

Partially-closes-bug: #2430

### DIFF
--- a/tests/src/test/scala/common/rest/WskRest.scala
+++ b/tests/src/test/scala/common/rest/WskRest.scala
@@ -1,0 +1,1458 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package common.rest
+
+import java.io.File
+import java.time.Clock
+import java.time.Instant
+import java.util.Base64
+
+import org.apache.commons.io.FileUtils
+import org.scalatest.Matchers
+import org.scalatest.FlatSpec
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.time.Span.convertDurationToSpan
+
+import scala.Left
+import scala.Right
+import scala.collection.JavaConversions.mapAsJavaMap
+import scala.collection.mutable.Buffer
+import scala.collection.immutable.Seq
+import scala.concurrent.duration.Duration
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.Future
+import scala.language.postfixOps
+import scala.util.Failure
+import scala.util.Success
+import scala.util.Try
+import scala.util.{ Failure, Success }
+
+import akka.http.scaladsl.model.StatusCode
+import akka.http.scaladsl.model.StatusCodes.Accepted
+import akka.http.scaladsl.model.StatusCodes.NotFound
+import akka.http.scaladsl.model.StatusCodes.BadRequest
+import akka.http.scaladsl.model.StatusCodes.OK
+import akka.http.scaladsl.model.HttpRequest
+import akka.http.scaladsl.model.HttpMethod
+import akka.http.scaladsl.model.HttpResponse
+import akka.http.scaladsl.model.headers.Authorization
+import akka.http.scaladsl.model.HttpEntity
+import akka.http.scaladsl.model.ContentTypes
+import akka.http.scaladsl.Http
+
+import akka.http.scaladsl.model.headers.BasicHttpCredentials
+import akka.http.scaladsl.model.Uri
+import akka.http.scaladsl.model.Uri.Path
+
+import akka.http.scaladsl.model.HttpMethods.DELETE
+import akka.http.scaladsl.model.HttpMethods.GET
+import akka.http.scaladsl.model.HttpMethods.POST
+import akka.http.scaladsl.model.HttpMethods.PUT
+
+import akka.stream.ActorMaterializer
+
+import spray.json._
+import spray.json.DefaultJsonProtocol._
+import spray.json.JsObject
+import spray.json.JsValue
+import spray.json.pimpString
+
+import common._
+import common.BaseDeleteFromCollection
+import common.BaseListOrGetFromCollection
+import common.HasActivation
+import common.RunWskCmd
+import common.TestUtils
+import common.TestUtils.SUCCESS_EXIT
+import common.TestUtils.DONTCARE_EXIT
+import common.TestUtils.ANY_ERROR_EXIT
+import common.TestUtils.DONTCARE_EXIT
+import common.TestUtils.RunResult
+import common.WaitFor
+import common.WhiskProperties
+import common.WskActorSystem
+import common.WskProps
+
+import whisk.core.entity.ByteSize
+import whisk.utils.retry
+
+class WskRest() extends RunWskRestCmd with BaseWsk {
+    override implicit val action = new WskRestAction
+    override implicit val trigger = new WskRestTrigger
+    override implicit val rule = new WskRestRule
+    override implicit val activation = new WskRestActivation
+    override implicit val pkg = new WskRestPackage
+    override implicit val namespace = new WskRestNamespace
+    override implicit val api = new WskRestApi
+}
+
+trait ListOrGetFromCollectionRest extends BaseListOrGetFromCollection {
+    self: RunWskRestCmd =>
+
+    /**
+     * List entities in collection.
+     *
+     * @param namespace (optional) if specified must be  fully qualified namespace
+     * @param expectedExitCode (optional) the expected exit code for the command
+     * if the code is anything but DONTCARE_EXIT, assert the code is as expected
+     */
+    override def list(namespace: Option[String] = None,
+                      limit: Option[Int] = None,
+                      nameSort: Option[Boolean] = None,
+                      expectedExitCode: Int = OK.intValue)(implicit wp: WskProps): RestResult = {
+        val (ns, name) = getNamespaceActionName(resolve(namespace))
+
+        val pathToList = s"$basePath/namespaces/$ns/$noun"
+        val entPath =
+            if (name != "") Path(s"$pathToList/$name/")
+            else Path(s"$pathToList")
+        val paramMap = Map[String, String]() ++ { Map("skip" -> "0".toString, "docs" -> true.toString) } ++ {
+            limit map { l =>
+                Map("limit" -> l.toString)
+            } getOrElse Map[String, String]("limit" -> "30".toString)
+        }
+        val resp = getWhiskEntity(entPath, paramMap).futureValue
+        val r = new RestResult(resp.status, getRespData(resp))
+        validateStatusCode(expectedExitCode, r.statusCode.intValue)
+        r
+    }
+
+    /**
+     * Gets entity from collection.
+     *
+     * @param name either a fully qualified name or a simple entity name
+     * @param expectedExitCode (optional) the expected exit code for the command
+     * if the code is anything but DONTCARE_EXIT, assert the code is as expected
+     */
+    override def get(name: String,
+                     expectedExitCode: Int = OK.intValue,
+                     summary: Boolean = false,
+                     fieldFilter: Option[String] = None,
+                     url: Option[Boolean] = None)(implicit wp: WskProps): RestResult = {
+        val (ns, entity) = getNamespaceActionName(name)
+        val entPath = Path(s"$basePath/namespaces/$ns/$noun/$entity")
+        val resp = getWhiskEntity(entPath)(wp).futureValue
+        val r = new RestResult(resp.status, getRespData(resp))
+        validateStatusCode(expectedExitCode, r.statusCode.intValue)
+        r
+    }
+}
+
+trait DeleteFromCollectionRest extends BaseDeleteFromCollection {
+    self: RunWskRestCmd =>
+
+    /**
+     * Deletes entity from collection.
+     *
+     * @param name either a fully qualified name or a simple entity name
+     * @param expectedExitCode (optional) the expected exit code for the command
+     * if the code is anything but DONTCARE_EXIT, assert the code is as expected
+     */
+    override def delete(entity: String, expectedExitCode: Int = OK.intValue)(implicit wp: WskProps): RestResult = {
+        val r = deleteEntity(entity)(wp)
+        validateStatusCode(expectedExitCode, r.statusCode.intValue)
+        r
+    }
+
+    def deleteEntity(name: String)(implicit wp: WskProps): RestResult = {
+        val (ns, entity) = getNamespaceActionName(name)
+        val path = Path(s"$basePath/namespaces/$ns/$noun/$entity")
+        val resp = deleteEntity(path)(wp)
+        new RestResult(resp.status, getRespData(resp))
+    }
+
+    /**
+     * Deletes entity from collection but does not assert that the command succeeds.
+     * Use this if deleting an entity that may not exist and it is OK if it does not.
+     *
+     * @param name either a fully qualified name or a simple entity name
+     */
+    override def sanitize(name: String)(implicit wp: WskProps): RestResult = {
+        delete(name, DONTCARE_EXIT)
+    }
+}
+
+trait HasActivationRest extends HasActivation {
+
+    /**
+     * Extracts activation id from invoke (action or trigger) or activation get
+     */
+    override def extractActivationId(result: RunResult): Option[String] = {
+        extractActivationIdFromInvoke(result.asInstanceOf[RestResult])
+    }
+
+    /**
+     * Extracts activation id from 'wsk action invoke' or 'wsk trigger invoke'
+     */
+    private def extractActivationIdFromInvoke(result: RestResult): Option[String] = {
+        Try {
+            val activationID =
+                if ((result.statusCode == OK) || (result.statusCode == Accepted)) result.getField("activationId") else ""
+            activationID
+        } toOption
+    }
+}
+
+class WskRestAction
+    extends RunWskRestCmd
+    with ListOrGetFromCollectionRest
+    with DeleteFromCollectionRest
+    with HasActivationRest
+    with BaseAction {
+
+    override protected val noun = "actions"
+    override implicit val config = PatienceConfig(100 seconds, 15 milliseconds)
+
+    /**
+     * Creates action. Parameters mirror those available in the REST.
+     *
+     * @param name either a fully qualified name or a simple entity name
+     * @param expectedExitCode (optional) the expected exit code for the command
+     * if the code is anything but DONTCARE_EXIT, assert the code is as expected
+     */
+    override def create(
+        name: String,
+        artifact: Option[String],
+        kind: Option[String] = None, // one of docker, copy, sequence or none for autoselect else an explicit type
+        main: Option[String] = None,
+        docker: Option[String] = None,
+        parameters: Map[String, JsValue] = Map(),
+        annotations: Map[String, JsValue] = Map(),
+        parameterFile: Option[String] = None,
+        annotationFile: Option[String] = None,
+        timeout: Option[Duration] = None,
+        memory: Option[ByteSize] = None,
+        logsize: Option[ByteSize] = None,
+        shared: Option[Boolean] = None,
+        update: Boolean = false,
+        web: Option[String] = None,
+        expectedExitCode: Int = OK.intValue)(implicit wp: WskProps): RestResult = {
+
+        val (namespace, actName) = getNamespaceActionName(name)
+        var exec = Map[String, JsValue]()
+        var code = ""
+        var kindType = ""
+
+        val (ext, artifactFile) = artifact map { a =>
+            (getExt(a), a)
+        } getOrElse (null, "")
+
+        ext match {
+            case ".jar" => {
+                kindType = "java:default"
+                val jar = FileUtils.readFileToByteArray(new File(artifactFile))
+                code = Base64.getEncoder.encodeToString(jar)
+            }
+            case ".zip" => {
+                val zip = FileUtils.readFileToByteArray(new File(artifactFile))
+                code = Base64.getEncoder.encodeToString(zip)
+            }
+            case ".js" => {
+                kindType = "nodejs:default"
+                code = FileUtils.readFileToString(new File(artifactFile))
+            }
+            case ".py" => {
+                kindType = "python:default"
+                code = FileUtils.readFileToString(new File(artifactFile))
+            }
+            case ".swift" => {
+                kindType = "swift:default"
+                code = FileUtils.readFileToString(new File(artifactFile))
+            }
+            case ".php" => {
+                kindType = "php:default"
+                code = FileUtils.readFileToString(new File(artifactFile))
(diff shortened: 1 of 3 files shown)


markusthoemmes:
Close #2480 ?

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1482 @@
...
+   *
+   * @param namespace (optional) if specified must be  fully qualified namespace
+   * @param expectedExitCode (optional) the expected exit code for the command
+   * if the code is anything but DONTCARE_EXIT, assert the code is as expected
+   */
+  override def list(namespace: Option[String] = None,
+                    limit: Option[Int] = None,
+                    nameSort: Option[Boolean] = None,
+                    expectedExitCode: Int = OK.intValue)(implicit wp: WskProps): RestResult = {
+    val NS = namespace map { ns =>
```
dubeejw: Why not use `resolve()` from BaseWsk here like Wsk.scala does?

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1482 @@
...
+  override def get(name: String,
+                   expectedExitCode: Int = OK.intValue,
+                   summary: Boolean = false,
+                   fieldFilter: Option[String] = None,
+                   url: Option[Boolean] = None)(implicit wp: WskProps): RestResult = {
+    val (ns, entity) = getNamespaceActionName(name)
+    val entPath = Path(s"$basePath/namespaces/$ns/$noun/$entity")
+    val resp = getWhiskEntity(entPath)(wp).futureValue
+    val r = new RestResult(resp.status, getRespData(resp))
+    if (expectedExitCode != DONTCARE_EXIT)
```
dubeejw: Looks like the check for expectedExitCode is repeated many times. Should construct a function that does the check instead.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1482 @@
...
+  }
+
+  /**
+   * Extracts activation id from 'wsk action invoke' or 'wsk trigger invoke'
+   */
+  private def extractActivationIdFromInvoke(result: RestResult): Option[String] = {
+    Try {
+      var activationID =
+        if ((result.statusCode == OK) || (result.statusCode == Accepted)) result.getField("activationId") else ""
+      activationID
```
dubeejw: This line isn't needed activationID should already be the last thing returned when it is defined.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1482 @@
...
+   * if the code is anything but DONTCARE_EXIT, assert the code is as expected
+   */
+  override def create(name: String,
+                      trigger: String,
+                      action: String,
+                      annotations: Map[String, JsValue] = Map(),
+                      shared: Option[Boolean] = None,
+                      update: Boolean = false,
+                      expectedExitCode: Int = SUCCESS_EXIT)(implicit wp: WskProps): RestResult = {
+    val path = Path(s"$basePath/namespaces/${wp.namespace}/$noun/$name")
```
dubeejw: Seems like `$basePath/namespaces/${wp.namespace}/$noun` is repeated several times and could be shared amongst functions.

tests/src/test/scala/common/rest/WskRest.scala
dubeejw: Also, why use capitals for NS? See https://docs.scala-lang.org/style/naming-conventions.html for naming conventions.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1482 @@
...
+    }
+  }
+
+  override def get(activationId: Option[String],
+                   expectedExitCode: Int = OK.intValue,
+                   fieldFilter: Option[String] = None,
+                   last: Option[Boolean] = None)(implicit wp: WskProps): RestResult = {
+    val path = activationId map { id =>
+      Path(s"${basePath}/namespaces/${wp.namespace}/$noun/$id")
+    } getOrElse Path("")
```
dubeejw: What happens if path is an empty string?

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1482 @@
...
+                      annotations: Map[String, JsValue] = Map(),
+                      parameterFile: Option[String] = None,
+                      annotationFile: Option[String] = None,
+                      shared: Option[Boolean] = None,
+                      update: Boolean = false,
+                      expectedExitCode: Int = OK.intValue)(implicit wp: WskProps): RestResult = {
+    val path = Path(s"$basePath/namespaces/${wp.namespace}/$noun/$name")
+    var bodyContent = JsObject("namespace" -> s"${wp.namespace}".toJson, "name" -> name.toJson)
+
+    val (params, annos) = this.getParamsAnnos(parameters, annotations, parameterFile, annotationFile)
```
dubeejw: Hmm, I don't understand why the parameter and annotations handling differs here base on action update/create.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1482 @@
...
+                      operation: Option[String] = None,
+                      action: Option[String] = None,
+                      apiname: Option[String] = None,
+                      swagger: Option[String] = None,
+                      responsetype: Option[String] = None,
+                      expectedExitCode: Int = SUCCESS_EXIT,
+                      cliCfgFile: Option[String] = None)(implicit wp: WskProps): RestResult = {
+    val (ns, actionName) = this.getNamespaceActionName(action map { a =>
+      a
+    } getOrElse "")
```
dubeejw: What happens if action name is an empty string?

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1482 @@
...
+    if (filePath.contains(sep)) filePath.substring(filePath.lastIndexOf(sep), filePath.length())
+    else null
+  }
+
+  def request(method: HttpMethod, uri: Uri, body: String = "", creds: BasicHttpCredentials): Future[HttpResponse] = {
+    val entity = HttpEntity(ContentTypes.`application/json`, body)
+    val request = HttpRequest(method, uri, List(Authorization(creds)), entity = entity)
+    Http().singleRequest(request)
+  }
+  def createEntity(path: Path, body: String, params: Map[String, String] = null, whiskUrl: Uri = whiskRestUrl)(
```
dubeejw: createEntity, postEntityParam, postEntity, deleteEntity, getWhiskEntity seems like they call all be rolled into one function that takes an HTTP verb like PUT, DELETE, GET, POST.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: I will get rid of val NS.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: It has been put in a validate method.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: If I remove activationID, there is going to be an syntax error.
Well, I can change activationID from var to val.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: Done.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: We should return a not found error.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: We will return notfound error in the response.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: We pick up either parameters or parameterFile as the parameters to create the action, and pick up either annotations or annotationFile as the annotations to create the action. 
When it comes to update the action, we only care about the parameters and annotations, not parameterFile or annotationFile any more. That is why we see the differences.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: done.

houshengbo:
PG 2158

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1437 @@
...
+    else null
+  }
+
+  def request(method: HttpMethod, uri: Uri, body: String = "", creds: BasicHttpCredentials): Future[HttpResponse] = {
+    val entity = HttpEntity(ContentTypes.`application/json`, body)
+    val request = HttpRequest(method, uri, List(Authorization(creds)), entity = entity)
+    Http().singleRequest(request)
+  }
+
+  def requestEntity(method: HttpMethod,
```
dubeejw: Nice. I think the goal is to avoid using null if possible. Can you use `Some` and `None` instead?

dubeejw:
@rabbah, @csantanapr, I think it might be possible to use the work here to provide a Scala OpenWhisk client at some point in the future.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1437 @@
...
+      }
+    } else {
+      ns = wp.namespace
+      entityName = name
+    }
+
+    (ns, entityName)
+  }
+
+  def invokeAction(name: String,
```
dubeejw: I think this function can be named more appropriately as it looks like `invokeAction` will handle all CRUD requests for actions.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1437 @@
...
+      case _ =>
+    }
+
+    kindType = docker map { d =>
+      "blackbox"
+    } getOrElse kindType
+
+    var (params, annos) = getParamsAnnos(parameters, annotations, parameterFile, annotationFile, web = web)
+
+    val inputKindType = kind map { k =>
```
dubeejw: Shouldn't need this as kind is defined as a `Some` or `None`.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1437 @@
...
+      k
+    } getOrElse null
+
+    inputKindType match {
```
dubeejw: Match on `Some` or `None` here for `kind`. Ex:
```
kind match {
	case Some(k) if k == "copy" => copy routine
	case Some(k) if k == "sequence" => sequence routine
	case _ => kind is None
}
```

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1437 @@
...
+    val resp =
+      if (update) requestEntity(PUT, path, Map("overwrite" -> "true"), bodyContent.toString)
+      else requestEntity(PUT, path, body = bodyContent.toString)
+    val result = new RestResult(resp.status, getRespData(resp))
+
+    val feedInput = feed map { f =>
+      f
+    } getOrElse null
+
+    if ((feed == null) || (result.statusCode != OK)) {
```
dubeejw: How would `feed` be null here if it is a `Some` or `None`?

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1437 @@
...
+        bodyContent = JsObject(bodyContent.fields + ("annotations" -> annos))
+      }
+    }
+
+    val resp =
+      if (update) requestEntity(PUT, path, Map("overwrite" -> "true"), bodyContent.toString)
+      else requestEntity(PUT, path, body = bodyContent.toString)
+    val result = new RestResult(resp.status, getRespData(resp))
+
+    val feedInput = feed map { f =>
```
dubeejw: Should need this, just match on feed like example I left above.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1437 @@
...
+   * @param expectedExitCode (optional) the expected exit code for the command
+   * if the code is anything but DONTCARE_EXIT, assert the code is as expected
+   */
+  def listActivation(filter: Option[String] = None,
+                     limit: Option[Int] = None,
+                     since: Option[Instant] = None,
+                     docs: Boolean = true,
+                     expectedExitCode: Int = SUCCESS_EXIT)(implicit wp: WskProps): RestResult = {
+    val entityPath = Path(s"${basePath}/namespaces/${wp.namespace}/$noun")
+    var paramMap = Map("skip" -> "0".toString, "docs" -> docs.toString) ++ {
```
dubeejw: `"0"` is already a string. Shouldn't need to use `toString` there.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: Done

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: invokeaction only invokes actions by POST the existing action.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: Done.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: It should be feedInput, but I will change part anyway.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: Done

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: Done

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: Done

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1430 @@
...
+   * @param namespace (optional) if specified must be  fully qualified namespace
+   * @param expectedExitCode (optional) the expected exit code for the command
+   * if the code is anything but DONTCARE_EXIT, assert the code is as expected
+   */
+  override def list(namespace: Option[String] = None,
+                    limit: Option[Int] = None,
+                    nameSort: Option[Boolean] = None,
+                    expectedExitCode: Int = OK.intValue)(implicit wp: WskProps): RestResult = {
+    val (ns, name) = getNamespaceActionName(resolve(namespace))
+
```
dubeejw: Newline here needed?

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1430 @@
...
+      activation.respBody
+    } map {
+      Right(_)
+    } getOrElse Left(s"Cannot find activation id from '$activation'")
+
+  }
+
+  def waitForActivationConsole(totalWait: Duration = 30 seconds, sinceTime: Instant)(
+    implicit wp: WskProps): RestResult = {
+    var result = new RestResult(NotFound, null)
```
dubeejw: Possible to get rid of `null` on these `new RestResult` calls?

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: removed.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: done.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1434 @@
...
+  override def create(name: String,
+                      trigger: String,
+                      action: String,
+                      annotations: Map[String, JsValue] = Map(),
+                      shared: Option[Boolean] = None,
+                      update: Boolean = false,
+                      expectedExitCode: Int = SUCCESS_EXIT)(implicit wp: WskProps): RestResult = {
+    val path = getNamePath(noun, name)
+    val annos = convertMapIntoKeyValue(annotations)
+    val published = shared map { s =>
```
dubeejw: Here you can do `shared.getOrElse(false)`

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1434 @@
...
+                      update: Boolean = false,
+                      expectedExitCode: Int = OK.intValue)(implicit wp: WskProps): RestResult = {
+
+    val (ns, triggerName) = this.getNamespaceActionName(name)
+    val path = Path(s"$basePath/namespaces/$ns/$noun/$triggerName")
+    val (params, annos) = getParamsAnnos(parameters, annotations, parameterFile, annotationFile, feed)
+    var bodyContent = JsObject("name" -> name.toJson, "namespace" -> s"$ns".toJson)
+
+    if (!update) {
+      val published = shared map { s =>
```
dubeejw: Here you can do `shared.getOrElse(false)`.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1434 @@
...
+                      annotationFile: Option[String] = None,
+                      shared: Option[Boolean] = None,
+                      update: Boolean = false,
+                      expectedExitCode: Int = OK.intValue)(implicit wp: WskProps): RestResult = {
+    val path = getNamePath(noun, name)
+    var bodyContent = JsObject("namespace" -> s"${wp.namespace}".toJson, "name" -> name.toJson)
+
+    val (params, annos) = this.getParamsAnnos(parameters, annotations, parameterFile, annotationFile)
+    if (!update) {
+      val published = shared map { s =>
```
dubeejw: Here you can do `shared.getOrElse(false)`.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1434 @@
...
+    extractActivationIdFromInvoke(result.asInstanceOf[RestResult])
+  }
+
+  /**
+   * Extracts activation id from 'wsk action invoke' or 'wsk trigger invoke'
+   */
+  private def extractActivationIdFromInvoke(result: RestResult): Option[String] = {
+    Try {
+      val activationID =
+        if ((result.statusCode == OK) || (result.statusCode == Accepted)) result.getField("activationId") else ""
```
dubeejw: Here it looks like the an empty string returns `Some()`. I think `None` should be returned instead.

dubeejw: Should only need this in the function: 
```
Try {
     if ((result.statusCode == OK) || (result.statusCode == Accepted)) Some(result.getField("activationId")) else None
}
```

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: done

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: done

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: done

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: We should remove try {}, or it will report error.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1426 @@
...
+  def getRespData(resp: HttpResponse): String = {
+    val timeout = 5.seconds
+    Try {
+      resp.entity.toStrict(timeout).map { _.data }.map(_.utf8String).futureValue
+    } getOrElse {
+      ""
+    }
+  }
+
+  def getNamespaceActionName(name: String)(implicit wp: WskProps): (String, String) = {
```
dubeejw: Is this equivalent?
```
def getNamespaceEntityName(quailfiedName: String)(implicit wp: WskProps): (Option[String], Option[String]) = {
	quailfiedName.split("/") match {
		case Array(empty, namespace, entityName) => (Some(namespace), Some(entityName))
		case Array(empty, namespace) => (Some(namespace), None)
		case Array(entityName) => (Some(wp.namespace), Some(entityName))
		case _ => (None, None)
	}
}
```

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1426 @@
...
+    val result = new RestResult(resp.status, getRespData(resp))
+    if (result.statusCode != OK) {
+      validateStatusCode(expectedExitCode, result.statusCode.intValue)
+      result
+    }
+    val r = feed map { f =>
+      // Invoke the feed
+      val (nsFeed, feedName) = this.getNamespaceActionName(f)
+      val path = Path(s"$basePath/namespaces/$nsFeed/actions/$feedName")
+      val paramMap = Map("blocking" -> "true".toString, "result" -> "false".toString)
```
dubeejw: Both `toString` calls are not needed here.

tests/src/test/scala/common/rest/WskRest.scala
dubeejw: Maybe throw an error on `_` case. What do you think?

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1426 @@
...
+   */
+  override def bind(provider: String,
+                    name: String,
+                    parameters: Map[String, JsValue] = Map(),
+                    annotations: Map[String, JsValue] = Map(),
+                    expectedExitCode: Int = OK.intValue)(implicit wp: WskProps): RestResult = {
+    val params = convertMapIntoKeyValue(parameters)
+    val annos = convertMapIntoKeyValue(annotations)
+
+    val (ns, packageName) = this.getNamespaceActionName(provider)
```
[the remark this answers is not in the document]
houshengbo: I renamed the function into getNamespaceEntityName, cos it is more appropriate.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1426 @@
...
+                      parameters: Map[String, JsValue] = Map(),
+                      annotations: Map[String, JsValue] = Map(),
+                      parameterFile: Option[String] = None,
+                      annotationFile: Option[String] = None,
+                      feed: Option[String] = None,
+                      shared: Option[Boolean] = None,
+                      update: Boolean = false,
+                      expectedExitCode: Int = OK.intValue)(implicit wp: WskProps): RestResult = {
+
+    val (ns, triggerName) = this.getNamespaceActionName(name)
```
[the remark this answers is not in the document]
houshengbo: I renamed the function into getNamespaceEntityName, cos it is more appropriate.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: Let me see how to rewrite it. 
There is no need to fo case _ => (None, None), because the worst case is to return wp.namespace and name.

Issue #2568, opened 2017-08-03
Set a User-Agent header in wsk cli requests


mdeuser:
The `wsk` CLI currently does not set a User-Agent header when sending requests to the OpenWhisk backend.  Setting this can help distinguish between various clients' access to the backend.

drcariel:
Do we have a preferred value for the User-Agent Header? Keep it simple?

```User-Agent: CLI```

or more complicated?

```User-Agent: <product> / <product-version> <comment>```

mdeuser:
i like the latter.  we'll want the cli version included to distinguish between various builds..

drcariel:
Got some code @ https://github.com/apache/incubator-openwhisk/pull/2591 if you wanna check it out

drcariel:
os.Setenv/os.Getenv was the easiest way for me to get access to the build version since it was intentionally excluded from the available Config

rabbah:
@drcariel this is complete no? closing as I believe so, reopen otherwise.

Issue #2571, opened 2017-08-03
Add user agent header for CLI http client


csantanapr:
#It will be useful for operators to distinguish http clients, this way we can see how many % of users/usage is using the CLI vs other type of http clients.

This is more useful for deployment and crud operations, to see the usage of the tools.

Set a default value something like "OpenWhisk-cli...", and allow to override this thru env variables in ansible for example, for operators compiling the code with modifications and be able to override it.

cc @mdeuser

markusthoemmes:
duplicate of #2568?

csantanapr:
closed by #2951

PR #2591, opened 2017-08-08
User-Agent CLI/version header

drcariel:
#2568

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -80,11 +80,14 @@ wsk namespace list -v
 ```
 ```
 REQUEST:
-[GET]	https://openwhisk.ng.bluemix.net/api/v1/namespaces
+[GET]  https://openwhisk.ng.bluemix.net/api/v1/namespaces
 Req Headers
 {
   "Authorization": [
     "Basic XXXYYYY"
+  ],
+  "User-Agent": [
+    "OpenWhisk-CLI/1.0 (2017-08-10T20:09:30+00:00)"
   ]
 }
 RESPONSE:Got response with code 200
@@ -99,7 +102,7 @@ Response body received:
 ["john@example.com_dev"]
 ```
 
-As you can see you the printed information provides the properties of the HTTP request, it performs a HTTP method `GET` on the URL `https://openwhisk.ng.bluemix.net/api/v1/namespaces` using a Basic Authorization header `Basic XXXYYYY`.
+As you can see you the printed information provides the properties of the HTTP request, it performs a HTTP method `GET` on the URL `https://openwhisk.ng.bluemix.net/api/v1/namespaces` using a User-Agent header `OpenWhisk-CLI/1.0 (<CLI-Build-version>)` and Basic Authorization header `Basic XXXYYYY`.
 Notice that the authorization value is your base64-encoded OpenWhisk authorization string.
 The response is of content type `application/json`.
 
--- a/tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
+++ b/tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
@@ -57,6 +57,7 @@ class WskBasicUsageTests extends TestHelpers with WskTestHelpers {
   implicit val wskprops = WskProps()
   val wsk = new Wsk
   val defaultAction = Some(TestUtils.getTestActionFilename("hello.js"))
+  val usrAgentHeaderRegEx = """\bUser-Agent\b": \[\s+"OpenWhisk\-CLI/1.\d+.*"""
 
   behavior of "Wsk CLI usage"
 
@@ -114,6 +115,11 @@ class WskBasicUsageTests extends TestHelpers with WskTestHelpers {
 
       wsk.action.invoke(fullQualifiedName).stdout should include(s"ok: invoked /$fullQualifiedName")
       wsk.action.get(fullQualifiedName).stdout should include(s"ok: got action ${packageName}/${actionName}")
+  }
+
+  it should "include CLI user agent headers with outbound requests" in {
+    val stdout = wsk.cli(Seq("list", "--auth", wskprops.authKey) ++ wskprops.overrides, verbose = true).stdout
+    stdout should include regex (usrAgentHeaderRegEx)
   }
 
   behavior of "Wsk actions"
--- a/tools/cli/go-whisk-cli/commands/commands.go
+++ b/tools/cli/go-whisk-cli/commands/commands.go
@@ -30,6 +30,7 @@ import (
 
 var client *whisk.Client
 const DefaultOpenWhiskApiPath string = "/api"
+var UserAgent string = "OpenWhisk-CLI"
 
 func setupClientConfig(cmd *cobra.Command, args []string) (error){
     baseURL, err := getURLBase(Properties.APIHost, DefaultOpenWhiskApiPath)
@@ -60,6 +61,7 @@ func setupClientConfig(cmd *cobra.Command, args []string) (error){
         Version:    Properties.APIVersion,
         Insecure:   flags.global.insecure,
         Host:       Properties.APIHost,
+        UserAgent:  UserAgent + "/1.0 (" + Properties.CLIVersion + ")",
     }
 
     // Setup client
--- a/tools/cli/go-whisk/whisk/api.go
+++ b/tools/cli/go-whisk/whisk/api.go
@@ -45,7 +45,6 @@ type ApiListRequestOptions struct {
     Skip            int       `url:"skip"`
     Docs            bool      `url:"docs,omitempty"`
 }
-//type ApiListResponse RetApiArray
 type ApiListResponse RetApiArray
 
 // wsk api get : Request, Response
@@ -105,7 +104,7 @@ type RetApiArray struct {
 type ApiItem struct {
     ApiId           string    `json:"id,omitempty"`
     QueryKey        string    `json:"key,omitempty"`
-    ApiValue        *RetApi `json:"value,omitempty"`
+    ApiValue        *RetApi   `json:"value,omitempty"`
 }
 
 type RetApi struct {
--- a/tools/cli/go-whisk/whisk/client.go
+++ b/tools/cli/go-whisk/whisk/client.go
@@ -76,6 +76,7 @@ type Config struct {
     Verbose   	bool
     Debug       bool     // For detailed tracing
     Insecure    bool
+    UserAgent   string
 }
 
 func NewClient(httpClient *http.Client, config *Config) (*Client, error) {
@@ -124,6 +125,10 @@ func NewClient(httpClient *http.Client, config *Config) (*Client, error) {
 
     if len(config.Version) == 0 {
         config.Version = "v1"
+    }
+
+    if len(config.UserAgent) == 0 {
+        config.UserAgent = "OpenWhisk-Go-Client"
     }
 
     c := &Client{
@@ -203,6 +208,8 @@ func (c *Client) NewRequest(method, urlStr string, body interface{}, includeName
         werr := MakeWskErrorFromWskError(errors.New(errStr), err, EXIT_CODE_ERR_GENERAL, DISPLAY_MSG, NO_DISPLAY_USAGE)
         return nil, werr
     }
+
+    req.Header.Add("User-Agent", c.Config.UserAgent)
 
     return req, nil
 }
@@ -651,6 +658,8 @@ func (c *Client) NewRequestUrl(
     } else {
         Debug(DbgInfo, "No auth header required\n")
     }
+
+    req.Header.Add("User-Agent", c.Config.UserAgent)
 
     return req, nil
 }


drcariel:
Looks like this in action:
```
"User-Agent": [
    "OpenWhisk-CLI/2017-08-09T19:14:44+00:00"
  ]
```

tools/cli/go-whisk-cli/main.go
```diff
@@ -48,6 +48,7 @@ func init() {
 
     // Rest of CLI uses the Properties struct, so set the build time there
     commands.Properties.CLIVersion = CLI_BUILD_TIME
+    os.Setenv("CLI_Version", CLI_BUILD_TIME)
```
mdeuser: avoid using "global" style variables.  instead set a value on the http config object during it's initialization

drcariel: I was attempting this at first, but isn't as simple. The CLIVersion is intentionally omitted from that config. I assume for good reason?

drcariel: I would either need to inject the CLIVersion properly into ```cli/go-whisk-cli/commands/property.go``` then import this into ```/cli/go-whisk/whisk/client.go```, or just import ```cli/go-whisk-cli/main.go``` into ```/cli/go-whisk/whisk/client.go```.

Not the greatest options imo

mdeuser: maybe this will work.
- add AgentHeader property to the whisk.Config struct
- in commands.go (line 54 or so), set the AgentHeader value in the client configuration init

drcariel:
Thanks for the input @mdeuser , not sure why I found it so difficult the first time round.

drcariel:
PG2~1935

tools/cli/go-whisk/whisk/api.go
```diff
@@ -43,7 +43,6 @@ type ApiListRequestOptions struct {
     Skip            int       `url:"skip"`
     Docs            bool      `url:"docs,omitempty"`
 }
-//type ApiListResponse RetApiArray
 type ApiListResponse RetApiArray
```
drcariel: Some lingering cleanup from the experimental-api removal PR

drcariel:
PG2~1937

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -326,7 +326,7 @@ trait WhiskWebActionsApi
     private val defaultCorsResponse = List(
         `Access-Control-Allow-Origin`(AllOrigins),
         `Access-Control-Allow-Methods`(OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH),
-        `Access-Control-Allow-Headers`(`Authorization`.name, `Content-Type`.name))
+        `Access-Control-Allow-Headers`(`Authorization`.name, `Content-Type`.name, `User-Agent`.name))
```
rabbah: do we want this for a web action?

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -121,7 +121,7 @@ protected[controller] object RestApiCommons {
  */
 protected[controller] trait RespondWithHeaders extends Directives {
     val allowOrigin = `Access-Control-Allow-Origin`(AllOrigins)
-    val allowHeaders = `Access-Control-Allow-Headers`("Authorization", "Content-Type")
+    val allowHeaders = `Access-Control-Allow-Headers`("Authorization", "Content-Type", "User-Agent")
```
rabbah: what is the value of the controller accepting the user-agent?
should nginx just drop this header?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
mdeuser: imho, i think giving web actions visibility into the agent header is a good thing

rabbah: But the controller does nothing with these user agents for POST. And accepting it here for web actions is just for preflight - I suppose it's fine (but for the cli this has no value there). If you want to keep it I think we need a doc update.

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
mdeuser: if the controller log the user agent header, log queries can see which clients and which client versions are accessing the controller.

rabbah: the controller does not log user agents - i don't think it should/needs to.
you get that from nginx.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
drcariel: rest_api.md has been updated

drcariel:
PG1~1942

tests/src/test/scala/whisk/core/cli/test/WskWebActionsTests.scala
```diff
@@ -47,94 +47,16 @@ import whisk.core.entity.Subject
...
+    with WskTestHelpers with RestUtil with BeforeAndAfterAll {
 
     val MAX_URL_LENGTH = 8192 // 8K matching nginx default
 
     val wsk = new Wsk
     private implicit val wskprops = WskProps()
     val namespace = wsk.namespace.whois()
 
-    protected val testRoutePath: String
+    protected val testRoutePath: String = "/api/v1/web"
```
drcariel: Is this appropriate here? or will testRoutePath be populated at this point?

drcariel:
PG2~1955

tools/cli/go-whisk/whisk/client.go
```diff
@@ -204,6 +209,8 @@ func (c *Client) NewRequest(method, urlStr string, body interface{}, includeName
         return nil, werr
     }
 
+    req.Header.Add("User-Agent", "OpenWhisk-CLI/" + c.Config.CLIVersion)
```
mdeuser: Use a constant for the "OpenWhisk-CLI/" string

tools/cli/go-whisk/whisk/client.go
```diff
@@ -126,6 +127,10 @@ func NewClient(httpClient *http.Client, config *Config) (*Client, error) {
         config.Version = "v1"
     }
 
+    if len(config.CLIVersion) == 0 {
+        config.CLIVersion = "not_set"
+    }
+
```
mdeuser: In your testing did you find this default setting to be needed?  i was thinking this check/set is not necessary due to the way the CLI version string is initialized

rabbah:
I'd like to raise again the question of what we are gaining by changing the preflight responses for webactions or why thosnchsnge has to affect the controller at all.

tools/cli/go-whisk/whisk/client.go
drcariel: didn't check, just followed suit with the others, will test it out

mdeuser:
@rabbah - to clarify your CORS related comment above.  this is in reference to a couple lines that look like
```val allowHeaders = `Access-Control-Allow-Headers`("Authorization", "Content-Type", "User-Agent")```

if this is not place, would web apps issuing cross origin api requests (implemented by web actions) still be allowed to provide a user agent header in the request?

File: tests/src/test/scala/whisk/core/cli/test/WskWebActionsTests.scala [hunk 8]
```diff
@@ -47,94 +47,16 @@ import whisk.core.entity.Subject
  * Tests web actions.
  */
 @RunWith(classOf[JUnitRunner])
-class WskWebActionsTestsV1 extends WskWebActionsTests {
```
dubeejw: There are still two versions of web actions. Why remove the tests for both versions?

tests/src/test/scala/whisk/core/cli/test/WskWebActionsTests.scala
```diff
@@ -250,7 +172,7 @@ trait WskWebActionsTests
                 response.statusCode shouldBe 200
                 response.header("Access-Control-Allow-Origin") shouldBe "*"
                 response.header("Access-Control-Allow-Methods") shouldBe "OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH"
-                response.header("Access-Control-Allow-Headers") shouldBe "Authorization, Content-Type"
+                response.header("Access-Control-Allow-Headers") shouldBe "Authorization, Content-Type, User-Agent"
```
dubeejw: Not sure how this is related to the CLI since web actions cannot be invoked using the CLI.

dubeejw: @mdeuser, see above.

mdeuser: true.  not related to cli, per-se, but web apps making api requests having web app specific user agent headers should be supported.  see related question above https://github.com/apache/incubator-openwhisk/pull/2591#issuecomment-322465029

dubeejw: @mdeuser, if the user wants to have access to the user-agent, he or she can already ad a custom CORS header.

dubeejw: @mdeuser, @drcariel see: https://github.com/apache/incubator-openwhisk/blob/master/docs/webactions.md#options-requests.

drcariel: Maybe I am confused, but I thought this user-agent header was for our own tracking data, not any thing that is exposed to the user for customization. Would enforcing this for web actions potentially hinder users from implementing their own user-agent header? and/or also force them to add a user-agent to all their web action calls?

mdeuser: @dubeejw - understood.  just wanted to be clear about the default behavior.  the docs currently do not tell the user that the Access-Control-Allow-Headers preflight response header is added
https://github.com/apache/incubator-openwhisk/blob/master/docs/webactions.md#options-requests

mdeuser: @drcariel - requests to the controller can come from the cli, gui, apps using the rest interface, and web action invocations.  each can and will have a different user agent header, so the controller must be application agnostic wrt user agent header handling.  the controller does not enforce that a user agent header be part of an incoming request.  if the user agent header is included, then the controller will handle it like any other benign header the controller does not explicitly act upon.

dubeejw: @drcariel, can you split the controller changes out into a separate PR since they are unrelated to the CLI changes?

docs/rest_api.md
```diff
@@ -80,11 +80,14 @@ wsk namespace list -v
...
 ```
 REQUEST:
-[GET]	https://openwhisk.ng.bluemix.net/api/v1/namespaces
+[GET]  https://openwhisk.ng.bluemix.net/api/v1/namespaces
 Req Headers
 {
   "Authorization": [
     "Basic XXXYYYY"
+  ],
+  "User-Agent": [
```
rabbah: back out changes to this file?

drcariel: I backed out the changes referencing the rest api requiring User-Agent, but this is an appropriate doc update for the CLI changes in my mind.

mdeuser: this is an example of a cli command with the `-v` which displays request header.  when the cli inserts its own agent-header, this will now appear in the `-v` output.. so, this update looks good.  was there something else you had in mind @rabbah ?

mdeuser:
Fixes #2568 
Fixes #2571

tools/cli/go-whisk/whisk/client.go
```diff
@@ -47,6 +47,7 @@ const (
     DoNotProcessTimeOut = false
     ExitWithErrorOnTimeout = true
     ExitWithSuccessOnTimeout = false
+    UserAgent = "OpenWhisk-CLI/"
```
mdeuser: @drcariel - after talking with @csantanapr, we think this would be better defined as a client.go externalized variable that could be overridden by client.go consumers.

dubeejw: Agreed. `go-whisk` does not know that it is part of a CLI.

tools/cli/go-whisk/whisk/client.go
```diff
@@ -47,6 +47,7 @@ const (
     DoNotProcessTimeOut = false
     ExitWithErrorOnTimeout = true
     ExitWithSuccessOnTimeout = false
+    UserAgent = "CloudFunctions-Client"
```
dubeejw: Use OpenWhisk-Client here or something simliar

tools/cli/go-whisk/whisk/client.go
```diff
@@ -76,6 +77,7 @@ type Config struct {
     Verbose   	bool
     Debug       bool     // For detailed tracing
     Insecure    bool
+    CLIVersion  string
```
dubeejw: The client should not know about the existence of a CLI as it is an abstraction.

dubeejw: You can just have a UserAgent value here that can be modified by the CLI.

tools/cli/go-whisk-cli/commands/commands.go
```diff
@@ -60,6 +60,8 @@ func setupClientConfig(cmd *cobra.Command, args []string) (error){
         Version:    Properties.APIVersion,
         Insecure:   flags.global.insecure,
         Host:       Properties.APIHost,
+        CLIVersion: Properties.CLIVersion,
```
dubeejw: How does CLIVersion get set?

tools/cli/go-whisk/whisk/client.go
```diff
@@ -204,6 +210,8 @@ func (c *Client) NewRequest(method, urlStr string, body interface{}, includeName
         return nil, werr
     }
 
+    req.Header.Add("User-Agent", c.Config.UserAgent + "/" + c.Config.CLIVersion)
```
dubeejw: Probably better to move CLIVersion logic into go-whisk-cli and just set the agent to `c.Config.UserAgent`.

tools/cli/go-whisk/whisk/client.go
```diff
@@ -126,6 +128,10 @@ func NewClient(httpClient *http.Client, config *Config) (*Client, error) {
         config.Version = "v1"
     }
 
+    if len(config.UserAgent) == 0 {
+        config.UserAgent = "Unknown"
```
dubeejw: Perhaps we could default to `OpenWhisk-Go-Client` or something like that.

mdeuser:
@drcariel New changes look good to me.  Will also need a peer PR for custom client user-agent value.

tools/cli/go-whisk-cli/commands/commands.go
```diff
@@ -60,6 +60,7 @@ func setupClientConfig(cmd *cobra.Command, args []string) (error){
         Version:    Properties.APIVersion,
         Insecure:   flags.global.insecure,
         Host:       Properties.APIHost,
+        UserAgent:  "OpenWhisk-CLI/" + Properties.CLIVersion,
```
mdeuser: @drcariel - define the value used by the cli in an exported string variable, then reference it here

tools/cli/go-whisk-cli/commands/commands.go
```diff
@@ -30,6 +30,7 @@ import (
 
 var client *whisk.Client
 const DefaultOpenWhiskApiPath string = "/api"
+var UserAgent string = "OpenWhisk-CLI"
```
dubeejw: Looks like this should be a const.

tools/cli/go-whisk/whisk/client.go
```diff
@@ -126,6 +127,10 @@ func NewClient(httpClient *http.Client, config *Config) (*Client, error) {
         config.Version = "v1"
     }
 
+    if len(config.UserAgent) == 0 || config.UserAgent == "/not set" {
```
dubeejw: Where does `/not set` come from?

drcariel: the CLI version, In my testing I saw the plugin will show `OW-Plugin/not set`, so I added it here to cover the missing UserAgent case, aka "/not set"

tools/cli/go-whisk-cli/commands/commands.go
drcariel: not a const so that it can be overwritten by the plugin etc

tools/cli/go-whisk/whisk/client.go
dubeejw: Do you think that is because the version isn't being set in that scenario? @mdeuser, see the above.

drcariel: the plugin doesn't use the same gradle build I believe (or maybe I didn't build it correctly), so yes, it is not being set. I was thinking I can check for this prior to appending the cli version in commands.go if we wanna move it.

mdeuser: from main.go
```
// CLI_BUILD_TIME holds the time of the CLI build.  During gradle builds,
// this value will be overwritten via the command:
//     go build -ldflags "-X main.CLI_BUILD_TIME=nnnnn"   // nnnnn is the new timestamp
var CLI_BUILD_TIME string = "not set"
```

dubeejw: Might be easier to just set the version in the plugin build.

mdeuser: yep.. it already is.  the "not set" CLI version is a result of a local build i.e. `go build main.go` without providing the CLI_BUILD_TIME linker flag

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -314,6 +314,11 @@ class WskBasicUsageTests
             }, waitBeforeRetry = Some(1.second), N = 5)
     }
 
+    it should "include CLI user agent headers with outbound requests" in {
+        val stdout = wsk.cli(Seq("action", "list", "-v", "--auth", wskprops.authKey) ++ wskprops.overrides).stdout
```
mdeuser: i recommend using the wsk.action.list function.  to do that, augment the ListOrGetFromCollection list() function with a `verbose: Boolean = false` parameter and explicitly pass that to the underlying cli() call.

tools/cli/go-whisk/whisk/client.go
mdeuser: oops. sorry,  as soon as i saw the "not set" i was thinking this was in reference to the cli version...

mdeuser: sooo.. i don't believe the user-agent var will every have a "not set" value.

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -314,6 +314,11 @@ class WskBasicUsageTests
...
+        stdout should include regex ("""\bUser-Agent\b": \[(\r\n|\r|\n)\s+"OpenWhisk\-CLI/201.*""".r)
```
dubeejw: Little simpler to just check for whitespace instead of the carriage returns and new lines. Also seems like the test will fail in 2020.`"""\bUser-Agent\b": \[\s+"OpenWhisk\-CLI/\d+.*""".r`

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -119,6 +119,10 @@ class WskBasicUsageTests
             wsk.action.get(fullQualifiedName).stdout should include(s"ok: got action ${packageName}/${actionName}")
     }
 
+    it should "include CLI user agent headers with outbound requests" in {
+        wsk.action.list(verboseFlag = Some(true)).stdout should include regex ("""(?i)User-Agent": \[(\r\n|\r|\n)\s+"OpenWhisk\-CLI/201.*""")
```
dubeejw: Little simpler to just check for whitespace instead of the carriage returns and new lines. Also seems like the test will fail in 2020. `"""\bUser-Agent\b": \[\s+"OpenWhisk\-CLI/\d+.*""".r`

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
drcariel: yes, you're right about 2020, I chose that to follow suit with the rest of the CLI version tests in this file. I will test a cleaner regex out

tests/src/test/scala/common/Wsk.scala
```diff
@@ -142,17 +142,20 @@ trait ListOrGetFromCollection extends FullyQualifiedNames {
...
         nameSort: Option[Boolean] = None,
-        expectedExitCode: Int = SUCCESS_EXIT)(
+        expectedExitCode: Int = SUCCESS_EXIT,
+        verboseFlag: Option[Boolean] = None)(
             implicit wp: WskProps): RunResult = {
         val params = Seq(noun, "list", resolve(namespace), "--auth", wp.authKey) ++
             { limit map { l => Seq("--limit", l.toString) } getOrElse Seq() } ++
-            { nameSort map { n => Seq("--name-sort") } getOrElse Seq() }
+            { nameSort map { n => Seq("--name-sort") } getOrElse Seq() } ++ 
+            { verboseFlag map { v => Seq("-v")} getOrElse Seq()}
```
mdeuser: use the cli() function provided `verbose` argument.

dubeejw: https://github.com/apache/incubator-openwhisk/blob/master/tests/src/test/scala/common/Wsk.scala#L1041

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
dubeejw: I see there are other tests that will fail in 2020 as well. We can fix those in a separate PR though.

dubeejw:
```
Syntax of a product token

The product identifier is described in the spec as

     product         = token["/' product-version]
     product-version = token
 
This is fairly self explanatory except for the fact that we have no idea what the syntax of a token is! If you dig around you will find that token is defined in httpbis part1 as,

        token     = 1*tchar 

Having fun yet?  So a token is one or more tchars.  A tchar is defined as,

     tchar     = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                    / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                    / DIGIT / ALPHA
                    ; any VCHAR, except delimiters

The terms DIGIT, ALPHA and VCHAR are defined in yet another IETF spec here, but what are these delimiters that we are not allowed to use?  The delimiters that are not allowed to appear in a token are defined as:

 (DQUOTE and "(),/:;<=>?@[\]{}")
For those of you still paying attention, you may be saying, but I see semi-colons in user agent strings all the time.  That is because they are in the comment not in the product token.  The “comment” has a different set of syntax rules.  The following is valid example of user-agent that is full of special characters,

user-agent: foo&bar-product!/1.0a$*+ (a;comment,full=of/delimiters@{fun})

```

http://bizcoder.com/the-much-maligned-user-agent-header

dubeejw:
Maybe replace `:` with `-`, or put the timestamp part after the date in a comment. @mdeuser, what do you think?

mdeuser:
@dubeejw - good catch.  there has been talk of adding semantic versioning to the various components, so perhaps this build timestamp is good to put inside a comment - leaving the version value open for a future semantic version value.

drcariel:
@dubeejw and @mdeuser can you show me a desired example? In my head, I am going to update this PR to show the User Agent like so:
```
"User-Agent": [
    "OpenWhisk-CLI/1.0 (2017-08-09T19:14:44+00:00)"
  ]
```
Is that what ya'll are looking for?

mdeuser:
@drcariel - is that error log message now gone?

drcariel:
Where would I see that log? In Kibana?

drcariel:
I could not find any logs pertaining to "Illegal header" in PG3~1051.

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -119,6 +119,10 @@ class WskBasicUsageTests
             wsk.action.get(fullQualifiedName).stdout should include(s"ok: got action ${packageName}/${actionName}")
     }
 
+    it should "include CLI user agent headers with outbound requests" in {
+        val stdout = wsk.action.list(verbose = true).stdout
+        stdout should include regex ("""\bUser-Agent\b": \[\s+"OpenWhisk\-CLI/1.\d+.*""")
```
rabbah: since the goal is just to run a cli command with verbose on, i'd rather you don't change the `list` method and install use `wsk.cli(Seq("list"), verbose = true)` directly.

mdeuser: @rabbah - oh.. i'm wondering why the entity functions don't have a `verbose` pass through to the `wsk.cli` fuction?

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -120,7 +120,7 @@ class WskBasicUsageTests
     }
 
     it should "include CLI user agent headers with outbound requests" in {
-        val stdout = wsk.cli(Seq("list"), verbose = true).stdout
+        val stdout = wsk.cli(Seq("list", "--auth", wskprops.authKey) ++ wskprops.overrides, verbose = true).stdout
         stdout should include regex ("""\bUser-Agent\b": \[\s+"CloudFunctions\-CLI/1.\d+.*""")
```
rabbah: CloudFunctions?

drcariel: Testing something out, will change back shortly.

rabbah:
Is the OS info part of the header?

dubeejw:
@rabbah, @drcariel, @mdeuser, looks like we can easily append the OS to the user-agent if we want to. To do so, import `runtime`, and call `runtime.GOOS`, which will return `windows`, `linux`, or `darwin`.

dubeejw:
The runtime package also provides the ability to retrieve the CPU architecture through `runtime.GOARCH`.

mdeuser:
i'd prefer to do that OS addition in a separate PR.

dubeejw:
I think this one should be ready for merge after Travis checks out.

PR #2592, opened 2017-08-08
Make the result of response consistent for both Wsk CLI and REST

houshengbo:
Rename CliActivationResponse & CliActivation, and change them from inner classes to classes

This PR renames the class CliActivation and CliActivationResponse
and makes them available to be accessed from another package, because
REST call can use both of the classes as well.

The PR also creates RestResult class to save the results of HTTP reopnses for REST.

Partially-closes-bug: #2430

### DIFF
--- a/tests/src/test/scala/common/WskTestHelpers.scala
+++ b/tests/src/test/scala/common/WskTestHelpers.scala
@@ -17,6 +17,10 @@
 
 package common
 
+import java.time.Instant
+
+import org.scalatest.Matchers
+
 import scala.collection.mutable.ListBuffer
 import scala.util.Failure
 import scala.util.Try
@@ -24,11 +28,93 @@ import scala.concurrent.duration.Duration
 import scala.concurrent.duration.DurationInt
 import scala.language.postfixOps
 
-import org.scalatest.Matchers
-
-import TestUtils._
 import spray.json._
-import java.time.Instant
+
+import TestUtils.RunResult
+import TestUtils.CONFLICT
+
+/**
+ * An arbitrary response of a whisk action. Includes the result as a JsObject as the
+ * structure of "result" is not defined.
+ *
+ * @param result a JSON object used to save the result of the execution of the action
+ * @param status a string used to indicate the status of the action
+ * @param success a boolean value used to indicate whether the action is executed successfully or not
+ */
+case class ActivationResponse(result: Option[JsObject], status: String, success: Boolean)
+
+object ActivationResponse extends DefaultJsonProtocol {
+    implicit val serdes = jsonFormat3(ActivationResponse.apply)
+}
+
+/**
+ * Activation record as it is returned from the OpenWhisk service.
+ *
+ * @param activationId a String to save the ID of the activation
+ * @param logs a list of String to save the logs of the activation
+ * @param response an Object of ActivationResponse to save the response of the activation
+ * @param start an Instant to save the start time of activation
+ * @param end an Instant to save the end time of activation
+ * @param duration a Long to save the duration of the activation
+ * @param cases String to save the cause of failure if the activation fails
+ * @param annotations a list of JSON objects to save the annotations of the activation
+ */
+case class ActivationResult(
+    activationId: String,
+    logs: Option[List[String]],
+    response: ActivationResponse,
+    start: Instant,
+    end: Instant,
+    duration: Long,
+    cause: Option[String],
+    annotations: Option[List[JsObject]]) {
+
+    def getAnnotationValue(key: String): Option[JsValue] = {
+        Try {
+            val annotation = annotations.get.filter(x => x.getFields("key")(0) == JsString(key))
+            assert(annotation.size == 1) // only one annotation with this value
+            val value = annotation(0).getFields("value")
+            assert(value.size == 1)
+            value(0)
+        } toOption
+    }
+}
+
+object ActivationResult extends DefaultJsonProtocol {
+    private implicit val instantSerdes = new RootJsonFormat[Instant] {
+        def write(t: Instant) = t.toEpochMilli.toJson
+
+        def read(value: JsValue) = Try {
+            value match {
+                case JsNumber(i) => Instant.ofEpochMilli(i.bigDecimal.longValue)
+                case _           => deserializationError("timetsamp malformed")
+            }
+        } getOrElse deserializationError("timetsamp malformed 2")
+    }
+
+    implicit val serdes = new RootJsonFormat[ActivationResult] {
+        private val format = jsonFormat8(ActivationResult.apply)
+
+        def write(result: ActivationResult) = format.write(result)
+
+        def read(value: JsValue) = {
+            val obj = value.asJsObject
+            obj.getFields("activationId", "response", "start") match {
+                case Seq(JsString(activationId), response, start) =>
+                    Try {
+                        val logs = obj.fields.get("logs").map(_.convertTo[List[String]])
+                        val end = obj.fields.get("end").map(_.convertTo[Instant]).getOrElse(Instant.EPOCH)
+                        val duration = obj.fields.get("duration").map(_.convertTo[Long]).getOrElse(0L)
+                        val cause = obj.fields.get("cause").map(_.convertTo[String])
+                        val annotations = obj.fields.get("annotations").map(_.convertTo[List[JsObject]])
+                        new ActivationResult(activationId, logs, response.convertTo[ActivationResponse],
+                            start.convertTo[Instant], end, duration, cause, annotations)
+                    } getOrElse deserializationError("Failed to deserialize the activation result.")
+                case _ => deserializationError("Failed to deserialize the activation ID, response or start.")
+            }
+        }
+    }
+}
 
 /**
  * Test fixture to ease cleaning of whisk entities created during testing.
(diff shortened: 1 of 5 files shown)


tests/src/test/scala/common/WskTestHelpers.scala
```diff
@@ -31,6 +31,55 @@ import spray.json._
...
+/**
+ * Activation record as it is returned by the CLI.
+ */
+ case class activationResult (
+        activationId: String,
+        logs: Option[List[String]],
+        response: activationResponse,
+        start: Instant,
+        end: Option[Instant], // This field is optional in responses of REST calls.
+        duration: Option[Long], // This field is optional in responses of REST calls.
```
markusthoemmes: How come those are optional?

tests/src/test/scala/common/WskTestHelpers.scala
```diff
@@ -31,6 +31,55 @@ import spray.json._
 import java.time.Instant
 
 /**
+ * An arbitrary response of a whisk action. Includes the result as a JsObject as the
+ * structure of "result" is not defined.
+ */
+case class activationResponse(result: Option[JsObject], status: String, success: Boolean)
```
markusthoemmes: Classes start with an uppercase letter.

tests/src/test/scala/common/WskTestHelpers.scala
```diff
@@ -258,3 +259,6 @@ trait WskTestHelpers extends Matchers {
         }
     }
 }
+
+
+
```
markusthoemmes: Remove the unnecessary whitespace.

markusthoemmes:
Can you explain how you want to tackle the migration from `Wsk` to `WskRest` or link a respective issue? I could not distill that from the open pull-requests which have no apparent connection. Would it make sense to merge some of those to one PR?

tests/src/test/scala/common/WskTestHelpers.scala
houshengbo: When we try to access the information of activation, sometimes the activation only has start available, no end or duration.
In this situation, CLI will fill in end and duration with some default values, while REST response will have them missing. When we to convertTo[activationResponse], there is no problem for CLI response, cos they will have values anyway. However, if we convert the http response into activationResponse, we need to tolerate end and duration missing. That is why they are optional.

houshengbo:
@markusthoemmes I will link all of the PRs regarding REST to an issue, with the explanation of how to do it step by step.

tests/src/test/scala/common/WskTestHelpers.scala
rabbah: This is a case where the cli encodes a schema incrroctly. We need to rid the cli of all of these. @dubeejw can comment.

markusthoemmes: That sounds like we should either do the same as the CLI on the API side or not make them optional on the API side. To me it seems like this is only valid for trigger/rule activations. Should we reflect that in the types?

houshengbo: I guess it would bring convenience if API is able to provide the information in a consistent manner, defining what is optional or not. What would be the default values for them if they are not available. Client side directly picks up what is in the http response.

tests/src/test/scala/common/WskTestHelpers.scala
houshengbo: Done.

tests/src/test/scala/common/WskTestHelpers.scala
houshengbo: Done.

tests/src/test/scala/common/WskTestHelpers.scala
```diff
@@ -31,6 +31,55 @@ import spray.json._
 import java.time.Instant
 
 /**
+ * An arbitrary response of a whisk action. Includes the result as a JsObject as the
+ * structure of "result" is not defined.
+ */
+case class ActivationResponse(result: Option[JsObject], status: String, success: Boolean)
```
rabbah: remind me why result is an option.
can you add java/scala doc for the parameters.

houshengbo: When a trigger is fired, there will be an activation linking to the rule, whose response does not have result available:
{
    "namespace": "whisk.system",
    "name": "testRule",
    "version": "0.0.1",
    "subject": "whisk.system",
    "activationId": "c2f410c141b9422ca32ab63a7c43d08e",
    "cause": "a8adb44ac0e24b8a9f0486617300b5ae",
    "start": 1502940356583,
    "end": 0,
    "duration": 0,
    "response": {
        "status": "success",
        "statusCode": 0,
        "success": true
    },
    "logs": [],
    "annotations": [],
    "publish": false
}

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,108 @@
...
+
+    def getFieldJsValue(obj: JsObject, key: String): JsValue = {
+        val list = obj.getFields(key)
+        if (list.size == 0) JsObject() else list(0).toJson
+    }
+
+    def getFieldListJsObject(obj: JsObject, key: String): Vector[JsObject] = {
+        val list = obj.getFields(key)
+        if (list.size == 0) Vector(JsObject()) else list(0).toJson.convertTo[Vector[JsObject]]
+    }
```
markusthoemmes: All of those helpers look a lot like old-school javaish way to write this, which can be fine but it's overly verbose and I think entirely unneeded in this case.

Is there a specific reason all of those have empty response like "fallbacks"? Could you give an example where that's necessary?

Some proposed equivalents working with scala idioms like `Option`:

```scala
obj.fields.get(key).map(_.compactPrint).getOrElse("") == getField(obj, key)
obj.fields.get(key).map(_.asJsObject).getOrElse(JsObject()) == getFieldJsObject(obj, key)
obj.fields.get(key).getOrElse(JsObject()) == getFieldJsValue(obj, key)
```

You can see, that those methods do compose quite nicely and I am not opposed to helper methods at all in general. In this case though, they seem a bit off.

tests/src/test/scala/common/WskTestHelpers.scala
```diff
@@ -31,6 +31,55 @@ import spray.json._
...
+/**
+ * Activation record as it is returned by the CLI.
+ */
+ case class ActivationResult (
+        activationId: String,
+        logs: Option[List[String]],
+        response: ActivationResponse,
+        start: Instant,
+        end: Option[Instant], // This field is optional in responses of REST calls.
+        duration: Option[Long], // This field is optional in responses of REST calls.
```
markusthoemmes: Could we not make those optional but have them deserialize to default values? I think if `end` is ommitted, the Rest API would give it an `Instant.EPOCH` while the CLI omits it. Duration I think is always `0` if ommitted.

houshengbo: Do we add the default values to end and duration on the rest api server side or fill in them on the client side when we receive the response?

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: Thanks. I can change it to the scala style.

I was thinking the "key" specified to call the method, can be random, and does not exist in the response, so add the coverage to take care when it is missing.

markusthoemmes: Sure, but what are tests supposed to do if it is missing? In Scala, the idiomatic way of handling a value that might be missing is using an `Option`.

In the cases above:

```scala
obj.fields.get(key).map(_.compactPrint) // Option[String]
```

tests/src/test/scala/common/WskTestHelpers.scala
markusthoemmes: I propose we fill them in in this test-class to smoothen out differences that way.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: For example, we have a case "create, and invoke an action that times out to ensure the proper response is received" in WskBasicTests. We invoke an action with blocking and result, what we expect is that it timeout not due to service failure, but the action timeout.

For rest, ActivationId should exist in the response. I verify it by checking if it is equal to "". Test will do anything, just check if they are equal.

houshengbo: I would like to use obj.fields.get(key).map(_.convertTo[String]).getOrElse("") instead of obj.fields.get(key).map(_.compactPrint).getOrElse("") for getField(obj, key), because I would like to pick the string instead of JsString.

tests/src/test/scala/common/WskTestHelpers.scala
```diff
@@ -31,6 +31,91 @@ import spray.json._
...
+            def write(result: ActivationResult) = format.write(result)
+
+            def read(value: JsValue) = {
+                val obj = value.asJsObject
+                obj.getFields("activationId") match {
+                    case Seq(JsString(activationId)) =>
+                        Try {
+                            val logs = obj.fields.get("logs").map(_.convertTo[List[String]])
+                            val response = obj.fields.get("response").map(_.convertTo[ActivationResponse]).get
+                            val start = obj.fields.get("start").map(_.convertTo[Instant]).get
```
markusthoemmes: those `get`s are unsafe. Can you move `response` and `start` to the sequence above?

tests/src/test/scala/common/WskTestHelpers.scala
```diff
@@ -17,18 +17,105 @@
...
+                assert(annotation.size == 1) // only one annotation with this value
+                val value = annotation(0).getFields("value")
+                assert(value.size == 1)
+                value(0)
+            } toOption
+        }
+}
+
+object ActivationResult extends DefaultJsonProtocol {
+        private implicit val instantSerdes = new RootJsonFormat[Instant] {
```
markusthoemmes: Formatting looks wrong here, did you install the scala formatting properties and formatted these files with it?

tests/src/test/scala/common/WskTestHelpers.scala
```diff
@@ -17,18 +17,105 @@
...
+                        Try {
+                            val logs = obj.fields.get("logs").map(_.convertTo[List[String]])
+                            val end = obj.fields.get("end").map(_.convertTo[Instant]).getOrElse(Instant.EPOCH)
+                            val duration = obj.fields.get("duration").map(_.convertTo[Long]).getOrElse(0L)
+                            val cause = obj.fields.get("cause").map(_.convertTo[String])
+                            val annotations = obj.fields.get("annotations").map(_.convertTo[List[JsObject]])
+                            new ActivationResult(activationId, logs, response.convertTo[ActivationResponse],
+                                start.convertTo[Instant], end, duration, cause, annotations)
+                        } getOrElse deserializationError("Failed to deserialize the activation result.")
+                    case _ => throw new DeserializationException("Missing activation ID.")
```
markusthoemmes: Please update the DeserializationException's message, it's not valid anymore

tests/src/test/scala/common/WskTestHelpers.scala
houshengbo: Done

tests/src/test/scala/common/WskTestHelpers.scala
houshengbo: Done

tests/src/test/scala/common/WskTestHelpers.scala
houshengbo: Done

PR #2593, opened 2017-08-09
Add BaseWsk class for Wsk and WskRest to inherit

houshengbo:


ansible/environments/mac/hosts
```diff
@@ -4,25 +4,25 @@
...
-172.17.0.1          ansible_host=172.17.0.1 ansible_connection=local
+192.168.99.100         ansible_host=192.168.99.100ansible_connection=local
 
 [redis]
-172.17.0.1          ansible_host=172.17.0.1 ansible_connection=local
+192.168.99.100         ansible_host=192.168.99.100ansible_connection=local
 
 [apigateway]
-172.17.0.1          ansible_host=172.17.0.1 ansible_connection=local
+192.168.99.100         ansible_host=192.168.99.100ansible_connection=local
```
csantanapr: don't change this file, docker-machine is the one using `192.168.99.100`

rabbah: pretty sure this is an accidental commit.

houshengbo: Fixed.

houshengbo:
@csantanapr I have fixed this patch.

tests/src/test/scala/common/BaseWsk.scala
```diff
@@ -0,0 +1,406 @@
...
+             expectedExitCode: Int = SUCCESS_EXIT,
+             cliCfgFile: Option[String] = None)(implicit wp: WskProps): RunResult
+}
+
+trait RunWskCmd extends Matchers {
+
+  /**
+   * The base command to run.
+   */
+  def baseCommand = Wsk.baseCommand
```
dubeejw: Here you make a call out to the Wsk singleton object that is defined in Wsk.scala. Since this file is the base for Wsk to inherit from, I don't think it should reference Wsk.scala at all.

houshengbo:
@dubeejw I have already made the changes. Please give a review.

dubeejw:
PG1 2038 🔵

tests/src/test/scala/common/BaseWsk.scala
```diff
@@ -0,0 +1,366 @@
...
+
+  def console(duration: Duration, since: Option[Duration] = None, expectedExitCode: Int = SUCCESS_EXIT)(
+    implicit wp: WskProps): RunResult
+
+  def logs(activationId: Option[String] = None, expectedExitCode: Int = SUCCESS_EXIT, last: Option[Boolean] = None)(
+    implicit wp: WskProps): RunResult
+
+  def result(activationId: Option[String] = None, expectedExitCode: Int = SUCCESS_EXIT, last: Option[Boolean] = None)(
+    implicit wp: WskProps): RunResult
+
```
dubeejw: New line here, but not in other traits.

tests/src/test/scala/common/Wsk.scala
```diff
@@ -189,10 +127,10 @@ trait ListOrGetFromCollection extends FullyQualifiedNames {
   }
 }
 
-trait DeleteFromCollection extends FullyQualifiedNames {
+trait DeleteFromCollectionCLI extends BaseDeleteFromCollection {
   self: RunWskCmd =>
 
-  protected val noun: String
+  override protected val noun: String
```
dubeejw: Does this value have to be overridden if it is never set in this trait?

tests/src/test/scala/common/Wsk.scala
```diff
@@ -101,43 +72,10 @@ class Wsk() extends RunWskCmd {
...
-    namespace getOrElse s"$sep${wp.namespace}"
-  }
-}
-
-trait ListOrGetFromCollection extends FullyQualifiedNames {
+trait ListOrGetFromCollectionCLI extends BaseListOrGetFromCollection {
   self: RunWskCmd =>
 
-  protected val noun: String
+  override protected val noun: String
```
dubeejw: Does this value have to be overridden if it is never set in this trait?

tests/src/test/scala/common/BaseWsk.scala
houshengbo: Done

tests/src/test/scala/common/Wsk.scala
houshengbo: I think we can remove this line here, since it is defined in BaseDeleteFromCollection.

tests/src/test/scala/common/Wsk.scala
houshengbo: I think we can remove this line here, since it is defined in BaseListOrGetFromCollection.

tests/src/test/scala/common/BaseWsk.scala
```diff
@@ -0,0 +1,365 @@
...
+  key: String =
+    WhiskProperties.getFileRelativeToWhiskHome("ansible/roles/nginx/files/openwhisk-client-key.pem").getAbsolutePath,
+  namespace: String = "_",
+  apiversion: String = "v1",
+  apihost: String = WhiskProperties.getEdgeHost,
+  token: String = "") {
+  def overrides = Seq("-i", "--apihost", apihost, "--apiversion", apiversion)
+  def writeFile(propsfile: File) = {
+    val propsStr =
+      s"NAMESPACE=$namespace\nAPIVERSION=$apiversion\nAUTH=$authKey\nAPIHOST=$apihost\nAPIGW_ACCESS_TOKEN=$token\n"
```
markusthoemmes: Should we break this into a multi-line string?

```scala
val propsStr = s"""NAMESPACE=$namespace
                              |APIVERSION=$apiversion
                              |AUTH=$authKey
                              |APIHOST=$apihost
                              |APIGW_ACCESS_TOKEN=$token""".stripMargin

File: tests/src/test/scala/common/BaseWsk.scala [hunk 7]
```diff
@@ -0,0 +1,365 @@
...
+   * @return resolved namespace
+   */
+  def resolve(namespace: Option[String])(implicit wp: WskProps) = {
+    val sep = "/" // Namespace.PATHSEP
+    namespace getOrElse s"$sep${wp.namespace}"
+  }
+}
+
+trait BaseListOrGetFromCollection extends FullyQualifiedNames {
+  self: BaseRunWsk =>
```
markusthoemmes: Do we need the extra qualification to `BaseRunWsk` @rabbah? As `self` is unused, I guess not?

tests/src/test/scala/common/BaseWsk.scala
```diff
@@ -0,0 +1,365 @@
...
+   */
+  def get(name: String,
+          expectedExitCode: Int = SUCCESS_EXIT,
+          summary: Boolean = false,
+          fieldFilter: Option[String] = None,
+          url: Option[Boolean] = None)(implicit wp: WskProps): RunResult
+}
+
+trait BaseDeleteFromCollection extends FullyQualifiedNames {
+  self: BaseRunWsk =>
```
markusthoemmes: Same as above.

tests/src/test/scala/common/BaseWsk.scala
```diff
@@ -0,0 +1,365 @@
...
+    jsonStr.substring(jsonStr.indexOf("\n") + 1).parseJson.asJsObject // Skip optional status line before parsing
+  }
+
+  def reportFailure(args: Buffer[String], ec: Integer, rr: RunResult) = {
+    val s = new StringBuilder()
+    s.append(args.mkString(" ") + "\n")
+    if (rr.stdout.nonEmpty) s.append(rr.stdout + "\n")
+    if (rr.stderr.nonEmpty) s.append(rr.stderr)
+    s.append("exit code:")
+  }
```
markusthoemmes: Do all of these methods apply for both Rest and Cli? At least `parseJsonString` doesn't look like it.

markusthoemmes:
Generally looks okay. The issue I see with it: Without the other extension point (WskRest), we cannot be sure if the abstraction is okay. This is more or less just creating a trait out of the existing implementation. Would it make sense to provide `WskRest` as well or would that make the PR too big?

I'm also wondering if we need the common base class or if we could just override the existing `Wsk` using rest based methods step by step while abstracting away helpers, that should not be surfaced because they correspond only to either CLI or REST.

Does that make sense?

houshengbo:
@markusthoemmes 
I tried very hard before to override the Wsk without creating the BaseWsk and the other basic entities, but I failed. When REST is implemented by akka client, we need to extend FlatSpec and add with WskActorSystem for RunWskRestCmd, so that WskRestAction and WskRestPackage can inherit it. However, if we extend the current RunWskCmd, it would report a contradiction of by adding "FlatSpec" and "WskActorSystem". I have to create a base class and the base entities, so both CLI and REST can extend with no conflicts.

If you want to see how it looks like in further PR, go to https://github.com/apache/incubator-openwhisk/pull/2589.

tests/src/test/scala/common/BaseWsk.scala
houshengbo: I would create a class named RestResult to save the REST results. In order to keep the changes in test cases as few as possible, I make RestResult extend RunResult, so WskTestHelpers can work for both CLI and REST without changes. RestResult can be converted into RunResult.

tests/src/test/scala/common/BaseWsk.scala
houshengbo: Done

tests/src/test/scala/common/BaseWsk.scala
houshengbo: We can remove this line.

tests/src/test/scala/common/BaseWsk.scala
houshengbo: We can remove this line.

PR #2595, opened 2017-08-09
Remove defaultLimits and have limits only

ningyougang:
regarding `distributed deployment`, because `distributed/group_vars/all` and `ansible/group_vars/all` also don't define `limits` variable, and `- name: (re)start container` ansible task has a env `"LIMITS_ACTIONS_INVOKES_PERMINUTE": "{{ limits.actions.invokes.perMinute }}"`, so the deployment wil be failed, the ansible task reports `the variable limits doesn't be defined`

Currently, `loca`, `mac`, `docker-machine` in its own all file also have variable `limits`, but for `distributed`,it doesn't have variable `limits`.

How to solve this question for `distributed` and has no influences on other envs, i think the best way is move the variable `limits` to `ansible/group_vars/all`, all envs can inherit `limits` value from that file.


markusthoemmes:
Did you verify it's still possible to override those limits per env then?

ningyougang:
Sorry, I will test it.

ningyougang:
@markusthoemmes , i have tested `local` and `distributed` env, it is ok. 
But for `mac` and `docker-machine`, i don't have these envs in my local, so can't test it.

But i am sure for `mac` and `docker-machine`, it will work well, why?
You can see variable `defaultLimits`  is defined in `ansible/groups_vars/all` only,
the `- name: (re)start controller` ansible task has a ENV `"DEFAULTLIMITS_ACTIONS_INVOKES_PERMINUTE": "{{ defaultLimits.actions.invokes.perMinute }}"`. so after modified for `limits`, `limits` and `defaultLimits` are symmetrical

markusthoemmes:
@ningyougang it is key though, that you are able to override limits in an environment, which is why the `defaultLimits` are global and `limits` are "local". I agree the existing scheme isn't perfect, in fact it is quite wrong with the recent changes now.

What we should do:
1. Remove `defaultLimits`, have only `limits`
2. Make each key in `limits` overridable, such as

```
limits:
  invocationsPerMinute: "{{ limit_invocations_per_minute | default(120) }}"
  concurrentInvocations: "{{ limit_invocations_concurrent | default(100) }}"
  concurrentInvocationsSystem:  "{{ limit_invocations_concurrent_system | default(5000) }}"
  firesPerMinute: "{{ limit_fires_per_minute | default(60) }}"
  sequenceLength: "{{ limit_sequence_length | default(50) }}"
```

That is: It adjusts the wording to conform with the `UserLimits` class. It also allows selective overriding per environment (by defining one of the `limit_...` variables), but keeps defaults up top.

What do you think?

ningyougang:
@markusthoemmes ,for `defaultLimits` and `limits`, i just know it is using for `benchmarks test`,  i don't know here why remove `defaultLimits`, if remove, below statements will also be removed
```
      "DEFAULTLIMITS_ACTIONS_INVOKES_PERMINUTE": "{{ defaultLimits.actions.invokes.perMinute }}"
      "DEFAULTLIMITS_ACTIONS_INVOKES_CONCURRENT": "{{ defaultLimits.actions.invokes.concurrent }}"
      "DEFAULTLIMITS_TRIGGERS_FIRES_PERMINUTE": "{{ defaultLimits.triggers.fires.perMinute }}"
      "DEFAULTLIMITS_ACTIONS_INVOKES_CONCURRENTINSYSTEM": "{{ defaultLimits.actions.invokes.concurrentInSystem }}"
      "DEFAULTLIMITS_ACTIONS_SEQUENCE_MAXLENGTH": "{{ defaultLimits.actions.sequence.maxLength }}"
```
Has any bad influences?

markusthoemmes:
That's the thing: We don't need both. We used to need both because we didn't have a good way to have per environment overridable settings. We do so now though, with the approach I outlined.

ningyougang:
@markusthoemmes  i have removed `defaultLimits` from `ansible/group_vars/all` and have `limits` only

markusthoemmes:
@ningyougang great, can you alter the scala code as well?

1. Throw out all usages of default limits
2. Throw out the fallback mechanism of `WhiskConfig`
3. Throw out optional arguments alltogether.

I can help you pick it up if you need assistance.

ningyougang:
@markusthoemmes , i have modified again, please review again ^^

ansible/README.md
```diff
@@ -282,20 +282,24 @@ the path to your OpenWhisk `ansible` directory contains spaces. To fix this, ple
...
+```
+limit_invocations_per_minute: 60
+limit_invocations_concurrent: 30
+concurrentInvocationsSystem: 5000
+firesPerMinute: 60
+```
+- The `limit_invocations_per_minute` represents the allowed namespace action invocations per minute.
+- The `limit_invocations_concurrent` represents the maximum concurrent invocations allowed per namespace.
+- The `concurrentInvocationsSystem` represents the maximum concurrent invocations the system will allow across all namespaces.
+- The `firesPerMinute` represents the allowed namespace trigger firings per minute.
```
markusthoemmes: The keys here don't match. Please use the snake_cased version for each of the limits here.

markusthoemmes:
@ningyougang This looks super awesome, exactly what I meant! 🎉

ansible/README.md
ningyougang: ^^, have modified.

ansible/README.md
```diff
@@ -282,20 +282,24 @@ the path to your OpenWhisk `ansible` directory contains spaces. To fix this, ple
 without spaces as there is no current fix available to this problem.
 
 #### Changing limits
-The system throttling limits can be changed by modifying the `group_vars` for your environment. For example,
-mac users will find the limits in this file [./environments/mac/group_vars/all](./environments/mac/group_vars/all):
+The system throttling limits's default value is configured in this file [./group_vars/all](./group_vars/all),
```
rabbah: The system throttling limits's default value is ->
The default system throttling limits are

rabbah: end this sentence with a period instead of comma.

ansible/README.md
```diff
@@ -282,20 +282,24 @@ the path to your OpenWhisk `ansible` directory contains spaces. To fix this, ple
...
-- The `concurrent` under `limits->actions->invokes` represents the maximum concurrent invocations allowed per namespace.
-- The `concurrentInSystem` under `limits->actions->invokes` represents the maximum concurrent invocations the system will allow across all namespaces.
-- The `perMinute` under `limits->triggers-fires` represents the allowed namespace trigger firings per minute.
+  invocationsPerMinute: "{{ limit_invocations_per_minute | default(120) }}"
+  concurrentInvocations: "{{ limit_invocations_concurrent | default(100) }}"
+  concurrentInvocationsSystem:  "{{ limit_invocations_concurrent_system | default(5000) }}"
+  firesPerMinute: "{{ limit_fires_per_minute | default(60) }}"
+  sequenceMaxLength: "{{ limit_sequence_max_length | default(50) }}"
+```
+Its value can be changed by modifying the `group_vars` for your environment. For example,
```
rabbah: Its value can -> These values may

tests/src/test/scala/limits/ThrottleTests.scala
```diff
@@ -65,9 +65,9 @@ class ThrottleTests
 
     val throttleWindow = 1.minute
 
-    val maximumInvokesPerMinute = getLimit("defaultLimits.actions.invokes.perMinute", "limits.actions.invokes.perMinute")
-    val maximumFiringsPerMinute = getLimit("defaultLimits.triggers.fires.perMinute", "limits.triggers.fires.perMinute")
-    val maximumConcurrentInvokes = getLimit("defaultLimits.actions.invokes.concurrent", "limits.actions.invokes.concurrent")
+    val maximumInvokesPerMinute = getLimit("limits.actions.invokes.perMinute", null)
```
rabbah: I suggest you change the impl of `getLimit` in this class and drop the `null` argument.

ningyougang:
@markusthoemmes @rabbah , have modified ^^

ningyougang:
@markusthoemmes @rabbah rebase again and test.

PR #2600, opened 2017-08-10
Remove databases of old ReplicatorTests that failed

cbickel:
This PR does a cleanup before the ReplicatorTests starts.

In addition, it fixes the retry method in the replicator-script.

PG4#661 started.

### DIFF
--- a/tests/src/test/scala/whisk/core/database/test/ReplicatorTests.scala
+++ b/tests/src/test/scala/whisk/core/database/test/ReplicatorTests.scala
@@ -130,6 +130,17 @@ class ReplicatorTests extends FlatSpec
         }
     }
 
+    // Do cleanups of possibly existing databases
+    val dbs = replicatorClient.dbs().futureValue
+    dbs shouldBe 'right
+    dbs.right.get.filter(dbname => dbname.contains(testDbPrefix)).map(dbName => removeDatabase(dbName))
+
+    val docs = replicatorClient.getAllDocs().futureValue
+    docs shouldBe 'right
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/whisk/core/database/test/ReplicatorTests.scala
```diff
@@ -130,6 +130,17 @@ class ReplicatorTests extends FlatSpec
...
 
+    // Do cleanups of possibly existing databases
+    val dbs = replicatorClient.dbs().futureValue
+    dbs shouldBe 'right
+    dbs.right.get.filter(dbname => dbname.contains(testDbPrefix)).map(dbName => removeDatabase(dbName))
+
+    val docs = replicatorClient.getAllDocs().futureValue
+    docs shouldBe 'right
+    docs.right.get.fields("rows").convertTo[List[JsObject]]
+        .filter(_.fields("id").convertTo[String].contains(testDbPrefix))
```
markusthoemmes: Should that be `startsWith(testDbPrefix)` for good measure? Might need to do `startsWith("backup_" + testDbProfix) || startsWith("continuous_" + testDbPrefix)`

cbickel: The problem is, that the test creates databases with the prefix "continous_testDbPrefix", "backup_timestamp_testDbPrefix" and "testDbPrefix".
The testDbPrefix is already a combination of the string "replicatorTests" and the dbPrefix of the environment.
That's why I just use contains instead of handling all cases.

PR #2601, opened 2017-08-10
Invalidate Caches between controllers with Kafka

cbickel:
This PR adds the ability to invalidate caches between controllers on updating and deleting the cache on one of the controllers.
It also adds the case class `CacheKey`, to use a particular key for the WhiskCache instead of `Any`.

This PR is dependant on #2338


cbickel:
Replaced by #2624

PR #2603, opened 2017-08-11
Revert "Temporalily use published swift3Action. (#2580)"

markusthoemmes:
Build is fine again and binaries are back online.

This reverts commit 49ac756064905da0bcba8ed9978a3e29dd8c04dc.


rabbah:
should we revert this? 
seems like a good way to start along the path of fewer runtimes to build.

markusthoemmes:
I'd still revert this and build a proper process around runtimes first. Your point is understandable and we definitly should attack it. Still the current situation is not where we want to end up.

csantanapr:
I agree, ideal would be we could have repo per runtime and test and publish using Travis, and then in core we pull them in thru ansible pinned images

PR #2609, opened 2017-08-12
Support json response in HTTP web actions without base64 encoding.

### LINKED ISSUE #2606: application-json no longer considered "binary" after migration from spray to akka
so we can support json without base64 encoding/decoding for http web actions.
this will make a lot of users happy.

we'll support both modes for a while and then drop the requirement since it's just too common a case.

rabbah:
Closes #2606.

The current patches allows an HTTP web action to respond with a JSON value without having to base64 encode it. See issue for more details.

@dubeejw can you take a look.
PG1/1945 & 1946 🔵

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
@@ -42,7 +42,7 @@ import akka.http.scaladsl.model.headers.`Content-Type`
 import akka.http.scaladsl.model.ContentType
 import akka.http.scaladsl.model.ContentTypes
 import akka.http.scaladsl.model.FormData
-import akka.http.scaladsl.model.HttpMethods.{ OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH}
+import akka.http.scaladsl.model.HttpMethods.{ OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH }
 import akka.http.scaladsl.model.HttpCharsets
 
 import spray.json._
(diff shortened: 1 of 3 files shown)


core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -241,7 +241,7 @@ protected[core] object WhiskWebActionsApi extends Directives {
 
             fields.get("body") map {
                 case JsString(str) => interpretHttpResponse(code, headers, str, transid)
-                case _             => terminate(BadRequest, Messages.httpContentTypeError)(transid)
+                case js            => interpretHttpResponseAsJson(code, headers, js, transid)
```
csantanapr: Those this case checks forts for js before it checks for base64 case? Making this code optimized for common case (i.e. Json string)

rabbah: I'd have to match against js object, array, etc. don't think it's necessary - jit should get there.

csantanapr: Ok

docs/webactions.md
```diff
@@ -102,11 +102,13 @@ function main(params) {
     return {
         statusCode: 200,
         headers: { 'Content-Type': 'application/json' },
-        body: new Buffer(JSON.stringify(params)).toString('base64'),
+        body: params
     };
 }
 ```
 
+The default content-type for an HTTP response is `application/json` and the body may be any allowed JSON value. The default content-type may be omitted from the headers.
```
dubeejw: There is a section in the doc entitled `Decoding binary body content from Base64`. Do we want to deprecate the Base64 functionality so that we can remove it at some point? @csantanapr, FYI.

rabbah: Content is still sent in base64 format for raw http actions. This is not affected by this pr.

dubeejw:
@csantanapr, want to merge?

PR #2616, opened 2017-08-14
Add retry to rate-throttle tests becasue of the possiblility, that it…

cbickel:
… gets refreshed during the test

In our pipeline, it happened, that the test `respect overridden rate-throttles of  1` failed, with the error `exit code: 0 was not equal to 173`.

The test started at 12:41:58.414 and was finished at 12:42:01.056.

So it failed, because the amount of invoked actions in the current minute has been reseted in the controller between the two calls.

PG4#670 started.

### DIFF
--- a/tests/src/test/scala/limits/ThrottleTests.scala
+++ b/tests/src/test/scala/limits/ThrottleTests.scala
@@ -45,6 +45,7 @@ import whisk.http.Messages._
 import whisk.utils.ExecutionContextFactory
 import org.scalatest.BeforeAndAfterAll
 import common.RunWskAdminCmd
+import whisk.utils.retry
 
 @RunWith(classOf[JUnitRunner])
 class ThrottleTests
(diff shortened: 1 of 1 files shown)


rabbah:
Is there a way we could attempt to settle the throttle instead?  

Hmm maybe if we provide the quota in the response in general it's useful and then the tests can deal with that accordingly.

rabbah:
Ah I misunderstood - so this is a case where we expected the request to be throttled and it wasn't. The calculation in the test setup then is wrong, it should compensate for this... iirc that was the intention.

markusthoemmes:
@rabbah these are the new tests I added for namespace specific throttles. They don't contain the settle/retry logic of the other ones.

markusthoemmes:
@cbickel the test is still race prone though because you can have overlapping fires where it'd still fail because the first one would get throttled then. Maybe we should fire twice regardless of return code and verify that at least one of them got the expected return code to guard against that?

PR #2619, opened 2017-08-14
Fix Illegal Header Warnings

dubeejw:
Closes https://github.com/apache/incubator-openwhisk/issues/2612

### DIFF
--- a/tests/dat/actions/base64Web.js
+++ b/tests/dat/actions/base64Web.js
@@ -1,0 +1,9 @@
+function main() {
+	  return {
+		  headers: {
+			  "content-type": "application/json"
+		  },
+		  statusCode: 200,
+		  body: new Buffer(JSON.stringify({'status': 'success'})).toString('base64')
+	  }
+}
--- a/tests/dat/actions/base64Web.js
+++ b/tests/dat/actions/base64Web.js
@@ -1,9 +1,9 @@
 function main() {
-	  return {
-		  headers: {
-			  "content-type": "application/json"
-		  },
-		  statusCode: 200,
-		  body: new Buffer(JSON.stringify({'status': 'success'})).toString('base64')
-	  }
+    return {
+        headers: {
+            "content-type": "application/json"
+        },
+        statusCode: 200,
+        body: new Buffer(JSON.stringify({'status': 'success'})).toString('base64')
+    }
 }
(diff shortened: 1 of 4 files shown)


dubeejw:
@rabbah, the configuration setting did not work. Had to filter the headers instead.

rabbah:
Ok. That's probably cleaner anyway. Thanks for knocking this off.

tests/dat/actions/base64Web.js
```diff
@@ -0,0 +1,9 @@
+function main() {
+	  return {
+		  headers: {
+			  "content-type": "application/json"
+		  },
+		  statusCode: 200,
+		  body: new Buffer(JSON.stringify({'status': 'success'})).toString('base64')
+	  }
+}
```
markusthoemmes: Contains tabs. Why does `scanCode` not catch this @mrutkows ?

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -788,7 +788,6 @@ trait WebActionsApiTests extends ControllerTestCommon with BeforeAndAfterEach wi
 
                         m(s"$testRoutePath/$path") ~> Route.seal(routes(creds)) ~> check {
                             status should be(OK)
-                            header("content-type").get.toString shouldBe "content-type: application/json"
```
rabbah: Ok - if we have other tests that assert the content type header, do we need the additional integration test?

dubeejw: The other unit tests look fine.

dubeejw:
PG4 691 ⏳ .

tests/dat/actions/base64Web.js
rabbah: the tests directory is exempted - chatted with @mrutkows to address.

mrutkows: @markusthoemmes Currently (and historically), the "tests" directory and its subdirs are listed in the "exclusions" section of the default scancode.cfg file.  Messaged Rodric that I would be happy to remove (and want to remove ALL exclusions), but have been waiting for dozens of PRs to merge to allow even considering this.  I will work to see what impact removing the "tests" folder exclusion and see if I can submit a PR.

PR #2621, opened 2017-08-15
Replace the test cases of action and package with REST implementation

houshengbo:
Partially-closes: #2430

### DIFF
--- a/tests/build.gradle
+++ b/tests/build.gradle
@@ -24,6 +24,17 @@ task testLean(type: Test) {
     exclude '**/*ThrottleTests*'
     exclude '**/MaxActionDurationTests*'
     exclude '**/*ApiGwTests*'
+    exclude '**/*Cli*'
+}
+
+task testLeanCli(type: Test) {
+    exclude '**/*Swift*'
+    exclude '**/*Python*'
+    exclude '**/*Java*'
+    exclude '**/*ThrottleTests*'
+    exclude '**/MaxActionDurationTests*'
+    exclude '**/*ApiGwTests*'
+    exclude '**/*Rest*'
 }
 
 // Add all images needed for local testing here
--- a/tests/src/test/scala/common/WskTestHelpers.scala
+++ b/tests/src/test/scala/common/WskTestHelpers.scala
@@ -280,9 +280,13 @@ trait WskTestHelpers extends Matchers {
     }
   }
 
-  def removeCLIHeader(response: String): String = response.substring(response.indexOf("\n"))
+  def removeCLIHeader(response: String): String = {
+    if (response.contains("\n")) response.substring(response.indexOf("\n")) else response
+  }
 
-  def getJSONFromCLIResponse(response: String): JsObject = removeCLIHeader(response).parseJson.asJsObject
+  def getJSONFromResponse(response: String, isCli: Boolean = false): JsObject = {
+    if (isCli) removeCLIHeader(response).parseJson.asJsObject else response.parseJson.asJsObject
+  }
 
   def getAdditionalTestSubject(newUser: String): WskProps = {
     val wskadmin = new RunWskAdminCmd {}
--- a/tests/src/test/scala/system/basic/WskActionTests.scala
+++ b/tests/src/test/scala/system/basic/WskActionTests.scala
@@ -24,6 +24,7 @@ import common.ActivationResult
 import common.JsHelpers
 import common.TestHelpers
 import common.TestUtils
+import common.BaseWsk
 import common.Wsk
 import common.WskProps
 import common.WskTestHelpers
@@ -33,10 +34,10 @@ import spray.json.JsObject
 import spray.json.pimpAny
 
 @RunWith(classOf[JUnitRunner])
-class WskActionTests extends TestHelpers with WskTestHelpers with JsHelpers {
+abstract class WskActionTests extends TestHelpers with WskTestHelpers with JsHelpers {
 
   implicit val wskprops = WskProps()
-  val wsk = new Wsk
+  val wsk: BaseWsk
 
   val testString = "this is a test"
   val testResult = JsObject("count" -> testString.split(" ").length.toJson)
@@ -140,8 +141,8 @@ class WskActionTests extends TestHelpers with WskTestHelpers with JsHelpers {
         action.create(copiedActionName, Some(origActionName), Some("copy"))
       }
 
-      val copiedAction = getJSONFromCLIResponse(wsk.action.get(copiedActionName).stdout)
-      val origAction = getJSONFromCLIResponse(wsk.action.get(copiedActionName).stdout)
+      val copiedAction = getJSONFromResponse(wsk.action.get(copiedActionName).stdout, wsk.isInstanceOf[Wsk])
+      val origAction = getJSONFromResponse(wsk.action.get(copiedActionName).stdout, wsk.isInstanceOf[Wsk])
 
       copiedAction.fields("annotations") shouldBe origAction.fields("annotations")
       copiedAction.fields("parameters") shouldBe origAction.fields("parameters")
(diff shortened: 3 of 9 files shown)


tests/src/test/scala/system/basic/WskRuleTests.scala
```diff
@@ -368,25 +369,32 @@ class WskRuleTests
...
 
+    def verifyRuleList(ruleListResult: RunResult, ruleNameEnable: String, ruleName: String) = {
+        val ruleList = ruleListResult.stdout
+        val listOutput = ruleList.lines
+        listOutput.find(_.contains(ruleNameEnable)).get should (include(ruleNameEnable) and include("active"))
+        listOutput.find(_.contains(ruleName)).get should (include(ruleName) and include("inactive"))
+        ruleList should not include ("Unknown")
+    }
 }
+
```
dubeejw: Remove comment.

tests/src/test/scala/system/basic/WskRuleTests.scala
```diff
@@ -89,7 +90,7 @@ class WskRuleTests
 
     behavior of "Whisk rules"
 
-    it should "invoke the action attached on trigger fire, creating an activation for each entity including the cause" in withAssetCleaner(wskprops) {
+    /*it should "invoke the action attached on trigger fire, creating an activation for each entity including the cause" in withAssetCleaner(wskprops) {
```
dubeejw: Does this test no longer work?

tests/src/test/scala/common/Wsk.scala
```diff
@@ -908,6 +771,15 @@ class WskApi()
             { apiname map { a => Seq("--apiname", a) } getOrElse Seq() } ++
             { swagger map { s => Seq("--config-file", s) } getOrElse Seq() } ++
             { responsetype map { t => Seq("--response-type", t) } getOrElse Seq() }
+        println(if (basepath == None) "" else basepath.get)
```
dubeejw: Remove `println()` calls.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -0,0 +1,1614 @@
...
+
+        val entPath = if (name != "") Path(s"$basePath/namespaces/$ns/$entityName/$name/") else Path(s"$basePath/namespaces/$ns/$entityName")
+        println("list path is " + entPath)
+        var paramMap = Map("skip" -> "0".toString, "docs" -> true.toString)
+        if (limit != None)
+            paramMap += ("limit" -> limit.get.toString)
+        else
+            paramMap += ("limit" -> "30".toString)
+        val resp = getWhiskEntity(entPath, paramMap).futureValue
+        //if (resp == None) {
```
dubeejw: Remove commetted out code and `println` calls from this file.

houshengbo:
PG 2174 in 1.

tests/build.gradle
```diff
@@ -24,6 +24,17 @@ task testLean(type: Test) {
     exclude '**/*ThrottleTests*'
     exclude '**/MaxActionDurationTests*'
     exclude '**/*ApiGwTests*'
+    exclude '**/*Cli*'
```
dubeejw: Hmm, don't think we should exclude the CLI tests until the CLI is ready to be removed from this repo.

dubeejw: I believe testLean is the task Travis runs when validating PRs.

tests/src/test/scala/common/WskTestHelpers.scala
```diff
@@ -280,9 +280,13 @@ trait WskTestHelpers extends Matchers {
...
   }
 
-  def removeCLIHeader(response: String): String = response.substring(response.indexOf("\n"))
+  def removeCLIHeader(response: String): String = {
+    if (response.contains("\n")) response.substring(response.indexOf("\n")) else response
+  }
 
-  def getJSONFromCLIResponse(response: String): JsObject = removeCLIHeader(response).parseJson.asJsObject
+  def getJSONFromResponse(response: String, isCli: Boolean = false): JsObject = {
+    if (isCli) removeCLIHeader(response).parseJson.asJsObject else response.parseJson.asJsObject
```
dubeejw: Seems like you can just call `removeCLIHeader()` without the need for the `isCli` flag.

tests/src/test/scala/system/basic/WskActionTests.scala
```diff
@@ -267,16 +269,6 @@ class WskActionTests extends TestHelpers with WskTestHelpers with JsHelpers {
     }
   }
 
-  it should "reject an invoke with the wrong parameters set" in withAssetCleaner(wskprops) { (wp, assetHelper) =>
```
dubeejw: I see why you moved this test to WskBasicUsageTests as it tests the functionality of the CLI. However, is there a similar test for just the REST interface?

houshengbo: This test is testing the CLI, so it will be moved to the repo of cli to test CLI. We do not need to have a REST counterpart to test, because it is an cli only functionality as "invoke with the wrong parameters".

tests/build.gradle
houshengbo: Because we need to keep the Cli tests inside the repo: we have a stage with both of the cli and rest tests co-existing: WskCliAction and WskRestAction, for example.

If we do not exclude either Cli or Rest, we will end up running the tests twice for the same purpose, which will take longer time than the Travis limit. That is the reason we keep an exclusion here. When Cli tests are decided to remove completely, we can remove the exclusion.

tests/src/test/scala/common/WskTestHelpers.scala
houshengbo: Since the Cli tests will still exist for some time, we need to make sure Cli tests also pass.

dubeejw: Ah, I see. The newline from the REST call might be anywhere in the output.

tests/src/test/scala/system/basic/WskActionTests.scala
dubeejw: Does there exist a test for testing the REST interface with bad parameters though?

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -888,6 +888,16 @@ class WskBasicUsageTests extends TestHelpers with WskTestHelpers {
       "Cannot save action sequences")
   }
 
+  it should "reject an invoke with the wrong parameters set" in withAssetCleaner(wskprops) { (wp, assetHelper) =>
+    val fullQualifiedName = s"/${wskprops.namespace}/samples/helloWorld"
+    val payload = "bob"
+    val rr = wsk.cli(
+      Seq("action", "invoke", fullQualifiedName, payload) ++ wskprops.overrides,
+      expectedExitCode = TestUtils.ERROR_EXIT)
```
dubeejw: Don't need to explicitly type `TestUtils.` here.

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -888,6 +888,16 @@ class WskBasicUsageTests extends TestHelpers with WskTestHelpers {
       "Cannot save action sequences")
   }
 
+  it should "reject an invoke with the wrong parameters set" in withAssetCleaner(wskprops) { (wp, assetHelper) =>
```
dubeejw: Actually this is a duplicate test and can be deleted. The same test occurs here https://github.com/apache/incubator-openwhisk/blob/master/tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala#L1554.

houshengbo: done

tests/build.gradle
```diff
@@ -24,6 +24,17 @@ task testLean(type: Test) {
...
+}
+
+task testLeanCli(type: Test) {
```
dubeejw: When is this task used?

PR #2623, opened 2017-08-16
WhiskProperties helper method to access whisk.api.localhost.name

mdeuser:


mdeuser:
pg2/1963

tests/src/test/scala/common/WhiskProperties.java
```diff
@@ -200,6 +200,8 @@ public static String getRouterHost() {
         return whiskProperties.getProperty("router.host");
     }
 
+    public static String getApiLocalHostName() { return whiskProperties.getProperty("whisk.api.localhost.name"); }
```
rabbah: if this is not used in open, note that you can always get a named property like this: `WhiskProperties.getProperty(name)`.

mdeuser:
no immediate use in open, so closing

PR #2625, opened 2017-08-16
Update CLI readme with cross platform information

akrabat:
Add information on how to build the CLI binaries for the other platforms.


akrabat:
As you implied, there's no point in having to ask the same question twice :)

PR #2626, opened 2017-08-16
Collapse multiple concurrent pulls of the same image.

markusthoemmes:
Multiple concurrent calls of `docker pull` to the same image create unnecessary network traffic. Instead, all of those calls are collapsed into the first calls. After that call finished, a subsequent pull will result in a `docker pull` again to enable updates to the same image, without having to work with tags for convenience.


markusthoemmes:
PG3 1970 🔵

PR #2628, opened 2017-08-16
Remove spray-caching dependency.

markusthoemmes:
Spray-caching is the caching module of the by us no longer used spray-http server. Our own implementation of the cache is there already anyway and keeping spray-caching binds us to Scala 2.11.


rabbah:
@cbickel apart from deleting the dead code, i presume this is independent of your cache work. yes?

common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -427,9 +427,15 @@ trait MultipleReadersSingleWriterCache[W, Winfo] {
...
  */
-private class ConcurrentMapBackedCache[V](store: ConcurrentMap[Any, Future[V]]) extends Cache[V] {
+private class ConcurrentMapBackedCache[V](store: ConcurrentMap[Any, Future[V]]) {
+    val cache = this
 
-    def get(key: Any) = Option(store.get(key))
+    def apply(key: Any) = new Keyed(key)
+
+    class Keyed(key: Any) {
+        def apply(magnet: ⇒ ValueMagnet[V])(implicit ec: ExecutionContext): Future[V] =
```
rabbah: convert to ascii?

PR #2631, opened 2017-08-17
Change controller cluster work mode configurable

ningyougang:
Currently, the default controller cluster work mode is `active/slave`.
if use `active/active` mode, can share more pressure on the request.
So i think it is necessary to change the controller cluster work mode configurable


ningyougang:
this PR may has relation with below PR:
https://github.com/apache/incubator-openwhisk/pull/2624

PR #2632, opened 2017-08-17
Use the stale parameter on listing activations

cbickel:
Use the parameter `stale=ok` on listing the activations.
This has the advantage, that activations are returned, even if the view in the activations db is not computed yet.

PG3#1016 is running.


common/scala/src/main/scala/whisk/core/database/ArtifactStore.scala
```diff
@@ -84,7 +84,7 @@ trait ArtifactStore[DocumentAbstraction] {
      * @param transid the transaction id for logging
      * @return a future that completes with List[JsObject] of all documents from view between start and end key (list may be empty)
      */
-    protected[core] def query(table: String, startKey: List[Any], endKey: List[Any], skip: Int, limit: Int, includeDocs: Boolean, descending: Boolean, reduce: Boolean)(
+    protected[core] def query(table: String, startKey: List[Any], endKey: List[Any], skip: Int, limit: Int, includeDocs: Boolean, descending: Boolean, reduce: Boolean, stale: Option[String])(
         implicit transid: TransactionId): Future[List[JsObject]]
```
rabbah: Could be Boolean and serialize to ok where it bottoms out into the app call.

sjfink:
`"slale" : "update_after"`  would be better I think

cbickel:
PG4#711 is 🔵

common/scala/src/main/scala/whisk/core/database/CouchDbRestClient.scala
```diff
@@ -174,6 +174,7 @@ class CouchDbRestClient(protocol: String, host: String, port: Int, username: Str
         endKey: List[Any] = Nil,
         skip: Option[Int] = None,
         limit: Option[Int] = None,
+        stale: Option[String] = None,
```
markusthoemmes: Shouldn't this be `StaleParameter`?

common/scala/src/main/scala/whisk/core/database/ArtifactStore.scala
```diff
@@ -28,6 +28,11 @@ import whisk.common.Logging
 import whisk.common.TransactionId
 import whisk.core.entity.DocInfo
 
+abstract class StaleParameter(val value: Option[String])
+case object StaleOk extends StaleParameter(Some("ok"))
+case object StaleUpdateAfter extends StaleParameter(Some("update_after"))
+case object NoStale extends StaleParameter(None)
```
markusthoemmes: Awesome!!

One nit: You can wrap the implementations into a companion object like:

```scala
object StaleParameter {
  case object Ok extends StaleParameter(Some("ok"))
  case object UpdateAfter extends StaleParameter(Some("update_after"))
  case object None extends StaleParameter(None)
}
```

I think that makes scoping a bit easier and its best practice.

common/scala/src/main/scala/whisk/core/entity/WhiskStore.scala
```diff
@@ -243,11 +249,12 @@ object WhiskEntityQueries {
         limit: Int,
         since: Option[Instant] = None,
         upto: Option[Instant] = None,
+        stale: StaleParameter,
```
markusthoemmes: Should this have a default value?

common/scala/src/main/scala/whisk/core/entity/WhiskStore.scala
```diff
@@ -259,11 +266,12 @@ object WhiskEntityQueries {
...
```
markusthoemmes: Should this have a default value?

common/scala/src/main/scala/whisk/core/entity/WhiskStore.scala
```diff
@@ -337,9 +348,10 @@ trait WhiskEntityQueries[T] {
         limit: Int,
         docs: Boolean = false,
         since: Option[Instant] = None,
-        upto: Option[Instant] = None)(
+        upto: Option[Instant] = None,
+        stale: StaleParameter)(
```
markusthoemmes: Default param?

markusthoemmes:
A few nits. Travis will suffice for the fixes. Very well done 👍

common/scala/src/main/scala/whisk/core/database/ArtifactStore.scala
```diff
@@ -28,6 +28,14 @@ import whisk.common.Logging
 import whisk.common.TransactionId
 import whisk.core.entity.DocInfo
 
+abstract class StaleParameter(val value: Option[String])
+
```
rabbah: Could have sealed this.

common/scala/src/main/scala/whisk/core/entity/WhiskStore.scala
```diff
@@ -205,12 +208,13 @@ object WhiskEntityQueries {
...
         namespace: EntityPath,
-        includeDocs: Boolean)(
+        includeDocs: Boolean,
+        stale: StaleParameter = StaleParameter.No)(
             implicit transid: TransactionId): Future[Map[String, List[JsObject]]] = {
         implicit val ec = db.executionContext
         val startKey = List(namespace.toString)
         val endKey = List(namespace.toString, TOP)
-        db.query(viewname(ENTITIES), startKey, endKey, 0, 0, includeDocs, descending = true, reduce = false) map {
+        db.query(viewname(ENTITIES), startKey, endKey, 0, 0, includeDocs, descending = true, reduce = false, stale = stale) map {
```
rabbah: I would have used just stale instead of stale = stale; since the name is indicative vs just a Boolean.

common/scala/src/main/scala/whisk/core/database/ArtifactStore.scala
markusthoemmes: Ugh, right. Should've looked up the whole pattern. Only relevant if you want to match against them and have the compiler check for an incomplete match.

PR #2641, opened 2017-08-17
Install composer in its own layer

akrabat:
Composer installation seems to fail sometimes in the CI build
environment. Move the installation of composer to a separate layer and
separate the download of the file via curl from the installation
process. Finally, run `composer --version` to ensure that it has been
installed correctly.

### DIFF
--- a/core/php7.1Action/Dockerfile
+++ b/core/php7.1Action/Dockerfile
@@ -27,12 +27,12 @@ RUN \
        bcmath \
        zip \
        gd \
-       soap \
+       soap
 
-   && \
-
-    # install Composer
-    cd /tmp && curl -sS https://getcomposer.org/installer | php  -- --install-dir=/usr/bin --filename=composer
+# install composer
+RUN curl -s -f -L -o /tmp/installer.php https://getcomposer.org/installer \
+    && php /tmp/installer.php --no-ansi --install-dir=/usr/bin --filename=composer \
+    && composer --ansi --version --no-interaction
 
 # create src directory to store action files
 RUN mkdir -p /action/src


csantanapr:
The issue is hard to explain when docker tries to use a cache layer, for some reason is not finding composer.
The only think I can think is that that there was a previous run that it didn't exit with error for the `| php ...`

The error looks like this:
```
Step 2/9 : RUN apk update && apk upgrade &&    apk add        postgresql-dev        icu        icu-libs        icu-dev        freetype-dev        libjpeg-turbo-dev        libpng-dev        libxml2-dev    &&    docker-php-ext-install        opcache        mysqli        pdo_mysql        pdo_pgsql        intl        bcmath        zip        gd        soap    &&     cd /tmp && curl -sS https://getcomposer.org/installer | php  -- --install-dir=/usr/bin --filename=composer
 ---> Using cache
 ---> 1612c7c99f05
Step 3/9 : RUN mkdir -p /action/src
 ---> Using cache
 ---> 2a8b4bd4b6c8
Step 4/9 : COPY composer.json /action
 ---> Using cache
 ---> d4459dc8fd95
Step 5/9 : RUN cd /action && /usr/bin/composer install --no-plugins --no-scripts --prefer-dist --no-dev -o && rm composer.lock
 ---> Running in 856b4de3df61
[91m/bin/sh: /usr/bin/composer: not found
The command '/bin/sh -c cd /action && /usr/bin/composer install --no-plugins --no-scripts --prefer-dist --no-dev -o && rm composer.lock' returned a non-zero code: 127
[0m
...
```

PR #2642, opened 2017-08-17
Update webhook.

rabbah:


rabbah:
@csantanapr confirmed this is working as desired.

PR #2643, opened 2017-08-17
Pretty print webaction JSON response.

### LINKED ISSUE #2635: webaction json response no longer pretty printed
```
> curl -k https://guest.openwhisk/default/echo.json
{"__ow_method":"get","__ow_headers":{"accept":"*/*","user-agent":"curl/7.43.0","host":"controllers","timeout-access":"<function1>","connection":"close"},"__ow_path":""}
```

@dubeejw

rabbah:
Fixes #2635.


csantanapr:
cc @dubeejw

rabbah:
@dubeejw we can combine this with your pr if e want to restore full pretty printing of json responses. I would actually do something slightly different in that case.

PR #2645, opened 2017-08-18
Restore Pretty Printing in Controller

dubeejw:
Relate to https://github.com/apache/incubator-openwhisk/issues/2635

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Actions.scala
@@ -76,6 +76,8 @@ trait WhiskActionsApi
     with ReferencedEntities {
     services: WhiskServices =>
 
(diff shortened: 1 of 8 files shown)


rabbah:
I was considering whether we wanted to restore his for the main API vs webactions. Which are more likely used with curl.

dubeejw:
@rabbah, I'd rather keep the pervious behavior. Any suggestions on testing that doesn't require testing every single path?

markusthoemmes:
Why would we need pretty printing either way? Isn't it more data in transport?

rabbah:
For webactions I think it's a useful convenience because one tends to use curl often - at least I do and we use these examples in our docs. 

For the mainline API I think it's less important because i think generally one interacts with that via the cli which pretty prints.

rabbah:
covered by https://github.com/apache/incubator-openwhisk/pull/2643.

PR #2646, opened 2017-08-18
Content type header improvements

### LINKED ISSUE #2630: JSON encoding is inconsistent for web actions
There's something very odd about JSON encoding and content types with web actions.

```php
<?php
{
    return [
        "body" => ["greeting" => "Hello world"],
        "headers" => ["Content-Type" => "application/json"],
    ];
}
```
gives:
```
$ curl -ki https://192.168.33.13/api/v1/web/guest/default/hello
HTTP/1.1 200 OK

{"greeting":"Hello world"}
```

All good.

However, for:

```php
<?php
{
    return [
        "body" => ["greeting" => "Hello world"],
        "headers" => ["Content-Type" => "application/json-patch+json"],
    ];
}
```
I get:
```
$ curl -ki https://192.168.33.13/api/v1/web/guest/default/hello
HTTP/1.1 400 Bad Request

{"error":"Response type in header did not match generated content type.","code":48}
```
Complete failure. Maybe it's because the new "no-need-to-base64-a-json-string" code is hard coded to only support the `application/json` content type?

However, if I base64 encode & stringify the JSON:

```php
<?php
{
    return [
        "body" => base64_encode(json_encode(["greeting" => "Hello world"])),
        "headers" => ["Content-Type" => "application/json-patch+json"],
    ];
}
```
then this happens:
```
$ curl -ki https://192.168.33.13/api/v1/web/guest/default/hello
HTTP/1.1 200 OK
Content-Type: application/json-patch+json

eyJncmVldGluZyI6IkhlbGxvIHdvcmxkIn0=
```
Content type is right, but the body isn't decoded!

My final test was using `application/problem+json` and this does something different again:

```php
<?php
{
    return [
        "body" => base64_encode(json_encode(["greeting" => "Hello world"])),
        "headers" => ["Content-Type" => "application/problem+json"],
    ];
}
```
results in:
```
$ curl -ki https://192.168.33.13/api/v1/web/guest/default/hello
HTTP/1.1 400 Bad Request

{"error":"Response did not specify a known content-type.","code":54}
```

It hasn't worked at all! Obviously `application/problem+json` is a valid content type as it's defined in [RFC7807](https://tools.ietf.org/html/rfc7807), so my guess is that it's not on the right list…  Clearly, we can't have a whitelist for content types.

As I say, something's odd here.

akrabat:
Fix the following issues with JSON content types in web actions:

1. base64 encoded bodies with a content-type of `application/patch-json+json` should work.
1. JSON dictionary bodies with a content-type of `application/patch-json+json` should work.
1. base64 encoded bodies with a content-type of `application/hal+json` should work.
1. JSON dictionary bodies with a content-type of `application/hal+json` should work.

I chose these two media types as one is "known" by Akka and one is "unknown" as the behaviour is currently different for each one. 

In practice, all media types that are of the correct format should be acceptable, so this PR fixes that which means that technically, we don't need both sets of unit tests. However, I used both sets to prove that the fix worked when developing it.

Fixes #2630

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
@@ -264,8 +264,9 @@ protected[core] object WhiskWebActionsApi extends Directives {
   }
 
   /**
-   * Finds the content-type in the header list and maps it to a known media type. If it is not
-   * recognized, construct a failure with appropriate message.
+   * Finds the content-type in the header list and ensures that it is a valid format. If it is not
+   * valid, construct a failure with appropriate message.
+   * If the content-type header is missing, then return the supplied defaultType
    */
   private def findContentTypeInHeader(headers: List[RawHeader],
                                       transid: TransactionId,
@@ -273,16 +274,15 @@ protected[core] object WhiskWebActionsApi extends Directives {
     headers.find(_.lowercaseName == `Content-Type`.lowercaseName) match {
       case Some(header) =>
         MediaType.parse(header.value) match {
-          case Right(mediaType: MediaType) =>
-            // lookup the media type specified in the content header to see if it is a recognized type
-            MediaTypes.getForKey(mediaType.mainType -> mediaType.subType).map(Success(_)).getOrElse {
-              // this is a content-type that is not recognized, reject it
-              Failure(RejectRequest(BadRequest, Messages.httpUnknownContentType)(transid))
-            }
-          case _ => Failure(RejectRequest(BadRequest, Messages.httpUnknownContentType)(transid))
+          case Right(mediaType: MediaType) => Success(mediaType)
+          case _                           => Failure(RejectRequest(BadRequest, Messages.httpUnknownContentType)(transid))
         }
       case None => Success(defaultType)
     }
+  }
+
+  private def isJsonFamily(mt: MediaType) = {
+    mt == `application/json` || mt.value.endsWith("+json")
   }
 
   private def interpretHttpResponseAsJson(code: StatusCode,
@@ -290,9 +290,22 @@ protected[core] object WhiskWebActionsApi extends Directives {
                                           js: JsValue,
                                           transid: TransactionId) = {
     findContentTypeInHeader(headers, transid, `application/json`) match {
-      case Success(mediaType) if (mediaType == `application/json`) =>
+      // use the default akka-http response marshaler for standard application/json
+      case Success(mediaType) if mediaType == `application/json` =>
         respondWithHeaders(removeContentTypeHeader(headers)) {
           complete(code, js)
+        }
+
+      // for all other json-family content-type, explicitly marshal the response;
+      // the order of the case statement matters; isJsonFamily returns true for application/json
+      case Success(mediaType) if isJsonFamily(mediaType) =>
+        respondWithHeaders(removeContentTypeHeader(headers)) {
+          complete(
+            code,
+            HttpEntity(
+              ContentType(
+                MediaType.customWithFixedCharset(mediaType.mainType, mediaType.subType, HttpCharsets.`UTF-8`)),
+              js.prettyPrint))
         }
 
       case _ =>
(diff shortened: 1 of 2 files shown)


core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -265,30 +265,33 @@ protected[core] object WhiskWebActionsApi extends Directives {
...
         headers.find(_.lowercaseName == `Content-Type`.lowercaseName) match {
             case Some(header) =>
                 MediaType.parse(header.value) match {
-                    case Right(mediaType: MediaType) =>
-                        // lookup the media type specified in the content header to see if it is a recognized type
-                        MediaTypes.getForKey(mediaType.mainType -> mediaType.subType).map(Success(_)).getOrElse {
-                            // this is a content-type that is not recognized, reject it
-                            Failure(RejectRequest(BadRequest, Messages.httpUnknownContentType)(transid))
-                        }
+                    case Right(mediaType: MediaType) => Success(mediaType) // mediaType is a valid format
```
dubeejw: @akrabat, please check to make sure this is correct even when an invalid media type is passed. If I recall correctly, there was a reason we were using `getForKey`.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -265,30 +265,33 @@ protected[core] object WhiskWebActionsApi extends Directives {
...
             case None => Success(defaultType)
         }
     }
 
     private def interpretHttpResponseAsJson(code: StatusCode, headers: List[RawHeader], js: JsValue, transid: TransactionId) = {
-        findContentTypeInHeader(headers, transid, `application/json`) match {
-            case Success(mediaType) if (mediaType == `application/json`) =>
+        findContentTypeInHeader(headers, transid, `application/json`).flatMap { mediaType =>
+            // A response is JSON if it's content type is "application/json" or ends in "+json"
+            if (MediaType.unapply(mediaType).getOrElse("").endsWith("+json") || mediaType == `application/json`) {
```
akrabat: I wasn't sure if I could use `mediaType.value` here as I couldn't work out if it's guaranteed to be a string. I'm pretty sure that `MediaType.unapply(mediaType).getOrElse("")` is safer though may be overkill?

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -890,7 +998,7 @@ trait WebActionsApiTests extends ControllerTestCommon with BeforeAndAfterEach wi
                         invocationsAllowed += 1
                         actionResult = Some(JsObject(
                             "headers" -> JsObject(
-                                "content-type" -> "xyz/bar".toJson),
+                                "content-type" -> "xyzbar".toJson),
```
dubeejw: Why the removal of the sub type here?

dubeejw: Is this related to the removal of `getForKey` above?

akrabat: > Why the removal of the sub type here?

I needed an invalid media type an this was the simplest one I could think of.

akrabat: > Is this related to the removal of getForKey above?

yes - `getForKey()` tests that the media type is known to Akka-Http which isn't very useful there are a lot of media types are not in their list. Hence I changed `findContentTypeInHeader` to only do the `parse()` which ensures that it the format is valid, but not do the check against the list Akka-Http knows about.

dubeejw: @akrabat, another way would be to create list of custom media types and content types for what you need.

Something like this for instance:
```
val jsonPatchMediaType = MediaType.applicationWithOpenCharset("application/json-patch+json") 
val jsonPatchContentType = jsonPatchMediaType withCharset HttpCharsets.`UTF-8` 
```

dubeejw: @rabbah, what do you think about the above?

akrabat: > another way would be to create list of custom media types and content types for what you need.

@dubeejw How would that work from a web action? The number of media types that can be used in the Content-Type header is essentially infinite.

For example, I may want to create `application/vnd.akrabat.bookshelf.v1+json` for a given project. This clearly won't be defined by OpenWhisk and it shouldn't prevent me from sending a content-type header with that media type.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
rabbah: does `mediaType.subType.endsWith("+json")` work?

dubeejw:
@akrabat, looks like you need to rebase this one.

akrabat:
Yes. The world changed under me  :) 

I've been a bit more snowed under than I realised, so haven't had a chance to sort it out. A quick check showed quite a bit of change required.

rabbah:
@akrabat I rebased your branch to master here https://github.com/apache/incubator-openwhisk/compare/master...rabbah:pr-2646?expand=1

This is a first cut to get your changes in - I need to look at this once more.

akrabat:
Superceded by https://github.com/apache/incubator-openwhisk/pull/2717.

akrabat:
Reopening as git get confused with others rebasing work.

rabbah:
PG2/2049 🔵

PR #2648, opened 2017-08-18
Overhaul logging

markusthoemmes:
**Note: Super duper WIP**

Profiling told me that the way we log today is quite a hotstop in the components. This is an attempt to adapt best practices in terms of logging, especially around String allocations.

- Using logback instead of log4j (It's supposed to be its successor, and its recommended by akka)
- Delaying `String` building until it is really needed.
- Will audit for log messages to be needed/at the right loglevel.

Opening early to gather feedback.


tysonnorris:
I've been confused by the need to use the whisk.common.Logging class at all - is there any reason not to use slf4j classes directly, like many projects do? e.g. instead of every class having to pass Logging around, use something like ```val logger: Logger = LoggerFactory.getLogger(this.getClass)``` (or one per package, instead of one per class, etc). This way you can a) get rid of Logging code in the codebase and b) enable logger customization for specific packages/classes using standard conventions (via logback config)

common/scala/src/main/resources/logback.xml
```diff
@@ -0,0 +1,16 @@
+<configuration>
+  <appender name="console" class="ch.qos.logback.core.ConsoleAppender">
```
tysonnorris: Recommend an AsyncAppender here; from the docs https://logback.qos.ch/manual/appenders.html
> This strategy has an amazingly favorable effect on performance at the cost of event loss.

(and its worth adding a comment to the config to advise on the lossy behavior when discarding TRACE/DEBUG/INFO events when writing to stream is backed up to 80% of queue size)

markusthoemmes:
@tysonnorris I looked into using plain slf4j loggers but using MDC seems to be a kinda non-starter because we don't specifically take control on a per Thread basis and slf4j's documentation says, that MDC should not be reset a bazillion times. I don't have much experience here though, if you have an idea on how to solve that I'd be very glad because as you said, going standard would be very nice!

common/scala/src/main/resources/logback.xml
markusthoemmes: Shouldn't give us any performance gain really, because our logging is already akka based and as such asynchronous from the code itself.

rabbah:
> I've been confused by the need to use the whisk.common.Logging class at all 

@tysonnorris primarily to make it easier to thread the transaction id throughout the system.

markusthoemmes:
@mhenke1 is doing all this work.

PR #2649, opened 2017-08-18
remove API GW V1 group vars

drcariel:
Should I be updating all api_gw_v2 vars to exclude v2 now as well? aka api_gw


drcariel:
PG4~705

drcariel:
PG4~707

PR #2650, opened 2017-08-18
Apply standard scala formatting.

markusthoemmes:
Formats all .scala files according to `scalafmt`'s (opinionated) style. Travis checks for correcly formatted code.

Let's discuss which style we want to adapt and enforce and then forget about misformatted code forever!

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,0 +1,6 @@
+style = intellij
+danglingParentheses = false
+maxColumn = 120
+docstrings = JavaDoc
+rewrite.rules = [SortImports]
+project.git = true
--- a/common/scala/src/main/scala/whisk/common/Config.scala
+++ b/common/scala/src/main/scala/whisk/common/Config.scala
@@ -42,113 +42,116 @@ import scala.util.Try
  * @param optionalProperties a Set of optional properties which may or not be defined.
  * @param env an optional environment to read from (defaults to sys.env).
  */
-class Config(
-    requiredProperties: Map[String, String],
-    optionalProperties: Set[String] = Set())(
-        env: Map[String, String] = sys.env)(
-            implicit logging: Logging) {
+class Config(requiredProperties: Map[String, String], optionalProperties: Set[String] = Set())(
+  env: Map[String, String] = sys.env)(implicit logging: Logging) {
 
-    private val settings = getProperties().toMap.filter {
-        case (k, v) => requiredProperties.contains(k) ||
-            (optionalProperties.contains(k) && v != null)
+  private val settings = getProperties().toMap.filter {
+    case (k, v) =>
+      requiredProperties.contains(k) ||
+        (optionalProperties.contains(k) && v != null)
+  }
+
+  lazy val isValid: Boolean = Config.validateProperties(requiredProperties, settings)
+
+  /**
+   * Gets value for key if it exists else the empty string.
+   * The value of the override key will instead be returned if its value is present in the map.
+   *
+   * @param key to lookup
+   * @param overrideKey the property whose value will be returned if the map contains the override key.
+   * @return value for the key or the empty string if the key does not have a value/does not exist
+   */
+  def apply(key: String, overrideKey: String = ""): String = {
+    Try(settings(overrideKey)).orElse(Try(settings(key))).getOrElse("")
+  }
+
+  /**
+   * Returns the value of a given key.
+   *
+   * @param key the property that has to be returned.
+   */
+  def getProperty(key: String): String = {
+    this(key)
+  }
+
+  /**
+   * Returns the value of a given key parsed as a double.
+   * If parsing fails, return the default value.
+   *
+   * @param key the property that has to be returned.
+   */
+  def getAsDouble(key: String, defaultValue: Double): Double = {
+    Try { getProperty(key).toDouble } getOrElse { defaultValue }
+  }
+
+  /**
+   * Returns the value of a given key parsed as an integer.
+   * If parsing fails, return the default value.
+   *
+   * @param key the property that has to be returned.
+   */
+  def getAsInt(key: String, defaultValue: Int): Int = {
+    Try { getProperty(key).toInt } getOrElse { defaultValue }
+  }
+
+  /**
+   * Converts the set of property to a string for debugging.
+   */
+  def mkString: String = settings.mkString("\n")
+
+  /**
+   * Loads the properties from the environment into a mutable map.
+   *
+   * @return a pair which is the Map defining the properties, and a boolean indicating whether validation succeeded.
+   */
+  protected def getProperties(): scala.collection.mutable.Map[String, String] = {
+    val required = scala.collection.mutable.Map[String, String]() ++= requiredProperties
+    Config.readPropertiesFromEnvironment(required, env)
+
+    // for optional value, assign them a default from the required properties list
+    // to prevent loss of a default value on a required property that may not otherwise be defined
+    val optional = scala.collection.mutable.Map[String, String]() ++= optionalProperties.map { k =>
+      k -> required.getOrElse(k, null)
     }
+    Config.readPropertiesFromEnvironment(optional, env)
 
-    lazy val isValid: Boolean = Config.validateProperties(requiredProperties, settings)
-
-    /**
-     * Gets value for key if it exists else the empty string.
-     * The value of the override key will instead be returned if its value is present in the map.
-     *
-     * @param key to lookup
-     * @param overrideKey the property whose value will be returned if the map contains the override key.
-     * @return value for the key or the empty string if the key does not have a value/does not exist
-     */
-    def apply(key: String, overrideKey: String = ""): String = {
-        Try(settings(overrideKey)).orElse(Try(settings(key))).getOrElse("")
-    }
-
-    /**
-     * Returns the value of a given key.
-     *
-     * @param key the property that has to be returned.
-     */
-    def getProperty(key: String): String = {
-        this(key)
-    }
-
-    /**
-     * Returns the value of a given key parsed as a double.
-     * If parsing fails, return the default value.
-     *
-     * @param key the property that has to be returned.
-     */
-    def getAsDouble(key: String, defaultValue: Double): Double = {
-        Try { getProperty(key).toDouble } getOrElse { defaultValue }
-    }
-
-    /**
-     * Returns the value of a given key parsed as an integer.
-     * If parsing fails, return the default value.
-     *
-     * @param key the property that has to be returned.
-     */
-    def getAsInt(key: String, defaultValue: Int): Int = {
-        Try { getProperty(key).toInt } getOrElse { defaultValue }
-    }
-
-    /**
-     * Converts the set of property to a string for debugging.
-     */
-    def mkString: String = settings.mkString("\n")
-
-    /**
-     * Loads the properties from the environment into a mutable map.
-     *
-     * @return a pair which is the Map defining the properties, and a boolean indicating whether validation succeeded.
-     */
-    protected def getProperties(): scala.collection.mutable.Map[String, String] = {
-        val required = scala.collection.mutable.Map[String, String]() ++= requiredProperties
-        Config.readPropertiesFromEnvironment(required, env)
-
-        // for optional value, assign them a default from the required properties list
-        // to prevent loss of a default value on a required property that may not otherwise be defined
-        val optional = scala.collection.mutable.Map[String, String]() ++= optionalProperties.map { k => k -> required.getOrElse(k, null) }
-        Config.readPropertiesFromEnvironment(optional, env)
-
-        required ++ optional
-    }
+    required ++ optional
+  }
 }
 
 /**
  * Singleton object which provides global methods to manage configuration.
  */
 object Config {
-    /**
-     * Reads a Map of key-value pairs from the environment -- store them in the
-     * mutable properties object.
-     */
-    def readPropertiesFromEnvironment(properties: scala.collection.mutable.Map[String, String], env: Map[String, String])(implicit logging: Logging) = {
-        for (p <- properties.keys) {
-            val envp = p.replace('.', '_').toUpperCase
-            val envv = env.get(envp)
-            if (envv.isDefined) {
-                logging.info(this, s"environment set value for $p")
-                properties += p -> envv.get.trim
-            }
-        }
-    }
 
-    /**
-     * Checks that the properties object defines all the required properties.
-     *
-     * @param required a key-value map where the keys are required properties
-     * @param properties a set of properties to check
-     */
-    def validateProperties(required: Map[String, String], properties: Map[String, String])(implicit logging: Logging): Boolean = {
-        required.keys.forall { key =>
-            val value = properties(key)
-            if (value == null) logging.error(this, s"required property $key still not set")
-            value != null
-        }
+  /**
+   * Reads a Map of key-value pairs from the environment -- store them in the
+   * mutable properties object.
+   */
+  def readPropertiesFromEnvironment(properties: scala.collection.mutable.Map[String, String], env: Map[String, String])(
+    implicit logging: Logging) = {
+    for (p <- properties.keys) {
+      val envp = p.replace('.', '_').toUpperCase
+      val envv = env.get(envp)
+      if (envv.isDefined) {
+        logging.info(this, s"environment set value for $p")
+        properties += p -> envv.get.trim
+      }
     }
+  }
+
+  /**
+   * Checks that the properties object defines all the required properties.
+   *
+   * @param required a key-value map where the keys are required properties
+   * @param properties a set of properties to check
+   */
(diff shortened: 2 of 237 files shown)


markusthoemmes:
@rabbah @cbickel @vvraskin @sven-lange-last @tysonnorris @ningyougang 

Could you give your thought on the formatting changes in this PR? You've been making most of the scala contributions until now, so you'll be impacted (if you feel impacted by this)

.scalafmt.conf
```diff
@@ -0,0 +1,2 @@
+maxColumn = 100
```
sven-lange-last: I just checked in my editor windows - 100 chars per line seems a little narrow. I would opt for 120 or even 130.

common/scala/src/main/scala/whisk/common/Config.scala
```diff
@@ -42,113 +42,120 @@ import scala.util.Try
...
-    /**
-     * Returns the value of a given key.
-     *
-     * @param key the property that has to be returned.
-     */
-    def getProperty(key: String): String = {
-        this(key)
-    }
+  lazy val isValid: Boolean =
+    Config.validateProperties(requiredProperties, settings)
```
sven-lange-last: This line seems to break a little early - statement readability would be improved in a single line.

sven-lange-last:
Travis checks whether all code complies with code formatting. If not, the build will fail - can you confirm?

If yes, it would be pretty inconvenient to wait for a Travis build just to learn that code formatting rules need to be applied. For this reason, we need something like a pre-commit hook that runs code formatting automatically including a description how to set up.

common/scala/src/main/scala/whisk/common/Config.scala
```diff
@@ -42,113 +42,120 @@ import scala.util.Try
...
-    optionalProperties: Set[String] = Set())(
-        env: Map[String, String] = sys.env)(
-            implicit logging: Logging) {
-
-    private val settings = getProperties().toMap.filter {
-        case (k, v) => requiredProperties.contains(k) ||
-            (optionalProperties.contains(k) && v != null)
-    }
+class Config(requiredProperties: Map[String, String], optionalProperties: Set[String] = Set())(
+    env: Map[String, String] = sys.env)(implicit logging: Logging) {
```
tysonnorris: seems like things are at 2 space indent - can wrapping be the same?

common/scala/src/main/scala/whisk/common/Config.scala
tysonnorris: +1

common/scala/src/main/scala/whisk/common/Config.scala
```diff
@@ -42,113 +42,120 @@ import scala.util.Try
...
+   */
+  protected def getProperties(): scala.collection.mutable.Map[String, String] = {
+    val required = scala.collection.mutable
+      .Map[String, String]() ++= requiredProperties
+    Config.readPropertiesFromEnvironment(required, env)
+
+    // for optional value, assign them a default from the required properties list
+    // to prevent loss of a default value on a required property that may not otherwise be defined
+    val optional = scala.collection.mutable
+      .Map[String, String]() ++= optionalProperties.map { k =>
```
tysonnorris: another early wrapping?

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -25,7 +25,7 @@ import java.util.concurrent.atomic.AtomicInteger
 import scala.math.BigDecimal.int2bigDecimal
 import scala.util.Try
 
-import akka.event.Logging.{ InfoLevel, WarningLevel }
+import akka.event.Logging.{InfoLevel, WarningLevel}
```
tysonnorris: Is there a way to force these imports to be separate? I'm not sure how others feel, but I like to have imports not use wildcards or have multiples on same line.

common/scala/src/main/scala/whisk/core/connector/MessageConsumer.scala
```diff
@@ -92,140 +92,151 @@ object MessageFeed {
...
-            false
-        }
+  }
+
+  private def shouldFillQueue(): Boolean = {
+    val occupancy = outstandingMessages.size
+    if (occupancy <= pipelineFillThreshold) {
+      logging.debug(
+        this,
+        s"$description pipeline has capacity: $occupancy <= $pipelineFillThreshold ($handlerCapacity)")
```
tysonnorris: is it possible to wrap only the last arg, instead of causing all args to appear wrapped? like:
```
logging.debug(this, 
  s"$description pipeline has capacity: $occupancy <= $pipelineFillThreshold ($handlerCapacity)")
```

common/scala/src/main/scala/whisk/common/Config.scala
```diff
@@ -42,113 +42,120 @@ import scala.util.Try
...
+   */
+  def validateProperties(required: Map[String, String], properties: Map[String, String])(
+      implicit logging: Logging): Boolean = {
+    required.keys.forall { key =>
+      val value = properties(key)
+      if (value == null)
+        logging.error(this, s"required property $key still not set")
+      value != null
     }
+  }
```
tysonnorris: FWIW I got a slightly different wrapping when trying the scalafmt plugin in intellij, using the same .scalafmt.conf:
```
  def validateProperties(required: Map[String, String], properties: Map[String, String])(
      implicit logging: Logging): Boolean = {
    required.keys.forall { key =>
      val value = properties(key)
      if (value == null) logging.error(this, s"required property $key still not set")
      value != null
    }
  }
```

common/scala/src/main/scala/whisk/common/TransactionId.scala
markusthoemmes: Hm, I'd argue against this. Wildcards are very handy especially in case of implicit imports.

common/scala/src/main/scala/whisk/common/Config.scala
markusthoemmes: For wrapping this is quite handy to distinguish the class body from the class header. I'd keep that as is (it's the standard config)

markusthoemmes:
@sven-lange-last @tysonnorris seems like the early wrappings are solved by the longer line length.

common/scala/src/main/scala/whisk/common/Config.scala
```diff
@@ -42,113 +42,116 @@ import scala.util.Try
...
-     * The value of the override key will instead be returned if its value is present in the map.
-     *
-     * @param key to lookup
-     * @param overrideKey the property whose value will be returned if the map contains the override key.
-     * @return value for the key or the empty string if the key does not have a value/does not exist
-     */
-    def apply(key: String, overrideKey: String = ""): String = {
-        Try(settings(overrideKey)).orElse(Try(settings(key))).getOrElse("")
-    }
+  private val settings = getProperties().toMap.filter {
```
rabbah: so we're going to 2 space indent instead of 4? I thought on the call you mentioned 4.

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -132,115 +135,115 @@ class PrintStreamLogging(outputStream: PrintStream = Console.out) extends Loggin
...
+    else simpleName
+  }
 }
 
 private object Emitter {
-    val timeFormat = DateTimeFormatter.
-        ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'").
-        withZone(ZoneId.of("UTC"))
+  val timeFormat =
+    DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'").withZone(ZoneId.of("UTC"))
```
rabbah: blugh - can we make that add curlies?

common/scala/src/main/scala/whisk/common/Scheduler.scala
```diff
@@ -33,99 +33,93 @@ import akka.actor.Props
...
+   *
+   * @param interval the time to wait between two runs
+   * @param alwaysWait always wait for the given amount of time or calculate elapsed time to wait
+   * @param closure the closure to be executed
+   */
+  private class Worker(initialDelay: FiniteDuration,
+                       interval: FiniteDuration,
+                       alwaysWait: Boolean,
+                       name: String,
+                       closure: () => Future[Any])(implicit logging: Logging, transid: TransactionId)
```
rabbah: can we make implicit break?
if this is a longer line, what'll look like?

common/scala/src/main/scala/whisk/common/Scheduler.scala
```diff
@@ -33,99 +33,93 @@ import akka.actor.Props
...
+   * @param interval the time to wait at most between two runs of the closure
+   * @param initialDelay optionally delay the first scheduled iteration by given duration
+   * @param f the function to run
+   */
+  def scheduleWaitAtMost(interval: FiniteDuration,
+                         initialDelay: FiniteDuration = Duration.Zero,
+                         name: String = "Scheduler")(f: () => Future[Any])(implicit system: ActorSystem,
+                                                                           logging: Logging,
+                                                                           transid: TransactionId =
+                                                                             TransactionId.unknown) = {
```
rabbah: I see, it'll look hideous :D

rabbah: good reason to stop using implicits :)

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -39,90 +39,107 @@ import whisk.core.entity.InstanceId
...
+   * @param message An additional message that is written into the log, together with the other information.
+   * @param logLevel The Loglevel, the message should have. Default is <code>InfoLevel</code>.
+   * @param endTime Manually set the timestamp of the end. By default it is NOW.
+   */
+  def finished(from: AnyRef,
+               startMarker: StartMarker,
+               message: String = "",
+               logLevel: LogLevel = InfoLevel,
+               endTime: Instant = Instant.now(Clock.systemUTC))(implicit logging: Logging) = {
+    val endMarker =
```
rabbah: curlies around these might help... if it's doable.

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -39,90 +39,107 @@ import whisk.core.entity.InstanceId
...
+  def deltaToStart = Duration.between(meta.start, Instant.now(Clock.systemUTC)).toMillis
+
+  /**
+   * Calculates the time between now and the startMarker that was returned by <code>starting</code>.
+   *
+   * @param startMarker <code>StartMarker</code> returned by a <code>starting</code> method.
+   * @param endTime Manually set the endtime. By default it is NOW.
+   */
+  def deltaToMarker(startMarker: StartMarker, endTime: Instant = Instant.now(Clock.systemUTC)) =
+    Duration.between(startMarker.start, endTime).toMillis
```
rabbah: comparing 140 to 131, i like 140 better for example.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -34,189 +34,195 @@ import whisk.common.Logging
...
+  val dockerPort = this(WhiskConfig.dockerPort)
+
+  val dockerImagePrefix = this(WhiskConfig.dockerImagePrefix)
+  val dockerImageTag = this(WhiskConfig.dockerImageTag)
+
+  val invokerContainerNetwork = this(WhiskConfig.invokerContainerNetwork)
+  val invokerContainerPolicy =
+    if (this(WhiskConfig.invokerContainerPolicy) == "") None
+    else Some(this(WhiskConfig.invokerContainerPolicy))
+  val invokerContainerDns =
```
rabbah: more {} preference

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -34,189 +34,195 @@ import whisk.common.Logging
...
+    if (this(WhiskConfig.invokerContainerDns) == "") Seq()
+    else this(WhiskConfig.invokerContainerDns).split(" ").toSeq
+  val invokerNumCore = this(WhiskConfig.invokerNumCore)
+  val invokerCoreShare = this(WhiskConfig.invokerCoreShare)
+
+  val wskApiHost = this(WhiskConfig.wskApiProtocol) + "://" + this(WhiskConfig.wskApiHostname) + ":" + this(
```
rabbah: eh

common/scala/src/main/scala/whisk/core/connector/Message.scala
```diff
@@ -31,89 +31,89 @@ import whisk.core.entity.WhiskActivation
...
-    cause: Option[ActivationId] = None)
+case class ActivationMessage(override val transid: TransactionId,
+                             action: FullyQualifiedEntityName,
+                             revision: DocRevision,
+                             user: Identity,
+                             activationId: ActivationId,
+                             activationNamespace: EntityPath,
+                             rootControllerIndex: InstanceId,
+                             content: Option[JsObject],
+                             cause: Option[ActivationId] = None)
```
rabbah: visually i like the old style better.

common/scala/src/main/scala/whisk/core/connector/MessageConsumer.scala
```diff
@@ -92,140 +92,149 @@ object MessageFeed {
...
-        } else {
-            logging.debug(this, s"$description pipeline must drain: $occupancy > $pipelineFillThreshold")
-            false
-        }
+  }
+
+  private def shouldFillQueue(): Boolean = {
+    val occupancy = outstandingMessages.size
+    if (occupancy <= pipelineFillThreshold) {
+      logging.debug(this,
```
rabbah: 180 characters? :D

common/scala/src/main/scala/whisk/common/Config.scala
akrabat: 2 space matches the [style guide](https://docs.scala-lang.org/style/indentation.html). Generally I'm in favour of following the dominant style for a given language (if there is one) as it reduces the length of discussions.

docs/dev/hooks.md
```diff
@@ -0,0 +1,23 @@
+# Git pre-commit hooks
+
+To minimize review overhead OpenWhisk uses tools to enforce our coding standards. This is a set of git pre-commit hooks which will automatically check for formatting errors, so you don't need to wait for a CI build to fail to find those.
```
rabbah: To minimize review overhead OpenWhisk uses tools to enforce our coding standards. ->
OpenWhisk uses tools to enforce coding standards and reduce review overhead.

rabbah: , so you don't need to wait for a CI build to fail to find those. ->
. This will save you from wasting Travis CI builds to find formatting errors.

docs/dev/hooks.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Scala
+
+Scala is formatted using [scalafmt](http://scalameta.org/scalafmt/). It's configuration can be found [here](../.scalafmt.conf). To automatically reformat **changed** Scala files upon commit, add the following to your `pre-commit` hook.
```
rabbah: It's -> Its

rabbah: replace 'here' with actual name of the file?

docs/dev/hooks.md
```diff
@@ -0,0 +1,23 @@
...
+
+git diff --name-only --cached | grep '\.scala'
+if [ "$?" -eq 0 ]; then
+    # reformat scala code iff a scala file changed
+    $root/gradlew scalafmtAll
+fi
+
+# readd all staged files
+git add "$(git diff --name-only --cached)"
+```
```
rabbah: should we mention scan code also (and link to that commit hook)?

markusthoemmes:
@tysonnorris @sven-lange-last any last opinions from you?

I prematurely added documentation here: https://github.com/apache/incubator-openwhisk/wiki/Contributing:-Git-guidelines#scala

common/scala/src/main/scala/whisk/common/TransactionId.scala
tysonnorris: Ok, ignoring wildcards, WDYT about changing these to separate imports? I think the `ExpandImportSelectors` rewrite in scalafmt will do it (while leaving wildcards alone) http://scalameta.org/scalafmt/#ExpandImportSelectors. 

I think it will provide a good change to get imports formatted consistently (along with `SortedImported`).

tysonnorris:
Wrapping seems better, at least in the example I ran into before. 
I would favor adding to the config:
- expand imports: ```rewrite.rules = [SortImports,ExpandImportSelectors]```

markusthoemmes:
@tysonnorris SortImports and ExpandImportSelectors are incompatible (I tried it). I'm also in favor of keeping the grouping as is, although I'd prefer the formatter to auto-group, which it doesn't.

tysonnorris:
OK, good enough. Thanks

markusthoemmes:
PG2 2042 🔵

PR #2584, opened 2017-08-07
SPI for Loadbalancer

tysonnorris:
To allow deployments to customize the Loadbalancer impls. Currently same Loadbalancer is used for all traffic, but future enhancements could use multiple load balancers for different traffic patterns.

Default is whisk.core.loadBalancer.LoadBalancerServiceProvider, which wraps the existing LoadBalancer impl  (Kafka + InvokerPool) behind the SPI for instantiation.

To decouple up the SPI interface for LoadBalancer from implementation details, no ArtifactStore stores are passed as parameters into LoadBalancerService anymore. CouchDbStoreProvider caches stores by name, so they can be fetched via SPI in multiple places and resolve to the same store instance.  Now LoadBalancerService can do `WhiskEntityStore.datastore(config)` to acquire the entityStore.

PR #2656, opened 2017-08-21
ActivationExecutor

tysonnorris:
*Note this PR builds upon #2584*

To support routing loads to a particular execution system based on kind, this PR introduces:
* ActivationExecutor - an interface that defines the execution process
* KindBasedLoadBalancer - a LoadBalancer impl that routes kinds to the appropriate ActivationExecutor

Existing behavior remains (LoadBalancerService implements both LoadBalancer and ActivationExecutor), but enables additional ActivationExecutors to be live, and determines the appropriate executor by:
* priority (in case multiple are available for the same kind)
* fallback to LoadBalancerService (in case no additional executors are defined for a particular kind)

This will allow things like isolating concurrent-request-tolerant actions to a specific kind, or other types of isolation (long timeout vs short timeout, etc).


markusthoemmes:
Would you elaborate on why we need to differentiate between a load-balancer and an ActivationExector?

tysonnorris:
@markusthoemmes I expect there is some need for system wide management (load balancer), separate from individual execution paths (ActivationExecutor). e.g. ActivationThrottler applies to system wide activity, while ActivationExecutors may have their own independent notion of throttling.

markusthoemmes:
I still don't grasp the difference. Couldn't we define the overall system behavior as the behavior of all loadbalancers in the system combined?

tysonnorris:
That is possible, but I'm not sure it makes sense. Taking throttling as an example, it may or may not make sense to throttle the system specifically using the sum of all executors concurrent activations. I would say there should either be global throttling + executor throttling OR only executor throttling.  Since there is currently only global throttling, it seemed prudent to leave that in place, but it's possible to merge the loadbalancer and executor and remove global throttling instead.

markusthoemmes:
I think it is best if we relay that throttling to each Loadbalancer really. As you said, different paths might induce different types of throttling. Each loadbalancer should get to decide if it throttles the user (which makes more sense than what we do today anyway, which is throttling after some number is reached vs. actually looking at capacity).

So I'd say, throttling is not based on the sum of all loadbalancers, but it is decided by each loadbalancer itself after the request has been dispatched to it.

tysonnorris:
closing in favor of #2688

PR #2658, opened 2017-08-22
Treat a timed out active ack as failed activation in invokerhealth protocol

cbickel:
The Loadbalancer has a buffer of, if the last activations of an invoker has been successful.
This PR treats an activation as unsuccessful, if the active ack times out.
This will take the invoker unhealthy, if it does not respond with active acks anymore.

PG4#714 is running

### DIFF
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
@@ -285,8 +285,9 @@ class InvokerActor(invokerInstance: InstanceId, controllerInstance: InstanceId) 
 
     // If the action is successful it seems like the Invoker is Healthy again. So we execute immediately
     // a new test action to remove the errors out of the RingBuffer as fast as possible.
-    // The actions that arrice while the invoker is unhealthy are most likely health actions.
-    // But it could be real actions as well, if they were already in the queue/in progress while the invoker turned unhealthy.
+    // The actions that arrive while the invoker is unhealthy are most likely health actions.
+    // It is possible they are normal user actions as well. This can happen if such actions were in the
+    // invoker queue or in progress while the invoker's status flipped to Unhealthy.
     if (wasActivationSuccessful && stateName == UnHealthy) {
       invokeTestAction()
     }
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
@@ -283,9 +283,10 @@ class InvokerActor(invokerInstance: InstanceId, controllerInstance: InstanceId) 
   private def handleCompletionMessage(wasActivationSuccessful: Boolean, buffer: RingBuffer[Boolean]) = {
     buffer.add(wasActivationSuccessful)
 
-    // If the current state is UnHealthy, then the active ack is the result of a test action.
-    // If this is successful it seems like the Invoker is Healthy again. So we execute immediately
+    // If the action is successful it seems like the Invoker is Healthy again. So we execute immediately
     // a new test action to remove the errors out of the RingBuffer as fast as possible.
+    // The actions that arrice while the invoker is unhealthy are most likely health actions.
+    // But it could be real actions as well, if they were already in the queue/in progress while the invoker turned unhealthy.
     if (wasActivationSuccessful && stateName == UnHealthy) {
       invokeTestAction()
     }
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
@@ -129,7 +129,7 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
                                 invoker: InstanceId): Unit = {
     val aid = response.fold(l => l, r => r.activationId)
 
-    // treat left as success (as it is the result a the message exceeding the bus limit)
+    // treat left as success (as it is the result of a message exceeding the bus limit)
     val isSuccess = response.fold(l => true, r => !r.response.isWhiskError)
 
     loadBalancerData.removeActivation(aid) match {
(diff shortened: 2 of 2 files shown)


markusthoemmes:
@rabbah I do believe though that this should result in a 429 or rather: This is a better indicator than the arbitrary system limit we set today, isn't it?

rabbah:
Pathologically can you end up where all invokers appear unhealthy (which result in 500 level errors instead of system overload)?

rabbah:
The pings should carry a count indicating forward progress. I think treating timed out active acks as a sign of an unhealthy invoker like this done here is problematic - I'll explain why when I get a chance.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -222,11 +228,8 @@ class LoadBalancerService(
         val raw = new String(bytes, StandardCharsets.UTF_8)
         CompletionMessage.parse(raw) match {
             case Success(m: CompletionMessage) =>
-                processCompletion(m.response, m.transid, false)
-                // treat left as success (as it is the result a the message exceeding the bus limit)
-                val isSuccess = m.response.fold(l => true, r => !r.response.isWhiskError)
+                processCompletion(m.response, m.transid, false, m.invoker)
```
rabbah: s/false/forced = false/

cbickel:
PG3#1116 is 🔵

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -286,9 +286,10 @@ class InvokerActor(invokerInstance: InstanceId, controllerInstance: InstanceId)
   private def handleCompletionMessage(wasActivationSuccessful: Boolean, buffer: RingBuffer[Boolean]) = {
     buffer.add(wasActivationSuccessful)
 
-    // If the current state is UnHealthy, then the active ack is the result of a test action.
-    // If this is successful it seems like the Invoker is Healthy again. So we execute immediately
+    // If the action is successful it seems like the Invoker is Healthy again. So we execute immediately
     // a new test action to remove the errors out of the RingBuffer as fast as possible.
+    // The actions that arrice while the invoker is unhealthy are most likely health actions.
```
rabbah: arrice -> arrive.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -286,9 +286,10 @@ class InvokerActor(invokerInstance: InstanceId, controllerInstance: InstanceId)
...
+    // But it could be real actions as well, if they were already in the queue/in progress while the invoker turned unhealthy.
```
rabbah: It is possible they are normal user actions as well. This can happen if such actions were in the invoker queue or in progress while the invoker's status flipped to Unhealthy.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -125,18 +125,34 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
...
   private def processCompletion(response: Either[ActivationId, WhiskActivation],
                                 tid: TransactionId,
-                                forced: Boolean): Unit = {
+                                forced: Boolean,
+                                invoker: InstanceId): Unit = {
     val aid = response.fold(l => l, r => r.activationId)
+
+    // treat left as success (as it is the result a the message exceeding the bus limit)
+    // treat timed out active ack as failure to let the invoker become unhealthy
+    val isSuccess = response.fold(l => true, r => !r.response.isWhiskError)
```
rabbah: timed out active ack === forced.
but this fold does not account for the forced value.
comment inconsistent.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -125,18 +125,34 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
...
+
     loadBalancerData.removeActivation(aid) match {
       case Some(entry) =>
         logging.info(this, s"${if (!forced) "received" else "forced"} active ack for '$aid'")(tid)
+        // If we receive an active ack (forced = false) we send the activationResult. If the timeout
```
rabbah: activationResult?

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -125,18 +125,34 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
...
+        // is executed earlier, we treat the activation as failed (forced = true).
+        // At this point no health actions are handled, because they don't have an entry in loadBalancerData.
+        invokerPool ! InvocationFinishedMessage(invoker, isSuccess && !forced)
```
rabbah: slight rewording for clarity:

Active acks that are received here are strictly from user actions - health actions are not part of the load balancer's activation map. Inform the invoker pool supervisor of the user action completion. If the active ack was forced, because the waiting period expired, treat it as a failed activation. A cluster of such failures will eventually turn the invoker unhealthy and suspend queuing activations to that invoker topic.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -125,18 +125,34 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
...
           entry.promise.tryFailure(new Throwable("no active ack received"))
         }
       case None =>
         // the entry was already removed
+        // This could happen if this is the active ack of an health action. Another reason is
+        // that this method is called twice. On activeAcks and on timeouts. The first hit removes the entry.
+        // We send the InvocationFinishedMessage only, when processCompletion has not been called by the
+        // timeout (forced = false). Health actions don't have a timeout, so there is no problem, if they
+        // are very long in the queue.
+        if (!forced) invokerPool ! InvocationFinishedMessage(invoker, isSuccess)
```
rabbah: we could probably distinguish better between activate acks matching None for the two cases described:
if the `response` is left, _it should not be a health action_. So I wonder if can add two case matches here:

```
case None if response.isLeft => // this is a clean up active ack, drop it
case /* None and response.isRight */ => // this is a health action

PR #2659, opened 2017-08-22
Containerfactory SPI

tysonnorris:
This SPI interface allows alternate container management strategies, e.g. docker, mesos, kubernetes, etc. 

Default impl is DockerContainerFactoryProvider, which has same behavior as before.


core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
```diff
@@ -0,0 +1,67 @@
...
+}
+
+class DockerContainerFactoryProvider extends ContainerFactoryProvider {
+    override def getContainerFactory(actorSystem: ActorSystem, logging: Logging, config: WhiskConfig): ContainerFactory =
+        new DockerContainerFactory()(actorSystem.dispatcher, logging)
+}
+
+object DockerContainerFactoryProvider extends SpiFactory[ContainerFactoryProvider] {
+    override def apply(dependencies: Dependencies): ContainerFactoryProvider = new DockerContainerFactoryProvider()
+}
```
markusthoemmes: I do still think that these SPI implementations are too verbose, the intermediary steps (including `SingletonSpiFactory`) seems unnecessary bloat. Shall we refactor those bits before getting a lot of interfaces in, that we want to change soon anyway?

common/scala/src/main/scala/whisk/core/containerpool/Container.scala
```diff
@@ -66,15 +74,3 @@ case class BlackboxStartupError(msg: String) extends ContainerStartupError(msg)
...
-    def duration = Duration.create(end.toEpochMilli() - start.toEpochMilli(), MILLISECONDS)
-}
-
-object Interval {
-    /** An interval starting now with zero duration. */
-    def zero = {
-        val now = Instant.now
-        Interval(now, now)
-    }
-}
```
markusthoemmes: Why did you move these bits?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
tysonnorris: What specifically do you have in mind? Are you sure this can be simplified further?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -120,7 +122,7 @@ object DockerContainer {
  * @param id the id of the container
  * @param ip the ip of the container
  */
-class DockerContainer(id: ContainerId, ip: ContainerIp)(
+class DockerContainer(val id: ContainerId, val ip: ContainerIp)(
```
markusthoemmes: Do these need to be public?

common/scala/src/main/scala/whisk/core/containerpool/Container.scala
tysonnorris: I'm suggesting that container management can be handled outside of an invoker, therefore Container related interfaces do not belong there.

markusthoemmes: Right, that's fine. But why move these classes out of `Container.scala`?

tysonnorris: I'm not sure why I moved them originally - moved back to Container.scala

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
tysonnorris: id/ip are needed by the HTTP API during initialize() and run() - latest changes have these as protected val, but also has the HTTP API methods initialize() and run() moved into a separate class so they can be shared - see ContainerApi.scala

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -72,31 +69,12 @@ class InvokerReactive(config: WhiskConfig, instance: InstanceId, producer: Messa
...
-    }
-    cleanup()
-    sys.addShutdownHook(cleanup())
 
     /** Factory used by the ContainerProxy to physically create a new container. */
-    val containerFactory = SpiLoader.get[ContainerFactoryProvider]().getContainerFactory(actorSystem, logging, config).createContainer _
+    val containerFactory = SpiLoader.get[ContainerFactoryProvider].getContainerFactory(actorSystem, logging, config, instance)
+    val containerFactoryFunction = containerFactory.createContainer _
+    containerFactory.cleanup()
+    sys.addShutdownHook(containerFactory.cleanup())
```
jcrossley3: I think this just effectively runs `cleanup` twice at startup. Shouldn't it be `sys.addShutdownHook { containerFactory.cleanup() }` instead?

tysonnorris: This is the same syntax that was there before; I tested it and it does the right thing - runs once at startup, and once at shutdown.

markusthoemmes: For posterity:

```scala
def addShutdownHook(body: => Unit): ShutdownHookThread = ShutdownHookThread(body)
```

`body` is a by-name argument, which is only evaluated once its used. I agree it looks like the function is actually called, but braces or parentheses do no make a difference in Scala here.

jcrossley3: Thanks for the explanation!

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
tysonnorris: Simplified variant implemented now (no more SpiFactory)

common/scala/src/main/scala/whisk/core/containerpool/ContainerApi.scala
```diff
@@ -0,0 +1,146 @@
...
+import whisk.common.Logging
+import whisk.common.LoggingMarkers
+import whisk.common.TransactionId
+import whisk.core.entity.ActivationResponse
+import whisk.core.entity.ActivationResponse.ContainerConnectionError
+import whisk.core.entity.ActivationResponse.ContainerResponse
+import whisk.core.entity.size._
+import whisk.http.Messages
+
+trait ContainerApi { container: Container =>
```
markusthoemmes: Is this type narrowing useful? I thought about doing the same abstraction a while ago aswell. Would it make sense to instead model it as 

```scala
abstract class HttpContainer(id: ContainerId, ip: ContainerIp)(implicit ec: ExecutionContext, logging: Logging) extends Container
```

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -81,53 +75,14 @@ class InvokerReactive(config: WhiskConfig, instance: InstanceId, producer: Messa
...
-        cpuShares = config.invokerCoreShare.toInt,
-        environment = Map("__OW_API_HOST" -> config.wskApiHost),
-        network = config.invokerContainerNetwork,
-        dnsServers = config.invokerContainerDns,
-        name = Some(name))
-    }
+    SpiLoader
+      .get[ContainerFactoryProvider]
+      .getContainerFactory(actorSystem, logging, config, instance)
+  val containerFactoryFunction = containerFactory.createContainer _
```
markusthoemmes: No need to create an additional `val` I think, just put `containerFactory.createContainer _` into `props` directly.

common/scala/src/main/scala/whisk/core/containerpool/ContainerApi.scala
```diff
@@ -0,0 +1,146 @@
...
+    }
+    Future {
+      http.post(path, body, retry)
+    }.map { response =>
+      val finished = Instant.now()
+      RunResult(Interval(started, finished), response)
+    }
+  }
+
+  def destroyApi()(implicit transid: TransactionId): Future[Unit] = {
```
markusthoemmes: Why `destroyApi` vs. just `destroy`?

common/scala/src/main/scala/whisk/core/containerpool/ContainerApi.scala
markusthoemmes: Oh got it, this is the "formalized" version of something the runtime needs to implement and stuff. Hmm. Might be sensible. Thinking about it.

common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
```diff
@@ -0,0 +1,74 @@
...
+    val dockerRunParameters = new HashMap[String, Set[String]] with MultiMap[String, String]
+    dockerRunParameters.addBinding("--cap-drop", "NET_RAW")
+    dockerRunParameters.addBinding("--cap-drop", "NET_ADMIN")
+    dockerRunParameters.addBinding("--ulimit", "nofile=1024:1024")
+    dockerRunParameters.addBinding("--pids-limit", "1024")
+    val dnsServers = config.invokerContainerDns
+    dnsServers.map(d => dockerRunParameters.addBinding("--dns", d))
+    dockerRunParameters
+  }
+}
```
markusthoemmes: Not sure if using a `Map` here is useful. For additional arguments I found it mostly useful to just use a flat `Seq` and pass stuff that way.

Alternatively we could have a type encapsulating all the options on a docker run and propagate it that way. But please do not use a mutable map here. If you think a Map is the right thing to do, let's at least use an immutable one (I know you can't use MultiMap then)

common/scala/src/main/scala/whisk/core/containerpool/ContainerApi.scala
tysonnorris: e.g. `MesosTask extends Container with ContainerApi` - an alternative is to move the HTTP contract definition into Container?

common/scala/src/main/scala/whisk/core/containerpool/ContainerApi.scala
tysonnorris: I wasn't really sure how best to call `ContainerApi.destroy()` from `DockerContainer.destroy()`, since DockerContainer is now defined `with ContainerApi`. 
Ideas?

common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
tysonnorris: At least partly I used Map because a) all the options I think that should be settable here are key-value and b) I happen to know in mesos we will have to translate this into a Map :(

I think the options will boil down to either using Seq (which at least some alternative container factories will have to decode to a Map) or writing an immutable version of MultiMap. I'm not wild about either of these prospects, let me know if you have other ideas.

common/scala/src/main/scala/whisk/core/containerpool/ContainerApi.scala
markusthoemmes: Right, we could just fold it into `Container` itself or have an intermediary type between `HttpContainer` which is exactly about the Http communication. Both sound sensible to me, cause I think all the container impls will just have `with ContainerApi`.

common/scala/src/main/scala/whisk/core/containerpool/ContainerApi.scala
markusthoemmes: Yeah I guess that's why I'd prefer it folded into either Container or have HttpContainer as an extension: Implementations can delegate to `super.destroy()`, which makes the impl. more transparent maybe?

common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
markusthoemmes: As seen in the method signature, `MultiMap` is just some sugar over `Map[String, Set[String]]`, so you can translate your mutable code to:

```scala
Map(
  "--cap-drop" -> Set("NET_RAW", "NET_ADMIN"),
  "--pids-limit" -> Set("1024"),
  "--dns" -> config.invokerContainerDns.toSet
)
```

Which I'd prefer to keep the interface immutable. MultiMap-sugar is arguably not helping much here anyway?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -66,27 +59,25 @@ object DockerContainer {
...
+    val environmentArgs = environment.flatMap {
       case (key, value) => Seq("-e", s"$key=$value")
-    }.flatten
+    }
 
-    val dnsArgs = dnsServers.map(Seq("--dns", _)).flatten
+    val params = dockerRunParameters.flatMap {
+      case (key, valueList) =>
+        valueList.map {
+          case (value) => Seq(key, value)
```
markusthoemmes: Drop the unnecessary parentheses around `value`.

common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
markusthoemmes: Also: Shouldn't this method exist only for the docker implementation?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -66,27 +59,25 @@ object DockerContainer {
...
-    val environmentArgs = environment.map {
+    val environmentArgs = environment.flatMap {
       case (key, value) => Seq("-e", s"$key=$value")
-    }.flatten
+    }
 
-    val dnsArgs = dnsServers.map(Seq("--dns", _)).flatten
+    val params = dockerRunParameters.flatMap {
+      case (key, valueList) =>
+        valueList.map {
```
markusthoemmes: Could be `flatMap` to not iterate again.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
```diff
@@ -0,0 +1,99 @@
...
+      cpuShares = config.invokerCoreShare.toInt,
+      environment = Map("__OW_API_HOST" -> config.wskApiHost),
+      network = config.invokerContainerNetwork,
+      dnsServers = config.invokerContainerDns,
+      name = Some(name),
+      parameters)
+  }
+
+  /** Perform cleanup on init */
+  override def init: Unit = cleanup()
```
markusthoemmes: Add parentheses to `init` and `cleanup`s definition please. Both are sideeffecting and not "get like" behavior with no-sideeffects.

common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
tysonnorris: Agree on immutable MultiMap, will make those changes.

Run parameters are possible in other non-docker-cli container management systems (e.g. mesos, guessing kube as well), so it should be useable by any SPI impl.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
tysonnorris: FlatMap on valueList (a Set) caused the duplicate keys to be removed - I changed it to `valueList.toList.flatMap` to avoid that, but I'm not sure that is what you're looking for, performance-wise. Let me know if you have suggestions.

common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
markusthoemmes: Still the method is called `dockerRunParameters` but it's part of the general `ContainerFactory`. It's not part of any interface, so it should live in `DockerContainerFactory`?

tysonnorris: Are you suggesting to rename it to `containerRunParameters`? or that there may be some non-docker container support, like lxc/oci/etc? 

The problem with "moving it to DockerContainerFactory" is that DockerContainerFactory shouldn't be used if some other container factory is used, but these parameters are still important. 

This may be confusing terminology, but currently the notion of "container" exists only in the form of "docker container", but the ContainerFactory is allowing multiple approaches to creating "docker containers" (as opposed to creating "other non-docker containers"). Its a difference of using docker cli to create a docker container, vs using some other system (which in turn uses docker cli) to create a docker container.

markusthoemmes:
PG1 2102 🔵

markusthoemmes:
@tysonnorris this is good to be merged but I wanted to squash it locally which failed miserably because of the merge commits inside. I got conflicts all about. Can you please rebase this on latest master and squash it to one commit with an appropriate title?

tysonnorris:
@markusthoemmes I did it this way, since I don't relish reapplying every merge conflict:
```
git reset --soft master
git commit -m "ContainerFactory SPI"
git push origin +containerfactory
```
Let me know if you think that is problematic?

markusthoemmes:
Great job, thanks for doing this 🎉

PR #2660, opened 2017-08-22
Reduce Memory Consumption Used to Invoke a Single Action

dubeejw:
Reduces number of call-by-value references to action code.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/actions/PostActionActivation.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/actions/PostActionActivation.scala
@@ -58,7 +58,18 @@ protected[core] trait PostActionActivation extends PrimitiveActions with Sequenc
                 invokeSequence(user, action, components, payload, waitForResponse, cause, topmost = true, 0).map(r => r._1)
             // a non-deprecated ExecutableWhiskAction
             case Some(executable) if !executable.exec.deprecated =>
-                invokeSingleAction(user, executable, payload, waitForResponse, cause)
+                invokeSingleAction(user,
+                    action.rev,
+                    action.parameters,
+                    action.namespace,
+                    action.name,
+                    action.version,
+                    executable.exec.pull,
+                    action.fullyQualifiedName(false).asString,
+                    action.limits.timeout.duration,
+                    payload,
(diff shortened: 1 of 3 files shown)


dubeejw:
@rabbah, @markusthoemmes, can you have a look. Haven't looked through sequences yet.

markusthoemmes:
Have you measured the impact of this?

Note that "call-by-value" for instances means: passing a reference, which is virtually free in terms of heap consumption.

markusthoemmes:
While it's true that, in theory, not holding a reference to the action code could make it possible to release the heap consumed by the action code (our cache though holds a weak-reference as well), in this case we would speak about milliseconds of delay pre-change vs. Post-change.

Thats is: If we're leaking a reference to a closure that lives for longer (we used to do that with activations), that would indeed be a real problem.

rabbah:
@starpit memory profiled a lot of the controller. Nick - thoughts?

starpit:
i suggest that, before we proceed (and, indeed, for any memory-related change, if not perhaps even for *any* controller change going forwards) that we should collect:

at the very least: max heap occupancy for a test which invokes a longish sequence N-way, in parallel. compare this value to that of the current master.

very important also: collect a class histogram after completion of the test, say within 30 seconds (very important: less than 5 minutes after test completion), and ensure that the number of WhiskActivation structures is very small; if N threads each execute a sequence of length M, and to his L times, then the number of WhiskActivation structures alive should be close to N-ish, some small multiple of it at worst.

ideally, we also want the instantaneous number of JSON structures to be close to N (if that is max concurrency for the load test), and the number should decrease to zero within 10 seconds of test completion.

beyond this, we want the number of JSON structures in heap to be constant, independent of the number of concurrently live actors. this will take quite a bit more work, but is probably the single biggest offender.

beyond WhiskActivation and JSON structures, we also want to establish the property that the number of code/blob strings in heap is zero. we should get to this point, and write tests that establish the verity of this condition, by using class histogram snapshots taken during such a mild load test run (i.e. we don't need to do a heavy duty load test in order to identify these problems, and to verify that they remain fixed). but it does require some amount of load, i.e. these are just unit tests.

PR #2662, opened 2017-08-23
Add judge condition for invoker deployment

ningyougang:
If invokers have been deployed success.
Then, it will report error when execute `ansible-playbook -i environments/<env> invoker.yml ` again.
because the invoker0 or invoker1 alreay exist, so below errors will appear.
```
TASK [invoker : start invoker using docker cli] *********************************************************************************************
...
...
You have to remove (or rename) that container to be able to reuse that name..\nSee 'docker run --help'.", "stderr_lines": ["docker: Error response from daemon: Conflict. The name \"/invoker0\" is already in use by container 1de4dc7d6ca3234d1b6590061e637c48367f9076766f767ef9271d1848e08620. You have to remove (or rename) that container to be able to reuse that name..", "See 'docker run --help'."], "stdout": "", "stdout_lines": []}
[FAILED]
```

So it is necessary to add the `judge condition` that whether the invoker has been deployed,
if already deployed, there has no need to deploy again.

### DIFF
--- a/ansible/roles/invoker/tasks/deploy.yml
+++ b/ansible/roles/invoker/tasks/deploy.yml
@@ -59,6 +59,10 @@
   with_items: "{{ invokerInfo.json }}"
   when: not invoker.allowMultipleInstances and item.Names[0] != "/invoker{{ groups['invokers'].index(inventory_hostname) }}"
 
+- name: get invoker container Names list
+  set_fact:
+    nameList: "{{ invokerInfo.json | map(attribute='Names') | list }}"
+
 - name: start invoker using docker cli
   shell: >
         docker run -d
@@ -104,6 +108,7 @@
         -p {{ invoker.port + groups['invokers'].index(inventory_hostname) }}:8080
         {{ docker_registry }}{{ docker.image.prefix }}/invoker:{{ docker.image.tag }}
         /bin/sh -c "exec /invoker/bin/invoker {{ groups['invokers'].index(inventory_hostname) }} >> /logs/invoker{{ groups['invokers'].index(inventory_hostname) }}_logs.log 2>&1"
+  when: "['/invoker{{ groups['invokers'].index(inventory_hostname) }}'] not in nameList"
 
 # todo: re-enable docker_container module once https://github.com/ansible/ansible-modules-core/issues/5054 is resolved
 


markusthoemmes:
Can we implement a recreate behavior like we have for all our components? We should somehow use ansible's capabilities to detect changed components and leave those alone that can live unchanged.

ningyougang:
@markusthoemmes 
Can we implement a recreate behavior like we have for all our components?
answer: if use `recreate` feature, because `invoker` is special, it will cause problem. 
as you know `user's action container is created by invoker`, if use `recreate`, those action containers will be removed when `invoker` is stoped.
So i think the best way to `redeploy invoker` using `-e mode=clean` first and `-e mode=deploy`.

ningyougang:
@markusthoemmes ,what do you think about this?

rabbah:
@sven-lange-last this is possibly related to enabled you've done to deploy individual invokers. It would be useful to document that capability as well. Do you mind taking a look at this pr?

sven-lange-last:
- Well, this PR just wants to introduce a guard that supports re-running the invoker playbook without affecting a deployed invoker.
- At the same time, @markusthoemmes is asking @ningyougang for a re-create behaviour.
- I think @ningyougang is right with his assessment that a simple re-creation would drop / impact running user invocations as well as dropping some invocations that have already been received from Kafka.
- A reasonable invoker re-create playbook has to include an invoker drain and re-enablement.
- @ningyougang - is my understanding correct that you want to improve robustness of the invoker playbook?
- @rabbah - I just have a way to run Ansible playbooks such that an already broken invoker can be refreshed without affecting the other invokers: `ansible-playbook -i environments/... -e docker_image_tag=$TAG --limit=$INVOKER_HOST openwhisk.yml -e mode=clean` and `ansible-playbook -i environments/... -e docker_image_tag=$TAG --limit=$INVOKER_HOST openwhisk.yml`.

sven-lange-last:
@ningyougang - what's the intention behind this PR? A way to add new invokers to a running system? Why do you want to run a deployment playbook again on an already deployed system?

ningyougang:
Why i did this?
becuase as you said
```
Well, this PR just wants to introduce a guard that supports re-running the invoker playbook without affecting a deployed invoker.
```

rabbah:
> A reasonable invoker re-create playbook has to include an invoker drain and re-enablement.

This is coming via https://github.com/apache/incubator-openwhisk/pull/3704.

markusthoemmes:
Closing due to inactivity. @ningyougang since this has conflicts and if this is still needed: Please fix the conflicts and reopen the PR. Thanks!

PR #2663, opened 2017-08-23
Simplify Spi implementations.

markusthoemmes:
Using `object`s instead of a hollow class as the "factory" for an Spi makes them being singleton by default, which reduces boilerplate to implement an Spi.

The `Dependencies` object is not used anywhere and there might be better alternatives which we can decide on once the need of them comes up. For now, the Spi's interfaces are relatively tightly coupled to their current default implementations anyway.


markusthoemmes:
@tysonnorris would you do me the honors?

tysonnorris:
Looks good - Thanks

rabbah:
does this require any changes to the SPI example doc?

PR #2676, opened 2017-08-31
Change response when entity is empty from 200 to 204.

### LINKED ISSUE #2674: controller should respond with 204 (no content) when a body is missing for HTTP web actions
For an http web action, the absence of a body results in an empty entity (content-length = 0).
The controller however uses a default status code if one is not specified, which is 200. Instead, this should be 204 No content.

Pseudo-code courtesy of @akrabat:
```php
if (webaction_response["status"] is empty) {
    if webaction_response["body"] is empty) {
        response_to_client["status"] = 204;
    } else {
        response_to_client["status"] = 200;
    }
}
```

rabbah:
This allows empty json response with http actions.
Add tests for handling partial result for http actions.

Fixes #2674.

- The first commit fixes a bug in the test actions (unrelated to this change, but got exposed by this pr).
- The second commit adds tests coverage for the various corner cases and changes the status code for the empty body response.
- The third commit treats the empty string and a null JSON value for body as the empty content.
- The last commit suppresses the akka-http content negotiation when a code is specified and no body.

FYI @akrabat.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
@@ -43,6 +43,7 @@ import akka.http.scaladsl.model.ContentTypes
 import akka.http.scaladsl.model.FormData
 import akka.http.scaladsl.model.HttpMethods.{DELETE, GET, HEAD, OPTIONS, PATCH, POST, PUT}
 import akka.http.scaladsl.model.HttpCharsets
+import akka.http.scaladsl.model.HttpResponse
 
 import spray.json._
 import spray.json.DefaultJsonProtocol._
@@ -222,6 +223,8 @@ protected[core] object WhiskWebActionsApi extends Directives {
         case _ => throw new Throwable("Invalid header")
       } getOrElse List()
 
+      val body = fields.get("body")
+
       val code = fields.get(rp.statusCode).map {
         case JsNumber(c) =>
           // the following throws an exception if the code is
@@ -229,22 +232,26 @@ protected[core] object WhiskWebActionsApi extends Directives {
           StatusCode.int2StatusCode(c.toIntExact)
 
         case _ => throw new Throwable("Illegal code")
-      } getOrElse (OK)
-
-      fields.get("body") map {
-        case JsString(str) => interpretHttpResponse(code, headers, str, transid)
-        case js            => interpretHttpResponseAsJson(code, headers, js, transid)
-      } getOrElse {
-        respondWithHeaders(removeContentTypeHeader(headers)) {
-          // note that if header defined a content-type, it will be ignored
-          // since the type must be compatible with the data response
-          complete(code, HttpEntity.Empty)
-        }
       }
+
+      body.collect {
+        case JsString(str) if str.nonEmpty   => interpretHttpResponse(code.getOrElse(OK), headers, str, transid)
+        case JsString(str) /* str.isEmpty */ => respondWithEmptyEntity(code.getOrElse(NoContent), headers)
+        case js if js != JsNull              => interpretHttpResponseAsJson(code.getOrElse(OK), headers, js, transid)
+      } getOrElse respondWithEmptyEntity(code.getOrElse(NoContent), headers)
+
     } getOrElse {
-      // either the result was not a JsObject or there was an exception validting the
+      // either the result was not a JsObject or there was an exception validating the
       // response as an http result
       terminate(BadRequest, Messages.invalidMedia(`message/http`))(transid, jsonPrettyPrinter)
+    }
+  }
+
+  private def respondWithEmptyEntity(code: StatusCode, headers: List[RawHeader]) = {
+    respondWithHeaders(removeContentTypeHeader(headers)) {
+      // note that if header defined a content-type, it will be ignored
+      // since the type must be compatible with the data response
+      complete(HttpResponse(code, entity = HttpEntity.Empty))
     }
   }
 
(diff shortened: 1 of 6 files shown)


tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -770,6 +805,74 @@ trait WebActionsApiTests extends ControllerTestCommon with BeforeAndAfterEach wi
...
+                            webApiDirectives.statusCode -> Created.intValue.toJson,
+                            "body" -> Base64.getEncoder.encodeToString {
+                                JsObject("field" -> "value".toJson).compactPrint.getBytes
+                            }.toJson))
+
+                        m(s"$testRoutePath/$path") ~> Route.seal(routes(creds)) ~> check {
+                            status should be(Created)
+                            responseAs[String] shouldBe actionResult.get.fields("body").convertTo[String]
+                            //??? it appears akka-http is ignoring the character set
+                            if (false) contentType shouldBe MediaTypes.`text/html`.withCharset(HttpCharsets.`UTF-8`)
```
rabbah: @dubeejw any insights on this?

rabbah:
- The first commit fixes a bug in the test actions (unrelated to this change, but got exposed by this pr).
- The second commit adds tests coverage for the various corner cases and changes the status code for the empty body response
- The third commit is debatable: should the controller suppress the akka-http content negotiation when a code is specified and no body? Otherwise, the current tests in this commit will fail because the response is Not Acceptable, and not the action given code.

rabbah:
Related comment from akka-http https://github.com/akka/akka-http/issues/366#issuecomment-252912853.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -216,32 +216,48 @@ protected[core] object WhiskWebActionsApi extends Directives {
...
                     // the following throws an exception if the code is
                     // not a whole number or a valid code
                     StatusCode.int2StatusCode(c.toIntExact)
 
                 case _ => throw new Throwable("Illegal code")
-            } getOrElse (OK)
+            }
+
+            body map {
+                case JsString(str) =>
```
markusthoemmes: Would it make sense to add the `str.nonEmpty` as a guard and let the rest fall through to the default case which is identical to the `else` case here?

markusthoemmes: Thinking about it, we could simplify it even more maybe:

```scala
body.collect {
    case JsString(str) if str.nonEmpty => 
        interpretHttpResponse(code.getOrElse(OK), headers, str, transid)
    case js if js != JsNull => 
        interpretHttpResponseAsJson(code.getOrElse(OK), headers, js, transid)
}.getOrElse(respondWithEmptyEntity(code.getOrElse(NoContent), headers))

Should be equivalent to the existing code.

File: tests/dat/actions/multipleHeaders.js [hunk 3]
```diff
@@ -3,6 +3,6 @@ function main() {
         headers: {
             "Set-Cookie": ["a=b", "c=d"]
         },
-        code: 200
+        statusCode: 200
```
markusthoemmes: Is this a breaking change wrt. the users? (The code itself doesn't appear so, just asking for extra confidence)

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
rabbah: After the first match the second will succeed. Since the type is JsString of an empty string. You need a second check.

tests/dat/actions/multipleHeaders.js
rabbah: That's what the property should have always been. The 'code' is otherwise ignored.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
rabbah:
```
            body.collect {
                case JsString(str) if str.nonEmpty   => interpretHttpResponse(code.getOrElse(OK), headers, str, transid)
                case JsString(str) /* str.isEmpty */ => respondWithEmptyEntity(code.getOrElse(NoContent), headers)
                case js if js != JsNull              => interpretHttpResponseAsJson(code.getOrElse(OK), headers, js, transid)
            } getOrElse respondWithEmptyEntity(code.getOrElse(NoContent), headers)
```

an alternative.

markusthoemmes: Meh. You're right of course. Worth it? (The last version?)

rabbah:
PG1/2016 🔵

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -699,20 +699,36 @@ trait WebActionsApiTests extends ControllerTestCommon with BeforeAndAfterEach wi
...
+                                    }
+                                }
 
-                        m(s"$testRoutePath/$path") ~> Route.seal(routes(creds)) ~> check {
-                            status should be(OK)
-                            response.entity shouldBe HttpEntity.Empty
-                            withClue(headers) {
-                                headers.length shouldBe 0
+                                // repeat with accept header, which should be ignored for content-negotiation
+                                m(s"$testRoutePath/$path") ~> addHeader("Accept", "application/json") ~> Route.seal(routes(creds)) ~> check {
```
dubeejw: Looks like we established a pattern of using ```MediaTypes.`application/json`.value``` instead of just using `application/json`. Six other of the same references below.

rabbah: I deliberately use the string when setting custom headers, but the constant when checking the expected result. I rather leave it this way since it mimics the actual header values one might set literally.

PR #2677, opened 2017-08-31
Add blocking indicator for activation request.

rabbah:
This allows non blocking requests to drop their results from the activeack response.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
@@ -90,6 +90,7 @@ case class WhiskActivation(
         }
     }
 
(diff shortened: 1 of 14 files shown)


common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
```diff
@@ -90,6 +90,7 @@ case class WhiskActivation(
         }
     }
 
+    def withoutLogsOrResponse = copy(response = ActivationResponse.success(), logs = ActivationLogs()).revision[WhiskActivation](rev)
```
markusthoemmes: Falling back to a success always breaks the health-protocol doesn't it? We can safely drop the user's response stuff, but we need to send the correct statuscode I think.

rabbah:
PG2/2027 🚀

rabbah:
PG2/2027 🔵

rabbah:
Squash all - thanks for asking.

PR #2680, opened 2017-08-31
do not use InstanceId.int within TransactionCounter

dgrove-oss:
abstract away the assumption that InstanceId must be
an integer in the range 0..numInstances-1 in TransactionCounter.
Small step in preparation for adding invoker elasticity.

### DIFF
--- a/common/scala/src/main/scala/whisk/common/TransactionId.scala
+++ b/common/scala/src/main/scala/whisk/common/TransactionId.scala
@@ -31,7 +31,6 @@ import spray.json.JsArray
 import spray.json.JsNumber
 import spray.json.JsValue
 import spray.json.RootJsonFormat
-import whisk.core.entity.InstanceId
 
 /**
  * A transaction id for tracking operations in the system that are specific to a request.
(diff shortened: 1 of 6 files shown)


markusthoemmes:
What's the desired goal here? Note that the current loadbalancer is strictly based numbering the instances from 0-(numberOfInstances-1). I don't like that, but that's how it is today.

rabbah:
The patch here addressed a bug: the invoker tids would be an infinite sequence of 0s. The invoker however does not generate tids (no rest interface). Separately, the naming/indexing of invokers can be addressed.

dgrove-oss:
to expand on what Rodric said:  
 invoker0 would generate tids 0,0,0,...
 invoker1 would generate 1,2,3,...
 invoker2 would generate 2,4,6,...
This fixes all of the invokers to generate tids 1,2,3,...

rabbah:
A side note that perhaps once the instances have unique, not ordinal, ids, the `tid` could be extended from `tid_NNN` to `tid_id_NNN` and avoid the delta increments entirely.

dgrove-oss:
rebased to resolve conflicts with Scala reformat.   @cbickel any review comments?

rabbah:
pg3/1067 🏃

dgrove-oss:
what's the next step with this PR?

rabbah:
pg3/1078 🏃 (because the last one has a build failure - unrelated).

PR #2684, opened 2017-09-01
Decorate throttled error response with count/limit.

### LINKED ISSUE #1385: improve throttling message
Our error message for throttling is the following:

Too many requests from user in a given amount of time. (code 186)

It should be enhanced to give more precise information:

"Too many requests   User has requested [x] actions in [time period y].  The limit is [z]"

rabbah:
Fixes #1385.

Reports error messages as:
`Too many requests in the last minute (count: 120, allowed: 60). (code ...)`
`Too many concurrent requests in flight (count: 30, allowed: 30). (code ...)`

PG2/2032 🔵

### DIFF
--- a/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
+++ b/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
@@ -59,10 +59,12 @@ object Messages {
   val resourceDoesNotExist = "The requested resource does not exist."
 
   /** Standard message for too many activation requests within a rolling time window. */
-  val tooManyRequests = "Too many requests in a given amount of time for namespace."
+  def tooManyRequests(count: Int, allowed: Int) =
+    s"Too many requests in the last minute (count: $count, allowed: $allowed)."
 
   /** Standard message for too many concurrent activation requests within a time window. */
-  val tooManyConcurrentRequests = "Too many concurrent requests in flight for namespace."
+  def tooManyConcurrentRequests(count: Int, allowed: Int) =
+    s"Too many concurrent requests in flight (count: $count, allowed: $allowed)."
 
   /** System overload message. */
   val systemOverloaded = "System is overloaded, try again later."
--- a/core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
+++ b/core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
@@ -21,6 +21,7 @@ import whisk.common.Logging
 import whisk.common.TransactionId
 import whisk.core.entity.Identity
 import whisk.core.loadBalancer.LoadBalancer
+import whisk.http.Messages
 
 /**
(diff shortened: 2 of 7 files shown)


core/controller/src/main/scala/whisk/core/entitlement/RateThrottler.scala
```diff
@@ -45,43 +46,41 @@ class RateThrottler(description: String, defaultMaxPerMinute: Int, overrideMaxPe
...
 /**
  * Tracks the activation rate of one subject at minute-granularity.
  */
 private class RateInfo {
-    var lastMin = getCurrentMinute
-    var lastMinCount = 0
-
-    def count() = lastMinCount
+    @volatile var lastMin = getCurrentMinute
+    var lastMinCount = new AtomicInteger()
```
markusthoemmes: These have been non threadsafe for all this time, have they? 🙈

rabbah: mmhmm.

core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
```diff
@@ -39,11 +40,11 @@ class ActivationThrottler(loadBalancer: LoadBalancer, defaultConcurrencyLimit: I
     /**
      * Checks whether the operation should be allowed to proceed.
      */
-    def check(user: Identity)(implicit tid: TransactionId): Boolean = {
+    def check(user: Identity)(implicit tid: TransactionId): RateLimit = {
         val concurrentActivations = loadBalancer.activeActivationsFor(user.uuid)
         val concurrencyLimit = user.limits.concurrentInvocations.getOrElse(defaultConcurrencyLimit)
-        logging.info(this, s"namespace = ${user.uuid.asString}, concurrent activations = $concurrentActivations, below limit = $concurrencyLimit")
-        concurrentActivations < concurrencyLimit
+        logging.info(this, s"namespace = ${user.uuid.asString}, concurrent activations = $concurrentActivations, limit = $concurrencyLimit")
```
mdeuser: for easier log searching, having the user namespace string included with the namespace uuid would be nice.

common/scala/src/main/scala/whisk/http/ErrorResponse.scala
```diff
@@ -57,10 +57,10 @@ object Messages {
     val resourceDoesNotExist = "The requested resource does not exist."
 
     /** Standard message for too many activation requests within a rolling time window. */
-    val tooManyRequests = "Too many requests in a given amount of time for namespace."
+    def tooManyRequests(count: Int, allowed: Int) = s"Too many requests in the last minute (count: $count, allowed: $allowed)."
```
dubeejw: Is the time period here configurable via the limits in Ansible? If so, it'd be nice to surface that value here instead of say `in the last minute`.

PR #2689, opened 2017-09-01
first stage of support for dynamic invoker id assignment

dgrove-oss:
A prototype of one possible approach for dynamic invoker registration.
The invoker's instanceId is not provided on the command line on
startup.  Instead, the invoker dynamically requests an id from
the controller during its startup via the message bus (kafka).
Once it receives an id, invoker startup continues unchanged.

On the plus side, this protocol maintains the property that invokers
are densely assigned ids starting from 0.  On the minus side, it takes
an extra round of messages thus mildly delaying system startup.

An alternative approach would be to use randomly generated UUIDs as
the invoker's id and adapt the load balancer algorithm accordingly.
Using UUIDs as invoker ids might make some aspects of operations
slightly more annoying as well, as the invokers would not have
stable (or short/easy to recognize) names.  This is a larger change,
but I think it is possible to implement without too much trouble.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
@@ -21,6 +21,8 @@ import scala.concurrent.Await
 import scala.concurrent.duration._
 import scala.util.Failure
 
+import com.redis.RedisClient
+
 import akka.actor.ActorSystem
 import akka.stream.ActorMaterializer
 import whisk.common.AkkaLogging
@@ -48,6 +50,7 @@ object Invoker {
       WhiskEntityStore.requiredProperties ++
       WhiskActivationStore.requiredProperties ++
       kafkaHost ++
+      redisHost ++
       wskApiHost ++ Map(
       dockerImageTag -> "latest",
       invokerNumCore -> "4",
@@ -55,12 +58,10 @@ object Invoker {
       invokerContainerPolicy -> "",
       invokerContainerDns -> "",
       invokerContainerNetwork -> null,
-      invokerUseRunc -> "true")
+      invokerUseRunc -> "true") ++
+      Map(invokerName -> null)
 
   def main(args: Array[String]): Unit = {
-    require(args.length == 1, "invoker instance required")
-    val invokerInstance = InstanceId(args(0).toInt)
-
     implicit val ec = ExecutionContextFactory.makeCachedThreadPoolExecutionContext()
     implicit val actorSystem: ActorSystem =
       ActorSystem(name = "invoker-actor-system", defaultExecutionContext = Some(ec))
@@ -86,6 +87,41 @@ object Invoker {
       abort()
     }
 
+    val proposedInvokerId: Option[Int] = args.headOption.map(_.toInt)
+    val assignedInvokerId = proposedInvokerId
+      .map { id =>
+        logger.info(this, s"invokerReg: using proposedInvokerId ${id}")
+        id
+      }
+      .getOrElse {
+        val invokerName = config.invokerName
+        val redisClient = new RedisClient(config.redisHostName, config.redisHostPort.toInt)
+        val assignedId = redisClient
+          .hget("controller:registar:idAssignments", invokerName)
+          .map { oldId =>
+            logger.info(this, s"invokerReg: invoker ${invokerName} was assigned its previous invokerId ${oldId}")
+            oldId.toInt
+          }
+          .getOrElse {
+            // If key not present, incr initializes to 0 before applying increment.
+            // Convert from 1-based to 0-based invokerIds by subtracting 1 from incr's result
+            val newId = redisClient
+              .incr("controller:registrar:nextInvokerId")
+              .map { id =>
+                id.toInt - 1
+              }
+              .getOrElse {
+                logger.error(this, "Failed to increment invokerId")
+                abort()
+              }
+            redisClient.hset("controller:registar:idAssignments", invokerName, newId)
+            logger.info(this, s"invokerReg: invoker ${invokerName} was assigned invokerId ${newId}")
+            newId
+          }
+        redisClient.quit
+        assignedId
+      }
+    val invokerInstance = InstanceId(assignedInvokerId);
     val msgProvider = SpiLoader.get[MessagingProvider]
     val producer = msgProvider.getProducer(config, ec)
     val invoker = new InvokerReactive(config, invokerInstance, producer)
--- a/ansible/roles/invoker/tasks/deploy.yml
+++ b/ansible/roles/invoker/tasks/deploy.yml
@@ -119,6 +119,8 @@
         -e PORT='8080'
         -e KAFKA_HOST='{{ groups['kafka']|first }}'
         -e KAFKA_HOST_PORT='{{ kafka.port }}'
+        -e REDIS_HOST='{{ groups['redis'] | first }}'
+        -e REDIS_HOST_PORT='{{ redis.port }}'
         -e DB_PROTOCOL='{{ db_protocol }}'
         -e DB_PROVIDER='{{ db_provider }}'
         -e DB_HOST='{{ db_host }}'
@@ -140,6 +142,7 @@
         -e INVOKER_NUMCORE='{{ invoker.numcore }}'
         -e INVOKER_CORESHARE='{{ invoker.coreshare }}'
         -e INVOKER_USE_RUNC='{{ invoker.useRunc }}'
+        -e INVOKER_NAME='{{ groups['invokers'].index(inventory_hostname) }}'
         -e WHISK_LOGS_DIR='{{ whisk_logs_dir }}'
         -v /sys/fs/cgroup:/sys/fs/cgroup
         -v /run/runc:/run/runc
--- a/ansible/templates/whisk.properties.j2
+++ b/ansible/templates/whisk.properties.j2
@@ -43,6 +43,7 @@ limits.actions.sequence.maxLength={{ limits.sequenceMaxLength }}
 
 edge.host={{ groups["edge"]|first }}
 kafka.host={{ groups["kafka"]|first }}
+redis.host={{ groups["redis"]|first }}
 router.host={{ groups["edge"]|first }}
 zookeeper.host={{ groups["kafka"]|first }}
 invoker.hosts={{ groups["invokers"] | map('extract', hostvars, 'ansible_host') | list | join(",") }}
@@ -51,6 +52,7 @@ edge.host.apiport=443
 zookeeper.host.port={{ zookeeper.port }}
 kafka.host.port={{ kafka.port }}
 kafkaras.host.port={{ kafka.ras.port }}
+redis.host.port={{ redis.port }}
 invoker.hosts.baseport={{ invoker.port }}
 
 controller.hosts={{ groups["controllers"] | map('extract', hostvars, 'ansible_host') | list | join(",") }}
--- a/common/scala/build.gradle
+++ b/common/scala/build.gradle
@@ -22,6 +22,7 @@ dependencies {
     compile 'ch.qos.logback:logback-classic:1.2.3'
     compile 'org.slf4j:jcl-over-slf4j:1.7.25'
     compile 'org.slf4j:log4j-over-slf4j:1.7.25'
+    compile 'net.debasishg:redisclient_2.11:3.4'
     compile 'commons-codec:commons-codec:1.9'
     compile 'commons-io:commons-io:2.4'
     compile 'commons-collections:commons-collections:3.2.2'
--- a/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
+++ b/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
@@ -65,6 +65,7 @@ class WhiskConfig(requiredProperties: Map[String, String],
   val invokerNumCore = this(WhiskConfig.invokerNumCore)
   val invokerCoreShare = this(WhiskConfig.invokerCoreShare)
   val invokerUseRunc = this.getAsBoolean(WhiskConfig.invokerUseRunc, true)
+  val invokerName = this(WhiskConfig.invokerName)
 
   val wskApiHost = this(WhiskConfig.wskApiProtocol) + "://" + this(WhiskConfig.wskApiHostname) + ":" + this(
     WhiskConfig.wskApiPort)
@@ -74,6 +75,8 @@ class WhiskConfig(requiredProperties: Map[String, String],
 
   val edgeHost = this(WhiskConfig.edgeHostName) + ":" + this(WhiskConfig.edgeHostApiPort)
   val kafkaHost = this(WhiskConfig.kafkaHostName) + ":" + this(WhiskConfig.kafkaHostPort)
+  val redisHostName = this(WhiskConfig.redisHostName)
+  val redisHostPort = this(WhiskConfig.redisHostPort)
 
   val edgeHostName = this(WhiskConfig.edgeHostName)
 
@@ -190,6 +193,7 @@ object WhiskConfig {
   val invokerNumCore = "invoker.numcore"
   val invokerCoreShare = "invoker.coreshare"
   val invokerUseRunc = "invoker.use.runc"
+  val invokerName = "invoker.name"
 
   val wskApiProtocol = "whisk.api.host.proto"
   val wskApiPort = "whisk.api.host.port"
@@ -205,9 +209,11 @@ object WhiskConfig {
 
   val kafkaHostName = "kafka.host"
   private val zookeeperHostName = "zookeeper.host"
+  val redisHostName = "redis.host"
 
   private val edgeHostApiPort = "edge.host.apiport"
   val kafkaHostPort = "kafka.host.port"
+  val redisHostPort = "redis.host.port"
   private val zookeeperHostPort = "zookeeper.host.port"
 
   val invokerHostsList = "invoker.hosts"
@@ -215,6 +221,7 @@ object WhiskConfig {
   val edgeHost = Map(edgeHostName -> null, edgeHostApiPort -> null)
   val invokerHosts = Map(invokerHostsList -> null)
   val kafkaHost = Map(kafkaHostName -> null, kafkaHostPort -> null)
+  val redisHost = Map(redisHostName -> null, redisHostPort -> null)
 
   val runtimesManifest = "runtimes.manifest"
 


dgrove-oss:
I can't add labels, but this is a WIP posted for discussion & feedback.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -89,6 +91,41 @@ object Invoker {
 
         val msgProvider = SpiLoader.get[MessagingProvider]()
         val producer = msgProvider.getProducer(config, ec)
+
+        val invokerOrdinal =
+          if (args.length == 0) {
+            val registrationUUID = UUID()
```
tysonnorris: If an invoker restarts on same host should it have a different UUID? Currently this will orphan messages queued for the ID just before it exited. I thought that the invoker on same host should have the same instance ID on restart. This implies that the ID is derived from the host (or cluster of hosts) somehow - we currently leverage the host IP (encoded as an int) for this purpose

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -89,6 +91,41 @@ object Invoker {
...
+            producer.send("invokerReg", InvokerIdRequestMessage(registrationUUID.asString)).andThen {
+              case Success(_) => {
+                logger.info(this, s"sent invokerReg message with UUID ${registrationUUID.asString}")
+              }
+              case Failure(t) => {
+                logger.error(this, s"failed to send invokerReg message: $t")
+                abort()
+              }
+            }
+            val topic = s"invokerReg-${registrationUUID.asString}"
```
tysonnorris: requires a new topic per restart of the same invoker?

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -332,3 +332,51 @@ object InvokerActor {
...
+
+    // State of the actor. It's important not to close over these
+    // references directly, so they don't escape the Actor.
+    var nextInvokerId = 0
+
+    def receive = {
+      case m : InvokerIdRequestMessage => {
+        val response = InvokerIdResponseMessage(m.uuid, nextInvokerId)
+        nextInvokerId += 1
+        sendIdToInvoker(response)
```
tysonnorris: This (sending an ID back to invoker) is only required because invoker id is currently an Int, and the registration was using UUID, right? Maybe we can change it to a String to be more flexible (@dragos and I just talked about this for other reasons as well). Then there is no need to explicitly send it an ID, AND maybe it can default to the host ip (and override with an "advertise ip" config).  This enforces the idea of "1 invoker per host" - not sure that breaks anything for kube.  

Side note: the case we ran into was where our invoker was misconfigured and kept restarting with a different ID each restart; eventually this crashed the *controller* because of the volume of health topic messages that were processed into a Map during controller startup, so separately it will be good to set the health topic consumer to receive only the latest n health messages. I'll log an issue for this.

dgrove-oss:
Keeping the same ID across restarts is a good point.  I was thinking we will need to have some mechanism for draining the topics of offline invokers, so we could just use that more frequently and not worry about reusing invoker ids across restarts. But that might not be optimal. 

Perhaps the path to pursue is allowing invoker IDs to be an arbitrary alphanumeric string (anything that is legal to embed in a kafka topic name) that the invoker computes _somehow_ without needing to communicate with the controller (by reading a command line, by computing a function on its ip address, by otherwise picking something out of its environment/filesystem, randomUUID, ...).  The controller/load balancer just uses the ID (provided to it implicitly by ping messages on the health topic) to create the kafka topic invoker-ID and the loadbalancer internally maps the ID to a dense numbering of currently active invokers to support hashing incoming work to invokers.

rabbah:
The invoker id could a uuid. When it registers with the load balancer, in the spirit of this pr, it is assigned a message topic to start draining. This topic id is thus orthogonal to the invoker id. This would make the protocol for restart of an invoker a bit more chatty (to recover a lost topic processor) but could be better longer term, and maybe gets us closer to work stealing.

ServoKvd:
Hello,
Im in favor of making the Invoker Stateless, I brought this up earlier on the slack chnnls- Am happy to see some momentum on this.   Im experimenting with  the following way of obraining the InstanceId from the Controller and calling this  API in the "Invoker" before it can bootstrap itself up with an ID ( ie. instead of reading from the properties/config file as it is doing now, it will read the Instance from teh query to the Controller) 

"curl controller.openwhisk:8080/invokers"  ---> This should give the current list of Invokers,
Get the length of current Invokers from here and create an Invoker with the next index or write a new
REST API  in Controller.scala, to get "availableIndex" for the Invoker. 
"avialable Indexes" can be a pool of Indexes/UUIDs maninatined by the Controller that may be assigned for new Invoker Instance creation. 

Any thoughts on this kind of solution? 

(Note:  Am having trouble with finding the routing path for the Controller REST APIS - so for now am modifying teh existing REST API( curl controller:openwhisk:8080/invokers) to return the data I need,
Any pointers as to how to add a new REST API path is apprecaited. 
(should be somehwere in some resoruces/*.json file, Im expecting)

dgrove-oss:
I've reworked the code based on the comments from @rabbah and @tysonnorris.  There is now the ability to provide a stable invokerUUID for an invoker instance via an environment variable.  The loadbalancer maintains a mapping of UUIDs to invokerIds to allow stable assignment of invoker topics across invoker restarts.   The invoker can also optionally suggest the invokerId it should be assigned (to preserve all the old behavior of statically assigning the invokerIds for each instance).

ServoKvd:
Hello,
I looked at the above patch, I understand all the changes, except I was wondering,  why we need to use Kfka Bus for InvokerIdRequest/Registration messages. Treating Adding/removing invokers as CRUD operations on the OpenWhisk Controller's Invenotry, I was wondering if adding a REST API as follows may be a viable design:
-------------------------------------------------------------------------------------
Invoker.scala:
import scalaj.http.{Http, HttpResponse}
// val invokerInstance = instanceId(args(0).toInt)
if(args.length ==1)
   val invokerInstance = instanceId(args(0).toInt)
else {
   val response: httpResponse[String] = Http("http://172.17.0.5:8000/invokerId").asString
   val invokerInstance = InstanceId(response.body.toInt)
}
-------------------------------------------------------------------------------------
Note:  Here,  172.17.0.5 is "openwhisk.controller" in myexample  implementation - trying to figure out how to expose the DNS mappings to Invoker, so that openwhisk.controller may resolve to 172.17.0.5.

dgrove-oss:
I decided to use the MessagingProvider (Kafka by default) because we already are assuming the controller and invokers can communicate with each other via this service.  So it seemed to add fewer constraints on how the system is deployed than having the invoker do a post to a controller REST API.

dgrove-oss:
I've squashed to a single commit and updated the commit comment to reflect the final design.

@rabbah I think this is ready for review and could have the WIP label removed.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -87,6 +89,38 @@ object Invoker {
 
     val msgProvider = SpiLoader.get[MessagingProvider]
     val producer = msgProvider.getProducer(config, ec)
+
+    val invokerUUID = sys.env.getOrElse("INVOKER_ASSIGNED_UUID", UUID().toString)
+    val proposedInvokerId = if (args.length == 1) { args(0).toInt } else { -1 }
```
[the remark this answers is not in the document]
dgrove-oss: True with current simplistic code.  I could document/enforce that if we think it is important to do so.

I think where we this is likely to end up is that the Registrar actor would get a little more complex and interact with the rest of the load balancing subsystem.  In particular, instead of blindly incrementing a counter if there is not match and no proposed id, it could first look to see if there is an abandoned topic (because of an unhealthy/offline invoker) and then prefer to assign that topic to the invoker instead of creating a new topic.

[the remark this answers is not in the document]
dgrove-oss: I added a check to the code to insist that either INVOKER_ASSIGNED_UUID is defined or a proposedId is provided on the command line.  If neither is provided, we abort the invoker.

DanLavine:
For Kubernetes, we would be able to leverage the hostname of the node each Invoker is running on assuming the Invoker is running as a DaemonSet. For example, in the yaml file for Invokers, you could add:

```
env:
- name: INVOKER_ASSIGNED_UUID
  valueFrom:
    fieldRef:
      fieldPath: spec.nodeName
```

I believe in Kuberentes each Kube Node has to have a unique name and so this should guarantee that the UUIDs are unique

bbrowning:
How does this loadbalancer bookkeeping work once we get multiple Controllers in the mix? There's some extra shared state and coordination required when assigning and remembering the UUID to Invoker id mapping?

If an Invoker dies, the only way to get an Invoker registered to that numerical Invoker id again is to use the proposedInvokerId? If I just start up a new Invoker without telling it which Invoker id to propose then it will get assigned a new topic in Kafka instead of draining from the orphaned one. So, in practice, my deployment tooling needs to keep up with Invoker ids just like it does today?

dgrove-oss:
For multiple controllers, the current code is assuming that every controller will see the messages on the invokerReg topic in the same order.  My understanding of kafka is superficial, but I believe that is what it provides.  Assuming it does, then the code in the Registrar agent should make exactly the same set of state updates in each controller process.  I've done a few dozen runs with multiple controllers and it seems to work, but that isn't proof (could be dumb luck still).   An alternative implementation could forgo relying on kafka and use Akka distributed data to maintain a shared counter and mapping data structure. 

If the invoker dies, but restarts with the same invokerUUID it had before, it will get the same invokerId assigned to it.  The main difference from what you need to do today is that although your deployment tooling still needs to assign UUIDs, they no longer have to be dense integers.  For example the invokerUUID can be the Kube nodeName or the IP address of the node (if you only deploy 1 invoker instance per IP address) or some other arbitrary alpha-numeric string.

bbrowning:
The controllers may see the messages in the same order, but that only works if the controllers all start together. If a controller needs to be restarted or a new one added how will the state get synchronized? Or will a new controller see all messages from the invokerReg topic since the beginning of time?

And, even with one controller, what happens if that controller process gets restarted? All of the invokers also have to be restarted to get the UUID -> mapping in-sync? Otherwise the controller would start handing out invoker ids starting from 0 again, which may overlap with ids already in use?

I do think allowing dynamic invoker adding and removal is a great thing. I'm just worried about things getting out of sync during regular restart, deployment, and scaling operations.

dgrove-oss:
Those are good points.  I hadn't thought carefully about a total controller restart.  Maybe I should rework to use an external store.  This is definitely cold-path code so we can afford to hit a database if we don't have a UUID=>invokerID mapping.  Then the in-memory structure is just a cache of the system of record.

ServoKvd:
I think on Controller restart, it should read back the last snapshot saved before going down.  the snapshot in this case can be an array of fixed Length with InvokerIds and their status( assigned/free/assigned-waiting-for-ack etc.)

dgrove-oss:
updated to persist the state for dynamic invoker id assignment in a Redis instance.

dgrove-oss:
pg1 2076 🏃

dgrove-oss:
pg 2076 🔴 investigating.

dgrove-oss:
pg1 2077 🏃

dgrove-oss:
pg1 2078 🔵

dgrove-oss:
rebased with master to resolve conflict in LoadBalancer.scala.

tysonnorris:
Is there an assumption that the redis instance data is persistent? If so, would it be better to use zookeeper for this state storage, which is definitely persisted (it's not clear whether the redis deployed via ansible is using persistence).

ddragosd:
BTW, what's preventing the Controller and the Invoker to use host's IP Address as ID ? Does it have to be a number ? The Invoker, in particular, is managing one host.

ddragosd:
Just a few weeks ago I experienced an issue with an OW controller that wouldn't start. It turns out it was due to Out Of Memory during startup ! Invokers were coming up, they would register in kafka, then they would be terminated due to some special conditions, but each invoker lived long enough to leave a trace in kafka. At one point the number of invokers was so big that the controller failed to start. IIRC the number was about 7M. The fix for this was to configure Kafka's `health` topic to delete messages older than `X`ms. 

Unless it's a business reason to use UUIDs for invokers, I don't find a strong argument for UUIDs otherwise. 

Since that event, our deployment moved to setting the ID from the IP of the host.

dgrove-oss:
Doesn't have to be UUIDs.  Just needs to be any stable unique alphanumeric String that can be used to distinguish invokers.  UUIDs were convenient but I don't really care if we use them or change to just be a String that might be a UUID if the deployment hasn't specified anything else.

dgrove-oss:
Rebased again to resolve conflicts with master.  

Assuming CI agrees, I think this PR could be ready for final review & merge.  Additional comments/feedback from anyone?

dgrove-oss:
pg1 2092 🏃

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -87,6 +89,45 @@ object Invoker {
...
+        case Success(_) => {
+          logger.info(this, s"sent invokerReg message with name ${invokerName} and proposedId ${proposedInvokerId}")
+        }
+        case Failure(t) => {
+          logger.error(this, s"failed to send invokerReg message: $t")
+          abort()
+        }
+      }
+      val topic = s"invokerReg-${invokerName}"
+      val consumer = msgProvider.getConsumer(config, topic, topic)
```
markusthoemmes: We should shut this consumer down after successfully receiving a message.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -87,6 +89,45 @@ object Invoker {
...
     val producer = msgProvider.getProducer(config, ec)
+
+    val proposedInvokerId = if (args.length == 1) { args(0).toInt } else { -1 }
+    val invokerName = sys.env.getOrElse("INVOKER_ASSIGNED_NAME", (() => {
+      if (proposedInvokerId == -1) {
+        logger.error(this, "Must either define INVOKER_ASSIGNED_NAME or provide proposedInvokerId on command line")
+        abort()
+      }
+      UUID().toString
+    })())
```
markusthoemmes: You don't need to use `-1` to indicate "It's not there". Scala has the `Option` type to model just that.

In this particular case, you can make use of some nice helpers to make use of Option's features to fold this nicely:

```
val proposedInvokerId: Option[Int] = args.headOption.map(_.toInt)
val invokerName: String = sys.env.getOrElse("INVOKER_ASSIGNED_NAME", {
  if(proposedInvokerId.isDefined) {
    logger.error(this, "Must either define INVOKER_ASSIGNED_NAME or provide proposedInvokerId on command line")
    abort()
  }
  UUID().toString
})
```

Note: You also don't need to build an anonymous self-calling function here, a block is sufficient. You can now model the sent message to have `Option` as well and handle the absence of a proposal in the controller accordingly.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -87,6 +89,45 @@ object Invoker {
...
+        }
+      }
+      val topic = s"invokerReg-${invokerName}"
+      val consumer = msgProvider.getConsumer(config, topic, topic)
+      val records = consumer.peek(30.second).toSeq
+      if (records.length < 1) {
+        logger.error(this, s"Failed to get a response from controller on expected topic: ${topic}")
+        abort()
+      }
+      val (_, _, _, bytes) = records(0)
```
markusthoemmes: You can enforce correctness via the compiler here:

```
val (_, _, _, bytes) = records.headOption.getOrElse {
  logger.error(this, s"Failed to get a response from controller on expected topic: ${topic}")
  abort()
}
```

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
```diff
@@ -347,3 +350,77 @@ object InvokerActor {
...
+  def receive = {
+    case m: InvokerIdRequestMessage => {
+      val assignedId =
+        if (m.proposedId != -1) {
+          // static id assignment prototcol -- always accept proposedId
+          logging.info(this, s"invoker with name ${m.instanceName} was assigned its proposedId ${m.proposedId}")
+          m.proposedId
+        } else {
+          // dynamic id assignment protocol
+          val priorId = redisClient.hget("controller:registar:idAssignments", m.instanceName)
```
markusthoemmes: It does seem like the redis client you're using has a blocking interface. If that's correct, this is quite dangerous, as threads will be blocked on our main dispatcher effectively reducing throughput while a registration is in progress.

It doesn't seem like we need an Actor in this case either as no state is encapsulated. Could the whole registration be handled by the message consumer then?

common/scala/src/main/scala/whisk/core/connector/Message.scala
```diff
@@ -118,3 +118,21 @@ object PingMessage extends DefaultJsonProtocol {
   def parse(msg: String) = Try(serdes.read(msg.parseJson))
   implicit val serdes = jsonFormat(PingMessage.apply _, "name")
 }
+
+case class InvokerIdRequestMessage(instanceName: String, proposedId: Int) extends Message {
```
markusthoemmes: As indicated below, `proposedId` could be `Option[Int]` to model possible absence of a proposal.

core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala
dgrove-oss: Good point.  When I changed things to externalize the state to Redis I could have eliminated the Actor.  I'll do that.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -87,6 +89,45 @@ object Invoker {
 
     val msgProvider = SpiLoader.get[MessagingProvider]
     val producer = msgProvider.getProducer(config, ec)
+
+    val proposedInvokerId = if (args.length == 1) { args(0).toInt } else { -1 }
+    val invokerName = sys.env.getOrElse("INVOKER_ASSIGNED_NAME", (() => {
```
rabbah: you could also read this via whisk config - we've generally avoid reading anything from sys env in the past.
it would fail if you start up the invoker without an id.

dgrove-oss:
update to address review comments from @rabbah and @markusthoemmes.

pg1 2100 🏃 

@markusthoemmes, is moving the blocking Redis calls into the handler really enough, or do I need to be finer-grained wrapping in Future { blocking { ....blocking_redis_call...} }

markusthoemmes:
@dgrove-oss unfortunately its not, I suggest we move to a non-blocking redis client (the current one is blocking by design as it seems).

https://github.com/etaty/rediscala seems like a great fit (its correctly licensed as well).

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -48,18 +50,17 @@ object Invoker {
...
+      redisHost ++
       wskApiHost ++ Map(
       dockerImageTag -> "latest",
       invokerNumCore -> "4",
       invokerCoreShare -> "2",
       invokerContainerPolicy -> "",
       invokerContainerDns -> "",
-      invokerContainerNetwork -> null)
+      invokerContainerNetwork -> null) ++
+      Map(invokerName -> null)
```
rabbah: could fold the invoker name into the previous map

```
invokerContainerNetwork -> null) ++
+      Map(invokerName -> null)
```

to 
```
invokerContainerNetwork -> null,
invokerName -> null)
```

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -85,6 +86,31 @@ object Invoker {
...
+    }.getOrElse {
+      val invokerName = config.invokerName
+      val redisClient = new RedisClient(config.redisHostName, config.redisHostPort.toInt)
+      val assignedId = redisClient.hget("controller:registar:idAssignments", invokerName).map { oldId =>
+        logger.info(this, s"invokerReg: invoker ${invokerName} was assigned its previous invokerId ${oldId}")
+        oldId.toInt
+      }.getOrElse {
+        // If key not present, incr initializes to 0 before applying increment.
+        // Convert from 1-based to 0-based invokerIds by subtracting 1 from incr's result
+        val newId = redisClient.incr("controller:registrar:nextInvokerId").map { id => id.toInt - 1 }.getOrElse {
```
rabbah: does this retry under the covers in case of contention (i.e., multiple invokers come up at once)?

dgrove-oss: my understanding from reading the docs is that this just wraps the Redis INCR command and that internally Redis is single threaded, so there cannot be a need to retry. (the various INCR operations will be applied sequentially by the single Redis thread).

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
dgrove-oss: sure.  Do you think that is better?  The previous map seems to be all docker related things so I was keeping logical separation.  Happy to do it either way.

rabbah: Thanks for the explanation. Makes sense.

dgrove-oss:
pg1 2101 🏃

dgrove-oss:
Now simplified as suggested by @markusthoemmes.  With the state for the id assignment externalized to redis, there is no need to involve the controller at all.  Each invoker can self-assign ids on startup using redis for coordination.

dgrove-oss:
pg1 2102 🔵

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -85,6 +86,41 @@ object Invoker {
...
+              }
+              .getOrElse {
+                logger.error(this, "Failed to increment invokerId")
+                abort()
+              }
+            redisClient.hset("controller:registar:idAssignments", invokerName, newId)
+            logger.info(this, s"invokerReg: invoker ${invokerName} was assigned invokerId ${newId}")
+            newId
+          }
+        redisClient.disconnect
```
markusthoemmes: Nitpick: Add `()` as this method is side-effecting. That's a common scala best-practice.

dgrove-oss: I realized it would be better to call quit instead of disconnect, so I changed that.

In either case, the compiler doesn't like it if I put ().  I think it is a property, not a method?

/Users/dgrove/code/openwhisk/openwhisk/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala:120: Boolean does not take parameters
        redisClient.quit()

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -85,6 +86,41 @@ object Invoker {
...
     }
 
+    val proposedInvokerId: Option[Int] = args.headOption.map(_.toInt)
+    val assignedInvokerId = proposedInvokerId
+      .map { id =>
+        logger.info(this, s"invokerReg: using proposedInvokerId ${id}")
+        id
+      }
+      .getOrElse {
+        val invokerName = config.invokerName
```
tysonnorris: Can invokerName come from an separate environment variable or arg? I think it would be possible from typesafe config, but don't think so with whisk Config. In our case, since the ID is an arg we can generate it with a script, but it cannot be the IP address directly since it is not an Int, or easily translated to an Int that is "small".  It would be better to use the invokerName, and leverage your approach, but we cannot do that with `config.invokerName` afaik without designating a name unique to each instance. 

This may be more of a problem of using IndexedSeq[Int] in the controller with padding where a large value for proposedInvokerId will cause havoc due to padding of invoker registration for each id lower than the proposed. 

It would be great to document this treatment of instanceid (I couldn't find any) to let devs know that using "any Int" is not acceptable; or else it would be good to have a workaround, I may be doing it wrong - we hoped to use an int derived from IP number, but the result is that, for example, the controller/invokers endpoint displays as massive quantity of invokers, all of which are offline, due to the padding. I'm not sure if it would be possible to change this type from Int to String in controllers invoker registration logic, but that would be one way to solve it, if using a referenced env var (e.g. `INVOKER_NAME=${SOME_VAR}`)  for invoker name is not possible.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -85,6 +86,41 @@ object Invoker {
       abort()
     }
 
+    val proposedInvokerId: Option[Int] = args.headOption.map(_.toInt)
+    val assignedInvokerId = proposedInvokerId
+      .map { id =>
```
rabbah: is this just a foreach?

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
rabbah: If I am reading the pr correctly that name and ordinal assigned as the id for load balancing are completely decoupled. The name could be any value/string. The increments through redis make sure you have a monotoicnally increasing value.

tysonnorris: My question is about whether `config.invokerName` has to be statically assigned as an environment variable `INVOKER_NAME=MyInvokerName`, or if it can be derived from either a system property or other environment variable, e.g. `INVOKER_NAME=${SOME_VALUE}` (where `SOME_VALUE` is another variable). Or, if it was possible to pass it as an arg, that would also be a way to use `$SOME_VALUE`, but currently it isn't.

rabbah: Ah. Thanks for the clarification. My understanding is as long as you can generate a stable name (across restarts) it should no longer matter. The reason the name has to be stable is for the redis negotiation. Every new name would bump the count otherwise. So I think the naming could be generalized accordingly.

ServoKvd: I was wondering fi a separate redis client to get the Invoker Id is an overkill.  Isn't just keeping a circular buffer of " InvokerIds& Names" in the Controller sufficient to assign a new id to the Invoker.  the same Circular buffer can log data from the Health stats and registration status. "Number of Invokers" per Controller can be set in Limits and higher-level scaling with HA replicated controllers.

rabbah: The redis instance is already deployed (for api gw) and I expect will have more uses as we evolve the control/data plane.

The proposed patch here is preserving current invariants which doesn’t require the loadbalancer to change the way the topics are mapped to invokers. But it breaks the naming and topic conflation which exists otherwise. I expect a future iteration to go further and delegate topic assignment to the loadbalancer on demand/as invokers register. This could even permit work stealing, with some imagination.

ServoKvd: Rabbah, Thank you for the clarification. So the idea of Redis is so that the Openwhisk configuration(/and)state can be reloaded, incase the controller goes down and comes back up?

rabbah: Here, the concern is primarily with the invokers going down and then recovering. Assuming a stable id, the invoker will pick up its previous active topic. As a result, there are no orphaned topics when an invoker recovers. I think @dgrove-oss has described somewhere a possible algorithm for back filing topics through the loadbalancer (where then it's closer to what you're alluding to in your previous comment).

ServoKvd: If  "stable id" == "same id as before going down", then I guess the Invoker needs a way to Identify itself ( similar  to a MAC/Ethernet Id that gets assigned an IP address on ARP resolution ),  when it comes back up; But since functions ( Invoker actions) are stateless by nature, we can just rerout(In kafka terminology I guess re-topic) the orhaned actions, instead of putting the burden on the Invoker to get its previous name - Could u point me to the algorithm that u r referring to.

rabbah: Baby steps - this is an non-invasive pr in that it preserves the current behavior but breaks the bijection between invoker name and topic id. @dgrove-oss will know better where he posted it, maybe the dev list. My recollection: the lb can detect an orphaned topic and assign it to a new invoker once it registers itself.

ServoKvd: OK,  makes sense. Thank you!, I will syncup w/ dgrove-oss et al  on the follow-up items from this PR, if any.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
markusthoemmes: Nope, get's `getOrElse`d below. Unfortunately, `Option` has no method to just apply a side-effect, passing the original value through (like `andThen` on Futures)

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
dgrove-oss: @tysonnorris re config.invokerName, this is being read from the environment variable INVOKER_NAME  (@rabbah had asked me to not read sys.env directly, but instead access the container's environment through the WhiskConfig object).   Does this work for your scenario?

dgrove-oss: @ServoKvd, I think I failed to actually post something about possible followons to this PR.  As Rodric said, this PR was meant as a small step to allow us to decouple the invoker name from the kafka topic that the invoker is assigned to service.  Useful for some of the kube work and to give some flexibility in deployment.  In the future the decoupling could allow a fancier load balancer that could more dynamically assign invokers to topics and also support an elastic pool of invokers that could grow and shrink depending on overall system load.  All details still to be worked out 😁

tysonnorris: @dgrove-oss unfortunately setting INVOKER_NAME doesn't work in our case, since the env value will be the same for each instance (when running via mesos/marathon). If it was possible to reference *another* env value as the value for INVOKER_NAME, it would work, e.g. `INVOKER_NAME=${SOME_OTHER_ENV}`, since there will be a few of these values set per instance automatically (but cannot be used to explicitly set some var unique per instance). 

I think a bigger change would be required to support this for our case, like either use TypeSafe Config (instead of whisk Config), or arg flags (e.g. you can specify either id OR name as an arg), or some extension for looking up the name that may be appropriate for particular environments (e.g. in mesos we could lookup a name from some other env vars; other envs may have other ways of associating names, etc).

dgrove-oss: @tysonnorris thanks; I understand the mesos/marathon constraints now.  They are different than kube 😞  Would it be acceptable for me to generalize in a followup PR? I have to admit I'd like to get this one merged without more iterations so we can unblock the DaemonSet PR on the kube deploy project.

I don't think it would be that hard for me to do a followup PR to allow a combination of arg flags and environment variables to be used.

tysonnorris: @dgrove-oss thanks, I understand. We can work with the current requirement to assign instance id, it is just very clunky :) If you get to a followup for the other combinations I'd be happy to test it.

dgrove-oss:
rebased to Oct 2 master.  pg4 836 🏃

dgrove-oss:
2 comments in favor of merging on dev list and no objections.  Good to go?

dgrove-oss:
rebased yet again to master to resolve conflicts.

PG1 2150 🏃 

Could we please finally merge once CI approves?

PR #2691, opened 2017-09-03
Remove redundant view which predates db split into activations and assets.

rabbah:
The views computed in the database have some redundancies. This will remove one entire view which is not necessary now that activations are not in the same database as the rest of the assets. It also removes a custom (read expensive) reduction for packages which we don't use either.


cbickel:
I really like the idea of cleaning up our views.
The only problem I see, is with already running production environments. Depending on the size of the database it could take a long time to compute the new view. In this time, the views would return nothing and customers could see errors on `action list`, ...
What do you think about a versioning our design documents?
The advantage would be, that everyone could create the v2-document in the production database, and after it is computed, you can configure OpenWhisk to use v2 of the Design doc.
After some period we could deprecate v1 of the design doc and remove it from the repository and database.
I think this would also help a lot for future Design doc changes.

PR #2694, opened 2017-09-04
Wait for view in AuthenticateTests.

markusthoemmes:
This test can be subject to a race condition, where the checks are run before the database has updated its views.


markusthoemmes:
PG2 2039 🔵

PR #2695, opened 2017-09-05
Logstore SPI

tysonnorris:
Notes: 
* builds on ContainerFactory SPI PR #2659
* This is a WIP
* The SplunkLogStore impl makes assumptions about the format of the logs - this PR does not address generating the expected format, which will require some change to the action container generating logs

LogStore SPI provides an abstraction for:
- how logs are collected from containers
- where logs are stored after collection
- how the API retrieves logs (e.g. to return directly, or display via CLI commands)

Default impl retains the same existing behavior (`docker logs` command to collect logs, stored with Activation entity into couchdb, retrieved via Activation entity).
An additional trait provided to use docker log driver, where collection is not possible, but retrieval can be done via whatever route the specific store supports, e.g. Splunk is accessible via REST API, via the SplunkLogStore impl.


rabbah:
Are we going to introduce splunk into the system or treat it as an external store that can be injected into a deployment? I'm confused by the description and the changes - thanks in advance for clarification. 

I would have expected an abstract interface for draining activation logs and a concrete implementation using the existing couchdb store. 

A word of caution: using docker logs to fetch logs is historically a recipe for docker daemon locks up and kernel panics. It doesn't work under load.

tysonnorris:
@rabbah We have no intention to introduce any specific dependency on splunk. Rather, there is an SPI for LogStore, and 2 impls are provided: the "traditional docker one", and "one based on collecting logs via docker log driver, and exposing logs via splunk API".  (This second one is easily extended to any log store that might be coupled to your docker log driver, FWIW)

Maybe the log driver makes this a bit confusing https://docs.docker.com/engine/admin/logging/overview/

In general, when using log drivers (other than the default json file) will not allow "draining activation logs", so there is no usage of the `docker logs` command at all when using a LogDriverLogStore (see `LogDriverLogStore.collectLogs`). Rather the retrieval of logs bypasses the couchdb activation entity "logs" field, and gets logs from the external system (splunk in our case).

rabbah:
@tysonnorris how shall I label this (or you can apply labels now)- is this a wip, requires discussion, ...?

tysonnorris:
@rabbah other than syncing with master and recent minor updates from #2695 I think its ok for review; will label WIP till #2695 is complete

vvraskin:
A follow up after the short discussion we had during the community interlock. 

In case we need to support both UI and CLI based logs, I think it would be advantageous to stream the logs out to some internal bus/buffer (e.g. Reddis, Elasticsearch) and keep it there for a short period of time, say 10 mins. 
Then it would be possible either to keep the previous api or create a separate component just for log streaming (a concept similar to CF loggregator), so the CLI and other log consumers could connect to it and redirect them to whatever backend is needed.

That would provide more flexibility for the end users, say everyone is able to select a particular logging backend within one OpenWhisk provider versus an OW operator defining one backend for the whole deployment. However I’m not sure if an SPI concept allows to specify several implementations to run in parallel.

rabbah:
> In case we need to support both UI and CLI based logs,

I for one don't think this is a question of "if". The immediate availability of logs is critical for a fast dev/debug cycle. You cannot be at the mercy of your downstream log drain to make progress. I think this is a unique advantage of openwhisk and we should _not_ give that up.

A short retention period - tens of minutes or hours - makes sense. You only need to scan AWS Lambda forums to see that among one of the complaints is that it takes a long time for log to appear in cloudwatch.

tysonnorris:
Thanks for the comments @vvraskin and @rabbah - there is nothing prevent an SPI impl from achieving the requirements you mention: stream to ES or redis, making logs available "immediately" (by tuning your log driver collection process and waiting on the activation to finish and log collection to complete, as it does now). The concern I have is the nobody will agree that ES or splunk or xyz is the best place to store logs, and mostly I think everyone agrees that couchdb is not the best place. WDYT? Would it help if there was a different impl available? Its not clear how you'd like to see this PR changed.

rabbah:
I haven't had a chance to review the PR - I was just compelled to clarify one aspect of the previous comment.

rabbah:
Just tried to scan quickly: Do you plan to rebase this to highlight the main changes? I think some of the commits here are not relevant to this PR but are carried over from previous staging commits?

tysonnorris:
@rabbah yes, will do that asap

tysonnorris:
squashed commits; still todo:
- [ ] additional tests
- [ ] remove some hard coded date logic in splunk impl

tysonnorris:
@rabbah @vvraskin any comments on this? I'm wondering what are your thoughts on if (and how) this should change to be acceptable.

ddragosd:
:+1: On the CI part, It would be great to have a separate conversation where we also discuss holistically how to run integration tests isolated from unit tests. If we do this, the builds would also be faster, and we would be able to test such implementations with other systems (Mesos, Splunk, MongoDB, etc). Such changes might  fit better in separate PR(s)?

tysonnorris:
@markusthoemmes thanks for the changes - LGTM

markusthoemmes:
PG2 2269 🔵

PR #2696, opened 2017-09-05
Bump replication timeout, abort test with correct message

markusthoemmes:
Replication in CouchDb is hard to control, thus increasing the timeout should not be harmful. The additional assert will make it easier to grasp why a test failed in the future.


markusthoemmes:
PG1 2019 🔵

PR #2697, opened 2017-09-05
Increase CouchDB to version 2.0

cbickel:
If the cache is disabled in the controller, the following tests are failing:
- Wsk actions should create, and get an action to verify parameter and annotation parsing
- Wsk packages should create, and get a package to verify parameter and annotation parsing
- Wsk triggers should create, and get a trigger to verify parameter and annotation parsing

The reason is, that the number `12345678912.123456789012` is rounded to `12345678912.123457` in the controller.
On writing `12345678912.123457` into CouchDB, the value `12345678912.123456955` is saved.
On getting the document again from CouchDB, the controller returns `12345678912.123456955`. That's why the test is failing.
The cache in the controller has hidden this issue until now.
Increasing to version 2.0 fixes this issue.

### DIFF
--- a/ansible/roles/couchdb/tasks/deploy.yml
+++ b/ansible/roles/couchdb/tasks/deploy.yml
@@ -1,6 +1,10 @@
 ---
 # This role will run a CouchDB server on the db group
 
+- name: "Set node name to couchdb{{ groups['db'].index(inventory_hostname) }}"
+  set_fact:
+    nodeName: "couchdb{{ groups['db'].index(inventory_hostname) }}"
+
 - name: check if db credentials are valid for CouchDB
   fail: msg="The db provider in your {{ inventory_dir }}/group_vars/all is {{ db_provider }}, it has to be CouchDB, pls double check"
   when: db_provider != "CouchDB"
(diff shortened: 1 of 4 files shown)


ansible/roles/couchdb/tasks/deploy.yml
```diff
@@ -40,18 +48,9 @@
...
     timeout: 60
 
-- name: create admin user
-  uri:
-    url: "{{ db_protocol }}://{{ ansible_host }}:{{ db_port }}/_config/admins/{{ db_username }}"
-    method: PUT
-    body: >
-        "{{ db_password }}"
-    body_format: json
-    status_code: 200
```
rabbah: this is no longer needed?
is that sub summed by the `env` setting above?

tests/src/test/scala/whisk/core/database/test/ReplicatorTests.scala
```diff
@@ -52,6 +52,7 @@ class ReplicatorTests
 
   val replicatorClient =
     new ExtendedCouchDbRestClient(dbProtocol, dbHost, dbPort.toInt, dbUsername, dbPassword, "_replicator")
+  replicatorClient.createDb()
```
rabbah: why is this new?

ansible/roles/couchdb/tasks/deploy.yml
cbickel: Yes, the admin user can be created with the env as well.

tests/src/test/scala/whisk/core/database/test/ReplicatorTests.scala
cbickel: In this version of CouchDB, the `_replication`-DB does not exist. So the initial cleanup fails with 404.

rabbah:
Thanks for the clarification. 👍

PR #2698, opened 2017-09-05
Fix false assertion, refactor CacheInvalidationTests

markusthoemmes:
If we need to retry the condition, `deleteAction` will always fail because the if it succeeded once it will never return `200` again.


rabbah:
Do you think this addresses https://github.com/apache/incubator-openwhisk/issues/2692

markusthoemmes:
@rabbah I think so, yes.

markusthoemmes:
PG1 2019 🔵

PR #2701, opened 2017-09-05
Fixed the docker image prefix for vagrant OW installation

sayanh:


csantanapr:
Hi @sayanh this is working as design.
This will brake, for vagrant on first run when creating the VM it will use the images from dockerhub, this makes it run very fast.

Then after this is user wants to use the vagrant machine to do development on whisk, then running `./gradlew distDocker` will use the prefix `whisk` and the default value use when deploying with ansible or wsdev it will use `whisk` as the prefix.

rabbah:
@csantanapr perhaps a comment in that file is warranted.

Issue #2706, opened 2017-09-06
Syntax error in tasks/patchKernel.yml?


kevhouck:
## Environment details:

* native distributed deployment on ubuntu 14.04.1 LTS (GNU/Linux 3.13.0-117-generic x86_64)

## Steps to reproduce the issue:

1. Started the [Ubuntu setup](https://github.com/apache/incubator-openwhisk/blob/master/tools/ubuntu-setup/README.md) until it mentions the distributed setup
2. Following the [distributed setup](https://github.com/apache/incubator-openwhisk/blob/master/ansible/README_DISTRIBUTED.md), at step:
`ansible-playbook -i environments/distributed prereq_build.yml`

## Provide the expected results and outputs:

It should complete successfully

## Provide the actual results and outputs:

```
ERROR! no action detected in task. This often indicates a misspelled module name, or incorrect module path.

The error appears to have been in '/users/khouck/openwhisk/ansible/roles/whiskhost/tasks/patchKernel.yml': line 35, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: wait until host comes back up, timeout after 10 minutes
  ^ here


The error appears to have been in '/users/khouck/openwhisk/ansible/roles/whiskhost/tasks/patchKernel.yml': line 35, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: wait until host comes back up, timeout after 10 minutes
  ^ here
```

## Additional information you deem important:
* Manually created hosts file, but `ansible all -i environments/local -m ping` is successful, this seems like a syntax error with ansible
* Could it be the wrong version of ansible is installed when running `(cd tools/ubuntu-setup && ./all.sh)`

dubeejw:
Closing as fixed.

PR #2707, opened 2017-09-07
fix syntax error in patchKernel.yml

devbv:
#2706


csantanapr:
Thanks for the fix @devbv

PR #2669, opened 2017-08-29
initial work on no root ansible

csantanapr:
This allows to deploy with no sudo/root access.

PR #2709, opened 2017-09-07
Fix Ansible playbook problems with missing variable interpolation

sven-lange-last:
Surround variables introduced with PR #2669 with double curly braces to force interpolation.
We observed Ansible playbook runs where `become` was not active because variables
were not interpolated such that the value was `false`.

### DIFF
--- a/ansible/roles/cli/tasks/clean.yml
+++ b/ansible/roles/cli/tasks/clean.yml
@@ -4,4 +4,4 @@
   file:
     path: "{{ cli.nginxdir }}"
     state: absent
-  become: cli.dir.become
+  become: "{{ cli.dir.become }}"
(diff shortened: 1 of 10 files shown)


sven-lange-last:
PG 2 / 2045

rabbah:
@sven-lange-last my fault for not catching it - thanks for the patch.
i think a pg isn't helpful since twice this has bit in the main build.

sven-lange-last:
@rabbah If `become: false` is acceptable in an environment, a test won't catch this problem. And in a clean environment, `become: false` is always acceptable.

csantanapr:
@sven-lange-last it's odd that it was silently evaluating to false, and giving an error about the  specific variable.

csantanapr:
PG2 2045 🔵

PR #2714, opened 2017-09-07
Disable explicit retry on container http connection.

rabbah:
Remove unused type.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala
@@ -96,7 +96,6 @@ protected[core] object ActivationResponse extends DefaultJsonProtocol {
    * Class of errors for invoker-container communication.
    */
   protected[core] sealed abstract class ContainerConnectionError
-  protected[core] case class NoHost() extends ContainerConnectionError
   protected[core] case class ConnectionError(t: Throwable) extends ContainerConnectionError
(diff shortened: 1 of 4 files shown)


rabbah:
pg1/2042 :large_blue_circle:

common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala
```diff
@@ -96,7 +96,6 @@ protected[core] object ActivationResponse extends DefaultJsonProtocol {
    * Class of errors for invoker-container communication.
    */
   protected[core] sealed abstract class ContainerConnectionError
-  protected[core] case class NoHost() extends ContainerConnectionError
```
rabbah: @markusthoemmes if i recall correctly, this is left over from old invoker.

core/invoker/src/main/scala/whisk/core/containerpool/docker/HttpUtils.scala
```diff
@@ -133,6 +136,7 @@ protected[core] class HttpUtils(hostname: String, timeout: FiniteDuration, maxRe
   private val connection = HttpClientBuilder.create
     .setDefaultRequestConfig(httpconfig)
     .useSystemProperties()
+    .disableAutomaticRetries()
```
rabbah: we don't use the automatic retries anyway and they don't apply for our operations which use `POST` that is not idempotent.

PR #2717, opened 2017-09-08
Ensure that base64 encoded '+json' bodies are accepted and decoded.

### LINKED ISSUE #2630: JSON encoding is inconsistent for web actions

rabbah:
For any base64 encoded string where the content-type is set to a valid
media type with `+json` at the end, then we need to decode it.

If the response from the web action is a JSON dictionary in the body and
the content-type header is set to a `+json` media type, then send this
is acceptable.

Do not whitelist media types in the content-type header
Instead, just ensure that they are of a valid format.

Replaces #2646.
Fixes #2630.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
@@ -264,8 +264,9 @@ protected[core] object WhiskWebActionsApi extends Directives {
   }
 
   /**
-   * Finds the content-type in the header list and maps it to a known media type. If it is not
-   * recognized, construct a failure with appropriate message.
+   * Finds the content-type in the header list and ensures that it is a valid format. If it is not
+   * valid, construct a failure with appropriate message.
+   * If the content-type header is missing, then return the supplied defaultType
    */
   private def findContentTypeInHeader(headers: List[RawHeader],
                                       transid: TransactionId,
@@ -273,16 +274,15 @@ protected[core] object WhiskWebActionsApi extends Directives {
     headers.find(_.lowercaseName == `Content-Type`.lowercaseName) match {
       case Some(header) =>
         MediaType.parse(header.value) match {
-          case Right(mediaType: MediaType) =>
-            // lookup the media type specified in the content header to see if it is a recognized type
-            MediaTypes.getForKey(mediaType.mainType -> mediaType.subType).map(Success(_)).getOrElse {
-              // this is a content-type that is not recognized, reject it
-              Failure(RejectRequest(BadRequest, Messages.httpUnknownContentType)(transid))
-            }
-          case _ => Failure(RejectRequest(BadRequest, Messages.httpUnknownContentType)(transid))
+          case Right(mediaType: MediaType) => Success(mediaType)
+          case _                           => Failure(RejectRequest(BadRequest, Messages.httpUnknownContentType)(transid))
         }
       case None => Success(defaultType)
     }
+  }
+
+  private def isJsonFamily(mt: MediaType) = {
+    mt == `application/json` || Option(mt.value).map(_.endsWith("+json")).exists(identity)
   }
 
   private def interpretHttpResponseAsJson(code: StatusCode,
@@ -290,9 +290,19 @@ protected[core] object WhiskWebActionsApi extends Directives {
                                           js: JsValue,
                                           transid: TransactionId) = {
     findContentTypeInHeader(headers, transid, `application/json`) match {
-      case Success(mediaType) if (mediaType == `application/json`) =>
+      case Success(mediaType) if mediaType == `application/json` =>
         respondWithHeaders(removeContentTypeHeader(headers)) {
           complete(code, js)
+        }
+
+      case Success(mediaType) if isJsonFamily(mediaType) =>
+        respondWithHeaders(removeContentTypeHeader(headers)) {
+          complete(
+            code,
+            HttpEntity(
+              ContentType(
+                MediaType.customWithFixedCharset(mediaType.mainType, mediaType.subType, HttpCharsets.`UTF-8`)),
+              js.prettyPrint))
         }
 
       case _ =>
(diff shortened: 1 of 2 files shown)


rabbah:
PG2/2049 🔵

rabbah:
@akrabat are you OK with the changes?
@dubeejw @markusthoemmes would you kindly lend 👀

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -264,37 +264,47 @@ protected[core] object WhiskWebActionsApi extends Directives {
...
-          case _ => Failure(RejectRequest(BadRequest, Messages.httpUnknownContentType)(transid))
+          case Right(mediaType: MediaType) => Success(mediaType)
+          case _                           => Failure(RejectRequest(BadRequest, Messages.httpUnknownContentType)(transid))
         }
       case None => Success(defaultType)
     }
   }
 
+  private def isJsonFamily(mt: MediaType) = {
+    mt == `application/json` || Option(mt.value).map(_.endsWith("+json")).exists(identity)
```
markusthoemmes: Can `mt.value` be `null`? 

I don't think it can. Should it then just be 

```scala
mt == `application/json` || mt.value.endsWith("+json")
```

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -264,37 +264,47 @@ protected[core] object WhiskWebActionsApi extends Directives {
...
       case Some(header) =>
         MediaType.parse(header.value) match {
-          case Right(mediaType: MediaType) =>
-            // lookup the media type specified in the content header to see if it is a recognized type
-            MediaTypes.getForKey(mediaType.mainType -> mediaType.subType).map(Success(_)).getOrElse {
-              // this is a content-type that is not recognized, reject it
-              Failure(RejectRequest(BadRequest, Messages.httpUnknownContentType)(transid))
-            }
-          case _ => Failure(RejectRequest(BadRequest, Messages.httpUnknownContentType)(transid))
+          case Right(mediaType: MediaType) => Success(mediaType)
```
markusthoemmes: To clarify: This skips the (arguably useless) doublecheck of the MediaType against akka-http's list of MediaTypes?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
rabbah: I admit that I didn't check - I carried over the unapply from the previous branch but turned it into an action. It does look like the parser will not do that.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
rabbah: It removes the check for media types which are understood/registered by akka. So for custom media types it is now acceptable where it would have otherwise failed. This is why the test which checks for an unknown media type changed.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -304,11 +317,13 @@ protected[core] object WhiskWebActionsApi extends Directives {
     findContentTypeInHeader(headers, transid, `text/html`).flatMap { mediaType =>
       val ct = ContentType(mediaType, () => HttpCharsets.`UTF-8`)
       ct match {
-        case _: ContentType.Binary | ContentType(`application/json`, _) =>
-          // base64 encoded json response supported for legacy reasons
-          Try(Base64.getDecoder().decode(str)).map(HttpEntity(ct, _))
+        // base64 encoded json response supported for legacy reasons
```
dubeejw: Comment seems wrong here.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -264,37 +264,50 @@ protected[core] object WhiskWebActionsApi extends Directives {
...
-      case Success(mediaType) if (mediaType == `application/json`) =>
+      // use the default akka-http response marshaler for standard application/json
+      case Success(mediaType) if mediaType == `application/json` =>
         respondWithHeaders(removeContentTypeHeader(headers)) {
           complete(code, js)
         }
 
+      // for all other json-family content-type, explicitly marshal the response;
+      // the order of the case statement matters; isJsonFamily returns true for application/json
+      case Success(mediaType) if isJsonFamily(mediaType) =>
```
dubeejw: `isJsonFamily()` checks for `application/json` here even though the type will never be `application/json` at this point.

rabbah: see my comment above - also this method is used elsewhere where both types are plausible.
"use the default akka-http response marshaler for standard application/json'

i debated collapsing both case matches but decided it was cleaner this way - i'm receptive to input.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
rabbah: is this clearer?

```scala
// base64 encoded json will appear as non-binary but it is excluded here for legacy reasons
```

rabbah: is this clearer?

// base64 encoded json will appear as non-binary but it is excluded here for legacy reasons

rabbah:
Don't merge this. @akrabat I'm afraid if this is merged I will get the commit. This just happened on another PR I rebased and merged eventhough it shouldn't have been mine.

PR #2729, opened 2017-09-11
add support for multiple versions of ubuntu

strajansebastian:
ubuntu 16.04 and any other distros than trusty may encounter some problems because of hardcoded values

### DIFF
--- a/tools/ubuntu-setup/docker.sh
+++ b/tools/ubuntu-setup/docker.sh
@@ -4,7 +4,7 @@ set -x
 
 sudo apt-get -y install apt-transport-https ca-certificates
 sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 36A1D7869245C8950F966E92D8576A8BA88D21E9
-sudo sh -c "echo deb https://apt.dockerproject.org/repo ubuntu-trusty main  > /etc/apt/sources.list.d/docker.list"
+sudo sh -c "echo deb https://apt.dockerproject.org/repo ubuntu-$(lsb_release -c -s) main  > /etc/apt/sources.list.d/docker.list"
 sudo apt-get -y update -qq
 
 sudo apt-get purge lxc-docker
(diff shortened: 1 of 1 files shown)


rabbah:
Thanks @strajansebastian for the suggestion. Do you have an Apache ICLA already?

strajansebastian:
@rabbah just sent to the secretary@apache.org the signed ICLA. Do I need to send you something else in order for the process to be complete?

strajansebastian:
any update on this?

tools/ubuntu-setup/docker.sh
```diff
@@ -14,7 +14,7 @@ sudo apt-cache policy docker-engine
 sudo apt-get -y install linux-image-extra-$(uname -r)
 
 # DOCKER
-sudo apt-get install -y --force-yes docker-engine=1.12.0-0~trusty
+sudo apt-get install -y --force-yes docker-engine
```
markusthoemmes: This is a bit dangerous. While OpenWhisk works on newer docker versions perfectly fine, it is not CIed as of today. We got bitten by docker version regressions so we very carefully update and today recommend using 1.12 for anything close to a production environment.

strajansebastian: The reason I did that was in order to bring the latest stable version provided by the defined repository for a specific Ubuntu OS.
Docker seems to change really frequently and if I remember well on the Ubuntu 16.04 VM that I made the tests that specific version wasn't available at that time in the repository only a newer one (docker-engine 17.05) .

I remade just know a test on Ubuntu 16.04 and it seems that the key repo isn't working anymore at least for Ubuntu 16.04, haven't tested on a Ubuntu 14.04 server yet.
According to the docker documentation for Ubuntu trusty/xenial and later the docker-engine package seems to have become docker-ce. https://docs.docker.com/engine/installation/linux/docker-ce/ubuntu/ 
Will come back with an update for this script if that is ok with you.

rabbah: Also related to https://github.com/apache/incubator-openwhisk/pull/2947#issuecomment-343937990 @csantanapr

rabbah:
Related to https://github.com/apache/incubator-openwhisk/issues/295.

csantanapr:
This is now Fix by https://github.com/apache/incubator-openwhisk/pull/3358

PR #2730, opened 2017-09-11
Reduce memory consumption for invocations of cached actions

dubeejw:
Creates new datatype called WhiskActionMetadata that drops code property from its Exec data type for action invocations. This prevents action code from being loaded into memory.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
+++ b/common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
@@ -36,6 +36,7 @@ import whisk.core.entity.DocInfo
 import whisk.core.entity.DocRevision
 import whisk.core.entity.WhiskDocument
 import whisk.http.Messages
+import whisk.core.entity.DocumentReader
 
 /**
  * Basic client to put and delete artifacts in a data store.
@@ -58,7 +59,8 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
   implicit system: ActorSystem,
   val logging: Logging,
   jsonFormat: RootJsonFormat[DocumentAbstraction],
-  materializer: ActorMaterializer)
+  materializer: ActorMaterializer,
+  docReader: DocumentReader)
     extends ArtifactStore[DocumentAbstraction]
     with DefaultJsonProtocol {
 
@@ -223,7 +225,13 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
       e match {
         case Right(response) =>
           transid.finished(this, start, s"[GET] '$dbName' completed: found document '$doc'")
-          val asFormat = jsonFormat.read(response)
+
+          val asFormat = try {
+            docReader.read(ma, response)
+          } catch {
+            case e: Exception => jsonFormat.read(response)
+          }
+
           if (asFormat.getClass != ma.runtimeClass) {
             throw DocumentTypeMismatchException(
               s"document type ${asFormat.getClass} did not match expected type ${ma.runtimeClass}.")
--- a/common/scala/src/main/scala/whisk/core/entity/Exec.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Exec.scala
@@ -45,13 +45,17 @@ import whisk.core.entity.size.SizeString
  *   main  : name of the entry point function, when using a non-default value (for Java, the name of the main class)" }
  */
 sealed abstract class Exec extends ByteSizeable {
-  override def toString = Exec.serdes.write(this).compactPrint
+  override def toString: String = Exec.serdes.write(this).compactPrint
 
   /** A type descriptor. */
   val kind: String
 
   /** When true exec may not be executed or updated. */
   val deprecated: Boolean
+}
+
+sealed abstract class ExecMetaDataBase extends Exec {
+  override def toString: String = ExecMetaDataBase.serdes.write(this).compactPrint
 }
 
 /**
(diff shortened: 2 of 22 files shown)


dubeejw:
Thinking that all of the duplicate data types can be used as bases to extend the corresponding types that require the code property.

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -150,6 +157,18 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](
           ErrorLevel))
   }
 
+  def newRead[A](ma: Manifest[A], value: JsValue) = {
```
dubeejw: TODO: move this.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -126,6 +178,31 @@ protected[core] case class CodeExecAsAttachment(manifest: RuntimeManifest,
...
+    extends CodeExec2[Attachment[String]] {
+  override val kind = manifest.kind
+  override val image = manifest.image
+  override val sentinelledLogs = manifest.sentinelledLogs.getOrElse(true)
+  override val deprecated = manifest.deprecated.getOrElse(false)
+  override val pull = false
+  override lazy val binary = true
+  override def codeAsJson = "".toJson //code.toJson
+
+  /*
```
dubeejw: TODO: delete commetted out code if not needed.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -144,12 +221,33 @@ protected[core] case class BlackBoxExec(override val image: ImageName,
...
 }
 
+protected[core] case class BlackBoxExec2(override val image: ImageName,
+                                        override val entryPoint: Option[String],
+                                        val native: Boolean)
+    extends CodeExec2[Option[String]] {
+  override val kind = Exec2.BLACKBOX
+  override val deprecated = false
+  override def codeAsJson = "".toJson //code.toJson
+  override lazy val binary = false //TODO code map { Exec.isBinaryCode(_) } getOrElse false
```
dubeejw: Can't determine binary type for actions with code property removed.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -278,3 +376,132 @@ protected[core] object Exec extends ArgNormalizer[Exec] with DefaultJsonProtocol
...
+        val main = a.entryPoint.map("main" -> JsString(_))
+        JsObject(base ++ main)
+
+      case s @ SequenceExec2(comp) =>
+        JsObject("kind" -> JsString(s.kind), "components" -> comp.map(_.qualifiedNameWithLeadingSlash).toJson)
+
+      case b: BlackBoxExec2 =>
+        val base =
+          Map("kind" -> JsString(b.kind), "image" -> JsString(b.image.publicImageName), "binary" -> JsBoolean(b.binary))
+        //val code = b.code.filter(_.trim.nonEmpty).map("code" -> JsString(_))
```
dubeejw: TODO: delete.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -222,29 +222,49 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with
    */
   override def activate(user: Identity, entityName: FullyQualifiedEntityName, env: Option[Parameters])(
     implicit transid: TransactionId) = {
+
+    /*parameter('blocking ? false, 'result ? false, 'timeout.as[FiniteDuration] ? WhiskActionsApi.maxWaitForBlockingActivation) { (blocking, result, waitOverride) =>
```
dubeejw: TODO: delete.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -222,29 +222,49 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with
...
+        logging.info(this, "here1")
+        getEntity(WhiskActionMini, entityStore, entityName.toDocId, Some {
+          act: WhiskActionMini =>
+            logging.info(this, "here2")
             // resolve the action --- special case for sequences that may contain components with '_' as default package
             val action = act.resolve(user.namespace)
-            onComplete(entitleReferencedEntities(user, Privilege.ACTIVATE, Some(action.exec))) {
+            onComplete(entitleReferencedEntities2(user, Privilege.ACTIVATE, Some(action.exec))) {
               case Success(_) =>
+                logging.info(this, "here3")
```
dubeejw: TODO: remove extra logging.

dubeejw:
TODO: need test in ActionsApiTests that ensures code property does not exist.

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
dubeejw: Test failures are a result of those hack. Need to figure out a better way.

dubeejw:
PG2 2104 ⏳

dubeejw:
Profiling of action invocations using the changes provided in this PR. The action is a 34 MB zipped Node action. 

screen shot 2017-09-26 at 2 46 09 pm

First action invocation is a cache miss causing the controller to fetch the action from the DB. This means the code associated with the action is download into the controller causing a spike in the used heap. Garbage collection is then performed to lower the used heap. The same action is then invoked three times without the previous heap increase since the action without the code property is cached, which does not cause a fetch from the DB.

markusthoemmes:
@dubeejw nice view, can you provide a profiling result from before so we can compare?

dubeejw:
PG3 1136 🔵

dubeejw:
@rabbah, want to review?

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -45,7 +45,7 @@ import whisk.core.entity.size.SizeString
  *   main  : name of the entry point function, when using a non-default value (for Java, the name of the main class)" }
  */
 sealed abstract class Exec extends ByteSizeable {
-  override def toString = Exec.serdes.write(this).compactPrint
+  override def toString: String = Exec.serdes.write(this).compactPrint
```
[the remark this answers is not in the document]
dubeejw: Would have to investigate to see when this method is used.

PR #2735, opened 2017-09-12
Remove redundant view which predates db split into activations and assets.

rabbah:
The views computed in the database have some redundancies. This will remove one entire view which is not necessary now that activations are not in the same database as the rest of the assets. It also removes a custom (read expensive) reduction for packages which we don't use either.

Replaces https://github.com/apache/incubator-openwhisk/pull/2691 - rebased to master and reformated code. Preserving old prs for the incremental commits and Travis ✅.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
@@ -19,12 +19,16 @@ package whisk.core.entity
 
 import java.time.Instant
 
+import scala.concurrent.Future
 import scala.util.Try
 
+import spray.json._
 import spray.json.DefaultJsonProtocol
 import spray.json.DefaultJsonProtocol._
-import spray.json._
+import whisk.common.TransactionId
+import whisk.core.database.ArtifactStore
 import whisk.core.database.DocumentFactory
+import whisk.core.database.StaleParameter
 
 /**
  * A WhiskActivation provides an abstraction of the meta-data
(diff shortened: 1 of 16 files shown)


rabbah:
pg4/796 🔵

common/scala/src/main/scala/whisk/core/entity/WhiskStore.scala
```diff
@@ -169,30 +169,26 @@ object WhiskEntityQueries {
   val TOP = "\ufff0"
   val WHISKVIEW = "whisks"
   val ALL = "all"
-  val ENTITIES = "entities"
```
rabbah: @cbickel i removed the view change from this pr but eliminated the reference to the view from the core.

common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
```diff
@@ -119,4 +124,28 @@ object WhiskActivation
...
+   *
+   * @return list of records as JSON object if docs parameter is false, as Left
+   *         and a list of the WhiskActivations if including the full record, as Right
+   */
+  def listActivationsMatchingName(db: ArtifactStore[WhiskActivation],
+                                  namespace: EntityPath,
+                                  name: EntityName,
+                                  skip: Int,
+                                  limit: Int,
+                                  docs: Boolean = false,
```
cbickel: What do think about renaming this to `includeDocs`.
Then it would be consistent with `listCollectionInNamespace`.

rabbah: yes. I missed it.

rabbah: Do you mind if I address this in the next dependent pr https://github.com/apache/incubator-openwhisk/pull/2760? There’s another docs to rename there.

cbickel: No problem.

PR #2737, opened 2017-09-12
Removing 'Swift 3.1.1' from build.

markusthoemmes:
Swift 3.1.1 build is safe to remove, since we're not running any tests there anyway and it buys 4 minutes of time in the process.

Kicked off a dev-list discussion here to solve the underlying root cause more efficiently: https://lists.apache.org/thread.html/f81150a4f19c3c6f0a4a4505ca13ac044e7b85bda01faf3f29f64614@%3Cdev.openwhisk.apache.org%3E


rabbah:
Travis for this pr at 40 minutes. Master is closer to 47-49.

PR #2738, opened 2017-09-13
increase the poll period for tests to survive slower indexing ti…

jeremiaswerner:
…mes in couchdb

### DIFF
--- a/tests/src/test/scala/common/WskTestHelpers.scala
+++ b/tests/src/test/scala/common/WskTestHelpers.scala
@@ -26,7 +26,6 @@ import scala.util.Failure
 import scala.util.Try
 import scala.concurrent.duration.Duration
 import scala.concurrent.duration.DurationInt
-import scala.language.postfixOps
 
 import spray.json._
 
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/system/basic/WskRuleTests.scala
```diff
@@ -27,6 +27,8 @@ import common.WskTestHelpers
 import spray.json._
 import spray.json.DefaultJsonProtocol._
 import java.time.Instant
+import scala.concurrent.duration.DurationInt
+import scala.language.postfixOps
```
markusthoemmes: Please drop `postfixOps` and use standard notation of `60.seconds`. You'll get super weird compile errors when using `postfixOps`.

To import duration, just do `import scala.concurrent.duration._`

markusthoemmes:
Would it be worth to fix this more globally? You can change the wait time for the polling globally in `WskTestHelpers.scala` and change the default value for `withActivationsFromEntity`

jeremiaswerner:
🔵 PG1/2048

PR #2739, opened 2017-09-13
WIP: make test var lazy for other components' tests to be able to overwrite

drcariel:
not sure if this is necessary entirely, doing some testing atm


dubeejw:
@drcariel, I don't think this one is necessary, so closing.

PR #2741, opened 2017-09-13
Display content-type in __ow_headers for web actions

dubeejw:
Manually adds content-type sent to web action into the web action's __ow_headers.

Closes https://github.com/apache/incubator-openwhisk/issues/2690

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
@@ -371,11 +371,14 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
     `Access-Control-Allow-Methods`(OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH),
     `Access-Control-Allow-Headers`(`Authorization`.name, `Content-Type`.name))
 
-  private def contentTypeFromEntity(entity: HttpEntity) =
+  private def contentTypeFromEntity(entity: HttpEntity) = {
+    val noContentType = ContentType(MediaType.customBinary("none", "none", MediaType.NotCompressible))
+
     entity.contentType match {
-      case ct if ct == ContentType(MediaType.customBinary("none", "none", MediaType.NotCompressible)) => Seq()
-      case ct                                                                                         => Seq(RawHeader(`Content-Type`.lowercaseName, ct.toString))
+      case ct if ct == noContentType => Seq()
+      case ct                        => Seq(RawHeader(`Content-Type`.lowercaseName, ct.toString))
     }
+  }
 
   /** Extracts the HTTP method, headers, query params and unmatched (remaining) path. */
   private val requestMethodParamsAndPath = {
--- a/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
@@ -371,13 +371,9 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
     `Access-Control-Allow-Methods`(OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH),
     `Access-Control-Allow-Headers`(`Authorization`.name, `Content-Type`.name))
 
-  private def contentTypeFromEntity(entity: HttpEntity) = {
-    val noContentType = ContentType(MediaType.customBinary("none", "none", MediaType.NotCompressible))
-
-    entity.contentType match {
-      case ct if ct == noContentType => Seq()
-      case ct                        => Seq(RawHeader(`Content-Type`.lowercaseName, ct.toString))
-    }
+  private def contentTypeFromEntity(entity: HttpEntity) = entity.contentType match {
+    case ct if ct == ContentTypes.NoContentType => None
+    case ct                                     => Some(RawHeader(`Content-Type`.lowercaseName, ct.toString))
   }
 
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -1430,17 +1439,19 @@ trait WebActionsApiTests extends ControllerTestCommon with BeforeAndAfterEach wi
...
+        s"$testRoutePath/$systemId/proxy/raw_export_c.json?$queryString",
+        HttpEntity(ContentTypes.`application/json`, str)) ~> Route.seal(routes(creds)) ~> check {
         status should be(OK)
         val response = responseAs[JsObject]
         response shouldBe JsObject(
           "pkg" -> s"$systemId/proxy".toJson,
           "action" -> "raw_export_c".toJson,
           "content" -> metaPayload(
             Post.method.name.toLowerCase,
-            Map(webApiDirectives.body -> str.toJson, webApiDirectives.query -> queryString.toJson).toJson.asJsObject,
```
dubeejw: I replicated the test manually using the CLI without using the web action changes in this PR. The body is actually Base64 encoded in the web action response. This test expected the body to be a JSON string instead. After changing the POST request here to use an `HttpEntity`, the body was encoded just like in my manual testing.

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -470,7 +470,8 @@ trait WebActionsApiTests extends ControllerTestCommon with BeforeAndAfterEach wi
           val content = JsObject("extra" -> "read all about it".toJson, "yummy" -> true.toJson)
           val p = if (path.endsWith("/")) "/" else ""
           invocationsAllowed += 1
-          m(s"$testRoutePath/$path", content) ~> Route.seal(routes(creds)) ~> check {
+          m(s"$testRoutePath/$path", HttpEntity(ContentTypes.`application/json`, content.compactPrint)) ~> Route.seal(
```
dubeejw: Tried to get rid of `compactPrint`, could not get the tests to compile by just passing the `JsObject`.

dubeejw:
PG2 2067 🔵

dubeejw:
@rabbah, ready for review.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -371,13 +371,22 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
     `Access-Control-Allow-Methods`(OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH),
     `Access-Control-Allow-Headers`(`Authorization`.name, `Content-Type`.name))
 
+  private def contentTypeFromEntity(entity: HttpEntity) = {
+    val noContentType = ContentType(MediaType.customBinary("none", "none", MediaType.NotCompressible))
```
rabbah: there's no singleton representation of the no content-type value that we can reuse and which akka-http provides?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -371,13 +371,22 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
...
+
+    entity.contentType match {
+      case ct if ct == noContentType => Seq()
```
rabbah: scala tip, this could be None and Some, which you can add to a seq because they're iterable.

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
rabbah: why do we need to set the content type explicitly on the entity?

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
rabbah: the test was incorrect?

dubeejw: Yeah, test was wrong.

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -1354,9 +1356,8 @@ trait WebActionsApiTests extends ControllerTestCommon with BeforeAndAfterEach wi
        *
        */
 
-      Post(s"$testRoutePath/$systemId/proxy/export_c.json", str) ~> addHeader(
-        "Content-type",
-        ContentTypes.`text/html(UTF-8)`.value) ~> Route.seal(routes(creds)) ~> check {
+      Post(s"$testRoutePath/$systemId/proxy/export_c.json", HttpEntity(ContentTypes.`text/html(UTF-8)`, str)) ~> Route
+        .seal(routes(creds)) ~> check {
```
rabbah: why do you need to set the content type explicitly here on the entity?

dubeejw: Just maintaining the previous functionality as the content-type use to be set with `addHeader()`. If the content-type isn't set explicitly, the received content-type ends up being `text/plain; charset=UTF-8` instead of text/html; charset=UTF-8`.

Got rid of all the calls that didn't need the content-type set explicitly.

PR #2743, opened 2017-09-14
allow to override runtimesManifest

### LINKED ISSUE #2742: Add ability to override runtimesManifest in ansible
There is no easy way to override the runtimesManifest via target environment in ansible.

The use case is if I want to deploy whisk with only supporting a specific runtime kinds, or add additional custom ones like a new kind `nodejs-acme:8.5` 😉

csantanapr:
Closes #2742

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -30,7 +30,7 @@ whisk:
 #   runtimes: set of language runtime families grouped by language (e.g., nodejs, python)
 #   blackboxes: list of pre-populated docker action images as "name" with optional "prefix" and "tag"
 #
-runtimesManifest:
+runtimesManifestDefault:
   defaultImagePrefix: "openwhisk"
   defaultImageTag: "latest"
   runtimes:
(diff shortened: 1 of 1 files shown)


ansible/group_vars/all
```diff
@@ -92,6 +92,8 @@ runtimesManifest:
   blackboxes:
     - name: "dockerskeleton"
 
+runtimesManifest: "{{ runtimes_Manifest | default(runtimesManifestDefault) }}"
```
rabbah: Why the capital M after underscore?

csantanapr: I like MegaMan 📣

ansible/group_vars/all
```diff
@@ -30,7 +30,7 @@ whisk:
 #   runtimes: set of language runtime families grouped by language (e.g., nodejs, python)
 #   blackboxes: list of pre-populated docker action images as "name" with optional "prefix" and "tag"
 #
-runtimesManifest:
+runtimesManifestDefault:
```
markusthoemmes: This change is I think not necessary. We can leave the name as is, the distinction is made via snake_case vs. pascalCase.

csantanapr: I'm confuse, I need a new name to hold the default values, take a look below where I define `runtimesManifest`

markusthoemmes: Oh, right. My bad I guess it was a bit early this morning.

csantanapr: Can you approve now?

csantanapr:
PG1 2059 🔵

PR #2365, opened 2017-06-13
Reference scancode from utils repo. but preserve local copy (phase 1)

mrutkows:
Phase 1: Reference Travis to new home of scancode, but preserve one here until confirmed.

PR #2744, opened 2017-09-14
Deploy kafka & zookeeper cluster with ansible

devbv:
This PR make you deploy kafka as a cluster with replication factor.
I requested similar one a few month ago, but recently I've rebased it to be suitable with current code base.


devbv:
Travis failed, but it is not related with what I did...

TASK [write whisk.properties template to openwhisk_home] ***********************
 [WARNING]: Unable to find '/home/travis/build/apache/incubator-
openwhisk/ansible/db_local.ini' in expected paths.
fatal: [ansible]: FAILED! => {"changed": false, "failed": true, "msg": "AnsibleError: {{ lookup('ini', 'db_protocol section=db_creds file={{ playbook_dir }}/db_local.ini') }}: An unhandled exception occurred while running the lookup plugin 'ini'. Error was a <type 'exceptions.IOError'>, original message: [Errno 2] No such file or directory: 'None'"}
	to retry, use: --limit @/home/travis/build/apache/incubator-openwhisk/ansible/logs.retry

devbv:
Enable deploying zookeeper & kafka in separate hosts to allow deploying on local environment.

devbv:
@jeremiaswerner now it works on local environment. check it when you're available.

ansible/group_vars/all
```diff
@@ -145,10 +146,22 @@ kafka:
       retentionBytes: 1073741824
       retentionMS: 300000
 
+kafka_connect_string: "{% set ret = [] %}
```
jeremiaswerner: This multiline script leads to a string with whitespaces at the beginning of the property that is mapped into the container. 
```
"KAFKA_HOSTS=     192.168.99.100:9092,192.168.99.100:9093",
"ZOOKEEPER_HOSTS=    192.168.99.100:2181,192.168.99.100:2182",
```
If we put it in one line by removing the spaces between braces, like `}{` the setting looks correct. Another option might be to run something like a trim() on the resulting string after the joinL

ansible/group_vars/all
```diff
@@ -145,10 +146,22 @@ kafka:
...
+                       {% for host in groups['kafkas'] %}
+                         {{ ret.append( hostvars[host].ansible_host + ':' + ((kafka.port+loop.index-1)|string) ) }}
+                       {% endfor %}
+                       {{ ret | join(',') }}"
+
 zookeeper:
   version: 3.4
   port: 2181
 
+zookeeper_connect_string: "{% set ret = [] %}
```
jeremiaswerner: same here

ansible/environments/local/hosts
```diff
@@ -10,8 +10,11 @@ ansible ansible_connection=local
 controller0         ansible_host=172.17.0.1 ansible_connection=local
 controller1         ansible_host=172.17.0.1 ansible_connection=local
 
-[kafka]
-172.17.0.1          ansible_host=172.17.0.1 ansible_connection=local
+[kafkas]
```
jeremiaswerner: we might want to run `local` and travis deployments with two kafka nodes to have better test coverage.

ansible/group_vars/all
```diff
@@ -125,8 +125,9 @@ kafka:
   version: 0.10.2.1
   port: 9092
   ras:
-    port: 9093
+    port: 8093
   heap: "{{ kafka_heap | default('1g') }}"
+  replicationFactor: 1
```
jeremiaswerner: The `replicationFactor` is an important setting for the resiliency of the kafka cluster. If it's set to `1` by default the system would not be resilient against failures of a kafka node.

Therefore I would prefer to handle the `replicationFactor` in a way that we can override it per environment similar to the `heap` setting above, i.e.

```
replicationFactor: {{ kafka_replicationFactor | default(1) }}
```

In a particular environment where we use multiple kafka nodes we would then be able to override the replicationFactor accordingly. Also, I think that's in general a good practice in our ansible playbooks.

ansible/environments/docker-machine/hosts.j2.ini
```diff
@@ -10,9 +10,12 @@ ansible ansible_connection=local
 controller0                 ansible_host={{ docker_machine_ip }}
 controller1                 ansible_host={{ docker_machine_ip }}
 
-[kafka]
+[kafkas]
 {{ docker_machine_ip }}     ansible_host={{ docker_machine_ip }}
```
jeremiaswerner: 👍 Due to the complaints of resource consumption for local development I think it's good just have a single kafka container in a docker-machine deployment.

jeremiaswerner:
@devbv Thanks a lot for this very valuable and appreciated contribution. I've played around with the change and have left a few minor comments. 

I did the following manual resiliency test on my docker-machine deployment with two kafka containers. To get the test run successful I needed to adopt the `replicationFactor: 2` (instead of `1`). See my comment below and on the setting.

Prereq:
 - Deploy this PR with two kafka nodes `kafka0, kafka1` in the docker-machine
 - Create a simple hello world action

Test:
1. invoke an action and expect it succeeds -> works as expected 👍 
2. stop kafka0
3. invoke an action and expect it succeeds -> works as expected 👍
4. stop kafka1
5. invoke an action and expected it fails -> works as expected 👍
6. start kafka1 and/or kafka0
7. invoke an action and expect it succeeds again -> works as expected 👍

If we don't adjust the `replicationFactor` to `2` the topics seem to be distributed across the two nodes of the cluster. That leads to the situation where the health topic lives on just a single node and if we kill that node the controller would assume all invokers are down because invokers are not able to send the health pings to the controller.

For HA and resiliency reasons I therefore vote to make the replicationFactor configurable per environment. I'am not 100% sure if the replicationFactor must be equal to the number of kafka nodes and/or if the replicationFactor has a negative impact on latency and throughput.

I'll therefore further test the latency and throughput in a distributed deployment with different replicationFactors and will post the results here as datapoints.

I think we need to have a automated test that checks a similar scenario I've posted above. You might want to checkout the following example to add an automated test `ShootKafka` 
https://github.com/apache/incubator-openwhisk/blob/master/tests/src/test/scala/ha/ShootComponentsTests.scala

Again, many thanks for the contribution and I'am looking forward to get this in.

rabbah:
Great contribution, and comments! +1 to testing in Travis with a clustered kafka (as we do with controller and invoker). I would make the default for local dev one if each component (note today it’s not and I opened an issue to address this).

rabbah:
Jeremias do you think we can add a test for the resiliency then? Like we have done with controller active/passive?

devbv:
# About latency from replication factor

To state the simple conclusion first, **there's no additional latency by setting replication factor in current openwhisk system.** Because current KafkaProducerConnector sets 'ACKS_CONFIG(acks)' as 1 which means that leader of kafka responds ack to client after write the record to its local log but without awaiting full acks from other followers. You can find the details on [Kafka Docs](https://kafka.apache.org/documentation/)
By the way, it could lead to lose some messages if leader of kafka fails before copying data to the followers.

If we want to get more guarantee that there's no message lose after receiving ack, we can set 'ACKS_CONFIG' to "all" which means leader of kafka responds message after finishing replication.
I changed 'ACKS_CONFIG" to "all" and roughly checked latency of producing message by getting duration between the "posting topic ... " log and "posted to ..." log. I ignored that first 20 logs because there's some latency of kafka warm-up. Here's the result.

- RF: 1 (ack = 1)
> min: less than 1ms
   avg: 0.4 ms
   max: 4.0 ms

- RF: 1 (ack = all)
> min: less than 1ms
   avg: 3.6 ms
   max: 25.0 ms

- RF: 2 (ack = all)
> min: less than 1ms
   avg: 10.4 ms
   max: 52.0 ms

- RF: 3 (ack = all)
> min: less than 1ms
   avg: 14.4ms
   max: 87.0ms

we can find that "ack=all" increase latency even in case of RF=1. Maybe we need to provide config to set ACKS_CONFIG value.

@jeremiaswerner

jeremiaswerner:
@devbv many thanks for the update on the review comments and the investigation on latency. I think what you described makes perfect sense and making the ACKS_CONFIG value configurable per environment is the right thing to do.

I've done some tests on latency and load of the overall system as well and was not possible to detect a regression. I've tested maximum throughput with a single user/action and with 100 different users. Further checked on latency for different runtimes. All numbers in the same ballpark.

Further I've performed some investigations on resiliency under load and found that even with `acks=1` the loss during the outage of a single node seems acceptable. 

Since OW is latency sensitive I would vote to go with `acks=1` as default and allow a "deployer" to override the default per deployment. We might need a comment on the setting to describe the tradeoff.

ansible/roles/kafka/tasks/clean.yml
```diff
@@ -3,16 +3,8 @@
 
 - name: remove kafka
```
jeremiaswerner: just recognized an issue with rollout into existing environments and while testing it on PGs. While we do transition we should keep a clean step to remove `kafka` container. Since the new clean task would only clean `kafkaX` containers but leave `kafka` containers which could result in a port conflict.

Proposal: pls keep a second `remove kafka` task in here. we could remove after some time. (same for zookeeper removal below)

ansible/roles/zookeeper/tasks/clean.yml
```diff
@@ -0,0 +1,10 @@
+---
+# Remove kafka and zookeeper containers.
+
+- name: remove zookeeper
```
jeremiaswerner: same as for kafka above. pls add a task to remove `zookeeper` container.

jeremiaswerner:
PG2 #2365 🔵

ansible/roles/kafka/tasks/deploy.yml
```diff
@@ -1,57 +1,40 @@
...
-    image: ches/kafka:{{ kafka.version }}
+    name: kafka{{ groups['kafkas'].index(inventory_hostname) }}
+    image: wurstmeister/kafka:{{ kafka.version }}
     state: started
     recreate: true
     restart_policy: "{{ docker.restart.policy }}"
-    links:
-      - "zookeeper:zookeeper"
     env:
+      "KAFKA_DEFAULT_REPLICATION_FACTOR": "{{ kafka.replicationFactor }}"
```
markusthoemmes: Don't we set all replication factors for topics explicitly? Do we need this as a fallback?

ansible/roles/kafka/tasks/deploy.yml
```diff
@@ -1,57 +1,40 @@
...
+      "KAFKA_BROKER_ID": "{{ groups['kafkas'].index(inventory_hostname) }}"
       "KAFKA_ADVERTISED_HOST_NAME": "{{ ansible_host }}"
+      "KAFKA_ADVERTISED_PORT": "{{ kafka.port + groups['kafkas'].index(inventory_hostname) }}"
       "KAFKA_HEAP_OPTS": "-Xmx{{ kafka.heap }} -Xms{{ kafka.heap }}"
+      "KAFKA_ZOOKEEPER_CONNECT": "{{ zookeeper_connect_string }}"
+      "KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR": "{{ kafka.replicationFactor }}"
```
markusthoemmes: What is this needed for? We didn't have it before IIRC. Are offsets not tracked in zookeeper?

jeremiaswerner:
@markusthoemmes can you pls have a final check as well pls

tests/src/test/scala/services/KafkaConnectorTests.scala
```diff
@@ -112,4 +123,35 @@ class KafkaConnectorTests extends FlatSpec with Matchers with WskActorSystem wit
...
+
+        commandComponent(kafkaHost, "start", s"kafka$i")
+        retry({
+          startLog.r
+            .findAllMatchIn(commandComponent(kafkaHost, "logs", s"kafka$i").stdout)
+            .length shouldBe prevCount + 1
+        }, 15, Some(1.second)) // wait until kafka is up
+      }
+    }
+  }
```
markusthoemmes: Great to have a test but there are no assertions in here wrt. "sending and consuming of messages still works". Could those be added?

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
```diff
@@ -269,13 +269,13 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers {
...
     assetHelper.withCleaner(wsk.action, name, confirmDelete = true) {
-      val allowedMemory = 256.megabytes
+      val allowedMemory = 128.megabytes
       val actionName = TestUtils.getTestActionFilename("memoryWithGC.js")
       (action, _) =>
         action.create(name, Some(actionName), memory = Some(allowedMemory))
     }
 
-    val run = wsk.action.invoke(name, Map("payload" -> 512.toJson))
+    val run = wsk.action.invoke(name, Map("payload" -> 256.toJson))
```
markusthoemmes: Bad push? Those changes are unrelated, no?

tests/src/test/scala/common/TestUtils.java
```diff
@@ -69,7 +69,7 @@
     public static final int CONFLICT        = 153;      // 409 - 256 = 153
     public static final int TOO_LARGE       = 157;      // 413 - 256 = 157
     public static final int THROTTLED       = 173;      // 429 (TOO_MANY_REQUESTS) - 256 = 173
-    public static final int APP_ERROR       = 246;      // 502 - 256 = 246
+    public static final int APP_ERROR       = 244;      // 500 - 256 = 244
```
markusthoemmes: Why does this need to be changed?

ansible/environments/local/hosts
markusthoemmes: Note that travis == local.

ansible/roles/kafka/tasks/deploy.yml
devbv: Not only fallback but also kafka tests don't set the topic replication explicitly.

ansible/roles/kafka/tasks/deploy.yml
devbv: New kafka image 'wurstmeister/kafka-docker' use 1 as default value for it. (Previous image, ches/kafka set its value as 3)
So even though we set replication factor for each topic, one instance fail makes the whole kafka cluster fail.

https://travis-ci.org/apache/incubator-openwhisk/builds/305709471
This is the test result without setting KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR. You can see that kafka just stucked.

tests/src/test/scala/common/TestUtils.java
devbv: APP_ERROR and TIMEOUT have same error code. So I couldn't figure out exactly what kind of error happened to controller after kafka shooting test done.
Anyway it is unrelated with this PR, I'm gonna rollback it.

tests/src/test/scala/services/KafkaConnectorTests.scala
devbv: Good idea, I'm gonna add the test you suggest.

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
devbv: Travis failed “Action limits should be aborted when exceeding its memory limits” after I increased kafka instance number to 2. (All good in my local environment & dist environment)

```
whisk.core.limits.ActionLimitsTests > Action limits should be aborted when exceeding its memory limits STANDARD_OUT
    check failed for activation 106409d09700435da409d09700835d69: {"duration":315,"name":"TestNodeJsMemoryExceeding","subject":"guest","activationId":"106409d09700435da409d09700835d69","publish":false,"annotations":[{"key":"limits","value":{"timeout":60000,"memory":256,"logs":10}},{"key":"path","value":"guest/TestNodeJsMemoryExceeding"},{"key":"kind","value":"nodejs:6"}],"version":"0.0.1","response":{"result":{"error":"An error has occurred: RangeError: Array buffer allocation failed"},"success":false,"status":"action developer error"},"end":1511323308912,"logs":["2017-11-22T04:01:48.83040658Z  stdout: helloEatMemory memory 512MB","2017-11-22T04:01:48.899278892Z stderr: RangeError: Array buffer allocation failed","2017-11-22T04:01:48.899342652Z stderr: at Buffer.Uint8Array (native)","2017-11-22T04:01:48.899350793Z stderr: at FastBuffer (buffer.js:11:5)","2017-11-22T04:01:48.89935681Z  stderr: at createUnsafeBuffer (buffer.js:38:12)","2017-11-22T04:01:48.899362782Z stderr: at new Buffer.alloc (buffer.js:130:12)","2017-11-22T04:01:48.899368458Z stderr: at eat (eval at <anonymous> (/nodejsAction/runner.js:76:109), <anonymous>:3:18)","2017-11-22T04:01:48.899379285Z stderr: at NodeActionRunner.main [as userScriptMain] (eval at <anonymous> (/nodejsAction/runner.js:76:109), <anonymous>:11:5)","2017-11-22T04:01:48.899386376Z stderr: at /nodejsAction/runner.js:95:45","2017-11-22T04:01:48.899391497Z stderr: at NodeActionRunner.run (/nodejsAction/runner.js:89:16)","2017-11-22T04:01:48.899397412Z stderr: at doRun (/nodejsAction/src/service.js:155:31)","2017-11-22T04:01:48.899402993Z stderr: at runCode (/nodejsAction/src/service.js:111:20)"],"start":1511323308597,"namespace":"guest"}
    Exception occurred during test execution: org.scalatest.exceptions.TestFailedException: "An error has occurred: RangeError: Array buffer allocation failed" was not equal to "The action exhausted its memory and was aborted."
```

I doubt that travis suffers low free memory problem caused by memory consumption of additional kafka&zookeeper instance. I also reduced kafka default heap size to 0.5g with same reason. 
But it had been changed before I changed kafka image, so I try rollback and run Travis again.

devbv:
The travis once succeed just by luck. I've found that NotLeaderForPartitionException should be handled 
 in client-side code when we use kafka cluster. 
I'll have a look at this in this weekend. So please hold off the PG until I finish it.

devbv:
Kafka issues are resolved, but HA mode of controller lead to fail travis. Many tests were designed for sequential command.

jeremiaswerner:
PG2/2381 🔵

tests/src/test/scala/services/KafkaConnectorTests.scala
```diff
@@ -65,17 +65,34 @@ class KafkaConnectorTests extends FlatSpec with Matchers with WskActorSystem wit
     super.afterAll()
   }
 
+  def commandComponent(host: String, command: String, component: String) = {
```
rabbah: @devbv Could reuse these methods here: 
https://github.com/apache/incubator-openwhisk/blob/e587bf8ec632320614f14114c4e85e67f39c06c2/tests/src/test/scala/ha/ShootComponentsTests.scala#L94

Sorry I'm a little late to the party.

devbv: @rabbah ok, I could fix it on another PR including enhancement of handling kafka errors.

PR #2745, opened 2017-09-14
Configure jmxremote

vvraskin:
Hello, 

this change will enable/make configurable jmx remote configuration for controllers. 
A couple of benefits it provides:
* change logback log level without JVM restart
* check akka cluster state
* JVM profiling with visualVM/JProfile

regards, Vadim.

### DIFF
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -17,17 +17,17 @@
   become: "{{ logs.dir.become }}"
 
 - name: copy jmxremote password file
-  when: jmxremote_enabled
+  when: jmx.enabled
   template:
-    src: "jmxremote.password"
-    dest: "{{ whisk_logs_dir }}/controller{{ groups['controllers'].index(inventory_hostname) }}"
+    src: "jmxremote.password.j2"
+    dest: "{{ whisk_logs_dir }}/controller{{ groups['controllers'].index(inventory_hostname) }}/jmxremote.password"
     mode: 0777
 
 - name: copy jmxremote access file
-  when: jmxremote_enabled
+  when: jmx.enabled
   template:
-    src: "jmxremote.access"
-    dest: "{{ whisk_logs_dir }}/controller{{ groups['controllers'].index(inventory_hostname) }}"
+    src: "jmxremote.access.j2"
+    dest: "{{ whisk_logs_dir }}/controller{{ groups['controllers'].index(inventory_hostname) }}/jmxremote.access"
     mode: 0777
 
 - name: check, that required databases exist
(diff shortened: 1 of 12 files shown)


starpit:
personally, i'd be wary of having JMX enabled on the production servers. but this is mostly from an excess of caution, rather than any egregious risks. this *will* open ports on the production machines. JMX is RMI based, making it a pain (though not impossible) to do anything remotely, with firewalls in the middle, etc.

and... as long as we're logging histograms and GC stats, and collecting heap dumps on OOM, i think we have most of our bases covered. 

still, as long as we are aware of the risks, though, i don't see any giant problems, here.

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -16,6 +16,20 @@
     mode: 0777
   become: "{{ logs.dir.become }}"
 
+- name: copy jmxremote password file
+  when: jmx.enabled
+  template:
+    src: "jmxremote.password.j2"
+    dest: "{{ whisk_logs_dir }}/controller{{ groups['controllers'].index(inventory_hostname) }}/jmxremote.password"
```
cbickel: Can we use a config directory here instead (you are creating this in another PR anyway).

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -98,11 +128,10 @@
       "CONFIG_kamon_statsd_port": "{{ metrics.kamon.port }}"
     volumes:
       - "{{ whisk_logs_dir }}/controller{{ groups['controllers'].index(inventory_hostname) }}:/logs"
-    ports:
-      - "{{ controller.basePort + groups['controllers'].index(inventory_hostname) }}:8080"
-      - "{{ controller.akka.cluster.basePort + groups['controllers'].index(inventory_hostname) }}:{{ controller.akka.cluster.bindPort }}"
+    ports: "{{ ports_to_expose }}"
     command: /bin/sh -c "exec /init.sh {{ groups['controllers'].index(inventory_hostname) }} >> /logs/controller{{ groups['controllers'].index(inventory_hostname) }}_logs.log 2>&1"
 
+
```
cbickel: Can we remove that line?

ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -159,6 +179,8 @@
         -v {{ dockerInfo["DockerRootDir"] }}/containers/:/containers
         -v {{ docker_sock | default('/var/run/docker.sock') }}:/var/run/docker.sock
         -p {{ invoker.port + groups['invokers'].index(inventory_hostname) }}:8080
+        {% if jmx.enabled %} -p {{ jmx.basePortInvoker + groups['invokers'].index(inventory_hostname) }}:{{ jmx.basePortInvoker + groups['invokers'].index(inventory_hostname) }} {% endif %}
+        {% if jmx.enabled %} -p {{ jmx.rmiBasePortInvoker + groups['invokers'].index(inventory_hostname) }}:{{ jmx.rmiBasePortInvoker + groups['invokers'].index(inventory_hostname) }} {% endif %}
```
cbickel: Are we completely sure, that a user is not able to reach this port from an useraction?

ansible/roles/controller/tasks/deploy.yml
vvraskin: I think we can leave it as it is, so we both invoker and controller use logs directory. Didn't want to introduce a new config dir in invoker just as a temp storage for jmx config.

ansible/roles/invoker/tasks/deploy.yml
vvraskin: As long as we have network isolation between management and user container plain on the invokers we must be safe. Now they are separated into 2 subnets 172.x.x.x and 192.x.x.x afaik.

ansible/roles/controller/tasks/deploy.yml
rabbah: Hmm I understand not wanting to introduce a new directory but this seems like abusing the purpose of the logs dir. I’d favor using a conf or seperate scratch space also (maybe repurposing the nginx conf dir to be more generic is easier and cleaner).

vvraskin: I understand what you mean, but the config won't be stored in the logs directory anyway. It will be moved into the root after we init the component, https://github.com/apache/incubator-openwhisk/pull/2745/files#diff-98fffb6dfa1b0bf01d222972465eeb2eR5. 
We do this kind of _trick_ because in some docker versions, all the files that will be mounted from the host into containers will be given `nouser` and `nogroup` priviledges, which implies that the root user inside of the container won't have permissions to change file if its copied with 0600 or even if the file has 777, the file permissions cannot be changed by root unless you copy it into another file. (Some docker specific behavior that I was not able to change, would be grateful if someone has another workaround) So what I do is copying the file to logs directory with max permission and then move the data with the correct permissions and owners into /root directory.

rabbah: Thanks for the explanation - I was asking something simpler though: could `{{ config_root_dir }}` be the base vs the `{{ whisk_logs_dir }}`?

mhenke1: @rabbah {{ whisk_logs_dir }} is now used to move the file

cbickel: As discussed personal: @mhenke1 will change this to the config dir.

cbickel:
Will be replaced by: #3163

PR #2750, opened 2017-09-14
Added --result to invoke action

GabrielNicolasAvellaneda:
Passing --result to invoke action is a better feedback to confirm that it actually executed the function and got back the expected result.

### DIFF
--- a/docs/about.md
+++ b/docs/about.md
@@ -37,7 +37,7 @@ wsk action create myAction action.js
 ```
 Done. Now we actually want to invoke that action:
 ```
-wsk action invoke myAction
+wsk action invoke myAction --result
 ```
 
 ## The internal flow of processing


rabbah:
This was a good suggestion.

GabrielNicolasAvellaneda:
@rabbah I was not sure enough because at the end of the tutorial it was talking about just showing how to do async calls on actions and I suppose that using --result is a synchronous call instead. That's why I closed this. Probably will be good to show a complete example at the beginning using async and showing the actual result returned by the function. Then goes the explanation of the workflow for a call.

rabbah:
I can work here to use `--result` --- but welcome further improvements to the doc to make things clearer.

GabrielNicolasAvellaneda:
I will try to do some improvements as I read.

rabbah:
Thanks. Do you have an Apache ICLA?

GabrielNicolasAvellaneda:
Hey @rabbah I don't think that I have one. Should I follow this guide to get one? https://www.apache.org/licenses/#clas

rabbah:
👍 thank you.

GabrielNicolasAvellaneda:
@rabbah I filled the ICLA and I've got listed on http://people.apache.org/unlistedclas.html

rabbah:
Thanks for following up!

GabrielNicolasAvellaneda:
@rabbah Thanks! :+1:

PR #2751, opened 2017-09-15
Remove "Thread.sleep" from log reading retries.

markusthoemmes:
"Thread.sleep" unnecessarily blocks the Thread instead of scheduling the retry.

Part of a general mission of reducing blocking in our Invoker code.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
@@ -19,6 +19,9 @@ package whisk.core.containerpool.docker
 
 import java.nio.charset.StandardCharsets
 import java.time.Instant
+
+import akka.actor.ActorSystem
+
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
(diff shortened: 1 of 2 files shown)


core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -250,8 +255,7 @@ class DockerContainer(id: ContainerId, ip: ContainerIp)(implicit docker: DockerA
 
           if (retries > 0 && !isComplete && !isTruncated) {
             logger.info(this, s"log cursor advanced but missing sentinel, trying $retries more times")
-            Thread.sleep(logsRetryWait.toMillis)
-            readLogs(retries - 1)
+            akka.pattern.after(logsRetryWait, as.scheduler)(readLogs(retries - 1))
```
rabbah: should we make read logs tail recursive?

markusthoemmes: I think it now effectively is trampolined since it's decoupled via the akka scheduler. It's not really tail-recursive though.

PR #2752, opened 2017-09-15
Use an asynchronous process runner to spawn container commands.

markusthoemmes:
**Note:** Measurements of the impacts of this are not yet done. I thought it makes sense to open the PR though for feedback.

`scala.sys.Process` only has a blocking interface which blocks one thread for the runtime of the command. Furthermore, each of those processes open another thread to handle the output streams of the started process. Under load, the system churns through a **lot** of threads, which can cause efficiency problems.

NuProcess is supposed to fix this by providing an epoll interface, to make waiting on process completion asynchronous and its own (small) threadpool to maintain state and handle streams.

### DIFF
--- a/core/invoker/build.gradle
+++ b/core/invoker/build.gradle
@@ -13,6 +13,8 @@ repositories {
 dependencies {
     compile "org.scala-lang:scala-library:${gradle.scala.version}"
     compile project(':common:scala')
+
+    compile "com.zaxxer:nuprocess:1.1.2"
 }
 
 tasks.withType(ScalaCompile) {
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/ProcessRunner.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/ProcessRunner.scala
@@ -17,11 +17,13 @@
 
 package whisk.core.containerpool.docker
 
-import scala.collection.mutable
-import scala.concurrent.ExecutionContext
-import scala.concurrent.Future
-import scala.concurrent.blocking
-import scala.sys.process._
+import java.nio.ByteBuffer
+
+import akka.util.ByteString
+import com.zaxxer.nuprocess.{NuAbstractProcessHandler, NuProcessBuilder}
+
+import scala.collection.JavaConverters._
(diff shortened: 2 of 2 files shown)


core/invoker/build.gradle
```diff
@@ -13,6 +13,8 @@ repositories {
 dependencies {
     compile "org.scala-lang:scala-library:${gradle.scala.version}"
     compile project(':common:scala')
+
+    compile "com.zaxxer:nuprocess:1.1.2"
```
rabbah: license?

core/invoker/src/main/scala/whisk/core/containerpool/docker/ProcessRunner.scala
```diff
@@ -35,19 +37,30 @@ trait ProcessRunner {
...
-      (exitCode, out.mkString("\n"), err.mkString("\n"))
-    }).flatMap {
-      case (0, stdout, _) =>
-        Future.successful(stdout)
-      case (code, stdout, stderr) =>
-        Future.failed(ProcessRunningException(code, stdout, stderr))
-    }
+  protected def executeProcess(args: String*)(implicit ec: ExecutionContext): Future[String] = {
+    val promise = Promise[String]()
+    val pb: NuProcessBuilder = new NuProcessBuilder(args.asJava)
```
rabbah: what's the advantage of this vs using a custom runner - like in TestUtils.runCmd?

core/invoker/build.gradle
markusthoemmes: Apache 2.0.

core/invoker/src/main/scala/whisk/core/containerpool/docker/ProcessRunner.scala
markusthoemmes: As said in the commit message, in a nutshell:

* Use epoll to detect finished processes --> thread consumption reduced
* Use a centralized thread pool for stream consuming. The current (scala.sys.Process) runner starts one unnamed Thread **per command** to do this. These are barebone threads, not manageable by any pool whatsoever.

Point for NuProcess: Akka uses it in its akka-stream contrib module for async processes.

PR #2604, opened 2017-08-11
Give more information in CLI Activation lists

underwoodb-sd-ibm:
This update is intended to add more descriptive and informative information to the output of `wsk activation list`. Currently, the output is:

```
$ wsk activation list
activations
9dd311dac5fb446ab18af3039e919d34                hello
09a09105c6ff4a819659e94a4d587d25                hello
1ee5202308e4474d80f2e42cd7c39835                hello
56fa426c50e248f6a2c49130ae230a99                zhelloRule
f17373e4ee824f5b8b8067f5cd5956be                ahelloRule
d71dcf45ae9047d594a8281e3206c3cd                helloRule
07e0dcb02817425697cd9a5632205dc0                helloTriggerhello
```

The new output would be something like:

```
$ wsk activation list
activations
activation id                     name          type     start date  start time  end time  duration (ms)  status
9dd311dac5fb446ab18af3039e919d34  hello         action   2017-08-11    15:30:09  15:30:09             10  success
09a09105c6ff4a819659e94a4d587d25  hello         action   2017-08-11    15:30:09  15:30:09             81  success
1ee5202308e4474d80f2e42cd7c39835  hello         action   2017-08-11    15:30:09  15:30:09             71  success
56fa426c50e248f6a2c49130ae230a99  zhelloRule    rule     2017-08-11    15:30:09  --:--:--              -  success
f17373e4ee824f5b8b8067f5cd5956be  ahelloRule    rule     2017-08-11    15:30:09  --:--:--              -  success
d71dcf45ae9047d594a8281e3206c3cd  helloRule     rule     2017-08-11    15:30:09  --:--:--              -  success
07e0dcb02817425697cd9a5632205dc0  helloTrigger  trigger  2017-08-11    15:30:09  --:--:--              -  success
6e40decb0e714ec3b7309c4f11b499c5  hello         action   2017-08-11    15:20:11  15:20:11             55  success
```
As discussed in the original issue, this update includes defining not just the name of the entity that was activated, but the type, start/end time including start date, the duration, and the status.

Closes #20

PR #2760, opened 2017-09-16
Add new design docs that remove older views that are no longer necessary

rabbah:
Removes a view from the main entities db that is no longer required. It was used to isolate activations which now reside in a separate database.

Also remove a completely redundant activation index.
Add additional metadata to the activation view that is useful for #2604 and other rich clients and eschews the results and logs which are large payloads.

This builds on #2735.


rabbah:
pg3/1102 🔵

rabbah:
Here are the maps for convenience:

```js
function (doc) {
  var PATHSEP = "/";
  var isAction = function (doc) { return (doc.exec !== undefined) };
  if (isAction(doc)) try {
    var ns = doc.namespace.split(PATHSEP);
    var root = ns[0];
    var date = doc.updated;
    var value = {
      namespace: doc.namespace,
      name: doc.name,
      version: doc.version,
      publish: doc.publish,
      annotations: doc.annotations
    };
    emit([doc.namespace, date], value);
    if (root !== doc.namespace) {
      emit([root, date], value);
    }
  } catch (e) {}
}
```

```
function (doc) {
  var isPackage = function (doc) {  return (doc.binding !== undefined) };
  if (isPackage(doc)) try {
    var date = doc.updated;
    emit([doc.namespace, date], {
      namespace: doc.namespace,
      name: doc.name,
      version: doc.version,
      publish: doc.publish,
      annotations: doc.annotations,
      binding: Object.keys(doc.binding).length !== 0
    });
  } catch (e) {}
}
```

```
function (doc) {
  var PATHSEP = "/";
  var isRule = function (doc) {  return (doc.trigger !== undefined) };
  if (isRule(doc)) try {
    var ns = doc.namespace.split(PATHSEP);
    var root = ns[0];
    var date = doc.updated;
    var value = {
      namespace: doc.namespace,
      name: doc.name,
      version: doc.version,
      publish: doc.publish,
      annotations: doc.annotations
    };
    emit([doc.namespace, date], value);
    if (root !== doc.namespace) {
      emit([root, date], value);
    }
  } catch (e) {}
}
```

```
function (doc) {
  var PATHSEP = "/";
  var isTrigger = function (doc) { return (doc.exec === undefined && doc.binding === undefined && doc.parameters !== undefined) };
  if (isTrigger(doc)) try {
    var ns = doc.namespace.split(PATHSEP);
    var root = ns[0];
    var date = doc.updated;
    var value = {
      namespace: doc.namespace,
      name: doc.name,
      version: doc.version,
      publish: doc.publish,
      annotations: doc.annotations
    };
    emit([doc.namespace, date], value);
    if (root !== doc.namespace) {
      emit([root, date], value);
    }
  } catch (e) {}
}
```

```
function (doc) {
  var PATHSEP = "/";

  var isPackage = function (doc) {  return (doc.binding !== undefined) };
  var isAction = function (doc) { return (doc.exec !== undefined) };
  var isTrigger = function (doc) { return (doc.exec === undefined && doc.binding === undefined && doc.parameters !== undefined) };
  var isRule = function (doc) {  return (doc.trigger !== undefined) };
  
  var collection = function (doc) {
    if (isPackage(doc)) return "packages";
    if (isAction(doc)) return "actions";
    if (isTrigger(doc)) return "triggers";
    if (isRule(doc)) return "rules";
    return undefined;
  };

  try {
    var type = collection(doc);
    if (type !== undefined) {
      var ns = doc.namespace.split(PATHSEP);
      var root = ns[0];
      var date = doc.updated;
      var value = {
        collection: type,
        namespace: doc.namespace,
        name: doc.name,
        version: doc.version,
        publish: doc.publish,
        annotations: doc.annotations
      };
      if (isPackage(doc)) {
        value.binding = Object.keys(doc.binding).length !== 0;
      }
      emit([root, date], value);
    }
  } catch (e) {}
}
```

rabbah:
and for activations

```js
function (doc) {
  var PATHSEP = "/";
  var isActivation = function (doc) { return (doc.activationId !== undefined) };
  var summarize = function (doc) {
    var endtime = doc.end !== 0 ? doc.end : undefined;
    return {
        namespace: doc.namespace,
        name: doc.name,
        version: doc.version,
        publish: doc.publish,
        annotations: doc.annotations,
        activationId: doc.activationId,
        start: doc.start,
        end: endtime,
        duration: endtime !== undefined ? endtime - doc.start : undefined,
        cause: doc.cause,
        statusCode: (endtime !== undefined && doc.response !== undefined && doc.response.statusCode !== undefined) ? doc.response.statusCode : undefined
      }
  };

  if (isActivation(doc)) try {
    var value = summarize(doc)
    emit([doc.namespace, doc.start], value);
    emit([doc.namespace+PATHSEP+doc.name, doc.start], value);
  } catch (e) {}
}
```

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -166,10 +167,12 @@ object WhiskConfig {
...
   val dbUsername = "db.username"
   val dbPassword = "db.password"
-  val dbWhisk = "db.whisk.actions"
-  val dbAuths = "db.whisk.auths"
   val dbPrefix = "db.prefix"
+  val dbAuths = "db.whisk.auths"
+  val dbWhisk = "db.whisk.actions"
   val dbActivations = "db.whisk.activations"
+  val dbWhiskDesignDoc = "db.whisk.actions.ddoc"
+  val dbActivationsDesignDoc = "db.whisk.activations.ddoc"
```
rabbah: added these here because this is the one place we consolidate all "env" vars.

rabbah:
PG2/2120 🔵

rabbah:
@markusthoemmes @cbickel what do you think about changing this emit in the activation view:

```js
    emit([doc.namespace+PATHSEP+doc.name, doc.start], value);
```

which doesn't permit filtering by a package to extract the package name from the path annotation, and instead:

```js
    let pkg = getPackageFromAnnotation(doc.annotations) || "default"
    emit([doc.namespace+PATHSEP+pkg+PATHSEP+doc.name, doc.start], value);
```

where we prefix the package name if it exists, or "default" to normalize the strings for a missing package.

this will allow matching/filtering on an action either in the default package or a named package (but not a binding, which we dont track in an activation).

rabbah:
Maybe it's worth separating the two emits in the activation view into separate design docs - to allow for future changes to the by name indexing or alternate implementations without requiring a full reindex if the ddoc is changed. Thoughts @cbickel?

mdeuser:
IMHO, we will want to avoid any breaking change to the REST API.. if that is what will happen with the new db view return value.

rabbah:
There are no breaking changes in this PR. And further changing the views to allow filtering activations by package _can_ be done without an API change. But I think there's more work to be done there to warrant a separate PR. That said, it is worth considering splitting the two views in the activations ddoc into their own ddocs, to permit independent changes in the future.

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
```diff
@@ -292,27 +306,56 @@ class ActivationsApiTests extends ControllerTestCommon with WhiskActivationsApi
...
+    waitOnView(activationStore, namespace.root, 4, WhiskActivation.view)
 
     whisk.utils.retry {
       Get(s"$collectionPath?name=xyz") ~> Route.seal(routes(creds)) ~> check {
         status should be(OK)
         val response = responseAs[List[JsObject]]
-        activations.length should be(response.length)
-        activations forall { a =>
+        val allActivations = activations ++ activationsInPackage
+        allActivations.length should be(response.length)
```
markusthoemmes:
```
allActivations should have size response.length
```

Should yield a more speaking error.

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
```diff
@@ -292,27 +306,56 @@ class ActivationsApiTests extends ControllerTestCommon with WhiskActivationsApi
...
+        allActivations forall { a =>
           response contains a.summaryAsJson
         } should be(true)
```
markusthoemmes: A check like this will result in `false was not true` which usually doesn't help much.

Would

```scala
response should contain theSameElementsAs allActivations.map(_.summaryAsJson)
```

work? (I think that might apply to lots of tests here, so it'd be okay not to do this on this PR)

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
rabbah: fixed these in view tests - there's a lot of these in the API tests, would rather address all of these separately as it's more than just this one suite.

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
rabbah: yes should work - i made this change in view tests but going to defer cleaning up the rest of theses in this package as there are many of them and we should change all of them.

rabbah:
@cbickel squashed commits and updated message.

@ddragosd @bbrowning this PR contains changes which we'll need to mirror in the alt-deployment repo. I can submit PR.

PR #2766, opened 2017-09-17
Remove synthetic header.

### LINKED ISSUE #2636: incorrect header value in webaction
```
> curl -k https://guest.openwhisk/default/echo.json
{"__ow_method":"get","__ow_headers":{"accept":"*/*","user-agent":"curl/7.43.0","host":"controllers","timeout-access":"<function1>","connection":"close"},"__ow_path":""}
```

notice timeout-access header is `<function1>`.
https://github.com/akka/akka-http/issues/64 after a quick search.

@dubeejw

rabbah:
Removes timeout-access from headers:

```
> curl -k -X POST https://guest.localhost/default/echo.json -H "Content-Type: application/json" -d '["a"]'
{
  "__ow_method": "post",
  "__ow_query": "",
  "__ow_body": "WyJhIl0=",
  "__ow_headers": {
    "host": "controllers",
    "user-agent": "curl/7.43.0",
    "accept": "*/*",
    "timeout-access": "<function1>" <---- this one
  },
  "__ow_path": ""
}
```

Fixes #2636.
For reference: https://github.com/akka/akka-http/issues/64

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
@@ -38,6 +38,7 @@ import akka.http.scaladsl.model.HttpEntity
 import akka.http.scaladsl.server.Route
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.model.headers.`Content-Type`
+import akka.http.scaladsl.model.headers.`Timeout-Access`
 import akka.http.scaladsl.model.ContentType
 import akka.http.scaladsl.model.ContentTypes
 import akka.http.scaladsl.model.FormData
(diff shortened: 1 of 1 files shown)


core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -108,7 +109,11 @@ private case class Context(propertyMap: WebApiDirectives,
   def metadata(user: Option[Identity]): Map[String, JsValue] = {
     Map(
       propertyMap.method -> method.value.toLowerCase.toJson,
-      propertyMap.headers -> headers.map(h => h.lowercaseName -> h.value).toMap.toJson,
+      propertyMap.headers -> headers
+        .filter(_.name != `Timeout-Access`.name)
+        .map(h => h.lowercaseName -> h.value)
```
markusthoemmes: Scala tip: To iterate only once you can use `collect` to do a `filter`/`map` in one step:

```scala
headers.collect { 
  case h if h.name != `Timeout-Access`.name => 
    h.lowercaseName -> h.value
}
```

rabbah: thanks for the tip - likely will be adding more exclusions - there are a few synthetic headers we should drop.

PR #2771, opened 2017-09-19
Echo Access-Control-Request-Headers as Access-Control-Allow-Headers 

### LINKED ISSUE #2653: Default CORS handling doesn't handle headers well
Given this web action:

```php
<?php
{
    return [
        'hello' => 'world'
    ];
}
```

This JS call in a browser will fail:

```html
<!doctype html>
<html>
    <head>
    </head>
    <body>
        <h1>Test</h1>

        <pre id="data">
        </pre>

        <script src="https://code.jquery.com/jquery-3.2.1.min.js"></script>
        <script type="text/javascript">
            $(document).ready(function() {
                jQuery.ajax({
                    url: "https://openwhisk.eu-gb.bluemix.net/api/v1/web/19FT_dev/default/cors.json",
                    type: "GET",
                    headers: {
                        'X-Clacks-Overhead': 'GNU Terry Pratchett'
                    },
                    contentType: 'application/json; charset=utf-8',
                    success: function(resultData) {
                        console.log(resultData)
                        $('#data').html("Action result is: " + JSON.stringify(resultData))
                    },
                    error : function(jqXHR, textStatus, errorThrown) {
                        console.log("Error: " + textStatus)
                        console.log(errorThrown)
                        console.log(jqXHR)
                        $('#data').html("Error")
                    },
                    timeout: 120000,
                });
            });
        </script>
    </body>
    </html>
```

With an error message in the Chrome console of:

> XMLHttpRequest cannot load https://openwhisk.eu-gb.bluemix.net/api/v1/web/19FT_dev/default/cors.json. Request header field X-Clacks-Overhead is not allowed by Access-Control-Allow-Headers in preflight response.

This is because [`defaultCorsResponse`](https://github.com/apache/incubator-openwhisk/blob/master/core/controller/src/main/scala/whisk/core/controller/WebActions.scala#L354)  sends back `Access-Control-Allow-Headers` of `Authorization ` and `Content-Type` only.

What should happen is that `Access-Control-Allow-Headers` should be whatever was sent in `Access-Control-Request-Headers`.

rabbah:
Instead of hardcoding the header value for `Access-Control-Allow-Headers`, echo back the request headers from the request when `Access-Control-Request-Headers` is present.

Closes #2653.
FYI @akrabat

### DIFF
--- a/docs/webactions.md
+++ b/docs/webactions.md
@@ -393,11 +393,13 @@ $ curl -k -H "content-type: application" -X POST -d "Decoded body" https://${API
 
 By default, an OPTIONS request made to a web action will result in CORS headers being automatically added to the
 response headers. These headers allow all origins and the options, get, delete, post, put, head, and patch HTTP verbs.
-The headers are shown below:
+In addition, the header `Access-Control-Request-Headers` is echoed back as the header `Access-Control-Allow-Headers`
+if it is present in the HTTP request. Otherwise, a default value is generated as shown below.
 
 ```
 Access-Control-Allow-Origin: *
 Access-Control-Allow-Methods: OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH
+Access-Control-Allow-Headers: Authorization, Content-Type
 ```
 
 Alternatively, OPTIONS requests can be handled manually by a web action. To enable this option add a
--- a/tests/src/test/scala/whisk/core/cli/test/WskWebActionsTests.scala
+++ b/tests/src/test/scala/whisk/core/cli/test/WskWebActionsTests.scala
@@ -188,11 +188,28 @@ class WskWebActionsTests extends TestHelpers with WskTestHelpers with RestUtil w
       action.create(name, file, web = Some("true"))
     }
 
-    val responses = Seq(
-      RestAssured.given().config(sslconfig).options(s"$url.http"),
-      RestAssured.given().config(sslconfig).get(s"$url.json"))
+    Seq(
+      RestAssured
+        .given()
+        .config(sslconfig)
+        .header("Access-Control-Request-Headers", "x-custom-header")
+        .options(s"$url.http"),
+      RestAssured
+        .given()
+        .config(sslconfig)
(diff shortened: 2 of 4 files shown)


docs/webactions.md
```diff
@@ -393,11 +393,13 @@ $ curl -k -H "content-type: application" -X POST -d "Decoded body" https://${API
 
 By default, an OPTIONS request made to a web action will result in CORS headers being automatically added to the
 response headers. These headers allow all origins and the options, get, delete, post, put, head, and patch HTTP verbs.
-The headers are shown below:
+In addition, the header `Access-Control-Request-Headers` is echoed back as the header `Access-Control-Allow-Headers`
```
dubeejw: Should explicitly state that the `Access-Control-Request-Headers` is not echoed back when `web-custom-options` is enabled.

tests/src/test/scala/whisk/core/cli/test/WskWebActionsTests.scala
```diff
@@ -188,11 +188,28 @@ class WskWebActionsTests extends TestHelpers with WskTestHelpers with RestUtil w
       action.create(name, file, web = Some("true"))
     }
 
-    val responses = Seq(
-      RestAssured.given().config(sslconfig).options(s"$url.http"),
-      RestAssured.given().config(sslconfig).get(s"$url.json"))
+    Seq(
```
dubeejw: Need test that ensures `Access-Control-Request-Headers` isn't echoed back when `web-custom-options` is enabled.

docs/webactions.md
rabbah: The next paragraph states:

```
Alternatively, OPTIONS requests can be handled manually by a web action. To enable this option add a
`web-custom-options` annotation with a value of `true` to a web action. When this feature is enabled, CORS headers will
not automatically be added to the request response. Instead, it is the developer's responsibility to append their
desired headers programmatically. Below is an example of creating custom responses to OPTIONS requests.
```

Sufficient?

dubeejw:
PG3 1114 🔵

PR #2772, opened 2017-09-19
Fix issue with trailing spaces and handle comments in the property file.

### LINKED ISSUE #2770: cli does not tolerate comment on api host in property file
```
> cat ~/.wskprops
APIHOST=http://localhost:10001 #http://192.168.99.100:10001
```

```
> wsk property get
runtime error: invalid memory address or nil pointer dereference
Application exited unexpectedly
```

GabrielNicolasAvellaneda:
@rabbah this closes #2770

### DIFF
--- a/tests/src/test/scala/whisk/core/cli/test/WskConfigTests.scala
+++ b/tests/src/test/scala/whisk/core/cli/test/WskConfigTests.scala
@@ -136,7 +136,7 @@ class WskConfigTests extends TestHelpers with WskTestHelpers {
     val tmpwskprops = File.createTempFile("wskprops", ".tmp")
     try {
       val writer = new BufferedWriter(new FileWriter(tmpwskprops))
-      writer.write(s"APIHOST=http://localhost:10001    # This is a comment!")
+      writer.write(s"APIHOST=http://localhost:10001    # This is a comment!   ")
       writer.close()
       val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
       val stdout = wsk.cli(Seq("property", "get", "-i", "--apihost"), env = env).stdout
--- a/tests/src/test/scala/whisk/core/cli/test/WskConfigTests.scala
+++ b/tests/src/test/scala/whisk/core/cli/test/WskConfigTests.scala
@@ -132,6 +132,20 @@ class WskConfigTests extends TestHelpers with WskTestHelpers {
     }
   }
 
+  it should "get apihost removing any trailing white spaces and line comments" in {
+    val tmpwskprops = File.createTempFile("wskprops", ".tmp")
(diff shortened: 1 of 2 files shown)


rabbah:
Thanks for the fix! Could probably add a test for this, using a temp file and `WSK_CONFIG_FILE` to point to the temp file.

GabrielNicolasAvellaneda:
@rabbah Yes, I will do, I actually wanted to do a test for this. I think that should be good to add Go tests for the cli project as unit tests. Didn't see any unit tests. I think that the Scala tests looks like more as integration tests, I am correct?

rabbah:
Your observation is right. The cli code is moving to its own repo and should have unit tests eventually replacing supplementing the current ones.

GabrielNicolasAvellaneda:
Got it. Sounds good!

GabrielNicolasAvellaneda:
@dubeejw Sure, will do it today.

GabrielNicolasAvellaneda:
@dubeejw Done, please take a look. Thanks!

GabrielNicolasAvellaneda:
@rabbah Did you have time to take a look at these changes. On the other hand I'm thinking about adding unit tests for the wsk-cli project, do you think that it will be best to first move to its own repo?

rabbah:
The repo exists already. I don’t think it has any unit tests yet.

dubeejw:
@GabrielNicolasAvellaneda, when the CLI update the property file, do existing comments stay intact?

GabrielNicolasAvellaneda:
Hey @dubeejw Good question! No, comments gets removed when you write any property. **This is exactly the original behaviour.**

ReadProps is ignoring comments, for example if you have a single line comment it will be ignored by this following line that checks for a format like KEY=VALUE
https://github.com/apache/incubator-openwhisk/blob/master/tools/cli/go-whisk-cli/commands/util.go#L936
Then when you write a property it is using the map returned by ReadProps as a basis for all the property values.

Please also look at this line: https://github.com/apache/incubator-openwhisk/blob/master/tools/cli/go-whisk-cli/commands/util.go#L960

We should create a new issue to address this because it will require to change the logic of how the properties are set because the whole file is overwritten.
I could also do that fix if you assign it to me. In this case for writing to the file we must locate the proper line where the key is and update just that line. If that key does not exist then we should append at the end of the file.

GabrielNicolasAvellaneda:
@rabbah @dubeejw Created this https://github.com/apache/incubator-openwhisk/issues/2806 to fix .wskprops formatting issues.

rabbah:
@dubeejw I think this is ok now with the test, wdyt?

rabbah:
@dubeejw nudge.

dubeejw:
PG4 908 ⏳

PR #2773, opened 2017-09-19
Make blackbox invoker fraction configurable via Ansible

sven-lange-last:
Extend Ansible deploy playbook for controller role such that it passes the blackbox fraction
setting as environment variable when running the controller container. Make the blackbox fraction
a required property of the LoadBalancerService such that the setting is read from the environment.
Allow overriding of the default blackbox fraction in Ansible.

All this is needed to vary the proportion between managed and blackbox invokers to accomodate
usage in production.

I tested the change by adding the setting `controller_blackbox_fraction: 0.5` to file `ansible/environments/docker-machine/group_vars/all` and re-deploying the controller. As a result, the controller creates the following log entry indicating that the setting was picked up properly:

```
[2017-09-15T12:55:35.122Z] [INFO] [#sid_120] [LoadBalancerService] blackboxFraction = 0.5
```

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -106,7 +106,7 @@ controller:
   basePort: 10001
   heap: "{{ controller_heap | default('2g') }}"
   arguments: "{{ controller_arguments | default('') }}"
-  blackboxFraction: 0.10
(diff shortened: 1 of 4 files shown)


sven-lange-last:
PG3 - 1106 running

PR #2778, opened 2017-09-19
Test JSON values as web action body arguments

dubeejw:
Adds a unit test that ensures JSON strings, booleans, numbers, and arrays are acceptable body arguments with a `application/json` content type.


dubeejw:
PG3 1107 ⏳

dubeejw:
PG3 1107 🔵

PR #2779, opened 2017-09-20
Keep Cache invalidation messages in Kafka only for 5 minutes

cbickel:
Don't keep messages for the Cache Invalidation Kafka, that are older than 5 minutes.


cbickel:
PG4#811 is 🔵

PR #2781, opened 2017-09-20
Add a noOp invoker to test the controller without Invoker overhead

cbickel:
This Invoker is only to test the Controller. Therefor this invoker only returns a succesful ActiveAck if an activation is dispatched to it.
This Invoker does not get the action from the DB and it does not start docker containers to execute anything.
It should not be used in production environments.

### DIFF
--- a/ansible/roles/invoker/tasks/deploy.yml
+++ b/ansible/roles/invoker/tasks/deploy.yml
@@ -154,6 +154,7 @@
         -e CONFIG_kamon_statsd_hostname='{{ metrics.kamon.host }}'
         -e CONFIG_kamon_statsd_port='{{ metrics.kamon.port }}'
         -e CONFIG_whisk_spi_LogStoreProvider='{{ userLogs.spi }}'
+        -e CONFIG_whisk_spi_Invoker='{{ invoker_spi | default() }}'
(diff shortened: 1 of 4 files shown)


bbrowning:
There's already an spi model for Containers. You could perhaps achieve something similar by using the existing Invoker with a new NoOpContainer implementation that doesn't actually start any containers. That would run through the full Invoker code and just short-circuit container lifecycles to be no-ops.

markusthoemmes:
@bbrowning The idea though is to not run through all of the Invoker's code but rather just respond immediatly like an Invoker would. The goal is to more easily test the Controller in isolation, while still keeping messaging overhead the same.

bbrowning:
@markusthoemmes Ahh, right. A no-op Container would only avoid the Docker overhead but would still touch the DB. You'd also need a no-op or in-memory ArtifactStore in that case.

I'm not opposed to creating an Invoker spi as well, as @rabbah mentioned. But spi is probably the right approach here instead of adding additional properties and logic to select the no-op invoker explicitly.

markusthoemmes:
Closing as abandoned. @cbickel please reopen if you want to pick this up again.

PR #2784, opened 2017-09-20
Make logging hot paths lighter.

markusthoemmes:
Logging is **the** hotspot in our application today. This takes a first stab at making the overhead a bit smaller by respecting the set log-level as early as currently possible and not do any extranous throwaway computation.

It also reduces a bit of unnecessary boxing for the sake of just trying to keep a single space out of the message.


markusthoemmes:
PG1 2087 🔵

rabbah:
With these prs can we record pre and post improvements?

PR #2785, opened 2017-09-20
Compute UUIDs in a non-blocking fashion.

### LINKED ISSUE #2747: Use non-blocking UUID generation in controller
While doing profiling of controller we came across the fact that UUID generation is a blocking operation, e.g. 
```
    synchronized public void nextBytes(byte[] bytes) {
        secureRandomSpi.engineNextBytes(bytes);
    }
```
We need to think of alternatives to the current approach considering the fact that we should enable several active controllers in the near future, e.g. both cross-thread/cross-controller unique UUIDs.  

cc @mhenke1 @markusthoemmes @rabbah @cbickel

markusthoemmes:
Java's UUID.randomUUID() is known the be `synchronized` and blocking which causes thread-contention and thus performance problems.

`SecureRandom.nextBytes` is the synchronized piece, so using a `ThreadLocal` instance resolves thread contention there. Switching Java's randomness source to the non-blocking `/dev/urandom` resolves blocking of the computation itself.

Switching to `/dev/urandom` is [not a problem](https://unix.stackexchange.com/a/324210).

Supposed to fix #2747.

### DIFF
--- a/common/scala/build.gradle
+++ b/common/scala/build.gradle
@@ -26,6 +26,7 @@ dependencies {
     compile 'org.apache.zookeeper:zookeeper:3.4.10'
     compile 'org.apache.kafka:kafka-clients:0.10.2.1'
     compile 'org.apache.httpcomponents:httpclient:4.4.1'
+    compile 'com.fasterxml.uuid:java-uuid-generator:3.1.3'
     compile 'com.github.ben-manes.caffeine:caffeine:2.4.0'
     compile 'com.google.code.findbugs:jsr305:3.0.2'
 }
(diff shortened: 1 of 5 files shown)


markusthoemmes:
Okay, this looks like a super bad idea. Let's revisit.

common/scala/build.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
     compile 'org.apache.zookeeper:zookeeper:3.4.10'
     compile 'org.apache.kafka:kafka-clients:0.10.2.1'
     compile 'org.apache.httpcomponents:httpclient:4.4.1'
+    compile 'com.fasterxml.uuid:java-uuid-generator:3.1.3'
```
markusthoemmes: Apache 2.0 license.

bwmcadams:
👍🏻
Looks good to me.

PR #2792, opened 2017-09-21
Add a controller simulator to drive invoker benchmarks.

markusthoemmes:
In order to test the invokers in isolation, it is important to measure correctly. This adds a stubbed out controller which produces messages as fast as possible and additionally reads the answers of the invoker to correctly measure throughput of a single invoker instance.

I know it is unpleasant that we have benchmark tooling in different places, but at the moment it is super inconvenient to build a dependency against the main repo from another repo, code-wise.

### DIFF
--- a/tools/benchmark/src/main/scala/whisk/core/benchmark/ControllerSimulator.scala
+++ b/tools/benchmark/src/main/scala/whisk/core/benchmark/ControllerSimulator.scala
@@ -21,6 +21,7 @@ import java.nio.charset.StandardCharsets
 import java.time.Instant
 
 import akka.actor.{ActorSystem, Props}
+import breeze.stats.DescriptiveStats
 import whisk.common.{AkkaLogging, Logging, TransactionId}
 import whisk.core.WhiskConfig
 import whisk.core.connector.{ActivationMessage, CompletionMessage, MessageFeed, MessagingProvider}
@@ -33,7 +34,7 @@ import scala.concurrent.duration._
 import scala.concurrent.{Await, ExecutionContext, Future, Promise}
 
 object ControllerSimulator {
-  def between(start: Instant, end: Instant): Duration =
+  def between(start: Instant, end: Instant): FiniteDuration =
     Duration.fromNanos(java.time.Duration.between(start, end).toNanos)
 
   def main(args: Array[String]): Unit = {
@@ -46,7 +47,8 @@ object ControllerSimulator {
     val controllerToSimulate = InstanceId(sys.env.get("CONTROLLER_ID").map(_.toInt).getOrElse(0))
     val invokerToUse = InstanceId(sys.env.get("INVOKER_ID").map(_.toInt).getOrElse(0))
 
-    val actionCode = sys.env.getOrElse("ACTION_CODE", "function main() { return {}; }")
+    val actionCode =
+      sys.env.getOrElse("ACTION_CODE", "function main() { return new Promise(resolve => setTimeout(resolve, 200)); }")
 
     val topic = s"invoker${invokerToUse.toInt}"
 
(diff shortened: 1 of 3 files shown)


tools/benchmark/src/main/scala/whisk/core/benchmark/ControllerSimulator.scala
```diff
@@ -0,0 +1,165 @@
...
+  def between(start: Instant, end: Instant): Duration =
+    Duration.fromNanos(java.time.Duration.between(start, end).toNanos)
+
+  def main(args: Array[String]): Unit = {
+    implicit val as: ActorSystem = ActorSystem()
+    implicit val ec: ExecutionContext = as.dispatcher
+    implicit val log: Logging = new AkkaLogging(as.log)
+    implicit val tid: TransactionId = TransactionId.controller
+
+    val messageCount = sys.env.get("INVOCATIONS").map(_.toInt).getOrElse(10)
```
tysonnorris: I'm hoping that the intention is for this to run automated with some more significant value? I would expect to test these in the thousands at least.

markusthoemmes: You do hope right. This is just a sensible default that doesn't flood a system.

tools/benchmark/src/main/scala/whisk/core/benchmark/ControllerSimulator.scala
```diff
@@ -0,0 +1,165 @@
...
+    val identity = Identity(Subject(), EntityName("test"), AuthKey(), Set())
+
+    ExecManifest.initialize(config)
+    val action = ExecManifest.runtimesManifest
+      .resolveDefaultRuntime("nodejs:6")
+      .map { manifest =>
+        new WhiskAction(
+          namespace = identity.namespace.toPath,
+          name = EntityName("noop"),
+          exec = CodeExecAsString(manifest, """function main() { return {}; }""", None))
```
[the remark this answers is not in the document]
markusthoemmes: Let's just make it configurable. I agree a pure noop isn't representing anything really.

tools/benchmark/src/main/scala/whisk/core/benchmark/ControllerSimulator.scala
```diff
@@ -0,0 +1,167 @@
...
+    implicit val as: ActorSystem = ActorSystem()
+    implicit val ec: ExecutionContext = as.dispatcher
+    implicit val log: Logging = new AkkaLogging(as.log)
+    implicit val tid: TransactionId = TransactionId.controller
+
+    val messageCount = sys.env.get("INVOCATIONS").map(_.toInt).getOrElse(10)
+    val controllerToSimulate = InstanceId(sys.env.get("CONTROLLER_ID").map(_.toInt).getOrElse(0))
+    val invokerToUse = InstanceId(sys.env.get("INVOKER_ID").map(_.toInt).getOrElse(0))
+
+    val actionCode = sys.env.getOrElse("ACTION_CODE", "function main() { return {}; }")
```
tysonnorris: It's good this is configurable, but I still think this noop code is not a good default. Obviously it doesn't matter much (since it can now be specified in env), but it's unfortunate to not have the default be a better test, I think, considering this seems more geared at testing throughput and not minimum possible latency (where a noop action may be more appropriate). WDYT?

markusthoemmes: I'm not too sure. As this takes the controller out of the game, I think this default makes the added overhead of the invoker apparent. It's also a default anybody can agree on, it's the smallest possible action.

All other defaults we'll have to debate what a sensible default is. I'd leave it this way and then reconfigure on a per-test-basis if necessary.

markusthoemmes:
Closing as abandoned.

PR #2793, opened 2017-09-21
Mount docker/runc binaries from the host machine.

### LINKED ISSUE #2788: docker for mac or CoreOS does not have runc and causes invoker errors
observed while investigating separate issue:

```
[2017-09-20T23:59:44.351Z] [INFO] [#sid_102] [RuncClient] running /usr/bin/docker-runc pause 98a22ef65d5aac2abdee8313f3216112356c7c01bda18a38367cab05baa09f79 [marker:invoker_runc.pause_start:170209551] 
[2017-09-20T23:59:44.359Z] [ERROR] [#sid_102] [RuncClient] code: 1, stdout: , stderr: json: cannot unmarshal object into Go value of type []string [marker:invoker_runc.pause_error:170209558:8] 
```

This causes a container to be disposed and hence there is no container reuse.

markusthoemmes:
Differing docker/runc server/client configurations can lead to unexpected results and/or to failing commands. Mounting the binaries from the host machine will effectively make them even no matter what.

Fixes #2788

### DIFF
--- a/ansible/roles/invoker/files/docker-runc
+++ b/ansible/roles/invoker/files/docker-runc
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+if $2 == 'resume'; then
+  docker unpause $3
+fi
+
+if $2 == 'pause'; then
+  docker pause $3
+fi
--- a/ansible/roles/invoker/tasks/deploy.yml
+++ b/ansible/roles/invoker/tasks/deploy.yml
@@ -95,6 +95,30 @@
   with_items: "{{ invokerInfo }}"
   when: not invoker.allowMultipleInstances and item.Names[0] != "/invoker{{ groups['invokers'].index(inventory_hostname) }}"
 
+- name: "determine 'docker' binary location"
+  shell: which docker
+  register: dockerBinary
+
+- name: "determine 'docker-runc' binary location"
+  shell: which docker-runc
+  register: runcBinary
+  failed_when: false
+
+- set_fact:
+    runcMount: "-v {{ runcBinary.stdout }}:/usr/bin/docker-runc"
+  when: runcBinary.rc == 0
+
+- name: copy runc script to remote
+  copy:
+    src: "files/docker-runc"
+    dest: "{{ nginx.confdir }}"
+    mode: 111
+  when: runcBinary.rc != 0
+
+- set_fact:
+    runcMount: "-v '{{ nginx.confdir }}/docker-runc:/usr/bin/docker-runc'"
+  when: runcBinary.rc != 0
+
 - name: start invoker using docker cli
   shell: >
         docker run -d
@@ -132,6 +156,8 @@
         -e INVOKER_NUMCORE='{{ invoker.numcore }}'
         -e INVOKER_CORESHARE='{{ invoker.coreshare }}'
         -e WHISK_LOGS_DIR='{{ whisk_logs_dir }}'
+        {{ runcMount }}
+        -v {{ dockerBinary.stdout }}:/usr/bin/docker
         -v /sys/fs/cgroup:/sys/fs/cgroup
         -v /run/runc:/run/runc
         -v {{ whisk_logs_dir }}/invoker{{ groups['invokers'].index(inventory_hostname) }}:/logs
--- a/core/invoker/Dockerfile
+++ b/core/invoker/Dockerfile
@@ -1,16 +1,5 @@
 FROM scala
 
-ENV DOCKER_VERSION 1.12.0
-
-# Uncomment to fetch latest version of docker instead: RUN wget -qO- https://get.docker.com | sh
-# Install docker client
-RUN wget --no-verbose https://get.docker.com/builds/Linux/x86_64/docker-${DOCKER_VERSION}.tgz && \
-tar --strip-components 1 -xvzf docker-${DOCKER_VERSION}.tgz -C /usr/bin docker/docker && \
-tar --strip-components 1 -xvzf docker-${DOCKER_VERSION}.tgz -C /usr/bin docker/docker-runc && \
-rm -f docker-${DOCKER_VERSION}.tgz && \
-chmod +x /usr/bin/docker && \
-chmod +x /usr/bin/docker-runc
-
 COPY build/distributions/invoker.tar ./
 RUN tar xf invoker.tar && \
 rm -f invoker.tar


ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -95,6 +95,30 @@
...
+  failed_when: false
+
+- set_fact:
+    runcMount: "-v {{ runcBinary.stdout }}:/usr/bin/docker-runc"
+  when: runcBinary.rc == 0
+
+- name: copy runc script to remote
+  copy:
+    src: "files/docker-runc"
+    dest: "{{ nginx.confdir }}"
```
rabbah: nginx?

ansible/roles/invoker/files/docker-runc
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
ddragosd: maybe we can rename it to `runc-simulator` or something similar as at a quick glance it may look like it's the real `runc` ?

ddragosd: Or this is what it should be used always according to https://github.com/apache/incubator-openwhisk/blob/master/core/invoker/src/main/scala/whisk/core/containerpool/docker/RuncClient.scala#L42 ?

markusthoemmes: We can certainly rename on mount, but yes: This is what should be used according to the code itself.

It might after all be more worth to do what you suggested and detect a missing runc and fallback to docker.

ansible/roles/invoker/tasks/deploy.yml
markusthoemmes: Just for quick prototyping, I needed a folder which was already there and setup.

ansible/roles/invoker/files/docker-runc
ddragosd: > It might after all be more worth to do what you suggested and detect a missing runc and fallback to docker.

and do this from within this script instead of the scala code ?

markusthoemmes: That would be scala code. The added benefit would be, that the logs clearly show you're using docker vs. runc which would surface config errors in production where you really want to use runc and not scratch your head why pause/unpause get slow.

ddragosd: 👍

ddragosd: so in that case I can turn my changes into a PR, and, if I'm reading the PR correctly and the `docker-runc` won't be added, things should work.

markusthoemmes: Sure you can! Go for it 😬

csantanapr:
@ddragosd 
Mounting `/usr/local/bin/docker` for some reason doesn't work on Mac

This is what I tried
```
😄  $ docker run --rm -it -v '/tmp/nginx/docker-runc:/usr/bin/docker-runc' -v '/usr/local/bin/docker:/usr/bin/docker' whisk/invoker:latest /bin/sh
docker: Error response from daemon: Mounts denied: EOF.
ERRO[0000] error waiting for container: context canceled
~/dev/whisk/git/apache/incubator-openwhisk (pr-2793)
😄  $ ls -l /usr/local/bin/docker
lrwxr-xr-x  1 csantana23  staff  70 Sep 18 15:01 /usr/local/bin/docker -> /Users/csantana23/Library/Group Containers/group.com.docker/bin/docker
😄  $ ls -l "/Users/csantana23/Library/Group Containers/group.com.docker/bin/docker"
lrwxr-xr-x  1 csantana23  staff  54 Sep 18 15:01 /Users/csantana23/Library/Group Containers/group.com.docker/bin/docker -> /Applications/Docker.app/Contents/Resources/bin/docker
😄  $ docker run --rm -it -v '/tmp/nginx/docker-runc:/usr/bin/docker-runc' -v '/Applications/Docker.app/Contents/Resources/bin/docker:/usr/bin/docker' whisk/invoker:latest /bin/sh
docker: Error response from daemon: Mounts denied:
The path /Applications/Docker.app/Contents/Resources/bin/docker
is not shared from OS X and is not known to Docker.
You can configure shared paths from Docker -> Preferences... -> File Sharing.
See https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.
😄  $ docker run --rm -it -v '/tmp/nginx/docker-runc:/usr/bin/docker-runc' -v '/Applications/Docker.app/Contents/Resources/bin/docker:/usr/bin/docker' whisk/invoker:latest /bin/sh
# /usr/bin/docker info
/usr/bin/docker: 1: /usr/bin/docker: ������: not found
/usr/bin/docker: 1: /usr/bin/docker: �: not found
/usr/bin/docker: 1: /usr/bin/docker: {��: not found
/usr/bin/docker: 1: /usr/bin/docker: ��: not found
/usr/bin/docker: 2: /usr/bin/docker: �H__PAGEZER__TEXT�
                                                       __text__TEXT0$���0$�__stubs__TEXT�ϊ���__stub_helper__TEXT�ъ�ъ�__rodata__TEXT�Ԋx�+�Ԋ__typelink__TEXT{�__itablink__TEXTP�p�__gosymtab__TEXT����__gopclntab__TEXT�� �>��__cstring__TEXT�$�l�$�__const__TEXTP,��P,�__unwind_info__TEXT
                                                                                                                             0��
                                                                                                                                0�__eh_frame__TEXT�2�-�2�h__DATA�: File name too long
/usr/bin/docker: 1: /usr/bin/docker: Syntax error: end of file unexpected (expecting ")")
```

ddragosd:
@csantanapr I'm experimenting with docker for mac with:

```
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - /var/lib/docker/containers:/containers
      - /usr/bin/docker:/usr/bin/docker
      - /run/runc:/run/runc
      - /usr/bin/runc:/usr/bin/docker-runc
```

My current problem is that `runc` doesn't see all the processes and believes containers are stopped. I'm trying to add `privileged`, `pid:host` and `userns:host` hoping it helps.

csantanapr:
@ddragosd hum this one works for you?
```
/usr/bin/docker:/usr/bin/docker
```

csantanapr:
make sure you are not installing docker cli in the invoker Dockerfile
```
$ cat core/invoker/Dockerfile
FROM scala

COPY build/distributions/invoker.tar ./
RUN tar xf invoker.tar && \
rm -f invoker.tar

EXPOSE 8080
```

csantanapr:
OK I'm on the same page now, we are mounting from the VM 
```
- /usr/bin/docker:/usr/bin/docker
- /run/runc:/run/runc
```

ddragosd:
I got it to work with:
```
    volumes:
      - ~/tmp/openwhisk/invoker/logs:/logs
      - /var/run/docker.sock:/var/run/docker.sock
      - /var/lib/docker/containers:/containers
      - /usr/bin/docker:/usr/bin/docker
      - /run/runc:/run/runc
      - /usr/bin/runc:/usr/bin/docker-runc
      - /sys/fs/cgroup/freezer:/sys/fs/cgroup/freezer
```

`/sys/fs/cgroup/freezer` was interesting to find out about.

ddragosd:
It works even with the existing invoker container. I'll open a PR in https://github.com/apache/incubator-openwhisk-devtools

ddragosd:
PR opened in https://github.com/apache/incubator-openwhisk-devtools/pull/56

ansible/roles/invoker/files/docker-runc
ddragosd: created PR https://github.com/apache/incubator-openwhisk/pull/2796

csantanapr:
@ddragosd your going to open a PR here to fix Docker for Mac?

ddragosd:
@csantanapr I've been testing docker for mac with docker-compose. I'm trying it now w/ ansible. if it works I should be able to submit the PR.

ddragosd:
@csantanapr I've opened PR https://github.com/apache/incubator-openwhisk/pull/2799 illustrating the minimum changes.

@markusthoemmes I didn't update the docker image with the invoker to avoid conflicts with this PR.

ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -95,6 +95,30 @@
   with_items: "{{ invokerInfo }}"
   when: not invoker.allowMultipleInstances and item.Names[0] != "/invoker{{ groups['invokers'].index(inventory_hostname) }}"
 
+- name: "determine 'docker' binary location"
+  shell: which docker
+  register: dockerBinary
+
+- name: "determine 'docker-runc' binary location"
+  shell: which docker-runc
```
ddragosd: I think we can also try ```which runc || which docker-runc```

markusthoemmes:
Closing this as @ddragosd and @dgrove-oss took over the issue.

PR #2795, opened 2017-09-21
enable concurrent activation processing

tysonnorris:
- requires rebuilding nodejs6Action
- considers all actions support concurrency
- log collection is disabled for now due to interleaved logging when running concurrent activations in a single container

To validate operation, I've tested with the throughput.sh script using an action that has an async response after 175ms. In this type of scenario, I see throughput 2x over previous sequential processing (e.g. 10-15 rps before, 20-40 rps after).

Note that as discussed offline, in current form, with kafka between controller and invoker, there is no way to account for non-concurrent actions blocking resource usage by those that support concurrency, so currently the code assumes that all actions support concurrency, and the ContainerPoolScheduler accounts for concurrency settings per action, while MessageFeed only accounts for concurrency across the invoker in general (via maximumHandlerCapacity)


markusthoemmes:
Read briefly over the code, good for a start 👍 

The changes in ContainerProxy look good to me on first glance. I wonder if  we can simplify on the scheduling bits though: What if we tracked the number of concurrent invocation always and only update to Busy, if that number reaches `max`. Everything else will put the container into `Free`.

In the non-concurrent case that number can be set to 1. I think we could get away without any changes to the scheduler in that case, since the ContainerPool itself keeps the state-tracking consistent. WDYT?

tysonnorris:
@markusthoemmes I think you mean to keep containers in freePool, and only move to busyPool when reaching max? (I'm not sure that WorkerState is used anymore?)

If we do that, then ContainerPool.remove needs to pay attention to the concurrent invocations, and there will be potentially a lot of shuffling between freePool and busyPool when traffic hovers around max. In the other case (where container moves to busy after 1), there is potentially a lot of shuffling when there is a stream of sequential events (1 at a time), so maybe this doesn't matter. 

I'll see if I can change the remove to respect the concurrent activation count and avoid complicating the scheduling as you describe.

dubeejw:
@tysonnorris, is this PR still valid?

tysonnorris:
@dubeejw yes still valid, but will need to be updated to resolve conflicts; I'll get back to it asap.

markusthoemmes:
Assigning to @tysonnorris until out of WIP.

dubee:
@tysonnorris, looks like this one needs a rebase.

upgle:
How is it coming along?

tysonnorris:
@markusthoemmes @style95 @upgle this is ready for review

performance/preparation/create-async.sh
```diff
@@ -0,0 +1,17 @@
...
+# Host to use. Needs to include the protocol.
+host=$1
+# Credentials to use for the test. USER:PASS format.
+credentials=$2
+# Name of the action to create and test.
+action=$3
+
+# create a noop action
+echo "Creating action $action"
+curl -k -u "$credentials" "$host/api/v1/namespaces/_/actions/$action" -XPUT -d '{"namespace":"_","name":"test","exec":{"kind":"nodejs:default","code":"function main(){ return new Promise(function(resolve, reject) {     setTimeout(function() {       resolve({ done: true });     }, 175);   }) }"}}' -H "Content-Type: application/json"
```
markusthoemmes: Should we rather parametrize the existing `create.sh` on the action-code and `cat` that from a file?

performance/wrk_tests/throughput-async.sh
```diff
@@ -0,0 +1,31 @@
...
+encodedAuth=$(echo "$credentials" | tr -d '\n' | base64 | tr -d '\n')
+docker run --pid=host --userns=host --rm -v "$currentDir":/data williamyeh/wrk \
+  --threads "$threads" \
+  --connections "$concurrency" \
+  --duration "$duration" \
+  --header "Authorization: basic $encodedAuth" \
+  "$host/api/v1/namespaces/_/actions/$action?blocking=true" \
+  --latency \
+  --timeout 10s \
+  --script post.lua
```
markusthoemmes: Same here, shall we parametrize on the action-file name so we can reuse one script?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -107,6 +106,7 @@ class ContainerProxy(
   implicit val ec = context.system.dispatcher
   implicit val logging = new AkkaLogging(context.system.log)
   var rescheduleJob = false // true iff actor receives a job but cannot process it because actor will destroy itself
+  var activeActivationCount = 0
```
markusthoemmes: WDYT about making this part of `WarmedData` or even `ContainerData`? That would eschew the need for the `ActivationCounter` and make other parts easier as well (I'll ad comments for posterity)

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -266,11 +273,14 @@ object ContainerPool {
...
    * @param pool a map of all free containers in the pool
+   * @param checkConcurrent if true only consider containers not processing other concurrent activations (otherwise disregard concurrent activation counts)
    * @return a container to be removed iff found
    */
-  protected[containerpool] def remove[A](pool: Map[A, ContainerData]): Option[A] = {
+  protected[containerpool] def remove[A](pool: Map[A, ContainerData], checkConcurrent: Boolean): Option[A] = {
+    // Try to find a Free container that does NOT have any active activations AND is initialized with any OTHER action
     val freeContainers = pool.collect {
-      case (ref, w: WarmedData) => ref -> w
+      case (ref, w: WarmedData) if !checkConcurrent || ActivationCounter.getOrElse(ref, 0) == 0 =>
```
markusthoemmes: This becomes

```scala
case (ref, w: WarmedData) if w.activationCount == 0 =>
```

And you can remove the boolean flag if we can add the counter to `WarmedData`.

performance/wrk_tests/throughput-async.sh
tysonnorris: you mean for `post.lua`? does this need to be parameterized?

tysonnorris: Maybe you meant to parameterize throughput.sh script... will do that

rabbah:
Tyson do we need to guard the annotation to apply only to runtimes their can support concurrent activations? Will this need an attribute in the runtime manifest?

csantanapr:
@tysonnorris we have user extending the nodejs as blackbok docker actions because they have node_modukes larger than 50MB, so this user might want to use concurrency. 

There might kinds that do not support concurrency. 
I would add tests in the respective runtimes tontest that runtimes images proxies defently can handle concurrent /runs

csantanapr:
+1 to have runtimemanifest attribute

tysonnorris:
For this incremental PR, operator who enables concurrency is responsible for:
- log collection by some other means
- only using appropriate images

I would address the additional kind validations, action annotations,  etc in separate PRs

rabbah:
Having the predicate visible in the runtime manifest would be convenient for clients (wsk cli) to validate and fail early if an annotation is added to an unsupported runtime. Black box I think is exempted, user beware.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -53,14 +52,24 @@ case object Paused extends ContainerState
 case object Removing extends ContainerState
 
 // Data
-sealed abstract class ContainerData(val lastUsed: Instant)
+sealed abstract class ContainerData(val lastUsed: Instant, val activeActivationCount: Int = 0)
```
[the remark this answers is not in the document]
tysonnorris: yes - fixing that

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -205,10 +213,35 @@ class ContainerProxy(
...
+
+      if (newData.activeActivationCount > 0) {
+        stay using newData
+      } else {
+        logging.info(this, "going to ready after load completed")
+        goto(Ready) using newData
+      }
+
+    case Event(job: Run, data: WarmedData)
+        if stateData.activeActivationCount < poolConfig.maxConcurrent && !rescheduleJob => //if there was a delay, or a failure on resume, skip the run
```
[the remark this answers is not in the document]
tysonnorris: These should be the same? I thought that data was only a matching convenience. Will change it but am curious if this is problematic?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -205,10 +213,35 @@ class ContainerProxy(
...
+
+      implicit val transid = job.msg.transid
+      val newData = data.incrementActive
+
+      initializeAndRun(data.container, job)
+        .map(_ => WarmedData(data.container, job.msg.user.namespace, job.action, Instant.now))
```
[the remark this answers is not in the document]
tysonnorris: This is a confusing part of `ContainerProxy` for me  - that `WarmedData` is used for both event _and_ state. These ones that are `.pipeTo` are events, and do not track the current state. At some point I would rename these events to something like `RunComplete` (with a `WarmedData` val), and still used `WarmedData` for state.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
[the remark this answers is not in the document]
markusthoemmes: They should, yes. I tend to prefer the local value though, makes it more decoupled from the "outside world".

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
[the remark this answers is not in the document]
markusthoemmes: Agreed.

common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
```diff
@@ -31,7 +31,12 @@ case class ContainerArgsConfig(network: String,
                                dnsServers: Seq[String] = Seq.empty,
                                extraArgs: Map[String, Set[String]] = Map.empty)
 
-case class ContainerPoolConfig(numCore: Int, coreShare: Int) {
+case class ContainerPoolConfig(numCore: Int,
+                               coreShare: Int,
+                               maxConcurrent: Int = 1,
+                               concurrentPeekFactor: Double = 0.5) {
```
markusthoemmes: Please don't define defaults in code. That will lead to multiple cases where defaults are applied which will rot over time. The ones in `application.conf` should be enough?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -127,8 +126,11 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
 
       createdContainer match {
         case Some(((actor, data), containerState)) =>
-          busyPool = busyPool + (actor -> data)
-          freePool = freePool - actor
+          //only move to busyPool if max reached
+          if (data.activeActivationCount + 1 >= poolConfig.maxConcurrent) {
```
markusthoemmes: Don't know where to put this, so doing this here:

I have one general question on the state-keeping: I think we update the state in the ContainerProxy asynchronously. That is, the proxy sends a "WarmedData" message, which we then use to update state.

Do we need to "predict" that update in the ContainerPool, as in: increment WarmedData on scheduling locally and updating that with the "actual" value once the event happens from ContainerProxy?

I'm mostly wondering if we otherwise have racy counting going on. Thoughts?

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -725,7 +824,9 @@ class ContainerProxyTests
   /**
    * Implements all the good cases of a perfect run to facilitate error case overriding.
    */
-  class TestContainer extends Container {
+  class TestContainer(initPromise: Option[Promise[Interval]] = None,
+                      runPromises: Option[Seq[Promise[(Interval, ActivationResponse)]]] = None)
```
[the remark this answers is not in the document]
tysonnorris: I guess so, it just seems clearer to use `Option` here. I can change it, I don't think it will break anything.

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -754,7 +855,11 @@ class ContainerProxyTests
       initializeCount += 1
       initializer shouldBe action.containerInitializer
       timeout shouldBe action.limits.timeout.duration
-      Future.successful(initInterval)
+
+      initPromise match {
+        case Some(promise) => promise.future
+        case None          => Future.successful(initInterval)
+      }
```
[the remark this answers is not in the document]
tysonnorris: nice thanks!

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -770,7 +875,12 @@ class ContainerProxyTests
       // a freshly computed deadline, as they get computed slightly after each other
       deadline should (be <= maxDeadline and be >= Instant.now)
 
-      Future.successful((runInterval, ActivationResponse.success()))
+      runPromises match {
+        case Some(promises) =>
+          promises(runCount - 1).future
+        case None =>
+          Future.successful((runInterval, ActivationResponse.success()))
+      }
```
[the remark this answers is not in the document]
tysonnorris: like it

common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
tysonnorris: yep 👍

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
tysonnorris: So the values in ContainerPool will always be stale. If we update the preemptively, they will be accurate momentarily, and inaccurate again after the first in-flight activation completes while the other in-flights are still processing. We could do that, but it seems "even stranger" to not just let the ContainerPool receive the updates. 

Checking for passing the concurrency threshold is "mostly important" for the case where maxConcurrent=1, and in other cases, your mileage will vary as far as "actual concurrency", but pool will try to prevent passing that value, but it will always be working with stale data so will not be 100% precise. I think this is OK. WDYT?

tysonnorris:
@markusthoemmes I have merged in the wsk api changes to support concurrency via `ActionLimits`.

I added a test (see `ConcurrencyTests`)  - which requires:
* enabling akka client
* enabling runtimes via env var (only respected by nodejs runtime currently)

I currently enabled this by adding
```-e container_pool_akka_client=true -e runtimes_enable_concurrency=true```
to the `setupSystem.sh` setup script.

This implies that the "REQUIRE_SYSTEM" tests will run with this config. 

The only other way I can think of is to enable a 4th travis job, that also runs system tests (may ONLY the concurrency tests?) that does the deploy with the variant that enables concurrency. 

I hate the idea of adding another job, just for the concurrency tests. 

WDYT?

**Comment by tysonnorris** [comment]
I'm seeing test failures related to akka client (enabled on system tests) + `actionproxy.py` + `Connection: close` headers; results in: 
```
[DockerContainer] running result: ConnectionError(akka.http.impl.engine.client.OutgoingConnectionBlueprint$UnexpectedConnectionClosureException: The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests)
```

I reproduced this locally (and via travis) with swift and python runtimes; investigating what we might need to change in actionproxy.py (ideas appreciated)

**Comment by tysonnorris** [comment]
`ConnectionError`s are resolved now, using keepalive = false (vs `Connection:close` header). 

@markusthoemmes please have a look when possible.

@rabbah @csantanapr this does NOT currently implement any change to runtimes manifest, but since none of this is enabled by default, it is low (or none) impact.

File: core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala [hunk 13]
```diff
@@ -137,7 +137,18 @@ import scala.util.{Failure, Success}
  * - If a loadbalancer leaves or joins the cluster, all state is removed and created from scratch. Those events should
  *   not happen often.
  */
-class ShardingContainerPoolBalancer(config: WhiskConfig, controllerInstance: ControllerInstanceId)(
+class ShardingContainerPoolBalancer(
+  config: WhiskConfig,
+  controllerInstance: ControllerInstanceId,
+  private val feedFactory: (ActorRefFactory, MessagingProvider, (Array[Byte]) => Future[Unit]) => ActorRef,
```
chetanmeh: @tysonnorris May be we can use traits here instead of function arguments. It makes it easy in IDE to see what is the implementation and move back and forth easily

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -302,8 +332,10 @@ object ContainerPool {
                                            invocationNamespace: EntityName,
                                            idles: Map[A, ContainerData]): Option[(A, ContainerData)] = {
     idles.find {
-      case (_, WarmedData(_, `invocationNamespace`, `action`, _)) => true
-      case _                                                      => false
+      case c @ (_, WarmedData(_, `invocationNamespace`, `action`, _, _))
+          if c._2.activeActivationCount < action.limits.concurrency.maxConcurrent =>
```
chetanmeh: Can be simplified a bit

```scala
      case (_, WarmedData(_, `invocationNamespace`, `action`, _, activeActivationCount))
          if activeActivationCount < action.limits.concurrency.maxConcurrent =>
```

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -132,14 +133,22 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
...
-                      .getOrElse(createContainer(r.action.limits.memory.megabytes.MB), "recreated")))
+                      .map(container => (container, "recreatedPrewarm", container._2.activeActivationCount))
+                      .getOrElse(createContainer(r.action.limits.memory.megabytes.MB), "recreated", 0)))
           } else None
 
         createdContainer match {
-          case Some(((actor, data), containerState)) =>
-            busyPool = busyPool + (actor -> data)
-            freePool = freePool - actor
+          case Some(((actor, data), containerState, activeActivations)) =>
```
chetanmeh: I think `activeActivationCount` in `data` which is `ContainerData` == `activeActivations`. In that case we can avoid extracting `activeActivationCount` in steps before

tysonnorris: will try to remove that 👍

tysonnorris:
@chetanmeh any other comments on this?

.travis.yml
```diff
@@ -75,11 +75,17 @@ jobs:
       name: "System Tests"
     - script:
         - ./tests/performance/preparation/deploy.sh
-        - TERM=dumb ./tests/performance/wrk_tests/latency.sh "https://172.17.0.1:10001" "$(cat ansible/files/auth.guest)" 2m
-        - TERM=dumb ./tests/performance/wrk_tests/throughput.sh "https://172.17.0.1:10001" "$(cat ansible/files/auth.guest)" 4 2 2m
+        - TERM=dumb ./tests/performance/wrk_tests/latency.sh "https://172.17.0.1:10001" "$(cat ansible/files/auth.guest)" ./performance/preparation/actions/noop.js 2m
```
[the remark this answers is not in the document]
tysonnorris: sorry, these are left over from when the performance directory was moved, long ago...

performance/gatling_tests/src/gatling/resources/data/nodeJSAsyncAction.js
```diff
@@ -0,0 +1,12 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more contributor
```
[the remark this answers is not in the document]
tysonnorris: deleted the dupe

tools/travis/setupSystem.sh
```diff
@@ -25,7 +25,7 @@ ROOTDIR="$SCRIPTDIR/../.."
 
 cd $ROOTDIR/ansible
 
-$ANSIBLE_CMD openwhisk.yml
+$ANSIBLE_CMD -e container_pool_akka_client=true -e runtimes_enable_concurrency=true openwhisk.yml
```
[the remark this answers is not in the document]
tysonnorris: Do you mean to define vars in `runSystemTests.sh`? or in travis settings? I'm not sure moving them to the other scripts is less confusing, but I can do that; not sure we should define them in travis settings though.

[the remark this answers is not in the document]
cbickel: No, I mean moving it to the environment in ansible, that is used in travis.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -212,10 +222,34 @@ class ContainerProxy(
     // and we keep it in case we need to destroy it.
     case Event(data: PreWarmedData, _) => stay using data
 
-    // Run was successful
-    case Event(data: WarmedData, _) =>
+    // Init was successful
+    case Event(data: WarmedData, _: PreWarmedData) =>
```
chetanmeh: Note for myself - In `Started` state following sequence happens

1. proxy transitions to `Running` with `PreWarmedData` (activeCount = 1). 
2. It also then sends msg `WarmedData` (count = 1) in `initializeAndRun` . Assuming this is first `Run` being processed
3. It also then sends msg `WarmedData` with count = 0 

So each call to `initializeAndRun` end up sending 2 `WarmedData` messages. One with count = 0 and one with count at time of start of message

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -212,10 +222,34 @@ class ContainerProxy(
...
+      //in case concurrency supported, multiple runs can begin as soon as init is complete
       context.parent ! NeedWork(data)
-      goto(Ready) using data
+      stay using data
+
+    // Run was successful
+    case Event(_: WarmedData, s: WarmedData) =>
```
chetanmeh: @tysonnorris  Here use of `WarmedData` both as message and state is bit confusing. May be we use new event type

1. For successful - `RunCompleted`
2. For init - `Init`

Further we set the `activeActivationCount` in `PreWarmedData` to 0 and make it constant i.e. not provide a way to change it from 0 for `PreWarmedData`.

tools/travis/setupSystem.sh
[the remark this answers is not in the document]
tysonnorris: ok, like this? https://github.com/apache/incubator-openwhisk/pull/2795/commits/ebf7a18e53b758f1a0b75d55673e1b8c24db3c39

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -101,15 +102,16 @@ class InvokerReactive(
...
-  private val consumer = msgProvider.getConsumer(
-    config,
-    topic,
-    topic,
-    maximumContainers,
-    maxPollInterval = TimeLimit.MAX_DURATION + 1.minute)
+
+  //number of peeked messages - increasing the concurrentPeekFactor improves concurrent usage, but adds risk for message loss in case of crash
+  private val maxPeek =
+    math.max(maximumContainers, (maximumContainers * limitsConfig.max * poolConfig.concurrentPeekFactor).toInt)
```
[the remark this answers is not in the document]
tysonnorris: I didn't quite follow your math, is there some typos? In any case, this will vary based on load, but the defaults should be better. 
I will change default max to 1 (concurrency disabled by default; only enabled by default in "local" env for test runs)

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -52,16 +52,27 @@ case object Paused extends ContainerState
...
+                                    val memoryLimit: ByteSize,
+                                    val activeActivationCount: Int = 0)
 case class NoData() extends ContainerData(Instant.EPOCH, 0.B)
 case class MemoryData(override val memoryLimit: ByteSize) extends ContainerData(Instant.EPOCH, memoryLimit)
-case class PreWarmedData(container: Container, kind: String, override val memoryLimit: ByteSize)
-    extends ContainerData(Instant.EPOCH, memoryLimit)
+case class PreWarmedData(container: Container,
+                         kind: String,
+                         override val memoryLimit: ByteSize,
+                         override val activeActivationCount: Int = 0)
```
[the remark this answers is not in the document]
tysonnorris: Agreed, will work on that; 
PreWarmedData may be 1 when emitted during cold start, to signal to parent that it is prewarmed, but already has active activations in flight, so that pool can use it asap, but also respect the concurrency limit.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
tysonnorris: for 3 it should send WarmedData with count 0 only to itself for completion notification, and to parent only in case where all active activations are completed. There is a confusing aspect here which is that WarmedData is used BOTH for signaling completion (active count is ignored) on internal state machine AND for signaling "NeedWork" to parent (active count is important).

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
tysonnorris: Yes, just mentioned that above - I will take a stab at replacing it with RunCompleted.

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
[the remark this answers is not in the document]
chetanmeh: Let me try again

```scala
val maximumContainers = (poolConfig.userMemory / MemoryLimit.minMemory).toInt
```

Here
* `MemoryLimit.minMemory` = 128 MB
* `poolConfig.userMemory` = x MB

So `maximumContainers` = x/128

Now
```scala
val maxPeek =
    math.max(maximumContainers, (maximumContainers * limitsConfig.max * poolConfig.concurrentPeekFactor).toInt)
```

Where earlier the defauls were
* `limitsConfig.max` = 500
* `poolConfig.concurrentPeekFactor` = 0.5

So 
```
maxPeek = max(x/128, x/128 * 500 * 0.5)
maxPeek = max(128x, ~ 2x)
```

So if `poolConfig.userMemory` is 32 GB = 32768 MB then `maxPeek` = 65536. Which is a high number. Considering a worst case of 1 MB per activation it results in loading 65 GB of data. So may be we introduce an explicit config for `maxPeek`?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -346,7 +381,8 @@ object ContainerPool {
   def props(factory: ActorRefFactory => ActorRef,
             poolConfig: ContainerPoolConfig,
             feed: ActorRef,
-            prewarmConfig: List[PrewarmingConfig] = List.empty) =
+            prewarmConfig: List[PrewarmingConfig] = List.empty,
+            maxConcurrent: Int = 1) =
```
chetanmeh: `maxConcurrent` not used ?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -197,11 +230,9 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
     // Container got removed
     case ContainerRemoved =>
       freePool = freePool - sender()
-      busyPool.get(sender()).foreach { _ =>
```
chetanmeh: This change does not looks like required for concurrency support. So better to retain the existing approach of sending message only if `busyPool` had `sender`

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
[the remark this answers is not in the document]
tysonnorris: Would it be better to just define a different value for concurrentPeekFactor? 

My point is that regardless of specifying directly a value for maxPeek, or indirectly a value based on existing concurrentPeekFactor, there is still going to be potential for exhausting the system. e.g. we could set it to 0.1 - if you think that specifying maxPeek=12800 is easier to understand than concurrentPeekFactor=0.1, we can do that. But in neither case does it directly translate to actual behavior in the system since the message peeking a) pays no attention to differences in actions or activations request/response memory impact and b) is required to be increased to allow concurrent activations to be processed. This is a known tradeoff of enabling concurrency for now.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
tysonnorris: oops leftover from refactoring

tysonnorris:
@dubee @cbickel can you run another PG for this PR?

tysonnorris:
@dubee any update on PG?

common/scala/src/main/scala/whisk/core/entity/Limits.scala
```diff
@@ -72,8 +75,9 @@ protected[core] object ActionLimits extends ArgNormalizer[ActionLimits] with Def
       val time = TimeLimit.serdes.read(obj.get("timeout") getOrElse deserializationError("'timeout' is missing"))
       val memory = MemoryLimit.serdes.read(obj.get("memory") getOrElse deserializationError("'memory' is missing"))
       val logs = obj.get("logs") map { LogLimit.serdes.read(_) } getOrElse LogLimit()
+      val concurrency = obj.get("concurrency") map { ConcurrencyLimit.serdes.read(_) } getOrElse ConcurrencyLimit()
```
rabbah: note to self: check there is a test for successful serdes of previous schema.

common/scala/src/main/scala/whisk/core/entity/Limits.scala
```diff
@@ -72,8 +75,9 @@ protected[core] object ActionLimits extends ArgNormalizer[ActionLimits] with Def
...
 
-      ActionLimits(time, memory, logs)
+      ActionLimits(time, memory, logs, concurrency)
```
rabbah: is there/will there be a downstream PR to add this to the CLI?

tysonnorris: indeed; starting with client-go: https://github.com/apache/incubator-openwhisk-client-go/pull/94

rabbah:
@tysonnorris thanks for being patient with this and working through all the nuances. Looking forward to feedback from production on this feature and what the user experience will be like from real codes.

cbickel:
@tysonnorris Thanks a lot for this PR :) Great job.

style95:
@tysonnorris Thank you for your efforts.
I think many people can benefit from this : ) 👍

markusthoemmes:
Thanks a lot @tysonnorris a huge piece of good work 🎉 .

Also thanks to @rabbah, @cbickel and @chetanmeh for reviewing and @dubee for handling pull-request mechanics. Great team effort 🚀

csantanapr:
Thanks @tysonnorris This is a great milestone to achieve !!! 🎉 
Also to all reviewers 👍

PR #2797, opened 2017-09-22
Use non-blocking computation of cache values.

markusthoemmes:
Currently, each time we try to get a value out of our cache `putIfAbsent` is called, which causes thread-contention. Instead, this uses `computeIfAbsent` which does a `get` first and thus is non-blocking for getting cached values under load.

## Testcode:

```scala
object Test {
  import scala.concurrent.ExecutionContext.Implicits.global

  implicit val as = ActorSystem()

    val cache: ConcurrentMapBackedCache[String] = new ConcurrentMapBackedCache(
      Caffeine
        .newBuilder()
        .asInstanceOf[Caffeine[Any, Future[String]]]
        .expireAfterWrite(5, TimeUnit.MINUTES)
        .softValues()
        .build()
        .asMap())

    (1 to 100000000).par.foreach { _ =>
      cache.apply("testKey", () => akka.pattern.after(3.seconds, as.scheduler)(Future.successful("test2")))
    }

    as.terminate()
  }
}
```

## With our current code:

image

image

## With the updated code:

image

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
+++ b/common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
@@ -22,20 +22,15 @@
 
 package whisk.core.database
 
-import java.util.concurrent.ConcurrentMap
-import java.util.concurrent.TimeUnit
+import java.util.concurrent.{ConcurrentMap, TimeUnit}
 import java.util.concurrent.atomic.AtomicReference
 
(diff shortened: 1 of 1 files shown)


markusthoemmes:
PG1 2099 🔵

ben-manes:
Just added the `putIfAbsent` optimization to Caffeine based on seeing this. Please do comment upstream when you find gaps like this. Thanks!

PR #2802, opened 2017-09-25
Akka Usage Cleanup for Actor State safety

bwmcadams:
( Relates to enhancement issue I opened #2803 )
Some suggested cleanups of Akka Actors to improve the safety of Actors' internal state. Should make the actor usage safer and be more in line with best practice.

Specifically, the use of immutable collections stored in mutable `var`s is safer than immutable `val`s with mutable collections. With an immutable variable (`val`) pointing to a mutable object (such as a collection, e.g., `Vector` or `Map`), should the variable be closed over or sent in a message accidentally, the Actor's internal state could be inadvertently mutated. With a *mutable* variable (`var`) pointing at an immutable object (collection, etc.) should this state accidentally leak in any fashion, it won't be changeable from the outside.


core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -258,7 +257,7 @@ protected[actions] object ActivationFinisher {
     // when the future completes, self-destruct
     promise.future.andThen { case _ => shutdown() }
 
-    val preemptiveMsgs: Buffer[Cancellable] = Buffer.empty
+    var preemptiveMsgs = Vector.empty[Cancellable]
```
markusthoemmes: Mega nit: Should this be an `immutable.Buffer` for its constant append time?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -105,8 +105,8 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
 
       container match {
         case Some((actor, data)) =>
-          busyPool.update(actor, data)
-          freePool.remove(actor)
+          busyPool = busyPool + (actor → data)
```
markusthoemmes: We don't use ASCII arrows in our codebase.

bwmcadams: Fixing; was the default setup for my editor. longterm, having a `scalafmt` style is probably a good idea. I know IntelliJ can import it, suspect Eclipse, etc can as well.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
bwmcadams: There isn't an Immutable Buffer in the Scala collection library. 

[`Vector` is effectively constant on all operations](http://docs.scala-lang.org/overviews/collections/performance-characteristics.html) and the best alternative in the immutable side.

markusthoemmes: Gnah, I of course meant the immutable.Queue. My bad.

bwmcadams: You're right; we are only appending to the datastructure. `Queue`'s append is constant time versus `Vector`s which is effectively constant.

bwmcadams: As you say, performance issue isn't there so I'd probably still go with a `Vector` personally.

common/scala/src/main/scala/whisk/core/connector/MessageConsumer.scala
```diff
@@ -202,7 +206,11 @@ class MessageFeed(description: String,
   private def sendOutstandingMessages(): Unit = {
     val occupancy = outstandingMessages.size
     if (occupancy > 0 && handlerCapacity > 0) {
-      val (topic, partition, offset, bytes) = outstandingMessages.dequeue()
+      // Easiest way with an immutable queue to cleanly dequeue
+      // Head is the first elemeent of the queue, desugared w/ a assignment pattern
```
rabbah: a assignment -> an assignment

rabbah:
+1: Thanks for your first contribution and welcome on board 🎉

common/scala/src/main/scala/whisk/core/connector/MessageConsumer.scala
bwmcadams: ? Not sure what you meant here.

rabbah: typo: desugared w/ a assignment -> desugared w/ an assignment

bwmcadams:
I'm having trouble figuring out why the build is failing here? The only error I see is a compiler timeout.

rabbah:
```
> travis 279666212
whisk.core.cli.test.WskEntitlementTests > Wsk Package Action should get and invoke an action from package FAILED
whisk.core.cli.test.WskEntitlementTests > Wsk Package Action should invoke an action sequence from package FAILED
whisk.core.cli.test.WskEntitlementTests > Wsk Package Action should invoke a packaged action not owned by the subject to get the subject's namespace FAILED
whisk.core.cli.test.WskEntitlementTests > Wsk Trigger Feed should not create a trigger with timeout error when feed fails to initialize FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk CLI usage should allow a 3 part Fully Qualified Name (FQN) without a leading '/' FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should invoke an action that exits during initialization and get appropriate error FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should invoke an action that hangs during initialization and get appropriate error FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should invoke an action that exits during run and get appropriate error FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should retrieve the last activation using --last flag FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should ensure timestamp and stream are stripped from log lines FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should ensure keys are not omitted from activation record FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should write the action-path and the limits to the annotations FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should create, and invoke an action that utilizes an invalid docker container with appropriate error FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should invoke an action using npm openwhisk FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should invoke an action receiving context properties FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should invoke an action successfully with options --blocking and --result FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should invoke an action that returns a result by the deadline FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should invoke an action twice, where the first times out but the second does not and should succeed FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should invoke action while not encoding &, <, > characters FAILED
whisk.core.cli.test.WskBasicUsageTests > Wsk actions should limit length of HTTP request and response bodies for --verbose FAILED
```

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -115,26 +115,31 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
 
     // Container is free to take more work
     case NeedWork(data: WarmedData) =>
-      freePool.update(sender(), data)
-      busyPool.remove(sender()).foreach(_ => feed ! MessageFeed.Processed)
+      freePool = freePool + (sender() -> data)
+      busyPool = busyPool - sender()
+      busyPool.foreach { _ =>
```
markusthoemmes: The logic here is wrong. Above code said: "remove this element, and if it was actually there, also send a message". So it's going to be more like:

```scala
if(busyPool.contains(sender()) {
  busyPool = busyPool - sender()
  feed ! MessageFeed.Processed
}
```

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -115,26 +115,31 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
...
 
     // Container got removed
     case ContainerRemoved =>
-      freePool.remove(sender())
-      busyPool.remove(sender()).foreach(_ => feed ! MessageFeed.Processed)
+      freePool = freePool - sender()
+      busyPool = busyPool - sender()
+      busyPool.foreach { _ =>
+        feed ! MessageFeed.Processed
+      }
```
markusthoemmes: Same as above.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
bwmcadams: oof. that was a stupid bug, but explains the behavior I was seeing in my failed tests.

bwmcadams:
Looks like the bugs are resolved and build passes, if anyone is up for revisiting this PR.

rabbah:
PG2/2258 🚀

PR #2808, opened 2017-09-27
Create the required CouchDB system databases

bbrowning:
As of CouchDB 2.0, system databases don't automatically get created on
startup and you have to do this manually. The process differs a bit
for single nodes vs clusters, but since everything else in the Ansible
scripts seems to assume single node this just sets things up that way.

See the CouchDB docs at
http://docs.couchdb.org/en/2.0.0/install/#single-node-setup for more
info.

This prevents errors constantly being logged by CouchDB with messages
like `Error in process <0.22878.7> on node 'couchdb@couchdb0' with
exit value: {database_does_not_exist ...` and `chttpd_auth_cache
changes listener died database_does_not_exist` which can end up
chewing through a lot of log disk space.

### DIFF
--- a/ansible/roles/couchdb/tasks/deploy.yml
+++ b/ansible/roles/couchdb/tasks/deploy.yml
@@ -60,3 +60,19 @@
     user: "{{ db_username }}"
     password: "{{ db_password }}"
     force_basic_auth: yes
+
+- name: create the couchdb system databases
+  uri:
+    url: "{{ db_protocol }}://{{ ansible_host }}:{{ db_port }}/{{ item }}"
+    method: PUT
+    body: >
+        "false"
+    body_format: json
+    status_code: 201
+    user: "{{ db_username }}"
+    password: "{{ db_password }}"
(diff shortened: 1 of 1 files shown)


bbrowning:
From what I can tell, it looks like for a single node database we'd use this change but for a clustered setup we'd use the changes in #2810. If we want to always default to clustered then we should merge 2810 and not this one. Or perhaps there needs to be some switch to choose single or clustered setup.

rabbah:
@bbrowning @style95 It's not clear to me that 2810 is ready. Does it make sense to accept this PR and sort out the clustering separately?

style95:
Actually #2810 can address the problem specified in this PR.
With #2810, no matter how many couchdb nodes(even single) are configured, those DB will be created.

In terms of feature, #2810 is almost ready.
I am defining test case for it.

rabbah:
@bbrowning since #2810 was merged shall we close this?

csantanapr:
#2810 addressed this

PR #2812, opened 2017-09-28
Implement bulk write of activations in the Invoker.

markusthoemmes:
Writing activations in the Invoker is not time critical and can be greatly optimized by batching the activations to reduce the number of connections used towards the database. That enables a lot of optimizations on the database side and is in general a good practice.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/Batcher.scala
+++ b/common/scala/src/main/scala/whisk/core/database/Batcher.scala
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.database
+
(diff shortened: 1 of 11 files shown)


ningyougang:
If use original deal method, with mutliple action request, the invoker's `controller<n>_logs.log` exist below error
```
[PUT] 'lambda-exp_activations' internal error, failure: 'DB request queue is full.' [marker:database_saveDocument_error:12941:22]
```

So i am very expected that this pr can be merged to improve the invoker/couchdb performance.

rabbah:
Should consider applying to sequence activations written in the controller as well.

markusthoemmes:
We did indeed. The first version of this PR included a latency penalty in the batching, which is why we started implementation in the invoker, where the penalty doesn't matter as much.

We're currently investigating an even deeper integration in the CouchDB client itself. If that doesn't work out, we'd "just" bring the `put` based implementation.

Want to wait for that and put it in this PR or are you fine with a second one improving on this?

markusthoemmes:
PG3 1156 🔵

common/scala/src/main/scala/whisk/core/entity/DocInfo.scala
```diff
@@ -147,4 +148,6 @@ protected[core] object DocInfo {
    */
   @throws[IllegalArgumentException]
   protected[core] def !(id: String, rev: String): DocInfo = DocInfo(DocId(id), DocRevision(rev))
```
rabbah: Still need !?

common/scala/src/main/scala/whisk/core/database/Batcher.scala
```diff
@@ -0,0 +1,84 @@
...
+
+ * @tparam T the type to be batched
+ * @tparam R return type of a single element after operation
+ */
+class Batcher[T, R](batchSize: Int, concurrency: Int)(operation: Seq[T] => Future[Seq[R]])(
+  implicit materializer: ActorMaterializer,
+  ec: ExecutionContext) {
+
+  private val stream = Source
+    .actorRef[(T, Promise[R])](Int.MaxValue, OverflowStrategy.dropNew)
```
rabbah: we should perhaps open an issue to refactor the relevant code so that we can backpressure the invoker feed when the activations can't be drained fast enough.

PR #2814, opened 2017-09-28
Add action kind to activation record.

rabbah:
This matches the activation annotations for sequences which have a kind sequence appended.

@markusthoemmes is there a good place to unit test this or shall I add an integration test instead?


ddragosd:
good idea @rabbah . LGTM

rabbah:
@csantanapr added a test.

PR #2815, opened 2017-09-28
WIP: adding api host param

narinearraul:
relates to https://github.com/apache/incubator-openwhisk-catalog/issues/224

### DIFF
--- a/ansible/tasks/installOpenwhiskCatalog.yml
+++ b/ansible/tasks/installOpenwhiskCatalog.yml
@@ -29,6 +29,6 @@
     version: "{{ version }}"
 
 - name: install the catalog from the catalog location
-  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ cli.path }} chdir="{{ catalog_location }}/packages"
+  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ cli.path }} {{ whisk_api_localhost_name | default(whisk_api_host_name) | default(whisk_api_localhost_name_default) }} chdir="{{ catalog_location }}/packages"
   environment:
     OPENWHISK_HOME: "{{ openwhisk_home }}"


ddragosd:
I'm going to take a look at this as I need the catalog for the work on https://github.com/apache/incubator-openwhisk/pull/1928

ddragosd:
@narinearraul how can we test that this works ? Is it a specific action in the catalog that needs this variable ?

ansible/tasks/installOpenwhiskCatalog.yml
```diff
@@ -5,6 +5,7 @@
     catalog_location={{ item.value.location }}
     catalog_repo_url={{ item.value.url }}
     api_host={{ groups['edge'] | first }}
+    whisk_api_host={{ whisk_api_host_name | groups['edge'] | first }}
```
ddragosd: IIRC actions should receive this info via env variable `__OW_API_HOST`. If there's an action inside the catalog that doesn't make use of this we could open an issue there ?

ddragosd: looking at the changes in https://github.com/apache/incubator-openwhisk-catalog/pull/243 I'm reading that this is needed only for installing the catalog ? I'm confused b/c the catalog installs correctly on the local env.

narinearraul: @ddragosd I am working on changing this such that it can use the already existing variable instead of creating a new one. Will update accordingly.

ansible/tasks/installOpenwhiskCatalog.yml
```diff
@@ -29,6 +30,6 @@
     version: "{{ version }}"
 
 - name: install the catalog from the catalog location
-  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ cli.path }} chdir="{{ catalog_location }}/packages"
+  shell: ./installCatalog.sh {{ catalog_auth_key }} {{ api_host }} {{ catalog_namespace }} {{ cli.path }} {{ whisk_api_host }} chdir="{{ catalog_location }}/packages"
```
ddragosd: @narinearraul the CLI already knows about the `APIHOST`. At a quick glance on `installCatalog.sh` I couldn't tell easily which script would require direct HTTP API Access vs `wsk` CLI. Do you happen to know this ?

ddragosd:
@narinearraul is this PR still WIP, or the other actions in the catalog are now using `__OW_API_HOST` env var ?

rabbah:
I'm going to close this - actions should use __OW_API_HOST.

PR #2818, opened 2017-09-29
Change logging backend from log4j to Logback

mhenke1:
This PR changes the logging back-end from log4j to [Logback](https://logback.qos.ch/) as part of the  'improve logging performance' work stream discussed [here](https://lists.apache.org/thread.html/6054d657d240c6c833135ebb88081ab12e9520eb0970ab5b35ea7289@%3Cdev.openwhisk.apache.org%3E)

Rational:
1) Logback showed improved performance in comparison to log4js in our test environments
2) Logback is the [recommended](https://doc.akka.io/docs/akka/2.5.4/scala/logging.html) back-end for Akka Logging
```In your runtime, you also need a SLF4J backend. We recommend Logback```

Remark:
Apache httpclient logs over commons-logging for it's log output.
The log output is also redirected to slf4j using the 'jcl-over-slf4j' library

### CODE REVIEW

File: common/scala/build.gradle [hunk 1]
```diff
@@ -18,19 +18,26 @@ dependencies {
...
     compile 'com.github.ben-manes.caffeine:caffeine:2.4.0'
     compile 'com.google.code.findbugs:jsr305:3.0.2'
 }
 
+configurations {
+    compile.exclude group: 'log4j'
+    compile.exclude group: 'commons-logging'
+    runtime.exclude group: 'log4j'
+    runtime.exclude group: 'commons-logging'
+}
```
markusthoemmes: Does it make sense to put this block on the offending libraries? e.g:

```groovy
compile('com.fasterxml.uuid:java-uuid-generator:3.1.3') {
    exclude group: 'log4j'
}
```

That'd make sure we definitly know what we exclude in the future and nothing breaks by accident?

common/scala/src/main/resources/logging.conf
```diff
@@ -1,4 +1,5 @@
 akka {
   loglevel = "INFO"
```
markusthoemmes: You can now safely set this to `DEBUG`, so the loglevel is completely controlled by the backend and not akka itself.

mhenke1: If I understand Akka logging right, setting the level to DEBUG will cause that all messages will be placed into the event stream and only afterward filtered out by Logback. Therefore I would prefer to adapt the log level after we have reduced the number of messages with subsequent changes.

common/scala/build.gradle
mhenke1: Definitely the better way. Done by amending the first commit

common/scala/src/main/resources/logging.conf
markusthoemmes: I tried it out. If you use the `LoggingFilter` while still setting "DEBUG" in akka, `isEnabled(DebugLevel)` on the logger returns `false` (it seems to take the value from the configured backend then).

And per documentation:

> You need to enable the Slf4jLogger in the loggers element in the configuration. Here you can also define the log level of the event bus. More fine grained log levels can be defined in the configuration of the SLF4J backend (e.g. logback.xml). You should also define akka.event.slf4j.Slf4jLoggingFilter in the logging-filter configuration property. It will filter the log events using the backend configuration (e.g. logback.xml) before they are published to the event bus.

mhenke1: Changed as amend to the existing commit

vvraskin:
Hi Martin, 

should we also pack the async appender with `discardingThreshold = 0` as part this PR?

Vadim.

eshepelyuk:
As a suggestion, don't you consider using `logback.groovy` as logback configuration file.
Our current production experience shows that using `groovy` for configuring logback gives a lot of advantages and flexibility.

markusthoemmes:
@eshepelyuk Any specific advantages? I mainly read it as a nicer DSL but necessarily superior in features.

vvraskin:
We haven't tested it without the async appender, it should conceal the issues if you increase the buffer high enough. 
I'd prefer to have a lever to be able to mitigate them.

eshepelyuk:
Well, you can literally write a code there. So you can configure whatever you need.
And the DSL - yes is more concise.
I'd compare this to `Maven`'s `pom.xml` vs `Gradle`'s `build.grade`

PR #2820, opened 2017-09-29
Fix links for couchdb and cloudant

suryajak:
Fixed links for couchdb and cloudant


rabbah:
Thanks for fixing these links! Do you have an Apache ICLA already?

suryajak:
Nope.

Issue #2791, opened 2017-09-21
Split runtimes into individual repos


csantanapr:
Take the builds our from the core repo into individual repos.

High level goals for short term
- Core repo will continue to have gradle project with a the same task `distDocker`
  - This is done to allow for transparent transition, alllow for current flows for dev, test, prod to continue to work with no changes.
  - Invoker and Ansible assumes that all runtimes use the same prefix and come from same 
- Each runtime will only have a `Dockerfile` that extends the runtime from a DockerHub image
  - The Dockerfile will have a pinned image 

Repos to be created by runtime type/language
  - incubator-openwhisk-runtime-nodejs
  - incubator-openwhisk-runtime-python
  - incubator-openwhisk-runtime-swift
  - incubator-openwhisk-runtime-java
  - incubator-openwhisk-runtime-php
  - incubator-openwhisk-runtime-docker

Each repo:
 - Will have same process and structure as in the core repo
    - Will use gradle to build and push images and running tests
 - Have tests and Travis enable
    - The actionContainer tests will be kept in both core an runtime for a while
 - Build and Push docker images
 - Will not have a hard dependency against core/main repository
    - Files will be duplicated (i.e. gradle, scala tests, etc..)

The `incubator-openwhisk-runtime-docker` will produce an image that swift and python will extend to get the common python proxy code.

markusthoemmes:
One (important) word of caution: Let's not build a dependency of those repositories against the main repository. They should be completely stand-alone. If we need to copy a file or two for that, I'd rather do that than building a dependency.

csantanapr:
Yes that's a goal, I will added explicitly.

ddragosd:
Also maybe we should zero in a bit into the individual repo structure that
highlights the directory structure for multiple runtime versions of that
language. I.e multiple NodeJS versions.

csantanapr:
@ddragosd that's also a goal.

As the first pass to be able to migrate the git history I would start with initial repo having same directory structure as today.

First pass for example for nodejs
```
core/nodejsActionBase/
core/nodejs6ActionBase/
core/nodejs8ActionBase/
gradle
tests/
gradlew
build.gradle
settings.gradle
```

Second pass
```
6/
8/
gradle
tests/
gradlew
build.gradle
settings.gradle
```

csantanapr:
OK after talking to @markusthoemmes I want to clarify what it means for independent from main/core repo

All the scala code for testing will not have any dependencies from core repo
[...]

PR #2821, opened 2017-09-29
Move java runtime into it's own repo

csantanapr:
Related to #2791 

Repo located here now: https://github.com/apache/incubator-openwhisk-runtime-java

PG1 2109 👍


csantanapr:
@ddragosd 
I tried my best to leave everything identical and keep the git history.

For java9 we could use `core/java9Action/Dockerfile`

My second option  would be `9/Dockerfile`

And then on the next opportunity to when we update java8 then make it consistent `8/Dockerfile`

rabbah:
Java repo should not use “bx” either.

core/javaAction/Dockerfile
```diff
@@ -1,38 +1 @@
...
-ADD proxy /javaAction
-
-RUN rm -rf /javaAction/.classpath /javaAction/.gitignore /javaAction/.gradle /javaAction/.project /javaAction/.settings /javaAction/Dockerfile /javaAction/build
-
-RUN cd /javaAction; ./gradlew oneJar
-
-RUN rm -rf /javaAction/src
-
-CMD ["java", "-jar", "/javaAction/build/libs/javaAction-all.jar"]
+FROM openwhisk/java8action:1.0.0
```
ddragosd: Do we still need to maintain this container ? IIUC it's probably being used for tests only ?

ddragosd:
> leave everything identical and keep the git history

👍 @csantanapr

core/javaAction/Dockerfile
csantanapr: it's a transition, until we get invoker and ansible bits more dynamic

csantanapr:
@rabbah 
>Java repo should not use “bx” either.

My bad some copy/paste from another repo :-)

I will fix in all of them.

Also noticed that new repos don't have issues enabled, I will open an INFRA ticket

core/javaAction/Dockerfile
ddragosd: 👍

rabbah:
@csantanapr i think it would make sense after all these prs are merged to have a runtimes top level directory (peer with core) and put all the dockerfiles there. thoughts?

Issue #2652, opened 2017-08-19
Indicate that process was killed due to OutOfMemory issue


tleyden:
Currently if OpenWhisk kills a process due to exceeding memory limits, it doesn't appear to indicate that anywhere.  In my case I was seeing a result of 

```
"response": {
        "status": "action developer error",
        "statusCode": 0,
        "success": false,
        "result": {
            "error": "The action did not produce a valid response and exited unexpectedly."
        }
    },
```

and it took quite a bit of digging to figure out that's why it was killed.  Would have been a big time-saver if there was another field in the result that somehow indicated that it was killed for this reason.

## Environment details:

* Bluemix

## Steps to reproduce the issue:

1.   Create actions + activations that have large amounts of output
2.   Modify [this monitoring code](https://github.com/tleyden/keynuker/blob/70b4d72476c205214caad66ad17c89cb323c622d/openwhisk_status.go#L84-L93) to set the `Limit` to 0 (unlimited).  This will cause it to pull hundreds of activations which include the full doc result of the activation, which may have hundreds of KB of output.  It all adds up to more than 256MB (or even 512MB) of RAM
3.  Install it to run on Bluemix using this [modified openwhisk-dockerskeleton](https://github.com/tleyden/openwhisk-dockerskeleton), which essentially runs a [python REST listener](https://github.com/tleyden/openwhisk-dockerskeleton/blob/master/actionproxy.py) that invokes a compiled Go executable and passes the JSON and receives the result over stdio.

## Provide the expected results and outputs:

```
"response": {
        "status": "action developer error",
        "statusCode": 0,
        "success": false,
        "result": {
            "error": "The action did not produce a valid response and exited unexpectedly.",
            "detail": "Killed due to hitting 256MB RAM limit"   <-- this is the enhancement request!
        }
    },
```

## Provide the actual results and outputs:

```
"response": {
        "status": "action developer error",
        "statusCode": 0,
        "success": false,
        "result": {
            "error": "The action did not produce a valid response and exited unexpectedly."
        }
    },
```

## Additional information you deem important:

* 100% reproducible each time
* I can submit a minimal reproducible test case if that will help
* I might be able to submit a PR if this is something folks want, and I was given a few hints about where in the codebase that this should be added

markusthoemmes:
Submitting a PR would be awesomely awesome! Could you elaborate on the hints you got? 😁  Maybe I can provide a few more.

rabbah:
[...]

PR #2827, opened 2017-10-01
Notify the user if a container is killed due to memory exhaustion.

markusthoemmes:
If a user-container runs out of memory, the HTTP connections is abruptly aborted and the user has no chance to get further evidence into why her action failed.

Docker actually provides that information so it can be checked on an abrupt connection termination whether the container was indeed aborted by the OOM killer.

Fixes: #2652

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
@@ -131,9 +131,9 @@ class DockerContainer(protected val id: ContainerId, protected val addr: Contain
   /** The last read-position in the log file */
   private var logFileOffset = 0L
 
-  protected val waitForLogs = 2.seconds
-  protected val waitForOomState = 2.seconds
-  protected val filePollInterval = 100.milliseconds
+  protected val waitForLogs: FiniteDuration = 2.seconds
+  protected val waitForOomState: FiniteDuration = 2.seconds
+  protected val filePollInterval: FiniteDuration = 100.milliseconds
 
   def suspend()(implicit transid: TransactionId): Future[Unit] = runc.pause(id)
   def resume()(implicit transid: TransactionId): Future[Unit] = runc.resume(id)
@@ -142,12 +142,22 @@ class DockerContainer(protected val id: ContainerId, protected val addr: Contain
     docker.rm(id)
   }
 
+  /**
+   * Was the container killed due to memory exhaustion?
+   *
+   * Retries because as all docker state-relevant operations, they won't
+   * be reflected by the respective commands immediatly but will take
+   * some time to be propagated.
+   *
+   * @param retries number of retries to make
+   * @return a Future indicating a memory exhaustion situation
+   */
   private def isOomKilled(retries: Int = (waitForOomState / filePollInterval).toInt)(
     implicit transid: TransactionId): Future[Boolean] = {
     docker.isOomKilled(id)(TransactionId.invoker).flatMap { killed =>
-      if (killed) Future.successful(killed)
+      if (killed) Future.successful(true)
       else if (retries > 0) akka.pattern.after(filePollInterval, as.scheduler)(isOomKilled(retries - 1))
-      else Future.successful(killed)
+      else Future.successful(false)
     }
   }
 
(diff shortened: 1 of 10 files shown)


markusthoemmes:
PG4 838 🔵

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -142,6 +142,44 @@ class DockerContainer(protected val id: ContainerId, protected val addr: Contain
     docker.rm(id)
   }
 
+  private def isOomKilled(retries: Int = (waitForOomState / filePollInterval).toInt)(
+    implicit transid: TransactionId): Future[Boolean] = {
+    docker.isOomKilled(id)(TransactionId.invoker).flatMap { killed =>
+      if (killed) Future.successful(killed)
+      else if (retries > 0) akka.pattern.after(filePollInterval, as.scheduler)(isOomKilled(retries - 1))
+      else Future.successful(killed)
```
rabbah: For clarity I would use true and false explicitly to complete the future.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -142,6 +142,44 @@ class DockerContainer(protected val id: ContainerId, protected val addr: Contain
...
+      http.post(path, body, retry)
+    }.flatMap { response =>
+      val finished = Instant.now()
+
+      response.left
+        .map {
+          case t: ConnectionError =>
+            isOomKilled().map {
+              case true =>
+                logging.info(this, "connection abruptly terminated, checking for memory exhaustion")
```
rabbah: Hmm. Log message say checking for oom but here already know it is.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -142,6 +142,44 @@ class DockerContainer(protected val id: ContainerId, protected val addr: Contain
     docker.rm(id)
   }
 
+  private def isOomKilled(retries: Int = (waitForOomState / filePollInterval).toInt)(
+    implicit transid: TransactionId): Future[Boolean] = {
+    docker.isOomKilled(id)(TransactionId.invoker).flatMap { killed =>
+      if (killed) Future.successful(killed)
```
rabbah: A comment here is warranted to explain the need for retry.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -142,6 +142,44 @@ class DockerContainer(protected val id: ContainerId, protected val addr: Contain
...
+          case t: ConnectionError =>
+            isOomKilled().map {
+              case true =>
+                logging.info(this, "connection abruptly terminated, checking for memory exhaustion")
+                MemoryExhausted()
+              case false => t
+            }
+          case t => Future.successful(t)
+        }
+        .fold(_.map(Left(_)), right => Future.successful(Right(right)))
```
rabbah: Feels like this fold is avoidable.

markusthoemmes: Any suggestions? It is to transform `Either[Future[ContainerConnectionError], ContainerResponse]` to `Future[Either[ContainerConnectionError, ContainerResponse]]`

I implemented it using just one `fold` (vs. using `left.map` and a `fold`). It's possible, but I found it more clear the way it is now.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
markusthoemmes: True, I removed the message as our existing logging catches the situation anyway:

```
[2017-10-04T06:13:16.909Z] [INFO] [#tid_63] [DockerContainer] running result: MemoryExhausted() [marker:invoker_activationRun_finish:780:110]
```

tleyden:
Exciting to see this go in!  What's the approximate round-trip time before getting deployed to Bluemix?

rabbah:
@tleyden ibm cloud functions picks up master frequently... worst case, a week.

PR #2799, opened 2017-09-22
ansible fix for #2788 for docker for mac 

ddragosd:
This relates to https://github.com/apache/incubator-openwhisk/pull/2793 by mounting the `docker` and `runc` binaries from the host.

It doesn't yet update the dockerfile with the invoker as that's handled in #2793 , but this shows what's needed to fix #2788 . We can discuss how we want to merge all changes eventually.

PR #2828, opened 2017-10-02
option to use docker pause/unpause instead of docker-runc pause/unpause

dgrove-oss:
Restore config option to use docker pause/unpause instead of
docker-runc pause/unpause in the invoker.  Works around problem
of incompatible versions of docker-runc in invoker container and
hosting environment.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
@@ -49,6 +49,7 @@ object DockerContainer {
    * @param network network to launch the container in
    * @param dnsServers list of dns servers to use in the container
    * @param name optional name for the container
+   * @param useRunc use docker-runc to pause/unpause container?
    * @return a Future which either completes with a DockerContainer or one of two specific failures
    */
   def create(transid: TransactionId,
@@ -60,6 +61,7 @@ object DockerContainer {
              network: String = "bridge",
              dnsServers: Seq[String] = Seq(),
              name: Option[String] = None,
+             useRunc: Boolean = true,
              dockerRunParameters: Map[String, Set[String]])(implicit docker: DockerApiWithFileAccess,
                                                             runc: RuncApi,
                                                             as: ActorSystem,
@@ -105,7 +107,7 @@ object DockerContainer {
           docker.rm(id)
           Future.failed(WhiskContainerStartupError(s"Failed to obtain IP address of container '${id.asString}'."))
       }
-    } yield new DockerContainer(id, ip)
+    } yield new DockerContainer(id, ip, useRunc)
   }
 }
 
@@ -119,12 +121,13 @@ object DockerContainer {
  * @param id the id of the container
  * @param addr the ip of the container
  */
-class DockerContainer(protected val id: ContainerId, protected val addr: ContainerAddress)(
-  implicit docker: DockerApiWithFileAccess,
-  runc: RuncApi,
-  as: ActorSystem,
-  protected val ec: ExecutionContext,
-  protected val logging: Logging)
+class DockerContainer(protected val id: ContainerId,
+                      protected val addr: ContainerAddress,
+                      protected val useRunc: Boolean)(implicit docker: DockerApiWithFileAccess,
+                                                      runc: RuncApi,
+                                                      as: ActorSystem,
+                                                      protected val ec: ExecutionContext,
+                                                      protected val logging: Logging)
     extends Container
     with DockerActionLogDriver {
 
@@ -135,8 +138,10 @@ class DockerContainer(protected val id: ContainerId, protected val addr: Contain
   protected val waitForOomState: FiniteDuration = 2.seconds
   protected val filePollInterval: FiniteDuration = 100.milliseconds
 
-  def suspend()(implicit transid: TransactionId): Future[Unit] = runc.pause(id)
-  def resume()(implicit transid: TransactionId): Future[Unit] = runc.resume(id)
+  def suspend()(implicit transid: TransactionId): Future[Unit] =
+    if (useRunc) { runc.pause(id) } else { docker.pause(id) }
+  def resume()(implicit transid: TransactionId): Future[Unit] =
+    if (useRunc) { runc.resume(id) } else { docker.unpause(id) }
   override def destroy()(implicit transid: TransactionId): Future[Unit] = {
     super.destroy()
     docker.rm(id)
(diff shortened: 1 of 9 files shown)


dgrove-oss:
A workaround to sidestep the problems of https://github.com/apache/incubator-openwhisk/issues/2788

ddragosd:
The problem is that when an operator configures to use `runc`, it may not work correctly. This is what I tried to fix in  https://github.com/apache/incubator-openwhisk/pull/2799.

csantanapr:
From what I recalled the point of using docker-runc/runc is performance, runc was the winner when we compared containerd vs. runc because pause/unpause with docker cli gave us terrible performance too many abstraction and management layers to do what we want when waking up a warm container

dgrove-oss:
I may have not done a good job of explaining the motivation for this PR.  

Getting a working docker-runc is touchy because you need an exact version match between the invoker container and the host.  @ddragosd is trying to fix that by mounting the binaries from the host.  I also experimented with that in the kube context, but just mounting the binaries isn't enough (at least on kube as made available on Bluemix). docker-runc from the kube worker node also has dependencies on libAppArmor, so there is more junk to drag in.  I haven't gotten that to work (and I'm not 100% sure it can be made to work reliably), so I thought bringing back the option to bypass runc to a lower performance option might be better to get some container reuse might be better than getting no reuse at all.   But maybe I'm wrong (is no reuse better than docker pause/unpase?)

markusthoemmes:
Should we fold this into the `DockerClient` to avoid needing to pass around the flag? The `DockerClient` could transparently use docker/runc then. That'd somewhat break abstraction but would be neater in terms of variable scoping.

ddragosd:
> Should we fold this into the DockerClient to avoid needing to pass around the flag

In this case the Invoker could optimistically try to use `runc` and, in case of error, fallback on `docker`, and save this information next time it needs to `pause`/`resume` a container. `pause`/`resume` would only work when the binary inside the Invoker container is compatible with `runc` on the host, otherwise we're getting: ```stderr: json: cannot unmarshal object into Go value of type []string``` . #2799 solves this error by mounting `runc` from the host, but it only works if `runc` doesn't have other dependent libraries.

@dgrove-oss in the current kube deploy that you're testing with, I assume the binary on the host is compatible with the binary in the invoker and you didn't get the `unmarshall` error ?

dgrove-oss:
In the kube deploy I'm testing, the host's docker-runc has dependencies on additional host libraries (libAppArmor.so), so doesn't run correctly from within the invoker's container.  The kube deployment already was also mounting libAppArmor.so.1, but that isn't working (yet).  Perhaps the problem is as simple as also needing to fool around with /etc/ld.so.conf.d in the invoker container, but I haven't gotten it to work yet.  There might be additional dependencies that would need to be dragged in.

How would we make the logic for error on docker-runc ==> fallback to docker logic robust?  One could imagine a spurious error "poisoning" the container causing us to always use docker even though docker-runc would work most of the time?

dgrove-oss:
And the host docker-runc is newer (incompatible with) the docker-runc in the invoker image. So I am seeing the deserialization errors if the invoker attempts to use the docker-runc that is built into its image.

markusthoemmes:
I suppose having an implicit switch is safer as noted by @dgrove-oss. Wouldn't want to rely on possibly flawed auto-detection in production.

dgrove-oss:
squashed to single commit.  PG3 1151 🏃

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
```diff
@@ -41,6 +41,7 @@ class DockerContainerFactory(config: WhiskConfig, instance: InstanceId, paramete
   /** Initialize container clients */
   implicit val docker = new DockerClientWithFileAccess()(ec)
   implicit val runc = new RuncClient(ec)
+  implicit val useRunc = config.invokerUseRunc
```
markusthoemmes: Please remove this implicit as well.

dgrove-oss: done.  sorry; had removed that at some point yesterday but lost the change.

dgrove-oss:
docker pull of redis image from dockerhub failed in travis build.  Can someone kick it again in a few minutes?

dgrove-oss:
PG3 1151 🔵

ddragosd:
Before we merge this and close the other PRs WDYT if we document the expected behavior introduced with this flag, and what to do when it doesn't work ? Should we document that an operator could mount `runc` in case there's no version parity, or document that the invoker requires version parity ?

dgrove-oss:
Documenting makes sense.  Where would we typically document such a thing?  Is adding a section about docker-runc to docs/deploy.md the right place?

markusthoemmes:
We also need to be able to actually use the flag from ansible.

dgrove-oss:
I implemented the flag to default to true (use docker-runc) if unspecified. This was mainly to avoid disturbing existing deployments.  Arguably it should default to false (use docker) since that is the conservative/safe option.   Should I flip it, or leave it as is?

markusthoemmes:
No the default is fine. RunC is stable and running in many environments. We need an ansible controllable flag though, which is threaded through to the controller which we can then alter per environment.

dgrove-oss:
PG3 1153

dgrove-oss:
PG3 1153 🔵

dgrove-oss:
rebased to master to resolve conflict.  
@markusthoemmes I added the ansible controllable flag yesterday that threads through to the invoker.  I think this is what we need (although I'm new to ansible...please check).

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -60,6 +61,7 @@ object DockerContainer {
              network: String = "bridge",
              dnsServers: Seq[String] = Seq(),
              name: Option[String] = None,
+             useRunc: Boolean = true,
```
ddragosd: WDYT about making `config` an implicit param to avoid method signature changes downstream ?

dgrove-oss: i originally made it an implicit and @rabbah requested that it not be an implicit, so I reworked.

rabbah: an implicit boolean is suspect. I think @ddragosd was perhaps thinking about config as in whisk config? I prefer the explicit parameter here. Is there an implication to the other container factories of concerns?

ddragosd: no implications to other container factories I believe. I just thought we can keep the PR smaller, but if it makes more sense to be explicit with this flag, 👍

dgrove-oss:
fixed scoping problem noticed by @markusthoemmes in code review.  Added test case.

dgrove-oss:
PG2 2167 🏃

dgrove-oss:
squashed down to single commit.  Please merge once Travis verifies the squashed commit.

PR #2831, opened 2017-10-04
Make console tests more robust

sven-lange-last:
Use activation start times as time base when polling for activations since a point in time - before, local system time was used. If there is a clock mismatch between the local system and the system which records activations (=invokers), poll results can be empty.

Other queries use relative times, i.e. return activations since the specified number of seconds. For these relative times, a start and end time is taken based on the local system time and the time difference in seconds is computed. It is important to use start and end time from the same time source.

In addition, use some time contigency to make tests more robust.

Provide more debug output in case an assertion fails.

### DIFF
--- a/tests/src/test/scala/system/basic/WskConsoleTests.scala
+++ b/tests/src/test/scala/system/basic/WskConsoleTests.scala
@@ -66,7 +66,8 @@ class WskConsoleTests extends TestHelpers with WskTestHelpers {
     }
 
     // Some contingency to make query more robust
-    val start = Instant.now.minusSeconds(1)
+    // Account for time differences between controller and invoker
+    val start = Instant.now.minusSeconds(5)
     val payload = new String("from the console!".getBytes, "UTF-8")
     val run = wsk.action.invoke(fullActionName, Map("payload" -> payload.toJson))
     withActivation(wsk.activation, run, totalWait = 30 seconds) { activation =>
@@ -89,7 +90,8 @@ class WskConsoleTests extends TestHelpers with WskTestHelpers {
 
     val count = 3
     // Some contingency to make query more robust
-    val start = Instant.now.minusSeconds(1)
+    // Account for time differences between controller and invoker
+    val start = Instant.now.minusSeconds(5)
     val run = wsk.action.invoke(name, Map("n" -> count.toJson))
     withActivation(wsk.activation, run) { activation =>
       // Time recorded by invoker, some contingency to make query more robust
--- a/tests/src/test/scala/system/basic/WskConsoleTests.scala
+++ b/tests/src/test/scala/system/basic/WskConsoleTests.scala
@@ -17,7 +17,6 @@
 
 package system.basic;
 
-import java.time.Clock
 import java.time.Instant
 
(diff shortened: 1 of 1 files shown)


sven-lange-last:
@rabbah - please attend to this PR.

PG 3 / 1160 still running.

tests/src/test/scala/system/basic/WskConsoleTests.scala
```diff
@@ -61,31 +65,46 @@ class WskConsoleTests extends TestHelpers with WskTestHelpers {
...
-    val duration = Some(30 seconds)
+    val duration = 30 seconds
     val payload = new String("from the console!".getBytes, "UTF-8")
     val run = wsk.action.invoke(fullActionName, Map("payload" -> payload.toJson))
-    withActivation(wsk.activation, run, totalWait = duration.get) { activation =>
-      val console = wsk.activation.console(10 seconds, since = duration)
-      console.stdout should include(payload)
+    withActivation(wsk.activation, run, totalWait = duration) { activation =>
+      val pollTime = 10 seconds
+      val console = wsk.activation.console(pollTime, since = Some(duration))
```
rabbah: should the duration here used in `since` actually account for the activation start time?

sven-lange-last: The `since` parameter passed to `wsk.activation.console`= `poll` is translated to `--since-seconds` and thus, is a relative parameter. So no absolute point in time.

The activation start and end times base on the invoker's system clock. The clock of the system running the test may differ from the invoker's system clock. For this reason, queries that use absolute point in times should always be run with times that base on activation start and end times to make them robust.

For relative times, we can measure how long an operation takes and use the measured time plus some contingency.

At the moment, we use a fixed time of 30 seconds instead of measuring. If the time from action invocation start until query execution takes more than 30 seconds, the test will fail. You are right that using a measured time would make the test more robust - thanks for pointing out.

At the same time, I would prefer to measure local system time for computing a `--since-seconds` time because this time difference will also contain network latencies, nginx, database, and controller times.

I will update the PR.

rabbah: Thanks for making the change. Looks better. There’s still an assumption here that the controller’s time is ahead at worst (not behind); it’s the controller that sets the since marker in the query. So there could be three skews to consider: invoker, controller and test machine?

sven-lange-last:
PG 3 / 1163 completed successfully.

tests/src/test/scala/system/basic/WskConsoleTests.scala
sven-lange-last: Good point to also consider clock skew between invoker and controller. In our test environments I've never seen a time difference larger than 3 seconds between invokers and controllers. BTW: this time difference even shows up for NTP synchronized VMs...

So I think with adding 5 seconds of contingency, we should be on the safe side. Extending the search time window should not match undesired activations because this PR introduces unique package / action names.

I will provide a commit with more contingency.

sven-lange-last:
PG 3 / 166 running.

PR #2832, opened 2017-10-04
Do not cache invalidate when adding a DB attachment

dubeejw:
Currently when an entity that has an attachment is created or updated, the corresponding entity is not cached. This change caches the created entity.

For instance, creating or updating a java action will create a database entry of the entity and the entity will be cached. Immediately after the jar file is added as an attachment in the DB and the previously cached entity is removed from the cache. That means quickly making a call after action creation, such as action get would have to fetch the entire entity from the DB instead of the local cache.

Instead these  changes will not remove the cached entity when a DB attachment is added to the entity.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
@@ -133,25 +133,25 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
    * and share common super types EntityRecord and WhiskEntity.
    *
    * @param db the datastore client to fetch entity from
-   * @param doc the entity to store
+   * @param entity the entity to store
    * @param transid the transaction id for logging
    * @param notifier an optional callback when cache changes
    * @return Future[DocInfo] with completion to DocInfo containing the save document id and revision
    */
-  def put[Wsuper >: W](db: ArtifactStore[Wsuper], doc: W)(
+  def put[Wsuper >: W](db: ArtifactStore[Wsuper], entity: W)(
     implicit transid: TransactionId,
     notifier: Option[CacheChangeNotification]): Future[DocInfo] = {
     Try {
       require(db != null, "db undefined")
-      require(doc != null, "doc undefined")
+      require(entity != null, "entity undefined")
     } map { _ =>
       implicit val logger = db.logging
       implicit val ec = db.executionContext
 
-      val key = CacheKey(doc)
+      val key = CacheKey(entity)
 
-      cacheUpdate(doc, key, db.put(doc) map { docinfo =>
-        doc match {
+      cacheUpdate(entity, key, db.put(entity) map { docinfo =>
+        entity match {
           // if doc has a revision id, update it with new version
           case w: DocumentRevisionProvider => w.revision[W](docinfo.rev)
         }
@@ -166,6 +166,7 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
 
   def attach[Wsuper >: W](
     db: ArtifactStore[Wsuper],
+    entity: W,
     doc: DocInfo,
     attachmentName: String,
     contentType: ContentType,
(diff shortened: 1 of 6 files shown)


common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
```diff
@@ -166,24 +166,29 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
...
+    entity: W,
     doc: DocInfo,
     attachmentName: String,
     contentType: ContentType,
     bytes: InputStream)(implicit transid: TransactionId, notifier: Option[CacheChangeNotification]): Future[DocInfo] = {
 
     Try {
       require(db != null, "db undefined")
+      require(entity != null, "entity undefined")
       require(doc != null, "doc undefined")
```
markusthoemmes: Should we get rid of all these null checks? They are for sanity, great, but are they really needed @rabbah ? I'm pretty sure we're `null` free in this area.

dubeejw:
PG4 854 ⏳

dubeejw:
PG2 2173 🔵

common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
```diff
@@ -166,24 +166,29 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
 
   def attach[Wsuper >: W](
     db: ArtifactStore[Wsuper],
-    doc: DocInfo,
+    doc: W,
+    docInfo: DocInfo,
```
rabbah: i think we can omit `docInfo` see below.

common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
```diff
@@ -166,24 +166,29 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
...
       implicit val logger = db.logging
       implicit val ec = db.executionContext
 
-      val key = CacheKey(doc.id.asDocInfo)
-      // invalidate the key because attachments update the revision;
-      // do not cache the new attachment (controller does not need it)
-      cacheInvalidate(key, {
-        val src = StreamConverters.fromInputStream(() => bytes)
-        db.attach(doc, attachmentName, contentType, src)
+      val key = CacheKey(doc)
```
rabbah: the `key` for `doc: WhiskEntity` is `docInfo`, so you could create a `DocInfo` instance out of the key.
alternatively, if `W` is bounded to `WhiskDocument` then you can get its `id.asDocInfo`.

dubeejw:
@rabbah, made updates based on your comments. Can you take another look?

dubeejw:
PG3 1406 🔵

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -315,7 +315,8 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
           val manifest = exec.manifest.attached.get
 
           for (i1 <- super.put(db, newDoc);
-               i2 <- attach[A](db, i1, manifest.attachmentName, manifest.attachmentType, stream)) yield i2
+               i2 <- attach[A](db, newDoc.revision(i1.rev), manifest.attachmentName, manifest.attachmentType, stream))
```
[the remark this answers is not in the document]
dubeejw: `i1` is of type `docinfo` and we need to pass a document as the second argument for `attach`.

dubeejw:
PG3 1417 🔵

common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
rabbah: Yeah we can have a separate cleanup that removes these which date back to the gson serdes.

common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
```diff
@@ -149,15 +108,12 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
...
+      val docInfo = doc.docinfo
 
-      cacheUpdate(doc, key, db.put(doc) map { docinfo =>
-        doc match {
-          // if doc has a revision id, update it with new version
-          case w: DocumentRevisionProvider => w.revision[W](docinfo.rev)
-        }
-        docinfo
+      cacheUpdate(doc, key, db.put(doc) map { newDocInfo =>
+        doc.revision[W](newDocInfo.rev)
```
rabbah: I know I used the copy but thinking if instead we should have a .revision on docinfo, or take the new doc info from the revised doc on line 114 to avoid an unnecessary copy.

common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
rabbah: The bound is now added to the trait.

common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
```diff
@@ -178,12 +134,13 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
...
-      // invalidate the key because attachments update the revision;
-      // do not cache the new attachment (controller does not need it)
-      cacheInvalidate(key, {
-        val src = StreamConverters.fromInputStream(() => bytes)
-        db.attach(doc, attachmentName, contentType, src)
+      val key = CacheKey(doc)
+      val docInfo = doc.docinfo
+      val src = StreamConverters.fromInputStream(() => bytes)
+
+      cacheUpdate(doc, key, db.attach(docInfo, attachmentName, contentType, src) map { newDocInfo =>
```
rabbah: I neee to check the locking on the to-be-updated doc; I think a write around is allowed to race to the db and this future would only succeed if the locked entity was updated. Since the attachment also holds the reference to the revision of the locked entry, the cached document will be consistent on success and invalidated on failure.

csantanapr:
it looks like reviews are done and approved, going to merge this

PR #2835, opened 2017-10-05
Move batching to CouchDbRestStore

cbickel:
Move the mechanism of batching activations to the CouchDbRestStore, to have it also in the Controller in place.
Currently batching is only enabled for activations.

Dependant on #2812

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
+++ b/common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
@@ -20,21 +20,22 @@ package whisk.core.database
 import scala.concurrent.Await
 import scala.concurrent.Future
 import scala.concurrent.duration._
+
 import akka.actor.ActorSystem
 import akka.event.Logging.ErrorLevel
 import akka.http.scaladsl.model._
+import akka.stream.ActorMaterializer
 import akka.stream.scaladsl._
 import akka.util.ByteString
 import spray.json._
 import whisk.common.Logging
 import whisk.common.LoggingMarkers
 import whisk.common.TransactionId
+import whisk.core.entity.BulkEntityResult
 import whisk.core.entity.DocInfo
 import whisk.core.entity.DocRevision
 import whisk.core.entity.WhiskDocument
 import whisk.http.Messages
-
-import scala.util.{Failure, Success, Try}
 
 /**
  * Basic client to put and delete artifacts in a data store.
@@ -47,13 +48,17 @@ import scala.util.{Failure, Success, Try}
  * @param dbName the name of the database to operate on
  * @param serializerEvidence confirms the document abstraction is serializable to a Document with an id
  */
-class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](
-  dbProtocol: String,
-  dbHost: String,
-  dbPort: Int,
-  dbUsername: String,
-  dbPassword: String,
-  dbName: String)(implicit system: ActorSystem, val logging: Logging, jsonFormat: RootJsonFormat[DocumentAbstraction])
+class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: String,
+                                                                  dbHost: String,
+                                                                  dbPort: Int,
+                                                                  dbUsername: String,
+                                                                  dbPassword: String,
+                                                                  dbName: String,
+                                                                  useBatching: Boolean = false)(
+  implicit system: ActorSystem,
+  val logging: Logging,
+  jsonFormat: RootJsonFormat[DocumentAbstraction],
+  materializer: ActorMaterializer)
     extends ArtifactStore[DocumentAbstraction]
     with DefaultJsonProtocol {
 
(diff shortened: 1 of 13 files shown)


common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -110,6 +134,39 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](
...
+
+    val f = client.putDocs(ds).map {
+      _ match {
+        case Right(response) =>
+          transid.finished(this, start, s"'$dbName' completed $count documents")
+
+          response.convertTo[Seq[BulkEntityResult]].map { singleResult =>
+            singleResult.error
+              .map {
+                case "conflict" => Left(DocumentConflictException("conflict on 'bulk_put'"))
```
dubeejw: If there is one conflicting entity in the bulk request do all the entities fail?

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -61,6 +67,13 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](
   private val client: CouchDbRestClient =
     new CouchDbRestClient(dbProtocol, dbHost, dbPort.toInt, dbUsername, dbPassword, dbName)
 
+  // This the the amount of allowed parallel requests for each entity, before batching starts. If there are already maxOpenDbRequests
+  // and more documents need to be stored, then all arriving documents will be put into batches (if enabled) to avoid a long queue.
+  private val maxOpenDbRequests = system.settings.config.getInt("akka.http.host-connection-pool.max-open-requests") / 2
```
dubeejw: Why half of `max-open-requests`?

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -61,6 +67,13 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](
...
+
+  private val batcher: Batcher[JsObject, Either[ArtifactStoreException, DocInfo]] =
+    new Batcher(500, maxOpenDbRequests)(put(_)(TransactionId.unknown))
```
dubeejw: Seems like the batch size should be configurable instead of hardcoded.

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
markusthoemmes: Nope, it's only a specific entity that fails. Behavior then is just like with any single request today.

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
markusthoemmes: To prevent fights of other put/reads vs. activation puts.

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
cbickel: No, all other entries will have no errors. So there is no need to retry, if there is only one document with a conflict.

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
cbickel: Due to the current design, we have an datastore (and a batcher) for each entity. That means in the controller we have 5 batchers. All of them are sharing the 128 connections, that are configured in akka (max-open-requests).
If the controller has to write a lot of activations now, only 64 of these connections will be used on setting `maxOpenDbRequests` to half of `max-open-requests`.
So if someone else comes and wants to invoke an action that is not in the cache, it has not to wait until all the activations are written into the database.
In the invoker that also means, that writing activations will not affect getting actions from the database.
In addition, @markusthoemmes and me did some performance meassurements and there was no big difference between using 64 or 128 requests in parallel to write away activations.

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
cbickel: Why do you think it should be configurable? Do you have a scenario in mind, where you want to configure this size?
Btw: It is only the maximum allowed batch size.

When I was chatting with Cloudant developpers a while ago, we were talking about such bulk requests. The scenario was a bit different, but very similar (write documents as fast as possible into a database with _bulk). This person suggested a batch size of 500.

common/scala/src/main/scala/whisk/core/entity/WhiskStore.scala
```diff
@@ -130,8 +131,8 @@ object WhiskActivationStore {
       dbPort -> null,
       dbActivations -> null)
 
-  def datastore(config: WhiskConfig)(implicit system: ActorSystem, logging: Logging) =
-    SpiLoader.get[ArtifactStoreProvider].makeStore[WhiskActivation](config, _.dbActivations)
+  def datastore(config: WhiskConfig)(implicit system: ActorSystem, logging: Logging, materializer: ActorMaterializer) =
```
dubeejw: Why not batch `WhiskEntityStore` as well?

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
dubeejw: Was thinking the value might change per deployment. Should at least add a comment stating why it's set to 500 here.

markusthoemmes:
PG2 2248 is all good.

PR #2847, opened 2017-10-10
Treat action code as attachments for created/updated actions

dubee:
Save action code as attachments in the database for new or updated actions.

Depends on https://github.com/apache/incubator-openwhisk/pull/2832 to get `whisk.core.controller.test.ActionsApiTests > Actions API should put and then get action from cache` test to pass.

I combined these changes with https://github.com/apache/incubator-openwhisk/pull/2730 and performed profiling on the controller.

Here is a graph of the heap monitoring when invoking a 34MB zipped action:

screen shot 2017-10-10 at 11 59 37 am

Previously invoking a 34MB non-cached action would result in a spike in heap usage when only using changes in https://github.com/apache/incubator-openwhisk/pull/2730:

30878095-c7623960-a2c9-11e7-9cc0-f17af3120f67

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/Exec.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Exec.scala
@@ -30,6 +30,8 @@ import whisk.core.entity.size.SizeInt
 import whisk.core.entity.size.SizeOptionString
 import whisk.core.entity.size.SizeString
 
+import java.util.Base64
+
 /**
  * Exec encodes the executable details of an action. For black
  * box container, an image name is required. For Javascript and Python
@@ -140,19 +142,24 @@ protected[core] case class CodeExecMetaDataAsString(manifest: RuntimeManifest,
 
 protected[core] case class CodeExecAsAttachment(manifest: RuntimeManifest,
                                                 override val code: Attachment[String],
-                                                override val entryPoint: Option[String])
+                                                override val entryPoint: Option[String],
+                                                override val binary: Boolean = false)
     extends CodeExec[Attachment[String]] {
   override val kind = manifest.kind
   override val image = manifest.image
   override val sentinelledLogs = manifest.sentinelledLogs.getOrElse(true)
   override val deprecated = manifest.deprecated.getOrElse(false)
   override val pull = false
-  override lazy val binary = true
   override def codeAsJson = code.toJson
 
   def inline(bytes: Array[Byte]): CodeExecAsAttachment = {
-    val encoded = new String(bytes, StandardCharsets.UTF_8)
-    copy(code = Inline(encoded))
+    val code = new String(bytes, StandardCharsets.UTF_8)
+
+    if (kind == "java" && !Exec.isBinaryCode(code)) {
+      val encoded = Base64.getEncoder.encodeToString(bytes)
+      copy(code = Inline(encoded))
+    } else
+      copy(code = Inline(code))
   }
 
   def attach: CodeExecAsAttachment = {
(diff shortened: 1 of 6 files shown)


dubeejw:
@csantanapr, FYI.

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -243,7 +242,7 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
           val newDoc = doc.copy(exec = exec.attach)
           newDoc.revision(doc.rev)
 
-          val stream = new ByteArrayInputStream(Base64.getDecoder().decode(code))
+          val stream = new ByteArrayInputStream(code.getBytes)
```
dubeejw: If this change is left in place, existing deployments will need to do a database migration in order to convert old Java action byte attachments to Base64 encoded attachments.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -241,21 +241,33 @@ protected[core] object Exec extends ArgNormalizer[Exec] with DefaultJsonProtocol
...
-                obj.fields.get("code").orElse(obj.fields.get("jar"))
+              // java actions once stored the attachment in "jar" instead of "code"
+              val code = obj.fields.get("code").orElse(obj.fields.get("jar"))
+
+              val binary: Boolean = code match {
+                case Some(JsString(c)) => isBinaryCode(c)
+                case _ =>
+                  obj.fields.get("binary") match {
+                    case Some(JsBoolean(b)) => b
+                    case _                  => false
```
dubeejw: Should probably error here.

tests/src/test/scala/whisk/core/controller/test/ActionsApiTests.scala
```diff
@@ -138,7 +138,7 @@ class ActionsApiTests extends ControllerTestCommon with WhiskActionsApi {
     Get(s"$collectionPath/${action.name}") ~> Route.seal(routes(creds)) ~> check {
       status should be(OK)
       val response = responseAs[WhiskAction]
-      response should be(action)
+      response.toJson should be(action.toJson)
```
dubeejw: Direct comparison of the WhiskAction here fails. However, comparing the actions as JSON seems to work.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -241,21 +241,33 @@ protected[core] object Exec extends ArgNormalizer[Exec] with DefaultJsonProtocol
...
+              }
+
+              val attachment: Attachment[String] = {
+                code
               } map {
                 attFmt[String].read(_)
               } getOrElse {
-                throw new DeserializationException(
-                  s"'code' must be a valid base64 string in 'exec' for '$kind' actions")
+                throw new DeserializationException(s"'code' must be a string defined in 'exec' for '$kind' actions")
```
dubeejw: Not sure we will ever get here.

dubeejw:
Can use test from https://github.com/apache/incubator-openwhisk/pull/2855 after that one is merged.

dubeejw:
whisk.core.database.test.CacheConcurrencyTests > the cache should support concurrent CRUD without bogus residual cache entries, iter 1 FAILED

Failure is a result of parallel action gets with updates. I think this test fails because there is now an extra database write for the attachment that causes the action get to fail as it does not have the correct document revision.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -278,21 +278,33 @@ protected[core] object Exec extends ArgNormalizer[Exec] with DefaultJsonProtocol
 
           manifest.attached
             .map { a =>
-              val jar: Attachment[String] = {
-                // java actions once stored the attachment in "jar" instead of "code"
-                obj.fields.get("code").orElse(obj.fields.get("jar"))
+              // java actions once stored the attachment in "jar" instead of "code"
+              val code = obj.fields.get("code").orElse(obj.fields.get("jar"))
```
rabbah: perhaps as part of the schema migration pr, you can normalize all `jar` occurrences to `code` and we can remove this legacy bit.

dubeejw:
@rabbah, I believe I had originally planned to have https://github.com/apache/incubator-openwhisk/pull/2855 merged first and update the tests in that PR for this one. I can add the tests here as well.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
dubeejw: As part of https://github.com/apache/incubator-openwhisk/pull/2938 I have normalized jar to code. Will have to remove the jar reference here.

dubeejw:
Pushed changes to use ansible/files/runtimes.json since the runtimes were removed from ansible/group_varz/all.

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -122,19 +124,24 @@ protected[core] case class CodeExecMetaDataAsString(manifest: RuntimeManifest) e
...
   override val pull = false
-  override lazy val binary = true
   override def codeAsJson = code.toJson
 
   def inline(bytes: Array[Byte]): CodeExecAsAttachment = {
-    val encoded = new String(bytes, StandardCharsets.UTF_8)
-    copy(code = Inline(encoded))
+    val code = new String(bytes, StandardCharsets.UTF_8)
+
+    if (kind == "java" && !Exec.isBinaryCode(code)) {
```
dubeejw: Allows backward compatibility with old Java actions that have attachments that are not base64.

markusthoemmes:
@dubeejw What's the status here? The latest Travis failed.

dubee:
@markusthoemmes, Travis is now passing. Awaiting more feed back.

dubee:
Closed in favor of https://github.com/apache/incubator-openwhisk/pull/3286

Issue #2837, opened 2017-10-06
Support reading param.json from stdin


alexkli:
It would be useful to support reading a parameter json file from stdin for various piping scenarios, using the convention of "-" as filename:

    cat some.json | envsubst | wsk action invoke my-action -P -

In this example we can leverage `envsubst` to replace some environment variables in the file. Without being able to read from stdin, one has to push the content into a temporary file, invoke and then delete the temp file again.

From looking at the go cli code, it seems `readFile()` in [util.go](https://github.com/apache/incubator-openwhisk/blob/26146368f1dd07f817062e662db64c73a8d486d6/tools/cli/go-whisk-cli/commands/util.go#L829) could be adapted to leverage `ioutil.ReadAll(os.Stdin)` in case `filename == "-"`. One probably has to check if that is ok for all other `readFile()` usages, or if that should be a special option reserved for `-P --param-file` only.

alexkli:
Provided pull request #2848.

Please have a look, thanks!

csantanapr:
Issue moved to [apache/incubator-openwhisk-cli #126](https://github.com/apache/incubator-openwhisk-cli/issues/126) via [**ZenHub**](https://www.zenhub.com/)

PR #2848, opened 2017-10-10
Support reading param.json from stdin #2837

alexkli:
For #2837. Adds support for `-P -` that reads the param json file from the stdin, as a common practice in *nix tools.

Disclaimer: My first shot at programming in go. Thus I blindly cargo-cult copied from the existing code base :)

I deliberately added the feature _only_ for `-P / --param-file`. Initially I tried to do it generically inside `readFile()` from `util.go`, but it seems it's not a good idea to enable this for all callers of `readFile()`, which apparently [include configuration files](https://github.com/apache/incubator-openwhisk/blob/26146368f1dd07f817062e662db64c73a8d486d6/tools/cli/go-whisk-cli/commands/api.go#L844) (i.e. file name is not passed as argument on the command line).

Example usage after the change:

    cat params.json | envsubst | wsk -i action invoke myaction -P -


alexkli:
I think this needs a documentation/usage output change.

alexkli:
Usage now says for all cases of `--param-file`:

      -P, --param-file FILE   FILE containing parameter values in JSON format, use '-' to read from stdin

Also added a [full example to docs/actions.md](https://github.com/alexkli/incubator-openwhisk/blob/4c058d9ca59d76e8d5aed0c6a85f8edda553e527/docs/actions.md#passing-parameters-to-an-action) (scroll down to 4.).

docs/actions.md
```diff
@@ -202,6 +202,39 @@ Parameters can be passed to the action when it is invoked.
       "payload": "Hello, Bernie from Vermont"
   }
   ```
+4.  It is possible to pipe JSON parameters to the `wsk` command by using the special filename `-` with `--params-file` or `-P`:
+  
+  ```
+  cat parameters.json | envsubst | wsk action invoke -r hello -P -
+  ```
+  
+  This example uses `envsubst` to replace environment variables in the JSON before passing them to the action invocation.
```
rabbah: for the sake of the example is `envsubst` necessary? I can see the counter argument it's already in a file why use `-`. How about using echo instead?

```bash
echo '{"name": "Alice"}' | wsk action invoke hello -r -P -
```

alexkli: I thought it's nice to have a more "real world" use case example.

The echo example can also be done simpler using `-p name Alice`. I chose `envsubst` as that's an actual case where you need to pipe from a file since some processing is going on, and I happened to be using `envsubst` in my case.

Do you think there is a need to be as simple as possible?

alexkli:
@rabbah How do I run the `WskBasicUsageTests.scala` easily and isolated?

dubeejw:
@alexkli, locally you can run `./gradlew tests:test -Dtest.single=WskBasicUsageTests` from the OpenWhisk root directory.

alexkli:
Ah, but that gives this error (I think I saw it before when trying to run tests):

```
* Where:
Build file '/Users/alex/Work/code/openwhisk/incubator-openwhisk/tests/build.gradle' line: 74

* What went wrong:
> java.io.FileNotFoundException: /Users/alex/Work/code/openwhisk/incubator-openwhisk/whisk.properties (No such file or directory)
```

Do I need to run a full local OpenWhisk and have the connection details in the `whisk.properties` before being able to run the tests?

dubeejw:
@alexkli, you do need a full OpenWhisk deployment setup locally.

alexkli:
@dubeejw Thanks. I have been using [devtools docker-compose](https://github.com/apache/incubator-openwhisk-devtools/tree/master/docker-compose), I was told that's easiest to setup (and it works fine). How do I get the `whisk.properties` file for it? What needs to go into it?

In the meantime I found `tools/build/redo props` which says it will `build whisk.properties file (required for tests)`. Had to install ansible using `brew install ansible` (which installed a lot of dependencies...).

However running `tools/build/redo props` does not create a `whisk.properties`, it outputs:
```
ansible-playbook -i environments/docker-machine properties.yml
 [WARNING]: Unable to parse /Users/alex/Work/code/openwhisk/incubator-openwhisk/ansible/environments/docker-machine as an inventory source

 [WARNING]: No inventory was parsed, only implicit localhost is available

 [WARNING]: Could not match supplied host pattern, ignoring: all

 [WARNING]: provided hosts list is empty, only localhost is available
...
skipping: no hosts matched
```

AFAICS, the wsk cli tests don't (all?) require a server? Is there no way to run them without?

alexkli:
While I have to wait for help with getting the tests to run, I had a look at `WskBasicUsageTests.scala`.

It seems there is not any positive test for `-P`, only failure tests if no filename is given or if it's an empty file. I feel like I am not familiar enough with scala and the code base here to know how to add such a positive test.

dubeejw:
@akrabat, check here for the valid `-P` tests.

https://github.com/apache/incubator-openwhisk/blob/master/tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala#L193

https://github.com/apache/incubator-openwhisk/blob/master/tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala#L920

https://github.com/apache/incubator-openwhisk/blob/master/tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala#L920

dubeejw:
@alexkli, I'm not familiar with the devtools project. Perhaps @rabbah has some suggestions.

rabbah:
Try wskdev setup props
If you have not run ansible before the hosts file may not exist and the error suggests this (i guess we have a missing depedendencd).

alexkli:
`wskdev setup props` fails similar to `tools/build/redo props`, as they seem to run the same and expect a running local ansible based OpenWhisk.

However, after some fiddling around, I was able to successfully run at least the WskBasicUsageTests

    ./gradlew tests:test -Dtest.single=WskBasicUsageTests

against my local OpenWhisk from the [docker-compose setup](https://github.com/apache/incubator-openwhisk-devtools/tree/master/docker-compose) using this minimal `whisk.properties` file in the root of the checkout:

```properties
openwhisk.home=$OPENWHISK_CHECKOUT_DIR

testing.auth=$USER_HOME/tmp/openwhisk/src/ansible/files/auth.guest
whisk.ssl.cert=$USER_HOME/tmp/openwhisk/apigateway/ssl/openwhisk-server-cert.pem

edge.host=$LOCAL_IP:443
whisk.api.host.proto=https
whisk.api.host.name=$LOCAL_IP
whisk.api.host.port=443

vcap.services.file=
use.cli.download=false
```

Replace the variables in there:
* `$OPENWHISK_CHECKOUT_DIR`: absolute path to the openwhisk git checkout
* `$USER_HOME`: absolute path to your user home (e.g. `/Users/alex`), tilde `~` does not work
* `$LOCAL_IP`: local IP (`localhost` might work too)

docs/actions.md
rabbah: OK - I was going for simpler example but after your explanation it's good to have a real use case.

alexkli:
@dubeejw @rabbah Thanks for the help. Once I got the tests running and the project in IntelliJ with Scala support, it was fairly straightforward.

PR updated with
* tests for `-P -` option for: action/package/trigger create, and action invoke with file and stdin cases
* updated the i18n json file

dubeejw:
PG3 1328 ⏳

rabbah:
Unfortunately this will require some git surgery. The go changes need to go in the cli repo now. The rest can stay here I think.

alexkli:
@rabbah are you asking something from my side? I am not familiar with the repo split.

rabbah:
I’ll try to spare you the overhead.

dubee:
Closing this one as the CLI has moved to incubator-openwhisk-cli and incubator-openwhisk-client-go.

PR #2852, opened 2017-10-12
invoker: use locally available image if docker pull fails

alexkli:
I have a local OW setup using devtools docker-compose and provide my own container via `--docker`. Currently, the invoker will always do `docker pull`, and if that fails to resolve, will fail the invocation, even if the image is available locally.

This patch will still do a `docker pull` first, to try and get the latest image, but if that fails, will check if the image is present by calling `docker inspect <image>` and check for the exit code (non-zero if image is not found). If it's not present, it will fail as before, but if present, it will continue with `docker run` (new behavior).

Beware, I am not very familiar with Scala. I feel this can probably be done better. It was just a quick proof of concept.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
@@ -102,6 +102,15 @@ class DockerClient(dockerHost: Option[String] = None)(executionContext: Executio
       runCmd("pull", image).map(_ => ()).andThen { case _ => pullsInFlight.remove(image) }
     })
 
+  def imagePresent(image: String)(implicit transid: TransactionId): Future[String] = {
+    val cmd = dockerCmd ++ Seq("inspect", "--format={{.Config.Image}}", image)
+    executeProcess(cmd: _*)/*.andThen {
+      case Success(_) => Future.successful(_)
+      case Failure(t) => Future.failed(_)
+    }*/
+    //runCmd("inspect", "--format={{.Config.Image}}", image).map(_ => ())
+  }
+
   def isOomKilled(id: ContainerId)(implicit transid: TransactionId): Future[Boolean] =
     runCmd("inspect", id.asString, "--format", "{{.State.OOMKilled}}").map(_.toBoolean)
 
@@ -180,6 +189,14 @@ trait DockerApi {
    * @return a Future completing once the pull is complete
    */
   def pull(image: String)(implicit transid: TransactionId): Future[Unit]
+
+  /**
+   * Checks if image is present locally.
+   *
+   * @param image the image to check
+   * @return a Future completing successful wif the image is present locally, or failing if not
+   */
+  def imagePresent(image: String)(implicit transid: TransactionId): Future[String]
 
   /**
    * Determines whether the given container was killed due to
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
@@ -103,12 +103,9 @@ class DockerClient(dockerHost: Option[String] = None)(executionContext: Executio
     })
 
   def imagePresent(image: String)(implicit transid: TransactionId): Future[String] = {
+    // --format only added to keep the output minimal
     val cmd = dockerCmd ++ Seq("inspect", "--format={{.Config.Image}}", image)
-    executeProcess(cmd: _*)/*.andThen {
-      case Success(_) => Future.successful(_)
-      case Failure(t) => Future.failed(_)
-    }*/
-    //runCmd("inspect", "--format={{.Config.Image}}", image).map(_ => ())
+    executeProcess(cmd: _*)
   }
 
   def isOomKilled(id: ContainerId)(implicit transid: TransactionId): Future[Boolean] =
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
@@ -89,7 +89,11 @@ object DockerContainer {
       params
     val pulled = if (userProvidedImage) {
       docker.pull(image).recoverWith {
-        case _ => Future.failed(BlackboxStartupError(s"Failed to pull container image '${image}'."))
+        case _ =>
+          // check if image exists locally and reuse that one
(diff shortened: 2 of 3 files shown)


alexkli:
This should include a unit test in `DockerContainerTests.scala`, but I am not quite sure how to write it in Scala.

rabbah:
Trying to understand the need for this - did you consider changing the manifest to inform the invoker of locally available images it should just use directly e.g., https://github.com/apache/incubator-openwhisk/blob/master/ansible/group_vars/all#L94

csantanapr:
You can push your docker action to dockerhub  and then specify the image $user/myimage

You can also push to another registry and configure access to your local environment 

If you don't want skip the pull all together then what @rabbah suggested should work

build your local image and tag it with the default prefix whisk
docker build . -t whisk/myblackbox 

Then in the manifest add "myblackbox" as a blackbox image 
If this doesn't work then we have a bug :-) and should fix

alexkli:
The need is local development with OpenWhisk and testing docker images (action containers) without having to publish them. It’s an unnecessary step to push and pull the (possibly large) image over the network after each change, and because it might be a proprietary image and setting up a private registry is extra work.

The suggestion with the manifest are also extra work and can’t be configured at runtime, i.e. if you have a new image or change its name, you’d have to redeploy the local OpenWhisk AFAIU.

alexkli:
@csantanapr I try the wisk/ prefix.

Another production problem I could see: if docker pull fails because of a temporary network error, the entire (cold) invocation will fail, even though the latest image is present locally, because it _always_ does docker pull first from what I see.

csantanapr:
I think adding a check for tags having the assumption that if image name contains a :tag the invoker can just do the docker CLI run if image is not already locally docker CLI will download and cache skipping the download on subsequent runs

csantanapr:
The corner case is when the tag is not specified or using latest we delegate the checking band compare to the docker pull to do it

csantanapr:
when the action gets updated is our indication that checking to see if there is a latest image is available and doing the docker pull is the route failing to use a probably old image might not be desired in some cases

csantanapr:
Today Using manifest to specify set of local images should be easy
You can specify multiple in advanced like 
Test1
Test2
Test3
Test4

Only the invoker needs to be redeploy if modifying that list  I believe 

My proposal on using tag then the action can be created with image tag making the pull to be skipped but this is a proposal not implented currently

rabbah:
For local dev, tagging the image a whisk/ should suffice.

csantanapr:
Yep I agree prefix whisk I think you meant, not tag 
Yeah should be trivial if not clear we can provide example

alexkli:
Using a "whisk" prefix alone does not work, I get
> Failed to pull container image 'wisk/proxyaction'

Looking at the invoker logs, I see it's still trying to do a pull first and bail out when that throws an error:

```
...[INFO] [#tid_10] [DockerClientWithFileAccess] running /usr/bin/docker pull wisk/proxyaction [marker:invoker_docker.pull_start:64] 
...[ERROR] [#tid_10] [DockerClientWithFileAccess] code: 1, stdout: Using default tag: latest, stderr: Error response from daemon: pull access denied for wisk/proxyaction, repository does not exist or may require 'docker login' [marker:invoker_docker.pull_error:1578:1514] 
```

alexkli:
FWIW, while I had a typo in my previous comment (`wisk`), using the correctly typed `whisk` as prefix fails as well.

rabbah:
You are right - and shame on me, I should have known because I changed this.
https://github.com/apache/incubator-openwhisk/blob/9be13e2c6d581f42b60351f05a11e4aa71ca31d9/common/scala/src/main/scala/whisk/core/entity/Exec.scala#L269 a container image is always pulled unless it's white listed in the manifest.

csantanapr:
As I said earlier you need to update manifest also
>Today Using manifest to specify set of local images should be easy
You can specify multiple in advanced like
Test1
Test2
Test3
Test4

alexkli:
@csantanapr  Sure, but fiddling with the manifest and having to restart OpenWhisk every time you make a change to the manifest can be a bit of a hurdle.

Is there anything problematic regarding production deployments that you see with this change, which is intended for development purposes?

As I can see, this fix might also help in case of temporary network errors or docker registry hiccups that would make docker pull fail, but would not make invocations fail unnecessarily as the image would be locally available.

rabbah:
A semantic change could be observed here - if an image is removed from docker hub, the pull will fail. However the action will continue to run in a particular invoker which has a previously pulled image, until that image is purged locally. This is a new behavior. 

We should separate the convenience of local development from resilience to network partitions. An option for the former is to restore an implicit pull bypass if the image prefix matches the local registry prefix. This could be done with a feature toggle through ansible as well for added control.

csantanapr:
>An option for the former is to restore an implicit pull bypass if the image prefix matches the local registry prefix. This could be done with a feature toggle through ansible as well for added control.

I agree with the compromise @rabbah +1

csantanapr:
I'm also OK with a more strict option in addition, to tell the invoker to not do any docker pull at all, to assume that any image required will be already available locally.

This is a theoretical  use case on a close down security deployment that no image should be pulled dynamically.

rabbah:
@alexkli what do you think about the compromise which would allow for the convenience for local testing but without changing the behavior more globally?

alexkli:
An easily configurable prefix would work as well.

I just generally find the notion of “check network first, if it fails, abort, and only then check locally” a bit weird. You’d expect the opposite, plus maybe some interval when it would check for new  versions. It would also save time when you avoid the docker pull done every time.

Also, I believe docker run will do a docker pull implicitly if an image is not found locally. So you’d only need docker pull for updates to existing images.

rabbah:
We break out the pull and the run for performance reasons: the pull can be independent of the run, and this pipelines requests better.

The issue with local first, network second is versioning and life cycle management. I'd rather we address this more systematically with an eager pull into a registry on `wsk action create` rather than dynamically at runtime.

alexkli:
> eager pull into a registry on `wsk action create` rather than dynamically at runtime

Agree, that would be useful. You'd still need worry about not aborting with an error if the (2nd) `pull` at invocation time fails.

The question is also how much control the action developer (that uses a custom docker image) has about updates to the image.

rabbah:
Today an update to the action forces a docker pull. Otherwise the action will pull as needed to bring up a new container. So the versioning is very loosely coupled. 

See #3052 for alternate fix for local development. If agreeable I suggest we close this patch.

rabbah:
Closing in favor of #3052 which was merged.

PR #2853, opened 2017-10-12
update kinds

csantanapr:


core/controller/src/main/resources/apiv1swagger.json
```diff
@@ -1554,13 +1554,10 @@
...
-                        "nodejs",
                         "nodejs:6",
-                        "nodejs:default",
-                        "python",
-                        "swift",
-                        "swift:3",
-                        "swift:default",
+                        "python:2",
+                        "python:3",
+                        "swift:3.1.1",
```
rabbah: dropped swift 3.

csantanapr: it's encouragement documentation for folks to use `swift:3.1.1` :-)
I can add back, and remove later when we deprecate and announce in mailing list

csantanapr:
@rabbah added `swift:3` please review

csantanapr:
PG 3 1342

PR #2855, opened 2017-10-13
Cache database attachments

dubee:
Caches attachments with corresponding actions when an action is created/updated. The attachment is also cached on a fetch if it is not already cached.

Depends on https://github.com/apache/incubator-openwhisk/pull/2832
  
Closes https://github.com/apache/incubator-openwhisk/issues/2672

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
@@ -120,12 +120,14 @@ trait DocumentFactory[W <: DocumentRevisionProvider] extends MultipleReadersSing
     }
   }
 
-  def attach[Wsuper >: W](
-    db: ArtifactStore[Wsuper],
-    doc: W,
-    attachmentName: String,
-    contentType: ContentType,
-    bytes: InputStream)(implicit transid: TransactionId, notifier: Option[CacheChangeNotification]): Future[DocInfo] = {
+  def attach[Wsuper >: W](db: ArtifactStore[Wsuper],
+                          doc: W,
+                          attachmentName: String,
+                          contentType: ContentType,
+                          bytes: InputStream,
+                          postProcess: Option[W => W] = None)(
+    implicit transid: TransactionId,
+    notifier: Option[CacheChangeNotification]): Future[DocInfo] = {
 
     Try {
       require(db != null, "db undefined")
@@ -137,10 +139,11 @@ trait DocumentFactory[W <: DocumentRevisionProvider] extends MultipleReadersSing
       val key = CacheKey(doc)
       val docInfo = doc.docinfo
       val src = StreamConverters.fromInputStream(() => bytes)
+      val cacheDoc = postProcess map { _(doc) } getOrElse doc
 
-      cacheUpdate(doc, key, db.attach(docInfo, attachmentName, contentType, src) map { newDocInfo =>
-        doc.revision[W](newDocInfo.rev)
-        doc.docinfo
+      cacheUpdate(cacheDoc, key, db.attach(docInfo, attachmentName, contentType, src) map { newDocInfo =>
+        cacheDoc.revision[W](newDocInfo.rev)
+        cacheDoc.docinfo
       })
     } match {
       case Success(f) => f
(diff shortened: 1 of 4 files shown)


common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -275,12 +278,10 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
...
           val b64s = Base64.getEncoder().wrap(boas)
 
-          getAttachment[A](db, action.docinfo, attachmentName, b64s).map { _ =>
+          getAttachment[A](db, action, action.docinfo, attachmentName, b64s, Some { a: WhiskAction =>
             b64s.close()
-            val newAction = action.copy(exec = exec.inline(boas.toByteArray))
-            newAction.revision(action.rev)
-            newAction
-          }
+            a.copy(exec = exec.inline(boas.toByteArray))
```
dubeejw: Need to update revision?

common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
```diff
@@ -245,26 +261,35 @@ trait DocumentFactory[W] extends MultipleReadersSingleWriterCache[W, DocInfo] {
...
+  def getAttachment[Wsuper >: W](
+    db: ArtifactStore[Wsuper],
+    entity: W,
+    doc: DocInfo,
+    attachmentName: String,
+    outputStream: OutputStream,
+    postProcess: Option[PostProcessEntity[W]] = None)(implicit transid: TransactionId, mw: Manifest[W]): Future[W] = {
 
     implicit val ec = db.executionContext
+    implicit val notifier: Option[CacheChangeNotification] = None
```
dubeejw: Cache invalidation callback should probably be assigned to the actual callback here.

dubeejw:
PG3 1186 🔵

dubeejw:
PG2 2466 🔵

markusthoemmes:
Does this mean we're going to cache attachments in the controller? Is that intended?

bbrowning:
I've observed some very poor performance for Java actions and with the help of @rabbah I think I've narrowed it down to the Invokers fetching the Jar attachment on every activation. So, I can confirm both that not caching database attachments is a problem and that this PR is needed.

However, I've as of yet been unable to verify that this PR works as intended. https://gist.github.com/bbrowning/6f18deb9f82282700cd3033deaf127b9#file-gistfile1-txt-L29 is a gist of me invoking an action twice in a row with this PR applied. You'll notice that on the highlighted line, the 2nd invocation still fetches the attachment directly from the database.

There are also a couple of test failures with the PR when applied to the current codebase. Those look to be related to the reduction of Info level logging and some of the log markers the tests are expecting no longer get logged. Here's the failing test output for reference - https://gist.github.com/bbrowning/727158c6c668d9e81d57543fe67f5ff0

bbrowning:
Another update - the tests can be fixed by just removing the "completed uploading attachment..." lines from the expected output checks in ActionsApiTests.

The caching works fine as-is when fetching attachments via commands like `wsk action get --save foo`. However, it does not cache actions when fetching attachments via the Invoker. The root cause of this is that the CacheKey wasn't taking the revision into account when updating the cache after calling `getAttachment` and can be fixed by https://gist.github.com/bbrowning/1eb6f24432af96fea264f0152d5cd222

The tests will need to be updated to verify the caching works when a revision is included when getting an action and its attachment. I have manually verified that this works when changed as I linked above.

dubee:
@bbrowning, do you think `key` needs to be change in `attach` and `put` as well?

bbrowning:
Speaking of performance improvements, I thought I'd share some basic throughput numbers here. On my test cluster with a simple Hello World Java action I was only able to sustain about 25 action activations per second across the entire cluster before this change. Afterwards, I was able to sustain about 410 action activations per second. The Invoker load, end-user activation latency, and waitTime as reported in the activation results dropped dramatically as a result of this change.

dubee:
@bbrowning, how many invokers are you using in your test cluster?

dubee:
I'm not very concerned about the controller caching Java actions. Mostly the action invoke path is used in the controller. With that action invoke API in the controller, only an action metadata is cached, which does not include the action code.

@chetanmeh, do you think fetching from actions from disk will be faster than fetching them from Cloudant?

chetanmeh:
> @chetanmeh, do you think fetching actions from disk will be faster than fetching them from CouchDB?

For our usecase we would be pulling attachments from Azure Blob Storage where latency may be higher (based on my previous experience with S3). Would have a benchmark to compare before proceeding with a disk based cache.

bbrowning:
@dubee I had 4 regular and 1 blackbox Invokers in that test cluster. But, everyone's results will vary based on hardware, resources allocated to Invokers, disk speed of CouchDB cluster, network speed, etc.

dubee:
@rabbah, @markusthoemmes, thoughts on this one? Seems like there's a fairly nice performance boost for invoking Java actions.

rabbah:
Can you exclude the caching only in the controller? Otherwise, yes this is worth having until we have the more general streaming support.

dubee:
PG4 1600 ⏳

dubee:
PG4 1600 🔵

dubee:
@rabbah, looks like caching is enabled/disabled on the `WhiskAction` singleton level, which is common code shared among the controller and invoker.

csantanapr:
@dubee @rabbah this ready to merge it would be a step forward

dubee:
@csantanapr, just need a 👍.

PR #2857, opened 2017-10-16
Emit metrics via kamon

mhenke1:
This PR implements the capability to send metric information via Kamon (http://kamon.io/)
to statsd (https://github.com/etsy/statsd) instead or additionally to the existing metrics information 
which is written into the system logs.

Per default Kamon metrics are disabled and metrics as log information is enabled. 
The system behavior is expected to be the same as before,
but can be changed via Ansible configuration by adding the following settings the
"group_vars/all" file of an environment:
 
- metrics_kamon   -> enable metrics via kamon (true/false with default false)
- metrics_kamon_statsd_host -> hostname (or ip address) of the statsd server
- metrics_kamon_statsd_port-> port number of the statsd server
- metrics_log: -> enable metrics via log file (true/false with default true)
   (with this PR all messages adjourning the metric logs will still be written to the log file to avoid
loss of information)

Example:
```
# Metrics configuration
metrics_kamon: false
metrics_kamon_statsd_host: '192.168.99.100'
metrics_kamon_statsd_port: '8125'
metrics_log: true
```

Local testing:
There is an integrated docker image available that installs statsd and grafana
https://github.com/kamon-io/docker-grafana-graphite


ansible/group_vars/all
```diff
@@ -105,7 +105,11 @@ limits:
 controller:
   basePort: 10001
   heap: "{{ controller_heap | default('2g') }}"
-  arguments: "{{ controller_arguments | default('') }}"
+  arguments: "{% if metrics.kamon == true %} 
+                {{ controller_arguments + ' -Dkamon.statsd.hostname=' + metrics_kamon_statsd_host + ' -Dkamon.statsd.port=' + metrics_kamon_statsd_port | default('') }} 
+              {% else %} 
+                {{ controller_arguments | default('') }} 
+              {% endif %}"
```
markusthoemmes: Let's fix these bits, I have an ENV->JVM System property adapter script stashed in another PR. I'll finish that off, then we can clean these up (We need system props for SPIs for example anyway).

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -55,7 +55,14 @@ case class TransactionId private (meta: TransactionMetadata) extends AnyVal {
    */
   def mark(from: AnyRef, marker: LogMarkerToken, message: String = "", logLevel: LogLevel = InfoLevel)(
     implicit logging: Logging) = {
-    logging.emit(logLevel, this, from, createMessageWithMarker(message, LogMarker(marker, deltaToStart)))
+
+    if (meta.metricsLog) {
+      logging.emit(logLevel, this, from, createMessageWithMarker(message, LogMarker(marker, deltaToStart)))
+    }
```
markusthoemmes: This is a dangerous change in behavior. I'd rather have this flagged by additional information like so:

```scala
if (message.nonEmpty) {
  val print = if(meta.metricsLog) createMessageWithMarker(message, LogMarker(marker, deltaToStart)) else message
  logging.emit(logLevel, this, from, print)
}

Assuming that you still want the message to be printed if it was given by the developer (we should probably clean that up later and log the lines explicitly vs. implicitly through this class, but it keeps the changes small for now)

File: common/scala/src/main/scala/whisk/common/Logging.scala [hunk 3]
```diff
@@ -178,6 +179,34 @@ object LogMarkerToken {
...
+  def emitMetric(wm: WhiskMetric, isFailed: Boolean = false) = wm match {
+    case WhiskCounterMetric(component, action) =>
+      Kamon.metrics
+        .counter(getMetricName(component, action, isFailed))
+        .increment(1)
+    case WhiskHistogramMetric(component, action, value) =>
+      Kamon.metrics
+        .histogram(getMetricName(component, action, isFailed))
+        .record(value)
+  }
```
markusthoemmes: Why not have an overloaded version of `emitMetric` and get rid of the intermediary types + non-needed allocations?

markusthoemmes:
Sorry oversaw the **WIP** tag, my bad.

common/scala/src/main/scala/whisk/common/TransactionId.scala
mhenke1: I changed the PR (via amend) to emit simple log message in the original debug level to avoid loss of information. I will tackle log reduction in a following commit

common/scala/src/main/scala/whisk/common/Logging.scala
mhenke1: I have simplified the MetricsEmitter (via amend) to reduce allocations

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -178,6 +180,26 @@ object LogMarkerToken {
   }
 }
 
+object MetricEmitter {
+
+  val metrics = Kamon.metrics
+
+  def getMetricName(component: String, action: String, state: String) = Seq(component, action, state).mkString("_")
```
markusthoemmes: Superfluous method now?

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -178,6 +180,26 @@ object LogMarkerToken {
...
+
+  def emitCounterMetric(token: LogMarkerToken) = {
+    println("************************MH:" + token.toString)
```
markusthoemmes: Bad push?

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -55,7 +55,19 @@ case class TransactionId private (meta: TransactionMetadata) extends AnyVal {
    */
   def mark(from: AnyRef, marker: LogMarkerToken, message: String = "", logLevel: LogLevel = InfoLevel)(
     implicit logging: Logging) = {
-    logging.emit(logLevel, this, from, createMessageWithMarker(message, LogMarker(marker, deltaToStart)))
+
+    if (TransactionId.metricsLog) {
+      logging.emit(logLevel, this, from, createMessageWithMarker(message, LogMarker(marker, deltaToStart)))
+    } else {
+      if (message.nonEmpty) {
```
markusthoemmes: Also known as 

```scala
} else if (message.nonEmpty) {
  ...
}
```

Applies to all occurences.

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -108,13 +141,27 @@ case class TransactionId private (meta: TransactionMetadata) extends AnyVal {
...
+        createMessageWithMarker(message, LogMarker(endMarker, deltaToStart, Some(deltaToMarker(startMarker)))))
+    } else {
+      if (message.nonEmpty) {
+        logging.emit(logLevel, this, from, message)
+      }
+    }
+
+    if (TransactionId.metricsKamon) {
+      val endTime = Instant.now(Clock.systemUTC)
+      val metricValue = deltaToMarker(startMarker, endTime)
```
markusthoemmes: `deltaToMarker` has a default parameter, already getting the correct endTime for you. No need to explicitly get it again.

PR #2858, opened 2017-10-16
Make Docker action container cleanup during invoker startup / shutdown more robust

sven-lange-last:
Remove all Docker containers on the system that match the naming filter `wsk<invoker instance number>_` at startup and shutdown. In the past, only running containers matching the filter were removed - paused containers are considered to be running. The new strategy will also remove containers that are stopped or could never be started properly for any reason. From time to time, we see action containers that are not running but should be cleaned up.

If Docker commands used to remove containers (`docker ps`, `docker-runc resume` and `docker rm`) do not return within 30 seconds during startup, abort startup and let Docker re-start the invoker container. We have seen systems with heavy load where Docker commands take too long to complete. In that situation, we need a defined behavior, i.e. re-start and re-try the cleanup. At the moment, an uncaught exception will end the main thread leaving the invoker inoperational.

Also changed the invoker initialization sequence such that Docker action container cleanup is performed as soon as possible - in particular, before starting other tasks / actors. At the moment, a cleanup timeout would end the main thread while the activation message feed is already running and consuming messsages that will never be processed.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
@@ -78,8 +78,7 @@ class DockerContainerFactory(config: WhiskConfig, instance: InstanceId, paramete
     try {
       removeAllActionContainers()
     } catch {
-      case e @ (_: TimeoutException | _: InterruptedException) =>
-        logging.error(this, s"Failed to remove action containers: ${e.getMessage}")
+      case e: Exception => logging.error(this, s"Failed to remove action containers: ${e.getMessage}")
     }
   }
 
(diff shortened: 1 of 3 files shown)


sven-lange-last:

sven-lange-last:
PG 1 / 2143 running.

sven-lange-last:
PG 1 / 2157 running with latest changes.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
```diff
@@ -69,27 +70,53 @@ class DockerContainerFactory(config: WhiskConfig, instance: InstanceId, paramete
...
-              docker.rm(id)(TransactionId.invokerNanny)
-            }
-        }
-        Future.sequence(removals)
+    implicit val transid = TransactionId.invoker
+    try {
+      removeAllActionContainers()
+    } catch {
+      case e @ (_: TimeoutException | _: InterruptedException) =>
+        logging.error(this, s"Failed to remove action containers: ${e.getMessage}")
```
markusthoemmes: Let's just use a generic catch here? Cleanup might fail because of lots of other reasons (failures to remove for instance), which would be helpful to have in the log as well?

sven-lange-last: Makes sense - done.

Issue #2863, opened 2017-10-17
Add the support of certificate checking for secure mode


houshengbo:
Currently in the CLI code, we only handle the cert and key for the insecure mode with -i. We definitely need to handle them for the secure mode without -i.

PR #2864, opened 2017-10-17
Add the support of certificate checking for secure mode

houshengbo:
Closes: #2863

### DIFF
--- a/tools/cli/go-whisk/whisk/client.go
+++ b/tools/cli/go-whisk/whisk/client.go
@@ -94,28 +94,6 @@ var DefaultObfuscateArr = []ObfuscateSet{
 
 func NewClient(httpClient *http.Client, config *Config) (*Client, error) {
 
-    // Disable certificate checking in the dev environment if in insecure mode
-    if config.Insecure {
-        Debug(DbgInfo, "Disabling certificate checking.\n")
-        var tlsConfig *tls.Config
-        if config.Cert != "" && config.Key != "" {
-            if cert, err := tls.LoadX509KeyPair(config.Cert, config.Key); err == nil {
-                tlsConfig = &tls.Config{
-                    Certificates: []tls.Certificate{cert},
-                    InsecureSkipVerify: true,
-                }
-            }
-        }else{
-            tlsConfig = &tls.Config{
-                InsecureSkipVerify: true,
-            }
-        }
-
-        http.DefaultClient.Transport = &http.Transport{
-            TLSClientConfig: tlsConfig,
-        }
-    }
-
     if httpClient == nil {
         httpClient = http.DefaultClient
     }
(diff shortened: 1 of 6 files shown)


houshengbo:
2154 passed.

tools/cli/go-whisk/whisk/client.go
```diff
@@ -94,30 +94,39 @@ var DefaultObfuscateArr = []ObfuscateSet{
...
+                tlsConfig = &tls.Config{
+                    Certificates: []tls.Certificate{cert},
+                }
+            }
         }
     }
 
-    if httpClient == nil {
-        httpClient = http.DefaultClient
+    httpClient.Transport = &http.Transport{
```
csantanapr: I think we have a different logic now
```
var tlsConfig *tls.Config
httpClient.Transport = &http.Transport{
    TLSClientConfig: tlsConfig,
}	
```
this will be the result when config.Insecure config.Cert and config.key are not defined.

houshengbo: Yes, I just changed the PR by adding 
tlsConfig = &tls.Config {
       InsecureSkipVerify: config.Insecure,
}
tlsConfig will hve a default value based on the config.Insecure, no matter cert or key is available.

tools/cli/go-whisk/whisk/client.go
```diff
@@ -94,30 +94,22 @@ var DefaultObfuscateArr = []ObfuscateSet{
...
+    }
 
-        http.DefaultClient.Transport = &http.Transport{
-            TLSClientConfig: tlsConfig,
+    tlsConfig := &tls.Config {
+        InsecureSkipVerify: config.Insecure,
+    }
+
+    if config.Cert != "" && config.Key != "" {
+        if cert, err := ReadX509KeyPair(config.Cert, config.Key); err == nil {
```
mdeuser: should check for `err != nil` case here.. log error, return failure with message.

houshengbo: We have a test case set the cert and key files via wsk property set --cert --key. If we raise an error if err is not nil, that test case will fail, because cert and key files are not available.

mdeuser: i think we should log the failure at a minimum as this information might lead to the root cause of the eventual failure.

i'm still not yet convinced that the cli should continue if it was configured to use some specific certificate and that certificate cannot be read/used for some reason.

if the command is `wsk property`, then we can ignore the cert/key when creating the http client if both are not specified.

tools/cli/go-whisk/whisk/client.go
```diff
@@ -67,14 +67,14 @@ type Client struct {
 }
 
 type Config struct {
```
dubeejw: Can you align all the properties in `Config`?

houshengbo: Done.

tools/cli/go-whisk/whisk/client.go
houshengbo: This is already done by the function: func (c *Client) LoadX509KeyPair() error .

tools/cli/go-whisk/whisk/client.go
```diff
@@ -162,11 +140,43 @@ func NewClient(httpClient *http.Client, config *Config) (*Client, error) {
     return c, nil
 }
 
+func (c *Client) LoadX509KeyPair() error {
+    tlsConfig := &tls.Config {
+        InsecureSkipVerify: c.Config.Insecure,
+    }
+
+    if c.Config.Cert != "" && c.Config.Key != "" {
```
mdeuser: should a warning/error be returned if the cert is partially configured - i.e. a key is specified, but not the cert or a cert is specified but not the key?  currently, this incorrect configuration is silently ignored.

houshengbo: done.

tools/cli/go-whisk/whisk/client.go
```diff
@@ -162,11 +140,62 @@ func NewClient(httpClient *http.Client, config *Config) (*Client, error) {
...
+            tlsConfig.Certificates = []tls.Certificate{cert}
+        } else {
+            errStr := wski18n.T("Unable to load the X509 key pair due to the following reason: {{.err}}",
+                map[string]interface{}{"err": err})
+            werr := MakeWskError(errors.New(errStr), EXIT_CODE_ERR_GENERAL, DISPLAY_MSG, NO_DISPLAY_USAGE)
+            return werr
+        }
+    } else if !c.Config.Insecure {
+    	if c.Config.Cert == "" {
+    	    warningStr := wski18n.T("Please configure the missing Cert file, if there is a security issue accessing the service.\n")
```
mdeuser: - Debug/tracing strings do not need to be enable for translation
- Can the same message be used for both debug/trace as well as the CLI response message?
- Nit.  Maybe consolidate conditional checking:
```
    if !Cert && Key {
       ...The Cert file is not configured...
    } else if Cert && !Key {
       ...The Key file is not configured...
    }
```

mdeuser:
include test automation for
- corrupt/invalid cert and/or key.  test `ReadX509KeyPair` failure
- secure connection && (Cert && !Key).  test for expected error message
- secure connection && (Key &7 !Cert).  test for expected error message

houshengbo:
@mdeuser Can we do the test in the new client go repo? Because it is easier for us to mock ReadX509KeyPair in test cases written in Go.

We do not have any unit test or automation test written in go in the openwhisk repo. It is worthy the effort adding them to the openwhisk repo, since it is the plan to add them in the new repos. If we want to add the test cases written in Scala, we cannot cross the language barrier to mock the go function.

tools/cli/go-whisk/whisk/client.go
houshengbo: I suggest to keep the current logic. We need to log a warning as long as c.Config.Cert is missing, no matter Key is missing or not.

houshengbo:
Test cases added to cover the code.

mdeuser:
@houshengbo - new tests look good!  is there a pg?

houshengbo:
PG 1 2154 passed without the test cases

mdeuser:
pg1/2219 includes new tests

PR #2868, opened 2017-10-18
Make redis optional on deployment

cbickel:
As a redis is optional for an invoker, the deployment should also work without any redis in the hosts file.


rabbah:
is this change actually sound? it didn't change the invoker code which negotiates with redis - and i suspect we wouldn't see it since we're not activating that part of the code yet.

PR #2870, opened 2017-10-18
Remove broken containers if docker run fails

sven-lange-last:
If `docker run` fails with exit code 125 and prints a valid container ID to stdout, a broken Docker container was left behind. This PR detects the described situation and immediately removes the broken container to save disk space. In addition, the Docker daemon seems to slow down with the number of containers, so cleaning up broken containers can also have a positive performance impact.

The main root cause for broken containers is a failure when extracting the container's file system - for example, due to missing disk space or exhausted inodes. This PR does fully resolve low disk space or exhausted inodes but it can help relax such situations. For example, if a container with a very large file system cannot be created, cleaning it up may enable the creation of a smaller container.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
@@ -36,8 +36,10 @@ import whisk.core.containerpool.ContainerId
 import whisk.core.containerpool.ContainerAddress
 
 object DockerContainerId {
+
+  val containerIdRegex = """^([0-9a-f]{64})$""".r
+
   def parse(id: String): Try[ContainerId] = {
-    val containerIdRegex = """^([0-9a-f]{64})$""".r
     id match {
       case containerIdRegex(_) => Success(ContainerId(id))
       case _                   => Failure(new IllegalArgumentException(s"Does not comply with Docker container ID format: ${id}"))
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
@@ -36,7 +36,7 @@ import whisk.core.containerpool.ContainerId
 import whisk.core.containerpool.ContainerAddress
 
 object DockerContainerId {
-  def convertToContainerId(id: String): Try[ContainerId] = {
+  def parse(id: String): Try[ContainerId] = {
     val containerIdRegex = """^([0-9a-f]{64})$""".r
     id match {
       case containerIdRegex(_) => Success(ContainerId(id))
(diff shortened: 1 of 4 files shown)


sven-lange-last:
PG 4 / 904 in queue.

sven-lange-last:

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
```diff
@@ -35,6 +35,16 @@ import scala.collection.concurrent.TrieMap
 import whisk.core.containerpool.ContainerId
 import whisk.core.containerpool.ContainerAddress
 
+object DockerContainerId {
+  def convertToContainerId(id: String): Try[ContainerId] = {
```
markusthoemmes: `parse`?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
```diff
@@ -63,8 +73,26 @@ class DockerClient(dockerHost: Option[String] = None)(executionContext: Executio
...
+      .map {
+        ContainerId(_)
+      }
+      .recoverWith {
+        // https://docs.docker.com/v1.12/engine/reference/run/#/exit-status
+        // Exit code 125 means an error reported by the Docker daemon.
+        // Examples:
+        // - Unrecognized option specified
+        // - Not enough disk space
+        case pre @ (_: ProcessRunningException) if pre.exitCode == 125 =>
```
markusthoemmes: `case pre: ProcessRunningException if pre.exitCode == 125`?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -98,7 +98,12 @@ object DockerContainer {
     for {
       _ <- pulled
       id <- docker.run(image, args).recoverWith {
-        case _ => Future.failed(WhiskContainerStartupError(s"Failed to run container with image '${image}'."))
+        case BrokenDockerContainer(brokenId, message) =>
+          docker.rm(brokenId)
+          Future.failed(
```
markusthoemmes: Do we want to wait for the remove to be finished or do we not care?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
sven-lange-last: Okay, will change the name.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
sven-lange-last: You are so right.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
sven-lange-last: I don't think we want to wait. If the removal fails, there is nothing we could do to recover and at this point, the Docker container creation has failed anyway so that we return a failed Future. Added a comment to the code.

A few lines below, we also remove the Docker container if `inspectIPAddress` fails and do not wait for the removal to complete.

sven-lange-last:
PG 4 / 906 running.

PR #2871, opened 2017-10-18
Invokers only require Redis if using dynamic Id assignment

dgrove-oss:
Make REDIS_HOST_NAME and REDIS_HOST_PORT optional for the invoker
and only trap as a configuration error when the invoker actually
tries to perform dynamicId assignment.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
@@ -50,7 +50,7 @@ object Invoker {
       WhiskEntityStore.requiredProperties ++
       WhiskActivationStore.requiredProperties ++
       kafkaHost ++
-      redisHost ++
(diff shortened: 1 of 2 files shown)


dgrove-oss:
Per discussion in slack with @jrcrossley3.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -94,6 +94,12 @@ object Invoker {
         id
       }
       .getOrElse {
+        if (config.redisHostName == "" || config.redisHostPort == "") {
```
rabbah: `== ""` could be `.isEmpty (or .nonEmpty for the opposite).
trim.isEmpty to avoid blanks?

dgrove-oss: changed to use trim.isEmpty

PR #2877, opened 2017-10-19
Better error controller message for authorization failure

### LINKED ISSUE #1684: better error message for authorization failure
```wsk action create --sequence cats /whisk.system/utils/echo,/someNamespace/list-documents -p dbname cats
error: Unable to create action: The supplied authentication is not authorized to access this resource. (code 6035175) 
```

Whenever we report an error message about "this resource",  we should instead explicitly name the resource that caused the problem in the error message.

Additionally, if "this resource" does not exist, we should produce an error message saying so explicitly.

mdeuser:
Fixes #1684

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/AuthorizedRouteDispatcher.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/AuthorizedRouteDispatcher.scala
@@ -26,6 +26,8 @@ import scala.concurrent.Future
 
 import akka.http.scaladsl.server.Directives
 import akka.http.scaladsl.model.HttpMethod
+import akka.http.scaladsl.model.StatusCodes.Forbidden
+import akka.http.scaladsl.model.StatusCodes.NotFound
 import akka.http.scaladsl.server.RequestContext
 import akka.http.scaladsl.server.RouteResult
 import akka.http.scaladsl.model.StatusCodes.InternalServerError
@@ -37,6 +39,7 @@ import whisk.core.entitlement._
 import whisk.core.entitlement.Resource
 import whisk.core.entity._
 import whisk.core.entity.size._
+import whisk.http.ErrorResponse
 import whisk.http.ErrorResponse.terminate
 import whisk.http.Messages
 
@@ -71,7 +74,16 @@ trait BasicAuthorizedRouteProvider extends Directives {
 
     onComplete(entitlementProvider.check(user, right, resource)) {
       case Success(_) => dispatchOp(user, right, resource)
-      case Failure(t) => handleEntitlementFailure(t)
+      case Failure(t) => t match {
+        case (r: RejectRequest) => r.code match {
+          case Forbidden =>
+            handleEntitlementFailure(RejectRequest(Forbidden, Some(ErrorResponse(Messages.notAuthorizedtoAccessResource(resource.fqname), transid))))
+          case NotFound =>
+            handleEntitlementFailure(RejectRequest(NotFound, Some(ErrorResponse(Messages.resourceDoesntExist(resource.fqname), transid))))
+          case _ => handleEntitlementFailure(t)
+        }
+        case _  => handleEntitlementFailure(t)
+      }
     }
   }
 
--- a/core/controller/src/main/scala/whisk/core/controller/AuthorizedRouteDispatcher.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/AuthorizedRouteDispatcher.scala
@@ -74,15 +74,21 @@ trait BasicAuthorizedRouteProvider extends Directives {
 
     onComplete(entitlementProvider.check(user, right, resource)) {
       case Success(_) => dispatchOp(user, right, resource)
-      case Failure(t) => t match {
-        case (r: RejectRequest) => r.code match {
-          case Forbidden =>
-            handleEntitlementFailure(RejectRequest(Forbidden, Some(ErrorResponse(Messages.notAuthorizedtoAccessResource(resource.fqname), transid))))
-          case NotFound =>
-            handleEntitlementFailure(RejectRequest(NotFound, Some(ErrorResponse(Messages.resourceDoesntExist(resource.fqname), transid))))
-          case _ => handleEntitlementFailure(t)
+      case Failure(t) =>
+        t match {
+          case (r: RejectRequest) =>
+            r.code match {
+              case Forbidden =>
+                handleEntitlementFailure(
+                  RejectRequest(
+                    Forbidden,
+                    Some(ErrorResponse(Messages.notAuthorizedtoAccessResource(resource.fqname), transid))))
+              case NotFound =>
+                handleEntitlementFailure(
+                  RejectRequest(NotFound, Some(ErrorResponse(Messages.resourceDoesntExist(resource.fqname), transid))))
+              case _ => handleEntitlementFailure(t)
+            }
         }
-      }
     }
   }
 
--- a/core/controller/src/main/scala/whisk/core/controller/AuthorizedRouteDispatcher.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/AuthorizedRouteDispatcher.scala
@@ -82,7 +82,6 @@ trait BasicAuthorizedRouteProvider extends Directives {
             handleEntitlementFailure(RejectRequest(NotFound, Some(ErrorResponse(Messages.resourceDoesntExist(resource.fqname), transid))))
           case _ => handleEntitlementFailure(t)
         }
-        case _  => handleEntitlementFailure(t)
       }
     }
   }
--- a/core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
+++ b/core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
@@ -214,9 +214,8 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
       Future.failed(
         RejectRequest(
           Forbidden,
-          Some(ErrorResponse(Messages.notAuthorizedtoAccessResource(resources.map(r => r.fqname).mkString(", ")), transid))
-        )
-      )
+          Some(ErrorResponse(
+            Messages.notAuthorizedtoAccessResource(resources.map(r => r.fqname).mkString(", ")), transid))))
     } else {
       Future.successful(false)
     }
@@ -233,9 +232,7 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
         RejectRequest(
           Forbidden,
           Some(ErrorResponse(
-            Messages.notAuthorizedtoAccessResource(resources.map(r => r.fqname).mkString(", ")), transid))
-        )
-      )
+            Messages.notAuthorizedtoAccessResource(resources.map(r => r.fqname).mkString(", ")), transid))))
     }
   }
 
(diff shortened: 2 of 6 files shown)


mdeuser:
pg2/2187

dubeejw:
@mdeuser, you need to run `gradle scalafmtAll` from OpenWhisk root directory then commit the changes in order to get Travis to pass.

core/controller/src/main/scala/whisk/core/controller/AuthorizedRouteDispatcher.scala
```diff
@@ -71,7 +74,21 @@ trait BasicAuthorizedRouteProvider extends Directives {
...
+                handleEntitlementFailure(
+                  RejectRequest(
+                    Forbidden,
+                    Some(ErrorResponse(Messages.notAuthorizedtoAccessResource(resource.fqname), transid))))
+              case NotFound =>
+                handleEntitlementFailure(
+                  RejectRequest(NotFound, Some(ErrorResponse(Messages.resourceDoesntExist(resource.fqname), transid))))
+              case _ => handleEntitlementFailure(t)
+            }
+        }
```
markusthoemmes: Does this disambiguate access to your own namespace vs. access to any other namespace? We shall not return `not found` if we don't even have access to that namespace because then we'd leak information from other peoples namespaces.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -56,6 +58,7 @@ protected[core] case class Resource(namespace: EntityPath,
                                     env: Option[Parameters] = None) {
   def parent = collection.path + EntityPath.PATHSEP + namespace
   def id = parent + (entity map { EntityPath.PATHSEP + _ } getOrElse (""))
+  def fqname = namespace + (entity map { EntityPath.PATHSEP + _ } getOrElse (""))
```
rabbah: namespace is not of type String - should use namespace.asString instead.

tests/src/test/scala/whisk/core/controller/test/PackageActionsApiTests.scala
```diff
@@ -591,7 +591,7 @@ class PackageActionsApiTests extends ControllerTestCommon with WhiskActionsApi {
     val auser = WhiskAuthHelpers.newIdentity()
     Get(s"/${provider.namespace}/${collection.path}/${provider.name}/${entity.name}") ~> Route.seal(routes(auser)) ~> check {
       status should be(Forbidden)
-      responseAs[ErrorResponse].error shouldBe Messages.notAuthorizedtoOperateOnResource
+      responseAs[ErrorResponse].error shouldBe Messages.notAuthorizedtoAccessResource(s"${namespace}/${provider.name}")
```
rabbah: don't need `{}` around a `namespace`: `$namespace`.

tests/src/test/scala/whisk/core/controller/test/EntitlementProviderTests.scala
```diff
@@ -534,25 +536,27 @@ class EntitlementProviderTests extends ControllerTestCommon with ScalaFutures {
...
     implicit val ep = entitlementProvider
 
+    val provider = WhiskPackage(someUser.namespace.toPath, MakeName.next(), None, publish = true)
+    val binding = WhiskPackage(guestUser.namespace.toPath, MakeName.next(), provider.bind)
+    val action = WhiskAction(binding.fullPath, MakeName.next(), jsDefault(""))
+    put(entityStore, provider)
+    put(entityStore, binding)
+    put(entityStore, action)
+
+    val resourceName = s"${binding.namespace}/${binding.name}"
```
rabbah: I suggest using `binding.fullyQualifiedName(false).asString` instead here and elsewhere a name is constructed explicitly.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -193,7 +199,14 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
...
       if (isAuthorized) Future.successful({})
-      else Future.failed(RejectRequest(Forbidden))
+      else
+        Future.failed(
+          RejectRequest(
+            Forbidden,
+            Some(
+              ErrorResponse(
+                Messages.notAuthorizedtoAccessResource(resources.map(r => r.fqname).mkString(", ")),
+                transid))))
```
rabbah: I don't think this is particularly helpful - the priv check is over a set of resources, and `isAuthorized` is a predicate that must hold over all of them. At this point, you can't disambiguate which of the resources was the one that failed the check. So the message is a blanket failure accessing all resources instead of the possibly just one that failed.

mdeuser: yep.  i wanted to only include the specific resource(s) in the set that were not accessible, but, as you mentioned, the lower level method performing the access check returned a single boolean as the all-or-nothing accessibility value.  if this is not a common use case, perhaps this message can be left as it currently is - without naming any particular resource; otherwise, additional changes will be needed to surface inaccessible resource(s).

core/controller/src/main/scala/whisk/core/controller/AuthorizedRouteDispatcher.scala
mdeuser: @markusthoemmes - not sure i completely understand the question.  if a user does not have authority to access the targeted `namespace + entity` or `namespace + package + entity`, a forbidden response will be returned - even if the namespace does not exist.  from what i've seen, the target resource name displayed will be dependent on the resource being checked.  that value might be `ns/pkg` or `ns` or `ns/entity`.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
rabbah: You could refactor the individual resource checks and resolve all the promises later where you can provide the extra information. I think that’s the sprit if the issue as it would apply to sequences.

mdeuser: @rabbah - is there a way to use the CLI to cause an auth check on a set of resources?  i think i have a refactored approach, but i'm not sure that the current automated tests exercise this code path.
/cc @dubeejw

rabbah: I'd create a unit test where it's easier to mock this instead of trying to do it through the CLI.

mdeuser:
@rabbah @dubeejw - code has been updated per comments.  let me know if there's any other needed changes

pg4/990

core/controller/src/main/scala/whisk/core/controller/AuthorizedRouteDispatcher.scala
rabbah: @markusthoemmes the entitlement should yield different rejection codes (forbidden or not found) depending on the subject and their rights. There should be tests for this.

rabbah:
I think the way sequence and rule entitlement works, we need to do more here for the case where an action does not exit. This could be a secondary PR. You can verify this by creating a test which creates a sequence with two actions `a` and `b` where only one of these exists.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -175,13 +181,34 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
...
+                }
+              }
+              .map { _.forall(identity) }
+          })
       } else Future.successful(true)
     } else if (right != REJECT) {
+      resources.map(r => inaccessibleResources += r)
       logging.info(
         this,
         s"supplied authkey for user '$subject' does not have privilege '$right' for '${resources.mkString(",")}'")
```
rabbah: i'd separate with `, ` (a space after comma) for better readability.

core/controller/src/main/scala/whisk/core/controller/AuthorizedRouteDispatcher.scala
mdeuser: _thinking out loud_  to limit the amount of information giving to an attacker, handing out a single response for either forbidden or not found seems like a good way to avoid resource name discovery.

rabbah:
In general yes but here the resources are explicitly named in the request. I’m not sure I see the point of hiding details beyond not found vs forbidden.

mdeuser:
@dubeejw - this one is ready to squash/merge.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -195,21 +200,45 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
...
-        logging.info(this, s"checking user '$subject' has privilege '$right' for '${resources.mkString(",")}'")
+        logging.info(this, s"checking user '$subject' has privilege '$right' for '${resources.mkString(", ")}'")
         checkSystemOverload(right)
           .flatMap(_ => checkUserThrottle(user, right, resources))
           .flatMap(_ => checkConcurrentUserThrottle(user, right, resources))
           .flatMap(_ => checkPrivilege(user, right, resources))
+          .flatMap(resourcePrivSet => {
+            Future
+              .sequence {
+                resourcePrivSet.map { resourcePriv =>
```
rabbah: you could filter the resource priv set on `._2` and check if the resultant list is empty, this would save the map+forall later and the construction of `inaccessibleResources` explicitly.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -232,23 +270,24 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
...
-              entitled(user.subject, right, resource)
-          }
+    Future.sequence {
+      resources.map { resource =>
+        resource.collection.implicitRights(user, defaultNamespaces, right, resource) flatMap {
+          case true => Future.successful(resource -> true)
+          case false =>
+            logging.info(this, "checking explicit grants")
+            entitled(user.subject, right, resource) flatMap {
+              case true  => Future.successful(resource -> true)
```
rabbah: This should just be `x => Future.successful(resource -> x)`.

core/controller/src/main/scala/whisk/core/controller/AuthorizedRouteDispatcher.scala
rabbah: In general yes but here the resources are explicitly named in the request. I’m not sure I see the point of hiding details beyond not found vs forbidden.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
mdeuser: i wasn't sure how to escape the Future flatmap cycle to produce a non-Future map.  i made an update close to what you've suggested

rabbah:
@dubeejw when you merge a pr that is not squashed can you sanitze the commit message so it doesn't come through like this one:

```
* Add resource name to Forbidden error message

* Comment updates and test case updates

* Fix scalafmt issue

* Reformat using scalafmt plugin

* Fix two more failing test cases

* Comment updates

* Compiles but tests are broken

* Add Set[Resources] test

* Comment updates

* Comment updates
```

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
rabbah: OK I will attempt this and can submit a patch.

dubeejw:
@rabbah, sure thing.

PR #2878, opened 2017-10-20
Streamingly read user-logs.

cbickel:
Use akka-streams and its various connectors to read a user's log file in a streaming fashion.

This will **not** contribute to memory-pressure relief as all the logs still need to be read into memory to write it to the database due to the existing schema.

Alters the existing limit calculation in that the limit is applied to the JSON structure of the log-file rather than only the log message. As a consequence, the perceived log-limit will be lower than the actually defined log-limit.


core/invoker/build.gradle
```diff
@@ -8,11 +8,16 @@ distDocker.dependsOn ':common:scala:distDocker', 'distTar'
 
 repositories {
     mavenCentral()
+    maven {
+        url "https://dl.bintray.com/akka/maven/"
```
tysonnorris: alpakka bits are in central, I don't think this is needed
https://mvnrepository.com/artifact/com.lightbend.akka/akka-stream-alpakka-file_2.11/0.14

markusthoemmes:
PG3 1307 🔵

common/scala/build.gradle
```diff
@@ -13,9 +13,9 @@ dependencies {
 
     compile 'io.spray:spray-json_2.11:1.3.3'
 
-    compile 'com.typesafe.akka:akka-actor_2.11:2.5.4'
-    compile 'com.typesafe.akka:akka-stream_2.11:2.5.4'
-    compile 'com.typesafe.akka:akka-slf4j_2.11:2.5.4'
+    compile 'com.typesafe.akka:akka-actor_2.11:2.5.6'
+    compile 'com.typesafe.akka:akka-stream_2.11:2.5.6'
```
markusthoemmes: This is useful, since the `FileIO.fromPath` source was very recently upgraded to a more performant one.

markusthoemmes:
PG4 1032 🔵

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerContainerTests.scala
```diff
@@ -630,45 +610,44 @@ class DockerContainerTests
...
-    val processedThirdLog = await(container.logs(limit = 1.MB, waitForSentinel = true))
+      awaitLogs(container.logs(limit = (secondRawLog.length - 1).bytes, waitForSentinel = false))
+    val processedThirdLog = awaitLogs(container.logs(limit = 1.MB, waitForSentinel = true))
 
     docker.rawContainerLogsInvocations should have size 3
     val (_, fromPos1) = docker.rawContainerLogsInvocations(0)
     fromPos1 shouldBe 0
     val (_, fromPos2) = docker.rawContainerLogsInvocations(1)
-    fromPos2 shouldBe firstRawLog.capacity() // second read should start behind full content of first read
+    fromPos2 shouldBe firstRawLog.length // second read should start behind full content of first read
```
sven-lange-last: If the requested limit is `firstRawLog.length - 1`, will `container.logs()` still fully read the truncated line? It should do so to make sure that the next `container.logs()` invocation won't start with the remainder from the truncated line.

common/scala/src/main/scala/whisk/core/containerpool/logging/DockerLogStore.scala
```diff
@@ -17,13 +17,37 @@
...
+protected[core] object LogLine extends DefaultJsonProtocol {
+  implicit val serdes = jsonFormat3(LogLine.apply)
+}
+
+object DockerLogStore {
+
+  /** Transforms chunked JsObjects into formatted strings */
+  val toFormattedString: Flow[ByteString, String, NotUsed] =
+    Flow[ByteString].map(_.utf8String.parseJson.convertTo[LogLine].toFormattedString)
+}
```
sven-lange-last: What's the behaviour of the `toFormattedString` flow if a truncated log line arrives? This may happen if a log limit is reached while reading.

markusthoemmes: Truncated log-lines never pass the framing stage.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -135,15 +142,14 @@ class DockerContainer(protected val id: ContainerId,
                                                       as: ActorSystem,
                                                       protected val ec: ExecutionContext,
                                                       protected val logging: Logging)
-    extends Container
-    with DockerActionLogDriver {
+    extends Container {
 
   /** The last read-position in the log file */
-  private var logFileOffset = 0L
+  private var logFileOffset = new AtomicLong(0)
```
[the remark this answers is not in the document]
markusthoemmes: It's being used by the stream, which is running in another thread. While that should be guaranteed to run sequentially, it's good measure to make variables that pass the async boundaries thread-safe.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -224,31 +230,73 @@ class DockerContainer(protected val id: ContainerId,
...
-          }
-        }
-        .andThen {
-          case Failure(e) =>
-            logging.error(this, s"Failed to obtain logs of ${id.asString}: ${e.getClass} - ${e.getMessage}")
-        }
+  def logs(limit: ByteSize, waitForSentinel: Boolean)(implicit transid: TransactionId): Source[ByteString, Any] = {
+    val source = docker
+      .rawContainerLogs(id, logFileOffset.get(), if (waitForSentinel) Some(filePollInterval) else None)
+      .via(Framing.delimiter(delimiter, Int.MaxValue))
```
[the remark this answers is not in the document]
markusthoemmes: Didn't use JSON Framing because it's a tad bit slower and we don't gain anything by it (failure scenarios stay the same). I might want to revisit the "tad bit slower" part, I'm quite sure it doesn't matter anymore.

Agree on additional tests for premature file end!

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerContainerTests.scala
```diff
@@ -429,34 +446,35 @@ class DockerContainerTests
...
-        Future.successful(readResults.dequeue())
+        Source.single(rawLog)
       }
     }
     implicit val runc = stub[RuncApi]
 
     val container = dockerContainer(id = containerId)()
     // Read with tight limit to verify that no truncation occurs
-    val processedLogs = await(container.logs(limit = expectedLogEntry.log.sizeInBytes, waitForSentinel = true))
+    val processedLogs = awaitLogs(container.logs(limit = (rawLog.length + 1).bytes, waitForSentinel = true))
```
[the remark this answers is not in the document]
markusthoemmes: Not needed, reverted.

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerContainerTests.scala
markusthoemmes: This test does make sure it does what you suggest, right? It drops the line but still starts **after** the first line for the next read?

common/scala/src/main/scala/whisk/core/containerpool/logging/DockerLogStore.scala
rabbah: note to self: is this tested or can be tested; should it be?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
```diff
@@ -135,62 +139,32 @@ class DockerClientWithFileAccess(
...
-      } finally {
-        if (fis != null) fis.close()
+  override def rawContainerLogs(containerId: ContainerId,
+                                fromPos: Long,
+                                pollInterval: Option[FiniteDuration]): AkkaSource[ByteString, Any] =
+    try {
+      // If there is no waiting interval, we can end the stream early by reading just what is there from file.
+      pollInterval match {
+        case Some(interval) => FileTailSource(containerLogFile(containerId).toPath, 8192, fromPos, interval)
+        case None           => FileIO.fromPath(containerLogFile(containerId).toPath, 8192, fromPos)
```
rabbah: lift 8K chunk size to a field const?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -17,24 +17,32 @@
...
 import whisk.core.entity.size._
+import akka.stream.scaladsl.{Framing, Source}
+import akka.stream.stage._
+import akka.util.ByteString
+import spray.json._
+import whisk.core.containerpool.logging.LogLine
+import whisk.http.Messages
 
 object DockerContainer {
+  // The action proxies inserts this line in the logs at the end of each activation for stdout/stderr
```
rabbah: should qualify the comment that this may not apply to blackbox/docker actions.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -215,40 +222,99 @@ class DockerContainer(protected val id: ContainerId,
...
+        case _: OccurrencesNotFoundException | _: FramingException =>
+          // Stream has already ended, since the log-source closed the stream, which caused a framing exception.
+          // This will be the last element of the stream.
+          ByteString(LogLine(Instant.now.toString, "stderr", Messages.logFailure).toJson.compactPrint)
+      }
+      .takeWithin(waitForLogs)
+  }
+
+  /* Delimiter used to split logs */
+  private val delimiter = ByteString("\n")
```
rabbah: this delimiter applies not to the user logs, but the JSON docker-log-driver file, correct?

markusthoemmes: correct.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -224,31 +230,73 @@ class DockerContainer(protected val id: ContainerId,
...
-        }
-        .andThen {
-          case Failure(e) =>
-            logging.error(this, s"Failed to obtain logs of ${id.asString}: ${e.getClass} - ${e.getMessage}")
-        }
+  def logs(limit: ByteSize, waitForSentinel: Boolean)(implicit transid: TransactionId): Source[ByteString, Any] = {
+    val source = docker
+      .rawContainerLogs(id, logFileOffset.get(), if (waitForSentinel) Some(filePollInterval) else None)
+      .via(Framing.delimiter(delimiter, Int.MaxValue))
+      .limitWeighted(limit.toBytes) { obj =>
```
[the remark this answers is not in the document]
rabbah: thanks for making this change - the Framing does allow for truncated results, which might be useful; since the format is well know, we could still parse it. Makes me also wonder if it may be generally better to eschew the more JSON parser 

```
    Flow[ByteString].map(_.utf8String.parseJson.convertTo[LogLine].toFormattedString)
```

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -369,10 +369,16 @@ class ContainerProxy(factory: (TransactionId, String, ImageName, Boolean, ByteSi
...
+        collectLogs(tid, container, job.action)
+          .map { logs =>
+            tid.finished(this, start)
+            activation.withLogs(logs)
+          }
+          .recover {
+            case _ =>
+              tid.failed(this, start)
+              activation.withLogs(ActivationLogs(Vector(Messages.logFailure)))
+          }
```
sven-lange-last: Thanks for adding the recovery path. If we end up in recovery: did we emit the error to the invoker log anywhere after the error occurred? Otherwise we have no change to understand what happened and don't even notice that activation logs cannot be processed.

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -375,7 +385,64 @@ class ContainerProxyTests
...
+    expectMsg(ContainerRemoved) // The message is sent as soon as the container decides to destroy itself
+    expectMsg(Transition(machine, Running, Removing))
+
+    awaitAssert {
+      factory.calls should have size 1
+      container.initializeCount shouldBe 1
+      container.runCount shouldBe 1
+      collector.calls should have size 1
+      container.destroyCount shouldBe 1
+      acker.calls(0)._2.response shouldBe ActivationResponse.success()
```
sven-lange-last: `acker.calls should have size 1`

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -375,7 +385,64 @@ class ContainerProxyTests
...
```
sven-lange-last: `acker.calls should have size 1`

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -370,24 +371,25 @@ class ContainerProxy(factory: (TransactionId, String, ImageName, Boolean, ByteSi
...
               tid.failed(this, start)
-              activation.withLogs(ActivationLogs(Vector(Messages.logFailure)))
+              val withLogs = t match {
+                case LogCollectingException(logs) => activation.withLogs(logs)
+                case _                            => activation.withLogs(ActivationLogs(Vector(Messages.logFailure)))
+              }
+              storeActivation(tid, withLogs)
+
+              Future.failed(ActivationLogReadingError(withLogs))
+          }
```
sven-lange-last: It would be helpful to have a comment here that the `recoverWith` path with always end up in a failed Future so that the following `flatMap` will be skipped.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -236,7 +236,7 @@ class DockerContainer(protected val id: ContainerId,
   def logs(limit: ByteSize, waitForSentinel: Boolean)(implicit transid: TransactionId): Source[ByteString, Any] = {
     docker
       .rawContainerLogs(id, logFileOffset.get(), if (waitForSentinel) Some(filePollInterval) else None)
-      .via(Framing.delimiter(delimiter, Int.MaxValue))
+      .via(Framing.delimiter(delimiter, limit.toBytes.toInt))
```
sven-lange-last: Should add a comment here that we can obtain a FramingException for two reasons:
1. the Source provides an incomplete line so that the delimiter cannot be found.
2. the Source provides more data than `limit.toBytes.toInt` which is effectively a truncation due to limit exceeded.

As a result we cannot decide in later stages whether we should issue a `log processing problem` or a `log truncated` message.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -371,18 +369,28 @@ class ContainerProxy(factory: (TransactionId, String, ImageName, Boolean, ByteSi
...
+        val start = tid.started(this, LoggingMarkers.INVOKER_COLLECT_LOGS)
+        collectLogs(tid, container, job.action)
+          .recoverWith {
+            case t =>
+              tid.failed(this, start, s"reading logs failed: $t")
+              val withLogs = t match {
+                case LogCollectingException(logs) => activation.withLogs(logs)
+                case _                            => activation.withLogs(ActivationLogs(Vector(Messages.logFailure)))
+              }
+              storeActivation(tid, withLogs)
```
rabbah: why are there two storeActivation calls? here and for the flatmap below.

rabbah: this flow is confusing - does it make sense to consolidate the `storeActivation` calls to just a single place?

common/scala/src/main/scala/whisk/core/containerpool/logging/DockerLogStore.scala
```diff
@@ -43,7 +69,22 @@ class DockerLogStore(system: ActorSystem) extends LogStore {
...
+
+    container
+      .logs(action.limits.logs.asMegaBytes, action.exec.sentinelledLogs)(transid)
+      .via(DockerLogStore.toFormattedString)
+      .runWith(Sink.seq)
+      .flatMap { seq =>
+        val errored = seq.lastOption.exists(last => possibleErrors.exists(last.contains))
+        val logs = ActivationLogs(seq.toVector)
+        if (!errored) {
+          Future.successful(logs)
```
rabbah: nit: the future here seems superfluous; this could have been an `either` type.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -371,18 +369,28 @@ class ContainerProxy(factory: (TransactionId, String, ImageName, Boolean, ByteSi
...
-      .andThen {
-        case Success(activation) => storeActivation(tid, activation)
-      }
-      .flatMap { activation =>
-        // Fail the future iff the activation was unsuccessful to facilitate
-        // better cleanup logic.
-        if (activation.response.isSuccess) Future.successful(activation)
-        else Future.failed(ActivationUnsuccessfulError(activation))
+        val start = tid.started(this, LoggingMarkers.INVOKER_COLLECT_LOGS)
+        collectLogs(tid, container, job.action)
```
rabbah: my nit about the superfluous future is related to this future continuation and recovery path which could be come slightly cleaner.

rabbah:
As subsequent improvements, as already noted: consider eschewing the JSON parsing with a bespoke parser because the format is quite trivial. Second, for the corner case where the last log line is truncated, consider amending it to the activation logs so that it is not lost. Should we open issues/enhancement requests if you concur?

markusthoemmes:
@rabbah as discussed in person, I don't see much value in adding a truncated last-line (note that we apply the log-limits on the entire JSON structure now, so the user has no way of knowing the exact translation of her log-lines vs. her log limits anyway).

Eschewing the json-parser we can do if necessary. While the format is simple, it ties us even closer to its implementation details (like order of fields).

rabbah:
The docker log driver will flush the bytes comprising the log line but does not necessarily flush a fully formed json object - we've observed this in production for example where eventually with retries the parsing will succeed, or times out. This could be exacerbated by logging large volumes to a single log line (since we don't control when docker does it buffer flush, _or_, how the action does its logging). For an end user, we should strive to pass on as many of the logs as possible, and I don't see the reason not to do that when we are forced to truncate a single line.

rabbah:
> Eschewing the json-parser we can do if necessary. While the format is simple, it ties us even closer to its implementation details (like order of fields).

For docker versions, which we change infrequently, I posit these formats change infrequently enough that if the performance gains are significant, it becomes worthwhile.

common/scala/src/main/scala/whisk/core/containerpool/logging/DockerLogStore.scala
markusthoemmes: Note the function is already returning a Future, this just makes it a bit more granular.

rabbah:
PG3/1935 🔵

rabbah:
Merged as https://github.com/apache/incubator-openwhisk/commit/83982f6687252e793c3e0862e8a23a656f76fc2c.

PR #2887, opened 2017-10-23
Check for correct number of minimal arguments on action create

dubeejw:
Currently the argument length check passes successfully when only one argument is passed to action create. Instead of failing the argument length check, a failure occurs in `parseAction` when only one argument is specified.

### DIFF
--- a/tools/cli/go-whisk-cli/commands/action.go
+++ b/tools/cli/go-whisk-cli/commands/action.go
@@ -75,7 +75,7 @@ var actionCreateCmd = &cobra.Command{
 
         if whiskErr := CheckArgs(
             args,
-            1,
+            2,
             2,
             "Action create",
             wski18n.T("An action name and code artifact are required.")); whiskErr != nil {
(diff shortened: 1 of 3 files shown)


dubeejw:
Looks like allowing one argument was on purpose when creating a docker action, but I don't think what is there is implemented correctly.
```
wsk -i action create dockerContainer --docker bogusihxfy6q5ssatqazy
```

dubeejw:
@mdeuser, looks like there is more at play here. Will make updates and see what you think.

dubeejw:
Currently when providing the sequence and docker flags during action create or update, an error message is displayed that could be more informative.
```
$ wsk action create dockerContainer3 --docker bogusihxfy6q5ssatqazy --sequence
error: Invalid argument(s). An action name and code artifact are required.
```

rabbah:
@csantanapr @jessealva something to pick up.

rabbah:
Closing as invalid now. This has to be opened against CLI repo.

PR #2888, opened 2017-10-24
Adapt trigger throttle for multiple controllers

vvraskin:
Divide per minute throttles through the number of controllers. 

* Required for https://github.com/apache/incubator-openwhisk/pull/2665
* PG2 2202 is green.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/entitlement/RateThrottler.scala
+++ b/core/controller/src/main/scala/whisk/core/entitlement/RateThrottler.scala
@@ -18,7 +18,6 @@
 package whisk.core.entitlement
 
 import scala.collection.concurrent.TrieMap
-
 import whisk.common.Logging
 import whisk.common.TransactionId
 import whisk.core.entity.Identity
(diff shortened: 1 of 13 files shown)


tests/src/test/scala/ha/ShootComponentsTests.scala
```diff
@@ -94,9 +94,11 @@ class ShootComponentsTests extends FlatSpec with Matchers with WskTestHelpers wi
 
       println(s"Done rerquests with responses: invoke: ${invokeExit.futureValue} and get: ${getExit.futureValue}")
 
-      // Do at most one action invocation per second to avoid getting 429s. (60 req/min - limit)
-      val wait = 1000 - (Instant.now.toEpochMilli - start.toEpochMilli)
-      Thread.sleep(if (wait < 0) 0L else if (wait > 1000) 1000L else wait)
+      // Do at most one action invocation per 2 seconds to avoid getting 429s. (30 req/min - half of the limit for
+      // when one controller is down)
+      val period = 2000
```
cbickel: Can we set this parameter dynamically depending on the number of controllers in this environment?

Reason is: If we add another controller, we are able to do more requests until we get throttled.

core/controller/src/main/scala/whisk/core/entitlement/RateThrottler.scala
```diff
@@ -43,14 +44,17 @@ class RateThrottler(description: String, defaultMaxPerMinute: Int, overrideMaxPe
...
    *
    * @param user the identity to check
    * @return true iff subject namespace is below allowed limit
    */
   def check(user: Identity)(implicit transid: TransactionId): RateLimit = {
     val uuid = user.uuid // this is namespace identifier
     val throttle = rateMap.getOrElseUpdate(uuid, new RateInfo)
-    val limit = overrideMaxPerMinute(user).getOrElse(defaultMaxPerMinute)
+    val limit =
+      Math.ceil((overrideMaxPerMinute(user).getOrElse(defaultMaxPerMinute).toDouble / diviser.toDouble) * 1.2).toInt
```
markusthoemmes: This computation needs to be done only iff controllerHA is active. I propose to adjust the values passed into the RateThrottlers based on the controllerHA value one level above instead of passing everything through here. Should leave the RateThrottler untouched.

PR #2891, opened 2017-10-24
Vagrant base image update to latest Ubuntu Xenial LTS

bwmcadams:
- Update Vagrant image to latest Ubuntu LTS release; adjust Docker tools accordingly
    - The version of Ubuntu being used was 14.04 (Trusty) which is EOL and 3+ years old
       - Updated to 16.04, Xenial, which is most current LTS
       - Moved from Ubuntu images to Chef Bento – Ubuntu's images after 14.04 started
         breaking Vagrant convention by not creating a "vagrant" user. Chef Bento images
         keep with Vagrant convention and obviate the need for extra "create user" scripts
    - Updated Ubuntu Tools Docker image accordingly
- Updated to retrieve same version of Scala as project Gradle
    - was 2.11.6, updated to 2.11.11

### DIFF
--- a/tools/ubuntu-setup/docker.sh
+++ b/tools/ubuntu-setup/docker.sh
@@ -20,15 +20,9 @@ sudo apt-get -y install linux-image-extra-$(uname -r) linux-image-extra-virtual
 # DOCKER
 sudo apt-get install -y --force-yes docker-ce
 
-# enable (security - use 127.0.0.1)
-#sudo -E bash -c 'echo '\''DOCKER_OPTS="-H tcp://0.0.0.0:4243 -H unix:///var/run/docker.sock --storage-driver=aufs"'\'' >> /etc/default/docker'
 sudo gpasswd -a `whoami` docker
 
-# Set DOCKER_HOST as an environment variable
-#sudo -E bash -c 'echo '\''export DOCKER_HOST="tcp://0.0.0.0:4243"'\'' >> /etc/bash.bashrc'
-#source /etc/bash.bashrc
-
-#sudo service docker restart
+sudo service docker restart
 
 # do not run this command without a vagrant reload during provisioning
 # it gives an error that docker is not up (which the reload fixes).
--- a/tools/ubuntu-setup/docker.sh
+++ b/tools/ubuntu-setup/docker.sh
@@ -3,28 +3,32 @@ set -e
 set -x
 
 sudo apt-get -y install apt-transport-https ca-certificates
-sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 36A1D7869245C8950F966E92D8576A8BA88D21E9
-sudo sh -c "echo deb https://apt.dockerproject.org/repo ubuntu-trusty main  > /etc/apt/sources.list.d/docker.list"
+
+# Docker GPG Key
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+
+sudo add-apt-repository \
+   "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
+   $(lsb_release -cs) \
+   stable"
+
 sudo apt-get -y update -qq
 
-sudo apt-get purge lxc-docker
-sudo apt-cache policy docker-engine
-
 # AUFS
-sudo apt-get -y install linux-image-extra-$(uname -r)
+sudo apt-get -y install linux-image-extra-$(uname -r) linux-image-extra-virtual
 
 # DOCKER
-sudo apt-get install -y --force-yes docker-engine=1.12.0-0~trusty
+sudo apt-get install -y --force-yes docker-ce
 
 # enable (security - use 127.0.0.1)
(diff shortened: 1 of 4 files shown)


bbrowning:
I was getting the same error as @akrabat in his gist at https://gist.github.com/akrabat/bc0bcbfd78c4f78686bac1fbc41e3054 and unable to start the Vagrant image locally. The changes in this PR fixed that error and the image now provisions successfully.

akrabat:
As @bbrowning has noted, the current Vagrant box does not successfully provision.

I've tested this PR and it works.

csantanapr:
Can we make a new set of instructions for Xenial, maybe under `tools/setup/ubuntu-setup/xenial/*` ?
We currently doesn't support Xenial but at the same time don't really care if docker works

I'm OK if the Vagrant scenario uses Xenial, I'm just worried about people using the current setup in ubuntu 14 and using the current scripts under `tools/ubuntu-setup/docker.sh`
then with this changes they will brake immediately. 

Also scala is currently configured for `2.8.11` not `2.11.11` as stated here: https://github.com/apache/incubator-openwhisk/blob/master/settings.gradle#L25

tools/ubuntu-setup/docker.sh
```diff
@@ -3,28 +3,32 @@ set -e
...
+
+sudo apt-get -y update -qq
 
 # AUFS
-sudo apt-get -y install linux-image-extra-$(uname -r)
+sudo apt-get -y install linux-image-extra-$(uname -r) linux-image-extra-virtual
 
 # DOCKER
-sudo apt-get install -y --force-yes docker-engine=1.12.0-0~trusty
+sudo apt-get install -y --force-yes docker-ce
```
csantanapr: I think we want to install docker 1.12 if possible since it's the version we support

tools/ubuntu-setup/docker.sh
```diff
@@ -3,28 +3,32 @@ set -e
...
 
 # enable (security - use 127.0.0.1)
-sudo -E bash -c 'echo '\''DOCKER_OPTS="-H tcp://0.0.0.0:4243 -H unix:///var/run/docker.sock --storage-driver=aufs"'\'' >> /etc/default/docker'
+#sudo -E bash -c 'echo '\''DOCKER_OPTS="-H tcp://0.0.0.0:4243 -H unix:///var/run/docker.sock --storage-driver=aufs"'\'' >> /etc/default/docker'
```
csantanapr: leaving commented code is kind ugly either remove it or leave it.

tools/ubuntu-setup/scala.sh
```diff
@@ -2,7 +2,8 @@
 set -e
 set -x
 
-wget www.scala-lang.org/files/archive/scala-2.11.6.deb -O /tmp/scala-2.11.6.deb
-sudo dpkg -i /tmp/scala-2.11.6.deb
+#wget www.scala-lang.org/files/archive/scala-2.11.6.deb -O /tmp/scala-2.11.6.deb
+wget https://downloads.lightbend.com/scala/2.11.11/scala-2.11.11.deb -O /tmp/scala-2.11.11.deb
```
csantanapr: Use `2.11.8`

markusthoemmes: Why? We're effectively on 2.11.11.

@bwmcadams Can you try to throw out Scala compiler installation? Gradle should handle that for us anyway.

tools/ubuntu-setup/docker.sh
bwmcadams: good catch; fixing that.

tools/ubuntu-setup/scala.sh
bwmcadams: Yup. Note this includes the REPL as well, but I don't think people should be doing REPL debugging on their host machine(s) anyway.

tools/ubuntu-setup/docker.sh
bwmcadams: I actually suspect I shouldn't have commented these out; testing with them reenabled otherwise will remove.

bwmcadams:
@csantanapr Ah, for some reason I thought I saw 2.11.11 used in a bunch of places. We *should* update that to latest 2.11.x series, but that seems like a separate set of PRs to me.

bwmcadams:
@csantanapr I am fine with splitting things off into Xenial subfolder, but I'd suggest if we are going to do that we also create a sep. one for 14. If we leave a "default" folder without clear versioning info it may lead to more confusion. The current folder would become a parent with just a short README explaining the subfolders/versioning.

Thoughts?

csantanapr:
bwmcadams 
yes I like the splitting of the folders, main readme explaining the differences.
```
tools/ubuntu-setup/v14/
tools/ubuntu-setup/v16/
tools/ubuntu-setup/README.md
```

csantanapr:
@markusthoemmes 
>Why? We're effectively on 2.11.11.

Then why do we have `2.11.8` in `setttings.gradle`?
Which version are we suppose to be using?
https://github.com/apache/incubator-openwhisk/blob/master/settings.gradle#L24-L26

csantanapr:
Closing as PR #3133 going to take care of this by detecting ubuntu 16 or later and installing docker-ce, the rest stays the same.

Once PR #3133 is merged we can switch the Vagran VM to use ubuntu 16

PR #2895, opened 2017-10-25
Update the test cases about rules with REST implementation

houshengbo:


tests/src/test/scala/system/basic/WskRestRuleTests.scala
```diff
@@ -0,0 +1,61 @@
...
+    for (rule <- rules) {
+      breakable {
+        if (RestResult.getField(rule, "name") == ruleName) {
+          found = true
+          break
+        }
+      }
+    }
+    found shouldBe true
+  }
```
markusthoemmes: I think that helper function is unnecessary. Using `exists` it can be one-lined:

```scala
rules.exists(rule => RestResult.getField(rule, "name") == ruleName)
```

houshengbo: Done.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -631,14 +632,14 @@ class WskRestActivation extends RunWskRestCmd with HasActivationRest with WaitFo
     var paramMap = Map("skip" -> "0", "docs" -> docs.toString) ++ {
       limit map { l =>
         Map("limit" -> l.toString)
-      } getOrElse Map("limit" -> "30")
+      } getOrElse Map("limit" -> "0")
```
dubeejw: `limit` should not be part of the payload if it is not set.

houshengbo: isn't 0 the default value, since I see cli always put 0 fr all requests?

dubeejw: I believe the CLI defaults to a limit of 30. Looks like the backend will default to 30 if 0 is provided for the limit query parameter. Looks like `skip` defaults to 0 in the CLI, but we don't have any tests for the skip functionality.

`wsk activation list` send this request: `https://apihost/api/v1/namespaces/_/activations?limit=30&skip=0`

houshengbo:
PG1 2209.
There are failures, but they are not related to the changes.

Issue #2872, opened 2017-10-18
invoker command line support for dynamicId assignment


dgrove-oss:
As discussed in the review of https://github.com/apache/incubator-openwhisk/pull/2689, for Mesos it is not feasible to use environment variables to communicate the unique invoker names.  For Mesos, we need to also support command-line flags style of specifying arguments.  For example:
  --name <some_name> --id <proposed_id>
For Kubernetes, we really would prefer environment variables.

So, the plan is to support both options.  Add flags-style command line argument processing to the invoker and read the environment if explicit command lines are not provided.

dgrove-oss:
Please label as enhancement and assign to me.  I plan to work on this next week.

dgrove-oss:
Implemented in PR #2939

dgrove-oss:
PR merged

PR #2902, opened 2017-10-25
make INVOKER_NAME optional

ddragosd:
This slipped in with the Redis changes and I don't think it's needed b/c the invoker already starts with an `id`. 

if `id arg` is provided, use it, and do NOT require INVOKER_NAME or REDIS_HOST configs; if `id arg` is NOT provided, then DO require INVOKER_NAME AND REDIS_HOST configs

Relates to: #2872 and https://github.com/apache/incubator-openwhisk-devtools/pull/62

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
@@ -60,7 +60,7 @@ object Invoker {
       invokerContainerDns -> "",
       invokerContainerNetwork -> null,
       invokerUseRunc -> "true") ++
-      Map(invokerName -> null)
+      Map(invokerName -> "")
 
   def main(args: Array[String]): Unit = {
     implicit val ec = ExecutionContextFactory.makeCachedThreadPoolExecutionContext()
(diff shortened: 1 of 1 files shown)


core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -60,7 +60,7 @@ object Invoker {
       invokerContainerDns -> "",
       invokerContainerNetwork -> null,
       invokerUseRunc -> "true") ++
-      Map(invokerName -> null)
+      Map(invokerName -> "")
```
tysonnorris: invokerName should be validated as non-empty below, same as redisHost (it will only be validated when the id arg is not present)

ddragosd: good point @tysonnorris. 
done

tysonnorris:
LGTM

ddragosd:
@rabbah , @markusthoemmes , @dgrove-oss I'd like to merge this to fix the build on `devtools` repo, unless you have any feedback.

PR #2906, opened 2017-10-26
Bump go cli to 1.8.4

mdeuser:
- Fix for security issue https://github.com/golang/go/issues/22125


mdeuser:
pg2/2219

markusthoemmes:
Seems per https://hub.docker.com/_/golang/ that 1.8 == 1.8.4? We should already be picking up that patch version I think.

mdeuser:
ah.. in that case, i think we should pin the version of go to a specific version.

PR #2908, opened 2017-10-26
Optionally fetch code when getting an action

dubeejw:
Provides optional query parameter for getting an action that specifies whether or not action code should be fetched and returned. When the query parameter is present, WhiskActionMetadata will be used to fetch the action. Otherwise, WhiskAction will be used to retrieve the action.

Fetch the action with the code as is today:
`https://192.168.99.100/api/v1/namespaces/_/actions/actionName`
or
`https://192.168.99.100/api/v1/namespaces/_/actions/actionName?code=true`

Fetch action metadata without code:
`https://192.168.99.100/api/v1/namespaces/_/actions/actionName?code=false`

Depends on https://github.com/apache/incubator-openwhisk/pull/2730.

### DIFF
--- a/tests/src/test/scala/whisk/core/controller/test/ActionsApiTests.scala
+++ b/tests/src/test/scala/whisk/core/controller/test/ActionsApiTests.scala
@@ -167,11 +167,13 @@ class ActionsApiTests extends ControllerTestCommon with WhiskActionsApi {
       status should be(OK)
       val response = responseAs[JsObject]
       response.fields("exec").asJsObject.fields should not(contain key "code")
+      responseAs[WhiskActionMetaData] shouldBe a[WhiskActionMetaData]
     }
     Get(s"$collectionPath/${action.name}?code=true") ~> Route.seal(routes(creds)) ~> check {
       status should be(OK)
       val response = responseAs[JsObject]
       response.fields("exec").asJsObject.fields("code") should be("??".toJson)
+      responseAs[WhiskAction] shouldBe a[WhiskAction]
     }
   }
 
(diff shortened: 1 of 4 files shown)


dubeejw:
@csantanapr, is this what you had in mind?

dubeejw:
@csantanapr, I believe @rabbah has concerns about this being a breaking API change.

rabbah:
Right. This will break all client integrations. We could add a no code flag for v1 but I think the schema and api change should be done on v2.

tests/src/test/scala/whisk/core/controller/test/ActionsApiTests.scala
```diff
@@ -159,6 +159,22 @@ class ActionsApiTests extends ControllerTestCommon with WhiskActionsApi {
...
+    put(entityStore, action)
+    Get(s"$collectionPath/${action.name}?code=false") ~> Route.seal(routes(creds)) ~> check {
+      status should be(OK)
+      val response = responseAs[JsObject]
+      response.fields("exec").asJsObject.fields should not(contain key "code")
+    }
+    Get(s"$collectionPath/${action.name}?code=true") ~> Route.seal(routes(creds)) ~> check {
+      status should be(OK)
+      val response = responseAs[JsObject]
+      response.fields("exec").asJsObject.fields("code") should be("??".toJson)
```
rabbah: you could tighten both these tests with additional checks that the JSON response can be demarshaled to a `WhiskAction` or `WhiskActionMetadata` respectively and check for equality as well.

rabbah:
Looks good - can you update the swagger for the new parameter?

dubeejw:
PG3 1483 🔵

PR #2915, opened 2017-10-31
AWS deployment

Shillaker:
Adds automated AWS provisioning of distributed OpenWhisk (similar to existing Openstack support).

I've not changed anything related to the existing distributed deployment work for fear of breaking it. To achieve this I've had to do a couple of work-arounds that could be avoided with a couple more tweaks:

1. Implement the `TODO` open in `environments/distributed/group_vars/all` to move instance properties to `group_vars`. I agree this would be useful and had to add a templating step in this PR that puts them in place
2. Reuse the new `disks` role I've created instead of the inline code in `provision_env_dist.yml`. There is another open `TODO` in that file mentioning that this would be useful.

Other than that I hope it should be pretty straightforward. I found the following config to be the minimum I could get away with for testing (set this in `ansible/environments/distributed/group_vars/all`):

```
instances:
  - name: registry
    num_instances: 1
    flavor: t2.medium
    volume:
      name: /dev/sda1
      size: 100
      fsmount: /

  - name: edge
    num_instances: 1
    flavor: t2.micro

  - name: controllers
    num_instances: 1
    flavor: t2.medium

  - name: kafka
    num_instances: 1
    flavor: t2.medium

  - name: invokers
    num_instances: 2
    flavor: t2.micro
    volume:
      name: /dev/sda1
      size: 50
      fsmount: /

  - name: db
    num_instances: 1
    flavor: t2.micro
    volume:
      name: /dev/xvdb
      size: 10
      fstype: ext4
      fsmount: /mnt/db
```

### DIFF
--- a/ansible/README_DISTRIBUTED.md
+++ b/ansible/README_DISTRIBUTED.md
@@ -65,6 +65,7 @@ cd ansible
 ansible-playbook -i environments/<environment> couchdb.yml
 ansible-playbook -i environments/<environment> initdb.yml
 ansible-playbook -i environments/<environment> wipe.yml
+ansible-playbook -i environments/<environment> apigateway.yml
 ansible-playbook -i environments/<environment> openwhisk.yml
 ansible-playbook -i environments/<environment> postdeploy.yml
 ```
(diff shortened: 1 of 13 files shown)


rabbah:
Thank you for this contribution - the distributed deployment is currently untested and while we'd love to have this AWS deployment scripts as well, a general concern is how do we maintain these in the long run and avoid the kinds of issues we had with the openstack scripts which rotted because of a lack of CI/CD. Do you have thoughts on how to test this? This may be a good disucssion for the Apache dev list where perhaps getting infra to help is a possibility. Do you mind initiating a discussion on the dev list?

Shillaker:
Hey, sorry for the late reply. I see your point, it's always tricky testing this sort of thing and, in reality, it'll only stay totally up-to-date if people actually use it regularly. I'll drop a message on the dev list now and see what the thoughts are. I've never personally tested Ansible scripts. I've tested lots of AWS integrations using a mocked boto library called moto which works well so perhaps there's a way. Will let you know

csantanapr:
there is no current way of testing this. I direction is to move away of using ansible VMs to deploy and instead use more native container/docker way by using kubernetes/docker-swarm type of deployments. 
Take a look at docker-compose in the devtools/experimental repo https://github.com/apache/incubator-openwhisk-devtools/tree/master/docker-compose

PR #2916, opened 2017-10-31
use ZooKeeper as persistent store for invokerId assignment

dgrove-oss:
Use ZooKeeper instead of Redis to persist mapping from invoker
names to invoker ids and the counter for the next available id.

Basic algorithm for dynamic invoker id assignment is unchanged, just
using a different persistent store whose durability is a better match
for this purpose (we want the invoker id assignment information to be
as durable as the kafka topics it is being used to assign).


dgrove-oss:
pg1 2216 🏃

rabbah:
@dgrove-oss do we need a companion pr against kube* repo?

dgrove-oss:
@rabbah here is the companion for the kube repo https://github.com/apache/incubator-openwhisk-deploy-kube/pull/78

The kube deployment is not using dynamic id assignment by default yet, so there isn't a real dependency.  But good point that we should update kube to match when this is merged.

core/invoker/build.gradle
```diff
@@ -13,6 +13,7 @@ repositories {
 dependencies {
     compile "org.scala-lang:scala-library:${gradle.scala.version}"
     compile project(':common:scala')
+    compile 'org.apache.curator:curator-recipes:2.12.0'
```
markusthoemmes: Any reason not to use the latest curator 4.0?

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -109,37 +111,47 @@ object Invoker {
...
-            // Convert from 1-based to 0-based invokerIds by subtracting 1 from incr's result
-            val newId = redisClient
-              .incr("controller:registrar:nextInvokerId")
-              .map { id =>
-                id.toInt - 1
-              }
-              .getOrElse {
-                abort("Failed to increment invokerId")
+        logger.info(this, s"invokerReg: creating zkClient to ${config.zookeeperHost}")
+        val retryPolicy = new RetryUntilElapsed(5000, 500)
```
markusthoemmes: Naming the parameters of `RetryUntilElapsed` would be beneficial for self-documenation, like:

```scala
val retryPolicy = new RetryUntilElapsed(maxElapsedTimeMs = 5000, sleepMsBetweenRetries = 500)
```

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -109,37 +111,47 @@ object Invoker {
...
+        val zkClient = CuratorFrameworkFactory.newClient(config.zookeeperHost, retryPolicy)
+        zkClient.start()
+        zkClient.blockUntilConnected();
+        logger.info(this, "invokerReg: connected to zookeeper")
+        val myIdPath = "/invokers/idAssignment/mapping/" + invokerName
+        val assignedId =
+          if (zkClient.checkExists().forPath(myIdPath) == null) {
```
markusthoemmes: Just to keep `null` out of the codebase: The "scala way" would be to wrap in an `Option` and match like:

```scala
Option(zkClient.checkExists().forPath(myIdPath)) match {
  case None => // == null branch
  case Some(_) => // already exists branch
}
```

It doesn't give us much in this specific case, but keeping `null` out of the codebase imho is the right thing to do.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -109,37 +111,47 @@ object Invoker {
...
+            // path doesn't exist ==> no previous mapping for this invoker
+            logger.info(this, s"invokerReg: no prior assignment of id for invoker $invokerName")
+            val idCounter = new SharedCount(zkClient, "/invokers/idAssignment/counter", 0)
+            idCounter.start()
+            def assignId(): Int = {
+              val current = idCounter.getVersionedValue()
+              return if (idCounter.trySetCount(current, current.getValue() + 1)) {
```
markusthoemmes: Extranous `return`. Note that `return` is actually dangerous in Scala: https://tpolecat.github.io/2014/05/09/return.html

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -109,37 +111,47 @@ object Invoker {
...
+                current.getValue()
```
markusthoemmes: Is `getValue()` a local operation or does it actually ask Zookeeper? If the latter, I think we need to cache this before the `if` to avoid a race condition.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -109,37 +111,47 @@ object Invoker {
...
+              } else {
+                assignId()
               }
-            redisClient.hset("controller:registar:idAssignments", invokerName, newId)
+            }
+            val newId = assignId()
+            idCounter.close()
+            zkClient.create().creatingParentContainersIfNeeded().forPath(myIdPath, newId.toString().getBytes())
```
markusthoemmes: I only briefly checked, but `scala.math.BigInt` supports byte-conversions, maybe that's a little bit more speaking?

```scala
zkClient.create().creatingParentContainersIfNeeded().forPath(myIdPath, BigInt(newId).toByteArray)
...
val oldId = BigInt(rawOldId).intValue
```

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
dgrove-oss: Purely local.  Insided the call to getVersionedValue() is where the ZooKeeper interaction happens.  The result of that call is an instance of VersionedValue, which is just a POJO with two final fields. getValue() is just a get accessor for the final field value

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
dgrove-oss: thanks, learn something every day 😁 Will fix. 

I'm familiar with non-local returns from Smalltalk & derivatives, but Scala is the first language I've seen that takes over the return keyword for it....most "sane" languages use a different syntax so it is more clear what you are doing.

core/invoker/build.gradle
dgrove-oss: We deploy zookeeper 3.4.  In theory, curator 4.0 supports both 3.4 and 3.5, but it didn't work for me (runtime errors due to unsupported operations).  I tried the gradle exclusion incantation from https://curator.apache.org/zk-compatibility.html to try to get 3.4 compatability, but it didn't help.

So, I stuck with 2.12.0 which is the most recent release of the zookeeper 3.4 compatible line of curator.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
dgrove-oss: added a comment;  trying to use keyword parameters failed to compile (Java API; I'm guessing the class file didn't have enough information to allow use of named parameters).

dgrove-oss:
Thanks @markusthoemmes; pushed changes to address your comments.

dgrove-oss:
squashed to single commit.  PG1 2229.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -22,10 +22,12 @@ import scala.concurrent.duration._
 import scala.concurrent.Future
 import scala.util.Failure
 
-import com.redis.RedisClient
```
markusthoemmes: Can we take the redis client dependency out now?

dgrove-oss: yep; good catch. Done.

dgrove-oss:
PG1 2231 (with redisclient removed).

dgrove-oss:
PG1 2231 completed successfully

PR #2917, opened 2017-10-31
Replace sequence with REST implementation

houshengbo:


houshengbo:
PG4 969 passed.

tests/src/test/scala/system/basic/WskConsoleTests.scala
```diff
@@ -70,7 +70,18 @@ class WskConsoleTests extends TestHelpers with WskTestHelpers {
     val start = Instant.now.minusSeconds(5)
     val payload = new String("from the console!".getBytes, "UTF-8")
     val run = wsk.action.invoke(fullActionName, Map("payload" -> payload.toJson))
-    withActivation(wsk.activation, run, totalWait = 30 seconds) { activation =>
+
+    val duration = Duration(Instant.now.minusMillis(start.toEpochMilli).toEpochMilli, MILLISECONDS)
```
dubeejw: Looks like this `duration` isn't used. the one below it is used.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -1149,10 +1146,15 @@ class RunWskRestCmd() extends FlatSpec with RunWskCmd with Matchers with ScalaFu
   }
   val connectionContext = new HttpsConnectionContext(SSL.nonValidatingContext, Some(sslConfig))
 
-  def validateStatusCode(expectedExitCode: Int, statusCode: Int) = {
+  def validateStatusCode(expectedExitCode: Int, statusCode: Int, backupExpectedExitCode: Int = -1) = {
```
dubeejw: What is the need for `backupExpectedExitCode`?

tests/src/test/scala/system/basic/WskConsoleTests.scala
houshengbo: Done.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: There is a special case for invokeaction.
If blocking or result is set to true, 200 is an acceptable status code, no matter what expectedExitCode is provided. I will change the code of invokeaction to narrate this case.

PR #2920, opened 2017-11-01
Replace basic runtimes with REST implementation

houshengbo:
Swift3, Swift311, Python, Node and Java are not cli related tests.
We do not have to maintain them as cli test suites, so a straight
replacement of wsk with wskrest will be sufficient for them.


houshengbo:
PG4 976 passed, which includes this PR as a base.

PR #2922, opened 2017-11-01
Replace non cli tests with REST implementation

houshengbo:


houshengbo:
@dubeejw I have updated the runtime tests.

houshengbo:
pg3/1295/

PR #2931, opened 2017-11-06
Replace the test cases under whisk.core.cli.test package

houshengbo:


tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -138,8 +141,11 @@ trait ListOrGetFromCollectionRest extends BaseListOrGetFromCollection {
                     expectedExitCode: Int = OK.intValue)(implicit wp: WskProps): RestResult = {
 
     val entPath = namespace map { ns =>
-      val (ns, name) = getNamespaceEntityName(resolve(namespace))
-      Path(s"$basePath/namespaces/$ns/$noun/$name/")
+      val (nspace, name) = getNamespaceEntityName(resolve(namespace))
+      name match {
```
rabbah: `if (name.isEmpty) ... else ...` is clearer imo.

tests/src/test/scala/whisk/core/cli/test/WskCliEntitlementTests.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import common.Wsk
+import common.TestUtils.FORBIDDEN
+import common.TestUtils.TIMEOUT
+import common.TestUtils.NOT_FOUND
+
+@RunWith(classOf[JUnitRunner])
+class WskCliEntitlementTests extends WskEntitlementTests {
+  override lazy val wsk = new Wsk
+  override def getForbiddenCode(): Int = FORBIDDEN
```
rabbah: you can override these as `val` instead of methods.

tests/src/test/scala/whisk/core/cli/test/WskEntitlementTests.scala
```diff
@@ -35,17 +33,22 @@ import whisk.core.entity.Subject
...
   lazy val defaultWskProps = WskProps()
   lazy val guestWskProps = getAdditionalTestSubject(Subject().asString)
 
   override def afterAll() = {
     disposeAdditionalTestSubject(guestWskProps.namespace)
   }
 
+  def getForbiddenCode(): Int
+  def getTimeoutCode(): Int
+  def getNotfoundCode(): Int
```
rabbah: these could be abstract values instead of methods.

rabbah: also you can drop the `get` prefix which is more scala idiomatic.

tests/src/test/scala/whisk/core/cli/test/WskCliEntitlementTests.scala
houshengbo: Done

tests/src/test/scala/whisk/core/cli/test/WskEntitlementTests.scala
houshengbo: Done.

tests/src/test/scala/common/rest/WskRest.scala
houshengbo: Done

PR #2934, opened 2017-11-07
Fix the example of default parameters

jnthas:
Update example that has place name overwritten by 'Washington, DC'


rabbah:
@jnthas do you have an Apache CLA on record already?

jnthas:
Hi! No, I don't have yet @rabbah. I can do if necessary.

rabbah:
It's ok for this change - for more substantial changes it will be necessary to have one on file.

PR #2938, opened 2017-11-07
Convert inline code to attachment in database

dubeejw:
Updates actions in database to move code from being inlined in the document to an attachment. The associated code for Java actions are already attachments. These changes Base64 encode Java attachments. In addition, the old exec.jar field is removed from Java actions.

Depends on https://github.com/apache/incubator-openwhisk/pull/2847

### DIFF
--- a/tools/db/README.md
+++ b/tools/db/README.md
@@ -138,3 +138,25 @@ Using that command will result in a replication for every database that matches 
 ### Replaying a snapshot
 
 To replay a snapshot, swap `--sourceDbUrl` and `--targetDbUrl` and call the script with the `replay` command. That command takes only 1 parameter: `--dbPrefix` to determine which backup to play back. Matching databases will be replicated back to the target database with the `backup_${TIMESTAMP_IN_SECONDS}_` removed, so they'd look just like the original database.
+
+## Database migration to new schema
+
+To reduce the memory consumption in the OpenWhisk controller, all code inlined in action documents has been moved to attachments. This change allows only metadata for actions to be fetched instead of the entire action. Though the OpenWhisk controller supports both mentioned schemas, it is ideal to update existing databases to use the new schema for memory consumption relief.
+
+1. The design document provided below will return all actions that are using the old schema. This design document needs to be manually to the related CouchDB
+```
+{
+  "_id": "_design/actionMigrate",
+  "language": "javascript",
+  "views": {
+    "actions": {
+      "map": "function (doc) {   var isAction = function (doc) {     return (doc.exec !== undefined)   };   var isMigrated = function (doc) {     return (doc._attachments !== undefined && doc._attachments.codefile !== undefined && typeof doc.code != 'string')   };   if (isAction(doc) && !isMigrated(doc)) try {     emit([doc.name]);   } catch (e) {} }"
+    }
+  }
+}
+```
+
+2. Run the `moveCodeToAttachment.py` to complete the schema update. Two parameters are required:
+
+* `--dbUrl`: Server URL of the database. E.g. 'https://xxx:yyy@domain.couch.com:443'.
+* `--dbName`: Name of the Database to update.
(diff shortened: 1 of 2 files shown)


dubeejw:
@csantanapr, FYI.

tools/db/moveCodeToAttachment.py
```diff
@@ -0,0 +1,100 @@
...
+            print('Document is an action: "{0}"'.format(id))
+
+            if doc['exec']['kind'] != 'java':
+                updated = updateNonJavaAction(db, doc, id)
+            else:
+                updated = updateJavaAction(db, doc, id)
+
+            if updated:
+                print('Updated action: "{0}"'.format(id))
+                time.sleep(.700)
```
cbickel: Why do we need this sleep here?

tools/db/moveCodeToAttachment.py
```diff
@@ -0,0 +1,100 @@
...
+    return updated
+
+def updateNonJavaAction(db, doc, id):
+    updated = False
+    code = doc['exec']['code']
+
+    if not isinstance(code, dict):
+        print('Updating action: "{0}"'.format(id))
+        db.put_attachment(doc, code, 'codefile', 'text/plain')
+        doc = db.get(id)
```
cbickel: Isn't this the doc, that is passed into this function?

cbickel:
I like this little script to update the database schema.
But one thing that comes into my mind is: If you have large databases, it could take a very long time to update the actions, because you are getting all the documents by its own.
What do you think about creating a design doc, that returns only the actions, thats still need to be modified?
Of cause, this design doc can be deleted afterwards.

Another aspect to make this script faster (if you want it to), would be batching.

tools/db/moveCodeToAttachment.py
dubeejw: It is, but the revision changes we the attachment is added. I thought it was safer to add the attachment first before deleting the inline code property incase one of the requests fails.

tools/db/moveCodeToAttachment.py
dubeejw: I suppose there doesn't need to be a timeout here. Was just concerned about overloading the machine this scripts gets ran on.

dubeejw:
Thanks for the review, @cbickel.

dubeejw:
Thinking something like this to have the database return the un-updated actions:```
{
  "_id": "_design/actionMigrate",
  "language": "javascript",
  "views": {
    "actions": {
      "map": "function (doc) {   var isAction = function (doc) {     return (doc.exec !== undefined)   };   var isMigrated = function (doc) {     return (doc._attachments !== undefined && doc_.attachments.codefile !== undefined)   };   if (isAction(doc) && !isMigrated(doc)) try {     emit([doc.namespace, doc.name]);   } catch (e) {} }"
    }
  }
}
```

http://192.168.99.100:5984/dockermachine_whisks/_design/actionMigrate/_view/actions

**Comment by dubeejw** [comment]
@cbickel, how are you thinking we should add the design doc to the database? Could use Ansible.

File: tools/db/moveCodeToAttachment.py (hunk 2 above)
Comment (cbickel) [reply tools/db/moveCodeToAttachment.py]: Oh yes you are right. I missed that.

**Comment by cbickel** [comment]
My suggestion for adding the Design doc would be, doing it manually. The reasons are:
- We need the doc only for a few days
- We have no task yet, that adds a design doc to a production environment (wipedb and initdb) delete the database as well. And if we add a job, this job would only be for this Design doc. Because all other Design docs should not be touched, otherwise they will be recomputed and a system with a large database has an outage (at least for the list commands)
- I think we don't need the doc in the pipeline (travis, ...) as the database is wiped there before a new build starts

Instead I'd suggest to write a step-by-step manual, how to update the database without any disruptions and take the creation and the deletion afterwards of the design doc as one step in this guide.

**Comment by dubeejw** [comment]
@cbickel, what do you think of having the migration script create the design doc?

File: tools/db/README.md [hunk 3]
```diff
@@ -138,3 +138,12 @@ Using that command will result in a replication for every database that matches
 ### Replaying a snapshot
 
 To replay a snapshot, swap `--sourceDbUrl` and `--targetDbUrl` and call the script with the `replay` command. That command takes only 1 parameter: `--dbPrefix` to determine which backup to play back. Matching databases will be replicated back to the target database with the `backup_${TIMESTAMP_IN_SECONDS}_` removed, so they'd look just like the original database.
+
+## Database migration to new schema
```
dubeejw: @csantanapr, @rabbah, what do you think about including the action migration doc in this README?

csantanapr:
@cbickel assigning to you since you already working with with @dubeejw on this PR.

tools/db/README.md
csantanapr: Yes add section in the this README also since the script is in the same folder.
It will be one place for db things ♻️

PR #2939, opened 2017-11-07
invoker cmd line support for dyanamicId assignment

dgrove-oss:
Add optional command line argument for invoker name
and invoker proposed id to support Mesos deployment
(see issue https://github.com/apache/incubator-openwhisk/issues/2872).

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
@@ -47,6 +47,8 @@ import whisk.spi.SpiLoader
 import whisk.utils.ExecutionContextFactory
 import whisk.common.TransactionId
 
+case class CmdLineArgs(name: Option[String] = None, id: Option[Int] = None)
+
 object Invoker {
 
   /**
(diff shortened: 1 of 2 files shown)


dgrove-oss:
PG4 989

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -104,8 +107,24 @@ object Invoker {
       abort("Bad configuration, cannot start.")
     }
 
-    val proposedInvokerId: Option[Int] = args.headOption.map(_.toInt)
-    val assignedInvokerId = proposedInvokerId
+    // process command line arguments
+    // We accept the command line grammar of:
+    // Usage: invoker [options] [<proposedInvokerId>]
+    //    --name <value>   a unique name to use for this invoker
```
tysonnorris: are there any restrictions on what characters can be passed here for --name? (I assume not, like dots and dashes should work?)

dgrove-oss: We're going to use invokerName as a piece of a zookeeper path.  So there are some modest restrictions https://zookeeper.apache.org/doc/r3.1.2/zookeeperProgrammers.html.  But I believe dots in the middle are fine (eg ip addresses or hostnames).  I didn't try it though.

PR #2932, opened 2017-11-07
refactoring loadbalacing service with an overflow queue

tysonnorris:
Unify overflow topic and revise invoker scheduling.
Includes:
- revert LoadBalancerData APIs to be synchronous (with updates pushed via replication)
- LoadBalancerService class replaced with LoadBalancerActorService (publish function sends a message)
- InvokerPool is configured outside of LoadBalancerServiceActor (and passed on construction; to make testing simpler)
- SpiLoader allows config injection (so that testkit ActorSystems can be injected with inline configs (instead of always reading application.conf)
- TestConnector enhanced to include a full TestMessagingProvider, to allow testing message flows

Note that one side affect is that a request arriving at controller0 that gets queued to overflow, MIGHT get routed to controller1 for invoker scheduling

This is supporting the "PR1" portion of https://cwiki.apache.org/confluence/display/OPENWHISK/Invoker+Activation+Queueing+Change

Issue #2859, opened 2017-10-16
LoadBalancerService does not consider concurrent requests when scheduling


tysonnorris:
Under concurrent requests to controller, LoadBalancerService will not consider concurrent requests when scheduling to an invoker. 

For example, consider `LoadBalacnerServiceObjectTests.scala`, with the following test
```
  it should "consider load immediately and jump to next invoker when overloaded" in {
    val invokerCount = 10
    val hash = 2
    val targetInvoker = hash % invokerCount

    val invs = invokers(invokerCount).updated(targetInvoker, (InstanceId(targetInvoker), Healthy, 0))
    val step = hashInto(LoadBalancerService.pairwiseCoprimeNumbersUntil(invokerCount), hash)

    LoadBalancerService.schedule(invs, 1, hash) shouldBe Some(InstanceId((hash + 0) % invs.size))
    LoadBalancerService.schedule(invs, 1, hash) shouldBe Some(InstanceId((hash + 1) % invs.size))

  }
```

We have:
* set the existing load to 0 to start with
* set the invokerBusyThreshold to 1 on `scheduler()` invocations
* schedule one activation with hash=2
* schedule second activation with hash=2

In this case, but activations are scheduled to InstanceId(2), but the second should be scheduled to next invoker since busy threshold is exceeded.

I believe this is related to the changing of `LoadbalancerData.activationCountPerInvoker()` from sync to async. Since LoadBalancerService is processed with multiple threads, the updates to the LoadBalancerData are not seen immediately by other threads.

vvraskin:
Good catch. 
I will try to reproduce it later, but I think you are right, the fact that we read async, doesn't allow us to see the update by the second call of `schedule`. 
Since we only write to the local instances, we could consider changing read to be synchronous, need to measure the performance overhead though. 
According to akka doc the use case you are showing should be supported:
> You will always read your own writes. For example if you send a Update message followed by a Get of the same key the Get will retrieve the change that was performed by the preceding Update message.

tysonnorris:
I think "reading your own writes" is only for the replicated data, which is inherently async (by way of messaging with replicator), so I'm not sure that will work (but great if it does!).

I was thinking about the following:
- revert to the LoadBalancerData to sync signatures (e.g. `def activationCountPerInvoker: Map[String, Int]`
- make DistributedLoadBalancerData keep 2 sets of data - one local, one replicated
- the views that DistributedLoadBalancerData exposes would be a composition of local + replicated data 

[...]

PR #2943, opened 2017-11-07
revert LoadBalancerData to be synchronous

tysonnorris:
Make LoadBalancerData API synchronous again. 
Partial fix for #2859 (rest of fix comes with making LoadBalancerService an Actor)
Partial impl for #2932


markusthoemmes:
In a brief chat with the Akka team they said that using `WriteLocal` you will always be able to read your own writes immediatly.

To me that seems like making the LoadBalancerData synchronous is not needed for the intended effect of being able to have consistent local data. Making scheduling sequential though, is of course a requirement to prevent stale reads.

tysonnorris:
Consider that LoadBalancerService will be changed to generate a Publish message to be handled by a LoadBalancerActor. 

I understand that the replicator enables reading local writes, but the issue is that assigning ActivationEntry to an invoker (via LoadBalancerData.putActivation) is done async from testing the current assigned activations (via LoadBalacnerData.activationCountPerInvoker), so while Publish messages can be made to arrive sequentially, the invoker selection process results in futures that complete at various times (without the ability to consider the futures that were generated previously). 

There would need to be a way to cause publish messages to be processed only when previous futures have reached a state of completion. 

I may not be stating this clearly, but is there some better way to wait for chooseInvoker() Future (+ schedule()) to complete before processing the next Publish message?

markusthoemmes:
@tysonnorris I think that is exactly what we need: `chooseInvoker()` needs to be sequentialized! Other parts of the Loadbalancer can even stay as-is.

This can for example be quite easily done by creating an akka-stream with a `mapAsync` stage with an asynchronity of 1. Alternatively we could buffer all requests and drain that buffer asynchronously one-by-one.

core/controller/src/main/scala/whisk/core/loadBalancer/DistributedLoadBalancerData.scala
```diff
@@ -17,69 +17,102 @@
...
-class DistributedLoadBalancerData(implicit actorSystem: ActorSystem, logging: Logging) extends LoadBalancerData {
+class DistributedLoadBalancerData(instance: InstanceId, monitor: Option[ActorRef] = None)(
+  implicit actorSystem: ActorSystem,
+  logging: Logging)
+    extends LoadBalancerData {
 
   implicit val timeout = Timeout(5.seconds)
   implicit val executionContext = actorSystem.dispatcher
-  private val activationsById = TrieMap[ActivationId, ActivationEntry]()
+  private val overflowKey = "overflow"
```
markusthoemmes: All things referencing "overflow" should not be part of this PR.

core/controller/src/main/scala/whisk/core/loadBalancer/DistributedLoadBalancerData.scala
```diff
@@ -17,69 +17,102 @@
...
+  private val activationsById = mutable.Map[ActivationId, ActivationEntry]()
```
markusthoemmes: Why remove the `TrieMap` and make this a (non-concurrent) `Map`?

core/controller/src/main/scala/whisk/core/loadBalancer/DistributedLoadBalancerData.scala
```diff
@@ -17,69 +17,102 @@
...
+    val shared = sharedDataNamespaces.getOrElse(namespace.toString, Map()).filter(_._1 != instance.toInt).values.sum
+    shared + localData.activationCountOn(namespace)
   }
 
-  def activationCountPerInvoker: Future[Map[String, Int]] = {
-    (sharedStateInvokers ? GetMap).mapTo[Map[String, BigInt]].map(_.mapValues(_.toInt))
+  def activationCountPerInvoker: Map[String, Int] = {
+    val shared = sharedDataInvokers.mapValues(_.filter(_._1 != instance.toInt).values.sum)
+    val local = localData.activationCountPerInvoker
+    local ++ shared.map { case (k, v) => k -> (v + local.getOrElse(k, 0)) }
```
markusthoemmes: My understanding is that this merges local + shared state on **each** request. Can we rather merge updates into the local state and have each request be just a `get`?

This comment is valid for all "requests"

core/controller/src/main/scala/whisk/core/loadBalancer/SharedDataService.scala
```diff
@@ -54,45 +63,53 @@ class SharedDataService(storageName: String) extends Actor with ActorLogging {
...
     case MemberRemoved(member, previousStatus) =>
       logging.warn(this, s"Member is Removed: ${member.address} after $previousStatus")
-
-    case c @ Changed(_) =>
+    case c @ Changed(e) =>
       logging.debug(this, "Current elements: " + c.get(storage))
+      val res = c.get(storage).entries.mapValues(_.entries.mapValues(_.toInt))
+      if (res.nonEmpty) {
+        res.values.foreach(_.values.foreach(i => {
+          require(i >= 0, s"values cannot be less than 0 ${res}")
```
markusthoemmes: Do we really want an exception to be thrown here? What's the supervision strategy for that?

core/controller/src/main/scala/whisk/core/loadBalancer/DistributedLoadBalancerData.scala
```diff
@@ -17,69 +17,102 @@
...
+import whisk.core.entity.UUID
 
 /**
  * Encapsulates data used for loadbalancer and active-ack bookkeeping.
  *
  * Note: The state keeping is backed by distributed akka actors. All CRUDs operations are done on local values, thus
  * a stale value might be read.
  */
-class DistributedLoadBalancerData(implicit actorSystem: ActorSystem, logging: Logging) extends LoadBalancerData {
+class DistributedLoadBalancerData(instance: InstanceId, monitor: Option[ActorRef] = None)(
```
markusthoemmes: What's `monitor` used for?

tysonnorris: monitor is used for testing, to test replication Changed events from SharedDataService:
```    case c @ Changed(e) =>
      logging.debug(this, "Current elements: " + c.get(storage))
      val res = c.get(storage).entries.mapValues(_.entries.mapValues(_.toInt))
      if (res.nonEmpty) {
        res.values.foreach(_.values.foreach(i => {
          require(i >= 0, s"values cannot be less than 0 ${res}")
        }))
        monitor ! Updated(storageName, res)
      }
```

core/controller/src/main/scala/whisk/core/loadBalancer/DistributedLoadBalancerData.scala
tysonnorris: I was going down the path of making lb an actor; but that can be a separate discussion, possibly.

core/controller/src/main/scala/whisk/core/loadBalancer/DistributedLoadBalancerData.scala
tysonnorris: I'm not sure - since the "shared" data is async and the local data is not, we need to assemble the data somehow to get an accurate map of values. 

For example, if we just use "shared" store to manage everything, that may be enough, but I'm not sure that gets accurate values (but maybe that is ok). I'll try it.

tysonnorris: Latest change (7b7f600) attempts this by a) extending LocalLoadBalancerData and b) resetting state periodically when replication updates are received. 

I need to test this more, and I'm not sure if there are better ways to do the update (currently does a clear() + foreach() to rebuild the TrieMaps) but let me know if you have thoughts on this approach.

tysonnorris:
@markusthoemmes I tried a new approach which is NOT to merge, but rather replace the TrieMap/AtomicIntegers each time the distributed data is updated (for activationByInvoker and activationByNamespaceId). 

I added a test in LoadBalancerDataTests ("quickly execute many activations") that demonstrates - note that the accuracy of counts fluctuates with DistributedData Changed events e.g.:

- TrieMaps are updated locally, with additional updates send to shared data
- until shared data update is received, TrieMap is accurate for local data, but not for cluster data
- once a shared data update is received, the TrieMap is rebuilt with data that is accurate for the cluster, but may overwrite some local state (which will arrive with the next cluster update)

Therefore test needs to wait (using Eventually) to receive counts for all the activations. 

Some rough testing numbers:
Running same test on previous (async) code takes 4.5s for 100k activations, and on this code takes ~1.2s for 100k activations; Meanwhile LocalLoadBalancerData takes ~1s in both cases (as expected). 

Let me know if you think this test is appropriate?

markusthoemmes:
Is it okay though to drop local data each time a cluster update comes in? Doesn't that mean that after the first cluster update, local counts are off again?

tysonnorris:
Yes, after *each* cluster update local counts will be off again, however the counts will always be off for remote/cluster data, unless we do the "merge per request" type of approach.  (to track which counts are contributed by remote instances and which are local)

markusthoemmes:
WDYT about the possibility of not using state-sharing at all but instead shard the invokers to avoid collisions as laid out in #3240?

tysonnorris:
As long as there are no cases where people expect the non-sharding lb to work properly, this won't be needed.

PR #2945, opened 2017-11-08
Be able to exclude databases by basename in replication

cbickel:
With our current replication skript, we are only able to exclude databases by its name, but not by its base-name. This means, that if a database has a suffix, it has to be excluded explicitly. 

With this PR, these databases can be excluded by their base name.

### DIFF
--- a/tools/db/replicateDbs.py
+++ b/tools/db/replicateDbs.py
@@ -24,6 +24,7 @@ import argparse
 import time
 import re
 import couchdb.client
+import functools
 
 def retry(fn, retries):
     try:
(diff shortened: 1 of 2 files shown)


tools/db/replicateDbs.py
```diff
@@ -42,6 +43,9 @@ def replicateDatabases(args):
     targetDb = couchdb.client.Server(args.targetDbUrl)
 
     excludedDatabases = args.exclude.split(",")
+    excludedBaseNames = []
+    if args.excludeBaseName:
+        excludedBaseNames = args.excludeBaseName.split(",")
```
markusthoemmes: Could be:

```python
filter(lambda x: x != "", args.excludeBaseName.split(","))
```

Same above.

PR #2947, opened 2017-11-08
disable runc usage in docker-machine environments by default

jeremiaswerner:
I was running into an issue with newer docker-machine version `17.06` where the invoker throws the following issue. 

```
[2017-11-08T16:47:55.884Z] [ERROR] [#sid_102] [RuncClient] code: 1, stdout: , stderr: json: cannot unmarshal object into Go value of type []string [marker:invoker_runc.pause_error:103484:238]
```

I therefore propose to disable `runc` by default in docker-machine environments.

### DIFF
--- a/ansible/environments/docker-machine/group_vars/all
+++ b/ansible/environments/docker-machine/group_vars/all
@@ -29,3 +29,4 @@ controller_arguments: '-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxre
 invoker_arguments: "{{ controller_arguments }}"
 
 invoker_allow_multiple_instances: true
+invoker_use_runc: false


rabbah:
Hmm can we make this optional? Some of us don't upgrade and develop/run locally almost exclusively.

rabbah:
Another consideration is that docker machine as documented is tied to an older version of docker. The new deployment for mac should be using docker for mac. So maybe we can just leave this alone for docker machine.

csantanapr:
@rabbah you can use `-e invoker_use_runc=true` for your use case.
Most folks will use newer docker

What problem do you hit by having it false in docker-machine? 
if runc is false, everything works fine using docker to unpause instead of runc

csantanapr:
Just thinking out loud:

Most users in the community using new docker (Docker for mac) are using newest Docker version (17.x) as it's easier to get the latest version the Docker for Mac auto updates.

For docker-machine is easier to install old version and keep it isolated via vm name `whisk` from the other development activities outside Whisk that an user will do in his computer.

In addition Docker for mac is kind of broken because runc is not possible because the hardcoded version in the invoker Docker file.

If we want to attract more community help we need to make it much easier to get their computer working out of the box, also most of the work people are contributing are not specific to having docker 1.12 installed.

I propose 3 things to compromise:
1. Have docker-machine with runc enable by default, assume for the most part people are using 1.12 
   Doc if user is using newer version of docker they need to disable runc via file or ansible override
2. Have Docker for Mac with runc always disable 
3. At some point we need to look into upgrading the version of docker we support, the more we wait more painful will be to upgrade, as more and more users in the wild will be using latest version when deploying

rabbah:
A previous pr added: `./group_vars/all:  useRunc: "{{ invoker_use_runc | default(true) }}"` which satisfies 1 but does not work for docker-for-mac because it shares the `local` environment.

rabbah:
Related: https://github.com/apache/incubator-openwhisk/pull/2799.

rabbah:
Closing since there is a flag to suppress runc per environment and a pr from @ddragosd for docker for mac.

PR #2948, opened 2017-11-08
Programmatic lazy creation of completedN and invokerN topics

dgrove-oss:
Each invoker and controller now ensures that their message topics are
created as part of their individual startup procedure. As a result, we
no longer need to eagerly create the topics during kafka startup
to ensure the topics have the desired config parameters.

This change is motivated by simplifying the deployment of OpenWhisk on
Kubernetes, but also generally enables more dynamic sizing of the
controller/invoker pools in any deployment.

The change has four main parts:
  1. addition of the ensureTopic method to the MessagingProvider SPI
  2. bump of kafka version to 0.11.0.1 (required to get AdminClient API).
  3. implementation of ensureTopic for kafka using AdminClient
  4. Changes to controller and invoker code and deployment files to
     use the ensureTopic SPI method during startup.

Note that although the various config parameters _may_ be overriden in
the controller/invoker environment, defining the variables is not
required. Defaults are defined that match the current ansible default
values.


core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -341,11 +341,17 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
...
-    kafkaHost ++ Map(
-      loadbalancerInvokerBusyThreshold -> null,
-      controllerBlackboxFraction -> null,
-      controllerLocalBookkeeping -> null,
-      controllerSeedNodes -> null)
+    kafkaHost ++
+      Map(
+        kafkaTopicsCompletedRetentionBytes -> (1024 * 1024 * 1024).toString(), // default is 1024 MB
+        kafkaTopicsCompletedRetentionMS -> (1 * 3600 * 1000).toString(), // default is 1 hour
+        kafkaTopicsCompletedSegmentBytes -> (512 * 1024 * 1024).toString() // default is 512 MB
```
markusthoemmes: We have some helpers here:

```scala
import whisk.core.entity.size._

1024.MB.toBytes
1.hour.toMillis
512.MB.toBytes
```

Also Scala nit: For non sideeffecting accessor-like methods, you can usually omit `()`, `toString` in this case.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -58,6 +58,11 @@ object Invoker {
       WhiskEntityStore.requiredProperties ++
       WhiskActivationStore.requiredProperties ++
       kafkaHost ++
+      Map(
+        kafkaTopicsInvokerRetentionBytes -> (1024 * 1024 * 1024).toString(), // default is 1024 MB
+        kafkaTopicsInvokerRetentionMS -> (48 * 3600 * 1000).toString(), // default is 48 hours
+        kafkaTopicsInvokerSegmentBytes -> (512 * 1024 * 1024).toString() // default is 512 MB
```
markusthoemmes: Same here.

common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
```diff
@@ -35,4 +45,30 @@ object KafkaMessagingProvider extends MessagingProvider {
...
+
+  def ensureTopic(config: WhiskConfig, topic: String, topicConfig: Map[String, String])(
+    implicit logging: Logging): Boolean = {
+    val props = new Properties
+    props.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, config.kafkaHost)
+    val client = AdminClient.create(props)
+    val numPartitions = topicConfig.getOrElse("numPartitions", "1").toInt
+    val replicationFactor = topicConfig.getOrElse("replicationFactor", "1").toShort
+    val nt = new NewTopic(topic, numPartitions, replicationFactor)
+      .configs(topicConfig.-("numPartitions", "replicationFactor").asJava)
```
markusthoemmes: Proposing: `.configs((topicConfig - ("numPartitions", "replicationFactor")).asJava)`

common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
```diff
@@ -35,4 +45,30 @@ object KafkaMessagingProvider extends MessagingProvider {
...
+    val results = client.createTopics(List(nt).asJava)
+    try {
+      results.values().get(topic).get()
+      logging.info(this, s"created topic $topic")
+      true
+    } catch {
+      case e: ExecutionException if e.getCause().isInstanceOf[TopicExistsException] =>
```
markusthoemmes: `getCause` is an accessor --> drop the `()`

tests/src/test/scala/whisk/core/loadBalancer/test/InvokerSupervisionTests.scala
```diff
@@ -206,7 +206,8 @@ class InvokerSupervisionTests
 
     sendActivationToInvoker
       .when(activationMessage, invokerInstance)
-      .returns(Future.successful(new RecordMetadata(new TopicPartition(invokerName, 0), 0L, 0L, 0L, 0L, 0, 0)))
+      .returns(Future.successful(
+        new RecordMetadata(new TopicPartition(invokerName, 0), 0L, 0L, 0L, java.lang.Long.valueOf(0L), 0, 0)))
```
markusthoemmes: Fairly interesting interface 🤔 

There's a scala version for `java.long.Long.valueOf` --> `Long.box(0L)`

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
dgrove-oss: Nice.  much easier to read.

tests/src/test/scala/whisk/core/loadBalancer/test/InvokerSupervisionTests.scala
dgrove-oss: yeah, someone is playing a game with overloading and deprecation.  Must be a story behind it in kafka land...

dgrove-oss:
PG3 1314

dgrove-oss:
PG3 1315

dgrove-oss:
squashed to single commit.  
PG2 2313 (in queue)

dgrove-oss:
@cbickel - ready for final review

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -182,6 +187,17 @@ object Invoker {
 
     val invokerInstance = InstanceId(assignedInvokerId)
     val msgProvider = SpiLoader.get[MessagingProvider]
+    if (!msgProvider.ensureTopic(
```
rabbah: should this happen inside the kafka messaging provider since this may not be applicable to an alternate bus implementation?

dgrove-oss: The keys of the Map are pretty kafka specific.  The values are both kafka specific and specific to the topic being created (see similar block in the controller where we read different fields of WhiskConfig).

I think using a Map[String,String] is probably workable in a generic SPI, but when populating the map you end up needing to know the SPI implementation you are talking to.

An alternative would be to not pass a Map and instead pass just the config object and the topic name. We could look at the topic prefix inside the ensureTopic implementation of the kafka messaging provider and use it to figure out which fields of the config to extract.  

I think this really doesn't help make it more generic, since the details of the messaging provider are still being leaked into the WhiskConfig.  I could make the change though without much effort.

dgrove-oss: The general question is (which maybe has been answered and I just don't know it), how are SPI implementations supposed to plumb their configuration values from the deploying environment through to the implementation code that needs them.  Adding them all to WhiskConfig seems wrong, but I don't know what else to do.

rabbah: @markusthoemmes @tysonnorris I think we’ve discussed this in the past. I have two comments: whisk config will get replaced and so the maps will likely be gone. I’ll link a PR as an example. Alternatively the Plugins would provide a requiredProperties method which get added to the whisk config object. It’s ok with me to defer this until to a separate pr down the line.

dgrove-oss:
pinging reviewers: @cbickel @rabbah.  I have kube work blocked on this PR; can we please progress it.

ansible/roles/kafka/tasks/deploy.yml
```diff
@@ -60,17 +60,3 @@
     settings: "{{ kafka.topics.health }}"
```
markusthoemmes: What about the health topic? Why not have them all try to create it and the first one wins?

dgrove-oss: I thought about that.  Same for the cacheInvalidation topic.  The controllers could just race to create them on startup and one of them will win.  If we think it is a good idea, I can do a followup PR to get all the topic logic out of ansible.

Should the followup PR wait on the pureconfig PR?  If not, I'll have to plumb more kafka-specific flags through the WhiskConfig.

common/scala/build.gradle
```diff
@@ -25,7 +25,7 @@ dependencies {
     compile 'commons-codec:commons-codec:1.9'
     compile 'commons-io:commons-io:2.4'
     compile 'commons-collections:commons-collections:3.2.2'
-    compile 'org.apache.kafka:kafka-clients:0.10.2.1'
+    compile 'org.apache.kafka:kafka-clients:0.11.0.1'
```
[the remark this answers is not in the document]
dgrove-oss: Starting with kafka 0.10.2, they support both forward & backward version mismatch of the server and clients.  So we don't have to be running the 0.11 server to use the 0.11 client.

I do think upgrading the kafka server is a good idea to keep current, so I am planning on submitting a separate PR that just does that.  By isolating that change it will be easier to roll it back and/or performance test it separately.

dgrove-oss:
anything else?

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
tysonnorris: Using typesafe config or pureconfig, I think, will make this issue moot since there will be no single "universal" config class; rather SPI impls can declare their own etc, and only that impl will reference its config. 

The only questions is whether there should be a universal config validation, which SPI impls should be able to contribute to? This I think can be added later, if at all. In the meantime, the difference is that "missing configs" required by SPI impls would cause failures ONLY when that impl is loaded. In all cases, impls should be able to require different config keys - but I don't think that is being questioned.

dgrove-oss: Once we have pureconfig and I can get the kafka-specific params internal to the SPI, I expect this block of code will reduce to

```
if (!msgProvider.ensureTopic(config,  "invoker" + assignedInvokerId) {
      abort(s"failure during msgProvider.ensureTopic for topic invoker$assignedInvokerId")
 }
```

PR #2951, opened 2017-11-09
Increase the number of max open requests

vvraskin:
At the moment we divide the number of parallel requests across 3 stores that use CouchDB client: entity, authstore and activationStore, thanks to the fact that the number of requests is shared across all flow materialisations as per this doc:
https://doc.akka.io/docs/akka-http/current/scala/http/configuration.html

This PR should help to prevent dropping the connections if the number of parallel requests to couchdb exceeds the max number of open connections.  
> Common causes of pool overload: 
There are peaks in the request rate (prevent peaks by tuning the client application or increase max-open-requests to buffer short-term peaks)
https://doc.akka.io/docs/akka-http/current/scala/http/client-side/pool-overflow.html

@bwmcadams please correct me if this is not correct.  

Thanks.

### DIFF
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -9,7 +9,7 @@ akka.http {
 
     host-connection-pool {
         max-connections = 128
-        max-open-requests = 128
+        max-open-requests = 512
     }
 }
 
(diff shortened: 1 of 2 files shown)


bwmcadams:
@vvraskin That sounds correct based on my recollection as well as review of the docs. ( I don't believe you can configure separate pools for each connection set which would be my first connection ).

Is 512 too much / not enough? Any benchmark concerns with it? (I don't see it causing a memory or performance limit, as the connection overhead should be low… call it curiosity)

markusthoemmes:
@bwmcadams we have a Buffer in front of the pools holding up to 16k outstanding requests anyway so.... I believe that new value isn't going to matter much?

bwmcadams:
@markusthoemmes No, it shouldn't. You're right. I'm just being paranoid.

jeremiaswerner:
PG3/1352 🔵

PR #2952, opened 2017-11-09
Re-implement the non-cli test cases with REST

violagao:
The test cases changed in this PR are non-CLI test cases, which should use
the REST only to access the openwhisk services. This PR replaces the Wsk
with WskRest in order to make it happen.


houshengbo:
PG 3 1295 passed for this PR.

PR #2956, opened 2017-11-10
Introduce separate Akka dispatchers for CouchDB and Kafka Clients

### LINKED ISSUE #2954: All Akka related code sharing a single dispatcher
Currently, all of the akka code (controller actors, invoker actors, akka-http actors, couchdb client, kafka client) are sharing a single dispatcher. This goes against best practices as it means that all of these separate features are sharing a single threadpool and competing for resources.

Different pieces will likely have different tuning requirements and priorities. It is best to separate them out into a few dispatcher pools that can be adjusted as needed… especially anywhere there is blocking code.

We've discussed this on Slack; I'm proposing leaving akka-http, controller, invoker, and the schedulers on the default system dispatcher for now but separating couchdb and kafka onto their own dispatchers. This allows those resources to be adjusted as needed and prevents any slow calls, blocking calls, etc from holding threads that may be needed by other pieces of the system.

bwmcadams:
- Rest of system continues to use the default system dispatcher
- Kafka and CouchDB clients each get their own dispatchers
  * This should prevent all of the akka based stuff from competing constantly for threads
  * Kafka and Couch can be tuned now based on individual needs
- Dispatchers are defined in common `reference.conf`, so config can be overridden by dependents

Closes #2954


markusthoemmes:
PG2 2348 🚀

markusthoemmes:
I took the liberty to rebase this, you'll need to override your local branch.

common/scala/src/main/resources/reference.conf
```diff
@@ -3,4 +3,36 @@ whisk.spi{
...
+
+      fork-join-executor {
+          # Min number of threads to cap factor-based parallelism number to
+          parallelism-min = 2
+          # Parallelism (threads) ... ceil(available processors * factor)
+          parallelism-factor = 2.0
+          # Max number of threads to cap factor-based parallelism number to
+          parallelism-max = 10
+      }
+  }
```
markusthoemmes: Are we at risk to starve this one? Kafka producing **can** be blocking and polling for sure is blocking although there should never be more than a couple of polls active.

bwmcadams: er, yeah. I think I copied this from docs; it looks like the default for Akka's default dispatchers is 64. I'll move it to that and we can always tweak specifically later.

bwmcadams:
@markusthoemmes Changed to leave fork-join-executor configs at the Akka default, until we can choose better default tuning for the separated dispatchers.

dubeejw:
@bwmcadams, is this PR still valid?

bwmcadams:
@dubeejw the intent is still there; I have to check the upstream changes though. Will update.

markusthoemmes:
@bwmcadams I'll close this one for now, please reopen once you get to it.

PR #2957, opened 2017-11-10
splunk logstore

tysonnorris:
includes tests to verify use of splunk api


common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -0,0 +1,158 @@
...
+
+  val splunkHost = config.getString("whisk.logstore.splunk.host")
+  val splunkPort = config.getInt("whisk.logstore.splunk.port")
+  private val splunkApi = "/services/search/jobs" //see http://docs.splunk.com/Documentation/Splunk/6.6.3/RESTREF/RESTsearch#search.2Fjobs
+  private val splunkUser = config.getString("whisk.logstore.splunk.user")
+  private val splunkPass = config.getString("whisk.logstore.splunk.password")
+  private val splunkIndex = config.getString("whisk.logstore.splunk.index")
+  private val logMessageFieldName = config.getString("whisk.logstore.splunk.log-message-field")
+  private val activationIdFieldName = config.getString("whisk.logstore.splunk.activation-id-field")
+  private val disableSNI = config.getBoolean("whisk.logstore.splunk.disableSNI")
```
markusthoemmes: I think it would be valuable to use #2576 (pureconfig) to parse all of these into a case-class. Will be much less verbose.

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -0,0 +1,158 @@
...
+
+  val log = actorSystem.log
```
markusthoemmes: Doesn't use `AkkaLogging` for the usual log format?

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -0,0 +1,158 @@
...
+        "latest_time" -> formatter.format(activation.end))).toEntity
+
+    log.debug("sending request")
+    queueRequest(
+      Post(splunkApi)
+        .withEntity(entity)
+        .withHeaders(List(Authorization(BasicHttpCredentials(splunkUser, splunkPass))))).flatMap(response => {
+      log.debug(s"splunk API response ${response}")
+      if (response.status.isSuccess()) {
+        Unmarshal(response.entity)
```
markusthoemmes: Make sure to **always** unmarshal the response, even if you choose to ignore the bytes. Otherwise the connection will hang.

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -0,0 +1,158 @@
...
+          .to[String]
```
markusthoemmes: Is it possible to define the log format statically and Unmarshal straight to that?

For example:

```scala
case class SplunkResponse(results: Seq[SplunkResult])
case class SpunkResult(logMessage: String)
```

You would then be able to use

```scala
Unmarshal(response.entity).to[SplunkResponse].map { response =>
  val messages = response.results.map(_.logMessage)
  ActivationLogs(messages)
}
```

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -0,0 +1,158 @@
...
+        Future.failed(new RuntimeException(s"failed to read logs from splunk ${response}"))
+      }
+    })
+
+  }
+
+  //based on http://doc.akka.io/docs/akka-http/10.0.6/scala/http/client-side/host-level.html
+  val queue =
+    Source
+      .queue[(HttpRequest, Promise[HttpResponse])](maxPendingRequests, OverflowStrategy.backpressure)
```
markusthoemmes: How are you applying backpressure? `queueRequest` will just continue to run, wouldn't it?

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * A Splunk based impl of LogDriverLogStore. Logs are routed to splunk via docker log driver, and retrieved via Splunk REST API
+ * @param actorSystem
+ * @param httpFlow Optional Flow to use for HttpRequest handling (to enable stream based tests)
+ */
+class SplunkLogStore(
+  actorSystem: ActorSystem,
+  httpFlow: Option[Flow[(HttpRequest, Promise[HttpResponse]), (Try[HttpResponse], Promise[HttpResponse]), Any]] = None)
+    extends LogDriverLogStore(actorSystem) {
```
markusthoemmes: If you apply the suggestion with `pureconfig` below, you can change the signature here to:

```scala
class SplunkLogStore(
  actorSystem: ActorSystem,
  httpFlow: Option[Flow[(HttpRequest, Promise[HttpResponse]), (Try[HttpResponse], Promise[HttpResponse]), Any]] = None,
  config: SplunkLogStoreConfig = loadConfigOrThrow[SplunkLogStoreConfig]("whisk.logstore.splunk"))
```

The huge benefit of this is, that in your tests you won't need to create a String containing the overriding config but you can rather construct your testconfig and pass it in as an override. I think that makes stuff a lot more convenient.

tests/src/test/scala/whisk/core/containerpool/logging/SplunkLogStoreTests.scala
```diff
@@ -0,0 +1,207 @@
...
+  val splunkPort = system.settings.config.getInt("whisk.logstore.splunk.port")
+
+  val testFlow: Flow[(HttpRequest, Promise[HttpResponse]), (Try[HttpResponse], Promise[HttpResponse]), NotUsed] =
+    Flow[(HttpRequest, Promise[HttpResponse])]
+      .mapAsyncUnordered(1) {
+        case (request, userContext) =>
+          //handle assertion errors so that stream does not swallow error info
+          try {
+            request.uri.path.toString() shouldBe "/services/search/jobs"
+            request.headers shouldBe List(Authorization.basic(splunkUser, splunkPwd))
```
markusthoemmes: Can these move into the inner try block to get away some nesting?

tests/src/test/scala/whisk/core/containerpool/logging/SplunkLogStoreTests.scala
```diff
@@ -0,0 +1,207 @@
...
+                    HttpResponse(
+                      StatusCodes.OK,
+                      entity = HttpEntity(
+                        ContentTypes.`application/json`,
+                        """{"preview":false,"init_offset":0,"messages":[],"fields":[{"name":"log_message"}],"results":[{"log_message":"some log message"},{"log_message":"some other log message"}], "highlighted":{}}"""))),
+                  userContext)
+              } catch {
+                case e: Throwable =>
+                  (Failure(e), userContext)
+              }
```
markusthoemmes: Do you need the second try/catch? Seems like this would also translate to:

```scala
Unmarshal(request.entity).to[FormData].map { form =>
  // asserts
  (Success(HttpResponse(...)), userContext)
}.recover { case e =>
  (Failure(e), userContext)
}
```

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
tysonnorris: Do you mean you think this should use whisk.common.Logging?

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
tysonnorris: I expect Source.queue() to backpressure at queueRequest once the queue buffer is full (hardcoded to 500 currently). SourceQueue.offer() docs indicate failure when backpressure mode is set.

markusthoemmes: I see. Since you're not retrying `offer`, does it make sense to use `dropNew` here (as we do in `CouchDbRestClient`) to explicitly drop the message? I wonder why one would even want the distinction 🤔 .

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
tysonnorris: I'm not sure, since statically defining the format implies that the log fields (and field names) are the same for everyone. I think this opens a bigger question I have, which is: can we require action developers to leverage structured loggers instead of just printing to stdout/stderr?

Currently this impl assumes "some structure", but does not dictate the field names that are used (they can be specified via config).

I'm open to ideas on this.

tests/src/test/scala/whisk/core/containerpool/logging/SplunkLogStoreTests.scala
tysonnorris: yes

tests/src/test/scala/whisk/core/containerpool/logging/SplunkLogStoreTests.scala
tysonnorris: yes, I was thrown off by my IDE incorrectly hinting a compile error at `(Failure(e), userContext)`. Ignoring that does wonders to simplify this (and it works fine too).

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
tysonnorris: I had expected offer to retry during backpressure mode, so I think you're right. Will change to dropNew.

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
tysonnorris: ah nice! 
I updated to do this.

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
tysonnorris: done

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
markusthoemmes: Hmm, does the fluentd driver not produce the same results for every line written to stdout/stderr? That is: Won't the message being printed to either stream always be captured in the `log_message` field regardless of what the user does?

Upon inspection, fluentd actually [hardcodes "log"](https://github.com/moby/moby/blob/master/daemon/logger/fluentd/fluentd.go#L148-L152) as its output field, maybe Splunk applies the transformation? In any case: How would the user be able to alter that?

common/scala/src/main/scala/whisk/core/containerpool/logging/LogDriverLogStore.scala
```diff
@@ -0,0 +1,70 @@
...
+ * Fetching logs from that external store is not provided in this trait.
+ */
+class LogDriverLogStore(actorSystem: ActorSystem,
+                        config: LogDriverLogStoreConfig =
+                          loadConfigOrThrow[ConsumerlessLogDriverLogStoreConfig]("whisk.logstore.log-driver"))
+    extends LogStore {
+  val logDriverMessage = config.message
+  val logParameters =
+    Map[String, Set[String]]("--log-driver" -> Set(config.dockerLogDriver)) ++
+      Map[String, Set[String]]("--log-opt" -> config.dockerLogDriverOpts)
```
markusthoemmes: Why use 2 `Map`s?

```scala
val logParameters = Map("--log-driver" -> Set(config.dockerLogDriver), "--log-opt" -> config.dockerLogDriverOpts)
```

Should work?

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -0,0 +1,167 @@
...
+              val jsObject = JsonParser(resultsString).asJsObject
+              //format of results is detailed here: http://docs.splunk.com/Documentation/Splunk/latest/RESTUM/RESTusing#Example_B:_JSON_response_format_example
+              val messages = jsObject
+                .fields("results")
+                .convertTo[JsArray]
+                .elements
+                .map(msgJsValue => {
+                  msgJsValue.asJsObject.fields(config.logMessageField).asInstanceOf[JsString].value
+                })
+              new ActivationLogs(messages)
```
markusthoemmes: No need for `new`

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -0,0 +1,167 @@
...
+              log.debug(s"splunk API results: ${resultsString}")
+              val jsObject = JsonParser(resultsString).asJsObject
+              //format of results is detailed here: http://docs.splunk.com/Documentation/Splunk/latest/RESTUM/RESTusing#Example_B:_JSON_response_format_example
+              val messages = jsObject
+                .fields("results")
+                .convertTo[JsArray]
+                .elements
+                .map(msgJsValue => {
+                  msgJsValue.asJsObject.fields(config.logMessageField).asInstanceOf[JsString].value
+                })
```
markusthoemmes: Try to never use `asInstanceOf`, there's usually a way around it. In this case:

```scala
resultsString.parseJson.convertTo[Seq[JsObject]].map(_.fields(config.logMessageField).convertTo[String])
```

Or to dense this down even more:

Use `Unmarshal` to do the heavy lifting for you, like so:

```scala
Unmarshal(response.entity).to[Seq[JsObject]].map(_.fields(config.logMessageField).convertTo[String])
```

tests/src/test/scala/whisk/core/containerpool/logging/LogDriverLogStoreTests.scala
```diff
@@ -0,0 +1,49 @@
...
+    "some message",
+    "fluentd",
+    Set("fluentd-address=localhost:24225", "tag=OW_CONTAINER"))
+
+  behavior of "LogDriver LogStore"
+
+  it should "fail when loading out of box configs (because whisk.logstore doesn't exist)" in {
+    assertThrows[ConfigReaderException[_]] {
+      val logDriverLogStore = new LogDriverLogStore(system)
+    }
```
markusthoemmes: FWIW there is: 

```scala
a[ConfigReaderException] should be thrownBy new LogDriverLogStore(system)`
```

ScalaTest's matchers usually result in better error messages vs. general asserts.

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -100,8 +100,8 @@ class SplunkLogStore(
         "exec_mode" -> "oneshot",
         "search" -> search,
         "output_mode" -> "json",
-        "earliest_time" -> formatter.format(activation.start),
-        "latest_time" -> formatter.format(activation.end))).toEntity
```
tysonnorris: I originally had this in there to limit the time range as much as possible for the splunk query, since the activation may have run days ago, lasted 10ms, and there may be millions of log events since then. 
I'm not positive that splunk does better with a narrower time range, but I assumed it does - WDYT? 
I could add it back in, but with end time + 5s to avoid any issues with collecting logs on short activations?

common/scala/src/main/scala/whisk/core/containerpool/logging/LogDriverLogStore.scala
```diff
@@ -21,34 +21,21 @@ import akka.actor.ActorSystem
...
-                        config: LogDriverLogStoreConfig =
-                          loadConfigOrThrow[ConsumerlessLogDriverLogStoreConfig]("whisk.logstore.log-driver"))
+                        config: ContainerArgsConfig =
+                          loadConfigOrThrow[ContainerArgsConfig]("whisk.container-factory.container-args"))
     extends LogStore {
-  val logDriverMessage = config.message
-  val logParameters = Map("--log-driver" -> Set(config.dockerLogDriver), "--log-opt" -> config.dockerLogDriverOpts)
 
-  override def containerParameters = logParameters.map(kv => (kv._1, kv._2.toSet)).toMap
+  override def containerParameters = config.extraArgs
```
tysonnorris: I think that loading the ContainerArgsConfig here will cause doubling of the args when the container launches? Historically, there were a handful of `docker run` args passed, in addition to special config handling for dns, network, and a couple others. Then LogStore SPI came, and added a couple other special handling. Then the ContainerArgsConfig provided "extraArgs" on top of those. 

So each of the ContainerArgsConfigs docker run flags should only be used in one place, I think, and separately there is a question in my mind about whether these should ALL reside under extraArgs (vs continuing to handle as "some args are special", which mostly means that those special ones should never be included in extraArgs).

So if we want people to add "--log-driver" via extra args, then LogDriverStore should just have an empty map for logParameters; and if we want people to use a special config for it, then it should not use extraArgs, but maybe either LogDriverConfig or ContainerArgsConfig.logDriverArgs.

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -0,0 +1,156 @@
...
+    val entity = FormData(
+      Map(
+        "exec_mode" -> "oneshot",
+        "search" -> search,
+        "output_mode" -> "json",
+        "earliest_time" -> formatter.format(activation.start),
+        "latest_time" -> formatter
+          .format(activation.end.plusSeconds(5)))).toEntity //add 5s to avoid a timerange of 0 on short-lived activations
+
+    log.debug("sending request")
```
markusthoemmes: Why not use `AkkaLogging` for consitency?

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -0,0 +1,156 @@
...
+        log.debug(s"splunk API response ${response}")
+        Unmarshal(response.entity)
+          .to[JsObject]
+          .map(r => {
+            ActivationLogs(
+              r.fields("results")
+                .convertTo[JsArray]
+                .elements
+                .map(_.asJsObject.fields(splunkConfig.logMessageField).convertTo[String]))
+          })
```
markusthoemmes: I still believe this should a bit more "modeled", as in:

```scala
case class SplunkResponse(results: Vector[JsObject])

...

Unmarshal(response.entity).to[SplunkResponse].map { r =>
  ActivationLogs(r.results.map(_.fields(splunkConfig.logMessageField).convertTo[String])
}
```

Gives you better compile-time guarantees and possibly better messages in an error case. WDYT?

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -0,0 +1,156 @@
...
+  override def fetchLogs(activation: WhiskActivation): Future[ActivationLogs] = {
+
+    //example curl request:
+    //    curl -u  username:password -k https://splunkhost:port/services/search/jobs -d exec_mode=oneshot -d output_mode=json -d "search=search index=\"someindex\" | spath=activation_id | search activation_id=a930e5ae4ad4455c8f2505d665aad282 |  table log_message" -d "earliest_time=2017-08-29T12:00:00" -d "latest_time=2017-10-29T12:00:00"
+    //example response:
+    //    {"preview":false,"init_offset":0,"messages":[],"fields":[{"name":"log_message"}],"results":[{"log_message":"some log message"}], "highlighted":{}}
+    val search =
+      s"""search index="${splunkConfig.index}"| spath ${splunkConfig.activationIdField}| search ${splunkConfig.activationIdField}=${activation.activationId.toString}| table ${splunkConfig.logMessageField}"""
+
+    val formatter = DateTimeFormatter.ofPattern("YYYY-MM-dd'T'hh:mm:ss").withZone(ZoneId.of("UTC"))
```
markusthoemmes: Isn't that exactly the format returned by `Instant.toString` anyway?

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -0,0 +1,156 @@
...
+class SplunkLogStore(
+  actorSystem: ActorSystem,
+  httpFlow: Option[Flow[(HttpRequest, Promise[HttpResponse]), (Try[HttpResponse], Promise[HttpResponse]), Any]] = None,
+  splunkConfig: SplunkLogStoreConfig = loadConfigOrThrow[SplunkLogStoreConfig]("whisk.logstore.splunk"))
+    extends LogDriverLogStore(actorSystem) {
+  implicit val as = actorSystem
+  implicit val ec = as.dispatcher
+  implicit val materializer = ActorMaterializer()
+
+  private val splunkApi = "/services/search/jobs" //see http://docs.splunk.com/Documentation/Splunk/6.6.3/RESTREF/RESTsearch#search.2Fjobs
```
markusthoemmes: Could use

```scala
private val splunkApi = Path / "services" / "search" / "jobs"
```

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * A Splunk based impl of LogDriverLogStore. Logs are routed to splunk via docker log driver, and retrieved via Splunk REST API
+ *
+ * @param actorSystem
+ * @param httpFlow Optional Flow to use for HttpRequest handling (to enable stream based tests)
+ */
+class SplunkLogStore(
+  actorSystem: ActorSystem,
+  httpFlow: Option[Flow[(HttpRequest, Promise[HttpResponse]), (Try[HttpResponse], Promise[HttpResponse]), Any]] = None,
+  splunkConfig: SplunkLogStoreConfig = loadConfigOrThrow[SplunkLogStoreConfig]("whisk.logstore.splunk"))
```
markusthoemmes: Please externalize the key to `ConfigKeys`

common/scala/build.gradle
```diff
@@ -43,6 +43,7 @@ dependencies {
     compile 'com.google.code.findbugs:jsr305:3.0.2'
     compile 'io.kamon:kamon-core_2.11:0.6.7'
     compile 'io.kamon:kamon-statsd_2.11:0.6.7'
+    compile 'com.github.pureconfig:pureconfig_2.11:0.8.0'
```
markusthoemmes: Should not be needed as it's already imported in this file?

common/scala/src/main/resources/application.conf
```diff
@@ -108,3 +108,15 @@ whisk {
...
+//    splunk {
+//        host = "splunk-host"
+//        port = 8080
+//        user = "splunk-user"
+//        password = "splunk-pass"
+//        index = "splunk-index"
+//        log-message-field = "log_message"
+//        activation-id-field = "activation_id"
+//    }
+//}
```
markusthoemmes: Why commented out? We should start building a `reference.conf` with sensible defaults. As long as the splunk logstore is not loaded, this will not take effect anyway?

core/invoker/src/main/resources/application.conf
```diff
@@ -27,4 +27,6 @@ whisk {
     extra-args: {}
 
   }
+  logstore {
+  }
```
markusthoemmes: Needed?

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
tysonnorris: I like it!

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
tysonnorris: I think `Instant` does not have the timezone.

markusthoemmes: It does, and it [defaults to UTC](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#now--) :)

common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
tysonnorris: Can, but have to use `Post(Uri(path = splunkApi))` in that case?

common/scala/src/main/resources/application.conf
tysonnorris: just for sample config, will remove till there is a better place to locate this

core/invoker/src/main/resources/application.conf
tysonnorris: not needed; will remove

PR #2959, opened 2017-11-13
remove hardcode prewarm nodejs:6 allow nodejs:default

csantanapr:


csantanapr:
currently working on `nodejs:8` this allow me to select which is prewarm based on ansible setting.

rabbah:
The indirection is dangerous here since you can end up affecting the setting for a majority of actions unintentially. Did you consider factoring out the prewarm kind(s) to ansible instead?

csantanapr:
@rabbah yes I considered moving to ansible.
Was thinking maybe an explicit kind selection, and not assume default for example a top level runtimeManifest property:
```yaml
runtimesManifest:
  prewarmKind:"nodejs:6"
```

what you think?

rabbah:
Instead of a new top level property you could add a property prewarm as part of the kind declaration. There is a nuance you also need to potentially declare the memory bound as well.

csantanapr:
Oh I forgot to reply.

Yes I agree, re defining the stem cell kind in configuring in runtimeManiFest JSON property, or another property, then can driven thru ansible or any config.

In terms of getting it done, 
not high priority, I thought I was going to be able to deploy in travis for the runtimes (i.e. swift, java) and not required nodejs runtime, but then I figured that the health protocol between controller and invoker requires a echo/hello action in nodejs.

I'm happy if you want to give it a stab @rabbah

PR #2960, opened 2017-11-13
Reduce the number of logs on info level

vvraskin:
This PR should provide a relief for controller while under load. We have reduced the number of log lines, which based on our performance measurements provides a 4x improvement on the throughput for long running tests and 2.5x throughput improvement for the short running tests. 

Here we also have made slight modification in the way log metrics are emitted. They could still be emitted in the following cases: 
* metricsLog is explicitely set to true
* the transaction timeout of 60 seconds is reached
* debug log level was selected
 
@mhenke1

### DIFF
--- a/common/scala/src/main/scala/whisk/common/TransactionId.scala
+++ b/common/scala/src/main/scala/whisk/common/TransactionId.scala
@@ -24,7 +24,7 @@ import java.util.concurrent.atomic.AtomicInteger
 
 import scala.math.BigDecimal.int2bigDecimal
 import scala.util.Try
-import akka.event.Logging.{InfoLevel, WarningLevel}
+import akka.event.Logging.{DebugLevel, InfoLevel, WarningLevel}
 import akka.event.Logging.LogLevel
 import spray.json._
 
@@ -53,11 +53,11 @@ case class TransactionId private (meta: TransactionMetadata) extends AnyVal {
    * @param message An additional message that is written into the log, together with the other information.
    * @param logLevel The Loglevel, the message should have. Default is <code>InfoLevel</code>.
    */
-  def mark(from: AnyRef, marker: LogMarkerToken, message: String = "", logLevel: LogLevel = InfoLevel)(
+  def mark(from: AnyRef, marker: LogMarkerToken, message: String = "", logLevel: LogLevel = DebugLevel)(
     implicit logging: Logging) = {
 
     if (TransactionId.metricsLog) {
-      logging.emit(logLevel, this, from, createMessageWithMarker(message, LogMarker(marker, deltaToStart)))
+      logging.emit(InfoLevel, this, from, createMessageWithMarker(message, LogMarker(marker, deltaToStart)))
     } else if (message.nonEmpty) {
       logging.emit(logLevel, this, from, message)
     }
@@ -79,11 +79,11 @@ case class TransactionId private (meta: TransactionMetadata) extends AnyVal {
    *
    * @return startMarker that has to be passed to the finished or failed method to calculate the time difference.
    */
-  def started(from: AnyRef, marker: LogMarkerToken, message: String = "", logLevel: LogLevel = InfoLevel)(
+  def started(from: AnyRef, marker: LogMarkerToken, message: String = "", logLevel: LogLevel = DebugLevel)(
     implicit logging: Logging): StartMarker = {
 
     if (TransactionId.metricsLog) {
-      logging.emit(logLevel, this, from, createMessageWithMarker(message, LogMarker(marker, deltaToStart)))
+      logging.emit(InfoLevel, this, from, createMessageWithMarker(message, LogMarker(marker, deltaToStart)))
     } else if (message.nonEmpty) {
       logging.emit(logLevel, this, from, message)
     }
@@ -107,7 +107,7 @@ case class TransactionId private (meta: TransactionMetadata) extends AnyVal {
   def finished(from: AnyRef,
                startMarker: StartMarker,
                message: String = "",
-               logLevel: LogLevel = InfoLevel,
+               logLevel: LogLevel = DebugLevel,
                endTime: Instant = Instant.now(Clock.systemUTC))(implicit logging: Logging) = {
 
     val endMarker =
@@ -116,7 +116,7 @@ case class TransactionId private (meta: TransactionMetadata) extends AnyVal {
 
     if (TransactionId.metricsLog) {
       logging.emit(
-        logLevel,
+        InfoLevel,
         this,
         from,
         createMessageWithMarker(message, LogMarker(endMarker, deltaToStart, Some(deltaToEnd))))
--- a/common/scala/src/main/scala/whisk/common/TransactionId.scala
+++ b/common/scala/src/main/scala/whisk/common/TransactionId.scala
@@ -57,6 +57,7 @@ case class TransactionId private (meta: TransactionMetadata) extends AnyVal {
     implicit logging: Logging) = {
 
     if (TransactionId.metricsLog) {
+      // marker received with a debug level will be emitted on info level
       logging.emit(InfoLevel, this, from, createMessageWithMarker(message, LogMarker(marker, deltaToStart)))
     } else if (message.nonEmpty) {
       logging.emit(logLevel, this, from, message)
@@ -83,6 +84,7 @@ case class TransactionId private (meta: TransactionMetadata) extends AnyVal {
     implicit logging: Logging): StartMarker = {
 
     if (TransactionId.metricsLog) {
+      // marker received with a debug level will be emitted on info level
       logging.emit(InfoLevel, this, from, createMessageWithMarker(message, LogMarker(marker, deltaToStart)))
     } else if (message.nonEmpty) {
       logging.emit(logLevel, this, from, message)
--- a/core/controller/src/main/scala/whisk/core/controller/ApiUtils.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/ApiUtils.scala
@@ -97,7 +97,7 @@ trait ReadOps extends Directives {
   protected def listEntities(list: Future[List[JsValue]])(implicit transid: TransactionId) = {
     onComplete(list) {
       case Success(entities) =>
-        logging.info(this, s"[LIST] entity success")
+        logging.debug(this, s"[LIST] entity success")
         complete(OK, entities)
       case Failure(t: Throwable) =>
         logging.error(this, s"[LIST] entity failed: ${t.getMessage}")
@@ -146,16 +146,16 @@ trait ReadOps extends Directives {
     ma: Manifest[A]) = {
     onComplete(factory.get(datastore, docid)) {
       case Success(entity) =>
-        logging.info(this, s"[GET] entity success")
+        logging.debug(this, s"[GET] entity success")
         postProcess map { _(entity) } getOrElse complete(OK, entity)
       case Failure(t: NoDocumentException) =>
-        logging.info(this, s"[GET] entity does not exist")
+        logging.debug(this, s"[GET] entity does not exist")
         terminate(NotFound)
       case Failure(t: DocumentTypeMismatchException) =>
-        logging.info(this, s"[GET] entity conformance check failed: ${t.getMessage}")
+        logging.debug(this, s"[GET] entity conformance check failed: ${t.getMessage}")
         terminate(Conflict, conformanceMessage)
       case Failure(t: ArtifactStoreException) =>
-        logging.info(this, s"[GET] entity unreadable")
+        logging.debug(this, s"[GET] entity unreadable")
         terminate(InternalServerError, t.getMessage)
       case Failure(t: Throwable) =>
         logging.error(this, s"[GET] entity failed: ${t.getMessage}")
@@ -183,16 +183,16 @@ trait ReadOps extends Directives {
     project: A => Future[JsObject])(implicit transid: TransactionId, format: RootJsonFormat[A], ma: Manifest[A]) = {
     onComplete(factory.get(datastore, docid)) {
       case Success(entity) =>
-        logging.info(this, s"[PROJECT] entity success")
+        logging.debug(this, s"[PROJECT] entity success")
         complete(OK, project(entity))
       case Failure(t: NoDocumentException) =>
-        logging.info(this, s"[PROJECT] entity does not exist")
+        logging.debug(this, s"[PROJECT] entity does not exist")
         terminate(NotFound)
       case Failure(t: DocumentTypeMismatchException) =>
-        logging.info(this, s"[PROJECT] entity conformance check failed: ${t.getMessage}")
+        logging.debug(this, s"[PROJECT] entity conformance check failed: ${t.getMessage}")
         terminate(Conflict, conformanceMessage)
       case Failure(t: ArtifactStoreException) =>
-        logging.info(this, s"[PROJECT] entity unreadable")
+        logging.debug(this, s"[PROJECT] entity unreadable")
         terminate(InternalServerError, t.getMessage)
       case Failure(t: Throwable) =>
         logging.error(this, s"[PROJECT] entity failed: ${t.getMessage}")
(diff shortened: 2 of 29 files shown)


style95:
+1 on this!!

daisy-ycguo:
+1. We do need to reduce logs to get better performance.

common/scala/src/main/scala/whisk/common/SimpleExec.scala
```diff
@@ -32,7 +32,7 @@ object SimpleExec {
    * @return a triple of (stdout, stderr, exitcode) from running the command
    */
   def syncRunCmd(cmd: Seq[String])(implicit transid: TransactionId, logging: Logging): (String, String, Int) = {
-    logging.info(this, s"Running command: ${cmd.mkString(" ")}")
+    logging.debug(this, s"Running command: ${cmd.mkString(" ")}")
```
rabbah: I think this is used only in the tests now - maybe it should be moved.

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -107,16 +107,20 @@ case class TransactionId private (meta: TransactionMetadata) extends AnyVal {
   def finished(from: AnyRef,
                startMarker: StartMarker,
                message: String = "",
-               logLevel: LogLevel = InfoLevel,
+               logLevel: LogLevel = DebugLevel,
                endTime: Instant = Instant.now(Clock.systemUTC))(implicit logging: Logging) = {
-
+    val timeout = 60
```
rabbah: suggest moving this outward to the singleton object.

common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
```diff
@@ -68,7 +68,7 @@ class KafkaProducerConnector(kafkahost: String,
 
   /** Closes producer. */
   override def close() = {
-    logging.info(this, "closing producer")
+    logging.debug(this, "closing producer")
```
rabbah: this is not an event that happens often - I would leave it at info level.

core/controller/src/main/scala/whisk/core/controller/ApiUtils.scala
```diff
@@ -166,16 +166,16 @@ trait ReadOps extends Directives {
                                                                                         ma: Manifest[A]) = {
     onComplete(factory.get(datastore, docid)) {
       case Success(entity) =>
-        logging.info(this, s"[GET] entity success")
+        logging.debug(this, s"[GET] entity success")
         postProcess map { _(entity) } getOrElse complete(OK, entity)
       case Failure(t: NoDocumentException) =>
-        logging.info(this, s"[GET] entity does not exist")
+        logging.warn(this, s"[GET] entity does not exist")
```
rabbah: the use of `warn` here will lead to a lot of spurious warnings that can be caused by bad client requests. I suggest using `debug` instead. This will apply to several of the error cases below.

core/controller/src/main/scala/whisk/core/controller/Rules.scala
```diff
@@ -152,13 +152,13 @@ trait WhiskRulesApi extends WhiskCollectionAPI with ReferencedEntities {
...
-                  logging.info(this, s"[POST] rule update conflict")
+                  logging.warn(this, s"[POST] rule update conflict")
                   terminate(Conflict, conflictMessage)
                 case IgnoredRuleActivation(ok) =>
-                  logging.info(this, s"[POST] rule update ignored")
+                  logging.warn(this, s"[POST] rule update ignored")
                   if (ok) complete(OK) else terminate(Conflict)
                 case _: NoDocumentException =>
-                  logging.info(this, s"[POST] the trigger attached to the rule doesn't exist")
+                  logging.warn(this, s"[POST] the trigger attached to the rule doesn't exist")
```
rabbah: Here also the use of `warn` is concerning. I think these should be reserved to system invariant warnings not errors that can be caused by the client. This is why these are at info level. Suggest `debug` instead.

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -215,6 +223,7 @@ object TransactionId {
   val loadbalancer = TransactionId(-120) // Loadbalancer thread
   val invokerHealth = TransactionId(-121) // Invoker supervision
   val controller = TransactionId(-130) // Controller startup
+  val longTransactionWarningTime = 60 // Threshold in seconds to log long-running transactions
```
markusthoemmes: `val longTransactionWarningTime = 60.seconds.toMillis` and then you can remove the arbitrary `* 1000` above.

core/controller/src/main/scala/whisk/core/controller/ApiUtils.scala
```diff
@@ -353,22 +353,22 @@ trait WriteOps extends Directives {
...
-        logging.info(this, s"[DEL] entity rejected with code $code: $message")
+        logging.debug(this, s"[DEL] entity rejected with code $code: $message")
         terminate(code, message)
       case Failure(t: DocumentTypeMismatchException) =>
-        logging.info(this, s"[DEL] entity conformance check failed: ${t.getMessage}")
+        logging.debug(this, s"[DEL] entity conformance check failed: ${t.getMessage}")
         terminate(Conflict, conformanceMessage)
       case Failure(t: ArtifactStoreException) =>
-        logging.info(this, s"[DEL] entity unreadable")
+        logging.debug(this, s"[DEL] entity unreadable")
```
markusthoemmes: This should be a proper `ERROR`, isn't it?

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -303,15 +303,17 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
   private def checkConcurrentUserThrottle(user: Identity, right: Privilege, resources: Set[Resource])(
     implicit transid: TransactionId): Future[Unit] = {
     if (right == ACTIVATE && resources.exists(_.collection.path == Collection.ACTIONS)) {
-      checkThrottleOverload(concurrentInvokeThrottler.check(user))
+      checkThrottleOverload(concurrentInvokeThrottler.check(user), user)
```
markusthoemmes: You don't need to pass the user through just for logging. Just use the `andThen` combinator:

```scala
checkThrottleOverload(concurrentInvokeThrottler.check(user)).andThen {
  case Failure(_) => logging.debug(this, s"user '${user.subject}' has exceeded its throttle limit")
}
```

vvraskin: What you propose is a valid option, however I'd like to leave it as it is to avoid duplicating the log entries. Since we call `checkThrottleOverload` at three places I think that would be better to keep the log entry inside of the method. You could argue that we can create three different log messages for each type of user throttle, but since the message on the debug level, I don't believe that would bring extra value. Passing user as a parameter should not add much overhead, right?

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -53,11 +53,11 @@ case class TransactionId private (meta: TransactionMetadata) extends AnyVal {
    * @param message An additional message that is written into the log, together with the other information.
    * @param logLevel The Loglevel, the message should have. Default is <code>InfoLevel</code>.
    */
-  def mark(from: AnyRef, marker: LogMarkerToken, message: String = "", logLevel: LogLevel = InfoLevel)(
+  def mark(from: AnyRef, marker: LogMarkerToken, message: String = "", logLevel: LogLevel = DebugLevel)(
     implicit logging: Logging) = {
 
     if (TransactionId.metricsLog) {
-      logging.emit(logLevel, this, from, createMessageWithMarker(message, LogMarker(marker, deltaToStart)))
+      logging.emit(InfoLevel, this, from, createMessageWithMarker(message, LogMarker(marker, deltaToStart)))
```
rabbah: a comment here is advisable: even though the mark log message is potentially logged at debug level, the time stamp is emitted at info level.

core/controller/src/main/scala/whisk/core/controller/Authenticate.scala
```diff
@@ -59,25 +59,25 @@ trait Authenticate {
       Try {
         // authkey deserialization is wrapped in a try to guard against malformed values
         val authkey = AuthKey(UUID(pw.username), Secret(pw.password))
-        logging.info(this, s"authenticate: ${authkey.uuid}")
+        logging.debug(this, s"authenticate: ${authkey.uuid}")
```
rabbah: this is useful to be able to correlate a tid to a uuid. i would revert this one.

core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
```diff
@@ -48,7 +48,7 @@ class ActivationThrottler(loadBalancer: LoadBalancer, defaultConcurrencyLimit: I
   def check(user: Identity)(implicit tid: TransactionId): Future[RateLimit] = {
     loadBalancer.activeActivationsFor(user.uuid).map { concurrentActivations =>
       val concurrencyLimit = user.limits.concurrentInvocations.getOrElse(defaultConcurrencyLimit)
-      logging.info(
+      logging.debug(
```
rabbah: these proved helpful in the past to know if a uuid was over its quota. can we revert?

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
rabbah: i think this is ok, but apropos my message above, having an info level message that user is over quota is helpful.
perhaps rather than user.subject, should the user.uuid instead.

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -53,11 +53,11 @@ case class TransactionId private (meta: TransactionMetadata) extends AnyVal {
    * @param message An additional message that is written into the log, together with the other information.
    * @param logLevel The Loglevel, the message should have. Default is <code>InfoLevel</code>.
    */
-  def mark(from: AnyRef, marker: LogMarkerToken, message: String = "", logLevel: LogLevel = InfoLevel)(
+  def mark(from: AnyRef, marker: LogMarkerToken, message: String = "", logLevel: LogLevel = DebugLevel)(
```
rabbah: i see in a few places below that you set the log level on the markers for to info level.
i hesitate with this global change - how many markers are affected and will now not log a message - can we avoid it?

core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
rabbah: great that will be helpful, is that in this pr or subsequent pr?

core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
```diff
@@ -60,7 +60,7 @@ class ActivationThrottler(loadBalancer: LoadBalancer, defaultConcurrencyLimit: I
    */
   def isOverloaded()(implicit tid: TransactionId): Future[Boolean] = {
     loadBalancer.totalActiveActivations.map { concurrentActivations =>
-      logging.info(
+      logging.debug(
```
rabbah: same for this message - if system is overloaded, log it.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -160,10 +160,10 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
    */
   protected[core] def checkThrottles(user: Identity)(implicit transid: TransactionId): Future[Unit] = {
 
-    logging.info(this, s"checking user '${user.subject}' has not exceeded activation quota")
+    logging.debug(this, s"checking user '${user.subject}' has not exceeded activation quota")
     checkSystemOverload(ACTIVATE)
-      .flatMap(_ => checkThrottleOverload(Future.successful(invokeRateThrottler.check(user))))
-      .flatMap(_ => checkThrottleOverload(concurrentInvokeThrottler.check(user)))
+      .flatMap(_ => checkThrottleOverload(Future.successful(invokeRateThrottler.check(user)), user))
+      .flatMap(_ => checkThrottleOverload(concurrentInvokeThrottler.check(user), user))
```
rabbah: could do the logging here since this goes through all the throttles.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -328,15 +328,17 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
...
   }
 
-  private def checkThrottleOverload(throttle: Future[RateLimit])(implicit transid: TransactionId): Future[Unit] = {
+  private def checkThrottleOverload(throttle: Future[RateLimit], user: Identity)(
+    implicit transid: TransactionId): Future[Unit] = {
     throttle.flatMap { limit =>
       if (limit.ok) {
         Future.successful(())
       } else {
+        logging.info(this, s"user '${user.namespace}' has exceeded its throttle limit")
```
rabbah: i'd drop "user" prefix, ` +        logging.info(this, s"'${user.namespace}' has exceeded its throttle limit") `

rabbah: here we don't know if it's the concurrent limit, or per minute limit - anyway to easily recover that? perhaps the RateLimit can have a descriptor to emit here?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
```diff
@@ -107,7 +107,7 @@ class DockerContainerFactory(config: WhiskConfig,
   private def removeAllActionContainers(): Unit = {
     implicit val transid = TransactionId.invoker
     val cleaning = docker.ps(filters = Seq("name" -> s"wsk${instance.toInt}_"), all = true).flatMap { containers =>
-      logging.info(this, s"removing ${containers.size} action containers.")
+      logging.debug(this, s"removing ${containers.size} action containers.")
```
rabbah: this is done at startup and shutdown, i think leave at info level.

core/controller/src/main/scala/whisk/core/entitlement/ActivationThrottler.scala
rabbah: ok i see it later on; see related comment there.

common/scala/src/main/scala/whisk/common/TransactionId.scala
rabbah: after looking through all the uses of `mark` this change is ok. 
here are the remaining ones for reference.

```
core/controller/src/main/scala/whisk/core/controller/Controller.scala:  TransactionId.controller.mark(
core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala:      transid.mark(
core/controller/src/main/scala/whisk/core/loadBalancer/InvokerSupervision.scala:      transid.mark(
common/scala/src/main/scala/whisk/common/TransactionId.scala:  def mark(from: AnyRef, marker: LogMarkerToken, message: String = "", logLevel: LogLevel = DebugLevel)(
```

rabbah: @vvraskin can you set the ones in the invoker supervisor to info level also? these should be infrequent.

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -338,7 +338,7 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
       if (limit.ok) {
         Future.successful(())
       } else {
-        logging.info(this, s"user '${user.namespace}' has exceeded its throttle limit")
+        logging.info(this, s"'${user.namespace}' has exceeded its throttle limit, ${limit.errorMsg}")
```
rabbah: nice - thanks!

common/scala/src/main/scala/whisk/common/TransactionId.scala
vvraskin: But warn level should be also fine right? If info is enabled, warn messages should be printed as well

rabbah: Ah - you're right. Thanks.

PR #2962, opened 2017-11-13
Clean up all the test cases

houshengbo:
The PRs before replace all the rest of the wsk inplementation with REST
implementation for test cases in openwhisk. This PR cleans up some
leftover for the WskRest migration work. After this PR, no more test
case needs to be changed any longer, and cli in openwhisk core is ready
to departure.


houshengbo:
PG 4 1010 launched.

PR #2963, opened 2017-11-13
Completely settle throttle after the API GW test suite is run

mdeuser:
- to avoid affecting any downstream tests that are action invocation heavy


mdeuser:
pg1/2270

PR #2964, opened 2017-11-13
Reject requests to create packages having the name 'default'

### LINKED ISSUE #1785: make package name "default" reserved

mdeuser:
Updates to existing packages named "default" are still permitted.

Fixes #1785

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Packages.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Packages.scala
@@ -76,25 +76,30 @@ trait WhiskPackagesApi extends WhiskCollectionAPI with ReferencedEntities {
    */
   override def create(user: Identity, entityName: FullyQualifiedEntityName)(implicit transid: TransactionId) = {
     parameter('overwrite ? false) { overwrite =>
-      entity(as[WhiskPackagePut]) { content =>
-        val request = content.resolve(entityName.namespace)
+      if (!overwrite && entityName.name.toString == "default") {
+        logging.info(this, "attempt to create a package named 'default'")
+        terminate(BadRequest, Messages.packageDefaultIsReserved)
+      } else {
+        entity(as[WhiskPackagePut]) { content =>
+          val request = content.resolve(entityName.namespace)
 
-        request.binding.map { b =>
-          logging.info(this, "checking if package is accessible")
-        }
-        val referencedentities = referencedEntities(request)
+          request.binding.map { b =>
+            logging.info(this, "checking if package is accessible")
+          }
+          val referencedentities = referencedEntities(request)
 
-        onComplete(entitlementProvider.check(user, Privilege.READ, referencedentities)) {
-          case Success(_) =>
-            putEntity(
-              WhiskPackage,
-              entityStore,
-              entityName.toDocId,
-              overwrite,
-              update(request) _,
-              () => create(request, entityName))
-          case Failure(f) =>
-            rewriteEntitlementFailure(f)
+          onComplete(entitlementProvider.check(user, Privilege.READ, referencedentities)) {
+            case Success(_) =>
+              putEntity(
+                WhiskPackage,
+                entityStore,
+                entityName.toDocId,
+                overwrite,
+                update(request) _,
+                () => create(request, entityName))
+            case Failure(f) =>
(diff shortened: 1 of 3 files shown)


mdeuser:
pg3/1333

core/controller/src/main/scala/whisk/core/controller/Packages.scala
```diff
@@ -76,25 +76,30 @@ trait WhiskPackagesApi extends WhiskCollectionAPI with ReferencedEntities {
    */
   override def create(user: Identity, entityName: FullyQualifiedEntityName)(implicit transid: TransactionId) = {
     parameter('overwrite ? false) { overwrite =>
-      entity(as[WhiskPackagePut]) { content =>
-        val request = content.resolve(entityName.namespace)
+      if (!overwrite && entityName.name.toString == "default") {
```
rabbah: iirc there’s an asString method which is preferred.

core/controller/src/main/scala/whisk/core/controller/Packages.scala
```diff
@@ -76,25 +76,30 @@ trait WhiskPackagesApi extends WhiskCollectionAPI with ReferencedEntities {
...
+        logging.info(this, "attempt to create a package named 'default'")
```
rabbah: We can probably do away with this log message or use debug level

tests/src/test/scala/whisk/core/controller/test/PackagesApiTests.scala
```diff
@@ -370,6 +370,20 @@ class PackagesApiTests extends ControllerTestCommon with WhiskPackagesApi {
     }
   }
 
+  it should "reject create package when package name is 'default'" in {
```
rabbah: Is there a test for update of an existing package named default?

tests/src/test/scala/whisk/core/controller/test/PackagesApiTests.scala
```diff
@@ -370,6 +370,20 @@ class PackagesApiTests extends ControllerTestCommon with WhiskPackagesApi {
...
+    implicit val tid = transid()
+    def defname() = MakeName.next("default")
+    val provider = WhiskPackage(namespace, defname(), None)
+    Put(s"$collectionPath/${provider.name}") ~> Route.seal(routes(creds)) ~> check {
+      status should be(BadRequest)
+      val response = responseAs[String]
+      println(s"Default failure response: '${response}'")
```
rabbah: Printing not necessary.

core/controller/src/main/scala/whisk/core/controller/Packages.scala
rabbah: I’d move the string default to a const on the whisk package Singleton.

mdeuser: :+1:  at the same time, i'll make it an array of reserved names that can be expanded as needed

tests/src/test/scala/whisk/core/controller/test/PackagesApiTests.scala
mdeuser: not currently; i'll add one

rabbah:
Hmm I think this update logic is incorrect - it would allow me to bypass the reserved names by just doing a PUT force=true. I believe the intention was to allow this _only_ if a package already exists with that name and otherwise to reject it. @mdeuser can you confirm and open a defect if you agree.

mdeuser:
@rabbah - yep.  a package update will allow the "default" name for the package.  need to add a check to confirm whether or not the package exists.  hopefully something simple exists :-)

dubeejw:
Looks like we can still create a package with the default name if the package is created by `wsk action update`.

dubeejw:
@mdeuser, seems like if we won't let a package to be created named as `default` we shouldn't let packages with the `default` name be updated either?

mdeuser:
@dubeejw - yep.  @rabbah mentioned the same issue.  I opened https://github.com/apache/incubator-openwhisk/pull/2964 to address it

mdeuser:
@dubeejw - if a package exists with a reserved name, i believe the intent is not to block updates to that package.

PR #2968, opened 2017-11-15
add additional metrics and logs

mhenke1:
This PR ads two additional metrics:
- loadbalancer_activations_<namespaceId>_count : counter for activation starts in load balancer
- invoker_container_start_<containerState>_<namespace>_<action>_ count

I also adds another info level log statement with container start information

### DIFF
--- a/common/scala/src/main/scala/whisk/common/Logging.scala
+++ b/common/scala/src/main/scala/whisk/common/Logging.scala
@@ -188,17 +188,18 @@ object MetricEmitter {
 
   val metrics = Kamon.metrics
 
-  def emitCounterMetric(token: LogMarkerToken) = {
+  def emitCounterMetric(token: LogMarkerToken): Unit = {
     metrics
       .counter(token.toString)
       .increment(1)
   }
 
-  def emitHistogramMetric(token: LogMarkerToken, value: Long) = {
+  def emitHistogramMetric(token: LogMarkerToken, value: Long): Unit = {
     metrics
       .histogram(token.toString)
       .record(value)
   }
+
 }
 
 object LoggingMarkers {
@@ -238,6 +239,8 @@ object LoggingMarkers {
   // Check invoker healthy state from loadbalancer
   val LOADBALANCER_INVOKER_OFFLINE = LogMarkerToken(loadbalancer, "invokerOffline", count)
   val LOADBALANCER_INVOKER_UNHEALTHY = LogMarkerToken(loadbalancer, "invokerUnhealthy", count)
+  def LOADBALANCER_ACTIVATION_START(namespaceId: String) =
+    LogMarkerToken(loadbalancer, s"activations_$namespaceId", count)
 
   // Time that is needed to execute the action
   val INVOKER_ACTIVATION_RUN = LogMarkerToken(invoker, "activationRun", start)
@@ -252,6 +255,8 @@ object LoggingMarkers {
   val INVOKER_ACTIVATION = LogMarkerToken(invoker, activation, start)
   def INVOKER_DOCKER_CMD(cmd: String) = LogMarkerToken(invoker, s"docker.$cmd", start)
   def INVOKER_RUNC_CMD(cmd: String) = LogMarkerToken(invoker, s"runc.$cmd", start)
+  def INVOKER_CONTAINER_START(actionName: String, namespaceName: String, containerState: String) =
+    LogMarkerToken(invoker, s"container_start_${containerState}_${namespaceName}_$actionName", count)
 
   /*
    * General markers
@@ -265,4 +270,5 @@ object LoggingMarkers {
   val DATABASE_QUERY = LogMarkerToken(database, "queryView", start)
   val DATABASE_ATT_GET = LogMarkerToken(database, "getDocumentAttachment", start)
   val DATABASE_ATT_SAVE = LogMarkerToken(database, "saveDocumentAttachment", start)
+  val DATABASE_BATCH_SIZE = LogMarkerToken(database, "batchSize", count)
 }
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/ContainerPoolBalancer.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/ContainerPoolBalancer.scala
@@ -163,7 +163,8 @@ class ContainerPoolBalancer(config: WhiskConfig, instance: InstanceId)(implicit 
         transid.mark(
           this,
           LoggingMarkers.LOADBALANCER_ACTIVATION_START(namespaceId.asString),
-          s"loadbalancer: activation started for namespace $namespaceId")
+          s"loadbalancer: activation started for namespace $namespaceId and activation $activationId",
+          logLevel = InfoLevel)
 
         // please note: timeoutHandler.cancel must be called on all non-timeout paths, e.g. Success
         ActivationEntry(
@@ -210,8 +211,7 @@ class ContainerPoolBalancer(config: WhiskConfig, instance: InstanceId)(implicit 
     val start = transid.started(
       this,
       LoggingMarkers.CONTROLLER_KAFKA,
-      s"posting topic '$topic' with activation id '${msg.activationId}'",
-      logLevel = InfoLevel)
+      s"posting topic '$topic' with activation id '${msg.activationId}'")
 
     producer.send(topic, msg).andThen {
       case Success(status) =>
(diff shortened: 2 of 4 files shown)


core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -188,6 +189,9 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
         processCompletion(Left(activationId), transid, forced = true, invoker = invokerName)
       }
 
+      if (TransactionId.metricsKamon) {
+        MetricEmitter.emitCounterMetric(LoggingMarkers.LOADBALANCER_ACTIVATION_START(namespaceId.toString))
```
markusthoemmes: Why not use `transid.mark` here?

mhenke1:
@markusthoemmes I have changed the code to use transid.mark.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -78,27 +76,56 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
...
+    val namespaceName = r.msg.user.namespace.name
+    val actionName = r.action.name.name
+    val activationId = r.msg.activationId.toString
+
+    logging.info(
+      this,
+      s"containerStart containerState: $containerState action: $actionName namespace: $namespaceName activationId: $activationId")(
+      r.msg.transid)
+    MetricEmitter.emitCounterMetric(LoggingMarkers.INVOKER_CONTAINER_START(actionName, namespaceName, containerState))
+  }
```
markusthoemmes: Why not `transid.mark`?

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -190,13 +190,20 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
...
-      }
+    loadBalancerData.putActivation(
+      activationId, {
+        actorSystem.scheduler.scheduleOnce(timeout) {
+          processCompletion(Left(activationId), transid, forced = true, invoker = invokerName)
+        }
 
-      ActivationEntry(activationId, namespaceId, invokerName, Promise[Either[ActivationId, WhiskActivation]]())
-    })
+        if (TransactionId.metricsKamon) {
```
markusthoemmes: Why the extra if?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
mhenke1: Code evolvement reasons. We added the logging out output late (and with it the transaction id) and did no realize that we now were able to merge the two things using txid.mark. Corrected.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
mhenke1: Removed

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -248,6 +251,8 @@ object LoggingMarkers {
   val INVOKER_ACTIVATION = LogMarkerToken(invoker, activation, start)
   def INVOKER_DOCKER_CMD(cmd: String) = LogMarkerToken(invoker, s"docker.$cmd", start)
   def INVOKER_RUNC_CMD(cmd: String) = LogMarkerToken(invoker, s"runc.$cmd", start)
+  def INVOKER_CONTAINER_START(actionName: String, namespaceName: String, containerState: String) =
+    LogMarkerToken(invoker, s"container_start_${containerState}_${namespaceName}_$actionName", count)
```
rabbah: will this allow you to count containers started for a namespace vs ones specific to an action? ie can you aggregate ignoring the action?

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -186,13 +187,19 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
...
+      activationId, {
+        actorSystem.scheduler.scheduleOnce(timeout) {
+          processCompletion(Left(activationId), transid, forced = true, invoker = invokerName)
+        }
 
-      ActivationEntry(activationId, namespaceId, invokerName, Promise[Either[ActivationId, WhiskActivation]]())
-    })
+        transid.mark(
+          this,
+          LoggingMarkers.LOADBALANCER_ACTIVATION_START(namespaceId.toString),
```
rabbah: use `asString` instead which is the convention we use throughout.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -78,27 +73,56 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
...
+
   def receive: Receive = {
     // A job to run on a container
     case r: Run =>
       val container = if (busyPool.size < maxActiveContainers) {
+
         // Schedule a job to a warm container
         ContainerPool
           .schedule(r.action, r.msg.user.namespace, freePool)
+          .map(container => {
```
rabbah: is this better done with an `andThen` @markusthoemmes ?

markusthoemmes: These are all `Option`, not `Future`. We unfortunately cannot move this outside of the innerts of the logic (now) because at the end you just get an ActorRef. You cannot determine from which branch it was.

We could, of course, create wrapping types for each possible return and log based on that at the very end of the decision.

common/scala/src/main/scala/whisk/common/Logging.scala
mhenke1: @rabbah Grafana Yes. Grafana allows to aggregate values on any 'regexable' - sub-string of the metric name

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
mhenke1: @markusthoemmes @rabbah I have moved the logging to the end of the method. I have used a simple tuple to transport the container state.

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -238,6 +239,8 @@ object LoggingMarkers {
   // Check invoker healthy state from loadbalancer
   val LOADBALANCER_INVOKER_OFFLINE = LogMarkerToken(loadbalancer, "invokerOffline", count)
   val LOADBALANCER_INVOKER_UNHEALTHY = LogMarkerToken(loadbalancer, "invokerUnhealthy", count)
+  def LOADBALANCER_ACTIVATION_START(namespaceId: String) =
```
rabbah: should this take a string type than string? entity name or identity perhaps.

common/scala/src/main/scala/whisk/common/Logging.scala
rabbah: same comment re param typing.

core/controller/src/main/scala/whisk/core/loadBalancer/ContainerPoolBalancer.scala
```diff
@@ -158,6 +160,11 @@ class ContainerPoolBalancer(config: WhiskConfig, instance: InstanceId)(implicit
           processCompletion(Left(activationId), transid, forced = true, invoker = invokerName)
         }
 
+        transid.mark(
+          this,
+          LoggingMarkers.LOADBALANCER_ACTIVATION_START(namespaceId.asString),
+          s"loadbalancer: activation started for namespace $namespaceId")
```
rabbah: do we need this vs 
```
val start = transid.started(
      this,
      LoggingMarkers.CONTROLLER_KAFKA,
      s"posting topic '$topic' with activation id '${msg.activationId}'",
      logLevel = InfoLevel)
```
?

common/scala/src/main/scala/whisk/common/Logging.scala
markusthoemmes: The alternative type here would be `EntityName` and I don't see much value in adding those.

common/scala/src/main/scala/whisk/common/Logging.scala
markusthoemmes: Same as below, would be `EntityName`. Is that much more valuable?

mhenke1:
@rabbah I have consolidated the two log messages when starting an activation into one that looks like this
```
[2018-02-05T10:49:55.757Z] [INFO] [#tid_3] [ContainerPoolBalancer] loadbalancer: activation started for namespace 23bc46b1-71f6-4ed5-8c54-816aa4f8c502 and activation d72c63deb6a84562ac63deb6a8f56244
```

Regarding the type related comment I will leave the parameters at 'String'  for now.

PR #2969, opened 2017-11-15
Fix the permission issues when creating the directories

houshengbo:


houshengbo:
@rabbah We can do it by:
```
ansible-playbook -i environments/local openwhisk.yml -e installation_mode=remote;  to download binaries.

ansible-playbook -i environments/local openwhisk.yml -e installation_mode=local; to copy local binaries.
```

rabbah:
fantastic - thank you!

ansible/group_vars/all
```diff
@@ -303,16 +303,16 @@ openwhisk_cli_tag: "{{ lookup('ini', 'git_tag section=openwhisk-cli file={{ open
 # remote mode or location to save the binaries of the OpenWhisk CLI in local mode.
 #
 
+installation_mode: remote
```
csantanapr: This should not be outside top level, it should be inside openwhisk_cli object.

csantanapr: Now that I look what you are doing I think we need more specific name and also a way to override via en environment.
How about this:
```
cli_installation_mode_default: local
cli_installation_mode: "{{ openwhisk_cli_installation_mode | default(cli_installation_mode_default) }}"
```
Then for example ins `environments/local/group_vars/all it can override via
```
openwhisk_cli_installation_mode: remote
```
In addition via cmd flag
```
ansible-playbook -i environments/local openwhisk.yml -e cli_installation_mode=remote;  to download binaries.

ansible-playbook -i environments/local openwhisk.yml -e cli_installation_mode=local; to copy local binaries.
```

csantanapr: Set the default to `local` we are not ready for `remote`, the `remote` will be together with the remove of cli source

ansible/roles/cli/tasks/copy_local_openwhisk_cli.yml
```diff
@@ -5,6 +5,20 @@
 - local_action: stat path={{ openwhisk_cli.local.location }}/{{ item }}
   register: binary_path
 
+- name: "ensure Nginx cli directory is writable"
+  file:
+    path: "{{ cli.nginxdir }}"
+    state: directory
+    mode: 0777
+  become: true
```
csantanapr: do not hardcode `become` use variable `cli.dir.become`
```
name: "ensure Nginx cli directory is writable"
   file:
   path: "{{ cli.nginxdir }}"
   state: directory
   mode: 0777
   become: "{{ cli.dir.become }}"
```

ansible/roles/cli/tasks/copy_local_openwhisk_cli.yml
```diff
@@ -5,6 +5,20 @@
...
+
 - name: "copy the local binaries from a local directory to Nginx directory"
   copy: src={{ openwhisk_cli.local.location }}/{{ item }} dest={{ cli.nginxdir }}/
   when: binary_path.stat.exists
+
+- name: "modify the mode of Nginx cli directory"
```
csantanapr: remove this since you are already setting permission to `0777` before on the same directory

ansible/roles/cli/tasks/download_openwhisk_cli.yml
```diff
@@ -48,6 +48,8 @@
   file:
     path: "{{ cli.nginxdir }}/{{ os }}/{{ arc }}"
     state: directory
+    mode: 0777
+  become: true
```
csantanapr: do not hardcode `become` use variable `cli.dir.become`
```
name: "ensure Nginx cli directory is writable"
   file:
   path: "{{ cli.nginxdir }}"
   state: directory
   mode: 0777
   become: "{{ cli.dir.become }}"
```

ansible/roles/cli/tasks/download_openwhisk_cli.yml
```diff
@@ -69,3 +71,9 @@
     chdir: "{{ cli.nginxdir }}/{{ os }}/{{ arc }}/"
   when: os == "linux"
 
+- name: "modify the mode of Nginx cli directory for ({{ os }}) {{ arc }}"
```
csantanapr: remove this since you are already setting permission to `0777` before on the same directory

houshengbo: It is good to keep the folder 0777, which means anyone can do anything. We set it to 0777 in order to upload the binaries. After it is done, do we still keep the 0777?

ansible/roles/cli/tasks/copy_local_openwhisk_cli.yml
houshengbo: Is it good to keep the folder 0777, which means anyone can do anything? We set it to 0777 in order to upload the binaries. After it is done, do we still keep the 0777?

ansible/roles/cli/tasks/copy_local_openwhisk_cli.yml
houshengbo: done

houshengbo:
@csantanapr The travis passed.

PR #2973, opened 2017-11-16
Change the confdir for the default environments.

cbickel:
This makes it much easier to delete all configuration with only one command.

### DIFF
--- a/ansible/environments/distributed/group_vars/all
+++ b/ansible/environments/distributed/group_vars/all
@@ -8,7 +8,7 @@ db_host: "{{ groups['db'] | first }}"
 db_prefix: whisk_distributed_
 
 whisk_version_name: distributed
-config_root_dir: /tmp
+config_root_dir: /tmp/wskconf
 whisk_logs_dir: /tmp/wsklogs
 registry_storage_dir: "/"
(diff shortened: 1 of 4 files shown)


rabbah:
FWIW we should docker cp the nginx comfy and stop mounting /tmp which has issues for vagrant and reboot where tmp gets wiped.

rabbah:
See https://github.com/apache/incubator-openwhisk/issues/32

csantanapr:
In general /tmp is a bad place to use as a persistent 
Maybe using a different mount point like /$HOME/.openwhisk/edgeconf
This is assuming that when the system reboots and dockers tries to start edge container works to mount the directory

cbickel:
Do you see this as part of this PR?

csantanapr:
Not required for this PR
This PR is a step forward, we can deal with default location or using cp in another PR

I’m not sure on using cp as I don’t understand as it will required for ansible to create container then to de the cp and then do a restart to pickup the new config? How would this trick will be done in docker compose or kube without requiring to build a new image with the config inside
Maybe the nginx supports passing a config string via CMD or env variable?

rabbah:
It’s just a cp no restart or new container.

PR #2974, opened 2017-11-16
Add LogStore which stores to database and file simultaneously.

markusthoemmes:
Includes #2878.

Adds a new LogStore which reads the whole file in memory and stores the logs in the activation entry. At the same time it pipes all user-written logs to a file in an enriched format (which is why the extra fields need to be added to the interface).

The written file can then be read by any other instance asynchronously.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
@@ -1,0 +1,156 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.containerpool.logging
+
+import java.nio.file.{Path, Paths}
+
+import akka.NotUsed
+import akka.actor.ActorSystem
+import akka.stream.alpakka.file.scaladsl.LogRotatorSink
+import akka.stream.{Graph, SinkShape, UniformFanOutShape}
+import akka.stream.scaladsl.{Broadcast, Flow, GraphDSL, Keep, MergeHub, Sink, Source}
+import akka.util.ByteString
+import whisk.common.TransactionId
+import whisk.core.containerpool.Container
+import whisk.core.entity.{ActivationLogs, ExecutableWhiskAction, Identity, WhiskActivation}
+import whisk.core.entity.size._
+import spray.json._
+import spray.json.DefaultJsonProtocol._
+import java.time.Instant
+
+import whisk.http.Messages
+
+import scala.concurrent.Future
+
+/**
+ * Docker based implementation of a LogStore.
+ *
+ * Relies on docker's implementation details with regards to the JSON log-driver. When using the JSON log-driver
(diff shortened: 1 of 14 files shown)


markusthoemmes:
@sven-lange-last I'd appreciate your 👀 on this PR.

markusthoemmes:
PG2 2443 🔵

common/scala/src/main/scala/whisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
```diff
@@ -0,0 +1,148 @@
...
+      }
+      .mkString(",")
+
+  /**
+   * Merges all file-writing streams into one globally buffered stream.
+   *
+   * This effectively decouples the time it takes to {@code collectLogs} from the time it takes to write the augmented
+   * logging data to a file on the disk.
+   */
+  val bufferSize = 100.MB
```
rabbah: should this be pulled out to a deployment config?

common/scala/src/main/scala/whisk/core/containerpool/logging/DockerToActivationLogStore.scala
```diff
@@ -67,16 +67,17 @@ class DockerLogStore(system: ActorSystem) extends LogStore {
...
 
-    val possibleErrors = Set(Messages.logFailure, Messages.truncateLogs(action.limits.logs.asMegaBytes))
-
     container
       .logs(action.limits.logs.asMegaBytes, action.exec.sentinelledLogs)(transid)
-      .via(DockerLogStore.toFormattedString)
+      .via(DockerToActivationLogStore.toFormattedString)
       .runWith(Sink.seq)
       .flatMap { seq =>
+        val possibleErrors = Set(Messages.logFailure, Messages.truncateLogs(action.limits.logs.asMegaBytes))
```
rabbah: should this be a singleton set?

common/scala/src/main/scala/whisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
```diff
@@ -0,0 +1,148 @@
...
+                           user: Identity,
+                           activation: WhiskActivation,
+                           container: Container,
+                           action: ExecutableWhiskAction): Future[ActivationLogs] = {
+
+    val logs = container.logs(action.limits.logs.asMegaBytes, action.exec.sentinelledLogs)(transid)
+
+    val additionalMetadata = Map(
+      "activationId" -> activation.activationId.asString.toJson,
+      "action" -> action.fullyQualifiedName(false).asString.toJson,
```
rabbah: isn't this in the WhiskActivation? Should we avoid closing over the Action itself?

markusthoemmes: We can't, we need limit values anyway (which are not part of the activation).

common/scala/src/main/scala/whisk/core/containerpool/logging/DockerToActivationLogStore.scala
markusthoemmes: Can't be, the truncation message is specific to the limit.

common/scala/src/main/scala/whisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
markusthoemmes: I wanted to avoid this for now, since this is an optional logstore.

common/scala/src/main/scala/whisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
rabbah: the limits are part of the activation also - we add both the name and the limits as annotations on every activation.

markusthoemmes: That means writing code that relies on annotations, which are optional by design, but are virtually always there. Feels like a super wrong thing to do to me honestly.

rabbah: annotations are optional on entities not activations which are system records generated by the invoker - we could abstract these with accessory methods on the WhiskActivation object and hide the implementation details otherwise. My concern was dragging the entire action around and lengthening its lifetime where it seems unnecessary; maybe there's an alternative or it's moot as this is going to become small in general and cached in the common case.

PR #2976, opened 2017-11-16
Support for updating feed triggers using CLI

abaruni:
Add support for `trigger update` with feed triggers in the CLI


abaruni:
@csantanapr fyi

PR #2979, opened 2017-11-17
Removal of openwhisk cli and cli related test cases

houshengbo:
This PR removes all the go code of cli and cli related test cases,
since it has been moved to a separate repo. The default cli installation
mode is switched to "remote". In addition, this PR updates the document
to configure the CLI installation mode.


houshengbo:
PG 4/1049/

ServoKvd:
Hi Vicent, Santana,
Today I notice the following files in openwhisk still has references to cli, though in some cases the build does nto complain:
1.)reference to cli settings.gradle in the root folder -
 ./gradlew distDocker executes fine, even though cli folder is not there, since it recursively checks for build.gradle and doesn seem to use settings.gradle to do the build itself. 
2.) Reference to cli in  tools/vagrant/Vagrantfile
Chetanm reported a build failure on teh slack channel today - so i guess this needs to be fixed.
3.) references to  "tools:cli" in bin/wskdev,  tools/admin/wskadmin
I guess wskdev & wdkadmin will also move to the cli folder soon?
4.) reference to tools:cli component in tools/build/redo 

I was thinking of submitted a patch for (1,2,4) above, but then I was wondering if the build scripts should  copy over the built component/image  from "cli repo" instead of simply ommitting that build steup. 

Any comments?

PR #2980, opened 2017-11-18
update feeds.md for with additional lifecycleEvent UPDATE

abaruni:


abaruni:
this is to accompany work done on the providers, cli, and js client. it's the same flow as `trigger get` and `trigger update`. for the cli we fetch the trigger and if it has the feed annotation, then we invoke the feed action passing in the `UPDATE` lifecycleEvent parameter. Otherwise, carry on as usual

abaruni:
wanted to update the docs here. work is completed for kafka and cloudant providers along with js client. working on getting the pr in for the cli

abaruni:
@csantanapr i moved the code changes to the cli to the cli repo. this is just the doc changes

PR #2668, opened 2017-08-28
multiple LoadBalancers

tysonnorris:
Builds upon #2584 to provide multiple loadbalancers.
ActivationThrottler behavior is moved into LoadBalancer.check
provided impls are resolved using LoadBalancerResolver; resolver impls are currently: SingleLoadBalancerResolver (use a single LB for everything) and KindBasedLoadBalancerResolver (use typesafe config to map kind -> LoadBalancer impl)

PR #2984, opened 2017-11-19
Dynamic LoadBalancer load using SpiLoader

kpavel:
Leveraging SpiLoader dynamic object loading adding capability to load LoadBalancer dynamically, based on a value specified in reference.conf. Similar to MessagingProvider, ArtifactStoreProvider...etc.

changes in code:

reference.conf: added LoadBalancerProvider

Controller: using SpiLoader to load LoadBalancer instance and its
requiredProperties

RestAPIs: use LoadBalancer trait instead of LoadBalancerService class

LoadBalancerService:
added LoadBalancerProvider trait implementing Spi
added allInvokers method to LoadBalancer trait

### DIFF
--- a/common/scala/src/main/resources/reference.conf
+++ b/common/scala/src/main/resources/reference.conf
@@ -3,5 +3,5 @@ whisk.spi{
   MessagingProvider = whisk.connector.kafka.KafkaMessagingProvider
   ContainerFactoryProvider = whisk.core.containerpool.docker.DockerContainerFactoryProvider
   LogStoreProvider = whisk.core.containerpool.logging.DockerToActivationLogStoreProvider
-  LoadBalancerProvider = whisk.core.loadBalancer.LoadBalancerService
+  LoadBalancerProvider = whisk.core.loadBalancer.ContainerPoolBalancer
 }
--- a/common/scala/src/main/resources/reference.conf
+++ b/common/scala/src/main/resources/reference.conf
@@ -3,4 +3,5 @@ whisk.spi{
   MessagingProvider = whisk.connector.kafka.KafkaMessagingProvider
   ContainerFactoryProvider = whisk.core.containerpool.docker.DockerContainerFactoryProvider
   LogStoreProvider = whisk.core.containerpool.logging.DockerToActivationLogStoreProvider
+  LoadBalancerProvider = whisk.core.loadBalancer.LoadBalancerService
 }
--- a/core/controller/src/main/scala/whisk/core/controller/Controller.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Controller.scala
@@ -138,10 +138,13 @@ class Controller(val instance: InstanceId,
    */
   private val internalInvokerHealth = {
     implicit val executionContext = actorSystem.dispatcher
-
     (path("invokers") & get) {
       complete {
-        loadBalancer.healthStatus
+        loadBalancer
+          .invokerHealth()
+          .map(_.map {
+            case i => s"invoker${i.id.toInt}" -> i.status.asString
+          }.toMap.toJson.asJsObject)
       }
     }
   }
--- a/core/controller/src/main/scala/whisk/core/controller/Controller.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Controller.scala
@@ -46,7 +46,7 @@ import whisk.core.entitlement._
 import whisk.core.entity._
 import whisk.core.entity.ActivationId.ActivationIdGenerator
 import whisk.core.entity.ExecManifest.Runtimes
-import whisk.core.loadBalancer.{LoadBalancerService}
+import whisk.core.loadBalancer.LoadBalancerProvider
 import whisk.http.BasicHttpService
 import whisk.http.BasicRasService
 import whisk.spi.SpiLoader
@@ -117,7 +117,8 @@ class Controller(val instance: InstanceId,
   })
 
   // initialize backend services
-  private implicit val loadBalancer = new LoadBalancerService(whiskConfig, instance, entityStore)
+  private implicit val loadBalancer =
+    SpiLoader.get[LoadBalancerProvider].loadBalancer(whiskConfig, instance)
   private implicit val entitlementProvider = new LocalEntitlementProvider(whiskConfig, loadBalancer)
   private implicit val activationIdFactory = new ActivationIdGenerator {}
   private implicit val logStore = SpiLoader.get[LogStoreProvider].logStore(actorSystem)
@@ -140,9 +141,7 @@ class Controller(val instance: InstanceId,
 
     (path("invokers") & get) {
       complete {
-        loadBalancer.allInvokers.map(_.map {
-          case (instance, state) => s"invoker${instance.toInt}" -> state.asString
-        }.toMap.toJson.asJsObject)
+        loadBalancer.healthStatus
       }
     }
   }
@@ -163,7 +162,7 @@ object Controller {
     Map(WhiskConfig.controllerInstances -> null) ++
       ExecManifest.requiredProperties ++
       RestApiCommons.requiredProperties ++
-      LoadBalancerService.requiredProperties ++
+      SpiLoader.get[LoadBalancerProvider].requiredProperties ++
       EntitlementProvider.requiredProperties
 
   private def info(config: WhiskConfig, runtimes: Runtimes, apis: List[String]) =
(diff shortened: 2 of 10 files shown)


rabbah:
@tysonnorris do you think you can give this a look?

tysonnorris:
I opened a similar PR in the past, and Markus and others had some concerns about LoadBalancer being inappropriate for an SPI. @markusthoemmes can you voice your opinion here?

rabbah:
@tysonnorris this one https://github.com/apache/incubator-openwhisk/pull/2668?

tysonnorris:
Yes

rabbah:
If I recall some of the discussion previously, it was around ending improving the existing load balancer vs creating plugins for experimental or alternate heuristics. But I think this is inevitable - we need to experiment with load balancers or even no load balancers for some architectures. Hence I think we should have an SPI for the load balancer heuristic.

rabbah:
@tysonnorris is this useful for you also? do you want to revive your pr or shall we consolidate with this one?

tysonnorris:
@rabbah this is useful IFF creating loadbalancers to do significantly different things is tolerable, like elevating container knowledge to the controller to provide, effectively, "invokerless" envs (e.g. mesos/kube/swarm envs don't need the notion of invokers, since they should have a cluster-wide view). The basic summary of changes that this requires is moving ContainerPool to a shared (common/scala) location, and allowing a case for a loadbalancer that is ContainerFactory-aware. 

So, yes, I'm all for doing this. This PR, I think, "only" creates the SPI, and does not create new impls, unless I'm missing something, which is fine, but I think some of the signatures need to change to isolate some of the SPI interface and more cleanly define what is the interface that loadbalancer defines. 

Will review in more details, but would like feedback on the above.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -79,6 +80,8 @@ trait LoadBalancer {
   def publish(action: ExecutableWhiskActionMetaData, msg: ActivationMessage)(
     implicit transid: TransactionId): Future[Future[Either[ActivationId, WhiskActivation]]]
 
+  def allInvokers: Future[IndexedSeq[(InstanceId, InvokerState)]]
```
tysonnorris: Why should `allInvokers` be part of this interface? It should only be referenced internally (except that it currently is referenced as a health check endpoint on controller `/invokers`, which should delegate to the loadbalancer IMHO.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -339,7 +342,23 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
...
 }
 
-object LoadBalancerService {
+/**
+ * An Spi for providing LoadBalancerService implementations
+ */
+trait LoadBalancerProvider extends Spi {
+  def requiredProperties: Map[String, String]
+
+  def loadBalancer(whiskConfig: WhiskConfig, instance: InstanceId, entityStore: EntityStore)(
```
tysonnorris: I don't think `entityStore` should be part of this interface, it is an impl detail of the invokerpool, afaik.

rabbah: The `entityStore` is here only to create the health check. I think this can be separated from the load balancer service. The load balancer doesn't need access to the `entityStore`. We should factor it out.

This creates an ordering issue: factor it out first then introduce the SPI, or can we introduce the SPI without the entity store but have a concrete implementation that used the entity store to preserve existing behavior today?

Since SPI changes are breaking API changes potentially, we should avoid introducing an SPI that we'll change soon there after.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -46,7 +46,7 @@ import whisk.core.entitlement._
 import whisk.core.entity._
 import whisk.core.entity.ActivationId.ActivationIdGenerator
 import whisk.core.entity.ExecManifest.Runtimes
-import whisk.core.loadBalancer.{LoadBalancerService}
+import whisk.core.loadBalancer.{LoadBalancerProvider}
```
rabbah: since importing only one type, drop `{}`?

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -84,15 +89,23 @@ trait LoadBalancer {
   def publish(action: ExecutableWhiskActionMetaData, msg: ActivationMessage)(
     implicit transid: TransactionId): Future[Future[Either[ActivationId, WhiskActivation]]]
 
+  /**
+   * Return a message indicating the health of the containers and/or container pool in general
+   * @return a Future[String] representing the heal response that will be sent to the client
```
rabbah: `Future[String]` -> `Future[JsObject]`
`the heal response that will be sent to the client` -> the health of the container pools managed by the load balancer.

We probably want to raise the JsObject into some abstract type e.g., `(InstanceId, InvokerState)` used below, that at least has the name of the resource, it's heath (up/down) vs the opaque JsValue now.

rabbah:
@tysonnorris I think your review comments are addressed in the latest revisions. When you can, please take another look.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -84,15 +89,23 @@ trait LoadBalancer {
   def publish(action: ExecutableWhiskActionMetaData, msg: ActivationMessage)(
     implicit transid: TransactionId): Future[Future[Either[ActivationId, WhiskActivation]]]
 
+  /**
```
tysonnorris: ^^^ (line 72) as discussed a while back in #2668, it may be worthwhile changing the LoadBalancer trait to have a generic "check()" function, which may do things other than a systemwide and per-namespace activation check. This requires some changes in RateThrottler etc.

I'd be OK waiting on this till later, but it will require changing the interface exposed by the SPI later - I'm not sure whether that is sooner vs later. WDYT @rabbah ?

common/scala/src/main/resources/reference.conf
```diff
@@ -3,4 +3,5 @@ whisk.spi{
   MessagingProvider = whisk.connector.kafka.KafkaMessagingProvider
   ContainerFactoryProvider = whisk.core.containerpool.docker.DockerContainerFactoryProvider
   LogStoreProvider = whisk.core.containerpool.logging.DockerToActivationLogStoreProvider
+  LoadBalancerProvider = whisk.core.loadBalancer.LoadBalancerService
```
tysonnorris: I don't have a strong opinion, but it may be good to have more specific names to the Spi impl, e.g. `whisk.core.loadBalancer.DefaultLoadBalancerProvider' (not sure of a great name here; maybe KafkaLoadBalancerProvider?)

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
rabbah: What’s the interface you have in mind? Looking at the previous proposal it was:

def check(user: Identity)(implicit tid: TransactionId): Future[Unit]

Which is fairly generic. We could include a “preprocess” method in the interface with an empty implementation in the default.

I’d have to think further about refactoring the throttle checks into the LB vs the current model where the LB exports some state and the controller uses it for throttling. I’d rather not refactor until it’s compelling and enables something new.

I also think we should accept further changes to the interface if there are many lb providers written that warrant it. 

Do you envision how this might be used in the overflow q implementation you’re working on?

common/scala/src/main/resources/reference.conf
rabbah: Hmm I don’t like kafka in the name. The lb should be using an abstract messaging interface not specifically Kafka’s. ContainerPoolLoadBalancer? We can pull the renaming into a seperate issue and come up with a better name. I agree the choice of LoadBalancerService may not have been the best (it dates back to when we had a seperate microservice to LB).

rabbah:
@tysonnorris I amended this PR with a couple of refactoring commits.

core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
tysonnorris: I don't think this would impact the overflow q, since the throttling is done upstream, it is just a way to make the throttling implementation specific to a loadbalancer, and possibly more comprehensive than what exists today (action specific limits, dynamic limits, etc).  

We can revisit that later as long as its ok to rev the interface (which I am ok with as well)

PR #2986, opened 2017-11-20
Remove WskRestConsoleTest.

markusthoemmes:
This test does not make sense to have as a REST based test, since it doesn't test any implementation other than the test's implementation, which is faulty.

The `pollFor` uses a different view than the underlying `listActivations` call, which results in intermittent failures.


rabbah:
Partially addresses https://github.com/apache/incubator-openwhisk/issues/2936

PR #2988, opened 2017-11-20
Revert "Programmatic lazy creation of completedN and invokerN topics …

dgrove-oss:
…(#2948)"

This reverts commit da65465464d9fec016d40234396342548644cd91.


dgrove-oss:
Investigating https://github.com/apache/incubator-openwhisk/issues/2987.  Can easily reproduce failure Rodric is seeing with large action from master.  Reverted the kafka change, failure is no longer reproducible.

Suggest reverting change while I attempt to isolate root cause.  Mostly likely an issue with the 0.11.0.1 client, not with the topic creation itself.

dgrove-oss:
actually, it is not clear it is actually related to the kafka PR.   When the test fails for me, the root cause appears to be the "disk" being full in the kafka container.

dgrove-oss:
when I was able to reproduce the failure rodric reported, it was also the case that the disk was full in the kafka container.   not clear reverting is actually useful.

PR #2990, opened 2017-11-21
Change log level dynamically

vvraskin:
This will allow to change log level on per request/user basis for OW transactions. 
If a specific header is provided in the http request to the controller, it will log messages on the debug level for that given transaction id. 
@mhenke1 

regards, Vadim and Martin.

### DIFF
--- a/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/ansible/roles/nginx/templates/nginx.conf.j2
@@ -36,6 +36,20 @@ http {
         keepalive 512;
     }
 
+    map "$remote_user" $special_user {
+      default not_special;
+{% for user in nginx.special_users %}
+      "{{ user }}" special;
+{% endfor %}
+    }
+
+    map "$special_user:$http_x_ow_extra_logging" $extra_logging {
+      default              "off";
+      "special:off" off;
+      "special:on" on;
+      "special:" on;
+      }
+
     server {
         listen 443 default ssl;
 
@@ -63,6 +77,7 @@ http {
             if ($namespace) {
               rewrite    /(.*) /api/v1/web/${namespace}/$1 break;
             }
+            proxy_set_header X-OW-EXTRA-LOGGING $extra_logging;
             proxy_pass http://controllers;
             proxy_read_timeout 75s; # 70+5 additional seconds to allow controller to terminate request
         }
@@ -72,6 +87,7 @@ http {
             if ($namespace) {
               rewrite    ^ /api/v1/web/${namespace}/public/index.html break;
             }
+            proxy_set_header X-OW-EXTRA-LOGGING $extra_logging;
             proxy_pass http://controllers;
             proxy_read_timeout 75s; # 70+5 additional seconds to allow controller to terminate request
         }
--- a/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/ansible/roles/nginx/templates/nginx.conf.j2
@@ -44,7 +44,7 @@ http {
     }
 
     map "$special_user:$http_x_ow_extra_logging" $extra_logging {
-      default              "off";
+      default "off";
       "special:off" off;
       "special:on" on;
       "special:" on;
--- a/common/scala/src/main/scala/whisk/common/Logging.scala
+++ b/common/scala/src/main/scala/whisk/common/Logging.scala
@@ -37,7 +37,11 @@ trait Logging {
    * @param message Message to write to the log
    */
   def debug(from: AnyRef, message: String)(implicit id: TransactionId = TransactionId.unknown) = {
-    emit(DebugLevel, id, from, message)
+    if (id.meta.extraLogging) {
+      emit(InfoLevel, id, from, message)
+    } else {
+      emit(DebugLevel, id, from, message)
+    }
   }
 
   /**
--- a/common/scala/src/main/scala/whisk/common/TransactionId.scala
+++ b/common/scala/src/main/scala/whisk/common/TransactionId.scala
@@ -24,13 +24,9 @@ import java.util.concurrent.atomic.AtomicInteger
 
 import scala.math.BigDecimal.int2bigDecimal
 import scala.util.Try
-
 import akka.event.Logging.{InfoLevel, WarningLevel}
 import akka.event.Logging.LogLevel
-import spray.json.JsArray
-import spray.json.JsNumber
-import spray.json.JsValue
-import spray.json.RootJsonFormat
+import spray.json.{JsArray, JsBoolean, JsNumber, JsValue, RootJsonFormat}
 
 /**
  * A transaction id for tracking operations in the system that are specific to a request.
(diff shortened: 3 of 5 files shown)


common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -37,6 +37,9 @@ trait Logging {
    * @param message Message to write to the log
    */
   def debug(from: AnyRef, message: String)(implicit id: TransactionId = TransactionId.unknown) = {
+    if (id.meta.extraLogging) {
+      emit(InfoLevel, id, from, message)
+    }
     emit(DebugLevel, id, from, message)
```
chetanmeh: May be put the DebugLevel log in else to avoid duplicate logs

chetanmeh:
Looks useful!. Logback provides support for [TurboFilters] which can be used to enable log for specific thread of execution. We used this in [Apache Sling][2] to enable log for specific logger for specific thread via HTTP Header and make those logs available in Chrome plugin. Something like below

```
curl -s -D - -H "X-OW-EXTRA-LOGGING : true" 
-H "X-OW-EXTRA-LOGGING-CONFIG : org.apache.kafka.clients;level=trace" \
 http://localhost:8000/api/v1/namespaces/_/actions/hello?overwrite=true -o /dev/null
```

This would allow us to say see what all outbound calls are made in specific request. Going further probably the log directive can be added to activation message and picked up by invoker those allowing tracing logs for specific category on invoker side to debug any potential issue

All this is just a possible suggestion/idea for future work in this area! For now we can keep it simple and start with current approach in this PR

[1]: https://logback.qos.ch/manual/filters.html#TurboFilter
[2]: https://sling.apache.org/documentation/bundles/log-tracers.html

mhenke1:
@chetanmeh Thank you for your feedback. Learning from proven solutions like Apache Sling is always a valuable asset to avoid making the same things wrong :-)
Indeed we also looked into using Turbofilter, but had to learn that in our case they do not work well. 
The reason is that we use Akka Logging. Log message are passed through the Akka Event Bus (we win asynchronicity on this way) before passed to Logback. When running on INFO level all DEBUG level messaged are not put into the event bus and therefore do not reach Logback and it's filters. The only way to exploit Logback filtering would have been to put all messages into the event bus and use Logback  to filter out the one with the appropriate level. But in this case we would would again see increased memory pressure due to bigger internal buffers needed by the event bus. So we decided for this simpler solution.

common/scala/src/main/scala/whisk/common/Logging.scala
mhenke1: Good catch

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -37,7 +37,11 @@ trait Logging {
    * @param message Message to write to the log
    */
   def debug(from: AnyRef, message: String)(implicit id: TransactionId = TransactionId.unknown) = {
-    emit(DebugLevel, id, from, message)
+    if (id.meta.extraLogging) {
+      emit(InfoLevel, id, from, message)
```
csantanapr: Are these reverse? 
When extraLogging is True do DebugLevel, when False to InfoLevel

mhenke1: @csantanapr They are not reversed. In the case of one transaction got marked with extra logging we elevate those messages from DEBUG to INFO. As I tried to explained in the comment above, this is the only way to get them placed into Akka logging's event bus when the overall log-level is set to INFO.

csantanapr: Thanks @mhenke1 for the explanation all messages will be printed with [INFO] is just when extra logging there will be more [INFO] printed not a mixed between [DEBUG] and [INFO]

mhenke1: @csantanapr Exactly. For "normal" transactions messages will be send according to the log level. For transactions with "extra logging"  additionally all DEBUG messages will be send as INFO level messages.

ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -36,6 +36,20 @@ http {
...
+      "{{ user }}" special;
+{% endfor %}
+    }
+
+    map "$special_user:$http_x_ow_extra_logging" $extra_logging {
+      default              "off";
+      "special:off" off;
+      "special:on" on;
+      "special:" on;
+      }
```
markusthoemmes: Formatting?

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -216,21 +212,22 @@ object TransactionId {
...
       val now = Instant.now(Clock.systemUTC())
-      TransactionId(TransactionMetadata(tid.toLong, now))
+      TransactionId(TransactionMetadata(tid.toLong, now, extraLogging))
     } getOrElse unknown
   }
 
   implicit val serdes = new RootJsonFormat[TransactionId] {
-    def write(t: TransactionId) = JsArray(JsNumber(t.meta.id), JsNumber(t.meta.start.toEpochMilli))
+    def write(t: TransactionId) =
+      JsArray(JsNumber(t.meta.id), JsNumber(t.meta.start.toEpochMilli), JsBoolean(t.meta.extraLogging))
```
markusthoemmes: Shall we make this a bit more efficient in that, if `extraLogging` is false, we don't write anything at all. That'd spare us writing the extra bytes to Kafka.

Code would look like this:

```scala
def write(t: TransactionId) = if(t.meta.extraLogging) {
  val meta = Seq(t.meta.id.toJson, t.meta.start.toEpochMilli.toJson)
  val extra = if(t.meta.extraLogging) Seq(t.meta.extraLogging.toJson) else Seq.empty
  (meta ++ extra).toJson
}

def read(value: JsValue) = Try {
  value match {
    case JsArray(Vector(JsNumber(id), JsNumber(start)) =>
      TransactionId(TransactionMetadata(id.longValue, Instant.ofEpochMilli(start.longValue), false))
    case JsArray(Vector(JsNumber(id), JsNumber(start), JsBoolean(extraLogging)) =>
      TransactionId(TransactionMetadata(id.longValue, Instant.ofEpochMilli(start.longValue), extraLogging))
  }
}.getOrElse(unknown)
```

WDYT?

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -24,13 +24,9 @@ import java.util.concurrent.atomic.AtomicInteger
...
 import scala.math.BigDecimal.int2bigDecimal
 import scala.util.Try
-
 import akka.event.Logging.{InfoLevel, WarningLevel}
 import akka.event.Logging.LogLevel
-import spray.json.JsArray
-import spray.json.JsNumber
-import spray.json.JsValue
-import spray.json.RootJsonFormat
+import spray.json.{JsArray, JsBoolean, JsNumber, JsValue, RootJsonFormat}
```
markusthoemmes: `spray.json._` maybe?

mhenke1: Changed in PR #3174 (replacing this one)

cbickel:
Replaced by #3174

PR #2991, opened 2017-11-21
controller creates health and cacheInvalidation topics

dgrove-oss:
Move creation of the health and cacheInvalidation topics
from ansible to controller startup using ensureTopic
from MessagingProvider SPI.


dgrove-oss:
PG2 2356 - queued

dgrove-oss:
PG2 2356 - completed.  2 test failures, but don't look related to me.

dgrove-oss:
PG2 2362 🔵

dgrove-oss:
rebased to resolve merge conflicts.  Ready for review.

dgrove-oss:
correction.  missed a piece of the kafka cluster change to carry through replication factor.  Not ready for review.

dgrove-oss:
Rebased to pick up pureconfig (thanks @markusthoemmes!) and cleaned up SPI/config by moving all the kafka topic specific configurations to new kafka.conf file.

dgrove-oss:
PG2 2413 🏃

common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
```diff
@@ -46,15 +57,14 @@ object KafkaMessagingProvider extends MessagingProvider {
   def getProducer(config: WhiskConfig, ec: ExecutionContext)(implicit logging: Logging): MessageProducer =
     new KafkaProducerConnector(config.kafkaHosts, ec)
 
-  def ensureTopic(config: WhiskConfig, topic: String, topicConfig: Map[String, String])(
-    implicit logging: Logging): Boolean = {
+  def ensureTopic(config: WhiskConfig, topic: String, topicConfig: String)(implicit logging: Logging): Boolean = {
```
rabbah: Is config still needed?

dgrove-oss: I'm still using it to read kafka.replicationFactor.   I think we should move that value to kafka.conf as well, but I didn't want to delay getting this PR merged to trace through all the deployment configs that might have set the replicationFactor to different values for HA throughout the pipelines.

common/scala/src/main/scala/whisk/core/database/RemoteCacheInvalidation.scala
```diff
@@ -37,6 +37,7 @@ import whisk.core.entity.CacheKey
 import whisk.core.entity.InstanceId
 import whisk.core.entity.WhiskAction
 import whisk.core.entity.WhiskActionMetaData
+import whisk.core.WhiskConfig
```
rabbah: odd - this is now needed?

common/scala/src/main/resources/kafka.conf
```diff
@@ -0,0 +1,24 @@
+kafka {
```
rabbah: should the corresponding values in the ansible group vars be removed?

dgrove-oss: yep.  good catch.  done.

dgrove-oss: removed.  I thought it was still needed because WhiskConfig is mentioned in the constructor down below, but I guess I don't understand Scala's import rules...only have to import if you actually do something with the type (not just passing it through?)

common/scala/src/main/scala/whisk/core/database/RemoteCacheInvalidation.scala
dgrove-oss: removed. I thought it was still needed because WhiskConfig is mentioned in the constructor down below, but I guess I don't understand Scala's import rules...only have to import if you actually do something with the type (not just passing it through?)

dgrove-oss:
rebased to pick up using single couchdb instance in travis. review comments addressed & squashed.

dgrove-oss:
PG2 2423 🔵

rabbah:
Is there a companion to this?

dgrove-oss:
yes. companion is 4399

common/scala/src/main/resources/kafka.conf
markusthoemmes: But only if those are valid for **each** environment?

common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
```diff
@@ -35,6 +35,17 @@ import whisk.core.connector.MessageConsumer
 import whisk.core.connector.MessageProducer
 import whisk.core.connector.MessagingProvider
 
+import pureconfig._
+
+case class TopicConfig(segmentBytes: Long, retentionBytes: Long, retentionMs: Long) {
+  def toMap(): Map[String, String] = {
```
markusthoemmes: Drop the parentheses on `toMap`. It's not side-effecting.

common/scala/src/main/resources/kafka.conf
cbickel: Why do we create a new kafka.conf? We could simply add a whisk.kafka section the the application.conf

common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
cbickel: Why do we keep the replication-factor in the WhiskConfig?

common/scala/src/main/resources/kafka.conf
dgrove-oss: It struck me as more modular to have a conf files for the various sub-components and include them in application.conf (similar to how logging is already done).  I don't have strong opinions on this, but we should decide what we want to do and document it.

@markusthoemmes, is there some write up of conventions for how we should be using pureconfig?  One of my todos for the next couple of weeks is to migrate as much as possible from ansible to pureconfig to reduce the number of build breaks in kube.  We had 4 separate kube build breaks last week due to variables been added/updated in ansible configs and not replicated to the kube project.

markusthoemmes: @dgrove-oss yep, there is: https://github.com/apache/incubator-openwhisk/blob/master/docs/dev/configuration.md

We should maybe add that all openwhisk related configuration should go under "whisk" to prevent potential clashes with other modules (reactive kafka in this case).

common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
dgrove-oss: as above.  Because @markusthoemmes was on vacation last week and I am not clear on the plan for how values that are very likely to be changed in different deployment need to be managed, I was leaving for followup PR.  I'm blocked on some other kube work (elimintating custom kafka image) until this PR gets merged.

common/scala/src/main/resources/kafka.conf
dgrove-oss: ok.  So all the overrides still happen via environment variables (in ansible or yaml files for kube, etc).  So there is less value for a separate .conf file for different subsystems.  Should I merge kafka.conf into application.conf as suggested by @cbickel then?

common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
cbickel: Is there a chance to move the replicationFactor from WhiskConfig to Pureconfig in this PR?

dgrove-oss: Since Markus is back to answer questions, yeah I can do that in this PR too.

dgrove-oss:
large enough changes to address review comments that I'm redoing PG testing.
PG2 2444 🏃

dgrove-oss:
PG3 1443 🏃

dgrove-oss:
needs companion label

common/scala/src/main/scala/whisk/core/connector/MessagingProvider.scala
```diff
@@ -34,6 +34,5 @@ trait MessagingProvider extends Spi {
                   maxPeek: Int = Int.MaxValue,
                   maxPollInterval: FiniteDuration = 5.minutes)(implicit logging: Logging): MessageConsumer
   def getProducer(config: WhiskConfig, ec: ExecutionContext)(implicit logging: Logging): MessageProducer
-  def ensureTopic(config: WhiskConfig, topic: String, topicConfig: Map[String, String])(
-    implicit logging: Logging): Boolean
+  def ensureTopic(config: WhiskConfig, topic: String, topicConfig: String)(implicit logging: Logging): Boolean
```
cbickel: Now, `WhiskConfig` can be removed, right?

ansible/group_vars/all
```diff
@@ -134,24 +134,8 @@ kafka:
...
-    invoker:
-      segmentBytes: 536870912
-      retentionBytes: "{{ kafka_topics_invoker_retentionBytes | default(1073741824) }}"
-      retentionMS: "{{ kafka_topics_invoker_retentionMS | default(172800000) }}"
-    cacheInvalidation:
-      segmentBytes: 536870912
-      retentionBytes: "{{ kafka_topics_cacheInvalidation_retentionBytes | default(1073741824) }}"
-      retentionMS: "{{ kafka_topics_cacheInvalidation_retentionMS | default(300000) }}"
+
+kafka_replicationFactor: "{{ (groups['kafkas']|length)|int }}"
```
cbickel: Is there a reason to change `kafka.replicationFactor` to `kafka_replicationFactor`?

common/scala/src/main/scala/whisk/core/connector/MessagingProvider.scala
dgrove-oss: No.  We need it to get to kafkaHosts (same as the other two SPI methods).  https://github.com/apache/incubator-openwhisk/blob/d69f8db02be70db0da1e07614124ea2e08dd6826/common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala#L52

ansible/group_vars/all
dgrove-oss: I thought flattening was required if we wanted to use it via pureconfig.  Maybe I am still not understanding something here.

dgrove-oss:
PG4 1117 🏃

dgrove-oss:
Final PG run good: PG1 2398

PR #2992, opened 2017-11-21
create SPI for SeedNodesProvider; MarathonSeedNodesProvider to provid…

tysonnorris:
…e marathon based seed node config

Added mockito dependency for tests, to enable mocking classes (e.g. Cluster) without empty constructors


markusthoemmes:
Assigning to @tysonnorris and attaching WIP. What's the status from your POV @tysonnorris?

tysonnorris:
Will review in context of mesos HA controller deployment. If this shared data aspect is going away, it can become embedded into mesos support for special purpose there.

tysonnorris:
FYI - this is now using akka management and akka cluster bootstrap to implement the non-seednodes case of cluster join. It is technically possible to leave the seed nodes usage in place alongside the cluster bootstrap (which will be ignored if seed nodes are specified), but this will result in warnings. 
For details see https://developer.lightbend.com/docs/akka-management/current/bootstrap.html#joining-mechanism-precedence

I may revisit this to avoid a new SPI, and rather make it a boolean config to either use seednodes, or cluster bootstrap, as the method for cluster joining.

PR #2998, opened 2017-11-23
Increase limits.invocationsPerMinute default value

ningyougang:
Sometimes, when run the test cases, the apigw request may exceed the default
value(60), so it is necessary to increase the limits.invocationsPerMinute
default value.

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -100,7 +100,7 @@ runtimesManifestDefault:
     - name: "dockerskeleton"
 
 limits:
-  invocationsPerMinute: "{{ limit_invocations_per_minute | default(60) }}"
+  invocationsPerMinute: "{{ limit_invocations_per_minute | default(120) }}"
   concurrentInvocations: "{{ limit_invocations_concurrent | default(30) }}"
   concurrentInvocationsSystem:  "{{ limit_invocations_concurrent_system | default(5000) }}"
   firesPerMinute: "{{ limit_fires_per_minute | default(60) }}"


ningyougang:
This question has relation with below issue:
https://github.com/apache/incubator-openwhisk/issues/2961#issuecomment-346542753

markusthoemmes:
These tests are supposed to settle the throttles accordingly though. Bumping the limit unfortunately only hides the issue of a buggy test.

ningyougang:
@markusthoemmes , have any better solution?

markusthoemmes:
Have you had a look into the tests implementation? There are helpers in `BaseApiGwTests.scala` which are supposed to settle the throttles after X invocations (where X is the allowed number of executions per minute in that env). Seems like something is wrong there.

ningyougang:
ok. i will look into it.

rabbah:
The 429 response contains enough information to know how much to settle the throttle which takes the guesswork out of the current tests. It would be great to take the response and add an appropriate delay. 

Example response: made 67 requests in last minute 60 allowed.

ningyougang:
Below is failed log
```
whisk.core.cli.test.ApiGwRestTests > Wsk api should verify delete basepath/path STANDARD_OUT
    Action invokes within last minute: 47
    Action invokes within last minute: 48
    Action invokes within last minute: 49
    Action invokes within last minute: 50

whisk.core.cli.test.ApiGwRestTests > Wsk api should verify delete basepath/path FAILED
    org.scalatest.exceptions.TestFailedException: 429 was not equal to 200

whisk.core.cli.test.ApiGwRestTests STANDARD_OUT

    Finished test Wsk api should verify delete basepath/path at 2017-11-23 15:09:57.163

    Starting test Wsk api should verify delete single operation from existing API basepath/path/operation(s) at 2017-11-23 15:09:57.163

whisk.core.cli.test.ApiGwRestTests > Wsk api should verify delete single operation from existing API basepath/path/operation(s) STANDARD_OUT
    Action invokes within last minute: 51
    Action invokes within last minute: 52
```

It seems the apiGw requests number doesn't reach to `default value(60), ` still export `429(too many reqeust)` error.

It seems has some bug on the codes.

ningyougang:
I added some more logs on in `WskRest.scala`'s `def invokeAction(...)`
```
   ....
    if (!isStatusCodeExpected(expectedExitCode, r.statusCode.intValue)) {
      if (blocking || result) {
        println("(blocking||result)========");
        println("invokeAction->msg:"+r.respData);
        println("invokerAction->statusCode:"+r.statusCode);
        println("OK.intValue:"+OK.intValue);
        validateStatusCode(OK.intValue, r.statusCode.intValue)
      }
```
And run tests case again. some part of the logs as below
```
    Starting test Wsk api should verify API with http response type at 2017-11-24 12:02:48.628

whisk.core.cli.test.ApiGwRestTests > Wsk api should verify API with http response type STANDARD_OUT
    Action invokes within last minute: 58
    (blocking||result)========
    invokeAction->msg:{
      "error": "Too many requests in the last minute (count: 70, allowed: 60).",
      "code": 2126
    }
    invokerAction->statusCode:429 Too Many Requests
    OK.intValue:200
    Action invokes within last minute: 59

whisk.core.cli.test.ApiGwRestTests > Wsk api should verify API with http response type FAILED
    org.scalatest.exceptions.TestFailedException: 429 was not equal to 200
```

It seems need to change the default value from `60` to `70`

ningyougang:
After test deeply,
if use this commit: https://github.com/apache/incubator-openwhisk/commit/1a4493f1facd301d34c2beb5dd011c52d3b2953c and test.
The problem gone.

If remove that commit and test again, the problem exist,
It seems that commit solve this problem fundamentally.

So close this issue.

ningyougang:
Oh, sadly, 
We use the latest codes, the problem is still exist(not very stable) which means happen somtimes and not happen sometimes.

rabbah:
Thanks for the catch. The proper fix is to settle the throttle properly in the test suite based on the controller report/error message. I would leave this limit alone (it hides the defect in the test suite).

ningyougang:
Can close

PR #3005, opened 2017-11-26
Update invoke.py to reflect changes in repository

### LINKED ISSUE #1832: document max response size
our limits dont list the max response size.

rabbah:
Also update docs for max result size.

Closes #1832.

### DIFF
--- a/actionRuntimes/actionProxy/invoke.py
+++ b/actionRuntimes/actionProxy/invoke.py
@@ -5,7 +5,7 @@
   by simulating invoker interactions. Use it in combination with
   docker run <image> which starts up the action proxy.
   Example:
-     docker run -i -t -p 8080:8080 openwhisk/dockerskeleton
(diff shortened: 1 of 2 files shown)


actionRuntimes/actionProxy/invoke.py
```diff
@@ -3,9 +3,9 @@
 
   This script is useful for testing the action proxy (or its derivatives)
   by simulating invoker interactions. Use it in combination with
-  delete-build-run.sh which builds and starts up the action proxy.
-  Examples:
-     ./delete-build-run.sh &
+  docker run <image> which starts up the action proxy.
+  Example:
+     docker run -i -t -p 8080:8080 openwhisk/dockerskeleton
```
csantanapr: should this be `dockerskeleton` or `whisk/dockerskeleton` when your working with local images?

rabbah: it could be just `dockerskeleton` too which would avoid the prefix/tag.

PR #3006, opened 2017-11-26
Tweak error message for failing to invoke an action

rabbah:
if its type of an action has changed or cannot be deserialized because of a runtime-manifest/schema mismatch with controller, report a different error that is more useful.


rabbah:
This error doesn't discriminate enough - a total failure to fetch will result in the new misleading error now.

rabbah:
proposed https://github.com/apache/incubator-openwhisk/pull/3139 to address error message.

PR #3009, opened 2017-11-27
Add the idle timeout configuration to the Akka client

houshengbo:
This PR sets the idle timeout to 5 minutes in order to give enough
time for sequence tests to finish.


markusthoemmes:
This has already been added by me to the relevant `application.conf` in 1a4493f1facd301d34c2beb5dd011c52d3b2953c. I might've used the wrong setting, but we should keep akka-related settings in the relevant config file, thoughts?

PR #3010, opened 2017-11-27
Adapt timeouts of controller and nginx.

cbickel:
Lower the timeouts of the controller to match the proxy_read_timeout in nginx.

If the nginx proxy_read_timeout is lower than the controller timeout, this can result in issues with long requests.

### DIFF
--- a/core/controller/src/main/resources/application.conf
+++ b/core/controller/src/main/resources/application.conf
@@ -22,7 +22,7 @@ akka.http {
     # The controller holds connections up to 60s for blocking invokes, and
     # all other operations are expected to complete quickly. We allow a grace
     # period in addition to the blocking invoke timeout.
-    request-timeout = 90s
+    request-timeout = 65s
 
     # The maximum number of concurrently accepted connections when using the
     # `Http().bindAndHandle` methods.
@@ -50,7 +50,7 @@ akka.http {
     #
     # Explaining the set value:
     # This must be greater than the request timeout.
-    idle-timeout = 120s
+    idle-timeout = 70s
 
     parsing {
       # This indirectly puts a bound on the name of entities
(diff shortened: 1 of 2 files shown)


houshengbo:
@cbickel 
Have you got a wrong statement:
```
If the nginx proxy_read_timeout is higher than the controller timeout, this can result in issues with long requests.
```

I think it should be:
```
If the nginx proxy_read_timeout is lower than the controller timeout, this can result in issues with long requests.
```

Based on your change,  proxy_read_timeout becomes 75s, and idle-timeout becomes 70s, lower than 75s.

core/controller/src/main/resources/application.conf
```diff
@@ -15,7 +15,7 @@ akka.http {
     # The controller holds connections up to 60s for blocking invokes, and
     # all other operations are expected to complete quickly. We allow a grace
     # period in addition to the blocking invoke timeout.
-    request-timeout = 90s
+    request-timeout = 65s
```
houshengbo: system.basic.WskRestSequenceTests.Wsk Sequence should execute a sequence in blocking fashion and finish execution even if longer than blocking response timeout

This test case sets the payload exactly to 65000 milliseconds. I am not sure request-timeout = 65s is sufficient for this test case to complete.

cbickel: Thanks for pointing out this test. I didn't have this in mind.
I'll take care.

cbickel: I looked at the tests. I think you are referring to the test, that transforms a blocking request into a non-blocking one.
The timeout I change has nothing to do with this test. If the tests starts failing with 503 now, it would instead fail with a 200 without my changes.

houshengbo: This used to be my concern. Thank you for the explanation.

cbickel:
PG2#2383 🔵

PR #3013, opened 2017-11-28
Change the idle timeout to 90 seconds for client connection settings

houshengbo:
This PR makes sure the idle timeout is long enough for the action invocation to complete. By default, the idle timeout for a client connection is 1 minute. This PR changes it to 90 seconds in order to give enough time for the action call to finish. As an Akka client,  the idle timeout needs to be configured for both of clientConnectionSettings and connectionPoolSettings.

Closes: #3016

### DIFF
--- a/tests/src/test/resources/application.conf
+++ b/tests/src/test/resources/application.conf
@@ -6,3 +6,5 @@ whisk.spi {
 
 # Blocking requests fall back to non-blocking after ~60s
 akka.http.client.idle-timeout = 90 s
+akka.http.host-connection-pool.idle-timeout = 90 s
+akka.http.host-connection-pool.client.idle-timeout = 90 s
--- a/tests/src/test/scala/common/rest/WskRest.scala
+++ b/tests/src/test/scala/common/rest/WskRest.scala
@@ -63,7 +63,6 @@ import akka.http.scaladsl.model.HttpMethods.POST
 import akka.http.scaladsl.model.HttpMethods.PUT
 import akka.http.scaladsl.HttpsConnectionContext
 import akka.http.scaladsl.settings.ConnectionPoolSettings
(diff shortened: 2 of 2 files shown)


rabbah:
description? what is the pr solving?

markusthoemmes:
This has already been added by me to the relevant application.conf in 1a4493f. I might've used the wrong setting, but we should keep akka-related settings in the relevant config file, thoughts?

houshengbo:
akka.http.client.idle-timeout = 90 s added to application.conf should have worked, but based on my observation, it works in travis for openwhisk, but the configuration for timeout and even other akka client settings does not work in the Jenkins build, which is the reason that I switch all the configurations of akka client for test cases into the REST framework in scala code. @markusthoemmes @rabbah

markusthoemmes:
@houshengbo I presume that's due to a missing application.conf on the Jenkins side then?

houshengbo:
@markusthoemmes Is this you can fix for Jenkins? Besides, we need to change the timeout of akka.http.host-connection-pool.idle-timeout and akka.http.host-connection-pool.client.idle-timeout as well.

@jeremiaswerner We are still seeing the timeout issue occasionally raised for sequence tests. IMO, we need to change
* akka.http.client.idle-timeout
* akka.http.host-connection-pool.idle-timeout
* akka.http.host-connection-pool.client.idle-timeout.

markusthoemmes:
@houshengbo sounds good, can you add those to the `application.conf` settings in `tests` and roll back the code changes?

markusthoemmes:
@houshengbo I took the liberty to add a commit and rebase the branch so you'll need to refetch your branch.

houshengbo:
@markusthoemmes Thank you. This works for me.

PR #3017, opened 2017-11-28
log exception when ensureTopic fails with unexpected exception

dgrove-oss:


dgrove-oss:
rebased to pick up single couchdb for travis

dgrove-oss:
does this need a PG before it is merged?  Seems trivial...

PR #3019, opened 2017-11-28
WIP: Trigger activation reduction

mdeuser:
- If trigger has at least one corresponding active rule
  - save activation record
  - return 202 (Accepted); response body has activation id
- If trigger has no corresponding active rule
  - do not save activation record
  - return 204 (No Content) response code; no response body

See
- https://lists.apache.org/thread.html/199c9a75349cc217f2d9731ceea14bc6caa738bde4dc9cb1e3de47c3@%3Cdev.openwhisk.apache.org%3E

- https://github.com/apache/incubator-openwhisk/issues/512

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
@@ -46,6 +46,7 @@ import spray.json.DefaultJsonProtocol.RootJsObjectFormat
 
 import whisk.common.TransactionId
 import whisk.core.database.CacheChangeNotification
+import whisk.core.database.NoDocumentException
 import whisk.core.entitlement.Collection
 import whisk.core.entity.ActivationResponse
 import whisk.core.entity.EntityPath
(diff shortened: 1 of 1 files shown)


mdeuser:
Early look at controller code for comments.  CLI changes are also likely needed.  Test updates forthcoming as well..

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -143,71 +144,81 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
+                  }
+              }
           }
 
           onComplete(saveTriggerActivation) {
             case Success(activationId) =>
-              complete(OK, activationId.toJsObject)
+              complete(Accepted, activationId.toJsObject)
+            case Failure(t: NoDocumentException) =>
+              terminate(NoContent)
```
rabbah: @jasonpet with this change what would the feeds do ie is 204 treated as success?

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -143,71 +144,81 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
+                        logging.warn(this, s"${rule.action} failed due to $error")
+                      }
+                  }
+              }
           }
 
           onComplete(saveTriggerActivation) {
             case Success(activationId) =>
-              complete(OK, activationId.toJsObject)
+              complete(Accepted, activationId.toJsObject)
```
rabbah: this is still going to wait for the trigger put to the db.

style95:
Normally I activate trigger before integrating it with rule(any actions) to check what kinds of parameters to be passed to the action. 

Is this pr for trigger which is invoked by external event but not related to any rules?

rabbah:
That's correct. See discussion on dev list: https://lists.apache.org/thread.html/199c9a75349cc217f2d9731ceea14bc6caa738bde4dc9cb1e3de47c3@%3Cdev.openwhisk.apache.org%3E

mdeuser:
had a chat with @jasonpet.  apparently many of the provider tests rely on the current behavior (i.e. fire trigger that has no rule, look in db for activation record).  i think i'll break this up into smaller prs to allow for more discussion on the trigger activation reduction behavior
1. async trigger activation writes to db.  return 202
2. use internal http interface to invoke the corresponding rule action
3. no longer write trigger activations for triggers without a corresponding active rule

rabbah:
How many tests are we talking about?  If it’s a handful I suggest fixing the tests first.

jasonpet:
yes, almost all provider tests will need to be updated and the kafka provider will need to accept 202s and 204s before any of the proposed changes can be committed.

mdeuser:
Closing due to having moved the implementation to several other PRs
- #3031 / #3062 
- #3187 
- #3262

PR #3031, opened 2017-11-29
Return 202 for trigger fire requests

mdeuser:
Currently, trigger fire requests are blocked by db writes (storing trigger activation).  In times of extreme db latency, a storm of trigger fire retries can occur - further stressing the system.  This PR relaxes the trigger fire response dependency on the trigger activation db write avoiding unnecessary trigger retries when activation writes are excessively backed up.

- Return 202 for trigger fire requests.  Still include activation id in response body as before
- asynchronously write the trigger activation record to the db
  - db write failures continue to be logged to the backend log

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
@@ -21,7 +21,6 @@ import java.time.Clock
 import java.time.Instant
 
 import scala.concurrent.Future
-import scala.util.{Failure, Success}
 
 import akka.actor.ActorSystem
 import akka.stream.ActorMaterializer
@@ -144,9 +143,8 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
             duration = None)
 
           logging.info(this, s"[POST] trigger activated, writing activation record to datastore: $triggerActivationId")
-          WhiskActivation.put(activationStore, triggerActivation) onComplete {
-            case Success(_) =>
-            case Failure(t: Throwable) =>
+          WhiskActivation.put(activationStore, triggerActivation) recover {
+            case t =>
               logging.error(this, s"[POST] storing trigger activation $triggerActivationId failed: ${t.getMessage}")
           }
 
@@ -172,9 +170,8 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
                 logging.info(
                   this,
                   s"[POST] rule ${ruleName} activated, writing activation record to datastore: $ruleActivationId")
-                WhiskActivation.put(activationStore, ruleActivation) onComplete {
-                  case Success(_) =>
-                  case Failure(t: Throwable) =>
+                WhiskActivation.put(activationStore, ruleActivation) recover {
+                  case t =>
                     logging.error(this, s"[POST] storing rule activation $ruleActivationId failed: ${t.getMessage}")
                 }
 
--- a/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
@@ -60,7 +60,6 @@ import whisk.core.entity.types.ActivationStore
 import whisk.core.entity.types.EntityStore
 import whisk.core.entity.Identity
 import whisk.core.entity.FullyQualifiedEntityName
-import whisk.http.ErrorResponse.terminate
 
 /** A trait implementing the triggers API. */
 trait WhiskTriggersApi extends WhiskCollectionAPI {
@@ -143,9 +142,12 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
             response = ActivationResponse.success(payload orElse Some(JsObject())),
             version = trigger.version,
             duration = None)
+
           logging.info(this, s"[POST] trigger activated, writing activation record to datastore: $triggerActivationId")
-          val saveTriggerActivation = WhiskActivation.put(activationStore, triggerActivation) map { _ =>
-            triggerActivationId
+          WhiskActivation.put(activationStore, triggerActivation) onComplete {
+            case Success(_) =>
+            case Failure(t: Throwable) =>
+              logging.error(this, s"[POST] storing trigger activation $triggerActivationId failed: ${t.getMessage}")
           }
 
           val url = Uri(s"http://localhost:${whiskConfig.servicePort}")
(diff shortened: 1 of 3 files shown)


mdeuser:
Travis found a couple other places that are impacted by the 202.  Will investigate.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -155,19 +155,25 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
+                  ruleActivationId,
                   Instant.now(Clock.systemUTC()),
                   Instant.EPOCH,
                   cause = Some(triggerActivationId),
                   response = ActivationResponse.success(),
                   version = trigger.version,
                   duration = None)
-                logging.info(this, s"[POST] rule ${ruleName} activated, writing activation record to datastore")
-                WhiskActivation.put(activationStore, ruleActivation)
+                logging.info(
```
csantanapr: make logging.debug if it going to leave the log in.
doing a lot of logs to info (default) affects performance

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -143,9 +141,11 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
             response = ActivationResponse.success(payload orElse Some(JsObject())),
             version = trigger.version,
             duration = None)
+
           logging.info(this, s"[POST] trigger activated, writing activation record to datastore: $triggerActivationId")
-          val saveTriggerActivation = WhiskActivation.put(activationStore, triggerActivation) map { _ =>
-            triggerActivationId
+          WhiskActivation.put(activationStore, triggerActivation) recover {
+            case t =>
+              logging.error(this, s"[POST] storing trigger activation $triggerActivationId failed: ${t.getMessage}")
```
dubeejw: How is the user informed of this error?

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -155,19 +155,25 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
                   version = trigger.version,
                   duration = None)
-                logging.info(this, s"[POST] rule ${ruleName} activated, writing activation record to datastore")
-                WhiskActivation.put(activationStore, ruleActivation)
+                logging.info(
+                  this,
+                  s"[POST] rule ${ruleName} activated, writing activation record to datastore: $ruleActivationId")
+                WhiskActivation.put(activationStore, ruleActivation) recover {
+                  case t =>
+                    logging.error(this, s"[POST] storing rule activation $ruleActivationId failed: ${t.getMessage}")
```
dubeejw: How is the user informed of this error?

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -205,13 +211,7 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
             }
           }
 
-          onComplete(saveTriggerActivation) {
```
dubeejw: Might want to keep this code block around so the user knows if a failure occurred.

rabbah: They won’t here. No worse than today. The subsequent improvements will address this deficiency. You can see a sketch of the implementation here https://github.com/apache/incubator-openwhisk/compare/master...rabbah:trigger-activations?expand=1

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
rabbah: I’d favor removing them also.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
rabbah: They won’t here. No worse than today. The subsequent improvements will address this deficiency. You can see a sketch of the implementation here https://github.com/apache/incubator-openwhisk/compare/master...rabbah:trigger-activations?expand=1

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
rabbah: Then the api can’t be asynchronous wrt writing the record to the database.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
csantanapr: I’m also ok with removing

dubeejw:
@mdeuser, you probably need to rebase to get Travis to pass.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
mdeuser: Same as today.  No activation record is available.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
mdeuser: 👍

mdeuser:
@dubeejw - any more comments?

mdeuser:
PG2/2436

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -155,19 +155,22 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
                   Instant.now(Clock.systemUTC()),
                   Instant.EPOCH,
                   cause = Some(triggerActivationId),
                   response = ActivationResponse.success(),
                   version = trigger.version,
                   duration = None)
-                logging.info(this, s"[POST] rule ${ruleName} activated, writing activation record to datastore")
-                WhiskActivation.put(activationStore, ruleActivation)
+                WhiskActivation.put(activationStore, ruleActivation) recover {
+                  case t =>
```
rabbah: I thought you were going to remove this one too.

tests/src/test/scala/whisk/core/controller/test/TriggersApiTests.scala
```diff
@@ -316,17 +317,20 @@ class TriggersApiTests extends ControllerTestCommon with WhiskTriggersApi {
...
       val JsString(id) = response.fields("activationId")
       val activationId = ActivationId(id)
       response.fields("activationId") should not be None
 
       val activationDoc = DocId(WhiskEntity.qualifiedName(namespace, activationId))
-      val activation = get(activationStore, activationDoc, WhiskActivation, garbageCollect = false)
-      del(activationStore, DocId(WhiskEntity.qualifiedName(namespace, activationId)), WhiskActivation)
-      activation.end should be(Instant.EPOCH)
-      activation.response.result should be(Some(content))
+      whisk.utils.retry({
```
rabbah: This could have used the withActivation helper I think. And below.

csantanapr:
@mdeuser I think we forgot to update the swagger corresponding to the change.
https://github.com/apache/incubator-openwhisk/blob/master/core/controller/src/main/resources/apiv1swagger.json#L931
```
                "responses": {
                    "200": {
                        "description": "Activation id",
                        "schema": {
                            "$ref": "#/definitions/ItemId"
                        }
                    },
```
Could you submit a new PR to update this?

tests/src/test/scala/whisk/core/controller/test/TriggersApiTests.scala
markusthoemmes: Isn't that more geared towards system level tests? It uses "real" communication if i'm not mistaken (using Wsk or WskRest) vs. akka-http "internal" tests here.

mdeuser:
@csantanapr - good catch.  will do.

tests/src/test/scala/whisk/core/controller/test/TriggersApiTests.scala
mdeuser: i looked at some of those helpers and it seemed to me that they depended on having a `Wsk` object to drive the requests, which was not the case for these tests

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
mdeuser: perhaps i misunderstood.  i thought the issue was in the `.info` logs since those would occupy more log space than the `.error` log entries.  as such, i removed the `.info` log for the rule activation persistence.

rabbah: It’s a moot now since this will go away anyway

Issue #3029, opened 2017-11-29
couchdb clustering not working


csantanapr:
Took a look at the couchdb clustering and attempted to fix.
I got very far but I'm getting an error which I don't know now how to fix.

```
TASK [couchdb : add remote nodes to the cluster] *********************************************************************************************************************
Wednesday 29 November 2017  15:33:30 -0500 (0:00:00.323)       0:00:36.662 ****
fatal: [db0]: FAILED! => {"cache_control": "must-revalidate", "changed": false, "connection": "close", "content": "{\"error\":\"conflict\",\"reason\":\"Document update conflict.\"}\n", "content_length": "58", "content_type": "application/json", "date": "Wed, 29 Nov 2017 20:33:31 GMT", "failed": true, "json": {"error": "conflict","reason": "Document update conflict."}, "msg": "Status code was not [201]: HTTP Error 409: Conflict", "redirected": false, "server": "CouchDB/2.1.1 (Erlang OTP/17)","status": 409, "url": "http://172.17.0.1:5984/_cluster_setup", "x_couch_request_id": "d6757efe56", "x_couchdb_body_time": "0"}

```

I think we should consider reverting the commit https://github.com/apache/incubator-openwhisk/pull/2810
until it get's properly fix.
What do you think @cbickel @style95 ?

csantanapr:
@cbickel @style95 you can see changes here: https://github.com/apache/incubator-openwhisk/pull/3030

style95:
I am checking on it.

style95:
@csantanapr Are you using dns name in db section of `hosts` file?

csantanapr:
Give it a try on your system. 
No dns use I try to use the ansible_host value

csantanapr:
I’m using docker for Mac
Running the following command
ansible-playbook -i environments/local couchdb.yml

style95:
I have discussed with @csantanapr.
Actually there is limitation in #2810.

With that change, we can't setup multiple couchdb nodes on a single machine.
This limitation comes from erlang epmd port.

It is briefly explained here: https://github.com/apache/incubator-openwhisk/pull/2810#issuecomment-343837080

When couchdb form a cluster it counts on erlang epmd to find other nodes.
All nodes will try to find each other with same port(`4369`).
Since it is not possible to differentiate this port for each nodes, we can't run multiple nodes on a single machine.

One idea is, make all nodes communicate with each other using container IP.
[...]

PR #3037, opened 2017-12-01
Add guide about limitation on couchdb clustering

style95:
I wrote down current limitation on CouchDB clustering.
This is to inform those who may face similar issue to this : https://github.com/apache/incubator-openwhisk/issues/3029


csantanapr:
Closes #3029

PR #3039, opened 2017-12-01
Disable controller.ha if only 1 controller is deployed.

cbickel:
Reverts the removal of the second controller for local env.


rabbah:
The commit message isn’t consistent with the change.

PR #3040, opened 2017-12-01
Adjust controller side action time-out to avoid invokers marked as unhealthy 

mhenke1:
This PR adjusts the controller side time out defining the allowed run-time for actions to handle 
edge case scenarios correctly in which the synchronization of controllers in a HA setup does not work fast enough.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
@@ -44,17 +44,10 @@ import whisk.core.connector.MessageFeed
 import whisk.core.connector.MessageProducer
 import whisk.core.connector.MessagingProvider
 import whisk.core.database.NoDocumentException
-import whisk.core.entity.{ActivationId, WhiskActivation}
-import whisk.core.entity.EntityName
-import whisk.core.entity.ExecutableWhiskActionMetaData
-import whisk.core.entity.Identity
-import whisk.core.entity.InstanceId
-import whisk.core.entity.UUID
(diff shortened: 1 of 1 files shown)


mhenke1:
@rabbah Let's take the two controller HA case. In case of non-perfect state synchronization between controllers more than 16 actions can be scheduled to one invoker.  16 of them get executed and the others queued. 

When all 16 executed actions run to their maximal run time and expire, the next batch (the one waiting in the queue) can start after around one minute elapsed time. If these next actions also run to their maximal run time they return after around two minutes 
(1 minute waiting for the first set of actions to stop plus  1 minute run time plus some overhead). 

At the moment the controller only allows two minutes for an action to finish. So we might have the case that actions complete shortly after that two minute time and get regarded as failed by the controller. If we have to many of those cases the given invoker will be regarded as unhealthy. 

During the last days we see a lot of these cases where invokers were marked as unhealthy and recovered after a short time. In the cases the invokers were busy with  batches of actions that were all timing-out nearly at the same time.

In the case we add more controller to the HA game, the summed up waiting time might go up even more. Lets take the most pathological case in which the HA state is not at all synced*. 
In this case all n controllers might place actions on one and the same invoker. The resulting overall wait time is the time for the first batch to be executed plus the time for the (n-1) batches  waiting in the queue. Therefore with this PR the wait time is calculated as (1 + (n-1)) minutes + some overhead.

*Of course this last case is unlikely and hopefully a theoretical edge case.

mhenke1:
@rabbah as discussed the standard time for action invocations is now the lower limit for the calculation of the allowed calculated runtime.

PR #3042, opened 2017-12-01
Make the log level configurable via ansible

mhenke1:
This PR makes the log level configurable via ansible.
The log level is configurable for the controller (`controller_loglevel) `
the invoker (`invoker_loglevel`) 
or for both components (`whisk_loglevel`).

The default log level is still set to INFO level.

### DIFF
--- a/common/scala/src/main/resources/logback.xml
+++ b/common/scala/src/main/resources/logback.xml
@@ -11,10 +11,8 @@
 
   <!-- Kafka -->
   <logger name="org.apache.kafka" level="ERROR" />
-
-  <variable name="WHISK_LOG_LEVEL" value="${WHISK_LOG_LEVEL:-INFO}" />
   
-  <root level="${WHISK_LOG_LEVEL}">
+  <root level="${logback.log.level:-INFO}">
     <appender-ref ref="console" />
   </root>
 </configuration>
--- a/common/scala/src/main/resources/logback.xml
+++ b/common/scala/src/main/resources/logback.xml
@@ -11,8 +11,10 @@
 
   <!-- Kafka -->
   <logger name="org.apache.kafka" level="ERROR" />
+
+  <variable name="WHISK_LOG_LEVEL" value="${WHISK_LOG_LEVEL:-INFO}" />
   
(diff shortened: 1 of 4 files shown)


rabbah:
is this something we can do with PureConfig given that we're heading in that direction vs ansible?

rabbah:
is this something we can do with PureConfig given that we're heading in that direction vs ansible?

mhenke1:
@rabbah As far as I understand the purpose of PureConfig, it is the meant to configure Scala/Akka code. This PR does not do any Scala/Akka related configuration. It only changes the logback configuration (see logback.xml) via a system property. So I do not see how PureConfig comes into play here.

common/scala/src/main/resources/logback.xml
```diff
@@ -11,8 +11,10 @@
 
   <!-- Kafka -->
   <logger name="org.apache.kafka" level="ERROR" />
+
+  <variable name="WHISK_LOG_LEVEL" value="${WHISK_LOG_LEVEL:-INFO}" />
   
-  <root level="info">
+  <root level="${WHISK_LOG_LEVEL}">
```
markusthoemmes: I guess what @rabbah was after, is that this could now be changed to:

```
// entirely omit the <variable> line
<root level="whisk.loglevel">
```

While defining:

```
-e CONFIG_whisk_loglevel=...
```

on the components.

Your call on what you prefer.

In any case: Is the double-definition even necessary in your case? Couldn't it be just:

```
<root level="${WHISK_LOG_LEVEL:-INFO}">
```
?

mhenke1:
@rabbah I realized that we have a misunderstanding regarding of the usage of the term PureConfig. 
I have now pushed a second commit that changed the passing of the log level as a JVM property using the 'PureConfig'-Format. 
As @markusthoemmes made me aware, this avoids to declare the variable in the logback.xml file and therefore keeps it more clean.

common/scala/src/main/resources/logback.xml
mhenke1: @markusthoemmes Changed (see my other comment)

PR #3044, opened 2017-12-01
add WhiskActivation.Outcome, hiding Either[ActivationId, WhiskActivation] impl

starpit:
Currently, the implementation of the datatype of the active ack return flow from the Invoker is transparent, and spread across a dozen files: `Either[ActivationId, WhiskActivation]`

this PR introduces `WhiskActivation.Outcome`, an opaque data type that hides the impl.

the goal of this PR is to enable future work that may change the impl, without those PRs each having to change a dozen files. let's pay that cost once, in this PR.

PG3 1424

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
@@ -123,6 +123,13 @@ object WhiskActivation
     with WhiskEntityQueries[WhiskActivation]
     with DefaultJsonProtocol {
 
+  /**
+   * The invoker's result for an activation might only be an ActivationId,
+   * e.g. in the case of RecordTooLargeException; see InvokerReactive
+   *
+   */
+  type Outcome = Either[ActivationId, WhiskActivation]
+
   private implicit val instantSerdes = new RootJsonFormat[Instant] {
     def write(t: Instant) = t.toEpochMilli.toJson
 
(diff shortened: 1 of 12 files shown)


common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
```diff
@@ -123,6 +123,13 @@ object WhiskActivation
     with WhiskEntityQueries[WhiskActivation]
     with DefaultJsonProtocol {
 
+  /**
+   * The invoker's result for an activation might only be an ActivationId,
+   * e.g. in the case of RecordTooLargeException; see InvokerReactive
+   *
```
rabbah: Or a non blocking invoke.

starpit:
i'm now a bit more skeptical of this PR. any change to `WhiskActivation.Outcome` that moves it away from a type alias will run afoul of the pervasive `Left` and `Right` idioms, not to mention the maps. i.e. the use points all assume that the impl is an `Either`, making it hard to refactor, unless we *also* abstract away the projections...

rabbah:
That was my point above. The point of using the  either type basically a union type here was to avoid additional overhead for serdes. It’s possibly to refactor the rest but you’re introducing effectively a new type.

starpit:
we could make a Left and Right constructor... i'll think about it a bit more. i think there's some value here

rabbah:
So you don’t have to change every location you can also just add a toEither method.

markusthoemmes:
@starpit what's the status of this? Is this still relevant?

markusthoemmes:
Closing this as abandoned for quite some time. @starpit let me know if you want to reopen.

PR #3045, opened 2017-12-01
Test updates needed in order to be able to test extra quote strings in the CLI

jessealva:


jessealva:
@dubeejw can you take a look at this when you get a chance? its some changes i need for apache/incubator-openwhisk-cli#173

PR #3047, opened 2017-12-01
cleanup: remove unused redis host/port from WhiskConfig

dgrove-oss:
I should have removed these when we switched to using zookeeper
as the persistent store for invoker id assignment.  They are not used.


rabbah:
pg?

dgrove-oss:
PG4 1096 🏃

dgrove-oss:
PG2 2438 🏃

dgrove-oss:
PG 2438 🔵

Issue #3022, opened 2017-11-29
services.KafkaConnectorTests.Kafka heisenbug


rabbah:
This issue is to collect the intermittent issues. We found three different types so far:
- [ ] `This server is not the leader for that topic-partition.`
- [ ] `services.KafkaConnectorTests > Kafka connector should send and receive a kafka message even after shutdown one of instances FAILED
    org.scalatest.exceptions.TestFailedException: 0 was not equal to 1`
- [ ] hanging while trying to retrieve a message https://github.com/apache/incubator-openwhisk/issues/3022#issuecomment-348948838

Observed one failure in jenkins:

```bash
services.KafkaConnectorTests.Kafka connector should send and receive a kafka message even after shutdown one of instances

Error Message

org.apache.kafka.common.errors.NotLeaderForPartitionException: This server is not the leader for that topic-partition.

Stacktrace

org.apache.kafka.common.errors.NotLeaderForPartitionException: This server is not the leader for that topic-partition.

Standard Output

[2017-11-28T15:07:19.923Z] [INFO] Slf4jLogger started
Received 1. Took 407 msec: List(Tue Nov 28 15:07:20 EST 2017)

Received 1. Took 4 msec: List(Tue Nov 28 15:07:20 EST 2017)

Received 1. Took 4 msec: List(Tue Nov 28 15:07:20 EST 2017)

Received 1. Took 4 msec: List(Tue Nov 28 15:07:20 EST 2017)

Received 1. Took 3 msec: List(Tue Nov 28 15:07:20 EST 2017)

Received 1. Took 4 msec: List(Tue Nov 28 15:07:20 EST 2017)

Received 2. Took 11 msec: List(Tue Nov 28 15:07:20 EST 2017, Tue Nov 28 15:07:31 EST 2017)

Received 3. Took 9 msec: List(Tue Nov 28 15:07:20 EST 2017, Tue Nov 28 15:07:31 EST 2017, Tue Nov 28 15:07:42 EST 2017)

Received 1. Took 4 msec: List(Tue Nov 28 15:07:42 EST 2017)

Received 1. Took 7 msec: List(Tue Nov 28 15:07:42 EST 2017)

Received 1. Took 10 msec: List(Tue Nov 28 15:07:42 EST 2017)

Received 1. Took 22 msec: List(Tue Nov 28 15:07:52 EST 2017)


```

jeremiaswerner:
@devbv I caught a stacktrace of a slightly different error but related
```
[...]

PR #3048, opened 2017-12-02
Reduce number of kafka instance temporarily

devbv:
Not to break master build status, reduce number of kafka to 1 until we
can ensure that we solve it.
There's heisenbug in KafkaConnectorTests(#3022) when we use two kafka instance and
also two kafka instances affect Travis with high load.


markusthoemmes:
Since we're fixing forward and this doesn't seem to impose a huge problem right now, closing for good. (Let me know if that's the wrong decision @devbv)

PR #3052, opened 2017-12-03
Allow for a docker pull bypass for docker actions where the prefix maches the invoker's docker prefix

rabbah:
This is a proposed patch to address the feature request in #2852.

The presence of an property `bypassPullForLocalImages` in the runtime manifest will toggle the feature on or off allowing a docker pull to be skipped if the docker action's image name has a prefix matches the local docker build prefix.

See https://github.com/apache/incubator-openwhisk/pull/3052/files#diff-d895f19f56edd5b725a1696b7b9d8877R196 for an example. 

@alexkli can you take a look?

### DIFF
--- a/ansible/environments/docker-machine/group_vars/all
+++ b/ansible/environments/docker-machine/group_vars/all
@@ -3,6 +3,7 @@ config_root_dir: /Users/Shared/wskconf
 whisk_logs_dir: /Users/Shared/wsklogs
 docker_registry: ""
 docker_dns: ""
+bypass_pull_for_local_images: true
 
 # The whisk_api_localhost_name is used to configure nginx to permit vanity URLs for web actions.
 # It is also used for the SSL certificate generation. For a local deployment, this is typically
--- a/ansible/environments/local/group_vars/all
+++ b/ansible/environments/local/group_vars/all
@@ -3,6 +3,7 @@ config_root_dir: /tmp/wskconf
 whisk_logs_dir: /tmp/wsklogs
 docker_registry: ""
 docker_dns: ""
+bypass_pull_for_local_images: true
 
 db_prefix: whisk_local_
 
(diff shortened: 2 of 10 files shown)


common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -122,26 +122,6 @@ object WhiskConfig {
    */
   def readFromEnv(key: String): Option[String] = sys.env.get(asEnvVar(key))
 
-  private def whiskPropertiesFile: File = {
```
rabbah: this file is now only used for tests and tests should use `WhiskProperties` which already have to discover this file.

ansible/environments/docker-machine/group_vars/all
```diff
@@ -3,6 +3,7 @@ config_root_dir: /Users/Shared/wskconf
 whisk_logs_dir: /Users/Shared/wsklogs
 docker_registry: ""
 docker_dns: ""
+bypass_pull_for_local_images: true
```
rabbah: @csantanapr what do you think about making this true for the local envs?

csantanapr: Yes make it true for local 👍

alexkli:
@rabbah I don't have a chance to test the PR, but IIUC:

1. inside the manifest (where is that btw, i have never configured it?):
   * set `bypassPullForLocalImages` to` true`
   * set `defaultImagePrefix` to e.g. `whisk` 
2. give the docker images you develop that prefix in their tag name, e.g. `whisk/*`
3. start OW
4. then it would never pull images with this prefix for `--docker` actions

I think that should work for the local dev/testing or demo use case.

rabbah:
If you’re using the local or Mac environments then 1 is done for you automatically.

PR #3053, opened 2017-12-04
Add init time to activation record.

rabbah:
Annotate an activation record with its init time if it's a cold start. This allows a client to know if an action was cold or warm and surfaces the time spent in init.


rabbah:
Rebased - conflict free.

common/scala/src/main/scala/whisk/core/containerpool/Container.scala
```diff
@@ -191,6 +191,7 @@ case class RunResult(interval: Interval, response: Either[ContainerConnectionErr
   def ok = response.right.exists(_.ok)
   def toBriefString = response.fold(_.toString, _.toString)
 }
+
 object Interval {
 
   /** An interval starting now with zero duration. */
```
rabbah: why is zero a def not a val?

rabbah:
@csantanapr can you review the doc change https://github.com/rabbah/openwhisk/blob/6daea94b81563e6c1ec61c8f9ffb7bfe1ab7a45c/docs/annotations.md#annotations-specific-to-activations for clarity?

PR #3058, opened 2017-12-05
Fix Kafka values in WhiskConfig

cbickel:
Kafka settings are not taken from the environment, because the Strings in WhiskConfig are wrong. This PR fixes this.

Setting`retention.ms` value with ansible and printing the map:

with master:
`Map(retention.bytes -> 1073741824, retention.ms -> 3600000, replicationFactor -> 1, numPartitions -> 1, segment.bytes -> 536870912)`

with this PR:
`Map(retention.bytes -> 1073741824, retention.ms -> 300000, replicationFactor -> 1, numPartitions -> 1, segment.bytes -> 536870912)`

### DIFF
--- a/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
+++ b/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
@@ -242,12 +242,12 @@ object WhiskConfig {
 
   val runtimesManifest = "runtimes.manifest"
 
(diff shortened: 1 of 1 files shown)


dgrove-oss:
#2991 moves the kafka configuration values to pureconfig.  I suggest we merge it first, then fix the values in kafka.conf.

Issue #3057, opened 2017-12-05
Provide a way to refer to extra jars in launch script classpath


chetanmeh:
Currently the launch scripts generated for controller and invoker refer to an explicit classpath which enumerates all the existing dependencies of these modules.

For certain scenarios its desirable to add custom jars to the classpath. For e.g. by default openwhisk bundles kamon-statsd but in our environment we need to use kamon-datadog. To enable that we need to include the kamon-datadog jars in controller classpath.

### Proposal

To enable that launchscript can provide a wildcard pattern 'extra-libs/*'. Then in our custom image we can add extra jars to this directory which would get picked up at runtime

    CLASSPATH=$APP_HOME/lib/akka-slf4j_2.11-2.5.6.jar:$APP_HOME/extra-libs/*

### Possible Approach

Checking in Gradle it does not appear to provide a simple way to add such custom entry. (Raised query [here][1]). So for now we can go with an approach like [here][2]

[1]: https://discuss.gradle.org/t/adding-wildcard-classpath-entry-in-launch-script/25010
[2]: http://mrhaki.blogspot.in/2015/04/gradle-goodness-alter-start-scripts.html

chetanmeh:
Open PR #3059 for this.

To see usage refer to [example][1] which provides a custom docker image which extends whisk/controller and adds Kamon JMX support to it

[1]: https://github.com/chetanmeh/ow-classpath-example

chetanmeh:
Done via #3059

PR #3059, opened 2017-12-05
Modify the unix script

chetanmeh:
Modifies the line containing `CLASSPATH=` in generated unix script file and add entries for

1. `$APP_HOME/ext-lib/*` - To be used to add extra jars to classpath
2. `$APP_HOME/config` - To be used to add custom application.conf

This PR is for issue #3057


build.gradle
```diff
@@ -10,4 +10,29 @@ buildscript {
...
+
+    afterEvaluate {
+        if (project.plugins.hasPlugin('application')) {
+            startScripts {
+                doLast {
+                    unixScript.text = configureUnixClasspath(unixScript)
+                }
+            }
+        }
+    }
```
markusthoemmes: I googled for a bit and stumbled upon:

```groovy
startScripts {
  classpath += files('ext-libs/*', 'config')
}
```

There seems to be a possibility to do this other than manipulating raw lines (which feels a bit rough). I'd need to look further into this, since gradle seems to apply some sanitizing unfortunately.

chetanmeh: Thanks @markusthoemmes  for looking into this. I also saw that but it does not work as expected. Have a look at links provided in #3057 for background and hence used this approach

markusthoemmes: @chetanmeh oh sorry, should've checked. Yeah your observations are exactly the same as mine, so we're clear on that 👍

build.gradle
```diff
@@ -10,4 +10,29 @@ buildscript {
...
+def configureUnixClasspath(File script) {
+    script
+        .readLines()
+        .collect { line ->
+            // Looking for the line that starts with CLASSPATH=
+            line = line.replaceAll(~/^CLASSPATH=.*$/) { original ->
+
+            // Get original line and append it
+            // with the configuration directory.
+            original += ":\$APP_HOME/ext-lib/*:\$APP_HOME/config"
```
markusthoemmes: Could use single quotes to circumenvent character escaping?

chetanmeh: Good point. This was leftover of previous attempt which was also using a variable for path separator. With that not used we can use single quote. Updated that.

docs/spi.md
```diff
@@ -62,3 +62,15 @@ Since SPI implementations are loaded from the classpath, and a specific implemen
 * Include all implementations, and only use the specified implementations.
 * Include some combination of defaults and alternative implementations, and use the specified implementations for the alternatives, and default implementations for the rest.
 
+## Including the implementation
+
+Base openwhisk docker images provide 2 extension points in classpath for including the implementation
```
markusthoemmes: - in *the* classpath
- missing `.` at the end.

docs/spi.md
```diff
@@ -62,3 +62,15 @@ Since SPI implementations are loaded from the classpath, and a specific implemen
...
+
+### Application Jars
+
+The application jars can be added to `$APP_HOME/ext-lib` for e.g. in `openwhisk/controller` image the implementation jars can be added to `/controller/ext-lib` folder and for `openwhisk/invoker` they can be added to `/invoker/ext-lib` folder
```
markusthoemmes: - to `/controller/ext-lib` ~~folder~~
- to `/invoker/ext-lib` ~~folder~~
- Missing `.` at the end.

docs/spi.md
```diff
@@ -62,3 +62,15 @@ Since SPI implementations are loaded from the classpath, and a specific implemen
...
+
+### Application Configuration
+
+The configuration files can be added to `$APP_HOME/config` folder
```
markusthoemmes: - ~~folder~~
- Missing `.` at the end.

chetanmeh:
Updated docs as per feedback

Issue #3060, opened 2017-12-05
Generate maven artifacts


chetanmeh:
To support custom SPI implementations which cannot be part of main repo it would be useful to generate maven artifacts for common, controller and invoker modules.

To start with we can support creating them locally and then figure out way to push the snapshots to [apache repository][1]

[1]: https://repository.apache.org/content/repositories/snapshots

chetanmeh:
Took a first try with #3061 Please share your feedback for the approach? Should we move maven related logic to gradle/maven.gradle and import that or use current approach

chetanmeh:
Maven artifacts are now getting generated but currently not pushed. Once the release process is implemented they would be pushed to Maven central. Resolving this issue for now

PR #3061, opened 2017-12-05
Add maven plugin to build.grade

chetanmeh:
- Adds `maven` plugin to build.gradle files for common, controller and
invoker.
- Enables support for generating source jars
- Uses `org.apache.openwhisk` as groupId
- Uses `0.0.1-SNAPSHOT` as current version

PR for #3060

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,8 +30,14 @@ subprojects {
                 from sourceSets.main.allSource
             }
 
+            task testSourcesJar(type: Jar, dependsOn: testClasses) {
+                classifier = 'test-sources'
+                from sourceSets.test.allSource
+            }
+
             artifacts {
                 archives sourcesJar
+                archives testSourcesJar
             }
         }
     }
--- a/build.gradle
+++ b/build.gradle
@@ -11,6 +11,9 @@ subprojects {
     apply plugin: 'scalafmt'
     scalafmt.configFilePath = gradle.scalafmt.config
 
+    group 'org.apache.openwhisk'
+    version '0.0.1-SNAPSHOT'
+
     afterEvaluate {
         if (project.plugins.hasPlugin('application')
                 && project.plugins.hasPlugin('scala')) {
@@ -18,6 +21,17 @@ subprojects {
                 doLast {
                     unixScript.text = configureUnixClasspath(unixScript)
                 }
+            }
+        }
+
+        if (project.plugins.hasPlugin('maven')) {
+            task sourcesJar(type: Jar, dependsOn: classes) {
+                classifier = 'sources'
+                from sourceSets.main.allSource
+            }
+
+            artifacts {
+                archives sourcesJar
             }
         }
     }
--- a/build.gradle
+++ b/build.gradle
@@ -40,6 +40,22 @@ subprojects {
                 archives testSourcesJar
             }
         }
+
+        if (project.plugins.hasPlugin('application')) {
+            //Ensure that dist archive name does not contain version
+            distTar {
+                archiveName = "${project.name}.tar"
+            }
+
+            //Avoid generating the zip files from maven installations
+            distZip {
+                enabled false
+            }
+
+            configurations.archives.with {
+                artifacts.remove artifacts.find { it.archiveTask.is distZip }
+            }
+        }
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -35,9 +35,15 @@ subprojects {
                 from sourceSets.test.allSource
             }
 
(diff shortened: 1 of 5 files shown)


markusthoemmes:
Awesome, thanks for taking this on! 🎉 

Any chance to do the same to our tests, so the external repos can rely on the published jar rather than the whole repository here?

chetanmeh:
> Any chance to do the same to our tests, so the external repos can rely on the published jar rather than the whole repository here?

Missed including that. Updated the PR to also publish jars for "tests" module. Its sources are published with classifier `test-sources`

csantanapr:
I was thinking using Jenkins build script that today pushes images on every master build already.

csantanapr:
Asked @chetanmeh to update the jenkinsfile with this PR or seperate PR

chetanmeh:
> Adding documentation on how to use these tasks would be very valuable so we can use it in the future to actually publish stuff. (+ I can verify locally that it works as the tasks are unused for now)

@markusthoemmes  The doc discussion was more for #3059. To try the change locally you just need to run. Later once its integrated with jenkins end users would typically not need to do it locally. So not sure if we need to document this in docs

    ./gradlew :core:controller:install

And then you can see controller artifacts at ~/.m2/repository/org/apache/openwhisk/controller

> Asked @chetanmeh to update the jenkinsfile with this PR or seperate PR

@csantanapr Sure would do this in a separate PR

chetanmeh:
Few references for changes in main build.gradle

The use of maven plugin somehow also enables the zipDist and hence gradle starts creating and publishing the distribution in zip format also in addition to tar. So to disable that 2 steps are required

```
  //Avoid generating the zip files from maven installations
            distZip {
                enabled false
            }

            configurations.archives.with {
                artifacts.remove artifacts.find { it.archiveTask.is distZip }
            }
```

First part disables distZip but then it cause issue in install. So zip artifact needs to be explicitly removed. See discussion [here][1]

[1]: https://discuss.gradle.org/t/how-to-exclude-zip-only-from-archives-for-maven-repo-upload/13354/2

chetanmeh:
Also opened #3082 to ensure that addition of new maven artifacts do not increase the build context for creating docker images

chetanmeh:
@markusthoemmes  Anything more to be done related to this PR before it can be merged?

chetanmeh:
@markusthoemmes  I have rebased the PR to current master. Can this PR be merged now?

markusthoemmes:
@csantanapr A quick glance from your release-experienced view please?

chetanmeh:
@csantanapr Can you have a look here so that this PR can be merged?

chetanmeh:
Note that for common/scala module the jar name is `scala-0.0.1-SNAPSHOT.jar`. Below is how this is referred in a different module

```
dependencies {
    compile "org.apache.openwhisk:scala:0.0.1-SNAPSHOT"
}
```
This is because of module dir name being `scala`. Should be change it to common-xxx.jar or something else?

csantanapr:
>This is because of module dir name being scala. Should be change it to common-xxx.jar or something else?

Yeah I think it would be better to call it `common`

chetanmeh:
After discussing with @csantanapr  modified the build to produce jars following pattern "openwhisk-<module>-xxx.jar". So with this change we produce following artifacts

1. commons/scala 
    - openwhisk-common-\<version\>.jar
    - openwhisk-common-\<version\>-sources.jar
    - openwhisk-common-\<version\>-tests.jar
    - openwhisk-common-\<version\>-test-sources.jar
2. core/invoker
    - openwhisk-invoker-\<version\>.jar
    - openwhisk-invoker-\<version\>-sources.jar
    - openwhisk-invoker-\<version\>-tests.jar
    - openwhisk-invoker-\<version\>-test-sources.jar 
3. core/controller 
    - openwhisk-controller-\<version\>.jar
    - openwhisk-controller-\<version\>-sources.jar
    - openwhisk-controller-\<version\>-tests.jar
    - openwhisk-controller-\<version\>-test-sources.jar 
4. tests 
    - openwhisk-tests-\<version\>.jar
    - openwhisk-tests-\<version\>-sources.jar
    - openwhisk-tests-\<version\>-tests.jar
    - openwhisk-tests-\<version\>-test-sources.jar 

With this running `gradlew install` would install these jars to _~/.m2/repository/org/apache/openwhisk_ folder

Note that this currently does not add support for deploy and publishing. That would need to be done in a separate PR. It would only install the artifacts in local maven repo

csantanapr:
Playground4/1287 🔵

PR #3062, opened 2017-12-05
Update trigger fire REST API swagger

mdeuser:
Successful POST `/namespaces/{namespace}/triggers/{triggerName}` responds with 202 instead of 200

Related to PR #3031


csantanapr:
FWIW we should have tests that validate this type of things, that API and swagger are sync/contract

PR #3063, opened 2017-12-05
use alpine linux for base image

tysonnorris:


markusthoemmes:
Good one 👍 

Two questions though:
1. How is the build time in relation to before? We mainly moved to trusty-curl because all the prereqs were already there and updating repositories took quite a while (relevant because we build this image each time in travis)
2. Is JVM tooling working under alpine? I.e: `jstack`.

If both of those are fine, I'd say it's ok to move.

tysonnorris:
1. I don't notice much, if any, difference in time locally; travis for this PR took 40 min 29 sec (which seems on par with other builds taking 39-42 minutes)
2. size is better, less than half:  controller 283mb (vs 711mb) invoker 298mb vs (726mb)
3. jstack works - I haven't tested "all tooling" though

core/controller/Dockerfile
```diff
@@ -1,13 +1,11 @@
...
 # Install swagger-ui
-RUN wget --no-verbose https://github.com/swagger-api/swagger-ui/archive/v2.1.4.tar.gz && \
+RUN curl -sSL -o swagger-ui-v2.1.4.tar.gz --no-verbose https://github.com/swagger-api/swagger-ui/archive/v2.1.4.tar.gz && \
     mkdir swagger-ui && \
-    tar zxf v2.1.4.tar.gz -C /swagger-ui --strip-components=2 swagger-ui-2.1.4/dist && \
-    rm v2.1.4.tar.gz && \
-    perl -pi -w -e  's{http://petstore.swagger.io/v2/swagger.json}{/api/v1/api-docs}g;' /swagger-ui/index.html
+    tar zxf swagger-ui-v2.1.4.tar.gz -C /swagger-ui --strip-components=2 swagger-ui-2.1.4/dist && \
+    rm swagger-ui-v2.1.4.tar.gz && \
+    sed -i s#http://petstore.swagger.io/v2/swagger.json#/api/v1/api-docs#g /swagger-ui/index.html
```
rabbah: i notice this is also in another pr #3064?

rabbah:
BUILD SUCCESSFUL in 3m 28s for me - but I already had the alpine image locally.

core/controller/Dockerfile
tysonnorris: #3064 was supposed to only update swagger version

tysonnorris:
also, in travis, last successful scala (base image) build is 41s; with alpine was 53s.

I expect this base image build to be the only significant build timing diff.

csantanapr:
@tysonnorris why not keep using debian based, and use a newer version?

common/scala/Dockerfile
```diff
@@ -1,8 +1,5 @@
-FROM buildpack-deps:trusty-curl
+FROM alpine:3.6
```
ddragosd: WDYT about extending from an existing JAVA based image from https://hub.docker.com/_/openjdk/ ?

tysonnorris: switching jdk from oracle to openjdk should be a separate topic, I think; I'm not sure all the ramifications there, I believe @bwmcadams ran into some issue with openjdk in recent times?

tysonnorris:
@csantanapr mainly size; alpine is significantly smaller

common/scala/Dockerfile
markusthoemmes: I tried OpenJDK once and it was fine at that time. FWIW you're right though, that should be an entirely seperate discussion.

rabbah:
@vvraskin you might want to check given your open pr on jmx. 
Otherwise I think this makes the image leaner, that’s generally good.

common/scala/Dockerfile
jonpspri: Agreed there needs to be a conversation around JDK.  Oracle doesn't support the s390x architecture I'm working on; I am working with openjdk there but haven't completed regression testing.

vvraskin:
@rabbah Thanks for the hint, the change seems to be orthogonal to jmx. I've rebased mine on the Tyson's branch, all works well.

tysonnorris:
Can one of you give this an official github review to allow merging?

tysonnorris:
@csantanapr can you merge this?

csantanapr:
Hum I need some bit of time to take a look at other images we use internally that we extend from common/scala image, that are failing when I tried this PR in our internal pipeline.

tysonnorris:
@csantanapr any update on this? please merge when you are ready

common/scala/Dockerfile
```diff
@@ -11,24 +8,68 @@ ENV VERSION 8
...
+           "$JAVA_HOME"/jre/lib/ext/jfxrt.jar \
+           "$JAVA_HOME"/jre/lib/ext/nashorn.jar \
+           "$JAVA_HOME"/jre/lib/oblique-fonts \
+           "$JAVA_HOME"/jre/lib/plugin.jar \
+           /tmp/* /var/cache/apk/* && \
+    echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \
+    # https://github.com/anapsix/docker-alpine-java/issues/18#issue-167437838
+    # If you want to change the default values here, also change the values in 042-set-jvm-ttl.sh
+    sed -i s/#networkaddress.cache.ttl=-1/networkaddress.cache.ttl=10/ ${JAVA_HOME}/jre/lib/security/java.security && \
+    sed -i s/networkaddress.cache.negative.ttl=10/networkaddress.cache.negative.ttl=0/ ${JAVA_HOME}/jre/lib/security/java.security
```
markusthoemmes: Any reason why we copy https://hub.docker.com/r/anapsix/alpine-java/ instead of just building `FROM` it? Seems like we have a 1:1 copy or am I missing something?

tysonnorris: Discussed with Markus: the alpine-java image does not have a version that is alpine 3.6 + jdk 8u141b15

I will create a separate PR to use that base image, however this will typically require revving alpine version AND jdk at same time. I'm OK with either approach, since typically neither will rev frequently.

PR #3072, opened 2017-12-07
Enhance kafka message provider

### LINKED ISSUE #3022: services.KafkaConnectorTests.Kafka heisenbug

devbv:
- Make sessionTimeout configurable
- Handle exception from Kafka broker down
- Reuse docker command functions

Resolve #3022

### DIFF
--- a/common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
+++ b/common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
@@ -19,23 +19,34 @@ package whisk.connector.kafka
 
 import java.util.Properties
 
-import scala.collection.JavaConversions.iterableAsScalaIterable
-import scala.collection.JavaConversions.seqAsJavaList
-import scala.concurrent.duration.Duration
-import scala.concurrent.duration.DurationInt
-import org.apache.kafka.clients.consumer.ConsumerConfig
-import org.apache.kafka.clients.consumer.KafkaConsumer
+import akka.actor.ActorSystem
+import org.apache.kafka.clients.consumer.{ConsumerConfig, KafkaConsumer}
+import org.apache.kafka.common.errors.{RetriableException, WakeupException}
 import org.apache.kafka.common.serialization.ByteArrayDeserializer
 import pureconfig.loadConfigOrThrow
 import whisk.common.Logging
 import whisk.core.ConfigKeys
 import whisk.core.connector.MessageConsumer
 
-class KafkaConsumerConnector(kafkahost: String,
-                             groupid: String,
-                             topic: String,
-                             override val maxPeek: Int = Int.MaxValue)(implicit logging: Logging)
+import scala.collection.JavaConversions.{iterableAsScalaIterable, seqAsJavaList}
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
+
+case class KafkaConsumerConfig(sessionTimeoutMs: Long)
+
+class KafkaConsumerConnector(
+  kafkahost: String,
+  groupid: String,
+  topic: String,
+  override val maxPeek: Int = Int.MaxValue)(implicit logging: Logging, actorSystem: ActorSystem)
     extends MessageConsumer {
+
+  implicit val ec: ExecutionContext = actorSystem.dispatcher
+  private val gracefulWaitTime = 100.milliseconds
+
+  // The consumer is generally configured via getProps. This configuration only loads values necessary for "outer"
+  // logic, like the wakeup timer.
+  private val cfg = loadConfigOrThrow[KafkaConsumerConfig](ConfigKeys.kafkaConsumer)
 
   /**
    * Long poll for messages. Method returns once message are available but no later than given
@@ -43,19 +54,51 @@ class KafkaConsumerConnector(kafkahost: String,
    *
    * @param duration the maximum duration for the long poll
    */
-  override def peek(duration: Duration = 500.milliseconds) = {
-    val records = consumer.poll(duration.toMillis)
-    records map { r =>
-      (r.topic, r.partition, r.offset, r.value)
-    }
+  override def peek(duration: FiniteDuration = 500.milliseconds,
+                    retry: Int = 3): Iterable[(String, Int, Long, Array[Byte])] = {
+
+    // poll can be infinitely blocked in edge-cases, so we need to wakeup explicitly.
+    val wakeUpTask = actorSystem.scheduler.scheduleOnce(cfg.sessionTimeoutMs.milliseconds + 1.second)(consumer.wakeup())
+
+    try {
+      consumer.poll(duration.toMillis).map(r => (r.topic, r.partition, r.offset, r.value))
+    } catch {
+      // Happens if the peek hangs.
+      case _: WakeupException if retry > 0 =>
+        logging.error(this, s"poll timeout occurred. Retrying $retry more times.")
+        Thread.sleep(gracefulWaitTime.toMillis) // using Thread.sleep is okay, since `poll` is blocking anyway
+        peek(duration, retry - 1)
+      case e: RetriableException if retry > 0 =>
+        logging.error(this, s"$e: Retrying $retry more times")
+        wakeUpTask.cancel()
+        Thread.sleep(gracefulWaitTime.toMillis) // using Thread.sleep is okay, since `poll` is blocking anyway
+        peek(duration, retry - 1)
+      // Every other error results in a restart of the consumer
+      case e: Throwable =>
+        recreateConsumer()
+        throw e
+    } finally wakeUpTask.cancel()
   }
 
   /**
    * Commits offsets from last poll.
    */
-  def commit() = consumer.commitSync()
+  def commit(retry: Int = 3): Unit =
+    try {
+      consumer.commitSync()
+    } catch {
+      case e: RetriableException =>
+        if (retry > 0) {
+          logging.error(this, s"$e: retrying $retry more times")
+          Thread.sleep(gracefulWaitTime.toMillis) // using Thread.sleep is okay, since `commitSync` is blocking anyway
+          commit(retry - 1)
+        } else {
+          throw e
+        }
+    }
 
-  override def close() = {
+  override def close(): Unit = {
+    consumer.close()
     logging.info(this, s"closing '$topic' consumer")
   }
 
@@ -79,9 +122,18 @@ class KafkaConsumerConnector(kafkahost: String,
     val keyDeserializer = new ByteArrayDeserializer
     val valueDeserializer = new ByteArrayDeserializer
     val consumer = new KafkaConsumer(props, keyDeserializer, valueDeserializer)
-    topics foreach { consumer.subscribe(_) }
+    topics.foreach(consumer.subscribe(_))
     consumer
   }
 
-  private val consumer = getConsumer(getProps, Some(List(topic)))
+  private def recreateConsumer(): Unit = {
+    val oldConsumer = consumer
+    Future {
+      oldConsumer.close()
+      logging.info(this, s"old consumer closed")
+    }
+    consumer = getConsumer(getProps, Some(List(topic)))
+  }
+
+  @volatile private var consumer = getConsumer(getProps, Some(List(topic)))
 }
--- a/common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
+++ b/common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
@@ -19,36 +19,38 @@ package whisk.connector.kafka
 
 import java.util.Properties
 
-import scala.concurrent.ExecutionContext
-import scala.concurrent.Future
-import scala.concurrent.Promise
-import scala.util.Failure
-import scala.util.Success
-import org.apache.kafka.clients.producer.Callback
-import org.apache.kafka.clients.producer.KafkaProducer
-import org.apache.kafka.clients.producer.ProducerConfig
-import org.apache.kafka.clients.producer.ProducerRecord
-import org.apache.kafka.clients.producer.RecordMetadata
-import org.apache.kafka.common.errors.NotLeaderForPartitionException
+import akka.actor.ActorSystem
+import akka.pattern.after
+import org.apache.kafka.clients.producer._
+import org.apache.kafka.common.errors.{
+  NotEnoughReplicasAfterAppendException,
+  RecordTooLargeException,
+  RetriableException,
+  TimeoutException
+}
 import org.apache.kafka.common.serialization.StringSerializer
-import whisk.common.Counter
-import whisk.common.Logging
-import whisk.core.connector.Message
-import whisk.core.connector.MessageProducer
-import whisk.core.entity.UUIDs
 import pureconfig._
+import whisk.common.{Counter, Logging, TransactionId}
 import whisk.core.ConfigKeys
+import whisk.core.connector.{Message, MessageProducer}
+import whisk.core.entity.UUIDs
 
-class KafkaProducerConnector(kafkahosts: String,
-                             implicit val executionContext: ExecutionContext,
-                             id: String = UUIDs.randomUUID().toString)(implicit logging: Logging)
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.util.{Failure, Success}
+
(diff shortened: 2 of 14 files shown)


markusthoemmes:
Can you provide tests for the failure scenarios you implemented here?

In general I have the feeling that we're moving into a direction here, which should not really be our business. That is: Write a resilient client to talk to Kafka/a Kafka cluster. Personally, I don't grasp all the error-cases implemented here and neither do I know if the code is complete in it's error handling.

A possibility is to move to a well tested client (like https://github.com/cakesolutions/scala-kafka-client or https://doc.akka.io/docs/akka-stream-kafka/current/home.html). Quite frankly that will be a larger operation than this so I'm a bit unsure on how to proceed. What do you think?

devbv:
@markusthoemmes @jeremiaswerner 
I can't write test case for almost scenarios because they are intermittent issues from Kafka fail-over and rebalancing and it is so hard to recreate it manually. 
My purpose was trying to success all user request by handling exceptions and adding fallbacks, even there are node fails in kafka cluster.
If it is ok to allow all errors and response user request as a internal error when kafka issues occur, I definitely agree with that we don't need to handle exceptions and it's quite large operation. So if you thought in that way, I would get rid of all error handling and retry logics!
And in that case, our kafka shooting up test should be changed to something like "Allow at most 3 errors for 10 request".
So give me opinons, I'll gladly follow yours.

markusthoemmes:
Okay, that makes sense.

The major difference from the better-known clients is, that their error behavior is more principled (i.e. controllable via a `SupervisionStrategy`).

In any case: If we make sure that:

1. Errors we know of  are handled
2. Errors we don't know of just result in a restart of the consumer/producer

I guess we are fine.

I'll review more thoroughly tomorrow. Testing new clients/completely moving away from our own ones is a much larger task that should be seen orthogonal.

devbv:
@markusthoemmes I think some types of RetriableException could duplicate the message if we tried to send a message again. I'm pretty sure that InvalidMetadata guarantees that we really fail to send a message, but cannot sure it when TimeoutException is raised.

And you can see here, producer retries will cause duplicates for NotEnoughReplicasAfterAppendException
https://kafka.apache.org/0100/javadoc/org/apache/kafka/common/errors/NotEnoughReplicasAfterAppendException.html

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -21,46 +21,106 @@ import java.util.Properties
...
    * @param duration the maximum duration for the long poll
    */
-  override def peek(duration: Duration = 500.milliseconds) = {
-    val records = consumer.poll(duration.toMillis)
-    records map { r =>
-      (r.topic, r.partition, r.offset, r.value)
+  override def peek(duration: Duration = 500.milliseconds,
+                    retry: Int = 3): Iterable[(String, Int, Long, Array[Byte])] = {
+    // Since kafka client can be infinitely blocked in poll(timeout), we should interrupt it by wakeup() method
+    val wakeUpTask = actorSystem.scheduler.scheduleOnce(sessionTimeout) {
```
markusthoemmes: Why `sessionTimeout`? Shouldn't we do `duration + someGraceTime`?

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -21,46 +21,106 @@ import java.util.Properties
...
+import org.apache.kafka.common.errors.{InvalidMetadataException, WakeupException}
+import akka.actor.ActorSystem
+import org.apache.kafka.common.KafkaException
 import whisk.common.Logging
 import whisk.core.connector.MessageConsumer
 
 class KafkaConsumerConnector(kafkahost: String,
                              groupid: String,
                              topic: String,
+                             actorSystem: ActorSystem,
```
markusthoemmes: You might want to pass this in as an implicit parameter.

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -21,46 +21,106 @@ import java.util.Properties
...
 
   /**
    * Commits offsets from last poll.
    */
-  def commit() = consumer.commitSync()
+  def commit(retry: Int = 3): Unit =
+    try {
+      consumer.commitSync()
+    } catch {
+      case error: InvalidMetadataException =>
```
markusthoemmes: Can we use `RetrieableException` here instead?

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -21,46 +21,106 @@ import java.util.Properties
...
+    // Since kafka client can be infinitely blocked in poll(timeout), we should interrupt it by wakeup() method
+    val wakeUpTask = actorSystem.scheduler.scheduleOnce(sessionTimeout) {
+      consumer.wakeup()
     }
+
+    try {
+      val records = consumer.poll(duration.toMillis)
+      records map { r =>
+        (r.topic, r.partition, r.offset, r.value)
+      }
```
markusthoemmes: FWIW: This can be 

```scala
consumer.poll(duration.toMillis).map(r => (r.topic, r.partition, r.offset, r.value))
```

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -62,6 +62,7 @@
       "CONFIG_whisk_kafka_topics_health_retentionBytes": "{{ kafka_topics_health_retentionBytes | default() }}"
       "CONFIG_whisk_kafka_topics_health_retentionMs": "{{ kafka_topics_health_retentionMS | default() }}"
       "CONFIG_whisk_kafka_topics_health_segmentBytes": "{{ kafka_topics_health_segmentBytes | default() }}"
+      "KAFKA_SESSIONTIMEOUT": "{{ kafka.sessionTimeout }}"
```
markusthoemmes: Please move that line below KAFKA_HOSTS.

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -21,46 +21,106 @@ import java.util.Properties
...
+       * 1. Error we know of can be handled by retry
+       *  - WakeupException: Exception we raise intentionally to prevent hanging in consumer.poll(timeout)
+       *  - InvalidMetadataException: An exception that may indicate the client's metadata is out of date,
+       *                              includes NetworkException, KafkaStorageException, LeaderNotAvailableException,
+       *                              NotLeaderForPartitionException, UnknownTopicOrPartitionException.
+       *
+       * 2. Errors we don't know of just result in a restart of the consumer/producer
+       *
+       */
+      case e: KafkaException =>
```
markusthoemmes: As discussed: Can we just go with `RetriableException`  in the KafkaConsumer?

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -21,46 +21,106 @@ import java.util.Properties
...
+        e match {
+          case _: WakeupException =>
+            if (retry > 0) {
+              logging.error(this, s"Poll timeout occurred, remain $retry retries to peek")
+              peek(duration, retry - 1)
```
markusthoemmes: Do we want/need to delay the retry for a bit?

common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
```diff
@@ -66,12 +70,33 @@ class KafkaProducerConnector(kafkahosts: String,
...
+       *
+       * 1. Error we know of can be handled by retry
+       *  - InvalidMetadataException: An exception that may indicate the client's metadata is out of date,
+       *                              includes NetworkException, KafkaStorageException, LeaderNotAvailableException,
+       *                              NotLeaderForPartitionException, UnknownTopicOrPartitionException.
+       *
+       * 2. Errors we don't know of just result in a restart of the consumer/producer
+       *
+       */
+      case e: KafkaException =>
```
markusthoemmes: Can we use `RetriableException` and exclude the ones we deem harmful?

```scala 
case _: NotEnoughReplicasAfterAppendException =>
  recreateProducer()
  produced.future
case _: RetriableException =>
  ...
case _ =>
  recreateProducer()
  produced.future
```

common/scala/src/main/scala/whisk/core/connector/MessagingProvider.scala
```diff
@@ -31,6 +33,7 @@ trait MessagingProvider extends Spi {
   def getConsumer(config: WhiskConfig,
                   groupId: String,
                   topic: String,
+                  actorSystem: ActorSystem,
```
markusthoemmes: Want to move this to the implicit section?

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -21,46 +21,106 @@ import java.util.Properties
...
+            } else {
+              recreateConsumer()
+              throw e
+            }
+          case _: InvalidMetadataException =>
+            if (retry > 0) {
+              logging.error(this, s"${e.getMessage}: remain $retry retries to peek")
+              wakeUpTask.cancel()
```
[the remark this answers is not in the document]
devbv: Not really, each retry creates own wakeUpTask so we should cancel it every retry.

common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
devbv: Great idea

devbv:
After rebase, a failed test would be resolved.
(https://github.com/apache/incubator-openwhisk/commit/50c0142e9543b38ba2807b1c734c5af6c3116d72)

devbv:
@markusthoemmes Everything is ok

markusthoemmes:
@rabbah can you please sanity check this?

ansible/group_vars/all
```diff
@@ -135,6 +135,7 @@ kafka:
     port: 8093
   heap: "{{ kafka_heap | default('1g') }}"
   replicationFactor: "{{ kafka_replicationFactor | default((groups['kafkas']|length)|int) }}"
+  sessionTimeout: 30000
```
rabbah: do these values make sense as pure config props as well?

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -21,46 +21,91 @@ import java.util.Properties
...
    * Commits offsets from last poll.
    */
-  def commit() = consumer.commitSync()
+  def commit(retry: Int = 3): Unit =
+    try {
+      consumer.commitSync()
+    } catch {
+      case e: RetriableException =>
+        if (retry > 0) {
+          logging.error(this, s"$e: retrying $retry more times")
```
rabbah: (nit warning ahead): how about `retries left: $retry` to avoid this reading `retrying 1 more times` here and elsewhere above and below.

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -89,9 +134,18 @@ class KafkaConsumerConnector(kafkahost: String,
...
+  private def recreateConsumer(): Unit = {
+    val oldConsumer = consumer
+    Future {
+      oldConsumer.close()
+      logging.info(this, s"old consumer closed")
+    }
+    consumer = getConsumer(getProps, Some(List(topic)))
+  }
+
+  @volatile private var consumer = getConsumer(getProps, Some(List(topic)))
```
rabbah: i'm not sure that this needs to be volatile (and hence incur the performance implications); in the architecture (controller's pings and active acks, or invoker activation feed) there is one actor that is connected to the message consumer and it serializes all the reads from the message bus.

common/scala/src/main/scala/whisk/core/database/RemoteCacheInvalidation.scala
```diff
@@ -60,8 +60,9 @@ class RemoteCacheInvalidation(config: WhiskConfig, component: String, instance:
   private val instanceId = s"$component${instance.toInt}"
 
   private val msgProvider = SpiLoader.get[MessagingProvider]
-  private val cacheInvalidationConsumer = msgProvider.getConsumer(config, s"$topic$instanceId", topic, maxPeek = 128)
-  private val cacheInvalidationProducer = msgProvider.getProducer(config, ec)
+  private val cacheInvalidationConsumer =
```
rabbah: no change on this line, can the RHS move back to same line?

rabbah:
LGTM in general; although I don't understand why the poll interrupt is needed and want to investigate that on my own.

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -21,46 +21,91 @@ import java.util.Properties
...
    * @param duration the maximum duration for the long poll
    */
-  override def peek(duration: Duration = 500.milliseconds) = {
-    val records = consumer.poll(duration.toMillis)
-    records map { r =>
-      (r.topic, r.partition, r.offset, r.value)
-    }
+  override def peek(duration: Duration = 500.milliseconds,
+                    retry: Int = 3): Iterable[(String, Int, Long, Array[Byte])] = {
+    // Since kafka client can be infinitely blocked in poll(timeout), we should interrupt it using the wakeup() method
```
markusthoemmes: Can you elaborate on why the poll could be blocked infinitely? In theory, the timeout should limit the time it can be blocked? Is this for edge cases (i.e. lost connection)?

markusthoemmes:
I clarified the remaining bits with Kafka experts. Consider this scheduled for merge after #3258 to minimize conflicts.

@devbv I'll take care of the rebase if you want.

devbv:
@markusthoemmes 
Sorry for late reply. Recently I'm focusing on the other project in company, so I don't have much time to take care of it. Please rebase and merge. :)

markusthoemmes:
PG3 1895 🔵

PR #3073, opened 2017-12-07
Add namespaceId to the activation record as well to enable correlation.

markusthoemmes:
Correlation should happen based on a unique identifier.


markusthoemmes:
PG3 1456 🔵

PR #3077, opened 2017-12-08
Enable ssl on the path between edge and controllers

vvraskin:
This PR will enable SSL between nginx and controllers. A couple of things that must be considered: 
* certs for local and docker-machine environments are self-signed
* certs are regenerated during OW deployment
* the default protocol will be https

@mhenke1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,9 +60,13 @@ ansible/db_local.ini*
 ansible/tmp/*
 ansible/roles/nginx/files/openwhisk-client*
 ansible/roles/nginx/files/*.csr
+ansible/roles/nginx/files/*.p12
 ansible/roles/nginx/files/*cert.pem
 ansible/roles/nginx/files/*p12
 ansible/roles/kafka/files/*
+ansible/roles/controller/files/*.pem
+ansible/roles/controller/files/*.key
+ansible/roles/controller/files/*.p12
 
 # .zip files must be explicited whitelisted
 *.zip
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -66,6 +66,25 @@ controller:
   loadbalancer:
     spi: "{{ controller_loadbalancer_spi | default('') }}"
   loglevel: "{{ controller_loglevel | default(whisk_loglevel) | default('INFO') }}"
+  protocol: "{{ controllerProtocolForSetup }}"
+  ssl:
+    cn: openwhisk-controllers
+    cert: "{{ controller_ca_cert | default('controller-openwhisk-server-cert.pem') }}"
+    key: "{{ controller_key | default('controller-openwhisk-server-key.pem') }}"
+    clientAuth: "{{ controller_client_auth | default('true') }}"
+    storeFlavor: PKCS12
+    keystore:
+      password: "{{ controllerKeystorePassword }}"
+      path: "/conf/{{ controllerKeystoreName }}"
+# keystore and truststore are the same as long as controller and nginx share the certificate
+    truststore:
+      password: "{{ controllerKeystorePassword }}"
+      path: "/conf/{{ controllerKeystoreName }}"
+# move controller protocol outside to not evaluate controller variables during execution of setup.yml
+controllerProtocolForSetup: "{{ controller_protocol | default('https') }}"
+controllerKeystoreName: "{{ controllerKeyPrefix }}openwhisk-keystore.p12"
+controllerKeyPrefix: "controller-"
+controllerKeystorePassword: openwhisk
 
 jmx:
   basePortController: 15000
--- a/common/scala/src/main/scala/whisk/common/Https.scala
+++ b/common/scala/src/main/scala/whisk/common/Https.scala
@@ -1,0 +1,75 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.common
+
+import java.io.{FileInputStream, InputStream}
+import java.security.{KeyStore, SecureRandom}
+import javax.net.ssl.{KeyManagerFactory, SSLContext, TrustManagerFactory}
+
+import akka.http.scaladsl.ConnectionContext
+import akka.stream.TLSClientAuth
+import com.typesafe.sslconfig.akka.AkkaSSLConfig
+import whisk.core.WhiskConfig
+import pureconfig._
+
+object Https {
+  case class HttpsConfig(keystorePassword: String,
+                         keystoreFlavor: String,
+                         keystorePath: String,
+                         truststorePath: String,
+                         truststorePassword: String,
+                         truststoreFlavor: String,
+                         clientAuth: String)
+  private val httpsConfig = loadConfigOrThrow[HttpsConfig]("whisk.controller.https")
+
+  def getCertStore(password: Array[Char], flavor: String, path: String): KeyStore = {
+    val certStorePassword: Array[Char] = password
+    val cs: KeyStore = KeyStore.getInstance(flavor)
+    val certStore: InputStream = new FileInputStream(path)
+    cs.load(certStore, certStorePassword)
+    cs
+  }
+
+  def connectionContext(config: WhiskConfig, sslConfig: Option[AkkaSSLConfig] = None) = {
+
+    val keyFactoryType = "SunX509"
+    val clientAuth = {
+      if (httpsConfig.clientAuth.toBoolean)
+        Some(TLSClientAuth.need)
+      else
+        Some(TLSClientAuth.none)
+    }
+
+// configure keystore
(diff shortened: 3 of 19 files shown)


ddragosd:
@vvraskin 👍 to the idea of adding SSL between edge -> controller. 
Did you get the chance to run some performance tests to compare HTTP with HTTPS (throughput and CPU usage) ?

style95:
Actually HTTPS has huge effects on performance.
Is there any reason to take such overhead?

common/scala/src/main/scala/whisk/common/Https.scala
```diff
@@ -0,0 +1,47 @@
...
+import java.security.{KeyStore, SecureRandom}
+import javax.net.ssl.{KeyManagerFactory, SSLContext, TrustManagerFactory}
+
+import akka.http.scaladsl.ConnectionContext
+import com.typesafe.sslconfig.akka.AkkaSSLConfig
+import whisk.core.WhiskConfig
+
+object Https {
+
+  def connectionContext(config: WhiskConfig, sslConfig: Option[AkkaSSLConfig] = None) = {
```
markusthoemmes: Can we move the configuration into PureConfig to avoid passing stuff around?

tests/src/test/scala/services/HeadersTests.scala
```diff
@@ -67,15 +66,19 @@ class HeadersTests extends FlatSpec with Matchers with ScalaFutures with WskActo
...
+  val url = Uri(s"https://${WhiskProperties.getBaseControllerAddress()}")
 
   def request(method: HttpMethod, uri: Uri, headers: Option[Seq[HttpHeader]] = None): Future[HttpResponse] = {
     val httpRequest = headers match {
       case Some(headers) => HttpRequest(method, uri, headers)
       case None          => HttpRequest(method, uri)
     }
+    val sslConfig = AkkaSSLConfig().mapSettings { s =>
+      s.withHostnameVerifierClass(classOf[AcceptAllHostNameVerifier].asInstanceOf[Class[HostnameVerifier]])
+    }
```
markusthoemmes: Can we do this in a generalized way since our Rest testcases need the same "functionality". Maybe we can have a `testReques(req: HttpRequest)` which does the same as 

```scala
Http().singleRequest(httpRequest, connectionContext = connectionContext)
```

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -343,4 +348,18 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
   private def completeAsTriggerResponse(trigger: WhiskTrigger): RequestContext => Future[RouteResult] = {
     complete(OK, trigger.withoutRules)
   }
+
+  /**
+   * Sends a request either over http or https depending on the configuration
+   * @param request
+   * @param config
+   * @return
```
markusthoemmes: Missing scaladoc.

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -80,7 +80,7 @@ protected[controller] class SwaggerDocs(apipath: Uri.Path, doc: String)(implicit
 
 protected[controller] object RestApiCommons {
   def requiredProperties =
-    Map(WhiskConfig.servicePort -> 8080.toString) ++
+    Map(WhiskConfig.servicePort -> 8080.toString, WhiskConfig.controllerProtocol -> "https") ++
```
markusthoemmes: Do we really want to set https to be the default here? It's harder to develop on, isn't it?

ansible/group_vars/all
```diff
@@ -124,6 +127,13 @@ controller:
       seedNodes: "{{ groups['controllers'] | map('extract', hostvars, 'ansible_host') | list }}"
   # We recommend to enable HA for the controllers only, if bookkeeping data are shared too. (localBookkeeping: false)
   ha: "{{ controller_enable_ha | default(True) and groups['controllers'] | length > 1 }}"
+  protocol: "{{ controller_protocol | default('https') }}"
```
csantanapr: Can we make the default http instead of https? Helps for local development environment 

You can pass override to ansible-playbook with -e flag in Travis if you want to enable in Travis testing deploy script

rabbah:
Do we need another genssl bash script (can we extend/repurpose https://github.com/apache/incubator-openwhisk/blob/master/ansible/roles/nginx/files/genssl.sh).

ansible/group_vars/all
rabbah: Agree. Flipping this default breaks our documented workaround for wsk -i which instructs a user to set the api host against the controller directly.

vvraskin:
It took some time to get back to the PR, sorry guys :)
@ddragosd I haven't noticed any impact on CPU, but with respect to throughput and latency its payload specific. With smaller payload (<128KB) the overhead is negligible, the backoff starts after that point and fluctuates around ~10-15%. 

@style95 I partially agree with you, but the overhead is not really huge. I think there could be some people who could decide to go with encryption due to compliance and or some other security related motives.

ansible/environments/local/group_vars/all
```diff
@@ -32,3 +32,4 @@ kafka_topics_invoker_retentionBytes: 104857600
 kafka_topics_invoker_retentionMS: 300000
 
 env_hosts_dir: "{{ playbook_dir }}/environments/local"
+controller_protocol: https
```
cbickel: Is there a reson for not setting the `distributed` environment to https as well?
If there is no reason, can we make the default to https and leave these variables in the environements' files?

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -98,6 +98,8 @@ class WhiskConfig(requiredProperties: Map[String, String],
   val controllerSeedNodes = this(WhiskConfig.controllerSeedNodes)
   val controllerLocalBookkeeping = getAsBoolean(WhiskConfig.controllerLocalBookkeeping, false)
   val controllerHighAvailability = getAsBoolean(WhiskConfig.controllerHighAvailability, false)
+  val controllerProtocol = this(WhiskConfig.controllerProtocol)
+  val controllerSSLEnabled = if (this.controllerProtocol == "https") true else false
```
cbickel: Can we move this to pureconfig as well?

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -161,7 +161,7 @@ object Controller {
   // a value, and whose values are default values.   A null value in the Map means there is
   // no default value specified, so it must appear in the properties file
   def requiredProperties =
-    Map(WhiskConfig.controllerInstances -> null) ++
+    Map(WhiskConfig.controllerInstances -> null, WhiskConfig.controllerProtocol -> null) ++
```
cbickel: Pure config as well.

But if you really need to keep it in WhiskConfig, it is already in `RestApiCommons.requiredProperties `. So you don't need to add it here.

tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -85,15 +81,42 @@ import common.WskActorSystem
...
   override def verify(s: String, sslSession: SSLSession): Boolean = true
 }
 
 object SSL {
+
+  val controllerProtocol = WhiskProperties.getProperty("controller.protocol")
+  val controllerKeyStorePath = WhiskProperties.getProperty("controller.ssl.keystore.path")
+  val controllerKeyStorePassword = WhiskProperties.getProperty("controller.ssl.keystore.password")
+  val controllerKeyStoreFlavor = WhiskProperties.getProperty("controller.ssl.keystore.flavor")
+  val controllerClientAuth = WhiskProperties.getProperty("controller.ssl.clientAuth").toBoolean
```
cbickel: You already put these variables into `application.conf`. Can we use them instead?

Afterwards you can remove the entries from `whisk.properties` again.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
vvraskin: I wanted to keep it in the Config, so we don't create two identical objects in Trigger and Controller or pass that property around.

tests/src/test/scala/common/rest/WskRest.scala
markusthoemmes: Please refer to https://github.com/apache/incubator-openwhisk/pull/3111/commits/04579a3b786c582d0ff35f679208e63069473b68 for our future path of solving this. I back that we should move these to pureconfig and write the `.conf` the same way we write whisk.properties today.

.gitignore
```diff
@@ -2,6 +2,7 @@
 nginx.conf
 whisk.properties
 default.props
+/tests/src/test/resources/application.conf
```
cbickel: Is there a reason for not tracking changes of this file?

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
cbickel: As we need to go the internal way for triggers (long term), I think it's more valuable to use pureconfig here instead of the old whiskConfig.

ansible/group_vars/all
```diff
@@ -65,6 +65,25 @@ controller:
...
+    storeFlavor: PKCS12
+    keystore:
+      password: "{{ controllerKeystorePassword }}"
+      path: "/conf/{{ controllerKeystoreName }}"
+# keystore and truststore are the same as long as controller and nginx share the certificate
+    truststore:
+      password: "{{ controllerKeystorePassword }}"
+      path: "/conf/{{ controllerKeystoreName }}"
+# move controller protocol outside to not evaluate controller variables during execution of setup.yml
+controllerProtocolForSetup: "{{ controller_protocol | default('https') }}"
```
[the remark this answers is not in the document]
vvraskin: The problem is that when we execute `setup.yml` ansible doesn't know about the `groups[controllers]` and it is complaining about the undefined vabiable. So I moved it to `controllerProtocolForSetup` so we don't evaluate all controller properties.

PR #3080, opened 2017-12-08
Remove unneeded code reference from WhiskActionMetaData test helper

dubeejw:
A code parameter is not needed for `js6MetaData()`.


dubeejw:
PG3 1481 🔵

PR #3084, opened 2017-12-09
Allow invoke helper to work with tgz files.

rabbah:
A zip or tgz (more generally archive format) is specific to a runtime. Allow this script to work with zip or tgz files.

See https://github.com/apache/incubator-openwhisk-runtime-nodejs/pull/16 for example.


csantanapr:
We should add a test "should invoke python package with tgz"

rabbah:
That will require changes in the runtimes and should be tested in those repos. This just makes the helper script work with zip or tgz.

csantanapr:
Ah missed that it was the helper/dev script

PR #3089, opened 2017-12-11
Fix "No such file or directory" problem when cleaning "veth*.ifstate"…

sven-lange-last:
… files

See #3081 for a discussion.

Let Ansible ignore errors from the `find` command because there is no option to let the `find` commmand ignore errors from the `-delete` action.

Root cause of the problem must be a race condition where files that have been identified during the `Find` phase have been removed before these files are processed in the `Action` phase.


sven-lange-last:
PG 1 / 2424 completed with one failed test that is in no way related to the change introduced by this PR.

sven-lange-last:
@csantanapr please attend to this PR.

PR #3090, opened 2017-12-11
Remove obsolete code to remove old config directory of nginx.

cbickel:
Some weeks ago, I moved all config files to `wskconf`. To delete config files in the old directory, I added the code, that I remove again in this PR.


cbickel:
PG3#1496 🔵

PR #3092, opened 2017-12-11
only consider healthy invokers when scheduling

### LINKED ISSUE #3091: if invoker id 0 is offline, invoker id 1 is ignored even if it is healthy
## Environment details:

* local deployment (all deployments I've tested exhibit the 

1.  deploy a single invoker, with id=1  
2.  restart controller+invoker
3.  invoke any action 

expected successful use of invoker1

```
[2017-12-11T17:41:24.665Z] [ERROR] [#sid_121] [LoadBalancerService] all invokers down
[2017-12-11T17:41:24.668Z] [ERROR] [#tid_14] [ActionsApi] [POST] action activation failed: no invokers available
```

probably only happens with single invoker, related to blackbox vs managed reservation

tysonnorris:
fixes #3091 

don't consider unhealthy/offline invokers when scheduling

### DIFF
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
@@ -324,7 +324,9 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
 
     loadBalancerData.activationCountPerInvoker.flatMap { currentActivations =>
       allInvokers.flatMap { invokers =>
-        val invokersToUse = if (action.exec.pull) blackboxInvokers(invokers) else managedInvokers(invokers)
+        val healthyInvokers = invokers.filter(_._2 == Healthy)
+        val invokersToUse =
+          if (action.exec.pull) blackboxInvokers(healthyInvokers) else managedInvokers(healthyInvokers)
         val invokersWithUsage = invokersToUse.view.map {
           // Using a view defers the comparably expensive lookup to actual access of the element
           case (instance, state) => (instance, state, currentActivations.getOrElse(instance.toString, 0))


core/controller/src/main/scala/whisk/core/loadBalancer/LoadBalancerService.scala
```diff
@@ -324,7 +324,9 @@ class LoadBalancerService(config: WhiskConfig, instance: InstanceId, entityStore
 
     loadBalancerData.activationCountPerInvoker.flatMap { currentActivations =>
       allInvokers.flatMap { invokers =>
-        val invokersToUse = if (action.exec.pull) blackboxInvokers(invokers) else managedInvokers(invokers)
+        val healthyInvokers = invokers.filter(_._2 == Healthy)
+        val invokersToUse =
+          if (action.exec.pull) blackboxInvokers(healthyInvokers) else managedInvokers(healthyInvokers)
```
markusthoemmes: Keeping Unhealthy/Down invokers in that list is deliberate:

The loadbalancer heavily relies on the ordering of invokers. If You shrink/grow this list continually (for example by an invoker going online/offline all the time) you completely destroy its heuristics, resulting in super bad scheduling behavior if you have a system with some invokers in a flipping state.

What's the motivation to do this change?

tysonnorris: currently if you deploy a single invoker, and it is not id 0, it is ignored; so while "ordering of invokers" may be important, it will require some way to make sure there is also a way to ignore invokers that should not be referenced in the scheduling. How should that be done, if not by referencing the healthy state?

tysonnorris:
per #3091 working as designed...

Issue #2152, opened 2017-04-18
OpenWhisk Docker Image for Edge component.


ghost:
I have seen that there were a bunch of OpenWhisk Docker images for all of the components which is pretty cool. It would be nice to also have an image for Nginx that has all of the extra pieces already baked into it like the CLI and SDKs. This we we would not have to build them and upload them as part of the deployment process.

rabbah:
@csantanapr is this done?

csantanapr:
not sure how to go about this one.
Do we create the edge docker image instead of using a mount volume we add the files to the image, what about the configuration do we pass this info as environment variables to replace the template on start of the container?

DanLavine:
@csantanapr the "Edge" image will need to mount in the `certs` and `nginx.conf` file since those are environment specific. the "Edge" Docker Image should just need to comprise of the public Nginx image plus any of the CLI/SDKs prebuilt into it.

This means that the current Ansible scripts should more or lest just work as they are. Minus the need to already have the cli prebuilt and build the BlackBox tarball. 

Then When deploying Kubernetes, you can take a look at this [README](https://github.com/apache/incubator-openwhisk-deploy-kube/blob/master/kubernetes/nginx/README.md) for how configuring those same values would go.

dgrove-oss:
The only barrier to using a vanilla docker image for nginx now is this stanza in nginx.conf:
```
        location /cli/go/download {
            autoindex on;
            root /etc/nginx;
        }

```

I can see four ways to deal with this:
  1. Redirect the generic /cli/go/download to the incubator-openwhisk-cli release page. (Ok if this route is intended for interactive human consumption.  Not so nice if we're scripting against it).
  2. Have stanzas for the CLI for each platform that we support that redirects the download to the right release tar ball (not nice if people are used to exploring the available downloads).
  3. Do both 1 and 2.
  4.  Decide that we really want to do auto-indexing and so commit to always needing to copy in the CLI binaries to /etc/nginx/cli/download/os/arch/cli.zip.

If we need to take path 4, I think I can do it with an InitContainer in Kubernetetes so we could use a vanilla nginx image + a custom image to do the download of the CLI to the right place in the running containers file system.

@rabbah @csantanapr   opinions on what this should look like when we are done?

csantanapr:
I like option 3
I have being thinking on updating upstream main to have this option on no longer copying the CLI and serving from nginx

We already cleaned up this way the swift sdk and the docker sdk

[...]

PR #3093, opened 2017-12-11
Make OpenWhisk CLI available only via redirect

dgrove-oss:
As was already done with the iOS starter app and the blackbox SDK,
switch to making the cli binaries available via a redirect to the
openwhisk-cli release page instead of copying cli binaries into
the edge container.  As discussed in issue #2152 this eliminates
the need for a custom OpenWhisk edge image for downstream deployments.

Also update GoCLINginxTests to check for redirects instead of direct
downloads and remove the cli role and related variable definitions
from ansible.


dgrove-oss:
Closes #2152

dgrove-oss:
Overly aggressive; I need to bring back support for downloading the cli to $OPENWHISK_HOME/bin.

dgrove-oss:
needs labels: review, companion.

PG1 2449 🏃

csantanapr:
@dgrove-oss This is not quite what we agreed.
I don't think we should remove all options, In my comment https://github.com/apache/incubator-openwhisk/issues/2152#issuecomment-350508205

I said that is OK to add a 3rd option `remote_redirect`, leaving the other 2 options in place.
We can update doc to mention that `local` and `remote` are deprecated, and eventually going to be removed.

Also that for kube deployment only the `remote_redirect` is the way the nginx is configured.

dgrove-oss:
will redo; easier to start over on new branch. closing.

PR #3094, opened 2017-12-12
Bound docker/runc commands in their allowed runtime.

markusthoemmes:
Docker can cause hanging commands which never finish. Essentially those break the invoker and it needs to be restarted to recover. This adds a timeout to each of those commands to detect this problem.


core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
```diff
@@ -49,6 +53,17 @@ object DockerContainerId {
   }
 }
 
+/**
+ * Configuration for docker client command timeouts.
+ */
+case class DockerClientTimeoutConfig(run: Duration,
```
rabbah: This is nice.

markusthoemmes:
PG2 2597 🔵

PR #3095, opened 2017-12-12
Decouple health checking from active ack messages.

markusthoemmes:
Active ack messages are inspected to check whether an invoker has a lot of system-level errors. Timeout of an active ack is treated as such an error. The active ack timeout time is coupled to the "promise" that there is no waiting queue on an invoker at all (diluted because of clustering). This promise is non-existent as of today and thus active ack timeout should not be seen as system-level error from the invoker.


rabbah:
Remove the comment couple of lines down? “        // For both cases, it looks like the invoker works again and we should send the status of”

PR #3100, opened 2017-12-13
add remote_redirect mode for openwhisk_cli installation

dgrove-oss:
Add a third remote_redirect option to the prior local and
remote options for cli installation.  In remote_redirect mode,
the CLI binaries are not installed into the edge server's
file system. All requests to download the CLI are instead
redirected to an external URL (for example to download
from the openwhisk-cli release page).  This follows the
pattern already set for the iOS starter app and blackbox SDK.

See discussion in issue #2152.

### DIFF
--- a/ansible/roles/cli/tasks/deploy.yml
+++ b/ansible/roles/cli/tasks/deploy.yml
@@ -1,14 +1,15 @@
 ---
 # Tasks for handling CLI customization and publishing
 
+- set_fact:
+    cli_installation_mode="{{ openwhisk_cli.installation_mode }}"
+
 - name: "ensure nginx directory for cli exists"
   file:
     path: "{{ cli.nginxdir  }}"
     state: directory
   become: "{{ cli.dir.become }}"
-
-- set_fact:
-    cli_installation_mode="{{ openwhisk_cli.installation_mode }}"
+  when: cli_installation_mode != "remote_redirect"
 
 - include: docker_login.yml
 
@@ -26,3 +27,7 @@
   when: cli_installation_mode == "local"
 
 - include: download_cli.yml
+  when: cli_installation_mode != "remote_redirect"
+
+- include: download_remote_redirect_cli.yml
+  when: cli_installation_mode == "remote_redirect"
--- a/ansible/roles/cli/tasks/download_remote_redirect_cli.yml
+++ b/ansible/roles/cli/tasks/download_remote_redirect_cli.yml
@@ -1,0 +1,38 @@
+# Download the cli package from the remote_redirect location to openwhisk_home/bin and unarchive it.
+
+- name: "set the file name to download for linux"
+  set_fact:
+    cli_download_name={{ openwhisk_cli.remote_redirect.package_name }}-{{ openwhisk_cli_tag }}-linux-amd64.tgz
+  when: ('environments/docker-machine' not in inventory_dir) and (ansible_os_family != "Darwin")
+
+- name: "set the file name to download for mac"
+  set_fact:
+    cli_download_name={{ openwhisk_cli.remote_redirect.package_name }}-{{ openwhisk_cli_tag }}-mac-amd64.zip
+  when: ('environments/docker-machine' in inventory_dir ) or (ansible_os_family == "Darwin")
+
+
+- name: "set the http header for downloading the binary packages"
+  set_fact:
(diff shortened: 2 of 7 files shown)


dgrove-oss:
Closes #2152 

This is a rework of PR #3093 based on feedback from @csantanapr

dgrove-oss:
PG1 2454 - queued

dgrove-oss:
needs labels: review, companion.

dgrove-oss:
@csantanapr @houshengbo -- reminder to review

ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -90,9 +90,25 @@ http {
             return 301 https://github.com/apache/incubator-openwhisk-client-swift/releases/download/0.2.3/starterapp-0.2.3.zip;
         }
 
+{% if openwhisk_cli.installation_mode == "remote_redirect" %}
+        # redirect requests for specific binaries to the matching one from the {{ openwhisk_cli_tag }} openwhisk-cli release.
+  {% for os in cli_os %}
+     {% for arch in cli_arch %}
```
jonpspri: This concerns me given changes I made in #3133 to introduce cli_os_arch.  It will be a simple update, depending on which goes first.

ansible/roles/cli/tasks/deploy.yml
```diff
@@ -26,3 +27,7 @@
   when: cli_installation_mode == "local"
 
 - include: download_cli.yml
+  when: cli_installation_mode != "remote_redirect"
```
jonpspri: My gut says this should be `cli_installation_mode == "remote"` because selecting local install is usually because the download is broken for a particular platform.  Of course, I'm not sure local is functionally complete in the configuration.  But, more to the point, we should separate making the CLI available for download via NGinx from sourcing the binary for /wsk/bin for local use.

ansible/roles/cli/tasks/download_remote_redirect_cli.yml
```diff
@@ -0,0 +1,38 @@
+# Download the cli package from the remote_redirect location to openwhisk_home/bin and unarchive it.
+
+- name: "set the file name to download for linux"
+  set_fact:
+    cli_download_name={{ openwhisk_cli.remote_redirect.package_name }}-{{ openwhisk_cli_tag }}-linux-amd64.tgz
```
jonpspri: This is a bit more complicated now, because there will be multiple linux architectures (amd64, s390x).  Is there a reason we can't reuse the existing ansible files for downloading the cli?  It seems configuring Nginx should be a separate problem from downloading a binary for "local" use.

dgrove-oss:
Thinking about this change more over break, it seems to me that it would be cleaner to not conflate local building vs. downloading of the CLU and embedding the CLI in the edge host vs. putting a redirect in the nginx.conf file.   So instead of having 3 modes for cli_installation_mode we would have two boolean properties: cli_build_cli_locally (default to false) and cli_download_redirect (default to true).    Thoughts?

jonpspri:
I was heading in the same direction.  My intention was to have build occur in the 'incubator-openwhisk-cli' project.  The build would actually create two artifacts:

1. The `wsk` binaries (as it does today).

1. A tarball for deployment in Nginx consisting of the binaries directory structure _and_ the `content.json` file.

Riffing on that, the CLI "download" could pull the tarball from a URL, and for local build the configuration could override that URL with `file://../../incubator-openwhisk-cli/build/<path-to-tarball>.tgz`.  So we only need to address the two Nginx configurations (download vs redirect) -- anyone wanting local build should be savvy enough to do the build themselves before overriding the URL.  Local CLI sourcing is similar -- is a URL override provided?  Get that.  Otherwise, get the 'default' CLI for the local OS and architecture.

dgrove-oss:
What do we want to do with this PR?  Should I close it in favor of the work @jonpspri is doing, or is there some piece of it that is worth pulling as a separate PR that complements his cli changes?

markusthoemmes:
@csantanapr ^^ WDYT wrt @dgrove-oss 's comment? How do we want to proceed?

csantanapr:
I like @jonpspri proposal
>So we only need to address the two Nginx configurations (download vs redirect)
https://github.com/apache/incubator-openwhisk/pull/3100#issuecomment-354289069

Closing this PR, if @jonpspri needs something from this PR he is free to take it here.

PR #3101, opened 2017-12-13
Disable reduce limits on all nodes

style95:
This pr is to disable `reduce_limit` on all nodes.
Currently, it is only disabled in coordinator node.

I thought this configuration is cluster-wide, but it was not.
So it should be disabled on all nodes respectively.


style95:
@cbickel Kindly review this.

PR #3102, opened 2017-12-13
Use alpakka's FileRotator instead of the copied one.

markusthoemmes:
A new alpakka release is out so we can remove the copied version of this file to make this clean again 🎉 .


markusthoemmes:
PG3 1614 🔵

PR #3104, opened 2017-12-13
Add TID stride configuration

dubeejw:
Allows the number of TIDs to be skipped in a controller to be configured.

### DIFF
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -43,7 +43,7 @@
       "JAVA_OPTS": "-Xmx{{ controller.heap }} -XX:+CrashOnOutOfMemoryError -XX:+UseGCOverheadLimit -XX:ErrorFile=/logs/java_error.log -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/logs"
       "CONTROLLER_OPTS": "{{ controller.arguments }}"
       "CONTROLLER_INSTANCES": "{{ controller.instances }}"
-      "CONFIG_whisk_controller_tidStrides": "{{ controller.tidStrides }}"
+      "CONFIG_whisk_controller_tidStrides": "{{ controller.tidStrides | default( controller.instances ) }}"
 
       "COMPONENT_NAME": "controller{{ groups['controllers'].index(inventory_hostname) }}"
       "PORT": 8080
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -43,7 +43,6 @@
       "JAVA_OPTS": "-Xmx{{ controller.heap }} -XX:+CrashOnOutOfMemoryError -XX:+UseGCOverheadLimit -XX:ErrorFile=/logs/java_error.log -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/logs"
       "CONTROLLER_OPTS": "{{ controller.arguments }}"
       "CONTROLLER_INSTANCES": "{{ controller.instances }}"
-      "CONFIG_whisk_controller_tidStrides": "{{ controller.tidStrides | default() }}"
 
       "COMPONENT_NAME": "controller{{ groups['controllers'].index(inventory_hostname) }}"
       "PORT": 8080
@@ -105,6 +104,8 @@
       "CONFIG_whisk_spi_LogStoreProvider": "{{ userLogs.spi }}"
       
       "CONFIG_logback_log_level": "{{ controller.loglevel }}"
+
+      "CONFIG_whisk_transactions_strides": "{{ transactions.strides | default() }}"
     volumes:
       - "{{ whisk_logs_dir }}/controller{{ groups['controllers'].index(inventory_hostname) }}:/logs"
     ports:
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -43,7 +43,7 @@
       "JAVA_OPTS": "-Xmx{{ controller.heap }} -XX:+CrashOnOutOfMemoryError -XX:+UseGCOverheadLimit -XX:ErrorFile=/logs/java_error.log -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/logs"
       "CONTROLLER_OPTS": "{{ controller.arguments }}"
       "CONTROLLER_INSTANCES": "{{ controller.instances }}"
-      "CONFIG_whisk_controller_tidStrides": "{{ controller.tidStrides | default( controller.instances ) }}"
+      "CONFIG_whisk_controller_tidStrides": "{{ controller.tidStrides | default() }}"
 
       "COMPONENT_NAME": "controller{{ groups['controllers'].index(inventory_hostname) }}"
       "PORT": 8080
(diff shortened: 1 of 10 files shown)


dubeejw:
PG2 2526 👍

rabbah:
We could have the tid also assigned by the edge router and passed to the controller as a header - perhaps as a subsequent change.

dubeejw:
@dgrove-oss, pushed a commit that defaults the controllerTidStrides value to controllerInstances if it is not present in the config.

dgrove-oss:
LGTM; thanks

markusthoemmes:
What do you think about moving both controllerInstances and the stride config to `application.conf` and load it via PureConfig? You could then have: `whisk.controller.instanceCount = 1` and `whisk.controller.tidStrides = ${whisk.controller.instanceCount}`?

dubeejw:
PG3 1542 🔵

dubeejw:
@markusthoemmes, I sent you a message bout it. Wouldn't that require updates from the dependent components that @dgrove-oss mentioned?

dubeejw:
@rabbah, I like the idea about using nginx to generate the TID. We would have to do probably do a fair amount of rewriting though to make it work with the current implementation. How would the invoker TID generation happen with this change?

dgrove-oss:
If the default is in application.conf (instead of in ansible), then the downstream projects don't need to be changed.  They'll get the new defaults automatically.  That's why we're going to migrate as much as possible to using pureconfig.

rabbah:
Invokers don't generate tids, they receive them from the activation messages. They should only have system ids.

dubeejw:
Pushed a commit that will use PureConfig. Will need someone to take a look.

common/scala/src/main/resources/application.conf
```diff
@@ -82,4 +82,7 @@ whisk {
         activations-ddoc = "whisks.v2"
         activations-filter-ddoc = "whisks-filters.v2"
     }
+
+    controller {
```
dubeejw: Had to add this so that `loadConfigOrThrow[ControllerConfig](ConfigKeys.controller)` does not fail.

dubeejw:
PG2 2615 🚀

dubeejw:
PG2 2615 🔵

common/scala/src/main/resources/application.conf
rabbah: I think this i s because you're using ansible still for the values.

dubeejw: tidStrides is an optional value in the controller section here. However, the controller section has to be defined to prevent a runtime here, @rabbah.

dgrove-oss: I think what you want here is:
```
    controller {
        tidStrides = ${?CONTROLLER_INSTANCES}
    }
```
that will default tidStrides to the value of the CONTROLLER_INSTANCES envar if CONFIG_whisk_controller_tidStrides isn't set.

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -43,6 +43,7 @@
       "JAVA_OPTS": "-Xmx{{ controller.heap }} -XX:+CrashOnOutOfMemoryError -XX:+UseGCOverheadLimit -XX:ErrorFile=/logs/java_error.log -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/logs"
       "CONTROLLER_OPTS": "{{ controller.arguments }}"
       "CONTROLLER_INSTANCES": "{{ controller.instances }}"
+      "CONFIG_whisk_controller_tidStrides": "{{ controller.tidStrides | default( controller.instances ) }}"
```
dgrove-oss: Should be:
```
"CONFIG_whisk_controller_tidStrides": "{{ controller.tidStrides | default( ) }}"
```

dubeejw:
Thanks, @dgrove-oss!

dubeejw:
PG2 2616 🚀

dubeejw:
PG2 2616 🔵
PG2 1633 🔵

common/scala/src/main/resources/application.conf
markusthoemmes: It feels a bit odd to add "controller" to the common configuration, does it? Also you don't want to use camelCased but rather kebab-cased fieldnames, otherwise the override through the env-convert script won't work.

Should we rather put this under a "transactions" key? `whisk.transactions.strides`

markusthoemmes: To add to dave's comment, if you want to get rid of the `Option` in the code itself you can do this:

```
whisk {
  transactions {
    strides = 1
    strides = ${?CONTROLLER_INSTANCES}
  }
}
```

rabbah: > whisk.transactions.strides
+1 good suggestion.

rabbah: the `option` makes sense with the code below where it default to number of instances; fits as a nice idiom.

dubeejw:
PG2 2623 🕐

dubeejw:
PG2 2623 🔵

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -104,6 +104,8 @@
       "CONFIG_whisk_spi_LogStoreProvider": "{{ userLogs.spi }}"
       
       "CONFIG_logback_log_level": "{{ controller.loglevel }}"
+
+      "CONFIG_whisk_transactions_strides": "{{ transactions.strides | default() }}"
```
rabbah: Can we eliminate the ansible setting entirely and use the controller index here

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -241,12 +245,15 @@ object TransactionId {
...
 trait TransactionCounter {
-  val numberOfInstances: Int
+  case class TransactionCounterConfig(strides: Int)
+
+  val transCounterConfig = loadConfigOrThrow[TransactionCounterConfig](ConfigKeys.transactions)
+  val strides = transCounterConfig.strides
   val instanceOrdinal: Int
 
-  private lazy val cnt = new AtomicInteger(numberOfInstances + instanceOrdinal)
+  private lazy val cnt = new AtomicInteger(strides + instanceOrdinal)
```
rabbah: Seems that having the instance ordinal and stride is duplicitous. Can we reduce to just strides

cbickel:
@rabbah Regarding your last comment: `If the point of the original counter change to use instances was to make non overlapping ids that worked because it separated even and odd numbers. But with more instances the numbers will overlap anyway.`

Not only odd and even numbers are used. The total number of controllers is also part of the tid generation. E.g for 3 controllers:
controller0 will have the following tids: 3, 6, 9, ...
controller1: 4, 7, 10, ...
controller2: 5, 8, 11, ...

```
  // Initial value: amountOfControllers + instanceOfThisController
  private lazy val cnt = new AtomicInteger(numberOfInstances + instanceOrdinal)

  def transid(): TransactionId = {
    // add amountOfControllers to the last tid
    TransactionId(cnt.addAndGet(numberOfInstances))
  }
```

rabbah:
Right! Thanks for that. I miss that initial seed of the atomic counter every time it!

common/scala/src/main/resources/application.conf
```diff
@@ -82,4 +82,9 @@ whisk {
         activations-ddoc = "whisks.v2"
         activations-filter-ddoc = "whisks-filters.v2"
     }
+    # transaction ID related configuration
+    transactions {
+        stride = 1
+        stride = ${?CONTROLLER_INSTANCES}
```
rabbah: since this is in the common conf, it feels wrong to have this tied to controller instances.
` +        stride = ${?CONTROLLER_INSTANCES} `?

PR #3106, opened 2017-12-14
Couchdb persistency

style95:
This pr is to introduce persistency on CouchDB.

Notable changes:
* Default configuration files are added.
* `db_datadir` is added to use volume mapping for CouchDB data.
* Cluster setup predicates are restored to work on single node setup.

I copied default configuration files from `apache/couchdb:2.1`

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -195,6 +195,9 @@ nginx:
 # The key db.whisk.auth is the name of the authentication database where all keys of all users are stored.
 # The db_prefix is defined for each environment on its own. The CouchDb credentials are also defined for each environment on its own.
 db:
+  confdir: "{{ config_root_dir }}/couchdb"
+  dir:
+    become: "{{ db_dir_become | default(false) }}"
   instances: "{{ groups['db'] | length }}"
   authkeys:
   - guest
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -195,7 +195,6 @@ nginx:
 # The key db.whisk.auth is the name of the authentication database where all keys of all users are stored.
 # The db_prefix is defined for each environment on its own. The CouchDb credentials are also defined for each environment on its own.
 db:
-  confdir: "{{ config_root_dir }}/couchdb"
   dir:
     become: "{{ db_dir_become | default(false) }}"
   instances: "{{ groups['db'] | length }}"
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -196,7 +196,7 @@ nginx:
 # The db_prefix is defined for each environment on its own. The CouchDb credentials are also defined for each environment on its own.
 db:
   dir:
-    become: "{{ db_dir_become | default(false) }}"
+    become: "{{ db_dir_become | default(true) }}"
   instances: "{{ groups['db'] | length }}"
   authkeys:
   - guest
--- a/ansible/roles/couchdb/files/default.d/README
+++ b/ansible/roles/couchdb/files/default.d/README
@@ -1,0 +1,11 @@
+CouchDB default configuration files
+
+Files found under the etc/default.d directory that end with .ini are
+parsed within couchdb(1) at startup.
+
+This directory is intended for distribution-specific overrides of
+CouchDB defaults. Package maintainers should be placing overrides in
+this directory.
+
+System administrator should place overrides in the etc/local.d directory
+instead.
--- a/ansible/roles/couchdb/files/local.d/README
+++ b/ansible/roles/couchdb/files/local.d/README
@@ -1,0 +1,8 @@
+CouchDB local configuration files
+
+Files found under the etc/local.d directory that end with .ini are parsed
+within couchdb(1) at startup.
+
+This directory is intended for system administrator overrides of CouchDB
+defaults. Package maintainers should be placing overrides in the
+etc/default.d directory instead.
--- a/ansible/roles/couchdb/tasks/deploy.yml
+++ b/ansible/roles/couchdb/tasks/deploy.yml
@@ -9,17 +9,26 @@
   fail: msg="The db provider in your {{ hosts_dir }}/group_vars/all is {{ db_provider }}, it has to be CouchDB, pls double check"
   when: db_provider != "CouchDB"
 
-- name: check for persistent disk
-  shell: df -h
-  register: disk_status
-  when: block_device is defined
-
 - name: "set the volume_dir"
-  vars:
-    instance: "{{instances | selectattr('name', 'equalto', 'db') | list | first}}"
   set_fact:
-    volume_dir: "{{ instance.volume.fsmount | default( '/mnt/' + group_names|first, true ) }}:/usr/local/var/lib/couchdb"
-  when: (block_device is defined) and (block_device in disk_status.stdout)
+    volume_dir: "{% if db_datadir is defined  %}[ '{{ db.confdir  }}:/opt/couchdb/etc', '{{ db_datadir  }}:/opt/couchdb/data' ]{% else %}{{ db.confdir  }}:/opt/couchdb/etc{% endif %}"
+
+- name: ensure CouchDB config directory exists
+  file:
+    path: "{{ db.confdir }}"
+    state: directory
+  become: "{{ db.dir.become }}"
+
+- name: copy default configuration files from local to remote in CouchDB config directory
+  copy:
+    src: "files/{{ item }}"
+    dest: "{{ db.confdir }}"
+  with_items:
+        - "default.d"
+        - "default.ini"
+        - "local.d"
+        - "local.ini"
+        - "vm.args"
 
 - name: "pull the apache/couchdb:{{ couchdb.version }} image"
   shell: "docker pull apache/couchdb:{{ couchdb.version }}"
@@ -72,7 +81,7 @@
     user: "{{ db_username }}"
     password: "{{ db_password }}"
     force_basic_auth: yes
-  when: (inventory_hostname == coordinator) and (db.instances|int >= 2)
+  when: inventory_hostname == coordinator
 
 - name: add remote nodes to the cluster
   uri:
@@ -98,4 +107,4 @@
     user: "{{ db_username }}"
     password: "{{ db_password }}"
     force_basic_auth: yes
-  when: (inventory_hostname == coordinator) and (db.instances|int >= 2)
+  when: inventory_hostname == coordinator
--- a/ansible/roles/couchdb/tasks/deploy.yml
+++ b/ansible/roles/couchdb/tasks/deploy.yml
@@ -11,24 +11,21 @@
 
 - name: "set the volume_dir"
   set_fact:
-    volume_dir: "{% if db_datadir is defined  %}[ '{{ db.confdir  }}:/opt/couchdb/etc', '{{ db_datadir  }}:/opt/couchdb/data' ]{% else %}{{ db.confdir  }}:/opt/couchdb/etc{% endif %}"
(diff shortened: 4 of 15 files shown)


style95:
It might be better to make volume mapping for `db.confdir` configurable as well.
Because we always need to delete existing conf dir.
It will be troublesome in case of repeated local deployment.

style95:
Default volume mapping for `db_confdir` is removed.
So there is no volume mapping for CouchDB by default.

And `cookie` is introduced.
Though one of CouchDB nodes is restarted, they will form a cluster.
https://github.com/apache/couchdb/blob/master/rel/overlay/etc/vm.args#L34

style95:
One thing to discuss is, default security mode of CouchDB is `admin_local`.
So anyone can read/write against `5984` port.
How about changing this to `admin_only`?
Then only admin can read/write CouchDB.

style95:
Travis status is green.

style95:
@csantanapr @cbickel 
Could you guys review this please?

style95:
Is there anyone who can review this please?

ansible/group_vars/all
```diff
@@ -203,6 +203,8 @@ nginx:
 # The key db.whisk.auth is the name of the authentication database where all keys of all users are stored.
 # The db_prefix is defined for each environment on its own. The CouchDb credentials are also defined for each environment on its own.
 db:
+  dir:
+    become: "{{ db_dir_become | default(false) }}"
```
rabbah: is this privilege escalation?

ansible/roles/couchdb/files/default.d/README
```diff
@@ -0,0 +1,11 @@
+CouchDB default configuration files
+
+Files found under the etc/default.d directory that end with .ini are
+parsed within couchdb(1) at startup.
+
+This directory is intended for distribution-specific overrides of
+CouchDB defaults. Package maintainers should be placing overrides in
```
rabbah: be placing -> place

ansible/roles/couchdb/files/local.d/README
```diff
@@ -0,0 +1,8 @@
+CouchDB local configuration files
+
+Files found under the etc/local.d directory that end with .ini are parsed
+within couchdb(1) at startup.
+
+This directory is intended for system administrator overrides of CouchDB
+defaults. Package maintainers should be placing overrides in the
```
rabbah: be placing -> place

ansible/roles/couchdb/tasks/deploy.yml
```diff
@@ -9,17 +9,23 @@
...
+    path: "{{ db_confdir }}"
+    state: directory
+  become: "{{ db.dir.become }}"
+  when: db_confdir is defined
+
+- name: copy default configuration files from local to remote in CouchDB config directory
+  copy:
+    src: "files/"
+    dest: "{{ db_confdir }}"
+  when: db_confdir is defined
```
rabbah: if this is predicated on the existence of the files, should we bother checking in template files at all if they're coming from couchdb templates as is?

ansible/roles/couchdb/tasks/deploy.yml
```diff
@@ -9,17 +9,23 @@
...
-  register: disk_status
-  when: block_device is defined
-
 - name: "set the volume_dir"
-  vars:
-    instance: "{{instances | selectattr('name', 'equalto', 'db') | list | first}}"
   set_fact:
-    volume_dir: "{{ instance.volume.fsmount | default( '/mnt/' + group_names|first, true ) }}:/usr/local/var/lib/couchdb"
-  when: (block_device is defined) and (block_device in disk_status.stdout)
+    volume_dir: "{% if db_datadir is defined and db_confdir is defined  %}[ '{{ db_confdir }}:/opt/couchdb/etc', '{{ db_datadir }}:/opt/couchdb/data' ]{% elif db_datadir is undefined  %}{{ db_confdir }}:/opt/couchdb/etca{% else %}{{ db_datadir }}:/opt/couchdb/data{%- endif %}"
```
cbickel: In the past, the persistent part was defined with the block_device.
Which changes have to be made, if someone already used this?
Does he only need to set db_datadir to the same like block_device before?

ansible/roles/couchdb/files/default.d/README
style95: I just coped it from CouchDB docker image.
But if required, I will update it accordingly.

ansible/group_vars/all
style95: Yes if needed.
I just imitated the configuration in nginx.

One more thing to note is, after we deploy couchdb, there would be configuration files.
And if we try to redeploy it, it will fail because couchdb map the configuration files with its own permission(`uid: 1000`). As of now, there is no privilege escalation applied for this.

So, in order to redeploy it, we must remove that directory first.
(If there is no volume mapping for configuration, it's fine.)

Regarding this behavior, I want to discuss which direction is more appropriate.

1. Manually delete all configuration before redeployment.
2. Explicitly delete all configuration before redeployment via Ansible
3. Keep the configuration files and just redeploy couchdb.

One of main purposes of this pr is to make CouchDB restartable.
For some reason, we will change configuration and want to restart it with keeping up existing cluster.

If we change configuration and restart CouchDB manually, there will be no problem.
But if we change and redeploy it via Ansible, we need to consider above situation.

If we decide to redeploy it via Ansible, [serial](http://docs.ansible.com/ansible/latest/playbooks_delegation.html#rolling-update-batch-size) also should be applied as 1, to keep the cluster.

ansible/roles/couchdb/tasks/deploy.yml
style95: Yes, they can just specify the directory in `db_datadir` or `db_confdir`.

And I think, as we are using `apache/couchdb`, `block_device` is no more working.
Because data directory of CouchDB is different; `/usr/local/var/lib/couchdb`.

style95:
I want to know what kinds of tests would be required.
Whether to check volume is correctly mapped between container and host?

If there are any similar test cases for nginx configuration volume mapping, I may refer to that.

style95:
After discussion with @cbickel, I added following changes.

* Add test case to check whether volume mapping is rotted or not
* Synchronize configuration files when deploying
* Change `default_security` to `admin_only`
* Add volume mapping for configuration directory by default.
* Configure serial value as 1.

Logics behind these changes are as follow:

### 1. Test scenario.

In the future, if configuration directory in `apache/couchdb` is changed, our configuration files will be mapped to wrong place. Since there will be already configuration files in the container by default, CouchDB will run without any problems though our configuration files will be located in different place.

Under this situation, `ConfigurationTests` will check if `default_security` is `admin_only`.
Since default value in `apache/couchdb` is `admin_local`, test will be failed, and we can get it volume mapping is rotted.

### 2. Set `serial` value as 1.

Main purpose of this pr is to make CouchDB restartable.
Since CouchDB cluster should be kept during restarting of individual node, it would be better to deploy CouchDB node one by one.


### 3. Synchronization of configuration files.

As volume mapping for configuration is enabled by default just like nginx, we need to synchronize configuration files from `src` directory at every redeployment. So when we change CouchDB configuration and redeploy it, we should delete existing directory first and copy all files from `src`.
This is because `apache/couchdb` change the ownership of configuration files to `uid: 1000` when begins. So I changed `copy` task to `synchronize` task to make these two steps into one.

Now configuration files will be synchronized to `dest` directory and if there are already existing files which is not in `src` directory, they will be deleted.
In this way, we can only concern `src` directory all the time.

If you need to change CouchDB configuration, you can just change it in `src` directory and just redeploy it. It will make sure both directory is same and there will be no any residue in `dest` directory.

style95:
It seems error is not related to this pr.

style95:
Currently there is an issue in Travis build.

CouchDB container changes ownership of working directory when begins via executing one shell script.
https://github.com/apache/couchdb-docker/blob/master/2.1.1/docker-entrypoint.sh#L29

But in Travis environment, it fails when changing ownership of mounted volume due to permission denied.
```
"stderr_lines": [
        "chown: changing ownership of '/opt/couchdb/etc/default.d/README': Operation not permitted", 
        "chown: changing ownership of '/opt/couchdb/etc/default.d': Operation not permitted", 
        "chown: changing ownership of '/opt/couchdb/etc/vm.args': Operation not permitted", 
        "chown: changing ownership of '/opt/couchdb/etc/default.ini': Operation not permitted", 
        "chown: changing ownership of '/opt/couchdb/etc/local.d/README': Operation not permitted", 
        "chown: changing ownership of '/opt/couchdb/etc/local.d/local.ini': Operation not permitted", 
        "chown: changing ownership of '/opt/couchdb/etc/local.d': Operation not permitted", 
        "chown: changing ownership of '/opt/couchdb/etc/local.ini': Operation not permitted", 
        "chown: changing ownership of '/opt/couchdb/etc': Operation not permitted"
    ], 
```

Since CouchDB container is started as root, that script is executed by root.
So it should not happens. 
But it fails. I am looking for any clue.

If anyone has any tips, kindly share that..

style95:
It seems CouchDB container does not run as root so it does not have permission to change ownership.

```
"stdout_lines": [
        "travis    8001  6732  6 16:57 pts/0    00:00:01 /opt/python/2.7.13/bin/python2.7 /home/travis/.local/bin/ansible-playbook -i environments/local -e docker_image_prefix=testing -vvv couchdb.yml", 
        "493216    8451  8433  0 16:57 ?        00:00:00 tini -- /docker-entrypoint.sh /opt/couchdb/bin/couchdb", 
        "493216    8474  8451  0 16:57 ?        00:00:00 /bin/bash /docker-entrypoint.sh /opt/couchdb/bin/couchdb", 
        "493216    8475  8474  4 16:57 ?        00:00:00 chown -R couchdb:couchdb /opt/couchdb", 
        "travis    8555  8001  0 16:57 pts/0    00:00:00 /opt/python/2.7.13/bin/python2.7 /home/travis/.local/bin/ansible-playbook -i environments/local -e docker_image_prefix=testing -vvv couchdb.yml", 
        "travis    8574  8573  0 16:57 pts/0    00:00:00 /bin/sh -c ps -ef | grep couchdb", 
        "travis    8576  8574  0 16:57 pts/0    00:00:00 grep couchdb"
    ]
```

style95:
It seems build is hanging, could anyone abort it please?

style95:
Because of travis limitation, it seems custom image for couchdb is required to pass travis build.
Would it be ok to go further in that direction?

rabbah:
For the custom image you’re thinking to add the dockerfile in this repo? If so I think that works then can publish the image to dockerhub. I’d check with the kube and mesos depencies though - might require downstream changes also.

dgrove-oss:
It would be nice if we could have a single openwhisk/couchdb image that would work for both the main project and the downstream projects like kube.   Right now kube is using its own image (https://github.com/apache/incubator-openwhisk-deploy-kube/tree/master/docker/couchdb).

style95:
@rabbah @dgrove-oss Thank you for comments, I missed them.
I will try with the one in kube project.

style95:
@dgrove-oss
It seems kube project no more maintain its own CouchDB image, how does it handle CouchDB persistency issue during Travis build?

dgrove-oss:
For kube deploy, we switched to using the vanilla apache/couchdb:2.1 image along with a Job using the openwhisk/kube-ansible-runner image to initialize the database.  We support three options in the kube-deploy project:
1. a non-replicated couchdb running in the kubernetes cluster without persistent storage.
2. a non-replicated couchdb running in the kubernetes cluster with persistent storage.
3. using an external (not running in kubernetes cluster) couchdb.  (The couchdb might actually be running in the kubernetes cluster, but it is not deployed by the openwhisk helm chart).

For travisCI, we are only testing options 1.  Our assumption is that most production deployments will probably use option 3 and configure/deploy their databases externally to the deployment of the core openwhisk system.

markusthoemmes:
@style95 any status on this?

style95:
@markusthoemmes I will open a new PR when I find a way to build CouchDB in Travis with persistence.

PR #3111, opened 2017-12-15
Replace our own clients with akka-streams-kafka.

markusthoemmes:
Inspired by @devbv's awesome contributions (thanks 🎉 ) with regards to Kafka clustering, we tried to replace our self-written Kafka abstractions with akka-streams based kafka clients from Lightbend.

This is the outcome of that experiment, it still contains a couple of todos but opening the PR early to gather feedback.

**Important:** As @devbv correctly noted, not all cases are handled by these clients either. They are rather passed to the user to handle. We're in discussions with Lightbend to contribute a more resilient version of their flows if those play out to be production-ready.

In general we found the interface to be much much nicer and it gets rid of a lot of code and tests that we'd need to do ourselves.

**TODO:**

- [ ] Backoff strategies for retries
- [ ] Handle `offer` error cases


dubee:
@markusthoemmes, is this still valid?

PR #3112, opened 2017-12-15
optional bypass of whisk.properties when setting GW_ values

dgrove-oss:
If GW_USER, GW_PWD, or GW_HOST_V2 are already set in the environment
when running installRouteMgmt.sh then use the value from the
environment bypassing the values in whisk.properties.
This is useful for downstream deployments such as kube to avoid
the need to generate a whisk.properties just to run this script.


dgrove-oss:
PG2 2537 🏃

dgrove-oss:
PG2 2357 🔵

Issue #3120, opened 2017-12-20
WhiskAction and ExecutableWhiskAction equality ignores DocRevision


rabbah:
Two actions are equal modulo their revision using the default case class equality method. The invoker uses action equality when matching actions to warm containers. That means creating an action, invoking it, then deleting/recreating/reinvoking it will reuse the previous container.

I don't think this was intended but is likely harmless since equality will require that code matches exactly between the two versions (parameters, annotations also).

I'm going to suggest a test to enforce this behavior should we keep it. Otherwise, we need to define an explicit equals method.

dubeejw:
@rabbah, can this one be closed?

PR #3121, opened 2017-12-20
Enforce WhiskAction equality mod DocRevision.

rabbah:
See #3120 for discussion.


dubeejw:
PG2 2571 🚀

rabbah:
PG failed due to CLI download error. Scheduled PG2 2574.

That said, I don't think a PG is needed since this doesn't change any code and only adds a unit tests that's validated by Travis CI.

PR #3122, opened 2017-12-20
Remove redundant property from activation message.

rabbah:
The property `activationNamespace` in an activation message is redundant with the `user` identity that is also part of the message. We don't need to pass in on the control plane as a result.


rabbah:
PG3/1563 🔵

PR #3125, opened 2017-12-20
Allow / in "host" portion of actionUrlPattern

dgrove-oss:
Allowing / in the "host" portion of the actionUrlPattern supports
deployments where URL rewriting in a front-end proxy is used to allow
multiple services to be hosted from a single hostname or IP address.
This is a common scenario for Kubernetes-based deployments where a
cluster may only have a single exposed IP address or DNS name and its
Ingress does URL rewriting to expose multiple services using different
paths.  For example, when deploying OpenWhisk on a kube cluster we
need to expose two aspects of OpenWhisk using a single host name
  mycluster.cloudprovider.com/openwhisk
  mycluster.cloudprovider.com/apigateway


dgrove-oss:
PG1 2472 🏃

dgrove-oss:
PG1 2472 🆗 - 6 test failures but ones that also happened on recent runs without this change.

rabbah:
@mdeuser @csantanapr can you review?

dgrove-oss:
@csantanapr -- ping

rabbah:
@mdeuser is this something you can review?

dgrove-oss:
@mdeuser -- ping.  We need this fix (or something equivalent) for deploying openwhisk on kube where we need to expose multiple services from a single host name using path prefixes.

dgrove-oss:
I don't mind adding another test to an existing test suite, but I'm not finding any unit tests for the parseActionUrl function.  Did I overlook something?

core/routemgmt/common/apigw-utils.js
```diff
@@ -577,15 +577,15 @@ function generateCliApiFromGwApi(gwApi) {
...
+ *           result[2] : host (i.e. myco.com, 1.2.3.4, myco.com/mywhisk)
  *           result[3] : namespace
  *           result[4] : package name
  *           result[5] : action name
  *           result[6] : action response type (i.e http, json, text, html, or svg)
  */
 function parseActionUrl(actionUrl) {
   console.log('parseActionUrl: parsing action url: '+actionUrl);
-  var actionUrlPattern = /(\w+):\/\/([:\w.\-]+)\/api\/v\d\/web\/([@\w .\-]+)\/([@\w .\-]+)\/([@\w .\-\/]+)\.(\w+)/;
+  var actionUrlPattern = /(\w+):\/\/([:/\w.\-]+)\/api\/v\d\/web\/([@\w .\-]+)\/([@\w .\-]+)\/([@\w .\-\/]+)\.(\w+)/;
```
mdeuser: need to escape that new `/` character?

dgrove-oss: it seems to work without escaping the / (because it is inside a [] ??), but  I can test with the / being escaped.   I'm not an expert of the grammar for JS regex...

dubeejw:
@mdeuser, are there tests for these files?

csantanapr:
Current tests should catches a regression.

PR #3126, opened 2017-12-21
Throttle log messages when rescheduling containers

vvraskin:
In certain cases when invoker is overloaded it will reschedule activations in the container pool. 
In order to prevent log message overflood we print out every N message and log additional information such as user namespace, action, retries count, etc. 

Playground2/2547/ is green

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
@@ -71,6 +71,7 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
   var freePool = immutable.Map.empty[ActorRef, ContainerData]
   var busyPool = immutable.Map.empty[ActorRef, ContainerData]
   var prewarmedPool = immutable.Map.empty[ActorRef, ContainerData]
+  val logRetriesInterval = 1000
 
   prewarmConfig.foreach { config =>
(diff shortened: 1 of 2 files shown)


core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -110,8 +111,15 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
           freePool = freePool - actor
           actor ! r // forwards the run request to the container
         case None =>
-          logging.error(this, "Rescheduling Run message, too many message in the pool")(r.msg.transid)
-          self ! r
+          if (r.retryCount % logRetriesInterval == 0) {
```
rabbah: does it make sense to do this time based? ie every 10seconds?

markusthoemmes:
@vvraskin any idea which triggers this situation? In theory it should **never** be happening (or at least that's the assumption in the code).

Is there some other bug to fix to prevent this from happening?

rabbah:
@markusthoemmes the bug is addressed by #3132.

dubeejw:
@vvraskin, needs rebase.

rabbah:
We should change this to be time based too.

cbickel:
Replaced by #3165

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
mhenke1: I made that change in the replacing PR #3165

PR #3127, opened 2017-12-21
fix PrimitiveActions drag of WhiskActivations

starpit:
the akka `after` implementation creates an actor, but opaquely. thus, we have no opportunity to cancel any timeout futures. this PR adjust the `after` impl to expose and utilize the Cancellables

the consequence of this drag is that WhiskActivations, and a bunch of related artifacts, hang around until various timeouts expire.

this is a descendent of #3118 
the relevant change for this PR is in ExecutionContextFactory.scala, i.e. https://github.com/apache/incubator-openwhisk/pull/3127/files#diff-37b2c2f6b7bfbdbd928baaa6696ba64d


starpit:
closed in favor of cleaner #3129

PR #3129, opened 2017-12-21
fix memory drag in PrimitiveActions, and clean up withAlternateAfterTimeout

starpit:
the akka after implementation creates an actor, but opaquely. thus, we have no opportunity to cancel any timeout futures. this PR adjusts the after impl to expose and utilize the Cancellables

the consequence of this drag is that WhiskActivations, and a bunch of related artifacts, hang around until various timeouts expire.

PG3 1567 🔵 
PG3 1568 🔵

### DIFF
--- a/common/scala/src/main/scala/whisk/utils/ExecutionContextFactory.scala
+++ b/common/scala/src/main/scala/whisk/utils/ExecutionContextFactory.scala
@@ -23,38 +23,61 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.Promise
 import scala.concurrent.duration.FiniteDuration
-import scala.util.Try
+import scala.util.control.NonFatal
 
 import akka.actor.ActorSystem
-import akka.pattern.{after => expire}
+import akka.actor.Cancellable
+import akka.actor.Scheduler
 
 object ExecutionContextFactory {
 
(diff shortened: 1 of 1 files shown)


common/scala/src/main/scala/whisk/utils/ExecutionContextFactory.scala
```diff
@@ -23,38 +23,61 @@ import scala.concurrent.ExecutionContext
...
+   *
+   */
+  def firstCompletedOf2[T](f1: Future[T], f2: CancellableFuture[T])(implicit executor: ExecutionContext): Future[T] = {
+    val p = Promise[T]()
+
+    f1 onComplete { result =>
+      p.tryComplete(result)
+      f2._1.cancel()
+    }
+    f2._2 onComplete p.tryComplete
```
markusthoemmes: Could use some destructuring for ease of reading:

```scala
val (f2Killswitch, f2) = f2Cancellable

f1.onComplete { result =>
  p.tryComplete(result)
  f2Killswitch.cancel()
}

f2.onComplete(p.tryComplete)
```

WDYT?

starpit: i'm happy to oblige. updated pushed

PR #3130, opened 2017-12-22
Remove obsolete filter when determining container to remove.

rabbah:
While collecting a container to remove, the filter checks the action and invocation namespace and excludes those matching the new action from the candidate list. But this check is not necessary - any container in the free pool matching the new action and namespace would be reused and the remove method is not reached in this case.

We could keep these checks as defensive checks. Otherwise it is deadcode and unnecessary overhead.

Tentatively remove, update comment and tests.

For historical context: https://github.com/apache/incubator-openwhisk/pull/2246 but this scheduling heuristic has changed since.


dubeejw:
PG2 2556 🚀

PR #3131, opened 2017-12-22
Updated couchdb docker image name to apache/couchdb

NaohiroTamura:
This is a follow-up patch to the below:
- Add couchdb clustering. (#2810)
  commit e587bf8ec632320614f14114c4e85e67f39c06c2

Please review it,
Thanks.

### DIFF
--- a/ansible/roles/couchdb/tasks/clean.yml
+++ b/ansible/roles/couchdb/tasks/clean.yml
@@ -4,6 +4,5 @@
 - name: remove CouchDB
   docker_container:
     name: couchdb
-    image: klaemo/couchdb
     state: absent
   ignore_errors: True


ansible/roles/couchdb/tasks/clean.yml
```diff
@@ -4,6 +4,6 @@
 - name: remove CouchDB
   docker_container:
     name: couchdb
-    image: klaemo/couchdb
+    image: apache/couchdb
```
csantanapr: This is a clean task to remove the container, I think the image name is ignored, we should try to remove this line

NaohiroTamura: Carlos,
Thanks for the review, I confirmed that "state: absent" doesn't do 'docker rmi', but 'docker rm'. So we should remove it.
--
Naohiro

NaohiroTamura:
Travis failed due to the following error which seems to have nothing to do with this patch.
How can we restart the Travis?

whisk.core.controller.test.ActivationsApiTests > Activations API should get summary activation by namespace and action name FAILED
    org.scalatest.exceptions.TestFailedException: 2 was not equal to 0
        at org.scalatest.Matchers$ShouldMethodHelper$.shouldMatcher(Matchers.scala:6668)

markusthoemmes:
Is this good to go in?

PR #3132, opened 2017-12-22
Separate container removal from job completion.

rabbah:
The current handshake between the container proxy and the container pool conflates container removal and job completion. The former does not necessarily indicate there is free capacity in the pool since there are conditions under which the job is rescheduled: resume failure, container aging, container reclamation.

pg3/1585 🔵

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
@@ -81,6 +81,10 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
 
   def receive: Receive = {
     // A job to run on a container
+    //
+    // Run messages are received either via the feed or from child containers which cannot process
+    // their requests and send them back to the pool for rescheduling (this may happen if "docker" operations
+    // fail for example, or a container has aged and was destroying itself when a new request was assigned)
     case r: Run =>
       val container = if (busyPool.size < maxActiveContainers) {
         // Schedule a job to a warm container
@@ -110,6 +114,9 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
           freePool = freePool - actor
           actor ! r // forwards the run request to the container
         case None =>
+          // this can also happen if createContainer fails to start a new container, or
+          // if a job is rescheduled but the container it was allocated to has not yet destroyed itself
+          // (and a new container would over commit the pool)
           logging.error(this, "Rescheduling Run message, too many message in the pool")(r.msg.transid)
           self ! r
       }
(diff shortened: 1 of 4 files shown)


core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -248,16 +250,20 @@ class ContainerProxy(
           // Sending the message to self on a failure will cause the message
           // to ultimately be sent back to the parent (which will retry it)
           // when container removal is done.
-          case Failure(_) => self ! job
+          case Failure(_) =>
+            rescheduleJob = true
```
dubeejw: Does `rescheduleJob` ever need to be reset to `false`?

rabbah: the transition is ultimately to Removing state which destroys the actor

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -248,16 +250,20 @@ class ContainerProxy(
...
         .pipeTo(self)
 
       goto(Running)
 
-    // timeout or removing
-    case Event(StateTimeout | Remove, data: WarmedData) => destroyContainer(data.container)
+    // container is reclaimed by the pool or it has become too old
+    case Event(StateTimeout | Remove, data: WarmedData) =>
+      rescheduleJob = true // to supress sending message to the pool and not double count
+      destroyContainer(data.container)
```
markusthoemmes: I don't think it's guaranteed that the job is in fact rescheduled in this case, which would lead to a nasty lock situation if too many containers are reclaimed due to timeouts.

In fact I think this specific situation needs to be resolved differently, since at the point of this transition you won't know whether or not you'll resend a job?

rabbah: If a container is reclaimed because of a timeout only, it does not reflect a change in the pool the capacity. It is sufficient to instruct the parent to remove the container from it’s free pool.

What do you mean by it’s not guaranteed that the job is rescheduled in case of a timeout (note there is a test specifically for this case)?

rabbah: > Ideally you'd want Line 272 to set the flag and then eventually send the right message. The message is sent before Line 272 might ever be used though.

I had this thought too, to send a `RescheduledRun` message instead of the current `Run` but the state machine transitions would require a more invasive change, I think, and I abandoned it. Want to try it?

markusthoemmes: Ah you're right indeed. The "timeout means the capacity is already freed earlier" is the trick here.

I might've expressed myself badly: Ideally you'd want Line 272, but it's impossible to do there.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -131,8 +138,21 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
...
+    // Container errored while to resume a warm container, could not process the job, and sent the job back
+    // The container aged, is destroying itself, and was assigned a job which it had to send back
+    // The container aged and is destroying itself
+    case JobRescheduled =>
+      freePool = freePool - sender()
+      busyPool.get(sender()).foreach { _ =>
+        // even though container was busy, no message is sent to the feed since there is no change in capacity yet
+        // (job still needs to be processed)
+        busyPool = busyPool - sender()
+      }
```
markusthoemmes: You can just go with: `busyPool = busyPool - sender()` here, since you don't need the handle on "if the element existed do xyz"

rabbah:
@dubeejw @markusthoemmes thanks for the review, I squashed the commits.

PR #3133, opened 2017-12-23
Subtle changes to support S390X deployment of OpenWhisk

jonpspri:
The intention of this PR is to merge in changes necessary to the OpenWhisk core code base for it to run on a Linux on Z environment. These changes are driven primarily by the unavailability of certain Docker images on the Linux on Z platform and the creation of substitute images (usually in a more recent base image). The hope is that with a move to Xenial/Alpine and (possibly) OpenJDK most remaining incompatibilities would disappear and multi-platform images would become available.

In the meantime, some override capabilities have been introduced into the ansible configuration and the image build has been adjusted to support conditional Dockerfiles. (At one point, I was using m4 and/or templating, but for the moment only pick-a-Dockerfile seems necessary).  An additional file 'ansible/extra/s390x.yml' is used to set overrides where needed for s390x.  It can be added to the ansible-playbook commandline as such:

```
ansible-playbook -i environment/<environment> -e @extra/s390x.yml <playbook>.yml
```

It was also necessary to fiddle with the CLI download capabilities, now that the CLI is being automatically compiled for s390x.  The result is not quite as dynamic as one might wish, yet, but it appears to get the job done.

I'll be on the usual Slack channels.

I have an active Apache CLA.


dgrove-oss:
Thanks for rebasing;  at first glance this looks reasonable to me.  Take a look at the pending change to add a remote_redirect mode for the CLI (#3100) too;  I think it should be fine with the changes you are proposing.

jonpspri:
@dgrove-oss Thanks for pointing out #3100.  I have two concerns, one of which is directly related to adding s390x, and one of which is something I've been scratching my head about anyway.

1. I did away with 'cli_arch' and only reluctantly kept 'cli_os' as ansible variables.  Instead, I introduced 'cli_os_arch' to list VALID combinations of os and architecture.  In particular, when I added 's390x' arch, 'mac' and 'windows' did not make sense as OSs.  Therefore I enumerated valid pairs.  In the new code in #3100, there's still a nested loop of cli_os and cli_arch to create the directory structure.  I can fix it if 3100 goes first, but perhaps we should unify on a strategy, which brings me to...

1. The treatment of `bin/wsk`.  I'm at a loss here -- what's the appropriate action for this binary in each of the three modes?  Here, the binary is being used locally to execute OpenWhisk actions, so we need to have it on the filesystem, but we either download it or copy it from the local build in the next project over.  I suppose the appropriate actions are [ local: copy, remote: download, remote_redirect: download ], but I'm not sure either of us has it that way right now.  I can tweak it if I go second, but it may be worth a second look at #3100.  I'm not sure I have rights to comment at the point over there, but I'll try to.

rabbah:
pg3/1597 🔵

ansible/extra/s390x.yml
```diff
@@ -0,0 +1,17 @@
+### S390X specifics (TODO: could I include what's above?) ###
+
+nginx_version: 1.12
```
csantanapr: if doesn't hurt, let's just update the default from 1.11 to 1.12 and no need to override.
As curiosity version 1.12 has something required to work on s390x?

ansible/extra/s390x.yml
```diff
@@ -0,0 +1,17 @@
...
+
+# Docker images to pull for S390X docker
+
+couchdb:
+    docker_image: "docker.xanophis.com/s390x/couchdb:latest"
+
+kafka:
+    docker_image: "docker.xanophis.com/s390x/wurstmeister-kafka:0.11"
```
csantanapr: Same comment, here where is this image coming from? Is it public, how do we know what's inside.

ansible/extra/s390x.yml
jonpspri: 1.12 docker image is available as multi-architecture image.  I can make the update in this PR, or in a separate PR after this is merged... your call.

ansible/extra/s390x.yml
```diff
@@ -0,0 +1,17 @@
...
+
+apigateway:
+    docker_image: "docker.xanophis.com/s390x/openwhisk-apigateway:latest"
```
csantanapr: apigateway is an image that OpenWhisk project controls, we should not have an image in some registry, we should public/push an apigateway for s390x from the apigateway repo to some public registry.

ansible/extra/s390x.yml
jonpspri: I compiled it and stashed it based on wurstmeister-kafka.  Github repository is at https://github.com/jonpspri/s390x-docker-builds.  We could consider creating a repository for incubator-openwhisk-multiarchitecture-docker-builds or something for tracability and library in DockerHub.  It's an unfortunate problem we have where multi-architecture builds are not yet available.

gradle/docker.gradle
```diff
@@ -29,6 +44,7 @@ ext {
     dockerRetries = project.hasProperty('dockerRetries') ? dockerRetries.toInteger() : 3
     dockerBinary = project.hasProperty('dockerBinary') ? [dockerBinary] : ['docker']
     dockerBuildArg = ['build']
+    dockerArch = project.hasProperty('dockerArch') ? dockerArch : detectDockerArch()
```
csantanapr: I don't get why we need a detect, the build machine might not be the machine running the containers.
I would just kip this detection, and just use dockerArch when it's define and passed in.

ansible/extra/s390x.yml
jonpspri: Agreed.  With your blessing, I can work with the apigateway team on how to create a multi-architecture build of their image as part of their build process.

gradle/docker.gradle
csantanapr: since there is only one arch for now, I would just make is explicit the parameter s390x as a flag/property

gradle/docker.gradle
```diff
@@ -42,14 +58,42 @@ if(project.hasProperty('dockerBuildArgs')) {
     }
 }
 
+task templateDockerfile (type: Copy) {
+  def directory = project.buildscript.sourceFile.getParentFile()
+  from directory
+  into directory
+  include 'Dockerfile.in'
```
csantanapr: I'm trying to grok this new gradle process.
Where is Dockerfile.in coming from?

From I grok the templateDockerfile or rmDockerfile is need it since you are passing `-f Dockerfile.s390x` when your building the specia image for the s390x image

gradle/docker.gradle
jonpspri: Hmm... I see your point, but it opens a big can of worms about where the build is running vs where the target machine is.  This worked great for logging onto the LoZ environment and typing './gradlew distDocker' before running the ansible local configuration.  The detect is useful because it ensure the architecture of the Dockerfile matches to where the build will be run.  The gradle script doesn't really have any insight into target architecture, and passing it a dockerArch that mis-matches the Docker it's running on will have core-generating results.

I can take it or leave it -- removing it just adds one more parameter to remember to get right when running './gradlew'.

gradle/docker.gradle
jonpspri: There are two options for Dockerfile.  'Dockerfile.s390x' if it exists will completely override 'Dockerfile'.  Option 2 is 'Dockerfile.in' would be run through the groovy template engine.  It's intended for small differences in the Dockerfile (e.g. Debian releases).  I'm not sure Option 2 is even in use here, and so perhaps I should just rip it out for the sake of readability/clarity.  I had left it in in case it was needed later.  At any rate, I need better comment documentation apparently.

jonpspri:
Okay.  Let me see if I can get a CLI-changes-only PR together.  Please review my responses to your other reactions and let me know your thoughts after.  I can then start on the needed cleanup.

gradle/docker.gradle
jonpspri: Also, 'ppc64le' is running close on the heels of 's390x'.  We could create a flag for each architecture, I suppose.

jonpspri:
Created #3145 with the CLI changes.  Will re-baseline this PR once #3145 is checked out and merged.

jonpspri:
Just rebased to apache:master to ease the next phases of the conversation.

tools/ubuntu-setup/docker.sh
```diff
@@ -14,7 +27,7 @@ sudo apt-cache policy docker-engine
 sudo apt-get -y install linux-image-extra-$(uname -r)
 
 # DOCKER
-sudo apt-get install -y --force-yes docker-engine=1.12.0-0~trusty
+sudo apt-get install -y --force-yes docker-ce
```
csantanapr: We only support docker 1.12
Maybe you can add a check if ubuntu14/trusty use `docker-engine=1.12.0-0~trusty`

csantanapr:
@jasonpet you have a conflict, also we only support docker 1.12

tools/ubuntu-setup/docker.sh
csantanapr: Need to address the docker1.12/xenial change

jonpspri: @sudswas Could you look into this and let me know what works for Xenial?

jonpspri:
Yeah, it's time to rebase.  Feel's like we are getting close, though.

tools/ubuntu-setup/docker.sh
sudswas: @jonpspri If you do a apt-get install docker.io on xenial - it comes with the following:

```docker version
Client:
 Version:	17.12.0-ce
 API version:	1.35
 Go version:	go1.9.2
 Git commit:	c97c6d6
 Built:	Wed Dec 27 20:08:17 2017
 OS/Arch:	linux/ppc64le
```

csantanapr: Yeah you need to override the defaults so you can get an old version of docker installed

sudswas: AFAIK, ppc64le does not have a distro/docker repo supported version of docker 1.12 - which makes it tricky.  I am not sure if the same problem exists for `s390x` as well? @jonpspri

core/invoker/Dockerfile.s390x
```diff
@@ -0,0 +1,15 @@
+FROM scala
+
+RUN apt-get update \
+ && DEBIAN_FRONTEND="noninteractive" apt-get install -y docker.io \
```
sudswas: @jonpspri @csantanapr is it ok to have any docker version here?

jonpspri: @sudswas  My gut says probably not.  I expect the Docker version should match the docker version being used to run OpenWhisk, which means locking in on a stable version of Docker CE to use.  So we should adjust the s390x and ppc64le Dockerfiles to do that.  Does it make sense for us to use the templating engine so we can provide the configuration variable externally?  I have lots of thoughts and not a clear direction yet.

jonpspri:
I made suggested changes to the 'ubuntu-tools' scripting, starting to pave the way for a move to Xenial as our supported Ubuntu.  I'm starting to feel like there are too many moving parts in this PR again, so I'm going to try to decompose the changes into more digestible pieces to get them moving into the build.  We may want to wait for Bionic -- it's due in 2 months.

tools/ubuntu-setup/all.sh
```diff
@@ -17,8 +18,14 @@ echo "*** install scala"
 "$SCRIPTDIR/scala.sh"
 
 echo "*** installing docker"
-"$SCRIPTDIR/docker.sh"
+if [ "${u_release%%.*}" -lt "16" ]; then
```
csantanapr: I like this is handle

tools/ubuntu-setup/java8.sh
```diff
@@ -1,9 +1,15 @@
...
+arch=`uname -m`
+if [ $arch == "ppc64le" ]
+then
+   sudo apt-get update -y
+   sudo apt-get install openjdk-8-jdk -y
+else
+   sudo apt-get install -y software-properties-common
+   sudo add-apt-repository -y ppa:webupd8team/java
+   sudo apt-get update -y
+   echo oracle-java8-installer shared/accepted-oracle-license-v1-1 select true | sudo /usr/bin/debconf-set-selections
```
csantanapr: I think is ok to use openjdk here for both and and not have a condition, just update java8.sh to do openjdk for all.

tools/ubuntu-setup/all.sh
```diff
@@ -17,8 +18,14 @@ echo "*** install scala"
 "$SCRIPTDIR/scala.sh"
```
csantanapr: Go ahead and remove scala from all.sh, it's not need it gradle takes care of this.

tools/ubuntu-setup/docker-xenial.sh
```diff
@@ -0,0 +1,49 @@
...
+sudo add-apt-repository \
+    "deb [arch=$(uname -m)] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+sudo apt-get -y update -qq
+
+sudo apt-get purge lxc-docker
+sudo apt-cache policy docker-engine
+
+# AUFS
+sudo apt-get -y install "linux-image-extra-$(uname -r)"
+
```
csantanapr: I think for AUF do you also need `linux-image-extra-virtual` I think
```
sudo apt-get -y install linux-image-extra-virtual
``

jonpspri: Actually, for xenial it looks like you don’t need any form of `linux-image-extra` package; the goodies have been compiled in.

core/invoker/Dockerfile.s390x
csantanapr: that's correct same docker version in invoker to match docker engine on the host.

csantanapr: Yes that's OK to templating with variables

ansible/extra/s390x.yml
csantanapr: This needs to be resolved first and fix apigateway first, we cant point to some personal image.

ansible/extra/s390x.yml
csantanapr: maybe instead of another repo, we can have the Dockerfile in this repo and build the image, and can also be posted .
If you could open an issue or send a PR to wurstmeister/kafka to also have s390x at the same time that would also help.
But it's ok to build the image ourselves for now.

rabbah:
@jonpspri is this still ongoing? It was titled `wip`.

abdasgupta:
Is it still going on? I need power support for openwhisk. If this PR is not progressing, I might have to make a new PR working on ppc64le only.

Issue #2927, opened 2017-11-06
bash kind action


style95:
Shell script is commonly used.
If we can create an action with shell script, many users can take advantage of OpenWhisk.
Even they can automate their bash logics or provide it as a REST API.

For example, if someone wants to use OpenStack client feature in OpenWhisk action, it is also possible to implement all logics with OpenStack sdk.
But it is too time-wasting job to implement all the features because there are huge number of  features.

Instead, user can define shell script which execute [openstack cli](https://docs.openstack.org/python-openstackclient/latest/cli/command-list.html) and delegate to it.
User can easily use all functionalities of OpenStack client and automate painful shell-accessing jobs.

It is just an example, I think there might be more use-cases for bash kind action.

In this scene, it would be important to define pre-installed packages.

rabbah:
@style95 this is already supported - the default docker skeleton in fact executes a bash script. Also this article shows how to use the bash support to support a Go action: https://www.ibm.com/blogs/bluemix/2017/01/docker-bluemix-openwhisk/

sungjunyoung:
The bash script has an problem reading openwhisk’s json-like parameters and responsing json

I have modified the code on dockerskeleton’s actionproxy.py  slightly, It allows you to access the value of json from the bash by supplying the json key as an environment variable.

You can visit [my repository](https://github.com/sungjunyoung/openwhisk-bash2action) :)

If you want to execute `cli` command that not in docker image, you just write install script in bash script (exec) when the script is called for the first time

style95:
@rabbah great documents! 
I knew we can run any bash script or binaries with custom docker skeleton but I couldn't imagine such use cases. That would be very useful when user want to add their own dependencies without pulling custom docker image.

For clarity, I wanted `--kind bash` in this pr.
User will be able to create an action like this:

```
wsk action create hello hello.sh
``` 

Then users can also more easily take advantage of what you mentioned to add dependencies.

@sungjunyoung 
It seems your changes will be valuable when user uses `bash` action.
How about creating a PR with your changes.
We can get more people involved and discuss your approach in depth.
It would be more easier to get your changes as well ;)

sungjunyoung:
I'm currently working to do that people can create custom action with `bash` script. I'll create PR In the near future. Thanks, @style95 !

rabbah:
@sungjunyoung thanks for sharing your changes - looks like the main change you made is to recursively project the JSON object's fields into env vars. Did you consider using `jq` instead?

```bash
[...]

PR #3138, opened 2018-01-01
Add bash kind action

upgle:
This is a new bash kind action. Related issue is #2927.

bash action runtime is based on `openwhisk/dockerskeleton` and action proxy is overrided for bash script. also added jq to handle json in bash script

## How to use
```
cat <<EOF > exec
#!/bin/bash
echo \$_test
EOF
```
It returns only the last line, supports auto boxing. (no need to make json object in script)

```
$ wsk -i action update test exec --kind bash
ok: updated action test
$ wsk -i action invoke hello -r -p test "Hello World!"
{
    "result": "Hello World!"
}
```
## Using parameters easily
The parameters are easy to use as environment variables in bash scripts. access parameters using environment variables is more intuitive and easier to use.

[sungjunyoung/openwhisk-bashAction](https://github.com/sungjunyoung/openwhisk-bashAction), This is the repository in which it is implemented

1. Object
```json
{
  "key":{
    "key1":"value1",
    "key2":"value2"
  }
}
```
```bash
#!/bin/bash
echo $_key_key1
echo $_key_key2
```
2. List
```
{
  "key":[
    {"first":"value1"},
    {"second":"value2"}
  ]
}
```
```bash
#!/bin/bash
echo $_key_0_first
echo $_key_1_second
```

3. also can be access theme using jq, jq is aleady installed.
```bash
#!/bin/bash
ARGS=$@
A=`echo "$ARGS" | jq '."a"'`
B=`echo "$ARGS" | jq '."b"'`
C=`echo "$ARGS" | jq '."c"[0]'`
RES=$(($B + $C))
```

## Testing
```
./gradlew tests:test --tests actionContainers.BashActionContainerTests
```

Closes https://github.com/apache/incubator-openwhisk/issues/2927

### DIFF
--- a/actionRuntimes/bashAction/actionproxy.py
+++ b/actionRuntimes/bashAction/actionproxy.py
@@ -1,0 +1,298 @@
+"""Executable Python script for a proxy service to bashAction.
+
+Provides a proxy service (using Flask, a Python web microframework)
+that implements the required /init and /run routes to interact with
+the OpenWhisk invoker service.
+
+The implementation of these routes is encapsulated in a class named
+ActionRunner which provides a basic framework for receiving code
+from an invoker, preparing it for execution, and then running the
+code when required.
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+"""
+
+import sys
+import os
+import json
+import subprocess
+import codecs
+import flask
+from gevent.wsgi import WSGIServer
+import zipfile
+import io
+import base64
+
+class ActionRunner:
+    """ActionRunner."""
+    LOG_SENTINEL = 'XXX_THE_END_OF_A_WHISK_ACTIVATION_XXX'
+
+    # initializes the runner
+    # @param source the path where the source code will be located (if any)
+    # @param binary the path where the binary will be located (may be the
+    # same as source code path)
+    def __init__(self, source=None, binary=None):
+        defaultBinary = '/action/exec'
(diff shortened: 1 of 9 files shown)


rabbah:
@csantanapr should we make this its own repo as with the other runtimes?

actionRuntimes/bashAction/actionproxy.py
```diff
@@ -0,0 +1,298 @@
+"""Executable Python script for a proxy service to bashAction.
```
rabbah: do you need actionproxy vs just overriding the `ActionRunner`? See https://github.com/apache/incubator-openwhisk-runtime-swift/blob/master/core/swift3.1.1Action/swift3runner.py#L113 for example.

ansible/group_vars/all
```diff
@@ -99,6 +99,13 @@ runtimesManifestDefault:
       deprecated: false
       image:
         name: "action-php-v7.1"
+    bash:
```
rabbah: need rebase - this manifest was moved to its file.

csantanapr:
I don't think we need a new image or a new runtime, dockerskeleton should be able to handle bash, perl, any scripting language

I think today you can do
```
wsk action create foo foo.sh --native
```
changing to use `--kind bash` is not shorter 
If you want `jq` support we should add that to the dockerskeleton image

If what you want is shorter syntax then we could update the CLI to sugar certain extentions (i.e. .bash, .sh, .pl)
To detect the `.sh` and remmap that to --native (docker kind with default dockerskeleton):

```
wsk action create foo foo.sh
```

csantanapr:
In addition as pointed by @rabbah we could add `perl` to the dockerskeleton image, and support perl scripts also with a single image.

In terms of the extra step of converting the json input to environments variables I don't feel good about it.
1. redundant work spent passing as env and argv
2. the custom convention of underscores and such not very attractive, I prefer the using jq to parse the json string.
3. I don't like passing all my parameters which would include api keys and passwords, secrets I could be using a 3rd party library or module and I don't want them sniffing to my private input parameters, I prefer to be explicit and when using this 3rd party pass enough to make them work.

csantanapr:
Recent changes to dockerSkeleton runtime now allows for better support for bash actions.
Using `--native` with a simple script is the recommend way for bash actions.

For example:

bash.sh reads  a line from \<STDIN\>
```
$ cat bash.sh
#!/bin/bash
read ARGS
A=`echo "$ARGS" | jq '."a"'`
B=`echo "$ARGS" | jq '."b"'`
C=`echo "$ARGS" | jq '."c"[0]'`
RES=$(($B + $C))
echo A is $A
```
```
wsk action update bash bash.sh --native
```
```
ok: updated action bash
```

```
wsk action invoke bash -p a A -p b 123 -p c '[1,2,3]' -r
```
```json
{
    "result": 124
}
```
```
wsk activation logs --last --strip
```
```
A is "A"
```

docs will update to reflect this new usage.

PR #3142, opened 2018-01-03
externalize runtimesManifestDefault to runtimes.json

dgrove-oss:
Move definition of default runtimes manifest out of
group_vars/all to an external .json file to make it
easier to share this configuration information with
non-ansible deployments such as kube/mesos/etc.

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -30,77 +30,9 @@ whisk:
 #   runtimes: set of language runtime families grouped by language (e.g., nodejs, python)
 #   blackboxes: list of pre-populated docker action images as "name" with optional "prefix" and "tag"
 #   bypassPullForLocalImages: optional, if true, allow images with a prefix that matches {{ docker.image.prefix }}
-#                             to skip docker pull in invoker even if the image is not part of the blackboxe set
+#                             to skip docker pull in invoker even if the image is not part of the blackbox set
 #
-runtimesManifest: "{{ runtimes_manifest | default(runtimesManifestDefault) }}"
-
-runtimesManifestDefault:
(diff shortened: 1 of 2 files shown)


dgrove-oss:
PG2 2577 🏃

dgrove-oss:
PG2 2577 🔵
note: this PR has a companion.

ansible/group_vars/all
```diff
@@ -30,77 +30,9 @@ whisk:
 #   runtimes: set of language runtime families grouped by language (e.g., nodejs, python)
 #   blackboxes: list of pre-populated docker action images as "name" with optional "prefix" and "tag"
 #   bypassPullForLocalImages: optional, if true, allow images with a prefix that matches {{ docker.image.prefix }}
-#                             to skip docker pull in invoker even if the image is not part of the blackboxe set
+#                             to skip docker pull in invoker even if the image is not part of the blackbox set
 #
-runtimesManifest: "{{ runtimes_manifest | default(runtimesManifestDefault) }}"
-
-runtimesManifestDefault:
-  bypassPullForLocalImages: "{{ bypass_pull_for_local_images | default(false) }}"
```
rabbah: @dgrove-oss when you externalized the manifest to its own file we lost this change/ability to override this setting.

Reported by @krugazor on Slack.
I'll open an issue - one option to restore this ability is to read this value from pure config rather than the manifest.

dgrove-oss: @rabbah, I think we should move it to pureconfig.

rabbah: Would be the easiest fix. Makes sense too since the prefix is driven by the local docker prefix not the one in the manifest.

PR #3144, opened 2018-01-03
Env variable instead of parameter

beemarie:


beemarie:
PG: https://openwhisk-jenkins.swg-devops.com:8443/view/Playgrounds/job/Playground2/2583/

PR #3146, opened 2018-01-04
Use cache invalidation policy based on access time

dubeejw:
Currently the cache uses an `expireAfterWrite` policy to invalidate entries. The current policy invalidates entities five minutes after entities are cached even if they are accessed within that time period. 

Using an `expireAfterAccess` policy with a time of five minutes will keep entities in the cache that are accessed within that timeframe. This prevents unnecessary database reads for entities that are used frequently.  

For documentation see: https://github.com/ben-manes/caffeine/wiki/Eviction.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
+++ b/common/scala/src/main/scala/whisk/core/database/MultipleReadersSingleWriterCache.scala
@@ -442,7 +442,7 @@ trait MultipleReadersSingleWriterCache[W, Winfo] {
     Caffeine
       .newBuilder()
       .asInstanceOf[Caffeine[Any, Future[Entry]]]
-      .expireAfterWrite(5, TimeUnit.MINUTES)
+      .expireAfterAccess(5, TimeUnit.MINUTES)
       .softValues()
       .build()
       .asMap())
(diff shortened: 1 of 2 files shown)


dubeejw:
PG3 1598 🔵

dubeejw:
Demonstration of `expireAfterWrite` policy while continuously invoking the same action: 
screen shot 2018-01-03 at 10 31 49 pm

Continuous invocation of the same action with `expireAfterAccess`: 
screen shot 2018-01-03 at 9 45 07 pm

rabbah:
+1 to making it policy configurable per cache. 
Having a source to invalidate identities would be nice (part of a general admin api perhaps) but requires both work here and integration with the iam source.

dubeejw:
Thanks @markusthoemmes, didn't think about the `Identity` cache.

dubeejw:
Pushed changes that allow the cache invalidation policy to be configured base on access time or write time. Where the access time is the default and the `Identity` cache uses the write time policy. 

Also made the `cache` definition lazy, which will presumably prevent the cache from being instantiated if `fromCache` is set to `false`.

dubeejw:
PG3 1605 👍

dubeejw:
@markusthoemmes, can you take another look?

dubeejw:
PG2 2589 🔵

PR #3147, opened 2018-01-04
Disallow create and update of package with reserved names.

rabbah:
Fixes #2964.

Disable update on package with a reserved name. A `get` and `delete` of the package is still allowed. Packages that preexist with a reserved name will continue to work: can add actions to them, or invoke actions in them. But the package itself can't be updated.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Packages.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Packages.scala
@@ -43,8 +43,6 @@ trait WhiskPackagesApi extends WhiskCollectionAPI with ReferencedEntities {
 
   protected override val collection = Collection(Collection.PACKAGES)
 
-  protected[core] val RESERVED_NAMES = Array("default")
-
   /** Database service to CRUD packages. */
   protected val entityStore: EntityStore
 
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/whisk/core/controller/test/PackagesApiTests.scala
```diff
@@ -371,28 +371,27 @@ class PackagesApiTests extends ControllerTestCommon with WhiskPackagesApi {
...
+          status should be(BadRequest)
+          val response = responseAs[String]
+          responseAs[ErrorResponse].error shouldBe Messages.packageNameIsReserved(reservedName)
         }
       }
     }
   }
 
-  it should "allow package update even when package name a reserved name" in {
+  it should "not allow package update even when package name is a reserved name" in {
```
dgrove-oss: Nit:  "not allow package update when package name is a reserved name"

core/controller/src/main/scala/whisk/core/controller/Packages.scala
```diff
@@ -60,6 +58,9 @@ trait WhiskPackagesApi extends WhiskCollectionAPI with ReferencedEntities {
   /** JSON response formatter. */
   import RestApiCommons.jsonDefaultResponsePrinter
 
+  /** Reserved package names. */
+  protected[core] val RESERVED_NAMES = Set("default")
```
ddragosd: Should we hardcode to “default” or make it configurable, with “default” as ... default :smile:

ddragosd: I confused this a bit with the “system” NS. I don’t have a strong pref, nor do I see cases where an operator would wanna change the “default” package. You can Ignore my prev comment.

rabbah: The rationale for reserving `default` is that it's tied to web actions in the "default" package. An explicitly named package with that name would be unreachable.

rabbah:
pg3/1606 🚀

PR #3148, opened 2018-01-04
Make memory limits as config variables

duynguyen:
Some actions requires more than the current hard limit of 512MB of memory.
This PR makes memory limits (min / max / default) configurable per system deployment, via environment variables.

### DIFF
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -82,6 +82,10 @@
       "LIMITS_TRIGGERS_FIRES_PERMINUTE": "{{ limits.firesPerMinute }}"
       "LIMITS_ACTIONS_SEQUENCE_MAXLENGTH": "{{ limits.sequenceMaxLength }}"
 
+      "CONFIG_whisk_memory_min": "{{ limit_action_memory_min | default() }}"
+      "CONFIG_whisk_memory_max": "{{ limit_action_memory_max | default() }}"
+      "CONFIG_whisk_memory_std": "{{ limit_action_memory_std | default() }}"
+
       "RUNTIMES_MANIFEST": "{{ runtimesManifest | to_json }}"
       "CONTROLLER_LOCALBOOKKEEPING": "{{ controller.localBookkeeping }}"
       "AKKA_CLUSTER_SEED_NODES": "{{seed_nodes_list | join(' ') }}"
--- a/common/scala/src/main/scala/whisk/core/entity/MemoryLimit.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/MemoryLimit.scala
@@ -27,10 +27,14 @@ import spray.json.JsValue
 import spray.json.RootJsonFormat
 import spray.json.deserializationError
 import whisk.core.entity.size.SizeInt
+import whisk.core.ConfigKeys
+import pureconfig._
+
+case class MemoryLimitConfig(min: String, max: String, std: String)
 
 /**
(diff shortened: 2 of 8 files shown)


duynguyen:
@tysonnorris , @ddragosd FYI

rabbah:
See https://github.com/apache/incubator-openwhisk/blob/master/docs/dev/configuration.md

duynguyen:
thanks @rabbah . I will update it to pure config.

ddragosd:
:+1: LGTM w @rabbah’s comments.

Would it benefit OW operators more if these limits would be associated with each “kind” in the  runtime_manifest ? If yes, we can consider adding another PR after this one.

rabbah:
Do you envision having different memory spreads per kind?
I like the idea of putting this in the runtime manifest though regardless.

common/scala/src/main/scala/whisk/core/entity/MemoryLimit.scala
```diff
@@ -41,12 +44,14 @@ import whisk.core.entity.size.SizeInt
 protected[entity] class MemoryLimit private (val megabytes: Int) extends AnyVal {}
 
 protected[core] object MemoryLimit extends ArgNormalizer[MemoryLimit] {
-  protected[core] val MIN_MEMORY = 128 MB
-  protected[core] val MAX_MEMORY = 512 MB
-  protected[core] val STD_MEMORY = 256 MB
+  private val memoryConfig = loadConfigOrThrow[MemoryLimitConfig]("whisk.memorylimit")
```
dgrove-oss: If #3150 gets merged first, let's add "whisk.memorylimit" into the PureConfigKeys singleton.

duynguyen:
@dgrove-oss afaics the tests are failing because those pureconfig values are not set in test setup. Do you have any pointer for setting them in tests?
(just asking as you're making the same pureconfig changes in ddoc :) )

dgrove-oss:
I think the problem is that you didn't add the default values to application.conf.   You need to add them in the "whisk" stanza that starts here: https://github.com/apache/incubator-openwhisk/blob/29273808b43fcdf1e2479d8f6b58c3daf1da5c04/common/scala/src/main/resources/application.conf#L52.

Also, your defaults should be default() instead of default("somevalue") so that it loads the default from application.conf

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -82,6 +82,10 @@
       "LIMITS_TRIGGERS_FIRES_PERMINUTE": "{{ limits.firesPerMinute }}"
       "LIMITS_ACTIONS_SEQUENCE_MAXLENGTH": "{{ limits.sequenceMaxLength }}"
 
+      "CONFIG_whisk_memorylimit_min": "{{ limit_action_memory_min | default('128M') }}"
```
dgrove-oss: all of these should be default()

duynguyen:
@dgrove-oss thanks, I have updated the PR according to the review.

dubeejw:
PG3 1629 🚀

dubeejw:
PG3 1630 🔴

rabbah:
@dubeejw any info on the red?

rabbah:
@duynguyen I rebased your branch to resolve the conflict. You'll have to repull.

rabbah:
PG2/2634 🏃

rabbah:
@dgrove-oss kube changes needed?

duynguyen:
thanks @rabbah @dgrove-oss

PR #3149, opened 2018-01-04
Cleanup test assets between unit tests using afterEach

rabbah:
No semantic change is intended here. This is cleaning up some incorrect uses of `after`.
Some additional unit tests are added for checking that empty activation views are handled correctly.
The commits are mutually exclusive for review.


rabbah:
I thought about it - they share the same underlying method, so I thought it's sufficient.
We should probably refactor these API tests so that there's a base common set that we can iterate over all the collections rather than clone and own the tests in each. Already there's a bunch of duplication.

rabbah:
pg2/2586 🚀

PR #3150, opened 2018-01-05
singleton to hold PureConfig keys

dgrove-oss:
As suggested by @rabbah in review of another PR, introduce
a singleton to hold the various String keys used to load
config data via PureConfig to make it easier to find all
uses of a particular piece of configuration data.


dgrove-oss:
PG2 2591 🏃

dgrove-oss:
PG2 2591 👌 -- 1 test failure that can't be due to this change.

PR #3152, opened 2018-01-06
Enrich entity views, and add count reducer

rabbah:
The view for actions is enriched to include limits, and if an action is binary or not. I imagine that as the exec blob is moved to an attachment, the view computation becomes simpler and includes the entire doc. This would avoid having to add some annotation on the document which then become redundant (`kind` for example). I've done this for rules in the proposed patch herein.

In addition, add a `count` reducer to all the views in service of providing an API to return entity (including activation) counts.

The first commit adds the new views as `v2.1.0` as they add fields but don't remove any existing ones.

PR #3155, opened 2018-01-07
Support ?count on collections

rabbah:
Builds on #3152 to provide counts for the collections. See https://github.com/apache/incubator-openwhisk/issues/3051#issuecomment-355751801.


common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -329,6 +329,44 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
...
 
+  protected[core] def count(table: String, startKey: List[Any], endKey: List[Any], skip: Int, stale: StaleParameter)(
+    implicit transid: TransactionId): Future[Long] = {
+
+    val (realStartKey, realEndKey) = (startKey, endKey)
+    val parts = table.split("/")
+
+    val start = transid.started(this, LoggingMarkers.DATABASE_QUERY, s"[COUNT] '$dbName' searching '$table")
+
+    val f = for (eitherResponse <- client.executeView(parts(0), parts(1))(
```
cbickel: Is there a special  reason why you are using a for-comprehension instead of map?

rabbah: Followed the other methods in the file. It’s the same just easier to read in some cases.

markusthoemmes: A map would be easier to follow here though.

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -329,6 +329,44 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
...
+              assert(rows.length == 1, s"result of reduced view contains more than one value: '$rows'")
+              rows.head.fields("value").convertTo[Long]
+            } else 0L
+
+            transid.finished(this, start, s"[COUNT] '$dbName' completed: count $out")
+            out
+
+          case Left(code) =>
+            transid.failed(this, start, s"Unexpected http response code: $code", ErrorLevel)
+            throw new Exception("Unexpected http response code: " + code)
```
markusthoemmes: I know the rest of the file does this differently, but would you mind adding a `flatMap` above and use an explicit `Future.failed` here?

tests/src/test/scala/whisk/core/controller/test/PackagesApiTests.scala
```diff
@@ -107,29 +123,31 @@ class PackagesApiTests extends ControllerTestCommon with WhiskPackagesApi {
...
+      val path = s"/$namespace/${collection.path}"
+      val auser = WhiskAuthHelpers.newIdentity()
+      val expected = providers.filter(p => p.namespace == namespace && p.publish) ++
+        references.filter(p => p.publish && p.binding == None)
+
+      checkCount(path, expected.length, auser)
+      Get(path) ~> Route.seal(routes(auser)) ~> check {
+        status should be(OK)
+        val response = responseAs[List[JsObject]]
+        response.length shouldBe expected.length
```
markusthoemmes: `response should have size expected.size` for a better error message? (applies to others)

rabbah:
PG2/2650 🔵

jthomas:
Do we still keep the API swagger current with changes like these? For the OW JS SDK, that file was super useful when implemeting the library. I'm not sure where else we are documenting the API if not there?

rabbah:
My oversight. We should update the swagger.

csantanapr:
Who is updating the swagger?

rabbah:
i think we need to move to swagger 3.0 to use `oneof` since this query changes the response schema.

csantanapr:
>I think we need to move to swagger 3.0 to use oneof since this query changes the response schema.

I don’t think there should be a problem with this. Do we have an issue tracking this work?

PR #3156, opened 2018-01-08
move PureConfig strings to PureConfigKeys singleton

dgrove-oss:


dgrove-oss:
PG1 2502 🏃

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -244,4 +244,7 @@ object PureConfigKeys {
   val whiskKafkaTopics = "whisk.kafka.topics"
 
   val whiskDB = "whisk.db"
+
+  val whiskDockerTimeouts = "whisk.docker.timeouts"
+  val whiskRuncTimeouts = "whisk.runc.timeouts"
```
markusthoemmes: I feel like we should drop `Pure` since it's an implementation detail and `whisk` from all the names, forming:

```scala
object ConfigKeys {
  val loadbalancer = "whisk.loadbalancer"

  val kafka = "whisk.kafka"
  val kafkaTopics = s"$kafka.topics"

  val db = "whisk.db"

  val docker = "whisk.docker"
  val dockerTimeouts = s"$docker.timeouts"
  val runc = "whisk.runc"
  val runcTimeouts = s"$runc.timeouts"
}
```

WDYT?

dgrove-oss: Makes sense to me.  I'll do a follow up PR to drop Pure and apply the s"$" cascading in your comment.

PR #3159, opened 2018-01-08
Sanitize error message returned by a failed container startup.

markusthoemmes:
The error message returned by a failing container startup contains internal information (like the internally configured images, various exception messages). Returning a sanitized string is better design in general here.

All error cases are logged fine for operators to find the root-cause of the failing run.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
@@ -147,7 +147,7 @@ class ContainerProxy(
             // the failure is either the system fault, or for docker actions, the application/developer fault
             val response = t match {
               case WhiskContainerStartupError(_) => ActivationResponse.whiskError(Messages.resourceProvisionError)
-              case BlackboxStartupError(_)       => ActivationResponse.applicationError(Messages.resourceProvisionError)
+              case BlackboxStartupError(msg)     => ActivationResponse.applicationError(msg)
               case _                             => ActivationResponse.whiskError(Messages.resourceProvisionError)
             }
             // construct an appropriate activation and record it in the datastore,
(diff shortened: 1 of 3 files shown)


dubeejw:
Looks like there is one test failure: ```

whisk.core.cli.test.WskBasicUsageTests > Wsk actions should create, and invoke an action that utilizes an invalid docker container with appropriate error FAILED
    org.scalatest.exceptions.TestFailedException: "Failed to provision resources to run the action." was not equal to "Failed to pull container image 'boguschgac47qtmfgbscq'."
```

File: core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala [hunk 1]
```diff
@@ -146,9 +146,9 @@ class ContainerProxy(
             // the container did not come up cleanly, so disambiguate the failure mode and then cleanup
             // the failure is either the system fault, or for docker actions, the application/developer fault
             val response = t match {
-              case WhiskContainerStartupError(msg) => ActivationResponse.whiskError(msg)
-              case BlackboxStartupError(msg)       => ActivationResponse.applicationError(msg)
-              case _                               => ActivationResponse.whiskError(t.getMessage)
+              case WhiskContainerStartupError(_) => ActivationResponse.whiskError(Messages.resourceProvisionError)
+              case BlackboxStartupError(_)       => ActivationResponse.applicationError(Messages.resourceProvisionError)
```
rabbah: @dubeejw this comes from the change here:

https://github.com/apache/incubator-openwhisk/pull/3159#issuecomment-356098093

rabbah: might be worth keeping the old message, more informative if it's truly a `pull` issue.

markusthoemmes: I agree, I'll update the inner messages accordingly.

Issue #3161, opened 2018-01-09
Add type property to entity documents


chetanmeh:
Currently all whisk entities are stored in "whisks" collection in couchdb. Most queries around these entities are of form [root namespace, date] invoked against entity specific views.

    GET /local_whisks/_design/whisks.v2/_view/packages?descending=true&endkey=["guest",0]&reduce=false&startkey=["guest","￰","￰"]&limit=30
    
The views are computed by "inferring" the entity type [0] based on document properties.This mode of querying works fine with couchdb which provides "computed" views. However it poses problem with storage system where query relies on declarative indexes.

As discussed on [mailing list][1] we should add a `type` property which is set to entity type.

Would open a separate issue for namespace

[0]
```
var isPackage = function (doc) {
        return (doc.binding !== undefined)
 };
var isAction = function (doc) {
     return (doc.exec !== undefined)
 };
 var isTrigger = function (doc) {
        return (doc.exec === undefined && doc.binding === undefined &&
doc.parameters !== undefined)
 };
 var isRule = function (doc) {
        return (doc.trigger !== undefined)
 };
```

[1]: https://lists.apache.org/thread.html/36891e1f3a37312e3923b90717599a4a8a5c8260aea66a0b0fc5f4db@%3Cdev.openwhisk.apache.org%3E

chetanmeh:
@rabbah  Opened PR #3162 with initial approach which uses reflection to determine the type property value based on entity class name. This approach has potential issues like

# Add slight overhead due to use of reflection
# The type value may change in entity class name is refactored later causing backwarnd compatibility issue

Other possible approaches can be

# Add type property explicitly with each type
# OR derive the property in `WhiskDocument` based on same logic as used in CouchDb views

Let me know which approach should be taken

rabbah:
At first glance I think adding type property explicitly is more direct and clean. An abstract method in WhiskEntity that all the subtypes must implement which returns a string and added to the properties written in the WhiskEntity base serializer.

chetanmeh:
Closed via #3162

PR #3162, opened 2018-01-09
Add type property to entity documents 

chetanmeh:
Adds a `type` property derived from entity class name

This PR is first attempt for #3161. See the issue for more details. It adds a `type` property based on entity class

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskStore.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskStore.scala
@@ -70,6 +70,14 @@ protected[core] trait WhiskDocument extends DocumentSerializer with DocumentRevi
   protected[core] final def docinfo: DocInfo = DocInfo(docid, rev)
 
   /**
+   * Determines entity type based on entity class name
+   */
+  protected[core] def documentType: String = {
+    val typeName = this.getClass.getSimpleName.toLowerCase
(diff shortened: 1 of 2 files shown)


common/scala/src/main/scala/whisk/core/entity/WhiskStore.scala
```diff
@@ -69,6 +69,14 @@ protected[core] trait WhiskDocument extends DocumentSerializer with DocumentRevi
    */
   protected[core] final def docinfo: DocInfo = DocInfo(docid, rev)
 
+  /**
+   * Determines entity type based on entity class name
+   */
+  protected[core] def documentType: String = {
+    val typeName = this.getClass.getSimpleName.toLowerCase
+    if (typeName.startsWith("whisk")) typeName.substring(5) else typeName
+  }
```
markusthoemmes: Should we rather override this property in the specific subtypes instead of relying on "magic"?

Alternatively, maybe we can get away with something like:

```scala
protected[core] def documentType: String = this match {
  case _: WhiskAction => "action"
  case _: WhiskRule => "rule"
  ...
}
```

Didn't test this though.

chetanmeh: Ack. Would work on a new PR

PR #3163, opened 2018-01-09
Configure jmxremote (updated)

mhenke1:
This PR replaces PR #2745 `Configure jmxremote`
Please check this PR for the discussions and reviews

Original text:
Hello,

this change will enable/make configurable jmx remote configuration for controllers.
A couple of benefits it provides:

    change logback log level without JVM restart
    check akka cluster state
    JVM profiling with visualVM/JProfile

regards, 
Vadim.

### DIFF
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -16,18 +16,25 @@
     mode: 0777
   become: "{{ logs.dir.become }}"
 
+- name: ensure controller config directory is created with permissions
+  file:
+    path: "{{ controller.confdir }}/controller{{ groups['controllers'].index(inventory_hostname) }}"
+    state: directory
+    mode: 0777
+  become: "{{ controller.dir.become }}"
+
 - name: copy jmxremote password file
   when: jmx.enabled
   template:
     src: "jmxremote.password.j2"
-    dest: "{{ whisk_logs_dir }}/controller{{ groups['controllers'].index(inventory_hostname) }}/jmxremote.password"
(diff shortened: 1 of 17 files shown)


ansible/roles/controller/tasks/deploy.yml
```diff
@@ -16,6 +16,27 @@
     mode: 0777
   become: "{{ logs.dir.become }}"
 
+- name: ensure controller config directory is created with permissions
+  file:
+    path: "{{ controller.confdir }}/controller{{ groups['controllers'].index(inventory_hostname) }}"
+    state: directory
+    mode: 0777
```
cbickel: Is 0777 necessary here?

cbickel: Same for the files above.

core/controller/init.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash
 
+if [[ $( ls /conf/jmxremote.* 2> /dev/null ) ]]
+then
+  mv /conf/jmxremote.* /root ; chmod 600 /root/jmxremote.*
```
cbickel: Woukd it make sense to create a new script in the scala container, that handles these jmx-files?
If doing it there, we would avoid the code duplication in the init script in controller and invoker.

ansible/roles/controller/tasks/deploy.yml
mhenke1: Yes. File move operations need these rights.

core/controller/init.sh
mhenke1: As suggested I have moved the common functionality into a  file that is copied into the Scala container

mhenke1:
PG3 1632 sucessful

Issue #3027, opened 2017-11-29
ActivationsApiTests fails with no results from list query


rabbah:
Observed in couple of Travis PR builds.

```
whisk.core.controller.test.ActivationsApiTests > Activations API should get summary activation by namespace and action name FAILED
    org.scalatest.exceptions.TestFailedException: 2 was not equal to 0
        at org.scalatest.MatchersHelper$.indicateFailure(MatchersHelper.scala:340)
        at org.scalatest.Matchers$ShouldMethodHelper$.shouldMatcher(Matchers.scala:6668)
        at org.scalatest.Matchers$AnyShouldWrapper.should(Matchers.scala:6704)
        at whisk.core.controller.test.ActivationsApiTests$$anonfun$11$$anonfun$apply$30$$anonfun$apply$31.apply(ActivationsApiTests.scala:328)
```

It failed on the `activations.length should be(response.length)`. 
The initial wait on view was ok; it confirmed at least one view was up to date. If couchdb is not clustered in [Travis](https://github.com/apache/incubator-openwhisk/blob/master/ansible/environments/local/hosts#L26), why does the query fail?

rabbah:
answering my question: the `waitOnView` is checking the wrong index. The filtering by name using a different view. So need to adjust the wait.

PR #3164, opened 2018-01-09
Add type property to entity documents - Take 2

chetanmeh:
Adds a `type` property to the document json based on `entityType` method

This PR is second attempt for #3161. See the issue for more details. It adds a type property based onexplicitly defined `entityType` method value

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
@@ -88,6 +88,8 @@ abstract class WhiskActionLike(override val name: EntityName) extends WhiskEntit
       parameters.definedParameters
     } else Set.empty[String]
 
+  def entityType = "action"
+
   def toJson =
     JsObject(
       "namespace" -> namespace.toJson,
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
@@ -72,7 +72,7 @@ case class WhiskActionPut(exec: Option[Exec] = None,
   }
 }
 
-abstract class WhiskActionLike(override val name: EntityName) extends WhiskEntity(name) {
+abstract class WhiskActionLike(override val name: EntityName) extends WhiskEntity(name, "action") {
   def exec: Exec
   def parameters: Parameters
   def limits: ActionLimits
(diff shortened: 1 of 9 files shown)


common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -88,6 +88,8 @@ abstract class WhiskActionLike(override val name: EntityName) extends WhiskEntit
       parameters.definedParameters
     } else Set.empty[String]
 
+  def entityType = "action"
```
rabbah: Could be a val.

common/scala/src/main/scala/whisk/core/entity/WhiskEntity.scala
```diff
@@ -60,11 +60,17 @@ abstract class WhiskEntity protected[entity] (en: EntityName) extends WhiskDocum
...
+
   /**
    * Returns a JSON object with the fields specific to this abstract class.
    */
   protected def entityDocumentRecord: JsObject =
-    JsObject("name" -> JsString(name.toString), "updated" -> JsNumber(updated.toEpochMilli()))
+    JsObject(
+      "name" -> JsString(name.toString),
+      "updated" -> JsNumber(updated.toEpochMilli()),
+      "type" -> JsString(entityType))
```
rabbah: I’d use the property kind instead of type, or entityType. The string type could cause problems in the future because it’s a reserved string in scala and would have to be escaped in a case class for example with default serdes.

chetanmeh: Ack. Used `entityType` as property name as `kind` is already being used for `exec` to avoid confusuion

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
chetanmeh: Ack. Made it an instance val. Used def earlier to avoid extra memory usage for in memory representation of entity. However val looks like the right way now

chetanmeh:
The build failure is in ActivationsApiTests which I think is not related to changes here per #3027

```
whisk.core.controller.test.ActivationsApiTests > Activations API should get summary activation by namespace and action name FAILED
    org.scalatest.exceptions.TestFailedException: 2 was not equal to 1
        at org.scalatest.Matchers$ShouldMethodHelper$.shouldMatcher(Matchers.scala:6668)
        at org.scalatest.Matchers$AnyShouldWrapper.should(Matchers.scala:6704)
```

rabbah:
Agree. One aspect that’s giving me pause is that activation records will inherit this but are in their own non ambiguous database. But refactoring the type hierarchy to also address other inconsistencies between activations and other entities is a seperate topic - that we’ll need to address.

chetanmeh:
> You could rename the test to "it should define the entityType property in its json representation" since that's what it asserts.

Makes sense. Updated the PR now

PR #3165, opened 2018-01-09
 Throttle log messages when rescheduling containers (update) 

mhenke1:
This PR replaces PR #3126 

In certain cases when invoker is overloaded it will reschedule activations in the container pool.
In order to prevent log message overflood we print out every 10 seconds and log additional information such as user namespace, action, retries count, etc.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
@@ -71,6 +71,8 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
   var freePool = immutable.Map.empty[ActorRef, ContainerData]
   var busyPool = immutable.Map.empty[ActorRef, ContainerData]
   var prewarmedPool = immutable.Map.empty[ActorRef, ContainerData]
+  var lastRetryLog: Option[Long] = None
+  val logMessageInterval = 10000;
 
   prewarmConfig.foreach { config =>
     logging.info(this, s"pre-warming ${config.count} ${config.exec.kind} containers")(TransactionId.invokerWarmup)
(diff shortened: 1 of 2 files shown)


core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -71,6 +71,8 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
   var freePool = immutable.Map.empty[ActorRef, ContainerData]
   var busyPool = immutable.Map.empty[ActorRef, ContainerData]
   var prewarmedPool = immutable.Map.empty[ActorRef, ContainerData]
+  var lastRetryLog: Option[Long] = None
```
markusthoemmes: I don't think we want to globally throttle, but throttle per message. The current code would different messages being swallowed.

That's actually why @vvraskin created the `retries` field of `Run`. In the same sense I think you'd want to create a `lastRetry: Instant` field, which can be `Instant.EPOCH` by default.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -71,6 +71,8 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
...
+  val logMessageInterval = 10000;
```
markusthoemmes: Can we use `FiniteDuration` here and work with higher level abstractions (`Instant`) instead of `Long`s?

mhenke1: I changed the logic to compare a given `Deadline`

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
mhenke1: Though we saw that only one message is looping at a given time, I agree that it is better to track the time for each run seperately. Changed

PR #191, opened 2016-04-07
Update README.md

lzbj:
Add one build tip.

PR #3167, opened 2018-01-09
Remove API for listing all entities in namespace

rabbah:
Proposed patch for https://github.com/apache/incubator-openwhisk/issues/3153.


rabbah:
@csantanapr The node runtimes have been updated, and the CLI as well. I think this is ready then.

rabbah:
pg5/104 🏃

cbickel:
In general this PR looks good to me.
But this will be a breaking change in our api, right?
Does this has to be communicated to the community and customers of all the OpenWhisk-installations before we merge it?
And what about the cli command `wsk list`? Will the CLI make 4 requests instead?

rabbah:
@cbickel https://lists.apache.org/thread.html/d49727a3057759bd2879b5b4e37a63e3b6d9b916143f9d315e20d2dc@%3Cdev.openwhisk.apache.org%3E

also I've already updated the CLI and npm openwhisk to sugar the call (and make 4 calls instead).

mrutkows:
See PR: https://github.com/apache/incubator-openwhisk-cli/pull/191
"Handle namespace API listing on the client. #191"

csantanapr:
@cbickel OK to merge this?

cbickel:
@csantanapr As you are the only Assignee and not one of the Reviewers, you can decide.

PR #3170, opened 2018-01-10
Add controller info route test

dubeejw:
Ensures that the controller info route returns the expected response.

### DIFF
--- a/tests/src/test/scala/system/rest/ControllerTests.scala
+++ b/tests/src/test/scala/system/rest/ControllerTests.scala
@@ -39,8 +39,8 @@ import whisk.core.WhiskConfig
 class ControllerTests extends FlatSpec with RestUtil with Matchers with ExecHelpers with StreamLogging {
 
   it should "ensure controller returns info" in {
-    val response = RestAssured.given().config(sslconfig).get(getServiceURL)
-    val runtimes = getRuntimesManifest
+    val response = RestAssured.given.config(sslconfig).get(getServiceURL)
+    val runtimes = runtimesManifest
     val config = new WhiskConfig(
       Map(
         WhiskConfig.actionInvokePerMinuteLimit -> null,
@@ -57,7 +57,7 @@ class ControllerTests extends FlatSpec with RestUtil with Matchers with ExecHelp
         "actions_per_minute" -> config.actionInvokePerMinuteLimit.toInt.toJson,
         "triggers_per_minute" -> config.triggerFirePerMinuteLimit.toInt.toJson,
         "concurrent_actions" -> config.actionInvokeConcurrentLimit.toInt.toJson))
-    response.statusCode() should be(200)
+    response.statusCode should be(200)
     response.body.asString.parseJson shouldBe (expectedJson)
   }
 }
--- a/tests/src/test/scala/system/rest/ControllerTests.scala
+++ b/tests/src/test/scala/system/rest/ControllerTests.scala
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/system/rest/ControllerTests.scala
```diff
@@ -0,0 +1,63 @@
...
+      "api_paths" -> JsArray("/api/v1".toJson),
+      "runtimes" -> runtimes.toJson,
+      "limits" -> JsObject(
+        "actions_per_minute" -> config.actionInvokePerMinuteLimit.toInt.toJson,
+        "triggers_per_minute" -> config.triggerFirePerMinuteLimit.toInt.toJson,
+        "concurrent_actions" -> config.actionInvokeConcurrentLimit.toInt.toJson))
+    response.statusCode() should be(200)
+    response.body.asString.parseJson shouldBe (expectedJson)
+  }
+}
```
markusthoemmes: Can this be written in a Unit rather than a system test?

rabbah: We need a system test as well, otherwise routers could be misconfigured and screw this up.

rabbah: Ah maybe I misunderstood. Do you mean does it need to be in the system package?

tests/src/test/scala/whisk/core/entity/test/ExecHelpers.scala
```diff
@@ -99,4 +99,6 @@ trait ExecHelpers extends Matchers with WskActorSystem with StreamLogging {
   protected def bb(image: String, code: String, main: Option[String] = None) = {
     BlackBoxExec(ExecManifest.ImageName(trim(image)), Some(trim(code)).filter(_.nonEmpty), main, false)
   }
+
+  protected def getRuntimesManifest() = ExecManifest.runtimesManifest
```
markusthoemmes: Should drop the "get" and `()` since this is just an accessor?

rabbah: Why not just use ExecManifest.runtimesManifest
directly?

tests/src/test/scala/system/rest/ControllerTests.scala
markusthoemmes: You didn't misunderstand but you're right. All tests under "system" are not to rely on `WhiskConfig`. Those tests need to be runnable with API information only IIRC.

rabbah: 👍 it’s enough to run this test outside the system package.

tests/src/test/scala/whisk/core/entity/test/ExecHelpers.scala
dubeejw: I believe `ExecManifest.runtimesManifest` can only be accessed from packages in the `core`. It is defined with an access modifier `protected[core] object ExecManifest {`.

rabbah: ok - rather than add another get method to change the package protection, widen it at the source in that case.

tests/src/test/scala/system/rest/ControllerTests.scala
dubeejw: We want to ensure that nginx routes us to the proper location, so a unit test wouldn't suffice.

markusthoemmes: Still this cannot be part of the system package due to the usage of WhiskConfig.

rabbah:
@dubeejw i think this is a simple fix to move the test to a different package, no?

dubeejw:
@rabbah, that is a possibility. Haven't had a chance to update this one yet.

dubeejw:
@rabbah, moved the test to another package? Anything else?

PR #3174, opened 2018-01-10
Change log level dynamically (updated)

mhenke1:
This PR replaces PR #2990 (please see this PR for discussions and comments)

This will allow to change log level on per request/user basis for OW transactions.
If a specific header is provided in the http request to the controller, it will log messages on the debug level for that given transaction id.
@vvraskin 

regards, Vadim and Martin.

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -161,7 +161,6 @@ nginx:
     verify_client: "{{ nginx_ssl_verify_client | default('off') }}"
   wpn:
     router: "{{ nginx_wpn_router | default('1') }}"
-  # an array of user ids to enable extra logging on the system
   special_users: "{{ nginx_special_users | default('[]') }}"
 
 # These are the variables to define all database relevant settings.
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -161,6 +161,8 @@ nginx:
     verify_client: "{{ nginx_ssl_verify_client | default('off') }}"
   wpn:
     router: "{{ nginx_wpn_router | default('1') }}"
+  # an array of user ids to enable extra logging on the system
+  special_users: "{{ nginx_special_users | default('[]') }}"
 
 # These are the variables to define all database relevant settings.
 # The authKeys are the users, that are initially created to use OpenWhisk.
--- a/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/ansible/roles/nginx/templates/nginx.conf.j2
@@ -49,7 +49,7 @@ http {
       "special:off" off;
       "special:on" on;
       "special:" on;
-      }
+    }
 
     server {
         listen 443 default ssl;
--- a/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/ansible/roles/nginx/templates/nginx.conf.j2
@@ -37,6 +37,20 @@ http {
         keepalive 512;
     }
 
+    map "$remote_user" $special_user {
+      default not_special;
+{% for user in nginx.special_users %}
+      "{{ user }}" special;
+{% endfor %}
+    }
+
+    map "$special_user:$http_x_ow_extra_logging" $extra_logging {
+      default              "off";
+      "special:off" off;
+      "special:on" on;
+      "special:" on;
+      }
+
     server {
         listen 443 default ssl;
 
@@ -65,6 +79,7 @@ http {
             if ($namespace) {
               rewrite    /(.*) /api/v1/web/${namespace}/$1 break;
             }
+            proxy_set_header X-OW-EXTRA-LOGGING $extra_logging;
             proxy_pass http://controllers;
             proxy_read_timeout 75s; # 70+5 additional seconds to allow controller to terminate request
         }
@@ -74,6 +89,7 @@ http {
             if ($namespace) {
               rewrite    ^ /api/v1/web/${namespace}/public/index.html break;
             }
+            proxy_set_header X-OW-EXTRA-LOGGING $extra_logging;
             proxy_pass http://controllers;
             proxy_read_timeout 75s; # 70+5 additional seconds to allow controller to terminate request
         }
--- a/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/ansible/roles/nginx/templates/nginx.conf.j2
@@ -45,7 +45,7 @@ http {
     }
 
     map "$special_user:$http_x_ow_extra_logging" $extra_logging {
-      default              "off";
+      default "off";
       "special:off" off;
       "special:on" on;
       "special:" on;
--- a/common/scala/src/main/scala/whisk/common/Logging.scala
+++ b/common/scala/src/main/scala/whisk/common/Logging.scala
@@ -37,7 +37,11 @@ trait Logging {
    * @param message Message to write to the log
    */
   def debug(from: AnyRef, message: String)(implicit id: TransactionId = TransactionId.unknown) = {
-    emit(DebugLevel, id, from, message)
+    if (id.meta.extraLogging) {
+      emit(InfoLevel, id, from, message)
+    } else {
+      emit(DebugLevel, id, from, message)
+    }
   }
 
   /**
--- a/common/scala/src/main/scala/whisk/common/TransactionId.scala
+++ b/common/scala/src/main/scala/whisk/common/TransactionId.scala
@@ -24,13 +24,9 @@ import java.util.concurrent.atomic.AtomicInteger
 
 import scala.math.BigDecimal.int2bigDecimal
 import scala.util.Try
-
 import akka.event.Logging.{InfoLevel, WarningLevel}
 import akka.event.Logging.LogLevel
-import spray.json.JsArray
-import spray.json.JsNumber
-import spray.json.JsValue
-import spray.json.RootJsonFormat
+import spray.json.{JsArray, JsBoolean, JsNumber, JsValue, RootJsonFormat}
 
 import whisk.core.ConfigKeys
 
@@ -202,7 +198,7 @@ case class StartMarker(val start: Instant, startMarker: LogMarkerToken)
  *           negative for system operation and zero when originator is not known
  * @param start the timestamp when the request processing commenced
  */
-protected case class TransactionMetadata(val id: Long, val start: Instant)
+protected case class TransactionMetadata(val id: Long, val start: Instant, val extraLogging: Boolean = false)
 
 object TransactionId {
 
@@ -221,21 +217,22 @@ object TransactionId {
   val controller = TransactionId(-130) // Controller startup
   val dbBatcher = TransactionId(-140) // Database batcher
 
-  def apply(tid: BigDecimal): TransactionId = {
+  def apply(tid: BigDecimal, extraLogging: Boolean = false): TransactionId = {
     Try {
       val now = Instant.now(Clock.systemUTC())
-      TransactionId(TransactionMetadata(tid.toLong, now))
+      TransactionId(TransactionMetadata(tid.toLong, now, extraLogging))
     } getOrElse unknown
   }
 
   implicit val serdes = new RootJsonFormat[TransactionId] {
-    def write(t: TransactionId) = JsArray(JsNumber(t.meta.id), JsNumber(t.meta.start.toEpochMilli))
+    def write(t: TransactionId) =
+      JsArray(JsNumber(t.meta.id), JsNumber(t.meta.start.toEpochMilli), JsBoolean(t.meta.extraLogging))
 
     def read(value: JsValue) =
       Try {
         value match {
-          case JsArray(Vector(JsNumber(id), JsNumber(start))) =>
-            TransactionId(TransactionMetadata(id.longValue, Instant.ofEpochMilli(start.longValue)))
+          case JsArray(Vector(JsNumber(id), JsNumber(start), JsBoolean(extraLogging))) =>
+            TransactionId(TransactionMetadata(id.longValue, Instant.ofEpochMilli(start.longValue), extraLogging))
         }
       } getOrElse unknown
   }
@@ -254,7 +251,7 @@ trait TransactionCounter {
   // seed the counter so transids do not overlap: instanceOrdinal + n * stride, start at n = 1
   private lazy val cnt = new AtomicInteger(instanceOrdinal + stride)
 
-  def transid(): TransactionId = {
-    TransactionId(cnt.addAndGet(stride))
+  def transid(extraLogging: Boolean = false): TransactionId = {
+    TransactionId(cnt.addAndGet(stride), extraLogging)
   }
 }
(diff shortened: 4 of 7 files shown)


mhenke1:
@rabbah (as discussed off-line)
 The thread opened by this PR is that somebody would be able elevate the system log level for calls to OpenWhisk by passing the X-OW-EXTRA-LOGGING header to the controller and impede the performance of the system.
This performance hit would be restricted by the rate limiting mechanisms and the limited number 
of system level logs, but could be notable nevertheless. 

The risk of such an attack would be very low since it would require a change the configuration of 
nginx (which blocks passing the header) which could be easily mitigated by redeployment   of the original configuration.  

In the light of the risk assessment I would like to go forward with the PR as is.

ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -37,6 +37,20 @@ http {
...
+      "{{ user }}" special;
+{% endfor %}
+    }
+
+    map "$special_user:$http_x_ow_extra_logging" $extra_logging {
+      default "off";
+      "special:off" off;
+      "special:on" on;
+      "special:" on;
+      }
```
rabbah: indentation.

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -37,7 +37,11 @@ trait Logging {
    * @param message Message to write to the log
    */
   def debug(from: AnyRef, message: String)(implicit id: TransactionId = TransactionId.unknown) = {
-    emit(DebugLevel, id, from, message)
+    if (id.meta.extraLogging) {
```
rabbah: should this be applied to _all_ the logger methods not just debug perhaps by making the change in the `emit` method instead?

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -221,21 +217,29 @@ object TransactionId {
...
     Try {
       val now = Instant.now(Clock.systemUTC())
-      TransactionId(TransactionMetadata(tid.toLong, now))
+      TransactionId(TransactionMetadata(tid.toLong, now, extraLogging))
     } getOrElse unknown
   }
 
   implicit val serdes = new RootJsonFormat[TransactionId] {
-    def write(t: TransactionId) = JsArray(JsNumber(t.meta.id), JsNumber(t.meta.start.toEpochMilli))
+    def write(t: TransactionId) = {
```
rabbah: could you add a test (SchemaTests perhaps) to confirm the intended behavior for the serdes in the presence/absence of the additional field?

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
```diff
@@ -90,7 +91,13 @@ trait BasicHttpService extends Directives with TransactionCounter {
   }
 
   /** Assigns transaction id to every request. */
-  protected val assignId = extract(_ => transid())
+  protected def assignId = HeaderDirectives.optionalHeaderValueByName(OW_EXTRA_LOGGING_HEADER) flatMap { headerValue =>
```
rabbah: add a comment? e.g. `extract headers from HTTP request that indicates if additional logging should be enabled for this transaction. Allowed values are ...`

ansible/roles/nginx/templates/nginx.conf.j2
mhenke1: Fixed

common/scala/src/main/scala/whisk/common/Logging.scala
mhenke1: The only message we do suppress when running on INFO level are the ones on debug level. 
No need to generalize in my opinion.

rabbah: is that true for warnings?

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
mhenke1: I have added a comment explaining the header's purpose

common/scala/src/main/scala/whisk/common/Logging.scala
markusthoemmes: I second this. We should apply this to all methods (possibly do it just in `emit`?)! If we decide to run on WARN or on ERROR, you still want *all* logs to appear for those special users, no?

mhenke1:
I have added a test case to test serialization behavior

common/scala/src/main/scala/whisk/common/Logging.scala
mhenke1: @rabbah  WARNINGS will be shown when running on INFO level
@markusthoemmes Do you really think running OW on a level higher than INFO is a good idea with the current logging output ?

common/scala/src/main/scala/whisk/common/TransactionId.scala
mhenke1: I have added the test

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
rabbah: This header will be visible to web actions - should we strip it?

tests/src/test/scala/whisk/core/entity/test/SchemaTests.scala
```diff
@@ -70,6 +69,46 @@ class SchemaTests extends FlatSpec with BeforeAndAfter with ExecHelpers with Mat
...
   }
 
+  behavior of "TransactionId"
+
+  it should "serdes a transaction id without extraLogging parameter" in {
+    val txId = 4711
+    val txWithoutExtraLogging = TransactionId(txId)
+
+    // test serialization
+    val serializedTxIdWithoutParameter = TransactionId.serdes.write(txWithoutExtraLogging)
```
rabbah:
```scala
TransactionId.serdes.write(txWithoutExtraLogging) shouldBe JsArray(Vector(JsNumber(id), JsNumber(tid.meta.start.toEpochMilli))
```

tests/src/test/scala/whisk/core/entity/test/SchemaTests.scala
```diff
@@ -70,6 +69,46 @@ class SchemaTests extends FlatSpec with BeforeAndAfter with ExecHelpers with Mat
...
+    serializedTxIdWithoutParameter match {
+      case JsArray(Vector(JsNumber(id), JsNumber(_))) =>
+        assert(id == txId)
+      case JsArray(Vector(JsNumber(_), JsNumber(_), JsBoolean(_))) =>
```
rabbah: `case _ => assert(false)` since anything else is not expected?

tests/src/test/scala/whisk/core/entity/test/SchemaTests.scala
```diff
@@ -70,6 +69,46 @@ class SchemaTests extends FlatSpec with BeforeAndAfter with ExecHelpers with Mat
...
+  behavior of "TransactionId"
+
+  it should "serdes a transaction id without extraLogging parameter" in {
+    val txId = 4711
+    val txWithoutExtraLogging = TransactionId(txId)
+
+    // test serialization
+    val serializedTxIdWithoutParameter = TransactionId.serdes.write(txWithoutExtraLogging)
+    serializedTxIdWithoutParameter match {
+      case JsArray(Vector(JsNumber(id), JsNumber(_))) =>
```
rabbah: `id shouldBe txWithoutExtraLogging.id`?

tests/src/test/scala/whisk/core/entity/test/SchemaTests.scala
```diff
@@ -70,6 +69,46 @@ class SchemaTests extends FlatSpec with BeforeAndAfter with ExecHelpers with Mat
     a[DeserializationException] should be thrownBy Privilege.serdes.read("???".toJson)
   }
 
+  behavior of "TransactionId"
+
+  it should "serdes a transaction id without extraLogging parameter" in {
+    val txId = 4711
```
rabbah: the `id` is accessible via `TransactionId.id`

tests/src/test/scala/whisk/core/entity/test/SchemaTests.scala
```diff
@@ -70,6 +69,46 @@ class SchemaTests extends FlatSpec with BeforeAndAfter with ExecHelpers with Mat
...
+
+  it should "serdes a transaction id with extraLogging parameter" in {
+    val txId = 4711
+    val txIdWithExtraLogging = TransactionId(txId, true)
+    // test serialization
+    val serializedTxIdWithParameter = TransactionId.serdes.write(txIdWithExtraLogging)
+    serializedTxIdWithParameter match {
+      case JsArray(Vector(JsNumber(_), JsNumber(_))) =>
+        assert(false)
+      case JsArray(Vector(JsNumber(id), JsNumber(_), JsBoolean(extraLogging))) =>
```
rabbah: you could add a method to get the start time out of the tid and then this check becomes simpler:

`serializedTxIdWithParameter shouldBe JsArray(... actual values...)`

tests/src/test/scala/whisk/core/entity/test/SchemaTests.scala
mhenke1: Changed to use ...meta.id

tests/src/test/scala/whisk/core/entity/test/SchemaTests.scala
mhenke1: Changed to use ...meta.id

tests/src/test/scala/whisk/core/entity/test/SchemaTests.scala
mhenke1: I have now added that case

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
mhenke1: Good catch. The header will be removed.

mhenke1:
@rabbah We have a added a short documentation file and added issue #3191 to cover your wskadmin proposal.

docs/logging.md
```diff
@@ -0,0 +1,41 @@
+# Logging in OpenWhisk
+
+## Default Logging Provider
+
+OpenWhisk uses Logback as default logging provider via slf4j.
+
+Logback can be configured in the configuration file [logback.xml](../common/scala/src/main/resources/logback.xml).
+
+Besides other things this configuration file defines the default log level for OpenWhisk. 
+Akka Logging inherits the log level define here.
```
rabbah: define here -> defined here

tests/src/test/scala/whisk/core/entity/test/SchemaTests.scala
```diff
@@ -70,6 +69,47 @@ class SchemaTests extends FlatSpec with BeforeAndAfter with ExecHelpers with Mat
...
+
+  it should "serdes a transaction id without extraLogging parameter" in {
+    val txIdWithoutParameter = TransactionId(4711)
+
+    // test serialization
+    val serializedTxIdWithoutParameter = TransactionId.serdes.write(txIdWithoutParameter)
+    serializedTxIdWithoutParameter match {
+      case JsArray(Vector(JsNumber(id), JsNumber(_))) =>
+        assert(id == txIdWithoutParameter.meta.id)
+      case JsArray(Vector(JsNumber(_), JsNumber(_), JsBoolean(_))) =>
```
rabbah: You can drop this case.

tests/src/test/scala/whisk/core/entity/test/SchemaTests.scala
```diff
@@ -70,6 +69,47 @@ class SchemaTests extends FlatSpec with BeforeAndAfter with ExecHelpers with Mat
...
+
+  it should "serdes a transaction id with extraLogging parameter" in {
+    val txIdWithParameter = TransactionId(4711, true)
+
+    // test serialization
+    val serializedTxIdWithParameter = TransactionId.serdes.write(txIdWithParameter)
+    serializedTxIdWithParameter match {
+      case JsArray(Vector(JsNumber(_), JsNumber(_))) =>
+        assert(false)
+      case JsArray(Vector(JsNumber(id), JsNumber(_), JsBoolean(extraLogging))) =>
```
rabbah: Can drop the first case.  Cover by next line.

rabbah:
I wonder if this is a better place to drop the header https://github.com/apache/incubator-openwhisk/blob/master/core/controller/src/main/scala/whisk/core/controller/WebActions.scala

docs/logging.md
mhenke1: Fixed. Thx

tests/src/test/scala/whisk/core/entity/test/SchemaTests.scala
mhenke1: Done

tests/src/test/scala/whisk/core/entity/test/SchemaTests.scala
mhenke1: Done

mhenke1:
@rabbah I think header removal is better placed at the current location, because header handling and removal is near to each other in the same file and therefore more understandable

rabbah:
Linking https://github.com/apache/incubator-openwhisk/commit/59a7355d1336414cade36a4095360bf3a1342dff#r26909692 from @chetanmeh on suggestion for jmx/logback.

ansible/group_vars/all
```diff
@@ -161,6 +161,8 @@ nginx:
     verify_client: "{{ nginx_ssl_verify_client | default('off') }}"
   wpn:
     router: "{{ nginx_wpn_router | default('1') }}"
+  # an array of user ids to enable extra logging on the system
```
rabbah: an id you mean a namespace?

common/scala/src/main/scala/whisk/common/Logging.scala
rabbah: this still feels wrong to me - it presupposes that you're logging at least at InfoLevel - where the intent is to enable logging, period. So if the log level is off, this flag generates no additional output. Is that the intend?

docs/logging.md
```diff
@@ -0,0 +1,41 @@
...
+run -b ch.qos.logback.classic:Name=default,Type=ch.qos.logback.classic.jmx.JMXConfigurator setLoggerLevel ROOT DEBUG
+run -b ch.qos.logback.classic:Name=default,Type=ch.qos.logback.classic.jmx.JMXConfigurator getLoggerLevel ROOT
+close
+```
+
+2. Issue the command with the created file like this:
+```
+java -jar jmxterm-1.0.0-uber.jar -n silent < filename
+```
+
```
rabbah: kindly remove the extra spaces.

ansible/group_vars/all
mhenke1: I have added an example to the documentation file

docs/logging.md
mhenke1: Done

Issue #3176, opened 2018-01-11
Repeated throttle check for sequences


tardieu:
The entitlement check for reference entities in sequences (https://github.com/apache/incubator-openwhisk/blob/master/core/controller/src/main/scala/whisk/core/controller/Actions.scala#L382-L401) include checks against throttles (https://github.com/apache/incubator-openwhisk/blob/master/core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala#L221-L223) instead of simply checking for read/activate privileges. This result in the sequence being incorrectly accounted for twice (if non-empty).

tardieu:
I opened a PR with the proposed fix in #3177.

tardieu:
The PR has been merged. A test has been added.

PR #3177, opened 2018-01-11
Avoid repeated throttle check for sequences

tardieu:
Suggested fix for #3176.

The entitlement check for reference entities in sequences (https://github.com/apache/incubator-openwhisk/blob/master/core/controller/src/main/scala/whisk/core/controller/Actions.scala#L382-L401) include checks against throttles (https://github.com/apache/incubator-openwhisk/blob/master/core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala#L221-L223) instead of simply checking for read/activate privileges. This result in the sequence being incorrectly accounted for twice (if non-empty).

This patch adds the ability to check privileges without impacting throttles and uses the new option for the checking referenced entities in sequences.

The added flexibility to the `Entitlement` `check` method is also necessary to correctly entitle action compositions in composer (https://github.com/tardieu/composer/tree/hybrid).

### DIFF
--- a/tests/src/test/scala/limits/ThrottleTests.scala
+++ b/tests/src/test/scala/limits/ThrottleTests.scala
@@ -327,8 +327,12 @@ class NamespaceSpecificThrottleTests
   val oneProps = getAdditionalTestSubject("oneSubject")
   wskadmin.cli(Seq("limits", "set", oneProps.namespace, "--invocationsPerMinute", "1", "--firesPerMinute", "1"))
 
+  // Create a subject where the rate limits are set to 1 for testing sequences
+  val oneSequenceProps = getAdditionalTestSubject("oneSequenceSubject")
+  wskadmin.cli(Seq("limits", "set", oneSequenceProps.namespace, "--invocationsPerMinute", "1", "--firesPerMinute", "1"))
+
   override def afterAll() = {
-    Seq(zeroProps, zeroConcProps, oneProps).foreach { wp =>
+    Seq(zeroProps, zeroConcProps, oneProps, oneSequenceProps).foreach { wp =>
       val ns = wp.namespace
       disposeAdditionalTestSubject(ns)
       withClue(s"failed to delete temporary limits for $ns") {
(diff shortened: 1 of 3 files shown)


markusthoemmes:
Looks sensible, adding a test would be good here. Can you do that?

tardieu:
I added a test (which fails without the patch).

tests/src/test/scala/limits/ThrottleTests.scala
```diff
@@ -327,8 +327,12 @@ class NamespaceSpecificThrottleTests
   val oneProps = getAdditionalTestSubject("oneSubject")
   wskadmin.cli(Seq("limits", "set", oneProps.namespace, "--invocationsPerMinute", "1", "--firesPerMinute", "1"))
 
+  // Create a subject where the rate limits are set to 1 for testing sequences
+  val oneSequenceProps = getAdditionalTestSubject("oneSequenceSubject")
```
rabbah: do you need an additional test subject vs reusing one `oneProps` subject above?

tardieu: Yes or I have to wait 1 minute.

tests/src/test/scala/limits/ThrottleTests.scala
```diff
@@ -400,6 +404,44 @@ class NamespaceSpecificThrottleTests
...
+      assetHelper.withCleaner(wsk.action, sequenceName) { (action, _) =>
+        action.create(sequenceName, Some(s"$actionName,$actionName"), kind = Some("sequence"))
+      }
+
+      val deployedControllers = WhiskProperties.getControllerHosts.split(",").length
+
+      // One invoke should be allowed.
+      wsk.action
+        .invoke(sequenceName, expectedExitCode = TestUtils.DONTCARE_EXIT)
+        .exitCode shouldBe TestUtils.SUCCESS_EXIT
```
rabbah: `.invoke(sequenceName, expectedExitCode = TestUtils.SUCCESS_EXIT)` would eschew the next check (`.exitCode shouldBe ...`)

tests/src/test/scala/limits/ThrottleTests.scala
rabbah: Ah - sneaky. Thanks.

rabbah: fwiw you only need the invocations per minute for the test.

rabbah: perhaps a comment :)

tests/src/test/scala/limits/ThrottleTests.scala
tardieu: It did not work for me. `expectedExitCode` is `ACCEPTED` and not `SUCCESS`. I decided to follow the pattern of the other tests, rather than improvising.

Issue #3178, opened 2018-01-11
Gradle tests updating the runtime dependencies


houshengbo:
When we run the tests by the command: ./gradlew tests:test, gradle will verify and update the dependencies. We should not do this, since tests would not make any sense, if they are able to modify our deployment.

https://github.com/apache/incubator-openwhisk/blob/master/tests/build.gradle#L41:
```
test.dependsOn([
    ':actionRuntimes:nodejs6Action:distDocker',
    ':actionRuntimes:actionProxy:distDocker',
    ':actionRuntimes:pythonAction:distDocker',
    ':actionRuntimes:python2Action:distDocker',
    ':actionRuntimes:javaAction:distDocker',
    ':actionRuntimes:swift3.1.1Action:distDocker',
    ':sdk:docker:distDocker',
    ':tests:dat:blackbox:badaction:distDocker',
    ':tests:dat:blackbox:badproxy:distDocker'
])
```

markusthoemmes:
I take it that no tests in the main repo are relying on the existence of those images anymore? Note that the tests do not modify the deployment but rather rely on those images being around (the ActionContainerTests for instance). Can you make sure, that all of those tests have been moved out to different repos?

rabbah:
https://github.com/apache/incubator-openwhisk/tree/master/tests/src/test/scala/actionContainers We still have all the tests in this repo and we run them.

rabbah:
(Iirc all of the tests and related skeleton have been cloned and owned by the runtime repos 😕)

PR #3179, opened 2018-01-11
Remove the dependencies when running tests

houshengbo:
Closes: #3178

### DIFF
--- a/tests/build.gradle
+++ b/tests/build.gradle
@@ -37,19 +37,6 @@ task testLeanCli(type: Test) {
     exclude '**/*Rest*'
 }
 
-// Add all images needed for local testing here
-test.dependsOn([
-    ':actionRuntimes:nodejs6Action:distDocker',
-    ':actionRuntimes:actionProxy:distDocker',
-    ':actionRuntimes:pythonAction:distDocker',
-    ':actionRuntimes:python2Action:distDocker',
-    ':actionRuntimes:javaAction:distDocker',
-    ':actionRuntimes:swift3.1.1Action:distDocker',
-    ':sdk:docker:distDocker',
-    ':tests:dat:blackbox:badaction:distDocker',
-    ':tests:dat:blackbox:badproxy:distDocker'
-])
-
 dependencies {
     compile "org.scala-lang:scala-library:${gradle.scala.version}"
     compile 'org.apache.commons:commons-lang3:3.3.2'


jonpspri:
Ooo!  I vote yes, because this was also mucking up the CLI tests.

rabbah:
We’re making harder and harder to develop across the entire stack. These dependencies made sure the container images were fresh before running the container unit tests. Rather than remove we should provide ways to bypass.

jonpspri:
Well, to make it all work in what I'm doing with 'incubator-openwhisk-cli', I had to add this stanza to 'settings.gradle'.  Maybe we should make this available in 'incubator-openwhisk' as a file to `apply`?  I'm on sleep deprivation, so there could be many reasons that NOT a good idea...

```
def owPath = System.getenv("OPENWHISK_HOME") ?: '../incubator-openwhisk'
def owDirectory = new File(owPath)

if (!owDirectory.exists()) {
    throw new GradleScriptException(
        "Environment variable OPENWHISK_HOME must point to a valid OpenWhisk build", null)
}

include 'clitests'; project(':clitests').projectDir = new File(owDirectory, 'tests')

[   'common:scala',
    'core:controller',
    'core:invoker',
    'actionRuntimes:nodejs6Action',
    'actionRuntimes:actionProxy',
    'actionRuntimes:pythonAction',
    'actionRuntimes:python2Action',
    'actionRuntimes:javaAction',
    'actionRuntimes:swift3.1.1Action',
    'sdk:docker',
    'tests:dat:blackbox:badaction',
    'tests:dat:blackbox:badproxy'
].each() { p ->
    include p
    project(':'+p).projectDir = new File(owDirectory, p.replaceAll(':','/'))
}

include 'tests'
```

csantanapr:
I'm OK removing since settings.gradle cover this projects, and when user builds which is before running any tests, he will build with `./gradlew distDocker` and the images will be there to be use by the tests.

markusthoemmes:
The tests rely on those images so the tests should have a dependency on them, no?

You can always create a specific test-target which excludes those dependencies, which I'd prefer here.

csantanapr:
That’s also good to create a new target for environments were a specific image tag want to be use and avoid the rebuild and trashed on what’s already prepared. Like a lean target that skips this dependencies assuming that deos are already taken care. 

The use case is for the CI/CD of Apache release automation where a specific tag of the runtimes would be use and build before running the tests.

markusthoemmes:
@houshengbo comments on my concerns above?

markusthoemmes:
Closing as abandonend. Please ping me to reopen if you want.

PR #3187, opened 2018-01-15
reduce rule activation records

mdeuser:
- no longer create/persist a rule activation record when trigger fires
- include the async action invocation response info in the trigger activation log
- adjust rule tests

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
@@ -413,12 +413,19 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
                          actionActivationId: Option[JsValue] = None,
                          errorMsg: Option[String] = None): JsObject = {
     JsObject(
-      "rule" -> JsString(ruleName.asString),
-      "action" -> JsString(actionName.asString),
-      "statusCode" -> JsNumber(statusCode),
-      "success" -> JsBoolean(statusCode == ActivationResponse.Success),
-      "activationId" -> actionActivationId.toJson,
-      "error" -> errorMsg.toJson)
+      Map(
+        "rule" -> JsString(ruleName.asString),
+        "action" -> JsString(actionName.asString),
+        "statusCode" -> JsNumber(statusCode),
+        "success" -> JsBoolean(statusCode == ActivationResponse.Success)) ++ {
+        actionActivationId map { id =>
+          Seq("activationId" -> id.toJson)
+        } getOrElse Seq()
+      } ++ {
+        errorMsg map { err =>
+          Seq("error" -> JsString(err))
+        } getOrElse Seq()
+      })
   }
 
   /** Common base bath for the controller, used by internal action activation mechanism. */
--- a/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
@@ -19,18 +19,22 @@ package whisk.core.controller
 
 import java.time.{Clock, Instant}
 
+import scala.collection.immutable.Map
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
-import akka.http.scaladsl.model.ContentTypes
 import akka.http.scaladsl.model.HttpMethods.POST
-import akka.http.scaladsl.model.StatusCodes._
-import akka.http.scaladsl.model.{HttpEntity, HttpRequest, MediaTypes, Uri}
+import akka.http.scaladsl.model.StatusCodes.{Accepted, BadRequest, InternalServerError, OK}
 import akka.http.scaladsl.model.Uri.Path
 import akka.http.scaladsl.model.headers.{Authorization, BasicHttpCredentials}
+import akka.http.scaladsl.model._
 import akka.http.scaladsl.server.{RequestContext, RouteResult}
 import akka.http.scaladsl.unmarshalling.{Unmarshal, Unmarshaller}
 import akka.stream.ActorMaterializer
+import spray.json.DefaultJsonProtocol._
 import spray.json._
 import whisk.common.TransactionId
 import whisk.core.controller.RestApiCommons.ListLimit
@@ -39,9 +43,6 @@ import whisk.core.entitlement.Collection
 import whisk.core.entity._
 import whisk.core.entity.types.{ActivationStore, EntityStore}
 import whisk.http.ErrorResponse
-
-import scala.collection.immutable.Map
-import scala.concurrent.Future
 
 /** A trait implementing the triggers API. */
 trait WhiskTriggersApi extends WhiskCollectionAPI {
@@ -62,8 +63,6 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
   protected val activationStore: ActivationStore
 
   /** JSON response formatter. */
-  import RestApiCommons.jsonDefaultResponsePrinter
-
   /** Path to Triggers REST API. */
   protected val triggersPath = "triggers"
   protected val url = Uri(s"http://localhost:${whiskConfig.servicePort}")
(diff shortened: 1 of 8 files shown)


core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -122,72 +126,135 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
-                        Unmarshal(response.entity).to[String].map { error =>
-                          logging.warn(this, s"${rule.action} failed due to $error")
+          // List of active rules associated with the trigger
+          val activeRules = trigger.rules map { _.filter((r) => r._2.status == Status.ACTIVE) } getOrElse Map.empty
+
+          if (activeRules.nonEmpty) {
+            val args = trigger.parameters.merge(payload)
+
+            // Iterate through each active rule; invoking each mapped action
+            val actionLogList = activeRules
```
rabbah: i suggest factoring this long method into at least three: mapping over the active rules, activating the action, and gathering the activation ids/updating the activation record.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -122,72 +126,135 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
+                  val actionNamespace = rule.action.path.root.asString
+                  val actionPath = {
+                    rule.action.path.relativePath.map { pkg =>
+                      (Path.SingleSlash + pkg.namespace) / rule.action.name.asString
+                    } getOrElse {
+                      Path.SingleSlash + rule.action.name.asString
+                    }
+                  }.toString
+
+                  val actionUrl = Path("/api/v1") / "namespaces" / actionNamespace / "actions"
```
rabbah: `Path("/api/v1") / "namespaces"` can be part of the base url path.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -122,72 +126,135 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
+                    .flatMap {
+                      response =>
+                        response.status match {
+                          case OK | Accepted =>
+                            Unmarshal(response.entity).to[JsObject].map { a =>
+                              logging.info(
+                                this,
+                                s"trigger-fired rule '${rule.action}' invoked with activation ${a.fields("activationId")}")
+                              (ruleName.asString -> triggerActivationLogMsg(
+                                entityName.asString,
```
rabbah: what's the difference between entity name and rule.action?

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -122,72 +126,135 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
+                                ruleName.asString,
+                                rule.action.asString,
+                                InfoLevel,
```
rabbah: what is the log level needed for?

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -122,72 +126,135 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
+                                s"'${rule.action}' activated ${a.fields("activationId")}"))
```
rabbah: this contains redundant information (rule.action already passed above).
why not just ass the activation id?

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -321,6 +388,33 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
+  /**
+   *
+   * @param triggerName
+   * @param ruleName
+   * @param actionName
+   * @param loglevel
+   * @param msg
+   * @return
+   */
+  private def triggerActivationLogMsg(triggerName: String,
```
rabbah: what's the point of this method?

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
rabbah: i see this is the return value of the future.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
mdeuser: this method creates the trigger activation log entry string.  if we need the format to change, it's centralized here.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
mdeuser: the `entityName` refers to the trigger name while the `rule.action` refers to the mapped action that was just invoked.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
mdeuser: i was thinking it would be good to flag invocation failures with a tag that can eventually be consumed by a logging service.  here's some sample output
```
  "logs": [
      "[2018-01-16T16:23:17.174Z] [INFO] [guest/wtrig] [guest/wrule] [guest/weather] 'guest/weather' activated \"3c2aaf37e07549cbaaaf37e07519cb19\"",
      "[2018-01-16T16:23:17.175Z] [INFO] [guest/wtrig] [guest/wrule2] [guest/echo-env-web] 'guest/echo-env-web' activated \"8e2b7c0f04084c53ab7c0f04087c5386\"",
      "[2018-01-16T16:23:16.948Z] [ERROR] [guest/wtrig] [guest/wrule3] [guest/badrule] 'guest/badrule' failed, action not found"
  ],
```

rabbah: thanks for the clarification.
we'll want this to be something easily consumable; @starpit what do think?
we can use a fixed format like this, or perhaps a json representation logged to a single line.

starpit: trigger activations will still have activation records, but rules will not? and if, e.g. in this case, the rule's action is not found, there will be a log entry in that trigger fire's activation record?  or are these separate log records?

mdeuser: with this pr, only two kinds of activation records will be created - one for the trigger and one for each associated action.  the rule activation will no longer be created.  the trigger activation log entries shown above are meant to assist in tracking down the action results that were generated in response to the trigger event.

starpit: thanks for the help in my understanding. at first glance, this seems like information better suited for a statusCode?

rabbah: Yes. But here are some caveats. We can’t use the trigger activation result because it’s already used. It records the triggers incoming payload. 

There are multiple results, one for each action that’s activated. 

I do think we can use the activation response schema to record the result of its action that’s activated. And store these in an array. The question is do we log these as log lines as in the current form, use annotations or use a single log line that is formatted as a json string.

starpit: "single log line ... as a json string". is the suggestion that we change the entire log record to a JsonFormat structure, or that we have a hybrid, with "what we have now" using the current positional encoding, and then subrecords formatted as JSON?

there is some nascent desire for allowing the actions themselves to communicate structured information to the user; e.g. internal timing information. so this is a somewhat broader topic.

rabbah: I'm strictly speaking about the trigger activation, not action logs.
The schema doesn't constrain log lines. It's a string. That we generate a `[timestamp] [stream] text` format today is a convenience generated by the invoker.

I'm suggesting that the timestamp and stream are meaningless here. So it doesn't make sense to follow these conventions. Instead, a single JSON string of the action post results is more appropriate in my view.

How logs generate their log lines, and how those are parsed, indexed, and enriched is a separate issue entirely in my opinion.

mdeuser: @csantanapr and i talked about this format a bit and we liked the idea of maintaining the log format of existing action invocation logs - iff we want to use the activation records `logs` field.  potentially a logging service, like logmet, could consume these and make trigger/action search/correlation easier.

mdeuser:
pg3/1695

mdeuser:
bumping travis

mdeuser:
good travis.  pg4/1260

mdeuser:
with these changes, the trigger activation record looks like
```
ok: got activation c40f435761cd49ab8f435761cdd9ab9d
{
    "namespace": "guest",
    "name": "wtrig",
    "version": "0.0.1",
    "subject": "guest",
    "activationId": "c40f435761cd49ab8f435761cdd9ab9d",
    "start": 1516483143448,
    "end": 0,
    "duration": 0,
    "response": {
        "status": "success",
        "statusCode": 0,
        "success": true,
        "result": {}
    },
    "logs": [
        "{\"statusCode\":0,\"ruleName\":\"guest/wrule\",\"success\":true,\"activationId\":\"68ddc2d06ec74ccf9dc2d06ec7dccfa0\",\"actionName\":\"guest/weather\"}",
        "{\"statusCode\":0,\"ruleName\":\"guest/wrule2\",\"success\":true,\"activationId\":\"948813595efa4e058813595efafe05c5\",\"actionName\":\"guest/echo-env-web\"}",
        "{\"statusCode\":1,\"ruleName\":\"guest/wrule3\",\"success\":false,\"error\":\"The requested resource does not exist.\",\"actionName\":\"guest/badrule\"}"
    ],
    "annotations": [],
    "publish": false
}
```

mdeuser:
@dubeejw - the `logs` field is defined as Vector[String], so any JSON must be converted to JSON strings

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
mdeuser: just to wrap up this topic.  @csantanapr, @rabbah and i spoke and decided to write out an array of json strings instead of the above log entries.  each json string would be a single json object representing the status/result of the async rule/action activation.  for example
```d
    "logs": [
        "{\"statusCode\":0,\"success\":true,\"activationId\":\"ae2d98c6b49b4a32ad98c6b49bba32c5\",\"rule\":\"guest/wrule\",\"action\":\"guest/echo-env-web\"}",
        "{\"statusCode\":0,\"success\":true,\"activationId\":\"32e3afff8a7f4c31a3afff8a7fbc319b\",\"rule\":\"guest/wrule2\",\"action\":\"guest/echo-env-web\"}",
        "{\"statusCode\":1,\"success\":false,\"rule\":\"guest/wrule3\",\"error\":\"The requested resource does not exist.\",\"action\":\"guest/nonexistent-action\"}"
    ]
```

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -390,46 +300,130 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
-    //   "rule": "my-rule",
-    //   "action": "my-action",
-    //   "statusCode": 0,
-    //   "status": "success",
-    //   "activationId": "90c84e7b33c84ceb884e7b33c8ecebf6",  // Optional
-    //   "error": "The requested resource does not exist."  // Optional
-    // }
-    JsObject(objMap)
+      "success" -> JsBoolean(statusCode == ActivationResponse.Success),
+      "activationId" -> actionActivationId.toJson,
```
rabbah: note: the unit tests (or integration tests) are not tight enough and don't enforce a schema on the response; the activationId and error properties here will emit as `null` if not present which changes the behavior relative to the previous commit.

I suggest a unit test which checks the expected schema exactly; this can be done for the intergration test as well; parse the array to json and check for equality.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -413,12 +413,19 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
-      "statusCode" -> JsNumber(statusCode),
-      "success" -> JsBoolean(statusCode == ActivationResponse.Success),
-      "activationId" -> actionActivationId.toJson,
-      "error" -> errorMsg.toJson)
+      Map(
+        "rule" -> JsString(ruleName.asString),
+        "action" -> JsString(actionName.asString),
+        "statusCode" -> JsNumber(statusCode),
+        "success" -> JsBoolean(statusCode == ActivationResponse.Success)) ++ {
+        actionActivationId map { id =>
```
rabbah: you can simplify this to:
```scala
    JsObject(
      Map(
        "rule" -> JsString(ruleName.asString),
        "action" -> JsString(actionName.asString),
        "statusCode" -> JsNumber(statusCode),
        "success" -> JsBoolean(statusCode == ActivationResponse.Success)) ++
        actionActivationId.map("activationId" -> _.toJson) ++
        errorMsg.map("error" -> JsString(_)))
```

rabbah:
rebase to drop second commit?

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
mdeuser: so many ways in scala to `map` :-)

mdeuser:
can i assume that since the `reduce fillInStackTrace expenses` commit is already in the master repo, that this same commit being in this PR will be ignored during merge?

rabbah:
I’d think so but if you rebased to master already and it’s still there then something else is off.

rabbah:
A note on the API test - I had missed an important detail. The actions are activated via the REST API. So this will not work with the unit test today which permits intercepting action activations but only through the internal mechanism. So in effect all actions will fail here if we tried. We should fix this as we move toward the internal path.

PR #3188, opened 2018-01-15
use /var/tmp on Ubuntu to prevent configuration files changing owner when rebooting

### LINKED ISSUE #32: nginx deployment fails if mount source is wiped
Observed this failure a few times and here's how to reproduce it.

```
sudo rm -rf /tmp/nginx
ant clean build deploy
ant teardown
sudo rm -rf /tmp/nginx # note that if vm reboots or on a vagrant reload, this directory is gone
ant deploy # would normally work without a rebuild modulo the nginx config
```

Now /tmp/nginx is owned by root instead of the user (e.g., vagrant); nginx deployment appears to work and the failure is detected much later because the API host will not respond. The nginx container during deployment comes up and responds to heartbeat before it detects a bad configuration (and dies).

The nginx deployment needs to assert the configuration directory exists and fail if it does not. This will prevent mounting a non-existent directory and making it root-owned. Should also consider using an alternate directory instead of /tmp.

Lastly, the nginx build needs to fail if it cannot copy the nginx conf to the target directory (which will now happen since that directory is owned by root and the user has to write permission).

sciabarracom:
I think I figure out what goes wrong when you reboot the Vagrant machine.
Files in /tmp change owner, probably because of some startup initialization in Ubuntu.
Since it does not happen using for example /var/tmp my patch is simply to change directory for those files. However since it can break Mac deployment I added a test to do it only for Ubunt. 
With this patch you can do vagrant reload safely or not worry about shutting down the vm.
It may also close #32

### DIFF
--- a/ansible/environments/local/group_vars/all
+++ b/ansible/environments/local/group_vars/all
@@ -1,6 +1,6 @@
 whisk_version_name: local
-config_root_dir: /tmp/wskconf
-whisk_logs_dir: /tmp/wsklogs
+config_root_dir: "{{ '/var/tmp' if ansible_distribution == 'Ubuntu' else '/tmp' }}/wskconf"
+whisk_logs_dir: "{{ '/var/tmp' if ansible_distribution == 'Ubuntu' else '/tmp' }}/wsklogs"
 docker_registry: ""
 docker_dns: ""
 bypass_pull_for_local_images: true


rabbah:
I think there's a prize for closing a two digit issue number!

csantanapr:
This is fix now, I moved the location of the mount point out of `/tmp`
https://github.com/apache/incubator-openwhisk/pull/3269

PR #3189, opened 2018-01-16
Make some CLI tests more stable.

cbickel:
The test i modified failed in a pipeline.

I think, that the problem was, that the cache invalidation was not fast enough. To not have this problem again and to improve debugging, I added some more information and a retry.

@dubeejw FYI

PG5#20 🏃

### DIFF
--- a/tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
+++ b/tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
@@ -17,36 +17,32 @@
 
 package whisk.core.cli.test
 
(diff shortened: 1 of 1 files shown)


rabbah:
Perhaps for convenience, the get method should accept an optional semver and internally retry until it gets is (assuming there’s a semver change say 0.0.1 to 0.0.2 because of an intervening update). More generally I think CRUD operations should have some namespace stickiness but requires an nginx change. 

Do you think we’re going to hit more of this?

markusthoemmes:
@cbickel the testcase is fixed in the meantime I think. Please open a new PR if you want to fix the test-action names still.

PR #3190, opened 2018-01-16
Code clean up

dubeejw:


tests/src/test/scala/system/rest/GoCLINginxTests.scala
```diff
@@ -32,10 +32,7 @@ import DefaultJsonProtocol._
  */
 @RunWith(classOf[JUnitRunner])
 class GoCLINginxTests extends FlatSpec with Matchers with RestUtil {
-  val ProductName = "OpenWhisk_CLI"
   val DownloadLinkGoCli = "cli/go/download"
-  val OperatingSystems = List("mac", "linux", "windows")
-  val Architectures = List("386", "amd64")
```
rabbah: Is this suite moving out?

dubeejw: content.json is intended to be a index file for all the downloadable CLIs hosted by nginx. This test is a quick way to verify that all the expected downloadables are actually hosted by the OpenWhisk backend. As long as we use OpenWhisk to host the downloadable CLIs, I think the suite has to stay.

PR #3193, opened 2018-01-17
No automatic topic creation in Kafka

sven-lange-last:
So far, we had automatic topic creation - the broker would automatically create a topic if a producer sent a message to a topic or a consumer tried to receive messages from a topic. In that case, topic would be created automatically with default settings. These default settings are not always what we need. In the end, this is a race condition during deployment. If we manage to create a topic with desired settings before it gets auto-created with default settings, we are fine. Otherwise, we may see undesired effects.

This change disables automatic topic creation.

### DIFF
--- a/ansible/roles/kafka/tasks/deploy.yml
+++ b/ansible/roles/kafka/tasks/deploy.yml
@@ -23,7 +23,7 @@
       "KAFKA_HEAP_OPTS": "-Xmx{{ kafka.heap }} -Xms{{ kafka.heap }}"
       "KAFKA_ZOOKEEPER_CONNECT": "{{ zookeeper_connect_string }}"
       "KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR": "{{ kafka.replicationFactor }}"
-      "KAFKA_AUTO_CREATE_TOPICS_ENABLE": "{{ kafka.autoCreateTopics }}"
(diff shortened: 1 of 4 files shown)


sven-lange-last:
PG2 / 2668 running.

@markusthoemmes Please attend to this PR.

ansible/roles/kafka/tasks/deploy.yml
```diff
@@ -23,6 +23,7 @@
       "KAFKA_HEAP_OPTS": "-Xmx{{ kafka.heap }} -Xms{{ kafka.heap }}"
       "KAFKA_ZOOKEEPER_CONNECT": "{{ zookeeper_connect_string }}"
       "KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR": "{{ kafka.replicationFactor }}"
+      "KAFKA_AUTO_CREATE_TOPICS_ENABLE": "{{ kafka.autoCreateTopics }}"
```
markusthoemmes: Honestly, I'd take the configuration out. We don't want auto-creation I think.

sven-lange-last: That's fair. I'm removing the code to make this configurable and will simply switch automatic topic creation off. The default is `off` anyway and no Ansible environment has it `on`.

sven-lange-last:
PG2 / 2669 🔵

PR #3196, opened 2018-01-17
Fixes #3195 - create config object for docker in the invoker and get …

### LINKED ISSUE #3195: When running Invoker on "Bare-Metal" it cannot read log files
* Using docker-compose as deployment
* Running the invoker directly in the OS ( ubuntu in this case )

1.   Deploy openwhisk with docker-compose
2.   Build the invoker
3.   Stop invoker in compose
4.   Start built invoker as root user ( needed to actually read the log files )

Open Whisk cannot read the log files as /containers is hardcoded into the invoker's docker container pool:
https://github.com/apache/incubator-openwhisk/blob/a6a782f7d3a409c542c157e4be46074ef69a67ce/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala#L43
and
https://github.com/apache/incubator-openwhisk/blob/a6a782f7d3a409c542c157e4be46074ef69a67ce/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala#L41-L42

* There is also a swallowed exception that happens if you run the Invoker as a non-root user, still trying to track that one down.

mcdan:
…the rootpath from it.


core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
```diff
@@ -17,30 +17,35 @@
...
+import scala.concurrent.Await
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
 
 class DockerContainerFactory(config: WhiskConfig, instance: InstanceId, parameters: Map[String, Set[String]])(
   implicit actorSystem: ActorSystem,
   ec: ExecutionContext,
   logging: Logging)
     extends ContainerFactory {
+  private val dockerConfig = pureconfig.loadConfigOrThrow[DockerContainerFactoryConfig](ConfigKeys.invokerDocker)
```
mcdan: Would like some advice on if this is the best pattern for the pureconfig setup. I think the DockerContainter factory is probably the right spot for reading some config but not sure how much I should be pulling into the case class and if the "key" is appropriate in WhiskConfig.scala.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -254,4 +254,8 @@ object ConfigKeys {
 
   val transactions = "whisk.transactions"
   val stride = s"$transactions.stride"
+  val invokerDocker = "whisk.docker"
```
tysonnorris: this is duplicating https://github.com/apache/incubator-openwhisk/blob/master/common/scala/src/main/scala/whisk/core/WhiskConfig.scala#L250

mcdan: yup.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
```diff
@@ -129,3 +134,5 @@ object DockerContainerFactoryProvider extends ContainerFactoryProvider {
                                    parameters: Map[String, Set[String]]): ContainerFactory =
     new DockerContainerFactory(config, instanceId, parameters)(actorSystem, actorSystem.dispatcher, logging)
 }
+
+case class DockerContainerFactoryConfig(rootpath: Option[String])
```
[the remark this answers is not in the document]
mcdan: renamed.

common/scala/src/main/scala/whisk/core/WhiskConfig.scala
```diff
@@ -254,4 +254,8 @@ object ConfigKeys {
...
+
+  val invokerDockerRootPath = s"$invokerDocker.rootpath"
```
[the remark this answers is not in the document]
mcdan: No, because I ended up going with the case class.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
mcdan: It seems at least that the above example works just fine -- meaning the case class just ignores "extra" parameters, quick test proves this out:

```scala
import com.typesafe.config.ConfigFactory
import pureconfig._
import com.typesafe.config.ConfigFactory
case class Holiday(where: String)
loadConfig[Holiday](ConfigFactory.parseString("{ where: cabo }")) // Works
loadConfig[Holiday](ConfigFactory.parseString("{ where: cabo, why: hagar }")) //Works just "cabo" available
loadConfig[Holiday](ConfigFactory.parseString("{ why: hagar }")) // "fails" because 'where' not defined.
```

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
```diff
@@ -129,3 +135,15 @@ object DockerContainerFactoryProvider extends ContainerFactoryProvider {
...
+case class DockerContainerFactoryConfig(dataroot: Option[String]) {
+  def containerPath: File =
+    Paths
+      .get({
+        if (dataroot.isEmpty)
+          "/containers"
+        else
+          dataroot.get + "/containers"
+      })
+      .toFile
```
markusthoemmes: Some Scala goodness: You can use `Option`s methods to transform this in a statically proven way.

```scala 
val containerPath: File = Paths.get(dataRoot.getOrElse("") + "/containers")
```

Also, I'd use a `val` since this is used quite often in production code and doesn't need to be recomputed.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -380,18 +382,25 @@ class ContainerProxy(
...
-          }
-          .map(logs => Right(activation.withLogs(logs)))
-          .recover {
-            case LogCollectingException(logs) =>
-              Left(ActivationLogReadingError(activation.withLogs(logs)))
-            case _ =>
-              Left(ActivationLogReadingError(activation.withLogs(ActivationLogs(Vector(Messages.logFailure)))))
-          }
+        Try(
+          collectLogs(tid, job.msg.user, activation, container, job.action)
```
markusthoemmes: Where is the exception thrown? Can we make `collectLogs` not throw an exception but rather return a `Future.failed` in that case?

If not: We should only wrap `collectLogs` into a `Try` or just use a `try/catch` here to not box unnecessary.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -380,18 +382,25 @@ class ContainerProxy(
...
+            .andThen {
+              case Success(_) => tid.finished(this, start)
+              case Failure(t) => tid.failed(this, start, s"reading logs failed: $t")
+            }
+            .map(logs => Right(activation.withLogs(logs)))
+            .recover {
+              case LogCollectingException(logs) =>
+                Left(ActivationLogReadingError(activation.withLogs(logs)))
+              case _ =>
+                Left(ActivationLogReadingError(activation.withLogs(ActivationLogs(Vector(Messages.logFailure)))))
```
tysonnorris: my scala game is not great, but can you remove either the `.recover` or the `case Failure(t)` below? i.e. instead of using Try/Success/Failure, can you just add `tid.failed(this, start, s"reading logs failed: $t")` to the previously existing `.recover case _`?

tysonnorris: @markusthoemmes can you comment on this when you have a sec?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
mcdan: Ah yes, should have seen that sleight of hand prepending an empty string, dang.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
mcdan: Sorry I'm a total novice with Akka so not sure how to proceed, the error i think is being thrown when Akka tries to "FileTailSource" https://github.com/apache/incubator-openwhisk/blob/85bb3fe38642076646954a8ed1243b76d7f0fee6/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala#L150
This file isn't readable as the user may not have permissions so it blows up. Still trying to figure out the right way to recover from the error.

chetanmeh: > Where is the exception thrown? Can we make collectLogs not throw an exception but rather return a Future.failed in that case?

@markusthoemmes  Problem here is that `collectLogs` is provided from outside. So we would need to rely on it returning a Future.failed. More details are provided [here][1]

[1]: https://github.com/apache/incubator-openwhisk/issues/3195#issuecomment-358734535

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
markusthoemmes: Hmm, no you can't. These errors are related to the inner `Future` while the block below catches the errors from the outer `Try`. I'll try to code something up to fold them both into one.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
markusthoemmes: @chetanmeh got it, thanks for the clarification.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
markusthoemmes: Here it comes, I propose the following diff for this file to narrow down the exception handling:

```diff
diff --git a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
index 23d09651..93399069 100644
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
@@ -380,7 +380,11 @@ class ContainerProxy(
     val activationWithLogs: Future[Either[ActivationLogReadingError, WhiskActivation]] = activation
       .flatMap { activation =>
         val start = tid.started(this, LoggingMarkers.INVOKER_COLLECT_LOGS)
-        collectLogs(tid, job.msg.user, activation, container, job.action)
+        val collected: Future[ActivationLogs] = try {
+          collectLogs(tid, job.msg.user, activation, container, job.action)
+        } catch { case t: Throwable => Future.failed(t) }
+
+        collected
           .andThen {
             case Success(_) => tid.finished(this, start)
             case Failure(t) => tid.failed(this, start, s"reading logs failed: $t")

```

It transforms a failure in `collectLogs` to a failed Future, so we can process it as is. WDYT?

chetanmeh: Much simpler! I was trying to unnecessarily use `Try` where good old `try` would have been better

markusthoemmes:
@mcdan What's the status of this? Most comments seem unaddressed and there are merge conflicts.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
markusthoemmes: You can always change to `catch { case NonFatal(t) => Future.failed(t) }`

mcdan:
@markusthoemmes sorry for the delay, was waiting on some of the other config stuff to settle down a bit to make sure I was putting the directory in the right place. This is all up to date and should address all concerns.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -261,6 +265,8 @@ class DockerContainer(protected val id: ContainerId,
           // FramingException can also mean exceeding the limits, we cannot decide which case happened so we resort
           // to the general error message. This will be the last element of the stream.
           ByteString(LogLine(Instant.now.toString, "stderr", Messages.logFailure).toJson.compactPrint)
+        case _: RuntimeException =>
+          ByteString(LogLine(Instant.now.toString, "stderr", "Ouch.").toJson.compactPrint)
```
markusthoemmes: Those two need to be removed I reckon?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala
```diff
@@ -24,19 +24,14 @@ import akka.actor.ActorSystem
...
-import whisk.common.TransactionId
-import whisk.core.containerpool.ContainerId
-import whisk.core.containerpool.ContainerAddress
+import whisk.common.{Logging, TransactionId}
+import whisk.core.containerpool.{ContainerAddress, ContainerId}
 
-import scala.io.Source
+import scala.concurrent.{blocking, ExecutionContext, Future}
 import scala.concurrent.duration.FiniteDuration
+import scala.io.Source
```
markusthoemmes: Can you remove the push on this file? Seems like no change intended anyway?

common/scala/src/main/resources/application.conf
```diff
@@ -131,4 +131,4 @@ whisk {
         max = 512 m
         std = 256 m
     }
-}
+}
```
markusthoemmes: Please remove that change as its not an intended change.

markusthoemmes:
Compile error and conflict.

```
/home/travis/build/apache/incubator-openwhisk/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClientWithFileAccess.scala:97: not found: value Source
      val source = Source.fromFile(configFile)
```

rabbah:
@mcdan bump?

markusthoemmes:
Closing this due to inactivity. @mcdan please fix the merge conflicts and open a new PR if this is still applicable. Thanks!

Issue #3110, opened 2017-12-14
INVOKER_CORESHARE overloaded?


tysonnorris:
## Environment details:

* local deployment
## Steps to reproduce the issue:

Currently INVOKER_CORESHARE config (which dictates WhiskConfig.invokerCoreShare) appears overloaded in some confusing way:

1. it is used to directly set the `--cpu-shares` value on `docker run` commands
2. it is used to indirectly set the max active containers in the ContainerPool

## Provide the expected results and outputs:

Expect container CpuShares to be some value approximating a portion of 1024.0, based on coreshare+numcores and 1024.0.

## Provide the actual results and outputs:

docker inspect locally shows e.g. `"CpuShares": 2` for stem cell containers when INVOKER_CORESHARE=2

## Additional information you deem important:
I think this came from the invoker refactoring a while back:
```
-            cpuShares = OldContainerPool.cpuShare(config),
 +            cpuShares = config.invokerCoreShare.toInt,
```
where ye ole ContainerPool.cpuShare was
```
    private val totalShare = 1024.0 // This is a pre-defined value coming from docker and not our hard-coded value.
    def cpuShare(config: WhiskConfig) = (totalShare / getDefaultMaxActive(config)).toInt
```

PR #3198, opened 2018-01-17
move network + dns container args to pureconfig

tysonnorris:
also add "extra" container args to pureconfig

The main motivation here was to allow all action containers to some additional env vars (same value used for all containers). To get there, I moved network + dns args to pureconfig as well. 

There are still more container args that could be refactored, or it would be possible to move everything to the "extra-args" config.  These are left as an exercise for another time.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
@@ -27,6 +27,10 @@ import whisk.core.entity.ExecManifest
 import whisk.core.entity.InstanceId
 import whisk.spi.Spi
 
+case class ContainerArgsConfig(network: String,
+                               dnsServers: Seq[String] = Seq(),
+                               extraArgs: Map[String, Set[String]] = Map())
+
 /**
  * An abstraction for Container creation
  */
--- a/core/invoker/src/main/resources/application.conf
+++ b/core/invoker/src/main/resources/application.conf
@@ -19,4 +19,12 @@ whisk {
     pause: 10 seconds
     resume: 10 seconds
   }
+
+  # args for 'docker run' to use
+  container-factory.container-args {
+    network: bridge
+    dns-servers: []
+    extra-args: {}
+
+  }
 }
--- a/core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
@@ -19,14 +19,11 @@ package whisk.core.invoker
 
 import java.nio.charset.StandardCharsets
 import java.time.Instant
-
 import scala.concurrent.Future
 import scala.concurrent.duration._
 import scala.util.Failure
 import scala.util.Success
-
 import org.apache.kafka.common.errors.RecordTooLargeException
-
 import akka.actor.ActorRefFactory
 import akka.actor.ActorSystem
 import akka.actor.Props
@@ -82,8 +79,7 @@ class InvokerReactive(config: WhiskConfig, instance: InstanceId, producer: Messa
         Map(
           "--cap-drop" -> Set("NET_RAW", "NET_ADMIN"),
           "--ulimit" -> Set("nofile=1024:1024"),
-          "--pids-limit" -> Set("1024"),
-          "--dns" -> config.invokerContainerDns.toSet) ++ logsProvider.containerParameters)
+          "--pids-limit" -> Set("1024")) ++ logsProvider.containerParameters)
   containerFactory.init()
   sys.addShutdownHook(containerFactory.cleanup())
 
--- a/tests/src/test/scala/whisk/core/containerpool/test/ContainerArgsConfigTest.scala
+++ b/tests/src/test/scala/whisk/core/containerpool/test/ContainerArgsConfigTest.scala
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.containerpool.test
+
+import org.junit.runner.RunWith
+import org.scalatest.FlatSpec
+import org.scalatest.Matchers
+import org.scalatest.junit.JUnitRunner
+import pureconfig._
+import whisk.core.ConfigKeys
+import whisk.core.containerpool.ContainerArgsConfig
+
+@RunWith(classOf[JUnitRunner])
+class ContainerArgsConfigTest extends FlatSpec with Matchers {
+
+  it should "use defaults for container args map" in {
+    val config = loadConfigOrThrow[ContainerArgsConfig](ConfigKeys.containerArgs)
+
+    //check defaults
+    config.network shouldBe "bridge"
+    config.dnsServers shouldBe Seq[String]()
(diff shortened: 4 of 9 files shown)


common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
```diff
@@ -57,3 +65,23 @@ trait ContainerFactoryProvider extends Spi {
                           instance: InstanceId,
                           parameters: Map[String, Set[String]]): ContainerFactory
 }
+
+object ContainerFactory {
+
+  val containerArgsConfig = {
+    //since typesafe (lightbend) Config does NOT support parsing complex objects from system properties, we need a ConfigReader
+    //see https://github.com/lightbend/config/blob/master/HOCON.md#java-properties-mapping for details on properties handling from typesafe Config
+    //see pureconfig docs at https://pureconfig.github.io/docs/combinators.html
```
markusthoemmes: Can you elaborate on this a bit?

tried this locally:

```scala
case class ContainerArgsConfig(network: String, dnsServers: Seq[String], extraArgs: Map[String, Set[String]])

object Testing {
  import pureconfig._

    val config = """
                   |network: bridge
                   |dns-servers: [google.com]
                   |extra-args: {
                   |  test: [a, b]
                   |  another: [c, d]
                   |}
                   | """.stripMargin

    println(loadConfigOrThrow[ContainerArgsConfig](ConfigFactory.parseString(config)))
  }
}
```

Which yields:

```
ContainerArgsConfig(bridge,List(google.com),Map(another -> Set(c, d), test -> Set(a, b)))
```

Do we need something different?

common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
```diff
@@ -18,15 +18,23 @@
 package whisk.core.containerpool
 
 import akka.actor.ActorSystem
+import com.typesafe.config.ConfigFactory
+import java.util.ArrayList
+import pureconfig.ConfigReader
+import pureconfig.loadConfigOrThrow
```
markusthoemmes: You want to import `pureconfig._` here to get all the implicits.

core/invoker/src/main/resources/application.conf
```diff
@@ -19,4 +19,12 @@ whisk {
     pause: 10 seconds
     resume: 10 seconds
   }
+
+  # args for 'docker run' to use
+  container-factory.container-args {
```
mcdan: This is great I can add the container root path from #3196 to this struct instead of where I put it.

common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
tysonnorris: I tried to be as specific as possible in the comment :)

It is this: _typesafe (lightbend) Config does NOT support parsing complex objects from system properties_ i.e. when you override values via system properties (e.g. the `CONFIG_*` env vars in our case), they are always read as String type, never converted to List/Map/Object/etc.

So, in your example, you did not override the config from system properties.

IMHO this is a real flaw in lightbend/typesafe config, and we just need to be very careful about what types are parsed from the config, and make sure the override via system property works as you expect. I saw some similar logged issues with lightbend config, along with some arguments as to why it won't be fixed...

In fact, I just notice the same flaw exists for the `dns-servers` property - I'll add a config reader for that as well. Now I'm also wonder if the DockerClientTimeoutConfig will exhibit the same issue, if you try to override via system properties.

tysonnorris: Actually for 'dns-servers' (type `Seq[String]`), there exists a way to handle this via typesafe java properties parser: https://github.com/lightbend/config/pull/70

You have to specify each element in the list as a separate system property whose key ends with the index number in the list, e.g.
```
    System.setProperty("whisk.container-factory.container-args.dns-servers.0", "google.com")
    System.setProperty("whisk.container-factory.container-args.dns-servers.1", "1.2.3.4")
```

Note that we _could_ alternatively choose to add a custom parser for this, so that we can use a single property like "dns-servers"="[google.com,1.2.3.4]", but I ran into this issue trying to use the typesafe parser https://github.com/lightbend/config/issues/460 so we'd need to parse the value in yet another way.

core/invoker/src/main/resources/application.conf
tysonnorris: There is a subtle difference here in that the container-args config is not something that is specific to _docker cli_, rather it will apply to non-docker ContainerFactory impls (mesos/kube), while the container root path is very specific to docker cli. 
So I would keep the configs separated to:
- `whisk.docker`
- `whisk.container-factory`

common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
markusthoemmes: Got it, that makes sense!

Actually using your workaround for the lists, overriding the extra args works exactly the same way:

```scala
case class ContainerArgsConfig(network: String, dnsServers: Seq[String], extraArgs: Map[String, Set[String]])

object Testing {
  import pureconfig._

    System.setProperty("network", "test")
    System.setProperty("dns-servers.0", "google.com")
    System.setProperty("extra-args.test.0", "a")
    System.setProperty("extra-args.test.1", "b")
    System.setProperty("extra-args.try.0", "a")

    println(loadConfigOrThrow[ContainerArgsConfig](""))
  }
}
```

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -83,7 +85,7 @@ class InvokerReactive(config: WhiskConfig, instance: InstanceId, producer: Messa
           "--cap-drop" -> Set("NET_RAW", "NET_ADMIN"),
           "--ulimit" -> Set("nofile=1024:1024"),
           "--pids-limit" -> Set("1024"),
-          "--dns" -> config.invokerContainerDns.toSet) ++ logsProvider.containerParameters)
+          "--dns" -> containerArgs.dnsServers.toSet) ++ logsProvider.containerParameters)
```
markusthoemmes: I'm wondering: The dns configuration seems duplicated (given it's set in the factory itself again as well). Can we remove it here and thus remove the dependency on the config from `InvokerReactive`?

tysonnorris: well, now I'm noticing a problem that _transformEnvironment.sh_ is ignoring my env vars that end with index, e.g. `CONFIG_whisk_containerFactory_containerArgs_dnsServers.0=8.8.8.8`; will look into that...

tysonnorris: oh, using `_<index>` works...

tysonnorris:
@markusthoemmes I updated to remove ContainerArgsConfig from InvokerReactive, and also verified that the dnsServers config is adjustable from env vars. 

In order to leave the existing ansible config for dns functional, I had to add some changes to the deploy.yml for invoker, specifically:
```        {% for item in (invoker_container_network_dns_servers | default()).split(' ')  %}
        -e CONFIG_whisk_containerFactory_containerArgs_dnsServers_{{loop.index0}}={{ item }}
        {% endfor %}
```
Let me know if you have concerns. Tested locally fine, FWIW.

tysonnorris:
Any concerns on this @markusthoemmes @rabbah ?

After this is merged I'll take a stab at fixing #3110 along with moving some (all?) of the other invoker configs to pureconfig.

markusthoemmes:
PR1 2545 🔵

tests/src/test/scala/whisk/core/containerpool/test/ContainerArgsConfigTest.scala
```diff
@@ -0,0 +1,58 @@
...
+    val config = loadConfigOrThrow[ContainerArgsConfig](ConfigKeys.containerArgs)
+
+    //check defaults
+    config.network shouldBe "bridge"
+    config.dnsServers shouldBe Seq[String]()
+    config.extraArgs shouldBe Map[String, Set[String]]()
+  }
+
+  it should "override defaults from system properties" in {
+    System.setProperty("whisk.container-factory.container-args.extra-args.label.0", "l1")
```
chetanmeh: May be the tests should revert changes in System property post run otherwise it may lead to side effects in other test

tysonnorris: agreed - since this just got merged, I will address in a future pr - this area will get some revisions as I work on some of the containerfactory details for mesos

PR #3199, opened 2018-01-18
One tid for the whole system.

cbickel:
With this PR, I want to introduce the usage about one tid to trace one request through the whole system. With this PR, I generate the tid in the nginx.

Long term, my plan is, to generate the tid already in the provider to trace one request through the whole system.

To realize this plan, the tid will be generated by nginx and passed to the components as header (name is configurable). The tid, that is generated by nginx is "generated from 16 random bytes, in hexadecimal" [1].

Things that are still TODO:
- [x] Change CLI to not handle error code as int. https://github.com/apache/incubator-openwhisk-client-go/pull/77
- [ ] Bump CLI https://github.com/apache/incubator-openwhisk-cli/pull/284

Things that are not possible:
- The tid, generated by nginx has the same format like our activation id. Unfortunately it is not possible to use this tid as transactionId, because of the following scenarios with several activations within one tid:
    - trigger, rule, activation
    - sequences

[1] http://nginx.org/en/docs/http/ngx_http_core_module.html


rabbah:
> if the header reaches nginx, it should not generate a new one

Can you clarify: If we’re using this header value as the tid we want it to be privileged and not set by a client, no?

rabbah:
Couple more thoughts: 

1. going from long to bigint might have some implication since our tid is pervasive: a long can be unboxed, a bigint is on the heap. It might be unavoidable though. 

2. We need to fix the trigger/action activation using an internal path and stop doing the post. As this might be a longer task than the change in this pr, might have to do what you’re doing as a stop gap. 

3. For dependent operation (trigger/action, or sequence/action), it would be nice to have correlation between parent and child but have two different ids.

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -254,7 +251,9 @@ trait TransactionCounter {
   // seed the counter so transids do not overlap: instanceOrdinal + n * stride, start at n = 1
   private lazy val cnt = new AtomicInteger(instanceOrdinal + stride)
 
-  def transid(): TransactionId = {
-    TransactionId(cnt.addAndGet(stride))
+  def transid(nginxRequestId: Option[String] = None): TransactionId = {
+    nginxRequestId
+      .map(id => TransactionId(new BigInteger(id, 16)))
```
rabbah: Should probably use a Try here and guard against a malformed tid. I’d rename the variable to drop nginx reference.

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -200,39 +204,37 @@ object TransactionId {
...
-        JsArray(JsNumber(t.meta.id), JsNumber(t.meta.start.toEpochMilli), JsBoolean(t.meta.extraLogging))
-      else
-        JsArray(JsNumber(t.meta.id), JsNumber(t.meta.start.toEpochMilli))
-    }
+    def write(t: TransactionId) =
+      JsArray(
+        JsString(t.meta.id),
+        JsNumber(t.meta.start.toEpochMilli),
+        JsBoolean(t.meta.extraLogging),
+        JsBoolean(t.meta.isSid))
```
rabbah: fine although: is there any case when a system id is actually serialized? i don't think there are.

common/scala/src/main/scala/whisk/http/ErrorResponse.scala
```diff
@@ -248,13 +247,27 @@ object ErrorResponse extends Directives with DefaultJsonProtocol {
   }
 
   implicit val serializer = new RootJsonFormat[ErrorResponse] {
-    def write(er: ErrorResponse) = JsObject("error" -> er.error.toJson, "code" -> er.code.meta.id.toJson)
+    def write(er: ErrorResponse) = {
+      val tidMeta = er.code.meta
+      // Convert the tid to a number. If this is not possible, use `unknown`-tid
```
rabbah: should this serialization of the tid be moved to the transaction id singleton?
also i'd note that the system id should never leak out in the response (there aren't any instances where it will as i remember).

ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -83,6 +89,8 @@ http {
         proxy_ssl_verify off;
         proxy_ssl_session_reuse on;
 
+        proxy_set_header OW-TID $ow_tid;
```
rabbah: can you add a comment here that clarifies whether it's possible for a client to override this internally set header?

tysonnorris:
@ddragosd WDYT?

common/scala/src/main/scala/whisk/common/TransactionId.scala
cbickel: Isn't this also used to put the TransactionId into the activationMessage that is sent from the controller to the invoker?
And the loadbalancer is invoking some health actions with a sid as TransactionId.

sven-lange-last:
Some background info: https://www.nginx.com/blog/application-tracing-nginx-plus/

ansible/roles/nginx/templates/nginx.conf.j2
ddragosd: Can we use a known header name: `X-Request-ID` instead of a custom header ?

ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -12,10 +12,16 @@ http {
 {# allow large uploads, need to thread proper limit into here #}
     client_max_body_size 50M;
 
+    map $http_ow_tid $ow_tid {
```
ddragosd: Can we use [set_secure_random_alphanum](https://github.com/openresty/set-misc-nginx-module#set_secure_random_alphanum) instead ?

ddragosd: And if we want to respect an ID sent by the client, then we can use a config similar to : https://github.com/adobe-apiplatform/apigateway/blob/master/api-gateway-config/conf.d/commons/common-headers.conf#L23-L30 which only sets an ID unless one is provided. or is there any reason to have this ID adhere to a strict format, instead of having it treated as a String ?

ddragosd:
💯  @cbickel ! this is very useful !
Thanks for working on it. I was in need of something similar recently, when I needed to trace a transaction through the system, GW included, in order to debug a failure. Can't wait to see this merged. I left a few comments above with some ideas.

ansible/roles/nginx/templates/nginx.conf.j2
sven-lange-last: @ddragosd `set_secure_random_alphanum` depends on `/dev/urandom` which may impact performance. Is there a requirement to use a secure random and not just the ngninx built-in `request_id`? We should also understand the performance of `set_secure_random_alphanum` compared to `request_id`.

ddragosd: @sven-lange-last 👍   NGINX's `request_id` should do it as well. I've been using the other `set_secure_random_alphanum` in the past years ( before NGINX 1.11.x added `request_id` ) and the performance implications were not visible. 

One thing to consider here is that OW should accept a variable length for this ID ( which IMHO should be fine) b/c NGINX might update this in the future; I'm not aware of a way to configure the size `16` or `32` or another size, unlike with the older `set_secure....`; Note that the article referenced says it's 32 chars, while the [docs](http://nginx.org/en/docs/http/ngx_http_core_module.html) say it's 16.

tysonnorris:
@cbickel @sven-lange-last any updates on this?

ansible/roles/nginx/templates/nginx.conf.j2
markusthoemmes: Agreed on the variable length. It shouldn't care at all (so nobody is forced to use this specific config as their front-door). Will be changed!

ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -59,6 +66,12 @@ http {
     }
 
     proxy_set_header X-OW-EXTRA-LOGGING $extra_logging;
+    # Set the request id generated by nginx as tid-header to the upstream.
+    # This tid is either the request-id generated by ngix or a tid, sent by the caller.
+    proxy_set_header OW-TID $ow_tid;
+
+    # Send the tid always back as header.
+    add_header OW-TID $ow_tid always;
```
markusthoemmes: As  @ddragosd  mentioned this will be changed to `X-Request-Id` (per https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Common_non-standard_request_fields it's a quasi standard).

The header that's picked in the controller will be configurable, so you can use whatever front-door generates these ids for you without relying on this too much.

tysonnorris:
@csantanapr I'm not sure what typescript spec may be referenced in the email thread. Can you point it out? Or is this OK to merge now?

csantanapr:
@tysonnorris it was @starpit who brought up the TypeScript spec. 
But I would not make that spec a blocker on this. 

What I worry a CLI (wskdeploy and wsk) breakage if CLI is not broken because of the change on type or field name change for “code” then I’m good 

@mdeuser could you test if CLI is ok?

rabbah:
Another case of https://github.com/apache/incubator-openwhisk-cli/issues/148? The cli shouldn’t really care what the type of the code is.

ansible/group_vars/all
```diff
@@ -96,8 +96,10 @@ jmx:
   jvmCommonArgs: "-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.local.only=false -Dcom.sun.management.jmxremote.authenticate=true -Dcom.sun.management.jmxremote.password.file=/root/jmxremote.password -Dcom.sun.management.jmxremote.access.file=/root/jmxremote.access"
   enabled: "{{ jmxremote_enabled | default('true') }}"
 
-transactions:
-  stride: "{{ groups['controllers'] | length }}"
+transaction:
+  header: "{{ transaction_header | default('X-Request-ID') }}"
+  # The tid can be any String between 1 and 64 characters.
+  regex: "{{ whisk_tid_regex | default('[a-zA-Z0-9_-]{1,64}') }}"
```
rabbah: excluding - would make it easier to copy/paste a code (the dash breaks double click to highlight a string).

ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -12,10 +12,17 @@ http {
 {# allow large uploads, need to thread proper limit into here #}
     client_max_body_size 50M;
 
+    map $http_{{ transaction.header | lower | replace('-', '_') }} $ow_tid {
```
rabbah: Nice.

ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -12,10 +12,17 @@ http {
...
+        # request_id: unique request identifier generated from 16 random bytes, in hexadecimal
+        default $request_id;
+        # If the caller has already set a tid into the header in the correct format, this one will be taken.
+        "~^(?<tid>{{ transaction.regex }})$" $tid;
```
rabbah: The tid is used in the cache logic - if the id isn’t unique (specified by a user, it will affect the cache behavior).

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -200,58 +196,42 @@ object TransactionId {
...
-  val dbBatcher = TransactionId(-140) // Database batcher
-
-  def apply(tid: BigDecimal, extraLogging: Boolean = false): TransactionId = {
-    Try {
-      val now = Instant.now(Clock.systemUTC())
-      TransactionId(TransactionMetadata(tid.toLong, now, extraLogging))
-    } getOrElse unknown
+  val generatorConfig = loadConfigOrThrow[TransactionGeneratorConfig](ConfigKeys.transactions)
+
+  val unknown = TransactionId("sid_unknown")
```
rabbah: Is there a mechanism that prevents a user specified sid?

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -200,58 +196,42 @@ object TransactionId {
...
-  val instanceOrdinal: Int
-
-  // seed the counter so transids do not overlap: instanceOrdinal + n * stride, start at n = 1
-  private lazy val cnt = new AtomicInteger(instanceOrdinal + stride)
-
-  def transid(extraLogging: Boolean = false): TransactionId = {
-    TransactionId(cnt.addAndGet(stride), extraLogging)
-  }
-}
+case class TransactionGeneratorConfig(header: String)
```
rabbah: I think for local testing it might still be nice to have a counter based id instead. Thoughts?

tests/src/test/scala/whisk/common/SchedulerTests.scala
```diff
@@ -127,7 +127,7 @@ class SchedulerTests extends FlatSpec with Matchers with WskActorSystem with Str
 
     waitForCalls()
     stream.toString.split(" ").drop(1).mkString(" ") shouldBe {
-      s"[ERROR] [#sid_1] [Scheduler] halted because $msg\n"
+      s"[ERROR] [${transid.toString}] [Scheduler] halted because $msg\n"
     }
```
rabbah: Don’t need toString.

tests/src/test/scala/whisk/core/controller/test/PackagesApiTests.scala
```diff
@@ -750,7 +750,6 @@ class PackagesApiTests extends ControllerTestCommon with WhiskPackagesApi {
       status should be(Conflict)
       val response = responseAs[ErrorResponse]
       response.error should include("Package not empty (contains 1 entity)")
-      response.code.id should be >= 1L
```
rabbah: Can you change this check to assert the id is not empty? This weakens the test to where a missing id is accepted.

tests/src/test/scala/whisk/core/database/test/DbUtils.scala
```diff
@@ -45,13 +44,14 @@ import whisk.core.entity.types.EntityStore
...
-trait DbUtils extends TransactionCounter {
+trait DbUtils {
   implicit val dbOpTimeout = 15 seconds
-  override val instanceOrdinal = 0
-  val instance = InstanceId(instanceOrdinal)
+  val instance = InstanceId(0)
   val docsToDelete = ListBuffer[(ArtifactStore[_], DocInfo)]()
   case class RetryOp() extends Throwable
 
+  def transid() = TransactionId.testing
```
rabbah: As noted earlier a sequential counter here would be useful when looking at a test log.

ansible/group_vars/all
cbickel: That's a good idea. This makes operations much easier :)

common/scala/src/main/scala/whisk/common/TransactionId.scala
cbickel: Good catch.
I've added a gate to the nginx and the controller.

common/scala/src/main/scala/whisk/common/TransactionId.scala
cbickel: I've addad a counter to the test. This might be helpful.

cbickel:
@mdeuser Can you point me to the correct position in the CLI where the integer for the code is required? Because I was not able to find it in the repositories https://github.com/apache/incubator-openwhisk-cli and https://github.com/apache/incubator-openwhisk-client-go

mdeuser:
@cbickel https://github.com/apache/incubator-openwhisk-client-go/blob/7671398729ab11fdaeb17e1f3ae308e3126f740a/whisk/client.go#L610

i think we should be careful of the sequence of the deployment of these changes.  i.e. first release a version of the cli/plugin that handles the `code` value being either an integer or string.  then deploy the backend.  i'm not sure if there are other sdks that might be impacted by this api change...

rabbah:
Makes sense @mdeuser - should use this as another opportunity to rip out another typed schema in the cli. Most values should be opaque and transient and their types inconsequential.

ddragosd:
:+1: @rabbah

tysonnorris:
Does this require additional work in cli? Or ready to merge?

common/scala/src/main/scala/whisk/common/TransactionId.scala
```diff
@@ -39,10 +34,10 @@ import pureconfig._
  */
 case class TransactionId private (meta: TransactionMetadata) extends AnyVal {
   def id = meta.id
-  override def toString = {
-    if (meta.id > 0) s"#tid_${meta.id}"
-    else if (meta.id < 0) s"#sid_${-meta.id}"
-    else "??"
+  override def toString = s"#tid_${meta.id}"
```
rabbah: will a tid now appear as `tid_sid_...` in the logs?

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
```diff
@@ -97,7 +105,20 @@ trait BasicHttpService extends Directives with TransactionCounter {
...
     }
-    extract(_ => transid(extraLogging))
+    extract { req =>
+      val tid =
+        req.request.headers
+          .find(_.name == TransactionId.generatorConfig.header)
+          .map(_.value)
+          .filterNot(_.startsWith(TransactionId.systemPrefix))
+          .getOrElse {
+            // As this is only a fallback, because the tid should be generated by nginx, this shouldn't be used.
```
rabbah: did you consider if we should add a log warning when a header is missing?

cbickel:
@tysonnorris No, everything is done in the meanwhile. I'll just resolve my conflict, work on the last two comments and then it should be good to merge :)

common/scala/src/main/scala/whisk/common/TransactionId.scala
cbickel: Yes, it will.
We thought about removing the `sid_` but we decided to leave it in. If we leave it in, the sids will  attract more attention. Otherwise you see only digits and letters in every tid, and you might miss the sids, as they are also only letters, which might not attract enough attention.

rabbah: heh... not entirely convinced but I don’t think you can easily remove it either so it is fine to let it be. If it becomes annoying we will see it.

PR #3202, opened 2018-01-19
Support action continuations in the controller

tardieu:
This PR introduces a new type of actions called compositions. Compositions are a kind of sequences with dynamically computed steps. While the components of a sequence must be specified before invocation, components of a composition can be decided as the composition is running.

Like the sequence implementation, the bulk of the composition implementation lies in the controller. It chains the invocations of the component actions in the composition. It keeps track of the execution state. It produces the final activation record for the composition invocation.

In contrast with sequences and in an attempt to keep the changes contained in the controller code, a composition is not a new kind of action but rather a action with a marker annotation. This means no CLI changes for instance.

A conductor action is an action with a defined “conductor” annotation. The value of the annotation does not matter. For simplicity, sequences cannot be conductors. The annotation is ignored on sequences. But sequences may be components of a composition.

Conductor actions drive the execution of compositions. A conductor action may either return a final result or a triplet { action, params, state }. In the latter case, the specified component action is invoked on the specified params object. Upon completion of this action the conductor action is reinvoked with a payload that combines the output of the action with the state returned by the previous conductor invocation. The composition result is the result of the final conductor invocation in the chain of invocations.

A composition is therefore a sequence of invocations alternating invocations of the (fixed) conductor action with invocations of the dynamically specified components of the composition. In a sense, the conductor action acts as a scheduler that decodes and executes a sequential program. Conductor actions make it possible to support dynamic composition in OpenWhisk without committing to any particular execution model (e.g., finite state machines of one kind or another), or invocation conventions.

Compositions may be nested, i.e, a step in a composition may itself be a composition. The controller recognizes this pattern and handles it efficiently.

Like a sequence, one composition is materialized by one action. Creating a composition requires the definition of one conductor action. Deleting the composition consists in deleting the corresponding action. Like a sequence, a composition refers to the composed actions by name (i.e., by reference).

The PR enables a much better implementation of the composer library for OpenWhisk (https://github.com/ibm-functions/composer) that offers faster performance, lower cost (no double billing), and dramatically improves usability. With this PR, a composition _is_ an action. There is no distinction between invoking a composition and an action. Both invocations returns an activation id that can be used to fetch the result of the execution of the action or composition. Both invocations can be blocking. Both can be web actions.

As for sequences, there is a configurable upper bound on the length of a composition (number of steps). In contrast with sequences, the limit is assessed at run time, and the composition invocation is aborted if the limit is exceeded. An entitlement check verify that each composed action is accessible prior to its execution. A composition invocation like a sequence invocation counts as one invocation viz. throttling. In particular, a composition will not be aborted during execution due to throttling. Therefore a composition designed to have fewer steps than the limit will not abort due to quotas once it has started execution.

This PR comes with unit tests for the controller as well as integration tests.

Example:
```
$ cat > increment.js
function main({ value }) { return { value: value + 1 } }
^D

$ cat > conductor.js
function main(params) {
    switch (params.$step || 0) {
        case 0: delete params.$step; return { params, action: 'increment', state: { $step: 1 } }
        case 1: delete params.$step; return { params, action: 'increment', state: { $step: 2 } }
        case 2: delete params.$step; return params
    }
}
^D

$ wsk action update increment increment.js
$ wsk action update conductor conductor.js -a conductor true

$ wsk action invoke conductor -br -p value 3
{
    "value": 5
}
```
This example demonstrates a simple composition with two increment steps. The progress of the execution is tracked via the `$step` field of the parameter object that is propagated from a conductor invocation to the next but hidden from the composed actions.


csantanapr:
worth an email to the mailing list, this is a sweet feature.

csantanapr:
docs? (i.e. .md)

tardieu:
The code is documented. I'll add a docs/compositions.md. But the real power of compositions will come from composer and I am not going to document this here.

tardieu:
I rewrote the commit history as requested.

cfjedimaster:
Am I wrong to be worried that this may confuse folks with Composer? Like - I assumed this was related and it seems to have some similarities but if it is truly unique, maybe the name "composition" should be changed?

cfjedimaster:
"The annotation is ignored on sequences"
No, no, no, no. Please no. I keep running into the bug where you can't do a default param for sequences and it is silently ignored.

tardieu:
We can support the annotation on sequences. But I would like to do sequences as a separate PR as this one is quite large already.

tardieu:
Composer is being rewritten to take advantage of this PR internally. The composer API is unchanged. The PR makes it possible to execute action compositions efficiently. Composer will continue to help author action compositions by automatically generating the conductor action code from a high-level specification. For example, code equivalent to the double increment `conductor.js` action code shown above can be generated by Composer from the much simpler code `composer.sequence('increment', 'increment')`.

cfjedimaster:
... interesting. So a user would have 2 main ways to make compositions. Either using Composer or "by hand". Of course, when I work with Composer I'm working with a local file anyway. It feels like this could be - potentially - a bit confusing for the end user. I'm not saying this is *bad* in anyway, but I worry about how I'd explain this to a group of devs, know what I mean?

Could you elaborate on when I'd use Composer vs compositions? (And since Compser makes compositions, you can see already it's a bit confusing. ;)

tardieu:
Always use Composer! It is much easier, less error prone, and there is no real drawback. There is a good reason I have not written docs for compositions. :D

cfjedimaster:
So if it fair to say: "This PR changes stuff beneath the covers and Ray you can stop paying attention."  Cuz I'm totally fine with ignoring this if so. :)

tardieu:
We have been discussing ways to run composer server-side, i.e., upload the composer source directly as opposed to having to compile the composition locally. I suspect we will end up supporting this mode of operation in some way.

tardieu:
Yes you can think of this PR as improving the Composer backend. The conductor action "language" is not really meant for end users. But the PR will have visible positive impact on Composer. First, the distinction between actions and apps will basically disappear. Second, REDIS will be an option (for compositions with large amounts of state) not a requirement.

cfjedimaster:
Thank you - I appreciate you taking the time to explain this to me! (I wanted to do more than just a thumbs up - I really am thankful.)

tardieu:
The distinction between composer and compositions is not gratuitous... This PR with the addition of action compositions is about adding a fundamental capability to the OpenWhisk controller with as little a bias as possible. It does not favor a specific approach to building composition. Composer on the other hand embodies many design choices. For instance, it currently targets Javascript. In contrast, the PR is language/runtime agnostic. While I don't think writing compositions by hand is a compelling proposition for OpenWhisk users, there could be many alternatives to Composer. All would benefit from the PR in the same way.

cfjedimaster:
Ah the JS bias in Composer is a very good point. This would definitely help the non-JS folks.

csantanapr:
@tardieu 
>The code is documented. I'll add a docs/compositions.md. But the real power of compositions will come from composer and I am not going to document this here.

This is a change that end users would not use directly, but is a change in the primitives to a core/kernel component (controller) to allow more efficient (store and forward state) and language agnostic applications. This is a feature to be use by someone building a framework or tooling for end user (i.e. downstream integrators/providers), but not end users (serverless developers).

So the docs should be written for a downstream integrator/provider, on how the primitive works and some examples how to construct an orchestration to illustrate the concepts.

rabbah:
@csantanapr it's not quite right in that openwhisk developers can use this feature directly, as the docs should show... as they do with sequences, and the `wsk` CLI. Another way of looking at it, you can write assembly code if you like... but why when you can write - gasp - node code.

tardieu:
I added a readme.

ddragosd:
@tardieu 💯 . This looks awesome ! 

> Second, REDIS will be an option (for compositions with large amounts of state) not a requirement.

So good, so good, I got ... a smaller barrier to create compositions ! 👍 
What does _large amounts of state_ mean ?

ddragosd:
@tardieu the `conductor` action is basically like an extension for OpenWhisk, b/c it is the one that knows about the constructs in a composition, right ? I'm not challenging the decision to implement this as an `action`, but I'd like to understand the benefits of developing this feature with actions, vs with SPI.

tardieu:
Large amount means that the combination of the action parameters and composition state exceeds the limit on parameter size (1 MB).

tardieu:
@ddragosd The PR is really only about chaining action invocations dynamically. It does not contain anything about composition constructs. It does not need to. This gives a lot a flexibility for designing composition frameworks. We don't want to special case a specific composition framework.

The motivation for making a composition an action follows from sequences. A composition, a sequence, and an action become interchangeable. As a user, I can invoke an action, a sequence, or a composition not having to know if it is one or the other.

tests/src/test/scala/system/basic/WskConductorTests.scala
```diff
@@ -0,0 +1,311 @@
...
+      action.create(
+        echo,
+        Some(TestUtils.getTestActionFilename("echo.js")),
+        annotations = Map("conductor" -> true.toJson))
+    }
+
+    // a normal result
+    val run = wsk.action.invoke(echo, Map("payload" -> testString.toJson))
+    withActivation(wsk.activation, run) { activation =>
+      activation.response.status shouldBe "success"
```
rabbah: Maybe we should _separately_ move https://github.com/apache/incubator-openwhisk/blob/master/common/scala/src/main/scala/whisk/core/entity/ActivationResult.scala#L66-L69 into a singleton.

tests/src/test/scala/system/basic/WskConductorTests.scala
```diff
@@ -0,0 +1,311 @@
...
+    assetHelper.withCleaner(wsk.action, conductor) { (action, _) =>
+      action.create(
+        conductor,
+        Some(TestUtils.getTestActionFilename("conductor.js")),
+        annotations = Map("conductor" -> true.toJson))
+    }
+
+    val step = "step" // step action
+    assetHelper.withCleaner(wsk.action, step) { (action, _) =>
+      action.create(step, Some(TestUtils.getTestActionFilename("step.js")), memory = Some(128 MB))
```
rabbah: is the 128MB here important? On the open deployment, I think 256MB is a stem cell whereas this would not be... not sure if you're trying to force a particular condition.

tests/src/test/scala/system/basic/WskConductorTests.scala
```diff
@@ -0,0 +1,311 @@
...
+   * checks logs for the activation of a conductor action (length/size and ids)
+   * checks that the cause field for nested invocations is set properly
+   * checks duration
+   * checks memory
+   */
+  private def checkConductorLogsAndAnnotations(activation: ActivationResult, size: Int) = {
+    activation.logs shouldBe defined
+    // check that the logs are what they are supposed to be (activation ids)
+    // check that the cause field is properly set for these activations
+    activation.logs.get.size shouldBe (size) // the number of activations in this sequence
```
rabbah: I think you can rewrite this as `activation.logs.get should have length size`

tests/src/test/scala/system/basic/WskConductorTests.scala
```diff
@@ -0,0 +1,311 @@
...
+    var totalTime: Long = 0
+    var maxMemory: Long = 0
+    for (id <- activation.logs.get) {
```
rabbah: `activation.logs.get.foreach`?

tests/src/test/scala/system/basic/WskConductorTests.scala
tardieu: Yes I am trying to test the max-memory-across-the-composition computation. I could change to a different value, other than 256 MB. What would be a better choice?

tests/src/test/scala/system/basic/WskConductorTests.scala
tardieu: OK. I will try.

tests/src/test/scala/system/basic/WskConductorTests.scala
tardieu: OK

tests/src/test/scala/system/basic/WskConductorTests.scala
```diff
@@ -0,0 +1,311 @@
...
+
+  behavior of "Whisk conductor controller"
+
+  it should "invoke a conductor action with no dynamic steps" in withAssetCleaner(wskprops) { (wp, assetHelper) =>
+    val echo = "echo" // echo conductor action
+    assetHelper.withCleaner(wsk.action, echo) { (action, _) =>
+      action.create(
+        echo,
+        Some(TestUtils.getTestActionFilename("echo.js")),
+        annotations = Map("conductor" -> true.toJson))
```
rabbah: in this test, you're validating that a conductor annotation does not modify the behavior of a "classic action" - is that correct?

tests/src/test/scala/system/basic/WskConductorTests.scala
```diff
@@ -0,0 +1,311 @@
...
+
+    // an error result
+    val secondrun = wsk.action.invoke(echo, Map("error" -> testString.toJson))
+    withActivation(wsk.activation, secondrun) { activation =>
+      activation.response.status shouldBe "application error"
+      activation.response.result shouldBe Some(JsObject("error" -> testString.toJson))
+      checkConductorLogsAndAnnotations(activation, 1) // echo
+    }
+
+    // a wrapped result { params: result } is unwrapped by the controller
```
rabbah: perhaps add to the comment... because of the conductor annotation?

tests/src/test/scala/system/basic/WskConductorTests.scala
```diff
@@ -0,0 +1,311 @@
...
+        JsString(compositionComponentInvalid(JsString(invalid))))
+      checkConductorLogsAndAnnotations(activation, 1) // echo
+    }
+
+    // an undefined action
+    val undefinedrun = wsk.action.invoke(echo, Map("payload" -> testString.toJson, "action" -> missing.toJson))
+    withActivation(wsk.activation, undefinedrun) { activation =>
+      activation.response.status shouldBe "application error"
+      activation.response.result.get.fields.get("error") shouldBe Some(JsString(compositionComponentNotFound(missing)))
+      checkConductorLogsAndAnnotations(activation, 1) // echo
```
rabbah: is there a separate  integration test for the not entitled case?
how about invoking an action which is actually not an action (eg say a trigger).

tests/src/test/scala/system/basic/WskConductorTests.scala
tardieu: More or less. A conductor annotation should not modify the behavior of an action that returns a dictionary which does not include a `params` field or an `action` field. I'll add a comment.

tests/src/test/scala/system/basic/WskConductorTests.scala
tardieu: ok

tests/src/test/scala/system/basic/WskConductorTests.scala
rabbah: perhaps a comment for clarity - otherwise fine then.

tests/src/test/scala/system/basic/WskConductorTests.scala
tardieu: I have a unit test for this scenario but no integration test. In fact I am tempted to remove the integration tests for error conditions that replicate unit tests because they take a long time. The unit tests only take a few seconds. The integration tests take more than a minute. Thoughts?

tests/src/test/scala/system/basic/WskConductorTests.scala
tardieu: I am not sure I understand the suggestion here.

dubeejw:
Hey @tardieu, does this work with web actions, and is there a test to ensure that an action composition can't be used to infinitely invoke a single action?

tardieu:
@dubeejw yes it works with web actions (no test), and yes there are unit tests for invocation limits.

docs/compositions.md
```diff
@@ -0,0 +1,144 @@
...
+Action compositions make it possible to dynamically build and invoke a series of actions. Action compositions are similar to sequences. However, whereas the components of a sequence must be specified before invoking the sequence, components of a composition can be decided as the composition is running.
+
+## Example
+
+Suppose we define an _increment_ action:
+
+```
+$ cat > increment.js
+function main({ value }) { return { value: value + 1 } }
+^D
```
csantanapr: What are this ^D about ? in the file

csantanapr: separate command from code, make code in it's own block with the language easy to copy paste out.
Also I think removing the `cat` command is better, I'm not sure if there is a `cat` in windows.

Example:
Suppose we define an _increment_ action: in a source file `increment.js`
```javascript
function main({ value }) { return { value: value + 1 } }
```
Create the action _increment_:
```
wsk action create increment increment.js
```

docs/compositions.md
```diff
@@ -0,0 +1,144 @@
+# Action compositions
+
+Action compositions make it possible to dynamically build and invoke a series of actions. Action compositions are similar to sequences. However, whereas the components of a sequence must be specified before invoking the sequence, components of a composition can be decided as the composition is running.
```
jpkessle: Action compositions make it possible to dynamically build and invoke a series of Actions, similar to sequences. However, whereas the components of a sequence must be specified before invoking the sequence, components of a composition can be decided as the composition is running.

docs/compositions.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Example
+
+Suppose we define an _increment_ action:
```
jpkessle: Suppose we define an `_increment_ action`:

docs/compositions.md
```diff
@@ -0,0 +1,144 @@
...
+
+```
+$ cat > increment.js
+function main({ value }) { return { value: value + 1 } }
+^D
+
+$ wsk action create increment increment.js
+```
+
+We can use this _increment_ action in a composition as follows:
```
jpkessle: We can use this `_increment_` Action in a composition as follows:

docs/compositions.md
tardieu: I am missing the diff here...

docs/compositions.md
```diff
@@ -0,0 +1,144 @@
...
+        case 1: delete params.$step; return { params, action: 'increment', state: { $step: 2 } }
+        case 2: delete params.$step; return { params }
+    }
+}
+^D
+
+$ wsk action create composition composition.js -a conductor true
+```
+
+The key to making this action a composition is the _conductor_ annotation, which we discuss in the next section.
```
jpkessle: The key to making the Action a composition is the _conductor_ annotation, which we discuss in the next section.

This example composition executes two _increment_ Actions in a sequence:

docs/compositions.md
tardieu: So what is the story for [a|A]ction [c|C]omposition and the [c|C]omposition shorthand?

docs/compositions.md
```diff
@@ -0,0 +1,144 @@
...
+
+This example composition executes two _increment_ actions in a sequence:
```
jpkessle: The following example composition executes two _increment_ Actions in a sequence:

docs/compositions.md
```diff
@@ -0,0 +1,144 @@
...
+
+In essence, this composition defines a program with three steps:
+
+- step 0: invoke increment on the input dictionary,
+- step 1: invoke increment on the output dictionary from step 1,
+- step 2: return the output dictionary from step 2.
+
+## Conductor Actions
+
+An action composition is driven by a _conductor action_. A _conductor action_, or in short a _conductor_,  is an action with a _conductor_ annotation. The value of the annotation is irrelevant.
```
jpkessle: An Action composition is driven by a _conductor action_. A _conductor action_, or in short a _conductor_,  is an Action with a _conductor_ annotation. The value of the annotation is irrelevant.

docs/compositions.md
```diff
@@ -0,0 +1,144 @@
...
+
+Because a conductor action is an action, it has all the attributes of an action (e.g., name, namespace, default parameters, limits...) and it can be managed as such, for instance using the `wsk action` CLI commands. It can be part of a package. It can be a web action. And so on.
```
jpkessle: Because a _conductor_ Action is an Action, it retains all of the attributes of an Action (e.g., name, namespace, default parameters, limits...), and it can be managed as such. For instance, using the `wsk action` CLI commands. It can be part of a package or be a web action and so on.

docs/compositions.md
tardieu: Updated

tardieu:
@jpkessle I see you want to capitalize Action. But isn't this completely inconsistent with the rest of the documentation, for instance the current document for actions: https://github.com/apache/incubator-openwhisk/blob/master/docs/actions.md? What about compositions, action compositions, composition actions, sequences, packages, etc. ?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+      duration = Some(session.duration))
+
+    // complete the promise if any
+    session.result.map { _.success(Right(activation)) }
+
+    logging.info(this, s"recording activation '${activation.activationId}'")
+    WhiskActivation.put(activationStore, activation)(transid, notifier = None) onComplete {
+      case Success(id) => logging.info(this, s"recorded activation")
+      case Failure(t) =>
+        logging.error(
```
rabbah: this should be logged in the entity store.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+        Parameters(WhiskActivation.kindAnnotation, JsString(Exec.SEQUENCE)) ++
+        Parameters(WhiskActivation.conductorAnnotation, JsBoolean(true)) ++
+        causedBy ++
+        sequenceLimits,
+      duration = Some(session.duration))
+
+    // complete the promise if any
+    session.result.map { _.success(Right(activation)) }
+
+    logging.info(this, s"recording activation '${activation.activationId}'")
```
rabbah: probably unnecessary - the entity store will log straddling the operation (now at debug level).

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -65,18 +76,67 @@ protected[actions] trait PrimitiveActions {
...
+   * The activation record for a composition also includes a specific annotation "conductor" with value true.
+   */
+  protected[actions] def invokeSingleAction(
+    user: Identity,
+    action: ExecutableWhiskActionMetaData,
+    payload: Option[JsObject],
+    waitForResponse: Option[FiniteDuration],
+    cause: Option[ActivationId])(implicit transid: TransactionId): Future[Either[ActivationId, WhiskActivation]] = {
+
+    if (action.annotations.get(WhiskActivation.conductorAnnotation).isDefined) {
```
rabbah: it's less confusing to actually check this value for `true` as it may be suprising for an annotation `conductor: false` to be treated as a conductor anyway. `asBool("conductor").exists(identity)` will do.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+        Promise[Either[ActivationId, WhiskActivation]]() // placeholder for result if blocking invoke
+      })
+
+    invokeConductor(user, payload, session)
+
+    // is caller waiting for the result of the activation?
+    waitForResponse
+      .map { timeout =>
+        // handle timeout
+        session.result.head.future
```
rabbah: why .head not .get on an option?
perhaps a comment that this is safe since if waitForResponse is defined, so the the session.result.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+        .getOrElse(payload)
+
+      // invoke conductor action
+      session.accounting.conductors += 1
+      val response =
+        invokeSimpleAction(
+          user,
+          session.action,
+          params,
+          Some(session.action.limits.timeout.duration + 1.minute), // wait for result
```
rabbah: i'd use 
```
..., 
waitForResponse = ...,
cause = ...)
```
in addition to/vs the comment.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+          completeAppActivation(user, session, response)
+        case Success(Left(activationId)) =>
+          // invocation timeout
+          session.logs += activationId
+          val response = ActivationResponse.whiskError(compositionActivationTimeout(activationId))
+          completeAppActivation(user, session, response)
+        case Success(Right(activation)) =>
+          // successful invocation
+          session.logs += activation.activationId
+          session.duration += activation.duration.getOrElse(activation.end.toEpochMilli - activation.start.toEpochMilli)
```
rabbah: a comment that duration must be defined here; (it's only undefined for triggers/rules).
probably should move this to the WhiskActivation class (as fyi, we didn't always store duration in the document, so the option on the duration was to permit for migration of schema).

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+
+          val result = activation.resultAsJson
+
+          // extract params from result
+          val params = result.getFields(WhiskActivation.paramsField).headOption.map { p =>
```
rabbah: should be same as 
` +          val params = result.fields.get(WhiskActivation.paramsField).map { p => `

here and elsewhere.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+            Try(p.asJsObject).getOrElse(JsObject(WhiskActivation.valueField -> p)) // ensure params is a dictionary
+          }
+
+          // update session state
+          session.state = result.getFields(WhiskActivation.stateField).headOption.flatMap { p =>
+            Try(Some(p.asJsObject)).getOrElse(None)
```
rabbah: hmm, if the state isn't a json object, should we just treat it as an error and abort vs continuing with no state?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+        case Success(Right(activation)) =>
+          // successful invocation
+          session.logs += activation.activationId
+          session.duration += activation.duration.getOrElse(activation.end.toEpochMilli - activation.start.toEpochMilli)
+
+          val result = activation.resultAsJson
+
+          // extract params from result
+          val params = result.getFields(WhiskActivation.paramsField).headOption.map { p =>
+            Try(p.asJsObject).getOrElse(JsObject(WhiskActivation.valueField -> p)) // ensure params is a dictionary
```
rabbah: it is likely cheaper to case match on p rather than try/catch on the cast exception.

rabbah: this auto boxing is worth documenting, here and in the docs.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
rabbah: or, why not box the state the way the value is boxed above?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+          }
+
+          // update session state
+          session.state = result.getFields(WhiskActivation.stateField).headOption.flatMap { p =>
+            Try(Some(p.asJsObject)).getOrElse(None)
+          }
+
+          // extract next action from result and invoke
```
rabbah: a nit... long methods are hard(er) to review :)

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+          result.getFields(WhiskActivation.actionField).headOption match {
+            case None =>
+              // no next action, end composition execution, return to caller
+              val response = ActivationResponse(activation.response.statusCode, Some(params.getOrElse(result)))
```
rabbah: why is the response result `params` else `result` and not just result?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+                  // resolve and invoke next action
+                  val fqn = (if (next.defaultPackage) EntityPath.DEFAULT.addPath(next) else next)
+                    .resolveNamespace(user.namespace)
+                    .toFullyQualifiedEntityName
+                  val resource = Resource(fqn.path, Collection(Collection.ACTIONS), Some(fqn.name.asString))
+                  entitlementProvider.check(user, Privilege.ACTIVATE, Set(resource), noThrottle = true).onComplete {
+                    case Failure(t) =>
+                      // failed entitlement check
+                      val response =
+                        ActivationResponse.applicationError(compositionComponentNotAccessible(next.toString))
```
rabbah: s/toString/asString/ which is defined on a fqn and preferred as the canonical string serializer.

rabbah: the entitlement check should fail with a RejectRequest exception which contains a reason for the failure. Did you consider use that instead of a new error message (which I'm ok with - just curious if you considered it).

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+                      completeAppActivation(user, session, response)
+                    case Success(_) =>
+                      // successful entitlement check
+                      WhiskActionMetaData.resolveActionAndMergeParameters(entityStore, fqn).onComplete {
```
rabbah: i hope all these failure modes are tested :)

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+                        case Failure(t) =>
+                          // resolution failure
+                          val response =
+                            ActivationResponse.applicationError(compositionComponentNotFound(next.toString))
```
rabbah: s/toString/asString.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+                  entitlementProvider.check(user, Privilege.ACTIVATE, Set(resource), noThrottle = true).onComplete {
+                    case Failure(t) =>
+                      // failed entitlement check
+                      val response =
+                        ActivationResponse.applicationError(compositionComponentNotAccessible(next.toString))
+                      completeAppActivation(user, session, response)
+                    case Success(_) =>
+                      // successful entitlement check
+                      WhiskActionMetaData.resolveActionAndMergeParameters(entityStore, fqn).onComplete {
+                        case Failure(t) =>
```
rabbah: Here, the resolve and merge operation if it fails should be an instance of an ArtifactStore exception, which may be a user error (action doesn't exist), or possibly an internal error (db connection partition). So some of these errors should be disambiguated for whisk errors.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+                          // resolution failure
+                          val response =
+                            ActivationResponse.applicationError(compositionComponentNotFound(next.toString))
+                          completeAppActivation(user, session, response)
+                        case Success(next) =>
+                          if (session.accounting.components >= actionSequenceLimit) {
```
rabbah: hmm should you just do this check before you fetch the action since it will just be rejected anyway?
would avoid the entitlement check, etc.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+   * @param session the session object for this composition
+   * @param transid a transaction id for logging
+   */
+  private def invokeComponent(user: Identity, action: WhiskActionMetaData, payload: Option[JsObject], session: Session)(
+    implicit transid: TransactionId) {
+
+    val exec = action.toExecutableWhiskAction
+    if (action.annotations.get("conductor").isDefined && exec.isDefined) {
+      // composition
+      invokeComposition(user, exec.head, payload, None, Some(session.activationId), Some(session)) // non-blocking
```
rabbah: s/head/get

rabbah: maybe this is clearer and avoids the gets.
```
action.toExecutableWhiskAction match {
   case Some(x) if !conductor => invoke simple action
   case Some(x) => invoke conductor
   case None => invoke sequence
}
```

rabbah: 🙏 for clarity, use argument names in the method calls, particularly when using None - makes it easier to remember what value that is for.

rabbah: have to go back and look, is the conductor accounted for in invoke composition?
edit: yes it is, perhaps a comment, since the component accounting is increased below - for clarity.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+            Some(action.limits.timeout.duration + 1.minute),
+            Some(session.activationId))
+        }
+        .getOrElse {
+          // sequence
+          val SequenceExecMetaData(components) = action.exec
+          invokeSequence(user, action, components, payload, None, Some(session.activationId), false, 0).map(r => r._1)
+        }
+        .onComplete {
+          case Failure(t) =>
```
rabbah: this blobs seems repetitive with a previous on complete - can this be consolidated?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+  private def completeAppActivation(user: Identity, session: Session, response: ActivationResponse)(
+    implicit transid: TransactionId): Unit = {
+
+    // compute max memory
+    val sequenceLimits = Parameters(
+      WhiskActivation.limitsAnnotation,
+      ActionLimits(session.action.limits.timeout, MemoryLimit(session.maxMemory MB), session.action.limits.logs).toJson)
+
+    // set causedBy if not topmost
+    val causedBy = if (session.cause.isDefined) {
```
rabbah: this is just a map on session.cause.map(...) simplifies the if/else some/none.

tests/src/test/scala/system/basic/WskConductorTests.scala
rabbah: was thinking to avoid hardcoding the strings for the codes... separate pr, affects a few places.

docs/compositions.md
```diff
@@ -4,38 +4,44 @@ Action compositions make it possible to dynamically build and invoke a series of
 
 ## Example
 
-Suppose we define an _increment_ action:
```
rabbah: should we call this more formally Action Continuations?

docs/compositions.md
```diff
@@ -4,38 +4,44 @@ Action compositions make it possible to dynamically build and invoke a series of
...
     }
 }
-^D
+```
+
+We create the composition action:
 
-$ wsk action create composition composition.js -a conductor true
+```
+wsk action create composition composition.js -a conductor true
```
rabbah: since the examples are using the value `true`, i think there is a reasonable expectation that the value matters.
hence, the reason for the suggested change in the controller logic.

docs/compositions.md
```diff
@@ -4,38 +4,44 @@ Action compositions make it possible to dynamically build and invoke a series of
...
 ```
 
 The key to making this action a composition is the _conductor_ annotation, which we discuss in the next section.
 
 This example composition executes two _increment_ actions in a sequence:
 
 ```
-$ wsk action invoke composition -br -p value 3
+wsk action invoke composition -br -p value 3
```
rabbah: -r is the same as -br, the -b is implied.

docs/compositions.md
```diff
@@ -73,11 +79,18 @@ For each conductor invocation (except for the first one) the _state_ produced by
...
-$ wsk action invoke composition -p value 3
+wsk action invoke composition -p value 3
+```
+```
 ok: invoked /_/composition with id 4f91f9ed0d874aaa91f9ed0d87baaa07
-
-$ wsk activation get 4f91f9ed0d874aaa91f9ed0d87baaa07
+```
+```
+wsk activation get 4f91f9ed0d874aaa91f9ed0d87baaa07
```
rabbah: you can avoid this and simplify the doc perhaps using `wsk activation get --last`
so you dont have to show generic ids, and makes it easier for someone to just copy/paste/run the commands.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: In practice, I use the `conductor` annotation to annotate the action with the json object encoding the composition, so it is not a Boolean value. I could check that the annotation value is thruthy (not null, zero, '', or undefined).

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+   * A mutable session object to keep track of the execution of one composition.
+   *
+   * NOTE:
+   * The session object is not shared between callers and callees.
+   * A callee has a reference to the session object for the caller.
+   * This permits the callee to return to the caller when done.
+   *
+   * @param activationId the activationId for the composition (ie the activation record for the composition)
+   * @param start the start time for the composition
+   * @param action the conductor action responsible for the execution of the composition
```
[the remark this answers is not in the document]
tardieu: A composition is not permitted to have more than one conductor.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+   *
+   * @param user the identity invoking the action
+   * @param payload the dynamic arguments for the activation
+   * @param session the session object for this composition
+   * @param transid a transaction id for logging
+   */
+  private def invokeConductor(user: Identity, payload: Option[JsObject], session: Session)(
+    implicit transid: TransactionId) = {
+
+    if (session.accounting.conductors > 2 * actionSequenceLimit) {
```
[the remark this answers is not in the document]
tardieu: I don't understand what you mean by static value. The maximum session length is configurable using the same configuration variable as the maximum sequence length. The limit cannot be enforced statically as the length of the composition is not known statically. The factor 2 here is somewhat arbitrary and chosen so that the limit on composition length is primarily enforced by counting the number of composed actions (as opposed to conductor actions) but to make sure nevertheless that no infinite recursion is possible. See later test in code: `if (session.accounting.components >= actionSequenceLimit)`.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: I'll add a comment.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: OK

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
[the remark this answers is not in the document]
rabbah: I mean lift the multiplication out to a static field. Separately we might want a different configuration parameter for continuations vs sequences to remove the 2x.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
rabbah: Truthy 👍

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: I'll add a comment.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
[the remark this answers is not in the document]
tardieu: I have two tests: `if (session.accounting.conductors > 2 * actionSequenceLimit)` and `if (session.accounting.components >= actionSequenceLimit)`. Should these two limits by independent additional configuration parameters?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: OK will fix throughout.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: I don't want to restrict the output schema of component actions. For instance, a component action should be permitted to return `{ action: 'foo' }` or `{ state: 'OK' }`. Hence, a composition should be permitted to return `{ action: 'foo' }` without triggering the execution of action `foo`. This is made possible by the code above (`params` case). A conductor may return `{ action: 'foo' }` to trigger the execution of action `foo` or `{ params: { action: 'foo' } }` to return `{ action: 'foo' }` without further processing. The default case (`result` case) is intended to handle error responses from conductor actions.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: There are tests for `compositionIsTooLong`, `compositionComponentInvalid`, `compositionComponentNotFound`, and `compositionComponentNotAccessible` errors. These are user/application errors.

There are no tests for `compositionActivationFailure` and `compositionActivationTimeout` errors which are internal errors. I did not find tests for the corresponding sequence failure scenarios (`sequenceActivationFailure`, `sequenceRetrieveActivationTimeout`).

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: OK... but why?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: Good point. Will fix.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+   * @param action the component action to invoke
+   * @param payload the dynamic arguments for the activation
+   * @param session the session object for this composition
+   * @param transid a transaction id for logging
+   */
+  private def invokeComponent(user: Identity, action: WhiskActionMetaData, payload: Option[JsObject], session: Session)(
+    implicit transid: TransactionId) {
+
+    val exec = action.toExecutableWhiskAction
+    if (action.annotations.get("conductor").isDefined && exec.isDefined) {
```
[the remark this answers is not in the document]
tardieu: OK for thruty test.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: OK for match.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -139,6 +199,356 @@ protected[actions] trait PrimitiveActions {
...
+            user,
+            exec,
+            payload,
+            Some(action.limits.timeout.duration + 1.minute),
+            Some(session.activationId))
+        }
+        .getOrElse {
+          // sequence
+          val SequenceExecMetaData(components) = action.exec
+          invokeSequence(user, action, components, payload, None, Some(session.activationId), false, 0).map(r => r._1)
```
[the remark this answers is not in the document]
tardieu: OK

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: OK

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
rabbah: code base avoids using toString to do type conversions to string and uses the `asX` idiom instead; leaving `toString` primarily for logging. In many cases the two are equal but for consistency, use `asString`.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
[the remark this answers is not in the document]
tardieu: This was inspired by the `storeSequenceActivation` code in the sequence implementation.
```
private def storeSequenceActivation(activation: WhiskActivation)(implicit transid: TransactionId): Unit = {
    logging.debug(this, s"recording activation '${activation.activationId}'")
    WhiskActivation.put(activationStore, activation)(transid, notifier = None) onComplete {
```

I will drop the log entry.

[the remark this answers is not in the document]
rabbah: drop or change to debug level to match recent log level tweaks.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: OK for boxing.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: OK (x2)

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: a bit shorter now...

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: You mean doing this?

```javascript
case Failure(t: RejectRequest) => ActivationResponse(t.code.intValue, Some(t.toString.toJson))
```

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: I refactored the code.

docs/compositions.md
tardieu: done

docs/compositions.md
tardieu: done

docs/compositions.md
tardieu: done

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -140,6 +200,387 @@ protected[actions] trait PrimitiveActions {
...
+   *
+   * NOTE:
+   * A sequence invocation counts as one invocation irrespective of the number of action invocations in the sequence.
+   * If one component of a composition is also a composition, the caller and callee share the same accounting object.
+   * The counts are shared between callers and callees so the limit applies globally.
+   *
+   * @param components the current count of component actions already invoked
+   * @param conductors the current count of conductor actions already invoked
+   */
+  private case class CompositionAccounting(var components: Int = 0, var conductors: Int = 0)
```
markusthoemmes: Why is the mutability needed? Could we keep the object immutable and create copies instead? In general that's much more readable and much easier to follow, flow-wise.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -140,6 +200,387 @@ protected[actions] trait PrimitiveActions {
...
+   * @param caller the session object for the parent composition (caller) if any
+   * @param result a placeholder for returning the result of the composition invocation (blocking invocation only)
+   */
+  private case class Session(activationId: ActivationId,
+                             start: Instant,
+                             action: ExecutableWhiskActionMetaData,
+                             cause: Option[ActivationId],
+                             var duration: Long,
+                             var maxMemory: Int,
+                             var state: Option[JsObject],
```
markusthoemmes: Same comment on mutability.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -140,6 +200,387 @@ protected[actions] trait PrimitiveActions {
...
+                             accounting: CompositionAccounting,
+                             logs: Buffer[ActivationId],
+                             caller: Option[Session],
+                             result: Option[Promise[Either[ActivationId, WhiskActivation]]])
```
markusthoemmes: I'll put the comment here, but the concern is around the other functions as well: Do you need to pass a Promise around vs. chaining all actions together using Futures?

When glancing over the code it was very hard to tell for me, where this promise is used and which of the other methods does which side-effects. They all return `Unit`. There is no way for me to safely tell what is going on if I do not read **all** of the source-code.

I believe this is a perfect use-case to make use of `Future`s and chain the methods together and I think the code will very much profit from that. It'll make it virtually impossible to miss cases and/or have dead-ends in the code.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
rabbah: I considered this more - I think your error is adequate. We can reconsider in the future. I don't think we want to leak the `code` here since it's an HTTP Status code which doesn't make sense here. The message is an option, so you'd have to map on that and get either the custom message or the default message for the `code` otherwise. I think this doesn't make the error reporting any more meaningful.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: Thanks to mutability, it is clear that there is only one CompositionAccounting object ever allocated in the course of the execution of a composition. Therefore there is only one set of counters. With object copies on the other hand, it is hard to keep track of the copies. Why are copies being made, where, when? Are the multiple sets of counters concurrently valid and useful? Moreover, immutability does not prevent forking the object, hence concurrency issues.

Give me ownership types and strong updates... and then I buy your argument. But without aliasing control, I am afraid we are discussing style more than safety.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: Same answer. I want one Session object for one session. I believe this one to one mapping makes the code easier to follow. Part of what constitutes a session-its state-changes over the course of the execution, so the corresponding fields of the object are mutable.

If I made this class immutable and used object copies instead as suggested, then the distinction between immutable session elements and mutable session elements would disappear from the type and become buried in the code and comments. I would argue this obfuscates intend and hinder readability.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: Ok I'll replace the promise with a future.

tardieu: Pushed.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -140,6 +200,362 @@ protected[actions] trait PrimitiveActions {
...
+          result.fields.get(WhiskActivation.actionField) match {
+            case None =>
+              // no next action, end composition execution, return to caller
+              Future.successful(ActivationResponse(activation.response.statusCode, Some(params.getOrElse(result))))
+            case Some(next) =>
+              Try(next.convertTo[EntityPath]) match {
+                case Failure(t) =>
+                  // parsing failure
+                  Future.successful(ActivationResponse.applicationError(compositionComponentInvalid(next)))
+                case Success(next) =>
```
rabbah: could save a level of nesting by doing this:

```
               case Success(next) if (session.accounting.components >= actionSequenceLimit) =>
                    Future.successful(ActivationResponse.applicationError(compositionIsTooLong)) 
               case Success(next) => the rest
```

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -140,6 +200,362 @@ protected[actions] trait PrimitiveActions {
...
+      invokeConductor(user, payload, session).map(response => Right(completeActivation(user, session, response)))
+
+    // is caller waiting for the result of the activation?
+    if (cause.isDefined) {
+      // ignore waitForResponse when not topmost
+      response
+    } else {
+      waitForResponse map { timeout =>
+        // blocking invoke, wait until timeout
+        response.withAlternativeAfterTimeout(waitForResponse.get, Future.successful(Left(session.activationId)))
```
rabbah: isn't `waitForResponse.get == timeout`?

tardieu: yes! will fix

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: yes!

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -140,6 +200,362 @@ protected[actions] trait PrimitiveActions {
...
+        Parameters(WhiskActivation.pathAnnotation, JsString(session.action.fullyQualifiedName(false).asString)) ++
+        Parameters(WhiskActivation.kindAnnotation, JsString(Exec.SEQUENCE)) ++
+        Parameters(WhiskActivation.conductorAnnotation, JsBoolean(true)) ++
+        causedBy ++
+        sequenceLimits,
+      duration = Some(session.duration))
+
+    logging.debug(this, s"recording activation '${activation.activationId}'")
+    WhiskActivation.put(activationStore, activation)(transid, notifier = None) onComplete {
+      case Success(id) => logging.info(this, s"recorded activation")
```
rabbah: logging.debug; to follow in general log reduction.

rabbah: or just drop it entirely, it's logged in the dbstore anyway.

tardieu: I am confused. The code already invokes `logging.debug`. Should this change?

rabbah: line 549:
 +      case Success(id) => logging.info(this, s"recorded activation")

tardieu: OK I see it.

tardieu:
I agree that the code with a future is simpler than the code with a promise so I am ok with the change. But, FWIW, there was a decent reason to use the more complex concept that I try to summarize below (before I forget).

The future-based implementation is now similar to the sequence implementation. It creates one pending future (ie pending continuation) for each level of nesting (sequence inside a sequence, composition inside a composition). In contrast the promise-based implementation only requires one pending continuation irrespective of nesting. It used instead a call stack (now removed caller field in session object) and unwinds the stack upon termination of a nested composition. Concretely, the callee resumes the execution of the caller explicitly with the promise implementation, whereas with the future implementation all the stack levels simultaneously exist as pending tasks in the scheduler. In short, I believe the future implementation is by nature more expensive. But since we cap the recursion depth to a low value, it is most likely irrelevant in practice.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -140,6 +200,349 @@ protected[actions] trait PrimitiveActions {
...
+          session.logs += activationId
+          Left(ActivationResponse.whiskError(compositionActivationTimeout(activationId)))
+        case Right(activation) => // successful invocation
+          session.logs += activation.activationId
+          // activation.duration should be defined but this is not reflected by the type so be defensive
+          session.duration += activation.duration.getOrElse(activation.end.toEpochMilli - activation.start.toEpochMilli)
+          activation.annotations.get("limits").foreach { limitsAnnotation =>
+            limitsAnnotation.asJsObject.getFields("memory") match {
+              case Seq(JsNumber(memory)) =>
+                session.maxMemory = Math.max(session.maxMemory, memory.toInt)
```
markusthoemmes: I believe this does not ignore the "not a sequence of number" cases but throws a `MatchException`.

You're looking for:

```scala
activation.annotations.get("limits")
  .flatMap(_.asJsObject.fields.get("memory"))
  .flatMap(memory => Try(memory.convertTo[Int]).toOption)
  .foreach(memory => session.maxMemory = memory max session.maxMemory
```

To be relatively sure it doesn't throw (asJsObject can still throw)

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -140,6 +200,349 @@ protected[actions] trait PrimitiveActions {
...
+
+    activationResponse
+      .map {
+        case Left(activationId) => // invocation timeout
+          session.logs += activationId
+          Left(ActivationResponse.whiskError(compositionActivationTimeout(activationId)))
+        case Right(activation) => // successful invocation
+          session.logs += activation.activationId
+          // activation.duration should be defined but this is not reflected by the type so be defensive
+          session.duration += activation.duration.getOrElse(activation.end.toEpochMilli - activation.start.toEpochMilli)
```
markusthoemmes: Maybe go for `end` - `start` straight away to avoid confusion?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -140,6 +200,349 @@ protected[actions] trait PrimitiveActions {
...
+   * @param accounting the global accounting object used to abort compositions requiring too many action invocations
+   * @param logs a mutable buffer that is appended with new activation ids as the composition unfolds
+   *             (in contrast with sequences, the logs of a hierarchy of compositions is not flattened)
+   */
+  private case class Session(activationId: ActivationId,
+                             start: Instant,
+                             action: ExecutableWhiskActionMetaData,
+                             cause: Option[ActivationId],
+                             var duration: Long,
+                             var maxMemory: Int,
```
markusthoemmes: Should this be a `ByteSize`?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -140,6 +200,349 @@ protected[actions] trait PrimitiveActions {
...
+   * @param state the json state object to inject in the parameter object of the next conductor invocation
+   * @param accounting the global accounting object used to abort compositions requiring too many action invocations
+   * @param logs a mutable buffer that is appended with new activation ids as the composition unfolds
+   *             (in contrast with sequences, the logs of a hierarchy of compositions is not flattened)
+   */
+  private case class Session(activationId: ActivationId,
+                             start: Instant,
+                             action: ExecutableWhiskActionMetaData,
+                             cause: Option[ActivationId],
+                             var duration: Long,
```
markusthoemmes: Should this be a `FiniteDuration`?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: I am not sure I understand the suggestion here: `duration` is not equal to `end - start` and we have to use the correct value. I'll update the comment to be more clear.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: No I don't think so. This current type follows from:

```scala
case class WhiskActivation(namespace: EntityPath,
                           ...
                           duration: Option[Long] = None)
```

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tardieu: The `memory` annotation in activation records is generated. If the annotation is present without the proper type, we should get an internal error. `MatchException` does that. I don't think silently ignoring the error makes more sense. We might want to have a more specific error message but I don't think this is needed either. The current code follows directly from the SequenceActions implementation.

docs/conductors.md
```diff
@@ -0,0 +1,253 @@
...
+
+We create a _conductor_ action by specifying the _conductor_ annotation:
+
+```
+wsk action create tripleAndIncrement tripleAndIncrement.js -a conductor true
+```
+
+A _conductor action_ is an action with a _conductor_ annotation with a value that is not _falsy_, i.e., a value that is different from zero, null, false, and the empty string.
+
+At this time, sequence actions cannot be conductor actions. The conductor annotation on sequence actions is ignored.
```
rabbah: should we phrase this differently? sequence actions are implicit conductors, and adding a conductor annotation has no effect/is ignored?

docs/conductors.md
```diff
@@ -0,0 +1,253 @@
...
+
+- step 0: invoke the _triple_ action on the input dictionary,
+- step 1: invoke the _increment_ action on the output dictionary from step 1,
+- step 2: return the output dictionary from step 2.
+
+At each step, the conductor action specifies how to continue or terminate the execution by means of a _continuation_. We explain continuations after discussing invocation and activations.
+
+## Invocation
+
+A conductor action is invoked like a regular action, for instance:
```
rabbah: perhaps link to actions.md section on working with actions?

docs/conductors.md
```diff
@@ -0,0 +1,253 @@
...
+a1f58ade9b1e4c26b58ade9b1e4c2614 triple
+3624ad829d4044afa4ad829d40e4af60 tripleAndIncrement
+4f91f9ed0d874aaa91f9ed0d87baaa07 tripleAndIncrement
+```
+
+There are six activations in this example, one for the _tripleAndIncrement_ action we invoked plus five additional activations _caused_ by this invocation. These five activations are:
+
+- one activation of the _triple_ action with input `{ value: 3 }` and output `{ value: 9 }`,
+- one activation of the _increment_ action with input `{ value: 9 }` and output `{ value: 10 }`,
+- three _secondary_ activations of the _tripleAndIncrement_ action.
```
rabbah: this is not immediately clear:
```
+3624ad829d4044afa4ad829d40e4af60 tripleAndIncrement
+4f91f9ed0d874aaa91f9ed0d87baaa07 tripleAndIncrement
```

`4f9` is the outermost activation record, whereas `362` is the activation of the conductor action itself. This is a little subtle and perhaps warrants a bit more explanation.

rabbah: i see it below - perhaps foreshadow this explanation then.

docs/conductors.md
```diff
@@ -0,0 +1,253 @@
...
+ 2. _triple_ activation,
+ 3. secondary _tripleAndIncrement_ activation,
+ 4. _increment_ activation,
+ 5. secondary _tripleAndIncrement_ activation.
+
+Intuitively, secondary activations of the conductor action decide which component actions to invoke by running before, in-between, and after the component actions.
+
+Only an internal error (invocation failure or timeout) may result in an even number of derived activations.
+
+### Primary activation
```
rabbah: plural.

docs/conductors.md
```diff
@@ -0,0 +1,253 @@
...
+
+While secondary activations of the conductor action reflect executions of the action code, the primary activation record for the invocation of a conductor action does not. It is a synthetic record similar to the activation record of a sequence action. Concretely in our example, the code in the main function of the _tripleAndIncrement_ conductor action runs exactly three times since there are three secondary activations.
+
+The primary activation record summarizes the series of derived activations:
```
rabbah: suggestion: flip primary to come first, secondary section to come second.

docs/conductors.md
```diff
@@ -0,0 +1,253 @@
...
+
+On the other hand, if the _action_ field is not defined in the output of the conductor action, the conductor action invocation ends. The output for the conductor action invocation is either the value of the _params_ field in the output dictionary of the last secondary activation if defined (auto boxed if necessary) or if absent the complete output dictionary.
+
+## Limits
+
+There are limits on the number of component action activations and secondary conductor activations in a conductor action invocation. These limits are assessed globally, i.e., if some components of a conductor action invocation are themselves conductor actions, the limits apply to the combined counts of activations across all the conductor action invocations.
+
+The maximum number _n_ of permitted component activations is equal to the maximum number of components in a sequence action. It is configured via the same configuration parameter. The maximum number of secondary conductor activations is _2n+1_.
```
rabbah: we should add this to references.md (i dont think we document sequence limit from our past discussion).
we can do that separately.

tardieu: ok

docs/conductors.md
tardieu: ok

docs/conductors.md
tardieu: ok

docs/conductors.md
tardieu: tweaking...

PR #3206, opened 2018-01-19
Refactor controller role to use variables for hostname and host index

dubeejw:
- Define variables for frequently referenced operations in controller role
- Move controller.jmxremote.jvmArgs to deploy.yml

### DIFF
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -3,6 +3,14 @@
 
 - include: docker_login.yml
 
+- name: get controller index
+  set_fact:
+    controller_index: "{{ groups['controllers'].index(inventory_hostname) }}"
+
+- name: get controller name
+  set_fact:
+    controller_name: "controller{{ controller_index }}"
+
 - name: "pull the {{ docker.image.tag }} image of controller"
   shell: "docker pull {{ docker_registry }}{{ docker.image.prefix }}/controller:{{ docker.image.tag }}"
   when: docker_registry != ""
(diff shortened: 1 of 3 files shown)


dubeejw:
@cbickel, can you look at this one when you have a chance? It's just refactoring the controller deployment scripts to make them easier to manage/read.

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -48,17 +56,17 @@
 
 - name: prepare controller ports
   set_fact:
-    ports_to_expose: ["{{ controller.basePort + groups['controllers'].index(inventory_hostname) }}:8080", "{{ controller.akka.cluster.basePort + groups['controllers'].index(inventory_hostname) }}:{{ controller.akka.cluster.bindPort }}"]
+    ports_to_expose: ["{{ controller.basePort + (controller_index | int) }}:8080", "{{ controller.akka.cluster.basePort + (controller_index | int) }}:{{ controller.akka.cluster.bindPort }}"]
```
cbickel: Is it possible to do the `int` conversation already on defining the variable? Or would this result in issues on gernerating the name-String?

dubeejw: @cbickel, unfortunately defining the `controller_index` as an `int` results in the variable type being `unicode`. See below.

Define `controller_index` as an int:
```
- name: get controller index
  set_fact:
    controller_index: "{{ groups['controllers'].index(inventory_hostname) | int }}"
```

Attempting to add `controller_index` to an int:
```
{{ controller.basePort + controller_index }}
```

Results in this error:
```
fatal: [192.168.99.100]: FAILED! => {"failed": true, "msg": "Unexpected templating type error occurred on ({{ controller.basePort + controller_index }}:8080): unsupported operand type(s) for +: 'int' and 'unicode'"}
```

Apparently using `set_fact` always produces a string: https://github.com/ansible/ansible/issues/4170.

dubeejw:
PG2 2754 ⌛️ 
PG5 64 ⌛️

dubeejw:
PG3 1821 🔵

dubeejw:
@cbickel, PG looks good.

PR #3209, opened 2018-01-22
Make payload limit configurable.

markusthoemmes:
The maximum payload allowed to flow through the system is hardcoded today. Several pieces (kafka for instance) need to be adjusted to allow a larger payload.

**Note:** After travis passes happily, the default can be reset to 1 MB to not change anything here.

### DIFF
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -53,6 +53,12 @@ whisk {
     # kafka related configuration
     kafka {
         replication-factor = 1
+
+        producer {
+            acks = 1
+            max-request-size = ${whisk.activation.payload.max}
+        }
+
         topics {
             cache-invalidation {
                 segment-bytes   =  536870912
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -70,9 +70,10 @@ whisk {
                 retention-ms    = 3600000
             }
             invoker {
-                segment-bytes   =  536870912
-                retention-bytes = 1073741824
-                retention-ms    = 172800000
+                segment-bytes     =  536870912
+                retention-bytes   = 1073741824
+                retention-ms      =  172800000
+                max-message-bytes = ${whisk.activation.payload.max}
             }
         }
     }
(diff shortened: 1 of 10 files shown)


common/scala/src/main/resources/application.conf
```diff
@@ -70,9 +70,10 @@ whisk {
                 retention-ms    = 3600000
             }
             invoker {
-                segment-bytes   =  536870912
-                retention-bytes = 1073741824
-                retention-ms    = 172800000
+                segment-bytes     =  536870912
+                retention-bytes   = 1073741824
+                retention-ms      =  172800000
+                max-message-bytes = ${whisk.activation.payload.max}
```
rabbah: Why is this needed?

common/scala/src/main/resources/application.conf
```diff
@@ -88,6 +89,12 @@ whisk {
         stride = ${?CONTROLLER_INSTANCES}
     }
 
+    activation {
+        payload {
+            max = 5242880 // 5 m not possible because cross-referenced to kafka configurations
```
rabbah: Vs just this.

common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
```diff
@@ -87,6 +87,9 @@ class KafkaProducerConnector(kafkahosts: String,
     val props = new Properties
     props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, kafkahosts)
     props.put(ProducerConfig.ACKS_CONFIG, 1.toString)
+    // Allow messages with up to 5 MiB size
+    // The target topic must support this message size as well because it's larger as the default
+    props.put(ProducerConfig.MAX_REQUEST_SIZE_CONFIG, 5242880.toString)
```
rabbah: Should the 5M also be a setting here from pure config ?

common/scala/src/main/resources/application.conf
markusthoemmes: Wanted to seperate kafka and general config, so the value is carried over by configuration not by hardcoding.

I also changed the topic's configurations to be generalized as in: You can provide all possible Kafka keys here and they are transformed to the respective Kafka string.

common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
markusthoemmes: Presumably, yes. Need to thing how to cleverly thread that through 👍

markusthoemmes:
PG2 2692 🔵

PR #3211, opened 2018-01-23
implement ContainerFactory.cpuShare to fix #3110

### LINKED ISSUE #3110: INVOKER_CORESHARE overloaded?

tysonnorris:
added test for DockerContainerFactory to verify proper args in docker run
fixes #3110

### DIFF
--- a/ansible/roles/invoker/tasks/deploy.yml
+++ b/ansible/roles/invoker/tasks/deploy.yml
@@ -192,9 +192,9 @@
         {% for item in (invoker_container_network_dns_servers | default()).split(' ')  %}
         -e CONFIG_whisk_containerFactory_containerArgs_dnsServers_{{loop.index0}}={{ item }}
         {% endfor %}
-        -e INVOKER_NUMCORE='{{ invoker.numcore }}'
-        -e INVOKER_CORESHARE='{{ invoker.coreshare }}'
-        -e INVOKER_USE_RUNC='{{ invoker.useRunc }}'
+        -e CONFIG_whisk_containerPool_numCore='{{ invoker.numcore }}'
+        -e CONFIG_whisk_containerPool_coreShare='{{ invoker.coreshare }}'
+        -e CONFIG_whisk_docker_containerFactory_useRunc='{{ invoker.useRunc }}'
         -e INVOKER_NAME='{{ groups['invokers'].index(inventory_hostname) }}'
         -e WHISK_LOGS_DIR='{{ whisk_logs_dir }}'
         -e METRICS_KAMON='{{ metrics.kamon.enabled }}'
--- a/common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
@@ -31,6 +31,23 @@ case class ContainerArgsConfig(network: String,
                                dnsServers: Seq[String] = Seq.empty,
                                extraArgs: Map[String, Set[String]] = Map.empty)
 
+case class ContainerPoolConfig(numCore: Int, coreShare: Int) {
+
+  /**
+   * The total number of containers is simply the number of cores dilated by the cpu sharing.
+   */
+  def maxActiveContainers = numCore * coreShare
+
+  /**
+   * The shareFactor indicates the number of containers that would share a single core, on average.
+   * cpuShare is a docker option (-c) whereby a container's CPU access is limited.
+   * A value of 1024 is the full share so a strict resource division with a shareFactor of 2 would yield 512.
+   * On an idle/underloaded system, a container will still get to use underutilized CPU shares.
+   */
+  private val totalShare = 1024.0 // This is a pre-defined value coming from docker and not our hard-coded value.
+  def cpuShare = (totalShare / maxActiveContainers).toInt
+}
+
 /**
  * An abstraction for Container creation
  */
@@ -41,7 +58,8 @@ trait ContainerFactory {
                       name: String,
                       actionImage: ExecManifest.ImageName,
                       userProvidedImage: Boolean,
-                      memory: ByteSize)(implicit config: WhiskConfig, logging: Logging): Future[Container]
+                      memory: ByteSize,
+                      cpuShares: Int)(implicit config: WhiskConfig, logging: Logging): Future[Container]
 
   /** perform any initialization */
   def init(): Unit
--- a/core/invoker/src/main/resources/application.conf
+++ b/core/invoker/src/main/resources/application.conf
@@ -18,6 +18,16 @@ whisk {
     unpause: 10 seconds
   }
 
+  docker.container-factory {
+    # Use runc (docker-runc) for pause/resume functionality in DockerContainerFactory
+    use-runc: true
+  }
+
+  container-pool {
+    num-core: 4      # used for computing --cpushares, and max number of containers allowed
+    core-share: 2    # used for computing --cpushares, and max number of containers allowed
+  }
+
   kubernetes {
     # Timeouts for k8s commands. Set to "Inf" to disable timeout.
     timeouts {
@@ -42,7 +52,7 @@ whisk {
   container-factory.container-args {
     network: bridge
     dns-servers: []
-    extra-args: {}
+    extra-args: {}   # to pass additional args to 'docker run'; format is `{key1: [v1, v2], key2: [v1, v2]}`
   }
 
   container-proxy {
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
@@ -36,26 +36,28 @@ import pureconfig._
 import whisk.core.ConfigKeys
 import whisk.core.containerpool.ContainerArgsConfig
 
-class DockerContainerFactory(config: WhiskConfig,
-                             instance: InstanceId,
+case class DockerContainerFactoryConfig(useRunc: Boolean)
+
+class DockerContainerFactory(instance: InstanceId,
                              parameters: Map[String, Set[String]],
-                             containerArgs: ContainerArgsConfig =
-                               loadConfigOrThrow[ContainerArgsConfig](ConfigKeys.containerArgs))(
+                             containerArgsConfig: ContainerArgsConfig =
+                               loadConfigOrThrow[ContainerArgsConfig](ConfigKeys.containerArgs),
+                             dockerContainerFactoryConfig: DockerContainerFactoryConfig =
+                               loadConfigOrThrow[DockerContainerFactoryConfig](ConfigKeys.dockerContainerFactory))(
   implicit actorSystem: ActorSystem,
   ec: ExecutionContext,
-  logging: Logging)
+  logging: Logging,
+  docker: DockerApiWithFileAccess,
+  runc: RuncApi)
     extends ContainerFactory {
-
-  /** Initialize container clients */
-  implicit val docker = new DockerClientWithFileAccess()(ec)
-  implicit val runc = new RuncClient()(ec)
 
   /** Create a container using docker cli */
   override def createContainer(tid: TransactionId,
                                name: String,
                                actionImage: ExecManifest.ImageName,
                                userProvidedImage: Boolean,
-                               memory: ByteSize)(implicit config: WhiskConfig, logging: Logging): Future[Container] = {
+                               memory: ByteSize,
+                               cpuShares: Int)(implicit config: WhiskConfig, logging: Logging): Future[Container] = {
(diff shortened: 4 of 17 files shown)


common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
```diff
@@ -44,6 +44,26 @@ trait ContainerFactory {
 
   /** cleanup any remaining Containers; should block until complete; should ONLY be run at startup/shutdown */
   def cleanup(): Unit
+
```
tysonnorris: This comes directly from the "old" ContainerPool:
https://github.com/apache/incubator-openwhisk/blob/e47cc71bcd73ebb3da352eeadcc4236630680e92/core/invoker/src/main/scala/whisk/core/container/ContainerPool.scala#L826-L844

which was deleted in
https://github.com/apache/incubator-openwhisk/commit/10fae73e15cf886bb3ad892e659b15a28cc4c80b

markusthoemmes:
Since you're on it anyway: What about moving those to pureconfig now?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
```diff
@@ -126,6 +129,13 @@ object DockerContainerFactoryProvider extends ContainerFactoryProvider {
...
-                                   parameters: Map[String, Set[String]]): ContainerFactory =
-    new DockerContainerFactory(config, instanceId, parameters)(actorSystem, actorSystem.dispatcher, logging)
+                                   parameters: Map[String, Set[String]]): ContainerFactory = {
+
+    new DockerContainerFactory(config, instanceId, parameters, dockerApiFactory)(
+      actorSystem,
+      actorSystem.dispatcher,
+      logging)
+  }
+  def dockerApiFactory(actorSystem: ActorSystem, logging: Logging) =
```
markusthoemmes: Why pull this out to a seperate function?

tysonnorris: wasn't sure how else to make it testable - as a mock https://github.com/apache/incubator-openwhisk/pull/3211/files#diff-16241f60cd0658f397e5c38d57963c95R63

Will take another look to see if I can simplify it a bit

tysonnorris: ok, removed the function; the args on DockerContainerFactory are a little out of control, but maybe this is a bit simpler

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerContainerFactoryTests.scala
```diff
@@ -0,0 +1,120 @@
...
+    //setup inspect expectation
+    (dockerApiStub
+      .inspectIPAddress(_: ContainerId, _: String)(_: TransactionId))
+      .expects(ContainerId("fakecontainerid"), "net1", *)
+      .returning(Future.successful { ContainerAddress("1.2.3.4", 1234) })
+    //setup rm expectation
+    (dockerApiStub
+      .rm(_: ContainerId)(_: TransactionId))
+      .expects(ContainerId("fakecontainerid"), *)
+      .returning(Future.successful(Unit))
```
markusthoemmes: I kinda wanted to get rid of `ScalaMock`. I found it super non-intuitive to use. What's your take on this?

In this specific case, I used self-rolled mocking (see `DockerContainerTests`) which I found much simpler to grasp.

ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -173,8 +173,8 @@
         {% for item in (invoker_container_network_dns_servers | default()).split(' ')  %}
         -e CONFIG_whisk_containerFactory_containerArgs_dnsServers_{{loop.index0}}={{ item }}
         {% endfor %}
-        -e INVOKER_NUMCORE='{{ invoker.numcore }}'
-        -e INVOKER_CORESHARE='{{ invoker.coreshare }}'
+        -e CONFIG_whisk_containerFactory_containerArgs_numCore='{{ invoker.numcore }}'
+        -e CONFIG_whisk_containerFactory_containerArgs_coreShare='{{ invoker.coreshare }}'
```
markusthoemmes: Hmm, should these live under container-args? Feels like they should live on `whisk.containerpool` instead?

tysonnorris: I'm not sure why - these are used primarily to set the `--cpushares` flag for docker run; it is from that perspective that the container pool also uses it to define "maximumContainers" (to define the pool size). 

FWIW in the future, I will be creating a PR to isolate these 2 items (cpushare vs maximumContainers) , since it does not make sense for cluster-aware deployments where maximumContainers value can fluctuate dynamically, and is not correlated to only the setting used for `--cpushares`.

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerContainerFactoryTests.scala
tysonnorris: I'm not sure any mocking framework is intuitive, but I prefer it over manually creating impls. One thing that I find helpful is the default behavior where "extra" calls to stub methods are not tolerated, so it's pretty simple to make sure you are executing exactly what you expect.

I would prefer to leave ScalaMock in place.

ansible/roles/invoker/tasks/deploy.yml
markusthoemmes: Its main purpose is to determine the maximum amount of containers to spawn. Secondarily, that value is used to also determine the cpushares parameter, but the main purpose is to tell the containerpool: How many cores and how do you want to split them.

tysonnorris: Ha. Ok moving it.

tysonnorris: Oh separately from config path, there is another issue, where ContainerPoolConfig should generally reside near ContainerPool which resides in invoker, but ContainerFactory resides in common. This is a problem because non-docker ContainerFactory will require use of this config (to set the cpushares), so for now I will use `whisk.container-factory.container-pool` (not `whisk.container-pool`) to store the config.

tysonnorris: nm, I can just leave it at `whisk.container-pool`, but the config class will be in ContainerFactory. This class can move into ContainerPool in the future if/when ContainerPool moves to common.

markusthoemmes:
@tysonnorris status? Travis failed.

tysonnorris:
@markusthoemmes this is all fixed

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
```diff
@@ -67,13 +70,13 @@ class DockerContainerFactory(config: WhiskConfig,
       image = image,
       userProvidedImage = userProvidedImage,
       memory = memory,
-      cpuShares = config.invokerCoreShare.toInt,
+      cpuShares = poolConfig.cpuShare,
```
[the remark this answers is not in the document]
tysonnorris: Philosophically I'm ok with this, but I'm not sure about how to update KubernetesContainerFactory to respect this change; @dgrove-oss WDYT?

For the time being, it may be better to leave this to a separate PR, since it is changing the ContainerFactory SPI interface.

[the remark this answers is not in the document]
tysonnorris: on second thought - I'll go ahead and put this in, and if there is a simple way to update kube to respect the cpu limit, that can be added later; for now, kube will do nothing with that parameter.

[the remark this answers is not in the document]
dgrove-oss: go ahead and put it in;  I will update KubernetesContainerFactory later.  You are right, currently we are only passing in a memory limit in KubernetesContainerFactory.  However, passing in a CPU limit will be small to the change in container creation in #3338

markusthoemmes:
@rabbah I'll need a second pair of 👀 on this change to make sure nothing seriously breaks.

tysonnorris:
@rabbah let me know if you have comments on this?

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
```diff
@@ -67,13 +69,13 @@ class DockerContainerFactory(config: WhiskConfig,
...
       environment = Map("__OW_API_HOST" -> config.wskApiHost),
-      network = containerArgs.network,
-      dnsServers = containerArgs.dnsServers,
+      network = containerArgsConfig.network,
+      dnsServers = containerArgsConfig.dnsServers,
       name = Some(name),
-      useRunc = config.invokerUseRunc,
-      parameters ++ containerArgs.extraArgs)
+      useRunc = dockerContainerFactoryConfig.useRunc,
+      parameters ++ containerArgsConfig.extraArgs.map { case (k, v) => ("--" + k, v) })
```
rabbah: `v` here is an an array of strings e.g. `[v1, v2]`per the comment above; is that what's intended here?

tests/src/test/scala/whisk/core/containerpool/docker/test/DockerContainerFactoryTests.scala
rabbah: hmm if the test is just to validate the cpu share, this seems like over engineering of the test (could you have tested just the config class itself in contrast for example?).

i'm ok keeping the test as is but the test name/comment struck me relative to the actual code.

tests/src/test/scala/whisk/core/containerpool/test/ContainerArgsConfigTest.scala
```diff
@@ -52,7 +52,7 @@ class ContainerArgsConfigTest extends FlatSpec with Matchers {
     config.dnsServers shouldBe Seq[String]("google.com", "1.2.3.4")
     //check map parsing of extra-args config
     config.extraArgs.get("label") shouldBe Some(Set("l1", "l2", "l3"))
-    config.extraArgs.get("environment") shouldBe Some(Set("e1", "e2"))
+    config.extraArgs.get("env") shouldBe Some(Set("e1", "e2"))
```
rabbah: i suggest a test for extra args following the comment above, or fix the comment.

dubee:
@tysonnorris, can you rebase. What's the status of this one?

core/invoker/src/main/resources/application.conf
```diff
@@ -39,7 +49,6 @@ whisk {
   container-factory.container-args {
     network: bridge
     dns-servers: []
-    extra-args: {}
-
+    extra-args: {}   # to pass additional args to 'docker run'; format is `{key1: [v1, v2], key2: [v1, v2]`
```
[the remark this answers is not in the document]
tysonnorris: yes the value has to be an array

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainerFactory.scala
tysonnorris: yes

tests/src/test/scala/whisk/core/containerpool/test/ContainerArgsConfigTest.scala
tysonnorris: I think this is addressed now

tysonnorris:
@dubee @markusthoemmes any issues left on this?

markusthoemmes:
@tysonnorris should be good. Could you rebase one more time? I'll get this in ASAP then.

tysonnorris:
hopefully this will do it

markusthoemmes:
PG5 232 🔵

PR #3213, opened 2018-01-23
Fixes #3212 - WhiskConfig should close the property file source

### LINKED ISSUE #3212: WhiskConfig should close the property file source
`WhiskConfig` currently opens a `Source` from the property file for reading purpose but it does not close the source instance leading to it being remain open.

chetanmeh:
Close the source instance with try-finally usage


markusthoemmes:
Thanks for the fix. I feel a bit uneasy about the testcase though. Seems rather low-level. In this case I could as well live without a testcase.

Thoughts? @rabbah

rabbah:
+1, can live without a test for this - since we're not checking that we're leaking a file resource anywhere else. We could enable resource leak checks more generally.

chetanmeh:
Okie would update PR with test case removed

PR #3214, opened 2018-01-23
Removing unused param

beemarie:
This parameter was originally used to optionally skip a package when there are two provided. We refactored this recently to instead use a better (more generic) solution of passing in environment variables for a catalog, which could be used for anything. PR #3144. This additional parameter can now be removed -- it's not set anywhere other than as a default value here, nor is it used anywhere.

### DIFF
--- a/ansible/tasks/installOpenwhiskCatalog.yml
+++ b/ansible/tasks/installOpenwhiskCatalog.yml
@@ -18,7 +18,6 @@
 
 - set_fact:
     skip_catalog_install="{{ item.value.skip | default(false) }}"
-    optional_package_skip="{{ item.value.optional_package_skip | default(true) }}"
 
 - set_fact:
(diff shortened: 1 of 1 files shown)


beemarie:
🔵 PG2/2694

markusthoemmes:
Want to add a description to the PR, briefly describing why this is safe to remove please?

beemarie:
@markusthoemmes sure, sorry about that.  This parameter was originally used to optionally skip a package when there are two provided.  We refactored this recently to instead use a better (more generic) solution of passing in environment variables for a catalog, which could be used for anything. PR #3144.  This additional parameter can now be removed -- it's not set anywhere other than as a default value here, nor is it used anywhere.

csantanapr:
This was suppose to be cleaned up with #3144

PR #3215, opened 2018-01-24
adding namePrefix config to ContainerArgsConfig

tysonnorris:
This is to enable a deployment case where we briefly want to allow an invoker to run local to another running invoker; To enable this, the "cleanup" should not kill the action containers created by the previously existing invoker. As long as each invoker has a different prefix, they won't destroy each other's containers at startup.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
@@ -29,8 +29,7 @@ import whisk.spi.Spi
 
 case class ContainerArgsConfig(network: String,
                                dnsServers: Seq[String] = Seq(),
-                               extraArgs: Map[String, Set[String]] = Map(),
-                               namePrefix: String)
+                               extraArgs: Map[String, Set[String]] = Map())
 
 /**
  * An abstraction for Container creation
@@ -49,6 +48,15 @@ trait ContainerFactory {
 
   /** cleanup any remaining Containers; should block until complete; should ONLY be run at startup/shutdown */
   def cleanup(): Unit
+}
+
+object ContainerFactory {
+
+  /** include the instance name, if specified and strip invalid chars before attempting to use them in the container name */
+  def containerNamePrefix(instanceId: InstanceId): String =
+    s"wsk${instanceId.name.getOrElse("")}${instanceId.toInt}"
+      .replaceAll("[^a-zA-Z0-9_\\.\\-]", "") // based on https://github.com/moby/moby/issues/3138 and https://github.com/moby/moby/blob/master/daemon/names/names.go
+
 }
 
 /**
--- a/common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
@@ -29,7 +29,8 @@ import whisk.spi.Spi
 
 case class ContainerArgsConfig(network: String,
                                dnsServers: Seq[String] = Seq(),
-                               extraArgs: Map[String, Set[String]] = Map())
+                               extraArgs: Map[String, Set[String]] = Map(),
+                               namePrefix: String)
 
 /**
  * An abstraction for Container creation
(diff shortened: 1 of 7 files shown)


rabbah:
Are both invokers using different instance ids? If so the existing prefix would address this I think. I must be missing something.

tysonnorris:
They are using dynamic instance ids, different invoker names, and different kafka brokers... so the instance ids will often be the same (but not guaranteed, I think). This gives us a way to guarantee the name (or specifically the prefix, which is used for cleanup purposes) is unique.

markusthoemmes:
Can you not already guarantee that by passing a unique id? If you need to generate something unique anyway?

tysonnorris:
We cannot pass a unique id, but we can pass a unique `--name` arg, which results in a non-unique id.

dgrove-oss:
Ideally this is exactly what separating invoker_name from invoker_id (aka topic) was supposed to enable. Maybe the fix is we should always be using invoker_name as the container prefix instead of the instanceId

tysonnorris:
_always be using invoker_name as the container prefix instead of the instanceId_ would work for us, if that is more appealing than a separate config for specifying the container name prefix.

rabbah:
+1 to no new config.

tysonnorris:
Ok, in the interest of "no new config" I added `name` to InstanceId, and use that for the containerNamePrefix.

common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
```diff
@@ -50,6 +50,15 @@ trait ContainerFactory {
   def cleanup(): Unit
 }
 
+object ContainerFactory {
+
+  /** include the instance name, if specified and strip invalid chars before attempting to use them in the container name */
+  def containerNamePrefix(instanceId: InstanceId): String =
+    s"wsk${instanceId.name.getOrElse("")}${instanceId.toInt}"
+      .replaceAll("[^a-zA-Z0-9_\\.\\-]", "") // based on https://github.com/moby/moby/issues/3138 and https://github.com/moby/moby/blob/master/daemon/names/names.go
```
rabbah: Since there’s a replace all subsequently should we remove this and trust the invariant is asserted elsewhere?

rabbah: Ah there’s a docker clean which would need this.

rabbah:
@dgrove-oss are you ok with these changes?

dgrove-oss:
fine with me.

rabbah:
PG4/1299 🚀

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -130,8 +130,7 @@ object Invoker {
         if (config.zookeeperHosts.startsWith(":") || config.zookeeperHosts.endsWith(":")) {
           abort(s"Must provide valid zookeeper host and port to use dynamicId assignment (${config.zookeeperHosts})")
         }
-        val invokerName = cmdLineArgs.name.getOrElse(config.invokerName)
-        if (invokerName.trim.isEmpty) {
+        if (invokerName.isEmpty || invokerName.get.trim.isEmpty) {
```
rabbah: this check is only done when the id is missing - so it is possible to have a name that is an empty (`""` or `" "`) string accepted as the invoker name (only `Option(null) == None`).

i don't think that's intended.

common/scala/src/main/scala/whisk/core/containerpool/ContainerFactory.scala
rabbah: per my note below, instanceId.name may be an empty string `"    "` - this is replaced here so is harmless but i don't think was intended.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
tysonnorris: ah, got it - I added another `isEmpty` check as
```
val invokerName = cmdLineArgs.name.orElse(if (config.invokerName.trim.isEmpty) None else Some(config.invokerName))
```
so, use the `--name` arg, OR config.invokerName (if its not empty)

rabbah:
Restarted Travis.

rabbah:
@markusthoemmes I'm going to merge this first - you'll need to update your pr.

PR #3217, opened 2018-01-24
Use case class instead of self-rolled class for InvokerHealth

markusthoemmes:
`case class` give us the wanted equality by default without the added boilerplate.


rabbah:
I also considered moving to json serialization into this class as well (from the controller where it is today).

markusthoemmes:
@rabbah A `trait` for the interface and a case class as an implementation then?

rabbah:
Maybe... are you concerned with the equals methods? Note there is a test already to cover it (in the supervisor tests) which checks list equality.

markusthoemmes:
It just felt weird, maybe I'm overpicky 😂

rabbah:
I suppose it could be a trait - not sure it's really worth it.

markusthoemmes:
You're right, it's nitty.

PR #3218, opened 2018-01-24
Stop-start invoker and retry more to ease testing

jonpspri:
When running repeated test cycles, it becomes necessary to 'docker rm -f invoker0' before rerunning the openwhisk.yml playbook.  This adds a step to ensure prior iterations are stopped before starting the new invoker.

Y'alls call whether you want to do it -- running 'mode=clean' before rerunning tests is another option; however for basic build testing it's nice to have the behavior of the playbook essentially the same for each run.  I won't take a 'no' personally (for this one...)

Yes, I've looked at:

https://github.com/apache/incubator-openwhisk/issues/1633
https://github.com/ansible/ansible-modules-core/issues/5054
https://github.com/ansible/ansible/issues/20648

BTW, both the ansible issues are closed without resolution; I can't divine the state of userns support.

-J

### DIFF
--- a/ansible/roles/invoker/tasks/deploy.yml
+++ b/ansible/roles/invoker/tasks/deploy.yml
@@ -129,6 +129,12 @@
   set_fact:
     invoker_args: "{{ invoker.arguments }} {{ invoker.jmxremote.jvmArgs }}"
 
+- name: stop and remove the invoke container if it's currently running
+  docker_container:
+      name: invoker{{ groups['invokers'].index(inventory_hostname) }}
+      state: absent
+      force_kill: yes
+
 - name: start invoker using docker cli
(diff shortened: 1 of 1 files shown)


markusthoemmes:
We have clean tasks which are specifically supposed to do this exact thing. Using the redo tool, you can compile, clean, deploy in a  single command: `redo invoker`.

rabbah:
redo -c
For list of components

redo deploy 
Is equivalent to redeploying all the components 

redo -n
For dry run and see what it runs under the covers

jonpspri:
Ah ha!  Had I actually read 'tools/build/README.md' I might have avoided some problems.  I'll close this issue.

PR #3219, opened 2018-01-24
Kubernetes ContainerFactoryProvider implementation

jcrossley3:
We've had this working in a fork for a few months now, but have held off submitting due to performance limits imposed by our inability to `suspend/resume` containers like the Docker provider. However, @dgrove-oss has some ideas around this and asked us to go ahead and open the PR.

One other limitation is that the k8s API doesn't distinguish between `stdout` and `stderr` for logged messages. So in our impl, all messages (even those sent to `stderr`) appear to go to `stdout`. See https://github.com/kubernetes/kubernetes/issues/28167

I squashed a bunch of commits for this but wanted to maintain a bit of history for posterity. If you'd prefer a single commit, that's cool, too.

### DIFF
--- a/core/invoker/Dockerfile
+++ b/core/invoker/Dockerfile
@@ -1,6 +1,9 @@
 FROM scala
 
 ENV DOCKER_VERSION 1.12.0
+ENV KUBERNETES_VERSION 1.6.4
+
+RUN apk add --update openssl
 
 # Uncomment to fetch latest version of docker instead: RUN wget -qO- https://get.docker.com | sh
 # Install docker client
@@ -10,6 +13,11 @@ tar --strip-components 1 -xvzf docker-${DOCKER_VERSION}.tgz -C /usr/bin docker/d
 rm -f docker-${DOCKER_VERSION}.tgz && \
 chmod +x /usr/bin/docker && \
 chmod +x /usr/bin/docker-runc
+
+# Install kubernetes client
+RUN wget --no-verbose https://storage.googleapis.com/kubernetes-release/release/v${KUBERNETES_VERSION}/bin/linux/amd64/kubectl && \
+chmod +x kubectl && \
+mv kubectl /usr/bin/kubectl
 
 COPY build/distributions/invoker.tar ./
 RUN tar xf invoker.tar && \
--- a/core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
@@ -1,0 +1,380 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.containerpool.kubernetes
+
+import java.io.{FileNotFoundException, IOException}
+import java.net.SocketTimeoutException
+import java.nio.file.Files
+import java.nio.file.Paths
+import java.time.{Instant, ZoneId}
+import java.time.format.DateTimeFormatterBuilder
+
+import akka.actor.ActorSystem
+import akka.event.Logging.{ErrorLevel, InfoLevel}
+import akka.http.scaladsl.model.Uri
+import akka.http.scaladsl.model.Uri.Path
+import akka.http.scaladsl.model.Uri.Query
+import akka.stream.{Attributes, Outlet, SourceShape}
+import akka.stream.scaladsl.Source
+import akka.stream.stage._
+import akka.util.ByteString
+import pureconfig.loadConfigOrThrow
+import whisk.common.Logging
+import whisk.common.LoggingMarkers
+import whisk.common.TransactionId
+import whisk.core.ConfigKeys
+import whisk.core.containerpool.ContainerId
+import whisk.core.containerpool.ContainerAddress
+import whisk.core.containerpool.docker.ProcessRunner
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.blocking
+import scala.concurrent.duration._
+import scala.util.Failure
+import scala.util.Success
+import scala.util.Try
+import spray.json._
+import spray.json.DefaultJsonProtocol._
+import io.fabric8.kubernetes.client.ConfigBuilder
+import io.fabric8.kubernetes.client.DefaultKubernetesClient
+import okhttp3.{Call, Callback, Request, Response}
+import okio.BufferedSource
+
+import scala.annotation.tailrec
+import scala.collection.mutable
+import scala.util.control.NonFatal
+
+/**
+ * Configuration for kubernetes client command timeouts.
+ */
+case class KubernetesClientTimeoutConfig(run: Duration, rm: Duration, inspect: Duration, logs: Duration)
+
+/**
+ * Serves as interface to the kubectl CLI tool.
+ *
+ * Be cautious with the ExecutionContext passed to this, as the
+ * calls to the CLI are blocking.
+ *
+ * You only need one instance (and you shouldn't get more).
+ */
+class KubernetesClient(
+  timeouts: KubernetesClientTimeoutConfig = loadConfigOrThrow[KubernetesClientTimeoutConfig](
+    ConfigKeys.kubernetesTimeouts))(executionContext: ExecutionContext)(implicit log: Logging, as: ActorSystem)
+    extends KubernetesApi
+    with ProcessRunner {
+  implicit private val ec = executionContext
+  implicit private val kubeRestClient = new DefaultKubernetesClient(
+    new ConfigBuilder()
+      .withConnectionTimeout(timeouts.logs.toMillis.toInt)
+      .withRequestTimeout(timeouts.logs.toMillis.toInt)
+      .build())
+
+  // Determines how to run kubectl. Failure to find a kubectl binary implies
+  // a failure to initialize this instance of KubernetesClient.
+  protected def findKubectlCmd(): String = {
+    val alternatives = List("/usr/bin/kubectl", "/usr/local/bin/kubectl")
+    val kubectlBin = Try {
+      alternatives.find(a => Files.isExecutable(Paths.get(a))).get
+    } getOrElse {
+      throw new FileNotFoundException(s"Couldn't locate kubectl binary (tried: ${alternatives.mkString(", ")}).")
+    }
+    kubectlBin
+  }
+  protected val kubectlCmd = Seq(findKubectlCmd)
+
+  def run(name: String, image: String, args: Seq[String] = Seq.empty[String])(
+    implicit transid: TransactionId): Future[ContainerId] = {
+    runCmd(Seq("run", name, s"--image=$image") ++ args, timeouts.run)
+      .map(_ => ContainerId(name))
+  }
+
+  def inspectIPAddress(id: ContainerId)(implicit transid: TransactionId): Future[ContainerAddress] = {
+    Future {
+      blocking {
+        val pod =
+          kubeRestClient.pods().withName(id.asString).waitUntilReady(timeouts.inspect.length, timeouts.inspect.unit)
+        ContainerAddress(pod.getStatus().getPodIP())
+      }
+    }.recoverWith {
+      case e =>
+        log.error(this, s"Failed to get IP of Pod '${id.asString}' within timeout: ${e.getClass} - ${e.getMessage}")
+        Future.failed(new Exception(s"Failed to get IP of Pod '${id.asString}'"))
+    }
+  }
+
+  def rm(id: ContainerId)(implicit transid: TransactionId): Future[Unit] =
+    runCmd(Seq("delete", "--now", "pod", id.asString), timeouts.rm).map(_ => ())
+
+  def rm(key: String, value: String)(implicit transid: TransactionId): Future[Unit] =
+    runCmd(Seq("delete", "--now", "pod", "-l", s"$key=$value"), timeouts.rm).map(_ => ())
+
+  def logs(id: ContainerId, sinceTime: Option[Instant], waitForSentinel: Boolean = false)(
+    implicit transid: TransactionId): Source[TypedLogLine, Any] = {
+
+    log.debug(this, "Parsing logs from Kubernetes Graph Stage…")
+
+    Source
+      .fromGraph(new KubernetesRestLogSourceStage(id, sinceTime, waitForSentinel))
+      .log("foobar")
+
+  }
+
+  private def runCmd(args: Seq[String], timeout: Duration)(implicit transid: TransactionId): Future[String] = {
+    val cmd = kubectlCmd ++ args
+    val start = transid.started(
+      this,
+      LoggingMarkers.INVOKER_KUBECTL_CMD(args.head),
+      s"running ${cmd.mkString(" ")} (timeout: $timeout)",
+      logLevel = InfoLevel)
+    executeProcess(cmd, timeout).andThen {
+      case Success(_) => transid.finished(this, start)
+      case Failure(t) => transid.failed(this, start, t.getMessage, ErrorLevel)
+    }
+  }
+}
+
+object KubernetesClient {
+
+  // Necessary, as Kubernetes uses nanosecond precision in logs, but java.time.Instant toString uses milliseconds
+  //%Y-%m-%dT%H:%M:%S.%N%z
+  val K8STimestampFormat = new DateTimeFormatterBuilder()
+    .parseCaseInsensitive()
+    .appendPattern("u-MM-dd")
+    .appendLiteral('T')
+    .appendPattern("HH:mm:ss[.n]")
+    .appendLiteral('Z')
+    .toFormatter()
+    .withZone(ZoneId.of("UTC"))
+
+  def parseK8STimestamp(ts: String): Try[Instant] =
+    Try(Instant.from(K8STimestampFormat.parse(ts)))
+
+  def formatK8STimestamp(ts: Instant): Try[String] =
+    Try(K8STimestampFormat.format(ts))
+}
+
+trait KubernetesApi {
+  def run(name: String, image: String, args: Seq[String] = Seq.empty[String])(
+    implicit transid: TransactionId): Future[ContainerId]
+
+  def inspectIPAddress(id: ContainerId)(implicit transid: TransactionId): Future[ContainerAddress]
+
+  def rm(id: ContainerId)(implicit transid: TransactionId): Future[Unit]
+
+  def rm(key: String, value: String)(implicit transid: TransactionId): Future[Unit]
+
+  def logs(containerId: ContainerId, sinceTime: Option[Instant], waitForSentinel: Boolean = false)(
+    implicit transid: TransactionId): Source[TypedLogLine, Any]
+}
+
+object KubernetesRestLogSourceStage {
+
+  import KubernetesClient.{formatK8STimestamp, parseK8STimestamp}
+
+  def constructPath(namespace: String, containerId: String): Path =
+    Path / "api" / "v1" / "namespaces" / namespace / "pods" / containerId / "log"
+
+  def constructQuery(sinceTime: Option[Instant], waitForSentinel: Boolean): Query = {
+
+    val sinceTimestamp = sinceTime.flatMap(time => formatK8STimestamp(time).toOption)
+
+    Query(Map("timestamps" -> "true") ++ sinceTimestamp.map(time => "sinceTime" -> time))
+
+  }
+
+  @tailrec
+  def readLines(src: BufferedSource,
+                lastTimestamp: Option[Instant],
+                lines: Seq[TypedLogLine] = Seq.empty[TypedLogLine]): Seq[TypedLogLine] = {
+
+    if (!src.exhausted()) {
+      (for {
+        line <- Option(src.readUtf8Line()) if !line.isEmpty
+        timestampDelimiter = line.indexOf(" ")
+        // Kubernetes is ignoring nanoseconds in sinceTime, so we have to filter additionally here
+        rawTimestamp = line.substring(0, timestampDelimiter)
+        timestamp <- parseK8STimestamp(rawTimestamp).toOption if isRelevantLogLine(lastTimestamp, timestamp)
+        msg = line.substring(timestampDelimiter + 1)
+        stream = "stdout" // TODO - when we can distinguish stderr: https://github.com/kubernetes/kubernetes/issues/28167
+      } yield {
+        TypedLogLine(timestamp, stream, msg)
+      }) match {
+        case Some(logLine) =>
+          readLines(src, Option(logLine.time), lines :+ logLine)
+        case None =>
+          // we may have skipped a line for filtering conditions only; keep going
+          readLines(src, lastTimestamp, lines)
+      }
+    } else {
+      lines
+    }
+
+  }
+
+  def isRelevantLogLine(lastTimestamp: Option[Instant], newTimestamp: Instant): Boolean =
+    lastTimestamp match {
+      case Some(last) =>
+        newTimestamp.isAfter(last)
+      case None =>
+        true
+    }
+
+}
+
+final class KubernetesRestLogSourceStage(id: ContainerId, sinceTime: Option[Instant], waitForSentinel: Boolean)(
+  implicit val kubeRestClient: DefaultKubernetesClient)
+    extends GraphStage[SourceShape[TypedLogLine]] { stage =>
+
+  import KubernetesRestLogSourceStage._
+
+  val out = Outlet[TypedLogLine]("K8SHttpLogging.out")
+
+  override val shape: SourceShape[TypedLogLine] = SourceShape.of(out)
+
+  override protected def initialAttributes: Attributes = Attributes.name("KubernetesHttpLogSource")
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogicWithLogging(shape) { logic =>
+
+      private val queue = mutable.Queue.empty[TypedLogLine]
+      private var lastTimestamp = sinceTime
+
+      def fetchLogs(): Unit =
+        try {
+          val path = constructPath(kubeRestClient.getNamespace, id.asString)
+          val query = constructQuery(lastTimestamp, waitForSentinel)
+
+          log.debug("* Fetching K8S HTTP Logs w/ Path: {} Query: {}", path, query)
+
+          val url = Uri(kubeRestClient.getMasterUrl.toString)
+            .withPath(path)
+            .withQuery(query)
+
+          val request = new Request.Builder().get().url(url.toString).build
+
+          kubeRestClient.getHttpClient.newCall(request).enqueue(new LogFetchCallback())
+        } catch {
+          case NonFatal(e) =>
+            onFailure(e)
+            throw e
+        }
+
+      def onFailure(e: Throwable): Unit = e match {
+        case _: SocketTimeoutException =>
+          log.warning("* Logging socket to Kubernetes timed out.") // this should only happen with follow behavior
+        case _ =>
+          log.error(e, "* Retrieving the logs from Kubernetes failed.")
+      }
+
+      val emitCallback: AsyncCallback[Seq[TypedLogLine]] = getAsyncCallback[Seq[TypedLogLine]] {
+        case firstLine +: restOfLines if isAvailable(out) =>
+          pushLine(firstLine)
+          queue ++= restOfLines
+        case lines =>
+          queue ++= lines
+      }
+
+      class LogFetchCallback extends Callback {
+
+        override def onFailure(call: Call, e: IOException): Unit = logic.onFailure(e)
+
+        override def onResponse(call: Call, response: Response): Unit =
+          try {
+            val lines = readLines(response.body.source, lastTimestamp)
+
+            response.body.source.close()
+
+            lines.lastOption.foreach { line =>
+              lastTimestamp = Option(line.time)
+            }
+
+            emitCallback.invoke(lines)
+          } catch {
+            case NonFatal(e) =>
+              log.error(e, "* Reading Kubernetes HTTP Response failed.")
+              logic.onFailure(e)
+              throw e
+          }
+      }
+
+      def pushLine(line: TypedLogLine): Unit = {
+        log.debug("* Pushing a chunk of kubernetes logging: {}", line)
+        push(out, line)
+      }
+
+      setHandler(
+        out,
+        new OutHandler {
+          override def onPull(): Unit = {
+            // if we still have lines queued up, return those; else make a new HTTP read.
+            if (queue.nonEmpty)
+              pushLine(queue.dequeue())
+            else
+              fetchLogs()
+          }
+        })
+    }
+}
+
+protected[core] final case class TypedLogLine(time: Instant, stream: String, log: String) {
+  import KubernetesClient.formatK8STimestamp
+
+  lazy val toJson: JsObject =
+    JsObject("time" -> formatK8STimestamp(time).getOrElse("").toJson, "stream" -> stream.toJson, "log" -> log.toJson)
+
+  lazy val jsonPrinted: String = toJson.compactPrint
+  lazy val jsonSize: Int = jsonPrinted.length
+
+  /**
+   * Returns a ByteString representation of the json for this Log Line
+   */
+  val toByteString = ByteString(jsonPrinted)
+
+  override def toString = s"${formatK8STimestamp(time).get} $stream: ${log.trim}"
+}
+
+protected[core] object TypedLogLine {
+
+  import KubernetesClient.{parseK8STimestamp, K8STimestampFormat}
+
+  def readInstant(json: JsValue): Instant = json match {
+    case JsString(str) =>
+      parseK8STimestamp(str) match {
+        case Success(time) =>
+          time
+        case Failure(e) =>
+          deserializationError(
+            s"Could not parse a java.time.Instant from $str (Expected in format: $K8STimestampFormat: $e")
+      }
+    case _ =>
+      deserializationError(s"Could not parse a java.time.Instant from $json (Expected in format: $K8STimestampFormat)")
+  }
+
+  implicit val typedLogLineFormat = new RootJsonFormat[TypedLogLine] {
+    override def write(obj: TypedLogLine): JsValue = obj.toJson
+
+    override def read(json: JsValue): TypedLogLine = {
+      val obj = json.asJsObject
+      val fields = obj.fields
+      TypedLogLine(readInstant(fields("time")), fields("stream").convertTo[String], fields("log").convertTo[String])
+    }
+  }
+
+}
--- a/core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
@@ -1,0 +1,158 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.containerpool.kubernetes
+
+import java.time.Instant
+import java.util.concurrent.atomic.AtomicReference
+
+import akka.stream.StreamLimitReachedException
+import akka.stream.scaladsl.Framing.FramingException
+import akka.stream.scaladsl.Source
+import akka.util.ByteString
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import whisk.common.Logging
+import whisk.common.TransactionId
+import whisk.core.containerpool.Container
+import whisk.core.containerpool.WhiskContainerStartupError
+import whisk.core.containerpool.ContainerId
+import whisk.core.containerpool.ContainerAddress
+import whisk.core.containerpool.docker.{CompleteAfterOccurrences, DockerContainer, OccurrencesNotFoundException}
+import whisk.core.entity.ByteSize
+import whisk.core.entity.size._
+import whisk.http.Messages
+
+object KubernetesContainer {
+
+  /**
+   * Creates a container running in kubernetes
+   *
+   * @param transid transaction creating the container
+   * @param image image to create the container from
+   * @param userProvidedImage whether the image is provided by the user
+   *     or is an OpenWhisk provided image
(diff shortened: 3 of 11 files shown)


core/invoker/Dockerfile
```diff
@@ -11,6 +12,11 @@ rm -f docker-${DOCKER_VERSION}.tgz && \
 chmod +x /usr/bin/docker && \
 chmod +x /usr/bin/docker-runc
 
+# Install kubernetes client
+RUN wget --no-verbose https://storage.googleapis.com/kubernetes-release/release/v${KUBERNETES_VERSION}/bin/linux/amd64/kubectl && \
+chmod +x kubectl && \
+mv kubectl /usr/bin/kubectl
+
```
dgrove-oss: I was expecting the client would talk directly to Kubernetes's API service using its REST API, not use kubectl.   I see that the KubernetesContainer has a mix of REST calls using the fabric.io client and kubectl commands.  Is due to ease of development, or is there a performance/functional reason to use kubectl to do the run command?

bbrowning: This is just an artifact of copying the original Docker implementation that shelled out to `docker`. We had to replace usage of `kubectl` with the API for performance reasons in some places but it looks like we never got around to replacing it everywhere.

dgrove-oss: makes sense; thanks.

dgrove-oss:
Happy to report this worked for me (once I figured out the one magic config variable to pass on the command line for my kube environment...).  I have a WIP PR in the deploy-kube that adds the kubernetes yml files to deploy an invoker that uses this new pool.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -37,7 +37,7 @@ import akka.stream.scaladsl.{Framing, Source}
 import akka.stream.stage._
 import akka.util.ByteString
 import spray.json._
-import whisk.core.containerpool.logging.LogLine
+import whisk.core.containerpool.logging.{LogLine => CoreLogLine}
 import whisk.http.Messages
```
rabbah: Why this import remapping?

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,222 @@
...
+      "Always",
+      "--limits",
+      "memory=256Mi") ++ environmentArgs ++ labelArgs
+
+    runCmd(runArgs, timeouts.run)
+      .map { _ =>
+        name
+      }
+      .map(ContainerId.apply)
+  }
```
rabbah: Why not flatten the two maps to:

.map(_ => ContainerId(name))

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,222 @@
...
+            JsObject("log" -> msg.toJson, "stream" -> stream.toJson, "time" -> ts.toJson)
+          }
+          .mkString("\n") + "\n" // trailing newline is necessary or the frame won't be decoded and break the akka stream
+      )
+    })
+  }
+
+  private def runCmd(args: Seq[String], timeout: Duration)(implicit transid: TransactionId): Future[String] = {
+    val cmd = kubectlCmd ++ args
+    val start = transid.started(
```
rabbah: Recent changes to this logger method means this is only emitted at debug level. FYI.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
```diff
@@ -0,0 +1,159 @@
...
+class KubernetesContainer(protected val id: ContainerId, protected val addr: ContainerAddress)(
+  implicit kubernetes: KubernetesApi,
+  protected val ec: ExecutionContext,
+  protected val logging: Logging)
+    extends Container {
+
+  /** The last read timestamp in the log file */
+  private val lastTimestamp = new AtomicReference("")
+
+  protected val logsRetryCount = 15
```
markusthoemmes: No longer used.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
```diff
@@ -0,0 +1,159 @@
...
+  protected val logsRetryWait = 100.millis
```
markusthoemmes: No longer used either.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,222 @@
...
+    runCmd(runArgs, timeouts.run)
+      .map { _ =>
+        name
+      }
+      .map(ContainerId.apply)
+  }
+
+  def inspectIPAddress(id: ContainerId)(implicit transid: TransactionId): Future[ContainerAddress] = {
+    Future {
+      blocking {
```
markusthoemmes: Is there a non-blocking client prodiving the features we need? (https://github.com/doriordan/skuber for example?)

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,222 @@
...
+          URLUtils.join(
+            kubeRestClient.getMasterUrl.toString,
+            "api",
+            "v1",
+            "namespaces",
+            kubeRestClient.getNamespace,
+            "pods",
+            id.asString,
+            "log?timestamps=true" ++ sinceTimePart))
+        val request = new Request.Builder().get().url(url).build
```
markusthoemmes: Would it make sense to use Akka's URI building capabilities here? Could get much more readable.

```scala
val sinceTime = Some("abc")

val path = Path / "api" / "v1" / "namespaces" / "namespace" / "pods" / "pod"
val query = Seq("timestamp" -> true.toString) ++ sinceTime.map(time => "sinceTime" -> time)
val url = Uri("https://google.de")
  .withPath(path)
  .withQuery(Query(query: _*))
```

That assumes that `sinceTime` is an `Option[String]` to denote the possible absence of it in the type. In general, it'll make your code much easier to grasp if you make use of `Option` whenever it is possible that something is missing rather than passing an empty String and checking for it to be empty.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,222 @@
...
+
+        val response = kubeRestClient.getHttpClient.newCall(request).execute
+        if (!response.isSuccessful) {
+          Future.failed(
+            new Exception(s"Kubernetes API returned HTTP status ${response.code} when trying to retrieve pod logs"))
```
markusthoemmes: This failed `Future` will be thrown away and not used anywhere. I think what you're looking for is:

```scala
Future {
  blocking {
    // your request code
    val response = kubeRestClient.getHttpClient.newCall(request).execute
    if(response.isSuccessful) {
      Future.successful(response.body.string)
    } else {
      Future.failed(
            new Exception(s"Kubernetes API returned HTTP status ${response.code} when trying to retrieve pod logs"))
    }
  }
}.flatMap(identity).map { output => ... }
```

An `if` is an expression in Scala, meaning it'll return something (namely: the value of the last line). The `Future { }` block in your case always returns `response.body.string`. If you want to fold the error case into the future chain, you'll explicitly need to return a `Success` for the good case and a `Failure` for the bad case. Since it makes chaining easier, I used the `Future.successful` and `Future.failed` variants, which are immediatly resolved Futures of the desired state. The return type would now be `Future[Future[String]]` though, since I returned a Future inside another Future. To flatten those (flatten is only available in Scala 2.12) you can use `flatMap(identity)`. `identity` is a predefined function. You can as well write `flatMap(x => x)`.

Hope that makes sense.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,222 @@
...
+    val environmentArgs = environment.flatMap {
+      case (key, value) => Seq("--env", s"$key=$value")
+    }.toSeq
+
+    val labelArgs = labels.map {
+      case (key, value) => s"$key=$value"
+    } match {
+      case Seq() => Seq()
+      case pairs => Seq("-l") ++ pairs
+    }
```
markusthoemmes: Do you need to add `"-l"` in front of every pair or just in front of all pairs? Currently it's the latter.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,222 @@
...
+ * Serves as interface to the kubectl CLI tool.
+ *
+ * Be cautious with the ExecutionContext passed to this, as the
+ * calls to the CLI are blocking.
+ *
+ * You only need one instance (and you shouldn't get more).
+ */
+class KubernetesClient(
+  timeouts: KubernetesClientTimeoutConfig = loadConfigOrThrow[KubernetesClientTimeoutConfig](
+    "whisk.kubernetes.timeouts"))(executionContext: ExecutionContext)(implicit log: Logging, as: ActorSystem)
```
markusthoemmes: Please externalize the config key to `ConfigKeys`

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
```diff
@@ -0,0 +1,159 @@
...
+
+    val activationMarkerCheck: ByteString ⇒ Boolean = { bs ⇒
+      if (bs.containsSlice(DockerContainer.ActivationSentinel)) {
+
+        val lastTS = bs.utf8String.lines.toSeq.lastOption
+          .getOrElse("""{"time":""}""")
+          .parseJson
+          .asJsObject
+          .fields("time")
+          .convertTo[String]
```
markusthoemmes: It's not necessary to match on lines and non-emptyness here, since we have successfully passed the Framing-delimiter we know we only have 1 line at hand.

What about:

```scala
lastTimestamp.set(bs.utf8String.convertTo[LogLine].time)
```

Parsing to `LogLine` should be safe since you created the objects accordingly above.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,222 @@
...
+      //var sentinelSeen = false
+      ByteString(
+        result
+          .map { line =>
+            val pos = line.indexOf(" ")
+            val ts = line.substring(0, pos)
+            val msg = line.substring(pos + 1)
+            // TODO - can we get the proper stream name from kubernetes? Some stuff is stderr
+            val stream = "stdout"
+            JsObject("log" -> msg.toJson, "stream" -> stream.toJson, "time" -> ts.toJson)
```
markusthoemmes: What about using `LogLine(msg, stream, ts).toJson.compactPrint` instead? That would ensure the proper format and fieldnames.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,222 @@
...
+        val response = kubeRestClient.getHttpClient.newCall(request).execute
+        if (!response.isSuccessful) {
+          Future.failed(
+            new Exception(s"Kubernetes API returned HTTP status ${response.code} when trying to retrieve pod logs"))
+        }
+        response.body.string
+      }
+    }.map { output =>
+      val original = output.lines.toSeq
+      val relevant = original.dropWhile(s => !s.startsWith(sinceTime))
```
markusthoemmes: Can you elaborate a bit on this piece? I don't quite grasp why you're dropping lines having the exact same timestamp as the sinceTime and then decide based on that what to keep. That is to say: A comment describing it would be good.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
jcrossley3: Looks like `DockerContainer` explicitly sets `InfoLevel` now. I'll update to match.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
dgrove-oss: All the clients are going to bottom out with making http put/get against the same k8s REST API.  So, I'm not sure if it really matters in the end if we have the blocking explicit in our code or hide it in the bowels of 3rd party client instead.  

I'd bias towards picking a client based on expected long-term maintenance/support, expectation of it tracking k8s API changes, and license compatibility.

markusthoemmes: Sure, that's why I added "providing the features we need". I'm all in for not just picking it for the sake of it. Still a non-blocking client is preferable over a blocking one (if it exists). This might matter if you want to scale the invoker to more than the 16 containers that we scale it today.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
jcrossley3: When selecting by multiple labels, you can either use multiple `"-l"`flags or a single `"-l"` with a comma-delimited list. The logic for the former is "any can match" and the logic for the latter is "all must match". At least I think that's how it works. ;)

Right now, I believe we only ever pass one label anyway, so I'm inclined to leave it as is for now.

dgrove-oss: fwiw, I'm prototyping a version where we create the pod for the user action container using the client (not kubectl).  That eliminates this code entirely.

markusthoemmes: If we only want this to work for one label, we should change `labels: Map[String, String]` to `label: (String, String)` or make the code work for multiple labels. Leaving a known bug seems odd to me and might become frustrating for others who want to enhance/refactor that code.

jcrossley3: I wouldn't call it a bug. The code does work for multiple labels, in the way I would expect if I did pass multiple labels. I get what you're saying about changing the code to work for one label since that's how we're currently using it, but to me that conflicts with how one might generically use `kubectl` itself. I don't feel that strongly about it, though, especially in light of @dgrove-oss attempt to obviate it. :)

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
jcrossley3: I don't seem to have those methods on `LogLine`. Is it enough to just `LogLine(msg, stream, ts)`?

markusthoemmes: `LogLine` is a case-class, you'll likely need to import it. But then, yes, it should just be that.

jcrossley3: One of us may be confused. Just to make sure you know what this code is doing: we are adapting the k8s log output, making it look like the docker log output, so that we can re-use the docker classes in `KubernetesContainer.logs`. I concede it's a hack, but I think it means we can't use `LogLine` at this point.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
dgrove-oss: This code would become something like: https://github.com/dgrove-oss/openwhisk/commit/7a72e1913b7e8a3e49f6ccfd5c69441fc90801db

I'm debating between constructing the pod like this (via code) vs. loading a template .yml file and then filling in a few holes (with limit, env, name).  I have a guess that this is a place where providers may want to customize, but maybe we should defer that decision until later.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
markusthoemmes: Right I'm aware. Using `LogLine` just **ensures** you're adapting correctly (because it's used for deserialization again). Both you're any my line should yield the same result, but using `LogLine` ensures you're using the right field names etc.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
jcrossley3: Very cool!

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
bbrowning: We're planning to scale each Invoker to way more than 16 containers, but we haven't done large-scale load testing of this yet. I'd say we're targeting at least 100 or more containers per Invoker.

dgrove-oss:
PG1 / 2574 🏃

dgrove-oss:
@jcrossley3 - I dug into the PG test failures this morning.  The problem is that the PR needs to be rebased to pick up the change from #3063  that moves common/scala/Dockerfile to use alpine linux as a base image.  Without that change in base image, it won't go through the PG testing pipeline.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,218 @@
...
+        result
+          .map { line =>
+            val pos = line.indexOf(" ")
+            val ts = line.substring(0, pos)
+            val msg = line.substring(pos + 1)
+            // TODO - when we can distinguish stderr: https://github.com/kubernetes/kubernetes/issues/28167
+            val stream = "stdout"
+            LogLine(ts, stream, msg).toJson.compactPrint
+          }
+          .mkString("\n") + "\n" // trailing newline is necessary or the frame won't be decoded and break the akka stream
```
markusthoemmes: Why not just add the `"\n"` on the line above:

```scala
LogLine(ts, stream, msg).toJson.compactPrint + "\n"
```

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,218 @@
...
+      // filter logs from previous activations on a warm runtime pod
+      val original = output.lines.toSeq
+      val relevant = original.dropWhile(s => !s.startsWith(sinceTime.getOrElse("")))
+      val result =
+        if (!relevant.isEmpty && original.size > relevant.size) {
+          // drop matching timestamp from previous activation
+          relevant.drop(1)
+        } else {
+          original
+        }
```
markusthoemmes: Hmm, I'm still not able to grasp this, let me try:

```scala
val original = output.lines
val relevant = sinceTime.map(time => original.dropWhile(!_.startsWith(time)).drop(1)).getOrElse(original)
```

`relevant` should no in any case contain what you need to continue processing, right? The following `if` is redundant, because if relevant is empty, the result should be empty as well?

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,218 @@
...
+      "run",
+      name,
+      "--image",
+      image,
+      "--generator",
+      "run-pod/v1",
+      "--restart",
+      "Always",
+      "--limits",
+      "memory=256Mi") ++ environmentArgs ++ labelArgs
```
markusthoemmes: 1. Looks to me like `memory=256Mi` should be passed in from the outside.
2. Why `restart Always`? For user-containers?
3. Do we need to make the generator configurable?

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainerFactory.scala
```diff
@@ -0,0 +1,81 @@
...
+      actionImage.localImageName(config.dockerRegistry, config.dockerImagePrefix, Some(config.dockerImageTag))
+    }
+
+    KubernetesContainer.create(
+      tid,
+      image = image,
+      userProvidedImage = userProvidedImage,
+      environment = Map("__OW_API_HOST" -> config.wskApiHost),
+      labels = Map("invoker" -> label),
+      name = Some(name))
```
markusthoemmes: Memory usage should be passed in here?

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
jcrossley3: If `relevant` is empty, it means the previous timestamp isn't in `original`. This will be the case for both cold pods and warm pods that haven't been used within the granularity of the timestamp.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
jcrossley3: Makes sense to me, I think. :)

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
jcrossley3: #1 makes sense. I'm gonna refer to @bbrowning for #2 and #3

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
jcrossley3: I should clarify about the warm pods. It's not necessarily that they haven't been used. It's that the log messages from the previous activation are within the server's granularity relative to the activation's last message. This is very rarely more than just the two stream sentinel markers.

jcrossley3: When I read my comments, I don't feel like they're helping you. :)

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
bbrowning: This may be specific to OpenShift, but the reason for this is so that these pods don't get treated as temporary pods with different lifecycle semantics, sometimes resulting in them being prematurely killed. By setting `restart` to `Always` we prevent them from being treated as temporary pods. However, if you only set that, then instead of creating a Pod object it ends up creating a replication controller. So, we also pass the `generator` to ensure it just creates a Pod.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
dgrove-oss: Circling back on the client question.  The fabric8io client appears to lack support for specifying pod affinity.  Skimming through the source code, I see no support for affinity (and several open issues on GitHub asking for affinity support). The skuber client looks like it does have affinity support.  I think affinity is a required feature (lets us tell kube to restrict creation of user action containers to designated invoker nodes).

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
bwmcadams: the "why" of the trailing newline is my fault; no logic to it and the suggested adjustment makes sense. 

If I'm understanding you correctly, @markusthoemmes… right now we're splitting the entire logging result based on the HTTP response, and pushing it out as a single item. Your suggestion is to break it out so the stream takes individual events for each line, yes?

markusthoemmes: @bwmcadams Yep, it goes through a framing stage immediatley anyway so I don't expect much difference in either way. It just feels clearer to have it immediately exploded into the specific events and squash them into the stream instead of creating a string just to split it immediatly.

markusthoemmes: Actually, thinking about it: If you choose to implement my proposal you can just drop the FramingStage below.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
```diff
@@ -0,0 +1,145 @@
...
+        false
+      }
+    }
+
+    kubernetes
+      .logs(id, lastTimestamp.get()) // todo - same sentinel check behavior as DockerContainer should be implemented?
+      .via(Framing.delimiter(delimiter, limit.toBytes.toInt))
+      .via(new CompleteAfterOccurrences(activationMarkerCheck, 2, waitForSentinel))
+      .recover {
+        case _: StreamLimitReachedException =>
```
markusthoemmes: Is this Exception relevant? You dropped the limitWeighted stage. Was that deliberate? You'll read/store unbounded data.

jcrossley3:
Ok, PR has been rebased against current master and unit tests have been added. There are 2 pending ones in there that revealed some issues @bwmcadams is in the process of addressing. Yay, unit tests! :)

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
```diff
@@ -0,0 +1,165 @@
...
+      "--generator",
+      "run-pod/v1",
+      "--restart",
+      "Always",
+      "--limits",
+      s"memory=${memory.toMB}Mi") ++ environmentArgs ++ labelArgs
+
+    for {
+      id <- kubernetes.run(podName, image, args).recoverWith {
+        case _ => Future.failed(WhiskContainerStartupError(s"Failed to run container with image '${image}'."))
```
rabbah: We recently (@markusthoemmes specifically) changes the content of this message to use
```
  val resourceProvisionError = "Failed to provision resources to run the action."
```
since the image name may be mangled internally for some images.

Perhaps we should use that here.
@markusthoemmes we should may just make that the default so we prevent this in the future from possibly leaking out to activation responses.

rabbah: Same for line 99 below.

jcrossley3: Done

rabbah:
@jcrossley3 I know you want to preserve the history in this PR so when @markusthoemmes approves/dismisses his review, can you adjust the last few commit and stash them appropriately? One of us can merge without squashing the full history.

markusthoemmes:
Regarding credits for all committers: Github introduced https://github.com/blog/2496-commit-together-with-co-authors. Would you be okay using that and still squash down to one commit?

rabbah:
nice! thanks for sharing it, I hadn't see that yet.

jcrossley3:
@rabbah do you want all of this PR's commits squashed into a single one? Or just those pushed in response to feedback?

rabbah:
With the git enhancement linked by Markus you don’t need to squash at all. We can just amend the commit message when merging.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
bwmcadams: @markusthoemmes consider most of this thread obsolete; I've done a rewrite with all of this handled as a custom `GraphStage` and it's much saner. It cuts out most of the hackiness of the existing code.

markusthoemmes:
@jcrossley3 since all feedback seems to be addressed, it'd still be good if you can squash down or at least ammend the first commit in a way to conform with the feature I linked. It seems to be a bit cumbersome to add the proper email adresses and everything. You'll have an easier time doing that than the reviewers/mergers.

jcrossley3:
@markusthoemmes I added more tests around the log handling and implemented your code suggestions in `KubernetesClient` which @bwmcadams is in the midst of changing as well. When he pushes his fixes, I'll squash the most recent commits with the email addresses.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
bwmcadams: With regards to the non-blocking, I've solved this down stream and am working on finishing testing. The fabric8 uses okhttp, which has an async mode. I've migrated to a custom `Source`, using the async callbacks, to eliminate most of the cruft and any blocking code. (I'm not sure about the pod affinity however)

dgrove-oss: cool!  Maybe we can help out by adding the pod affinity support downstream as well.  I don't think it should be too hard.

markusthoemmes:
@jcrossley3 You'll need to rebase. Are all commit from @bwmcadams in?

jcrossley3:
Not quite yet @markusthoemmes. In fixing the tests, @bwmcadams took it upon himself to refactor things to remove the blocking calls. Once he pushes that change and you guys approve it, I'll rebase.

jcrossley3:
Ok, we've rebased and changed the logging stuff quite a bit, but I think we have good unit test coverage now, at least parity with our Docker counterparts. Once you guys approve, let me know how much you'd like to squash. I say probably at least back to 3ef07bf and maybe even farther.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
bwmcadams: After reviewing, with regards to this section of code, I can't see a way with the fabric8 java client to *not* block here. Even if you avoid the utility method for waiting that we use in this code, every layer all the way down the client blocks at some point for it. and since this isn't able to just be a simple REST call like the log fetch, I can't refactor it to be async. The blocking is wrapped and marked inside the future, however, and will be handled safely.

dgrove-oss:
Started PG1/2632 to see how things look.

dgrove-oss:
PG1/2632 looked clean (2 test failed; not specific to this PR (also failing on other PG runs).

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
dgrove-oss: Makes sense to me.  Fundamentally we need to wait somehow until the pod is ready before we can inspect its Status object to get the PodIP and similar info.

tests/src/test/scala/whisk/core/containerpool/kubernetes/test/KubernetesClientTests.scala
```diff
@@ -0,0 +1,272 @@
...
+  val commandTimeout = 500.milliseconds
+  def await[A](f: Future[A], timeout: FiniteDuration = commandTimeout) = Await.result(f, timeout)
+
+  val kubectlCommand = "kubectl"
+
+  /** Returns a KubernetesClient with a mocked result for 'executeProcess' and 'fetchHTTPLogs' */
+  def kubernetesClient(fixture: => Future[String]) = new KubernetesClient()(global) {
+    override def findKubectlCmd() = kubectlCommand
+    override def executeProcess(args: Seq[String], timeout: Duration)(implicit ec: ExecutionContext, as: ActorSystem) =
+      fixture
```
dgrove-oss: fwiw, I don't think mocking out executeProcess is going to be a useful long-term tool for unit tests.  As we replace kubectl commands with calls to the kubernertes REST API, we lose executeProcess as a hook point.  For example, in a branch based off this PR, I've switched to doing pod creation via the REST API and as a result lost this hook point for all the "run" unit tests (so I disabled them).

jcrossley3: DISABLED THE TESTS WHAT!?!?! ;)
Yeah, @bwmcadams  had to set up dummy `Source` objects to validate the log parsing logic. I guess we'll need to figure out something similar as we convert the other calls to the API.

dgrove-oss: open for suggestions on how to unit test run when it becomes just a thin veneer over the kubernetesClient API.  Code is sitting here if you want a preview: https://github.com/dgrove-oss/openwhisk/tree/k8s-invokeragent-20180219 (I've cleaned up & consolidated my older branches into 3 logical commits when I rebased them onto the latest code from this PR).

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,324 @@
...
+  def constructPath(namespace: String, containerId: String): Path =
+    Path / "api" / "v1" / "namespaces" / namespace / "pods" / containerId / "log"
+
+  def constructQuery(sinceTime: Option[LocalDateTime], waitForSentinel: Boolean): Query = {
+
+    val qB = Map.newBuilder[String, String]
+    qB += "timestamps" → "true"
+    qB ++= sinceTime.map(time => "sinceTime" -> time.format(KubernetesClient.K8SDateTimeFormat))
+
+    Query(qB.result())
```
markusthoemmes: Why use a manual `Builder` here?

Why not:

```scala
Map("timestamps" -> "true") ++ sinceTime.map(time => "sinceTime" -> time.format(KubernetesClient.K8SDateTimeFormat))
```

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,324 @@
...
+object KubernetesClient {
+
+  //%Y-%m-%dT%H:%M:%S.%N%z
+  val K8SDateTimeFormat = new DateTimeFormatterBuilder()
+    .parseCaseInsensitive()
+    .appendPattern("u-MM-dd")
+    .appendLiteral('T')
+    .appendPattern("HH:mm:ss[.n]")
+    .appendLiteral('Z')
+    .toFormatter()
```
markusthoemmes: This looks like the default `Instant.toString` format. There's also `Instant.parse` for the inverse. Is a manual pattern needed here?

Can we change the whole code to use `Instant` instead of `LocalDateTime`?

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,324 @@
...
+        l <- Option(src.readUtf8Line()) if !l.isEmpty
+        p = l.indexOf(" ")
+        // Kubernetes is ignoring nanoseconds in sinceTime, so we have to filter additionally here
+        ts = l.substring(0, p)
+        tsDate <- Try(LocalDateTime.parse(ts, K8SDateTimeFormat)).toOption if isRelevantLogLine(lastTimestamp, tsDate)
+        msg = l.substring(p + 1)
+        stream = "stdout" // TODO - when we can distinguish stderr: https://github.com/kubernetes/kubernetes/issues/28167
+      } yield {
+        tsDate -> JsObject("log" -> msg.toJson, "stream" -> stream.toJson, "time" -> ts.toJson)
+      }) match {
```
markusthoemmes: Some more descriptive variable names might help here:

```
line vs. l
timestampDelimiter vs. p
timestamp vs. ts
parsedTimestamp vs. tsDate
```

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -0,0 +1,324 @@
...
+        case Some((latestTS, js)) =>
+          readLines(src, Some(latestTS), lines :+ js)
+        case None =>
+          // we may have skipped a line for filtering conditions only; keep going
+          readLines(src, lastTimestamp, lines)
+      }
+    } else {
+      src.close()
```
markusthoemmes: Should the caller be responsible of closing the source? Since it isn't opened by `readLines`

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
```diff
@@ -0,0 +1,171 @@
...
+
+    val activationMarkerCheck: ByteString ⇒ Boolean = { bs ⇒
+      val str = bs.utf8String
+      val parsed = str.parseJson.convertTo[LogLine]
+      // If this line is a sentinel line, remember the timestamp for lastTimestamp
+      if (bs.containsSlice(DockerContainer.ActivationSentinel)) {
+        Try(LocalDateTime.parse(parsed.time, K8SDateTimeFormat)).foreach(time => lastTimestamp.set(Option(time)))
+        true
+      } else false
+    }
```
markusthoemmes: Is it not feasible to alter the entire stream for the kube-case to avoid to reparse all the loglines just to set the latest timestamp?

You could, for instance, return a fully parsed object (`LogLine` or something like a new `TypedLogLine(Instant, Stream, String)`?) from `kubernetes.logs`, drop the Framing, adjust limitWeighted and then have a `.map(_.toJson.compactPrint)` as the very last stage to be pluggable into the logstores.

You're parsing everything anyway, why not keep it that way? Is it just test-parity?

tests/src/test/scala/whisk/core/containerpool/kubernetes/test/KubernetesClientTests.scala
```diff
@@ -0,0 +1,272 @@
...
+  implicit val materializer: ActorMaterializer = ActorMaterializer()
+
+  /** Reads logs into memory and awaits them */
+  def awaitLogs(source: Source[ByteString, Any], timeout: FiniteDuration = 1000.milliseconds): Vector[LogLine] =
+    Await
+      .result(
+        source
+          .via(Framing.delimiter(ByteString("\n"), 1024))
+          .via(Flow[ByteString].map(l => Try(l.utf8String.parseJson.convertTo[LogLine]).getOrElse(LogLine("", "", ""))))
+          .filter(_ != LogLine("", "", ""))
```
markusthoemmes: Why not `mapConcat(l => Try(l.utf8String.parseJson.convertTo[LogLine])` and drop the following filter?

tests/src/test/scala/whisk/core/containerpool/kubernetes/test/KubernetesContainerTests.scala
```diff
@@ -0,0 +1,550 @@
...
+
+  import KubernetesClientTests.TestKubernetesClient
+
+  override def beforeEach() = {
+    stream.reset()
+  }
+
+  implicit val materializer: ActorMaterializer = ActorMaterializer()
+
+  def instantDT(instant: Instant): LocalDateTime = LocalDateTime.from(instant.atZone(ZoneId.of("GMT+0")))
```
markusthoemmes: "UTC"?

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
bwmcadams: Looks like that's an artifact that survived a refactor and didn't get cleaned up; will fix.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
bwmcadams: It appears you are correct in the formatting; originally we had two split formatters for posting and reading because the REST endpoint doesn't accept nanos (at the time, I thought it just ignored the whole date, but turns out it just skips the nanos). Switching this to instant makes perfect sense.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
bwmcadams: It struck me at the time as the sanest approach, since the HTTP library will drop lots of exceptions and warnings in  the logs if you don't close the source. But it isn't clean; i'll move it back out to the caller of `readLines`

tests/src/test/scala/whisk/core/containerpool/kubernetes/test/KubernetesContainerTests.scala
bwmcadams: UTC? This is one of those European things, like Celsius, and Kilometers, right?

glikson:
should have "kubernetes" label?

PR #3220, opened 2018-01-25
Move log message into a closure to defer string interpolation.

rabbah:
A lot of log messages are logged at debug level and use string interpolation. By moving the message into a closure, for debug level messages which are generally not logged, the interpolation is deferred and then eschewed.

Apply the same to other logging methods and transaction markers.

### DIFF
--- a/common/scala/src/main/scala/whisk/common/Logging.scala
+++ b/common/scala/src/main/scala/whisk/common/Logging.scala
@@ -34,9 +34,9 @@ trait Logging {
    * Prints a message on DEBUG level
    *
    * @param from Reference, where the method was called from.
-   * @param message Message to write to the log
+   * @param message Message to write to the log if not empty
    */
-  def debug(from: AnyRef, message: String)(implicit id: TransactionId = TransactionId.unknown) = {
+  def debug(from: AnyRef, message: => String)(implicit id: TransactionId = TransactionId.unknown) = {
     if (id.meta.extraLogging) {
       emit(InfoLevel, id, from, message)
     } else {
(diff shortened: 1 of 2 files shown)


markusthoemmes:
Easier as I thought it will be. For those curious if this really defers even the interpolation:

```scala
object Test {
    def test(arg: => String, print: Boolean): Unit = {
      if (print) println(arg) else ()
    }

    var i = 0
    test(s"${i += 1}", false)
    test(s"${i += 1}", true)
    println(i) // prints 1
  }
}
```

rabbah:
Made a few more tweaks.

rabbah:
@starpit if you load test this, can you give it a 👍 or 👎, 🙏.

cbickel:
👍  Great PR. 
I did some load measurements and saw an increase of trigger get of 11%.

mbehrendt:
excellent. @cbickel can you pls post the absolute numbers of your measurements?

starpit:
i discovered this by looking at stack traces, after a bunch of other tuning opportunities. i suspect that the benefit of this will grow, as we start shaving off other, larger, inefficiencies. 

right now, on my 44-way system, for echo invokes, i'm seeing no major difference in either throughput, p25, p50, p90, p99, or max. there is a small benefit in p25, under light load. master, for reference, gives me 2500 echo invokes per second, which is pretty far short of the 7000 or so i was getting after a bunch of other changes.

this is using the defaults for all settings, except specifying the use of `runc`, and upping the limits.

i'll start portioning in some other changes.

PR #3221, opened 2018-01-26
Fix overload of invoker.

cbickel:
Due to duplicate names of variables, the loadbalancer thought, that the home invoker of an action has still some free slots. But the controller asked for the wrong invoker, so the invoker was totally overloaded.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/ContainerPoolBalancer.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/ContainerPoolBalancer.scala
@@ -43,9 +43,9 @@ import scala.util.{Failure, Success}
 
 case class LoadbalancerConfig(blackboxFraction: Double, invokerBusyThreshold: Int)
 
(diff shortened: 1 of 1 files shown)


core/controller/src/main/scala/whisk/core/loadBalancer/ContainerPoolBalancer.scala
```diff
@@ -299,7 +303,7 @@ class ContainerPoolBalancer(config: WhiskConfig, instance: InstanceId)(implicit
         val invokersToUse = if (action.exec.pull) blackboxInvokers(invokers) else managedInvokers(invokers)
         val invokersWithUsage = invokersToUse.view.map {
           // Using a view defers the comparably expensive lookup to actual access of the element
-          case invoker => (invoker.id, invoker.status, currentActivations.getOrElse(instance.toString, 0))
+          case invoker => (invoker.id, invoker.status, currentActivations.getOrElse(invoker.id.toString, 0))
```
cbickel: This line was the issue.
Instead of the invoker-instance, the controller instance has been used.

PR #3223, opened 2018-01-26
Remove kafka related debug message

mhenke1:
With enhanced kafka message sizes this debug message writes an enormous amount of data.
Since enough information is printed out in error cases, we can simply remove it.


dubeejw:
PG3 1743 ⏳

PR #3225, opened 2018-01-27
reduce fillInStackTrace expenses due to ArgNormalizer

starpit:
ArgNormalizer uses a try/catch pattern when parsing entity names. in two common cases, the arg to be normalized is known to be a string. we can fast-path this case. i called this `DocId.direct()`

the expense of this can be witnessed from call stack samples: you will see `fillInStackTrace` due to exceptions being thrown by spray's json parser.

preliminary experiments show that this results in more stable performance, i.e. with improvements principally to p99 latency numbers.

```
           p25   p50   p75    p90    p99     max
master     20ms  23ms  28ms   40ms   270ms  1.7s
this PR    18ms  21ms  26ms   35ms   85ms   1.4s
```

test protocol: using wrk -t8 -c800; wipe.yml, then warmup 60s, then timed run 120s
endpoint: blocking invokes of echo action
deployment: environments/local, local couchdb, 44-way with 48GB memory; 12 invokers
load driver: 16-way located in the same rack

PG2/2715  🔵

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/DocInfo.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/DocInfo.scala
@@ -37,7 +37,7 @@ import whisk.core.entity.ArgNormalizer.trim
  *
  * @param id the document id, required not null
  */
-protected[core] class DocId private (val id: String) extends AnyVal {
+protected[core] class DocId(val id: String) extends AnyVal {
   def asString = id // to make explicit that this is a string conversion
   protected[core] def asDocInfo = DocInfo(this)
   protected[core] def asDocInfo(rev: DocRevision) = DocInfo(this, rev)
--- a/common/scala/src/main/scala/whisk/core/entity/FullyQualifiedEntityName.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/FullyQualifiedEntityName.scala
@@ -48,7 +48,7 @@ protected[core] case class FullyQualifiedEntityName(path: EntityPath, name: Enti
    */
   def add(n: EntityName) = FullyQualifiedEntityName(path.addPath(name), n)
 
-  def toDocId = DocId(qualifiedName)
+  def toDocId = new DocId(qualifiedName)
   def namespace: EntityName = path.root
   def qualifiedNameWithLeadingSlash: String = EntityPath.PATHSEP + qualifiedName
   def asString = path.addPath(name) + version.map("@" + _.toString).getOrElse("")
--- a/core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
@@ -159,7 +159,7 @@ protected[actions] trait PrimitiveActions {
     // 2. failing active ack (due to active ack timeout), fall over to db polling
     // 3. timeout on db polling => converts activation to non-blocking (returns activation id only)
     // 4. internal error message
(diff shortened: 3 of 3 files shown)


starpit:
for future reference, here is an example call stack for one of the two call sites:

```
	at java.lang.Throwable.fillInStackTrace(Native Method)
	at java.lang.Throwable.fillInStackTrace(Throwable.java:783)
	- locked <0x00000000e2187818> (a spray.json.JsonParser$ParsingException)
	at java.lang.Throwable.<init>(Throwable.java:265)
	at java.lang.Exception.<init>(Exception.java:66)
	at java.lang.RuntimeException.<init>(RuntimeException.java:62)
	at spray.json.JsonParser$ParsingException.<init>(JsonParser.scala:31)
	at spray.json.JsonParser.fail(JsonParser.scala:213)
	at spray.json.JsonParser.value(JsonParser.scala:64)
	at spray.json.JsonParser.parseJsValue(JsonParser.scala:43)
	at spray.json.JsonParser$.apply(JsonParser.scala:28)
	at spray.json.PimpedString.parseJson(package.scala:45)
	at whisk.core.entity.ArgNormalizer$$anonfun$factory$1.apply(ArgNormalizer.scala:30)
	at whisk.core.entity.ArgNormalizer$$anonfun$factory$1.apply(ArgNormalizer.scala:30)
	at scala.util.Try$.apply(Try.scala:192)
	at whisk.core.entity.ArgNormalizer$class.factory(ArgNormalizer.scala:30)
	at whisk.core.entity.DocId$.factory(DocInfo.scala:107)
	at whisk.core.entity.ArgNormalizer$class.apply(ArgNormalizer.scala:45)
	at whisk.core.entity.DocId$.apply(DocInfo.scala:107)
	at whisk.core.entity.FullyQualifiedEntityName.toDocId(FullyQualifiedEntityName.scala:51)
```

mbehrendt:
great finding! Thx @starpit

markusthoemmes:
Woa. That actually looks like loads of overhead generated without us knowing it. @rabbah could we possibly just ditch `ArgNormalizer`? We shouldn't have any `null` values anyway and I'm not aware of any values where we'd need to trim?

rabbah:
Yup. I made the same comment to @starpit.

common/scala/src/main/scala/whisk/core/entity/DocInfo.scala
```diff
@@ -105,6 +105,7 @@ case class BulkEntityResult(id: String,
 }
 
 protected[core] object DocId extends ArgNormalizer[DocId] {
```
rabbah: Does it make sense instead to make constructor public and call it directly?

dubeejw:
Quick reference of everything that uses ArgNormalizer: https://github.com/apache/incubator-openwhisk/search?p=1&q=%22ArgNormalizer%22&type=&utf8=%E2%9C%93.

Double whammy here? `val activationId = Try { ActivationId(n) }`
https://github.com/apache/incubator-openwhisk/blob/20a5a6083d1630037740b4cf62f0d4e8fbc5c829/core/controller/src/main/scala/whisk/core/controller/Activations.scala#L95

starpit:
made the constructor public, and got rid of `DocId.direct`. i didn't change any other call sites (they don't matter, for performance, yet at least)

starpit:
@dubeejw this is a problem, though indirectly. i've opened a separate issue for it: #3228

i'll work out a solution tomorrow. it should offer a nice improvement, if my investigations are on target.

PR #3231, opened 2018-01-29
update docs for ulimit and pid-limit

csantanapr:


docs/reference.md
```diff
@@ -484,14 +484,14 @@ The following table lists the default limits for actions.
...
+* The maximum number of open files is 1024 (for both hard and soft limits).
+* The docker run command use the argument `--ulimit nofile=1024:1024`.
 * For more information about the ulimit for open files see the [docker run](https://docs.docker.com/engine/reference/commandline/run) documentation.
 
-### Per Docker action processes ulimit (Fixed: 512:512)
-* The maximum number of processes available to a user is 512 (for both hard and soft limits).
-* The docker run command use the argument `--ulimit nproc=512:512`.
+### Per Docker action processes ulimit (Fixed: 1024)
+* The maximum number of processes available to the action container is 1024.
+* The docker run command use the argument `--pids-limit 1024`.
```
jpkessle: Should the argument contain an `=` sign ? Like this:

The docker run command use the argument `--pids-limit=1024`.

csantanapr: It doesn't have `=` it's space

csantanapr:
>@csantanapr, seems like this value should be configurable during deployment here https://github.com/apache/incubator-openwhisk/blob/master/core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala#L82. Can you open an issue?

@dubeejw issue created here https://github.com/apache/incubator-openwhisk/issues/3235

good to merge?

PR #3232, opened 2018-01-29
Making prewarm kind (and count) configurable

Himavanth:
This is to make configurable the prewarm kind (ex:nodejs:6) and prewarm count (number of containers the invoker prewarms) which is hardcoded to 2 right now.

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -36,6 +36,11 @@ whisk:
 #
 runtimesManifest: "{{ runtimes_manifest | default(lookup('file', '{{ openwhisk_home }}/ansible/files/runtimes.json') | from_json) }}"
 
+##
+# Prewarm configuration
+prewarmKind: "{{prewarm_kind | default('nodejs:6')}}"
+prewarmCount: "{{prewarm_count | default('2')}}"
+
 limits:
   invocationsPerMinute: "{{ limit_invocations_per_minute | default(60) }}"
   concurrentInvocations: "{{ limit_invocations_concurrent | default(30) }}"
--- a/ansible/templates/whisk.properties.j2
+++ b/ansible/templates/whisk.properties.j2
@@ -35,6 +35,9 @@ whisk.api.vanity.subdomain.parts=1
 
 runtimes.manifest={{ runtimesManifest | to_json }}
 
+prewarm.kind={{ prewarmKind | default('nodejs:6') }}
+prewarm.count={{ prewarmCount | default('2') }}
+
 limits.actions.invokes.perMinute={{ limits.invocationsPerMinute }}
 limits.actions.invokes.concurrent={{ limits.concurrentInvocations }}
 limits.actions.invokes.concurrentInSystem={{ limits.concurrentInvocationsSystem }}
--- a/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
+++ b/common/scala/src/main/scala/whisk/core/WhiskConfig.scala
@@ -90,6 +90,9 @@ class WhiskConfig(requiredProperties: Map[String, String],
   val mainDockerEndpoint = this(WhiskConfig.mainDockerEndpoint)
 
   val runtimesManifest = this(WhiskConfig.runtimesManifest)
+  val prewarmKind = this(WhiskConfig.prewarmKind)
+  val prewarmCount = this(WhiskConfig.prewarmCount)
+
   val actionInvokePerMinuteLimit = this(WhiskConfig.actionInvokePerMinuteLimit)
   val actionInvokeConcurrentLimit = this(WhiskConfig.actionInvokeConcurrentLimit)
   val triggerFirePerMinuteLimit = this(WhiskConfig.triggerFirePerMinuteLimit)
@@ -218,6 +221,8 @@ object WhiskConfig {
   val zookeeperHosts = Map(zookeeperHostList -> null)
 
   val runtimesManifest = "runtimes.manifest"
+  val prewarmKind = "prewarm.kind"
+  val prewarmCount = "prewarm.count"
 
   val actionSequenceMaxLimit = "limits.actions.sequence.maxLength"
   val actionInvokePerMinuteLimit = "limits.actions.invokes.perMinute"
--- a/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
@@ -67,7 +67,8 @@ object Invoker {
       invokerNumCore -> "4",
       invokerCoreShare -> "2",
       invokerUseRunc -> "true") ++
-      Map(invokerName -> "")
+      Map(invokerName -> "") ++
+      Map(prewarmKind -> "nodejs:6", prewarmCount -> "2")
 
   def main(args: Array[String]): Unit = {
     Kamon.start()
--- a/core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
@@ -141,7 +141,8 @@ class InvokerReactive(config: WhiskConfig, instance: InstanceId, producer: Messa
   val childFactory = (f: ActorRefFactory) =>
     f.actorOf(ContainerProxy.props(containerFactory.createContainer, ack, store, logsProvider.collectLogs, instance))
 
-  val prewarmKind = "nodejs:6"
+  val prewarmKind = config.prewarmKind
+  val prewarmCount = config.prewarmCount.toInt
   val prewarmExec = ExecManifest.runtimesManifest
     .resolveDefaultRuntime(prewarmKind)
     .map { manifest =>
@@ -155,7 +156,7 @@ class InvokerReactive(config: WhiskConfig, instance: InstanceId, producer: Messa
       maximumContainers,
       maximumContainers,
       activationFeed,
-      Some(PrewarmingConfig(2, prewarmExec, 256.MB))))
+      Some(PrewarmingConfig(prewarmCount, prewarmExec, 256.MB))))
 
   /** Is called when an ActivationMessage is read from Kafka */
   def processActivationMessage(bytes: Array[Byte]): Future[Unit] = {


ansible/group_vars/all
```diff
@@ -36,6 +36,11 @@ whisk:
 #
 runtimesManifest: "{{ runtimes_manifest | default(lookup('file', '{{ openwhisk_home }}/ansible/files/runtimes.json') | from_json) }}"
 
+##
+# Prewarm configuration
+prewarmKind: "{{prewarm_kind | default('nodejs:6')}}"
```
rabbah: we've increasingly moved away from ansible configs to use pure config: refer to this doc https://github.com/apache/incubator-openwhisk/blob/master/docs/dev/configuration.md

it makes it more convenient in general for kube and mesos based deployments downstream.

ansible/templates/whisk.properties.j2
```diff
@@ -35,6 +35,9 @@ whisk.api.vanity.subdomain.parts=1
 
 runtimes.manifest={{ runtimesManifest | to_json }}
 
+prewarm.kind={{ prewarmKind | default('nodejs:6') }}
```
rabbah: don't think we'll need these until they play into a test which shouldn't be the case.

rabbah:
I'd suggest in addition making the configuration a set of stem cells (vs a single kind). It might look something like this: `prewarm: [{ "kind": "nodejs:6", "stemcells": 2}, { "kind": "python:3", "stemcells": 2}]`

rabbah:
Thanks for the contribution - do you have an Apache ICLA on file?

hsane2001:
This will be a great addition. I believe the prewarm conditions include memory limits besides kind of container, can this be configurable as well? Will this code be checked in?

ningyougang:
@rabbah @Himavanth 
Should we support multiple prewarm kinds?(It seems current code supports only one kind)

rabbah:
Yes. That was my suggestion above (https://github.com/apache/incubator-openwhisk/pull/3232#issuecomment-361311688).

rabbah:
We should seperate the multiple cell support into a seperate PR. Here it sufficies to export the kind to a pure config setting but make no semantic changes in the invoker.

Himavanth:
Thanks @rabbah for letting me know about the PureConfig doc. I have made the changes required.
I have also submitted the ICLA to secretary@apache.org I am hoping thats the right one and waiting for a reply.

hsane2001:
Hi, Thanks for this change. I am looking to be able to keep prewarmed containers spawned from custom images alive. Currently these come up as kind:blackbox and I cant employ the same method above to keep them warm.

ansible/templates/whisk.properties.j2
rabbah: @markusthoemmes should we eschew these?

ansible/group_vars/all
rabbah: @markusthoemmes should we include these as examples?

common/scala/src/main/resources/application.conf
```diff
@@ -118,7 +118,11 @@ whisk {
         bypass-pull-for-local-images = false
         local-image-prefix = "whisk"
     }
-
+    # prewarm configuration
+    prewarm {
```
rabbah: i think we should nest these under `invoker`.

rabbah: or above under runtimes.

rabbah:
@hsane2001  can you open an issue to discuss the desired feature. I think the approach here establishes a bit of the foundation to do what you like.

hsane2001:
Done -> https://github.com/apache/incubator-openwhisk/issues/3389

jiangpengcheng:
is this pr alive?

rabbah:
Yes. We’d like to get this in for sure.

rabbah:
Thinking through this a bit more - because it will have breaking changes to downstream projects, I think we should do the break change once. Let's move the stem cells to the runtime manifest. 

We could do this by adding an additional property for a kind, `stemCells: [ {memory, count} ]` since the kind already specifies the image name. The invoker's `preWarmConfig` is:
```
case class PrewarmingConfig(count: Int, exec: CodeExec[_], memoryLimit: ByteSize)
```
so it is sufficient to add the fields I note above.

jiangpengcheng:
Hi @rabbah , do you mean adding an additional property in `runtime.manifests` ?

after that the `runtime.manifests` will be like:

```json
{
    "runtimes": {
        "nodejs": [
            {
                "kind": "nodejs",
                "image": {
                    "name": "nodejsaction"
                },
                "deprecated": true,
                "stemCells": [{"memory": 256, "count": 1}, {"memory": 512, "count": 1}]
            }
        ]
    }
}
```

ningyougang:
It is better to support multiple prewarm kind like @rabbah already said.
for example like below:
[{ "kind": "nodejs:6", "stemcells": 2}, { "kind": "python:3", "stemcells": 2}]

Himavanth:
@rabbah Thanks for the feedback. Do you think the prewarming should be for every kind or only the default one under each family.

tysonnorris:
I think it would be useful to have both per-kind and multiple. Obviously we can't guess all the variants that will be used realistically, but implementing these gives significant flexibility, e.g. to tune the deployment to have lots of small prewarm, plus a few large ones, etc. 

The potential problem for multiple prewarms that vary by memory is that you may end up running "small" (memory footprint) actions on a "large" container, if you only test whether the action memory < prewarm memory. This may have some odd results, like creating many large footprint containers that are only used for small actions, and causing large actions to wait for cold starts. 

It may be worthwhile to add another config property, like `minMemory` where the prewarm will ONLY be used if `prewarm.minMemory < action.limits.memory < prewarm.memory'. In this case, you could have a few large prewarms around, that will actually be reserved for running actions that are close in memory footprint requirements. 

In the spirit of "change it once", this may be a good time to introduce action resource attributes as well. I think this is needed in actions so that, for example, an action can signal that it needs to run on a gpu host, or consumes a relatively large amount of cpu, etc. If an action can signal these, they will need to be configurations within the runtime manifest as well. This is a bigger change though, so may be worthwhile to settle for "change it twice"? I will open up an issue for this (if one doesn't exist)

csantanapr:
@Himavanth what’s the latest on this PR, I’m interested on this feature

common/scala/src/main/resources/application.conf
```diff
@@ -141,11 +141,11 @@ whisk {
         bypass-pull-for-local-images = false
         local-image-prefix = "whisk"
     }
-    # prewarm configuration
-    prewarm {
-        kind = "nodejs:6"
-        count = 2
+
+    user-events {
+        enabled = false
```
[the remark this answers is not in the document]
Himavanth: I believe this is from a previous merge. https://github.com/apache/incubator-openwhisk/pull/3552 .

Himavanth:
@rabbah I have added the tests. Do you think this would suffice?
@csantanapr Glad to know. I was working on tests and scalafmt issues(new here). Done now.

csantanapr:
Thanks for the updates @Himavanth 👍

rabbah:
@csantanapr this will need a PG. I'll go through it today or this weekend but it's good enough for jenkins.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -204,26 +204,25 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
...
-            freePool = freePool + (ref -> data)
-            prewarmedPool = prewarmedPool - ref
-            // Create a new prewarm container
-            prewarmContainer(config.exec, config.memoryLimit)
+  def takePrewarmContainer(action: ExecutableWhiskAction): Option[(ActorRef, ContainerData)] = {
+    val kind = action.exec.kind
+    val memory = action.limits.memory.megabytes.MB
+    prewarmedPool
+      .find {
+        case (_, PreWarmedData(_, `kind`, `memory`)) => true
```
tysonnorris: This will use the prewarm IFF the prewarm container is an exact match for memory req. Is that intended? 

I wonder if it should be `if prewarm-memory > action memory > prewarm-memory - memory-match-threshold`
So that
* there DOES NOT need to be an exact match of action memory req + prewarm memory
* there  DOES need to be a "close match", dictated by a configurable memory-match-threshold (so that a 20MB action does not land in a 4GB prewarm)

I'm not sure that configuration should be global vs per-prewarm config (which would require additional change in the manifest)

WDYT?

csantanapr:
From reading the source code, it will use the kind set by default.
So it looks what needs to be specified in the ansible manifest is for `nodejs` json structure a `stemCells` with `count` and `memory` like (2, 256M)

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
rabbah: @tysonnorris this is semantic preserving - the stem cell today is only used if the memory slot matches exactly. I suggest opening a separate issue to configure a memory slop factor for the stem cell matching.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -204,26 +204,25 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
...
+        case _                                       => false
+      }
+      .map {
+        case (ref, data) =>
+          // Move the container to the usual pool
+          freePool = freePool + (ref -> data)
+          prewarmedPool = prewarmedPool - ref
+          // Create a new prewarm container
+          prewarmContainer(action.exec, memory)
```
rabbah: i don't like the use of action.exec here - it leaves the door open to leaking init code.

rabbah:
@Himavanth take a look at this branch:
https://github.com/apache/incubator-openwhisk/compare/master...rabbah:stemcells?expand=1

I rebased this PR to master, resolved conflicts, and made a few changes, organized as separate commits. If you're agreeable, I can open a PR to supersede this one.

Himavanth:
Sounds good. Thanks @rabbah 👍

PR #3240, opened 2018-02-01
Add a loadbalancer with local state and horizontal invoker sharding.

markusthoemmes:
The current ContainerPoolBalancer suffers a couple of problems and bottlenecks:

1. **Inconsistent state:** The data-structures keeping the state for that loadbalancer are not thread-safely handled, meaning there can be queuing to some invokers even though there is free capacity on other invokers.
2. **Asynchronously shared state:** Sharing the state is needed for a high-available deployment of multiple controllers and for horizontal scale in those. Said state-sharing makes point 1 even worse and isn't anywhere fast enough to be able to efficiently schedule quick bursts.
3. **Bottlenecks:** Getting the state from the outside (like for the ActivationThrottle) is a very costly operation (at least in the sharedstate case) and actually bottlenecks the whole invocation path. Getting the current state of the invokers is a second bottleneck, where one request is made to the corresponding actor for each invocation.

This new implementation aims to solve the problems mentioned above as follows:

1. **All state is local:** There is no shared state. Resources are managed through horizontal sharding. Horizontal sharding means: The invokers' slots are evenly divided between the loadbalancers in existence. If we deploy 2 loadbalancers and each invoker has 16 slots, each of the loadbalancers will have access to 8 slots on each invoker.
2. **Slots are given away atomically:** When scheduling an activation, the slot is immediatly assigned to that activation (implemented through Semaphores). That means: Even in concurrent schedules, there will not be an overload on an invoker as long as there is capacity left on that invoker.
3. **Asynchronous updates of slow data:** Slowly changing data, like a change in the invoker's state, is asynchronously handled and updated to a local version of the state. Querying the state is as cheap as it can be.

### Todo:

- [x] Tests for the Semaphore

### DIFF
--- a/common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
+++ b/common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
@@ -40,6 +40,7 @@ class ForcableSemaphore(maxAllowed: Int) {
 
     def permits: Int = getState
 
+    /** Merely copied from `java.util.concurrent.Semaphore` and translated into a tail-recursive method */
     @tailrec
     override final def tryReleaseShared(releases: Int): Boolean = {
       val current = getState
@@ -54,6 +55,7 @@ class ForcableSemaphore(maxAllowed: Int) {
       }
     }
 
+    /** Merely copied from `java.util.concurrent.Semaphore` and translated into a tail-recursive method */
     @tailrec
     final def nonFairTryAcquireShared(acquires: Int): Int = {
       val available = getState
@@ -65,6 +67,10 @@ class ForcableSemaphore(maxAllowed: Int) {
       }
     }
 
+    /**
+     * Basically the same as `nonFairTryAcquireShared`, but does bound to a minimal value of 0 so permits can get
+     * negative.
+     */
     @tailrec
     final def forceAquireShared(acquires: Int): Unit = {
       val available = getState
--- a/common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
+++ b/common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
@@ -40,12 +40,13 @@ class ForcableSemaphore(maxAllowed: Int) {
 
     def permits: Int = getState
 
+    /** Try to release a permit and return whether or not that operation was successful. */
     @tailrec
     override final def tryReleaseShared(releases: Int): Boolean = {
       val current = getState
       val next = current + releases
-      if (next < current) { // overflow
-        throw new Error("Maximum permit count exceeded")
+      if (next < current) { // integer overflow
+        throw new Error("Maximum permit count exceeded, permit variable overflowed")
       }
       if (compareAndSetState(current, next)) {
         true
@@ -54,6 +55,10 @@ class ForcableSemaphore(maxAllowed: Int) {
       }
     }
 
+    /**
+     * Try to acquire a permit and return whether or not that operation was successful. Requests may not finish in FIFO
+     * order, hence this method is not necessarily fair.
+     */
     @tailrec
     final def nonFairTryAcquireShared(acquires: Int): Int = {
       val available = getState
@@ -88,6 +93,7 @@ class ForcableSemaphore(maxAllowed: Int) {
    * @return `true`, iff the internal semaphore's number of permits is positive, `false` if negative
    */
   def tryAcquire(acquires: Int = 1): Boolean = {
+    require(acquires > 0, "cannot acquire negative or no permits")
     sync.nonFairTryAcquireShared(acquires) >= 0
   }
 
@@ -99,6 +105,7 @@ class ForcableSemaphore(maxAllowed: Int) {
    * @param acquires the number of permits to get
    */
   def forceAcquire(acquires: Int = 1): Unit = {
+    require(acquires > 0, "cannot force acquire negative or no permits")
     sync.forceAquireShared(acquires)
   }
 
@@ -108,6 +115,7 @@ class ForcableSemaphore(maxAllowed: Int) {
    * @param acquires the number of permits to release
    */
   def release(acquires: Int = 1): Unit = {
+    require(acquires > 0, "cannot release negative or no permits")
     sync.releaseShared(acquires)
   }
 
--- a/common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
+++ b/common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
@@ -40,7 +40,6 @@ class ForcableSemaphore(maxAllowed: Int) {
 
     def permits: Int = getState
 
-    /** Merely copied from `java.util.concurrent.Semaphore` and translated into a tail-recursive method */
     @tailrec
     override final def tryReleaseShared(releases: Int): Boolean = {
       val current = getState
@@ -55,7 +54,6 @@ class ForcableSemaphore(maxAllowed: Int) {
       }
     }
 
-    /** Merely copied from `java.util.concurrent.Semaphore` and translated into a tail-recursive method */
     @tailrec
     final def nonFairTryAcquireShared(acquires: Int): Int = {
       val available = getState
--- a/common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
+++ b/common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
@@ -1,0 +1,112 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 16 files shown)


markusthoemmes:
Note for the kind reviewers:

The code to process active acks etc. is mostly unchanged. Only the bits to reflect updates to state are updated. I have however reordered them to give the file a bit more structure.

core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -0,0 +1,439 @@
...
+   * Updates the size of a cluster. Throws away all state for simplicity.
+   *
+   * This is okay to not happen atomically, since a dirty read of the values set are not dangerous. At worst the
+   * scheduler works on outdated invoker-load data which is acceptable.
+   *
+   * It is important that this method does not run concurrently to itself and/or to `updateState`
+   */
+  def updateCluster(newSize: Int): Unit = {
+    val actualSize = newSize max 1 // if a cluster size < 1 is reported, falls back to a size of 1 (alone)
+    val newTreshold = totalInvokerThreshold / actualSize
```
vvraskin: Considering a hypothetical case when the number of available controllers get > then the threshold, then we won't be able to schedule. Should we also set the value to `val newTreshold = (totalInvokerThreshold / actualSize) max 1` as above?

markusthoemmes: Ah good one! Yes we should definitly document this and safe-guard the code! Should also add to the docu that the number **must** be divisible.

core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -0,0 +1,439 @@
...
+  override def invokerHealth(): Future[IndexedSeq[InvokerHealth]] = Future.successful(schedulingState.invokers)
+  override def activeActivationsFor(namespace: UUID): Future[Int] =
+    Future.successful(activationsPerNamespace.get(namespace).map(_.intValue()).getOrElse(0))
+  override def totalActiveActivations: Future[Int] = Future.successful(totalActivations.intValue())
+
+  /** 1. Publish a message to the loadbalancer */
+  override def publish(action: ExecutableWhiskActionMetaData, msg: ActivationMessage)(
+    implicit transid: TransactionId): Future[Future[Either[ActivationId, WhiskActivation]]] = {
+
+    val invokersToUse = if (action.exec.pull) schedulingState.blackboxInvokers else schedulingState.managedInvokers
```
rabbah: Flip for common case?

core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -0,0 +1,439 @@
...
+  private def sendActivationToInvoker(producer: MessageProducer,
+                                      msg: ActivationMessage,
+                                      invoker: InstanceId): Future[RecordMetadata] = {
+    implicit val transid: TransactionId = msg.transid
+
+    val topic = s"invoker${invoker.toInt}"
+    val start = transid.started(
+      this,
+      LoggingMarkers.CONTROLLER_KAFKA,
+      s"posting topic '$topic' with activation id '${msg.activationId}'",
```
rabbah: Pick up message from @mhenkes ‘s pr?

common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
```diff
@@ -0,0 +1,118 @@
...
+ * negative value. Getting permits using `tryAcquire` will only be possible once the permits value is in a positive
+ * state again.
+ *
+ * As this is (now) only used for the loadbalancer's scheduling, this does not implement the "whole" Java Semaphore's
+ * interface but only the methods needed.
+ *
+ * @param maxAllowed maximum number of permits given away by `tryAcquire`
+ */
+class ForcableSemaphore(maxAllowed: Int) {
+  class Sync extends AbstractQueuedSynchronizer {
```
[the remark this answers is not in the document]
markusthoemmes: It's nested to hide the innerts of `AbstractQueuedSynchronizer` (which are nasty) from the general interface.

tests/src/test/scala/whisk/common/ForcableSemaphoreTests.scala
```diff
@@ -0,0 +1,76 @@
...
+  it should "allow to acquire the defined amount of permits only" in {
+    val s = new ForcableSemaphore(2)
+    s.tryAcquire() shouldBe true // 1 permit left
+    s.tryAcquire() shouldBe true // 0 permits left
+    s.tryAcquire() shouldBe false
+
+    val s2 = new ForcableSemaphore(4)
+    s2.tryAcquire(5) shouldBe false // only 4 permits available
+    s2.tryAcquire(3) shouldBe true // 1 permit left
+    s2.tryAcquire(2) shouldBe false // only 1 permit available
```
vvraskin: To fortify the test coverage should we also try to acquire 0 and negative permits here?

common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
```diff
@@ -0,0 +1,116 @@
...
+      if (next < current) { // overflow
+        throw new Error("Maximum permit count exceeded")
+      }
+      if (compareAndSetState(current, next)) {
+        true
+      } else {
+        tryReleaseShared(releases)
+      }
+    }
+
```
vvraskin: I think short comments would be a boon here, what do you think?

common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
```diff
@@ -0,0 +1,116 @@
...
+    setState(maxAllowed)
+
+    def permits: Int = getState
+
+    @tailrec
+    override final def tryReleaseShared(releases: Int): Boolean = {
+      val current = getState
+      val next = current + releases
+      if (next < current) { // overflow
+        throw new Error("Maximum permit count exceeded")
```
vvraskin: As far as I understood, this case will only happen when `releases` is a negative value. Maybe better to return false meaning that the counter was not updated, or clarify the exception message?

common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
```diff
@@ -0,0 +1,116 @@
...
+      } else {
+        tryReleaseShared(releases)
+      }
+    }
+
+    @tailrec
+    final def nonFairTryAcquireShared(acquires: Int): Int = {
+      val available = getState
+      val remaining = available - acquires
+      if (remaining < 0 || compareAndSetState(available, remaining)) {
```
vvraskin: I think it would make sense to split this condition and add a nested _if_ for `compareAndSetState(available, remaining))`.
Given that we have 4 acquires available,say we ask for 5, given that there was no concurrent write it will update the state and return the `remaining` which is -1. It could cause to deadlock when no releases would follow, what do you think?

common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
markusthoemmes: As the comment indicates, this catches an integer overflow (overflowing numbers will become negative), so I'd say an exception is warranted.

common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
vvraskin: Nice, was not aware of short circuiting 👍

ningyougang:
LGTM

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -82,42 +80,63 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
...
+   * limit, so it needs to be divided between the parties who want to perform that check.
+   *
+   * @param defaultLimit the default limit across the whole system
+   * @param user the user to apply that limit to
+   * @return a calculated limit
+   */
+  private def calculateIndividualLimit(defaultLimit: Int, overrideLimit: Identity => Option[Int])(
+    user: Identity): Int = {
+    val limit = calculateLimit(defaultLimit, overrideLimit)(user)
+    limit / loadBalancer.clusterSize
```
cbickel: If the limit is 1 and we have more than 1 controller, the user won't be able to invoke any actions.

markusthoemmes: Good one!

markusthoemmes:
PG3 1842 🔵

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -82,42 +79,69 @@ protected[core] abstract class EntitlementProvider(config: WhiskConfig, loadBala
...
-      _.limits.firesPerMinute.map(dilateThrottle))
-  private val concurrentInvokeThrottler = new ActivationThrottler(
-    loadBalancer,
-    config.actionInvokeConcurrentLimit.toInt,
-    config.actionInvokeSystemOverloadLimit.toInt)
+      calculateIndividualLimit(config.triggerFirePerMinuteLimit.toInt, _.limits.firesPerMinute))
+
+  private val activationThrottleCalculator = loadBalancer match {
+    // This loadbalancer applies sharding and does not share any state
+    case _: ShardingContainerPoolBalancer => calculateIndividualLimit _
```
tysonnorris: This should be part of the loadbalancer spi interface, instead of testing for a specific impl?

PR #3252, opened 2018-02-06
bump docker runtime for large input, perl

csantanapr:


csantanapr:
PG 3 1800

csantanapr:
Doc updates would be cover by different PR.
I want to do refactor the whole native/docker section.

PR #3253, opened 2018-02-06
Efficiently parse CompletionMessage.

### LINKED ISSUE #3228: Expensive fillInStackTrace when parsing CompletionMessage
The controller spends a large number of CPU cycles filling in stack traces, due to the nature of the `CompletionMessage` and spray's deserialization of scala's `Either`.

First, note how spray uses a try/catch pattern for `Either`, starting with the `Left`, then falling back to the `Right`: https://github.com/spray/spray-json/blob/master/src/main/scala/spray/json/AdditionalFormats.scala#L107

Next, observe that CompletionMessage has an `Either[ActivationId, WhiskActivation]`: https://github.com/apache/incubator-openwhisk/blob/master/common/scala/src/main/scala/whisk/core/connector/Message.scala#L84

The common case, for blocking invokes that complete in less than 60 seconds, is the `Right` of the `Either`.  The implication, for these fast blocking invokes, is an expensive `fillInStackTrace` for every invoke.

The proposed fix is to use a custom serdes for `CompletionMessage`, perhaps reversing the order (trying `Right` first). Better, we should find a way to avoid the try/catch pattern to parsing altogether.

markusthoemmes:
The parsing of `Either` relies on parsing failures in its generalized form. In this specialized case we can circumenvent this and parse only once, without throwing any forced error messages.

Fixes #3228

@starpit Want to have a look? I took the liberty to open the PR.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/connector/Message.scala
+++ b/common/scala/src/main/scala/whisk/core/connector/Message.scala
@@ -95,6 +95,20 @@ case class CompletionMessage(override val transid: TransactionId,
 }
 
 object CompletionMessage extends DefaultJsonProtocol {
+  implicit def eitherResponse =
+    new JsonFormat[Either[ActivationId, WhiskActivation]] {
+      def write(either: Either[ActivationId, WhiskActivation]) = either match {
+        case Right(a) => a.toJson
+        case Left(b)  => b.toJson
+      }
+
+      def read(value: JsValue) = value match {
+        case _: JsString => Left(value.convertTo[ActivationId])
+        case _: JsObject => Right(value.convertTo[WhiskActivation])
+        case _           => deserializationError("could not read CompletionMessage")
+      }
+    }
+
   def parse(msg: String): Try[CompletionMessage] = Try(serdes.read(msg.parseJson))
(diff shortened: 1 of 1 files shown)


common/scala/src/main/scala/whisk/core/connector/Message.scala
```diff
@@ -95,6 +95,20 @@ case class CompletionMessage(override val transid: TransactionId,
...
 object CompletionMessage extends DefaultJsonProtocol {
+  implicit def eitherResponse =
+    new JsonFormat[Either[ActivationId, WhiskActivation]] {
+      def write(either: Either[ActivationId, WhiskActivation]) = either match {
+        case Right(a) => a.toJson
+        case Left(b)  => b.toJson
+      }
+
+      def read(value: JsValue) = value match {
+        case _: JsString => Left(value.convertTo[ActivationId])
```
rabbah: Can you add a comment that we guarantee activation id to be a string even if it’s contains only digits?

markusthoemmes: Done!

starpit:
looks good. i tested this out on my own fork a week ago, and it seemed to remedy the targeted problem.

PR #3254, opened 2018-02-06
Preserve rule status when rule is updated

dubeejw:
Fixes two rule bugs:
1. Updating a rule with `wsk rule update` always returned a status of `active`.
2. After updating a rule with `wsk rule update` the status would be `inactive` when the rule is fetched with `wsk rule get`.

Closes https://github.com/apache/incubator-openwhisk/issues/1840

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Rules.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Rules.scala
@@ -85,20 +85,31 @@ trait WhiskRulesApi extends WhiskCollectionAPI with ReferencedEntities {
         val request = content.resolve(entityName.namespace)
         onComplete(entitlementProvider.check(user, Privilege.READ, referencedEntities(request))) {
           case Success(_) =>
-            putEntity(WhiskRule, entityStore, entityName.toDocId, overwrite, update(request) _, () => {
-              create(request, entityName)
-            }, postProcess = Some { rule: WhiskRule =>
-              val getRuleWithStatus = getTrigger(rule.trigger) map { trigger =>
-                getStatus(trigger, FullyQualifiedEntityName(rule.namespace, rule.name))
-              } map { status =>
-                rule.withStatus(status)
-              }
+            putEntity(
+              WhiskRule,
+              entityStore,
+              entityName.toDocId,
+              overwrite,
+              update(request) _,
+              () => {
+                create(request, entityName)
+              },
+              postProcess = Some { rule: WhiskRule =>
+                if (overwrite == true) {
+                  val getRuleWithStatus = getTrigger(rule.trigger) map { trigger =>
+                    getStatus(trigger, FullyQualifiedEntityName(rule.namespace, rule.name))
+                  } map { status =>
+                    rule.withStatus(status)
+                  }
 
-              onComplete(getRuleWithStatus) {
-                case Success(r) => completeAsRuleResponse(rule, r.status)
-                case Failure(t) => terminate(InternalServerError)
-              }
-            })
+                  onComplete(getRuleWithStatus) {
+                    case Success(r) => completeAsRuleResponse(rule, r.status)
+                    case Failure(t) => terminate(InternalServerError)
+                  }
+                } else {
+                  completeAsRuleResponse(rule, Status.ACTIVE)
+                }
+              })
           case Failure(f) =>
             handleEntitlementFailure(f)
         }
--- a/core/controller/src/main/scala/whisk/core/controller/Rules.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Rules.scala
@@ -286,6 +286,7 @@ trait WhiskRulesApi extends WhiskCollectionAPI with ReferencedEntities {
     val oldTriggerName = rule.trigger
 
     getTrigger(oldTriggerName) flatMap { oldTriggerOpt =>
+      val status = getStatus(oldTriggerOpt, ruleName)
       val newTriggerEntity = content.trigger getOrElse rule.trigger
       val newTriggerName = newTriggerEntity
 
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/system/basic/WskRuleTests.scala
```diff
@@ -88,6 +88,40 @@ abstract class WskRuleTests extends TestHelpers with WskTestHelpers {
...
+      .parseJson
+      .asJsObject
+      .fields
+      .get("status") shouldBe (Some("active".toJson))
+
+    assetHelper.withCleaner(wsk.trigger, triggerName2) { (trigger, name) =>
+      trigger.create(name)
+    }
+
+    assetHelper.withCleaner(wsk.action, actionName2) { (action, name) =>
```
rabbah: You could reuse the same action as trigger one.

markusthoemmes:
I agree, this should be moved to be a Unit test instead.

dubeejw:
Tried to do a unit test originally, but had trouble since the Rule route also needs access to the Trigger route. I'll give it another shot when I get time.

rabbah:
Thanks for the kick - same comment as here: https://github.com/apache/incubator-openwhisk/pull/3187#issuecomment-363464521

Something for us to fix in the future.

core/controller/src/main/scala/whisk/core/controller/Rules.scala
```diff
@@ -313,7 +314,7 @@ trait WhiskRulesApi extends WhiskCollectionAPI with ReferencedEntities {
             WhiskTrigger.put(entityStore, oldTrigger.removeRule(ruleName))
           }
 
-          val triggerLink = ReducedRule(actionName, Status.INACTIVE)
+          val triggerLink = ReducedRule(actionName, status)
```
rabbah: Perhaps we should add a test to enforce an inactive rule does not become active with update which changes the trigger.

active rule: update with new trigger, or update with new action -> rule stays active
inactive rule: update with new trigger, or update with new action -> rule stays inactive

rabbah: i could argue that updating the rule with a new trigger should enable it, matching create.

dubeejw: The `create()` method is made in such a way that `active` is always returned for the rule status even if the rule is not actually in an active state. The updated test will fail because of this. If I used `wsk.rule.get()` instead of checking the `wsk.rule.create()` response, the tests would pass. Maybe we need to use `getStatus()` in the `create()` method.

dubeejw: Referring to this line: https://github.com/dubeejw/openwhisk/blob/5814698cdff360e4d15e451545467982a2d4221f/core/controller/src/main/scala/whisk/core/controller/Rules.scala#L91.

tests/src/test/scala/system/basic/WskRuleTests.scala
```diff
@@ -88,6 +88,61 @@ abstract class WskRuleTests extends TestHelpers with WskTestHelpers {
...
+      .stdout
+      .parseJson
+      .asJsObject
+      .fields
+      .get("status") shouldBe (Some("active".toJson))
+
+    wsk.rule.disable(ruleName)
+
+    // Ensure status stays inactive when rule is updated with its current trigger
+    wsk.rule
```
dubeejw: Fails here, but `wsk.rule.get` would return the proper response.

core/controller/src/main/scala/whisk/core/controller/Rules.scala
rabbah: If creating a rule then it doesn’t exist so get status will not work. Did you mean something else?

dubeejw: `create()` calls `update()` in the controller. We would only check the rule status during an update.

csantanapr:
Bump @dubeejw 
Can you take a look into Travis failure ?
It would be good to get this bug fix landed it’s affecting the wskdeploy tool now

core/controller/src/main/scala/whisk/core/controller/Rules.scala
dubeejw: @rabbah, fixed the tests. Can take care of the problem above when this is done https://github.com/apache/incubator-openwhisk/issues/3274.

core/controller/src/main/scala/whisk/core/controller/Rules.scala
```diff
@@ -85,11 +85,31 @@ trait WhiskRulesApi extends WhiskCollectionAPI with ReferencedEntities {
...
+              WhiskRule,
+              entityStore,
+              entityName.toDocId,
+              overwrite,
+              update(request) _,
+              () => {
+                create(request, entityName)
+              },
+              postProcess = Some { rule: WhiskRule =>
+                if (overwrite == true) {
```
dubeejw: Fetch the rule status only when an overwrite occurs.

dubeejw:
Turned out to be two different bugs.
1. Updating a rule with `wsk rule update` always returned a status of active. 
2. After updating a rule with `wsk rule update` the status would be inactive when the rule is fetched with `wsk rule get`.

tests/src/test/scala/system/basic/WskRuleTests.scala
```diff
@@ -88,6 +88,35 @@ abstract class WskRuleTests extends TestHelpers with WskTestHelpers {
...
+    val inactive = Some("inactive".toJson)
+    val statusPermutations =
+      Seq((triggerName, active), (triggerName, inactive), (triggerName2, active), (triggerName2, inactive))
+
+    ruleSetup(Seq((ruleName, triggerName, (actionName, actionName, defaultAction))), assetHelper)
+    assetHelper.withCleaner(wsk.trigger, triggerName2) { (trigger, name) =>
+      trigger.create(name)
+    }
+
+    statusPermutations.foreach {
```
dubeejw: Makes sure `wsk rule update` and `wsk rule get` return the proper status.

tests/src/test/scala/whisk/core/controller/test/RulesApiTests.scala
```diff
@@ -563,7 +563,7 @@ class RulesApiTests extends ControllerTestCommon with WhiskRulesApi {
         WhiskRuleResponse(
           namespace,
           rule.name,
-          Status.ACTIVE,
+          Status.INACTIVE,
```
dubeejw: Status is set to `INACTIVE` in this file because the trigger route is not accessible in the unit test, resulting in an `INACTIVE` status when a rule is updated.

PR #3255, opened 2018-02-06
Only log outgoing requests.

cbickel:
This commit removes the logging of incoming requests.
The information of these logs are moved to the log of the outgoing requests.
This results in a small performance boost of a few percent.


rabbah:
Can we enable this for debug log levels instead of removing completely?

PR #3256, opened 2018-02-06
Refactor some bits of the triggers API.

markusthoemmes:
Some bits of the triggers API did include duplicated code.  Didn't get around to post comments to the specific PRs so here are my proposed fixes.

@mdeuser would you have a look please?

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
@@ -21,13 +21,12 @@ import java.time.{Clock, Instant}
 
 import scala.collection.immutable.Map
 import scala.concurrent.Future
-import scala.util.{Failure, Success}
-
+import scala.util.{Failure, Try}
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.model.HttpMethods.POST
-import akka.http.scaladsl.model.StatusCodes.{Accepted, BadRequest, InternalServerError, OK}
+import akka.http.scaladsl.model.StatusCodes.{Accepted, BadRequest, InternalServerError, OK, ServerError}
 import akka.http.scaladsl.model.Uri.Path
 import akka.http.scaladsl.model.headers.{Authorization, BasicHttpCredentials}
 import akka.http.scaladsl.model._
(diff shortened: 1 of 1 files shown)


core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -305,59 +292,56 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
       case (ruleName, rule) =>
         // Invoke the action. Retain action results for inclusion in the trigger activation record
-        val actionActivationResult: Future[JsObject] = postActivation(user, rule, args)
+        postActivation(user, rule, args)
           .flatMap { response =>
             response.status match {
               case OK | Accepted =>
                 Unmarshal(response.entity).to[JsObject].map { activationResponse =>
-                  val activationId: JsValue = activationResponse.fields("activationId")
+                  val activationId = activationResponse.fields("activationId").convertTo[ActivationId]
```
rabbah: not sure it's worth it here to convert to ActivationId only to convert it back to json.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -305,59 +292,56 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
-                    } else {
+              case code =>
+                Unmarshal(response.entity).to[String].map { error =>
+                  val failureType = code match {
+                    case _: ServerError => ActivationResponse.WhiskError // all 500s are to be considered whisk errors
+                    case _              => ActivationResponse.ApplicationError
+                  }
+                  val errorMessage: String = Try(error.parseJson.convertTo[ErrorResponse])
+                    .map { e =>
+                      def logMsg = s"trigger-fired action '${rule.action}' failed to invoke with ${e.error}, ${e.code}"
```
rabbah: I'd prefer not to expose the `tid` here since that's an artifact of the current implementation and will go away. If there is a code, the error response should be sufficiently informative.

markusthoemmes: Note this is a refactor. This is copied from the current code.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
markusthoemmes: It's mainly to gain strong types in `RuleActivationResult`, which I'd assume will stay when we refactor to use the internal path?

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
rabbah: You are right - this is just for the log message not the response logged in the activation. Sorry about the confusion.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
rabbah: OK - sold 💰

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -372,58 +356,36 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
-        "action" -> JsString(actionName.asString),
-        "statusCode" -> JsNumber(statusCode),
-        "success" -> JsBoolean(statusCode == ActivationResponse.Success)) ++
-        actionActivationId.map("activationId" -> _.toJson) ++
-        errorMsg.map("error" -> JsString(_)))
+  /** Contains the result of invoking a rule */
+  case class RuleActivationResult(statusCode: Int,
+                                  ruleName: FullyQualifiedEntityName,
+                                  actionName: FullyQualifiedEntityName,
+                                  response: Either[String, ActivationId]) {
```
mdeuser: today these values are mutually exclusive, but perhaps it may be possible for a trigger response to include the activation id as well as an error message..?

markusthoemmes: Well, let's write code for today!

csantanapr:
@markusthoemmes PG?

markusthoemmes:
@csantanapr PG4 1344 🔵

PR #3258, opened 2018-02-07
Encrypt data in motion - Kafka

vvraskin:
Hi,

this is a part of the initiative to encrypt data in motion in OW. 
This PR enables SSL between Controllers, Kafka and Invokers. 
It has a feature flag, so you are not forced to deploy it, if you don’t need it. 

regards, Vadim.


common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
```diff
@@ -84,4 +102,14 @@ object KafkaConfiguration {
...
+
+  def useSSL(props: Properties, httpsConfig: KafkaSslConfig) = {
+    props.put(CommonClientConfigs.SECURITY_PROTOCOL_CONFIG, "SSL")
+    props.put(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, httpsConfig.truststorePath)
+    props.put(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, httpsConfig.truststorePassword)
+  }
+  def authenticateUser(props: Properties, httpsConfig: KafkaSslConfig) = {
+    props.put(SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG, httpsConfig.keystorePath)
+    props.put(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG, httpsConfig.keystorePassword)
+  }
```
markusthoemmes: I lately did an effort to generalize all the kafka settings for the producer into `whisk.kafka.producer`. Could we introduce a `whisk.kafka.clients` or `whisk.kafka.common` which includes settings needed for both the producer and consumer?

I modeled it in a way so that we can just throw in the keys from the Kafka config itself and override anything we want from application.conf without needing code changes. Do you think that's feasible here?

vvraskin: Yeap, sounds feasible. I could have probably seen it somewhere in the code, will update it

common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
```diff
@@ -44,9 +41,17 @@ case class KafkaConfig(replicationFactor: Short)
  * A Kafka based implementation of MessagingProvider
  */
 object KafkaMessagingProvider extends MessagingProvider {
+
+  case class KafkaSslConfig(enabled: String,
+                            authentication: String,
+                            keystorePath: String,
+                            keystorePassword: String,
+                            truststorePath: String,
+                            truststorePassword: String)
```
markusthoemmes: Not needed anymore.

ansible/environments/local/group_vars/all
```diff
@@ -33,3 +33,5 @@ kafka_topics_invoker_retentionBytes: 104857600
 kafka_topics_invoker_retentionMS: 300000
 
 env_hosts_dir: "{{ playbook_dir }}/environments/local"
+
+kafka_protocol: SSL
```
markusthoemmes: Please revert the default to non-ssl.

PR #3259, opened 2018-02-07
use PureConfig for misc. runtimesManifest configuration

### LINKED ISSUE #3245: bypassPullForLocalImages setting no longer overridable
As noted on this PR, https://github.com/apache/incubator-openwhisk/pull/3142#pullrequestreview-93706400 factoring out the runtime manifest to a separate file does not in its current form permit overriding any of the values, including `bypassPullForLocalImages` (this value permits docker pull bypass for locally tagged images).

@dgrove-oss, I think we can fix this a couple of ways: mangle the manifest after reading it from file, or read from pureconfig. What do you think?

Reported by @krugazor on slack.

dgrove-oss:
Refactor runtimes.json so that the misc. simple configuration
fields are specified using PureConfig (whisk.runtimes),
leaving only the more complex arrays of runtimes and blackboxes
in runtime.json.  Since we expect deployments to frequently
override the set of runtimes, it makes sense to leave them as
being defined by a json file that is not included in the
Invoker/Controller docker image (simplify override).  The flags
moved to PureConfig are simple fields, and thus easier to override
via envvars if necessary.

Restores the ability to easily override bypassPullForLocalImages.

Fixes #3245.


ansible/group_vars/all
```diff
@@ -24,16 +24,8 @@ whisk:
...
-#   defaultImagePrefix: the default image prefix when not given explicitly
-#   defaultImageTag: the default image tag
-#   runtimes: set of language runtime families grouped by language (e.g., nodejs, python)
-#   blackboxes: list of pre-populated docker action images as "name" with optional "prefix" and "tag"
-#   bypassPullForLocalImages: optional, if true, allow images with a prefix that matches {{ docker.image.prefix }}
-#                             to skip docker pull in invoker even if the image is not part of the blackbox set
-#
+
+# list of supported runtimes (see whisk.core.entity.ExecManifest for schema)
 runtimesManifest: "{{ runtimes_manifest | default(lookup('file', '{{ openwhisk_home }}/ansible/files/runtimes.json') | from_json) }}"
```
rabbah: is this not moved because it points to a file instead of an inline value?

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -125,6 +125,11 @@
       "CONFIG_whisk_activation_payload_max": "{{ limit_activation_payload | default() }}"
 
       "RUNTIMES_MANIFEST": "{{ runtimesManifest | to_json }}"
+      "CONFIG_whisk_runtimes_defaultImagePrefix": "{{ runtimes_default_image_prefix | default() }}"
+      "CONFIG_whisk_runtimes_defaultImageTag": "{{ runtimes_default_image_tag | default() }}"
+      "CONFIG_whisk_runtimes_bypassPullForLocalImages": "{{ runtimes_bypass_pull_for_local_images | default() }}"
```
rabbah: fwiw we don't actually need this setting in the controller; ok to keep it though.

common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
```diff
@@ -41,11 +43,11 @@ protected[core] object ExecManifest {
...
    *
    * @param config a valid configuration
-   * @param localDockerImagePrefix optional local docker prefix, permitting images matching prefix to bypass docker pull
    * @return the manifest if initialized successfully, or an failure
    */
-  protected[core] def initialize(config: WhiskConfig, localDockerImagePrefix: Option[String] = None): Try[Runtimes] = {
-    val mf = Try(config.runtimesManifest.parseJson.asJsObject).flatMap(runtimes(_, localDockerImagePrefix))
+  protected[core] def initialize(config: WhiskConfig): Try[Runtimes] = {
+    val rmc = loadConfigOrThrow[RuntimeManifestConfig](ConfigKeys.runtimes)
+    val mf = Try(config.runtimesManifest.parseJson.asJsObject).flatMap(runtimes(_, rmc))
```
rabbah: if the manifest itself is moved to pureconfig, then can nuke whisk config here entirely - is that feasible?

dgrove-oss: I think moving the manifest itself to pure config isn't a good idea.  We override it in our deployments and overriding a complex structure via an environment variable is going to be annoying.  Better to keep the manifest itself out of the docker images.

ansible/roles/controller/tasks/deploy.yml
dgrove-oss: yeah, wasn't sure what the best policy was here.  Seems like a latent bug to not wire up the overrides (just in case some later change needs them), but if we don't consistently do that for other flags then I could remove these.

common/scala/src/main/scala/whisk/core/entity/ExecManifest.scala
rabbah: i was thinking more about making the value a path rather than the structure itself - but don't know if that's feasible and if the file path resolution would be an issue.

dgrove-oss:
PG1 / 2590 🏃

dgrove-oss:
PG1 / 2604 🏃

dgrove-oss:
rebased to latest master; PG1/2605 🏃

dgrove-oss:
PG3 / 1841 🔵

PR #3260, opened 2018-02-08
Enhance the RestResult with more information to debug failing tests e…

cbickel:
…asier.

I had to debug a failing test case. But most information from the RestResult was hidden. Only the information of the parent (RunResult) was shown.

### DIFF
--- a/tests/src/test/scala/common/rest/WskRest.scala
+++ b/tests/src/test/scala/common/rest/WskRest.scala
@@ -1531,6 +1531,12 @@ class RestResult(var statusCode: StatusCode, var respData: String = "", blocking
       respData,
       RestResult.convertHttpResponseToStderr(respData)) {
 
+  override def toString: String = {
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/common/rest/WskRest.scala
```diff
@@ -1531,6 +1531,13 @@ class RestResult(var statusCode: StatusCode, var respData: String = "", blocking
       respData,
       RestResult.convertHttpResponseToStderr(respData)) {
 
+  override def toString: String = {
+    s"""${super.toString}statusCode: $statusCode
+respData: $respData
+blocking: $blocking
+    """
+  }
```
markusthoemmes: How about:

```scala
override def toString: String = {
  s"""${super.toString}
      |statusCode: $statusCode
      |respData: $respData
      |blocking: $blocking""".stripMargin
}
```

cbickel:
PG5#95 🔵

PR #3262, opened 2018-02-08
Handle trigger activations with inactive rules

### LINKED ISSUE #3304: firing a trigger with no rules will return an activaiton id but no activation record will be written
recent changes and refactoring in the trigger handler will cause an activation id to be returned but no activation record to be created if the trigger has no matching rules.

mdeuser:
- log trigger activation entry for disabled rules (but at least one active rule is needed to create a trigger activation record)
- if trigger has no associated active rule(s); return a 204.  also no trigger activation record is created

Closes #3304.

### DIFF
--- a/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
+++ b/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
@@ -111,6 +111,11 @@ object Messages {
   val notAllowedOnBinding = "Operation not permitted on package binding."
   def packageNameIsReserved(name: String) = s"Package name '$name' is reserved."
 
+  /** Error messages for triggers */
+  def triggerWithInactiveRule(rule: String, action: String) = {
+    s"Rule ${rule} is inactive; action ${action} was not activated."
+  }
+
   /** Error messages for sequence activations. */
   def sequenceRetrieveActivationTimeout(id: ActivationId) =
     s"Timeout reached when retrieving activation $id for sequence component."
--- a/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
+++ b/common/scala/src/main/scala/whisk/http/ErrorResponse.scala
@@ -113,7 +113,7 @@ object Messages {
 
   /** Error messages for triggers */
   def triggerWithInactiveRule(rule: String, action: String) = {
-    s"Rule ${rule} is inactive; action ${action} was not activated."
+    s"Rule '$rule' is inactive, action '$action' was not activated."
   }
 
   /** Error messages for sequence activations. */
--- a/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Triggers.scala
@@ -26,7 +26,7 @@ import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.model.HttpMethods.POST
-import akka.http.scaladsl.model.StatusCodes.{Accepted, BadRequest, InternalServerError, OK, ServerError}
+import akka.http.scaladsl.model.StatusCodes.{Accepted, BadRequest, InternalServerError, NoContent, OK, ServerError}
 import akka.http.scaladsl.model.Uri.Path
 import akka.http.scaladsl.model.headers.{Authorization, BasicHttpCredentials}
 import akka.http.scaladsl.model._
@@ -146,9 +146,14 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
                 case Failure(t) =>
                   logging.error(this, s"[POST] storing trigger activation $triggerActivationId failed: ${t.getMessage}")
               }
+            complete(Accepted, triggerActivationId.toJsObject)
+          } else {
+            logging
+              .info(
+                this,
+                s"[POST] trigger without an active rule was activated; no trigger activation record created for $triggerActivationId")
+            complete(NoContent)
           }
-
-          complete(Accepted, triggerActivationId.toJsObject)
       })
     }
   }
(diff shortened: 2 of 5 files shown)


core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -300,13 +305,21 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
   /**
-   * Iterates through each active rule and invoke each mapped action.
+   * Iterates through each rule and invoking each active rule's mapped action.
    */
   private def activateRules(user: Identity,
                             args: JsObject,
                             rulesToActivate: Map[FullyQualifiedEntityName, ReducedRule])(
     implicit transid: TransactionId): Iterable[Future[JsObject]] = {
     rulesToActivate.map {
+      case (ruleName, rule) if (rule.status != Status.ACTIVE) =>
```
rabbah: could be `rule.status == Status.INACTIVE` no?

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -300,13 +305,21 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
+        Future {
+          ruleResult(
+            ActivationResponse.ApplicationError,
+            ruleName,
+            rule.action,
+            errorMsg = Some(s"Rule ${ruleName.asString} is inactive; action ${rule.action.asString} was not activated"))
```
rabbah: should move this error message to `whisk.http.Messages`.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
mdeuser: possibly.  since the prior filter used `status == Status.ACTIVE`, i was thinking all rules without an _active_ status would be handled by this case.  that said, i see that rules only have two statuses currently defined - ACTIVE and INACTIVE.  so either will work, with the former automatically handling any other non-active state - either for good or for bad.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -129,7 +129,7 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
 
           if (activeRules.nonEmpty) {
             val args: JsObject = trigger.parameters.merge(payload).getOrElse(JsObject())
-            val actionLogList: Iterable[Future[JsObject]] = activateRules(user, args, activeRules)
+            val actionLogList: Iterable[Future[JsObject]] = activateRules(user, args, trigger.rules.get)
```
rabbah: I can't be absolutely certain here that trigger.rules.get is safe.
@markusthoemmes in your refactoring, I think this becomes avoidable. Should I merge that first?

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -300,13 +305,21 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
-   * Iterates through each active rule and invoke each mapped action.
+   * Iterates through each rule and invoking each active rule's mapped action.
    */
   private def activateRules(user: Identity,
                             args: JsObject,
                             rulesToActivate: Map[FullyQualifiedEntityName, ReducedRule])(
     implicit transid: TransactionId): Iterable[Future[JsObject]] = {
     rulesToActivate.map {
+      case (ruleName, rule) if (rule.status != Status.ACTIVE) =>
+        Future {
```
markusthoemmes: Please use `Future.successful`. Using `Future.apply` (as done here) results in a forced async-dispatch which is not needed here.

rabbah:
@mdeuser can you take a look at #3256 I think we should get that one in first as it will benefit the enhancement in this PR.

mdeuser:
rebased on top of #3256

mdeuser:
Updated `logs` field in the trigger activation record will look like the following for a trigger with one active and one inactive rule:
```
    "logs": [
        "{\"statusCode\":0,\"success\":true,\"activationId\":\"a0e1a50ba6074eeca1a50ba6072eecdc\",\"rule\":\"guest/activerule1\",\"action\":\"guest/web-echo-env\"}",
        "{\"statusCode\":1,\"success\":false,\"rule\":\"guest/inactiverule1\",\"error\":\"Rule guest/inactiverule1 is inactive; action guest/web-echo-env was not activated\",\"action\":\"guest/web-echo-env\"}"
    ],
```

tests/src/test/scala/system/basic/WskBasicTests.scala
```diff
@@ -748,6 +748,67 @@ class WskBasicTests extends TestHelpers with WskTestHelpers {
...
+            "action" -> JsString(ns + "/" + actionName1)),
+          JsObject(
+            "statusCode" -> JsNumber(1),
+            "success" -> JsBoolean(false),
+            "error" -> JsString(s"Rule ${ns}/$ruleName2 is inactive; action ${ns}/$actionName2 was not activated"),
+            "rule" -> JsString(ns + "/" + ruleName2),
+            "action" -> JsString(ns + "/" + actionName2)))
+        logs shouldBe expectedLogs
+      }
+  }
```
markusthoemmes: Test for the 204 case?

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
markusthoemmes: Just go with a `getOrElse(Map.empty)`. `activateRules` just won

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -129,7 +129,7 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
 
           if (activeRules.nonEmpty) {
```
markusthoemmes: I'll suggest to go for the following:

```scala
val rules = trigger.rules.getOrElse(Map.empty) // use this in belows call for activateRules

if(rules.exists(_._2 == Status.ACTIVE)) {
   that code
} else {
  complete(NoContent)
}
```

common/scala/src/main/scala/whisk/http/ErrorResponse.scala
```diff
@@ -111,6 +111,11 @@ object Messages {
   val notAllowedOnBinding = "Operation not permitted on package binding."
   def packageNameIsReserved(name: String) = s"Package name '$name' is reserved."
 
+  /** Error messages for triggers */
+  def triggerWithInactiveRule(rule: String, action: String) = {
+    s"Rule ${rule} is inactive; action ${action} was not activated."
```
rabbah: i suggest small change here:
```scala
 +    s"Rule '$rule' is inactive; action '$action' was not activated." 
```

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -146,9 +147,14 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
                 case Failure(t) =>
                   logging.error(this, s"[POST] storing trigger activation $triggerActivationId failed: ${t.getMessage}")
               }
+            complete(Accepted, triggerActivationId.toJsObject)
+          } else {
+            logging
+              .info(
```
rabbah: set to debug level.

core/controller/src/main/scala/whisk/core/controller/Triggers.scala
```diff
@@ -146,9 +147,14 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
+                this,
+                s"[POST] trigger without an active rule was activated; no trigger activation record created for $triggerActivationId")
+            complete(NoContent)
```
rabbah: we could add a message here so that the HTTP resonse status code is NoContent and the message is informative: Trigger has no associated rules that are active.

rabbah:
This will require a CLI change as well.

PR #3265, opened 2018-02-09
Change recently added metric key name. (very simple change)

upgle:
This PR is related to #2968(merged 2days ago) and thank you for your commit @mhenke1 

I think it is more common to use dotted notation format(.) for recording metric.

Most time series db like Graphite, OpenTSDB and InfluxDB normally use a dot-delimited notation. (e.g. `sys.cpu.user`, `invoker_docker.rm_start`)

Even in the openwhisk code, metric keys for recording docker cmd use dot delimiter.
```scala
def INVOKER_RUNC_CMD(cmd: String) = LogMarkerToken(invoker, s"runc.$cmd", start)
```

If you are using kamon-statsd with graphite, a query is a better way. 
For example, a query like `invoker_containerStart.prewarmed.*` will fill the variable with all possible values that exist in the wildcard position

### Before
```
invoker_container_start_prewarmed_guest_hello_count
```

> if action name is `foo_bar` and namespace is `baz_qux` then this marker will look ambiguous. `invoker_container_start_prewarmed_baz_qux_foo_bar_count`

### After
```
invoker_containerStart.prewarmed.guest.hello_count
```

> but in this case, dotted format is not ambiguous and can be seperated by dot(.) delimiter in Grafana, `invoker_containerStart.prewarmed.baz_qux.foo_bar_count`

> with grafana, It is possible to query using wildcard(*) easily `invoker_containerStart.prewarmed.*`

image

And I also changed naming convention, 
because most of the metric key names already used in the code is using camelcase like `activationRun`, `invokerOffline` and `blockingActivation`,

### DIFF
--- a/common/scala/src/main/scala/whisk/common/Logging.scala
+++ b/common/scala/src/main/scala/whisk/common/Logging.scala
@@ -243,7 +243,7 @@ object LoggingMarkers {
   val LOADBALANCER_INVOKER_OFFLINE = LogMarkerToken(loadbalancer, "invokerOffline", count)
   val LOADBALANCER_INVOKER_UNHEALTHY = LogMarkerToken(loadbalancer, "invokerUnhealthy", count)
   def LOADBALANCER_ACTIVATION_START(namespaceId: String) =
-    LogMarkerToken(loadbalancer, s"activations_$namespaceId", count)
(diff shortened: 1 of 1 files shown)


markusthoemmes:
@mhenke1 would you please review?

upgle:
I will close this PR because these metrics removed from #3284 PR.

PR #3266, opened 2018-02-09
Use always all available controllers.

cbickel:
If several controllers are specified, use always all of them.
This PR deprecates the availability of an hot standby controller.

PG3#1839 🔵 .


markusthoemmes:
Doesn't this straight-down remove the possibility of a hot-standby controller? Please write a mail to the dev-list first.

PR #3272, opened 2018-02-11
improve loadbalancer's schedule algorithm

ningyougang:
When all invokers's load is saturated, the lb will chose the head invoker
to execute the remaining action only, so it is necessary to make sure all
invokers have equal opportunity to exectue actions even if all invoker's
load is saturated.

### DIFF
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/ContainerPoolBalancer.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/ContainerPoolBalancer.scala
@@ -18,6 +18,7 @@
 package whisk.core.loadBalancer
 
 import java.nio.charset.StandardCharsets
+import java.util.concurrent.ThreadLocalRandom
 
 import akka.actor.{ActorSystem, Props}
 import akka.cluster.Cluster
@@ -343,7 +344,11 @@ object ContainerPoolBalancer extends LoadBalancerProvider {
         .find(_._3 < invokerBusyThreshold)
         .orElse(invokerProgression.find(_._3 < invokerBusyThreshold * 2))
         .orElse(invokerProgression.find(_._3 < invokerBusyThreshold * 3))
-        .orElse(invokerProgression.headOption)
+        .orElse(
+          if (invokerProgression.isEmpty)
+            None
+          else
+            Some(invokerProgression(ThreadLocalRandom.current().nextInt(invokerProgression.size))))
         .map(_._1)
     } else None
   }
--- a/tests/src/test/scala/whisk/core/loadBalancer/test/ContainerPoolBalancerObjectTests.scala
+++ b/tests/src/test/scala/whisk/core/loadBalancer/test/ContainerPoolBalancerObjectTests.scala
@@ -143,11 +143,14 @@ class ContainerPoolBalancerObjectTests extends FlatSpec with Matchers {
     ContainerPoolBalancer.schedule(invs, 16, hash) shouldBe Some(InstanceId(0))
   }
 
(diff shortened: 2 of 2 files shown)


rabbah:
This is generally a good idea. There an alternate loadbalancer  https://github.com/apache/incubator-openwhisk/pull/3240 that you might be interested in reviewing. Note that it also applies a random assignment when overloaded.

core/controller/src/main/scala/whisk/core/loadBalancer/ContainerPoolBalancer.scala
```diff
@@ -345,7 +345,8 @@ object ContainerPoolBalancer extends LoadBalancerProvider {
         .find(_._3 < invokerBusyThreshold)
         .orElse(invokerProgression.find(_._3 < invokerBusyThreshold * 2))
         .orElse(invokerProgression.find(_._3 < invokerBusyThreshold * 3))
-        .orElse(invokerProgression.headOption)
+        .orElse(
+          if (invokerProgression.size == 0) None else invokerProgression.lift(Random.nextInt(invokerProgression.size)))
```
markusthoemmes: The extra check for `size == 0` is redundant I think, since `lift` will automatically bound-check anyway.

It is **crucial** to use `ThreadLocalRandom` vs. `Random` because getting randomness is synchronized and thus will result in thread-contention.

ningyougang: `invokerProgression` here,  doesn't have `nonEmpty` method.

ningyougang: @markusthoemmes , can you look at this again?

ningyougang:
@markusthoemmes ,can you loot at this again?

markusthoemmes:
@ningyougang would you mind adding a test? You can have a look at the PR @rabbah mentioned. I tested the randomness by checking that every invoker is scheduled to in 100 schedules and by asserting that the unhealthys are not in that sequence even once.

ningyougang:
@markusthoemmes , it seems `random` is not a good idea, in long time, it is `fairly`, but in a short time, `random` may not `failly`, may lead to some invoker nodes's action is more, but some others is `less`.

How do i found  the problem?
Becuase i am testing the `zerodowntime deployment`, during `disable one invoker` short time, above  problem exist.

So i suggest to improve like below
```
      invokerProgression
        .find(_._3 < invokerBusyThreshold)
        .orElse(invokerProgression.find(_._3 < invokerBusyThreshold * 2))
        .orElse(invokerProgression.find(_._3 < invokerBusyThreshold * 3))
        .orElse(
          if (invokerProgression.isEmpty)
            None
          else
            Some(invokerProgression.minBy(_._3)))
        .map(_._1)
```
How do you think?

markusthoemmes:
As @rabbah would put it: Randomness generally helps against pathologies. Your `minBy` makes sense in homogenous load but once you start getting activations with different lengths, runtimes etc. using `minBy` might create a pathology.

That said, it's the overflow case anyway so we cannot optimize that much.

ningyougang:
ok, i think so, i will add the test case

ningyougang:
@markusthoemmes ,i have added the test case, please review again.

markusthoemmes:
PG2 2836 running.

markusthoemmes:
@ningyougang can you please rebase?

ningyougang:
@markusthoemmes , i have rebased the codes.

ningyougang:
@markusthoemmes ,can be merged?

markusthoemmes:
I will get to it, don't you worry.

PR #3273, opened 2018-02-11
Wait for logs based on intervals not based on total processing time.

markusthoemmes:
Writing a large chunk of logs can take quite some time to process. The standard timeout for this process is 2 seconds today. It is bounded, because an action developer might break the action proxy to make sentinels not appear at all which would cause us to infinitely wait on sentinels.

As we process logs after an activation has run though, we can safely rely on the time **between** two logs not exceeding a certain threshold. That way, the complete processing is not bounded by some arbitrary timeout (which can even be too short for large volumes) and is still tight enough to exit early if sentinels really are missing.

Furthermore, an error line is inserted if this timeout hits to inform the user that something might've gone wrong.


markusthoemmes:
PG3 1832 🔵

PR #3276, opened 2018-02-13
Use proper API host and reduce logging noise.

markusthoemmes:
The API host was hardcoded to `WhiskProperties.getApiHost` which is not correct for all tests. It should use the one defined in `WskProps`.

It's unnecessary to open a connection-pool per request. A single request should be sufficient.

Co-authored-by: cbickel <github@cbickel.de>
Co-authored-by: jeremiaswerner <jeremias.werner@gmail.com>


markusthoemmes:
PG2 2801 🔵

PR #3278, opened 2018-02-13
Fix NullPointerException in ActivationFinisher.

markusthoemmes:
The ActivationFinisher's code is racy. If shutdown and postStop are called concurrently a NullPointerException is possible.

Stopping the actor itself should call `postStop` and thus the logic there seems redundant.


markusthoemmes:
This proves that `postStop` will also be called from an internal `context.stop` (and that the context is always there then, the NPE was actually in `shutdown`).

```scala
object Testing {
  import scala.concurrent.Await

  class TestActor extends Actor {
    override def receive: Receive = {
      case "stop" => context.stop(self)
    }

    override def postStop(): Unit = {
      println(context)
      println("Yep, I was called :)")
    }
  }

    implicit val as = ActorSystem()

    val testActor1 = as.actorOf(Props[TestActor])
    val testActor2 = as.actorOf(Props[TestActor])

    as.stop(testActor1)
    testActor2 ! "stop"

    Await.ready(as.terminate(), 10.seconds)
  }
}
```

PR #3280, opened 2018-02-13
Replace asString and asBool with generic method.

markusthoemmes:
Mainly aims to reduce boilerplate in code which tries to extract non-primitive data from parameters/annotations. Any supported json-format (for example case classes) can be loaded safely from parameters.


markusthoemmes:
PG4 1359 🔵

PR #3282, opened 2018-02-14
Disable createKeystore if test execution is disabled

chetanmeh:
Currently the `createKeystore` in tests is linked with `compileTestScala`. This poses problem in cases where the build is triggered with test execution disabled.

```
* What went wrong:
> java.io.FileNotFoundException: /path/to/openwhisk/whisk.properties (No such file or directory)
```
This would be the case for example where core is build just to install the test artifacts to Maven repository and where whisk.properties is yet not generated. (see apache/incubator-openwhisk-runtime-docker#17)

This PR disables this task if `test` is disabled

### DIFF
--- a/tests/build.gradle
+++ b/tests/build.gradle
@@ -85,8 +85,14 @@ task createKeystore(dependsOn: deleteKeystore) {
     doLast {
         Properties props = new Properties()
         props.load(new FileInputStream(file('../whisk.properties')))
+        keystorePath.parentFile.mkdirs()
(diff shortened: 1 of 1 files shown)


markusthoemmes:
Good thought! Is there a possibility to hook stuff up differently? I think the issue here is that `createKeystore` is run after compiling. Is it maybe possible to hook it up to "preRun" or something? That'd make this more clear.

chetanmeh:
> Is it maybe possible to hook it up to "preRun" or something? 

Good point. Updated the PR to use a different approach which binds the createKeystore with all test tasks. So now it would only be triggered if any test task is invoked

csantanapr:
@markusthoemmes could you review this one and merge if ready when you have a chance

PR #3283, opened 2018-02-14
Add implicit ClassTag for entity type

chetanmeh:
For implementing custom `ArtifactStore` and computing views on client side for databases other than CouchDB the `ArtifactStore` would need access to the entity type. 

The logic in custom ArtifactStoreProvider would be similar to one present [here][1] just that this PR uses `ClassTag` instead of `Manifest` as [Manifest are now deprecated post Scala 2.10][2]

To enable that `ArtifactStoreProvider` should have an implicit parameter for `ClassTag`

[1]: https://github.com/apache/incubator-openwhisk/commit/9be13e2c6d581f42b60351f05a11e4aa71ca31d9#diff-9399daa47aa3257bb6bcb49425f4389dR110
[2]: https://docs.scala-lang.org/overviews/reflection/typetags-manifests.html#typetags-and-manifests

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/ArtifactStoreProvider.scala
+++ b/common/scala/src/main/scala/whisk/core/database/ArtifactStoreProvider.scala
@@ -25,13 +25,15 @@ import whisk.core.WhiskConfig
 import whisk.spi.Spi
 import whisk.core.entity.DocumentReader
 
+import scala.reflect.ClassTag
+
 /**
  * An Spi for providing ArtifactStore implementations
  */
 trait ArtifactStoreProvider extends Spi {
-  def makeStore[D <: DocumentSerializer](config: WhiskConfig,
-                                         name: WhiskConfig => String,
-                                         useBatching: Boolean = false)(
+  def makeStore[D <: DocumentSerializer: ClassTag](config: WhiskConfig,
+                                                   name: WhiskConfig => String,
+                                                   useBatching: Boolean = false)(
     implicit jsonFormat: RootJsonFormat[D],
     docReader: DocumentReader,
     actorSystem: ActorSystem,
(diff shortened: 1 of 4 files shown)


markusthoemmes:
> In Scala 2.10, scala.reflect.ClassManifest are deprecated, and it is planned to deprecate scala.reflect.Manifest in favor of TypeTags and ClassTags in an upcoming point release. Thus, it is advisable to migrate any Manifest-based APIs to use Tags.

Manifest is not deprecated but planned to be deprecated at some point. Can you provide an implementation example of where a `ClassTag` would be needed?

chetanmeh:
> Can you provide an implementation example of where a ClassTag would be needed?

Have a look at [MongoDbStoreProvider][1] (currently work in progress) which needs this info to [compute fields](chetanmeh/openwhisk-mongo#8)

[1]: https://github.com/chetanmeh/openwhisk-mongo/blob/master/common/mongo/src/main/scala/whisk/core/database/mongo/MongoDbStoreProvider.scala#L55

markusthoemmes:
Okay makes sense.

This is a bit smelly though. It suggests we have a generalized API for something which shouldn't be generalized? Since we already have the `WhiskAuthStore`, `WhiskEntityStore`, `WhiskActivationStore` objects to construct entity stores, would it make sense to rather widen the constructor of `ArtifactStore`? What is the type of `WhiskHandler` for instance?

chetanmeh:
> It suggests we have a generalized API for something which shouldn't be generalized?

Its kind of true as with CouchDB most of query logic is implemented as code in CouchDB side which can be treated as implementation detail for CouchDbRestStore where the (designDoc, viewName) parameter maps to the server side logic.

For non CouchDB stores this would need to be implemented as part of store implementation. Another usecase is type of index being created which would differ for various stores and would differ based on persistent store.

Hence using entity/document type as selector for such logic

chetanmeh:
@markusthoemmes  @rabbah Does the approach looks fine here or some aspect needs to be improved?

rabbah:
LGTM. I think we’ll need to rethink the artifact store as we’ve made several singnifcant changes since initial conception and can simplify some of the generic typing which gets in the way; this led to a whole of code duplication in WhiskActionMetadata for example.

chetanmeh:
Can this be merged

PR #3284, opened 2018-02-14
Remove finely granulated metrics.

cbickel:
We introduced some very finely granulated metrics.
On using a commonly used monitoring backend, it can be easily overloaded with a huge amount of users and different actions on the OpenWhisk system.


cbickel:
PG5#97 looks good.

PR #3286, opened 2018-02-14
Treat action code as attachments for created/updated actions

dubee:
Supersedes https://github.com/apache/incubator-openwhisk/pull/2847.

### DIFF
--- a/ansible/files/runtimes.json
+++ b/ansible/files/runtimes.json
@@ -6,7 +6,11 @@
                 "image": {
                     "name": "nodejsaction"
                 },
-                "deprecated": true
+                "deprecated": true,
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
+                }
             },
             {
                 "kind": "nodejs:6",
@@ -14,7 +18,11 @@
                 "image": {
                     "name": "nodejs6action"
                 },
-                "deprecated": false
+                "deprecated": false,
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
+                }
             },
             {
                 "kind": "nodejs:8",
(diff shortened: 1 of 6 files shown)


markusthoemmes:
@dubeejw can we close this PR? You can setup travis on your own repository if you want to have full travis coverage.

dubee:
@rabbah, @markusthoemmes, a cache concurrency test fails on this one. I have figured out the failure is due the need of two DB writes for saving attachments. Where the first write is for the DB document and the second write is for the DB document attachment. 

When only caching Java actions as done in https://github.com/apache/incubator-openwhisk/pull/2855 and implementing the same concurrency test for Java actions, the test will fail as well.
```

Failure output shown here. The test fails because `update` does two writes as stated above, so a concurrent `get` request can fail as `update` does not finish as quickly for attachments.
```
whisk.core.database.test.CacheConcurrencyTests > the cache should support concurrent CRUD without bogus residual cache entries, iter 1 FAILED
    org.scalatest.exceptions.TestFailedException: update+get: failed for testy4 404 was not equal to 200
```

dubee:
I suppose another option would be to change the test to accept a 200 or 404 from the `get` requests.

dubee:
Closing https://github.com/apache/incubator-openwhisk/pull/2847 in favor of this PR as it was easier to rebase with the merged caching changes.

dubee:
@chetanmeh, do you think these changes will be useful in the future?

chetanmeh:
@dubee Yes this would be useful as document size limitation is impacting nodejs kind also for us. So would like to convert all types to attachments. With support for Inlined attachments #3709 switching to attachment should not have negative impact on small functions.

chetanmeh:
@dubee  Do you plan to work on this otherwise I can look into it. Given recent changes in master the `ArtifactStore` does not make use of attachmentName attribute from manifest. So was thinking to do away with manifest entry for name attribute.

For attachment type I am not sure - For e.g. if its a simple code snippet then type can be `text/plain` otherwise if its a zip (say js code packaged as zip) then it would be `application/octet-stream`.

Further with inlining support now we can **always** store the code for all kinds as attachments by design and let `ArtifactStore` inline the binary if its small. So no need to different between attachment for various kinds

chetanmeh:
@dubee Ping

chetanmeh:
PR #3945 supersedes this one

PR #3288, opened 2018-02-14
Run container pool logging tests

dubeejw:
The following test suites were not being run: SplunkLogStoreTests, LogDriverLogStoreTests, DockerToActivationLogStoreTests, DockerToActivationFileLogStoreTests.

PR #3289, opened 2018-02-14
add @RunWith annotation for splunk logstore tests

tysonnorris:


tysonnorris:
already fixed in #3288

PR #3290, opened 2018-02-14
Check all testfiles to actually contain the `RunWith` annotation.

markusthoemmes:
We had a couple of rogue tests in the repository which didn't have the annotation `RunWith`, so they haven't been run in CI/CD at all.

Note: Misses one file on purpose to test the fail.


markusthoemmes:
PG2 2823 🔵

csantanapr:
Oh my 😱  Good find @markusthoemmes @cbickel  👍

PR #3292, opened 2018-02-15
Remove expected duration from Swift test

dubeejw:
This test is prone to failure because of the expected test duration exceeds 45 seconds. The time limit instead should be based on the action's `timeout` property, which by default is 60 seconds. Also, don't count the time it takes to fetch the action's activation.

Test failure example:
`Test duration exceeds expectation (ms) 48629 was not less than or equal to 45000`


dubeejw:
PG2 2820 ⏳

PR #3293, opened 2018-02-16
Script to extract view functions as js files

chetanmeh:
This PR adds a module to host some utility scripts.

To start with this PR adds a utility script to extract the view functions which are embedded in design docs json files in ansible/files folder to formatted js files. This helps in understanding the design docs better

    ./gradlew -p tools/dev couchdbViews

The views would created under _tools/dev/build/views_ directory with name set to \<design doc name>-\<view name>.js


rabbah:
Would be nice to have as a `wskdev` target as well.

chetanmeh:
> Would be nice to have as a wskdev target as well.

@rabbah Would just having wskdev target to invoke this gradle command would be fine? Or should it also move the rendered js files to some other place

rabbah:
> Would just having wskdev target to invoke this gradle command would be fine?

that is all i was suggesting.

chetanmeh:
Opened #3317 for that

PR #3296, opened 2018-02-16
Bump swift test-timeout to 2 minutes.

markusthoemmes:
- This test is prone to errors because compiling swift apparently is subject to a lot variance (seen up to 1 minute).
- Refactor of some test helpers.


markusthoemmes:
We definitly should, yes. @rabbah

PR #3298, opened 2018-02-17
Attempt to resolve action name from conductor to a fully qualified name for added convenience.

rabbah:
I updated the document for conductor actions to specify that the action name for a continuation should be a fully qualified name. I however also added a conversion from string to a FQN with sensible defaults. This makes for slightly more convenient return results and avoids potentially confusing error messages when referencing a packages action but neglecting to specify the namespace. We have in the past resisted doing this and required the clients to resolve the names fully. 

@tardieu can you review and comment?

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/FullyQualifiedEntityName.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/FullyQualifiedEntityName.scala
@@ -96,4 +96,36 @@ protected[core] object FullyQualifiedEntityName extends DefaultJsonProtocol {
         case Failure(t)                           => deserializationError("fully qualified name malformed")
       }
   }
+
+  /**
+   * Converts the name to a fully qualified name.
+   * There are 3 cases:
+   * - name is not a valid EntityPath => error
+   * - name is a valid single segment with a leading slash => error
+   * - name is a valid single segment without a leading slash => map it to user namespace, default package
+   * - name is a valid multi segment with a leading slash => treat it as fully qualified name (max segments allowed: 3)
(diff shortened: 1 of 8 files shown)


csantanapr:
@rabbah check on Travis and do PG

rabbah:
pg2/2829 already 🏃

rabbah:
pg2/2831 🔵

common/scala/src/main/scala/whisk/core/entity/FullyQualifiedEntityName.scala
```diff
@@ -96,4 +96,36 @@ protected[core] object FullyQualifiedEntityName extends DefaultJsonProtocol {
...
+   * The action should use a fully qualified result to avoid the ambiguity.
+   *
+   * @param name name of the action to fully qualify
+   * @param namespace the user namespace for the simple resolution
+   * @return Some(FullyQualifiedName) if the name is valid otherwise None
+   */
+  protected[core] def resolveName(name: JsValue, namespace: EntityName): Option[FullyQualifiedEntityName] = {
+    name match {
+      case v @ JsString(s) =>
+        Try(v.convertTo[EntityPath]).toOption
```
markusthoemmes: Is the nested matching needed? Why not directly go with `Try(name.convertTo[EntityPath])`?

rabbah: Once the value is converted the leading slash is gone/stripped.

markusthoemmes: gnah. ugly. got it, thanks!

csantanapr:
@tardieu bump

tardieu:
:thumbsup:

PR #3303, opened 2018-02-19
Actually retry package remove on conflict.

markusthoemmes:
The AssetHelper is supposed to retry a package deletion if it returns a conflic. This can happen, if the package contained other entities (like an action), which got deleted first but the package's view does not yet reflect that deletion. The DELETE call will then complain about "entities still in package". A retry resolves that issue.

- Some cleanup of the helpers file.
- Refactoring of the annotations helper.

### DIFF
--- a/tests/src/test/scala/common/WskTestHelpers.scala
+++ b/tests/src/test/scala/common/WskTestHelpers.scala
@@ -28,6 +28,7 @@ import scala.concurrent.duration.Duration
 import scala.concurrent.duration.DurationInt
 
 import spray.json._
+import spray.json.DefaultJsonProtocol._
 
 import TestUtils.RunResult
 import TestUtils.CONFLICT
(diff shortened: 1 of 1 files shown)


jasonpet:
pg3: 1888

csantanapr:
@cbickel PG3 1888 had problems 🔴 
```
Pruning sources from previous analysis, due to incompatible CompileSetup.
/home/cusina/workspace/Playground3/blue/bluemix/tests/src/test/scala/bluemix/billing/BluemixWskBillingTests.scala:343: constructor cannot be instantiated to expected type;
 found   : (T1, T2)
 required: Unit
        val timeWindows = activations.map { case (result, namespace) => calculateTimeWindowFromInvocationResponse(result.asInstanceOf[String]) }.distinct
                                                 ^
/home/cusina/workspace/Playground3/blue/bluemix/tests/src/test/scala/bluemix/billing/BluemixWskBillingTests.scala:353: constructor cannot be instantiated to expected type;
 found   : (T1, T2)
 required: Unit
            case (activation, namespace) =>
                 ^
```

cbickel:
I missed that. #3308 will fix this.

PR #3309, opened 2018-02-20
Rework rule tests to not rely on views anymore.

markusthoemmes:
Relying on views is brittle and will fail if the database is under high load. We don't need to rely on those anymore.


markusthoemmes:
PG4 1386 🔵

tests/src/test/scala/system/basic/WskRuleTests.scala
```diff
@@ -164,17 +162,12 @@ abstract class WskRuleTests extends TestHelpers with WskTestHelpers {
 
     withActivation(wsk.activation, run) { triggerActivation =>
       triggerActivation.cause shouldBe None
-      triggerActivation.logs.get.size shouldBe (1)
```
cbickel: Can we add that check again?

PR #3310, opened 2018-02-20
Refactor activation finisher without actor.

markusthoemmes:
We frequently create/tear-down actors on this hot-path. Maybe we can get around that.


markusthoemmes:
PG3 1892 🔵

PR #3311, opened 2018-02-20
Bump default retry timeout to 50 milliseconds.

markusthoemmes:
The default retry timeout is fairly low today (1 millisecond). Rather than adjusting each value locally, we should use a more decent value to start with globally.


markusthoemmes:
PG4 1387 🔵

Issue #3312, opened 2018-02-20
Configure Gradle build scan plugin


chetanmeh:
Gradle provide a [hosted service][1] where build results can be published. This requires [build scan][2] plugin to be configured. 

- https://docs.gradle.com/build-scan-plugin
- https://gradle.com/build-scans
- Sample report from JUnit Project - https://scans.gradle.com/s/pgjgssca2kkli/tests
- Introduction Blog https://blog.gradle.org/introducing-build-scans

This would be specially useful when CI fails for some random test issues and also get an insight on which test is taking more time 

[1]: https://scans.gradle.com/get-started
[2]: https://plugins.gradle.org/plugin/com.gradle.build-scan/1.11

chetanmeh:
As per of setup one needs to [accept terms of service][1] via 

```
buildScan {
    termsOfServiceUrl = 'https://gradle.com/legal/terms-of-service'
    termsOfServiceAgree = 'yes'
    publishAlwaysIf(System.getenv('CI') != null)
}
```
So has some legal implication. Checking the usage in Apache project it appears that [Groovy project has this configured][2] so it should be ok to use

[1]: https://docs.gradle.com/build-scan-plugin/#accept_the_scans_gradle_com_terms_of_service
[2]: https://github.com/apache/groovy/blob/270709706102e4efef5a59da0f185f48436a8402/build.gradle#L54

chetanmeh:
Here is an output based on build from the fork https://gradle.com/s/cw3dfrmz237ig

markusthoemmes:
Sweeeet, didn't know this existed. Can we hook this up with travis as indicated by 

> This would be specially useful when CI fails for some random test issues and also get an insight on which test is taking more time

chetanmeh:
To find the scan report url look for log entry in travis like below

```
Publishing build scan...
https://gradle.com/s/reldo4qqlg3ka
```

chetanmeh:
Closed via #3314

PR #3314, opened 2018-02-20
Configure Gradle build scan plugin

chetanmeh:
PR for #3312. With this the scan result would only be published on CI setups

However with this change now following warning is seen

```
WARNING: The build scan plugin was applied after other plugins.
The captured data is more comprehensive when the build scan plugin is applied first.

Please see https://gradle.com/scans/help/plugin-late-apply for how to resolve this problem.
```

From changes done in this PR its ensured that this plugin comes first but still the warning is seen. For now it appears a [bug in gradle](gradle/gradle#3341) which was fixed in 4.3 (we use 4.1). I tried with current 4.5 and warning disappears.

@markusthoemmes Not sure sure if changes here affect down stream projects (like the issue seen with coverage pr)

Apart from the noisy warning things work fine

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,6 +7,16 @@ buildscript {
     }
 }
 
+plugins {
+    id "com.gradle.build-scan" version "1.12.1"
+}
+
+buildScan {
+    termsOfServiceUrl = 'https://gradle.com/terms-of-service'
+    termsOfServiceAgree = 'yes'
+    publishAlwaysIf(System.getenv('CI') != null)
+}
+
 subprojects {
     apply plugin: 'scalafmt'
     scalafmt.configFilePath = gradle.scalafmt.config
--- a/tools/build/README.md
+++ b/tools/build/README.md
@@ -68,6 +68,19 @@ The logs are saved to `./B-build.log` and can be reprocessed using `citool` with
 citool -i -b B cat -s -g "tid_124" whisk/logs N
 ```
 
+## Gradle Build Scan Integration
+
+OpenWhisk builds on CI setups have [Gradle Build Scan][1] integrated. Each build on travis pushes scan reports to
+[Gradle Scan Community Hosted Server][2]. To see the scan report you need to check the travis build logs for lines like
+below 
+
+```
+Publishing build scan...
+https://gradle.com/s/reldo4qqlg3ka
+```
+
+The url above is the scan report url and is unique per build
+
 ## Troubleshooting
 
 If you encounter an error `ImportError: No module named pkg_resources` while running `redo`, try the workaround below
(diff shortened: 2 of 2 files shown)


build.gradle
```diff
@@ -7,6 +7,16 @@ buildscript {
...
 }
 
+plugins {
+    id "com.gradle.build-scan" version "1.12.1"
+}
+
+buildScan {
+    termsOfServiceUrl = 'https://gradle.com/terms-of-service'
+    termsOfServiceAgree = 'yes'
+    publishAlwaysIf(System.getenv('CI') != null)
```
markusthoemmes: Do we ever set this variable anywhere?

markusthoemmes:
I don't think the dependent repos should break because this affects a file which isn't loaded by those.

build.gradle
chetanmeh: This is set in Travis env [by default][1]

[1]: https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

markusthoemmes:
@csantanapr could you check possible implications (legally) because we're an Apache project?

build.gradle
csantanapr: Oh I didn’t know this

csantanapr:
Apache wise this is OK, and it’s not a hard dependency to the main functionality of the core componets

csantanapr:
@chetanmeh can you take a look into Travis failure, it might be that we are going over the budget 50 minute limit.
>The job exceeded the maximum time limit for jobs, and has been terminated.

I think @houshengbo told me that we are able to have use a Travis matrix and be able to run two jobs in parallel maybe we can run the scan in one and the tests in another one.

@houshengbo do you have an example of such Travis setup?

chetanmeh:
@csantanapr  Same build in [my fork][1] took took ~ 42 mins compared to 49 mins on this build. May be a temporary load issue. Have triggered a re build

[1]: https://travis-ci.org/chetanmeh/incubator-openwhisk/builds/343949594

chetanmeh:
Latest build took  40 min 22 sec. So last failure would be a temporary state

tools/build/README.md
```diff
@@ -76,3 +89,6 @@ or see [these instructions](https://pypi.python.org/pypi/setuptools/0.9.8#instal
 ```
 pip install --upgrade setuptools
 ```
+
+[1]: https://gradle.com/build-scans
+[2]: https://scans.gradle.com
```
markusthoemmes: We usually inline links with the `[text](link)` notation. Would you mind changing that?

chetanmeh: Done now

csantanapr:
PG 4 1390 ⏳

csantanapr:
Playground4/1390/ 🔵

csantanapr:
found the tests scan https://scans.gradle.com/s/ig5d2gnawa6ss/tests 👍

PR #3316, opened 2018-02-21
Add conditional statement to check metricsKamon flag.

upgle:
Recently added all MetricEmitter logic does not check metricsKamon config is enabled.

```scala
if (TransactionId.metricsKamon) {
        MetricEmitter.emitHistogramMetric(token, tid.deltaToStart)
        MetricEmitter.emitCounterMetric(token)
}
```
https://github.com/apache/incubator-openwhisk/blob/925500cf8d34bb75bebd077ea057af236eba0b01/common/scala/src/main/scala/whisk/http/BasicHttpService.scala#L127

### DIFF
--- a/common/scala/src/main/scala/whisk/common/Logging.scala
+++ b/common/scala/src/main/scala/whisk/common/Logging.scala
@@ -192,15 +192,19 @@ object MetricEmitter {
   val metrics = Kamon.metrics
 
   def emitCounterMetric(token: LogMarkerToken): Unit = {
-    metrics
-      .counter(token.toString)
-      .increment(1)
+    if (TransactionId.metricsKamon) {
+      metrics
+        .counter(token.toString)
+        .increment(1)
+    }
   }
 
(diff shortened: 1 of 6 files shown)


rabbah:
@markusthoemmes this looks right to me - but please verify and merge.

markusthoemmes:
@mhenke1 can you double-check please?

markusthoemmes:
@upgle Is there a possibility to move this check inside the `MetricEmitter`? More importantly: What's the behavior of this if you don't have Kamon setup? Any error logs?

upgle:
@markusthoemmes 
yes it is possible to move inside the MetricEmitter.

And Kamon always starts, regardless of `metrics_kamon: true` options (Kamon setup), it tries to send the meaningless metric without proper setting.
So it should be checked `TransactionId.metricsKamon` always.

Also, an error occurred in our custom kamon-backend metric system without setting like host, auth... (it is not statsd backend.)

markusthoemmes:
PG5 112 🔵

PR #3318, opened 2018-02-21
Remove restriction on updating a trigger with a new feed.

rabbah:
Feedback solicited.

Since a feed is already handled entirely on the client side today, should we just remove the restriction on the backend entirely?

Note the existing code is buggy - you can update an existing trigger with an attached feed, and de-associate it from the feed by specifying a new annotation. Further, if there's no feed attached, why prohibit an update that attaches a feed?

So either we should tighten the backend check, or remove it entirely, since clients like the CLI currently handle feed differently and route some API calls directly to the feed vs the controller. See https://github.com/apache/incubator-openwhisk-cli/issues/225 for example.

Closes https://github.com/apache/incubator-openwhisk/issues/3297


csantanapr:
+1 to remove concern from backend.
Document better behavior and let clients tools/frameworks/etc.. handle it

csantanapr:
@rabbah PG?

rabbah:
PG3/1901 🔵

PR #3323, opened 2018-02-21
Normalize a missing body for raw actions to JsString.empty.

rabbah:
Proposed fix for https://github.com/apache/incubator-openwhisk/issues/3321.

I partially confirmed the reported issue. The code and previous test confirms a missing body is an instance of JsObject, not JsArray. I'm not sure yet where the JsArray comes from; it may be PHP actions specific. A node test shows the type is not an array.

@akrabat FYI for node and python below.

screen shot 2018-02-21 at 10 27 59 am

screen shot 2018-02-21 at 10 34 55 am

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
@@ -129,11 +129,11 @@ private case class Context(propertyMap: WebApiDirectives,
     // if the body is a json object, merge with query parameters
     // otherwise, this is an opaque body that will be nested under
     // __ow_body in the parameters sent to the action as an argument
-    val bodyParams = body match {
+    val bodyParams: Map[String, JsValue] = body match {
       case Some(JsObject(fields)) if !boxQueryAndBody => fields
       case Some(v)                                    => Map(propertyMap.body -> v)
       case None if !boxQueryAndBody                   => Map.empty
-      case _                                          => Map(propertyMap.body -> JsObject())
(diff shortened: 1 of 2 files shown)


markusthoemmes:
Ohhhh ofcourse. It's an array in PHP because PHP has associative arrays 🤦‍♂️. Good one!

rabbah:
pg3/1912 🔵

akrabat:
> Ohhhh ofcourse. It's an array in PHP because PHP has associative arrays 🤦‍♂️. Good one!

To follow up on this so for completeness, PHP's JSON decoder can decode to associative arrays (dictionaries) or to objects (PHP's stdClass). However, PHP's associative array handling is so strong that decoding JSON to associated arrays is more common and so that's what I made PHP Runtime do. 

As this came up, I've doubled checked by looking at a few major PHP frameworks and they all decode JSON to associative arrays. I also asked a few PHP developers that I know about what they would expect JSON decoding to look like on the PHP side and they all confirmed that associative arrays is what they would expect rather than stdClass objects.

i.e. we're doing it right for PHP, even though it's not necessarily what other languages would do.

csantanapr:
Thanks Rob for the investigation 👍

markusthoemmes:
@akrabat sure, I didn't meant to say that this is the wrong thing to do. I just completely forgot that PHP has this thing so I was wondering how on earth it could become an empty array 😂 No offense at all.

akrabat:
Oh, I wasn't offended :) It wasn't until this came up that I thought about it properly, so wanted to double check I wasn't an outlier in the PHP world on this point!

PR #3326, opened 2018-02-22
Make SplunkLogStore tests more resilient.

markusthoemmes:
Timeouts of 1 second are a bit narrow and that narrowness is not necessary here. Bumped to 10 seconds.

Also refactored some bit for a more density.

### DIFF
--- a/tests/src/test/scala/whisk/core/containerpool/logging/SplunkLogStoreTests.scala
+++ b/tests/src/test/scala/whisk/core/containerpool/logging/SplunkLogStoreTests.scala
@@ -17,63 +17,31 @@
 
 package whisk.core.containerpool.logging
 
+import java.time.ZonedDateTime
+
 import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.http.javadsl.model.headers.Authorization
-import akka.http.scaladsl.model.ContentTypes
-import akka.http.scaladsl.model.FormData
-import akka.http.scaladsl.model.HttpEntity
-import akka.http.scaladsl.model.HttpRequest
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/whisk/core/containerpool/logging/SplunkLogStoreTests.scala
```diff
@@ -17,59 +17,40 @@
...
     extends TestKit(ActorSystem("SplunkLogStore"))
     with FlatSpecLike
     with Matchers
     with ScalaFutures
     with StreamLogging {
+
+  def await[T](awaitable: Future[T], timeout: FiniteDuration = 10.seconds) = Await.result(awaitable, timeout)
+
   val testConfig = SplunkLogStoreConfig(
     "splunk-host",
```
sven-lange-last: I suggest to specify a bogus IP address instead of a bogus host name. A bogus host name will require DNS name resolution which takes much longer than connecting the bogus IP address directly.

We could use 127.0.0.1 together with a port that will fail when connected - for example, port 7 (echo protocol).

tysonnorris: I expect there should not be any network interaction (dns or otherwise) with this test - is hostname resolution actually happening? The test should only create an HttpRequest and use an alternate mock flow to create a canned response.

tysonnorris:
I'm ok with chaning the timeout to 10s, but am surprised to see this be a problem since I expect no network activity in the test? If it is simply a problem of running on a slow machine that is ok, but let me know if you think there is network related logic leaking into the test?

PR #3327, opened 2018-02-22
Add retries to rule tests to prevent race condition

vvraskin:
In some deployments where cloudant is being used instead of couchdb, one could get Rule update conflict due a sluggish processing on the cloudant side, thus updating rules one after another could potentially lead to test failure. 
In my case the failed put request was 1.5s seconds apart, assuming that 5s buffer implied in the retry should be enough. 

Failed mainBluewhisk builds:
7887
7292 

@cbickel could you please review and merge?


vvraskin:
Forget it, its already there in the latest master

PR #3328, opened 2018-02-22
Fix oversight in resolving /_ to namespace.

rabbah:


csantanapr:
Pg?

rabbah:
Travis should be sufficient.

csantanapr:
Bingo bango 🖖

PR #3330, opened 2018-02-23
Update Log Store Fetch

dubeejw:
Pass HTTP request and user `Identity` to log store's `fetchLogs()`

### DIFF
--- a/common/scala/src/main/scala/whisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
@@ -18,6 +18,7 @@
 package whisk.core.containerpool.logging
 
 import java.nio.file.{Path, Paths}
+import java.time.Instant
 
 import akka.NotUsed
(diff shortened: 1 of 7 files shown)


dubeejw:
PG3 1906 ⏳

rabbah:
Rather than pass http headers should we instead allow the identity type to be extensible with additional iam details?

markusthoemmes:
I'd rather we pass in the entire `HttpRequest`. Query-string or body might be interesting to implementors as well?

@rabbah this is a lot about staging changes here and I think as an MVP it makes most sense to give `fetchLogs` access to the entire `HttpRequest`.

For anybody: The background of this is, that any backend Logstore might be subject to a different authentication than OpenWhisk itself. Rather than entangle everything and anything in OpenWhisk authentication means it might be beneficial to provide a "custom" API.

Thinking about it, just the headers might be fine as a first step.

dubeejw:
@markusthoemmes, added a second commit that send the HTTP request to the log store instead of just the HTTP headers.

dubeejw:
PG3 1908 🔵

PR #3333, opened 2018-02-23
Make parameters with defined values final 

rabbah:
Web actions can carry a final annotation which makes their parameters with predefined values immutable. This patch makes the same behavior applicable for a POST invoke as well.

The relevant change is the third commit. The first two, staged this way, introduce semantic preserving changes and a new test. Suggest https://github.com/apache/incubator-openwhisk/pull/3333/files?w=1

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Actions.scala
@@ -230,7 +230,7 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with 
                 // incoming parameters may not override final parameters (i.e., parameters with already defined values)
                 // on an action once its parameters are resolved across package and binding
                 val allowInvoke = payload
-                  .map(_.fields.keySet.intersect(actionWithMergedParams.immutableParameters).isEmpty)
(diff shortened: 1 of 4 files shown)


csantanapr:
PG?

rabbah:
pg1/2649 🔵

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -226,38 +226,17 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with
...
-                    logging.debug(this, s"[POST] action rejected with code $code: $message")
-                    terminate(code, message)
-                  case Failure(t: Throwable) =>
-                    logging.error(this, s"[POST] action activation failed: ${t.getMessage}")
-                    terminate(InternalServerError)
+
+                // incoming parameters may not override final parameters (i.e., parameters with already defined values)
+                // on an action once its parameters are resolved across package and binding
+                val allowInvoke = payload
+                  .map(_.fields.keySet.intersect(actionWithMergedParams.immutableParameters).isEmpty)
```
markusthoemmes:
```scala
_.fields.keySet.forall(key => !actionWithMergedParams.immutableParameters.contains(key))
```

for an early exit on violation instead of computing the whole intersection?

PR #3335, opened 2018-02-23
Allow container tests to run without a proxy.

rabbah:
Running container unit tests on docker-for-mac does not work because the 172. network is not reachable. Instead detect "local" mode and use port forwarding instead.

@akrabat FYI.

### DIFF
--- a/tests/src/test/scala/actionContainers/ActionContainer.scala
+++ b/tests/src/test/scala/actionContainers/ActionContainer.scala
@@ -135,19 +135,31 @@ object ActionContainer {
       case (k, v) => s"-e ${k}=${v}"
     } mkString (" ")
 
-    // We create the container...
-    val runOut = awaitDocker(s"run --name $name $envArgs -d $imageName", 10 seconds)
-    assert(runOut._1 == 0, "'docker run' did not exit with 0: " + runOut)
+    // We create the container... and find out its IP address...
+    def createContainer(portFwd: Option[Int] = None): Unit = {
+      val runOut = awaitDocker(
+        s"run ${portFwd.map(p => s"-p $p:8080").getOrElse("")} --name $name $envArgs -d $imageName",
+        10 seconds)
+      assert(runOut._1 == 0, "'docker run' did not exit with 0: " + runOut)
+    }
 
     // ...find out its IP address...
-    val ipOut = awaitDocker(s"""inspect --format '{{.NetworkSettings.IPAddress}}' $name""", 10 seconds)
-    assert(ipOut._1 == 0, "'docker inspect did not exit with 0")
-    val ip = ipOut._2.replaceAll("""[^0-9.]""", "")
+    val (ip, port) = if (WhiskProperties.getProperty("whisk.version.name") == "local") {
+      val p = 8988 // port must be available or docker run will fail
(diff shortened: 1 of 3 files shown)


rabbah:
@csantanapr so.... all the runtime repos cloned and owned this code :)

akrabat:
I copied the changed `ActionContainer.scala` file to my clone of incubator-openwhisk-runtime-php (Note that the package name is different in the version of this file in incubator-openwhisk-runtime-php…)

The tests ran. See https://gist.github.com/akrabat/60c1a0d7cc9b7d4acdb0a64b7da81230

rabbah:
There may be a heisenbug with my sleep hack - several travis builds failed in a row. Will have to investigate that.

tests/src/test/scala/actionContainers/ActionContainer.scala
```diff
@@ -135,19 +135,32 @@ object ActionContainer {
       case (k, v) => s"-e ${k}=${v}"
     } mkString (" ")
 
-    // We create the container...
-    val runOut = awaitDocker(s"run --name $name $envArgs -d $imageName", 10 seconds)
-    assert(runOut._1 == 0, "'docker run' did not exit with 0: " + runOut)
+    // We create the container... and find out its IP address...
+    def createContainer(portFwd: Option[Int] = None) = {
+      val runOut = awaitDocker(
+        s"run ${portFwd.map(p => s"-p $p:8080").getOrElse("")} --name $name $envArgs -d $imageName",
```
csantanapr: @rabbah I thought we try at one point to remove the proxy but we ran into trouble, and that's we have the proxy.

can't recall now why didn't push forward for a solution.

But this would be great if we can get rid off the proxy.

csantanapr:
Yeah we need to see if we can re-use this class in the runtime repos. Can't remember why I copied the class. I guess I didn't know much scala then.

rabbah:
pg5/115 🔵

csantanapr:
>There may be a heisenbug with my sleep hack - several travis builds failed in a row. Will have to investigate that.

@rabbah how confident are you on this?
You did some investigate to ruled out some of the previous failures?

rabbah:
At 1.5s it seems to be stable. But I can’t be 100% certain. I thought about alternate solutions, for example, retrying the /init route with an empty response to get an OK. This would work for most cases except there are some tests that deal with a non response that would have to change. 

I think this is the better solution though (port mapping) vs using a proxy so I’m inclined to say let’s see if there are heisenbugs in the long run and we can address them.

Of course I’m open to alternate solutions and welcome suggestions.

csantanapr:
@rabbah sounds good just wanted to pick your brain on possible solutions down the road if we start seen issues

I agree 💯 dealing with the proxy every day for me is  PITA 😿 

>for example, retrying the /init route with an empty response to get an OK. This would work for most cases except there are some tests that deal with a non response that would have to change.

Let’s move forward with this solution refactor the  runtime repos to reuse this class from core.

ddragosd:
💯   from me as well. This should make it easier for any new developer.

csantanapr:
_¡que Dios nos coja confesados!_ merging...

PR #3337, opened 2018-02-23
Put active-ack consumers in their own consumer-groups.

markusthoemmes:
Just like with the invoker consumers, it doesn't make sense to have those in one group as crash of one will cause a rebalancing pause for the other.


markusthoemmes:
PG4 1415 greenybeeny

ningyougang:
Oh.good solution, i also found the problem one month again. but use another solution.
anyway, this solution is better.

PR #3338, opened 2018-02-23
implement suspend/resume for KubernetesContainer

dgrove-oss:
We can implement suspend and resume for KubernetesContainerPool by delegating the operations to an invoker-agent that runs as a DaemonSet on every invoker node.  The invoker-agent interacts with the local docker/containerd engine to pause/unpause containers as directed by the invoker.

This PR contains the OpenWhisk core changes to implement this design.  It assumes an invoker-agent as implemented in https://github.com/apache/incubator-openwhisk-deploy-kube/pull/155.  The main changes in this PR are:
1.  Add configuration for the invoker-agent to application.conf (default to disabled).
2. Implement suspend/resume by delegating to the invoker-agent
3. Change the implementation of container creation to use the kubeRestClient to create the containers instead of using the kubectl command line.  As part of creation we now extract and save the workerNode IP information in the KubernetesContainer to cache the information needed to communicate with the invoker-agent.
4. Update test cases for changes in the KubernetesClient API.

If you look into the invoker-agent code in openwhisk-deploy-kube code, you will see that it also support log consolidation.   I plan to submit a separate PR for the openwhisk core side of the logging changes (because it also includes a modest refactor to the logging implementation that should be reviewed separately from the Kubernetes-centric changes in this PR).

### DIFF
--- a/core/invoker/src/main/resources/application.conf
+++ b/core/invoker/src/main/resources/application.conf
@@ -15,6 +15,7 @@ whisk {
   }
 
   kubernetes {
+    namespace: openwhisk
     # Timeouts for k8s commands. Set to "Inf" to disable timeout.
     timeouts {
       run: 1 minute
--- a/core/invoker/src/main/resources/application.conf
+++ b/core/invoker/src/main/resources/application.conf
@@ -14,16 +14,18 @@ whisk {
     unpause: 10 seconds
   }
 
-  # Timeouts for k8s commands. Set to "Inf" to disable timeout.
   kubernetes {
+    # Timeouts for k8s commands. Set to "Inf" to disable timeout.
     timeouts {
       run: 1 minute
       rm: 1 minute
       inspect: 1 minute
       logs: 1 minute
     }
-    invoker-agent: false
-    invoker-agent-port: 3233
+    invoker-agent {
+      enabled: false
+      port: 3233
+    }
   }
 
   # Timeouts for runc commands. Set to "Inf" to disable timeout.
@@ -39,5 +41,4 @@ whisk {
     extra-args: {}
 
   }
-
 }
--- a/core/invoker/src/main/resources/application.conf
+++ b/core/invoker/src/main/resources/application.conf
@@ -15,11 +15,15 @@ whisk {
   }
 
   # Timeouts for k8s commands. Set to "Inf" to disable timeout.
-  kubernetes.timeouts {
-    run: 1 minute
-    rm: 1 minute
-    inspect: 1 minute
-    logs: 1 minute
+  kubernetes {
+    timeouts {
+      run: 1 minute
+      rm: 1 minute
+      inspect: 1 minute
+      logs: 1 minute
+    }
+    invoker-agent: false
+    invoker-agent-port: 3233
   }
 
   # Timeouts for runc commands. Set to "Inf" to disable timeout.
@@ -35,4 +39,5 @@ whisk {
     extra-args: {}
 
   }
+
 }
--- a/core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
@@ -26,12 +26,11 @@ import java.time.format.DateTimeFormatterBuilder
 
 import akka.actor.ActorSystem
 import akka.event.Logging.{ErrorLevel, InfoLevel}
-import akka.http.scaladsl.model.Uri
+import akka.http.scaladsl.model.{HttpRequest, HttpResponse, Uri}
 import akka.http.scaladsl.model.Uri.Path
 import akka.http.scaladsl.model.Uri.Query
 import akka.stream.{Attributes, Outlet, SourceShape}
 import akka.http.scaladsl.Http
-import akka.http.scaladsl.model.HttpRequest
 import akka.stream.ActorMaterializer
 import akka.stream.scaladsl.Source
 import akka.stream.stage._
@@ -172,8 +171,10 @@ class KubernetesClient(
     runCmd(Seq("delete", "--now", "pod", container.id.asString), config.timeouts.rm).map(_ => ())
   }
 
-  def rm(key: String, value: String)(implicit transid: TransactionId): Future[Unit] = {
-    if (config.invokerAgent.enabled) {
+  def rm(key: String, value: String, ensureUnpaused: Boolean = false)(implicit transid: TransactionId): Future[Unit] = {
+    if (ensureUnpaused && config.invokerAgent.enabled) {
+      // The caller can't guarantee that every container with the label key=value is already unpaused.
+      // Therefore we must enumerate them and ensure they are unpaused before we attempt to delete them.
       Future {
         blocking {
           kubeRestClient
@@ -184,8 +185,11 @@ class KubernetesClient(
             .getItems
             .asScala
             .map { pod =>
-              // Call destroy to ensure container is resumed before it is deleted.
-              toContainer(pod).destroy()
+              val container = toContainer(pod)
+              container
+                .resume()
+                .recover { case _ => () } // Ignore errors; it is possible the container was not actually suspended.
+                .map(_ => rm(container))
             }
         }
       }.flatMap(futures =>
@@ -199,9 +203,7 @@ class KubernetesClient(
 
   def suspend(container: KubernetesContainer)(implicit transid: TransactionId): Future[Unit] = {
     if (config.invokerAgent.enabled) {
-      // Forward command to invoker-agent daemonset instance on container's worker node
-      Http()
-        .singleRequest(HttpRequest(uri = agentCommand("suspend", container)))
+      agentCommand("suspend", container)
         .map { response =>
           response.discardEntityBytes()
         }
@@ -212,9 +214,7 @@ class KubernetesClient(
 
   def resume(container: KubernetesContainer)(implicit transid: TransactionId): Future[Unit] = {
     if (config.invokerAgent.enabled) {
-      // Forward command to invoker-agent daemonset instance on container's worker node
-      Http()
-        .singleRequest(HttpRequest(uri = agentCommand("resume", container)))
+      agentCommand("resume", container)
         .map { response =>
           response.discardEntityBytes()
         }
@@ -245,12 +245,14 @@ class KubernetesClient(
     new KubernetesContainer(id, addr, workerIP, nativeContainerId)
   }
 
-  private def agentCommand(command: String, container: KubernetesContainer): Uri = {
-    Uri()
+  // Forward a command to invoker-agent daemonset instance on container's worker node
+  private def agentCommand(command: String, container: KubernetesContainer): Future[HttpResponse] = {
+    val uri = Uri()
       .withScheme("http")
       .withHost(container.workerIP)
       .withPort(config.invokerAgent.port)
       .withPath(Path / command / container.nativeContainerId)
+    Http().singleRequest(HttpRequest(uri = uri))
   }
 
   private def runCmd(args: Seq[String], timeout: Duration)(implicit transid: TransactionId): Future[String] = {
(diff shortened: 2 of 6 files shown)


dgrove-oss:
@jcrossley3 @bbrowning @bwmcadams  Please review proposed k8s changes.  Thanks.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -41,17 +45,19 @@ import whisk.core.ConfigKeys
...
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.blocking
-import scala.concurrent.duration._
 import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
 import spray.json._
 import spray.json.DefaultJsonProtocol._
+import collection.JavaConversions._
```
markusthoemmes: Please use `JavaConverters`. They'll allow you to write `asJava` or `asScala` after a usual `Map` for example. `JavaConversions` makes heavy use of implicits and are discouraged.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -99,43 +111,133 @@ class KubernetesClient(
...
+    }.toSeq
+
+    val pod = new PodBuilder()
+      .withNewMetadata()
+      .withName(name)
+      .addToLabels("name", name)
+      .addToLabels(mapAsJavaMap(labels))
+      .endMetadata()
+      .withNewSpec()
+      .withRestartPolicy("Always")
```
markusthoemmes: Do we really want user-pods restarted?

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -99,43 +111,133 @@ class KubernetesClient(
...
+        Future.failed(new Exception(s"Failed to create pod '$name'"))
     }
   }
 
-  def rm(id: ContainerId)(implicit transid: TransactionId): Future[Unit] =
-    runCmd(Seq("delete", "--now", "pod", id.asString), timeouts.rm).map(_ => ())
+  def rm(container: KubernetesContainer)(implicit transid: TransactionId): Future[Unit] = {
+    // Pod deletion will never complete if the pod contains a paused container.
+    // Therefore issue a resume before the delete (resuming a non-suspended container is harmless).
+    resume(container).map { _ =>
```
markusthoemmes: This should be taken care of by the container lifecycle.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -99,43 +111,133 @@ class KubernetesClient(
...
+    }
+  }
 
-  def logs(id: ContainerId, sinceTime: Option[Instant], waitForSentinel: Boolean = false)(
+  def suspend(container: KubernetesContainer)(implicit transid: TransactionId): Future[Unit] = {
+    if (config.invokerAgent) {
+      // Forward command to invoker-agent daemonset instance on container's worker node
+      Http()
+        .singleRequest(HttpRequest(
+          uri = "http://" + container.workerIP + ":" + config.invokerAgentPort + "/suspend/" + container.nativeContainerId))
```
markusthoemmes: Could make use of `akka-http`s `Uri` API:

```scala
Uri().withScheme("http").withHost(container.workerIp).withPort(config.invokerAgentPort).withPath(/ "suspend" / container.nativeContainerId)
```

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -99,43 +111,133 @@ class KubernetesClient(
...
+            .reduce((a, b) => a.flatMap(_ => b))
+        }
+      }
+    } else {
+      runCmd(Seq("delete", "--now", "pod", "-l", s"$key=$value"), config.timeouts.rm).map(_ => ())
+    }
+  }
 
-  def logs(id: ContainerId, sinceTime: Option[Instant], waitForSentinel: Boolean = false)(
+  def suspend(container: KubernetesContainer)(implicit transid: TransactionId): Future[Unit] = {
```
markusthoemmes: Should we move those to the container instead of "faking" them in the client? Seems weird to pass the whole container to the client.

core/invoker/src/main/resources/application.conf
```diff
@@ -15,11 +15,15 @@ whisk {
...
-    logs: 1 minute
+  kubernetes {
+    timeouts {
+      run: 1 minute
+      rm: 1 minute
+      inspect: 1 minute
+      logs: 1 minute
+    }
+    invoker-agent: false
+    invoker-agent-port: 3233
```
markusthoemmes: Should we nest the invoker agent config in something like 

```
invoker-agent {
  use: false
  port: 3233
}
```

?

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
dgrove-oss: Preserving the behavior that was there.  I agree we don't actually want them to be restarted but @bbrowning said in #3219 that on OpenShift they need to set the restart policy to Always as a hack to prevent the containers from being arbitrarily killed by the system.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
dgrove-oss: Makes sense; I restructured.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
dgrove-oss: I think the implementations belong in the KubernetesClient, since they need access to some of its state too (its config object). 

I agree passing the entire KubernetesContainer object as a parameter looks a bit backwards.  I decided to do that (instead of passing a few of its fields individually) as a hedge against future churn and because it felt like a nice logical unit (vs. passing containerId, workerIP, nativeContainerId).   But I don't feel that strongly about this.

dgrove-oss:
PG3 / 1913 🔵

dgrove-oss:
rebased to latest master and consolidated small commits made while addressing review comments.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -99,43 +116,139 @@ class KubernetesClient(
...
+      .withEnv(envVars.asJava)
+      .addNewPort()
+      .withContainerPort(8080)
+      .withName("action")
+      .endPort()
+      .endContainer()
+      .endSpec()
+      .build()
+
+    kubeRestClient.pods.inNamespace("openwhisk").create(pod)
```
jcrossley3: I don't love the hardcoded namespace here. If we must have it, can we at least make it configurable?

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -99,43 +116,139 @@ class KubernetesClient(
...
 
-  def inspectIPAddress(id: ContainerId)(implicit transid: TransactionId): Future[ContainerAddress] = {
     Future {
       blocking {
-        val pod =
-          kubeRestClient.pods().withName(id.asString).waitUntilReady(timeouts.inspect.length, timeouts.inspect.unit)
-        ContainerAddress(pod.getStatus().getPodIP())
+        val createdPod = kubeRestClient.pods
+          .inNamespace("openwhisk")
```
jcrossley3: Here, too

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -99,43 +116,139 @@ class KubernetesClient(
...
+  def rm(container: KubernetesContainer)(implicit transid: TransactionId): Future[Unit] = {
+    runCmd(Seq("delete", "--now", "pod", container.id.asString), config.timeouts.rm).map(_ => ())
+  }
+
+  def rm(key: String, value: String)(implicit transid: TransactionId): Future[Unit] = {
+    if (config.invokerAgent.enabled) {
+      Future {
+        blocking {
+          kubeRestClient
+            .inNamespace("openwhisk")
```
jcrossley3: And here :)

dgrove-oss: good point.  Moved the namespace intoapplication.conf.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
bwmcadams: Agreed on this one; I'm not a fan of "magic" when it comes to collection behavior.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
bwmcadams: Would it make sense to have this be a user configurable (perhaps as a future update)? For now, obviously our focus is on the OpenShift side but longterm `Always` isn't-as pointed out-going to work for everyone.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
bwmcadams: With regards to configurable vs. hardcoded, I'd argue the same with ports and memory limits. One size won't fit all.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
```diff
@@ -57,10 +57,10 @@ object KubernetesContainer {
              image: String,
              userProvidedImage: Boolean = false,
              memory: ByteSize = 256.MB,
-             environment: Map[String, String] = Map(),
-             labels: Map[String, String] = Map())(implicit kubernetes: KubernetesApi,
-                                                  ec: ExecutionContext,
-                                                  log: Logging): Future[KubernetesContainer] = {
+             environment: Map[String, String] = Map.empty,
+             labels: Map[String, String] = Map.empty)(implicit kubernetes: KubernetesApi,
```
bwmcadams: 👍 
I much prefer `Map.empty` as it, in my eyes, is cleared reading wise.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -99,43 +117,144 @@ class KubernetesClient(
...
 
   }
 
+  private def toContainer(pod: Pod): KubernetesContainer = {
+    val id = ContainerId(pod.getMetadata.getName)
+    val addr = ContainerAddress(pod.getStatus.getPodIP)
+    val workerIP = pod.getStatus.getHostIP
+    // Extract the native (docker or containerd) containerId for the container
+    // By convention, kubernetes adds a docker:// prefix when using docker as the low-level container engine
+    val nativeContainerId = pod.getStatus.getContainerStatuses.get(0).getContainerID.stripPrefix("docker://")
```
glikson: Would it make sense to make the invoker-agent take care of the low-level details, and make the logic here agnostic of specific container engine implementation? For example, if explicit docker interaction is replaced with CRI?

dgrove-oss: not a bad idea, but let's defer that to a later PR.  I'd like to get this one merged quickly, so I can submit the logging implementation I have that builds on it. Then can rethink how we pass the nativeContainerId to the invokerAgent.

PR #3339, opened 2018-02-23
Add the NOTICE file for future development

houshengbo:


rabbah:
can this be a markdown file?

houshengbo:
@rabbah Apache does not define the suffix of the license and notice files. I need to confirm whether we can use .md files.

csantanapr:
empty file WTF?

PR #3340, opened 2018-02-23
Default activation response to empty JsObject.

markusthoemmes:
When using a blocking call, reduced to return only the response the unsafe conversion creates an uncaught exception.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
@@ -75,13 +75,13 @@ case class WhiskActivation(namespace: EntityPath,
   require(end != null, "end undefined")
   require(response != null, "response undefined")
 
-  def toJson = WhiskActivation.serdes.write(this).asJsObject
+  def toJson: JsObject = WhiskActivation.serdes.write(this).asJsObject
 
   /**
    * This the activation summary as computed by the database view.
    * Strictly used in view testing to enforce alignment.
    */
-  override def summaryAsJson = {
+  override def summaryAsJson: JsObject = {
     import WhiskActivation.instantSerdes
 
     def actionOrNot() = {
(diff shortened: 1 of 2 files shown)


rabbah:
This is OK - where are we using the response result though incorrectly; isn't there another bug?

markusthoemmes:
I discovered it in a loopback invoker test where I stubbed the invoker out. Was a mere accident. I was thinking though if we should change the type of `reponse` to not be `Option`. Didn't try what that'd mean though.

rabbah:
could be... or remove the option and use an empty object. None is used on the non-blocking path so the result doesn’t matter.

markusthoemmes:
No PG needed here I'd say?

common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
```diff
@@ -97,13 +97,13 @@ case class WhiskActivation(namespace: EntityPath,
...
-        cause.map(("cause" -> _.toJson)) ++
+        cause.map("cause" -> _.toJson) ++
         actionOrNot())
   }
 
-  def resultAsJson = response.result.toJson.asJsObject
+  def resultAsJson: JsObject = response.result.map(_.toJson.asJsObject).getOrElse(JsObject.empty)
 
-  def toExtendedJson = {
+  def toExtendedJson: JsObject = {
```
rabbah: can you check this one as well

rabbah:
Travis is enough - The result = None is not observable unless there's an invariant violation; returning result = None and a full activation instead of an id (I suppose the loopback return a WhiskActivation() here for example). 

It would be nice to tighten but I think it's equally OK to treat a missing result as `{}`... until we want to support `JsNull` as a possible response as well (related to sequence semantics).

common/scala/src/main/scala/whisk/core/entity/WhiskActivation.scala
rabbah: i checked it - it's fine.

markusthoemmes:
Closing this as abandoned.

Issue #3168, opened 2018-01-09
Rate limit requests to download CLI, Docker, and iOS SDKs


dubeejw:
Nginx should be configured so that requests to download the CLI, Docker and iOS SDKs are rate limited.

Nginx config: https://github.com/apache/incubator-openwhisk/blob/master/ansible/roles/nginx/templates/nginx.conf.j2
Rate limiting doc here: https://www.nginx.com/blog/rate-limiting-nginx/

mptap:
@dubeejw: Please could you review? https://github.com/apache/incubator-openwhisk/pull/3342 Assumed key = $binary_remote_addr, zone = 10MB, rate = 10 reqs/sec and burst = 20 as in the example docs. Let me know if you want me to use other values.

dubeejw:
@mptap, I'll take a look. Thanks for working on this issue!

PR #3342, opened 2018-02-24
 #3168 Rate limit requests to download CLI, Docker and iOS SDKs

mptap:
Nginx configured so that requests to download the CLI, Docker and iOS SDKs are rate limited.

### DIFF
--- a/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/ansible/roles/nginx/templates/nginx.conf.j2
@@ -9,6 +9,9 @@ events {
 }
 
 http {
+{# nginx rate limit config parameters #}
+    limit_req_zone $binary_remote_addr zone=mylimit:10m rate=10r/s;
+    
 {# allow large uploads, need to thread proper limit into here #}
     client_max_body_size 50M;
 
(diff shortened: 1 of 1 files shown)


cbickel:
@mptap 
Thanks a lot for your contribution. I like the idea of limiting the download of some attributes, to prevent dos attacks.

But I also have two concerns: 
- If this nginx is not the front-door, (e.g. Openwhisk is only one of many offerings of one company), than the rate limit would apply on the frontdoor in front of this nginx, right? Then users may experience issues in downloading the CLI, because others do this as well.
- If a company or university wants to introduce OpenWhisk, the students/employees may try to download the CLI at once. In most cases, this is one IP as well. Then the first impression for some of them may be a bad experience.

A agree, that these two examples are only edge-cases. But I think as this is the first experience of a new user, we have to think about it.

One solution to work around this (especially the first problem) may be to make the limits configurable (and add the ability to disable them).

Do you think it makes sense to write a short test, to check that the limits are really applied? (Also to avoid, that they are not applied at some day anymore, because of some configuration changes by other contributors)

ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -94,18 +97,26 @@ http {
...
         }
         # leaving this for a while for clients out there to update to the new endpoint
         location /blackbox-0.1.0.tar.gz {
+            limit_req zone=mylimit burst=20 nodelay;
+            
             return 301 /blackbox.tar.gz;
         }
 
         location /OpenWhiskIOSStarterApp.zip {
+            limit_req zone=mylimit burst=20 nodelay;
```
cbickel: As we only return a 301 to a github address: Do we really need to throttle this request (and the others with 301)?

rabbah:
@mptap are you interested in making the rate limits configurable to a deployment?

PR #3343, opened 2018-02-24
Support time-series DB using Kamon Tag

upgle:
This PR is for supporting time-series database using Kamon Tag like OpenTSDB, Datadog.

and I also added granular metric again which was removed from #3284 PR. (but it is **optional**, default is False) because we need this feature for tracking activities per namespace, and already have built up enough infrastructure.

### Reference

http://opentsdb.net/docs/build/html/user_guide/writing.html
> OpenTSDB handles things a bit differently by introducing the idea of 'tags'. Each time series still has a 'metric' name, but it's much more generic, something that can be shared by many unique time series. Instead, the uniqueness comes from a combination of tag key/value pairs that allows for flexible queries with very fast aggregations.

https://docs.datadoghq.com/agent/tagging/
> Tagging is used throughout the Datadog product to make it easier to subset and query the machines and metrics that you have to monitor.

## Metrics tested in real environment.
_* It is tested on both OpenTSDB and Graphite_

### 1. OpenTSDB (Tag based aggregation)
In OpenTSDB, you can use regexable aggregation query with tag only.
e.g. `host=regexp(web[0-9]+.lax.mysite.com)`

image

### 2. Graphite (Metric name based aggregation)
In Graphite, you can use regexable aggregation query with a metric name only.
e.g. `invoker_containerStart_*`

image

image


csantanapr:
If it’s optional and and is false by default I don’t see an issue to add in the grand scheme of things. 
👍

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -164,27 +164,44 @@ private object Logging {
...
   override def toString() = component + "_" + action + "_" + state
 
   def asFinish = copy(state = LoggingMarkers.finish)
   def asError = copy(state = LoggingMarkers.error)
 }
 
 object LogMarkerToken {
+
+  def apply(component: String, action: String, state: String, tags: Map[String, String]): LogMarkerToken = {
+    if (!TransactionId.metricsKamonTags) {
```
markusthoemmes: Should this be handled by the `LogMarkerToken`'s `toString`?

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -193,20 +210,17 @@ object MetricEmitter {
 
   def emitCounterMetric(token: LogMarkerToken): Unit = {
     if (TransactionId.metricsKamon) {
-      metrics
-        .counter(token.toString)
-        .increment(1)
+      if (token.tags.nonEmpty) metrics.counter(token.toString, token.tags.get).increment(1)
+      else metrics.counter(token.toString).increment(1)
```
markusthoemmes: Should be:

```scala
token.tags match {
  case Some(tags) => metrics.counter(token.toString, token.tags).increment(1)
  case None => metrics.counter(token.toString).increment(1)
}
```

Same below.

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -259,11 +273,14 @@ object LoggingMarkers {
 
   // Time in invoker
   val INVOKER_ACTIVATION = LogMarkerToken(invoker, activation, start)
-  def INVOKER_DOCKER_CMD(cmd: String) = LogMarkerToken(invoker, s"docker.$cmd", start)
-  def INVOKER_RUNC_CMD(cmd: String) = LogMarkerToken(invoker, s"runc.$cmd", start)
-  def INVOKER_CONTAINER_START(containerState: String) =
-    LogMarkerToken(invoker, s"container_start_${containerState}", count)
-  def INVOKER_KUBECTL_CMD(cmd: String) = LogMarkerToken(invoker, s"kubectl.$cmd", start)
+  def INVOKER_DOCKER_CMD(cmd: String) = LogMarkerToken(invoker, "docker", start, Map("cmd" → cmd))
```
markusthoemmes: Please remove all ASCII arrows and replace them with `->` (memo to self: automate this check)

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -164,27 +164,44 @@ private object Logging {
...
   }
-
 }
 
 private object Emitter {
   val timeFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'").withZone(ZoneId.of("UTC"))
 }
 
-case class LogMarkerToken(component: String, action: String, state: String) {
+case class LogMarkerToken(component: String, action: String, state: String, tags: Option[Map[String, String]] = None) {
```
markusthoemmes: I've been looking through all the PR, and it feels a bit unclear to me to grasp all the branches on when something will be logged/sent and how. Maybe we can simplify:

What do you think about creating 

```scala
LogMarkerToken(component: String, action: String, state: String, macroTags: Map[String, String] = Map.empty, microTags: Map[String, String] = Map.empty)
```

Then we could have

```scala
override def toString = component + "_" + action + "_" + state
override def toStringWithTags = component + "_" + action + macroTags.mkString(".", ".", "") + "_" + state
```

And call them in the specific call-site accordingly:

```scala
def emitCounterMetric(token: LogMarkerToken): Unit = {
  if (TransactionId.metricsKamon) {
    if (TransactionId.metricsKamonTags) {
      metrics.counter(token.toString, token.macroTags ++ token.microTags).increment(1)
    } else {
      metrics.counter(token.toStringWithTags).increment(1)
    }
  }
}
```

We maybe don't even need the granular option then. If the backend supports tags, we're fine anyway, right? We just need to make sure we only log the macro tags if the backend does not support "true" tags.

WDYT?

common/scala/src/main/scala/whisk/common/Logging.scala
markusthoemmes: Edit: Might be ignorable, see lengthy comment.

common/scala/src/main/scala/whisk/common/Logging.scala
upgle: Yeah, I agree with this all and think it is a very clear solution. 
I will modify the code, test it, and then commit it again.

Thank you for your kind review!

upgle: @markusthoemmes 
I changed the code, but I'm worried also the scaleable issue that occurs when using a low-capacity and low-performance Kamon-backend that supports tags.

Should we make a granular option again? 
What do you think..?

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -164,27 +164,34 @@ private object Logging {
...
   }
-
 }
 
 private object Emitter {
   val timeFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'").withZone(ZoneId.of("UTC"))
 }
 
-case class LogMarkerToken(component: String, action: String, state: String) {
+case class LogMarkerToken(component: String,
```
markusthoemmes: A ScalaDoc comment here would be great, stating that `macroTags` should be used for tags with a bounded number of permutations only while microTags can be used for whatever granularity you might need.

common/scala/src/main/scala/whisk/http/BasicHttpService.scala
```diff
@@ -121,7 +121,11 @@ trait BasicHttpService extends Directives with TransactionCounter {
 
       val name = "BasicHttpService"
 
-      val token = LogMarkerToken("http", s"${m.toLowerCase}.${res.status.intValue}", LoggingMarkers.count)
+      val token = LogMarkerToken(
+        "http",
+        s"${m.toLowerCase}",
```
markusthoemmes: Doesn't need string interpolation anymore. Just `m.toLowerCase` should suffice.

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -259,11 +269,16 @@ object LoggingMarkers {
...
-  def INVOKER_CONTAINER_START(containerState: String) =
-    LogMarkerToken(invoker, s"container_start_${containerState}", count)
-  def INVOKER_KUBECTL_CMD(cmd: String) = LogMarkerToken(invoker, s"kubectl.$cmd", start)
+  def INVOKER_DOCKER_CMD(cmd: String) = LogMarkerToken(invoker, "docker", start, Map("cmd" -> cmd))
+  def INVOKER_RUNC_CMD(cmd: String) = LogMarkerToken(invoker, "runc", start, Map("cmd" -> cmd))
+  def INVOKER_KUBECTL_CMD(cmd: String) = LogMarkerToken(invoker, s"kubectl", start, Map("cmd" -> cmd))
+  def INVOKER_CONTAINER_START(containerState: String, namespaceName: String, actionName: String) =
+    LogMarkerToken(
+      invoker,
+      s"containerStart",
```
markusthoemmes: No string interpolation needed. Just `containerStart` is fine.

upgle: Modified. thank you

upgle:
@markusthoemmes could you please review again?

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -164,27 +164,47 @@ private object Logging {
...
+                          action: String,
+                          state: String,
+                          macroTags: Map[String, String] = Map.empty,
+                          microTags: Map[String, String] = Map.empty) {
+
   override def toString() = component + "_" + action + "_" + state
 
+  // folderLeft is used instead of mkString(".", ".", ""),
+  // because Map.empty.mkString returns "."
+  def toStringWithTags = component + "_" + action + getTags.values.foldLeft("")((a, v) => a + "." + v) + "_" + state
```
markusthoemmes: You can simplify to

```scala
def toStringWithTags(includeMicroTags: Boolean) = {
  val tags = if(includeMicroTags) macroTags ++ microTags else macroTags
  val tagString = tags.values.foldLeft("")(_ + "." + _) // graceful handling of an empty map
  component + "_" + action + tagString + "_" + state
}
```

**Note:** I'd remove the entanglement of this instance to some arbitrary global value and pass it in instead for better control (in tests for instance).

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -193,20 +213,23 @@ object MetricEmitter {
 
   def emitCounterMetric(token: LogMarkerToken): Unit = {
     if (TransactionId.metricsKamon) {
-      metrics
-        .counter(token.toString)
-        .increment(1)
+      if (TransactionId.metricsKamonTags) {
+        metrics.counter(token.toString, token.getTags).increment(1)
+      } else {
+        metrics.counter(token.toStringWithTags).increment(1)
```
markusthoemmes: Changes to

```scala
metrics.counter(token.toStringWithTags(includeMicroTags = false)).increment(1)
```

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -193,20 +213,23 @@ object MetricEmitter {
 
   def emitCounterMetric(token: LogMarkerToken): Unit = {
     if (TransactionId.metricsKamon) {
-      metrics
-        .counter(token.toString)
-        .increment(1)
+      if (TransactionId.metricsKamonTags) {
+        metrics.counter(token.toString, token.getTags).increment(1)
```
markusthoemmes: Changes to

```scala
metrics.counter(token.toString, if(TransactionId.granularMetric) token.allTags else token.macroTags).increment(1)
```

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -164,27 +164,47 @@ private object Logging {
...
+                          state: String,
+                          macroTags: Map[String, String] = Map.empty,
+                          microTags: Map[String, String] = Map.empty) {
+
   override def toString() = component + "_" + action + "_" + state
 
+  // folderLeft is used instead of mkString(".", ".", ""),
+  // because Map.empty.mkString returns "."
+  def toStringWithTags = component + "_" + action + getTags.values.foldLeft("")((a, v) => a + "." + v) + "_" + state
+  def getTags = if (TransactionId.granularMetric) macroTags ++ microTags else macroTags
```
markusthoemmes: How about `def allTags = macroTags ++ microTags` and use it from the outside?

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -164,27 +164,47 @@ private object Logging {
...
 }
 
-case class LogMarkerToken(component: String, action: String, state: String) {
+/**
+ * Used to record log message and make a metric name.
+ *
+ * @param component Component like invoker, controller, and docker. It is defined in LoggingMarkers.
+ * @param action Action of the component.
+ * @param state State of the action.
+ * @param macroTags macroTags should be used for tags with a bounded number of permutations.
```
mhenke1: Please add an example for such a bounded tag (e.g. http staus code)

upgle: ok, I'll add the annotation now. and how about changing naming convention to `container_start` to `containerStart`?

markusthoemmes:
@upgle it's taking me so long because there is a test failure in RuncClient tests. It is currently not run by travis due to a bug (will be resolved shoon #3367). Can you please take a look?

upgle:
@markusthoemmes has it been resolved?

markusthoemmes:
@upgle it's merged yep, you can now rebase and try if the test passes (i think it won't).

upgle:
The test is passed, and some test code has modified to use log parser which changed the signature.
now if there are some tags, log parser for unit test uses LogMarkerToken as a hint to parsing.
please review again. @markusthoemmes

upgle:
conflicts have been resolved.

markusthoemmes:
PG3 1939 🔵

PR #3347, opened 2018-02-26
Generic Pooling REST Client

dubeejw:
Moves the REST client from CouchDbRestClient to its own class that can be reused by other classes that need a generic REST client.

### DIFF
--- a/common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
+++ b/common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
@@ -17,9 +17,6 @@
 
 package whisk.http
 
-import java.net.URLEncoder
-import java.nio.charset.StandardCharsets
-
 import scala.concurrent.Future
 import scala.concurrent.Promise
 import scala.util.{Failure, Success}
@@ -30,7 +27,6 @@ import akka.http.scaladsl.Http.HostConnectionPool
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.marshalling._
 import akka.http.scaladsl.model._
-import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.unmarshalling._
 import akka.stream.ActorMaterializer
 import akka.stream.OverflowStrategy
@@ -40,14 +36,13 @@ import akka.stream.scaladsl._
 import spray.json._
 
 /**
- * This class only handles the basic communication to the proper endpoints
- *  ("JSON in, JSON out"). It is up to its clients to interpret the results.
- *  It is built on akka-http host-level connection pools; compared to single
- *  requests, it saves some time on each request because it doesn't need to look
- *  up the pool corresponding to the host. It is also easier to add an extra
- *  queueing mechanism.
+ * This class only handles the basic communication to the proper endpoints.
+ *  It is up to its clients to interpret the results. It is built on akka-http
+ *  host-level connection pools; compared to single requests, it saves some time
+ *  on each request because it doesn't need to look up the pool corresponding
+ *  to the host. It is also easier to add an extra queueing mechanism.
  */
-class RestClient(protocol: String, host: String, port: Int)(implicit system: ActorSystem) {
+class PoolingRestClient(protocol: String, host: String, port: Int, queueSize: Int)(implicit system: ActorSystem) {
   require(protocol == "http" || protocol == "https", "Protocol must be one of { http, https }.")
 
   implicit val context = system.dispatcher
@@ -65,9 +60,8 @@ class RestClient(protocol: String, host: String, port: Int)(implicit system: Act
   // Additional queue in case all connections are busy. Should hardly ever be
   // filled in practice but can be useful, e.g., in tests starting many
   // asynchronous requests in a very short period of time.
-  private val QUEUE_SIZE = 16 * 1024;
   private val requestQueue = Source
-    .queue(QUEUE_SIZE, OverflowStrategy.dropNew)
+    .queue(queueSize, OverflowStrategy.dropNew)
     .via(pool.mapMaterializedValue { x =>
       poolPromise.success(x); x
     })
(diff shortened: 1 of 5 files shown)


dubeejw:
PG3 1029 ⏳

common/scala/src/main/scala/whisk/http/RestClient.scala
```diff
@@ -0,0 +1,171 @@
...
+  } else {
+    Http().cachedHostConnectionPoolHttps[Promise[HttpResponse]](host = host, port = port)
+  }
+
+  private val poolPromise = Promise[HostConnectionPool]
+
+  // Additional queue in case all connections are busy. Should hardly ever be
+  // filled in practice but can be useful, e.g., in tests starting many
+  // asynchronous requests in a very short period of time.
+  private val QUEUE_SIZE = 16 * 1024;
```
markusthoemmes: This should be made configurable.

common/scala/src/main/scala/whisk/http/RestClient.scala
```diff
@@ -0,0 +1,171 @@
...
+    .toMat(Sink.foreach({
+      case ((Success(response), p)) => p.success(response)
+      case ((Failure(error), p))    => p.failure(error)
+    }))(Keep.left)
+    .run
+
+  // Properly encodes the potential slashes in each segment.
+  protected def uri(segments: Any*): Uri = {
+    val encodedSegments = segments.map(s => URLEncoder.encode(s.toString, StandardCharsets.UTF_8.name))
+    Uri(s"/${encodedSegments.mkString("/")}")
```
markusthoemmes: Should we investigate if `Uri.Path` would give us the right behavior straight away? I think this is an implementation detail of the CouchDB store (because we have slashes in documents there) so we should consider moving it out of the generalization.

common/scala/src/main/scala/whisk/http/RestClient.scala
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * This class only handles the basic communication to the proper endpoints
+ *  ("JSON in, JSON out"). It is up to its clients to interpret the results.
+ *  It is built on akka-http host-level connection pools; compared to single
+ *  requests, it saves some time on each request because it doesn't need to look
+ *  up the pool corresponding to the host. It is also easier to add an extra
+ *  queueing mechanism.
+ */
+class RestClient(protocol: String, host: String, port: Int)(implicit system: ActorSystem) {
```
markusthoemmes: Should we name this more specifically? `PoolingRestClient` for instance?

common/scala/src/main/scala/whisk/http/RestClient.scala
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  // Headers common to all requests.
+  val baseHeaders: List[HttpHeader] = List(Accept(MediaTypes.`application/json`))
+
+  // Prepares a request with the proper headers.
+  def mkRequest0(method: HttpMethod,
+                 uri: Uri,
+                 body: Future[MessageEntity],
+                 headers: Option[List[HttpHeader]] = None): Future[HttpRequest] = {
```
markusthoemmes: Should headers be an Option? Wouldn't an empty list suffice (the implementation defaults to that anyway)

common/scala/src/main/scala/whisk/http/RestClient.scala
```diff
@@ -0,0 +1,171 @@
...
+    .run
+
+  // Properly encodes the potential slashes in each segment.
+  protected def uri(segments: Any*): Uri = {
+    val encodedSegments = segments.map(s => URLEncoder.encode(s.toString, StandardCharsets.UTF_8.name))
+    Uri(s"/${encodedSegments.mkString("/")}")
+  }
+
+  // Headers common to all requests.
+  val baseHeaders: List[HttpHeader] = List(Accept(MediaTypes.`application/json`))
```
markusthoemmes: I wonder if it'd be better to relay those to the implementors. It's a bit weird to read the code and always have to assume that base-headers are passed in by magic.

dubeejw:
PG2 2862 🔵

common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
```diff
@@ -0,0 +1,154 @@
...
+      case ((Success(response), p)) => p.success(response)
+      case ((Failure(error), p))    => p.failure(error)
+    }))(Keep.left)
+    .run
+
+  // Prepares a request with the proper headers.
+  def mkRequest0(method: HttpMethod,
+                 uri: Uri,
+                 body: Future[MessageEntity],
+                 headers: List[HttpHeader] = List()): Future[HttpRequest] = {
```
markusthoemmes: Please use `List.empty` instead. Much less garbage and more clear.

common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
```diff
@@ -0,0 +1,154 @@
...
+    body.map { b =>
+      HttpRequest(method = method, uri = uri, headers = headers, entity = b)
+    }
+  }
+
+  protected def mkRequest(method: HttpMethod, uri: Uri, headers: List[HttpHeader] = List()): Future[HttpRequest] = {
```
markusthoemmes: Please use `List.empty` instead. Much less garbage and more clear.

common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
```diff
@@ -0,0 +1,154 @@
...
+    mkRequest0(method, uri, Future.successful(HttpEntity.Empty), headers)
+  }
+
+  protected def mkJsonRequest(method: HttpMethod,
+                              uri: Uri,
+                              body: JsValue,
+                              headers: List[HttpHeader] = List()): Future[HttpRequest] = {
```
markusthoemmes: Please use `List.empty` instead. Much less garbage and more clear.

dubeejw:
PG3 1925 🔵

PR #3348, opened 2018-02-26
Allow wskadmin user block/unblock to accept a list of subject to block

rabbah:
This patch allows wskadmin to block or unblock a list of subjects, separated by a comma (which is not a valid token for subject names). The first commit is a semantic preserving change. The second commit adds:

```
+    for subject in args.subject.split(","):
+        subject = subject.strip()
+        if len(subject) > 0:
```

around the existing code: https://github.com/apache/incubator-openwhisk/pull/3348/files?w=1

### DIFF
--- a/tools/admin/wskadmin
+++ b/tools/admin/wskadmin
@@ -111,8 +111,8 @@ def parseArgs():
     subcmd = subparser.add_parser('block', help='block one or more users')
     subcmd.add_argument('subject', help='comma separated list of users to block')
 
-    subcmd = subparser.add_parser('unblock', help='unblock a user')
-    subcmd.add_argument('subject', help='the user to unblock')
+    subcmd = subparser.add_parser('unblock', help='unblock one or more users')
+    subcmd.add_argument('subject', help='comma separated list of users to unblock')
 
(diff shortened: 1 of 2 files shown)


tools/admin/wskadmin
```diff
@@ -108,11 +108,11 @@ def parseArgs():
     subcmd = subparser.add_parser('whois', help='identify user from an authorization key')
     subcmd.add_argument('authkey', help='the credentials to look up')
 
-    subcmd = subparser.add_parser('block', help='block a user')
-    subcmd.add_argument('subject', help='the user to block')
+    subcmd = subparser.add_parser('block', help='block one or more users')
+    subcmd.add_argument('subject', help='comma separated list of users to block')
```
markusthoemmes: Should this be

```python
subcmd.add_argument('subjects', nargs='+', help='comma separated list of users to block')
```

That would make the following command possible:

```
$ wskadmin user block subject1 subject2 subject3
```

which is more in line with bash standards I think. it also makes the "manual" splitting unnecessary (`args.subjects` will be a list already) and it'll force the user to give at least one subject.

WDYT?

rabbah:
> A test would be pretty nice

i thought you might say that 😅

PR #3349, opened 2018-02-26
Download cli while cli_installation_mode=local too

jiangpengcheng:
While using local mode to get cli, we'd better use ansible to auto detect
the os and arch of deploy machine and download the correct cli just like
remote mode.


jiangpengcheng:
already fixed in https://github.com/apache/incubator-openwhisk/commit/347d4b72116fcde7fb2988f550220daba28c0959

PR #3351, opened 2018-02-26
Fix intermitent test failures due to non-unique entity names

mdeuser:
- leverage the `withTimestamp` method in a few more tests
- move `withTimestamp` into common module and remove copies


mdeuser:
pg4/1417 in progress

rabbah:
@mdeuser is this abandoned?

PR #3357, opened 2018-02-27
Remove usages of replace and replaceAll on the hotpath.

markusthoemmes:
replace(String, String) and replaceAll(String, String) will compile regex patterns implicitly. This creates a lot of unnecessary garbage.

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
@@ -427,9 +427,14 @@ object ContainerProxy {
    * @param suffix the container name's suffix
    * @return a unique container name
    */
-  def containerName(instance: InstanceId, prefix: String, suffix: String) =
-    s"${ContainerFactory.containerNamePrefix(instance)}_${containerCount.next()}_${prefix}_${suffix}"
-      .replaceAll("[^a-zA-Z0-9_]", "")
+  def containerName(instance: InstanceId, prefix: String, suffix: String): String = {
+    def isAllowed(c: Char): Boolean = c.isLetterOrDigit || c == '_'
+
+    val sanitizedPrefix = prefix.filter(isAllowed)
+    val sanitizedSuffix = suffix.filter(isAllowed)
+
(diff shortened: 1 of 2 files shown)


core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -429,7 +429,7 @@ object ContainerProxy {
    * @return a unique container name
    */
   def containerName(instance: InstanceId, prefix: String, suffix: String) =
-    s"wsk${instance.toInt}_${containerCount.next()}_${prefix}_${suffix}".replaceAll("[^a-zA-Z0-9_]", "")
+    s"wsk${instance.toInt}_${containerCount.next()}_${prefix.filter(_.isLetterOrDigit)}_${suffix.filter(_.isLetterOrDigit)}"
```
rabbah: dropping the `_` on the suffix is harmless.

common/scala/src/main/scala/whisk/core/entity/ActivationId.scala
```diff
@@ -42,7 +42,7 @@ import whisk.http.Messages
  */
 protected[whisk] class ActivationId private (private val id: java.util.UUID) extends AnyVal {
   def asString = toString
-  override def toString = id.toString.replaceAll("-", "")
+  override def toString = id.toString.filterNot(_ == '-')
```
sven-lange-last: For some reasons, we suspect that the `toString` and `asString` methods are run rather often. So it's probably no pre-mature optimisation to compute the filter upon construction and have a `val idString` that is returned instead of computing on the fly.

rabbah: You'll have to remove `AnyVal` and make the activation id not a value to do that.

markusthoemmes:
Superseded by #3362

markusthoemmes:
@rabbah updated to only contain removal of `replaceAll` in container naming. Not as pressing but still good to resolve I think.

markusthoemmes:
PG4 1427 🔵

PR #3362, opened 2018-02-27
Make ActivationId implementation leaner.

markusthoemmes:
ActivationId is today essentially a wrapper around `UUID`. This is not necessary though. Subsequent operations (serdes for instance) are unnecessarily costly because we keep the UUID around and `toString` + "clean" it each time.

We only need UUID semantics at creation time though. After that, we can happily continue to only process the cleaned `String` representation of the UUID since we don't care about it's semantics anymore.

- Cleans up the whole interface
- Introduce clearer naming of methods


markusthoemmes:
PG4 1428 🔵

PR #3364, opened 2018-02-27
Add retries to some entity "update" tests

mdeuser:
Fix for the following kind of automated test failure recently observed
```
system.basic.WskBasicTests > Wsk Package REST should create, update, get and list a package FAILED
    org.scalatest.exceptions.TestFailedException: "0.0.[1]" was not equal to "0.0.[2]"
```
Add some retries to allow for entity update to settle


mdeuser:
@markusthoemmes - here's the update per our discussion.  would you mind reviewing when you have the chance.  thanks!

markusthoemmes:
PG?

Issue #3356, opened 2018-02-27
Add method to delete attachment to ArtifactStore


chetanmeh:
As discussed on [mailing list][1] a method should be added to delete attachment to allow deletions in ArtifactStore implementation which store binary data in a separate store. To enable that

1. Add a method to ArtifactStore
```
  protected[core] def deleteAttachment[T](doc: DocInfo, name: String)(
    implicit transid: TransactionId): Future[Boolean]
```
2. Have this method exposed in `DocumentFactory`
3. Have the method in (2) invoked by `WhiskAction`
4. In `CouchDBStore` add a no op implementation which returns a successful future as delete is handled implicitly in `del` call for the entity

[1]: https://lists.apache.org/thread.html/b5c46924e9c9c951cb68492b1ecc101aacbccce2ab1cd74fc3bbdafa@%3Cdev.openwhisk.apache.org%3E

chetanmeh:
Currently when any entity is deleted we do not have entity instance. So this would make it tricky to implement `deleteAttachment(docInfo, name)`. Instead we can have a method like

```
protected[core] def deleteAttachments[T](doc: DocInfo)(
    implicit transid: TransactionId): Future[Boolean]
```
Here the `ArtifactStore` implementation can remove all attachments related to the given docId. This requires that
1. custom `AttachmentStore` (probably a new SPI for later) stores attachment with blob id set to "${doc.id.id}/$name"
2. Delete operation can be implemented by removing all blobs with id starting with document id

This may involve multiple calls depending on `AttachmentStore`  implementation. 

### S3

S3 does not provide a direct call to [delete by prefix][1]. Instead we need to 

1. first [list blobid][2] matching given prefix 
2. Delete then via [multiple delete support][3]

### Azure Blob Storage

It does not support performing multiple deletes in single call. Instead we would need to 

1. [List Blobs][4] matching given prefix
2. Delete then [one by one][5]

[1]: https://stackoverflow.com/questions/3178685/how-to-delete-lot-of-objects-named-with-common-prefix-from-s3-bucket
[2]: https://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html
[3]: https://docs.aws.amazon.com/AmazonS3/latest/API/multiobjectdeleteapi.html
[4]: https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs
[5]: https://docs.microsoft.com/en-us/rest/api/storageservices/delete-blob

[...]

PR #3366, opened 2018-02-28
Adds method deleteAttachments to ArtifactStore

chetanmeh:
PR which implements the proposal at #3356

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/ArtifactStore.scala
+++ b/common/scala/src/main/scala/whisk/core/database/ArtifactStore.scala
@@ -129,6 +129,11 @@ trait ArtifactStore[DocumentAbstraction] {
   protected[core] def readAttachment[T](doc: DocInfo, name: String, sink: Sink[ByteString, Future[T]])(
     implicit transid: TransactionId): Future[(ContentType, T)]
 
+  /**
+   * Deletes an attachments linked to given document
+   */
+  protected[core] def deleteAttachments[T](doc: DocInfo)(implicit transid: TransactionId): Future[Boolean]
+
   /** Shut it down. After this invocation, every other call is invalid. */
   def shutdown(): Unit
 }
--- a/common/scala/src/main/scala/whisk/core/database/ArtifactStore.scala
+++ b/common/scala/src/main/scala/whisk/core/database/ArtifactStore.scala
@@ -130,7 +130,7 @@ trait ArtifactStore[DocumentAbstraction] {
     implicit transid: TransactionId): Future[(ContentType, T)]
 
   /**
-   * Deletes an attachments linked to given document
+   * Deletes all attachments linked to given document
    */
   protected[core] def deleteAttachments[T](doc: DocInfo)(implicit transid: TransactionId): Future[Boolean]
 
--- a/common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
+++ b/common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
@@ -456,6 +456,11 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
           ErrorLevel))
   }
 
+  override protected[core] def deleteAttachments[T](doc: DocInfo)(implicit transid: TransactionId): Future[Boolean] =
+    //For CouchDB attachments lifecycle is tied to document. So if the
+    //document is deleted then associated attachments are also removed
+    Future.successful(true)
+
   override def shutdown(): Unit = {
     Await.ready(client.shutdown(), 1.minute)
   }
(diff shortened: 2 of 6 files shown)


tests/build.gradle
```diff
@@ -66,7 +66,8 @@ dependencies {
     compile 'org.scalamock:scalamock-scalatest-support_2.11:3.4.2'
     compile 'com.typesafe.akka:akka-testkit_2.11:2.4.16'
     compile 'com.typesafe.akka:akka-http-testkit_2.11:10.0.10'
-    compile 'com.github.java-json-tools:json-schema-validator:2.2.8';
+    compile 'com.github.java-json-tools:json-schema-validator:2.2.8'
+    compile "org.mockito:mockito-core:2.15.0"
```
markusthoemmes: We already have `scalamock` in the repo, can you use that?

chetanmeh: Checked scalamock but did not found any way to [spy][1] on existing object which I require in the test. 

Scalamock can mock any interface or class but looks like does not provide a way to proxy an existing object and check interactions that happen with that object

[1]: http://static.javadoc.io/org.mockito/mockito-core/2.15.0/org/mockito/Mockito.html#13

markusthoemmes: You might be right, in this case you'd need to "implement" the method of the mock.

common/scala/src/main/scala/whisk/core/database/ArtifactStore.scala
```diff
@@ -129,6 +129,11 @@ trait ArtifactStore[DocumentAbstraction] {
   protected[core] def readAttachment[T](doc: DocInfo, name: String, sink: Sink[ByteString, Future[T]])(
     implicit transid: TransactionId): Future[(ContentType, T)]
 
+  /**
+   * Deletes an attachments linked to given document
+   */
+  protected[core] def deleteAttachments[T](doc: DocInfo)(implicit transid: TransactionId): Future[Boolean]
```
markusthoemmes: Can you clarify: Create one or all attachements linked?

chetanmeh: It deletes all linked attachments. See #3356 for more details on the API part

chetanmeh:
@dubeejw Can you share your feedback on the PR

common/scala/src/main/scala/whisk/core/database/ArtifactStore.scala
rabbah: >  +   * Deletes an attachments linked to given document 

-> Deletes all attachments linked to a given document.

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -456,6 +456,11 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
           ErrorLevel))
   }
 
+  override protected[core] def deleteAttachments[T](doc: DocInfo)(implicit transid: TransactionId): Future[Boolean] =
+    //For CouchDB attachments lifecycle is tied to document. So if the
```
rabbah: I prefer a stronger warning here:

```
// NOTE: this method is not intended for standalone use for CouchDB.
// To delete attachments, it is expected that the entire document is deleted.
```

chetanmeh:
Updated PR based on review feedback

rabbah:
PG1/2672 🏃

chetanmeh:
@rabbah has the PG passed for this PR?

rabbah:
PG1/2682 is ok.

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -376,6 +376,24 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
...
+    implicit transid: TransactionId,
+    notifier: Option[CacheChangeNotification]): Future[Boolean] = {
+    Try {
+      require(db != null, "db undefined")
+      require(doc != null, "doc undefined")
+    }.map { _ =>
+      val fa = super.del(db, doc)
+      implicit val ec = db.executionContext
+      fa.flatMap { _ =>
+        super.deleteAttachments(db, doc)
```
rabbah: as a future enhancement this could be done only when there is an attachment present and this code possibly merged into the document factory instead.

chetanmeh: @rabbah Any pointers on how to determine if an attachment is present ... or you meant expose a method in `DocumentFactory` like `hasAttachement` and use that and have that overridden in `WhiskAction`

rabbah: Thinking strictly about an action, today you’d know what based on the exec type and attached details. Eventually we will store all blobs as attachments so it may be inconsequential.

PR #3368, opened 2018-02-28
Soften DockerExampleContainerTests.

markusthoemmes:
This test fails in different environments with different exceptions. This commit only aims to resolve the immediate issue. A proper RCA for test-hardening will need to be conducted.


rabbah:
Heisenbug?

markusthoemmes:
PG2 2869 🔵

PR #3371, opened 2018-02-28
Make AuthKey parsing leaner.

markusthoemmes:
Like in the case of ActivationIds, for AuthKeys we do not care if the user really gives us a UUID. For the system internal it doesn't matter.

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/AuthKey.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/AuthKey.scala
@@ -17,12 +17,8 @@
 
 package whisk.core.entity
 
-import scala.util.Try
-
-import spray.json.JsString
-import spray.json.JsValue
-import spray.json.RootJsonFormat
-import spray.json.deserializationError
+import spray.json._
+import spray.json.DefaultJsonProtocol._
 
 /**
  * Authentication key, consisting of a UUID and Secret.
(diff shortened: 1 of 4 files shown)


markusthoemmes:
PG5 125 🔵

common/scala/src/main/scala/whisk/core/entity/AuthKey.scala
```diff
@@ -31,14 +27,14 @@ import spray.json.deserializationError
...
  */
 protected[core] class AuthKey private (private val k: (UUID, Secret)) extends AnyVal {
-  def uuid = k._1
-  def key = k._2
+  def uuid: UUID = k._1
+  def key: Secret = k._2
   def revoke = new AuthKey(uuid, Secret())
-  def compact = s"$uuid:$key"
-  override def toString = uuid.toString
+  def compact: String = s"$uuid:$key"
```
rabbah: fwiw i tend to skip the signature where it's obvious (ie one liners, here .toString, or s"") but OK for uniformity.

common/scala/src/main/scala/whisk/core/entity/AuthKey.scala
```diff
@@ -64,37 +60,22 @@ protected[core] object AuthKey {
...
+   * @return AuthKey if argument is properly formatted
    * @throws IllegalArgumentException if argument is not well formed
    */
   @throws[IllegalArgumentException]
   protected[core] def apply(str: String): AuthKey = {
-    val (k, v) = split(str)
-    new AuthKey(UUID(k), Secret(v))
-  }
+    val (uuid, secret) = str.split(':').toList match {
+      case k :: v :: _ => (k, v)
```
rabbah: trim `k` and `v` (preserve existing split) or treat pedantically (new split)?

common/scala/src/main/scala/whisk/core/entity/Secret.scala
```diff
@@ -33,12 +30,12 @@ import spray.json.deserializationError
...
 protected[core] class Secret private (val key: String) extends AnyVal {
-  protected[core] def asString = toString
-  protected[entity] def toJson = JsString(toString)
-  override def toString = key
+  protected[core] def asString: String = toString
+  protected[entity] def toJson: JsString = JsString(toString)
+  override def toString: String = key
 }
 
-protected[core] object Secret extends ArgNormalizer[Secret] {
```
rabbah: should we open an issue to remove arg normalizer?

common/scala/src/main/scala/whisk/core/entity/AuthKey.scala
markusthoemmes: IntelliJ complains about no signature on public members, so I lazily fix files throughout.

common/scala/src/main/scala/whisk/core/entity/Secret.scala
markusthoemmes: I think it's okay to do it lazily/when it gets in the way.

PR #3378, opened 2018-03-02
LogStoreProvider using invokerAgent for KubernetesContainerPool

dgrove-oss:
An implementation of a forwarding LogStoreProvider for Kubernetes that uses the invokerAgent to remotely extract the user action logs and enrich them with additional per-activation metadata to prepare them for ingestion by a logging service.  Logs are never processed by the invoker; all processing is remote.

Also includes a minor refactoring to enable the log enrichment functionality to be shared between multiple log-forwarding LogStoreProviders.


dgrove-oss:
PG4 / 1440 🏃

dgrove-oss:
PG4 / 1440 🔵

PR #3379, opened 2018-03-02
Make action-time-limit's system bounds configurable.

markusthoemmes:
In the same vain the memory limits are settable, this should be decidable per deployment.


markusthoemmes:
PG2 2881 🔵

PR #3380, opened 2018-03-02
Generate build relative files local

ningyougang:
Ansible copy module works on local mode default which means copy
local files to remote node, but when execute on condition that
copy remote node's src to dest, will report Unable to find '{}' in expected paths error. So it is necessary to generate relative build
files on localhost.

### DIFF
--- a/ansible/roles/cli/tasks/deploy.yml
+++ b/ansible/roles/cli/tasks/deploy.yml
@@ -14,6 +14,7 @@
   file:
     path: "{{ openwhisk_build_dir }}/{{ openwhisk_cli.archive_name }}"
     state: directory
+  delegate_to: localhost
 
 #
 #  Why are we unarchiving into the build directory instead of directly into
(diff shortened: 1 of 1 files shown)


ningyougang:
Please refer to: https://github.com/ansible/ansible/issues/19994

ningyougang:
Build this

ningyougang:
@markusthoemmes ,can you look at this, if `deployment node and target node are in same machine`, 
below ansible task has no problem
```
- name: "Copy expanded archive to final configuration directory"
  copy:
    #  WARNING:  The trailing slash is significant, signalling to copy contents
    src: "{{ openwhisk_build_dir }}/{{ openwhisk_cli.archive_name }}/"
    dest: "{{ openwhisk_cli.nginxdir.name }}"
```
but if deployment node and target node are in different machines, will report `report Unable to find '{}' in expected paths error.`

So submit a patch to solve the problem.

ningyougang:
This problem is solved by this patch: https://github.com/apache/incubator-openwhisk/pull/3406
So closed.

PR #1966, opened 2017-03-07
update web actions documentation

rabbah:

PR #3382, opened 2018-03-02
Cleanup script for unused entities in the whisks database.

steffenrost:
Python script for cleanup that runs over all documents in the whisks database and checks if its namespace  still exists. The document will be marked for deletion or deleted if already marked for certain period of time if its namespace does not exists.

### DIFF
--- a/tools/db/cleanUpWhisks.py
+++ b/tools/db/cleanUpWhisks.py
@@ -1,0 +1,150 @@
+#!/usr/bin/env python
+"""Python script to delete whisks entries having none existent ns.
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


rabbah:
@lue-kle can you please provide a description for your pull request.

lue-kle:
Playground3 (Build #1966) run successfully on this fork/branch.

rabbah:
@lue-kle do you have an Apache CLA?

tools/db/cleanUpWhisks.py
```diff
@@ -0,0 +1,150 @@
...
+        self.data = []
+        self.maxsize = size * 2
+
+    def append(self, ns, bool):
+
+        self.index=(self.index+2)%self.maxsize
+
+        if len(self.data) < self.maxsize:
+            self.data.append(ns)
+            self.data.append(bool)
```
markusthoemmes: Python supports tuples. Should we store tuples of (String, Boolean) instead of relying on a fixed order here?

rabbah:
As a general note: It would be useful in the long run to consolidate this and scripts like it into wskadmin (and rearchitect that accordingly).

lue-kle:
@rabbah yes I have signed the CLA. I will talk to @cbickel regarding your suggestion for consolidation.

tools/db/cleanUpWhisks.py
lue-kle: sure we can use tuples, would than require two checks if the ns exists in the cache. I would suggest to do this change when adapting the script to be used also for kafka and cloudant trigger dbs.

PR #3383, opened 2018-03-02
Add the license headers

houshengbo:
Based on Apache release policy, every source file needs to contain the license header at the top. This PR adds the license header to the source files missing it.

Partially-Closes: https://github.com/apache/incubator-openwhisk-release/issues/54.


markusthoemmes:
Please add a description, which gives context on why this is needed and link to a relevant issue in the releases repo.

PR #3388, opened 2018-03-03
Update require-whisk-auth behavior to secure web action

mdeuser:
- Setting web action annotation `require-whisk-auth` with string or integer value secures the web action
  - Incoming requests must have the `X-Require-Whisk-Auth` header set to the target web action's require-whisk-auth value
  - When the annotation value matches the header value, the action is invoked
  - When the header is not present or the values do not match, a 401 is returned
- Existing handling of `require-whisk-auth` with a boolean value is unchanged
  - When `true` the subject in the authorization header is authenticated
  - When `false` no authentication is performed


core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -483,7 +483,24 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
               provide(fullyQualifiedActionName(actionName)) { fullActionName =>
                 onComplete(verifyWebAction(fullActionName, onBehalfOf.isDefined)) {
                   case Success((actionOwnerIdentity, action)) =>
-                    if (!action.annotations.getAs[Boolean]("web-custom-options").exists(identity)) {
+                    val requireWebAuthIsBool = (action.annotations.getAs[Boolean]("require-whisk-auth") != None)
```
rabbah: use `.isTruthy` on annotation instead.

csantanapr:
Just in case 💯 awesomeness

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
mdeuser: @rabbah - i think `isTruthy` would return true for a "true" boolean value, non-zero integers and non-empty strings; whereas this check is just determining if the web action is configured with a boolean (true or false) to determine if the existing `require-whisk-auth` behavior should be enforced.

markusthoemmes: Does that mean that the distinction of new vs. old behavior is determined on the type of the annotation? If it's boolean, use the old way, if it's either string or int use the new way?

Could you add a bit to the description describing old vs. new behavior please?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -483,7 +483,24 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
...
+                    val annotationRequireWebAuthIsIntOrString = ((action.annotations
+                      .getAs[String]("require-whisk-auth") != None) || (action.annotations
+                      .getAs[Int]("require-whisk-auth") != None))
+                    val annotationRequireWebAuth = (action.annotations
+                      .getAs[Int]("require-whisk-auth")
+                      .getOrElse(action.annotations.getAs[String]("require-whisk-auth").getOrElse("")))
+                      .toString
+                    val enforceWhiskAuthHdr = (!requireWebAuthIsBool && annotationRequireWebAuthIsIntOrString)
+                    val headerWhiskAuthSeq = context.headers.filter(_.lowercaseName == "x-require-whisk-auth")
```
markusthoemmes: If you want to get only one element which matches the predicate, use `find` over `filter`. It will return an `Option[Header]` which more closely reflects what you're after.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -483,7 +483,24 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
...
+                    // If the require-whisk-auth annotation is either an integer or a string, secure the web action by enforcing that
+                    // the require-whisk-auth annotation value == request header x-require-whisk-auth value
+                    if (enforceWhiskAuthHdr && (headerWhiskAuthSeq.isEmpty || headerWhiskAuthSeq(0).value != annotationRequireWebAuth)) {
```
markusthoemmes: With the `find => Option[Header]` change thise becomes obsolete, but for posterity: Sequences, Lists etc. generally have a `.headOption` method, which returns `Some[A]` if the sequence is not empty or `None` if it is.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -483,7 +483,24 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
...
                   case Success((actionOwnerIdentity, action)) =>
-                    if (!action.annotations.getAs[Boolean]("web-custom-options").exists(identity)) {
+                    val requireWebAuthIsBool = (action.annotations.getAs[Boolean]("require-whisk-auth") != None)
+                    val annotationRequireWebAuthIsIntOrString = ((action.annotations
+                      .getAs[String]("require-whisk-auth") != None) || (action.annotations
+                      .getAs[Int]("require-whisk-auth") != None))
+                    val annotationRequireWebAuth = (action.annotations
+                      .getAs[Int]("require-whisk-auth")
+                      .getOrElse(action.annotations.getAs[String]("require-whisk-auth").getOrElse("")))
+                      .toString
```
markusthoemmes: `Option` has another combinator called `orElse` which takes another `Option` if the first one was `None`. In your case this would look like this:

```scala
val annotationRequireWebAuth: Option[String] = action.annotations
  .getAs[Int]("require-whisk-auth").map(_.toString)
  .orElse(action.annotations.getAs[String]("require-whisk-auth"))
```

If you leave out the default of `""` you can also use this value to determine if its Int or String by using `isDefined`. That makes `annotationRequireWebAuthIsIntOrString` obsolete.

mdeuser: i like it :+1:

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
mdeuser: @markusthoemmes that is correct.  i've updated the description to make this more clear.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
mdeuser: yes, that works much better.  thanks!

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -483,7 +483,20 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
               provide(fullyQualifiedActionName(actionName)) { fullActionName =>
                 onComplete(verifyWebAction(fullActionName, onBehalfOf.isDefined)) {
                   case Success((actionOwnerIdentity, action)) =>
-                    if (!action.annotations.getAs[Boolean]("web-custom-options").exists(identity)) {
+                    val requireWebAuthIsBool = (action.annotations.getAs[Boolean](WhiskAction.requireWhiskAuthAnnotation) != None)
+                    val annotationRequireWebAuth = action.annotations
+                      .getAs[Int](WhiskAction.requireWhiskAuthAnnotation).map(_.toString)
+                      .orElse(action.annotations.getAs[String](WhiskAction.requireWhiskAuthAnnotation))
+                    val enforceWhiskAuthHdr = (!requireWebAuthIsBool && annotationRequireWebAuth.isDefined)
```
markusthoemmes: Please rename to `enforceWhiskAuthHdr`, took me a full 10 seconds to figure out it's supposed to mean "header" vs. High Dynamic Range.

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -313,6 +313,9 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
 
   override val cacheEnabled = true
 
+  val requireWhiskAuthAnnotation = "require-whisk-auth"
```
dubeejw: I don't think these values belong here since `WhiskAction` does not know what a Web Action is.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -483,7 +483,29 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
...
                   case Success((actionOwnerIdentity, action)) =>
-                    if (!action.annotations.getAs[Boolean]("web-custom-options").exists(identity)) {
+                    // If the require-whisk-auth annotation is either an integer or a string, secure the web action by enforcing
+                    //   require-whisk-auth annotation value == request header x-require-whisk-auth value
+                    // If the require-whisk-auth annotation is a boolean, skip the request header x-require-whisk-auth check
+                    val requireWhiskHeaderAuthenticationFailed = action.annotations
+                      .get(WhiskAction.requireWhiskAuthAnnotation)
+                      .flatMap {
+                        case JsString(authStr) => Some(authStr)
+                        case JsNumber(authNum) => Some(authNum.toInt.toString)
```
rabbah: do you need .toInt?

rabbah: do you need .toInt?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -483,7 +483,29 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
               provide(fullyQualifiedActionName(actionName)) { fullActionName =>
                 onComplete(verifyWebAction(fullActionName, onBehalfOf.isDefined)) {
                   case Success((actionOwnerIdentity, action)) =>
-                    if (!action.annotations.getAs[Boolean]("web-custom-options").exists(identity)) {
+                    // If the require-whisk-auth annotation is either an integer or a string, secure the web action by enforcing
+                    //   require-whisk-auth annotation value == request header x-require-whisk-auth value
+                    // If the require-whisk-auth annotation is a boolean, skip the request header x-require-whisk-auth check
+                    val requireWhiskHeaderAuthenticationFailed = action.annotations
```
rabbah: can you refactor the handing of the auth string to a new method?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -719,7 +741,8 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
   private def confirmExportedAction(actionLookup: Future[WhiskActionMetaData], authenticated: Boolean)(
     implicit transid: TransactionId): Future[WhiskActionMetaData] = {
     actionLookup flatMap { action =>
-      val requiresAuthenticatedUser = action.annotations.getAs[Boolean]("require-whisk-auth").exists(identity)
+      val requiresAuthenticatedUser =
+        action.annotations.getAs[Boolean](WhiskAction.requireWhiskAuthAnnotation).exists(identity)
```
rabbah: I was thinking... not sure if it works... to make the auth check a truthy, so that then you can consolidate the authentication failure. The `authenticated` bit could be computed by the earlier check of the header (or the whisk key as done now).

docs/annotations.md
```diff
@@ -48,7 +48,7 @@ and must be present and explicitly set to `true` to have an affect. The annotati
...
-* `require-whisk-auth`: This annotation protects the web action so that it is only accessible to an authenticated subject. It is important to note that the _owner_ of the web action will still incur the cost of running them in the system (i.e., the _owner_ of the action also owns the activations record).
+* `require-whisk-auth`: This annotation protects the web action so that it is only invoked by requests that provide appropriate authentication credentials. When set to a boolean value, it controls whether or not the request's Basic Authentication subject will be authenticated - a value of `true` will authenticate the subject, a value of `false` will invoke the action without any authentication. When set to an integer or a string, this value must match the request's `X-Require-Whisk-Auth` header value. In both cases, it is important to note that the _owner_ of the web action will still incur the cost of running them in the system (i.e., the _owner_ of the action also owns the activations record).
```
rabbah: can you further elaborate that the basic auth credentials would be valid WHISK API keys.

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -197,7 +200,9 @@ trait WebActionsApiBaseTests extends ControllerTestCommon with BeforeAndAfterEac
               annotations ++
                 Parameters("web-export", JsBoolean(true)) ++ {
                 if (requireAuthentication) {
-                  Parameters("require-whisk-auth", JsBoolean(true))
+                  Parameters(
+                    "require-whisk-auth",
```
rabbah: we should probably start pulling all the annotations to an Annotations singleton.

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -379,21 +386,51 @@ trait WebActionsApiBaseTests extends ControllerTestCommon with BeforeAndAfterEac
...
+                "content" -> metaPayload(
+                  m.method.name.toLowerCase,
+                  JsObject(),
+                  creds,
+                  pkgName = "proxy",
+                  headers = List(RawHeader("X-Require-Whisk-Auth", requireAuthenticationKey))))
+              response.fields("content").asJsObject.fields(webApiDirectives.namespace) shouldBe user.namespace.toJson
+            }
+          } else {
+            m(s"$testRoutePath/${path}.json") ~> Route.seal(routes(creds)) ~> check {
```
rabbah: can you add the expected header but with the wrong value as well.

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
mdeuser: since a web action is actually a "kind" of `WhiskAction, and since the annotation is set within the context of an action create/update, imho it makes sense to define all action annotation constants as part of `WhiskAction`

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
mdeuser: i just played a bit with the BigDecimal data type, and i think that `toInt` should be removed to support any json number, not just integers.

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
rabbah: I also suggested below that we should instead consolidate all the annotations into a singleton - they're scattered right now.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
mdeuser: let's discuss after the refactoring into a separate method.. @markusthoemmes wanted to be sure we minimized the deserialization of the json values.

csantanapr:
From @dubeejw 
>Can we open an issue at least to update the doc with examples of using the web action auth?

csantanapr:
PG4 1446 🏃

mdeuser:
rebased + added doc update with an example

rabbah:
I dont think you need a pg.

dubeejw:
Can't get a PG to work.

dubeejw:
PG3 1969 ⏳

mdeuser:
or pg4/1451

csantanapr:
pg4/1451 🔵

PR #3391, opened 2018-03-05
Add namespace-blacklist to invoker.

cbickel:
This commit adds a check to the invoker, if the current namespace is allowed to execute actions.
This check is already done in the controller, but if the action is already in the queue and throttled afterwards, there is no chance to get this action out of the queue.
With this commit, the invoker checks as well if the action is allowed to be executed. If it is not, it will be refused immediately. This will clean up the queue very fast.

### DIFF
--- a/ansible/files/namespace_throttlings_design_document_for_subjects_db.json
+++ b/ansible/files/namespace_throttlings_design_document_for_subjects_db.json
@@ -2,7 +2,7 @@
   "_id": "_design/namespaceThrottlings",
   "views": {
     "blockedNamespaces": {
-      "map": "function (doc) {\n  if (doc._id.indexOf(\"/limits\") >= 0) {\n    if (doc.concurrentInvocations === 0 || doc.invocationsPerMinute === 0) {\n      var namespace = doc._id.replace(\"/limits\", \"\");\n      emit(namespace, 1);\n    }\n  }\n}",
+      "map": "function (doc) {\n  if (doc._id.indexOf(\"/limits\") >= 0) {\n    if (doc.concurrentInvocations === 0 || doc.invocationsPerMinute === 0) {\n      var namespace = doc._id.replace(\"/limits\", \"\");\n      emit(namespace, 1);\n    }\n  } else if (doc.subject && doc.namespaces && doc.blocked) {\n    doc.namespaces.forEach(function(namespace) {\n      emit(namespace.name, 1);\n    });\n  }\n}",
       "reduce": "_sum"
     }
   },
--- a/ansible/files/namespace_throttlings_design_document_for_subjects_db.json
+++ b/ansible/files/namespace_throttlings_design_document_for_subjects_db.json
@@ -1,0 +1,10 @@
+{
+  "_id": "_design/namespaceThrottlings",
+  "views": {
+    "blockedNamespaces": {
+      "map": "function (doc) {\n  if (doc._id.indexOf(\"/limits\") >= 0) {\n    if (doc.concurrentInvocations === 0 || doc.invocationsPerMinute === 0) {\n      var namespace = doc._id.replace(\"/limits\", \"\");\n      emit(namespace, 1);\n    }\n  }\n}",
+      "reduce": "_sum"
+    }
+  },
+  "language": "javascript"
+}
--- a/core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
+++ b/core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
@@ -20,35 +20,26 @@ package whisk.core.invoker
 import java.nio.charset.StandardCharsets
 import java.time.Instant
 
-import scala.concurrent.Future
-import scala.concurrent.duration._
-import scala.util.Failure
-import scala.util.Success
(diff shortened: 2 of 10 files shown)


ansible/files/namespace_throttlings_design_document_for_subjects_db.json
```diff
@@ -0,0 +1,10 @@
+{
+  "_id": "_design/namespaceThrottlings",
+  "views": {
+    "blockedNamespaces": {
+      "map": "function (doc) {\n  if (doc._id.indexOf(\"/limits\") >= 0) {\n    if (doc.concurrentInvocations === 0 || doc.invocationsPerMinute === 0) {\n      var namespace = doc._id.replace(\"/limits\", \"\");\n      emit(namespace, 1);\n    }\n  }\n}",
+      "reduce": "_sum"
```
rabbah: How about also checking if the subject is blocked entirely?

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -87,6 +86,17 @@ class InvokerReactive(config: WhiskConfig, instance: InstanceId, producer: Messa
   /** Initialize needed databases */
   private val entityStore = WhiskEntityStore.datastore(config)
   private val activationStore = WhiskActivationStore.datastore(config)
+  private val authStore = WhiskAuthStore.datastore(config)
+
+  private val namespaceBlacklist = new NamespaceBlacklist(authStore)
+
+  Scheduler.scheduleWaitAtMost(5.minutes) { () =>
```
rabbah: Consider moving poll interval to a pure config deployment setting.

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -87,6 +86,17 @@ class InvokerReactive(config: WhiskConfig, instance: InstanceId, producer: Messa
...
+    logging.info(this, "running background job to update blacklist")
```
rabbah: Debug level?

mdeuser:
@cbickel - is the procedure for updating the controller's action-blacklist documented anywhere?

markusthoemmes:
@mdeuser there is no procedure, you use the same tools as today (setting the namespace limits). The view will pick those up who are throttled down to 0 and handle them accordingly.

core/invoker/src/main/scala/whisk/core/invoker/NamespaceBlacklist.scala
```diff
@@ -0,0 +1,70 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+import spray.json.DefaultJsonProtocol._
+
+/**
+ * The Namespace blacklist gets all namespaces that are throttled to 0 from the database.
+ * The caller is responsible for a periodically update of the blacklist with `refreshBlacklist`.
+ *
+ * @param authStore Subjects database with the limit-documents.
+ */
+class NamespaceBlacklist(authStore: AuthStore) {
```
mdeuser: might this functionality be also used by the controller (i.e. belong under the `common` package)?

markusthoemmes: Not that I'm aware of. The controller already checks all of these limits during its entitlement.

cbickel:
PG5#140 is OK.

rabbah:
Thanks @chetanmeh, used the gradle task you added, here is the view function fwiw:

```javascript
function (doc) {
  if (doc._id.indexOf("/limits") >= 0) {
    if (doc.concurrentInvocations === 0 || doc.invocationsPerMinute === 0) {
      var namespace = doc._id.replace("/limits", "");
      emit(namespace, 1);
    }
  } else if (doc.subject && doc.namespaces && doc.blocked) {
      emit(namespace.name, 1);
    });
  }
}

```

ansible/files/namespace_throttlings_design_document_for_subjects_db.json
```diff
@@ -0,0 +1,10 @@
+{
+  "_id": "_design/namespaceThrottlings",
+  "views": {
+    "blockedNamespaces": {
+      "map": "function (doc) {\n  if (doc._id.indexOf(\"/limits\") >= 0) {\n    if (doc.concurrentInvocations === 0 || doc.invocationsPerMinute === 0) {\n      var namespace = doc._id.replace(\"/limits\", \"\");\n      emit(namespace, 1);\n    }\n  } else if (doc.subject && doc.namespaces && doc.blocked) {\n    doc.namespaces.forEach(function(namespace) {\n      emit(namespace.name, 1);\n    });\n  }\n}",
```
rabbah: is there anyway for the index of `/limits` to be `0`? ie should this be a strictly `> 0` test?

markusthoemmes: I don't think so. Slashes are generally not allowed in namespaces so a namespace cannot possibly start with it.

The better check here would've been an `endWith` (ofcourse we'd need to emulate it through an indexOf + offset check). But we decided against that for improved readability.

Issue #3395, opened 2018-03-06
Generate ActivationId string directly without going through UUID.


markusthoemmes:
ActivationIDs are generated via UUID.randomUUID() today. Looking through the code of that you'll notice that it only generates a random hexadecimal String, delimited with dashes. We then proceed to cut out the dashes to end up at our ActivationId format.

We can straight up jump to generate the String ourselves. This even has the added benefit of adding more entropy to that String, because the UUID contains some bytes to determine its version (which are fixed bytes).

There are even libraries out there which make UUID serdes as fast as possible because it's seemingly slow and generates lots of garbage (https://github.com/jchambers/fast-uuid).

PR #3396, opened 2018-03-06
Self-generate random strings without going through a UUID.

markusthoemmes:
## Description

ActivationIDs are generated via UUID.randomUUID() today. Looking through the code of that you'll notice that it only generates a random hexadecimal String, delimited with dashes. We then proceed to cut out the dashes to end up at our ActivationId format.

We can straight up jump to generate the String ourselves. This even has the added benefit of adding more entropy to that String, because the UUID contains some bytes to determine its version (which are fixed bytes).

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3395)

- [ ] API
- [X] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

- [X] "Bug fix" (generally a non-breaking change which closes an issue).

## Checklist:

- [X] ~~I added tests to cover my changes.~~ (Not necessary as caught by existing tests)
- [X] ~~My changes require further changes to the documentation.~~

### DIFF
--- a/common/scala/src/main/scala/whisk/common/RandomString.scala
+++ b/common/scala/src/main/scala/whisk/common/RandomString.scala
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.common
+
+import java.security.SecureRandom
+
+import org.apache.commons.codec.binary.Hex
+
+object RandomString {
+  private val generator = new ThreadLocal[SecureRandom] {
+    override def initialValue() = new SecureRandom()
+  }
+
+  val lowercase: IndexedSeq[Char] = 'a' to 'z'
+  val uppercase: IndexedSeq[Char] = 'A' to 'Z'
+  val digits: IndexedSeq[Char] = '0' to '9'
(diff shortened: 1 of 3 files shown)


markusthoemmes:
Based on a discussion with @rabbah I benchmarked the code (and vastly improved it). Here are some numbers:

```scala
object Bench {
  def test(name: String, count: Int, generator: => String) = {
    println(s"running $name")
    val start = System.currentTimeMillis()
    Seq.fill(count)(generator)
    println(s"took ${System.currentTimeMillis() - start}")
  }

    val secure = new scala.util.Random(new SecureRandom())

    println("warmup")

    test("uuids", 10000, UUIDs.randomUUID().toString.filterNot(_ == '-'))
    test("randomHex", 10000, RandomString.generateHexadecimal(32))

    test("uuids", 10000000, UUIDs.randomUUID().toString.filterNot(_ == '-'))
    test("randomHex", 10000000, RandomString.generateHexadecimal(32))
  }
}
```

```
warmup
running uuids
took 142
running randomHex
took 181
running uuids
took 41861
running randomHex
took 31486
```

You can see that the `randomHex` code is indeed faster now. This is more about that code not allocating as much garbage though.

@rabbah WDYT? I'm a bit unsure if this is even worth it 😁.

rabbah:
bigger 🐟 in the noise...

common/scala/src/main/scala/whisk/common/RandomString.scala
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+package whisk.common
+
+import java.security.SecureRandom
+
+import org.apache.commons.codec.binary.Hex
+
+object RandomString {
+  private val generator = new ThreadLocal[SecureRandom] {
```
chetanmeh: May be use [ThreadLocalRandom][1]. As we set `java.security.egd=file:/dev/./urandom` using normal Random may be just fine

[1]:  https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadLocalRandom.html

markusthoemmes: AFAIK `ThreadLocalRandom` uses a pseudo-random generator vs. `SecureRandom` actually uses `java.security.egd=file:/dev/./urandom`

chetanmeh: Checked again ... ThreadLocalRandom uses pseudo-random. So makes sense to implement a custom ThreadLocal impl

rabbah:
Leaving as UUID factory based for now. Reopen if you disagree.

PR #3409, opened 2018-03-07
Remove namespace configuration; use invoker's instead

jcrossley3:
Making this configurable -- and defaulting to something that may not
be the ns in which the invoker is running -- is more trouble than it's
worth, I think.


dgrove-oss:
PG1 / 2688 🏃

PR #3416, opened 2018-03-09
Refactor and unit test PoolingRestClient

dubee:
Make it possible to unit test the PoolingRestClient, and adds some unit tests.

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [x] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
+++ b/common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
@@ -83,7 +83,7 @@ class PoolingRestClient(
   // Enqueue a request, and return a future capturing the corresponding response.
   // WARNING: make sure that if the future response is not failed, its entity
   // be drained entirely or the connection will be kept open until timeouts kick in.
-  def request0(futureRequest: Future[HttpRequest]): Future[HttpResponse] = {
+  def request(futureRequest: Future[HttpRequest]): Future[HttpResponse] = {
     futureRequest flatMap { request =>
       val promise = Promise[HttpResponse]
 
@@ -109,7 +109,7 @@ class PoolingRestClient(
 
   // Runs a request and returns either a JsObject, or a StatusCode if not 2xx.
   def requestJson[T: RootJsonReader](futureRequest: Future[HttpRequest]): Future[Either[StatusCode, T]] = {
-    request0(futureRequest) flatMap { response =>
+    request(futureRequest) flatMap { response =>
       if (response.status.isSuccess()) {
         Unmarshal(response.entity.withoutSizeLimit()).to[T].map { o =>
           Right(o)
@@ -142,21 +142,18 @@ class PoolingRestClient(
 
 object PoolingRestClient {
 
-  def mkRequest0(method: HttpMethod, uri: Uri, body: Future[MessageEntity], headers: List[HttpHeader] = List.empty)(
-    implicit ec: ExecutionContext): Future[HttpRequest] = {
+  def mkRequest(method: HttpMethod,
+                uri: Uri,
+                body: Future[MessageEntity] = Future.successful(HttpEntity.Empty),
+                headers: List[HttpHeader] = List.empty)(implicit ec: ExecutionContext): Future[HttpRequest] = {
     body.map { b =>
       HttpRequest(method, uri, headers, b)
     }
   }
 
-  def mkRequest(method: HttpMethod, uri: Uri, headers: List[HttpHeader] = List.empty)(
-    implicit ec: ExecutionContext): Future[HttpRequest] = {
-    mkRequest0(method, uri, Future.successful(HttpEntity.Empty), headers)
-  }
-
   def mkJsonRequest(method: HttpMethod, uri: Uri, body: JsValue, headers: List[HttpHeader] = List.empty)(
     implicit ec: ExecutionContext): Future[HttpRequest] = {
     val b = Marshal(body).to[MessageEntity]
-    mkRequest0(method, uri, b, headers)
+    mkRequest(method, uri, b, headers)
   }
 }
--- a/common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
+++ b/common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
@@ -17,10 +17,9 @@
 
 package whisk.http
 
-import scala.concurrent.Future
-import scala.concurrent.Promise
-import scala.util.{Failure, Success}
-import scala.util.Try
+import scala.concurrent.{Future, Promise}
+import scala.util.{Failure, Success, Try}
+import scala.concurrent.ExecutionContext
 
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
@@ -142,28 +141,21 @@ class PoolingRestClient(
 }
 
 object PoolingRestClient {
-  private implicit val system = ActorSystem()
-  private implicit val materializer = ActorMaterializer()
-  private implicit val executionContext2 = system.dispatcher
 
-  // Prepares a request with the proper headers.
-  def mkRequest0(method: HttpMethod,
-                 uri: Uri,
-                 body: Future[MessageEntity],
-                 headers: List[HttpHeader] = List.empty): Future[HttpRequest] = {
+  def mkRequest0(method: HttpMethod, uri: Uri, body: Future[MessageEntity], headers: List[HttpHeader] = List.empty)(
+    implicit ec: ExecutionContext): Future[HttpRequest] = {
     body.map { b =>
       HttpRequest(method, uri, headers, b)
     }
   }
 
-  def mkRequest(method: HttpMethod, uri: Uri, headers: List[HttpHeader] = List.empty): Future[HttpRequest] = {
+  def mkRequest(method: HttpMethod, uri: Uri, headers: List[HttpHeader] = List.empty)(
+    implicit ec: ExecutionContext): Future[HttpRequest] = {
     mkRequest0(method, uri, Future.successful(HttpEntity.Empty), headers)
   }
 
-  def mkJsonRequest(method: HttpMethod,
-                    uri: Uri,
-                    body: JsValue,
-                    headers: List[HttpHeader] = List.empty): Future[HttpRequest] = {
+  def mkJsonRequest(method: HttpMethod, uri: Uri, body: JsValue, headers: List[HttpHeader] = List.empty)(
+    implicit ec: ExecutionContext): Future[HttpRequest] = {
     val b = Marshal(body).to[MessageEntity]
     mkRequest0(method, uri, b, headers)
   }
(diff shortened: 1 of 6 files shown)


common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
```diff
@@ -55,16 +63,18 @@ class PoolingRestClient(protocol: String, host: String, port: Int, queueSize: In
...
 
   // Additional queue in case all connections are busy. Should hardly ever be
   // filled in practice but can be useful, e.g., in tests starting many
   // asynchronous requests in a very short period of time.
   private val requestQueue = Source
     .queue(queueSize, OverflowStrategy.dropNew)
-    .via(pool.mapMaterializedValue { x =>
-      poolPromise.success(x); x
-    })
+    .via(httpFlow.getOrElse(defaultHttpFlow))
```
chetanmeh: Interesting approach to make client testable!

common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
```diff
@@ -81,14 +91,14 @@ class PoolingRestClient(protocol: String, host: String, port: Int, queueSize: In
     }
   }
 
-  protected def mkRequest(method: HttpMethod, uri: Uri, headers: List[HttpHeader] = List.empty): Future[HttpRequest] = {
+  def mkRequest(method: HttpMethod, uri: Uri, headers: List[HttpHeader] = List.empty): Future[HttpRequest] = {
     mkRequest0(method, uri, Future.successful(HttpEntity.Empty), headers)
   }
```
markusthoemmes: If we make all of those public, would it make sense to pull them out in a `RestClient` object? They don't need the context of a class and are merely small plumbing helpers.

tests/src/test/scala/whisk/http/PoolingRestClientTests.scala
```diff
@@ -0,0 +1,220 @@
...
+    with ScalaFutures
+    with StreamLogging
+    with BeforeAndAfterEach {
+  implicit val ec: ExecutionContext = system.dispatcher
+  implicit val materializer: ActorMaterializer = ActorMaterializer()
+
+  var requestHttpResponse = HttpResponse(OK, entity = HttpEntity.Empty)
+  var requestUri = Uri./
+  var requestHeaders: List[HttpHeader] = List.empty
+  var requestPayload = ""
```
markusthoemmes: Can we get rid of the mutability?

I think you could generate a new `testFlow` per test, which defines the "body" of the flow for each test anew (since you likely need to test different things.

Like:

```scala
val testFlow(test: HttpRequest => Future[HttpResponse]): Flow[(HttpRequest, Promise[HttpResponse]), (Try[HttpResponse], Promise[HttpResponse]), NotUsed] =
  Flow[(HttpRequest, Promise[HttpResponse])]
    .mapAsyncUnordered(1) {
      case (request, userContext) => 
        test(request)
          .map(response => (Success(response), userContext))
          .recoverWith {
            case t => (Failure(t), userContext)
          }
    }
```

In that test method, you could define assertions as you please and also mock the outcome of the flow. WDYT?

tests/src/test/scala/whisk/http/PoolingRestClientTests.scala
```diff
@@ -0,0 +1,220 @@
...
+
+  behavior of "Pooling REST Client"
+
+  it should "error when configuration protocol is invalid" in {
+    a[IllegalArgumentException] should be thrownBy new PoolingRestClient("invalid", "host", 443, 1, Some(testFlow))
+  }
+
+  it should "get a non-200 status code when performing a request" in {
+    requestHttpResponse = HttpResponse(InternalServerError)
+    val expectedResponse = HttpResponse(InternalServerError, List.empty, HttpEntity.Empty, HttpProtocol("HTTP/1.1"))
```
markusthoemmes: Where does the HttpProtocol come from?

common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
```diff
@@ -42,7 +44,13 @@ import spray.json._
  *  on each request because it doesn't need to look up the pool corresponding
  *  to the host. It is also easier to add an extra queueing mechanism.
  */
-class PoolingRestClient(protocol: String, host: String, port: Int, queueSize: Int)(implicit system: ActorSystem) {
+class PoolingRestClient(
+  protocol: String,
+  host: String,
+  port: Int,
+  queueSize: Int,
+  httpFlow: Option[Flow[(HttpRequest, Promise[HttpResponse]), (Try[HttpResponse], Promise[HttpResponse]), Any]] = None)(
```
markusthoemmes: Have you thought about overriding `requestQueue` instead of passing in an arbitrary flow here?

dubee:
@markusthoemmes, thanks for the comments. Simplified the testFlow code. Will address other comments later.

dubee:
PG1 2154 ⏳

dubee:

markusthoemmes:
@dubee Have you thought about https://github.com/apache/incubator-openwhisk/pull/3416#discussion_r173771805?

dubee:
@markusthoemmes, I think that could be done in a different PR if we want to make requestQueue configurable. What do you think that advantages are there?

dubee:
PG3 2154 🔵

common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
```diff
@@ -162,3 +139,24 @@ class PoolingRestClient(
...
+object PoolingRestClient {
+
+  def mkRequest0(method: HttpMethod, uri: Uri, body: Future[MessageEntity], headers: List[HttpHeader] = List.empty)(
+    implicit ec: ExecutionContext): Future[HttpRequest] = {
+    body.map { b =>
+      HttpRequest(method, uri, headers, b)
+    }
+  }
+
+  def mkRequest(method: HttpMethod, uri: Uri, headers: List[HttpHeader] = List.empty)(
```
markusthoemmes: Can you fold `mkRequest0` into this with default arguments?

```scala
def mkRequest(method: HttpMethod, uri: Uri, body: Future[MessageEntity] = Future.successful(HttpEntity.Empty), headers: List[HttpHeader] = List.empty)
```

common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
```diff
@@ -131,7 +108,7 @@ class PoolingRestClient(
   }
 
   // Runs a request and returns either a JsObject, or a StatusCode if not 2xx.
-  protected def requestJson[T: RootJsonReader](futureRequest: Future[HttpRequest]): Future[Either[StatusCode, T]] = {
+  def requestJson[T: RootJsonReader](futureRequest: Future[HttpRequest]): Future[Either[StatusCode, T]] = {
```
markusthoemmes: I can't put a comment there, so leaving it here:

Would it make sense to rename `request0` to `request`? Not sure why it was `request0` in the first place and I think from an interface standpoint `request` makes sense? WDYT?

tests/src/test/scala/whisk/http/PoolingRestClientTests.scala
```diff
@@ -0,0 +1,154 @@
...
+  implicit val materializer: ActorMaterializer = ActorMaterializer()
+
+  def testFlow(httpResponse: HttpResponse = HttpResponse(), httpRequest: HttpRequest = HttpRequest())
+    : Flow[(HttpRequest, Promise[HttpResponse]), (Try[HttpResponse], Promise[HttpResponse]), NotUsed] =
+    Flow[(HttpRequest, Promise[HttpResponse])]
+      .mapAsyncUnordered(1) {
+        case (request, userContext) =>
+          request shouldBe httpRequest
+          Future.successful((Success(httpResponse), userContext))
+      }
```
markusthoemmes: What about adding a test, which makes the Future fail? Is that possible? Will the queue mechanism still be intact after such a "terminal" failure or will the whole client break down?

tests/src/test/scala/whisk/http/PoolingRestClientTests.scala
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  it should "return a status code on request failure" in {
+    val httpResponse = HttpResponse(NotFound)
+    val httpRequest = HttpRequest(entity = HttpEntity(ContentTypes.`application/json`, JsObject().compactPrint))
+    val poolingRestClient = new PoolingRestClient("https", "host", 443, 1, Some(testFlow(httpResponse, httpRequest)))
+    val request = mkJsonRequest(GET, Uri./, JsObject(), List.empty)
+
+    await(poolingRestClient.requestJson[JsObject](request)) shouldBe Left(NotFound)
+  }
```
markusthoemmes: Could you add a test on what happens if unmarshalling fails?

tests/src/test/scala/whisk/http/PoolingRestClientTests.scala
dubee: @markusthoemmes, I added a test with the Future failing. According to the resultant error message, it like the entire connection pool shutdowns when the Future is failed.

```
[ERROR] [05/02/2018 14:22:39.550] [PoolingRestClientTests-akka.actor.default-dispatcher-24] [akka://PoolingRestClientTests/system/pool-master] connection pool for PoolGateway(hcps = HostConnectionPoolSetup(host,443,ConnectionPoolSetup(ConnectionPoolSettings(128,0,5,512,1,90 seconds,ClientConnectionSettings(Some(User-Agent: akka-http/10.1.1),10 seconds,90 seconds,512,None,WebSocketSettings(<function0>,ping,Duration.Inf,<function0>),List(),ParserSettings(2048,16,64,64,8192,64,8388608,256,52428800,Strict,RFC6265,false,Full,Error,Map(If-Range -> 0, If-Modified-Since -> 0, If-Unmodified-Since -> 0, default -> 12, Content-MD5 -> 0, Date -> 0, If-Match -> 0, If-None-Match -> 0, User-Agent -> 32),false,true,<function1>,<function1>,<function2>),None,TCPTransport),New,1 second),akka.http.scaladsl.HttpsConnectionContext@4d2b3fa6,akka.event.MarkerLoggingAdapter@370a978d))) has shut down unexpectedly
```

tests/src/test/scala/whisk/http/PoolingRestClientTests.scala
dubee: Test added.

dubee:
@markusthoemmes, anything else?

tests/src/test/scala/whisk/http/PoolingRestClientTests.scala
markusthoemmes: Hm, for the test there shouldn't even be a connectionpool on the path 🤔 . I'm wondering what this error message means then.

dubee: Not sure where the error comes from. Can that be investigated separate from this PR?

rabbah:
@dubee is this pending anything other than clearing @markusthoemmes's review?

dubee:

PR #3418, opened 2018-03-09
support cleanup of cloudant/kafka trigger dbs

steffenrost:
Python script for cleanup that runs over all documents in specified whisk database and checks if its namespace still exists. The document will be marked for deletion or deleted if already marked for certain period of time if its namespace does not exists. Supported database types are whisks, cloudanttrigger and kafkatrigger.

### DIFF
--- a/tests/src/test/scala/whisk/core/database/test/CleanUpWhisksDbSkriptTests.scala
+++ b/tests/src/test/scala/whisk/core/database/test/CleanUpWhisksDbSkriptTests.scala
@@ -293,9 +293,9 @@ class CleanUpWhisksDbSkriptTests
 
     // Create document/action with random namespace
     val documents = Map(
-      ":AHA04676_dev:vision-cloudant-trigger" -> JsObject("dbname" -> JsString("openwhisk-darkvision-dev")),
-      ":AHA04676_stage:vision-cloudant-trigger" -> JsObject("dbname" -> JsString("openwhisk-darkvision-stage")),
-      ":AHA04676_prod:vision-cloudant-trigger" -> JsObject("dbname" -> JsString("openwhisk-darkvision-prod")))
+      ":dev_cluster:vision-cloudant-trigger" -> JsObject("dbname" -> JsString("openwhisk-darkvision-dev")),
+      ":stage_cluster:vision-cloudant-trigger" -> JsObject("dbname" -> JsString("openwhisk-darkvision-stage")),
+      ":prod_cluster:vision-cloudant-trigger" -> JsObject("dbname" -> JsString("openwhisk-darkvision-prod")))
 
     documents.foreach {
       case (id, document) =>
@@ -366,15 +366,15 @@ class CleanUpWhisksDbSkriptTests
 
     // Create document/action with random namespace
     val documents = Map(
-      "/AJackson@uk.ibm.com_dev/badgers" -> JsObject(
+      "/User@ibm.com_dev/badgers" -> JsObject(
         "triggerURL" -> JsString(
-          "https://xxx:xxx@openwhisk.ng.bluemix.net/api/v1/namespaces/AJackson%40uk.ibm.com_dev/triggers/badgers")),
-      "/AJackson@uk.ibm.com_dev/badgers2" -> JsObject(
+          "https://xxx:xxx@openwhisk.ng.bluemix.net/api/v1/namespaces/User%40ibm.com_dev/triggers/badgers")),
+      "/User@ibm.com_dev/badgers2" -> JsObject(
         "triggerURL" -> JsString(
-          "https://xxx:xxx@openwhisk.ng.bluemix.net/api/v1/namespaces/AJackson%40uk.ibm.com_dev/triggers/badgers2")),
-      "/AJackson@uk.ibm.com_dev/badgers3" -> JsObject(
+          "https://xxx:xxx@openwhisk.ng.bluemix.net/api/v1/namespaces/User%40ibm.com_dev/triggers/badgers2")),
+      "/User@ibm.com_dev/badgers3" -> JsObject(
         "triggerURL" -> JsString(
-          "https://xxx:xxx@openwhisk.ng.bluemix.net/api/v1/namespaces/AJackson%40uk.ibm.com_dev/triggers/badgers3")))
+          "https://xxx:xxx@openwhisk.ng.bluemix.net/api/v1/namespaces/User%40ibm.com_dev/triggers/badgers3")))
 
     documents.foreach {
       case (id, document) =>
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/whisk/core/database/test/CleanUpWhisksDbSkriptTests.scala
```diff
@@ -257,19 +266,262 @@ class CleanUpWhisksDbSkriptTests
...
+  }
+
+  it should "mark documents for deletion in cloudanttrigger if namespace does not exist" in {
+    // Create whisks db
+    val dbName = dbPrefix + "cleanup_cloudanttrigger_test_mark_for_deletion"
+    val client = createDatabase(dbName, None)
+
+    // Create document/action with random namespace
+    val documents = Map(
+      ":AHA04676_dev:vision-cloudant-trigger" -> JsObject("dbname" -> JsString("openwhisk-darkvision-dev")),
```
rabbah: please remove and use more generic names.

tests/src/test/scala/whisk/core/database/test/CleanUpWhisksDbSkriptTests.scala
```diff
@@ -257,19 +266,262 @@ class CleanUpWhisksDbSkriptTests
...
+  it should "mark documents for deletion in kafkatrigger if namespace does not exist" in {
+    // Create whisks db
+    val dbName = dbPrefix + "cleanup_kafkatrigger_test_mark_for_deletion"
+    val client = createDatabase(dbName, None)
+
+    // Create document/action with random namespace
+    val documents = Map(
+      "/AJackson@uk.ibm.com_dev/badgers" -> JsObject(
+        "triggerURL" -> JsString(
+          "https://xxx:xxx@openwhisk.ng.bluemix.net/api/v1/namespaces/AJackson%40uk.ibm.com_dev/triggers/badgers")),
```
rabbah: remove and sanitize all references to ids.

tools/db/cleanUpWhisks.py
```diff
@@ -95,56 +98,108 @@ def deleteWhisk(dbWhisks, wdoc):
...
+#
+# get delimiters for namespace retrieval by database type
+#
+def getDelimiters(whiskDatabaseType):
+
+    if whiskDatabaseType == "whisks":
+        # "_id": "whisk.system/utils",
+        return None, '/'
+    elif whiskDatabaseType == "cloudanttrigger":
+        # "_id": ":AHA04676_dev:vision-cloudant-trigger",
```
rabbah: this seems like over specialization - The script should be generic and all details like the database type below passed in as parameters.

steffenrost: @rabbah didn't get your last point, the database type is passed as parameter (`--whiskDBType`) and by that the delimiters for the namespace extract from the document id are determined.

rabbah: Right, thanks for following up. I am asking if the delimiter can be generic parameter (or the parts relayed differently) as opposed to teaching this script about the various databases for providers which are orthogonal to the openwhisk backend.

steffenrost: @rabbah ah ok, @cbickel and me discussed that point and thought it would be better to hide the internals and let the caller only specify the type of the whisk database that needs to be cleaned up, otherwise you would require knowledge about how the documents look like and what the script will done in order to obtain the namespace to check. I will discuss with @cbickel tomorrow about your point.

cbickel: The purpose of this script is, that an operator is able to cleanup the whisks-databases of his installation he is managing. He necessarily doesn't know, how the id of the documents is structured for each database.
So I think it would be valid to just pass the database name and the database type to hide these internals.
Do you agree, Rodric?

tests/src/test/scala/whisk/core/database/test/CleanUpWhisksDbSkriptTests.scala
```diff
@@ -58,26 +58,35 @@ class CleanUpWhisksDbSkriptTests
...
         }.toList
       }
 
-    println(s"marked:  $marked")
-    println(s"deleted: $deleted")
-    println(s"skipped: $skipped")
-    println(s"kept:    $kept")
+    println(s"marked for deletion: $marking")
+    println(s"already marked:      $marked")
+    println(s"deleted:             $deleting")
```
cbickel: As we already print out every single line above, can we leave these printlns away, to not make these tests too noisy?

tests/src/test/scala/whisk/core/database/test/CleanUpWhisksDbSkriptTests.scala
```diff
@@ -257,19 +266,262 @@ class CleanUpWhisksDbSkriptTests
...
+      ":prod_cluster:vision-cloudant-trigger" -> JsObject("dbname" -> JsString("openwhisk-darkvision-prod")))
+
+    documents.foreach {
+      case (id, document) =>
+        client.putDoc(id, document).futureValue
+    }
+
+    // execute script
+    val (marked, _, _, _) = runScript(dbUrl, dbName, authDBName, "cloudanttrigger")
+    println(s"marked: $marked")
```
cbickel: As the marked documents are already printed twice in `runScript`, can we remove the println here?

tests/src/test/scala/whisk/core/database/test/CleanUpWhisksDbSkriptTests.scala
steffenrost: Ok I will remove the code fragment to print out every single line (I used it only for debugging), so we are then back to the original behavior of the script.

cbickel:
@jasonpet @abaruni @csantanapr 
Could you please also have a look, to not delete valid documents in the trigger databases?
Thanks a lot in advance :)

rabbah:
@cbickel I think you’re already hinting at the concern I have with this script. Why should it assume anything about the providers, where they store their data, or the format. Each of the providers should handle cleaning up their own assets. They can reuse this script if necessary, override and extend it. The providers should be fundamentally viewed as separate microservices. I’m not convinced we should conflate this repo with dependencies for the providers. Albeit a small dependence here but if the format changes by any of the microservices we’d need to fix this script. 

Is there another example where we do things here that are related to the providers?

cbickel:
@rabbah We discussed again, and we will move the script somewhere else.

rabbah:
Thanks @cbickel for the followup - Really my issue with this method

https://github.com/apache/incubator-openwhisk/pull/3418/files#diff-212d21ddb36d91d8394a7e6de713df59R124

Where I think you can solve this by having an additional delimiter argument to the script perhaps.

mdeuser:
is this part of a broader "clean up" effort to rid the environment of all entities (i.e. actions, triggers, rules, apis, etc) associated with a deleted subject?

PR #3432, opened 2018-03-14
Introduce jmx for kafka

ningyougang:
- jmx is already introduced into controller and invoker, in the actual
  scene, we not only collect kafka producer and consumer metrics but also
  collect kafka server metrics. So it is necessary to introduce jmx for
  kakfa also.
- Fix bug:jmx support for controller/invoker

## Related issue and scope

~~- [ ] I opened an issue to propose and discuss this change (#????)~~

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

- [x] Bug fix (jmx support for invoker/controller).
- [x] Enhancement or new feature (introduce jmx support for kafka).

## Checklist:

~~- [ ] I added tests to cover my changes.~~
~~- [ ] My changes require further changes to the documentation.~~
~~- [ ] I updated the documentation where necessary.~~

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -91,6 +91,8 @@ jmx:
   rmiBasePortController: 16000
   basePortInvoker: 17000
   rmiBasePortInvoker: 18000
+  basePortKafka: 19000
+  rmiBasePortKafka: 20000
   user: "{{ jmxuser | default('jmxuser') }}"
   pass: "{{ jmxuser | default('jmxpass') }}"
   jvmCommonArgs: "-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.local.only=false -Dcom.sun.management.jmxremote.authenticate=true -Dcom.sun.management.jmxremote.password.file=/root/jmxremote.password -Dcom.sun.management.jmxremote.access.file=/root/jmxremote.access"
@@ -103,6 +105,7 @@ registry:
   confdir: "{{ config_root_dir }}/registry"
 
 kafka:
+  confdir: "{{ config_root_dir }}/kafka"
   ssl:
     client_authentication: required
     keystore:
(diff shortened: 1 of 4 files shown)


ningyougang:
How strange, in my local building env, don't export below error
```
/root/jmxremote.password\nError: Password file not readable: /root/jmxremote.password\nError: Password file not readable: 
```
I use below ansible task to make the files's permission is right
```
- name: copy jmxremote password file
  when: jmx.enabled
  template:
    src: "jmxremote.password.j2"
    dest: "{{ kafka.confdir }}/kafka{{ groups['kafkas'].index(inventory_hostname) }}/jmxremote.password"
    mode: 0600

- name: copy jmxremote access file
  when: jmx.enabled
  template:
    src: "jmxremote.access.j2"
    dest: "{{ kafka.confdir }}/kafka{{ groups['kafkas'].index(inventory_hostname) }}/jmxremote.access"
    mode: 0600
```

I don't why `travis-ci` exported the `/root/jmxremote.password\nError: Password file not readable` error

ningyougang:
Any suggestion?

mhenke1:
@ningyougang 
The reason for your problems is that the uid of the root user on the vm (where Ansible does work) and the uid of the root user inside the container (which is used at runtime) are different . 
So setting the access rights in Ansible does not work. 
For the same reason setting the access rights  of a mounted file system does not work 
when issued from the container.

We encountered the same problem with containers and invokers. 
Our solution was to write a container startup script (`open/common/scala/copyJMXFiles.sh`) 
which copies the jmx related files into the container filesystem and changes a the access rights afterwards.

ningyougang:
thanks.
regarding `controller/invoker`, when build, will copy `copyJMXFiles.sh` to controller/invoker's image.
but regarding `kafka`, because the image is not controller by us, so it seems can't copy `CopyJMXFiles.sh` to `kafka`

Any good idea to solve the problem?

ningyougang:
Closed

PR #3436, opened 2018-03-14
move cleanup logic of cloudant/kafka trigger dbs to a different location

steffenrost:
Cleanup Python script is removed at this location.


rabbah:
Why remove this?

cbickel:
@rabbah We are just moving it.

cbickel:
PG1#2716 is OK.

PR #3441, opened 2018-03-15
Add swift41 settings gradle

csantanapr:
include swift41 project in settings so it's picked up by distDocker


markusthoemmes:
Should we add this to the exclude list of travis?

csantanapr:
Ah you mean the swift41 exclude to distDocker yes I agree I missed that would give use back a minute or 2

csantanapr:
@markusthoemmes #3444

PR #3451, opened 2018-03-16
Fix several loadbalancer bugs.

cbickel:
1. `padTo` (used when generating the list of semaphores for all invokers) does not recompute the value each time it adds an element to the underlying list. That resulted in **all** invokers-state being backed by the same Semaphore and thus not doing anything properly really.
2. The stepsizes need to be calculated for the list of invokers they are used to step to. Therefore we need a seperate list of step-sizes for the managed and the blackbox invoker list.
3. The assumption should be, that we freed our resources (and thus updated the state) **before** we return the request to the user. A minor change in code order is warranted here.
4. In the overload case we need to take the index we want to schedule to out of the `healthyInvokers` list, rather than using the `random` value directly. The `healthyInvokers` list might be shorter/in a different order than the list of all underlying invokers.

Furthermore, this adds a useful metric to determine (from the loadbalancers point-of-view) how many activations are currently running in the system. That can be used to determine overall system utilization.

Co-authored-by: Markus Thömmes <markusthoemmes@me.com>

## Related issue and scope

- [ ] API
- [X] Controller
- [X] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

PG5#185 is 🔵 .

### DIFF
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
@@ -18,8 +18,8 @@
 package whisk.core.loadBalancer
 
 import java.nio.charset.StandardCharsets
-import java.util.concurrent.atomic.LongAdder
 import java.util.concurrent.ThreadLocalRandom
+import java.util.concurrent.atomic.LongAdder
 
 import akka.actor.{Actor, ActorSystem, Props}
 import akka.cluster.ClusterEvent._
@@ -28,6 +28,7 @@ import akka.event.Logging.InfoLevel
 import akka.stream.ActorMaterializer
 import org.apache.kafka.clients.producer.RecordMetadata
 import pureconfig._
+import whisk.common.LoggingMarkers._
 import whisk.common._
 import whisk.core.WhiskConfig._
 import whisk.core.connector._
@@ -70,6 +71,10 @@ class ShardingContainerPoolBalancer(config: WhiskConfig, controllerInstance: Ins
 
   /** State needed for scheduling. */
   private val schedulingState = ShardingContainerPoolBalancerState()()
+
+  actorSystem.scheduler.schedule(0.seconds, 10.seconds) {
+    MetricEmitter.emitHistogramMetric(LOADBALANCER_ACTIVATIONS_INFLIGHT(controllerInstance), totalActivations.longValue)
+  }
 
   /**
    * Monitors invoker supervision and the cluster to update the state sequentially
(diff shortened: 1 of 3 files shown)


core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -71,6 +72,10 @@ class ShardingContainerPoolBalancer(config: WhiskConfig, controllerInstance: Ins
   /** State needed for scheduling. */
   private val schedulingState = ShardingContainerPoolBalancerState()()
 
+  actorSystem.scheduler.schedule(0.seconds, 10.seconds) {
```
vvraskin: It might be better to use 60s interval on par with the statsd interval due to the way kamon ingests the metrics to statsd. It sends both histogram and gauges as timers, which forces our statsd plugin to take an average over all received values since the last push which might cause some noise in the resulting metric value, e.g. result in non-integer values after rounding.

core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -247,17 +254,17 @@ class ShardingContainerPoolBalancer(config: WhiskConfig, controllerInstance: Ins
 
     activations.remove(aid) match {
       case Some(entry) =>
+        totalActivations.decrement()
+        activationsPerNamespace.get(entry.namespaceId).foreach(_.decrement())
+        schedulingState.invokerSlots.lift(invoker.toInt).foreach(_.release())
```
rabbah: this is moved up to release the slot as early as possible?

core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -331,9 +338,10 @@ object ShardingContainerPoolBalancer extends LoadBalancerProvider {
           val healthyInvokers = invokers.filter(_.status == Healthy)
           if (healthyInvokers.nonEmpty) {
             // Choose a healthy invoker randomly
-            val random = ThreadLocalRandom.current().nextInt(healthyInvokers.size)
-            dispatched(random).forceAcquire()
-            Some(healthyInvokers(random).id)
+            val random = healthyInvokers(ThreadLocalRandom.current().nextInt(healthyInvokers.size)).id
+            dispatched(random.toInt).forceAcquire()
+            logging.warn(this, s"system is overloaded. Chosen invoker${random.toInt} by random.")
```
rabbah: Chosen invoker${random.toInt} by random. -> Chose invoker${random.toInt} by random assignment.
(or just randomly).

core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
cbickel: We took the 10 seconds interval by purpose.
This is a value that changes every few milliseconds (depending how long the actions run). By sending several values and taking an average, the value, that will be displayed in a Dashboard later is much more accurate than a single snapshot which is taken only once a minute.

core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
cbickel: Yes, we wanted to free up the slots earlier to avoid unnecessary overload situations under very high load.

cbickel:
PG5#185 is 🔵 .

rabbah:
@cbickel I don’t see you in the co-authored list. Do you need to be? I only see Markus.

cbickel:
@rabbah Do I have to be in the co-authored list? Because my name and image are at the commits anyway, because I'm the author.

Issue #3450, opened 2018-03-16
Introduce a AttachmentStore SPI


chetanmeh:
Currently `ArtifactStore` implementation based on CouchDB stores the binaries as [attachments][1] to main document. For larger binaries (say > 1MB) it would be better to store them in some binary store. Further if we are deploying a custom OW setup in cloud then it would be better to store the binaries in cloud specific binary storage like S3/Azure BlobStore/IBM Object Storage

To support such a case we should introduce a new SPI in OW `AttachmentStore`. Some key points wrt to such an spi

1. Used internally by ArtifactStore and need not be exposed to layers above.
2. A default memory based implementation along with TCK to validate the contract
3. Later implement custom implementation for S3/Azure BlobStore/IBM Object Storage which can use the newly refactored PoolingRestClient and have the TCK test run against them
4. Allow using such a stored for CouchDbRestStore also via config

As an example  `MongoDBStore` implementation uses GridFS to store attachements via [GridFSAttachmentStore][2].

Would work on a PR to discuss the SPI interface and other aspects

[1]: http://docs.couchdb.org/en/2.0.0/api/document/attachments.html
[2]: https://github.com/chetanmeh/openwhisk-mongo/blob/master/common/mongo/src/main/scala/whisk/core/database/mongo/attachment/GridFSAttachmentStore.scala

chetanmeh:
All planned work for this issue done. Closing the issue

PR #3453, opened 2018-03-16
Introduce a AttachmentStore SPI

chetanmeh:
Introduces a new SPI `AttachmentStore` with a `MemoryAttachmentStore` 

## Description

This PR proposes a new `AttachmentStore` SPI with a `MemoryAttachmentStore` as an initial implementation. As part of this PR following aspect would be covered

1. Capture the SPI contract as part of `AttachmentStoreBehaviors` test specification. This suite would need to be passed by all SPI implementations
2. Refactor the current `CouchDbRestStore` to make use of an _optional_ `AttachmentStore`. 
3. Possibly implement another store based on S3/Azure/IBM Objects etc in a separate PR

## Key design aspects

Currently the attachments are only used for `WhiskEntities`. A setup can use same `AttachmentStore` for all entities or use different stores. 

### BlobId

A possible implementation for say S3 can use same S3 bucket to store attachments for all stores. It would use following format for blobId 

    <namespace>/<docId>/<name>

* namespace - Derived from entity classType like `whiskauth`, `whiskactivation`. `AttachmentStoreProvider` can determine this based on `ClassTag`
* docId - DocumentId of the document with which the attachment is attached
* name - Name of attachment

### Revisions

`AttachmentStore` would not support revision based storage and would only work with `DocId`. To enable such a design attachments were made immutable (#3502)

### Compatibility

Refactored `CouchDBRestStore` is backward compatible. For e.g. if a system has been running so far with `CouchDBRestStore` and storing attachments in Couch and later configured with an `S3AttachmentStore` then old attachments would still be read from CouchDB and only newer attachments which were stored in S3 would be read from there.

This is made possible by encoding the attachment storage as scheme in attachmentNames stored as part of entity.

### Digest and Length

`AttachmentStore` would need to compute the `SHA-256` digest and length of the attachment as part of upload and return it as part of `attached` call

These metadata would be later used to say decide how and when to cache attachments on disk.

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3450)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


chetanmeh:
@rabbah @markusthoemmes  This is a work in progress work and shared now with the intention to gather initial feedback. Would be helpful if you can take a brief look and provide some feedback on approach taken and trait api so as to confirm if over all approach is fine or not

I would continue work on adding some more test and further refactoring

chetanmeh:
All the test (non containers ones) [passed][0] with [use of `MemoryAttachmentStore`][1]

[0]: https://travis-ci.org/chetanmeh/incubator-openwhisk/builds/355356744
[1]: https://github.com/chetanmeh/incubator-openwhisk/compare/attachment-store...chetanmeh:attachment-store-memory

common/scala/src/main/scala/whisk/core/database/AttachmentStore.scala
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Retrieves a saved attachment, streaming it into the provided Sink.
+   */
+  protected[core] def readAttachment[T](doc: DocInfo, name: String, sink: Sink[ByteString, Future[T]])(
+    implicit transid: TransactionId): Future[(ContentType, T)]
+
+  /**
+   * Deletes all attachments linked to given document
+   */
+  protected[core] def deleteAttachments(doc: DocInfo)(implicit transid: TransactionId): Future[Boolean]
```
rabbah: rename to `detach`?

chetanmeh: Can be done ... currently its named as per method introduced with apache/incubator-openwhisk#3366. `deleteAttachments` conveys the intention clearly while detach is bit cryptic i.e. it may be interpreted as  that attachment still exist just link with main document broken

chetanmeh:
>  dont follow the need for the volatile attachments.

The updates to state are performed via stream stage and hence is single threaded. However reads can happen concurrently in other threads so the reference to map is marked as `var` and `volatile`

chetanmeh:
As a next step there is now an implementation for S3 aka S3AttachementStore (see [here][1] for the diff). Once this PR is merged would create a PR for the S3 support

[1]: https://github.com/chetanmeh/incubator-openwhisk/compare/chetanmeh:attachment-store...chetanmeh:attachment-store-s3

common/scala/src/main/scala/whisk/core/database/AttachmentStore.scala
rabbah: fair.

common/scala/src/main/scala/whisk/core/database/couchdb/CouchDbAttachmentStore.scala
```diff
@@ -0,0 +1,171 @@
...
+      config.protocol,
+      config.host,
+      config.port,
+      config.username,
+      config.password,
+      config.databaseFor[D])
+  }
+}
+
+class CouchDbAttachmentStore(
```
rabbah: I didn't scrutinize this class - is it the code from above moved/refactored here as is without semantic change?

common/scala/src/main/scala/whisk/core/database/couchdb/CouchDbAttachmentStore.scala
```diff
@@ -0,0 +1,171 @@
...
+  override protected[core] def deleteAttachments(doc: DocInfo)(implicit transid: TransactionId): Future[Boolean] =
+    // NOTE: this method is not intended for standalone use for CouchDB.
+    // To delete attachments, it is expected that the entire document is deleted.
+    Future.successful(true)
+
+  override def shutdown(): Unit = {
+    Await.ready(client.shutdown(), 1.minute)
+  }
+
+  private def reportFailure[T, U](f: Future[T], onFailure: Throwable => U): Future[T] = {
```
rabbah: could you move report failure to a singleton and use it in the store above also to avoid clone/own?

common/scala/src/main/scala/whisk/core/database/memory/MemoryAttachmentStore.scala
```diff
@@ -0,0 +1,185 @@
...
+import scala.reflect.ClassTag
+
+object MemoryAttachmentStoreProvider extends AttachmentStoreProvider {
+  override def makeStore[D <: DocumentSerializer: ClassTag]()(implicit actorSystem: ActorSystem,
+                                                              logging: Logging,
+                                                              materializer: ActorMaterializer): AttachmentStore =
+    new MemoryAttachmentStore(implicitly[ClassTag[D]].runtimeClass.getSimpleName.toLowerCase)
+}
+
+class MemoryAttachmentStore(dbName: String)(implicit system: ActorSystem,
```
rabbah: Can you add a comment describing this store, what it's supposed to do, how it does it, etc.

common/scala/src/main/scala/whisk/core/database/memory/MemoryAttachmentStore.scala
```diff
@@ -0,0 +1,185 @@
...
+
+  private sealed trait Action {
+    def doc: DocInfo
+  }
+  private case class AddAction(doc: DocInfo, name: String, attachment: Attachment) extends Action
+  private case class DelAction(doc: DocInfo) extends Action
+  private case class Attachment(bytes: ByteString, contentType: ContentType)
+
+  @volatile
+  private var attachments: Map[String, Attachment] = Map()
```
rabbah: in particular, `attachments` here appears to be intended as a cache (would be nice to mention in the class docs).
I have two concerns with this Map - while you use volatile, I think you might want to consider using a TrieMap instead http://www.scala-lang.org/api/2.12.1/scala/collection/concurrent/TrieMap.html.

Second, attachments can be large. This Map has no bound on its size and can grow to OOM the controller.
Instead, maybe it's possible to tie into the cache implementation that already exists which uses soft references, and furthermore has a time to live bound and max key count (still not ideal for attachments where a size bound is preferred).

common/scala/src/main/scala/whisk/core/database/couchdb/CouchDbAttachmentStore.scala
chetanmeh: Yes just moved the existing logic as is

common/scala/src/main/scala/whisk/core/database/memory/MemoryAttachmentStore.scala
chetanmeh: Done

common/scala/src/main/scala/whisk/core/database/couchdb/CouchDbAttachmentStore.scala
chetanmeh: Moved them to `StoreUtils`

common/scala/src/main/scala/whisk/core/database/memory/MemoryAttachmentStore.scala
chetanmeh: Did not knew about `TrieMap`. Thanks for the tip!. Now the `MemoryAttachmentStore` implementation is much simpler. 

> Second, attachments can be large. This Map has no bound on its size and can grow to OOM the controller

Purpose of memory store is to facilitate testing and there also each test would have its own instance. So we can keep it simple with assumption that tests would not be storing very large binaries in it.

To make it more explicit (i.e. it is meant for test) I have moved the logic to tests module

chetanmeh:
Blocked on #3502

chetanmeh:
@rabbah This PR is now ready for review.

csantanapr:
PG5 380 looks good

csantanapr:
PG1 3010 looks good

common/scala/src/main/scala/whisk/core/database/AttachmentStore.scala
```diff
@@ -0,0 +1,70 @@
...
+  def makeStore[D <: DocumentSerializer: ClassTag]()(implicit actorSystem: ActorSystem,
+                                                     logging: Logging,
+                                                     materializer: ActorMaterializer): AttachmentStore
+}
+
+case class AttachResult(digest: String, length: Long)
+
+trait AttachmentStore {
+
+  protected[core] def scheme: String
```
rabbah: can you add a comment for `scheme`?

common/scala/src/main/scala/whisk/core/database/AttachmentStore.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  /** Execution context for futures */
+  protected[core] implicit val executionContext: ExecutionContext
+
+  /**
+   * Attaches a "file" of type `contentType` to an existing document.
+   */
+  protected[core] def attach(doc: DocId, name: String, contentType: ContentType, docStream: Source[ByteString, _])(
```
rabbah: the attachment is also streaming from the source?

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -52,7 +52,8 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
                                                                   dbPassword: String,
                                                                   dbName: String,
                                                                   useBatching: Boolean = false,
-                                                                  val inliningConfig: InliningConfig)(
+                                                                  val inliningConfig: InliningConfig,
+                                                                  val attachmentStore: Option[AttachmentStore])(
```
rabbah: attachmentStore not in scaladoc.

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -357,22 +359,58 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
...
-        Future.successful(a)
-      }
-      i1 <- put(update(d, attached))
-      i2 <- if (isInlined(uri)) { Future.successful(i1) } else {
-        attach(i1, uri.path.toString, attached.attachmentType, combinedSource(bytes, tailSource))
-      }
-    } yield (i2, attached)
+    attachmentStore match {
+      case Some(as) =>
+        val asJson = d.toDocumentRecord
```
rabbah: do yo mind factoring the code for some and none into their own methods?
a comment on what happens if no store is defined would be helpful otherwise (code is inlined).

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -421,8 +459,24 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
 
   override protected[core] def readAttachment[T](doc: DocInfo, attached: Attached, sink: Sink[ByteString, Future[T]])(
     implicit transid: TransactionId): Future[T] = {
-
     val name = attached.attachmentName
+    val attachmentUri = Uri(name)
+    attachmentUri.scheme match {
+      case AttachmentInliner.MemScheme =>
+        memorySource(attachmentUri).runWith(sink)
+      case s if s == couchScheme || attachmentUri.isRelative =>
```
rabbah: can you add examples for this case or a comment?

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
chetanmeh: It is currently not part of API but instead used by`CouchDBStoreBehaviorBase` for test purpose. Code dealing with `ArtifactStore` is not aware of `AttachmentStore` for now

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
chetanmeh: Done

common/scala/src/main/scala/whisk/core/database/AttachmentStore.scala
chetanmeh: You mean is it being stream as part of action update call directly from client? Currently not as the attachment content are read completely as for POST parameter processed and Source is created from an in memory byte array

PR #3466, opened 2018-03-20
Support alternative path for python in whiskprops

tz70s:
By default, the path of python2.7 is located at `/usr/bin/`
instead of `/usr/local/bin/`. Since folks may desire to use homebrew
installation and the path _might_ be relocated at `/usr/local/bin/`. Support path
from `/usr/bin` both is better to avoid python3 override the default
`python` without any penalty.

Signed-off-by: Tzu-Chiao Yeh <su3g4284zo6y7@gmail.com>

## Related issue and scope

Not sure the small change is needed to open an additional issue? Didn't find similar issue, though.

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/tests/src/test/scala/common/WhiskProperties.java
+++ b/tests/src/test/scala/common/WhiskProperties.java
@@ -319,9 +319,9 @@ public class WhiskProperties {
     }
 
     /**
-     * where is python 2.7?
+     * python interpreter.
      */
-    public static final String python = findPython();
+    public static final String python = "python";
 
     protected static File findFileRecursively(String dir, String needle) {
         if (dir != null) {
@@ -360,15 +360,6 @@ public class WhiskProperties {
             }
         }
         return props;
-    }
-
-    private static String findPython() {
-        File p27 = new File("/usr/local/bin/python2.7");
-        if (p27.exists()) {
-            return "/usr/local/bin/python2.7";
-        } else {
-            return "python";
-        }
     }
 
     private static int getConcurrentTestCount(String count) {


rabbah:
This change is ok but I'm wondering if we actually need this still. I a few scripts are still in python (wskdamin, wskdev) but may be python 3 compatible. The reason for python 2.7 was the previous version of the wsk CLI which wasn't python 3 ready.

tz70s:
In my experience, `wskadmin` is not python 3 compatible. For example, `httplib` is particular a python2.7 name, `http.client` in python3. 
This makes me meet some problems under #3474.

However, the `wskadmin` script currently is fined bc I'm not sure `python2.7` binary is available on all platform. 
But in the `WhiskProperties` case, it might be worth to provide the path additionally.

tz70s:
@rabbah , this helps me a lot, thanks!

tz70s:
@rabbah , I found that there's plenty of places that should not be compatible.
If we really desire to have compatibility, as you mentioned, I think I can work on 2to3 scan associated with a travis build for checking all python 2, 3 code got passed. Should we have an additional branch on here for a visible testing result?

rabbah:
+1 python 3 is a decade old :) (@cclauss this will make you happy :)

tz70s:
Additional opinion, I think we can block this pr, once all processes finish.

rabbah:
`wskadmin`, `replicateDbs`, `cleanUpActivations`, and `deleteLogsFromActivations` use the python path in whisk properties.

tz70s:
I've checked for `cleanUpActivations` and `deleteLogsFromActivations` in pending #3483 . I can't successfully tested with `replicateDbs` seems caused by some additional problems in my settings/mac. (It didn't throw any python error, though) Waiting for travis check.

Maybe some additional utilities? i.e. `scanCode` (I'm sure it's not currently compatible)

rabbah:
@cclauss first i'd like to get the code compatible, then we can update the docs to make python 3 the default.

cclauss:
@rabbah I was not talking about which is default.  I was talking about how we discover the path to the executable.

rabbah:
Ah - thanks for clarifying. I'm thinking we can just nuke that method entirely.

PR #3467, opened 2018-03-20
Remove actionControllers tests from incubator-openwhisk repo

jonpspri:
## Description

When the various action controllers were spun out into their own repositories, they took their test suites with them.  However, the original source code for the test suites was left in `incubator-openwhisk` .  That was fine until we started compiling JAR files to support testing and leaving them in local maven.  Now the JAR files including the actionController tests can create classpath conflicts with the test being compiled from the individual runtimes, challenging the validity of the tests.

This PR deletes the not-canonical versions of the tests from incubator-runtime.  There are certain utility classes that are duplicated among the runtime repos that should maybe eventually be factored further into an SDK, but for the moment simplicity in the test process is needed to ensure test validity.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [X] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


rabbah:
We will need to designate a single place as the canonical root for the common code. Maybe it can still be this repo.

jonpspri:
Maybe it can be, but should we move it into a different namespace from the runtime-specific code (maybe also out of the tests directory since it will at that point be more of a library)? That would reduce the likelihood of collisions and make it clear when a given runtime has migrated to the new paradigm (whatever that may be).

jonpspri:
Marking as WIP because apparently, the what’s in the runtime repo and what’s in the core repo was even more messy than I originally anticipated.  Currently moving in the direction of “Make the runtimes self-sufficient for `actionContainer.*` classes”, then I’ll figure out what to factor back into the common libraries.

jonpspri:
Removing the WIP tag, because I’ve resolved the namespace oversights as I’ve moved through the runtimes.  It’s necessary to go through the tests in individual runtimes and make sure all are namespaced to `package runtimes.actionContainers`.  There were a couple oversights in the `master` build chain for runtimes (e.g. Java) that should start breaking in Travis builds once this change is committed.  I’ll try to file PRs for pre-emotive strikes on those.

A transition period will be needed to remove deploy builds from the Travis files, after which the shared classes can be factored back into `incubator-openwhisk`.  Or maybe we’ll do that first.  Too many repos and moving parts — time to put some stakes in the ground.

rabbah:
> Too many repos

_blame_ @csantanapr  😅

rabbah:
@jonpspri can you clarify: where will the canonical common code be located after this patch?

csantanapr:
Common test code should live in this repo for now. 

I started refactoring the runtimes repo to not have a copy of this code and leverage the common-test jar produced by maven and installed locally.

jonpspri:
@csantanapr @rabbah 

To be precise, these are the modules of concern (all from tests/src/test/scala/actionContainers):

- ActionContainer.scala
- ActionProxyContainerTests.scala (?)
- ResourceHelpers.scala

I think there's a path to moving them into the openwhisk-common-tests library, but it will require a little extra juggling with all the other changes impact them.  Let me think about it for a bit.

jonpspri:
Naming WIP again.  I intend to find a home for at least `ActionContainer.scala` and `ResourceHelpers.scala` in the `incubator-openwhisk` repository.

jonpspri:
@csantanapr @rabbah Please review.  Moving ActionContainer and ResourceHelpers into a 'common.actionContainers' package in the test library seemed to me to be the least disruptive course toward the shared library we're all looking for.

Now each runtime will need to 'import common.actionContainers._' and remove the redundant scala from the local repo.  Thoughts?

tests/src/test/scala/actionContainers/ActionProxyContainerTests.scala
```diff
@@ -1,392 +0,0 @@
...
-  }
-
-  testNotReturningJson(codeNotReturningJson, checkResultInLogs = true)
-  testEcho(stdCodeSamples)
-  testUnicode(stdUnicodeSamples)
-  testEnv(stdEnvSamples)
-  testLargeInput(stdLargeInputSamples)
-}
-
-trait BasicActionRunnerTests extends ActionProxyContainerTestUtils {
```
rabbah: @jonpspri can you moe this trait to the common package also - this is a boilerplate to run a few standard tests as shown just above on lines 243-247.

rabbah:
actually I think tests/src/test/scala/actionContainers/ActionProxyContainerTests.scala can be made more generic and moved to the common package. The idea being that if one wants to test a specific runtime, they can specify the docker image name and some code samples, leaving the rest of the suite the same.

I can work with you seperately on making this trait the base of all the other runtime tests.

jonpspri:
Yeah, I ran across the trait in question in a couple places during my works on the runtimes.  For the moment, I'll factor it over into ActionContainers.scala, which is where the other traits used by the tests live.  I also want to break the hard dependency on `whisk.properties` that's in there.  WIll probably post in a few hours.

jonpspri:
Also tinkering with how the Docker URL is discovered, because the newer runtime build Gradles know their Docker hosts and can provide them as part of the environment.

rabbah:
@csantanapr I rebased this PR and resolved the conflict. I believe the latest changes address your concerns as well.

tests/src/test/scala/actionContainers/ActionContainer.scala
```diff
@@ -104,19 +113,25 @@ object ActionContainer {
       sys.props
         .get("docker.host")
         .orElse(sys.env.get("DOCKER_HOST"))
+        .orElse {
+          // Check if we are running on docker-machine env.
+          Option(WhiskProperties.getProperty("whisk.version.name")).filter(_.toLowerCase.contains("mac")).map {
+            case _ => s"tcp://${WhiskProperties.getMainDockerEndpoint}"
+          }
+        }
```
csantanapr: I don't like this, I don't want the runtimes test to rely on WhiskProperties to have a `whisk.properties` file present.
If user is using docker-machine the `DOCKER_HOST` env variable needs to be set and use.

chetanmeh:
With this change travis build time for unit test now reduces to 18 mins of which test run is only ~6 mins!

image

Issue #3032, opened 2017-11-30
HA controller leads to fail some tests


devbv:
Since Nginx does proxy in Round-Robin way, some tests which were designed to run sequentially in one controller are failed.

Let me explain one example, travis of latest commit(0737826c06c73a904c49977910d103d1eed79174) failed the test "Wsk Package should allow updating a package". Because the result of package update was 409, not 200

See the code below
``` 
it should "allow updating a package" in withAssetCleaner(wskprops) { (wp, assetHelper) =>
    val name = "simplepackagetoupdate"
    assetHelper.withCleaner(wsk.pkg, name) { (pkg, _) =>
      pkg.create(name, params1)
      pkg.create(name, params2, update = true)
    }
  }
```

It might runs in this way
```
1. delete pkg when withCleaner starts -> controller0 delete package if it exists
2. pkg.create(name, params1) -> controller1 put package with rev A
3. pkg.create(name, params1) -> controller0 put package with rev B
4. delete pkg when withCleaner ends -> controller1 del package with rev A
```
Package is finally stored with rev B, but controller1 try to delete it with rev A, and then CouchDB response last operation as a conflict(409)
It could be success, if 'cacheInvalidation' msg arrived between 3 & 4. But we can't guarantee that.

I found that many tests were failed by similar scenario while running Travis.

Any good idea?

[tarvis log](https://travis-ci.org/apache/incubator-openwhisk/builds/309179905?utm_source=github_status&utm_medium=notification)
[controller0_logs.log](https://github.com/apache/incubator-openwhisk/files/1516179/controller0_logs.log)
[controller1_logs.log](https://github.com/apache/incubator-openwhisk/files/1516250/controller1_logs.log)

rabbah:
Thanks @devbv for the cross link - I think having a sticky session to route CRUD operations to the same controller will help alleviate this issue vs round robin spraying the operations between controllers. cc @cbickel

Ahmed-elsayed-mahmoud:
@rabbah @devbv Does round-robin handle load balancing between controllers and requests passing to them? Could you tell me where round-robin is implemented?
Thanks!

rabbah:
It’s done at the router (nginx).

Ahmed-elsayed-mahmoud:
which class ? in controller module?

rabbah:
See https://github.com/apache/incubator-openwhisk/pull/3266 for the nginx change.

rabbah:
[...]

PR #3480, opened 2018-03-22
Retry in feed tests to avoid race condition

vvraskin:
It provides a fix for test failure on mainWhisk/8244. 
Given current round robin load balancing for controllers and cache invalidation, a race condition might occur for the mentioned test. For example, a trigger for a broken feed was deleted on one controller, but cache invalidation didn't yet make it and we already check for 404 response code. 

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

### DIFF
--- a/tests/src/test/scala/whisk/core/cli/test/WskEntitlementTests.scala
+++ b/tests/src/test/scala/whisk/core/cli/test/WskEntitlementTests.scala
@@ -31,6 +31,8 @@ import spray.json._
 import spray.json.DefaultJsonProtocol._
 import whisk.core.entity.Subject
 import whisk.core.entity.WhiskPackage
+import whisk.utils.retry
+import scala.concurrent.duration._
 
 @RunWith(classOf[JUnitRunner])
 abstract class WskEntitlementTests extends TestHelpers with WskTestHelpers with BeforeAndAfterAll {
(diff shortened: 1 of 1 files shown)


mdeuser:
@vvraskin - is it possible that a customer might also experience similar issues due the controller cache invalidation delays?  if so, i think we might need to better understand this distributed cache behavior.. document for support purposes or maybe even to customers.. ?

rabbah:
@mdeuser https://github.com/apache/incubator-openwhisk/issues/3032 the fact there are multiple controllers shouldn’t really leak out and the backend should alleviate the races. Namespace stickiness to serialize all crud operations is one approach.

mdeuser:
@vvraskin - it sounds like the fix for #3032 is the appropriate remedy for the mainWhisk/8244 test failure reported by this issue.  imho, masking the root cause in each test may not be the way to go as these will serve to validate the eventual #3032 fix.

vvraskin:
Agree that this failure should go away with the session stickiness or alike. Given the change is there, I would expect a test that covers this behaviour. 
With the system as it is now, the assumption '404 must be returned right after the trigger deletion' is not valid and I think that the fix corrects it. Otherwise the test is occasionally false positive.

mdeuser:
@vvraskin - i understand your point, but if i understand correctly, the reason the test failed was due to cross-controller cache consistency and not due to an incorrect test implementation.. ?

vvraskin:
well, the test was correct for a deployment with a single controller :)

rabbah:
I think Mark’s point is that you may observe more of these failure in general because the root cause is outlined in the referenced issue.

rabbah:
In fact see https://github.com/apache/incubator-openwhisk/issues/3021 for longer discusison and examples.

Issue #3500, opened 2018-03-28
Make attachments immutable


chetanmeh:
As part of work to enable AttachmentStore SPI(#3450) we need to make attachments immutable. More details are provided in [mail thread][1]. In brief key point is that each attachment must be give a unique name and it should be created in following way

1. Upload the attachment and have ArtifactStore return a generated name
2. Then update the document with attachmentName set to name returned in previous step
3. Then delete the old attachment after #2 completes successfully

The approach would allow an AttachmentStore to

1. Ensure that attachments are uniquely named
2. Possibly inline small attachments by encoding the attachment content in attachment name say if 1KB doc is to be inlined the attachment name is hex encoded doc content

[1]: https://lists.apache.org/thread.html/f90129cf991023c5a2c4622658d37d0eb14ecbc561f970e55bbd0035@%3Cdev.openwhisk.apache.org%3E

chetanmeh:
I tried to implement this but the way CouchDB handled attachment I am not able to find a generic approach to implement the above flow. 

With CouchDB following problems are seen

1. Attaching attachment requires the document to be present - Its not possible to upload an attachment and then link to the document. So for initial creation we need to first create the document
2. Updating document post attach causes the attachment to be lost - If a document with attachment is updated later then attachment gets lost. Looks like one needs to [pass in a stub][1] of attachment for such an update to work
    > When you want to update the document (e.g. through PUT) you must include the "stub", otherwise Couch will remove the attachment.

Given these limitation it looks like we cannot implement proposed logic in a generic way. For now I see following option

#### A - Store specific flow in DocumentFactory

Use current approach when just `CouchDBRestStore` is configured and use the proposed approch when any other `AttachmentStore` is configured

#### B - Introduce a new method in ArtifactStore

Another option would be to be add another method in ArtifactStore

```scala
  protected[database] def putWithAttachment(d: DocumentAbstraction,
                              attachedUpdater: (Attached, DocumentAbstraction) => DocumentAbstraction,
                              contentType: ContentType,
                              docStream: Source[ByteString, _],
                              oldAttached: Option[Attached])(implicit transid: TransactionId): Future[DocInfo]
```

where
* `attachedUpdater` - Method which would update the `DocumentAbstraction` with new attachment details
* `oldAttached` - Details about old attachment. We can also just accept an attachmentName string also here

Such a method can then be efficiently implemented by respective `ArtifactStore`  for e.g. with
* `CouchDBRestStore`
[...]

PR #3502, opened 2018-03-29
Immutable attachments via putAndAttach

chetanmeh:
This PR implements the approach B of #3500 by adding a new `putAndAttach` method to `ArtifactStore`

## Description

This PR adds a new `ArtifactStore#putAndAttach` method and removes the `ArtifactStore#attach` method. Following are the main changes

#### A - Refactor the flow to pass in old document

For old attachment to be removed we need access to old attachment name. To facilitate that refactoring has been done to pass in old document to `put` method. As for update we were anyway reading existing document first (to know the revision) this change does not introduce any extra calls

For `WhiskActivation` refactoring was reduced by passing `None` for old document as so activations are not updated

#### B - Cache Update flow

See inline comments in code for details

#### C - Attachment names as Uri

The attachmentName stored in document would be a Uri consisting of scheme and path. So the format would be

    <scheme>:<attachment name/path >

Where

* `scheme` - Would indicate where the attachment is stored or attachment store type. It can have values like
    * `mem` - If attachment content is inlined in name itself (TBD)
    * `couch` - If attachment is stored in CouchDB
    * `s3` - If attachment is  stored in S3
* attachment name/path - Is the actual attachment name in the storage

For e.g. with CouchDB the attachmentName Uri would be something like `couch:73d6c4cc-3670-4328-96c4-cc3670a328ac`

```javascript
"code": {
      "attachmentName": "couch:73d6c4cc-3670-4328-96c4-cc3670a328ac",
      "attachmentType": "application/java-archive"
    }
```

However the actual name of attachment would just be the path part.

```
"_attachments": {
    "73d6c4cc-3670-4328-96c4-cc3670a328ac": {
      "content_type": "application/java-archive",
      "revpos": 4,
      "digest": "md5-6PkVAZ00zzZKPoquAJj5Lw==",
      "length": 5,
      "stub": true
    }
```

The `readAttachment` flow would make use of scheme  to determine from where to read the attachment. This would help to support lazy migration of attachments to a different `AttachmentStore`

For e.g. we have an existing setup where attachments are stored in CouchDB. So attachment names would be like `jarfile` or `couch:some-uuid`. Now we can configure a `S3AttachmentStore` in this setup such that new attachments are stored in S3 while existing remain in CouchDB. So  in such a setup

* Newer attachment writes would be routed to S3
* On read side - Logic would check if scheme is `couch` then it would read it from CouchDB while if it is something else it would read from configured `AttachmentStore`

#### D - Attachement Meta

The `Attached` class now supports 2 more attributes

``` scala
  case class Attached(attachmentName: String,
                      attachmentType: ContentType,
                      length: Option[Int] = None,
                      digest: Option[String] = None)
      extends Attachment[Nothing]
```

* length - Captures the length of attachment
* digest - Digest of attachment with format "<algorithm>-<digest value>" e.g. "md5-6PkVAZ00zzZKPoquAJj5Lw=="

##### Length and Digest for CouchDB AttachmentStore

As for CouchDB attachment storage we set the attachment properties first in the document and then upload the attachment its not possible to  set these value. For non CouchDB `AttachmentStores` this would be lot simpler as there its first upload and then attach. So there these values would be populated.

So for CouchDB these values are extracted at time of deserialization by having the deserializer look into `_attachments` sub document and extract  it from there. 

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3500)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
@@ -95,9 +95,10 @@ trait DocumentFactory[W <: DocumentRevisionProvider] extends MultipleReadersSing
    * @param doc the entity to store
    * @param transid the transaction id for logging
    * @param notifier an optional callback when cache changes
+   * @param old an optional old document in case of update
    * @return Future[DocInfo] with completion to DocInfo containing the save document id and revision
    */
-  def put[Wsuper >: W](db: ArtifactStore[Wsuper], doc: W)(
+  def put[Wsuper >: W](db: ArtifactStore[Wsuper], doc: W, old: Option[W])(
     implicit transid: TransactionId,
     notifier: Option[CacheChangeNotification]): Future[DocInfo] = {
     Try {
--- a/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
@@ -141,12 +141,12 @@ trait DocumentFactory[W <: DocumentRevisionProvider] extends MultipleReadersSing
 
       val key = CacheKey(doc)
       val src = StreamConverters.fromInputStream(() => bytes)
-      val cacheDoc = postProcess map { _(doc) } getOrElse doc
 
       db.putAndAttach[W](doc, update, contentType, src, oldAttachment) flatMap {
         case (newDocInfo, attached) =>
           val newDoc = update(doc, attached).revision[W](newDocInfo.rev)
           val cacheDoc = postProcess map { _(newDoc) } getOrElse doc
+          cacheDoc.revision[W](newDocInfo.rev)
           cacheUpdate(cacheDoc, key, Future.successful(newDocInfo))
       }
 
--- a/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
@@ -196,6 +196,15 @@ trait DocumentFactory[W <: DocumentRevisionProvider] extends MultipleReadersSing
     doc: DocId,
     rev: DocRevision = DocRevision.empty,
     fromCache: Boolean = cacheEnabled)(implicit transid: TransactionId, mw: Manifest[W]): Future[W] = {
+    getWithAttachment(db, doc, rev, fromCache, None)
+  }
+
+  protected def getWithAttachment[Wsuper >: W](
+    db: ArtifactStore[Wsuper],
+    doc: DocId,
+    rev: DocRevision = DocRevision.empty,
+    fromCache: Boolean,
+    attachmentHandler: Option[(W, Attached) => W])(implicit transid: TransactionId, mw: Manifest[W]): Future[W] = {
     Try {
       require(db != null, "db undefined")
     } map {
@@ -203,7 +212,7 @@ trait DocumentFactory[W <: DocumentRevisionProvider] extends MultipleReadersSing
       implicit val ec = db.executionContext
       val key = doc.asDocInfo(rev)
       _ =>
-        cacheLookup(CacheKey(key), db.get[W](key), fromCache)
+        cacheLookup(CacheKey(key), db.get[W](key, attachmentHandler), fromCache)
     } match {
       case Success(f) => f
       case Failure(t) => Future.failed(t)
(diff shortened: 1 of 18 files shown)


common/scala/src/main/scala/whisk/core/database/DocumentFactory.scala
```diff
@@ -135,13 +139,13 @@ trait DocumentFactory[W <: DocumentRevisionProvider] extends MultipleReadersSing
...
 
-      cacheUpdate(doc, key, db.attach(docInfo, attachmentName, contentType, src) map { newDocInfo =>
-        doc.revision[W](newDocInfo.rev)
-        doc.docinfo
-      })
+      db.putAndAttach[W](doc, update, contentType, src, oldAttachment) flatMap {
+        case (newDocInfo, attached) =>
+          val newDoc = update(doc, attached).revision[W](newDocInfo.rev)
+          cacheUpdate(newDoc, key, Future.successful(newDocInfo))
+      }
```
chetanmeh: Had to change the way cache is updated. So far the cache entry was created in `DocumentFactory` at start of `put` with the document set as value and a future registered to updates its revision post db operation gets completed. This follows the similar patterns as in get and delete. This worked so far as the document which is added just needs the revision to be updated. 

With `putAndAttach` we need to update the `exec` also and thus its not possible to follow current approach as `exec` is immutable (compared to `rev` which is mutable). So now the cache is updated *after* db operation gets completed. 

From what I understand this may ok for put as current approach is more useful for lookups where intention is to ensure that first request that arrives for a certain cache key causes a future to be put into the cache which all later requests then “hook into”

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -330,18 +330,24 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
...
 
-          val newDoc = doc.copy(exec = exec.attach)
-          newDoc.revision(doc.rev)
-
           val stream = new ByteArrayInputStream(Base64.getDecoder().decode(code))
           val manifest = exec.manifest.attached.get
-
-          for (i1 <- super.put(db, newDoc);
-               i2 <- attach[A](db, newDoc.revision(i1.rev), manifest.attachmentName, manifest.attachmentType, stream))
-            yield i2
```
chetanmeh: Not sure but probably existing flow had a race condition where the `super.put(db, newDoc)` would have "published" a cache value which is yet not complete. If any thread see this value and tries to read the attachment the call would fail. So may be cache should only be updated with attach

chetanmeh:
@rabbah  This is initial implementation of approach B (#3500) to check if this looks worth pursuing. If it looks ok would work further to add more test coverage.

chetanmeh:
#### Attachment Inlining

This feature would allow more natural inlining of attachment content solely on the basis of attachment size compared to current runtime manifest based approach. Under this if any   

1. attachment length \>= `maxInlineSize` - Then its saved as proper attachment
2. attachment length < `maxInlineSize` - Then its content would be converted to URL encoded base64 string and would be stored as part of `attachmentName` within the action document. Thus it can read without going to attachment storage

This is currently implemented as a [separate branch][1]. I am not sure if this should be treated as separate PR or included in current PR (it changes `readAttachment` api signature)

Probably we can include it in current PR and thus be done with all planned changes in `ArtifactStore` API for once as of now

[1]: https://github.com/apache/incubator-openwhisk/compare/chetanmeh:immutable-attachments...chetanmeh:immutable-attachments-inline

chetanmeh:
This PR is now ready for review. Attachment inlining is done in a [separate branch][1] and can possibly be reviewed as part of separate PR.

[1]: https://github.com/apache/incubator-openwhisk/compare/chetanmeh:immutable-attachments...chetanmeh:immutable-attachments-inline

chetanmeh:
Finally its green now!

rabbah:
@dubee @csantanapr @markusthoemmes needs a pg.

markusthoemmes:
PG1 2783 🔵

csantanapr:
What’s the latest status on this PR?
@dubee Does this affect or related to your ArtifactStore work?

chetanmeh:
Rebased to current master. @rabbah and @dubee would be helpful if you can review the proposed changes now

chetanmeh:
Can this PR be reviewed and approved? 

As this is required for some other PRs #3453, Attachment In lining and S3AttachmentStore (branch exist but PR pending)

PR #3503, opened 2018-03-29
move CLI node tests to CLI repo and sync node default tests

drcariel:
## Description

sister PR https://github.com/apache/incubator-openwhisk-cli/pull/262

## Related issue and scope

- [x] I opened an issue to propose and discuss this change 
https://github.com/apache/incubator-openwhisk-cli/issues/259
see issue for more details

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


tests/src/test/scala/system/basic/WskRuleTests.scala
```diff
@@ -83,43 +81,6 @@ abstract class WskRuleTests extends TestHelpers with WskTestHelpers {
 
   behavior of "Whisk rules"
 
-  it should "preserve rule status when a rule is updated" in withAssetCleaner(wskprops) { (wp, assetHelper) =>
```
dubee: Why does this test need to be moved?

drcariel: @dubee I moved this test due to the way the response is evaluated in REST vs the CLI. This was an update that was inconsistent across the "synced" `WskRuleTests` files. When I left this here, in it's current state, I was met with this exception from the CLI repo tests:
```
system.basic.WskCliRuleTests STANDARD_OUT
    Starting test Whisk rules should preserve rule status when a rule is updated at 2018-03-29 17:25:27.372
system.basic.WskCliRuleTests > Whisk rules should preserve rule status when a rule is updated STANDARD_OUT
    Exception occurred during test execution: spray.json.JsonParser$ParsingException: Unexpected character 'o' at input index 0 (line 1, position 1), expected JSON Value:
    ok: updated rule r1to1-1522344327373
    ^
system.basic.WskCliRuleTests > Whisk rules should preserve rule status when a rule is updated FAILED
    spray.json.JsonParser$ParsingException: Unexpected character 'o' at input index 0 (line 1, position 1), expected JSON Value:
    ok: updated rule r1to1-1522344327373
    ^
```
It expects a JSON response, while the CLI returns that string.

Do you think this would be better off merged with https://github.com/apache/incubator-openwhisk-cli/pull/262/files#r178365760 ?

I can try to mash them together and work some evaluation logic to handle the 2 cases if we think that is appropriate.

tests/src/test/scala/system/basic/WskRuleTests.scala
```diff
@@ -91,7 +92,7 @@ abstract class WskRuleTests extends TestHelpers with WskTestHelpers {
     val active = Some("active".toJson)
     val inactive = Some("inactive".toJson)
     val statusPermutations =
-      Seq((triggerName, active), (triggerName, inactive), (triggerName2, active), (triggerName, inactive))
+      Seq((triggerName, active), (triggerName, inactive), (triggerName2, active), (triggerName2, inactive))
```
drcariel: I believe this was a typo that I have fixed here?

tests/src/test/scala/system/basic/WskRuleTests.scala
drcariel: I have merged the tests to use the `cli` boolean flag now that I can override in the CLI extentions

tests/src/test/scala/system/basic/WskRuleTests.scala
dubee: Nope needs to be `triggerName` otherwise it's just running the first two permutations again.

tests/src/test/scala/system/basic/WskRuleTests.scala
```diff
@@ -102,21 +103,28 @@ abstract class WskRuleTests extends TestHelpers with WskTestHelpers {
...
-              .parseJson
-              .asJsObject
-              .fields
-              .get("status") shouldBe status
-
-            wsk.rule.get(ruleName).stdout.parseJson.asJsObject.fields.get("status") shouldBe status
-          },
-          10,
-          Some(1.second))
+        if (cli) {
```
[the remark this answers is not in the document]
dubee: As part of future work, the base tests shouldn't know or care whether the the REST or CLI is being used. We should be able to provide a special function to strip out the CLI header that have the same interface that the REST tests use to prevent adding this special logic in the base tests. Maybe we can override `RunResult` for the CLI tests that has a method that removes the header.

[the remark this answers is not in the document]
dubee: As part of future work, the base tests shouldn't know or care whether the the REST or CLI is being used. We should be able to provide a special function to strip out the CLI header that have the same interface that the REST tests use to prevent adding this special logic in the base tests. Maybe we can override RunResult for the CLI tests that has a method that removes the header.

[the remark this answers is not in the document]
dubee: Also, as part of future work, we should be able to move `Wsk.scala` to incubator-openwhisk-cli.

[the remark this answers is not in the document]
drcariel: would you like to add that to the issue @dubee ? or maybe create a new one? thankfully `wsk.scala` isn't being duplicated at the moment, so it may be more appropriate as a lower priority down the road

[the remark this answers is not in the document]
mdeuser: example cli header removal method
https://github.com/apache/incubator-openwhisk/blob/f64722498bc2f5eadc356b3ae26ebb85de1bcdbf/tests/src/test/scala/common/WskTestHelpers.scala#L278

tests/src/test/scala/system/basic/WskSequenceTests.scala
```diff
@@ -56,7 +56,7 @@ abstract class WskSequenceTests extends TestHelpers with ScalatestRouteTest with
   val allowedActionDuration = 120 seconds
   val shortDuration = 10 seconds
 
-  val whiskConfig = new WhiskConfig(Map(WhiskConfig.actionSequenceMaxLimit -> null))
+  val whiskConfig: WhiskConfig
```
dubee: Remove `assert(whiskConfig.isValid)` here. That can be put into `WskRestSequenceTests.scala`.

PR #3507, opened 2018-03-30
allow use of string for controller id

tysonnorris:
Allows starting controller with string type instance id, e.g.
```/init.sh ctrl-A```

## Description

This is useful when controller instances don't get unique configs to startup, such as in mesos/kube envs. 

Downstream impact is that transaction ids take on a slightly different form as:
```transaction id that shows in logs = <instandid> + <transactionid>```
Internally, transactionid is numberic, but when emitted it is prefixed with the instance id. 

Without changing `ansible/controller/tasks/deploy.yml`, the output in logs will be: changed by having the numeric instance prefixed on the transaction number of the log message:
if controller starts with `0`, then `[INFO] [#tid_3]` -> `[INFO] [#tid_03]`

It currently does NOT impact log messages with transaction ids that are _hard coded_ e.g. invokerHealth still shows in logs as `[INFO] [#sid_121]`

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### CODE REVIEW

**Comment by rabbah** [comment]
@cbickel's pr to factor out the TID to NGINX is probably better and avoid some of the underlying issue (see #3199).

**Comment by tysonnorris** [comment]
Sounds good RE transaction id; we would still need a string controller instanceid after that

**Comment by tysonnorris** [comment]
once #3199 is done, will update this accordingly

File: common/scala/src/main/scala/whisk/core/entity/InstanceId.scala [hunk 1]
```diff
@@ -22,7 +22,20 @@ import spray.json.DefaultJsonProtocol
 case class InstanceId(val instance: Int, name: Option[String] = None) {
   def toInt: Int = instance
 }
-
+case class StringInstanceId(instance: String) {
```
tysonnorris: It might be worth considering naming this `ControllerInstanceId`? (and `InstanceId` -> `InvokerInstanceId`)
WDYT?

tysonnorris:
@cbickel @rabbah @dubee any comments on this?

tysonnorris:
Looking for a review on this so I can get it merged

rabbah:
I’ll take it.

common/scala/src/main/scala/whisk/core/entity/InstanceId.scala
rabbah: is there a reason to distinguish the two - in other words, could they both be StringInstanceId? (in which case can rename StringInstanceId to InstanceId)

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -202,7 +202,7 @@ object Controller {
 
     // if deploying multiple instances (scale out), must pass the instance number as the
     require(args.length >= 1, "controller instance required")
-    val instance = args(0).toInt
+    val instance = StringInstanceId(args(0).trim)
```
rabbah: since there's a need to validate the string (to be a legal topic name), can we `require` the regex matches on construction instead of failing later on `topicName`?

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -113,13 +113,13 @@ class InvokerReactive(
...
                      blockingInvoke: Boolean,
-                     controllerInstance: InstanceId,
+                     controllerInstance: StringInstanceId,
                      userId: UUID) => {
     implicit val transid: TransactionId = tid
 
     def send(res: Either[ActivationId, WhiskActivation], recovery: Boolean = false) = {
       val msg = CompletionMessage(transid, res, instance)
-      producer.send(s"completed${controllerInstance.toInt}", msg).andThen {
+      producer.send(topic = controllerInstance.topicName("completed"), msg).andThen {
```
rabbah: this is done on every active ack - which now requires a regex match --- we should avoid this (moving the match to the constructor would achieve that).

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -1747,7 +1747,7 @@ trait WebActionsApiBaseTests extends ControllerTestCommon with BeforeAndAfterEac
   }
 
   class TestingEntitlementProvider(config: WhiskConfig, loadBalancer: LoadBalancer)
-      extends EntitlementProvider(config, loadBalancer, InstanceId(0)) {
```
rabbah: if we can make all InstanceId essentially a StringInstanceId, then an `apply(int)` instance would provide backward compatibility so you don't have to change all the cons sites.

tests/src/test/scala/whisk/core/entity/test/StringInstanceIdTests.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  it should "strip unusable characters when creating a topic name" in {
+    val id = StringInstanceId("1234&&^^%%5$$7890._-123")
+    id.topicName("prefix") shouldBe "prefix123457890._-123"
+  }
+
+  it should "throw an exception when the topic name length exceeds the limit" in {
+    //250 chars is too long...
+    val id = StringInstanceId(
+      "1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890" +
```
rabbah: you can replace the string with: `"1" * 250`

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
rabbah: in fact if we move the validation to the constructor, then the string interpolation can remain as is, with a change from toInt to asString.

common/scala/src/main/scala/whisk/core/entity/InstanceId.scala
tysonnorris: Currently InvokerSupervision does update by index - where the index is directly mapped from the instance id using:
```
status = status.updated(instance.toInt, new InvokerHealth(instance, currentState))
```
I think this would have to change to use a `Map` instead of `IndexedSeq`

core/controller/src/main/scala/whisk/core/controller/Controller.scala
tysonnorris: Sure, it just seemed more clear to make it fail by the code that cares about it most. I can move it here though.

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
tysonnorris: Good point - I'll move it.

common/scala/src/main/scala/whisk/core/entity/InstanceId.scala
rabbah: We could preserve the duality with toInt returning the hashcode of the string?

rabbah: A map may be fine as well...

tysonnorris: If it's a map, we can get rid of the invoker `--name` arg, I think? @dgrove-oss WDYT? If the instance id is always a string for invoker, is there a need to map the `--name` arg to an int anymore?

tysonnorris: can use a `ListMap` to preserve order of invoker ids - I believe the current loadbalancer is sensitive to the preserved order of invokers during scheduling

tysonnorris: Need some ideas here - the current invoker list is an `IndexesSeq` in loadbalancer, and the stepping behavior relies on the order being preserved even while invokers are added with gaps in the ids. 
If using a `Map` etc, this will break the stepping logic, since adding elements will change the result of stepping to next key. 

This was why I didn't change it originally...

BTW using hashcode could work, except that there is a possibility of collisions. That seems negligible, but possible - we could simply reject any invoker that is added with a colliding hashcode, but this seems bad form.

I'm back to thinking this should be a separate instanceId for controller - `StringInstanceId`, or rename it to `ControllerInstanceId`

rabbah: OK - incremental works. I'll defer to @markusthoemmes on the invoker list.

rabbah: Thanks for trying @tysonnorris.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
tysonnorris: Actually, cannot do it on construction unless we restrict to some specific prefix length - some instance id values will work with certain topic prefix, depending on prefix length. We could only reserve some fixed size for prefix, say 10 chars, and on construction require that id is < than MAX_LENGTH - 10?

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
tysonnorris: Based on problems moving validation to constructor, I can instead cache the regex+validation result instead?

core/controller/src/main/scala/whisk/core/controller/Controller.scala
rabbah: that's reasonable to me - i think we can be prescriptive about the string ids (regex, length). 
would that be adequate for your purpose?

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
rabbah: how strict do we want this validation to be? a `require` can be elided at compile time.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
tysonnorris: Regex validation? may be problematic. But max length < 230 validation would be fine. (currently the regex doesn't validate, per se, just replaces invalid-for-kafka chars)

common/scala/src/main/scala/whisk/core/entity/InstanceId.scala
dgrove-oss: Sorry, late to the party.

I agree we still need both --name and --id independent of whether they are of type String or Int unless we want to make some deeper changes in the loadbalancer.  Might be nice to try to simplify at some point, but I don't think it is worth holding up this PR to accomplish it.

markusthoemmes:
Existential question: Should we have `ControllerInstanceId` and `InvokerInstanceId`? I think with landing the TransactionId header support and the cluster bootstrap support, the controller id can indeed be just a `String`.

However, while we are on the current loadbalancer, the `InvokerInstanceId` will need to be an `Int`.

I think, having a distinct type for each of them will also greatly improve code readability, because then you know from the method's signature which instance we have at hand.

Thoughts?

tysonnorris:
classes renamed, let me know if you have other comments?

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package whisk.core.entity
+
+import spray.json.DefaultJsonProtocol
+
+case class InvokerInstanceId(val instance: Int, name: Option[String] = None) {
```
rabbah: hmm this file contains both invoer and controller ids... would leave it as intance.id or break it up into two files as the better approach?

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
```diff
@@ -0,0 +1,42 @@
...
+  def toInt: Int = instance
+}
+
+case class ControllerInstanceId(private val instance: String) {
```
rabbah: if we don't match on the case class anywhere, might as well make this as regular class which will actually hide  `instance`.

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
```diff
@@ -0,0 +1,42 @@
...
+  //keep instance private, since we will replace illegal chars for kafka usage
+  //for kafka topic legal chars see https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/internals/Topic.java#L29
+  val ILLEGAL_CHARS = "[^a-zA-Z0-9._-]".r
```
rabbah: does this need to be public?

rabbah: also should probably move to singleton.

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
```diff
@@ -0,0 +1,42 @@
...
+  //validate once on construction - reserve 20 (arbitrary) chars for prefix to be prepended to create topic names
+  private val MAX_NAME_LENGTH = 249 - 20
```
rabbah: this too can be in the singleton.

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
```diff
@@ -0,0 +1,42 @@
...
+  val replaced = ILLEGAL_CHARS.replaceAllIn(instance, "")
+  require(replaced.length <= MAX_NAME_LENGTH, s"topic name can be at most ${MAX_NAME_LENGTH} ${replaced}")
```
rabbah: don't need `{}` in string interpenetration here.

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -202,7 +202,7 @@ object Controller {
 
     // if deploying multiple instances (scale out), must pass the instance number as the
     require(args.length >= 1, "controller instance required")
-    val instance = args(0).toInt
+    val instance = ControllerInstanceId(args(0).trim)
```
rabbah: i think trim is not necessary since you will replace space it in the constructor.

tests/src/test/scala/whisk/core/entity/test/ControllerInstanceIdTests.scala
```diff
@@ -0,0 +1,44 @@
...
+import org.scalatest.junit.JUnitRunner
+import whisk.core.entity.ControllerInstanceId
+
+@RunWith(classOf[JUnitRunner])
+class ControllerInstanceIdTests extends FlatSpec with Matchers {
+
+  behavior of "StringInstanceId"
+
+  it should "strip unusable characters when creating a topic name" in {
+    val id = ControllerInstanceId("1234&&^^%%5$$7890._-123")
```
rabbah: add trailing space for `trim` test.

rabbah:
should get a PG.

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
markusthoemmes: Why do we need the `name: Option[String]` here?

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
markusthoemmes: I tend to disagree, case classes provide equals and everything.

Can we name `instance` to `asString`? Then you can make it public and leave it as is.

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  override def toString() = replaced
```
markusthoemmes: Please don't use `toString` for places where you programmatically need a value to be a `String`. It should be used for logging only.

We have a convention of using `asString` for places where you really need a string in code.

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
rabbah: Also instance is guarded because it should not be used as it hasn’t been sanitized. The intent is to only expose asString. An alternate is to move the require and cleaning to “replaced” into a singleton constructor method so that it’s not possible to retrieve the original string.

markusthoemmes: Should it be a [value class](https://docs.scala-lang.org/overviews/core/value-classes.html) then? Would give us the benefit of 0 allocation.

Code would translate to:

```scala
class ControllerInstanceId(val asString: String) extends AnyVal
```

I'm wondering if we should fail out early if the controller name containers bad characters vs. implicitly removing them. Any bad characters that kube/mesos add by default?

dgrove-oss: Don't think anything kube adds any bad characters.  The argument passed in on the command line as the controller name will be a DNS-legal hostname [a-z0-9.].  Invoker ids will either by the ipaddress of the workernode [0-9.] or a DNS legal hostname like invoker0 [a-z0-9.].

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
tysonnorris: It is used for container name prefixing, when specified

core/controller/src/main/scala/whisk/core/controller/Controller.scala
tysonnorris: I would typically leave this in place, since I think it is good to always trim input strings. But I can remove it since it is technically true that it won't make a difference here.

tests/src/test/scala/whisk/core/entity/test/ControllerInstanceIdTests.scala
tysonnorris: will add some additional spaces too

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
tysonnorris: ok, changing the refs to use `asString`. Note that the log markers are also changing to use this, not sure if that is a problem, but I think it is appropriate.

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
markusthoemmes: My take was to leave the sanitization out completely. Was wondering if we should just leave it fail on topic creation and tell the operator to choose a better name then?

rabbah: How about this resolution: https://github.com/apache/incubator-openwhisk/commit/3b9e52f1d82ef7159ba51b3d83fcc161afee129b

Simplifies the name checking as @markusthoemmes suggested (no sanitization) and still checks the controller instance consists of valid characters.

tysonnorris:
LGTM 👍

tysonnorris:
@markusthoemmes any concerns here?

rabbah:
🤦‍♂️ i thought we merged this already.

tests/src/test/scala/whisk/core/entity/test/ControllerInstanceIdTests.scala
```diff
@@ -15,14 +15,32 @@
...
+import org.scalatest.junit.JUnitRunner
+import whisk.core.entity.ControllerInstanceId
 
-case class InstanceId(val instance: Int, name: Option[String] = None) {
-  def toInt: Int = instance
-}
+@RunWith(classOf[JUnitRunner])
+class ControllerInstanceIdTests extends FlatSpec with Matchers {
+
+  behavior of "StringInstanceId"
```
markusthoemmes: Should be `behavior of "ControllerInstanceId"`

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
```diff
@@ -0,0 +1,47 @@
...
+import whisk.core.entity.ControllerInstanceId.MAX_NAME_LENGTH
+
+case class InvokerInstanceId(val instance: Int, name: Option[String] = None) {
+  def toInt: Int = instance
+}
+
+case class ControllerInstanceId(val asString: String) {
+  require(
+    asString.length <= MAX_NAME_LENGTH && asString.matches(LEGAL_CHARS),
+    "Controller instance id contains invalid characters")
```
markusthoemmes: I still dislike the regex matching here, IIRC the ControllerInstanceId is passed via an ActivationMessage, so this is checked over and over again in serdes for no real value at all. On top of that, it's a concern of kafka, not of the instanceId per-se.

common/scala/src/main/scala/whisk/core/entity/InvokerInstanceId.scala
```diff
@@ -0,0 +1,47 @@
+/*
```
markusthoemmes: This file should be called `InstanceId` since it includes multiple implementations.

markusthoemmes:
In a quick discussion with @rabbah, he made the very valid argument that my solution would only check for InstanceIds generated by the `main` method, but you'd be free to create other InstanceIds as you please.

Something to chew on (but not for this pull-request): Instead of passing the ControllerInstanceId down to the invoker, we should rather somewhere convert it to `KafkaTopic` and pass that down as `resultTopic`, because that's what the invoker really cares about, it doesn't care it's a controller (I believe).

Something for a later PR, I'm fine with this as-is! 🎉

chetanmeh:
@tysonnorris Should we also make controller id stable with an approach similar to one taken in Invoker based on zookeeper? As currently it leads to creation of many topics if the id is based on say ip

chetanmeh:
Opened #3858 to track this

PR #3515, opened 2018-04-02
Introduce separate Akka dispatchers for CouchDB and Kafka Clients (#2954)

bwmcadams:
- Rest of system continues to use the default system dispatcher

Refs #2954

## Description

This reopens the previous  #2956 PR; fork-join-executor wasn't the right choice previously and it is using Thread Pool dispatchers now, which are the correct way to handle this kind of IO separation with Akka. Defaults are set to what should be a reasonable starting point, especially given that couch and kafka now have their own separate execution pools from the "Default"/main system.

 Kafka and CouchDB clients each get their own dispatchers
  * This should prevent all of the akka based stuff from competing constantly for threads
  * Kafka and Couch can be tuned now based on individual needs
  * Using thread-pool dispatchers, which are best suited for these processes (default is fork-join)
- Dispatchers are defined in common `reference.conf`, so config can be overridden by dependents

Currently, OpenWhisk is sharing a single fork-join-executor across the board for the entire `ActorSystem`, which is less than ideal especially when dealing with potentially blocking IO. 

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#2954 )

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

- [x] Bug fix (generally a non-breaking change which closes an issue). (sort of a bug fix)

## Checklist:

- [ ] I added tests to cover my changes. (not easily testable in unit/integration)

### DIFF
--- a/common/scala/src/main/resources/reference.conf
+++ b/common/scala/src/main/resources/reference.conf
@@ -8,3 +8,56 @@ whisk.spi {
   LogStoreProvider = whisk.core.containerpool.logging.DockerToActivationLogStoreProvider
   LoadBalancerProvider = whisk.core.loadBalancer.ShardingContainerPoolBalancer
 }
+
+dispatchers {
+  # Custom dispatcher for CouchDB Client. Tune as needed.
+  couch-dispatcher {
+    type = Dispatcher
+    executor = "thread-pool-executor"
+
+    # Underlying thread pool implementation is java.util.concurrent.ThreadPoolExecutor
+    thread-pool-executor {
+      # Min number of threads to cap factor-based corePoolSize number to
+      core-pool-size-min = 2
+
+      # The core-pool-size-factor is used to determine corePoolSize of the
+      # ThreadPoolExecutor using the following formula:
+      # ceil(available processors * factor).
+      # Resulting size is then bounded by the core-pool-size-min and
+      # core-pool-size-max values.
+      core-pool-size-factor = 2.0
+
+      # Max number of threads to cap factor-based corePoolSize number to
+      core-pool-size-max = 32
+    }
+    # Throughput defines the number of messages that are processed in a batch
+    # before the thread is returned to the pool. Set to 1 for as fair as possible.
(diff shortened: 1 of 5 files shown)


rabbah:
@bwmcadams @chetanmeh this might be of use for other datastores so I wonder if we should rename the couch configuration to data store.

chetanmeh:
> this might be of use for other datastores so I wonder if we should rename the couch configuration to data store.

+1

bwmcadams:
Makes sense to me. I know it occurred to me while I was putting this
together, and tunable dispatchers for each component like this is a good
thing.

chetanmeh:
@bwmcadams  #3779 introduces a new `S3AttachmentStore` where I was thinking to use the dedicated dispatcher. Do you plan to rename the dispatcher from `couch-dispatcher` to `db-dispatcher`?

bwmcadams:
@chetanmeh good question. I think the quandary is whether there's a single shared `db-dispatcher` or a specific one per database implementation. My preference/instinct would be to have one per database implementation. The way a particular database driver is implemented can vary vastly from one to the other in terms of what the best dispatcher tuning is for it. Sharing a single dispatcher may not be ideal.

chetanmeh:
@bwmcadams There are 2 aspects

1. Have a dedicated separate dispatcher for db operations
2. Dispatcher being tuned per DB driver implementation

Currently we have few storage impls which use Akka HTTP stack like CouchDB, S3 and ElasticSearch. So for them I think it would be better to have a generic `db-dispatcher` which is used by default. If any setup needs specific tuning then config can still be overridden for that deployment and later we can see if it makes sense to use a dedicated dispatcher for that kind of db.

So to start with lets have a generic `db-dispatcher` and later based on learning we can define specific one. This would allow us to at least meet 1 and still enabling option for 2.

Thoughts?

PR #3546, opened 2018-04-13
Fix vagrant not use runc

csantanapr:
- minimum build (only builds nodejs)
- do not use runc by default (readme instruction on how to get runc working)
- build cli from source
- cleaned up the readme a bit

Travis doesn't support vagrant, I tried linux and osx.


chetanmeh:
May be we can pin Docker version to one where runc works. As while working on Mac if I need to  try/test out any runc related aspect I rely on Vagrant image. So if it works with some older version we can pin Docker to  that and keep runc enabled

csantanapr:
>May be we can pin Docker version to one where runc works. As while working on Mac if I need to try/test out any runc related aspect I rely on Vagrant image. So if it works with some older version we can pin Docker to that and keep runc enabled

The problem vagrant is using ubuntu16 and installs latest stable, for example 18.03.0-ce, so the invoker image would need to be rebuild with the matching version.

If you are trying and testing runc related things, then is better to experiment with the latest stable release.

I added docs for this special case, for an user to experiment with runc.

The version we recommend and we currently test in Travis and have it pinned in the invoker is an older version 1.12

csantanapr:
I was able to figure out how to increase the size of the disk with a vagrant plugin.

@mdeuser check if the plugin works on windows.

csantanapr:
@mdeuser can you test this branch one more time.

I remove the disksize plugin since you reported problems running on Windows.

I just a got a clean deploy
```
    default: Swagger UI URL: https://192.168.33.16/api/v1/docs/index.html?url=/api/v1/api-docs
    default: ++ echo 'Swagger UI URL: https://192.168.33.16/api/v1/docs/index.html?url=/api/v1/api-docs'
    default: ++ wsk action invoke /whisk.system/utils/echo -p message hello --result
    default: {
    default:     "message": "hello"
    default: }
    default: +++ date
    default: ++ echo 'Wed Apr 18 01:21:10 UTC 2018: build-deploy-end'
```
skipping the runtimes I was able to not run out of disk space
```
vagrant@ubuntu-xenial:~$ df -h
Filesystem              Size  Used Avail Use% Mounted on
udev                    2.0G     0  2.0G   0% /dev
tmpfs                   396M  6.3M  389M   2% /run
/dev/sda1               9.7G  5.4G  4.3G  56% /
tmpfs                   2.0G     0  2.0G   0% /dev/shm
tmpfs                   5.0M     0  5.0M   0% /run/lock
tmpfs                   2.0G     0  2.0G   0% /sys/fs/cgroup
tmpfs                   396M     0  396M   0% /run/user/1000
vagrant                 466G  412G   55G  89% /vagrant
home_vagrant_openwhisk  466G  412G   55G  89% /home/vagrant/openwhisk
```

chetanmeh:
@csantanapr Are the steps to change box size similar to one posted by @neerajmangal [here][1]. In that case can we at least leave the config in `Vagrantfile` and just document steps on how to increase the size

[1]: https://github.com/apache/incubator-openwhisk/issues/3533#issuecomment-381154231

csantanapr:
@chetanmeh not the same procedure 

At least in OSX I was able to use a vagrant plugin to which requires one line in Vagrant :-) tonset new size of disk like 50GB
But this didn’t work on Windows :-(  will need to dig with @mdeuser on why it doesn’t work

Another option is to create a second disk and configure docker to use it for images, but I have not figure out how easy to do that. 

My intention is to merge this PR to unblock folks and deal with building all images later if required.

tools/vagrant/README.md
```diff
@@ -176,46 +186,46 @@ cd ${OPENWHISK_HOME}
 ./gradlew distDocker
 ```
 
-## Safe Re-deploy (after VM restart)
+## Using docker-runc
+Only for experimental use:
+To use docker-runc the docker-runc CLI use in the invoker needs to match the version on the docker engine host.
```
mdeuser: "To use docker-runc`,` the docker-runc CLI `used` ~in~ `by` the invoker"

tools/vagrant/README.md
```diff
@@ -156,6 +158,14 @@ vagrant ssh
 wsk action invoke /whisk.system/utils/echo -p message hello --result
 ```
 
+## Other Runntimes
+The default vagrant deploy only deploys nodejs:6 runtime kind, as the image runs out of space if all runtimes are built.
```
mdeuser: "runtime kind ~, as~`because` the image runs"

tools/vagrant/README.md
```diff
@@ -98,8 +98,10 @@ machine. The IP address of the virtual machine accessible from outside is
 address will conflict, use `vagrant suspend` before starting another VM with the
 same IP address.
 
-The CLI is available in `../../bin`. There you will find binaries specific to
-various operating systems and architectures (e.g. `../../bin/mac/amd64/wsk`).
+The CLI is available in `../../bin`.
+The CLI in `../../bin/wsk` would be for Linux amd64.
```
mdeuser: "The CLI ~in~ `../../bin/wsk` ~would be~ is"

tools/vagrant/README.md
```diff
@@ -176,46 +186,46 @@ cd ${OPENWHISK_HOME}
 ./gradlew distDocker
 ```
 
-## Safe Re-deploy (after VM restart)
+## Using docker-runc
+Only for experimental use:
+To use docker-runc the docker-runc CLI use in the invoker needs to match the version on the docker engine host.
+Get the version of the docker engine like the following:
```
mdeuser: is this to find the version of the `docker-runc` CLI used for deploying openwhisk docker images or the one used by the invoker?  i'm thinking this is the former.. probably should say so for clarity

tools/vagrant/README.md
```diff
@@ -176,46 +186,46 @@ cd ${OPENWHISK_HOME}
...
+```
+$ docker version | grep Version
+Version:	18.03.0-ce
+```
+Update the Invoker Dockerfile to download a newer version of the docker-cli
```
mdeuser: i think this sentence is implying 
1. compare the docker-runc version obtained on the local system against the docker-runc configured for the invoker
2. if the versions are different, only then do you need to update the invoker dockerfile to point to the matching docker download

tools/vagrant/README.md
```diff
@@ -176,46 +186,46 @@ cd ${OPENWHISK_HOME}
...
-the OpenWhisk deployment. You can do this in a way that does not reload the data
-store container.
+Edit the [core/invoker/Dockefile](../../core/invoker/Dockefile)
+Update the variable with the version
+```
+ENV DOCKER_VERSION 18.03.0-ce
+```
+Then update line with the curl download command like
+```
+RUN curl -sSL -o docker-${DOCKER_VERSION}.tgz https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz && \
```
mdeuser: just wondering.  the current line is ```RUN curl -sSL -o docker-${DOCKER_VERSION}.tgz https://get.docker.com/builds/Linux/x86_64/docker-${DOCKER_VERSION}.tgz && \```   it's interesting that the docker 1.12.0 version has a different download template URL (i.e. host name is different, some path segments are different), so i'm wondering if this path may be docker version specific..?

tools/vagrant/README.md
```diff
@@ -176,46 +186,46 @@ cd ${OPENWHISK_HOME}
...
 cd ${ANSIBLE_HOME}
-# teardown all deployed containers
-ansible-playbook -i environments/local teardown.yml
-# deploy couchdb container
-ansible-playbook -i environments/local couchdb.yml
-# initialize db with guest/system keys
-ansible-playbook -i environments/local initdb.yml
-# recreate main db for entities
-ansible-playbook -i environments/local wipe.yml
+# teardown all containers expect couchdb container
```
mdeuser: "~expect~ `except`" typo?

tools/vagrant/Vagrantfile
```diff
@@ -92,57 +93,68 @@ Vagrant.configure('2') do |config|
         fi
 
         cd /home/vagrant/openwhisk/ansible
-        su vagrant -c 'ansible-playbook -i environments/local setup.yml'
-        su vagrant -c 'ansible-playbook -i environments/local prereq.yml'
+        su vagrant -c 'ansible-playbook -i environments/vagrant setup.yml'
+        su vagrant -c 'ansible-playbook -i environments/vagrant prereq.yml'
```
mdeuser: currently, the `-i environments/local` is used for vagrant installations.  will that continue to work with this new vagrant installation/environment?

tools/vagrant/README.md
```diff
@@ -176,46 +186,46 @@ cd ${OPENWHISK_HOME}
...
+```
+ENV DOCKER_VERSION 18.03.0-ce
+```
+Then update line with the curl download command like
+```
+RUN curl -sSL -o docker-${DOCKER_VERSION}.tgz https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz && \
+```
+Then update the ansible configuration to enable the use of runc, edit [](../../ansible/environments/vagrant/group_vars/all)
+```
+invoker_use_runc: true
```
mdeuser: will the following Vagrantfile command update in this PR
```su vagrant -c 'ansible-playbook -i environments/vagrant openwhisk.yml -e invoker_use_runc=False'```
override that ansible property value?

csantanapr: yes

csantanapr:
@mdeuser updated the readme with your comments.

PR #3559, opened 2018-04-19
add support for mesos attribute constraints 

tysonnorris:
For use when launching container using MesosContainerFactory

## Description

added config properties to MesosConfig:
```
whisk.mesos{
        constraints = [] //placement constraint strings to use for managed containers e.g. ["att1 LIKE v1", "att2 UNLIKE v2"]
        blackbox-constraints = [] //placement constraints to use for blackbox containers
        constraint-delimiter = " "//used to parse constraint strings
        teardown-on-exit = true //set to true to disable the mesos framework on system exit; set for false for HA deployments
}
```
## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -162,5 +162,9 @@ whisk {
         role = "*" //see http://mesos.apache.org/documentation/latest/roles/#associating-frameworks-with-roles
         failover-timeout = 0 seconds  //Timeout allowed for framework to reconnect after disconnection.
         mesos-link-log-message = true //If true, display a link to mesos in the static log message, otherwise do not include a link to mesos.
+        constraints = [] //placement constraint strings to use for managed containers e.g. ["att1 LIKE v1", "att2 UNLIKE v2"]
+        blackbox-constraints = [] //placement constraints to use for blackbox containers
+        constraint-delimiter = " "//used to parse constraint strings
+        teardown-on-exit = true //set to true to disable the mesos framework on system exit; set for false for HA deployments
     }
 }
--- a/common/scala/src/main/scala/whisk/core/mesos/MesosContainerFactory.scala
+++ b/common/scala/src/main/scala/whisk/core/mesos/MesosContainerFactory.scala
@@ -20,10 +20,14 @@ package whisk.core.mesos
 import akka.actor.ActorRef
 import akka.actor.ActorSystem
 import akka.pattern.ask
+import com.adobe.api.platform.runtime.mesos.Constraint
+import com.adobe.api.platform.runtime.mesos.LIKE
+import com.adobe.api.platform.runtime.mesos.LocalTaskStore
 import com.adobe.api.platform.runtime.mesos.MesosClient
 import com.adobe.api.platform.runtime.mesos.Subscribe
 import com.adobe.api.platform.runtime.mesos.SubscribeComplete
 import com.adobe.api.platform.runtime.mesos.Teardown
+import com.adobe.api.platform.runtime.mesos.UNLIKE
 import java.time.Instant
 import pureconfig.loadConfigOrThrow
 import scala.concurrent.Await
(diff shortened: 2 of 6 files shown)


common/scala/src/main/resources/application.conf
```diff
@@ -162,5 +162,9 @@ whisk {
         role = "*" //see http://mesos.apache.org/documentation/latest/roles/#associating-frameworks-with-roles
         failover-timeout = 0 seconds  //Timeout allowed for framework to reconnect after disconnection.
         mesos-link-log-message = true //If true, display a link to mesos in the static log message, otherwise do not include a link to mesos.
+        constraints = [] //placement constraint strings to use for managed containers e.g. ["att1 LIKE v1", "att2 UNLIKE v2"]
+        blackbox-constraints = [] //placement constraints to use for blackbox containers
```
ddragosd: I'm wondering whether the difference is in `managed containers` vs `blackbox containers` to decide where to spin up new action containers, or if it should be in some other place. I.e if I want to run containers on `gpu` VMs vs regular VMs, regardless of them being `blackbox` or not, where can we define this ? The reason I'm asking is b/c if the decision point should be influenced by other factors, maybe we don't _have_ to differentiate here between `blackbox` vs `non-blackbox` containers ?

tysonnorris:
You can always use the same placement constraints for black box and
managed. I just thought someone might find it useful to corral their black
box containers separately, unrelated to resource requirements.

Currently there is no notion of actions requiring specific resources other
than memory(I think?), which is a separate problem.

Is that what you mean?

ddragosd:
Thanks @tysonnorris .

Provided we leave these 2 settings for now, I'm wondering if the blackbox ratio setting in the controller must be updated to correctly match the VMs allocated for blackbox when using _different_ placement constraints ? For example, given a cluster with `100` VMs, out of which `30` VMs have special attributes for running blackbox actions, should the ratio config in the controller be updated as well to `0.3` ?

ddragosd:
Precisely for this uncertainty of whether the blackbox ratio matters or not I was coming from with the thought that maybe we don't need to differentiate between blackbox vs non-blackbox right now. Intuitively I'm thinking that this would change once we introduce semantics for actions that require GPU resources, being blackbox or non-blackbox. 

If you feel strongly about keeping the separation as-is in this PR, I would 👍 , waiting to solve the GPU part to get to the actual solution.

ddragosd:
Given that this is only touching the Mesos parts, do we need to wait for a PG before merging ? 

cc @markusthoemmes @mrutkows @rabbah

dgrove-oss:
I kicked off PG2/3043 for you (might be faster than getting an answer....)

rabbah:
LGTM.

dgrove-oss:
PG2 3403 🔵

ddragosd:
thanks a lot @dgrove-oss !

common/scala/src/main/scala/whisk/core/mesos/MesosContainerFactory.scala
```diff
@@ -126,9 +139,28 @@ class MesosContainerFactory(config: WhiskConfig,
...
+    constraintStrings.flatMap(cs => {
+      val parts = cs.split(mesosConfig.constraintDelimiter)
+      require(parts.length == 3, "constraint must be in the form <attribute><delimiter><operator><delimiter><value>")
+      Seq(LIKE, UNLIKE).find(_.toString == parts(1)) match {
+        case Some(o) => Some(Constraint(parts(0), o, parts(2)))
+        case _ =>
+          logging.warn(this, s"ignoring unsupported constraint operator ${parts(1)}")
+          None
+      }
+    })
```
markusthoemmes: Just to leave the note, this could make use of Scala's pattern matching, as in:

```scala
cs.split(mesosConfig.constraintDelimiter).toList match {
  case a :: LIKE.toString :: b   => Some(Constraint(a, LIKE, b)
  case a :: UNLIKE.toString :: b => Some(Constraint(a, UNLIKE, b))
  case a ::o :: b                =>
    logging.warn(this, s"ignoring unsupported constraint operator $o")
    None
  case failure =>
    logging.warn(this, s"couldn't parse constraint $failure")
    None
}
```

This wouldn't throw an exception in any case and is typechecked throughout (`parts(i)` will throw on an out-of-bounds exception)

PR #3562, opened 2018-04-20
ArtifactStore implementation for CosmosDB

chetanmeh:
This PR provides a `CosmosDBArtifactStore` implementation to enable using CosmosDB for storing subjects, whisks and activation in [Azure CosmosDB][1]

This is currently work in progress and opening this early so as to get feedback on design progress

## Description

Currently OpenWhisk supports CouchDB for storing various entities like actions, rules, subjects, activations etc. This PR provides a [CosmosDB][1] based implementation for ArtifactStore SPI

### Usage of CosmosDB Java SDK

CosmosDB supports various modes to connect to it. For our usage we have two options

1. [REST API][2]
2. [Async Java SDK][3]
3. [Sync Java SDK][6]

Compared to CouchDB performing queries against CosmosDB requires client side computation which involves sending queries to each partition and then collect and merge the result set. The Async Java SDK takes care of all these interactions and provides a simplified reactive api based on [RxJava][5]

Given the complexity involved in performing various operations against CosmosDB this PR uses the Java SDK to simplify and speed up the implementation

#### Partitioning Strategy

TBD

#### Query Execution

TBD

#### Attachment Storage

TBD

#### Testing Approach

TBD

- [X] Tests
- [x] Deployment
- [X] Documentation

## Types of changes

## Checklist:

[1]: https://docs.microsoft.com/en-us/azure/cosmos-db/
[2]: https://docs.microsoft.com/en-us/rest/api/documentdb/restful-interactions-with-documentdb-resources
[3]: https://docs.microsoft.com/en-us/azure/cosmos-db/sql-api-sdk-async-java
[4]: https://github.com/Azure/azure-cosmosdb-java
[5]: https://github.com/ReactiveX/RxJava
[6]: https://github.com/Azure/azure-documentdb-java


chetanmeh:
<a name="test-suite-result"></a>
### Test suite run result

In a [separate branch][1] travis is configured to run whole OW test suite as being run currently for master with default `ArtifactStoreProvider` switched to `CosmosDBArtifactStoreProvider`. In such a run [some test got failed][2]. Below is the analysis for them

* `WskRestEntitlementTests` - These test make use of `wskadmin` to provision a custom user. Currently `wskadmin` does not support CosmosDB. Hence for now test being ignored
* `ActionLimitsTests` - Some of the test here got failed due to Azure/azure-cosmosdb-java#33. PR raised for that. For now being ignored
* `SequenceMigrationTests` - These are passing now

[1]: https://github.com/chetanmeh/incubator-openwhisk/tree/artifact-store-cosmos-ci
[2]: https://scans.gradle.com/s/nftphfau7o4a6/tests/failed

csantanapr:
Makes sense to update wskadmin to also handle momdodb/cosmosdb ?

chetanmeh:
> Makes sense to update wskadmin to also handle momdodb/cosmosdb ?

Yes this aspect was discussed sometime back [here][1]. Need to figure out right way to abstract that in python and refactor wskadmin. Would do that in a separate PR

[1]: https://lists.apache.org/thread.html/921a0a6350a7ec3a2dc77564612de59104995622f8417583291f20bc@%3Cdev.openwhisk.apache.org%3E

chetanmeh:
## Indexing Policy

CosmosDB provides rich [indexing options][1] which support

* Indexing mode - consistent or lazy
* Hash Index - For equality check
* Range Index - For range check and order by
* Control Precision - Allows to make trade-offs between index storage overhead and query performance
* Boolean properties are indexed by default - As per [create document api][3] _Booleans and nulls are automatically indexed_ (See dataType)

Currently the view based indexes in CouchDB are eventually consistent and all the test around queries account for that. Based on usage following indexes are configured

### Subjects

* /uuid/? - Authentication query
* /subject/?
* /namespaces/[]/uuid/? - Authentication query
* /namespaces/[]/name/?
* /concurrentInvocations/? - For blacklist check query
* /invocationsPerMinute/? - For blacklist check query

Key points
* Indexing mode is set to consistent
* All indexes have precision set to -1 for now. Later based on metrics we can determine if this needs to be tweaked
* `key` is not being indexed - Authentication query is performed on uuid and key. Of this index is only created for `uuid` and `key` check should be taken care by filtering in query evaluation. As `uuid` is very much unique number of rows scanned for filtering should be small
* `concurrentInvocations` and `invocationsPerMinute` are used in "namespace name/limits" documents which are then used to find out blacklisted namespace if either of them is zero. We can possibly remove this by setting `blocked` to true even for limits docs

### Whisks

* /entityType/? - Hash
* /namespace/? - Hash
* /\_c/rootns/? - Hash
* /updated/? - Range

Key points
* Indexing mode set to consistent

### Activations

* /namespace/?
* /_c/nspath/?
* /start/?

Key points
* Indexing mode set to lazy - For activations the rate of writes would be much higher compared to those on `whisks` and `subjects`. So here it would be better to keep indexing mode lazy

### Index policy updates

Current logic upon system start would query for existing indexes configured on the collection. If there is a mismatch between what is expected and what is currently found it would provision the new indexes. 

CosmosDB currently shows a strange behavior where it is [creating extra indexes][4] for same path. For e.g. `/namespace/?` is configured with hash index. But post creation if indexes are checked its found that it has both `Range` and `Hash` index configured. To account for these extra indexes in index check this PR uses a custom `IndexingPolicy` type which accounts for this type of behaviour while check if indexes are same.

CosmosDB supports [online index policy changes][5] where by change in index definition would be performed asynchronously and online allowing the collection to remain available for writes while transformation is in progress. However if the application makes query relying on new indexes then those queries would fail untill indexing completes based on new index definition.

So in future if any change in index definition is done then it should be provisioned prior to deploying new code relying on that index. Current approach of checking for index definition would ensure that index definitions are consistent 



[1]: https://docs.microsoft.com/en-us/azure/cosmos-db/indexing-policies
[2]: https://docs.microsoft.com/en-us/azure/cosmos-db/indexing-policies#index-precision
[3]: https://docs.microsoft.com/en-us/rest/api/cosmos-db/create-a-collection
[4]: https://stackoverflow.com/questions/43775750/documentdb-creating-extra-indexes
[5]: https://azure.microsoft.com/en-in/blog/update-your-documentdb-indexing-policies-online/

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala
```diff
@@ -0,0 +1,104 @@
...
+
+    new CosmosDBArtifactStore(dbName, config, clientRef, handler, viewMapper)
+  }
+
+  private def handlerAndMapper[D](entityType: ClassTag[D])(
+    implicit actorSystem: ActorSystem,
+    logging: Logging,
+    materializer: ActorMaterializer): (String, DocumentHandler, CosmosDBViewMapper) = {
+    entityType.runtimeClass match {
+      case x if x == classOf[WhiskEntity] =>
```
tysonnorris: is there a better way? does ` case x:WhiskEntity =>` not work?

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala
```diff
@@ -0,0 +1,104 @@
...
+        ("whisks", WhisksHandler, WhisksViewMapper)
+      case x if x == classOf[WhiskActivation] =>
+        ("activations", ActivationHandler, ActivationViewMapper)
+      case x if x == classOf[WhiskAuth] =>
+        ("subjects", SubjectHandler, SubjectViewMapper)
+    }
+  }
+
+  private def getOrCreateReference(config: CosmosDBConfig) = synchronized {
```
tysonnorris: I get the feeling this reference is special - can you outline the goals of `ReferenceCounted` and `synchronized`?

nm: I see the "Shutdown Handling" comment

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala
```diff
@@ -0,0 +1,104 @@
...
+case class CosmosDBConfig(endpoint: String, key: String, db: String)
+
+case class ClientHolder(client: AsyncDocumentClient) extends Closeable {
+  override def close(): Unit = client.close()
+}
+
+object CosmosDBArtifactStoreProvider extends ArtifactStoreProvider {
+  type DocumentClientRef = ReferenceCounted[ClientHolder]#CountedReference
+  private lazy val config = loadConfigOrThrow[CosmosDBConfig](ConfigKeys.cosmosdb)
+  private var clientRef: ReferenceCounted[ClientHolder] = _
```
tysonnorris: does `config` need to be lazy? is there any reason not to make it non-lazy, and create the clientRef immediately here? (then no need for the special `getOrCreateReference` function)

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBSupport.scala
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  protected def querySpec(id: String) =
+    new SqlQuerySpec("SELECT * FROM root r WHERE r.id=@id", new SqlParameterCollection(new SqlParameter("@id", id)))
+
+  /**
+   * CosmosDB id considers '/', '\' , '?' and '#' as invalid. EntityNames can include '/' so
+   * that need to be escaped. For that we use '|' as the replacement char
+   */
+  protected def escapeId(id: String): String = id.replace("/", "|")
```
tysonnorris: I think this requires that '|' cannot be used in the id at all?  (that is ok by me, but should be validated in `escapeId` and `unescapeId`, I think)

common/scala/build.gradle
```diff
@@ -47,6 +47,10 @@ dependencies {
     compile 'io.kamon:kamon-statsd_2.11:0.6.7'
     //for mesos
     compile 'com.adobe.api.platform.runtime:mesos-actor:0.0.7'
+
+    compile 'io.reactivex:rxscala_2.11:0.26.5'
+    compile 'io.reactivex:rxjava-reactive-streams:1.2.1'
+    compile 'com.microsoft.azure:azure-cosmosdb:1.0.0'
```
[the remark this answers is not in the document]
chetanmeh: Below is the dependency tree being pulled in (see [here][1] for full tree)


```
+--- io.reactivex:rxscala_2.11:0.26.5
     |    +--- org.scala-lang:scala-library:2.11.8 -> 2.11.12
     |    \--- io.reactivex:rxjava:1.2.4 -> 1.3.3
     +--- io.reactivex:rxjava-reactive-streams:1.2.1
     |    +--- io.reactivex:rxjava:1.2.2 -> 1.3.3
     |    \--- org.reactivestreams:reactive-streams:1.0.0 -> 1.0.2
     \--- com.microsoft.azure:azure-cosmosdb:1.0.0
          +--- com.fasterxml.jackson.core:jackson-databind:2.9.4 (*)
          +--- com.fasterxml.uuid:java-uuid-generator:3.1.4
          +--- org.json:json:20140107
          +--- commons-io:commons-io:2.5 -> 2.6
          +--- com.github.davidmoten:rxjava-extras:0.8.0.11
          |    \--- io.reactivex:rxjava:1.3.3
          +--- io.reactivex:rxjava:1.3.3
          +--- io.reactivex:rxjava-string:1.1.1
          |    \--- io.reactivex:rxjava:1.2.3 -> 1.3.3
          +--- io.reactivex:rxnetty:0.4.20
          |    +--- io.reactivex:rxjava:1.2.2 -> 1.3.3
          |    +--- io.netty:netty-codec-http:4.1.5.Final -> 4.1.20.Final
          |    |    \--- io.netty:netty-codec:4.1.20.Final
          |    |         \--- io.netty:netty-transport:4.1.20.Final
          |    |              +--- io.netty:netty-buffer:4.1.20.Final
          |    |              |    \--- io.netty:netty-common:4.1.20.Final
          |    |              \--- io.netty:netty-resolver:4.1.20.Final
          |    |                   \--- io.netty:netty-common:4.1.20.Final
          |    +--- io.netty:netty-handler:4.1.5.Final -> 4.1.20.Final
          |    |    +--- io.netty:netty-buffer:4.1.20.Final (*)
          |    |    +--- io.netty:netty-transport:4.1.20.Final (*)
          |    |    \--- io.netty:netty-codec:4.1.20.Final (*)
          |    \--- org.slf4j:slf4j-api:1.7.6 -> 1.7.25
          +--- io.netty:netty-codec-http:4.1.20.Final (*)
          +--- io.netty:netty-handler:4.1.20.Final (*)
          +--- io.netty:netty-transport:4.1.20.Final (*)
          +--- org.slf4j:slf4j-api:1.7.6 -> 1.7.25
          \--- org.apache.commons:commons-lang3:3.5
```  

Further did a diff on jars being packaged in controller.tar

```diff
30a31
> azure-cosmosdb-1.0.0.jar
34a36
> commons-lang3-3.5.jar
41,43c43,45
< jackson-annotations-2.7.0.jar
< jackson-core-2.7.7.jar
< jackson-databind-2.7.7.jar
---
> jackson-annotations-2.9.0.jar
> jackson-core-2.9.4.jar
> jackson-databind-2.9.4.jar
46c48
< java-uuid-generator-3.1.3.jar
---
> java-uuid-generator-3.1.4.jar
52a55
> json-20140107.jar
69a73,79
> netty-buffer-4.1.20.Final.jar
> netty-codec-4.1.20.Final.jar
> netty-codec-http-4.1.20.Final.jar
> netty-common-4.1.20.Final.jar
> netty-handler-4.1.20.Final.jar
> netty-resolver-4.1.20.Final.jar
> netty-transport-4.1.20.Final.jar
78a89,94
> rxjava-1.3.3.jar
> rxjava-extras-0.8.0.11.jar
> rxjava-reactive-streams-1.2.1.jar
> rxjava-string-1.1.1.jar
> rxnetty-0.4.20.jar
> rxscala_2.11-0.26.5.jar

```      

So following existing dependency versions are getting updated

1. jackson - 2.7 -> 2.9 - (minor). So far jackson was being used by kubernetes-client
2. java-uuid-generator - 3.1.3 -> 3.1.4 (micro) 

Following new set of library being included

1. netty - 4.1.20 - 7 jars
2. rxjava  - 6
3. commons-lang3
4. json - This would get replaced with jackson post Azure/azure-cosmosdb-java#29

In total post this change 

1. Number of jars in controller increase from 92 -> 108
2. Size increases from 89M -> 94M     

[1]: https://gist.github.com/chetanmeh/ab354bb8c320020b8c3d4233bf8f3f62

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala
chetanmeh: That would work if x is an instance of `WhiskEntity`. However here x is class representing `WhiskEntity`. So current approach has to be taken

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala
chetanmeh: For some of the test setups I am using different database name and thus have a separate method which takes config explicitly. Making it  non lazy would cause issue if default config is not complete causing the class load to fail.

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBSupport.scala
chetanmeh: Good point. So far this was checked in `EntityName#entityNameMatcher` however would be good to check the invariant in util method also

Added a check with test for this logic

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala
tysonnorris: Isn't that OK? (to fail in case of invalid config?) It would only happen if CosmosDB is enabled, and the config is invalid. The tests can always provide a valid non-default config.

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala
tysonnorris: `ArtifactStore` has a `shutdown()` function - is there a reason we cannot use that instead of the specialized ref counting? (client is responsible for shutdown). I don't see any non-test usages of `shutdown` currently - what is the hazard, if any, in either treating this the same as couchdb (no shutdown called, except for test cases) OR requiring that clients invoke shutdown (e.g. via `sys.addShutdownHook(store.shutdown())` in controller/invoker)? 

Part of the diff in client handling from couchdb is the client sharing - you mention that stores _share the same AsyncDocumentClient instance_, but it's not clear why?

Just looking for ways to simply this

tysonnorris: After reading some of the cosmosdb doco (sorry, cosmos n00b here), I see now that the "databases" are modeled as "multiple collections within a single db" which partly explains the client sharing, but I'm still wondering what is the impact if we create per-collection clients?

common/scala/build.gradle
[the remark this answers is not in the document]
tysonnorris: Only concern is the jump in netty version - when I build controller.tar from your branch I get both netty versions:
```tnorris@tnorris-macOS ~/git/forks/adobe-apiplatform/incubator-openwhisk-mac -  (chetanmeh-artifact-store-cosmos) $ tar -tvf core/controller/build/distributions/controller.tar  | grep netty
-rw-r--r--  0 0      0      413674 May 18 13:27 controller/lib/rxnetty-0.4.20.jar
-rw-r--r--  0 0      0      562559 May 18 13:27 controller/lib/netty-codec-http-4.1.22.Final.jar
-rw-r--r--  0 0      0      374389 May 18 13:27 controller/lib/netty-handler-4.1.22.Final.jar
-rw-r--r--  0 0      0      316316 May 18 13:27 controller/lib/netty-codec-4.1.22.Final.jar
-rw-r--r--  0 0      0      456605 May 18 13:27 controller/lib/netty-transport-4.1.22.Final.jar
-rw-r--r--  0 0      0      270720 May 18 13:27 controller/lib/netty-buffer-4.1.22.Final.jar
-rw-r--r--  0 0      0       32276 May 18 13:27 controller/lib/netty-resolver-4.1.22.Final.jar
-rw-r--r--  0 0      0      575748 May 18 13:27 controller/lib/netty-common-4.1.22.Final.jar
-rw-r--r--  0 0      0     1292696 Sep 13  2017 controller/lib/netty-3.10.6.Final.jar```

**Comment by tysonnorris** [comment]
This seems good to me, would like some others to chime in. My questions are mainly around:
- client connection sharing across ArtifactStores (can we just create separate connections?)
- multiple stores currently getting created (is there any risk?)

On this last one, I think it is a defect of `WhiskStore` that multiple calls to `WhiskEntityStore.datastore()` will return multiple unique instances - I think these should return the same instance (invoke `makeStore[WhiskEntity]()` once ever). I don't know if it impacts cosmos (client is shared with multiple separate instances of the same store?), but if so, maybe we can fix it here? otherwise can fix it separately.

File: common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala (hunk 2 above)
Comment (chetanmeh) [reply common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala]: Each `AsyncDocumentClient` encapsulates a HttpClient which own a pool of HTTP Connections to CosmosDB. As for all 3 collections we are connecting to same url at backend it would be better to share the pool.

Compared to this in `CouchDBRestStore` the pool is managed on per `ActorSystem` level which by design is then shared across various `ArtifactStore` implementations. In [PoolingRestClient#shutdown][1] method you would see a note around not closing the pool assuming shutdown is only relevant for tests. So ref counted approach used in this PR is meant to solve that 

[1]: https://github.com/apache/incubator-openwhisk/blob/master/common/scala/src/main/scala/whisk/http/PoolingRestClient.scala#L150

**Comment by chetanmeh** [comment]
> On this last one, I think it is a defect of WhiskStore that multiple calls to WhiskEntityStore.datastore() will return multiple unique instances

One benefit of this approach is that future test using `MemoryArtifactStore` can work without any side effect of other tests. Caching the result of call of `makeStore` would make `shutdown` handling tricky. Currently each eventual owner of store instance is responsible for shutdown. Probably in that case we need to have a shutdown support at provider level

File: common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala (hunk 3 above)
Comment (chetanmeh) [reply common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala]: Even in that case `getOrCreateReference` may be required as currently there can be multiple shutdown calls made. In case of object level instance we would need to change the shutdown approach and move it to provider level

Comment (tysonnorris) [reply common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala]: I guess I favor multiple shutdown calls vs the reference tracking. 
I tested changing the provider to:
```
  private val config = loadConfigOrThrow[CosmosDBConfig](ConfigKeys.cosmosdb)
  private val clientRef = createClient(config)
```
and it seems to run fine in `CosmosDBArtifactStoreTests` - it ends up calling `client.close()` multiple times, as expected, and seems to operate properly.

**Comment by tysonnorris** [comment]
I think adding `ArtifactStoreProvider.shutdown()` seems like a fine addition, if calling `client.close()` multiple times is not reliable. The tests will need to know that some artifact stores have client bound to provider, and others have client bound to store, which I think is not great, but it is OK given that nothing is invoking shutdown currently, and the behavior is different amongst the stores currently.

File: common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala (hunk 3 above)
Comment (chetanmeh) [reply common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala]: I think post `close` no other operation is performed so it does not show up any issue. All clients are closed in `affterAll` i.e. once all methods in suite have run

**Comment by chetanmeh** [comment]
@tysonnorris I agree current ref count based shutdown handling is not desirable and we should relook into how `ArtifactStore` are closed. May be we deal with this in a separate issue and use existing approach in this PR for now?

**Comment by chetanmeh** [comment]
I am working on a separate PR which would add new stages to run full test suite with Cosmos. That would allow complete test coverage against this new store. Would raise it once this PR is merged

**Comment by chetanmeh** [comment]
Needs a PG run

**Comment by cbickel** [comment]
PG with commit hash 7eb9d23b8e8940b3630f98519f4cefa194983e42 is 🔵

File: common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala (hunk 1 above)
Comment (rabbah) [reply common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala]: given the matching and `if` predicates, might as well use if/else directly for brevity.

File: common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala (hunk 2 above)
Comment (rabbah) [reply common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala]: getOrCreateReference is called once at create time - is there concurrency here in making the stores that requires synchronization or is this mostly defensive to get the counts right?

File: common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBSupport.scala [hunk 6]
```diff
@@ -0,0 +1,85 @@
...
+  protected def viewMapper: CosmosDBViewMapper
+
+  def initialize(): (Database, DocumentCollection) = {
+    val db = getOrCreateDatabase()
+    (db, getOrCreateCollection(db))
+  }
+
+  private def getOrCreateDatabase(): Database = {
+    client
+      .queryDatabases(querySpec(config.db), null)
```
rabbah: is it possible to avoid `null` and use None (viz Option) instead?
we've generally stayed away from null.

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStoreProvider.scala
rabbah: can you add a comment explaining the synchronized fwiw.

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBSupport.scala
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  private def newDatabase = {
+    val databaseDefinition = new Database
+    databaseDefinition.setId(config.db)
+    databaseDefinition
+  }
+
+  protected def querySpec(id: String) =
+    new SqlQuerySpec("SELECT * FROM root r WHERE r.id=@id", new SqlParameterCollection(new SqlParameter("@id", id)))
```
rabbah: comment? what is this for...

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBUtil.scala
```diff
@@ -0,0 +1,108 @@
...
+    val m = fields.foldLeft(Map.empty[String, Any]) { (map, name) =>
+      addToMap(name, map)
+    }
+    val withId = addToMap(cid, m)
+    val json = asJsonLikeString(withId)
+    s"$json AS $alias"
+  }
+
+  private def addToMap(name: String, map: Map[String, _]): Map[String, Any] = name.split('.').toList match {
+    case Nil     => sys.error(s"Should not reach here $name")
```
rabbah: throw exception directly instead?

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBUtil.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * Prepares the json like select clause
+   * {{{
+   *   Seq("a", "b", "c.d.e") =>
+   *   { "a" : r['a'], "b" : r['b'], "c" : { "d" : { "e" : r['c']['d']['e']}}, "id" : r['id']} AS view
+   * }}}
+   * Here it uses {{{r['keyName']}}} notation to avoid issues around using reserved words as field name
+   */
+  def prepareFieldClause(fields: Iterable[String]): String = {
```
rabbah: should this be wrapped with a Try for failures (via one of the invariant violations below)?

rabbah: note to self: check unit tests for this trait.

common/scala/src/main/scala/whisk/core/database/cosmosdb/ReferenceCounted.scala
```diff
@@ -0,0 +1,49 @@
...
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger}
+
+private[cosmosdb] case class ReferenceCounted[T <: AutoCloseable](private val inner: T) {
+  private val count = new AtomicInteger(0)
+
+  private def inc(): Unit = count.incrementAndGet()
+
+  private def dec(): Unit = {
+    val newCount = count.decrementAndGet()
+    if (newCount == 0) {
```
rabbah: defensively `<=`?

common/scala/src/main/scala/whisk/core/database/cosmosdb/ReferenceCounted.scala
```diff
@@ -0,0 +1,49 @@
...
+      inner.close()
+      count.decrementAndGet()
```
rabbah: can you comment the second decrement.

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBSupport.scala
chetanmeh: `queryDatabases` is java api of Azure CosmosDB driver which uses the `null` based approach. So not possible to use Option here

chetanmeh: > getOrCreateReference is called once at create time

Added some comments. This method may be called concurrently multiple times at startup as various store instances are initialized. Hence need to `synchronize` it to ensure all those stores are backed by same client

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBUtil.scala
chetanmeh: String split should never result in empty list. Changed it to throw ISE

common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBUtil.scala
chetanmeh: The invariant should never fail as string split can not result in empty list. So should be fine without using `Try`

common/scala/src/main/scala/whisk/core/database/cosmosdb/ReferenceCounted.scala
chetanmeh: Second decrement make it negative which ensures that future `reference` calls result in error

chetanmeh:
Thanks a lot @rabbah and @tysonnorris for reviewing this big PR ✨. Opened #3872 to track documentation related work

PR #3563, opened 2018-04-20
Skip log collection if the log-limit is set to "0".

markusthoemmes:
Enables the per-action log-limit configuration to be read from config (min/max/std values respectively). It also skips the entire log-collecting process if that limit is set to 0 (no logs allowed).

## TODO

- [x] A test 👅 

- [ ] API
- [X] Controller
- [ ] Loadbalancer
- [X] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -156,7 +156,7 @@ whisk {
         std = 256 m
     }
 
-    # action memory configuration
+    # action log-limit configuration
     log-limit {
         min = 0 m
         max = 10 m
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -156,6 +156,13 @@ whisk {
         std = 256 m
     }
 
+    # action memory configuration
+    log-limit {
+        min = 0 m
+        max = 10 m
+        std = 10 m
+    }
+
     mesos {
         master-url = "http://localhost:5050" //your mesos master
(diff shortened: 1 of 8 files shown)


common/scala/src/main/resources/application.conf
```diff
@@ -156,6 +156,13 @@ whisk {
         std = 256 m
     }
 
+    # action memory configuration
```
rabbah: Wrong comment.

markusthoemmes:
@tysonnorris @bbrowning Could you have a quick look at this from a Mesos/Kube perspective? It's skipping `collectLogs` entirely for a specific case, just wanted to quickly make sure it doesn't break Mesos/Kube in a way.

markusthoemmes:
PG1 2805 🔵

dgrove-oss:
don't see any obvious kube problems; it should smoothly pick up the default from application.conf.

markusthoemmes:
@tysonnorris The idea behind this:
1. Optimization
2. You're able to disable logging for certain actions at **runtime**.

Min and Std values are used to control the bounds + default behavior of the system. You can set Std to 1Meg, then every action created by a user will get only 1 megabyte, but it'll still allow you to go upto Max.

tysonnorris:
@markusthoemmes got it thanks

PR #3588, opened 2018-04-26
Make count with skip to work with CouchDB

### LINKED ISSUE #3560: Count with skip is not working for CouchDB
As mentioned [here][1] in #3517 count with skip seems to be not working for CouchDB. As a fix we can instead compute the count by subtracting skip from overall count in CouchDBReststore. 

Once this is fixed the test case `ArtifactStoreQueryBehaviors` should be enabled again

[1]: https://github.com/apache/incubator-openwhisk/pull/3517#issuecomment-381031566

chetanmeh:
Fixes #3560

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3560)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
+++ b/common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
@@ -258,6 +258,8 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
                                      stale: StaleParameter)(implicit transid: TransactionId): Future[List[JsObject]] = {
 
     require(!(reduce && includeDocs), "reduce and includeDocs cannot both be true")
+    require(skip >= 0, "skip should be non negative")
+    require(limit >= 0, "limit should be non negative")
 
     // Apparently you have to do that in addition to setting "descending"
     val (realStartKey, realEndKey) = if (descending) {
@@ -309,6 +311,7 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
 
   protected[core] def count(table: String, startKey: List[Any], endKey: List[Any], skip: Int, stale: StaleParameter)(
     implicit transid: TransactionId): Future[Long] = {
+    require(skip >= 0, "skip should be non negative")
 
     val Array(firstPart, secondPart) = table.split("/")
 
--- a/common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
+++ b/common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
@@ -320,9 +320,10 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
         case Right(response) =>
           val rows = response.fields("rows").convertTo[List[JsObject]]
 
-          val out = if (!rows.isEmpty) {
+          val out = if (rows.nonEmpty) {
             assert(rows.length == 1, s"result of reduced view contains more than one value: '$rows'")
-            rows.head.fields("value").convertTo[Long] - skip
+            val count = rows.head.fields("value").convertTo[Long]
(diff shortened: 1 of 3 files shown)


chetanmeh:
@rabbah Please review

rabbah:
@csantanapr @dubee do you mind a PG 🚀 for this? It should be OK but better to be sure.

csantanapr:
PG2 3080 🏃

csantanapr:
PG2 3081 🔵

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -315,19 +315,14 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
...
-        reduce = true)
+      .executeView(firstPart, secondPart)(startKey = startKey, endKey = endKey, stale = stale, reduce = true)
       .map {
         case Right(response) =>
           val rows = response.fields("rows").convertTo[List[JsObject]]
 
           val out = if (!rows.isEmpty) {
             assert(rows.length == 1, s"result of reduced view contains more than one value: '$rows'")
-            rows.head.fields("value").convertTo[Long]
+            rows.head.fields("value").convertTo[Long] - skip
```
dubee: @chetanmeh, what happens if skip > value?

chetanmeh: Good point!. That case was not handled and would have resulted in negative count. Updated that to return 0 and also added a test to check that scenario

rabbah: nice catch... shame on me for missing it.

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -315,19 +315,15 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
...
       .map {
         case Right(response) =>
           val rows = response.fields("rows").convertTo[List[JsObject]]
 
-          val out = if (!rows.isEmpty) {
+          val out = if (rows.nonEmpty) {
             assert(rows.length == 1, s"result of reduced view contains more than one value: '$rows'")
-            rows.head.fields("value").convertTo[Long]
+            val count = rows.head.fields("value").convertTo[Long]
+            if (count > skip) count - skip else 0L
```
rabbah: skip may be negative also - although it is better to reject.a negative skip at the API handler (validate the value there) but defensively code against that as well here?

chetanmeh: Makes sense. Added a check for both `limit` and `skip` to  be non negative in both `query` and `count`

rabbah:
@dubee https://github.com/apache/incubator-openwhisk-cli/issues/187

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
dubee: @rabbah, are you saying to terminate a route if those values are negative at the top API level?

rabbah: yes - we can just validate and reject early (we do some other validations, so could extend to count and skip)... orthogonal to this PR.

dubee: Make sense. With these changes, `There was an internal server error.` is returned when a negative `skip` value is specified. A content-type of `text/plain; charset=UTF-8` is returned as well.

Issue #2147, opened 2017-04-15
Parse `+json` content types suffix for web action bodies


akrabat:
It would be nice if I could post with a content type of `vnd.akrabat.com/api+json` and have the web action automatically decode the JSON encoded body of my HTTP message.

The +json suffix if specified in [RFC 6839](https://tools.ietf.org/html/rfc6839#section-3.1).

rabbah:
This is for `.http` web actions, where the controller passes the body intact to the action as __ow_body. It would recover the behavior we provide for `.json` web actions for example where the body is passes as a top level JSON object to the action as an argument.

@akrabat did I understand correctly?

akrabat:
Yes. For `.http` web actions, a `Content-Type` of `application/json` and `xxx/yyy+json` should behave the same.

duglin:
I just ran into this today (I think its either the same, or related, issue).  My function is in a docker container and the __ow_body is sometimes base64 encoded and sometimes not based on these two content-type values:
`application/cloudevents+json`
`application/cloudevents+json; charset=UTF-8`
The UTF one is given to me as raw json, while the other one is base64 encoded. IMO both should be raw json. 

Additionally, it would be really nice if there was a flag (perhaps as a header) given to the function so it knows whether the body is base64 encoded or not. While I know it might be possible to look at the content-type and figure it out, you're asking the function to mimic the logic in OW to make that determination and that's not only error prone but could be a lot of work in the case where the function may not actually know about all of the various content-types and its acting more like a pass-thru. It seems like it would be really nice (from a UX) perspective if there was some flag someplace that the function could look at and know (w/o a doubt and w/o a lot of code) whether it needs to decode the body or not.

/cc @csantanapr

csantanapr:
@rabbah what are the next steps to fix this in your opinion?

rabbah:
On the utf8 non handling - that should be fixed; there is already an isJsonFamily method which is why one of them works but not the other. 

We can set a header or introduce another property to indicate b64 encoding. There is another related issue which is that extensions are not visible to the functions. If we do add a property we should address both at the same time.

duglin:
@markusthoemmes that will definitely address my immediate problem so thanks!!

I do still think having a flag someplace that the function can check to know when to decode the __ow_body would be useful as well.  (yes I'm demanding 😄  )

csantanapr:
Demand with PRs :-)

rabbah:
[...]

PR #3609, opened 2018-05-04
Support "+json" content-types as WebAction input.

markusthoemmes:
WebActions support various different input formats (like FormData). Only allowing `application/json` to be parsed and handled as JSON is an unnecessary limitation since there are many "+json" custom formats out there which boil down to JSON as well.

## Related issue and scope

- [ ] I opened an issue to propose and discuss this change (#2147)

- [X] API
- [X] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
@@ -21,7 +21,6 @@ import java.util.Base64
 
 import scala.concurrent.Future
 import scala.util.{Failure, Success, Try}
-
 import akka.http.scaladsl.model.HttpEntity.Empty
 import akka.http.scaladsl.server.Directives
 import akka.http.scaladsl.model.HttpMethod
@@ -36,7 +35,6 @@ import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.model.Uri.Query
 import akka.http.scaladsl.model.HttpEntity
 import akka.http.scaladsl.server.Route
-import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.model.headers.`Content-Type`
 import akka.http.scaladsl.model.headers.`Timeout-Access`
 import akka.http.scaladsl.model.ContentType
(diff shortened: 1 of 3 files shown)


markusthoemmes:
@jcrossley3 the `LenientSprayJsonSupport` I introduce here could be relevant if you want to make our general API more lenient.

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -1726,6 +1727,23 @@ trait WebActionsApiBaseTests extends ControllerTestCommon with BeforeAndAfterEac
         status should be(BadRequest)
       }
     }
+
+    it should s"support json (including +json subtypes) (auth? ${creds.isDefined})" in {
```
rabbah: There are already tests for this (+json) - what is different here is the explicit charset; i would tweak test name accordingly.

rabbah: Or merge tests?

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -574,7 +570,7 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
         case Empty =>
           process(None, isRawHttpAction)
 
-        case HttpEntity.Strict(ContentTypes.`application/json`, json) if !isRawHttpAction =>
+        case HttpEntity.Strict(ct, json) if WhiskWebActionsApi.isJsonFamily(ct.mediaType) && !isRawHttpAction =>
```
jcrossley3: How does this relate to the `customRejectionHandler` defined in `BasicHttpService`? I'm still unclear on the benefit of checking the content-type at all. Why not simply assume that all payloads are json regardless of content-type and return `HTTP 415 Unsupported Media Type` (with useful logging) in the event of a parse error?

markusthoemmes: This is for the JSON part of web-actions specifically. We **need** to check that here, since we also support FormData etc.

This doesn't apply for the general API where we can do what you suggest.

markusthoemmes:
PG2 3101 🔵

PR #3619, opened 2018-05-07
Provide an activation store SPI

dubee:
Allows for deployments to have a custom activation store through the use of a service provider interface (SPI)

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


chetanmeh:
@dubee  Can you share some details on requirement of a new SPI as the `ActivationStoreProvider` introduced in this PR is having similar feature set as `ArtifactStore` based implementation. Is the intention to store the activation in a store other than CouchDB for std  deployment

dubee:
@chetanmeh, yes we want to be able to use a different store to fetch activations from other than CouchDB. I plan to send some more information to the dev-list on this topic soon.

common/scala/src/main/scala/whisk/core/entity/ActivationStore.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  def store(activation: WhiskActivation)(implicit transid: TransactionId,
+                                         notifier: Option[CacheChangeNotification]): Future[DocInfo]
+
+  def get(activationId: ActivationId)(implicit transid: TransactionId): Future[WhiskActivation]
+
+  def delete(activationId: ActivationId)(implicit transid: TransactionId,
+                                         notifier: Option[CacheChangeNotification]): Future[Boolean]
+
+  def response(activationId: ActivationId)(implicit transid: TransactionId): Future[WhiskActivation]
```
chetanmeh: Whats the difference between `get` and `response`

dubee:
@markusthoemmes, any review comments?

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
```diff
@@ -92,43 +92,58 @@ class ActivationsApiTests extends ControllerTestCommon with WhiskActivationsApi
...
-          a.getFields("for") match {
-            case Seq(JsString(n)) => n == actionName.asString
-            case _                => false
+    try {
+      notExpectedActivations foreach {
+        storeActivation(_)
+      }
+      activations foreach {
+        storeActivation(_)
+      }
```
markusthoemmes: You should be able to shorthand this into:

```scala
(notExpectedActivations ++ activations).foreach(storeActivation)
```

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
```diff
@@ -92,43 +92,58 @@ class ActivationsApiTests extends ControllerTestCommon with WhiskActivationsApi
...
+        status should be(Forbidden)
+      }
+
+    } finally {
+      notExpectedActivations foreach { activation =>
+        deleteActivation(ActivationId(activation.docid.asString))
+      }
+      activations foreach { activation =>
+        deleteActivation(ActivationId(activation.docid.asString))
+      }
```
markusthoemmes: Can likewise be shorthanded

core/controller/src/main/scala/whisk/core/controller/Activations.scala
```diff
@@ -159,40 +157,16 @@ trait WhiskActivationsApi extends Directives with AuthenticatedRouteProvider wit
...
         countEntities {
-          WhiskActivation.countCollectionInNamespace(
-            activationStore,
-            name.flatten.map(p => namespace.addPath(p)).getOrElse(namespace),
-            skip.n,
-            since,
-            upto,
-            StaleParameter.UpdateAfter,
-            viewName = name.flatten.map(_ => WhiskActivation.filtersView).getOrElse(WhiskActivation.view))
+          activationStore.countActivationsInNamespace(name.flatten, namespace, skip.n, since, upto)
```
dubee: For some reason `name` is defined as a `Option[Option[EntityName]]`...

common/scala/src/main/scala/whisk/core/entity/ActivationId.scala
```diff
@@ -35,7 +35,7 @@ import scala.util.{Failure, Success, Try}
  *
  * @param asString the activation id
  */
-protected[whisk] class ActivationId private (val asString: String) extends AnyVal {
+protected[whisk] case class ActivationId private (val asString: String) extends AnyVal {
```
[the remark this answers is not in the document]
dubee: https://docs.scala-lang.org/overviews/core/value-classes.html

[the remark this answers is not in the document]
dubee: From what I read, `AnyVal` types are only instantiated when the below occurs. I think use a `case class` to extend `AnyVal` should preserve the non-allocating behavior of `ActivationID`.
```
Allocation Summary
A value class is actually instantiated when:

1. a value class is treated as another type.
2. a value class is assigned to an array.
3. doing runtime type tests, such as pattern matching.
```

dubee:

core/controller/src/main/scala/whisk/core/controller/Activations.scala
markusthoemmes: That's because of

`'name.as[Option[EntityPath]] ?` which makes the Option optionable. @rabbah any reason why this is the case?

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
```diff
@@ -530,10 +558,14 @@ class ActivationsApiTests extends ControllerTestCommon with WhiskActivationsApi
...
+    storeActivation(activation)
+    try {
 
-    Get(s"$collectionPath/${activation.activationId.asString}/bogus") ~> Route.seal(routes(creds)) ~> check {
-      status should be(NotFound)
+      Get(s"$collectionPath/${activation.activationId.asString}/bogus") ~> Route.seal(routes(creds)) ~> check {
+        status should be(NotFound)
+      }
+    } finally {
+      deleteActivation(ActivationId(activation.docid.asString))
```
markusthoemmes: Good one on deleting all the assets 👍

common/scala/src/main/scala/whisk/core/entity/ArtifactActivationStore.scala
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    res
+  }
+
+  def get(activationId: ActivationId)(implicit transid: TransactionId): Future[WhiskActivation] = {
+    WhiskActivation.get(artifactStore, DocId(activationId.asString))
+  }
+
+  def delete(activationId: ActivationId)(implicit transid: TransactionId,
```
markusthoemmes: Might want to add a comment saying that this is not used in production/performance relevant code. (since the get/delete cycle is kinda overheady)

common/scala/src/main/scala/whisk/core/entity/ActivationStore.scala
```diff
@@ -0,0 +1,67 @@
...
+import spray.json.JsObject
+import whisk.common.{Logging, TransactionId}
+import whisk.core.database.CacheChangeNotification
+import whisk.spi.Spi
+
+import scala.concurrent.Future
+
+trait ActivationStore {
+
+  def store(activation: WhiskActivation)(implicit transid: TransactionId,
```
markusthoemmes: Please add ScalaDoc to all the methods.

tests/src/test/scala/whisk/core/controller/test/ControllerTestCommon.scala
```diff
@@ -28,8 +28,7 @@ import org.scalatest.Matchers
 import common.StreamLogging
 import akka.http.scaladsl.testkit.ScalatestRouteTest
 import akka.http.scaladsl.testkit.RouteTestTimeout
-import spray.json.DefaultJsonProtocol
-import spray.json.JsString
+import spray.json.{DefaultJsonProtocol, JsObject, JsString}
```
markusthoemmes: Please import `spray.json._`

chetanmeh:
@dubee I had provided some feedback on the [DL mail. thread][1]. 

> In some cases we are seeing large activation responses when requested
by client (with lots of concurrent request) put pressure on Controller
and Invoker heaps. As in most cases the logic in controller or invoker
does not "interpret" the activation record it would be efficient to
have an API which can just stream the json content to output response.

> So if the `ActivationStore` provides a get by id method which accepts
a Sink[ByteString, Future[T]] like we have in
ArtifactStore#readAttachment then that can possibly be plugging to
HTTP response sink and thereby avoid buffering the whole json in
memory. This would reduce heap pressure for web actions and other
cases where lots of clients are concurrently requesting activation

Would it be possible to have a stream based API instead of one which returns a `Future[List[_]]`

[1]: https://lists.apache.org/thread.html/ee87b42d1f516e65bb1b291c509da667f8a2a33c311b1ee01e27922e@%3Cdev.openwhisk.apache.org%3E

markusthoemmes:
@chetanmeh that should indeed be possible, will require quite some work on top of this (i.e. the list would need to be rewritten while not buffering too much).

Rewriting a stream from CouchDB could be done with https://developer.lightbend.com/docs/alpakka/current/data-transformations/json.html#streaming-of-nested-structures for example.

In general, I see this as a subsequent PR though. I agree with the general idea, I believe we can do this separately though.

chetanmeh:
> Rewriting a stream from CouchDB could be done with https://developer.lightbend.com/docs/alpakka/current/data-transformations/json.html#streaming-of-nested-structures for example.

Interesting stuff @markusthoemmes ! I was also looking for a way to dump CouchDB export via bulk api and this would be very helpful.

> In general, I see this as a subsequent PR though. I agree with the general idea, I believe we can do this separately though.

Works for me. In general we should leverage streams more in our stores. Something we can look into going forward

dubee:
PG2 3237 🔵

common/scala/src/main/scala/whisk/core/entity/ActivationStore.scala
```diff
@@ -0,0 +1,126 @@
...
+   *
+   * @param namespace namespace to query
+   * @param name entity name to query
+   * @param skip number of activations to skip
+   * @param limit maximum number of activations to list
+   * @param includeDocs return document with each activation
+   * @param since timestamp to retrieve activations after
+   * @param upto timestamp to retrieve activations before
+   * @param transid transaction ID for request
+   * @return Future containing a List of activations in JSON format or List of WhiskActivation
```
chetanmeh: @dubee In which case it would be `JsObject` and not `WhiskActivation`?

markusthoemmes: @chetanmeh I think that's the general list interface currently, where the `Right` is with includeDocs and `Left` not. I don't like that `Either` (no pun intended). But it's due to the design of the ArtifactStore of today. An improvement is very welcome but imho not necessary in this PR.

ningyougang:
@markusthoemmes 

Currently, `ArtifactActivationStoreProvider` is for `couchdb`,
so do we have any plan write a new ActivationStoreProvider for `elasticsearch`, e.g. ElasticsearchActivationStoreProvider?

markusthoemmes:
@dubee can give you more information on plans here.

ningyougang:
### 1. Why need to write a new ActivationStoreProvider for `elasticsearch`?
Currently, `activations` are saved into `couchdb` using `ArtifactActivationStore`

couchdb has below disadvantage for saving activations
*  It is not convenience to  `delete big activations`(activation data is growing very fast, when disk is full, need to delete, but should do compression at first, otherwise, disk usage will increase) 
* Couchdb's search feature is not stronger than `elasticsearch`.

Since we provide `Provide an activation store SPI` in this patch, so i think it is necessary to implement a new ActivationStoreProvider  for `elasticsearch`

### 2.How to implement it?
* Implement this trait `ActivationStore`, for example:
```
class ElasticsearchActivationStore(actorSystem: ActorSystem, actorMaterializer: ActorMaterializer, logging: Logging) extends ActivationStore
```
* In class ElasticsearchActivationStore inner, we can use `https://github.com/sksamuel/elastic4s/` to implement it.

@markusthoemmes ,how do you think?

markusthoemmes:
@ningyougang sure it makes sense. @dubee is working on it already though if I'm not mistaken, hence I'd like his input here.

ningyougang:
@markusthoemmes @dubee 

I found the source codes here. https://github.com/dubee/openwhisk/commits/elasticsearch-activationstore

After review and tested in my local, the `ArtifactElasticSearchActivationStore` also store activation to couchdb(not store it to elasticsearch).

dubee:
@ningyougang, I am no longer using that branch. `es-as` is the new one.

ningyougang:
@dubee ,https://github.com/dubee/openwhisk/blob/es-as/common/scala/src/main/scala/whisk/core/database/ArtifactElasticSearchActivationStore.scala#L432
Regarding `ArtifactElasticSearchActivationStore`,
It seems when save activation, it will call super method, and save activation to couchdb.
but its query method is `query from elasticsearch`

So when use `ArtifactElasticSearchActivationStore`, how does `activation` save to elasticsearch?

PR #3620, opened 2018-05-07
Remove Eclipse license header and IBM copyright from a test file

daisy-ycguo:
Remove Eclipse license header and IBM copyright from a test file `tests/src/test/scala/common/Pair.java`

### DIFF
--- a/tests/src/test/scala/common/Pair.java
+++ b/tests/src/test/scala/common/Pair.java
@@ -15,16 +15,6 @@
  * limitations under the License.
  */
 
-/*******************************************************************************
- * Copyright (c) 2002 - 2006 IBM Corporation.
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v1.0
- * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
- *
- * Contributors:
- *     IBM Corporation - initial API and implementation
- *******************************************************************************/
 package common;
 
 import java.util.Iterator;


houshengbo:
@rabbah We have three Java files under tests/src/test/scala/common: Pair.java, TestUtils.java and WhiskProperties.java. Have we already had substitutes for them? Pair.java is used by TestUtils.java and WhiskProperties.java. Do we need to rewrite all of them?
We need to make sure every source code file is compliant with the Apache policy regarding headers. IBM copyright needs to be removed from the source code files. We can write the copyright inside NOTICE file if necessary.

mrutkows:
@daisy-ycguo @houshengbo This license header appears to indicate that this code (entire file) may have been copied from another source location.  I do not feel comfortable removing it without knowing the provenance of the source code.

mrutkows:
@daisy-ycguo it appears this file has existed in the public domain for some time.  For example, this file also exists in other projects such as:
[https://github.com/wala/WALA/blob/master/com.ibm.wala.util/src/com/ibm/wala/util/collections/Pair.java](https://github.com/wala/WALA/blob/master/com.ibm.wala.util/src/com/ibm/wala/util/collections/Pair.java)


We cannot remove the original license header; in fact we need to note it as an exception and likely remove the ASF header.

rabbah:
Indeed this came from WALA. We can replace all this code quickly in this repo although I don’t know how many other repos this will touch.

mrutkows:
@daisy-ycguo this file either needs to be replaced or removed from the release package/distribution if a suitable replacement cannot be found (since the files license is not compat. with Apache.).  Please open a tracking issue under the "release" repo. to track the removal/replacement and link this issue to it for history/commentary.

PR #3628, opened 2018-05-08
Revert "Replace WALA Pair with clean new Pair."

markusthoemmes:
Reverts apache/incubator-openwhisk#3624

This broke some downstream repositories. Will fix those first before merging this (soon!)

### DIFF
--- a/tests/src/test/scala/common/Pair.java
+++ b/tests/src/test/scala/common/Pair.java
@@ -15,14 +15,87 @@
  * limitations under the License.
  */
 
+/*******************************************************************************
+ * Copyright (c) 2002 - 2006 IBM Corporation.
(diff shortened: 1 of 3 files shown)


mrutkows:
@csantanapr Carlos, what do we need to do in IBM PG to fix this in order to merge in open?

csantanapr:
downstreams already fixed

csantanapr:
@mrutkows in this case was not about PG. 
it was that a common Class “common.Pair” was being removed from the common tests utilities. 

And there other repos like the package-cloudant repo that depended on it. 
In addition in the IBM downstream was also using this Class. 

What needed to happen with that PR was to communicate more broadly about the braking change and fix first the downstreams like the cloudant package first before removing from main repo.

Issue #3637, opened 2018-05-10
Recognize code as binary


KeonHee:
## Environment details:

* CentOS distributed environment
* Mac OS cli

## Steps to reproduce the issue:

1.   Create a `test.js` file below. 
```
functionmain
```

2. And create action using this file.
```
$ wsk action create test test.js
```

3. Get `test` action.
```
$ wsk action get test
```    

## Provide the expected results and outputs:
```
{
    "namespace": "geonhui",
    "name": "test",
    "version": "0.0.1",
    "exec": {
        "kind": "nodejs:6",
        "binary": false
    },
    "annotations": [
        {
            "key": "exec",
            "value": "nodejs:6"
        }
    ],
    "limits": {
        "timeout": 60000,
        "memory": 256,
        "logs": 10
    },
    "publish": false
}
```

## Provide the actual results and outputs:

The binary attribute is true.
```
{
    "namespace": "geonhui",
    "name": "test",
    "version": "0.0.1",
    "exec": {
        "kind": "nodejs:6",
        "binary": true
    },
    "annotations": [
        {
            "key": "exec",
            "value": "nodejs:6"
        }
    ],
    "limits": {
        "timeout": 60000,
        "memory": 256,
        "logs": 10
    },
    "publish": false
}
```

## Additional information you deem important:
* The regular expression below seems to recognize the string as binary.
* https://github.com/apache/incubator-openwhisk/blob/a2dcace3d4cbef737cb5875c758f7e5d2b6d9f3c/common/scala/src/main/scala/whisk/core/entity/Exec.scala#L335

PR #3644, opened 2018-05-11
Allow more special characters in EntityName.

markusthoemmes:
There's no reason not to allow these characters in EntityName, so let's not be narrower on our users than we need to be.

- [X] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


common/scala/src/main/scala/whisk/core/entity/EntityPath.scala
```diff
@@ -189,13 +189,15 @@ protected[core] class EntityName private (val name: String) extends AnyVal {
 
 protected[core] object EntityName {
   protected[core] val ENTITY_NAME_MAX_LENGTH = 256
+  protected[core] val allowedSpecialCharacters = "@.&%!?=#-"
```
rabbah: really?!?

csantanapr:
@markusthoemmes Is this a fix for #3637 ?

common/scala/src/main/scala/whisk/core/entity/EntityPath.scala
csantanapr: all this valid/safe in URL? and CouchDB `_id` ?

rabbah: https://github.com/apache/incubator-openwhisk/issues/2830

rather than more arbitrary characters - since this isn't the first time the list has gotten larger, how about allowing the largest set of characters instead.

markusthoemmes: proposing:

```
protected[core] val REGEX = raw"\A([a-zA-Z0-9]|[a-zA-Z0-9][^/]{0,${ENTITY_NAME_MAX_LENGTH - 2}}[^/ ])\z"
```

Will allow anything but "/" since that gets nasty in URLs.

rabbah: worth a dev list heads up.

markusthoemmes:
@rabbah pushed a change to allow all characters except "/" (to avoid ambiguous URLs).

Fixing the underscore problem mentioned is a lot more involved and is API breaking theory (since we'd disallow what we're allowing today). I'd delay that to a later change.

rabbah:
Thanks for weighing in. Should we factor the char set into a deployment configuration then? The set proposed first by the pr  "@.&%!?=#-" already includes ? for example.

cbickel:
@rabbah There are already some tests, to check for forbidden characters. Nevertheless, I've added some more, for more special characters.

@chetanmeh Thanks for your hint. I've changed the PR to the proposal of rodirc. Do you agree with this set of characters?

rabbah:
To clarify the set in my comment above was the original PR. It has to exclude ?# to not break other dbs. Maybe only & for now?

chetanmeh:
@cbickel Currently the regex is `\w@ .&-` (earlier `\w@ .-`). This should be fine. 

The scaladoc would need to be fixed though

cbickel:
PG5#319 is 🔵

Issue #3655, opened 2018-05-15
the availablePermits of an invokerSlot may exceed the maxAllowed value when number of controller changed


jiangpengcheng:
## Environment details:

* native CentOS 7.3.1611 with docker 17.05-ce
* 4 controllers and 9 invokers, with invokerThreShold=16 and numCore * coreShare=20

## Steps to reproduce the issue:

1. invoke a single action in a infinitely loop in every one second

action code is:
```js
function main() {
    sleep(50000)
    return {payload: 'Hello world'};
}

function sleep(d){
  for(var t = Date.now();Date.now() - t <= d;);
}
```

and the loop script is:

```shell
#!/bin/bash
while true
do
  wsk -i action invoke hello --result
  sleep 1s
done
```

as the `hello` action's execution time is about 50-51 seconds and it's invoked in every one second, there will be around 50 running containers in the cluster(make sure no other actions are invoked), which will make the home invoker(let's say it's invoker5) overloaded, while the cluster is still capable

2. stop a controller manually to simulate a failure and terminate the loop script in step 1

3. wait for all running actions finished, and then execute the loop script in step1 again

## Provide the expected results and outputs:

in the home invoker(invoker5), the total running containers should be:

```shell
# docker-runc list | grep running | wc -l
19
```
*19 = 16(running containers) + 2(prewarmed containers) + 1(invoker container)*

## Provide the actual results and outputs:

```
# docker-runc list | grep running | wc -l
23
```
*23 = 20(running containers) + 2(prewarmed containers) + 1(invoker container)*

## Explain
the reason is while cluster is changed, the `ShardingContainerPoolBalancer` will throw away the old `invokerSlots` and create a new sequence to replace it:

```scala
def updateCluster(newSize: Int): Unit = {
  val actualSize = newSize max 1 // if a cluster size < 1 is reported, falls back to a size of 1 (alone)
  if (_clusterSize != actualSize) {
    _clusterSize = actualSize
    val newTreshold = (totalInvokerThreshold / actualSize) max 1 // letting this fall below 1 doesn't make sense
    currentInvokerThreshold = newTreshold
    _invokerSlots = _invokerSlots.map(_ => new ForcableSemaphore(currentInvokerThreshold))

    logging.info(
      this,
      s"loadbalancer cluster size changed to $actualSize active nodes. invokerThreshold = $currentInvokerThreshold")(
      TransactionId.loadbalancer)
  }
}
```
[...]

PR #3656, opened 2018-05-15
Adjust the behavior when update controller cluster

jiangpengcheng:
Set the max permits for each invokerSlot instead of creating new ForcibleSemaphore. New ForcibleSemaphore will ignore acquired permits, so after these permits are released, the max permits of invokerSlot will exceed the `currentInvokerThreshold`, which is not what we want

## Description
in some case, the `availablePermits` may exceed the  `currentInvokerThreshold`, which may cause some invokers take more invoking requests than the number they can handle and other invokers are still idle

## Related issue and scope
- [x] I opened an issue to propose and discuss this change (#3655)

- [ ] API
- [ ] Controller
- [x] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/common/ForcibleSemaphore.scala
+++ b/common/scala/src/main/scala/whisk/common/ForcibleSemaphore.scala
@@ -85,6 +85,7 @@ class ForcibleSemaphore(maxAllowed: Int) {
   }
 
   val sync = new Sync
+  private var maxPermits = maxAllowed
 
   /**
    * Acquires the given numbers of permits.
@@ -121,4 +122,18 @@ class ForcibleSemaphore(maxAllowed: Int) {
 
   /** Returns the number of currently available permits. Possibly negative. */
   def availablePermits: Int = sync.permits
+
+  /** Set the max permits for sync. */
+  def setMaxPermits(newMaxPermits: Int): Unit = synchronized {
+    require(newMaxPermits > 0, "maxPermits cannot be negative")
+    newMaxPermits - maxPermits match {
+      case 0 =>
+      case delta if delta > 0 =>
+        sync.releaseShared(delta) // increase the max permits
+      case delta if delta < 0 =>
+        sync.forceAquireShared(-delta) // reduce the max permits
+    }
+
+    maxPermits = newMaxPermits
+  }
 }
--- a/tests/src/test/scala/whisk/common/ForcibleSemaphoreTests.scala
+++ b/tests/src/test/scala/whisk/common/ForcibleSemaphoreTests.scala
@@ -85,4 +85,18 @@ class ForcibleSemaphoreTests extends FlatSpec with Matchers {
       acquires should contain theSameElementsAs result
     }
   }
+
+  it should "set the max allowed permits dynamically" in {
+    val s = new ForcibleSemaphore(10)
+    s.tryAcquire(2) shouldBe true // 8 permits left
+
+    s.setMaxPermits(5) // reduce max permits
+    s.availablePermits shouldBe 3
+
+    s.setMaxPermits(10) // increase max permits
+    s.availablePermits shouldBe 8
+
+    s.setMaxPermits(10) // nothing changed
+    s.availablePermits shouldBe 8
+  }
 }
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
@@ -604,7 +604,7 @@ case class ShardingContainerPoolBalancerState(
         totalInvokerThreshold / actualSize
       }
       currentInvokerThreshold = newTreshold
-      _invokerSlots = _invokerSlots.map(_ => new ForcibleSemaphore(currentInvokerThreshold.toMB.toInt))
+      _invokerSlots.foreach(_.setMaxPermits(currentInvokerThreshold.toMB.toInt))
 
       logging.info(
         this,
--- a/tests/src/test/scala/whisk/core/loadBalancer/test/ShardingContainerPoolBalancerTests.scala
+++ b/tests/src/test/scala/whisk/core/loadBalancer/test/ShardingContainerPoolBalancerTests.scala
@@ -127,7 +127,8 @@ class ShardingContainerPoolBalancerTests extends FlatSpec with Matchers with Str
     state.invokerSlots.head.availablePermits shouldBe (memory - memoryPerSlot).toMB
 
     state.updateCluster(2)
-    state.invokerSlots.head.availablePermits shouldBe memory.toMB / 2 // state reset + divided by 2
+    // state reset + divided by 2 and minus the already acquired shared
+    state.invokerSlots.head.availablePermits shouldBe memory.toMB / 2 - memoryPerSlot.toMB.toInt
   }
 
   it should "fallback to a size of 1 (alone) if cluster size is < 1" in {


rabbah:
@jiangpengcheng this is plausible but I think not complete. If you exhaust all the slots, the current implementation will queue (over commits the semaphores) but your patch puts a hard max on the queue if i understood the change.

rabbah:
BTW do you mind fixing `ForcableSemaphore` -> `ForcibleSemaphore`

jiangpengcheng:
@rabbah, yes, I set the `currentInvokerThreshold` as the hard max, so the total slots for an invoker won't exceeded that value, normally this won't happen, but when the number of controllers is changed, things are different.

let's say we have 4 as the number of slot for each invoker, and there are 2 running actions in invokerN, then the controller cluster changed, and the `currentInvokerThreshold` is changed to 8, the `invokerSlots` will be recreated:

```
def updateCluster(newSize: Int): Unit = {
  val actualSize = newSize max 1 // if a cluster size < 1 is reported, falls back to a size of 1 (alone)
  if (_clusterSize != actualSize) {
    _clusterSize = actualSize
    val newTreshold = (totalInvokerThreshold / actualSize) max 1 // letting this fall below 1 doesn't make sense
    currentInvokerThreshold = newTreshold
    _invokerSlots = _invokerSlots.map(_ => new ForcableSemaphore(currentInvokerThreshold))

    logging.info(
      this,
      s"loadbalancer cluster size changed to $actualSize active nodes. invokerThreshold = $currentInvokerThreshold")(
      TransactionId.loadbalancer)
  }
}
```

so there are 8 free slots for invokerN

but there are still two running actions in invokerN, after they are finished, they will release the 2 slots they occupied before, then the number of free slot for invokerN will be changed to 10, which exceeded the `currentInvokerThreshold`, so I set a hard max value to prevent this.

common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
```diff
@@ -34,7 +34,7 @@ import scala.annotation.tailrec
  *
  * @param maxAllowed maximum number of permits given away by `tryAcquire`
  */
-class ForcableSemaphore(maxAllowed: Int) {
+class ForcableSemaphore(private var maxAllowed: Int) {
```
markusthoemmes: Does this need to be volatile now?

common/scala/src/main/scala/whisk/common/ForcableSemaphore.scala
```diff
@@ -121,4 +139,11 @@ class ForcableSemaphore(maxAllowed: Int) {
 
   /** Returns the number of currently available permits. Possibly negative. */
   def availablePermits: Int = sync.permits
+
+  /** Set the new maxAllowed value. */
+  def setMaxAllowed(newMaxAllowed: Int): Unit = {
```
markusthoemmes: Do we need to synchronize here? Especially since `forceChangeState` and resetting maxAllowed are non-atomic operations.

Can we reuse bits from here: http://blog.teamlazerbeez.com/2009/04/20/javas-semaphore-resizing/?

I guess your `forceChangeState` method is roughly the same as `reducePermits` in the original `Semaphore` impl. modulo you calculate an absolute number whereas they pass in a difference, so that should be fine.

jiangpengcheng: thanks! learned a lot

jiangpengcheng:
@markusthoemmes updated

common/scala/src/main/scala/whisk/common/ForcibleSemaphore.scala
```diff
@@ -121,4 +122,18 @@ class ForcibleSemaphore(maxAllowed: Int) {
...
+  /** Set the max permits for sync. */
+  def setMaxPermits(newMaxPermits: Int): Unit = synchronized {
+    require(newMaxPermits > 0, "maxPermits cannot be negative")
+    newMaxPermits - maxPermits match {
+      case 0 =>
+      case delta if delta > 0 =>
+        sync.releaseShared(delta) // increase the max permits
+      case delta if delta < 0 =>
+        sync.forceAquireShared(-delta) // reduce the max permits
+    }
```
markusthoemmes: It doesn't seem like a `match` block helps here.

```scala
val delta = newMaxPermits - maxPermits
if(delta > 0) {
  sync.releaseShared(delta)
} else if(delta < 0) {
  sync.forceAquireShared(-delta)
}
```

Reads a bit better imo?

tests/src/test/scala/whisk/common/ForcibleSemaphoreTests.scala
```diff
@@ -85,4 +85,18 @@ class ForcibleSemaphoreTests extends FlatSpec with Matchers {
...
+    s.tryAcquire(2) shouldBe true // 8 permits left
+
+    s.setMaxPermits(5) // reduce max permits
+    s.availablePermits shouldBe 3
+
+    s.setMaxPermits(10) // increase max permits
+    s.availablePermits shouldBe 8
+
+    s.setMaxPermits(10) // nothing changed
+    s.availablePermits shouldBe 8
```
tysonnorris: Please update the test to include:
* hitting the case where `if delta < 0 =>` (I think in the case where permits are exhausted, and then there is also a reduction in maxPermits?)
* concurrently executing tryAquire/forceAquire/release while setMaxPermits is in flight (I think that you will still end up with inaccurate permits, e.g. if release is invoked while setMaxPermits is incomplete)

markusthoemmes: Sounds valid, thanks for catching this!

jiangpengcheng: @tysonnorris 

- for the `if delta < 0 =>`, I think the `s.setMaxPermits(5) // reduce max permits` match the case

- for the concurrent situations, I added below tests

```
  it should "set the max allowed permits while call release concurrently" in {
    (0 until 100).foreach { _ =>
      val s = new ForcibleSemaphore(32)
      s.tryAcquire(32) // 0 permits left

      (0 until 64).par.map{ i =>
        if ( i == 16)
          s.setMaxPermits(16) // force acquire 16 permits
        else
          s.release()
      }

      s.availablePermits shouldBe 47 // 0 - 16 + 63
    }

    (0 until 100).foreach { _ =>
      val s = new ForcibleSemaphore(32)
      s.tryAcquire(32) // 0 permits left

      (0 until 64).par.map{ i =>
        if ( i == 16)
          s.setMaxPermits(64) // release 32 permits
        else
          s.release()
      }

      s.availablePermits shouldBe 95 // 0 + 32 + 63
    }
  }

  it should "set the max allowed permits while call forceAcquire concurrently" in {
    (0 until 100).foreach { _ =>
      val s = new ForcibleSemaphore(32)

      (0 until 64).par.map{ i =>
        if ( i == 16)
          s.setMaxPermits(16) // force acquire 16 permits
        else
          s.forceAcquire()
      }

      s.availablePermits shouldBe -47 // 32 - 16 - 63
    }
    (0 until 100).foreach { _ =>
      val s = new ForcibleSemaphore(32)

      (0 until 64).par.map{ i =>
        if ( i == 16)
          s.setMaxPermits(64) // release 32 permits
        else
          s.forceAcquire()
      }

      s.availablePermits shouldBe 1 // 32 + 32 - 63
    }
  }

  it should "set the max allowed permits while call tryAcquire concurrently" in {
    (0 until 100).foreach { _ =>
      val s = new ForcibleSemaphore(32)

      (0 until 64).par.map{ i =>
        if ( i == 16)
          s.setMaxPermits(16) // force acquire 16 permits
        else
          s.tryAcquire()
      }

      s.availablePermits shouldBe ??? // not sure how many tryAcquire succeed
    }

    (0 until 100).foreach { _ =>
      val s = new ForcibleSemaphore(32)

      (0 until 64).par.map{ i =>
        if ( i == 16)
          s.setMaxPermits(64) // release 32 permits
        else
          s.tryAcquire()
      }

      s.availablePermits shouldBe ??? // not sure how many tryAcquire succeed
    }
  }
```

the first two tests will succeed while the last one will not, I believe the reason is that the `nonFairTryAcquireShared` will not change the state while `available - acquires < 0`, so we can not make sure a `tryAcquire` is success or not while `setMaxPermits` is in flight, any advice?

```
    @tailrec
    final def nonFairTryAcquireShared(acquires: Int): Int = {
      val available = getState
      val remaining = available - acquires
      if (remaining < 0 || compareAndSetState(available, remaining)) {
        remaining
      } else {
        nonFairTryAcquireShared(acquires)
      }
    }
```

PR #3661, opened 2018-05-15
Restrict allowed namespaces when creating action of certain kinds

asteed:
## Description

Currently, all kinds supported by the system are able to be created within any namespace to which a client has access. As new action kinds are added to the system, there is a need to restrict the creation of new actions to a particular _whitelist_ of accepted namespaces.

## Related issue and scope

- [ ] API
- [X] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Actions.scala
@@ -187,8 +187,11 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with 
     parameter('overwrite ? false) { overwrite =>
       entity(as[WhiskActionPut]) { content =>
         val request = content.resolve(user.namespace)
+        val checkAdditionalPrivileges = entitleReferencedEntities(user, Privilege.READ, request.exec).flatMap {
+          case _ => entitlementProvider.check(user, content.exec)
+        }
 
-        onComplete(entitleReferencedEntities(user, Privilege.READ, request.exec)) {
+        onComplete(checkAdditionalPrivileges) {
           case Success(_) =>
             putEntity(WhiskAction, entityStore, entityName.toDocId, overwrite, update(user, request) _, () => {
               make(user, entityName, request)
--- a/core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
+++ b/core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
@@ -204,12 +204,12 @@ protected[core] abstract class EntitlementProvider(
   private val kindRestrictor = new KindRestrictor()
 
   /**
-    * Checks if an action's kind is restricted for a given subject
-    *
-    * @param user    the identity to check for restrictions
-    * @param exec    the action's executable details
-    * @return a promise that completes with success iff the user's action kind is not restricted
-    */
+   * Checks if an action's kind is restricted for a given subject
+   *
+   * @param user    the identity to check for restrictions
+   * @param exec    the action's executable details
+   * @return a promise that completes with success iff the user's action kind is not restricted
+   */
   protected[core] def check(user: Identity, exec: Option[Exec])(implicit transid: TransactionId): Future[Unit] = {
     exec
       .map {
--- a/core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
+++ b/core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
@@ -47,18 +47,21 @@ package object types {
  * Resource is a type that encapsulates details relevant to identify a specific resource.
  * It may be an entire collection, or an element in a collection.
  *
- * @param namespace the namespace the resource resides in
+ * @param namespace  the namespace the resource resides in
  * @param collection the collection (e.g., actions, triggers) identifying a resource
- * @param entity an optional entity name that identifies a specific item in the collection
- * @param env an optional environment to bind to the resource during an activation
+ * @param entity     an optional entity name that identifies a specific item in the collection
+ * @param env        an optional environment to bind to the resource during an activation
  */
 protected[core] case class Resource(namespace: EntityPath,
                                     collection: Collection,
                                     entity: Option[String],
                                     env: Option[Parameters] = None) {
   def parent: String = collection.path + EntityPath.PATHSEP + namespace
+
   def id: String = parent + entity.map(EntityPath.PATHSEP + _).getOrElse("")
+
   def fqname: String = namespace.asString + entity.map(EntityPath.PATHSEP + _).getOrElse("")
+
   override def toString: String = id
 }
 
@@ -92,13 +95,14 @@ protected[core] abstract class EntitlementProvider(
    * controllers
    */
   private def overcommit(clusterSize: Int) = if (clusterSize > 1) 1.2 else 1
+
   private def dilateLimit(limit: Int): Int = Math.ceil(limit.toDouble * overcommit(loadBalancer.clusterSize)).toInt
 
   /**
    * Calculates a possibly dilated limit relative to the current user.
    *
    * @param defaultLimit the default limit across the whole system
-   * @param user the user to apply that limit to
+   * @param user         the user to apply that limit to
    * @return a calculated limit
    */
   private def calculateLimit(defaultLimit: Int, overrideLimit: Identity => Option[Int])(user: Identity): Int = {
@@ -113,7 +117,7 @@ protected[core] abstract class EntitlementProvider(
    * limit, so it needs to be divided between the parties who want to perform that check.
    *
    * @param defaultLimit the default limit across the whole system
-   * @param user the user to apply that limit to
+   * @param user         the user to apply that limit to
    * @return a calculated limit
    */
   private def calculateIndividualLimit(defaultLimit: Int, overrideLimit: Identity => Option[Int])(
@@ -154,8 +158,8 @@ protected[core] abstract class EntitlementProvider(
   /**
    * Grants a subject the right to access a resources.
    *
-   * @param subject the subject to grant right to
-   * @param right the privilege to grant the subject
+   * @param user     the subject to grant right to
+   * @param right    the privilege to grant the subject
    * @param resource the resource to grant the subject access to
    * @return a promise that completes with true iff the subject is granted the right to access the requested resource
    */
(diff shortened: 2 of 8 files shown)


asteed:
Sample whitelist to be used in `common/scala/src/main/resources/application.confapplication.conf`
```
whisk {
    ...
    runtimes {
        ...
        whitelists = {
          "nodejs:6": ["guest"],
          python: ["supergroup"],
          blackbox: ["ADMIN"]
        }
    }
}
```

rabbah:
I can see how a need for this arises - the integration should be through the entitlement abstraction however, because I can imagine subjects and privileges are likely dynamically changing, and because this cleary is a privilege check for a subject operating on a resource.

asteed:
Is this a privilege check on a subject or on a resource only? 
The entirety of the check is that the particular target is supportive of housing the requested kind. 

My initial, albeit inexperienced concern with placing it within Entitlement is exposing the _action-specific_ notion of kinds within Entitlement. I'd be happy to know others' thoughts, however

csantanapr:
>there is a need to restrict the creation of new actions to a particular whitelist of accepted namespaces.


I don’t understand why there is a need to restrict?

Do you have a more real use case on this applicable ? private deployment, or public multi tenant, etc

rabbah:
My rationale is that this is just one example of an entitlement that can arise: longer limits on duration, higher memory reservation, longer sequences or compositions to name obvious ones. All of these should be addressed in a canonical way and fit logically as part of entitlement in my view. You are right to point out that the interface for entitlement is too narrow today because it ignores the kind or additional resource requirements I noted. 

I’d favor encapsulating these checks in a well defined interface vs integrating it with the action api handler. 

Further, note that the approach of white listing namespaces a done with a deployment configuration makes it more static whereas I view such lists as something that will be dynamic: the privileges should perhaps be part of the identity record (like other limits today) and we should provide an interface to black list and white list (or rather change the rights of the subject). 

I know this broadens scope much more than you’ve done. We can discuss the design and work toward that but allow a transient path to make progress in the meantime.

csantanapr:
So this issue about implementing identity (i.e. IAM) with scopes (i.e. oauth) and these pre-defined scopes associated with certain permissions or exclusions?

asteed:
@csantanapr here's are two more explicit use cases.
1) The OW system currently supports nodejs kinds in the system manifest. However, as a platform provider there is a need to periodically add support for additional kinds without allowing use across the entire system. Initially we want to restrict these kinds to certain namespaces which are owned and controlled by maintainers of the system.
2) We need to restrict the use of `blackbox` kinds to a namespace containing provider-owned actions which are usable by others within the system. Only actions in these namespaces should be allowed to be created as blackbox kinds.

Eventually, this could be handled at an IAM/ACL/oauth level. However, there is no such construct available today. The use cases do not require a per-subject set of ACLs for this particular PR.

I see two separate tasks here:
1) Support system-level configuration of how new functionality can be rolled out.
2) Support per-subject IAM and oauth for more granular permissions and entitlement.

asteed:
@rabbah @csantanapr I've made changes to move the check into Entitlement.

Any additional thoughts regarding my prior comment on separating these two issues?

core/controller/src/main/scala/whisk/core/entitlement/Entitlement.scala
```diff
@@ -208,8 +222,12 @@ protected[core] abstract class EntitlementProvider(
    * @param resource the resource the subject requests access to
    * @return a promise that completes with success iff the subject is permitted to access the requested resource
    */
-  protected[core] def check(user: Identity, right: Privilege, resource: Resource)(
-    implicit transid: TransactionId): Future[Unit] = check(user, right, Set(resource))
+  protected[core] def checkResource(
```
asteed: I ran into an issue with these two `check()` methods, in that from my research it appears only one method is allowed to contain a default parameter at a given position in the method's signature [0]

I was unable to find a way around it short of more explicitly naming them. I'm open to suggestions for handling this scenario.

[0] https://groups.google.com/forum/#!msg/scala-user/FyQK3-cqfaY/fXLHr8QsW_0J

rabbah:
I thought I’d mention that we do have a mechanism in place today related to quotas which is tied to the Identity. This is done by attachment a limits document to the subject. It has the advtange of being dynamically adaptable (eg wskadmin) vs a white list that’s configured once at deployment.

rabbah:
@asteed is this a `wip` or ready for review (modulo failing travis)?

core/controller/src/main/scala/whisk/core/entitlement/KindRestrictor.scala
```diff
@@ -0,0 +1,69 @@
...
+class KindRestrictor(allowedKinds: DefaultAllowedKinds = loadConfigOrThrow[DefaultAllowedKinds](ConfigKeys.runtimes))(
+  implicit logging: Logging) {
+
+  logging.info(this, s"allowedKinds = $allowedKinds")(TransactionId.controller)
+
+  private def execKindAllowed(exec: Option[Exec], kinds: Set[String]): Boolean = {
+    exec match {
+      case Some(e) if kinds.contains(e.kind) => true
+      case _                                 => false
+    }
```
markusthoemmes: This is equivalent to:

```scala
exec.map(e => kinds.contains(e.kind)).getOrElse(false)
```

In prose it says: If exec has a value, see if it is contained in the set and `get` that value. If exec has no value (`orElse`) default to false.

core/controller/src/main/scala/whisk/core/entitlement/KindRestrictor.scala
```diff
@@ -0,0 +1,69 @@
...
+      case None => {
+        // 2. Check for a default setting
+        allowedKinds.whitelist match {
+          case Some(defaultKinds) => execKindAllowed(exec, defaultKinds)
+          case None               =>
+            // 3. Fallback to allowing all system-accepted kinds
+            true
+        }
+      }
+    }
```
markusthoemmes: This is a very nice one :tada:, the equivalent with combinators would be:

```scala
user.limits.allowedKinds
  .orElse(allowedKinds.whitelist)
  .map(allowed => execKindAllowed(exec, allowed)
  .getOrElse(true)
```

It essentially says: If userlimits is defined, pass that value on, if not pass the whitelist on. If either of those two is defined, check if the kind is allowed and get that value. If not, default to true.

asteed:
Thanks @markusthoemmes, that definitely helps express the logic in a concise way

asteed:
@rabbah It looks like I'm unable to trigger another TravisCI run now that it's been addressed with a new commit. Any suggestions?

rabbah:
i kicked off travis again.

rabbah:
I rebased your PR and I kicked off travis again. You'll have to pull your branch from origin.

asteed:
Hi @markusthoemmes! Do you have any additional review comments or feedback that need to be addressed prior to merging this PR ?

rabbah:
I will take this one for review.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -118,15 +118,19 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with
   protected override def innerRoutes(user: Identity, ns: EntityPath)(implicit transid: TransactionId) = {
     (entityPrefix & entityOps & requestMethod) { (segment, m) =>
       entityname(segment) { outername =>
-        pathEnd {
+        (pathEnd & put & entity(as[WhiskActionPut])) { (content) =>
```
rabbah: @asteed can you add a comment explaining why you're short-circuiting the main dispatcher?

asteed: @rabbah the issue at hand was to be able to extract the request payload (specifically for the `kind` property) and marshall it into `entitlementProvider.checkResource` so that it can be used to determine the kind at the time of action creation. I'd appreciate any guidance if you know of any alternatives to this approach

core/controller/src/main/scala/whisk/core/entitlement/KindRestrictor.scala
```diff
@@ -0,0 +1,60 @@
...
+  logging.info(this, s"allowedKinds = $allowedKinds")(TransactionId.controller)
+
+  private def execKindAllowed(exec: Option[Exec], kinds: Set[String]): Boolean = {
+    exec.map(e => kinds.contains(e.kind)).getOrElse(false)
+  }
+
+  /**
+   * Checks whether a given kind is allowed for a given user.
+   */
+  def check(user: Identity, exec: Option[Exec])(implicit tid: TransactionId): Boolean = {
```
rabbah: the whitelist != the list of runtime kinds specified in the manifest?

rabbah: I believe the intent here to entitle a subject to a runtime if:
  1. they are explicitly granted rights to it in their identity record, or
  2. the runtime is whitelisted

The implementation assumes if there are no white listed runtimes (ie == `None`), then all runtimes are white listed.
Is that a correct understanding of the intent?

asteed: Yes, that is correct. And similarly, if an empty set is provided as a whitelist, it means access to no runtimes unless explicitly set for a subject

tests/src/test/scala/whisk/core/controller/test/EntitlementProviderTests.scala
```diff
@@ -681,4 +689,25 @@ class EntitlementProviderTests extends ControllerTestCommon with ScalaFutures {
...
   }
+
+  it should "restrict access to disallowed action kinds for a subject" in {
+    implicit val tid = transid()
+    implicit val ep = entitlementProvider
+    val subject = WhiskAuthHelpers.newIdentity().copy(limits = UserLimits(allowedKinds = Some(allowedKinds)))
+
+    disallowedKinds.foreach(k =>
+      intercept[RejectRequest] {
+        Await.result(ep.check(subject, Some(getExec(k))), 1.seconds)
```
rabbah: Can you add a check on the reject request error message as well?

asteed: Good catch. I've added the message checker

PR #3671, opened 2018-05-18
Activation id in header

### LINKED ISSUE #3582: [Feature Request] Include activation ID in web action header.
## Environment details:

* Any

1.  Invoke a web action.

```javascript
function main(params) { 
    return { response: params };
}
```

```bash
curl -X GET 'localhost/api/v1/web/myNamespace/hello.json'
{
  "response": {
    "__ow_method": "get",
    "__ow_headers": {
      "accept": "application/json, text/plain, */*",
      "activation_id": "ebc60b3553c14046860b3553c1c0464a"  # This is required
       .
       .
       .
    },
    "_": "1524725954543",
    "__ow_path": "",
  }
}
```

```
curl -X GET 'localhost/api/v1/web/myNamespace/hello.json'
{
  "response": {
    "__ow_method": "get",
    "__ow_headers": {
      "accept": "application/json, text/plain, */*",
       .
       .
       .
    },
    "_": "1524725954543",
    "__ow_path": "",
  }
}
```

Currently activation ID is not passed when the action is a web action.
When user want to retrieve more detail information about the activation, it's not easy to pick the right one from `activation list` API if there are subsequent action invocation calls.
So it would be great if server returns activation ID in headers or via any other ways.

style95:
This closes #3582 

## Description

This change will add activation id in the response headers.
Since same logic needs to be used in both collection API and web action API, I created dedicated trait for custom headers.

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3582)

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Actions.scala
@@ -256,24 +256,28 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with 
     onComplete(invokeAction(user, actionWithMergedParams, payload, waitForResponse, cause = None)) {
       case Success(Left(activationId)) =>
         // non-blocking invoke or blocking invoke which got queued instead
-        complete(Accepted, activationId.toJsObject)
+        respondWithActivationIdHeader(activationId) {
+          complete(Accepted, activationId.toJsObject)
+        }
       case Success(Right(activation)) =>
         val response = if (result) activation.resultAsJson else activation.toExtendedJson
 
-        if (activation.response.isSuccess) {
-          complete(OK, response)
-        } else if (activation.response.isApplicationError) {
-          // actions that result is ApplicationError status are considered a 'success'
-          // and will have an 'error' property in the result - the HTTP status is OK
-          // and clients must check the response status if it exists
-          // NOTE: response status will not exist in the JSON object if ?result == true
-          // and instead clients must check if 'error' is in the JSON
-          // PRESERVING OLD BEHAVIOR and will address defect in separate change
-          complete(BadGateway, response)
-        } else if (activation.response.isContainerError) {
-          complete(BadGateway, response)
-        } else {
-          complete(InternalServerError, response)
+        respondWithActivationIdHeader(activation.activationId) {
+          if (activation.response.isSuccess) {
+            complete(OK, response)
+          } else if (activation.response.isApplicationError) {
+            // actions that result is ApplicationError status are considered a 'success'
+            // and will have an 'error' property in the result - the HTTP status is OK
+            // and clients must check the response status if it exists
+            // NOTE: response status will not exist in the JSON object if ?result == true
+            // and instead clients must check if 'error' is in the JSON
+            // PRESERVING OLD BEHAVIOR and will address defect in separate change
+            complete(BadGateway, response)
+          } else if (activation.response.isContainerError) {
+            complete(BadGateway, response)
+          } else {
+            complete(InternalServerError, response)
+          }
         }
       case Failure(t: RecordTooLargeException) =>
         logging.debug(this, s"[POST] action payload was too large")
--- a/core/controller/src/main/scala/whisk/core/controller/Entities.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Entities.scala
@@ -66,7 +66,7 @@ protected[controller] trait ValidateRequestSize extends Directives {
 }
 
 protected trait CustomHeaders extends Directives {
-  val ActivationIdHeader = "X-OpenWhisk-Activation-Id"
+  val ActivationIdHeader = "x-openwhisk-activation-id"
 
   /** Add activation ID in headers */
   protected def respondWithActivationIdHeader(activationId: ActivationId): Directive0 = {
--- a/core/controller/src/main/scala/whisk/core/controller/Entities.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Entities.scala
@@ -20,9 +20,9 @@ package whisk.core.controller
 import scala.concurrent.Future
 import scala.language.postfixOps
 import scala.util.Try
-
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.model.StatusCodes.RequestEntityTooLarge
+import akka.http.scaladsl.model.headers.RawHeader
 import akka.http.scaladsl.server.Directive0
 import akka.http.scaladsl.server.Directives
 import akka.http.scaladsl.server.RequestContext
@@ -65,6 +65,15 @@ protected[controller] trait ValidateRequestSize extends Directives {
   protected val fieldDescriptionForSizeError = "Request"
 }
 
+protected trait CustomHeaders extends Directives {
+  val ActivationIdHeader = "X-OpenWhisk-Activation-Id"
+
+  /** Add activation ID in headers */
+  protected def respondWithActivationIdHeader(activationId: ActivationId): Directive0 = {
+    respondWithHeader(RawHeader(ActivationIdHeader, activationId.asString))
+  }
+}
+
 /** A trait implementing the basic operations on WhiskEntities in support of the various APIs. */
 trait WhiskCollectionAPI
     extends Directives
@@ -72,8 +81,8 @@ trait WhiskCollectionAPI
     with AuthorizedRouteProvider
     with ValidateRequestSize
     with ReadOps
-    with WriteOps {
-
+    with WriteOps
+    with CustomHeaders {
   /** The core collections require backend services to be injected in this trait. */
   services: WhiskServices =>
 
--- a/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/WebActions.scala
@@ -353,7 +353,7 @@ protected[core] object WhiskWebActionsApi extends Directives {
     headers.filter(_.lowercaseName != `Content-Type`.lowercaseName)
 }
 
-trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostActionActivation {
+trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostActionActivation with CustomHeaders {
   services: WhiskServices =>
 
   /** API path invocation path for posting activations directly through the host. */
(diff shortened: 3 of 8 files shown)


core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -257,24 +256,31 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with
     onComplete(invokeAction(user, actionWithMergedParams, payload, waitForResponse, cause = None)) {
       case Success(Left(activationId)) =>
         // non-blocking invoke or blocking invoke which got queued instead
-        complete(Accepted, activationId.toJsObject)
+
+        respondWithActivationIdHeader(activationId) {
+          // note that if header defined a content-type, it will be ignored
+          // since the type must be compatible with the data response
```
rabbah: I don’t think this comment makes sense here.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -634,37 +634,40 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
                               responseType: MediaExtension)(implicit transid: TransactionId) = {
     onComplete(queuedActivation) {
       case Success(Right(activation)) =>
-        val result = activation.resultAsJson
+        respondWithActivationIdHeader(activation) {
```
rabbah: I do not think the activation id should be reported in webaction. It leaks information. For normal POSTS it is ok because there’s an explicit subject key. For webactions, there is not.

style95: This is to cope with the inconvenience which is described in #3582.
This will give users a better option to retrieve right activation results.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
style95: removed.

style95:
@rabbah 
Activation id header checking is added in existing tests.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
rabbah: A webaction is called by an anonymous subject. The activation id is information specific the subject owning the action.

The webaction can return its own header to identify the action. Which could be the activation id if it chooses - I do not think the system should blanket return the activation id for webactions.

core/controller/src/main/scala/whisk/core/controller/Entities.scala
```diff
@@ -65,14 +65,25 @@ protected[controller] trait ValidateRequestSize extends Directives {
   protected val fieldDescriptionForSizeError = "Request"
 }
 
+protected trait CustomHeaders extends Directives {
+
+  /** Add activation ID in headers */
+  protected def respondWithActivationIdHeader(activation: WhiskActivation): Directive0 =
```
cbickel: Do we need to override this method?
Wouldn't it be easier to always pass the activationId? You have it anyway, if you have the activation.

ddragosd:
Given that the API Gateway sets the `X-Request-Id` as per #3199 ,   wouldn't it be easier to have the Gateway add it to the response header as well ?

rabbah:
@ddragosd the request id isn't the activation id though --- although I raised the question in the related issue: should the two be the same in which case, the gateway or nginx generates the id and adds the header on the response.

ddragosd:
> should the two be the same

Any reason they can't be the same ?

ddragosd:
makes sense @rabbah . I'd say we can safely assume that the Gateway generates unique activation IDs.

core/controller/src/main/scala/whisk/core/controller/Entities.scala
style95: Updated accordingly

tests/src/test/scala/whisk/core/controller/test/ControllerTestCommon.scala
```diff
@@ -91,6 +91,8 @@ protected trait ControllerTestCommon
   val authStore = WhiskAuthStore.datastore()
   val logStore = SpiLoader.get[LogStoreProvider].logStore(actorSystem)
 
+  val ActivationIdHeaderInLowercase = "x-openwhisk-activation-id"
```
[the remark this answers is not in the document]
style95: Updated accordingly.

Regarding case, I followed scala constant naming convention(Upper camel case).
Is it OW convention(start with a lowercase letter)?

cbickel:
@rabbah @ddragosd 
I also thought about making tid and activationid the same, but I see the following problems:
- If a trigger is invoked, you have a trigger activation and several action activations. Which of these activations should get the tid as activation id
- Same for sequences. Which of the actions should get the tid as activation-id.
- One of the ideas of sending the tid to the controller, was to generate the tid in one of the providers and pass it with the request. With this ability you would be able to trace a request from end to end. But this also adds the risk, that it could not be unique anymore.

But maybe there is an easy solution for all of these concerns ;)

rabbah:
For the first parts:

For a trigger, the request id is the triggers activation id. Every activated action generates a new activation id as we do today. 

For a sequence the outermost action (the sequence itself) gets the id all the subcomponents get new ids (as today). 

Same for conductor actions.

The uniqueness is the part I’m concerned about.

style95:
Let me recap the intermediate conclusion.

1. Activation ID should not be exposed by the system instead, user can decide to whether expose it or not using existing activation id parameter in case the action is a webaction.
2. Use request ID generated by nginx or gateway as Activation ID as well.


Regarding 1, I initially raised this issue in the aspect of UI. When providing the functionality to display  activation results and logs after invocation of an action in UI, it's not easy to select right activation when the action is webaction.

In case of normal action also, activation logs are collected asynchronously, it's not that simple.
But anyway we can look it up because it returns activation ID in the response.

For webactions, it's not easy to choose a specific activation, and display logs for web actions without system support.

Any suggestions for this?

With regard to 2, so request ID would be Transaction ID and Activation ID.
More precisely Transaction id will be same with activation id. Would it be ok?

I just want to confirm.
As @ddragosd  said, if we can safely assume it is unique, would it be ok for this PR to take that in?

rabbah:
@cbickel does https://github.com/apache/incubator-openwhisk/pull/3671#issuecomment-392007532 address your concerns?

if we adopt the nginx generated id as the activation id, then this pr becomes a lot simpler.
we can also remove the double headers currently generated by the controller (and only emit it if an id isn't received).

cbickel:
@rabbah 
Not completely. I agree, that this would be the best solution, but the problem is, that on firing a trigger, the controller is calling itself with the transaction-Id.
To my knowledge, the controller doesn't recognise today, that the action is part of a rule or a normal action. Or am I wrong?
And the uniqueness is still a problem. If the tids are not unique, it would be bad for debugging, but if activation-Ids are not unique anymore, users wont't get activation-records anymore.

tests/src/test/scala/whisk/core/controller/test/ActionsApiTests.scala
```diff
@@ -1043,13 +1043,15 @@ class ActionsApiTests extends ControllerTestCommon with WhiskActionsApi {
       status should be(Accepted)
       val response = responseAs[JsObject]
       response.fields("activationId") should not be None
+      headers.exists(_.is(ActivationIdHeaderInLowercase)) should be(true)
```
markusthoemmes: headers are immutable and per the line above you know the correct activation id. You should be able to rewrite this to something like:

```scala
headers should contain(RawHeader(ActivationIdHeader, response.fields("activationId").convertTo[String]))
```

That'll give you a more explanatory response in tests. WDYT?

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -932,7 +934,10 @@ trait WebActionsApiBaseTests extends ControllerTestCommon with BeforeAndAfterEac
 
             m(s"$testRoutePath/$path") ~> Route.seal(routes(creds)) ~> check {
               status should be(statusCode)
-              headers shouldBe List(RawHeader("Set-Cookie", "a=b"))
+              headers.exists(header => {
+                header.is("set-cookie") && header.value() == "a=b"
+              }) shouldBe true
+              headers.exists(_.is(ActivationIdHeaderInLowercase)) shouldBe true
```
markusthoemmes: See comment above, please rewrite both of these into "contains" checks.

tests/src/test/scala/whisk/core/controller/test/ActionsApiTests.scala
```diff
@@ -1153,13 +1159,16 @@ class ActionsApiTests extends ControllerTestCommon with WhiskActionsApi {
         status should be(OK)
         val response = responseAs[JsObject]
         response should be(activation.withoutLogs.toExtendedJson)
+        headers.exists(_.is(ActivationIdHeaderInLowercase)) should be(true)
       }
 
       // repeat invoke, get only result back
       Post(s"$collectionPath/${action.name}?blocking=true&result=true") ~> Route.seal(routes(creds)) ~> check {
+        headers
```
markusthoemmes: bad push?

markusthoemmes:
Is there something actionable in here? Should we take the discussion out to the dev-list? At least the bit on generating activationIds from something outside of the controller seems worth discussing in a broader audience.

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
style95: got it

style95:
I think the point is, if we allow activationID generation in multiple components(nginx, controller), we cannot guarantee the activationID is unique(though the possibility of duplicate activationID is less.)

To use the same ID for `TransactionID` and `ActivationID`, it should be generated in nginx.
But also, ActivationID should be generated in controllers as a trigger is fired by controllers.

style95:
@rabbah 
Regarding exposure of ActivationID of webaction in response header, how about having a configuration to control it by OW operator?
In some cases, it would be worth to expose it though it could be a kind of small data breaches.

rabbah:
I think I've changed perspective on the activation id in the header response of web actions, after some further consideration. I can explain in a longer comment later.

style95:
How about separating ActivationID addition in headers and using same ID for `TransactionID` and `ActivationID`?
I think there is not big correleation between the two.

How about taking these changes and discussing the second one in dev-list or another PR separately?

tests/src/test/scala/whisk/core/controller/test/ControllerTestCommon.scala
```diff
@@ -89,6 +89,8 @@ protected trait ControllerTestCommon
   val logStore = SpiLoader.get[LogStoreProvider].instance(actorSystem)
   val activationStore = SpiLoader.get[ActivationStoreProvider].instance(actorSystem, materializer, logging)
 
+  val ActivationIdHeaderInLowercase = ActivationIdHeader.toLowerCase
+
```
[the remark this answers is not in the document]
style95: I think it would be better to keep it because I restored the previous `exist` logic.
Unlike a normal action, we are unable to get the value of `ActivationID` from the response as so unable to use `contain` logic that Markus suggested.

```scala
headers should contain(RawHeader(ActivationIdHeader, _)) // does not work
```

style95:
Travis build has passed.

tests/src/test/scala/whisk/core/controller/test/ControllerTestCommon.scala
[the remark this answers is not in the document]
markusthoemmes: I'd propose to make the global setting lowercased then, and use that throughout. WDYT?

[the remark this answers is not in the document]
style95: Sorry, what does global setting mean?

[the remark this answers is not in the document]
style95: The value in ‘CustomHeaders’?

style95:
Updated accordingly

core/controller/src/main/scala/whisk/core/controller/Entities.scala
```diff
@@ -65,15 +65,24 @@ protected[controller] trait ValidateRequestSize extends Directives {
   protected val fieldDescriptionForSizeError = "Request"
 }
 
+protected trait CustomHeaders extends Directives {
+  val ActivationIdHeader = "x-openwhisk-activation-id"
```
rabbah: I think @mhenke1 introduced a header with X-OW as the prefix. API gateway uses x-openwhisk. We should normalize and collocate these. Fwiw nginx also generates x-request-id header.

style95: @rabbah Agree.
How about using x-openwhisk?
Since x- header is a custom header specific to the application, it would be better to use the detailed one.

style95: I am not sure changes to use the same OpenWhisk prefix for all custom headers should be included in this PR.

style95:
Any comments on this?

rabbah:
I think we should still pursue id consolidation - I don't think we need both an activation id and a transaction id for top level activations, and we should further then reduce the number of headers we're communicating back (it would just be one id that means both). Deferring to another discussion/PR.

rabbah:
I used this today. 💪

style95:
Sounds great to me :)

Issue #3477, opened 2018-03-21
Support extending the environment variables for docker in ansible 


chetanmeh:
Currently the ansible deploy scripts for invoker and controller have a specific enumerated set of environment variables configured. For enabling usage of non default ArtifactStore and other aspects it should be possible to have a way to plugin a custom set of properties based on some environment setting.

PR #3689, opened 2018-05-23
Enable extending environment variables of Controller and Invoker

chetanmeh:
Provides a way to extend the environment variable set for Controller and Invoker

## Description

As mentioned in #3477 for some set of deployments like those using CosmosDB we need to add some extra environment variables like `CONFIG_whisk_spi_ArtifactStoreProvider`. To support that ansible playbook now supports add a set of environment variables conditionally.

### Usage

To add extra environment variables user need to define them under `group_vars/all` for specific environment

```yaml
cosmos:
    CONFIG_whisk_spi_ArtifactStoreProvider: whisk.core.database.cosmosdb.CosmosDBArtifactStoreProvider
    CONFIG_whisk_cosmosdb_endpoint : "https://<account name>.documents.azure.com:443/"
    CONFIG_whisk_cosmosdb_key : "some secret"
    CONFIG_whisk_cosmosdb_db : openwhisk
```

Here `cosmos` is the name of the environment variable set related to CosmosDB. To enable this set user needs to set an environment vairable

```
OPENWHISK_EXTRA_ENV=cosmos
```

Here `OPENWHISK_EXTRA_ENV` refers to a comma separated list of environment variable set which need to be added. If configured then all variables defined under `cosmos` key would be added to env of controller and invoker

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3477)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [x] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


ansible/roles/controller/tasks/deploy.yml
```diff
@@ -236,6 +236,14 @@
 
       "CONFIG_whisk_transactions_header": "{{ transactions.header }}"
 
+- name: merge extra env variables
+  set_fact:
+    controller_env: "{{ controller_env | default({}) | combine(item) }}"
+  with_items: "{{ extraEnvSets.split(',') | map('trim') | map('extract', vars)| list }}"
+  loop_control:
+    label: redacted
```
chetanmeh: Ansible would by default emit the item value to stdout. As the values may contain secrets set the label to a constant

ansible/environments/local/group_vars/all
```diff
@@ -2,6 +2,7 @@
 # license agreements; and to You under the Apache License, Version 2.0.
 
 whisk_version_name: local
+extraEnvSets: "{{ lookup('env', 'OPENWHISK_EXTRA_ENV_SETS')|default(false, true) }}"
```
csantanapr: We should not assume that is one top level set and force on both controller and invoker
Maybe user wants to set certain config or secret that is intended for controller but not invoker.
Or you want to set the same variable ie apikey to the db and is different for both controller and invoker.

What about having namespace with option to override
```
invoker:
   extraEnv: "{{ invoker_extra_env | lookup('env', 'OPENWHISK_EXTRA_ENV') | default(false, true) }}"

controller:
   extraEnv: "{{ controller_extra_env | lookup('env', 'OPENWHISK_EXTRA_ENV') | default(false, true) }}"
```

chetanmeh: Makes sense. Update PR to support specific env for controller and invoker. By default they both rely on `OPENWHISK_EXTRA_ENV`. However one can change the setting in environment specific file.

Not sure on best practices for placing the vars in ansible ... for now I have added them to `local/group_vars/all` such that one can use a different setting in specific env

```
ansible/environments/custom/group_vars
controller:
   extraEnv: "cosmos,splunk"
```

csantanapr:
Thanks @chetanmeh for the changes.
PG1 2954 ⏳

csantanapr:
Hey @chetanmeh please review the commit I just pushed https://github.com/apache/incubator-openwhisk/pull/3689/commits/fd358716d3ba7ba20960ed9b802a142d67345569
It's was what I was proposing we do is much simpler, it allows override from an environment.
Also it does proper combine using `dict` in ansible

Now you are able to edit your own environment like `environment/local` and add something like
```
controller_extraEnv:
  CONFIG_whisk_cosmosdb_endpoint : "https://<account name>.documents.azure.com:443/"
  CONFIG_whisk_cosmosdb_key : "some secret"
  CONFIG_whisk_cosmosdb_db : openwhisk
invoker_extraEnv:
  CONFIG_whisk_cosmosdb_endpoint : "https://<account name>.documents.azure.com:443/"
  CONFIG_whisk_cosmosdb_key : "some secret"
  CONFIG_whisk_cosmosdb_db : openwhisk
```
You can do it like this ^^ or in your environment load an environment variable into the environment file, or override using `ansible-playbook -e`

chetanmeh:
@csantanapr  I needed this support for travis specially. There I want to have a way where just with a env variable switch I can make test run against Cosmos. So earlier plan was to define pre configured env set in local


```yaml
cosmos:
    CONFIG_whisk_spi_ArtifactStoreProvider: whisk.core.database.cosmosdb.CosmosDBArtifactStoreProvider
    CONFIG_whisk_cosmosdb_endpoint : {{ lookup('env','COSMOSDB_ENDPOINT') }}
    CONFIG_whisk_cosmosdb_key : "{{ lookup('env','COSMOSDB_KEY') }}"
    CONFIG_whisk_cosmosdb_db : "{{ lookup('env','COSMOSDB_KEY') }}"
```

These env variables would then be configured via travis ui. And then in cosmos db stage of travis set
```
export OPENWHISK_EXTRA_ENV=cosmos
```

This way with minimal changes specific to  Cosmos I can run test against it. Any pointers on how to achieve that with approach you proposed?

csantanapr:
PG 3 2312 🏃

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -236,6 +236,11 @@
 
       "CONFIG_whisk_transactions_header": "{{ transactions.header }}"
 
+- name: merge extra env variables
+  set_fact:
+    controller_env: "{{ controller_env | default({}) | combine(controller.extraEnv) }}"
+  when: controller.extraEnv | default(false)
```
markusthoemmes: Why is the `when` needed in the first place? Shouldn't we be able to specify: `extraEnv: "{{ controller_extraEnv | default({}) }}"` and run the combine step no matter what?

csantanapr: I assumed we need it the extra safety, your proposing just dropping the `when` line?
 since in group_vars we already default to `default({})` and extra check is not need it.

csantanapr: I see define the default and then drop the `when` I will give this a test and update

csantanapr:
PG1 2974 ⌛️

csantanapr:
PG 1 2976 🔵

PR #3694, opened 2018-05-23
Add retries to CLI test framework for network errors

dubee:
Allows the CLI test framework to perform retries when a network error occurs.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/tests/src/test/scala/common/TestUtils.java
+++ b/tests/src/test/scala/common/TestUtils.java
@@ -54,11 +54,12 @@ import junit.runner.Version;
 public class TestUtils {
     protected static final Logger logger = Logger.getLogger("basic");
 
-    public static final int SUCCESS_EXIT    = 0;
-    public static final int ERROR_EXIT      = 1;
-    public static final int MISUSE_EXIT     = 2;
-    public static final int DONTCARE_EXIT   = -1;       // any value is ok
-    public static final int ANY_ERROR_EXIT  = -2;       // any non-zero value is ok
+    public static final int SUCCESS_EXIT        = 0;
+    public static final int ERROR_EXIT          = 1;
+    public static final int MISUSE_EXIT         = 2;
+    public static final int DONTCARE_EXIT       = -1;       // any value is ok
+    public static final int ANY_ERROR_EXIT      = -2;       // any non-zero value is ok
+    public static final int NETWORK_ERROR_EXIT  = 3;
 
     public static final int ACCEPTED        = 202;      // 202
     public static final int BAD_REQUEST     = 144;      // 400 - 256 = 144
--- a/tests/src/test/scala/common/Wsk.scala
+++ b/tests/src/test/scala/common/Wsk.scala
@@ -1036,13 +1036,26 @@ trait RunWskCmd extends BaseRunWsk {
     val args = baseCommand
     if (verbose) args += "--verbose"
     if (showCmd) println(args.mkString(" ") + " " + params.mkString(" "))
-    val rr = TestUtils.runCmd(
-      DONTCARE_EXIT,
-      workingDir,
-      TestUtils.logger,
-      sys.env ++ env,
-      stdinFile.getOrElse(null),
-      args ++ params: _*)
+    val rr: RunResult =
+      retry(
+        {
+          val rr = TestUtils.runCmd(
+            DONTCARE_EXIT,
+            workingDir,
+            TestUtils.logger,
+            sys.env ++ env,
+            stdinFile.getOrElse(null),
+            args ++ params: _*)
+
+          if (rr.exitCode == NETWORK_ERROR_EXIT) {
+            println(rr)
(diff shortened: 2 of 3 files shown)


tests/src/test/scala/common/TestUtils.java
```diff
@@ -54,11 +54,12 @@
...
-    public static final int ERROR_EXIT      = 1;
-    public static final int MISUSE_EXIT     = 2;
-    public static final int DONTCARE_EXIT   = -1;       // any value is ok
-    public static final int ANY_ERROR_EXIT  = -2;       // any non-zero value is ok
+    public static final int SUCCESS_EXIT        = 0;
+    public static final int ERROR_EXIT          = 1;
+    public static final int MISUSE_EXIT         = 2;
+    public static final int DONTCARE_EXIT       = -1;       // any value is ok
+    public static final int ANY_ERROR_EXIT      = -2;       // any non-zero value is ok
+    public static final int NETWORK_ERROR_EXIT  = 3;
```
markusthoemmes: Which errors are part of this? Does the CLI return a "3" on network issues as a convention?

markusthoemmes:
General existential question: I'm unclear if this changes any semantics in our tests. The main issue I could see (and I'm not sure if its really a problem) is that our requests are not necessarily idempotent. That is: A connection error might happen at any stage of the connection so you'll never know if it's safe to retry.

This might still make things better but it could certainly harm debuggability in tests. It might make sense to add a logline stating it's retrying because of a network issue to be able to tell later if an inconsistency was caused by that. Thoughts?

tests/src/test/scala/common/TestUtils.java
mdeuser: yes, when detected.
https://github.com/apache/incubator-openwhisk-client-go/blob/df32dca4aecca4ec30add11ce89eac72928f95f4/whisk/wskerror.go#L22

tests/src/test/scala/common/Wsk.scala
```diff
@@ -1036,13 +1036,26 @@ trait RunWskCmd extends BaseRunWsk {
...
+          val rr = TestUtils.runCmd(
+            DONTCARE_EXIT,
+            workingDir,
+            TestUtils.logger,
+            sys.env ++ env,
+            stdinFile.getOrElse(null),
+            args ++ params: _*)
+
+          if (expectedExitCode != NETWORK_ERROR_EXIT && rr.exitCode == NETWORK_ERROR_EXIT) {
+            println("A network error occurred.")
```
mdeuser: maybe add a bit more context to this output line... i.e. that the cli will be retried.. maybe the retry count if available..

tests/src/test/scala/common/Wsk.scala
```diff
@@ -1036,13 +1036,28 @@ trait RunWskCmd extends BaseRunWsk {
...
+          val rr = TestUtils.runCmd(
+            DONTCARE_EXIT,
+            workingDir,
+            TestUtils.logger,
+            sys.env ++ env,
+            stdinFile.getOrElse(null),
+            args ++ params: _*)
+
+          if (expectedExitCode != NETWORK_ERROR_EXIT) {
+            withClue(hideStr(reportFailure(args ++ params, expectedExitCode, rr).toString(), hideFromOutput)) {
```
mdeuser: is the `withClue` line still needed now that the `retry` is displaying the failure message?

dubee: Yes, `withClue` will print the stack trace if the final retry fails.

mdeuser: will each retry will have a `withClue` stack trace message, followed by a final `withClue` message (line 1062) if all retries fail?

dubee: Just a final retry that fails will have the stack trace.

dubee:
@mdeuser, anything else on this one?

dubee:
PG4 1807 ⏳

mdeuser:
pg :large_blue_circle: :+1:

tests/src/test/scala/common/Wsk.scala
rabbah: This change in my opinion is wrong and should be backed out.

The retry method used here requires the function under retry to throw an exception to be retried, and use of `retry` here makes it impossible to exit the retry loop subject to the expected exit code unless that code is `NETWORK_ERROR_EXIT`.

I suggest backing this out and then fixing it properly, with a test for the `cli` command itself.

FYI @houshengbo

rabbah: I'm addressed this in #3776.

PR #3698, opened 2018-05-24
Make amount of parallel docker runs configurable.

cbickel:
This PR makes the amount of parallel docker runs configurable.
In addition, it reduces the amount of checks, if the database already exists during deployment.

PG5#346 is running.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [x] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -84,6 +84,7 @@
     - "{{ db.whisk.actions }}"
     - "{{ db.whisk.auth }}"
     - "{{ db.whisk.activations }}"
+  run_once: true
 
 - name: prepare controller port
   set_fact:
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -84,7 +84,6 @@
     - "{{ db.whisk.actions }}"
     - "{{ db.whisk.auth }}"
     - "{{ db.whisk.activations }}"
-  run_once: true
 
 - name: prepare controller port
   set_fact:
--- a/core/invoker/src/main/resources/application.conf
+++ b/core/invoker/src/main/resources/application.conf
@@ -10,23 +10,25 @@ whisk {
     poll-interval: 5 minutes
   }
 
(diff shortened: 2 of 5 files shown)


ansible/roles/controller/tasks/deploy.yml
```diff
@@ -82,6 +82,7 @@
     - "{{ db.whisk.actions }}"
     - "{{ db.whisk.auth }}"
     - "{{ db.whisk.activations }}"
+  run_once: true
```
rabbah: ?

cbickel: The intent of this check is to verify, that the databases exist.
If you have a system with hundred of invokers, I think it's not necessary that each invoker checks, if the database exists. It should be enough, if one invoker does this check.
What do you think?

markusthoemmes: Should we take this out of this PR and open a new one for it? Doesn't seem to have anything to do with the actual change?

core/invoker/src/main/resources/application.conf
```diff
@@ -10,6 +10,14 @@ whisk {
     poll-interval: 5 minutes
   }
 
+  # Docker < 1.13.1 has a known problem: if more than 10 containers are created (docker run)
+  # concurrently, there is a good chance that some of them will fail.
+  # See https://github.com/moby/moby/issues/29369
+  # Use a semaphore to make sure that at most 10 `docker run` commands are active
+  # the same time.
+  # 0 means that there are infinite parallel runs.
+  docker.parallel-runs: 10
```
markusthoemmes: Should we refactor the config a bit? I'm envisioning something like:

```
docker.client {
  parallel-runs: 10,
  timeouts {
    ..
  }
}
```

Which would change the code to:

```scala
case class DockerClientConfig(parallelRuns: Int, timeouts: DockerClientTimeoutConfig)
```

And makes it a bit more clean in general. WDYT?

cbickel:
PG2#3492 looks good.

PR #3704, opened 2018-05-24
Invoker graceful shutdown and drain mode

dubee:
Allows an invoker to manually be marked as down so that maintenance can be performed on the invoker. When in maintenance mode, the invoker will finish executing all actions currently running and finish processing its queued actions. Since an invoker in maintenance mode will appear as down to the load balancer, new actions will not be queued up for the invoker. The invoker will exit maintenance mode when it is restarted.

Related to https://github.com/apache/incubator-openwhisk/issues/3681

## Related issue and scope

- [x] API
- [x] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


chetanmeh:
I believe this is first management api over REST which allows change of state of system. Does this api has some form of  security enabled?

rabbah:
+1 @chetanmeh 

The administrative interfaces should have authentication in place. While at it, we should apply the same to the `/invokers` route in the controller perhaps.

markusthoemmes:
What do we gain over `docker stop`ing the container in this case?

Have you thought about using kill-signal based functionality here to integrate better with tools like kube?

csantanapr:
I vote for not using url I consider this anti pattern. 

I preferred os.Signal for use cases that “docker CLI” doesn’t cover

chetanmeh:
Both [Kubernetes][1] and [Mesos][2] support graceful shutdown via `SIGTERM` followed by some kind of graceful shutdown timeout (~ 30 secs and configurable). Only caveat here is that the shell shipped by default with Alpine Linux does not pass signals to children so we would need to launch with bash (currently we do install bash in common Docker base image)

So if we can ensure that Invoker can finish up shutdown in this time window then we should rely on this support. If at all we need to expose a url endpoint then we can secure it via leveraging same credentials which we use for JMX authentication (either directly or probably via some form of JAAS auth)

[1]: https://pracucci.com/graceful-shutdown-of-kubernetes-pods.html
[2]: https://jira.mesosphere.com/browse/MARATHON-2314

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -188,5 +188,20 @@ object Invoker {
...
+
+    Signal.handle(
+      new Signal("USR2"),
+      new SignalHandler() {
+        override def handle(signal: Signal) = {
+          signal.getName match {
+            case "USR2" =>
+              logger.info(this, s"Stopping health scheduler")
+              actorSystem.stop(healthScheduler)
+              while (true) {}
```
dubee: For a graceful shutdown, we would need to make sure user' containers are not running and possibly check that the queue is empty.

rabbah: By queue do you mean the internal message feed or the kafka topic? The topic doesn't need to be empty as that can take an indeterminate amount of time --- bringing up another invoker to take over the topic would address the messages held in the topic.

rabbah:
@chetanmeh 30s may not be enough time to drain all the request live in memory since each request could be for the max allowed duration. This assumes draining means execute the outstanding requests. It could also mean post these requests back to the topic and let some other invoker deal with them. For that, 30s is likely more suitable.

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
dubee: For the Kafka topic, I mean. Need to find a way to make the invoker stop consuming messages from the topic. Will take a look.

dubee: For the above, might just be able to shutdown the `activationFeed` actor.

rabbah: I thought the feed had a shutdown hook. It is worth adding one.

dubee: Looks like there is a shutdown hook for `containerFactory`. Perhaps maintenance mode is not necessary if we have a graceful shutdown.

markusthoemmes:
Correct, maintenance mode and graceful shutdown are different things. What's the benefit of this extra kill signal vs. just killing the whole invoker? Running actions will die either way.

dubee:
@markusthoemmes, as long as the thread is busy where we hook the signal the invoker will not shutdown. Currently there is just an infinite loop there, but a graceful shutdown would keep the thread busy until a shutdown can happen.

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -284,4 +291,57 @@ class InvokerReactive(
       })
   }
 
+  val healthScheduler = Scheduler.scheduleWaitAtMost(1.seconds)(() => {
+    producer.send("health", PingMessage(instance)).andThen {
+      case Failure(t) => logging.error(this, s"failed to ping the controller: $t")
+    }
+  })
+
+  Signal.handle(
```
dubee: Code is still a little clunky, but I think the behavior is right. Ie, stopping the health scheduler, stopping the Kafka message consumer, and waiting for the user containers to finish.

dubee:
@dgrove-oss, do you think these changes will work with Kube's graceful shutdown feature now?

dubee:
@dgrove-oss, I am using `docker kill --signal=TERM invoker0` to kick off the invoker graceful shutdown. This should be equivalent of `kill 1` from within the invoker container.

rabbah:
Max duration isn’t enough since there could be requests queued still in memory (feeds are double buffered).

markusthoemmes:
I'm wondering if there is a race condition possible: Can the pool be intermittently non-busy while getting new messages from the feed? That is: Should we attach this condition to the feed as in:

1. Shut down the feed.
2. Wait until the feed has no more messages **and** has received a "Processed" message for every message it already handed out.

rabbah:
There’s no other way to know the invoker has quiesced - but a processed message isn’t enough since the pool releases the resource (sends the message) before all the book keeping has finished. For example outstanding db operations might still be in flight to record the activation metadata.

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -284,4 +290,50 @@ class InvokerReactive(
...
+    } catch {
+      case e: akka.pattern.AskTimeoutException => logging.info(this, "Activation feed failed to shutdown gracefully")
+    }
+
+    implicit val timeout = Timeout(5 seconds)
+
+    while (Await.result(pool ? Busy, timeout.duration).asInstanceOf[Boolean] == true) {
+      logging.info(this, s"Container pool is busy")
+      Thread.sleep(1000)
+    }
```
markusthoemmes: WDYT about applying some Future composition here to simplify the flow, like

```scala
/** Polls the pools status and returns a future which completes once the pool is idle. */
def waitForContainerPoolIdle(pool: ActorRef): Future[Unit] = {
  (pool ? Busy).mapTo[Boolean].flatMap {
    case false => akka.pattern.after(1.second, system.scheduler)(waitForContainerPoolIdle(pool))
    case true => Future.successful(())
  }.recoverWith { case _ => akka.pattern.after(1.second, system.scheduler) }
}

val shutdowns = Seq(
  gracefulStop(healthScheduler, 5.seconds).recover { case _ => logging.info(this, "Health communication failed to shutdown gracefully")}, 
  gracefulStop(activationFeed, 5.seconds).recover { case _ => logging.info(this, "Activation feed failed to shutdown gracefully")}, 
  waitForContainerPoolIdle(pool))

// Allow the shutdown to take a maximum of 3 times the maximum action runtime since the 
// feed can be buffered and we want to allow for some grace period.
Await.result(shutdowns, TimeLimit.MAX_DURATION * 3)
```

Does that make sense?

dubee: I like it!

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -284,4 +290,63 @@ class InvokerReactive(
...
+  }
+
+  private val termSignal = "TERM"
+
+  Signal.handle(
+    new Signal(termSignal),
+    new SignalHandler() {
+      override def handle(signal: Signal) = {
+        signal.getName match {
+          case `termSignal` =>
```
markusthoemmes: Is there even a chance that a different signal makes it here since the handler is only attached to a specific signal? could possibly drop all the matching.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -178,6 +178,9 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
     case RescheduleJob =>
       freePool = freePool - sender()
       busyPool = busyPool - sender()
+
+    case Busy =>
+      sender ! (busyPool.size != 0)
```
markusthoemmes: Should be `sender ! busyPool.nonEmpty`

It's valuable to make it a habit to replace all `size > 0` and `size == 0` with the respective `nonEmpty` or `isEmpty` equivalents. It can be very costly (require a whole iteration and thus be `O(n)`) to calculate the size of a datastructure while `isEmpty`/`nonEmpty` can always be calculated in `O(1)`

Doesn't matter here, but a good habit to foster.

dubee:
Looks like there is another queue in `MessageConsumer` that should be drained before `activationFeed` is completely terminated.

dubee:
@rabbah, @markusthoemmes, made some changes to prevent the mentioned race condition with the activation feed. Now the feed is being polled until it is idle instead of just terminating the feed.

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
rabbah: Can you emit a log message while waiting for quiescence?

core/invoker/src/main/scala/whisk/core/containerpool/ContainerPool.scala
```diff
@@ -178,6 +178,9 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
     case RescheduleJob =>
       freePool = freePool - sender()
       busyPool = busyPool - sender()
+
+    case Busy =>
+      sender ! (!busyPool.isEmpty)
```
rabbah: busyPool.nonEmpty?

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -284,4 +290,84 @@ class InvokerReactive(
       })
   }
 
+  val healthScheduler = Scheduler.scheduleWaitAtMost(1.seconds)(() => {
+    producer.send("health", PingMessage(instance)).andThen {
+      case Failure(t) => logging.error(this, s"failed to ping the controller: $t")
```
rabbah: the controller -> controllers

common/scala/src/main/scala/whisk/core/connector/MessageConsumer.scala
```diff
@@ -163,6 +171,26 @@ class MessageFeed(description: String,
...
+    case _ => stay
+  }
+
+  when(GracefulShutdownDrain) {
+    case Event(Processed, _) =>
+      updateHandlerCapacity()
+      sendOutstandingMessages()
+      stay
+
+    case Event(FillCompleted(messages), _) =>
```
dubee: Still concerned about a race condition here. There might be a chance that the `Busy` event returns `false` while `fillPipeline()` is running and before the `FillCompleted` event is triggered.

rabbah: right - if filling, you need to wait.
you'll also need to prevent a further fill.

dubee: Hmm, do we know if events are handled synchronous or asynchronous in the Akka FSM?

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -284,4 +290,93 @@ class InvokerReactive(
...
+            case _ => logging.info(this, "Activation feed failed to shutdown gracefully")
+          }
+        } yield {
+          logging.info(this, "Successfully shutdown health scheduler, activation feed, and container pool")
+        }
+
+        try {
+          // Allow the shutdown to take a maximum of 3 times the maximum action runtime since the feed can be
+          // buffered and we want to allow for some grace period.
+          Await.result(shutdowns, TimeLimit.MAX_DURATION * 3)
```
chetanmeh: May be we return a different exit code in case of unclean shutdown

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -284,4 +290,93 @@ class InvokerReactive(
...
+        } finally {
+          containerFactory.cleanup()
+          logging.info(this, "Shutting down invoker")
+          System.exit(0)
```
chetanmeh: Given its a `System.exit` we should catch and log exception of `Await` operation

core/invoker/src/main/scala/whisk/core/invoker/InvokerReactive.scala
```diff
@@ -284,4 +290,93 @@ class InvokerReactive(
...
+        case false =>
+          Future.successful(())
+      }
+      .recoverWith { case _ => after(delay, actorSystem.scheduler)(waitForActivationFeedIdle(feed)) }
+  }
+
+  // Capture SIGTERM signals to gracefully shutdown the invoker. When gracefully shutting down, the health scheduler is
+  // shutdown preventing additional actions from being scheduler to the invoker, then the invoker processes its buffered
+  // messages from the activation feed, and waits for its user containers to finish running before the process exits.
+  Signal.handle(
```
chetanmeh: Do we need to rely on `Signal` here or normal shutdown hook would work as its also invoked in response to  `SIGTERM`

dubee: From experimenting, order of shutdown hooks is not guaranteed. That means the actor system may shutdown before the graceful shutdown hook is executed. However, the actor system needs to be running in order to perform the graceful shutdown.

chetanmeh: Okie. Reading [this comment][1] and [Signal Source Docs][2] it appeared that JVM would ignore any handler for signals for  which it registers its own handler like SIGTERM. 

```
 * Java code cannot register a handler for signals that are already used
 * by the Java VM implementation. The <code>Signal.handle</code>
 * function raises an <code>IllegalArgumentException</code> if such an attempt
 * is made.
```

So was not sure if current approach would always work or not

[1]: https://github.com/eclipse/openj9/issues/40#issuecomment-330820366
[2]: http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/sun/misc/Signal.java

markusthoemmes:
@dubee what's the status of this? Seems like you've run into some dead-ends/race conditions?

dubee:
Closing this as there seem to be race conditions.

PR #3710, opened 2018-05-29
Make REST communication with action containers more robust

sven-lange-last:
On systems with high load, POSTing the action container `/init` endpoint occasionally fails with `NoRouteToHostException`. Retry if this exception occurs.

## Description
* So far, the HTTP client retried when the connection to an action container failed with `HttpHostConnectException`. This is safe because no data has been transmitted yet when this exception occurs. Usually, this exception will occur if the action container startup is slow such that the target socket is not listening yet.
* This change adds `NoRouteToHostException` as retryable exception. Root cause for this exception is high system load. It is safe to retry after this exception occurred because it implies that no data has been transmitted yet.
* Needed to restructure the code to add the second retryable exception.
* Lowered the time between retries from 100 ms to 10 ms because this shorter time should be sufficient to have routing available / the target socket listening.
* Added logging so that retries can be seen in system logs.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/core/containerpool/HttpUtils.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/HttpUtils.scala
@@ -125,20 +125,16 @@ protected[core] class HttpUtils(hostname: String, timeout: FiniteDuration, maxRe
       case t: NoRouteToHostException => Failure(RetryableConnectionError(t))
     } match {
       case Success(response) => response
-      case Failure(t: RetryableConnectionError) =>
-        if (retry) {
-          val sleepTime = 10.milliseconds
-          if (timeout.length > 0) {
-            logging.info(this, s"POST failed with ${t} - retrying after sleeping ${sleepTime}.")
-            Thread.sleep(sleepTime.toMillis)
-            val newTimeout = timeout - sleepTime
-            execute(request, newTimeout, retry = true)
-          } else {
-            logging.warn(this, s"POST failed with ${t} - no retry because timeout exceeded.")
-            Left(Timeout(t))
-          }
+      case Failure(t: RetryableConnectionError) if retry =>
+        val sleepTime = 10.milliseconds
+        if (timeout > Duration.Zero) {
+          logging.info(this, s"POST failed with ${t} - retrying after sleeping ${sleepTime}.")
+          Thread.sleep(sleepTime.toMillis)
+          val newTimeout = timeout - sleepTime
+          execute(request, newTimeout, retry = true)
         } else {
-          Left(ConnectionError(t))
+          logging.warn(this, s"POST failed with ${t} - no retry because timeout exceeded.")
+          Left(Timeout(t))
         }
       case Failure(t: Throwable) => Left(ConnectionError(t))
     }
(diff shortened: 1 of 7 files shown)


common/scala/src/main/scala/whisk/core/containerpool/HttpUtils.scala
```diff
@@ -105,15 +112,33 @@ protected[core] class HttpUtils(hostname: String, timeout: FiniteDuration, maxRe
...
-      case Failure(t: HttpHostConnectException) if retry =>
-        if (timeoutMsec > 0) {
-          Thread sleep 100
-          val newTimeout = timeoutMsec - 100
-          execute(request, newTimeout, retry)
+      case Success(response) => response
+      case Failure(t: RetryableConnectionError) =>
+        if (retry) {
+          val sleepTime = 10.milliseconds
+          if (timeout.length > 0) {
```
markusthoemmes: This should be `if(timeout > Duration.Zero)`

common/scala/src/main/scala/whisk/core/containerpool/HttpUtils.scala
```diff
@@ -141,14 +166,15 @@ protected[core] class HttpUtils(hostname: String, timeout: FiniteDuration, maxRe
...
+  def post(host: String, port: Int, endPoint: String, content: JsValue)(implicit logging: Logging,
+                                                                        tid: TransactionId): (Int, Option[JsObject]) = {
     val connection = new HttpUtils(s"$host:$port", 90.seconds, 1.MB)
     val response = connection.post(endPoint, content, retry = true)
     connection.close()
     response match {
       case Right(r)                   => (r.statusCode, Try(r.entity.parseJson.asJsObject).toOption)
       case Left(NoResponseReceived()) => throw new IllegalStateException("no response from container")
-      case Left(Timeout())            => throw new java.util.concurrent.TimeoutException()
+      case Left(Timeout(t))           => throw new java.util.concurrent.TimeoutException()
```
markusthoemmes: Could drop the value (i.e. use `Timeout(_)`)

common/scala/src/main/scala/whisk/core/containerpool/HttpUtils.scala
```diff
@@ -68,20 +71,24 @@ protected[core] class HttpUtils(hostname: String, timeout: FiniteDuration, maxRe
...
 
-    execute(request, timeout.toMillis.toInt, retry)
+    execute(request, timeout, retry)
   }
 
-  private def execute(request: HttpRequestBase,
-                      timeoutMsec: Integer,
-                      retry: Boolean): Either[ContainerHttpError, ContainerResponse] = {
+  // Used internally to wrap all exceptions for which the request can be retried
+  private case class RetryableConnectionError(t: Throwable) extends Exception(t) with NoStackTrace
```
markusthoemmes: Nice!

common/scala/src/main/scala/whisk/core/containerpool/HttpUtils.scala
```diff
@@ -105,15 +112,33 @@ protected[core] class HttpUtils(hostname: String, timeout: FiniteDuration, maxRe
...
+      case t: NoRouteToHostException => Failure(RetryableConnectionError(t))
     } match {
-      case Success(r) => r
-      case Failure(t: HttpHostConnectException) if retry =>
-        if (timeoutMsec > 0) {
-          Thread sleep 100
-          val newTimeout = timeoutMsec - 100
-          execute(request, newTimeout, retry)
+      case Success(response) => response
+      case Failure(t: RetryableConnectionError) =>
```
markusthoemmes: Could add the `if retry` as a guard here, the `else` branch and the more general `Left` case are equivalent.

common/scala/src/main/scala/whisk/core/containerpool/HttpUtils.scala
sven-lange-last: Makes perfect sense. I will change.

common/scala/src/main/scala/whisk/core/containerpool/HttpUtils.scala
sven-lange-last: Good catch. Done.

PR #3731, opened 2018-06-07
Relax connection retry in invoker

sven-lange-last:
## Description

* In the past, sleep time between connection retries has been 100 ms. This has been lowered to 10 ms with #3710. On a loaded invoker, 50 ms makes more sense.
* Reduce log volume on loaded invoker.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/core/containerpool/HttpUtils.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/HttpUtils.scala
@@ -126,9 +126,8 @@ protected[core] class HttpUtils(hostname: String, timeout: FiniteDuration, maxRe
     } match {
       case Success(response) => response
       case Failure(t: RetryableConnectionError) if retry =>
-        val sleepTime = 10.milliseconds
+        val sleepTime = 50.milliseconds
         if (timeout > Duration.Zero) {
-          logging.info(this, s"POST failed with ${t} - retrying after sleeping ${sleepTime}.")
           Thread.sleep(sleepTime.toMillis)
           val newTimeout = timeout - sleepTime
           execute(request, newTimeout, retry = true)


sven-lange-last:
PG 1 / 2995 👍

rabbah:
travis not enugh?

common/scala/src/main/scala/whisk/core/containerpool/HttpUtils.scala
```diff
@@ -126,9 +126,8 @@ protected[core] class HttpUtils(hostname: String, timeout: FiniteDuration, maxRe
     } match {
       case Success(response) => response
       case Failure(t: RetryableConnectionError) if retry =>
-        val sleepTime = 10.milliseconds
+        val sleepTime = 50.milliseconds
```
csantanapr: @sven-lange-last Can this hardcoded value value be be a configuration setting that can be overload thru prureconfig/environment-variable?

It was originally to retry every 100ms, then it was changed to 10ms, and now every 50ms, 
In case 50ms also makes the invokers slower overall and need to be adjusted again, doing on config is more manageable.

sven-lange-last: @csantanapr Let me put it that way: it may be tempting to make this value configurable. I think that this setting should not be tampered with to save an invoker. This value is used in a `Thread.sleep()` which occupies the current worker thread. Increasing the value may have pretty undesired effects. We should find a reasonable value and don't change it any more.

It would be much more important to replace the `Thread.sleep()` with something that frees the thread while waiting / sleeping.

PR #3743, opened 2018-06-08
Fixes #3742 - just remove the time range part of the query, each acti…

### LINKED ISSUE #3742: Splunk Log Sore may return an empty result set if the time zones are not the same
## Environment details:

* local deployment, vagrant, native ubuntu, Mac OS, Bluemix, ...   
* version of docker, vagrant, ubuntu, ...

1.   Set the logging spi to splunk: `CONFIG_whisk_spi_LogStoreProvider=whisk.core.containerpool.logging.SplunkLogStoreProvider`
2.   Ensure the splunk user used for querying and the controller have different time zones 
3.   Check the controller logs:

```
{"error":"Object is missing required member 'results'","code":"P4VoKRECg6iECBf0SGILQm7G5rqZMNma"}
```

No error and you should get the logs back.

mcdan:
…vation is keyed with its activation ID already.

## Description

Just remove the time range from the query. The activation Id should be sufficient anyway. If the system has two activations with the same ID we have other problems.

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3742 )

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

- [ x] Bug fix (generally a non-breaking change which closes an issue).

## Checklist:

- [x ] I signed an [Apache CLA](https://github.com/apache/incubator-openwhisk/blob/master/CONTRIBUTING.md).
- [ x] I reviewed the [style guides](https://github.com/apache/incubator-openwhisk/wiki/Contributing:-Git-guidelines#code-readiness) and followed the recommendations (Travis CI will check :).

### DIFF
--- a/common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
@@ -112,11 +112,7 @@ class SplunkLogStore(
       Map(
         "exec_mode" -> "oneshot",
         "search" -> search,
-        "output_mode" -> "json",
-        "earliest_time" -> activation.start.toString, //assume that activation start/end are UTC zone, and splunk events are the same
-        "latest_time" -> activation.end
-          .plusSeconds(5) //add 5s to avoid a timerange of 0 on short-lived activations
-          .toString)).toEntity
+        "output_mode" -> "json")).toEntity
 
     logging.debug(this, "sending request")
     queueRequest(


common/scala/src/main/scala/whisk/core/containerpool/logging/SplunkLogStore.scala
```diff
@@ -112,11 +112,7 @@ class SplunkLogStore(
       Map(
         "exec_mode" -> "oneshot",
         "search" -> search,
-        "output_mode" -> "json",
-        "earliest_time" -> activation.start.toString, //assume that activation start/end are UTC zone, and splunk events are the same
-        "latest_time" -> activation.end
-          .plusSeconds(5) //add 5s to avoid a timerange of 0 on short-lived activations
-          .toString)).toEntity
+        "output_mode" -> "json")).toEntity
```
tysonnorris: I think the "no results" case also needs to be handled, regardless of whether the query is correct. 

It is still not clear why the timestamp queries wouldn't work - my concern is that this could be problematic depending on the config of the splunk index; do you have example logs and queries where this fails?

markusthoemmes:
What's the status of this? Any updates?

rabbah:
@mcdan @tysonnorris is this still live?

tysonnorris:
closing this since `SplunkLogStore` query is working fine; @mcdan let's discuss in #3742 if you think this is still an issue?

PR #3752, opened 2018-06-12
Move uuid into namespace to decouple it from authkey

mhenke1:
This PR makes the uuid an explicit part of a namespace  object containing name and uuid of a namespace.

## Description
In many places the uuid stored in the authkey is now used as the unique way to identify
a namespace. This behavior entangles the authorization mechanisms
with other unrelated parts of the code, making changes and extensions of the authentication mechanism unnecessarily hard.

This PR makes the uuid a member of a new namespace object containing the name and the
uuid of the namespace. All references to the uuid that are not used on the context of authorization are using the new namespace object as source.

[PG3:2375 succceded]

## Related issue and scope
N/A

## My changes affect the following components>
- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [x] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/Identity.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Identity.scala
@@ -19,7 +19,6 @@ package whisk.core.entity
 
 import scala.concurrent.Future
 import scala.util.Try
-
 import spray.json._
 import types.AuthStore
 import whisk.common.Logging
@@ -38,12 +37,18 @@ object UserLimits extends DefaultJsonProtocol {
   implicit val serdes = jsonFormat3(UserLimits.apply)
 }
 
+protected[core] case class Namespace(name: EntityName, uuid: UUID)
+
+protected[core] object Namespace extends DefaultJsonProtocol {
+  implicit val serdes = jsonFormat2(Namespace.apply)
+}
+
 protected[core] case class Identity(subject: Subject,
-                                    namespace: EntityName,
+                                    namespace: Namespace,
                                     authkey: AuthKey,
                                     rights: Set[Privilege],
                                     limits: UserLimits = UserLimits()) {
-  def uuid = authkey.uuid
+  def uuid = namespace.uuid
 }
 
 object Identity extends MultipleReadersSingleWriterCache[Identity, DocInfo] with DefaultJsonProtocol {
@@ -126,7 +131,12 @@ object Identity extends MultipleReadersSingleWriterCache[Identity, DocInfo] with
         val JsString(uuid) = value("uuid")
         val JsString(secret) = value("key")
         val JsString(namespace) = value("namespace")
-        Identity(subject, EntityName(namespace), AuthKey(UUID(uuid), Secret(secret)), Privilege.ALL, limits)
+        Identity(
+          subject,
+          Namespace(EntityName(namespace), UUID(uuid)),
+          AuthKey(UUID(uuid), Secret(secret)),
+          Privilege.ALL,
+          limits)
       case _ =>
         logger.error(this, s"$viewName[$key] has malformed view '${row.compactPrint}'")
         throw new IllegalStateException("identities view malformed")
--- a/common/scala/src/main/scala/whisk/core/entity/Identity.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Identity.scala
@@ -47,9 +47,7 @@ protected[core] case class Identity(subject: Subject,
                                     namespace: Namespace,
                                     authkey: AuthKey,
                                     rights: Set[Privilege],
-                                    limits: UserLimits = UserLimits()) {
-  def uuid = namespace.uuid
-}
+                                    limits: UserLimits = UserLimits())
 
 object Identity extends MultipleReadersSingleWriterCache[Identity, DocInfo] with DefaultJsonProtocol {
 
--- a/common/scala/src/main/scala/whisk/core/entity/WhiskAuth.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/WhiskAuth.scala
@@ -25,18 +25,18 @@ import scala.util.Try
  * database. Each namespace has its own key which is used to determine
  * the {@ Identity} of the user calling.
  */
-protected[core] case class WhiskNamespace(name: EntityName, authkey: AuthKey)
+protected[core] case class WhiskNamespace(namespace: Namespace, authkey: AuthKey)
 
 protected[core] object WhiskNamespace extends DefaultJsonProtocol {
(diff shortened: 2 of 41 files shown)


common/scala/src/main/scala/whisk/core/entity/Identity.scala
```diff
@@ -38,13 +37,17 @@ object UserLimits extends DefaultJsonProtocol {
...
 
+protected[core] case class Namespace(name: EntityName, uuid: UUID)
+
+protected[core] object Namespace extends DefaultJsonProtocol {
+  implicit val serdes = jsonFormat2(Namespace.apply)
+}
+
 protected[core] case class Identity(subject: Subject,
-                                    namespace: EntityName,
+                                    namespace: Namespace,
```
rabbah: By having two uuids you allow for them to be different. Is that intended?

common/scala/src/main/scala/whisk/core/entity/Identity.scala
```diff
@@ -38,13 +37,17 @@ object UserLimits extends DefaultJsonProtocol {
   implicit val serdes = jsonFormat3(UserLimits.apply)
 }
 
+protected[core] case class Namespace(name: EntityName, uuid: UUID)
```
rabbah: Do we need the nesting vs: add uuid to Identity and replace authkey with secret then?

common/scala/src/main/scala/whisk/core/entity/Identity.scala
mhenke1: The PR has to two motivations.
The first one is to make the (mis-)usage of the uuid as unique namespace alias explicit. 
The change  makes unrelated parts of the system (like for example authentication and user metrics)  also unrelated on the code level.  This could of course be done by providing new access functions to the Identity object. 
My second - not fully outspoken -  motivation for the change is to pave the path
to make the authentication extensible and/or exchangeable. For both cases it is imaginable 
the another authentication implementation  might not depend (fully) on the subject db but (additional) on an external identity and access management system that will not provide a uuid.

I plan to discuss a proposal for extending authentication and authorization on the dev list in the very near future. Since I saw this change more as driven by hygienic reasons (see my first motivation) I put it out separately.

common/scala/src/main/scala/whisk/core/entity/Identity.scala
mhenke1: please see my comment above

rabbah:
Thanks for the explanation and the context is more clear now. Definitely bring this to the dev list as there is broader interest wrt IAM integration. 

The PR also suggests the promotion of the namespace type more first class rather than an alias for EntityName.

common/scala/src/main/scala/whisk/core/entity/Identity.scala
```diff
@@ -38,13 +37,17 @@ object UserLimits extends DefaultJsonProtocol {
...
+  implicit val serdes = jsonFormat2(Namespace.apply)
+}
+
 protected[core] case class Identity(subject: Subject,
-                                    namespace: EntityName,
+                                    namespace: Namespace,
                                     authkey: AuthKey,
                                     rights: Set[Privilege],
-                                    limits: UserLimits = UserLimits()) {
-  def uuid = authkey.uuid
```
rabbah: This method could stay then and refers to namespace.uuid as a slight convenience.

common/scala/src/main/scala/whisk/core/entity/WhiskAuth.scala
```diff
@@ -25,18 +25,18 @@ import scala.util.Try
...
     def write(w: WhiskNamespace) =
-      JsObject("name" -> w.name.toJson, "uuid" -> w.authkey.uuid.toJson, "key" -> w.authkey.key.toJson)
+      JsObject("name" -> w.namespace.name.toJson, "uuid" -> w.namespace.uuid.toJson, "key" -> w.authkey.key.toJson)
 
     def read(value: JsValue) =
       Try {
         value.asJsObject.getFields("name", "uuid", "key") match {
           case Seq(JsString(n), JsString(u), JsString(k)) =>
-            WhiskNamespace(EntityName(n), AuthKey(UUID(u), Secret(k)))
+            WhiskNamespace(Namespace(EntityName(n), UUID(u)), AuthKey(UUID(u), Secret(k)))
```
rabbah: this implies a database schema change in the database.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -187,7 +187,7 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with
   override def create(user: Identity, entityName: FullyQualifiedEntityName)(implicit transid: TransactionId) = {
     parameter('overwrite ? false) { overwrite =>
       entity(as[WhiskActionPut]) { content =>
-        val request = content.resolve(user.namespace)
+        val request = content.resolve(user.namespace.name)
```
rabbah: should we instead change these method signature to accept the new Namespace type instead?

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
```diff
@@ -245,7 +245,7 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers {
     val allowedSize = ActivationEntityLimit.MAX_ACTIVATION_ENTITY_LIMIT.toBytes
 
     // Needs some bytes grace since activation message is not only the payload.
-    val args = Map("p" -> ("a" * (allowedSize - 700).toInt).toJson)
+    val args = Map("p" -> ("a" * (allowedSize - 750).toInt).toJson)
```
rabbah: ? this change seems unrelated

mhenke1: Not really. Serializing the uuid as an additional part of the namespace has added some crucial bytes to the message so that the "grace bytes" where not enough to get the message passed.

common/scala/src/main/scala/whisk/core/entity/Identity.scala
mhenke1: I was asked by Markus to remove that fuction, to make more explicit that the uuid is retrieved from the namespace.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
mhenke1: Addressed in latest commit

Issue #3759, opened 2018-06-13
routemgmt.yml playbook will fail silently if the wsk binary is missing


rabbah:
If you manage to run the playbook (for example with mode=clean) with a missing `wsk` binary the playbook succeeds even though the script failed. This is because it swallows the error incorrectly.

line 39 suppresses the failure:
https://github.com/apache/incubator-openwhisk/blob/cd7196a8e16aef6e848add239f6dec83c208e5f7/ansible/roles/routemgmt/files/uninstallRouteMgmt.sh#L39-L41

rabbah:
@mdeuser there's a problem with the dependencies in this playbook.

the routemgmt playbook runs last; when deploying this is ok since it will download `wsk` and install the package.

but if you teardown the system, this playbook cannot uninstall the package since the deployment is down at that point.

the playbook fails silently because a failed status code from `wsk` in the uninstall script is accepted.

mdeuser:
you are correct.  the `clean` task depends on a live backend.  i like the idea of pulling it out of openwhisk.yml and explicitly calling it in the correct order.  i prefer to keep it out of the apigateway.yml playbook to keep these concerns separate.

rabbah:
OK let me try this:

```
openwhisk 
download cli
routemgmt
```

to teardown

```
routemgmt
remove cli
openwhisk
```

rabbah:
@mdeuser there is a postdeploy playbook - it could also go in there; that playbook installs the various packages (aka catalog).

rabbah:
@mdeuser why not put it in the api gateway playbook? Logically it won't work without it.

rabbah:
I did improve this some time ago.

PR #3760, opened 2018-06-13
Separate routemgmt.yml out from  openwhisk.yml playbook

rabbah:
API gateway is missing from `redo` aka `wskdev`. Add it and also separate the `wsk` download into its own playbook. Fix bug in script which failed silently if `binary` did not exist.

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3759)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [x] Deployment
- [ ] CLI
- [x] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -150,27 +150,31 @@ cd ansible
 ansible-playbook -i environments/<environment> couchdb.yml
 ansible-playbook -i environments/<environment> initdb.yml
 ansible-playbook -i environments/<environment> wipe.yml
-ansible-playbook -i environments/<environment> apigateway.yml
 ansible-playbook -i environments/<environment> openwhisk.yml
+
+# installs a catalog of public packages and actions
 ansible-playbook -i environments/<environment> postdeploy.yml
+
+# to use the API gateway
+ansible-playbook -i environments/<environment> apigateway.yml
+ansible-playbook -i environments/<environment> routemgmt.yml
 ```
 
-You need to run `initdb.yml` **every time** you do a fresh deploy CouchDB to initialize the subjects database.
-The playbooks `wipe.yml` and `postdeploy.yml` should be run on a fresh deployment only, otherwise all transient
-data that include actions and activations are lost.
+- You need to run `initdb.yml` **every time** you do a fresh deploy CouchDB to initialize the subjects database.
+- The `wipe.yml` playbook should be run on a fresh deployment only, otherwise actions and activations will be lost.
+- Run `postdeploy.yml` after deployment to install a catalog of useful packages.
+- To use the API Gateway, you'll need to run `apigateway.yml` and `routemgmt.yml`.
+- Use `ansible-playbook -i environments/<environment> openwhisk.yml` to avoid wiping the data store. This is useful to start OpenWhisk after restarting your Operating System.
 
 #### Limitation
 
-You can not run multiple CouchDB nodes on a single machine.
-This limitation comes from Erlang EPMD.
-When CouchDB forms a cluster, it counts on EPMD to find other nodes.
-If we want to run multiple nodes on a single machine, we must differentiate EPMD port(`4369`) for each nodes. But if this port is different on each nodes, they cannot find each other.
-So if you want to deploy multiple CouchDB nodes, all nodes should be placed on different machines respectively.
+You can not run multiple CouchDB nodes on a single machine. This limitation comes from Erlang EPMD which CouchDB relies on to find other nodes.
+To deploy multiple CouchDB nodes, they should be placed on different machines respectively otherwise their ports will clash.
 
 
 ### Deploying Using Cloudant
--   Make sure your `db_local.ini` file is set up for Cloudant. See [Setup](#setup)
--   Then execute
+-   Make sure your `db_local.ini` file is set up for Cloudant. See [Setup](#setup).
+-   Then execute:
 
 ```
 cd <openwhisk_home>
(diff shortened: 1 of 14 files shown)


rabbah:
@mdeuser can you take a look?

rabbah:
@mdeuser can you take a look at this PR - I separated the routemgmt package installation as discussed on Slack.

tools/vagrant/Vagrantfile
```diff
@@ -149,8 +149,9 @@ Vagrant.configure('2') do |config|
     echo "`date`: deploy-start" >> /tmp/vagrant-times.txt
     cd ${ANSIBLE_HOME}
     su vagrant -c 'ansible-playbook -i environments/vagrant wipe.yml'
-    su vagrant -c 'ansible-playbook -i environments/vagrant apigateway.yml'
     su vagrant -c 'ansible-playbook -i environments/vagrant openwhisk.yml -e invoker_use_runc=False'
+    su vagrant -c 'ansible-playbook -i environments/vagrant apigateway.yml'
```
mdeuser: maybe even move after `postdeploy.yml`?

ansible/README.md
```diff
@@ -176,14 +178,20 @@ ansible-playbook -i environments/<environment> initdb.yml
...
+ansible-playbook -i environments/<environment> routemgmt.yml
 ```
 
 You need to run `initdb` on Cloudant **only once** per Cloudant database to initialize the subjects database.
 The `initdb.yml` playbook will only initialize your database if it is not initialized already, else it will skip initialization steps.
 
-The playbooks `wipe.yml` and `postdeploy.yml` should be run on a fresh deployment only, otherwise all transient
-data that include actions and activations are lost.
+The playbooks `wipe.yml` should be run on a fresh deployment only, otherwise all data are lost, including actions and activations.
+The `postdeploy.yml` is generally run once after a fresh deployment to install a catalog of generally useful packages.
```
mdeuser: "Run `postdeploy.yml` after deployment to install a catalog of useful packages"

mdeuser:
@rabbah - i think you have it all nicely covered.  the only concern is the breaking change for folk who expect the api web actions to be installed without the extra step.  just a communication thing though..

rabbah:
@mdeuser I pushed a commit to merge the routemgmt playbook installation with managing the API gateway itself. This should backward compatible. If someone wasn't using the apigw playbook then they will need to add the routemgmt to their deployment.

csantanapr:
@mdeuser can you take this reviewed it, this might have impacts to our downstream CI for the ordering of `gateway.yml` playbook now needs to be done after `openwhisk.yml`
Also apigateway repo would need to be adjusted here https://github.com/apache/incubator-openwhisk-apigateway/blob/master/tools/travis/build.sh#L51

rabbah:
@csantanapr @mdeuser I created this PR https://github.com/apache/incubator-openwhisk-apigateway/pull/314 which should be OK to merge now (Travis will verify) and will work after this PR is merged.

rabbah:
@csantanapr the API gateway build may be failing because the openwhisk image on dockerhub is several days old (@dgrove-oss observed similar issue on kube repo - is this kicked yet?)

csantanapr:
I fixed controller image on dockerhub today

ansible/apigateway.yml
```diff
@@ -8,3 +8,5 @@
 - hosts: apigateway
   roles:
   - apigateway
+
+- import_playbook: routemgmt.yml
```
mdeuser: in environments where a non-local api gw is used, these actions will still need to be separately installed.

rabbah:
@mdeuser anything for me to address on this PR?

rabbah:
@mdeuser is there a repo I need to fix downstream your concerned about?

mdeuser:
@rabbah - my apologies for the delay.  i've been in interrupt mode recently..  would you be ok with leaving the routemgmt.yml outside of the apigateway.yml.

rabbah:
does it make sense to do that - can you explain the rationale? My thinking is:

- if someone deploys the gateway, they need the package.
- if someone doesn't deploy the gateway, they have the option of not running the playbook (which avoids installing the packages as well).
- if someone is using a third party gateway, they just need to run the package to install the route management package.

mdeuser:
i guess i was seeing if we can avoid, as much as possible, changes that require some folks to tweak their deployment procedures. currently, the apigateway.yml deployment has been agnostic as to whether or not openwhisk has been deployed; by including routemgmt in that .yml, it now requires openwhisk to have been deployed.  also, any ci/cd pipelines that use a local api gw deployment in some environments, but use an already deployed production api gw in other environments will need new conditional logic when deploying openwhisk to avoid deploying routemgmt twice - a second deployment won't cause a failure, just would take some extra time.

rabbah:
By pulling out routemgmt.yml from the openwhisk playbook there's already a need to adjust a deployment: if you need the package, you have to add it to the deployment explicitly. So it's move the apigw playbook down, or leave it where it is and then add a second playbook. I'm not really seeing a big benefit here since either way - there's a change.

Also none of our playbooks can be deployed again without tear down - what this PR does is fix the bug in the playbook so you can properly uninstall and reinstall the packages.

rabbah:
As in this PR, this is not a big change:
https://github.com/apache/incubator-openwhisk-apigateway/pull/314/files#diff-183ae0496c3df4528d793134f32a6955R50

Maybe there are other scenarios I'm missing.

mdeuser:
> By pulling out routemgmt.yml from the openwhisk playbook there's already a need to adjust a deployment
true.  but imho only needing to make the routemgmt.yml change is a bit simpler in a dynamic pipeline than handling the combined apigw/routemgmt.  the routemgmt is always needed - no need for conditional deployment checks based on the target environment.

rabbah:
I don't follow - are you saying you have a pipeline where you deploy the apigateway once but the routemgmt package many times?

rabbah:
Noting result of discussion on slack: reverted the last commit. The rationale is that the apigw playbook is deploying a component (like couchdb) and routemgmt is related to configuration (like initdb/wipedb) and consistent with how we handle the lifecycle for those component, the two playbooks herein should be separate.

Thanks @mdeuser for the debate.

mdeuser:
pg4/1965

rabbah:
@mdeuser status?

csantanapr:
@mdeuser Can this be merge?

Issue #3769, opened 2018-06-15
Update of java action resulted in 503 failure and action remains unmanagable


jeremiaswerner:
## Environment details:

* Bluemix 

## Steps to reproduce the issue:

1. Delete a Java action  
2. Create a Java action

## Provide the actual results and outputs:

```
$ bx wsk action invoke actionX --result
{
    "error": "Action could not be found or may have been deleted."
}
$ bx wsk action delete actionX
error: Unable to delete action 'actionX': The connection failed, or timed out. (HTTP status code 404)
Run 'wsk --help' for usage.
```

The error shown in the controller looks as follows:
```
[2018-06-15T11:08:17.039Z] [ERROR] [#tid_xxxxxx] [CouchDbRestStore] [ATT_PUT] 'xxxxxx' internal error, name: xxxxx', doc: 'id: xxxxxx, rev: xxxxxx', failure: 'Substream Source cannot be materialized more than once'

[2018-06-15T11:06:56.634Z] [INFO] [#tid_xxxxxx] [WhiskAction] write initiated on existing cache entry, invalidating CacheKey(xxxxxx/xxxxxx), tid_xxxxxx, state WriteInProgress

[2018-06-15T11:06:56.581Z] [INFO] [#tid_xxxxxx] [WhiskAction] write all done, caching CacheKey(xxxxxx/xxxxxx) Cached

[2018-06-15T11:06:56.581Z] [INFO] [#tid_xxxxxx] [WhiskAction] write initiated on new cache entry

[2018-06-15T11:06:55.590Z] [INFO] [#tid_xxxxxx] [RestAPIVersion] authenticate: xxxxxx
```

## Additional information you deem important:
* The system runs with two controller instances
* The issue is reproducible
* The user is not able to delete nor update the action

chetanmeh:
The issue appears to  be coming due to recent changes around supporting attachment inlining #3709 which causes slowdown in attachment upload (see #3770). As an immediate fix I would disable inlining by default by setting `max-inline-size` = 0.

Once #3770 is fixed inlining would be enabled again.

PR #3777, opened 2018-06-18
Avoid converting ByteString to bytes in attachment inlining flow

### LINKED ISSUE #3770: Improve attachment inlining logic
Current attachment inlining approach as implemented in #3709 introduces lots of overhead and thus slows down the upload flow which leads to failures as seen in #3769

Currently it does following things

1. Convert the Source of `ByteString` to Source of `byte`
2. Check if overall bytes are less than maxInlineSize or not 
3. If not then re assemble the bytes to create a `ByteString`

Due to this a 100kb attachment which takes 25 ms without inline flow takes 60 secs with inline flow. As a fix we need to avoid this conversion to from ByteString -> byte -> ByteString and work at ByteString level only. 

As a temporary fix by default inlining would be disabled and once the performance issues are addressed it would be enabled again

chetanmeh:
This PR modifies the current attachment inlining logic to work at `ByteString` level and avoid converting ByteString to bytes and back. This fixes #3770

## Description

So far the inlining logic was converting a stream of `ByteString` to byte array and then working at byte level. This was causing the upload to slow down and also reverse conversion of stream of bytes to `ByteString` was not working as expected.

Instead of doing that it now uses approach mentioned [here](https://groups.google.com/forum/#!topic/akka-user/FdQAAzcwrqM) and works at `ByteString` level. 

See [this commit](https://github.com/apache/incubator-openwhisk/pull/3777/commits/a36bfb14a47bf3773b5dce13790c57f159874467) for this specific change

1. Do a `prefixAndTail` of 1 element
2. If extracted byteString prefix 
    1. is empty then stream is considered inline
    2. else merge the byteString prefix with `previousPrefix` and see if the size is still less than maxInlineSize or not
        1. If yes then recursively call `inlineAndTail` with new source and bytestring accumulated so far
        2. otherwise stream cannot be inlined and return a combined source of prefix extracted so far with tail

```scala 
  protected[database] def inlineAndTail(docStream: Source[ByteString, _],
                                        previousPrefix: ByteString = ByteString.empty)(
    implicit ec: ExecutionContext): Future[Either[ByteString, Source[ByteString, _]]] = {
    docStream.prefixAndTail(1).runWith(Sink.head).flatMap {
      case (prefix, tail) =>
        prefix.headOption.fold {
          Future.successful[Either[ByteString, Source[ByteString, _]]](Left(previousPrefix))
        } { prefix =>
          val completePrefix = previousPrefix ++ prefix
          if (completePrefix.size < maxInlineSize.toBytes) {
            inlineAndTail(tail, completePrefix)
          } else {
            Future.successful(Right(tail.prepend(Source.single(completePrefix))))
          }
        }
    }
  }
```

With new impl the time taken to attach is similar to setup where inlining is disabled. 

Some other aspects

1. Re-enables the attachment inlining by setting `max-inline-size` to 16 kb (same as one before #3769)
2. Removed the `chunk-size` config as its not needed
3. Refactors the logic related to uploading attachment to an `AttachmentStore` to `AttachmentSupport` trait as that remains same irrespective of `ArtifactStore`
4. Also includes a test which uploads 5MB attachment so as to check working with mid size attachments

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3770)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
+++ b/common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
@@ -55,16 +55,14 @@ trait AttachmentSupport[DocumentAbstraction <: DocumentSerializer] extends Defau
                                         previousPrefix: ByteString = ByteString.empty)(
     implicit ec: ExecutionContext): Future[Either[ByteString, Source[ByteString, _]]] = {
     docStream.prefixAndTail(1).runWith(Sink.head).flatMap {
-      case (prefix, tail) =>
-        prefix.headOption.fold {
-          Future.successful[Either[ByteString, Source[ByteString, _]]](Left(previousPrefix))
-        } { prefix =>
-          val completePrefix = previousPrefix ++ prefix
-          if (completePrefix.size < maxInlineSize.toBytes) {
-            inlineAndTail(tail, completePrefix)
-          } else {
-            Future.successful(Right(tail.prepend(Source.single(completePrefix))))
-          }
+      case (Nil, _) =>
+        Future.successful(Left(previousPrefix))
+      case (Seq(prefix), tail) =>
+        val completePrefix = previousPrefix ++ prefix
+        if (completePrefix.size < maxInlineSize.toBytes) {
+          inlineAndTail(tail, completePrefix)
+        } else {
+          Future.successful(Right(tail.prepend(Source.single(completePrefix))))
         }
     }
   }
--- a/common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
+++ b/common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
@@ -152,7 +152,7 @@ trait AttachmentSupport[DocumentAbstraction <: DocumentSerializer] extends Defau
 
     for {
       bytesOrSource <- inlineOrAttach(docStream)
-      uri <- Future.successful(uriOf(bytesOrSource, UUID().asString))
+      uri = uriOf(bytesOrSource, UUID().asString)
       attached <- {
         // Upload if cannot be inlined
         bytesOrSource match {
--- a/common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
+++ b/common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
@@ -51,8 +51,13 @@ trait AttachmentSupport[DocumentAbstraction <: DocumentSerializer] extends Defau
   /** Materializer required for stream processing */
   protected[core] implicit val materializer: Materializer
 
-  protected[database] def inlineAndTail(docStream: Source[ByteString, _],
-                                        previousPrefix: ByteString = ByteString.empty)(
+  /**
+   * Given a ByteString source it determines if the source can be inlined or not by returning an
+   * Either - Left(byteString) containing all the bytes from the source or Right(Source[ByteString, _])
+   * if the source is large
+   */
+  protected[database] def inlineOrAttach(docStream: Source[ByteString, _],
+                                         previousPrefix: ByteString = ByteString.empty)(
     implicit ec: ExecutionContext): Future[Either[ByteString, Source[ByteString, _]]] = {
     docStream.prefixAndTail(1).runWith(Sink.head).flatMap {
       case (Nil, _) =>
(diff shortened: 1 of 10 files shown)


common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
```diff
@@ -0,0 +1,189 @@
...
+  def maxInlineSize: ByteSize = inliningConfig.maxInlineSize
+
+  protected def inliningConfig: InliningConfig
+
+  /**
+   * Attachment scheme name to use for non inlined attachments
+   */
+  protected def attachmentScheme: String
+
+  protected def executionContext: ExecutionContext
```
markusthoemmes: 1. Please move all of those members to the top of the file for easier readability.
2. If you already include executionContext here, can we remove it from `inlineOrAttach`'s signature?

common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+   * See {{ ArtifactStore#put }}
+   */
+  protected[database] def put(d: DocumentAbstraction)(implicit transid: TransactionId): Future[DocInfo]
+
+  private def encode(bytes: Seq[Byte]): String = {
```
markusthoemmes: Can this now take `Array[Byte]` to avoid the cast?

common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
```diff
@@ -0,0 +1,189 @@
...
+    attachmentStore: AttachmentStore)(implicit transid: TransactionId): Future[(DocInfo, Attached)] = {
+
+    val asJson = doc.toDocumentRecord
+    val id = asJson.fields("_id").convertTo[String].trim
+
+    implicit val ec = executionContext
+
+    for {
+      bytesOrSource <- inlineOrAttach(docStream)
+      uri <- Future.successful(uriOf(bytesOrSource, UUID().asString))
```
markusthoemmes: I think this can be just `uri = uriOf(bytesOrSource, UUID().asString)` (same for the CouchDBRestStore)

common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
```diff
@@ -0,0 +1,189 @@
...
+        Future.successful(Left(previousPrefix))
+      case (Seq(prefix), tail) =>
+        val completePrefix = previousPrefix ++ prefix
+        if (completePrefix.size < maxInlineSize.toBytes) {
+          inlineOrAttach(tail, completePrefix)
+        } else {
+          Future.successful(Right(tail.prepend(Source.single(completePrefix))))
+        }
+    }
+  }
```
markusthoemmes: Super smart approach 👍 🎉

common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
chetanmeh: Done

common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
chetanmeh: bytes passed is `ByteString` which needs to be then converted to array for Java based api to work

common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
chetanmeh: Tried that but it does not work as datatype here needs to [support operations][1] like `map`, `flatMap`. Hence need to wrap it in `Future`

[1]: https://docs.scala-lang.org/tutorials/FAQ/yield.html#translating-for-comprehensions

markusthoemmes: Note the `=` vs the `<-`

chetanmeh: Good point ... missed that aspect. Updated it now

chetanmeh:
Rebased to latest master

chetanmeh:
Need a PG for this

common/scala/src/main/scala/whisk/core/database/AttachmentSupport.scala
```diff
@@ -0,0 +1,190 @@
...
+
+  /**
+   * Constructs a source from inlined attachment contents
+   */
+  protected[database] def memorySource(uri: Uri): Source[ByteString, NotUsed] = {
+    require(uri.scheme == MemScheme, s"URI $uri scheme is not $MemScheme")
+    Source.single(ByteString(decode(uri)))
+  }
+
+  protected[database] def isInlined(uri: Uri): Boolean = uri.scheme == MemScheme
```
markusthoemmes: I think this method is now unused? Delete?

markusthoemmes:
PG3 2416 👍

PR #3798, opened 2018-06-21
Allow for activation store to accept user and request information

dubee:
Other ActivationStore implementations, Elasticsearch for instance, will need user and request information.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


dubee:
PG4 1863 ⏳

common/scala/src/main/scala/whisk/core/entity/ActivationStore.scala
```diff
@@ -48,7 +49,8 @@ trait ActivationStore {
    * @param transid transaction ID for request
    * @return Future containing the retrieved WhiskActivation
    */
-  def get(activationId: ActivationId)(implicit transid: TransactionId): Future[WhiskActivation]
+  def get(activationId: ActivationId, user: Option[Identity] = None, request: Option[HttpRequest] = None)(
```
markusthoemmes: Why make `user` and `request` an Option? When would you not pass these values?

dubee:
@markusthoemmes, very concerned about this polling behavior. https://github.com/apache/incubator-openwhisk/blob/1d79fa395de32c6a774efeb4f53563689e9cbb7e/core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala#L585

markusthoemmes:
@dubee concerned in which way? I believe this is the database fallback polling in case an active ack goes missing.

common/scala/src/main/scala/whisk/core/entity/ActivationStore.scala
dubee: @markusthoemmes, since there exists situations were the `user` and `request` information may not be know, DB polling for blocking invocations, do you think we should leave those parameters as optional?

dubee: @markusthoemmes, certain activation stores may need to maintain a separate DB to track activations for blocking invocations given the DB polling. When `user` or `request` is not provided to the activation store, this could signal that a blocking invocation is taking place.

chetanmeh:
@dubee Can you share some more details on how Elasticsearch implementation would make use of user and request data. Would it like to use any aspect of request or we can extract explicit parts from request and pass it on. 

For now it appears that passing request param is making SPI usage more coupled to web requests

dubee:
@chetanmeh, certain Elasticsearch deployments require additional authentication which can be passed through the controller via HTTP headers. This allows the controller to communicate to Elasitcsearch for users making requests. Currently our Elasticsearch log store works in the same way. When required headers are specified via the OpenWhisk Ansible configuration for the log store, the log store will pass those headers from the client to Elasticsearch to perform a query.

See the log store for example:
https://github.com/apache/incubator-openwhisk/blob/92a64c291156a2cd3d6b304babc2a193a46d0699/common/scala/src/main/scala/whisk/core/containerpool/logging/ElasticSearchLogStore.scala#L103

chetanmeh:
@dubee Now its more clear. So we pass on some sort of "user context" to ES. May be we make that notion explicit via having a `UserContext` as parameter type

```
case class UserContext(user:Option[Identity], request: Option[HttpRequest])
-  def get(activationId: ActivationId, user: Option[Identity] = None, request: Option[HttpRequest] = None)(
-  def get(activationId: ActivationId, userContext: Option[UserContext] = None)(
```

dubee:
PG2 3475 🔵

dubee:

dubee:
@chetanmeh, where do you think the UserContext case class should live?

dubee:
Added UserContext in the third commit. The case class definition should probably be moved elsewhere though.

dubee:
PG4 2057 🔵

chetanmeh:
> where do you think the UserContext case class should live

Adding it to `ActivationStore` looks fine

dubee:
@markusthoemmes, any other comments on this one?

dubee:
@mdeuser, can you review this one?

common/scala/src/main/scala/whisk/core/containerpool/logging/DockerToActivationLogStore.scala
```diff
@@ -29,6 +28,7 @@ import whisk.common.TransactionId
 import whisk.core.containerpool.Container
 import whisk.core.entity.{ActivationLogs, ExecutableWhiskAction, Identity, WhiskActivation}
 import whisk.http.Messages
+import whisk.core.database.UserContext
```
mdeuser: is this class tightly coupled with the `database` package?  on the surface, it seems like it could go into some place else.

common/scala/src/main/scala/whisk/core/database/ActivationStore.scala
```diff
@@ -38,8 +41,9 @@ trait ActivationStore {
    * @param notifier cache change notifier
    * @return Future containing DocInfo related to stored activation
    */
-  def store(activation: WhiskActivation)(implicit transid: TransactionId,
-                                         notifier: Option[CacheChangeNotification]): Future[DocInfo]
+  def store(activation: WhiskActivation, context: UserContext)(
```
mdeuser: update api comment with new param

common/scala/src/main/scala/whisk/core/database/ActivationStore.scala
```diff
@@ -48,7 +52,7 @@ trait ActivationStore {
    * @param transid transaction ID for request
    * @return Future containing the retrieved WhiskActivation
    */
-  def get(activationId: ActivationId)(implicit transid: TransactionId): Future[WhiskActivation]
+  def get(activationId: ActivationId, context: UserContext)(implicit transid: TransactionId): Future[WhiskActivation]
```
mdeuser: update api comment with new param

common/scala/src/main/scala/whisk/core/database/ActivationStore.scala
```diff
@@ -58,8 +62,9 @@ trait ActivationStore {
    * @param notifier cache change notifier
    * @return Future containing a Boolean value indication whether the activation was deleted
    */
-  def delete(activationId: ActivationId)(implicit transid: TransactionId,
-                                         notifier: Option[CacheChangeNotification]): Future[Boolean]
+  def delete(activationId: ActivationId, context: UserContext)(
```
mdeuser: update api comment with new param

common/scala/src/main/scala/whisk/core/database/ActivationStore.scala
```diff
@@ -76,7 +81,8 @@ trait ActivationStore {
                                   name: Option[EntityPath] = None,
                                   skip: Int,
                                   since: Option[Instant] = None,
-                                  upto: Option[Instant] = None)(implicit transid: TransactionId): Future[JsObject]
+                                  upto: Option[Instant] = None,
+                                  context: UserContext)(implicit transid: TransactionId): Future[JsObject]
```
mdeuser: update api comment with new param

common/scala/src/main/scala/whisk/core/database/ActivationStore.scala
```diff
@@ -114,13 +121,14 @@ trait ActivationStore {
...
-                                 upto: Option[Instant] = None)(
-    implicit transid: TransactionId): Future[Either[List[JsObject], List[WhiskActivation]]]
+  def listActivationsInNamespace(
+    namespace: EntityPath,
+    skip: Int,
+    limit: Int,
+    includeDocs: Boolean = false,
+    since: Option[Instant] = None,
+    upto: Option[Instant] = None,
+    context: UserContext)(implicit transid: TransactionId): Future[Either[List[JsObject], List[WhiskActivation]]]
```
mdeuser: update api comment with new param

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -517,6 +518,8 @@ protected[actions] trait PrimitiveActions {
   private def completeActivation(user: Identity, session: Session, response: ActivationResponse)(
     implicit transid: TransactionId): WhiskActivation = {
 
+    val context = UserContext(user)
```
mdeuser: is the original request context needed here?

common/scala/src/main/scala/whisk/core/containerpool/logging/DockerToActivationLogStore.scala
dubee: @mdeuser, the ActivationStore was moved to the database package recently. Where do you think `UserContext` should live?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
dubee: Yes, the user information needs to be passed to the ActivationStore `store()` method. This allows ActivationStore SPIs to extract user information if need be,

mdeuser: but what about the original request context (i.e. request headers, payload, etc)..  since there's is a write to the activation store (like in other places) i was thinking the request context would also be needed (like in other places)

dubee: I don't have a use case for passing the request to the activation store here. Would rather keep the invocation path lean unless we really need the HTTP header in the store method.

PR #3799, opened 2018-06-21
Display proper error when sequence invocation fails due to missing component

dubee:
When invoking a sequence that has a missing component an error is generated in the controller logs and the sequence activation is never stored.

Steps to reproduce:
```
$ wsk action create comp ~/actions/logs.js
$ wsk action create seq --sequence comp
$ wsk action delete comp
$ wsk action invoke seq
ok: invoked /_/seq with id 2373488081eb4ed3b3488081ebbed31e
$ wsk activation get 2373488081eb4ed3b3488081ebbed31e
error: Unable to get activation '2373488081eb4ed3b3488081ebbed31e': The requested resource does not exist. (code fb60fd27845202c2af0e97b4157b8bd5)

```
Log error:
```
[ActionsApi] sequence activation failed: not found on 'get'
```
## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
@@ -230,13 +230,13 @@ protected[actions] trait SequenceActions {
    * @return a future which resolves with the accounting for a sequence, including the last result, duration, and activation ids
    */
   private def invokeSequenceComponents(
-                                        user: Identity,
-                                        seqAction: WhiskActionMetaData,
-                                        seqActivationId: ActivationId,
-                                        inputPayload: Option[JsObject],
-                                        components: Vector[FullyQualifiedEntityName],
-                                        cause: Option[ActivationId],
-                                        atomicActionCnt: Int)(implicit transid: TransactionId): Future[SequenceAccounting] = {
+    user: Identity,
+    seqAction: WhiskActionMetaData,
+    seqActivationId: ActivationId,
+    inputPayload: Option[JsObject],
+    components: Vector[FullyQualifiedEntityName],
+    cause: Option[ActivationId],
+    atomicActionCnt: Int)(implicit transid: TransactionId): Future[SequenceAccounting] = {
 
     // For each action in the sequence, fetch any of its associated parameters (including package or binding).
     // We do this for all of the actions in the sequence even though it may be short circuited. This is to
@@ -251,24 +251,30 @@ protected[actions] trait SequenceActions {
     }
 
     // this holds the initial value of the accounting structure, including the input boxed as an ActivationResponse
-    val initialAccounting = SequenceAccounting(atomicActionCnt, ActivationResponse.payloadPlaceholder(inputPayload))
+    val initialAccounting = Future.successful {
+      SequenceAccounting(atomicActionCnt, ActivationResponse.payloadPlaceholder(inputPayload))
+    }
 
     // execute the actions in sequential blocking fashion
     resolvedFutureActions
-      .foldLeft(Future.successful(initialAccounting)) { (accountingFuture, futureAction) =>
+      .foldLeft(initialAccounting) { (accountingFuture, futureAction) =>
         accountingFuture.flatMap { accounting =>
           if (accounting.atomicActionCnt < actionSequenceLimit) {
-            invokeNextAction(user, futureAction, accounting, cause).flatMap { accounting =>
-              if (!accounting.shortcircuit) {
-                Future.successful(accounting)
-              } else {
-                // this is to short circuit the fold
-                Future.failed(FailedSequenceActivation(accounting)) // terminates the fold
+            invokeNextAction(user, futureAction, accounting, cause)
+              .flatMap { accounting =>
+                if (!accounting.shortcircuit) {
+                  Future.successful(accounting)
+                } else {
+                  // this is to short circuit the fold
+                  Future.failed(FailedSequenceActivation(accounting)) // terminates the fold
+                }
               }
-            }.recoverWith {
-              case _: NoDocumentException =>
-                Future.successful(initialAccounting.fail(ActivationResponse.applicationError(sequenceComponentNotFound), None))
-            }
+              .recoverWith {
+                case _: NoDocumentException =>
+                  val updatedAccount =
+                    accounting.fail(ActivationResponse.applicationError(sequenceComponentNotFound), None)
+                  Future.failed(FailedSequenceActivation(updatedAccount)) // terminates the fold
+              }
           } else {
             val updatedAccount = accounting.fail(ActivationResponse.applicationError(sequenceIsTooLong), None)
             Future.failed(FailedSequenceActivation(updatedAccount)) // terminates the fold
--- a/core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
@@ -277,7 +277,7 @@ protected[actions] trait SequenceActions {
         // since all throwables are recovered with a failed accounting instance and this is
         // in turned boxed to FailedSequenceActivation
         case FailedSequenceActivation(accounting) => Future.successful(accounting)
-        case e: NoDocumentException =>
+        case _: NoDocumentException =>
           Future.successful(
             initialAccounting.fail(ActivationResponse.applicationError(sequenceComponentNotFound), None))
       }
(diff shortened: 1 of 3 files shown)


dubee:
PG4 1864 ⏳

tests/src/test/scala/system/basic/WskSequenceTests.scala
```diff
@@ -60,6 +56,27 @@ abstract class WskSequenceTests extends TestHelpers with ScalatestRouteTest with
...
+      val seqName = "seq"
+      val file = TestUtils.getTestActionFilename("echo.js")
+
+      wsk.action.create(seqCompName, Some(file))
+
+      assetHelper.withCleaner(wsk.action, seqName) { (action, seqName) =>
+        action.create(seqName, Some(seqCompName), kind = Some("sequence"))
+      }
+
+      wsk.action.delete(seqCompName)
```
dubee: Maybe should move the retry around delete.

dubee:
PG4 1864 🔵

dubee:
@rabbah, can you take a look at this one when you are free?

rabbah:
will do thanks for the ping.

dubee:
@rabbah, bump.

tests/src/test/scala/system/basic/WskSequenceTests.scala
```diff
@@ -48,6 +49,26 @@ class WskSequenceTests extends TestHelpers with WskTestHelpers with StreamLoggin
...
+
+      retry {
+        (withActivation(wsk.activation, wsk.action.invoke(seqName)) { activation =>
+          activation.response.result shouldBe Some(JsObject("error" -> sequenceComponentNotFound.toJson))
+        })
```
markusthoemmes: Why is the retry needed here? `withActivation` already does the polling I think?

tests/src/test/scala/whisk/core/controller/test/SequenceApiTests.scala
```diff
@@ -53,6 +49,22 @@ class SequenceApiTests extends ControllerTestCommon with WhiskActionsApi {
 
   val allowedActionDuration = 120 seconds
 
+  it should "produce proper error when sequence component does not exist" in {
```
dubee: @rabbah, good call on using a unit test instead. 👍

core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
```diff
@@ -280,6 +276,9 @@ protected[actions] trait SequenceActions {
         // since all throwables are recovered with a failed accounting instance and this is
         // in turned boxed to FailedSequenceActivation
         case FailedSequenceActivation(accounting) => Future.successful(accounting)
+        case _: NoDocumentException =>
+          Future.successful(
+            initialAccounting.fail(ActivationResponse.applicationError(sequenceComponentNotFound), None))
```
rabbah: at a quick glance i dont think this is quite right - it's possible there were components in the sequence executed before you reach the point of failure due to a missing action in the sequence. Using the initial accounting object here throws away the intermediate record up to that point in this patch.

I suggest a recover with after `resolvedFutureActions` and before `foldLeft` otherwise convince me this is right with a test :)

tests/src/test/scala/whisk/core/controller/test/SequenceApiTests.scala
rabbah: much simpler - thanks.

core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
rabbah: I'm wrong - looking deeper:

```
val resolvedFutureActions = resolveDefaultNamespace(components, user) map { c =>
      WhiskActionMetaData.resolveActionAndMergeParameters(entityStore, c)
    }
```

Will resolve all the components first and so if the future fails with a no doc exception, it is safe to use the initial accounting object. The placement of the recover with is dangerous though (for the reason I initially thought it was incorrect).

core/controller/src/main/scala/whisk/core/controller/actions/SequenceActions.scala
```diff
@@ -247,40 +245,47 @@ protected[actions] trait SequenceActions {
...
-                Future.successful(accounting)
-              } else {
-                // this is to short circuit the fold
-                Future.failed(FailedSequenceActivation(accounting)) // terminates the fold
-              }
+    val initialAccounting = SequenceAccounting(atomicActionCnt, ActivationResponse.payloadPlaceholder(inputPayload))
+
+    Future
+      .sequence(resolvedFutureActions)
+      .flatMap(actions =>
```
dubee: Gave refactoring a shot. Turned `resolvedFutureActions` into a `Future[List[resolvedFutureActions]]`. My understanding is that the `flapMap` block won't be executed if one or more of the actions does not exist and instead execution will jump to the second `recoverWith` on line 277.

dubee: @rabbah, what do you think?

rabbah: i think this is safer but now we wait for all the futures to resolve on the fetch - and don't pipeline any of the invokes, so worse on performance for a long sequence.

i think we need
1. additional test to cover missing package
2. additional test to cover missing bindings
3. a test with 2 sequence components, where the first is fetched but the second fails, the accounting object should show the execution that accounts for the first component's execution.

1 & 2 to make sure all no doc exceptions are handles consistently.
3 will require refactoring the code in the function so that you can test the independent components without having to mess with the data store. This is what I meant by refactoring.

markusthoemmes:
I think this fix has moved a bit in the wrong direction. The mode @rabbah wants to keep is lazy evaluation of the Futures. While it is true, that all actions will be fetched at the same time, invocation will only wait for the first Future to complete before it starts invoking the sequence. The failing operation will be `invokeNextAction`. We can either attach `recoverWith` to the `invokeNextAction` call or even recover this within `invokeNextAction`. We can then use the current `accounting` object (since that's known when calling `invokeNextAction`) to apply the failure there. That should keep state up to that point (another thing mentioned by @rabbah) and should keep the change simple.

Thoughts?

rabbah:
Thanks @markusthoemmes. That's much simpler and preserves the pipelining between fetches and invokes.

dubee:
PG3 2579 🔵

tests/src/test/scala/whisk/core/controller/test/SequenceApiTests.scala
```diff
@@ -53,6 +49,22 @@ class SequenceApiTests extends ControllerTestCommon with WhiskActionsApi {
 
   val allowedActionDuration = 120 seconds
 
+  it should "produce proper error when sequence component is missing during sequence invocation" in {
+    implicit val tid = transid()
+    val seqName = s"${aname()}_seq"
+    val compName = s"${aname()}_comp"
```
rabbah: the fix now looks correct above - can you improve the test however: the sequence should have more than one component and the result should show that the sequence partially activated.

dubee: Hmm, I think we need to create a integration test if we want to rely on the activation to show the sequence partially activated. Using a unit test would block for 60 seconds while the first action activated as the action doesn't actually get invoked in a unit test, but instead times out.

rabbah: I'd think something like this will work https://github.com/apache/incubator-openwhisk/blob/acbef6aa714f65b29c5fe67b68c7fd2ffbce159a/tests/src/test/scala/whisk/core/controller/test/ActionsApiTests.scala#L1239 (create a mock active ack). I didn't try so may be missing a detail why it doesn't.

dubee: Fancy, fancy. :)

dubee:
@rabbah, @markusthoemmes, thanks for all the feedback!!!

rabbah:
@dubee thanks for the improve test.

one last thought: perhaps the error message should tell you the component name that was missing; so that instead of "Sequence component does not exist.", it would say "Sequence component 'xyx' does not exist."

dubee:
@rabbah, makes sense to add the component name. I'll do some refactoring.

dubee:
@rabbah, how about I add the component names in an separate PR? Looks like there needs to be some refactoring around `NoDocumentException` to include entity names.

rabbah:
Thanks for looking into that.

PR #3806, opened 2018-06-25
add register and until to ansible invoker pull runtime images tasks

csantanapr:
`retries` and `delays` in ansible only applies to tasks/plugins that have `until` present
The `retries` and `delays` curently present are being ignored, this PR adds the `until` and changes to the new form of `loop` from `whith_items`

Ref: https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#do-until-loops

`If the until parameter isn’t defined, the value for the retries parameter is forced to 1.`

### DIFF
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -13,10 +13,10 @@
       "{{ (controller_index_base|int) + host_group.index(inventory_hostname) }}"
 
 - name: "pull the {{ docker.image.tag }} image of controller"
-  shell:
-    docker pull
-    {{ docker_registry~docker.image.prefix }}/controller:{{docker.image.tag}}
+  shell: "docker pull {{docker_registry}}{{ docker.image.prefix }}/controller:{{docker.image.tag}}"
   when: docker_registry != ""
+  register: result
+  until: (result.rc == 0)
   retries: "{{ docker.pull.retries }}"
   delay: "{{ docker.pull.delay }}"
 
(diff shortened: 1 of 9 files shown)


ansible/roles/controller/tasks/deploy.yml
```diff
@@ -13,10 +13,10 @@
       "{{ (controller_index_base|int) + host_group.index(inventory_hostname) }}"
 
 - name: "pull the {{ docker.image.tag }} image of controller"
-  shell:
-    docker pull
-    {{ docker_registry~docker.image.prefix }}/controller:{{docker.image.tag}}
+  shell: "docker pull {{docker_registry}}{{ docker.image.prefix }}/controller:{{docker.image.tag}}"
   when: docker_registry != ""
+  register: result
+  until: (result.rc == 0)
```
rabbah: How do these retries interact with existing retries?

csantanapr: I'm confuse, the task only has one entry for retires `retries: "{{ docker.pull.retries }}"`

Which other retries you are referring?

csantanapr:
Playground2/3332/ 🏃

csantanapr:
Playground2/3333/ 🔵

ansible/roles/controller/tasks/deploy.yml
rabbah: this is too broad --- if the image isn't found, this will retry 10 times to no avail. Can we restrict the error code for network failures instead?

PR #3812, opened 2018-06-26
ContainerClient + akka http alternative to HttpUtils

tysonnorris:
## Description

HttpUtils (http client for invoker -> action container) uses org.apache.http client that is synchronous and poor performing for concurrent requests. I ran into problems using it with concurrent activation support. Instead of trying to force that client to work, this PR is to work towards replacing it (or re-replacing it) with akka http based client.

This PR provides:
* add a ContainerClient trait to define the http client interface 
* refactoring of HttpUtils to implement this trait (HttpUtils remains default for now)
* add a PoolingContainerClient to provide a akka http based impl (based on PoolingRestClient)

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

Initial tests using this client to support this PR (for adding concurrency support in nodejs container) https://github.com/apache/incubator-openwhisk-runtime-nodejs/pull/41 were good, where tests requiring coordinated completion of 128 concurrent requests succeeded, while same tests with existing HttpUtils/org.apache.http client failed with as few as 3 concurrent requests.

More tests will be done, but wanted to get early feedback on this in general (wip labelled).


common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,205 @@
...
+    // - attaching a timeout to fail the promise
+    // - create a function to enqueue the request
+    // - retry (using same function) on StreamTcpException (only if retry == true)
+
+    val promise = Promise[HttpResponse]
+
+    // Timeout includes all retries.
+    as.scheduler.scheduleOnce(timeout) {
+      promise.tryFailure(new TimeoutException(s"Request to ${endpoint} could not be completed in time."))
+    }
```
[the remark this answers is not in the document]
tysonnorris: I've tried adding `.completionTimeout()` stage at the pool and the queue with no luck (was expecting a failure in that case, but don't get any success or failure...). I expected it to work at the pool. Any tips here?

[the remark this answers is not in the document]
tysonnorris: So I tried to do this using https://github.com/paypal/squbs/blob/master/squbs-ext/src/main/scala/org/squbs/streams/Timeout.scala#L268

I'm not wild about:
* dragging in squbs artifacts (seems heave handed for "just" adding a timeout)
* the `Try[HttpResponse]` is wrapped as a `Try[Try[HttpResponse]]` this seems awkward.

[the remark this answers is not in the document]
tysonnorris: ok idle-timeout is working, removed these changes and the `promise.tryFailure`

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,205 @@
...
+            // TCP error (e.g. connection couldn't be opened)
+            // Note: this is REQUIRED since the container may start, but ports are not listening before requests are made.
+            as.scheduler.scheduleOnce(retryInterval) { tryOnce() }
+          case t: Throwable =>
+            // Other error. We fail the promise.
+            promise.tryFailure(t)
+        }
+      }
+
+    tryOnce()
```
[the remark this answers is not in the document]
tysonnorris: much nicer! Thanks!

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,240 @@
...
+    //End retry handling
+
+    //map the HttpResponse to ContainerResponse
+    val r = promise.future
+      .flatMap({ response =>
+        val contentLength = response.entity.contentLengthOption.getOrElse(0l)
+        if (contentLength <= maxResponse.toBytes) {
+          Unmarshal(response.entity.withSizeLimit(maxResponse.toBytes)).to[String].map { o =>
+            //handle 204 as NoResponseReceived for parity with HttpUtils client
+            if (response.status == StatusCodes.NoContent) {
```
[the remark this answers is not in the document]
tysonnorris: This is to satisfy a test case that I transferred from `ContainerConnectionTests` to `PoolingContainerClientTests` - see `handle empty entity response`. 

Now this test is arguably wrong, compare to `not truncate responses within limit` (or one of them, at least), which returns a `null` and empty string as test responses (with a 200, not 204). 

`HttpUtils` (or `org.apache.http`) seems to vary from akka http in its handling for this case, where the `response.getEntity` is null on `HttpUtils`, but only when there is a 204 (not when there is a `null` sent as the response entity...)

I agree this is weird, but wanted to keep the tests at parity between `HttpUtils` and `PoolingContainerClient` at least for now.

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,240 @@
...
+        }
+      }
+
+    tryOnce()
+    //End retry handling
+
+    //map the HttpResponse to ContainerResponse
+    val r = promise.future
+      .flatMap({ response =>
+        val contentLength = response.entity.contentLengthOption.getOrElse(0l)
```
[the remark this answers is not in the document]
tysonnorris: Yes good catch; fixed

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,212 @@
...
+            }
+          case _ =>
+            //handle missing Content-Length as NoResponseReceived
+            //also handle 204 as NoResponseReceived, for parity with HttpUtils client
+            Future { Left(NoResponseReceived()) }
+      })
+      .recover {
+        case t: StreamTcpException => Left(Timeout(t))
+        case t: TimeoutException   => Left(Timeout(t))
+        case t: Throwable          => Left(ConnectionError(t))
```
chetanmeh: May be use `case NonFatal(t) => Left(ConnectionError(t))` to avoid handling of fatal errors

tysonnorris: sounds good!

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,213 @@
...
+                Right(ContainerResponse(response.status.intValue, s, Some(contentLength.B, maxResponse)))
+              }
+            }
+          case _ =>
+            //handle missing Content-Length as NoResponseReceived
+            //also handle 204 as NoResponseReceived, for parity with HttpUtils client
+            Future { Left(NoResponseReceived()) }
+      })
+      .recover {
+        case t: StreamTcpException => Left(Timeout(t))
```
tysonnorris: BTW - this is also parity with HttpUtils, but seems wrong. If there are retries on `StreamTcpException`, and after retrying until reaching the timeout period is still failing, we should broadcast the same exception, e.g. as `Left(ConnectionError(t))` right? 

AFAIK this is not checked anywhere, so I would prefer to change it

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,212 @@
...
+          akka.pattern.after(retryInterval, as.scheduler)(retryingRequest(req, newTimeout, retry))
+        } else {
+          logging.warn(this, s"POST failed with $t - no retry because timeout exceeded.")
+          Future.failed(t)
+        }
+      case t => Future.failed(t)
+    }
+  }
+
+  private def truncated(responseBytes: Source[ByteString, _],
```
chetanmeh: May be move it to `object PoolingContainerClient` and then have a test for this logic. Per [current test coverage][1] some flows are not covered

[1]: https://codecov.io/gh/apache/incubator-openwhisk/pull/3812/diff#D10-153

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,212 @@
...
+  port: Int,
+  timeout: FiniteDuration,
+  maxResponse: ByteSize,
+  queueSize: Int,
+  retryInterval: FiniteDuration = 100.milliseconds)(implicit logging: Logging, as: ActorSystem)
+    extends PoolingRestClient("http", hostname, port, queueSize, timeout = Some(timeout))
+    with ContainerClient
+    with AutoCloseable {
+
+  def close() = shutdown()
```
chetanmeh: `PoolingRestClient.shutdown` returns a `Future` while `ContainerClient.close` is defined to return a Unit. Should we wait for the result completion or change contract for `ContainerClient`?

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,212 @@
...
+        //ignore the tail (MUST CONSUME ENTIRE ENTITY!)
+        tail.runWith(Sink.ignore).map(_ => previouslyCaptured.utf8String)
+      case (Seq(prefix), tail) =>
+        val truncatedResponse = previouslyCaptured ++ prefix
+        if (truncatedResponse.size < maxResponse.toBytes) {
+          truncated(tail, truncatedResponse)
+        } else {
+          //ignore the tail (MUST CONSUME ENTIRE ENTITY!)
+          //captured string MAY be larger than the max response, so take only maxResponse bytes to get the exact length
+          tail.runWith(Sink.ignore).map(_ => truncatedResponse.take(maxResponse.toBytes.toInt).utf8String)
```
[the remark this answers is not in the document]
tysonnorris: It is as safe as truncation can be - client may get an error, but the `ActivationResponse` will end up with some info regarding the truncation.

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
tysonnorris: @markusthoemmes WDYT? This may affect use of `AutoCloseable` - for now I will return a Unit

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,223 @@
...
+              }
+            } else {
+              truncated(response.entity.dataBytes).map { s =>
+                Right(ContainerResponse(response.status.intValue, s, Some(contentLength.B, maxResponse)))
+              }
+            }
+          case _ =>
+            //handle missing Content-Length as NoResponseReceived
+            //also handle 204 as NoResponseReceived, for parity with HttpUtils client
+            Future { Left(NoResponseReceived()) }
```
markusthoemmes: Never use `Future.apply` if you already have the value of the Future at hand. It will schedule the value to the ExecutionContext unnecessarily.

Use `Future.successful(Left(NoResponseReceived())` instead (note how it doesn't require an ExecutionContext)

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,223 @@
...
+        }
+      })
+      .recover {
+        case t: StreamTcpException => Left(Timeout(t))
```
markusthoemmes: Is a StreamTcpException always a timeout?

common/scala/src/main/scala/whisk/core/containerpool/HttpUtils.scala
```diff
@@ -80,20 +85,17 @@ protected class HttpUtils(hostname: String, timeout: FiniteDuration, maxResponse
...
+    implicit tid: TransactionId): Future[Either[ContainerHttpError, ContainerResponse]] = {
     val entity = new StringEntity(body.compactPrint, StandardCharsets.UTF_8)
     entity.setContentType("application/json")
 
     val request = new HttpPost(baseUri.setPath(endpoint).build)
     request.addHeader(HttpHeaders.ACCEPT, "application/json")
     request.setEntity(entity)
 
-    execute(request, timeout, maxConcurrent, retry)
+    Future { execute(request, timeout, maxConcurrent, retry) }
```
markusthoemmes: Should add `blocking` here as well. This is using sync IO.

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
tysonnorris: No, but this is parity with HttpUtils - when timeout after retries, the `Timeout` response is used. I agree this is wrong, but wasn't sure how to otherwise make it compatible. If this compatibility is not a problem, I would change it?

common/scala/src/main/scala/whisk/core/containerpool/HttpUtils.scala
tysonnorris: Instead of `Future.successful`?

markusthoemmes: Yes, in this case you actually want another thread to take over, so you use `Future()`, but you also include `blocking` to denote that this is a blocking operation that might grab a Thread indefinitly. It gives the ExecutionContext the chance to adapt accordingly (create more threads)

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
tysonnorris: There is at least one test already that covers truncation: https://github.com/apache/incubator-openwhisk/pull/3812/files#diff-70e8c471d9056bda26d602a05f7ad091R180

Is codecov.io updated on each build? Can you tell why it wouldn't show in coverage? Theses tests `PoolingContainerClientTests` are using `ContainerClient` directly, no mocks etc, so I'm not sure why the coverage would not reflect?

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
tysonnorris: So I changed this so that on retry timeout, we don't rethrow a StreamTcpException, but rather a TimeoutException (with the message from StreamTcpException); this way we can project the Timeout consistency, but not imply that a mid-flight StreamTcpException is any indication of a timeout - WDYT?

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,218 @@
...
+   * @param body the JSON value to post (this is usually a JSON objecT)
+   * @param retry whether or not to retry on connection failure
+   * @return Left(Error Message) or Right(Status Code, Response as UTF-8 String)
+   */
+  def post(endpoint: String, body: JsValue, retry: Boolean)(
+    implicit tid: TransactionId): Future[Either[ContainerHttpError, ContainerResponse]] = {
+
+    //create the request
+    val req = Marshal(body).to[MessageEntity].map { b =>
+      //DO NOT reuse the connection (in case of paused containers)
```
markusthoemmes: In all cases actually, not just "in case of paused containers".

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,218 @@
...
+
+}
+
+/**
+ * This HTTP client is used only in the invoker to communicate with the action container.
+ * It allows to POST a JSON object and receive JSON object back; that is the
+ * content type and the accept headers are both 'application/json.
+ * The reason we still use this class for the action container is a mysterious hang
+ * in the Akka http client where a future fails to properly timeout and we have not
+ * determined why that is.
```
markusthoemmes: Please update the ScalaDoc.

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,205 @@
...
+import whisk.common.Logging
+import whisk.common.TransactionId
+import whisk.core.entity.ActivationResponse.ContainerHttpError
+import whisk.core.entity.ActivationResponse._
+import whisk.core.entity.ByteSize
+import whisk.core.entity.size.SizeLong
+import whisk.http.PoolingRestClient
+
+trait ContainerClient {
+  def close(): Unit
```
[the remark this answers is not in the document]
markusthoemmes: Should we remove the `close` method from the trait then and instead

```scala
trait ContainerClient extends AutoCloseable {
...
```
?

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,218 @@
...
+      // - http://github.com/akka/akka-http/tree/v10.1.3/akka-http-core/src/test/scala/akka/http/impl/engine/rendering/ResponseRendererSpec.scala#L470-L571
+      HttpRequest(HttpMethods.POST, endpoint, entity = b).withHeaders(Connection("close"))
+    }
+
+    retryingRequest(req, timeout, retry)
+      .flatMap(r => {
+        if (r._2 > 0) {
+          logging.info(this, s"completed after ${r._2} retries")
+        }
+        val response = r._1
```
markusthoemmes: You can unpack `response` and `retries` directly in the flapMap, like:

```scala
.flatMap { case (response, retries) =>
```

To avoid the tuple accessors.

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,218 @@
...
+                Right(ContainerResponse(response.status.intValue, o, None))
+              }
+            } else {
+              truncated(response.entity.dataBytes).map { s =>
+                Right(ContainerResponse(response.status.intValue, s, Some(contentLength.B, maxResponse)))
+              }
+            }
+          case _ =>
+            //handle missing Content-Length as NoResponseReceived
+            //also handle 204 as NoResponseReceived, for parity with HttpUtils client
```
markusthoemmes: I think we still need to drain the entity, like:

```scala
response.discardEntityBytes().future.map(_ => Left(NoResponseReceived())
```

since this case is also reached by an unknown Content-Length.

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,218 @@
...
+  }
+  //returns a Future HttpResponse -> Int (where Int is the retryCount)
+  private def retryingRequest(req: Future[HttpRequest],
+                              timeout: FiniteDuration,
+                              retry: Boolean,
+                              retryCount: Int = 0): Future[(HttpResponse, Int)] = {
+    request(req)
+      .map((_, retryCount))
+      .recoverWith {
+        case t: akka.stream.StreamTcpException if retry =>
```
markusthoemmes: Import the exception for readability?

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,212 @@
...
+
+  /** A helper method to post one single request to a connection. Used for container tests. */
+  def post(host: String, port: Int, endPoint: String, content: JsValue, timeout: Duration = 30.seconds)(
+    implicit logging: Logging,
+    as: ActorSystem,
+    ec: ExecutionContext,
+    tid: TransactionId): (Int, Option[JsObject]) = {
+    val connection = new PoolingContainerClient(host, port, 90.seconds, 1.MB, 1)
+    val response = executeRequest(connection, endPoint, content)
+    connection.close()
```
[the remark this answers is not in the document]
markusthoemmes: Need to move below the await then?

common/scala/src/main/scala/whisk/core/containerpool/Container.scala
```diff
@@ -166,16 +170,20 @@ trait Container {
     implicit transid: TransactionId): Future[RunResult] = {
     val started = Instant.now()
     val http = httpConnection.getOrElse {
-      val conn = new HttpUtils(s"${addr.host}:${addr.port}", timeout, 1.MB)
+      val conn = if (config.poolingClient) {
```
markusthoemmes: `poolingClient` seems an odd name since `HttpUtils` does pool as well. Should we rename to `newContainerClient`? (akka themselves did the same when they implemented the new ConnectionPool)

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,218 @@
...
+ * in the Akka http client where a future fails to properly timeout and we have not
+ * determined why that is.
+ *
+ * @param hostname the host name
+ * @param timeout the timeout in msecs to wait for a response
+ * @param maxResponse the maximum size in bytes the connection will accept
+ * @param queueSize once all connections are used, how big of queue to allow for additional requests
+ * @param retryInterval duration between retries for TCP connection errors
+ */
+protected class PoolingContainerClient(
```
markusthoemmes: Should we name this `AkkaContainerClient` or `NewContainerClient` for more clarity? Can we also rename `HttpUtils` to something more meaningful now, like `ApacheBlockingContainerClient`?

common/scala/src/main/scala/whisk/core/containerpool/ContainerClient.scala
```diff
@@ -0,0 +1,218 @@
...
+import spray.json._
+import whisk.common.Logging
+import whisk.common.TransactionId
+import whisk.core.entity.ActivationResponse.ContainerHttpError
+import whisk.core.entity.ActivationResponse._
+import whisk.core.entity.ByteSize
+import whisk.core.entity.size.SizeLong
+import whisk.http.PoolingRestClient
+
+trait ContainerClient {
```
markusthoemmes: The trait should be placed in its own file (or both ContainerClients should be placed in this one file, but I'd prefer one file for the trait, one for the akka based and one for the apache based)

common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
```diff
@@ -43,18 +44,24 @@ class PoolingRestClient(
...
   require(protocol == "http" || protocol == "https", "Protocol must be one of { http, https }.")
 
   protected implicit val context: ExecutionContext = system.dispatcher
   protected implicit val materializer: ActorMaterializer = ActorMaterializer()
 
+  //if specified, override the ClientConnection idle-timeout value
+  private val timeoutSettings = {
+    val ps = ConnectionPoolSettings(system.settings.config)
+    timeout.map(t => ps.withUpdatedConnectionSettings(_.withIdleTimeout(t))).getOrElse(ps)
+  }
```
markusthoemmes: Neat! 😁

core/invoker/src/main/resources/application.conf
```diff
@@ -29,6 +29,7 @@ whisk {
   container-pool {
     num-core: 4      # used for computing --cpushares, and max number of containers allowed
     core-share: 2    # used for computing --cpushares, and max number of containers allowed
+    pooling-client:  true # if true, use PoolingContainerClient for HTTP from invoker to action container (otherwise use HttpUtils)
```
markusthoemmes: Should we default to false for now? For safety?

common/scala/src/main/scala/whisk/http/PoolingRestClient.scala
tysonnorris: Thanks @chetanmeh !

tysonnorris:
@chetanmeh this is close - let me know if you think I missed any of your comments? I think I got them all. 

RE codecov - still some mysteries in the summary at top of PR conversation, but the report on codecov.io looks right to me. One anomaly in the PR conversation, I'm not sure where this is coming from?
```
...core/database/cosmosdb/RxObservableImplicits.scala | 0% <0%> (-100%)
```

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -190,29 +190,33 @@ class DockerContainer(protected val id: ContainerId,
     implicit transid: TransactionId): Future[RunResult] = {
     val started = Instant.now()
     val http = httpConnection.getOrElse {
-      val conn = new HttpUtils(s"${addr.host}:${addr.port}", timeout, ActivationEntityLimit.MAX_ACTIVATION_ENTITY_LIMIT)
+      val conn = new ApacheBlockingContainerClient(
+        s"${addr.host}:${addr.port}",
+        timeout,
+        ActivationEntityLimit.MAX_ACTIVATION_ENTITY_LIMIT)
```
markusthoemmes: Shouldn't this also check the feature toggle and use the correct client as configured?

tests/src/test/scala/whisk/core/containerpool/test/ContainerPoolTests.scala
```diff
@@ -296,7 +301,7 @@ class ContainerPoolTests
     val (containers, factory) = testContainers(2)
     val feed = TestProbe()
 
-    val pool = system.actorOf(ContainerPool.props(factory, ContainerPoolConfig(2, 2), feed.ref))
+    val pool = system.actorOf(ContainerPool.props(factory, ContainerPoolConfig(2, 2, false), feed.ref))
```
markusthoemmes: I think we should externalize building of the ContainerPoolConfig into a method so we don't need to adjust the values not needed for these tests continually. Check https://github.com/apache/incubator-openwhisk/pull/3767/files#diff-d00e1ef9ea3255332a28c35676361e29 for an impl. I did in another PR of exactly the same issue. I think it keeps the testcases clearer and reduces the diff in future PRs.

tests/src/test/scala/whisk/core/containerpool/kubernetes/test/KubernetesClientTests.scala
```diff
@@ -188,6 +189,7 @@ object KubernetesClientTests {
   implicit def strToInstant(str: String): Instant =
     strToDate(str).get
 
+  implicit val as = ActorSystem("kubernetes-client-tests-actor-system")
```
markusthoemmes: This actorSystem is leaked I think. Could you make the `TestKubernetesClient` take the actorSystem as an implicit parameter, so you can use the one imported (and closed) by the tests above?

Or maybe even move the class into the testclass. Makes it even easier?

tests/src/test/scala/whisk/core/containerpool/docker/test/ApacheBlockingContainerClientTests.scala
```diff
@@ -106,20 +113,41 @@ class ContainerConnectionTests
...
+    val end = Instant.now()
+    val waited = end.toEpochMilli - start.toEpochMilli
+    result should be('left)
+    result.left.get shouldBe a[Timeout]
+    result.left.get.asInstanceOf[Timeout].t shouldBe a[RetryableConnectionError]
+    result.left.get
+      .asInstanceOf[Timeout]
+      .t
+      .asInstanceOf[RetryableConnectionError]
+      .t shouldBe a[HttpHostConnectException]
```
markusthoemmes: This could be rewritten a little less verbose like:

```scala
result match {
  case Left(Timeout(RetryableConnectionError(_: HttpHostConnectException))) => // all good
  case _ =>
    fail(s"$result was not a Timeout(RetryableConnectionError(HttpHostConnectException)))")
}
```

I was not able to reproduce this locally though, the test always failed with a ConnectError in both implementations. Is this intermittent?

tests/src/test/scala/whisk/core/containerpool/docker/test/AkkaContainerClientTests.scala
```diff
@@ -0,0 +1,213 @@
...
+    }
+  }
+
+  it should "truncate large responses that exceed limit" in {
+    val timeout = 1.minute.toMillis
+    //use a limit large enough to not fit into a single ByteString as response entity is parsed into multiple ByteStrings
+    //seems like this varies, but often is ~64k or ~128k
+    val limit = 300.KB
+    val connection = new AkkaContainerClient(httpHost.getHostName, httpHost.getPort, timeout.millis, limit, 100)
+    Seq(true, false).foreach { code =>
```
markusthoemmes: Could you rename `code` to `success` or similar? Threw me off quite a bit when reading through (also in other occurences please).

tests/src/test/scala/whisk/core/containerpool/docker/test/AkkaContainerClientTests.scala
```diff
@@ -0,0 +1,213 @@
...
+  it should "retry till timeout on StreamTcpException" in {
+    val timeout = 5.seconds
+    val connection = new AkkaContainerClient("0.0.0.0", 12345, timeout, 1.B, 100)
+    val start = Instant.now()
+    val result = Await.result(connection.post("/init", JsObject.empty, retry = true), 10.seconds)
+    val end = Instant.now()
+    val waited = end.toEpochMilli - start.toEpochMilli
+    result should be('left)
+    result.left.get shouldBe a[Timeout]
+    result.left.get.asInstanceOf[Timeout].t shouldBe a[TimeoutException]
```
markusthoemmes: Same as below, could be rewritten to:

```scala
result match {
  case Left(Timeout(_ : TimeoutException) => // good
  case _ => fail(...)
}
```

IMO that pronounces the nesting of the exceptions more and is clearer in what you're actually testing. WDYT?

tests/src/test/scala/whisk/core/containerpool/docker/test/AkkaContainerClientTests.scala
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  it should "truncate large responses that exceed limit" in {
+    val timeout = 1.minute.toMillis
+    //use a limit large enough to not fit into a single ByteString as response entity is parsed into multiple ByteStrings
+    //seems like this varies, but often is ~64k or ~128k
+    val limit = 300.KB
+    val connection = new AkkaContainerClient(httpHost.getHostName, httpHost.getPort, timeout.millis, limit, 100)
+    Seq(true, false).foreach { code =>
+      Seq("0123456789" * 100000).foreach { r =>
```
markusthoemmes: Why even use `Seq.foreach` if you only pass in one value?

Please add a comment on what that value is supposed to be, i.e.

```scala
// Generate a response that's 1MB
val response = "0123456789" * 1024 * 1024
```

To make the numbers less magic.

common/scala/src/main/scala/whisk/core/containerpool/AkkaContainerClient.scala
```diff
@@ -0,0 +1,216 @@
...
+              s"POST failed after $retryCount retries with $t - no more retries because timeout exceeded.")
+            Future.failed(new TimeoutException(t.getMessage))
+          }
+      }
+  }
+
+  private def truncated(responseBytes: Source[ByteString, _],
+                        previouslyCaptured: ByteString = ByteString.empty): Future[String] = {
+    responseBytes.prefixAndTail(1).runWith(Sink.head).flatMap {
+      case (Seq(prefix), tail) =>
```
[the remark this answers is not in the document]
tysonnorris: I removed it since I was not able to establish a test that actually exercised that code path; can add it back for defense 👍

tests/src/test/scala/whisk/core/containerpool/kubernetes/test/KubernetesClientTests.scala
tysonnorris: I will give it a try, but it isn't clear why this test was setup this way? @dgrove-oss @jcrossley3 ?

tests/src/test/scala/whisk/core/containerpool/docker/test/ApacheBlockingContainerClientTests.scala
tysonnorris: nice - much better! (worked for me)

tests/src/test/scala/whisk/core/containerpool/docker/test/AkkaContainerClientTests.scala
tysonnorris: agreed

tests/src/test/scala/whisk/core/containerpool/docker/test/AkkaContainerClientTests.scala
tysonnorris: agreed

tests/src/test/scala/whisk/core/containerpool/kubernetes/test/KubernetesClientTests.scala
tysonnorris: The `TestKubernetesClient` is shared with `KubernetesContainerTests` - so for now, changed to implicit ActorSystem (and left object/classes in current places). Good?

dgrove-oss: I don't remember any particular reason;  I think it may be as simple as that since we hadn't needed to have our hands on an ActorSystem before in the stubbed out test client it wasn't plumbed through.

tysonnorris:
@chetanmeh @markusthoemmes I think I have addressed all comments

common/scala/src/main/scala/whisk/core/database/CouchDbRestStore.scala
```diff
@@ -513,7 +514,7 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
       .getOrElse(Future.successful(true)) // For CouchDB it is expected that the entire document is deleted.
 
   override def shutdown(): Unit = {
-    Await.ready(client.shutdown(), 1.minute)
+    Await.result(client.shutdown(), 30.seconds)
```
chetanmeh: Do we need to change this for this PR?

tysonnorris:
@markusthoemmes I added one more forgotten required change: enable ActionContainer based tests to explicitly use akka vs apache http client. While it is true that this could be accomplished by coercing test's akka config to include `akka-client: true`,  I think that until akka is the default client, tests should not be forced to run with multiple configs, but we still need to be able to force tests to run via a specific client (e.g. concurrency tests will fail on apache client, but not the akka client...).
For example forcing use of akka client will be done in container tests by changing: 
`withContainer(nodejsContainerImageName, env)(code)` to 
`withContainer(nodejsContainerImageName, env, true)(code)`
WDYT?

rabbah:
If the additional parameter doesn’t cause breaking downstream changes 👍 otherwise a separate method that tests can opt into.

markusthoemmes:
I'm a bit unsure about a "per-test" flag here. I'd be okay to make the new akka-client the default for all those tests straight away (after a decent amount of local test runs to squash out obvious heisenbugs). No need to gradually enable them one after the other. WDYT?

tysonnorris:
@rabbah we would need to run the same test cases twice for each test, with a different config each run. I'm not sure how to simply enable this structure without rewriting the tests?

Separately, if that is possible, we still need a way to disable some clients from being used in some tests - the reason I arrived at creating the new client is the concurrency tests simply don't work with the old apache client (at least they don't work in travis).

tysonnorris:
@markusthoemmes @rabbah I'm ok to switch all the tests to use akka http (and not apache http) after running them locally to verify, if that works? (we can either run all akka, or a mix of akka and apache, but NOT all apache)

rabbah:
i'd say run them all with the new client -- if it can't handle the unit tests, we have a problem ;)

tysonnorris:
Testing runtime repos, I ran into (so far) an issue with the proxy test in incubator-openwhisk-runtime-docker, where instead of receiving a `Timeout` we now (with akka client) receive `java.lang.IllegalStateException: The http server closed the connection unexpectedly before delivering responses for 1 outstanding requests`

tysonnorris:
So for docker "bad proxy" and "bad action" tests, to keep behavior parity, we need to change retry logic from:
```
case t: StreamTcpException if retry
```
to 
```
case t: RuntimeException if retry
``` 
because in some cases, akka http throws `UnexpectedConnectionClosureException`, which is private but extends `RuntimeException`. 
Some discussion is here: https://github.com/akka/akka-http/issues/768 in case you wonder about catching this specific exception...

Alternatively, we can update test cases for this edgy behavior handling?

tysonnorris:
OK, tested all runtime repo tests; one change required in incubator-openwhisk-runtime-docker (see https://github.com/apache/incubator-openwhisk-runtime-docker/pull/54)

This is OK to merge once @markusthoemmes approves.

tysonnorris:
@markusthoemmes any other concerns here?

markusthoemmes:
PG2 3412 🔵

PR #3813, opened 2018-06-27
Recover image pulls by trying to run the container anyways.

markusthoemmes:
A `docker pull` can fail due to various reasons. One of them is network throttling by the image registry. Since we try to pull on each blackbox invocation, high volume load can cause lots of errors due to failing pulls unnecessarily.

Instead, we try to run the container even if the pull failed in the first place. If the image is available locally, the container will start just fine and recover the error gracefully. If the image is not available locally, the run will fail as well and return the same error as the docker pull would've returned.

This behavior will only be enabled for blackbox actions that specify a tag. Blackbox actions not using a tag *or* using "latest" as a tag will exhibit the very same behavior as today. That is: There will always be a pull before each container start and a failing pull will result in an error reported to the user. This is to enable rapid prototyping on images and enable determinism in the workflow. Updating the action will then force a pull and will fail early if that pull fails. With the new behavior, the developer might be surprised by the image not being updated because the pull is swallowed.

For production workload it is considered best-practice to version images through labels, thus we can "safely" assume that we can fall back to a local image in case the pull fails.

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [X] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
@@ -39,6 +39,7 @@ import akka.stream.stage._
 import akka.util.ByteString
 import spray.json._
 import whisk.core.containerpool.logging.LogLine
+import whisk.core.entity.ExecManifest.ImageName
 import whisk.http.Messages
 
 object DockerContainer {
@@ -54,9 +55,7 @@ object DockerContainer {
    * Creates a container running on a docker daemon.
    *
    * @param transid transaction creating the container
-   * @param image image to create the container from
-   * @param userProvidedImage whether the image is provided by the user
-   *     or is an OpenWhisk provided image
+   * @param image either a user provided (Left) or OpenWhisk provided (Right) image
    * @param memory memorylimit of the container
    * @param cpuShares sharefactor for the container
    * @param environment environment variables to set on the container
@@ -67,8 +66,7 @@ object DockerContainer {
    * @return a Future which either completes with a DockerContainer or one of two specific failures
    */
   def create(transid: TransactionId,
-             image: String,
-             userProvidedImage: Boolean = false,
+             image: Either[ImageName, String],
              memory: ByteSize = 256.MB,
              cpuShares: Int = 0,
              environment: Map[String, String] = Map.empty,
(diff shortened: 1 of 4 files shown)


core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -104,22 +104,29 @@ object DockerContainer {
...
       name.map(n => Seq("--name", n)).getOrElse(Seq.empty) ++
       params
+
     val pulled = if (userProvidedImage) {
-      docker.pull(image).recoverWith {
-        case _ => Future.failed(BlackboxStartupError(Messages.imagePullError(image)))
-      }
-    } else Future.successful(())
+      // Always recover the Future so docker run is tried below in any case, in case we have an image available locally
+      docker.pull(image).map(_ => true).recover { case _ => false }
```
rabbah: https://github.com/apache/incubator-openwhisk/pull/2852/files

not sure if this is better but this was an old attempt to skip docker pull if image is present locally using docker inspect.

docs/actions-docker.md
```diff
@@ -105,6 +105,10 @@ The instructions that follow show you how to use the OpenWhisk Docker skeleton.
   Notice the use of `--docker` when creating an action. Currently all Docker images are assumed
   to be hosted on Docker Hub.
 
+  *Note:* It is considered best-practice for production images to be versioned via docker image tags. The tag "latest" or no tag will guarantee to always be on the latest state when creating new containers, but that contains the possibility of failing because pulling the image fails. For tagged images however, the system will allow a failing pull and gracefully recover it by using the image that is already locally available, making it much more resilient against Dockerhub outages etc.
```
rabbah: Suggested rewrite for the first sentence:

The absence of a tag will be treated the same as using the tag "latest" which will pull an image from Dockerhub. Please note that "latest" doesn't mean newest tag, but rather "latest" is an alias for any image built without an explicit tag.

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -104,22 +102,44 @@ object DockerContainer {
...
-      }
-    } else Future.successful(())
+
+    val imageToUse = image.fold(_.publicImageName, identity)
+
+    val pulled = image match {
+      case Left(userProvided) if userProvided.tag.map(_ == "latest").getOrElse(true) =>
+        // Iff the image tag is "latest" explicitly (or implicitly because no tag is given at all), failing to pull will
+        // fail the whole container bringup process, because it is expected to pick up the very latest version every
+        // time.
```
rabbah: the very latest "untagged" version every time.

PR #3821, opened 2018-06-28
Avoid regenerating auth key multiple times

chetanmeh:
Scala based `user create foo` command which sets a random auth key is printing wrong key in output.

## Description
This happens because in case authkey is not provided then current impl was generating the key multiple times. So they key which gets saved in db differs from one printed in output.

This PR fixes that by generating the key only once if none provided

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [x] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/tests/src/test/scala/whisk/core/database/UserCommandTests.scala
+++ b/tests/src/test/scala/whisk/core/database/UserCommandTests.scala
@@ -62,6 +62,12 @@ class UserCommandTests extends FlatSpec with WhiskAdminCliTestBase {
     admin.executeCommand().futureValue.right.get shouldBe key.compact
   }
 
+  it should "create a user with default key" in {
+    val subject = newSubject()
+    val generatedKey = resultOk("user", "create", subject)
+    resultOk("user", "get", subject) shouldBe generatedKey
+  }
+
   it should "add namespace to existing user" in {
     val subject = newSubject()
     val key = AuthKey()
--- a/tools/admin/src/main/scala/whisk/core/database/UserCommand.scala
+++ b/tools/admin/src/main/scala/whisk/core/database/UserCommand.scala
@@ -85,9 +85,7 @@ class UserCommand extends Subcommand("user") with WhiskCommand {
 
     def isUUID(u: String) = Try(UUID.fromString(u)).isSuccess
 
-    def desiredNamespace = Namespace(EntityName(namespace.getOrElse(subject()).trim), authKey.uuid)
-
-    def authKey: AuthKey = auth.map(AuthKey(_)).getOrElse(AuthKey())
+    def desiredNamespace(authKey: AuthKey) = Namespace(EntityName(namespace.getOrElse(subject()).trim), authKey.uuid)
   }
 
   val create = new CreateUserCmd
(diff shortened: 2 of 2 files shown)


tools/admin/src/main/scala/whisk/core/database/UserCommand.scala
```diff
@@ -85,9 +85,7 @@ class UserCommand extends Subcommand("user") with WhiskCommand {
 
     def isUUID(u: String) = Try(UUID.fromString(u)).isSuccess
 
-    def desiredNamespace = Namespace(EntityName(namespace.getOrElse(subject()).trim), authKey.uuid)
-
-    def authKey: AuthKey = auth.map(AuthKey(_)).getOrElse(AuthKey())
+    def desiredNamespace(authKey: AuthKey) = Namespace(EntityName(namespace.getOrElse(subject()).trim), authKey.uuid)
```
rabbah: subject is a `val`, should drop the `()` here and elsewhere.

rabbah: probably makes sense to add the trim on `val subject = ...trim`?

tests/src/test/scala/whisk/core/database/UserCommandTests.scala
```diff
@@ -62,6 +62,12 @@ class UserCommandTests extends FlatSpec with WhiskAdminCliTestBase {
     admin.executeCommand().futureValue.right.get shouldBe key.compact
   }
 
+  it should "create a user with default key" in {
```
rabbah: can you audit the existing wskadmin tests, are there any other tests we should add here?

tools/admin/src/main/scala/whisk/core/database/UserCommand.scala
```diff
@@ -171,22 +169,26 @@ class UserCommand extends Subcommand("user") with WhiskCommand {
...
+          val auth =
+            WhiskAuth(Subject(create.subject()), Set(WhiskNamespace(create.desiredNamespace(authKey), authKey)))
+          authStore.put(auth).map(_ => Right(authKey.compact))
       }
-    }
-  }.recoverWith {
-    case _: NoDocumentException =>
-      val auth =
-        WhiskAuth(Subject(create.subject()), Set(WhiskNamespace(create.desiredNamespace, create.authKey)))
-      authStore.put(auth).map(_ => Right(create.authKey.compact))
```
rabbah: ha - good one... if `authkey` is a `val` instead of a `def` might have saved you too?

tools/admin/src/main/scala/whisk/core/database/UserCommand.scala
chetanmeh: `subject` here is of type [ScallopOption[String]][1] and `subject()` invokes the [apply][2] method to extract the underlying value. Hence need to use `subject()` approach

[1]: http://scallop.github.io/scallop/org/rogach/scallop/ScallopOption.html
[2]: http://scallop.github.io/scallop/org/rogach/scallop/ScallopOption.html#apply():A

tests/src/test/scala/whisk/core/database/UserCommandTests.scala
chetanmeh: sure would review them

tools/admin/src/main/scala/whisk/core/database/UserCommand.scala
chetanmeh: Yes. Probably a `lazy val` should have worked as initialization needs to be deferred untill parsing is done

tools/admin/src/main/scala/whisk/core/database/UserCommand.scala
rabbah: aha.

PR #3838, opened 2018-07-03
add /numHealthyInvokers route to controller

dgrove-oss:
Add a route to easily get a count of healthy invokers. This route can
be used to as a check for the system being ready to actually execute
actions (for example as a readiness probe in a Helm chart or a
TravisCI built script). Although this information could be extracted by
client-side processing of the result of the /invokers route, it is
convenient to do the math on the server side and simply return the count.

## Related issue and scope

- [ ] API
- [ x ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/Controller.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/Controller.scala
@@ -43,7 +43,7 @@ import whisk.core.entitlement._
 import whisk.core.entity._
 import whisk.core.entity.ActivationId.ActivationIdGenerator
 import whisk.core.entity.ExecManifest.Runtimes
-import whisk.core.loadBalancer.LoadBalancerProvider
+import whisk.core.loadBalancer.{Healthy, LoadBalancerProvider}
 import whisk.http.BasicHttpService
 import whisk.http.BasicRasService
 import whisk.spi.SpiLoader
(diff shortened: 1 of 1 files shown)


core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -151,6 +151,25 @@ class Controller(val instance: ControllerInstanceId,
...
+   * @return number of healthy invokers
+   */
+  private val internalInvokerHealthCount = {
+    implicit val executionContext = actorSystem.dispatcher
+    (path("numHealthyInvokers") & get) {
+      complete {
+        loadBalancer
+          .invokerHealth()
+          .map(_.foldLeft(0) { (r, i) =>
+            if (i.status == Healthy) { r + 1 } else { r }
```
rabbah: can we do this with a query param on the existing route, or perhaps better as a sub resource in the invoker path?

`/invokers`, `/invokers/healthy`

markusthoemmes: Also: Scala conveniently has a `count` method on its collections, so you can simplify this to:

```scala
loadBalancer.invokerHealth()map(_.count(_._2 == Healthy).toJson)
```

dgrove-oss: as discussed on slack, I am going to change to use the route `/invokers/healthy/count`

PR #3843, opened 2018-07-04
Properly consume and optimize reading of http entities, drop connection reuse.

markusthoemmes:
**Notice:** This has been quite an odyssey to fix and we needed to roll back the change the last time. Please consider trying the changes out in an environment where you can assert latency and stability for different runtimes etc.

I myself have run:
- 3x Latency simulation (important, caught a regression last time) => **no regression**
- 3x Warm action simulation => **no regression**
- 1x Cold action simulation => **no regression**

## Description

- Explicitly consume and close the response's entity in any case (even if ignored) to make sure the connections get released properly and are not leaked.
- Use optimized path for consuming the entire entity into a string if its length is within bounds.
- Don't reuse connections.

To the last point: Reusing connections when a runtime doesn't support it adds a significant latency overhead when closing the response's entity. That's likely due to some sort of mismatched behavior (client wants to keep the connection open, the server doesn't even know the concept). The latency overhead in those cases (10-20ms of latency added in latency tests) seem far higher than the latency overhead of establishing a new connection (not measurable in latency tests).
Dropping connection reuse also solves any issues that might be encountered due to the pause/resume cycles of the container, in which sockets don't react to any event at all.

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [X] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/build.gradle
+++ b/common/scala/build.gradle
@@ -51,7 +51,7 @@ dependencies {
     compile 'commons-io:commons-io:2.6'
     compile 'commons-collections:commons-collections:3.2.2'
     compile 'org.apache.kafka:kafka-clients:0.11.0.1'
-    compile ('org.apache.httpcomponents:httpclient:4.4.1') {
+    compile ('org.apache.httpcomponents:httpclient:4.5.5') {
         exclude group: 'commons-logging'
     }
     compile ('com.fasterxml.uuid:java-uuid-generator:3.1.3') {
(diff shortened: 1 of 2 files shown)


chetanmeh:
@markusthoemmes Any pointers on how you ran the performance tests.

markusthoemmes:
@chetanmeh See https://github.com/apache/incubator-openwhisk/tree/master/tests/performance#gatling. I used the Gatling simulations in there.

Issue #3852, opened 2018-07-09
CouchDB member won't rejoin the old cluster automatically after restart


jiangpengcheng:
see this https://github.com/apache/couchdb/issues/1421

The reason is that each CouchDB container has a different [erlang magic cookie](http://erlang.org/doc/reference_manual/distributed.html#security),  which make connections  being rejected

The solution is set an identical cookie for every CouchDB container

style95:
resolved by #3853

PR #3853, opened 2018-07-09
Set erlang magic cookie for couchdb

jiangpengcheng:
Set identical erlang magic cookie for couchdb

## Description
This commit try to set an identical erlang magic cookie for each couchdb container, so that they  can rejoin the cluster after being restarted

## Related issue and scope
- [x] I opened an issue to propose and discuss this change (#3852)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/ansible/roles/couchdb/tasks/deploy.yml
+++ b/ansible/roles/couchdb/tasks/deploy.yml
@@ -7,6 +7,10 @@
   set_fact:
     coordinator: "{{ groups['db'][0] }}"
 
+- name: "Set the volumes"
+  set_fact:
+    volumes: []
+
 - name: check if db credentials are valid for CouchDB
   fail: msg="The db provider in your {{ hosts_dir }}/group_vars/all is {{ db.provider }}, it has to be CouchDB, pls double check"
   when: db.provider != "CouchDB"
@@ -20,8 +24,16 @@
   vars:
     instance: "{{instances | selectattr('name', 'equalto', 'db') | list | first}}"
   set_fact:
-    volume_dir: "{{ instance.volume.fsmount | default( '/mnt/' + group_names|first, true ) }}:/usr/local/var/lib/couchdb"
+    volumes: "{{ volumes }} + [ '{{ instance.volume.fsmount | default( '/mnt/' + group_names|first, true ) }}:/usr/local/var/lib/couchdb' ]"
   when: (block_device is defined) and (block_device in disk_status.stdout)
+
+- include_tasks: gen_erl_cookie.yml
+  when: (db.instances|int >= 2)
+
+- name: "set the erlang cookie volume"
+  set_fact:
+    volumes: "{{ volumes }} + [ '{{ config_root_dir }}/erlang.cookie:/opt/couchdb/.erlang.cookie' ]"
+  when: (db.instances|int >= 2)
 
 - name: "(re)start CouchDB from '{{ couchdb_image }} ' "
   vars:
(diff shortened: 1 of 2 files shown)


ansible/roles/couchdb/tasks/deploy.yml
```diff
@@ -23,6 +23,22 @@
     volume_dir: "{{ instance.volume.fsmount | default( '/mnt/' + group_names|first, true ) }}:/usr/local/var/lib/couchdb"
   when: (block_device is defined) and (block_device in disk_status.stdout)
 
+- name: "generate erlang cookie"
+  local_action: command openssl rand -base64 32op
+  register: erlang_cookie
+  run_once: true
+
+- name: ensuer config root dir exists
```
markusthoemmes: typo: "ensure"

ansible/roles/couchdb/tasks/deploy.yml
```diff
@@ -23,6 +23,22 @@
     volume_dir: "{{ instance.volume.fsmount | default( '/mnt/' + group_names|first, true ) }}:/usr/local/var/lib/couchdb"
   when: (block_device is defined) and (block_device in disk_status.stdout)
 
+- name: "generate erlang cookie"
```
vvraskin: Thanks for submitting this PR!
Do we need to generate the cookie even if a single couchdb node is deployed?

jiangpengcheng: don't need to

ansible/roles/couchdb/tasks/deploy.yml
```diff
@@ -23,6 +23,35 @@
...
+    path: "{{ config_root_dir }}"
+    state: directory
+
+# when enable uid namespace mode, couchdb container doesn't have permission to change the owner of files which mounted
+# from host, so the container will be failed to start. Use a temporary container here to create the file erlang.cookie
+# with the correct user 'couchdb' as its owner
+- name: "create the erlang cookie file on remote"
+  vars:
+    couchdb_image: "{{ couchdb.docker_image | default('apache/couchdb:' ~ couchdb.version ) }}"
+  command: "docker run --rm -v /tmp:/tmp -u couchdb {{ couchdb_image }} sh -c 'echo {{ erlang_cookie }} >> /tmp/erlang.cookie'"
```
vvraskin: The same `db.clustering.enabled` could also safe some time for couchdb deployment

vvraskin: Or maybe checking this condition would be enough as well: 
`when: (db.instances|int >= 2)`

jiangpengcheng: updated

jiangpengcheng:
recheck

vvraskin:
Should the cookie generation task run for the first cluster node only?
e.g. `when: (inventory_hostname == coordinator) and (db.instances|int >= 2)`

vvraskin:
Or should the cookie be always unique? What if we don't generate it each time when couchdb is deployed and use a static property instead?

jiangpengcheng:
@vvraskin thanks for the test
1. it's hard to find out the real uid of user couchdb on the host when docker enable the user namespace feature,  so I think the `become: true` is the right way, and it's normal for prompting for the sudo password, we can either use option `--ask-sudo-pass`, host var `ansible_sudo_pass` or enable the NOPASSWD for users on db hosts
 
2. I add an option for users to use their own value erlang cookie:
```
- name: "generate erlang cookie"
  local_action: command openssl rand -base64 32op
  register: random_stdout
  run_once: true
  when: erlang_cookie is not defined

- set_fact:
    erlang_cookie: "{{ random_stdout.stdout }}"
  when: erlang_cookie is not defined
``` 
we can pass an extra var `-e erlang_cookie=xxx`

PR #3855, opened 2018-07-09
Add invoker container name to health protocol

vvraskin:
This PR aims to add invoker container name into the health protocol for the case when the name differs from the Kafka topic name.
## Description

Previous response from the invokers endpoint:
```
curl https://localhost:8080/invokers -k
{"invoker0":"up"}
```
New response if invoker container name differs from the kafka topic name:
```
curl https://localhost:8080/invokers -k
{"invoker0/invoker-j8ptk":"up"}
```

## Related issue and scope

- [x] Controller
- [x] Invoker

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/core/entity/InstanceId.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/InstanceId.scala
@@ -21,7 +21,7 @@ import spray.json.DefaultJsonProtocol
 import whisk.core.entity.ControllerInstanceId.LEGAL_CHARS
 import whisk.core.entity.ControllerInstanceId.MAX_NAME_LENGTH
 
-case class InvokerInstanceId(val instance: Int, name: Option[String] = None) {
+case class InvokerInstanceId(val instance: Int, name: Option[String] = None, hostname: Option[String] = None) {
   def toInt: Int = instance
 }
 
@@ -32,7 +32,7 @@ case class ControllerInstanceId(val asString: String) {
 }
 
 object InvokerInstanceId extends DefaultJsonProtocol {
-  implicit val serdes = jsonFormat2(InvokerInstanceId.apply)
+  implicit val serdes = jsonFormat3(InvokerInstanceId.apply)
 }
 
 object ControllerInstanceId extends DefaultJsonProtocol {
--- a/common/scala/src/main/scala/whisk/core/entity/InstanceId.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/InstanceId.scala
@@ -21,8 +21,17 @@ import spray.json.DefaultJsonProtocol
 import whisk.core.entity.ControllerInstanceId.LEGAL_CHARS
 import whisk.core.entity.ControllerInstanceId.MAX_NAME_LENGTH
 
-case class InvokerInstanceId(val instance: Int, name: Option[String] = None, hostname: Option[String] = None) {
+// instance - a numeric value used for the load balancing and Kafka topic creation
+// uniqueName - is used for dynamic instance assignment by Zookeeper
+// displayedName - is used in health protocol to correlate Kafka topics with invoker container names
+case class InvokerInstanceId(val instance: Int,
+                             uniqueName: Option[String] = None,
+                             displayedName: Option[String] = None) {
   def toInt: Int = instance
+
+  def getOrEmpty(str: Option[String]): String = str.map(n => "/" + n).getOrElse("")
+
+  override def toString: String = "invoker" + instance + getOrEmpty(uniqueName) + getOrEmpty(displayedName)
 }
 
 case class ControllerInstanceId(val asString: String) {
--- a/common/scala/src/main/scala/whisk/core/entity/InstanceId.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/InstanceId.scala
@@ -21,17 +21,19 @@ import spray.json.DefaultJsonProtocol
 import whisk.core.entity.ControllerInstanceId.LEGAL_CHARS
 import whisk.core.entity.ControllerInstanceId.MAX_NAME_LENGTH
 
-// instance - a numeric value used for the load balancing and Kafka topic creation
-// uniqueName - is used for dynamic instance assignment by Zookeeper
-// displayedName - is used in health protocol to correlate Kafka topics with invoker container names
+/**
+ * An instance id representing an invoker
+ *
+ * @param instance a numeric value used for the load balancing and Kafka topic creation
+ * @param uniqueName an identifier required for dynamic instance assignment by Zookeeper
+ * @param displayedName an identifier that is required for the health protocol to correlate Kafka topics with invoker container names
+ */
 case class InvokerInstanceId(val instance: Int,
                              uniqueName: Option[String] = None,
                              displayedName: Option[String] = None) {
   def toInt: Int = instance
 
-  def getOrEmpty(str: Option[String]): String = str.map(n => "/" + n).getOrElse("")
-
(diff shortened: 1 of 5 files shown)


core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -291,7 +291,9 @@ class ShardingContainerPoolBalancer(config: WhiskConfig, controllerInstance: Con
     val start = transid.started(
       this,
       LoggingMarkers.CONTROLLER_KAFKA,
-      s"posting topic '$topic' with activation id '${msg.activationId}'",
+      invoker.hostname
+        .map(h => s"posting topic '$topic' to invoker '$h' with activation id '${msg.activationId}'")
+        .getOrElse(s"posting topic '$topic' with activation id '${msg.activationId}'"),
```
markusthoemmes: Could we also print the whole instanceId?

As in: `s"posting to $invoker using topic '$topic' with activation id '${msg.activationId}'"`

Should then log the whole case class with its respective members.

rabbah:
hostname first would be more natural no?

core/controller/src/main/scala/whisk/core/controller/Controller.scala
```diff
@@ -146,7 +146,12 @@ class Controller(val instance: ControllerInstanceId,
         complete {
           loadBalancer
             .invokerHealth()
-            .map(_.map(i => s"invoker${i.id.toInt}" -> i.status.asString).toMap.toJson.asJsObject)
+            .map(_.map(i => {
+              i.id.hostname
+                .map(hostname => s"invoker${i.id.toInt}/$hostname" -> i.status.asString)
+                .orElse(Some(s"invoker${i.id.toInt}" -> i.status.asString))
+                .get
```
markusthoemmes: `get` is unsafe and can throw exceptions. In this case you can statically guarantee no exceptions by using `getOrElse` like:

```scala
_.map(i => {
  val key = i.id.name.map(name => s"invoker${i.id.toInt}/$name").getOrElse(invoker${i.id.toInt})
  key -> i.status.asString
}).toMap.toJson.asJsObject
```

core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -164,11 +164,15 @@ object Invoker {
         zkClient.close()
         assignedId
       }
+    val topicBaseName = "invoker"
+    val topicName = topicBaseName + assignedInvokerId
 
-    val invokerInstance = InvokerInstanceId(assignedInvokerId, invokerName)
+    // Define invoker hostname for the health protocol to distinguish invoker topics in Kafka from the actual container names
+    val invokerHostname = scala.util.Properties.envOrNone("HOSTNAME").filter(name => name != topicName)
+    val invokerInstance = InvokerInstanceId(assignedInvokerId, invokerName, invokerHostname)
```
markusthoemmes: As discussed in slack, let's investigate if we can achieve the wanted functionality by using the already existing `name` Option.

common/scala/src/main/scala/whisk/core/entity/InstanceId.scala
```diff
@@ -21,7 +21,7 @@ import spray.json.DefaultJsonProtocol
 import whisk.core.entity.ControllerInstanceId.LEGAL_CHARS
 import whisk.core.entity.ControllerInstanceId.MAX_NAME_LENGTH
 
-case class InvokerInstanceId(val instance: Int, name: Option[String] = None) {
+case class InvokerInstanceId(val instance: Int, name: Option[String] = None, hostname: Option[String] = None) {
   def toInt: Int = instance
```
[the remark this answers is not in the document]
tysonnorris: Yes please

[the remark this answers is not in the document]
vvraskin: @tysonnorris do you use unique invoker container id as the 'name' together with dynamic kafka topic assignment?

I hesitate to use the 'name' here because of the dynamic Kafka topic assignment and its edge cases. For example in case of kube image reload, an invoker will get a new instance id which will cause a new topic creation in Kafka. This could potentially lead to some left over activations on the previous topic.

rabbah:
can you use a command line argument like the "unique name" and instance id? also what's the difference between a display name and a unique name --- more specifically why do we need two instead of just one.

vvraskin:
Unique name[1] is used for dynamic kafka topic assignment while display name[2] is for the mapping between invoker container name and its kafka topic. 
For the case of Kube based deployments, we use worker node ip addresses as [1]. This allows invoker topic stickiness for the case of worker OS reload. 
[2] is the invoker docker container name in kube. If one deploys Invokers as k8s "Deployment" each node gets it container name dynamically in a form `invoker-j8ptk`, currently each invoker reports its status via health protocol in a form `invoker0`. And we cannot say which invoker topic belongs to which invoker container in the deployment. Thus [2] is a conventional mapping between invoker topic and actual container name we see in kube.

In theory I can pack [2] into [1], but then each time the worker node gets OS reloaded, the invoker container will receive a new container name, which means that a new kafka topic will be created. This could potentially cause an issue, e.g. if Kafka topic of the invoker was not empty before the reload its activations will hang on that topic.

rabbah:
Thanks @vvraskin for the improvements and addressing the comments. The comment about kafka topics may be related to @markusthoemmes' point here https://github.com/apache/incubator-openwhisk/issues/3858#issuecomment-404059138.

core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -291,7 +291,7 @@ class ShardingContainerPoolBalancer(config: WhiskConfig, controllerInstance: Con
     val start = transid.started(
       this,
       LoggingMarkers.CONTROLLER_KAFKA,
-      s"posting topic '$topic' with activation id '${msg.activationId}'",
+      s"posting topic '$topic' to '${invoker.toString}' with activation id '${msg.activationId}'",
```
rabbah: since using `toString` this can become `$invoker`.

common/scala/src/main/scala/whisk/core/entity/InstanceId.scala
```diff
@@ -21,8 +21,19 @@ import spray.json.DefaultJsonProtocol
...
+ * @param instance a numeric value used for the load balancing and Kafka topic creation
+ * @param uniqueName an identifier required for dynamic instance assignment by Zookeeper
+ * @param displayedName an identifier that is required for the health protocol to correlate Kafka topics with invoker container names
+ */
+case class InvokerInstanceId(val instance: Int,
+                             uniqueName: Option[String] = None,
+                             displayedName: Option[String] = None) {
   def toInt: Int = instance
+
+  override def toString: String = (Seq("invoker" + instance) ++ uniqueName ++ displayedName).mkString("/")
```
rabbah: perhaps add `asString` as well (asString = toString) to use here instead https://github.com/apache/incubator-openwhisk/pull/3855/files#diff-21928219385e4e0fe9b1d25e26bf2eb1R149.

dgrove-oss:
This looks good to me now.  Since we are deploying on several different container orchestration systems, I think it is useful have separate "names" to distinguish between a transient container name vs. a logical invoker name vs. a kafka topic used by the load balancer (which because of the current load balancer implementation is effectively constrained to be a densely back natural number).

vvraskin:
PG2/3369 is :large_blue_circle:

PR #3874, opened 2018-07-12
Fix warnings in python scripts.

markusthoemmes:
See title.

Also enabled Flake8 to exit non-zero if there are warnings, seems like there is no harm to that right now. We can always disable that again if it really gets in the way.


tools/build/redo
```diff
@@ -278,12 +278,12 @@ Components = [
                   'build a runtime action container, matching name using the regex; NOTE: must use --dir for path to runtime directory',
                   yaml = False,
                   gradle = 'core:$1:distDocker'),
-
```
rabbah: how did these sneak in here with scancode anyway!?

PR #3875, opened 2018-07-12
Throttle the system based on active-ack timeouts.

markusthoemmes:
This is a fairly big change but bear with me!

Today, we have an arbitrary system-wide limit of maximum concurrent connections. In general that is fine, but it doesn't have a direct correlation to what's actually happening in the system.

This adds a new state to each monitored invoker: Overloaded. An invoker will go into overloaded state if active-acks are starting to timeout. Eventually, if the system is really overloaded, all Invokers will be in overloaded state which will cause the loadbalancer to return a failure. This failure now results in a `503 - System overloaded` message back to the user.

- [ ] API
- [X] Controller
- [X] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
+++ b/core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
@@ -353,8 +353,18 @@ class ShardingContainerPoolBalancer(config: WhiskConfig, controllerInstance: Con
                                 invoker: InvokerInstanceId): Unit = {
     val aid = response.fold(l => l, r => r.activationId)
 
-    // treat left as success (as it is the result of a message exceeding the bus limit)
-    val isSuccess = response.fold(_ => true, r => !r.response.isWhiskError)
+    val invocationResult = if (forced) {
+      InvocationFinishedResult.Timeout
+    } else {
+      // If the response contains a system error, report that, otherwise report Success
+      // Left generally is considered a Success, since that could be a message not fitting into Kafka
+      val isSystemError = response.fold(_ => false, _.response.isWhiskError)
+      if (isSystemError) {
+        InvocationFinishedResult.SystemError
+      } else {
+        InvocationFinishedResult.Success
+      }
+    }
 
     activations.remove(aid) match {
       case Some(entry) =>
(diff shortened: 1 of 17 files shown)


rabbah:
I haven’t looked at the changes yet- based on the description:

- is this heuristic affected by a network partition which results in no active acks 
- timing out active acks can happen in the controller side if it can’t drain active acks fast enough (too slow, too many, gc); is the heuristic affected?
l

markusthoemmes:
@rabbah, in theory, yes!

to 1.: If a network partition causes no active-acks, this network partition will also either lead to the controller not being able to write to kafka or the invoker not being able to read/write from kafka. These are valid cases to consider an invoker unusable?

to 2.: Again, I think this is a valid case of general overload. If there are too many active-acks for whatever reason, its safer to call the system done than continue to process with more crpytic errors (timeouts, crashes etc.)

WDYT?

markusthoemmes:
Summarizing a discussion with @rabbah: We agree that this is not optimal and might be subject to heuristics. It is however better what we have today and it serves as a catch-all around the invoker to catch any errors possible.

markusthoemmes:
PG4 2003 🔵

core/controller/src/main/scala/whisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -373,12 +393,12 @@ class ShardingContainerPoolBalancer(config: WhiskConfig, controllerInstance: Con
...
         // Active acks that are received here are strictly from user actions - health actions are not part of
         // the load balancer's activation map. Inform the invoker pool supervisor of the user action completion.
-        invokerPool ! InvocationFinishedMessage(invoker, isSuccess)
+        invokerPool ! InvocationFinishedMessage(invoker, invocationResult)
       case None if !forced =>
         // the entry has already been removed but we receive an active ack for this activation Id.
         // This happens for health actions, because they don't have an entry in Loadbalancerdata or
         // for activations that already timed out.
-        invokerPool ! InvocationFinishedMessage(invoker, isSuccess)
+        invokerPool ! InvocationFinishedMessage(invoker, invocationResult)
```
cbickel: Sending active-acks to the invokerPool, after they already did timeout will recover the invoker too early, as it might have still a queue, that is too large.
But this case also handles the active-acks of the health actions, which still need to be sent to the invokerpool.

markusthoemmes: Brilliant find! Indeed this kinda breaks the protocol because in an overloaded scenario, the invokers will swap between Healthy and Overloaded continuously.

As discussed in person, we now no longer send the result of an activation that finished after it was forced into the invokerPool 👍

PR #3878, opened 2018-07-13
Validate the Controller Swagger spec matches the impl

bbrowning:
## Description

This change wires up all existing API tests that use WskRestOperations to validate the HTTP requests and responses against the Controller API's Swagger spec.

It also adds a new Gradle task - :tests:testSwaggerCodegen - that does a basic validation of the Swagger spec and generates a basic Java client from that spec and verifies that the client compiles. CI is wired up to run this new test task.

The result of both of the above are a number of changes to the Swagger spec, also included here. This brings our Swagger spec much closer to matching the reality of the Controller API, and any additional coverage in the API test suite will further increase coverage of the Swagger spec.

There are some existing tests that purposely test API responses when passed invalid data, so the new SwaggerValidator trait includes a small whilelist of invalid requests and responses to accept so the build does not fail in these scenarios.

## Related issue and scope

http://mail-archives.apache.org/mod_mbox/openwhisk-dev/201806.mbox/%3CCAOpbpxHbSOdPpdR5t-WWTKkrTyuP%2B1c3_qeMY7vrhkfXmBYcQw%40mail.gmail.com%3E

- [x] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [x] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/tests/src/test/scala/common/rest/SwaggerValidator.scala
+++ b/tests/src/test/scala/common/rest/SwaggerValidator.scala
@@ -22,7 +22,6 @@ import scala.collection.JavaConverters._
 import akka.http.scaladsl.model.HttpEntity
 import akka.http.scaladsl.model.HttpRequest
 import akka.http.scaladsl.model.HttpResponse
-import akka.util.ByteString
 import com.atlassian.oai.validator.SwaggerRequestResponseValidator
 import com.atlassian.oai.validator.model.SimpleRequest
 import com.atlassian.oai.validator.model.SimpleResponse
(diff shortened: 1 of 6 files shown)


rabbah:
This is great!

tests/src/test/scala/common/rest/SwaggerValidator.scala
```diff
@@ -0,0 +1,109 @@
...
+                                 requestBody: Option[String],
+                                 response: HttpResponse,
+                                 responseBody: String): Seq[String] = {
+    var specRequestBuilder = new SimpleRequest.Builder(request.method.value, request.uri.path.toString())
+    for (header <- request.headers) {
+      specRequestBuilder = specRequestBuilder.withHeader(header.name, header.value)
+    }
+    for ((key, value) <- request.uri.query().toMap) {
+      specRequestBuilder = specRequestBuilder.withQueryParam(key, value)
+    }
```
markusthoemmes: Scala tip: You can get around the variable here by using `foldLeft`, like:

```scala
val specRequest = {
  val builder = new SimpleRequest.Builder(request.method.value, request.uri.path.toString)
  val withHeaders = request.headers.foldLeft(builder)((b, header) => b.withHeader(header.name, header.value))
  val andQuery =
    request.uri.query().foldLeft(withHeaders) { case (b, (key, value)) => b.withQueryParam(key, value) }
  val andBody = requestBody
    .map(builder.withBody(_).withHeader("Content-Type", request.entity.contentType.value))
    .getOrElse(andQuery)

  andBody.build()
}
```

`foldLeft` runs over a list and applies each element to a seed value that you pass to it.

This also gets around using `Option.get`, which can throw an exception. Using combinators on Option (like map and getOrElse) you get a fully compiletime checked version of all the types without needing to wrap them in if/else statements.

tests/src/test/scala/common/rest/SwaggerValidator.scala
```diff
@@ -0,0 +1,109 @@
...
+        .withBody(requestBody.get)
+        .withHeader("content-type", request.entity.contentType.value)
+    }
+    val responseCopy = response.copy(entity = HttpEntity.Strict(response.entity.contentType, ByteString(responseBody)))
+    var specResponseBuilder = SimpleResponse.Builder
+      .status(response.status.intValue())
+      .withBody(responseBody)
+    for (header <- response.headers) {
+      specResponseBuilder = specResponseBuilder.withHeader(header.name, header.value)
+    }
```
markusthoemmes: Same as above for avoiding the `var`

tests/src/test/scala/common/rest/WskRestOperations.scala
```diff
@@ -1188,7 +1189,20 @@ trait RunRestCmd extends Matchers with ScalaFutures {
       hostWithScheme.withPath(path).withQuery(Query(params)),
       List(Authorization(creds)),
       entity = body.map(b => HttpEntity(ContentTypes.`application/json`, b)).getOrElse(HttpEntity.Empty))
-    Http().singleRequest(request, connectionContext).futureValue
+    val response = Http().singleRequest(request, connectionContext).futureValue
```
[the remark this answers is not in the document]
bbrowning: I don't love creating the copy either. I'll try eagerly creating Strict entities as you suggested to see if that avoids the need.

bbrowning:
I've updated the implementation based on feedback and rebased on top of latest master as a conflict had been introduced since I originally opened this.

markusthoemmes:
I ran a PG and it caught issues in `NamespaceSpecificThrottleTests`. I then realized, that these aren't running in travis, hence #3933.

The error message there is:

```
org.scalatest.exceptions.TestFailedException: HTTP request or response did not match the Swagger spec.
Request: HttpRequest(HttpMethod(POST),https://openwhisk-pg4-vm1.swg.usma.ibm.com/api/v1/namespaces/zeroSubject/actions/zeroAction?blocking=false&result=false,List(Authorization: Basic ...),HttpEntity.Strict(application/json,{}),HttpProtocol(HTTP/1.1))
Response: HttpResponse(429 Too Many Requests,List(Server: nginx, Date: Wed, 01 Aug 2018 00:47:16 GMT, Connection: keep-alive, X-Request-ID: 35d8ba338053dc13c54da44052939a15, Access-Control-Allow-Origin: *, Access-Control-Allow-Headers: Authorization, Content-Type, OpenWhisk_Version_No: pg4-2023, X-Request-ID: 35d8ba338053dc13c54da44052939a15),HttpEntity.Strict(application/json,{"error":"Too many requests in the last minute (count: 1, allowed: 0).","code":"35d8ba338053dc13c54da44052939a15"}),HttpProtocol(HTTP/1.1))
Validation Error: ArrayBuffer(ERROR - Response status 429 not defined for path '/namespaces/{namespace}/actions/{actionName}'.: [])
```

bbrowning:
I pushed another commit to address the HTTP 429 from `NamespaceSpecificThrottleTests` and was able to verify locally that they pass with this change.

It's good this failure happened, since it shows the kind of error message output you'll see if a change in the Swagger spec or tests causes the test request/response to not match the spec.

markusthoemmes:
PG approved this now, having trouble to push travis through though :/

bbrowning:
Thanks!

PR #3880, opened 2018-07-16
Modify that web action in the bound package can be accessed.

upgle:
## Description

Recently, many users want to share their web actions with the package.
but, the web action in the bound package can't be accessed in the browser. 

Because when it is invoked in the browser, it doesn't resolve an action to find original fully qualified action name.  so, I added the action name resolver.

#### HOW TO TEST.
bind a package that contains web actions and access web action which is in the bound package using your browser, you may see errors like the image below.

image

## Fixed
Now, the bound web action can be accessed in the browser.

image

## Related issue and scope

- [x] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [x] Documentation

## Types of changes

## Checklist:


rabbah:
Thanks for trying this. There are some issues to be aware of packaged actions as web actions via bindings. The lack of support isn’t a bug (it was intentional at the time). This has come up before so it’s worthwhile to try and enable this. 

1. If an action isn’t already a web action in the package, it can’t be accessed as a webaction through the biding. That is a subject that binds the action can’t make it a web action (unless they already own it). So the creator/shares of the package will have to decide which of the actions to make available as web actions. 

2. The entitlement check for a web action today is straightforward: namespace must own both and so implicitly entitled to them. With a binding you need to check if the binding continues to have rights to the bound package.

3. And of course you’ll need to add tests for the change/new feature.

upgle:
@rabbah 
Thank you for your feedback and sorry that I thought it was a bug. 
I will add some test codes including the rights check and suggest it again.

upgle:
@rabbah 

I modified it to check the entitlement and added tests.
Now, It returns the authentication error if a web action has no entitlement.
```
{
  "error": "The supplied authentication is not authorized to access 'whisk.system/share'.",
  "code": "NmSIY4dJanMT7YS3MPwFoOPL4Np72g5B"
}
```

Could you please remove WIP label and review again?

upgle:
I fixed a conflict. please review it.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -700,9 +713,11 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
    */
   private def actionLookup(actionName: FullyQualifiedEntityName)(
     implicit transid: TransactionId): Future[WhiskActionMetaData] = {
-    getAction(actionName) recoverWith {
-      case _: ArtifactStoreException | DeserializationException(_, _, _) =>
-        Future.failed(RejectRequest(NotFound))
+    resolveAction(actionName) flatMap { resolveAction =>
+      getAction(resolveAction) recoverWith {
```
rabbah: i dont believe this is right - the resolveAction method returns just the name of the action in the base package, and so getting that action means you'll miss the bound parameters. I think you want to use resolveActionAndMergeParameters instead.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -743,6 +758,16 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
...
   }
 
+  /**
+   * Checks if an action is executable.
+   */
+  private def checkEntitlement(identity: Identity, action: WhiskActionMetaData)(
+    implicit transid: TransactionId): Future[Unit] = {
+    val resource =
+      Resource(action.namespace.root.toPath, Collection(Collection.PACKAGES), Some(action.namespace.last.toString))
+    entitlementProvider.check(identity, Privilege.READ, resource)
```
rabbah: this check also checks the throttles - which means there's now a redundant check for throttles.
the code (before this pr), did not check entitlement through the provider because it can implicitly infer the owner (there's no indirection through a binding) so it only checked throttles.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
upgle: Oh... I understand I'll modify it soon.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
upgle: Thank you for your kind review. I've removed existing entitlementProvider.checkThrottles(), then modified to use only entitlementProvider.check ().

upgle:
@rabbah Maybe it's done. could you please review it again? 😅

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -552,8 +552,11 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
         if (a.namespace.defaultPackage) {
           Future.successful(a)
         } else {
-          pkgLookup(a.namespace.toFullyQualifiedEntityName) map { pkg =>
-            (a.inherit(pkg.parameters))
+          // if action is not in the default package, then check entitlement
+          checkEntitlement(actionOwnerIdentity, a) flatMap { _ =>
+            pkgLookup(a.namespace.toFullyQualifiedEntityName) map { pkg =>
+              (a.inherit(pkg.parameters))
+            }
```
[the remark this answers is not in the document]
upgle: @rabbah 
yes, I agree that it looks complicated. so I've changed some codes. (removed `pkgLookup` method which is not necessary anymore)

1. action in a default package no need to resolve, just check entitlement and if it isexported.
2. action in a bound package(new) is handled by `resolveActionAndMergeParameters()`, it would resolve action and merge parameters. 
2. check whether action is in a proper package -> It check whether an action is exported by `confirmExportedAction()` method
3. and all types of action is checked for entitlement and throttle by `checkEntitlement()` method

please check changed code here

```scala
private def verifyWebAction(actionName: FullyQualifiedEntityName, authenticated: Boolean)(
    implicit transid: TransactionId) = {

    // lookup the identity for the action namespace
    identityLookup(actionName.path.root) flatMap { actionOwnerIdentity =>
      confirmExportedAction(actionLookup(actionName), authenticated) flatMap { a =>
        checkEntitlement(actionOwnerIdentity, a) map { _ => (actionOwnerIdentity, a)}
      }
    }
  }
```

https://github.com/apache/incubator-openwhisk/pull/3880/files#diff-a356e6b010b8aeb5ea7b6f7fec395038R535

junoyoon:
@upgle is it done?

rabbah:
The last commit looks a lot better now - I'll review more thoroughly.

upgle:
@junoyoon no, i'm waiting for review.

@rabbah ok, tell me if I missed something.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -440,21 +441,14 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
...
-  }
-
   /**
    * Gets action from datastore.
-   * This method is factored out to allow mock testing.
+   * if it is contained in the package, then resolve and merge parameters
+   * This method is factored out to allow mock testing
    */
-  protected def getAction(actionName: FullyQualifiedEntityName)(
+  protected def resolveActionAndMergeParameters(actionName: FullyQualifiedEntityName)(
```
rabbah: We should add a comment that while parameters are merged, annotations are not. This is important because the web annotation is on an action not an entire package. We should separately add a test for this invariant.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -537,27 +531,15 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
...
-      // package.params -> action.params -> query.params -> request.entity (body) -> augment arguments (namespace, path)
-      action <- confirmExportedAction(actionLookup(actionName), authenticated) flatMap { a =>
-        if (a.namespace.defaultPackage) {
-          Future.successful(a)
-        } else {
-          pkgLookup(a.namespace.toFullyQualifiedEntityName) map { pkg =>
-            (a.inherit(pkg.parameters))
-          }
+    // lookup the identity for the action namespace
+    identityLookup(actionName.path.root) flatMap { actionOwnerIdentity =>
```
rabbah: here the identity is tied to how the action is invoked: if the path is relative to the package, it's the owner of the package. But if it's via the binding, the identity is the owner binding. The two identities are not the same. It might also not make sense to invoke an action in a public package since it might require a binding to work correctly.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -684,33 +666,14 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
...
-  private def pkgLookup(pkg: FullyQualifiedEntityName)(implicit transid: TransactionId): Future[WhiskPackage] = {
-    getPackage(pkg).filter {
-      _.binding.isEmpty
-    } recoverWith {
-      case _: ArtifactStoreException | DeserializationException(_, _, _) =>
-        // if the package lookup fails or the package doesn't conform to expected invariants,
-        // fail the request with BadRequest so as not to leak information about the existence
-        // of packages that are otherwise private
-        logging.debug(this, s"package which does not exist")
-        Future.failed(RejectRequest(NotFound))
```
rabbah: we should add some tests for `resolveActionAndMergeParameters` to preserve the behavior represented by these request rejections (missing package or malformed package).

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
rabbah: another aspect to document: a web action in a public package cannot be disabled so any (private) binding of the package will have the web action exposed.

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
rabbah: need to update the docs here: https://github.com/apache/incubator-openwhisk/blob/master/docs/webactions.md#additional-features the precedence order would now also include package binding parameters.

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -233,7 +238,7 @@ trait WebActionsApiBaseTests extends ControllerTestCommon with BeforeAndAfterEac
       if (actionName.path.defaultPackage) {
         Future.successful(theAction)
       } else {
-        getPackage(actionName.path.toFullyQualifiedEntityName) map (_ => theAction)
+        getPackage(actionName.path.toFullyQualifiedEntityName) map (pkg => theAction.inherit(pkg.parameters))
```
rabbah: there's no test for actually using a package binding here (if there is, it would fail since the parameter inheritance does not include the binding).

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
upgle: ok I'll update the docs soon.

rabbah: thank you - it might be worth adding a section in the docs about web actions for packages and bindings (taking in some of my comments below). what do you think?

ningyougang:
@upgle ,can you rebase?

upgle:
@ningyougang thank you, i'll rebase it.

rabbah:
I rebased this and resolved the conflict.

docs/webactions.md
```diff
@@ -485,6 +485,12 @@ $ curl https://${APIHOST}/api/v1/web/guest/default/custom-options.http -kvX OPTI
 < Access-Control-Allow-Origin: example.com
 ```
 
+## Sharing web actions
+
```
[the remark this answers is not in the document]
upgle: @rabbah i've checked what you changed. thank you very much!

core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -697,22 +672,15 @@ trait WhiskWebActionsApi
...
-      case _: NoSuchElementException =>
-        logging.debug(this, s"'$pkg' is a binding")
-        Future.failed(RejectRequest(NotFound))
```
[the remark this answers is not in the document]
upgle: @rabbah 
test cases for 2, 3 have been added. https://github.com/apache/incubator-openwhisk/pull/3880/commits/e970d3a54a207fb60c41b7c56b2d0e802434428c

rabbah:
Thanks - will check this out today or tomorrow.

upgle:
@rabbah ok, thank you

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -126,36 +122,47 @@ trait WebActionsApiBaseTests extends ControllerTestCommon with BeforeAndAfterEac
...
+  var testParametersInInvokeAction = true // toggle to test parameter in invokeAction
   var requireAuthenticationKey = "example-web-action-api-key"
-  var customOptions = true // toogle web-custom-options annotation on action
   var invocationCount = 0
   var invocationsAllowed = 0
 
+  override def beforeAll() = {
+    implicit val tid = transid()
+
+    put(entityStore, stubPackage, garbageCollect = false)
```
rabbah: this is why the stubs arent collected: garbageCollect is set to false.

rabbah:
I fixed the test fixture cleanup and also tweaked one test.

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
upgle: yes, it seems other test case remove all document after tests.

https://github.com/apache/incubator-openwhisk/blob/master/tests/src/test/scala/whisk/core/controller/test/ControllerTestCommon.scala#L203

upgle:
@rabbah test is passed. thank you for your support 👍

rabbah:
Thank you @upgle for your contribution.

PR #3882, opened 2018-07-16
Use invoker id for metrics instead of the hostname.

cbickel:
Today, the hostname of the invoker is used for the metrics, sent by kamon. This PR changes this behaviour to use the instance of the invoker instead.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/common/Logging.scala
+++ b/common/scala/src/main/scala/whisk/common/Logging.scala
@@ -210,10 +210,7 @@ object LogMarkerToken {
 
 }
 
-object MetricEmitter {
-
-  val metrics = Kamon.metrics
-
+object MetricSupport {
   def initKamon(instanceName: String): Unit = {
     // Replace the hostname of the invoker to the assigned id of the invoker.
     val newKamonConfig = Kamon.config
@@ -222,6 +219,11 @@ object MetricEmitter {
         ConfigValueFactory.fromAnyRef(instanceName))
     Kamon.start(newKamonConfig)
   }
+}
+
+object MetricEmitter {
+
+  val metrics = Kamon.metrics
 
   def emitCounterMetric(token: LogMarkerToken): Unit = {
     if (TransactionId.metricsKamon) {
--- a/common/scala/src/main/scala/whisk/common/Logging.scala
+++ b/common/scala/src/main/scala/whisk/common/Logging.scala
@@ -18,12 +18,10 @@
 package whisk.common
 
 import java.io.PrintStream
-import java.time.format.DateTimeFormatter
 import java.time.{Clock, Instant, ZoneId}
-
+import java.time.format.DateTimeFormatter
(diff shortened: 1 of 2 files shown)


core/invoker/src/main/scala/whisk/core/invoker/Invoker.scala
```diff
@@ -125,6 +124,14 @@ object Invoker {
 
         new InstanceIdAssigner(config.zookeeperHosts).getId(invokerUniqueName.get)
       }
+
+    // Replace the hostname of the invoker to the assigned id of the invoker.
+    val newKamonConfig = Kamon.config
```
chetanmeh: Nit can this be refactored to a method `initKamon(hostId:String)`. Further should this be moved to a util method in `whisk.common` such it is also used in Controller which also has this notion of id to identify it.

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -212,6 +214,15 @@ object MetricEmitter {
 
   val metrics = Kamon.metrics
```
chetanmeh: Not sure if it would be safe to add to `MetricEmitter` as first reference to `MetricEmitter` object itself would trigger loading of `metrics`  and we would probably need to have config provided and Kamon started before that happens. 

So would be safer to have a new object hosting this method

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -208,6 +210,17 @@ object LogMarkerToken {
 
 }
 
+object MetricSupport {
```
vvraskin: @chetanmeh wondering if we indeed need this object since we only use it for invoker. In case of controller we don't really override it in this PR.

chetanmeh: I thought similar support would be required for controller also so as to distinguish the metrics easily and having this change would simplify that subsequently. Its fine to drop this and have a method in `Invoker` itself for now.

vvraskin: The problem with controllers would appear if we specify from the outside whether its only for activations or for crud operations. This is typically reflected in the container hostname which makes the metrics from all controller instances distinct. Given the current implementation, it would require some additional refactoring in the controller code to follow the same approach as in the invoker...

chetanmeh: Makes sense. Lets keep it simple then and have this method in invoker. If we rework Controller identity then this aspect can be revisited

cbickel:
PG5#532 is 🔵

PR #3884, opened 2018-07-17
Document metrics generated within OpenWhisk

chetanmeh:
OpenWhisk generates quite a few metrics via [Kamon](http://kamon.io). This PR aims to document all currently generated metrics such that system administrators can configure the right metrics for monitoring

See [here](https://github.com/chetanmeh/incubator-openwhisk/blob/metrics-docs/docs/metrics.md#metric-details) for the rendered HTML version of doc

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [x] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -75,6 +75,183 @@ The docker image exposes StatsD via the (standard) port 8125 and a Grafana dashb
 
 The address of your docker host has to be configured in the `metrics_kamon_statsd_host` configuration property.
 
+### Metric Names
+
+All metric names have to prefixed by a prefix that you specify and are subject to modification by graphite, datadog, or statsd. For example if prefix used is `openwhisk` then metric names would be like `openwhisk.counter.controller_activation_start`. This document assumes that metric name prefix is `openwhisk`
+
+Currently OpenWhisk emits following types of metrics
+
+#### Counter
+
+Counter [record the count](http://kamon.io/documentation/0.6.x/kamon-core/metrics/instruments/#counters) of metric and there names are prefixed with `openwhisk.counter`. For example `openwhisk.counter.controller_activation_start`. Counters just counts and resets to zero upon each flush.
+
+#### Histograms
+
+Histogram record the [distribution](http://kamon.io/documentation/0.6.x/kamon-core/metrics/instruments/#histograms) of given metric and there names are prefixed with `openwhisk.histogram`. For example `openwhisk.histogram.controller_activation_finish`. A histogram metrics may result in multiple values at the metric aggregator level. For example in [Datadog](https://docs.datadoghq.com/developers/metrics/histograms/) for each histogram metric following values are record
+
+* `my_metric.avg` - Average of aggregated values during the flush interval.
+* `my_metric.count` - Count of aggregated values during the flush interval.
+* `my_metric.median` - Median of aggregated values during the flush interval.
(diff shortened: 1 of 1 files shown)


rabbah:
Fantastic!

rabbah:
@vvraskin @mhenke1 would you review this? 

@chetanmeh this is very useful thanks for doing this.

docs/metrics.md
```diff
@@ -75,6 +75,183 @@ The docker image exposes StatsD via the (standard) port 8125 and a Grafana dashb
...
+  * `openwhisk.counter.invoker_docker.unpause_start`
+  * `openwhisk.counter.invoker_docker.unpause_error`
+  * `openwhisk.histogram.invoker_docker.unpause_finish`
+  * `openwhisk.histogram.invoker_docker.unpause_error`
+
+#### Kafka Metrics
+
+Metrics below are emitted per kafka topic.
+
+* `openwhisk.histogram.kafka_<topic name>.delay_start` - Time delay between when a message was pushed to kafka and when it is read within a consumer.
```
vvraskin: should we also mention that Delay is being emitted for each pool by Invoker, while Queue metric is emitted every 10 seconds?

docs/metrics.md
```diff
@@ -75,6 +75,183 @@ The docker image exposes StatsD via the (standard) port 8125 and a Grafana dashb
...
+
+Below are some of the important metrics emitted by OpenWhisk setup
+
+#### Controller metrics
+
+Metrics below are emitted from with a Controller instance.
+
+##### Controller Startup
+
+* `openwhisk.counter.controller_startup<controller_id>_count` (counter)
```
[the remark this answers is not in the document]
chetanmeh: Checked the stats in our setup and they are named like `openwhisk.counter.controller_startup0_count`. Also in code the prefix is `startup$id`

https://github.com/apache/incubator-openwhisk/blob/8b5abe7bbeb4464af586c1993fc3590e0fe8d516/common/scala/src/main/scala/whisk/common/Logging.scala#L257

docs/metrics.md
chetanmeh: Missed that it was be done via a scheduled task. Per default config it should be emitted every 60 seconds

vvraskin:
@chetanmeh 
Which backend do you use for the system metrics?
If you use Grafana+InfluxDB, it allows us to aggregate metrics of several components using regex filters. I'm wondering whether the same approach would be feasible...

chetanmeh:
@vvraskin  We use Datadog and it [does not seem][2] to support [regex in metric names][1]. Which makes it tricky to aggregate metrics across various invokers. 

Does Grafana+Influx support tags? Looking for common denominator here

[1]: https://docs.datadoghq.com/graphing/faq/when-i-query-can-i-use-wildcards-in-metric-names-and-events/
[2]: https://trello.com/c/WtJXrGQj/18-datadog-wildcard-regex-matches-for-metric-names-1

vvraskin:
Our statsd integration with influx won't support tagging, but I think that we already have tagging option for Kamon introduced in this PR https://github.com/apache/incubator-openwhisk/commit/6f1a445f22f827bd91f75b01bc82988054febbc8

chetanmeh:
Opened #3917 to discuss possible options for supporting setups which do not support regex

PR #3892, opened 2018-07-23
Use the authz provider to retrieve the identity of the namespace

mhenke1:
Unauthenticated web actions will use the authentication provider configured in the SPI to create the namespace based Identity

## Description
Unauthenticated web actions do create a namespace based identity that is used to entitle the call.
This change makes them use the configured authentication provider to construct this identity.

## Related issue and scope

- [x] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
@@ -320,9 +320,28 @@ class RestAPIVersion(config: WhiskConfig, apiPath: String, apiVersion: String)(
 }
 
 trait AuthenticationDirectiveProvider extends Spi {
+
+  /**
+   * Returns an authentication directive used to validate the
+   * passed user credentials.
+   * At runtime the directive returns an user identity
+   * which is passed to the following routes
+   *
+   * @return autehtication directive used to verify the user credentials
+   */
   def authenticate(implicit transid: TransactionId,
                    authStore: AuthStore,
                    logging: Logging): AuthenticationDirective[Identity]
 
-  def namespaceIdentity(namespace: EntityName)(implicit transid: TransactionId, authStore: AuthStore): Future[Identity]
+  /**
+   * Retrieves an Identity based on a given namespace name.
+   *
+   * For use-cases of anonymous invocation (i.e. WebActions),
+   * we need to an identity based on a given namespace-name to
+   * give the invocation all the context needed.
+   * @param namespace the namespace that the identity will be based on
+   * @return identity based on the given namespace
+   */
+  def identityByNamespace(namespace: EntityName)(implicit transid: TransactionId,
+                                                 authStore: AuthStore): Future[Identity]
 }
--- a/core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
@@ -40,7 +40,7 @@ import whisk.core.{ConfigKeys, WhiskConfig}
 import whisk.http.Messages
 import whisk.spi.{Spi, SpiLoader}
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
 /**
@@ -323,4 +323,6 @@ trait AuthenticationDirectiveProvider extends Spi {
   def authenticate(implicit transid: TransactionId,
                    authStore: AuthStore,
                    logging: Logging): AuthenticationDirective[Identity]
+
+  def namespaceIdentity(namespace: EntityName)(implicit transid: TransactionId, authStore: AuthStore): Future[Identity]
 }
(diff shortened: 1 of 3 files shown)


core/controller/src/main/scala/whisk/core/controller/WebActions.scala
```diff
@@ -457,7 +458,7 @@ trait WhiskWebActionsApi extends Directives with ValidateRequestSize with PostAc
    * This method is factored out to allow mock testing.
    */
   protected def getIdentity(namespace: EntityName)(implicit transid: TransactionId): Future[Identity] = {
-    Identity.get(authStore, namespace)
+    SpiLoader.get[AuthenticationDirectiveProvider].namespaceIdentity(namespace)(transid, authStore)
```
markusthoemmes: You'll want to cache the call to `SpiLoader.get[AuthenticationDirectiveProvider]`. That's not something which should be executed on a per request basis.

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -323,4 +323,6 @@ trait AuthenticationDirectiveProvider extends Spi {
   def authenticate(implicit transid: TransactionId,
                    authStore: AuthStore,
                    logging: Logging): AuthenticationDirective[Identity]
+
+  def namespaceIdentity(namespace: EntityName)(implicit transid: TransactionId, authStore: AuthStore): Future[Identity]
```
markusthoemmes: Should we rename this to something more descriptive, like `identityByNamespace`? A little bit of ScalaDoc would help for context, i.e.:

```scala
/**
 * Retrieves an Identity based on a given namespace name.
 *
 * For use-cases of anonymous invocation (i.e. WebActions), 
 * we need to an identity based on a given namespace-name to 
 * give the invocation all the context needed.
 */
```

A little bit of ScalaDoc for `authenticate` would be great in the same vain.

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -320,7 +320,28 @@ class RestAPIVersion(config: WhiskConfig, apiPath: String, apiVersion: String)(
 }
 
 trait AuthenticationDirectiveProvider extends Spi {
+
+  /**
+   * Returns an authentication directive used to validate the
+   * passed user credentials.
+   * At runtime the directive returns an user identity
+   * which is passed to the following routes
```
rabbah: please terminate comment with a `.`

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -320,7 +320,28 @@ class RestAPIVersion(config: WhiskConfig, apiPath: String, apiVersion: String)(
...
+   *
+   * @return autehtication directive used to verify the user credentials
```
rabbah: typo.

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -320,7 +320,28 @@ class RestAPIVersion(config: WhiskConfig, apiPath: String, apiVersion: String)(
...
                    authStore: AuthStore,
                    logging: Logging): AuthenticationDirective[Identity]
+
+  /**
+   * Retrieves an Identity based on a given namespace name.
+   *
+   * For use-cases of anonymous invocation (i.e. WebActions),
+   * we need to an identity based on a given namespace-name to
+   * give the invocation all the context needed.
+   * @param namespace the namespace that the identity will be based on
```
rabbah: Add new line before `@param` 🙏

rabbah:
> I thought in the same direction but then wondered if multiple identities for a namespace is in fact an edge-case of today' authentication rather than a common property of authentication, if that makes sense?

any organization that has a shared space is subject to this --- is that really an edge case?

markusthoemmes:
> any organization that has a shared space is subject to this --- is that really an edge case?

A different authentication model might generate a token to access the namespace on the fly rather than providing the first key found in a list of keys allowed to a namespace.

Does that make sense?

rabbah:
> A different authentication model might generate a token to access the namespace on the fly rather than providing the first key found in a list of keys allowed to a namespace.

Hence the need to add a test - to enforce that if there is ambiguity there is an agreed resolution and the contract to the controller is the same.

rabbah:
Another way of restating: the standard authentication (couchdb) returns one of the identities where there are multiples, i.e., it is not an error. For all SPIs should the choice of error or return _an_ identity be consistent or left up to the implementation?

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
mhenke1: Thx. Done

core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
mhenke1: Thx. Done

markusthoemmes:
Per Slack discussion with @rabbah:

`getIdentity` is a wrapper around the `identityByNamespace`. It just delegates to that method directly, so I claim no test needed for that. We have a test for a failing future and the signature forces you to return just one object. It’s up to the implementation to decide if duplicates are an error or not.

Testcoverage wise, we're good.

PR #3908, opened 2018-07-28
Prevent Kafka poll from hanging indefinitely 

dubee:
## Description

Kafka's poll method can hang indefinitely. Instead of infinitely blocking the thread that is polling Kafka, only allow poll to run for a specified time period.

For more information: https://cwiki.apache.org/confluence/display/KAFKA/KIP-266%3A+Fix+consumer+indefinite+blocking+behavior.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
+++ b/common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
@@ -30,7 +30,7 @@ import whisk.core.connector.MessageConsumer
 
 import scala.collection.JavaConverters._
 import scala.concurrent.duration._
-import scala.concurrent.{blocking, ExecutionContext, Future}
+import scala.concurrent.{blocking, Await, ExecutionContext, Future}
 
 case class KafkaConsumerConfig(sessionTimeoutMs: Long, metricFlushIntervalS: Int)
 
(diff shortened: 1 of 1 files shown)


common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -69,7 +69,7 @@ class KafkaConsumerConnector(
     val wakeUpTask = actorSystem.scheduler.scheduleOnce(cfg.sessionTimeoutMs.milliseconds + 1.second)(consumer.wakeup())
 
     try {
-      val response = consumer.poll(duration.toMillis).asScala
+      val response = Await.result(Future(consumer.poll(duration.toMillis).asScala), 30.seconds)
```
markusthoemmes: How does this relate to the wakeup call 4 lines above? The comment states that it's supposed to wake up the consumer in case poll is infinitely blocked. Does that not work as expected?

dubee: @markusthoemmes, I don't trust `wakeup()` to always work when the invoker machine is under heavy memory pressure. Perhaps this isn't the solution, but an extra safeguard here would be nice.

markusthoemmes: It's very important to note here that a failing Future does **not** cancel the scheduled action. The poll will be hanging around in the background.

Is there a reason why wakeup wouldn't work under memory pressure, where the new timeout would work better?

dubee:
Closing this in favor of https://github.com/apache/incubator-openwhisk/pull/3913.

PR #3912, opened 2018-07-30
Fix max.poll.interval.ms setting of KafkaConsumer.

markusthoemmes:
The `max.poll.interval.ms` setting needs to be high enough to not be thrown off by all actions in the system consuming the maximum duration of their respective action timeout, ideally also adding enough slack to account for image pulls taking long, or docker runs being very slow.

The name of the setting was also misspelled so we've been running on the default.

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -72,7 +72,7 @@ whisk {
             heartbeat-interval-ms = 10000
             enable-auto-commit = false
             auto-offset-reset = earliest
-            max-poll-interval = 360000
+            max-poll-interval-ms = 900000 // 20 minutes
             // This value controls the server-side wait time which affects polling latency.
             // A low value improves latency performance but it is important to not set it too low
             // as that will cause excessive busy-waiting.
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -72,7 +72,12 @@ whisk {
             heartbeat-interval-ms = 10000
             enable-auto-commit = false
             auto-offset-reset = earliest
-            max-poll-interval-ms = 900000 // 20 minutes
+
+            // request-timeout-ms always needs to be larger than max-poll-interval-ms per
+            // https://kafka.apache.org/documentation/#upgrade_1010_notable
+            max-poll-interval-ms = 1800000 // 30 minutes
+            request-timeout-ms = 1860000 // 31 minutes
+
             // This value controls the server-side wait time which affects polling latency.
             // A low value improves latency performance but it is important to not set it too low
             // as that will cause excessive busy-waiting.
--- a/common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
+++ b/common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
@@ -112,10 +112,10 @@ object KafkaConfiguration {
   def verifyConfig(config: Map[String, String], validKeys: Set[String])(implicit logging: Logging): Boolean = {
     val passedKeys = config.keySet
     val knownKeys = validKeys intersect passedKeys
-    val unknownSettings = config -- knownKeys
+    val unknownKeys = passedKeys -- knownKeys
 
-    if (unknownSettings.nonEmpty) {
-      logging.warn(this, s"potential misconfiguration, unknown settings: ${unknownSettings.mkString(",")}")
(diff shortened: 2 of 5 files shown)


common/scala/src/main/resources/application.conf
```diff
@@ -72,7 +72,7 @@ whisk {
             heartbeat-interval-ms = 10000
             enable-auto-commit = false
             auto-offset-reset = earliest
-            max-poll-interval = 360000
+            max-poll-interval-ms = 900000 // 20 minutes
```
chetanmeh: May be change it to use duration as a unit. `max-poll-interval = 20 minutes`

markusthoemmes: These values are passed directly down to the kafka-setting which don't support durations per my knowledge.

The configuration here is build to be passed directly to the respective consumer/producer without touching it in our code at all.

chetanmeh: Did not knew that. So configs as mentioned [here][1] are passed as is. May be we can later add a testcase which uses [AdminClient.describeTopic][2] to assert the expected config

[1]: https://kafka.apache.org/documentation/#newconsumerconfigs
[2]: http://kafka.apache.org/0110/javadoc/index.html?org/apache/kafka/clients/admin/AdminClient.html

sven-lange-last:
The Kafka new Java consumer documentation contains the following hint in the [notables changes section for 0.10.1.0](https://kafka.apache.org/documentation/#upgrade_1010_notable):

```
The new Java Consumer now supports heartbeating from a background thread. There is a new configuration max.poll.interval.ms which controls the maximum time between poll invocations before the consumer will proactively leave the group (5 minutes by default). The value of the configuration request.timeout.ms must always be larger than max.poll.interval.ms because this is the maximum time that a JoinGroup request can block on the server while the consumer is rebalancing, so we have changed its default value to just above 5 minutes.
```

At the moment, we have `request.timeout.ms` < `max.poll.interval.ms` and this PR will also keep the value smaller.

I guess that this should not harm in our situation with a single consumer group per topic, a single partition per topic and a single consumer process in the consumer group. But I don't understand all implications...

common/scala/src/main/resources/application.conf
sven-lange-last: Very good finding! And: 900,000 ms are 15 minutes :-)

markusthoemmes:
Fair point @sven-lange-last, added the relevant timeout + doc link.

@chetanmeh I added a check at startup of either Consumer/Producer that verifies the config and prints an error if all set keys are known.

sven-lange-last:
From [Kafka documentation - New Consumer Configs](https://kafka.apache.org/documentation/#newconsumerconfigs):

* `request.timeout.ms`: The configuration controls the maximum amount of time the client will wait for the response of a request. If the response is not received before the timeout elapses the client will resend the request if necessary or fail the request if retries are exhausted.
* `max.poll.interval.ms`: The maximum delay between invocations of poll() when using consumer group management. This places an upper bound on the amount of time that the consumer can be idle before fetching more records. If poll() is not called before expiration of this timeout, then the consumer is considered failed and the group will rebalance in order to reassign the partitions to another member.

common/scala/src/main/scala/whisk/connector/kafka/KafkaMessagingProvider.scala
```diff
@@ -101,4 +101,24 @@ object KafkaConfiguration {
...
+   * @param validKeys known valid keys to configure
+   * @return true if all configuration keys are known, false if at least one is unknown
+   */
+  def verifyConfig(config: Map[String, String], validKeys: Set[String])(implicit logging: Logging): Boolean = {
+    val passedKeys = config.keySet
+    val knownKeys = validKeys intersect passedKeys
+    val unknownSettings = config -- knownKeys
+
+    if (unknownSettings.nonEmpty) {
+      logging.warn(this, s"potential misconfiguration, unknown settings: ${unknownSettings.mkString(",")}")
```
chetanmeh: May be just log key names as value may potentially contain sensitive info

sven-lange-last: Sounds reasonable.

PR #3919, opened 2018-07-30
Allow raw JSON string as response for webaction without base64 encoding.

rabbah:
For legacy reasons, we required web actions to return a base 64 encoded string when returning JSON. This PR removes the requirement to base64 encode JSON if returning the result as string. This is based on a slack discussion with @csantanapr (as reported from an end-user).

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/routemgmt/common/apigw-utils.js
+++ b/core/routemgmt/common/apigw-utils.js
@@ -529,8 +529,6 @@ function addEndpointToSwaggerApi(swaggerApi, endpoint, responsetype) {
   responsetype = responsetype || 'json';
   console.log('addEndpointToSwaggerApi: operationid = '+operationId);
   try {
-    var auth_base64 = Buffer.from(endpoint.action.authkey,'ascii').toString('base64');
-
     // If the relative path already exists, append to it; otherwise create it
     if (!swaggerApi.paths[endpoint.gatewayPath]) {
       swaggerApi.paths[endpoint.gatewayPath] = {};
@@ -890,7 +888,7 @@ function replaceNamespaceInUrl(url, namespace) {
  *     {
  *        statusCode: 502,    <- signifies an application error
  *        headers: {'Content-Type': 'application/json'},
- *        body: Base64 encoded JSON error string
+ *        body: JSON object or JSON string
  *     }
  * Otherwise, the action was invoked as a regular OpenWhisk action
  * (i.e. https://OW-HOST/api/v1/namesapces/NS/actions/ACTION) and the
@@ -909,18 +907,16 @@ function makeErrorResponseObject(err, isWebAction) {
     return err;
   }
 
-  var bodystr;
+  var bodystr = err;
   if (typeof err === 'string') {
-    bodystr = JSON.stringify({
+    bodystr = {
       "error": JSON.parse(makeJsonString(err)),  // Make sure err is plain old string to avoid duplicate JSON escaping
-    });
-  } else {
-    bodystr = JSON.stringify(err);
+    };
   }
   return {
     statusCode: 502,
     headers: { 'Content-Type': 'application/json' },
-    body: new Buffer(bodystr).toString('base64'),
+    body: bodystr
   };
 }
 
@@ -935,7 +931,7 @@ function makeErrorResponseObject(err, isWebAction) {
  *     {
  *        statusCode: 200,    <- signifies a successful action
  *        headers: {'Content-Type': 'application/json'},
- *        body: Base64 encoded JSON error string
+ *        body: JSON object or JSON string
  *     }
  * Otherwise, the action was invoked as a regular OpenWhisk action
  * (i.e. https://OW-HOST/api/v1/namesapces/NS/actions/ACTION) and the
@@ -950,20 +946,18 @@ function makeErrorResponseObject(err, isWebAction) {
 function makeResponseObject(resp, isWebAction) {
   console.log('makeResponseObject: isWebAction: '+isWebAction);
   if (!isWebAction) {
-    console.log('makeErrorResponseObject: not called as a web action');
+    console.log('makeResponseObject: not called as a web action');
     return resp;
   }
 
-  var bodystr;
+  var bodystr = resp;
   if (typeof resp === 'string') {
-    bodystr = makeJsonString(resp);
-  } else {
-    bodystr = JSON.stringify(resp);
+    bodystr = JSON.parse(makeJsonString(resp));
   }
   retobj = {
     statusCode: 200,
     headers: { 'Content-Type': 'application/json' },
-    body: new Buffer(bodystr).toString('base64')
+    body: bodystr
   };
   return retobj;
 }
--- a/tests/dat/actions/echo-web-http.js
+++ b/tests/dat/actions/echo-web-http.js
@@ -2,11 +2,9 @@
 // license agreements; and to You under the Apache License, Version 2.0.
 
 function main(params) {
-    var bodyobj = params || {};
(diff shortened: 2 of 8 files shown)


rabbah:
Note that we continue to encode JSON for raw http actions receiving `__ow_body`. For symmetry, we should stop doing that as well. This would mean functions would `JSON.parse(param.__ow_body)` instead of `JSON.parse(decode64(param.__ow_body))`.

csantanapr:
This change will allow the following to work, this is a feature request from an user that is calling an external API, this API returns JSON as string, and the web action want to return it as is.

For example this will work and not give an error by the controller
```
function main() {
  return {
    headers: {
     'Content-Type': 'application/json'
    },
    body: '{"foo":"bar"}'
  }
}
```

rabbah:
API gateway actions need to be updated @csantanapr.

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -985,53 +984,7 @@ trait WebActionsApiBaseTests extends ControllerTestCommon with BeforeAndAfterEac
       }
     }
 
-    it should s"handle http web action with base64 encoded known '+json' response (auth? ${creds.isDefined})" in {
```
dubee: @rabbah, should these `+json` tests still be valid for raw web actions?

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
```diff
@@ -840,7 +846,7 @@ trait WebActionsApiBaseTests extends ControllerTestCommon with BeforeAndAfterEac
         }
     }
 
-    it should s"handle http web action with base64 encoded JSON response (auth? ${creds.isDefined})" in {
+    it should s"handle http web action with JSON object as string response (auth? ${creds.isDefined})" in {
```
dubee: Should this test emulate what `invoke web action ensuring JSON value body arguments are received as is ` does? Meaning running thse as the output ` Seq("this is a string".toJson, JsArray(1.toJson, "str str".toJson, false.toJson), true.toJson, 99.toJson)`?

rabbah: Not necessary since all of these would be returned as string. So no extra coverage gained.

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
rabbah: they would be yes - I could change the test that’s removed to behave a raw web instead although note that I opened an issue to make the encoding just be a string in the future.

rabbah: don't think so - these are testing b64 action returns. can you clarify what are you thinking?

csantanapr:
@dubee no need to update docs, encoding base64 json is not documented.

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
dubee: Hmm, I don't understand why they would all be a string?

rabbah: in the test, the action is returning JSON values as string.

rabbah: ` "body" -> JsObject("field" -> "value".toJson).compactPrint.toJson` the type of `body` is JsString regardless of the actual JSON value.

tests/dat/actions/echo-web-http.js
```diff
@@ -2,11 +2,9 @@
 // license agreements; and to You under the Apache License, Version 2.0.
 
 function main(params) {
-    var bodyobj = params || {};
-    bodystr = JSON.stringify(bodyobj);
     return {
         statusCode: 200,
         headers: { 'Content-Type': 'application/json' },
-        body: new Buffer(bodystr).toString('base64')
+        body: params
```
rabbah: should this be `JSON.stringify(params)` instead?

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
dubee: Make sense!

tests/src/test/scala/whisk/core/controller/test/WebActionsApiTests.scala
dubee: Nevermind, I thought raw web actions were going to still return base64, but that doesn't seem to be the case.

tests/dat/actions/echo-web-http.js
csantanapr: `body` can be `string` or `object`, today we support `object` and want to make sure we continue to support it.
With this PR what we are doing is in addition of supporting a `dict`/`object`  to also allow `string`.

or I'm missing something else?

rabbah: im just trying to recover the original behavior where body was returning a string instead of a JSON object.

core/routemgmt/common/apigw-utils.js
```diff
@@ -909,18 +907,16 @@ function makeErrorResponseObject(err, isWebAction) {
...
-    });
-  } else {
-    bodystr = JSON.stringify(err);
+    };
   }
   return {
     statusCode: 502,
     headers: { 'Content-Type': 'application/json' },
-    body: new Buffer(bodystr).toString('base64'),
+    body: bodystr
```
mdeuser: update comment above to remove the base64 business
`*        body: Base64 encoded JSON error string`

core/routemgmt/common/apigw-utils.js
```diff
@@ -954,16 +950,14 @@ function makeResponseObject(resp, isWebAction) {
...
-    bodystr = makeJsonString(resp);
-  } else {
-    bodystr = JSON.stringify(resp);
+    bodystr = JSON.parse(makeJsonString(resp));
   }
   retobj = {
     statusCode: 200,
     headers: { 'Content-Type': 'application/json' },
-    body: new Buffer(bodystr).toString('base64')
+    body: bodystr
```
mdeuser: update comment above to remove the base64 business
* body: Base64 encoded JSON error string

core/routemgmt/common/apigw-utils.js
```diff
@@ -954,16 +950,14 @@ function makeResponseObject(resp, isWebAction) {
     return resp;
   }
 
-  var bodystr;
```
mdeuser: not exactly related.. but simple update...
change line 953/940
`makeErrorResponseObject` -> `makeResponseObject`

cjelger:
As briefly discussed on Slack, this change "breaks" code that used to base64-encode JSON body responses, because this is no longer being decoded. Is this a regression and shall I open a bug for that, or is this intentional? Of course it's way better that we don't have to base64-encode JSON responses, but I'd have expected that the decoding is still done in order to be backwards compatible with actions that did base64-encode JSON responses. WDYT?

rabbah:
#3973 restores the legacy support. This will be deprecated so plan on upgrading your functions.

PR #3921, opened 2018-07-31
Synchronize access to the KafkaConsumer.

markusthoemmes:
The KafkaConsumer is **not** thread-safe, thus concurrent access to it needs to be avoided.

Per https://twitter.github.io/scala_school/concurrency.html, a synchronized block should be quite cheap as long as there is low contention. The contention in this case is controllable via the cadence the monitoring code is running (default 60 seconds). The cost of this synchronization thus should be relatively low.

## Types of changes

## Checklist:


markusthoemmes:
@rabbah the monitoring thread that's asking for endOffsets is not synchronized. I agree, the rest of those should already be pretty safe. Synchronization shouldn't cost much given the contention is probably very low (of poll vs. the relatively slowly cadenced monitoring call)

markusthoemmes:
In a discussion with @sven-lange-last, we concluded to delay this until all the other robustness fixes are in because this one is the least promising in terms of actually adding robustness.

markusthoemmes:
Getting back to this, @sven-lange-last added some log-lines to edge-cases and it became apparent that this actually **is** causing issues, at least with consumer lag reporting. It's plausible that it could cause other unwanted side-effects. Will run this PR through a couple of latency/throughput tests to verify it doesn't regress.

sven-lange-last:
These log entries show that concurrent usage of the Kafka consumer is causing problems - up to the point that Kafka consumer re-recreation failed in the past (before #3922).

```
# tail -20 /tmp/invoker-log.txt
 [2018-08-03T07:57:01.882Z] [INFO] [#tid_sid_invokerNanny] [DockerClientWithFileAccess] running /usr/bin/docker unpause 6698ecf7d8304d66a18f0077223c7e037901ed9d02ac9a6efae93e339ceed172 (timeout: 10 seconds)
 [2018-08-03T07:57:01.912Z] [INFO] [#tid_sid_invokerNanny] [DockerClientWithFileAccess] running /usr/bin/docker rm -f 6698ecf7d8304d66a18f0077223c7e037901ed9d02ac9a6efae93e339ceed172 (timeout: 1 minute)
 [2018-08-03T07:57:17.630Z] [INFO] [#tid_sid_unknown] [KafkaConsumerConnector] lag metric reporting failed for topic 'invoker0': java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access
 2018-08-03T07:57:20.253+0000: 15030.119: [GC (Allocation Failure) [PSYoungGen: 27200K->160K(27648K)] 56080K->29056K(100352K), 0.0035975 secs] [Times: user=0.01 sys=0.00, real=0.00 secs]
 2018-08-03T07:57:48.659+0000: 15058.526: [GC (Allocation Failure) [PSYoungGen: 27296K->96K(29184K)] 56192K->28992K(101888K), 0.0041650 secs] [Times: user=0.00 sys=0.00, real=0.01 secs]
 2018-08-03T07:58:17.253+0000: 15087.119: [GC (Allocation Failure) [PSYoungGen: 27232K->96K(27648K)] 56128K->29000K(100352K), 0.0042057 secs] [Times: user=0.01 sys=0.00, real=0.00 secs]
 [2018-08-03T07:58:17.650Z] [INFO] [#tid_sid_unknown] [KafkaConsumerConnector] lag metric reporting failed for topic 'invoker0': java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access
 2018-08-03T07:58:45.996+0000: 15115.863: [GC (Allocation Failure) [PSYoungGen: 27232K->64K(29184K)] 56136K->28968K(101888K), 0.0032574 secs] [Times: user=0.01 sys=0.00, real=0.00 secs]
 [2018-08-03T07:59:03.161Z] [INFO] [#tid_sid_unknown] [InvokerReactive] updated blacklist to 0 entries
 2018-08-03T07:59:13.253+0000: 15143.120: [GC (Allocation Failure) [PSYoungGen: 27200K->360K(27648K)] 56104K->29280K(100352K), 0.0032846 secs] [Times: user=0.00 sys=0.00, real=0.00 secs]
 [2018-08-03T07:59:17.670Z] [INFO] [#tid_sid_unknown] [KafkaConsumerConnector] lag metric reporting failed for topic 'invoker0': java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access
 2018-08-03T07:59:41.755+0000: 15171.621: [GC (Allocation Failure) [PSYoungGen: 27496K->96K(28672K)] 56416K->29016K(101376K), 0.0036743 secs] [Times: user=0.00 sys=0.00, real=0.00 secs]
 2018-08-03T08:00:10.253+0000: 15200.119: [GC (Allocation Failure) [PSYoungGen: 27232K->96K(27648K)] 56152K->29024K(100352K), 0.0075139 secs] [Times: user=0.00 sys=0.00, real=0.01 secs]
 [2018-08-03T08:00:17.689Z] [INFO] [#tid_sid_unknown] [KafkaConsumerConnector] lag metric reporting failed for topic 'invoker0': java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access
 2018-08-03T08:00:38.001+0000: 15227.867: [GC (Allocation Failure) [PSYoungGen: 27232K->96K(28672K)] 56160K->29024K(101376K), 0.0040564 secs] [Times: user=0.00 sys=0.00, real=0.00 secs]
 2018-08-03T08:01:06.506+0000: 15256.372: [GC (Allocation Failure) [PSYoungGen: 27232K->96K(27648K)] 56160K->29040K(100352K), 0.0037148 secs] [Times: user=0.00 sys=0.00, real=0.00 secs]
 [2018-08-03T08:01:17.709Z] [INFO] [#tid_sid_unknown] [KafkaConsumerConnector] lag metric reporting failed for topic 'invoker0': java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access
 2018-08-03T08:01:35.253+0000: 15285.120: [GC (Allocation Failure) [PSYoungGen: 27232K->96K(28672K)] 56176K->29056K(101376K), 0.0031375 secs] [Times: user=0.00 sys=0.00, real=0.00 secs]
 2018-08-03T08:02:03.631+0000: 15313.498: [GC (Allocation Failure) [PSYoungGen: 27232K->64K(27648K)] 56192K->29024K(100352K), 0.0033097 secs] [Times: user=0.00 sys=0.00, real=0.01 secs]
 [2018-08-03T08:02:17.729Z] [INFO] [#tid_sid_unknown] [KafkaConsumerConnector] lag metric reporting failed for topic 'invoker0': java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access
```

```
# grep “lag metric reporting failed for topic” /tmp/invoker-log.txt | grep “java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access” | wc -l
251
```

```
 [2018-08-03T06:15:15.792Z] [ERROR] [#tid_sid_unknown] [KafkaConsumerConnector] poll returned with failure. Recreating the consumer. Exception: java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access
 [2018-08-03T06:15:15.792Z] [ERROR] [#tid_sid_unknown] [KafkaConsumerConnector] failed to close old consumer while recreating: java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access
 [2018-08-03T06:15:15.848Z] [ERROR] [#tid_sid_dispatcher] [MessageFeed] exception while pulling new activation records: java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access
```

markusthoemmes:
@rabbah Yes this is coming from the monitor "thread". An alternative for intra-connector synchronization is indeed moving the monitoring bits to the MessageFeed and trigger the monitoring cadence from there.

I'm slightly preferring the way of making the consumer synchronized for now since the change is more straightforward and per @sven-lange-last's finding shouldn't impact performance at all, since there never was concurrency in place at all.

Also, the way the MessageFeed is built currently, we'd get quite complex in terms of state-transitions, because (like the ContainerProxy), we'll need interim states like "Pulling", "Monitoring" etc. where the respective calls are mutually excluded etc etc. This seems a lot safer to me, WDYT?

rabbah:
If there’s no perf impact, simpler is better.

chetanmeh:
If I see the code path within `consumer.endOffsets` API call then I do not see any non threadsafe usage. It uses `Fetcher` which internally uses [ConsumerNetworkClient][1] which is marked as thread safe. And it does not uses current consumer state in any place.

One issue with current code is that `offset` is accessed in unsafe manner and which can end up in wrong metric reporting. 

But as the `consumer` is non thread safe we should address that. However it may not fix the root cause

[1]: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerNetworkClient.java#L53

markusthoemmes:
@chetanmeh I feel like the logs posted by @sven-lange-last proof that the KafkaConsumer does indeed have an issue when calling `endOffsets` with `poll`,`commit` or `close`.

I don't think the access to `offsets` is unsafe, because it's only a reading access. Writing is synchronized via the MessageFeed which should only ever call poll once.

> But as the consumer is non thread safe we should address that. However it may not fix the root cause

Can you clarify this? Which root-cause do you refer to?

chetanmeh:
Sorry for the confusion. I was reading [0.11.0.1][1] code (yet not latest master) and that does not had locks but latest [2.0.0][2] code does have locks. 

So  synchronized should fix that issue 

[1]: https://github.com/apache/kafka/blob/c2a0d5f9b1f45bf59552b360fc2b39884357a698/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1534
[2]: https://github.com/apache/kafka/blob/afe00effe2d79e89e159972cf1f5f7ffeb0c6e97/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L2040

markusthoemmes:
Ran a bunch of Latency and Warm Performance tests (as checked in into this repository) and they all indicated the very same performance as before this change. Removing the "blocked" flag therefore.

sven-lange-last:
@rabbah @chetanmeh are you ok with merging this PR or do you think changes are required?

common/scala/src/main/scala/whisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -182,10 +185,11 @@ class KafkaConsumerConnector(
       blocking {
         if (offset > 0) {
           val topicAndPartition = new TopicPartition(topic, 0)
-          consumer.endOffsets(Set(topicAndPartition).asJava).asScala.get(topicAndPartition).foreach { endOffset =>
-            // endOffset could lag behind the offset reported by the consumer internally resulting in negative numbers
-            val queueSize = (endOffset - offset).max(0)
-            MetricEmitter.emitHistogramMetric(queueMetric, queueSize)
+          synchronized(consumer.endOffsets(Set(topicAndPartition).asJava)).asScala.get(topicAndPartition).foreach {
```
rabbah: Here you’re synchronizing again given the outer sync. Is this because of the blocking block?

sven-lange-last: I don't see what you mean by "synchronizing again given the outer sync"? There is no outer sync for this scheduled task. The `blocking` block is just a hint to the underlying thread pool that this Future may block such that the creation of a new thread is desirable to not block too many threads.

rabbah: Referring to Line 159.

sven-lange-last: This line also puzzled me at the first sight. If you expand the section of compared lines in this diff view, you will see that the `synchronized` in line 159 belongs to method `recreateConsumer()` while the `synchronized` you commented on belongs to a totally different scope in the scheduled task.

PR #3923, opened 2018-07-31
Bump Kafka's fetch.max.wait duration and adjust poll's timeout.

markusthoemmes:
Kafka's `fetch.max.wait` duration determines how long the server will wait to send a response if *no record is present*.
The timeout set on `poll` determines how long the client blocks if *no record is present*.

Note that both of these only block if no records are present. Latency should not be impacted, if records are arriving quickly, as both of these timeouts will not apply if records are already there.

The `fetch.max.wait` duration determines how often Kafka is queried though. Big systems can slow down Kafka even by the amount of polling that is happening, even if all the topics are empty.

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -81,7 +81,7 @@ whisk {
             // This value controls the server-side wait time which affects polling latency.
             // A low value improves latency performance but it is important to not set it too low
             // as that will cause excessive busy-waiting.
-            fetch-max-wait-ms = 20
+            fetch-max-wait-ms = 500
             metric-flush-interval-s = 60
         }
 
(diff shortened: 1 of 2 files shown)


sven-lange-last:
[Kafka documentation section `New Consumer Configs`](https://kafka.apache.org/documentation/#newconsumerconfigs):

> `fetch.max.wait.ms`: The maximum amount of time the server will block before answering the fetch request if there isn't sufficient data to immediately satisfy the requirement given by `fetch.min.bytes`.

The default value for `fetch.max.wait.ms` is 500 ms.

> `fetch.min.bytes`: The minimum amount of data the server should return for a fetch request. If insufficient data is available the request will wait for that much data to accumulate before answering the request. The default setting of 1 byte means that fetch requests are answered as soon as a single byte of data is available or the fetch request times out waiting for data to arrive. Setting this to something greater than 1 will cause the server to wait for larger amounts of data to accumulate which can improve server throughput a bit at the cost of some additional latency.

markusthoemmes:
Closing for now. Will be reopened once performance benefits have been measured out and proven.

PR #3932, opened 2018-07-31
eliminate setting of COMPONENT_NAME

dgrove-oss:
Small cleanup: As best as I can tell, this envvar is set when deploying the
invoker/controller but the value is never actually used.


csantanapr:
I thought it’s used on log lines

dgrove-oss:
Any recollection of how?  I did some recursive greps that came up empty, but that might not be conclusive.

markusthoemmes:
The "COMPONENT" part of log-lines is extracted from the logline itself and is essentially the class that logs (i.e. "CouchDbRestStore")

PR #3934, opened 2018-08-01
Adjust Kafka's config to not contain any more warnings.

markusthoemmes:
Currently, the KafkaConsumer warns about a misconfiguration because we added a fantasy value to the kafka config itself. There is a value to control the window size of metrics reporting which roughly fits the needs of the value we need, thus switching to that value also gets rid of the warning in the log.

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -58,6 +58,10 @@ whisk {
     kafka {
         replication-factor = 1
 
+        // Used to control the cadence of the consumer lag check interval
+        consumer-lag-check-interval = 60 seconds
+
+        // The following settings are passed "raw" to the respective Kafka client. Dashes are replaced by dots.
         common {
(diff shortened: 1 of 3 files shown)


common/scala/src/main/resources/application.conf
```diff
@@ -81,7 +81,9 @@ whisk {
             // A low value improves latency performance but it is important to not set it too low
             // as that will cause excessive busy-waiting.
             fetch-max-wait-ms = 20
-            metric-flush-interval-s = 60
+
+            // Used to control the window size of kafka metric reporting and cadence of consumer lag checking
+            metrics-sample-window-ms = 60000 // 60 seconds
```
markusthoemmes: @sven-lange-last made me aware of the fact that this a really bad idea because of conflating two things. Will update.

markusthoemmes:
PG2 3444 👍

Issue #3918, opened 2018-07-30
on "application error", return success=false, but should not destroy the container


tysonnorris:
## Environment details:

* local deployment 

## Steps to reproduce the issue:

1.   create an action that returns as: 
```
return {error: "this was a failure..."}
```
2.   invoke the action repeatedly
3.   Note that each invocation returns an "initTime", indicating no container reuse, and also in invoker logs, container is destroyed, despite a 200 response:
```[2018-07-30T17:39:14.144Z] [INFO] [#tid_Ebq8zgXiJVIQJyJJVxk57kOu6jal0BUB] [DockerContainer] running result: ok [marker:invoker_activationRun_finish:546:11]
[2018-07-30T17:39:14.145Z] [INFO] [#tid_Ebq8zgXiJVIQJyJJVxk57kOu6jal0BUB] [ContainerProxy]  [marker:invoker_collectLogs_start:547]
[2018-07-30T17:39:14.156Z] [INFO] [#tid_Ebq8zgXiJVIQJyJJVxk57kOu6jal0BUB] [InvokerReactive] posted completion of activation 89e79852c3fd4768a79852c3fd47688d
[2018-07-30T17:39:14.175Z] [INFO] [#tid_Ebq8zgXiJVIQJyJJVxk57kOu6jal0BUB] [ContainerProxy]  [marker:invoker_collectLogs_finish:578:31]
[2018-07-30T17:39:14.178Z] [INFO] [#tid_Ebq8zgXiJVIQJyJJVxk57kOu6jal0BUB] [CouchDbRestStore] [PUT] 'whisk_local_activations' saving document: 'id: guest/89e79852c3fd4768a79852c3fd47688d, rev: null' [marker:database_saveDocument_start:581]
[2018-07-30T17:39:14.180Z] [INFO] [#tid_sid_dbBatcher] [CouchDbRestStore] 'whisk_local_activations' saving 1 documents [marker:database_saveDocumentBulk_start:38009]
[2018-07-30T17:39:14.192Z] [INFO] [#tid_sid_invokerNanny] [DockerClientWithFileAccess] running /usr/bin/docker rm -f 4090ec441622f9eedd13c3b244b1f1fa5f3d9d7a74e3d92dad087665a7ff8e79 (timeout: 1 minute) [marker:invoker_docker.rm_start:38018]
``` 

## Provide the expected results and outputs:

Container reuse should be possible on an error that was gracefully handled by the action. 

[...]

PR #3941, opened 2018-08-02
to address #3918, reuse a container on applicationError

tysonnorris:
Reuse a container on applicationError (graceful error from action), only during /run (any error during /init still destroys the container)

## Description
In the Future handling of `ContainerProxy.initializeAndRun()`, this adjustment will allow container to NOT be destroyed when applicationError is returned from the /run request.

This affects warm container reuse in the case where an anticipated error should produce an error result, but should have no impact on container reuse (container is still valid to process next activation). 
Example use case is parameter validation where some sanity check is performed at the start of the action code, and immediately returns `{error: "invalid parameter"}`.

This will
* NOT affect the API/client results (error is still seen on any error from container)
* improve container reuse (since preemptively failing based on user input will no longer cause container destruction)

I considered, but did NOT implement, a change to `WhiskActivation` to add a field that indicates /init failure - since I think it can only be detected in activation logs whether the applicationError response was during /init vs /run, and it may be useful for action developers to have easier access to this bit of info. Sone of the code is convoluted to track the distinction between _applicationError on /init_ and _applicationError on /run_ (e.g. `WhiskActivation` becomes`(WhiskActivation, Boolean)`).

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3918 )

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

- [] Bug fix (generally a non-breaking change which closes an issue).

## Checklist:


core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -371,57 +371,65 @@ class ContainerProxy(
...
-    activationWithLogs.map(_.fold(_.activation, identity)).foreach(storeActivation(tid, _))
+    activationWithLogs.map(_.fold(_.activation, _._1)).foreach(storeActivation(tid, _))
 
     // Disambiguate activation errors and transform the Either into a failed/successful Future respectively.
     activationWithLogs.flatMap {
-      case Right(act) if !act.response.isSuccess => Future.failed(ActivationUnsuccessfulError(act))
-      case Left(error)                           => Future.failed(error)
-      case Right(act)                            => Future.successful(act)
+      //if non-successful, init failures should fail, and non-applicationErrors should also fail
+      case Right((act, initFailure)) if !act.response.isSuccess && (initFailure || !act.response.isApplicationError) =>
```
markusthoemmes: To make this a little bit more straightforward and reduce the size of the change, does it make sense to change https://github.com/apache/incubator-openwhisk/blob/master/common/scala/src/main/scala/whisk/core/containerpool/Container.scala#L114 to be a `containerError`? That way, I think all init failures are non-application errors and thus you can just check for `isSuccess | isApplicationError` here vs. having to thread the extra boolean flag through.

WDYT?

tysonnorris: Sounds reasonable! 👍

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -385,6 +390,70 @@ class ContainerProxyTests
...
+    run(machine, Ready)
+
+    awaitAssert {
+      factory.calls should have size 1
+      container.initializeCount shouldBe 1
+      container.runCount shouldBe 2
+      collector.calls should have size 2
+      container.suspendCount shouldBe 0
+      acker.calls should have size 2
+      store.calls should have size 2
```
markusthoemmes: Should there be a `remove` check that equals 0?

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -385,6 +390,70 @@ class ContainerProxyTests
...
+
+      val initErrorActivation = acker.calls(0)._2
```
rabbah: what are you testing here? I don't follow the rest of the checks here.

tysonnorris: These are the same (except the first run failure) assertions as in the `run an action and continue with a next run without pausing the container` test; I added a comment to indicate this

tests/src/test/scala/whisk/core/containerpool/test/ContainerProxyTests.scala
tysonnorris: Yes! done

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -419,9 +419,10 @@ class ContainerProxy(
 
     // Disambiguate activation errors and transform the Either into a failed/successful Future respectively.
     activationWithLogs.flatMap {
-      case Right(act) if !act.response.isSuccess => Future.failed(ActivationUnsuccessfulError(act))
-      case Left(error)                           => Future.failed(error)
-      case Right(act)                            => Future.successful(act)
+      case Right(act) if !act.response.isSuccess && !act.response.isApplicationError =>
```
dubee: @tysonnorris, is it possible to just check for `!act.response.isApplicationError` here instead?

mgencur:
Hi, I'm wondering what's the rationale behind throwing different types of errors when there's a timeout during init and run phases. Some thoughts:
* how is timeout during initialization different from the timeout during run that we want to treat it differently?
* can the timeout during init be caused by something else than a developer writing the function in a wrong way?
* when I look at the code and the variables it makes sense: ContainerError during init phase and ApplicationError during run phase, but then later the ContainerError translates to "action developer error" which starts to get confusing
Thanks

markusthoemmes:
Good questions @mgencur.

I agree, a timeout on run should be made a container-error (aka action-developer-error) as well. We can see it as an uncaught exception as the action can and should be aware of its own time limits.

On the containerError -> application-developer-error confusion: I agree, feel free to rename the internal methods as you see fit (developerError maybe fo readability?). The translation is weird and non-obvious.

rabbah:
It’s better - would you want to reuse the container if it timed out? You’d end up with a concurrent activation for example. +1

tysonnorris:
Is there a test existing for timeout on run? I don't see one but will try to add it (please let me know if you think it already exists)

I can take a stab at renaming containerError -> developerError, but I think this is also used for docker and other host-related errors:
* memory errors
* container failed to start

I'm not sure these are strictly under control of developer - does that matter?

rabbah:
I think there is such a test I’ll try to find it. 
For the second part I don’t think it matters.

tysonnorris:
I guess this is the test: https://github.com/apache/incubator-openwhisk/blob/8f1dc2e9c848deb9f72da5b636895cbc5c565adb/tests/src/test/scala/whisk/core/containerpool/docker/test/DockerContainerTests.scala#L477

rabbah:
See 
https://github.com/apache/incubator-openwhisk/blob/f6046721acd801ea1777cb5cb040dfaa6922a18e/tests/src/test/scala/whisk/core/limits/MaxActionDurationTests.scala

tysonnorris:
In renaming `containerError` to `developerError`, it suddenly becomes strange to me that the other error is `applicationError` (seems too much like developer Error

tysonnorris:
All the tests are updated; let me know if you have other comments?

markusthoemmes:
Whoops this fell through the cracks, sorry. Any last words @rabbah?

markusthoemmes:
@tysonnorris can you please rebase this to the latest and greatest?

drcariel:
@tysonnorris is the the extent of the change needed for the CLI to facilitate this PR? or do I need to worry about all the pre-existing `ApplicationError` logic?
https://github.com/apache/incubator-openwhisk-cli/pull/364

tysonnorris:
@drcariel AFAIK `ApplicationError` logic should remain as-is, I don't see any logic around ContainerError (now DeveloperError), so I think cli should be fine (aside from the tests, fixed in your PR). Thanks!

Issue #3944, opened 2018-08-06
Store action code as attachments 


chetanmeh:
See [dev thread][1] for background. All action code should be saved as attachments. This work would be done in following steps

- [ ] Support storing action code as attachment as per manifest
- [ ] Remove attachment information from manifest and always store code as attachment
- [ ] Support attachments for blackbox action
- [ ] Simplify existing code to remove parallel hierarchy of `MetaData` classes for various `Exec` types 

[1]: https://lists.apache.org/thread.html/290fbc7cfa54071754afdb62b383018a52586a0953bddebcde500072@%3Cdev.openwhisk.apache.org%3E

PR #3709, opened 2018-05-28
Inlined attachments

chetanmeh:
This PR enables inlining attachment content as part of attachment name for smaller attachments

## Description

Currently [runtime manifest][1] determine if the code related to any runtime is stored as attachment or as string property as part of action document itself. As part of #3450 plan is introduce a new `AttachmentStore` SPI and move code storage to a separate store. In doing that reading an action with code would invoke 2 remote calls. 

To reduce this overhead for smaller codes this PR enables inlining of attachments as part of attachment name itself.

Under this if any   

1. attachment length \>= `maxInlineSize` - Then its saved as proper attachment
2. attachment length < `maxInlineSize` - Then its content would be converted to URL encoded base64 string and would be stored as part of `attachmentName` within the action document. Thus it can read without going to attachment storage

Attachment name for an inlined attachment would be a URI like below

    mem:T3BlbldoaXNrIEZUVw==

Here 

* Scheme - `mem`
* Path - Would be base64 encoded attachment content

### Defaults

This PR sets the default value of `max-inline-size` to 16 KB

## Related issue and scope

- [ ] I opened an issue to propose and discuss this change (#????)

## My changes affect the following components

- [ ] API
- [ ] Controller
- [ ] Message Bus (e.g., Kafka)
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Intrinsic actions (e.g., sequences, conductors)
- [X] Data stores (e.g., CouchDB)
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

- [ ] Bug fix (generally a non-breaking change which closes an issue).
- [x] Enhancement or new feature (adds new functionality).
- [ ] Breaking change (a bug fix or enhancement which changes existing behavior).

## Checklist:

- [x] I signed an [Apache CLA](https://github.com/apache/incubator-openwhisk/blob/master/CONTRIBUTING.md).
- [x] I reviewed the [style guides](https://github.com/apache/incubator-openwhisk/wiki/Contributing:-Git-guidelines#code-readiness) and followed the recommendations (Travis CI will check :).
- [x] I added tests to cover my changes.
- [ ] My changes require further changes to the documentation.
- [ ] I updated the documentation where necessary.

[1]: https://github.com/apache/incubator-openwhisk/blob/master/ansible/files/runtimes.json
[2]: https://console.bluemix.net/docs/services/Cloudant/offerings/bluemix.html#request-and-document-size-limits

PR #3945, opened 2018-08-06
Treat action code as attachments

chetanmeh:
Treats action code as attachment for all kinds in default runtime.

Supersedes #3286. 

## Description

This PR builds on #3286 and enables use of attachment mode for all kinds. Some key aspects are

1. Binary code (like zip/jar) is stored in raw form (NOT Base64 encoded) in attachment
2. Change is backward compatible in reading code
3. Upon update or new action being created the code would be stored as attachment
4. `attachmentType` property of runtime.json is now not used. Instead one of the below is used
    - `text/plain(UTF-8)` - For plain string which are not base64 encoded
    - `application/octet-stream` - For codes which are Base64 string. Such codes are stored in raw form in persistent store

Note that with support for attachment inlining (#3709) it may happen that small code is still stored as part of same document (but in attachment object format)

### Forms of code

Code related to an action can be stored in following forms

#### String property

```json
{
  "exec": {
    "kind": "java",
    "code": "ZHViZWU=",
    "binary": true,
    "main": "hello"
  }
}
```

For Blackbox action also the code is stored as string property

```json
 {
 "exec": {
    "kind": "blackbox",
    "image": "docker-custom.com/openwhisk-runtime/magic/nodejs:0.0.1",
    "binary": false,
    "code": "\"use strict\";\n\nvar fs = require('fs');..."
  }
  }
```

#### Attachment

```json
{
  "exec": {
    "kind": "java",
    "code": {
      "attachmentName": "couch:f41be90d-5656-47f5-9be9-0d5656f7f555",
      "attachmentType": "application/java-archive",
      "length": 32768,
      "digest": "sha256-5a373715b8cdcd608059debe9dae2ad95087eb5322321d1d0b862f8330a0e54d"
    },
    "binary": true,
    "main": "hello"
  }
}
```

### Potential Performance Impact

Currently for non java kinds the code is stored as part of same document. With this change the code may be (if size > `max-inline-size`) stored as attachment. Due to this in case of action invocation if there is a cache miss then it would require 2 remote calls to load the whole action compared to 1 remote call currently.

Note that post #2855 the attachments are now also cached. So if more new containers need to be spinned then they would not incur any further performance impact till action is cached.

### Benefits

With this change it would be possible to store larger codes for setups using store which place restriction on maximum document size.

Further it also improves the heap usage for normal cases where the code is not required to be loaded. See #2847

### TODO

- [x] Review `moveCodeToAttachment.py`
- [x] Add more compatability tests
- [ ] Review default inline size
- [x] Restore support for `jar` property name
- [x] Use `application/octet-stream` for binary code. Even for jars and `text/plain` for non binary code

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3944)

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/ansible/files/runtimes.json
+++ b/ansible/files/runtimes.json
@@ -170,6 +170,10 @@
                     "prefix": "openwhisk",
                     "name": "action-php-v7.1",
                     "tag": "latest"
+                },
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
                 }
             },
             {
@@ -180,6 +184,10 @@
                     "prefix": "openwhisk",
                     "name": "action-php-v7.2",
                     "tag": "latest"
+                },
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
                 }
             }
         ],
--- a/ansible/files/runtimes.json
+++ b/ansible/files/runtimes.json
@@ -8,7 +8,11 @@
                     "name": "nodejsaction",
                     "tag": "latest"
                 },
-                "deprecated": true
+                "deprecated": true,
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
+                }
             },
             {
                 "kind": "nodejs:6",
@@ -19,6 +23,10 @@
                     "tag": "latest"
                 },
                 "deprecated": false,
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
+                },
                 "stemCells": [{
                     "count": 2,
                     "memory": "256 MB"
@@ -32,7 +40,11 @@
                     "name": "action-nodejs-v8",
                     "tag": "latest"
                 },
-                "deprecated": false
+                "deprecated": false,
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
+                }
             }
         ],
         "python": [
@@ -43,7 +55,11 @@
                     "name": "python2action",
                     "tag": "latest"
                 },
-                "deprecated": false
+                "deprecated": false,
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
+                }
             },
             {
                 "kind": "python:2",
@@ -53,7 +69,11 @@
                     "name": "python2action",
                     "tag": "latest"
                 },
-                "deprecated": false
+                "deprecated": false,
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
+                }
             },
             {
                 "kind": "python:3",
@@ -62,7 +82,11 @@
                     "name": "python3action",
                     "tag": "latest"
                 },
-                "deprecated": false
+                "deprecated": false,
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
+                }
             }
         ],
         "swift": [
@@ -73,7 +97,11 @@
                     "name": "swiftaction",
                     "tag": "latest"
                 },
-                "deprecated": true
+                "deprecated": true,
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
+                }
             },
             {
                 "kind": "swift:3",
@@ -82,7 +110,11 @@
                     "name": "swift3action",
                     "tag": "latest"
                 },
-                "deprecated": true
+                "deprecated": true,
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
+                }
             },
             {
                 "kind": "swift:3.1.1",
@@ -91,7 +123,11 @@
                     "name": "action-swift-v3.1.1",
                     "tag": "latest"
                 },
-                "deprecated": false
+                "deprecated": false,
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
+                }
             },
             {
                 "kind": "swift:4.1",
@@ -101,7 +137,11 @@
                     "name": "action-swift-v4.1",
                     "tag": "latest"
                 },
-                "deprecated": false
+                "deprecated": false,
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
+                }
             }
         ],
         "java": [
@@ -115,8 +155,8 @@
                 },
                 "deprecated": false,
                 "attached": {
-                    "attachmentName": "jarfile",
-                    "attachmentType": "application/java-archive"
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
                 },
                 "requireMain": true
             }
@@ -148,6 +188,10 @@
                 "kind": "ruby:2.5",
                 "default": true,
                 "deprecated": false,
+                "attached": {
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
+                },
                 "image": {
                     "prefix": "openwhisk",
                     "name": "action-ruby-v2.5",
--- a/common/scala/src/main/scala/whisk/core/entity/Exec.scala
+++ b/common/scala/src/main/scala/whisk/core/entity/Exec.scala
@@ -308,8 +308,9 @@ protected[core] object Exec extends ArgNormalizer[Exec] with DefaultJsonProtocol
           }
 
           manifest.attached
-            .map { a =>
-              val code = obj.fields.get("code")
+            .map { _ =>
+              // java actions once stored the attachment in "jar" instead of "code"
+              val code = obj.fields.get("code").orElse(obj.fields.get("jar"))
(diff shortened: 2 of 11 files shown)


common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -302,21 +309,29 @@ protected[core] object Exec extends ArgNormalizer[Exec] with DefaultJsonProtocol
...
+              val binary: Boolean = code match {
+                case Some(JsString(c)) => isBinaryCode(c)
+                case _ =>
+                  obj.fields.get("binary") match {
+                    case Some(JsBoolean(b)) => b
+                    case _                  => false
+                  }
+              }
+              val attachment: Attachment[String] = {
+                code
```
markusthoemmes: If I read this correctly, an undefined `code` results in a DeserializationException? If that's the case, why even bother checking for the "binary" field above if code is `None`? The binary field also feels redundant, is it ever used?

As I read it, this could be simplified to:

```scala
val (binary, attachment) = code match {
  case Some(JsString(c)) => 
    (isBinaryCode(c), attFmt[String].read(c))
  case _ => 
    throw new DeserializationException(s"'code' must be a string defined in 'exec' for '$kind' actions")
}
```

Am I missing something?

markusthoemmes: Chatted with @chetanmeh, I was indeed missing something:

`code` can be a JsObject, that breaks my simplification of course.

chetanmeh:
This PR is ready for review. One aspect which is still to be addressed is should we raise the default `max-inline-size` (defaults 16 KB) to some higher value to offset any potential perf impact on small function.

common/scala/src/main/scala/whisk/core/entity/WhiskAction.scala
```diff
@@ -333,23 +335,27 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
...
+        case exec @ CodeExecAsAttachment(_, Inline(code), _, binary) =>
           implicit val logger = db.logging
           implicit val ec = db.executionContext
 
-          val stream = new ByteArrayInputStream(Base64.getDecoder().decode(code))
-          val manifest = exec.manifest.attached.get
+          val (bytes, attachmentType) = if (binary) {
+            (Base64.getDecoder().decode(code), ContentTypes.`application/octet-stream`)
+          } else {
+            (code.getBytes("UTF-8"), ContentTypes.`text/plain(UTF-8)`)
```
markusthoemmes: Please replace "UTF-8" with `StandardCharsets.UTF_8`

common/scala/src/main/scala/whisk/core/entity/Exec.scala
```diff
@@ -300,33 +300,28 @@ protected[core] object Exec extends ArgNormalizer[Exec] with DefaultJsonProtocol
             case None    => throw new DeserializationException(s"kind '$kind' not in $runtimes")
           }
 
+          manifest.attached.map { _ =>
+            }
+
```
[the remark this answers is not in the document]
chetanmeh: Aah ... missed checking prior to push. Fixed it now

chetanmeh:
@dubee new actions would be stored with `code` as object. That format was already being used by Java actions so should work fine. 

```json
{
"code": {
      "attachmentName": "couch:f41be90d-5656-47f5-9be9-0d5656f7f555",
      "attachmentType": "application/java-archive",
      "length": 32768,
      "digest": "sha256-5a373715b8cdcd608059debe9dae2ad95087eb5322321d1d0b862f8330a0e54d"
    }
}
```

Is that you were looking for or you meant something else?

tests/src/test/scala/whisk/core/controller/test/ActionsApiTests.scala
```diff
@@ -790,82 +790,91 @@ class ActionsApiTests extends ControllerTestCommon with WhiskActionsApi {
...
-    val cacheKey = s"${CacheKey(action)}".replace("(", "\\(").replace(")", "\\)")
-    val expectedPutLog =
-      Seq(s"uploading attachment '[\\w-]+' of document 'id: ${action.namespace}/${action.name}", s"caching $cacheKey")
-        .mkString("(?s).*")
-    val notExpectedGetLog = Seq(
-      s"finding document: 'id: ${action.namespace}/${action.name}",
-      s"finding attachment '[\\w-/:]+' of document 'id: ${action.namespace}/${action.name}").mkString("(?s).*")
+    val nodeAction = WhiskAction(namespace, aname(), jsDefault(nonInlinedCode(entityStore)), Parameters("x", "b"))
+    val swiftAction = WhiskAction(namespace, aname(), swift3(nonInlinedCode(entityStore)), Parameters("x", "b"))
+    val actions = Seq((javaAction, JAVA_DEFAULT), (nodeAction, NODEJS6), (swiftAction, SWIFT3))
```
rabbah: Are there defaults for the other runtimes to use as well? Node 6 and Swift 3 might have short shelf life. Not that imprtant can fix later if there isn’t a default.

tests/src/test/scala/whisk/core/controller/test/ActionsApiTests.scala
```diff
@@ -1063,6 +1073,73 @@ class ActionsApiTests extends ControllerTestCommon with WhiskActionsApi {
     stream.reset()
   }
 
+  it should "ensure old and new action schemas are supported" in {
+    implicit val tid = transid()
```
rabbah: Great! Thanks.

tests/src/test/scala/whisk/core/entity/test/ExecTests.scala
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+
+  it should "read code stored as attachment" in {
+    val json = """{
+                 |  "kind": "java",
+                 |  "code": {
+                 |    "attachmentName": "foo:bar",
+                 |    "attachmentType": "application/java-archive",
+                 |    "length": 32768,
```
rabbah: Thanks for adding this test.

rabbah:
lgtm. As for the performance impact, sure two fetches will be slower than one. The size of the inline chunk could be controlled to tune the performance hit (ie based on expected size of code for example). Also whatever compression in performance will likely not match a production environment and hence skewed. Lastly this is the direction we need to go toward for streaming in any case.

ansible/files/runtimes.json
```diff
@@ -115,8 +155,8 @@
                 },
                 "deprecated": false,
                 "attached": {
-                    "attachmentName": "jarfile",
-                    "attachmentType": "application/java-archive"
+                    "attachmentName": "codefile",
+                    "attachmentType": "text/plain"
```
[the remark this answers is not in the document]
chetanmeh: Yes `attachmentName` and `attachmentType` are not used in actual flow. ContentType is just stored but not interpreted so far for any purpose

chetanmeh:
> you may want to alert the dev-list about this change as it is rather significant.

I have sent a mail [on dev][1] sometime back about this change ... would that be ok?. I can also highlight this on wednesday call

> have you run performance tests against this to detect any cold start regressions that may arise from the double database fetch?

@dubee I modified the `ColdBlockingInvokeSimulation` to support tweaking the code size. So one can run this like below which would result in using a 100kb code size and thus use of attachmentMode. Note that default `max-inline-size` is 16kb so we need a value higher than that default size

```
OPENWHISK_HOST="server" USERS="1" REQUESTS_PER_SEC="1" SECONDS="30" CODE_SIZE="100000" ./gradlew gatlingRun-ColdBlockingInvokeSimulation
```

I tried to run the tests but the numbers are not stable

| Pre Change | Post Change |
|---|---|
|image|image|

Numbers for

* 1MB code
* 30 sec run

Below is aggregated stats

image

As these are run on a local setup with all OpenWhisk component running on same host (client on another) they would not convey any real value. To get true sense of impact we would need to run same test on a more prod like setup where CouchDB is remote and also bit populated otherwise it would be streaming stuff from memory!

Unfortunately I do not have access to such a setup so would not be able to get such numbers.

[1]: https://lists.apache.org/thread.html/290fbc7cfa54071754afdb62b383018a52586a0953bddebcde500072@%3Cdev.openwhisk.apache.org%3E

dubee:
@chetanmeh, I'll run some performances tests on this.

rabbah:
What are you hoping to show? The real factor is the size of the Inlined code. If you have to do a fetch for the attachment it’s self evident the cost is additive. 

Having the size of the average code size and perhaps one std dev would be more informative imo.

chetanmeh:
@dubee Good point. Update php runtime entry also to use attachment

chetanmeh:
@dubee Is there anything else to be addressed in this PR? 

Would like to have this merged early next week as there are 3 more PR to be done after this to complete #3944

PR #3950, opened 2018-08-08
Extend system testsuite

mgencur:
This PR adds more tests for Actions, Activations, Packages. 

## Description
This PR adds more tests for Actions, Activations, Packages. Mostly some corner cases which were not covered by the test suite. It also makes possible logging request/response when running the tests. The package tests revealed a problem with Swagger spec which was missing responses with 403 error code. This is fixed with a specific commit.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


tests/src/test/scala/common/rest/WskRestOperations.scala
```diff
@@ -1295,6 +1323,11 @@ trait RunRestCmd extends Matchers with ScalaFutures with SwaggerValidator {
     Try(resp.entity.toStrict(timeout).map { _.data }.map(_.utf8String).futureValue).getOrElse("")
   }
 
+  def getReqData(req: HttpRequest): String = {
+    val timeout = 5.seconds
+    Try(req.entity.toStrict(timeout).map { _.data }.map(_.utf8String).futureValue).getOrElse("")
+  }
```
markusthoemmes: The duplication here seems unnecessary. HttpResponse and HttpRequest have a common supertype, HttpMessage. Changing getResData to, for example, `entityToString` taking a HttpMessage might make sense.

tests/src/test/scala/system/basic/WskActionTests.scala
```diff
@@ -71,6 +71,49 @@ class WskActionTests extends TestHelpers with WskTestHelpers with JsHelpers with
...
+          "end",
+          "logs",
+          "name",
+          "namespace",
+          "publish",
+          "response",
+          "start",
+          "subject",
+          "version")
+        .size shouldBe 12
```
markusthoemmes: When checking for size, ScalaTest has a matcher for that: `should have size 12` will print the whole sequence on failure and thus give you a more explanatory error output than `11 was not 12`

tests/src/test/scala/system/basic/WskActionTests.scala
```diff
@@ -210,6 +253,32 @@ class WskActionTests extends TestHelpers with WskTestHelpers with JsHelpers with
...
+        activation.logs.get.mkString(" ") should include(s"hello, $testString")
+      }
+
+      wsk.action.create(name, Some(TestUtils.getTestActionFilename("hello.py")), update = true)
+
+      val run2 = wsk.action.invoke(name)
+      withActivation(wsk.activation, run2) { activation =>
+        activation.response.status shouldBe "success"
+        activation.logs.get.mkString(" ") should include(s"Hello $testString")
+      }
```
markusthoemmes: Do we need to run these actions? Do we even need a system test for it? Feels like an API level test "just" updating the action and then checking that "parameters" remain unchanged is sufficient here, WDYT?

tests/src/test/scala/system/basic/WskActivationTests.scala
```diff
@@ -74,4 +75,39 @@ class WskActivationTests extends TestHelpers with WskTestHelpers with WskActorSy
...
+    }
+
+    val count = 1
+    val run = wsk.action.invoke(name, Map("n" -> count.toJson))
+    withActivation(wsk.activation, run) { activation =>
+      val res = wsk.activation.get(None, last = Some(true))
+      val jsonResponse = wsk.parseJsonString(res.stdout)
+      jsonResponse.getFields("name").head.toString should include("countdown")
+    }
+  }
```
markusthoemmes: Can these two be api level tests? AFAIK, this testclasses main purpose is to verify the activations API works with several backend ActivationStores.

markusthoemmes:
@mgencur greate someone is looking into our testcoverage, very much appreciated 😊 . I should've created a single review, sorry for the notification spam.

In general: We should move as many tests to the "API level" tests (whisk.core...) as possible as you noted in the issue.

tests/dat/actions/argsPrint.js
```diff
@@ -1,6 +1,8 @@
 // Licensed to the Apache Software Foundation (ASF) under one or more contributor
```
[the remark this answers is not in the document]
mgencur: applicationError.js was not used anywhere. I deleted it and also created argsPrint.js. Github thinks I renamed it.

tests/src/test/scala/common/rest/WskRestOperations.scala
```diff
@@ -1202,6 +1220,16 @@ trait RunRestCmd extends Matchers with ScalaFutures with SwaggerValidator {
...
   }
 
+  def logRequestResponse(req: HttpRequest, res: HttpResponse) {
+    logger.debug(this, s"Request Method: ${req.method}")
+    logger.debug(this, s"Request URI: ${req.uri}")
+    logger.debug(this, s"Request Headers: ${req.headers.mkString(",")}")
+    logger.debug(this, s"Request Body: ${getReqData(req)}")
+    logger.debug(this, s"Response Status: ${res.status.intValue()}")
+    logger.debug(this, s"Response Headers: ${res.headers.mkString(",")}")
+    logger.debug(this, s"Response Body: ${getRespData(res)}")
```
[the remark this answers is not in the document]
mgencur: Yeah. I don't know what I was thinking. Perhaps I wanted to clarify what is headers, body, etc.

tests/src/test/scala/system/basic/WskActionTests.scala
mgencur: Thanks for the tip;)

tests/src/test/scala/system/basic/WskActionTests.scala
mgencur: Perhaps we don't need to invoke it first time but we need it the second time in order to verify that the function can correctly consume the params even if it's a different runtime/language. This is more like end-to-end test.

mgencur:
@markusthoemmes thanks for the quick review. I will address most of the concerns. But I'd like to know what you mean by "API level" tests and how they differ from the system-level tests from your POV. 
My view is that most of the tests from whisk.core (API level as you say) need the whole system to be running. The tests there create and invoke actions like those in the system.basic package. So I don't see a different aside from different package naming.
I'd be happy to move some tests from system.basic to whisk.core as long as their requirements for running components were different. Otherwise it would be nice to have most of the tests in a single package - system.basic, IMO. (actually, I suggested this change in https://github.com/apache/incubator-openwhisk/issues/3737). Having very similar tests in different places is confusing and makes it more difficult to run a subset of the test suite.
What's your view?

markusthoemmes:
@mgencur As far as I'm aware, the API level tests (ActionsApiTests.scala for example) are heavily stubbed out through `ControllerTestCommon`. For example, for the activations API you can just write a couple of activations to the database and then assert that the skip parameter works correctly. For the actions API, the loadbalancer is stubbed out and it's responses are controllable from the test.

It is true that we don't have many unit tests in the true sense but believe we should always test the least amount of components needed. Invoking multiple actions to be able to assert that activations can be skipped while listing only adds to the (already too long) duration of our tests for no apparent benefit.

mgencur:
@markusthoemmes thanks, I can see the API tests now. Will try to move some tests there.

tests/src/test/scala/common/rest/WskRestOperations.scala
```diff
@@ -733,13 +738,23 @@ class RestActivationOperations(implicit val actorSystem: ActorSystem)
...
-    val rr = activationId match {
+    val actId = activationId match {
+      case Some(id) => activationId
+      case None =>
+        last match {
+          case Some(b) if b == true => {
+            val activations = pollFor(N = 1, entity = None, limit = Some(1))
+            require(activations.size <= 1)
+            if (activations.isEmpty) None else Some(activations.head)
+          }
```
markusthoemmes: The pattern match can match the "true" directly, as in:

```scala
last match {
  case Some(true) =>
  ...
```

tests/src/test/scala/system/basic/WskActionTests.scala
```diff
@@ -71,6 +71,49 @@ class WskActionTests extends TestHelpers with WskTestHelpers with JsHelpers with
...
+          "logs",
+          "name",
+          "namespace",
+          "publish",
+          "response",
+          "start",
+          "subject",
+          "version") should have size 12
+
+  }
```
markusthoemmes: I think this test should move to the API-level, as it only verifies the schema of the response vs. doing something that requires to go through the whole system, is that correct?

tests/src/test/scala/system/basic/WskActivationTests.scala
```diff
@@ -21,6 +21,7 @@ import common.rest.WskRestOperations
 import org.junit.runner.RunWith
 import org.scalatest.junit.JUnitRunner
 import common._
+
```
markusthoemmes: Please drop this file from the PR.

tests/src/test/scala/system/basic/WskRestBasicTests.scala
```diff
@@ -320,6 +320,40 @@ class WskRestBasicTests extends TestHelpers with WskTestHelpers with WskActorSys
...
+    }
+
+    //check first param overridden by params file and second param supplied by the file
+    val argsInvoke3 = Some(TestUtils.getTestActionFilename("argsInvokeOverride.json"))
+    val expectedOutput3 = JsObject("param1" -> JsString("value3"), "param2" -> JsString("value2"))
+    val run3 = wsk.action.invoke(name, parameterFile = argsInvoke3)
+    withActivation(wsk.activation, run3) { activation =>
+      activation.response.result shouldBe Some(expectedOutput3)
+    }
+  }
```
markusthoemmes: Parameter files are not even implemented by the API IIRC, as in: The test utility just reads the file and transforms the parameters to be included in the call like every other parameter as well. As such, I think this is only useful as a CLI test.

tests/src/test/scala/whisk/core/cli/test/WskEntitlementTests.scala
```diff
@@ -160,6 +160,46 @@ abstract class WskEntitlementTests extends TestHelpers with WskTestHelpers with
...
+    wsk.pkg.get(s"/${invalidNamespace}/utils", expectedExitCode = forbiddenCode)(wp).stderr should include(
+      "not authorized")
+  }
+
+  it should "reject getting invalid package from valid namespace" in withAssetCleaner(guestWskProps) {
+    (wp, assetHelper) =>
+      val invalidPackage = "utilssss"
+      wsk.pkg.get(s"/whisk.system/${invalidPackage}", expectedExitCode = forbiddenCode)(wp).stderr should include(
+        "not authorized")
+  }
```
markusthoemmes: Should all of these be moved to `PackagesApiTests.scala`?

EDIT: After a discussion via Slack, we agreed to keep these in here for the moment since they revealed missing pieces in the swagger documentation.

Does this only apply to the last two? Could we move the first ones to the API-level test and keep the others in here leaving a TODO to move once the API-level tests can validate swagger conformity as well?

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
```diff
@@ -462,6 +462,58 @@ class ActivationsApiTests extends ControllerTestCommon with WhiskActivationsApi
...
+        aname(),
+        creds.subject,
+        ActivationId.generate(),
+        start = Instant.now,
+        end = Instant.now)
+    }.toList
+
+    try {
+      activations.foreach(storeActivation)
+      waitOnListActivationsInNamespace(namespace, 3)
```
markusthoemmes: The "3" can be derived fromt the activations list: `waitOnListActivationsInNamespace(namespace, activations.size)`

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
```diff
@@ -462,6 +462,58 @@ class ActivationsApiTests extends ControllerTestCommon with WhiskActivationsApi
...
+
+      Get(s"$collectionPath?skip=1") ~> Route.seal(routes(creds)) ~> check {
+        status should be(OK)
+        val activations = responseAs[List[JsObject]]
+        activations.length should be(2)
```
markusthoemmes: Can we narrow this check by checking for the correct activations? Like:

```scala
responseActivations shouldBe activations.map(_.toJson).drop(1)
```

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
```diff
@@ -462,6 +462,58 @@ class ActivationsApiTests extends ControllerTestCommon with WhiskActivationsApi
...
+
+    try {
+      activations.foreach(storeActivation)
+      waitOnListActivationsInNamespace(namespace, 3)
+
+      Get(s"$collectionPath?limit=1") ~> Route.seal(routes(creds)) ~> check {
+        status should be(OK)
+        val activations = responseAs[List[JsObject]]
+        activations.length should be(1)
+      }
```
markusthoemmes: Same comments as above, we should narrow the check and not hardcode the "3" in the waitOn expression.

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
mgencur: There are two problems with this solution:
1) the two json results are not completely equal (one of them has more fields)
2) it looks like the activations might be returned in different order so it might be different every time
Comparing just the size of the results is more robust

markusthoemmes: Hmm, that sounds fishy. If we define the timestamps for the activations that we save, the order should be guaranteed.

I buy the schema difference though, should we then just check for the activation id for example?

mgencur: Well, I'm just reporting what I'm seeing:)
Running this code twice:
`
Get(s"$collectionPath?skip=1") ~> Route.seal(routes(creds)) ~> check {
        status should be(OK)
        val activations = responseAs[List[JsObject]].map(_.fields("name"))
        println(s"Resp: $activations")
        }
`
Gives me:
`
1) Resp: List("activations_tests_name15", "activations_tests_name13")
2) Resp: List("activations_tests_name14", "activations_tests_name15")
`

markusthoemmes: You might need to specifically set the start/end respectively to have each activation after the other. Granularity is milliseconds so it's plausible they all get the same timestamp in this test.

mgencur:
Hi, @markusthoemmes , the CI failure is fixed. This is ready for another review. I think only the last three commits are new. Let me know if I need to squash some commits. Thanks!

tests/dat/actions/argsInvokeOverride.json
```diff
@@ -0,0 +1,4 @@
+{
+  "param1": "value3",
+  "param2": "value2"
+}
```
markusthoemmes: I think these 3 json files are unused, no?

tests/src/test/scala/common/WskCliOperations.scala
```diff
@@ -495,6 +497,10 @@ class CliActivationOperations(val wsk: RunCliCmd) extends ActivationOperations w
       since map { i =>
         Seq("--since", i.toEpochMilli.toString)
       } getOrElse Seq.empty
+    } ++ {
+      since map { i =>
```
rabbah: Since? Should this be skip instead.

tests/src/test/scala/common/WskOperations.scala
```diff
@@ -313,6 +313,7 @@ trait ActivationOperations {
               entity: Option[String],
               limit: Option[Int] = None,
               since: Option[Instant] = None,
+              skip: Option[Int] = Some(0),
```
rabbah: Why doesn’t this default to none?

tests/src/test/scala/common/WskCliOperations.scala
```diff
@@ -480,12 +480,14 @@ class CliActivationOperations(val wsk: RunCliCmd) extends ActivationOperations w
...
    * @param limit (optional) the maximum number of activation to return
    * @param since (optional) only the activations since this timestamp are included
+   * @param skip (optional) the number of activations to skip
    * @param expectedExitCode (optional) the expected exit code for the command
    * if the code is anything but DONTCARE_EXIT, assert the code is as expected
    */
   def list(filter: Option[String] = None,
            limit: Option[Int] = None,
            since: Option[Instant] = None,
+           skip: Option[Int] = Some(0),
```
rabbah: Default to none instead of 0?

tests/src/test/scala/common/rest/WskRestOperations.scala
```diff
@@ -707,18 +709,21 @@ class RestActivationOperations(implicit val actorSystem: ActorSystem)
...
    * @param since (optional) only the activations since this timestamp are included
+   * @param skip (optional) the number of activations to skip
    * @param retries the maximum retries (total timeout is retries + 1 seconds)
    * @return activation ids found, caller must check length of sequence
    */
   override def pollFor(N: Int,
                        entity: Option[String],
                        limit: Option[Int] = Some(30),
                        since: Option[Instant] = None,
+                       skip: Option[Int] = Some(0),
```
rabbah: Same q why is default some 0 not none.

tests/src/test/scala/system/basic/WskActionTests.scala
```diff
@@ -71,6 +71,21 @@ class WskActionTests extends TestHelpers with WskTestHelpers with JsHelpers with
     }
   }
 
+  it should "invoke an action that throws an uncaught exception and returns correct status code" in withAssetCleaner(
```
rabbah: System tests are intended to for sniffing the health of the system. This seems unnecessary here and the test could be moved to the core actions test suite, no?

tests/src/test/scala/system/basic/WskActionTests.scala
rabbah: So here you are testing that the bound parameter is not lost? Did you check the actions api tests for a comparable test? If there’s one there do we need to add this? If so, should it also be out of the system suite.

tests/src/test/scala/whisk/core/cli/test/WskEntitlementTests.scala
rabbah: This suggests that the swagger validation should be added to the unit tests as well - of which there are many. Is this something we might be able to do? Otherwise we might end up adding a lot of system level tests that do the same thing as the unit tests. 

I’m not entirely sure we can integrate swagger validation though in the unit tests - at least not in the way Ben did it for the system tests. But I have not researched this. I’m mentioning it in case of a black hole here that’s hard to get out of.

tests/src/test/scala/system/basic/WskActionTests.scala
mgencur: This is to test that during execution of the action the correct status is returned. I'm not sure how this could be tested without actually invoking the action.

tests/src/test/scala/common/rest/WskRestOperations.scala
```diff
@@ -733,13 +738,23 @@ class RestActivationOperations(implicit val actorSystem: ActorSystem)
                    fieldFilter: Option[String] = None,
                    last: Option[Boolean] = None,
                    summary: Option[Boolean] = None)(implicit wp: WskProps): RestResult = {
-    val rr = activationId match {
+    val actId = activationId match {
+      case Some(id) => activationId
+      case None =>
+        last match {
```
rabbah: Caution here is that last night be stale.

tests/src/test/scala/common/rest/WskRestOperations.scala
```diff
@@ -733,13 +738,23 @@ class RestActivationOperations(implicit val actorSystem: ActorSystem)
                    fieldFilter: Option[String] = None,
                    last: Option[Boolean] = None,
                    summary: Option[Boolean] = None)(implicit wp: WskProps): RestResult = {
-    val rr = activationId match {
+    val actId = activationId match {
+      case Some(id) => activationId
```
rabbah: @markusthoemmes orElse?

tests/src/test/scala/whisk/core/controller/test/ActivationsApiTests.scala
```diff
@@ -462,6 +462,57 @@ class ActivationsApiTests extends ControllerTestCommon with WhiskActivationsApi
     }
   }
 
+  it should "skip activations and return correct ones" in {
+    implicit val tid = transid()
```
rabbah: @chetanmeh do you mind taking a look at these skip related tests as it’s something you’ve tweaked in the past.

rabbah:
Thanks for the extra tests. I think I got through the comments that maybe of of the goals is better swagger coverage.

tests/src/test/scala/common/rest/WskRestOperations.scala
mgencur: Sorry?

tests/src/test/scala/common/rest/WskRestOperations.scala
```diff
@@ -733,13 +738,23 @@ class RestActivationOperations(implicit val actorSystem: ActorSystem)
...
+      case None =>
+        last match {
+          case Some(true) => {
+            val activations = pollFor(N = 1, entity = None, limit = Some(1))
+            require(activations.size <= 1)
+            if (activations.isEmpty) None else Some(activations.head)
+          }
+          case _ => None
```
markusthoemmes: Right, this can be rewritten to:

```scala
val actId = activationId.orElse(last.filter(_ == true).flatMap { _ =>
  val activations = pollFor(N = 1, entity = None, limit = Some(1))
  require(activations.size <= 1)
  activations.headOption
})
```

`orElse` returns the value of the first option (activationId in your case) or falls back to a second option.
`filter` on an option returns a None if the filter is not fullfilled.
`headOption` returns the head of a list, if its non-empty, otherwise None.

Optional to implement, just mentioning how combinators could be used here.

mgencur:
Added a few minor fixes.

tests/src/test/scala/whisk/core/cli/test/WskEntitlementTests.scala
markusthoemmes: I looked briefly into this. Ben's interface essentially needs an HttpRequest and its corresponding HttpResponse, both of which are available in the unit tests as well. @mgencur already opened an issue (#3959) to track that work. We agreed to defer that to a followup PR.

tests/src/test/scala/system/basic/WskActionTests.scala
rabbah: That’s fine. There are “basic” action tests in core that do the same. Historically the system package implied tests might be used to check the health of a deployment.

tests/src/test/scala/common/rest/WskRestOperations.scala
rabbah: The value of the “last” activation might be the last id indexed not necessarily the last id stored. This is fine for your test related changes from what I scanned but felt it necessary to mention it as this behavior of last isn’t generally obvious.

tests/src/test/scala/system/basic/WskActionTests.scala
rabbah: fair - I think the api test iirc (didn’t verify again) checks with the parameters with a get instead of invoke.

markusthoemmes:
@mgencur is this ready to go from your PoV?
@rabbah seems like there are some bits of your feedback unaddressed?

tests/src/test/scala/system/basic/WskActionTests.scala
rabbah: `run2` could just be `run`? minor point.

rabbah:
This has a conflict which needs to resolve. Will try to do it.

rabbah:
Resolved conflict, fixed variable. This is ready to merge.

mgencur:
Thanks for handling the small fix and rebase.

PR #3951, opened 2018-08-08
Docs fixes for action invocations

mgencur:
More accurate description of blocking calls and the situation when _application error_ is returned upon invoking an action.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [x] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -237,9 +237,10 @@ Each action invocation results in an activation record which contains the follow
 - `response`: A dictionary that defines the following keys
   - `status`: The activation result, which might be one of the following values:
     - *"success"*: the action invocation completed successfully.
-    - *"application error"*: the action was invoked, but returned an error value on purpose, for instance because a precondition on the arguments was not met. This status code is also returned when the action exceeds its time limit.
+    - *"application error"*: the action was invoked, but returned an error value on purpose, for instance because a precondition on the arguments was not met.
     - *"action developer error"*: the action was invoked, but it completed abnormally, for instance the action did not detect an exception, or a syntax error existed. This status code is also returned under specific conditions such as:
-      - the action failed to initialize for any reason (including time out)
+      - the action failed to initialize for any reason
+      - the action exceeded its time limit during the init or run phase
       - the action specified a wrong docker container name
       - the action did not properly implement the expected [runtime protocol](actions-new.md)
     - *"whisk internal error"*: the system was unable to invoke the action.
@@ -347,8 +348,7 @@ is skipped if an action is dispatched to a previously initialized container --- 
 You can tell if an [invocation was a warm activation or a cold one requiring initialization](annotations.md#annotations-specific-to-activations)
 by inspecting the activation record.
 - An action runs for a bounded amount of time. This limit can be configured per action, and applies to both the
-initialization and the execution separately. If the action time limit is exceeded during initialization, the activation's
-response status is _action developer error_. If the action time limit is exceeded during the run phase it is _application error_.
+initialization and the execution separately. If the action time limit is exceeded during the initialization or run phase, the activation's response status is _action developer error_.
 - Functions should follow best practices to reduce [vulnerabilities](security.md) by treating input as untrusted,
 and be aware of vulnerabilities they may inherit from third-party dependencies.
 
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -216,10 +216,11 @@ ok: invoked /whisk.system/samples/greeting with id 5975c24de0114ef2b5c24de0118ef
 A blocking invocation request will _wait_ for the activation result to be available. The wait period
 is the lesser of 60 seconds or the action's configured
 [time limit](reference.md#per-action-timeout-ms-default-60s).
-The result of the activation is returned if it is available within the wait period.
-Otherwise, the activation continues processing in the system and an activation ID is returned
-so that one may check for the result later, as with non-blocking requests
-(see [here](#watching-action-output) for tips on monitoring activations).
+The result of the action is successfully returned if it is available within the wait period.
+Otherwise, the result is an error caused by the action timeout and the response status is _application error_.
+The action continues processing in the system for a while until it is terminated. 
+Subsequent calls to `wsk activation get <activationId> --summary` will return the original error even if
+the action finished its execution before it was terminated.
 
 ### Understanding the activation record
 
@@ -340,7 +341,7 @@ is skipped if an action is dispatched to a previously initialized container --- 
 You can tell if an [invocation was a warm activation or a cold one requiring initialization](annotations.md#annotations-specific-to-activations)
 by inspecting the activation record.
 - An action runs for a bounded amount of time. This limit can be configured per action, and applies to both the
-initialization and the execution separately.
+initialization and the execution separately. If the action time limit is exceeded the activation's response status is _application error_.
 - Functions should follow best practices to reduce [vulnerabilities](security.md) by treating input as untrusted,
 and be aware of vulnerabilities they may inherit from third-party dependencies.
 
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -235,10 +235,10 @@ Each action invocation results in an activation record which contains the follow
   - `status`: The activation result, which might be one of the following values:
     - *"success"*: the action invocation completed successfully.
     - *"application error"*: the action was invoked, but returned an error value on purpose, for instance because a precondition on the arguments was not met. This status code is also returned under specific conditions such as:
+      - action timed out
+    - *"action developer error"*: the action was invoked, but it completed abnormally, for instance the action did not detect an exception, or a syntax error existed. This status code is also returned under specific conditions such as:
       - wrong docker container name
       - initialization of a function failed
-      - action timed out
-    - *"action developer error"*: the action was invoked, but it completed abnormally, for instance the action did not detect an exception, or a syntax error existed.
     - *"whisk internal error"*: the system was unable to invoke the action.
   - `success`: Is *true* if and only if the status is *"success"*.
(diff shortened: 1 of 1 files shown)


docs/actions.md
```diff
@@ -216,10 +216,11 @@ ok: invoked /whisk.system/samples/greeting with id 5975c24de0114ef2b5c24de0118ef
 A blocking invocation request will _wait_ for the activation result to be available. The wait period
 is the lesser of 60 seconds or the action's configured
 [time limit](reference.md#per-action-timeout-ms-default-60s).
-The result of the activation is returned if it is available within the wait period.
-Otherwise, the activation continues processing in the system and an activation ID is returned
-so that one may check for the result later, as with non-blocking requests
-(see [here](#watching-action-output) for tips on monitoring activations).
+The result of the action is successfully returned if it is available within the wait period.
+Otherwise, the result is an error caused by the action timeout and the response status is _application error_.
```
rabbah: This is not quite true. The API response is 202 (accepted) and an activation id, not an application error (which would imply the activation completed) if the blocking period (60s) expires before a result is ready.

docs/actions.md
```diff
@@ -233,7 +234,10 @@ Each action invocation results in an activation record which contains the follow
 - `response`: A dictionary that defines the following keys
   - `status`: The activation result, which might be one of the following values:
     - *"success"*: the action invocation completed successfully.
-    - *"application error"*: the action invocation was successful, but the action returned an error value on purpose, for instance because a precondition on the arguments was not met.
+    - *"application error"*: the action was invoked, but returned an error value on purpose, for instance because a precondition on the arguments was not met. This status code is also returned under specific conditions such as:
+      - wrong docker container name
+      - initialization of a function failed
+      - action timed out
```
rabbah: some of these are about to change; see #3941.

docs/actions.md
```diff
@@ -340,7 +344,7 @@ is skipped if an action is dispatched to a previously initialized container ---
 You can tell if an [invocation was a warm activation or a cold one requiring initialization](annotations.md#annotations-specific-to-activations)
 by inspecting the activation record.
 - An action runs for a bounded amount of time. This limit can be configured per action, and applies to both the
-initialization and the execution separately.
+initialization and the execution separately. If the action time limit is exceeded the activation's response status is _application error_.
```
rabbah: This changes in #3941 - @tysonnorris @markusthoemmes please confirm.

mgencur: I think it only changes for the init phase.

docs/actions.md
mgencur: This is about blocking invocation. Here's my test:
`wsk action create sleep tests/dat/actions/sleep.js -t 100`
`wsk action invoke sleep --param sleepTimeInMs 1100 --blocking`
The response code is 502 and response itself is:
`ok: invoked /guest/sleep with id 75043a3d2eec4556843a3d2eec855620
{
    "activationId": "75043a3d2eec4556843a3d2eec855620",
    "annotations": [
        {
            "key": "path",
            "value": "guest/sleep"
        },
        {
            "key": "waitTime",
            "value": 1014
        },
        {
            "key": "kind",
            "value": "nodejs:6"
        },
        {
            "key": "limits",
            "value": {
                "logs": 10,
                "memory": 256,
                "timeout": 100
            }
        },
        {
            "key": "initTime",
            "value": 152
        }
    ],
    "duration": 255,
    "end": 1533718832936,
    "logs": [],
    "name": "sleep",
    "namespace": "guest",
    "publish": false,
    "response": {
        "result": {
            "error": "The action exceeded its time limits of 100 milliseconds."
        },
        "success": false
    },
    "start": 1533718832681,
    "subject": "guest",
    "version": "0.0.1"
}
`

docs/actions.md
mgencur: OK. Will change it.

docs/actions.md
markusthoemmes: Ah, the text reads as if you're referring to the blocking wait period. We wait for 60s at most for a blocking invocation to finish, after that we fall back to a 202 (actions can be configured to run longer than 60s). That's what @rabbah is referring to.

mgencur: I see. Let me rephrase it then.

mgencur:
Hey @rabbah @markusthoemmes , what do you think now? I've fixed the description a little bit in separate commits.

docs/actions.md
```diff
@@ -216,10 +216,17 @@ ok: invoked /whisk.system/samples/greeting with id 5975c24de0114ef2b5c24de0118ef
 A blocking invocation request will _wait_ for the activation result to be available. The wait period
 is the lesser of 60 seconds or the action's configured
 [time limit](reference.md#per-action-timeout-ms-default-60s).
-The result of the activation is returned if it is available within the wait period.
+
+The result of the activation is returned if it is available within the blocking wait period (this wait period is by default 60 seconds as well).
```
rabbah: does it make sense to fold this line into the sentence on 216 - a bit redundant IMO.

docs/actions.md
```diff
@@ -216,10 +216,17 @@ ok: invoked /whisk.system/samples/greeting with id 5975c24de0114ef2b5c24de0118ef
...
 Otherwise, the activation continues processing in the system and an activation ID is returned
 so that one may check for the result later, as with non-blocking requests
 (see [here](#watching-action-output) for tips on monitoring activations).
+If execution of the action exceeds the action's configured time limit within the blocking wait period
```
rabbah: If this is strictly intended to explain the activation's response status, then I suggest dropping the sentence on line 224 because what you write next is true for _all_ activations that exceed their configured duration.

docs/actions.md
```diff
@@ -216,10 +216,17 @@ ok: invoked /whisk.system/samples/greeting with id 5975c24de0114ef2b5c24de0118ef
...
+the result of the invocation is an error caused by the action timeout and the response status is _application error_.
```
rabbah: I don't think we've explained application error or response status at this point; this happens in the next section (understanding the activation record). How about this:

When an action exceeds its configured time lime, the activation record will indicate this error. See [understanding the activation record] for more details.

or something to that effect.

docs/actions.md
```diff
@@ -233,8 +240,11 @@ Each action invocation results in an activation record which contains the follow
 - `response`: A dictionary that defines the following keys
   - `status`: The activation result, which might be one of the following values:
     - *"success"*: the action invocation completed successfully.
-    - *"application error"*: the action invocation was successful, but the action returned an error value on purpose, for instance because a precondition on the arguments was not met.
-    - *"action developer error"*: the action was invoked, but it completed abnormally, for instance the action did not detect an exception, or a syntax error existed.
+    - *"application error"*: the action was invoked, but returned an error value on purpose, for instance because a precondition on the arguments was not met. This status code is also returned under specific conditions such as:
+      - action timed out
```
rabbah: since there's only one other condition, how about:

This status code is also set when the action exceeds its configured time limit.

docs/actions.md
```diff
@@ -233,8 +240,11 @@ Each action invocation results in an activation record which contains the follow
...
+    - *"action developer error"*: the action was invoked, but it completed abnormally, for instance the action did not detect an exception, or a syntax error existed. This status code is also returned under specific conditions such as:
+      - wrong docker container name
+      - initialization of a function failed
```
rabbah: - the function failed to initialize for any reason (including time out)
      - a Docker action specified the wrong docker container name
      - a Docker action that does not properly implement the expected runtime protocol (link to actions-new.md)?

docs/actions.md
rabbah: I suggest:

If the action time limit is exceeded during initialization, the activation's response status is ... otherwise it is ...

rabbah:
Thanks for improving the docs - this is important and appreciated.

docs/actions.md
mgencur: AFAIK, the action default timeout is 60 seconds, the blocking wait timeout is by default 60 as well. It could certainly be worded differently but I think it's good as it is.

docs/actions.md
mgencur: Unless the configured time limit is longer than 60 seconds and the blocking wait time is the default. In this case it returns 202 as you said previously? Curious...

docs/actions.md
rabbah: just seems redundant as line 217 already says 60 seconds (you can add this is the default there) and strictly not correct since an action with a timeout less than 60s for an action, it's a little confusing as worded.

docs/actions.md
mgencur: Could be.

docs/actions.md
mgencur: ok

docs/actions.md
mgencur: ok

docs/actions.md
rabbah: There are two things at play:
1. the HTTP response to the client: if the action has not finished in 60 seconds, the status code is 202 and an activation id is returned
2. in all cases (unless the system fails to generate one), the activation record for an action is created and recorded, and if the action timed out, the response status will be application error (for run) or developer error (for init).

Is that clearer?

mgencur:
hey @rabbah, wdyt now?

rabbah:
LGTM now - I’ll leave it to Markus to clarify the init vs run timeout.

markusthoemmes:
@rabbah Clarified in the other PR, I don't see a reason not to make them both behave the same. One could argue that an action timeout is a form of an uncaught exception (as the action is aware of its own deadline). Making them both non-application errors makes sense to me.

mgencur:
@rabbah @markusthoemmes alright, thanks. So let's wait with this PR until the other one is completely sorted out. Then I can make a little change to this one if necessary.

mgencur:
hey, I've updated the description of errors in the event of timeouts, after https://github.com/apache/incubator-openwhisk/pull/3941 was integrated.

PR #3963, opened 2018-08-13
k8s: implement invoker-node affinity and eliminate usage of kubectl

dgrove-oss:
1. Upgrade to latest released version of the fabric8 Kubernetes client
       to get access to an implementation of node affinity. Use that implementation
       to optionally add a scheduling affinity to the pods created for actions
       to bind them to worker nodes labeled as invoker nodes.
    
    2. implement the container removal operation via the kube rest client
       instead of via an exec to the kubectl cli.  This eliminates the last
       usage of kubectl in the KubernetesClient and allows the kubectl
       executable to be removed from the invoker Docker image.

### DIFF
--- a/core/invoker/Dockerfile
+++ b/core/invoker/Dockerfile
@@ -4,7 +4,6 @@
 FROM scala
 
 ENV DOCKER_VERSION 1.12.0
-ENV KUBERNETES_VERSION 1.6.4
 
 RUN apk add --update openssl
 
@@ -16,11 +15,6 @@ tar --strip-components 1 -xvzf docker-${DOCKER_VERSION}.tgz -C /usr/bin docker/d
 rm -f docker-${DOCKER_VERSION}.tgz && \
 chmod +x /usr/bin/docker && \
 chmod +x /usr/bin/docker-runc
-
-# Install kubernetes client
-RUN wget --no-verbose https://storage.googleapis.com/kubernetes-release/release/v${KUBERNETES_VERSION}/bin/linux/amd64/kubectl && \
-chmod +x kubectl && \
-mv kubectl /usr/bin/kubectl
 
 ADD build/distributions/invoker.tar ./
 
--- a/core/invoker/src/main/resources/application.conf
+++ b/core/invoker/src/main/resources/application.conf
@@ -46,13 +46,16 @@ whisk {
     # Timeouts for k8s commands. Set to "Inf" to disable timeout.
     timeouts {
       run: 1 minute
-      rm: 1 minute
-      inspect: 1 minute
       logs: 1 minute
     }
     invoker-agent {
       enabled: false
       port: 3233
+    }
+    user-pod-node-affinity {
+      enabled: true
+      key: "openwhisk-role"
+      value: "invoker"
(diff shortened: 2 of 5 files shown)


dgrove-oss:
PG2 3499 🏃

markusthoemmes:
@dgrove-oss what prevents us from removing the binary here?

dgrove-oss:
I was going to stage it in a separate PR to make it easier to find and revert just that piece in case we wanted the binary back later (or it was being used in openshift in some way I wasn't aware of).  I could remove the binary from the Dockerfile in this PR if you'd prefer a single PR.

dgrove-oss:
pushed a second commit to remove the kubectl executable from the Dockerfile.

core/invoker/src/main/resources/application.conf
```diff
@@ -54,6 +54,11 @@ whisk {
       enabled: false
       port: 3233
     }
+    invoker-node-affinity {
```
markusthoemmes: Should this be "pod-node-affinity" or "user-pod-node-affinity"?

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -166,11 +173,27 @@ class KubernetesClient(
...
+          .inNamespace(kubeRestClient.getNamespace)
+          .pods()
+          .withName(container.id.asString)
+          .delete()
+      }
+    }.map(_ => ())
   }
 
   def rm(key: String, value: String, ensureUnpaused: Boolean = false)(implicit transid: TransactionId): Future[Unit] = {
-    runCmd(Seq("delete", "--now", "pod", "-l", s"$key=$value"), config.timeouts.rm).map(_ => ())
```
markusthoemmes: Are the timeouts still in use? If not: Delete from the config?

core/invoker/Dockerfile
```diff
@@ -17,11 +16,6 @@ rm -f docker-${DOCKER_VERSION}.tgz && \
 chmod +x /usr/bin/docker && \
 chmod +x /usr/bin/docker-runc
 
-# Install kubernetes client
```
ddragosd: :+1:

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -75,10 +71,17 @@ case class KubernetesClientTimeoutConfig(run: Duration, rm: Duration, inspect: D
  */
 case class KubernetesInvokerAgentConfig(enabled: Boolean, port: Int)
 
+/**
+ * Configuration for invoker node affinity for user action containers
+ */
+case class KubernetesInvokerNodeAffinity(enabled: Boolean, key: String, value: String)
```
ddragosd: Can we document what key,value are used for ?

dgrove-oss:
addressed review comments & squashed to single commit.  Should be ready to merge once CI completes.

PR #3969, opened 2018-08-15
use KeepAlive(false) (SocketOption) instead of Connection:close (HTTP header)

tysonnorris:
Some intermittent build failures in various runtimes that use `ActionContainer` for tests (which uses the akka http client as of recently) shows some errors where connections are prematurely terminated. 

## Description

In separate testing, I think this is related to use of `Connection: close` request header, to dissuade connection reuse.

This change:
* removes `Connection:close` header
* uses instead the `SO.KeepAlive(false)` (SocketOption); similar to the apache client setting to disable connection reuse

Since Akka client is not enabled by default in existing code except the `ActionContainer` test base class, this change and its affects are not obvious, until runtimes using this test class are running on this change. Locally testing, I was able to intermittently reproduce problems with the previous version, and have not seen problems with this new version.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/core/containerpool/AkkaContainerClient.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/AkkaContainerClient.scala
@@ -27,7 +27,6 @@ import akka.http.scaladsl.model.MediaTypes
 import akka.http.scaladsl.model.MessageEntity
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.model.headers.Accept
-import akka.http.scaladsl.model.headers.Connection
 import akka.http.scaladsl.unmarshalling.Unmarshal
 import akka.stream.StreamTcpException
 import akka.stream.scaladsl.Sink
@@ -56,6 +55,7 @@ import whisk.http.PoolingRestClient
  * It allows to POST a JSON object and receive JSON object back; that is the
  * content type and the accept headers are both 'application/json.
  * This implementation uses the akka http host-level client API.
+ * NOTE: Keepalive is disabled to prevent issues with paused containers
  *
  * @param hostname the host name
  * @param port the port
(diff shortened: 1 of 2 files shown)


markusthoemmes:
> Locally testing, I was able to intermittently reproduce problems with the previous version, and have not seen problems with this new version.

Can you specify what these problems were? The setting of the Apache client is not necessarily related to the underlying socket but rather instructs the connection pool not to reuse connections at all.

What you're disabling here is [TCP Keepalive probing](https://www.tldp.org/HOWTO/html_single/TCP-Keepalive-HOWTO/#whatis) and not necessarily relevant to HTTP connection reuse at all.

tysonnorris:
Here is an example of the problem https://github.com/apache/incubator-openwhisk/pull/2795#issuecomment-412958409

Previously, when using "Connection: close" header, akka client would periodically get these connection errors (but not always, and not with all client tests). I changed the keepalive setting based on the keepalive handling added to `ApacheBlockingContainerClient`, but I agree this impl is not related to HTTP connection pooling.

Some options I see are:
* determine a more appropriate way to disable connection reuse besides keepalive SocketOption (this was the only approach I found)
* determine why `Connection:close` causes intermittent failures with akka client (have had no luck there so far )
* don't use `PoolingRestClient` or `Connection: close` in `AkkaContainerClient`, rather force a new connection on each request using `Http().outgoingConnection(<container ip>)`; I'm not sure the real performance impact here, since [docs](https://doc.akka.io/docs/akka-http/current/client-side/connection-level.html) mention it is 'slow'
* use `PoolingRestClient`, but not `Connection:close`, and allow max connection reuse, but add `client.close()` as part of the `Container.pause()` workflow (to dispose connections when container is paused). Connection pool will be recreated on unpause.

markusthoemmes:
The thing is: Does disabling TCP keepAlive also disable reuse of that connection since that's what we want to impact.

tysonnorris:
@markusthoemmes given that I'm not sure how to verify this, what about the last approach I mentioned (and was there a reason to not do this when the other client was changed to not reuse connectioned?):

> use PoolingRestClient, but not Connection:close, and allow max connection reuse, but add client.close() as part of the Container.pause() workflow (to dispose connections when container is paused). Connection pool will be recreated on unpause.

PR #3970, opened 2018-08-15
Fix macOS bash update instructions in docs/cli.md

apjanke:
## Description

The existing description of updating bash on macOS tells users to install an updated bash by installing `bash-completion` via brew. The `bash-completion` package is not bash itself; it is a set of completion definitions for various programs. To actually get a newer bash, users should do `brew install bash` instead.

This PR also adds a line to clarify that the updated bash is installed in `/usr/local/bin` (as opposed to updating the system default bash).

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [X] Documentation

## Types of changes

## Checklist:

- [ ] I added tests to cover my changes. *This is a doco-only change; no relevant tests to add.*

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -59,12 +59,15 @@ To verify your CLI setup, try [creating and running an action](./samples.md).
 ### Configure command completion for Openwhisk CLI
 
 For bash command completion to work, bash 4.1 or newer is required. The most recent Linux distributions should have the correct version of bash but Mac users will most likely have an older version.
+
 Mac users can check their bash version and update it by running the following commands:
 
 ```
 bash --version
-brew install bash-completion
+brew install bash
 ```
+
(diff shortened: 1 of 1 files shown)


rabbah:
Thanks for this fix. We should also include the bits for installing bash-completion as a separate step --- isn't that still required?

apjanke:
>  We should also include the bits for installing bash-completion as a separate step --- isn't that still required?

I don't think so. The bash completion mechanism itself is part of core bash. The bash-completion project just supplies completion definitions for various external commands, and doesn't include OpenWhisk completions. It looks to me like the OpenWhisk bash completion definitions are supplied by OpenWhisk itself, in the generated `wsk_cli_bash_completion.sh` file.

PR #3909, opened 2018-07-29
Refactor sentinel message into shared place.

tz70s:
Previously, kubernetes container depends on sentinel message defined by
docker container object. Not ideal for separated SPI impl, put this to a
share place instead.

Signed-off-by: Tzu-Chiao Yeh <su3g4284zo6y7@gmail.com>

## Related issue and scope

- [ ] I opened an issue to propose and discuss this change (#????)

## My changes affect the following components

- [ ] API
- [ ] Controller
- [ ] Message Bus (e.g., Kafka)
- [ ] Loadbalancer
- [x] Invoker
- [ ] Intrinsic actions (e.g., sequences, conductors)
- [ ] Data stores (e.g., CouchDB)
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

- [ ] Bug fix (generally a non-breaking change which closes an issue).
- [x] Enhancement or new feature (adds new functionality).
- [ ] Breaking change (a bug fix or enhancement which changes existing behavior).

## Checklist:

- [x] I signed an [Apache CLA](https://github.com/apache/incubator-openwhisk/blob/master/CONTRIBUTING.md).
- [x] I reviewed the [style guides](https://github.com/apache/incubator-openwhisk/wiki/Contributing:-Git-guidelines#code-readiness) and followed the recommendations (Travis CI will check :).
- [x] I added tests to cover my changes.
- [ ] My changes require further changes to the documentation.
- [ ] I updated the documentation where necessary.

PR #3974, opened 2018-08-16
Support for start-of-activation log sentinel

dgrove-oss:
## Description

Add support for filtering out an optional start-of-activation log sentinel
line that will contain the activation id in addition to the
sentinel marker.  For runtimes that inject this sentinel line into their
outputs, it becomes quite simple to do an asynchronous stream processing
operation on their log streams and forward log lines enriched with activation ids
to a platform logging service.

Also add a Fluent Bit configuration/script and Dockerfile to process logs in this format.

Appended is an example from an action that writes one line to stdout and no lines to stderr:

stdout stream:
```
XXX_THE_START_OF_A_WHISK_ACTIVATION_XXX with id cafca5b74be94eb8bca5b74be9beb80f
Here's a friendly message
XXX_THE_END_OF_A_WHISK_ACTIVATION_XXX
```

```
stderr stream:
XXX_THE_START_OF_A_WHISK_ACTIVATION_XXX with id cafca5b74be94eb8bca5b74be9beb80f 
XXX_THE_END_OF_A_WHISK_ACTIVATION_XXX
```


dgrove-oss:
A proof of concept implementation of the runtime changes for nodejs is in https://github.com/apache/incubator-openwhisk-runtime-nodejs/pull/81

dgrove-oss:
rebased to resolve conflicts with merged PR #3909

markusthoemmes:
This has been sitting in WIP, got stale and is inactive since 2 months. Any update?

dgrove-oss:
Not much has happened.   I do still think having a start sentinel is potentially useful, but there were some additional wrinkles I hadn't considered that came up in discussion of this PR.
1.  We can't trust the `namespace` information in the start sentinel (malicious users could write actions that generate many spurious start sentinels to their action's logs targeting another namespace).  So we still need an additional mechanism to associate the namespace for the log.
2. Depending on implementation details of the underlying container engine, we may need a way to block deletion of the container until log processing has completed (or we could lose user logs).  This means there could still need to be some coupling between log processing and the invoker's container management state machine, which reduces the potential simplification of having purely streaming logs.

dgrove-oss:
since it was easy, I rebased to resolve conflicts & squashed back to a single commit.

dgrove-oss:
Closing this PR as it seems more fruitful to push on full integration with platform logging and/or fully structured logging instead of this partial solution.

PR #3976, opened 2018-08-16
close connections and reset pool on Container.pause

tysonnorris:
## Description
Instead of restricting connection reuse, destroy connection pool on Container.pause.
Partly from discussion at https://github.com/apache/incubator-openwhisk/pull/3969#issuecomment-413590542
and partly to enable connection reuse to prep for concurrent activations (#2795)
we should not constraint connection handling at the client, but rather dispose the connection pool at time of pause (since it may leave the client pool in a bad state if the container is paused then resumed) 

(replaces the fix added in #3969 )

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/core/containerpool/Container.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/Container.scala
@@ -70,7 +70,10 @@ trait Container {
   protected var httpConnection: Option[ContainerClient] = None
 
   /** Stops the container from consuming CPU cycles. */
-  def suspend()(implicit transid: TransactionId): Future[Unit]
+  def suspend()(implicit transid: TransactionId): Future[Unit] = {
+    //close connection and reset the connection pool
+    closeConnections().andThen { case _ => httpConnection = None }
+  }
 
   /** Dual of halt. */
   def resume()(implicit transid: TransactionId): Future[Unit]
@@ -80,7 +83,7 @@ trait Container {
 
   /** Completely destroys this instance of the container. */
   def destroy()(implicit transid: TransactionId): Future[Unit] = {
-    Future.successful(httpConnection.foreach(_.close()))
+    closeConnections()
   }
 
   /** Initializes code in the container. */
@@ -188,6 +191,9 @@ trait Container {
         val finished = Instant.now()
         RunResult(Interval(started, finished), response)
       }
+  }
+  private def closeConnections(): Future[Unit] = {
+    Future.successful(httpConnection.foreach(_.close()))
   }
 }
 
--- a/common/scala/src/main/scala/whisk/core/containerpool/Container.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/Container.scala
@@ -196,10 +196,7 @@ trait Container {
       }
   }
   private def closeConnections(toClose: Option[ContainerClient]): Future[Unit] = {
-    toClose match {
-      case Some(conn) => conn.close()
-      case None       => Future.successful(Unit)
-    }
+    toClose.map(_.close()).getOrElse(Future.successful(()))
   }
 }
 
--- a/common/scala/src/main/scala/whisk/core/containerpool/Container.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/Container.scala
@@ -71,8 +71,11 @@ trait Container {
 
   /** Stops the container from consuming CPU cycles. */
   def suspend()(implicit transid: TransactionId): Future[Unit] = {
-    //close connection and reset the connection pool
-    closeConnections().andThen { case _ => httpConnection = None }
+    //close connection first, then close connection pool
+    //(testing pool recreation vs connection closing, time was similar - so using the simpler recreation approach)
+    val toClose = httpConnection
+    httpConnection = None
+    closeConnections(toClose)
   }
 
   /** Dual of halt. */
@@ -83,7 +86,7 @@ trait Container {
 
   /** Completely destroys this instance of the container. */
   def destroy()(implicit transid: TransactionId): Future[Unit] = {
-    closeConnections()
+    closeConnections(httpConnection)
   }
 
   /** Initializes code in the container. */
@@ -192,8 +195,11 @@ trait Container {
         RunResult(Interval(started, finished), response)
       }
   }
-  private def closeConnections(): Future[Unit] = {
-    Future.successful(httpConnection.foreach(_.close()))
+  private def closeConnections(toClose: Option[ContainerClient]): Future[Unit] = {
+    toClose match {
+      case Some(conn) => conn.close()
+      case None       => Future.successful(Unit)
+    }
   }
 }
 
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
@@ -177,9 +177,7 @@ class DockerContainer(protected val id: ContainerId,
   protected val filePollInterval: FiniteDuration = 5.milliseconds
 
   override def suspend()(implicit transid: TransactionId): Future[Unit] = {
-    super.suspend().andThen {
-      case _ => if (useRunc) { runc.pause(id) } else { docker.pause(id) }
-    }
+    super.suspend().flatMap(_ => if (useRunc) runc.pause(id) else docker.pause(id))
   }
   def resume()(implicit transid: TransactionId): Future[Unit] =
     if (useRunc) { runc.resume(id) } else { docker.unpause(id) }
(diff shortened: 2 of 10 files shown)


common/scala/src/main/scala/whisk/core/containerpool/Container.scala
```diff
@@ -189,6 +192,9 @@ trait Container {
         RunResult(Interval(started, finished), response)
       }
   }
+  private def closeConnections(): Future[Unit] = {
+    Future.successful(httpConnection.foreach(_.close()))
+  }
```
markusthoemmes: Should we break the `AutoClosable` usage (I know, I'm guilty here) and change the interface of `ContainerClient`'s `close` method to be `Future[Unit]`? That way, we can make use of akka's asynchronous closing without blocking the thread on that operation. Apache would of course still be blocking. WDYT?

Edit: Reading through some of the code, this is actually a `materializer.close()` for the Akka based client. That means we'll churn through Materializers in certain scenarios. Can you run the test suites against this change to see if that has any impact? IIRC Materializers are rather expensive to create.

common/scala/src/main/scala/whisk/core/containerpool/Container.scala
```diff
@@ -70,7 +70,10 @@ trait Container {
   protected var httpConnection: Option[ContainerClient] = None
 
   /** Stops the container from consuming CPU cycles. */
-  def suspend()(implicit transid: TransactionId): Future[Unit]
+  def suspend()(implicit transid: TransactionId): Future[Unit] = {
+    //close connection and reset the connection pool
+    closeConnections().andThen { case _ => httpConnection = None }
```
markusthoemmes: I think this should set `httpConnection = None` before running `closeConnections`. Otherwise you might get connections running on a closing client when you unpause directly.

common/scala/src/main/scala/whisk/core/containerpool/Container.scala
markusthoemmes: Looked through it briefly:

For the Apache based client, the ConnectionManager has a `closeIdleConnections()` method, which's timeout you can set to 0, so it should force-close all connections.

For the Akka based client, you need to get access to the `Http.HostConnectionPool`. You'll then have access to its underlying gateway which has a `shutdown` and a corresponding `startPool` method, which is documented to play just nice in tandem with `shutdown`. 
Alternatively, we can opt to just regenerate the hostConnectionPool itself for this case, if we think that's the safer alternative.

Measuring will tell if this is necessary optimisation though.

tysonnorris:
> Is there a way to flush all connections from the respective pool instead of regenerating the whole thing?
I didn't see one for akka http - how do you want to proceed?

> get access to the Http.HostConnectionPool 
It is not clear how to do this, starting with pool, which is type `Flow[(HttpRequest, T), (Try[HttpResponse], T), HostConnectionPool]`

tysonnorris:
@markusthoemmes one thing we can do here is rely on `host-connection-pool.idle-timeout`, instead of explicitly calling `pool.shutdown` (still not sure how to get the materialized instance of the existing pool...). If we set that `idle-timeout` to the `pause-grace` value - the pool should go away around the time that pause occurs. If it happens after the pause (since we loose control to wait for it during pause processing), hopefully that is ok? If it is not ok, we could always set the timeout to `pause-grace - 10ms` (and in some cases the pool will be recreated before pause ever occurs).  This is definitely less precise, but it may be an option if the explicit pool shutdown is not accessible.

markusthoemmes:
Yeah, streams and materialized values tend to get a little tricky. You can do this, to get the pool out of stream, along with the queue:

```scala
private val (requestQueue, connectionPool) = Source
    .queue(queueSize, OverflowStrategy.dropNew)
    .viaMat(httpFlow.getOrElse(pool))(Keep.both)
    .to(Sink.foreach({
      case (response, p) => p.complete(response)
    }))
    .run
```

(The type of the `httpFlow` value needs to change to have the `HostConnectionPool` as a materialized value as well. We should be able to put a `null`  there in the tests I think)

The idle-timeout idea is great as well! If the above doesn't work reliably, we can try that. A third way is to pass the ActorMaterializer into the client (like it arguably should do anyway) and fix the `shutdown` method to actually just shutdown the connection pool.

Note: This is all based on the assumption that destroying/recreating the whole thing is too costly. It might make sense to proof that if the optimization gets too ugly.

tysonnorris:
I made a test in `PoolingRestClientTests` that completely recreates client for each of 1000 requests, recreating the whole thing seems to add <2ms to requests on avg, in my machine. So let's say this would add ~2ms latency to unpause workflow (at first run on a paused container). 

I think this is tolerable - WDYT?

I wanted to test with the `pool.shutdown()` approach as well, but after your hint above, I'm then stuck at wondering how to deal with the signature on the flow either being `Any` or `Http.HostConnectionPool`.  With `Any` on the flow sig, the connectionPool ends up as `Any` type, and with `Http.HostConnectionPool` on the flow sig, the connectionPool ends up as the same type, but I'm not sure how to mock `Http.HostConnectionPool` in the test flows (created in test classes). 
Maybe `Any` type is ok here, but will require some casting - is there some better way?

common/scala/src/main/scala/whisk/core/containerpool/Container.scala
```diff
@@ -189,6 +195,12 @@ trait Container {
         RunResult(Interval(started, finished), response)
       }
   }
+  private def closeConnections(toClose: Option[ContainerClient]): Future[Unit] = {
+    toClose match {
+      case Some(conn) => conn.close()
+      case None       => Future.successful(Unit)
+    }
```
markusthoemmes: Nit: No need to match in this case imho.

```scala
toClose.map(_.close()).getOrElse(Future.successful(())
```

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -176,8 +176,11 @@ class DockerContainer(protected val id: ContainerId,
   protected val waitForOomState: FiniteDuration = 2.seconds
   protected val filePollInterval: FiniteDuration = 5.milliseconds
 
-  def suspend()(implicit transid: TransactionId): Future[Unit] =
-    if (useRunc) { runc.pause(id) } else { docker.pause(id) }
+  override def suspend()(implicit transid: TransactionId): Future[Unit] = {
+    super.suspend().andThen {
+      case _ => if (useRunc) { runc.pause(id) } else { docker.pause(id) }
+    }
```
markusthoemmes: We shouldn't use `andThen` here, because the Future should only complete after the pause operation has finished, `andThen` doesn't do that.

```scala
super.suspend().flatMap(_ => if(useRunc) runc.pause(id) else docker.pause(id))
```

Should be the way to go. We shouldn't ignore a failure of `super.suspend`, should we?

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
```diff
@@ -103,7 +103,11 @@ class KubernetesContainer(protected[core] val id: ContainerId,
 
   protected val waitForLogs: FiniteDuration = 2.seconds
 
-  def suspend()(implicit transid: TransactionId): Future[Unit] = kubernetes.suspend(this)
+  override def suspend()(implicit transid: TransactionId): Future[Unit] = {
+    super.suspend().andThen {
+      case _ => kubernetes.suspend(this)
+    }
+  }
```
markusthoemmes: Same as above on `andThen`

tysonnorris:
Yes tested apache client as well, similar ~2ms for recreating the client vs reusing the client. BTW, I left the `NoConnectionReuseStrategy` in place for apache client, but in theory that can be removed, if it is only there to avoid issues during pause/resume. We can adjust it later, or leave it if it is not causing problems.

Other comments addressed - thanks for the feedback!

markusthoemmes:
@tysonnorris as a matter of fact, relying on connection reuse caused issues with several runtimes (namely Java and Python). It might be worthwile checking when you do the concurrency work there (if even needed) if these handle reused connections correctly. +1 for leaving that in for now.

tysonnorris:
For the java+python runtimes - was the issue around paused containers? or around reusing connections within an unpaused container? I'm not sure how you could tell, but regardless we can test with java/python and concurrent activations.

markusthoemmes:
@tysonnorris My theory at the time was, that they just didn't support the "Connection" headers properly and confused the ConnectionPool.

PR #3980, opened 2018-08-21
Log activation retrieval from activation store for block invocations

dubee:
## Description

Adds a log message when an activation record is retrieved from the activation store for blocking invocations. This will help us determine how often the DB polling flow is used to retrieve blocking invocation activations.

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/common/Logging.scala
+++ b/common/scala/src/main/scala/whisk/common/Logging.scala
@@ -260,6 +260,7 @@ object LoggingMarkers {
   // Time of the activation in controller until it is delivered to Kafka
   val CONTROLLER_ACTIVATION = LogMarkerToken(controller, activation, start)
   val CONTROLLER_ACTIVATION_BLOCKING = LogMarkerToken(controller, "blockingActivation", start)
+  val CONTROLLER_ACTIVATION_BLOCKING_FROM_DB_POLL = LogMarkerToken(controller, "blockingActivationPoll", count)
 
   // Time that is needed load balance the activation
   val CONTROLLER_LOADBALANCER = LogMarkerToken(controller, loadbalancer, start)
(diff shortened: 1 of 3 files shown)


chetanmeh:
Would be useful to also capture it as a metric which would simplify get insight into rate of this flow happening

rabbah:
Perhaps obvious but will mention anyway that this will be highly dependent on offered load.

common/scala/src/main/scala/whisk/common/Logging.scala
```diff
@@ -260,6 +260,7 @@ object LoggingMarkers {
   // Time of the activation in controller until it is delivered to Kafka
   val CONTROLLER_ACTIVATION = LogMarkerToken(controller, activation, start)
   val CONTROLLER_ACTIVATION_BLOCKING = LogMarkerToken(controller, "blockingActivation", start)
+  val CONTROLLER_ACTIVATION_BLOCKING_ACTIVATION_STORE = LogMarkerToken(controller, activation, count)
```
markusthoemmes: "controller_activation_count" doesn't feel like the right metric to use here. `controller_activationPoll_count` maybe?

dubee:
Okay, added the metric and added a more appropriate metric name.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -613,7 +613,13 @@ protected[actions] trait PrimitiveActions {
     if (!result.isCompleted && retries < maxRetries) {
       val schedule = actorSystem.scheduler.scheduleOnce(wait(retries)) {
         activationStore.get(ActivationId(docid.asString), context).onComplete {
-          case Success(activation)             => result.trySuccess(Right(activation))
+          case Success(activation) =>
```
rabbah: do you just want to count the number of times the result is retrieved from the DB directly vs active ack, or the total number of times this poller is sheduled?

dubee: @rabbah, I just want to count how many times the polling mechanism is actually used to get the activation instead of the active ack.

chetanmeh:
Lets add this new metric to [metrics docs][1]

[1]: https://github.com/apache/incubator-openwhisk/blob/master/docs/metrics.md#metric-details

csantanapr:
It looks like all review comments are address 🚀

PR #3984, opened 2018-08-23
Simplify configuring OpenWhisk

chetanmeh:
Enables support for passing typesafe config string as part of environment variables. 

## Description

Currently to configure OpenWhisk settings in Controller and Invoker one needs to reverse map the typesafe config to environment variable following specific convention. For example setting `whisk.loadbalancer.invoker-busy-threshold` would need env name like `CONFIG_whisk_loadbalancer_invokerBusyThreshold`. This currently works but poses following problems

1. Mapping config name to env name has some transformation involved and is a manual process
2. Flattening the config tree in env variables obscures the config and makes it tricky to get the context
3. Mapping list entries in config to env variable is tricky

It would be better to have a way where we can pass in the config from a file in typesafe config format to the container via some environment variable. This brings back all the [benefits of typesafe config](https://github.com/lightbend/config/blob/master/HOCON.md#goals--background) for custom config.

### Usage

This PR currently does not convert the existing ansible setup to switch to config format and just enables someone to use the new approach.

#### Step 1 - Config in File

Define the required config in a file. This can also be a jinja template

```
include classpath("application.conf")

whisk {
  couchdb {
    protocol = "{{ db.protocol }}"
    host     = "{{ db.host }}"
    port     = "{{ db.port }}"
    username = "{{ db.username }}"
    password = "{{ db.password }}"
    databases {
      WhiskAuth       = "{{ db.whisk.auth }}"
      WhiskEntity     = "{{ db.whisk.actions }}"
      WhiskActivation = "{{ db.whisk.activations }}"
    }
  }
}

```

#### Step 2 - Configure env variable

Configure env variable `OPENWHISK_CONFIG` with content of  above file

```yml
- name: Load config from template
  set_fact:
      openwhisk_config: "{{ lookup('template', 'config.j2') }}"

- name: populate environment variables for controller
  set_fact:
    env:
      "OPENWHISK_CONFIG": "{{ openwhisk_config }}"
```

With this at runtime config would also include config provided from the config file in addition to config set via env variables

### Implementation Detail 

This PR modifies the `transformEnvironment.sh` to copy the content of `OPENWHISK_CONFIG` to file `/openwhisk.conf` on file system. It also configures the system property `-Dconfig.file='/openwhisk.conf` which makes [typesafe config logic load](https://github.com/lightbend/config#standard-behavior) this file. 
 

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/transformEnvironment.sh
+++ b/common/scala/transformEnvironment.sh
@@ -36,8 +36,9 @@ props=()
 
 if [ -n "$OPENWHISK_CONFIG" ]
 then
-   printf "%s" "$OPENWHISK_CONFIG" > "/openwhisk.conf"
-   props+=("-Dconfig.file='/openwhisk.conf'")
+    location="/config.conf"
+    printf "%s" "$OPENWHISK_CONFIG" > "$location"
+    props+=("-Dconfig.file='$location'")
 fi
 
 for var in $configVariables
--- a/common/scala/transformEnvironment.sh
+++ b/common/scala/transformEnvironment.sh
@@ -34,6 +34,12 @@ configVariables=$(compgen -v | grep $prefix)
 
 props=()
 
+if [ -n "$OPENWHISK_CONFIG" ]
+then
+   printf "%s" "$OPENWHISK_CONFIG" > "/openwhisk.conf"
+   props+=("-Dconfig.file='/openwhisk.conf'")
+fi
+
 for var in $configVariables
 do
     value=$(printenv "$var")


common/scala/transformEnvironment.sh
```diff
@@ -34,6 +34,12 @@ configVariables=$(compgen -v | grep $prefix)
 
 props=()
 
+if [ -n "$OPENWHISK_CONFIG" ]
+then
+   printf "%s" "$OPENWHISK_CONFIG" > "/openwhisk.conf"
+   props+=("-Dconfig.file='/openwhisk.conf'")
+fi
```
markusthoemmes: Can we generalize this a bit? As of now, this script is very non-OpenWhisk-dependent. What about:

```bash
if [ -n "$CONFIGFILE_CONTENT" ]
then
    location="/config.conf"
    echo "$CONFIGFILE_CONTENT" > "$location"
    props+=("-Dconfig.file='$location'")
fi
```

Note: I switched from `printf` to `echo`. Any reason to use `printf` here?
Note 2: Please fix the indenting from 3 to 4 spaces.

chetanmeh: Used `printf` based on [this answer][1] on SO.

[1]: https://stackoverflow.com/a/49418406/1035417

chetanmeh: > CONFIGFILE_CONTENT

I preferred `OPENWHISK_CONFIG` as it makes it explicit that its an OpenWhisk related config and not generic one. So once we document its usage it would make it obvious to end user.

markusthoemmes: As I said, personally I didn't like making this file OpenWhisk specific. I guess that's a very minor thing anyway though.

`printf` seems sensible thanks.

chetanmeh:
@markusthoemmes Done the formatting change. I can try to make this script generic by passing in the env name to lookup as parameter but that would introduce some complexity. I have changed the file name to `config.conf`

PR #3999, opened 2018-09-03
Mark activationRun as failed if there is an ContainerConnectionError

jiangpengcheng:
Just as the title said

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/core/containerpool/Container.scala
+++ b/common/scala/src/main/scala/whisk/core/containerpool/Container.scala
@@ -20,7 +20,7 @@ package whisk.core.containerpool
 import java.time.Instant
 
 import akka.actor.ActorSystem
-import akka.event.Logging.InfoLevel
+import akka.event.Logging.{ErrorLevel, InfoLevel}
 import akka.stream.scaladsl.Source
 import akka.util.ByteString
 import pureconfig._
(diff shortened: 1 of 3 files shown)


common/scala/src/main/scala/whisk/core/containerpool/Container.scala
```diff
@@ -142,12 +142,16 @@ trait Container {
...
       .andThen { // never fails
         case Success(r: RunResult) =>
-          transid.finished(
-            this,
-            start.copy(start = r.interval.start),
-            s"running result: ${r.toBriefString}",
-            endTime = r.interval.end,
-            logLevel = InfoLevel)
+          r.response match {
+            case Left(_) => transid.failed(this, start, s"run failed with ${r.toBriefString}", logLevel = ErrorLevel)
```
markusthoemmes: Can this be caused by user behavior or is this guaranteed to be an error of the system?

jiangpengcheng: I think so as the `ContainerConnectionError` here only occurred while connecting to action containers, or is there any thing users can do to terminate/prevent such connections?

markusthoemmes: For blackbox actions for example, the run path can be broken.

common/scala/src/main/scala/whisk/core/containerpool/Container.scala
```diff
@@ -142,12 +142,16 @@ trait Container {
     callContainer("/run", body, timeout, retry = false)
       .andThen { // never fails
         case Success(r: RunResult) =>
```
vvraskin: Did you consider adding a similar behaviour for the `/init` call as well? I see nearly the same code in https://github.com/apache/incubator-openwhisk/blob/master/common/scala/src/main/scala/whisk/core/containerpool/Container.scala#L102 

or maybe even extracting it into a separate method would help here...

common/scala/src/main/scala/whisk/core/containerpool/Container.scala
jiangpengcheng: ok, but that should be marked as error too in my opinion , as the activation is failed anyway

markusthoemmes: It is not an ERROR in the sense of the system is at fault though. Ideally, you'd want to monitor "ERROR" log messages in your production system and judge based on that if something is wrong. That loglevel should never be in the hand of the user of the system imo.

PR #4000, opened 2018-09-03
Put error message as tag into failed trace.

tz70s:
Prior error message doesn't put error tag into failed trace.
This is nice to have and useful on trouble shooting within request context via modern UI in Jaegar or Zipkin.

With a minor change in primitive action flow which didn't enclose with failed trace.

I'm not sure folks are consistent with tracing system because of conflicting goals that we already have transid in log. IMHO, at least, it shouldn't make traces confused once if user decides to take it into system: whether there's an accurate error message existed or not, we should know which one get failed.

Semantics followed by: 
https://github.com/opentracing/specification/blob/master/semantic_conventions.md

Snapshots in Jaegar:

firstpage
second

Signed-off-by: Tzu-Chiao Yeh <su3g4284zo6y7@gmail.com>

## Related issue and scope

- [x] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/common/tracing/OpenTracingProvider.scala
+++ b/common/scala/src/main/scala/whisk/common/tracing/OpenTracingProvider.scala
@@ -80,7 +80,7 @@ class OpenTracer(val tracer: Tracer, tracingConfig: TracingConfig, ticker: Ticke
    * @param transactionId
    */
   override def finishSpan(transactionId: TransactionId): Unit = {
-    clear(transactionId)
+    clear(transactionId, withErrorMessage = None)
   }
 
   /**
@@ -88,8 +88,8 @@ class OpenTracer(val tracer: Tracer, tracingConfig: TracingConfig, ticker: Ticke
    *
    * @param transactionId
    */
-  override def error(transactionId: TransactionId): Unit = {
-    clear(transactionId)
+  override def error(transactionId: TransactionId, message: => String): Unit = {
+    clear(transactionId, withErrorMessage = Some(message))
   }
 
   /**
@@ -122,17 +122,24 @@ class OpenTracer(val tracer: Tracer, tracingConfig: TracingConfig, ticker: Ticke
     }
   }
 
-  private def clear(transactionId: TransactionId): Unit = {
+  private def clear(transactionId: TransactionId, withErrorMessage: Option[String]): Unit = {
     spanMap.get(transactionId.meta.id).foreach {
       case head :: Nil =>
+        withErrorMessage.foreach(setErrorTags(head, _))
         head.finish()
         spanMap.remove(transactionId.meta.id)
         contextMap.remove(transactionId.meta.id)
       case head :: tail =>
+        withErrorMessage.foreach(setErrorTags(head, _))
         head.finish()
         spanMap.put(transactionId.meta.id, tail)
       case Nil =>
     }
+  }
+
+  private def setErrorTags(span: Span, message: => String): Unit = {
+    span.setTag("error", true)
+    span.setTag("message", message)
   }
 
   private def configureCache[T, R](): collection.concurrent.Map[T, R] =
@@ -149,7 +156,7 @@ class OpenTracer(val tracer: Tracer, tracingConfig: TracingConfig, ticker: Ticke
 trait WhiskTracer {
   def startSpan(logMarker: LogMarkerToken, transactionId: TransactionId): Unit = {}
   def finishSpan(transactionId: TransactionId): Unit = {}
-  def error(transactionId: TransactionId): Unit = {}
+  def error(transactionId: TransactionId, message: => String): Unit = {}
   def getTraceContext(transactionId: TransactionId): Option[Map[String, String]] = None
   def setTraceContext(transactionId: TransactionId, context: Option[Map[String, String]]): Unit = {}
 }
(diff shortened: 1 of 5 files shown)


chetanmeh:
Surfacing error state in trace would be useful. @sandeep-paliwal Thoughts?

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -174,23 +175,24 @@ protected[actions] trait PrimitiveActions {
 
     val postedFuture = loadBalancer.publish(action, message)
 
-    postedFuture.flatMap { activeAckResponse =>
-      // successfully posted activation request to the message bus
-      transid.finished(this, startLoadbalancer)
-
+    postedFuture andThen {
+      case Success(_) => transid.finished(this, startLoadbalancer)
+      case Failure(e) => transid.failed(this, startLoadbalancer, e.getMessage, logLevel = ErrorLevel)
```
vvraskin: Should the error log level be eliminated here? An activation failure doesn't always imply a system failure

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -174,23 +175,24 @@ protected[actions] trait PrimitiveActions {
...
-            .andThen { case _ => transid.finished(this, startActivation) }
         }
         .getOrElse {
           // no, return the activation id
-          transid.finished(this, startActivation)
           Future.successful(Left(message.activationId))
         }
+    } andThen {
+      case Success(_) => transid.finished(this, startActivation)
+      case Failure(e) => transid.failed(this, startActivation, e.getMessage, logLevel = ErrorLevel)
```
vvraskin: The same question regarding the log level...

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
tz70s: Thanks for reminding, we can use the default warn level.

Should this get changed as well?
https://github.com/apache/incubator-openwhisk/blob/0242a60aeb976d070abda76a8ac2aa7e3306e393/core/controller/src/main/scala/whisk/core/controller/Actions.scala#L292

Or I can match exception types for different levels. However, I think logging here is quite redundant b.c. we've already logged this at `doInvoke`.

BTW, in opentracing, there's no standardized warn tag provided, is this fine?

sandeep-paliwal:
@chetanmeh agree. Having error information in trace for failed Invocations will be very useful.

common/scala/src/main/scala/whisk/common/tracing/OpenTracingProvider.scala
```diff
@@ -122,19 +122,26 @@ class OpenTracer(val tracer: Tracer, tracingConfig: TracingConfig, ticker: Ticke
...
+        withErrorMessage.foreach(setErrorTags(head, _))
         head.finish()
         spanMap.put(transactionId.meta.id, tail)
       case Nil =>
     }
   }
 
+  private def setErrorTags(span: Span, message: => String): Unit = {
+    span.setTag("error", true)
+    span.setTag("message", message)
```
sandeep-paliwal: Tag is usually added to make it easy to search for a specific span. Since key in this case is always going to be "error" do you think its more appropriate case for `span.log()` ?

tz70s: Hi @sandeep-paliwal, thanks for suggestion.
I think the big difference between tag and log is timestamp; as you mentioned, the tag is much easier to search for a specific span, or make it explicit on what's going on.

I've tried [capturing errors guidance in open tracing data model standard](
https://github.com/opentracing/specification/blob/master/semantic_conventions.md#captured-errors). This is not that useful to know that what's really going on in Jaegar. 

Some code snippet:
```Scala
private def setErrorTags(span: Span, message: => String): Unit = {
  span.log(Map("event" -> "error", "message" -> message).asJava)
}
```
Snapshot:

The error is not explicit.



Am I missing something? If not, I think an alternative way is still use error as tag, and attach the log that we can eliminate the message tag field, if this is necessary. WDYT?

i.e.
```Scala
private def setErrorTags(span: Span, message: => String): Unit = {
  span.setTag("error", true)
  span.log(Map("event" -> "error", "message" -> message).asJava)
}
```

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
vvraskin: I think that using the standard warn level would be fine. 

The snippet you provided is depicting the system error, so the error log level in that place is fine IMO.
 
As long as it doesn't disturb anyone else, I'm fine with leaving the `err` tag for the tracing. My main point was to avoid the log overflood with the messages produced by users, since it could significantly hamper the performance.

tz70s: Ah yes, it's a system error, thanks!

common/scala/src/main/scala/whisk/common/tracing/OpenTracingProvider.scala
sandeep-paliwal: Yeah I see that just adding log to trace is not that helpful in UI. I guess using tag has a better option then. Is there any use case to search for all failed invocations with a specific error? In that case error message as tag will also be useful compared to log().

Issue #3965, opened 2018-08-14
Provide command to update namespace key in wskadmin-next


chetanmeh:
Currently wskadmin (and next) provide way to add a namespace with specific auth key. 

     wskadmin user create -ns foo --auth uuid:key

In some cases we need a way to update the namespace key with a specific auth even if it already exists. One way would be to add a `-f`, `--force` flag to enable this override behaviour

    wskadmin user create -ns foo --auth uuid:key -f

rabbah:
Do you mean to set the uuid and key or just the key?

Can we add a revoke flag to change just the key? This is handy.

chetanmeh:
> Do you mean to set the uuid and key or just the key?

For our case both. Objective here is that credentials for some system namespaces are stored in separate store and deployment logic would ensure that one in db matches the one in credential store used by deployment

> Can we add a revoke flag to change just the key? This is handy.

That would be useful (cc @duynguyen)

PR #4005, opened 2018-09-06
Allow updating auth key for user with wskadmin-next

duynguyen:
This PR enhances the [wskadmin-next](https://github.com/apache/incubator-openwhisk/blob/master/tools/admin/README-NEXT.md) tooling with `--force` and `--revoke` flags that allow updating user auth key.

## Description
Currently wskadmin (and next) allow adding a namespace with or without specific auth key.
In some cases, it is handy to allow updating existing namespace with specific auth key, or regenerating new key.
There are 2 flags being added in this PR:
- `--revoke`, abbreviated `-r`: regenerate a new key for the user / namespace, UUID remains the same
- `--force`, abbreviated `-f`:
    - with `--auth uuid:key`: the new auth value is updated for the user
    - without `--auth`: an auth of new UUID and key is randomly generated

## Related issue and scope
- #3965 

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [x] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/tests/src/test/scala/whisk/core/database/UserCommandTests.scala
+++ b/tests/src/test/scala/whisk/core/database/UserCommandTests.scala
@@ -68,6 +68,23 @@ class UserCommandTests extends FlatSpec with WhiskAdminCliTestBase {
     resultOk("user", "get", subject) shouldBe generatedKey
   }
 
+  it should "force update an existing user" in {
+    val subject = newSubject()
+    val oldKey = resultOk("user", "create", "--force", subject)
+    resultOk("user", "get", subject) shouldBe oldKey
+
+    // Force update with provided auth uuid:key
+    val key = BasicAuthenticationAuthKey()
+    val newKey = resultOk("user", "create", "--auth", key.compact, "--force", subject)
+    resultOk("user", "get", subject) shouldBe newKey
+    newKey shouldBe key.compact
+
+    // Force update without auth, uuid:key is randomly generated
+    val generatedKey = resultOk("user", "create", "--force", subject)
+    generatedKey should not be newKey
+    generatedKey should not be oldKey
+  }
+
   it should "create a user or update an existing user with revoke flag" in {
     val subject = newSubject()
     val oldKey = resultOk("user", "create", "--revoke", subject)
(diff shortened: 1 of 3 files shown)


duynguyen:
@chetanmeh , @rabbah : this is for the discussion in #3965

tests/src/test/scala/whisk/core/database/UserCommandTests.scala
```diff
@@ -68,6 +68,35 @@ class UserCommandTests extends FlatSpec with WhiskAdminCliTestBase {
     resultOk("user", "get", subject) shouldBe generatedKey
   }
 
+  it should "force update an existing user" in {
+    val subject = newSubject()
+    val oldKey = resultOk("user", "create", "--force", subject)
```
rabbah: is this `--force` intended?

tests/src/test/scala/whisk/core/database/UserCommandTests.scala
```diff
@@ -68,6 +68,35 @@ class UserCommandTests extends FlatSpec with WhiskAdminCliTestBase {
...
+
+    // Force update without auth, uuid:key is randomly generated
+    val generatedKey = resultOk("user", "create", "--force", subject)
+    generatedKey should not be newKey
+    generatedKey should not be oldKey
+  }
+
+  it should "create a user or update an existing user with revoke flag" in {
+    val subject = newSubject()
+    val oldKey = resultOk("user", "create", "--revoke", subject)
```
rabbah: should there be a `--revoke` on the first command?

tests/src/test/scala/whisk/core/database/UserCommandTests.scala
duynguyen: The idea is to tell "if user is not existing, enabling force should behave the same as without it".
Not sure if it's necessary?

tests/src/test/scala/whisk/core/database/UserCommandTests.scala
duynguyen: Like above, the idea is to tell "if user is not existing, enabling revoke should behave the same as without it".

tools/admin/src/main/scala/whisk/core/database/UserCommand.scala
```diff
@@ -164,14 +172,23 @@ class UserCommand extends Subcommand("user") with WhiskCommand {
     authStore
       .get[ExtendedAuth](DocInfo(create.subject()))
       .flatMap { auth =>
+        val nsToUpdate = create.desiredNamespace(authKey).name
+        var newNS = auth.namespaces.filter(_.namespace.name != nsToUpdate)
```
chetanmeh: Avoid use of `var`. Instead name it as `existingNS` and then later just add to it and get `newNS`

tools/admin/src/main/scala/whisk/core/database/UserCommand.scala
```diff
@@ -164,14 +172,23 @@ class UserCommand extends Subcommand("user") with WhiskCommand {
...
         if (auth.isBlocked) {
           Future.successful(Left(IllegalState(CommandMessages.subjectBlocked)))
-        } else if (auth.namespaces.exists(_.namespace.name == create.desiredNamespace(authKey).name)) {
-          Future.successful(Left(IllegalState(CommandMessages.namespaceExists)))
-        } else {
-          val newNS = auth.namespaces + WhiskNamespace(create.desiredNamespace(authKey), authKey)
+        } else if (!auth.namespaces.exists(_.namespace.name == nsToUpdate) || create.force.isSupplied) {
+          newNS = newNS.+(WhiskNamespace(create.desiredNamespace(authKey), authKey))
```
chetanmeh: Instead of `.+(` just simply use `+` like `newNS + WhiskNamespace(create.desiredNamespace(newAuthKey), newAuthKey)`

tools/admin/src/main/scala/whisk/core/database/UserCommand.scala
```diff
@@ -164,14 +172,23 @@ class UserCommand extends Subcommand("user") with WhiskCommand {
...
           val newAuth = WhiskAuth(auth.subject, newNS).revision[WhiskAuth](auth.rev)
           authStore.put(newAuth).map(_ => Right(authKey.compact))
+        } else if (create.revoke.isSupplied) {
+          val updatedAuthKey = auth.namespaces.find(_.namespace.name == nsToUpdate).get.authkey
```
chetanmeh: What if namespace to revoke does not exist? 

Okie it would then just add a new namespace entry with new AuthKey i.e. enter the previous flow in if. Should we return an error if namespace to revoke does not exist?

duynguyen: I don't have a strong opinion on this, but the former sounds a bit simpler and consistent with `-f` behavior. wdyt?

rabbah:
https://github.com/apache/incubator-openwhisk/blob/master/tools/admin/README.md

https://github.com/apache/incubator-openwhisk/blob/master/tools/admin/README-NEXT.md

tests/src/test/scala/whisk/core/database/UserCommandTests.scala
rabbah: to be clear: use create X fails if  X already exists, but --force allows it to succeed and updates the uuid and key. is my understanding correct? if so, yes i'm ok with that.

chetanmeh:
@dubee  The dcos update is [part of this PR][1]. `README.md` is not being updated as the new feature is only in scala based implementation and hence documented in READEME-NEXT.md. Please have a look

[1]: https://github.com/apache/incubator-openwhisk/pull/4005/files#diff-d1651969e21b0b9c401fac9b00d2833d

tests/src/test/scala/whisk/core/database/UserCommandTests.scala
duynguyen: @rabbah yes that's correct

duynguyen:
@dubee documentation update was done, as justified by @chetanmeh . Could you please review?

PR #4010, opened 2018-09-08
CORS headers work

### LINKED ISSUE #4009: enhancement: add user-agent to allowed CORS headers, and extend allow-methods
Right now, we limit accepted headers via CORS to 

```json
{
  "Access-Control-Allow-Headers": [
    "Authorization, Content-Type"
  ]
}
```

to allow for browsers to communicate directly with openwhisk API endpoints, it is helpful to add User-Agent to the list of allowed headers. many popular request libraries tack on default user agents, including our own openwhisk npm.

we also will need to expand the default Access-Control-Allow-Methods, which, by default, limits clients to GET operations.

starpit:
## Description

This PR updates the REST and Web Action endpoints to change the Allowed-Headers response to `"*"`, and to add an Allowed-Methods header. It also updates the related tests to check for conformance.

Fixes #4009

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#4009)

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [x] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
@@ -19,6 +19,7 @@ package whisk.core.controller
 
 import akka.actor.ActorSystem
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+import akka.http.scaladsl.model.HttpMethods.{DELETE, GET, HEAD, POST, PUT}
 import akka.http.scaladsl.model.StatusCodes._
 import akka.http.scaladsl.model.Uri
 import akka.http.scaladsl.model.headers._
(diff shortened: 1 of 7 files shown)


starpit:
i believe that the trade-off we need to consider is 

Positive: this change allows for purely serverless clients

Negative: i think there is only one exposed attack surface. if the client exposes the openwhisk AUTH as a javascript global variable, and the HTML origin CORS allows it to make outbound requests, then untrusted content can pick it up and begin invoking openwhisk (or any other endpoint) with that key. with the REST API CORS as it stands, at least this poorly designed client would be protected from third party calls (from the client) against openwhisk (though not from other third party endpoints).

i think this is a pretty fine trade-off; to expound a bit: a well-designed client would keep the AUTH in either localStorage, sessionStorage, or cookies -- all three of which are protected by SAME-ORIGIN, independent of endpoint CORS settings.

furthermore, it is worth noting that one can work around the REST API limitation by using a web action to proxy invoke and create requests (web actions offer better CORS support, though still with improper header limitations -- also fixed by this PR); however, this web action proxy will need to be multi-tenant; so, by forcing client developers to use a web action proxy, i claim that there is a large net negative in security.

markusthoemmes:
Thanks a lot for clarifying @starpit! SGTM

starpit:
it does expose a rather fascinating aspect of serverless that i hadn't much considered till now: a conventional client-with-server-proxy architecture will certainly require some level of multi-tenancy; either shared in-process (for e.g. a java/scala-based endpoint controller), or at least at the OS level (e.g. for run-and-done PHP-style controllers).

whereas, with this CORS change, we can enable a path from client to API endpoints that needn't have a multi-tenant intermediary -- especially so if one stands up a private openwhisk.

PR #4038, opened 2018-09-24
Don't lookup config on every container-creation.

cbickel:
On profiling the invoker I saw, that the lookup of the config blocks the invoker-threads on the beginning of a burst.
The reason is, that the config is read on every container-creation. It is enough to read this configuration once.
I also verified, that the threads are not blocked after the change.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


chetanmeh:
Good find @cbickel ! So with future changes we must be careful that we do not perform config lookups on critical paths

PR #4040, opened 2018-09-24
Prevent Artifact Store Polling for Large Activation Records

dubee:
## Description

Differentiates between KAFKA_MESSAGE_MAX_BYTES and whisk.activation.payload.max. KAFKA_MESSAGE_MAX_BYTES should be larger than whisk.activation.payload.max to account for data being added to the activation record when it is serialized by the Kafka producer. Right now the data added during serialization is not accounted for, so truncation of larger responses results in the truncated response not being able to be sent through the Kafka completed topic.

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


dubee:
@csantanapr, @mdeuser, @jasonpet, need a review one this. 

@mdeuser, as requested on the dev list, I have split the DB polling removal out from the action response size defect. Added some comments here to address your review comments in the related PR.

ansible/group_vars/all
```diff
@@ -138,6 +140,9 @@ kafka:
   offsetsTopicReplicationFactor: "{{ kafka_offsetsTopicReplicationFactor | default(kafka_replicationFactor) | default((groups['kafkas']|length)|int) }}"
   # adapt this param for production deployments depending on the number of kafka consumers
   networkThreads: "{{ kafka_network_threads | default(3) }}"
+  # whisk.activation.payload.max + 5% of that value to account for data added by Kafka producer serialization
+  payload:
+    max: "{{ limit_activation_payload + (limit_activation_payload * 0.05) | int | default() }}"
```
rabbah: i think the 5% is not adequate in general. We can actually determine a fixed overhead because (all?) the fields have limits on size.

common/scala/src/main/resources/application.conf
```diff
@@ -172,7 +177,7 @@ whisk {
 
     activation {
         payload {
-            max = 1048576 // 5 m not possible because cross-referenced to kafka configurations
+            max = 1048576 // value is cross-referenced to kafka configurations for whisk.kafka.payload.max
```
rabbah: can we avoid having this value defined twice? here and as `limit_activation_payload: 1048576 # Maximum action invocation payload and response size` in the environment?

dubee:
@rabbah, I like the idea, but I don't think it is possible to add values in `application.conf`.

dubee:
There are also two tests that already exists in the ActionLimitsTests suite. `successfully invoke an action with a payload close to the limit` and `succeed but truncate result, if result exceeds its limit`. However, these two tests currently are impacted by the defect and take at least 15 seconds to complete.

dubee:
PG3 2851 🔵

tests/src/test/scala/whisk/core/limits/ActionLimitsTests.scala
```diff
@@ -248,7 +248,9 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
 
     // Needs some bytes grace since activation message is not only the payload.
     val args = Map("p" -> ("a" * (allowedSize - 750).toInt).toJson)
+    val start = Instant.now
```
rabbah: this change here looks unrelated to the rest of the PR.

dubee: This test currently takes 15 seconds to invoke an action as serialization size is not account for even when the invoker truncates the action response.

```
whisk.core.limits.ActionLimitsTests > Action limits should successfully invoke an action with a payload close to the limit (5 MB) STANDARD_ERROR
    org.scalatest.exceptions.TestFailedException: 19204 was not less than 15000
```
```
whisk.core.limits.ActionLimitsTests > Action limits should succeed but truncate result, if result exceeds its limit (blocking: true) STANDARD_ERROR
    org.scalatest.exceptions.TestFailedException: 15754 was not less than 15000
```

dubee: @rabbah, see above.

dubee:
@csantanapr, want to merge?

PR #4046, opened 2018-09-27
update Access-Control-Allow-Headers to enumerate rather than wildcard

### LINKED ISSUE #4045: Access-Control-Allow-Headers CORS bug
safari and firefox don't (yet) support wildcards on Access-Control-Allow-Headers, despite this being adopted by whatwg in 2016 😡 :

firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1309358
webkit: https://bugs.webkit.org/show_bug.cgi?id=165508

chromium's tracker for this was closed last december as Fixed. and i can testify that chrome works well with Access-Control-Allow-Headers: *

therefore, we need to change the Access-Control-Allow-Headers behavior away from a wildcard to a enumerated list as per https://stackoverflow.com/a/33704645

starpit:
As summarized in #4045, the previous #4010 PR, which changed Access-Control-Allow-Headers from a 2-element list to a wildcard was misguided. Despite whatwg introducing wildcard support in 2016, only Chrome as of yet supports this feature. So, for now, we need to back off to a specifically enumerated list of allowed headers. This PR reverts the wildcard change, but expands the list from the initial 2 to include a few more common headers, e.g. User-Agent.

Fixes #4045

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#4045)

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [x] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
+++ b/core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
@@ -19,10 +19,8 @@ package whisk.core.controller
 
 import akka.actor.ActorSystem
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
-import akka.http.scaladsl.model.HttpMethods.{DELETE, GET, HEAD, POST, PUT}
 import akka.http.scaladsl.model.StatusCodes._
 import akka.http.scaladsl.model.Uri
-import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.server.directives.AuthenticationDirective
 import akka.http.scaladsl.server.{Directives, Route}
 import akka.stream.ActorMaterializer
@@ -149,11 +147,7 @@ protected[controller] object RestApiCommons {
  * A trait for wrapping routes with headers to include in response.
  * Useful for CORS.
  */
-protected[controller] trait RespondWithHeaders extends Directives {
-  val allowOrigin = `Access-Control-Allow-Origin`.*
-  val allowHeaders = `Access-Control-Allow-Headers`("*")
-  val allowMethods =
-    `Access-Control-Allow-Methods`(GET, DELETE, POST, PUT, HEAD)
+protected[controller] trait RespondWithHeaders extends Directives with CorsSettings.RestAPIs {
   val sendCorsHeaders = respondWithHeaders(allowOrigin, allowHeaders, allowMethods)
 }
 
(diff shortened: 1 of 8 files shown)


core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala
```diff
@@ -151,7 +151,13 @@ protected[controller] object RestApiCommons {
...
 protected[controller] trait RespondWithHeaders extends Directives {
   val allowOrigin = `Access-Control-Allow-Origin`.*
-  val allowHeaders = `Access-Control-Allow-Headers`("*")
+  val allowHeaders = `Access-Control-Allow-Headers`(
+    "Authorization",
+    "Origin",
+    "X-Requested-With",
+    "Content-Type",
+    "Accept",
+    "User-Agent")
```
markusthoemmes: Should we make this list configurable?

starpit: probably the ultimate right course of action is to make both fields (allow-origin, allow-headers) configurable.

the reason i've put in this short-term fix is that my previous PR probably broke some use cases; whereas previously web actions would work from any browser (assuming only those two previously allowed headers were passed), now web actions will only work from chrome (but will support *any* headers)

with this PR, we restore functionality on all browsers for web actions, with a somewhat expanded list of allow headers)

i would suggest that we do the right thing in a subsequent PR, but this PR i feel fills an important a short-term regression fix until we can get the right PR in place?

markusthoemmes: SGTM

markusthoemmes: Can we at least define the list of headers in a centralized place so we don't end up copying them around?

starpit: sounds good. i'll work on this in a bit.

starpit:
i have updated the PR to introduce a new file CorsSettings, and cleaned up RestAPIs and WebActions to pull the constants from this file.

in a subsequent PR, perhaps we can allow CorsSettings to be populated from the ansible/whatever level of deploy-time configuration.

starpit:
further refined to use traits

starpit:
ping

PR #4058, opened 2018-10-08
Add protect feature to avoid update or delete actions by mistake

ningyougang:
As more and more production system uses openwhisk,
Users will need some features to protect their service safe from mistake.
If we have this feature, user can protect their actions from deletion by mistake.

When users create action use like below(add annotation "lock":true)
```
curl  -X PUT -H "Content-type: application/json"  --user 23bc46b1-71f6-4ed5-8c54-816aa4f8c502:123zO3xZCLrMN6v2BKK1dXYFpXlPkccOFqm12CdAsMgRU4VrNZ9lyGVCGuMDGIwP 
-d '{"namespace":"guest","name":"hello","annotations":[{"key":"lock","value":true}],"exec":{"kind":"nodejs:default","code":"// Licensed to the Apache Software Foundation (ASF) under one or more contributor\r\n// license agreements; and to You under the Apache License, Version 2.0.\r\n\r\n/**\r\n * Hello, world.\r\n */\r\nfunction main() {\r\n return {\"payload\":\"Hello world\"};\r\n}\r\n"}}' 
'http://xxx.xxx.xxx.xxx:port/api/v1/namespaces/guest/actions/hello?'
```
The lock field will be added in couchdb's  annotation, like below
```
...
  "annotations": [
    {
      "key": "exec",
      "value": "nodejs:6"
    },
    {
      "key": "lock",
      "value": true
    }
...
```
So this action can't be deleted/updated until the protection is updated to `false`

## Related issue and scope

- [x] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
@@ -367,30 +367,8 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
 
   val defaultPermissions = "rwxr-x"
 
-  // notes on users, just have 2 type users,
-  // 1. the action's owner
-  // 2. the user (not the owner) who used the shared action directly(e.g. get, invoke), we call it "the shared user"
-  //
-  // Notes on permission control
-  // 1. the owner has read(or download) permission on any situation, but for the shared user,
-  //    in spite of has read permission on any situation, but can set it undownloadable or downloadable
-  // 2. the shared user can't update/delete the action on any situation.
-  // 3. the owner's permission can affect the shared user's permission, e.g
-  //    if the owner is not given execute permission, the shared user can't have execute permission as well.
-  //
-  // Notes on permission values, include below permission value
-  //  1. permission code:rwxr-x: owner:read(yes)/write(yes)/execute(yes)|the shared action's user:download(yes)/write(no)/execute(yes), this is default
-  //  2. permission code:rwxr--: owner:read(yes)/write(yes)/execute(yes)|the shared action's user:download(yes)/write(no)/execute(no)
-  //  3. permission code:r-xr-x: owner:read(yes)/write(no)/execute(yes)|the shared action's user:download(yes)/write(no)/execute(yes)
-  //  4. permission code:r-xr--: owner:read(yes)/write(no)/execute(yes)|the shared action's user:download(yes)/write(no)/execute(no)
-  //  5. permission code:r--r--: owner:read(yes)/write(no)/execute(no)|the shared action's user:download(yes)/write(no)/execute(no)
-  //  6. permission code:rw-r--: owner:read(yes)/write(yes)/execute(no)|the shared action's user:download(yes)/write(no)/execute(no)
-  //  7. permission code:rwx--x: owner:read(yes)/write(yes)/execute(yes)|the shared action's user:download(no)/write(no)/execute(yes)
-  //  8. permission code:rwx---: owner:read(yes)/write(yes)/execute(yes)|the shared action's user:download(no)/write(no)/execute(no)
-  //  9. permission code:r-x--x: owner:read(yes)/write(no)/execute(yes)|the shared action's user:download(no)/write(no)/execute(yes)
-  // 10. permission code:r-x---: owner:read(yes)/write(no)/execute(yes)|the shared action's user:download(no)/write(no)/execute(no)
-  // 11. permission code:r-----: owner:read(yes)/write(no)/execute(no)|the shared action's user:download(no)/write(no)/execute(no)
-  // 12. permission code:rw----: owner:read(yes)/write(yes)/execute(no)|the shared action's user:download(no)/write(no)/execute(no)
+  // For more information about permission, please refer to the online manual at
+  // https://github.com/apache/openwhisk/blob/master/docs/actions.md#action-permission-management
   val permissionList = List(
     defaultPermissions,
     "rwxr--",
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
@@ -76,6 +76,16 @@ case class WhiskActionPut(exec: Option[Exec] = None,
       case _ => this
     } getOrElse this
   }
+
+  protected[core] def getPermissions(): Option[String] = {
+    annotations match {
+      case Some(value) =>
+        value
+          .get(WhiskAction.permissionsFieldName)
+          .map(value => value.convertTo[String])
+      case None => None
+    }
+  }
 }
 
 abstract class WhiskActionLike(override val name: EntityName) extends WhiskEntity(name, "action") {
@@ -351,6 +361,49 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
 
   val execFieldName = "exec"
   val requireWhiskAuthHeader = "x-require-whisk-auth"
+
+  // annotation permission key name
+  val permissionsFieldName = "permissions"
+
+  val defaultPermissions = "rwxr-x"
+
+  // notes on users, just have 2 type users,
+  // 1. the action's owner
+  // 2. the user (not the owner) who used the shared action directly(e.g. get, invoke), we call it "the shared user"
+  //
+  // Notes on permission control
+  // 1. the owner has read(or download) permission on any situation, but for the shared user,
+  //    in spite of has read permission on any situation, but can set it undownloadable or downloadable
+  // 2. the shared user can't update/delete the action on any situation.
+  // 3. the owner's permission can affect the shared user's permission, e.g
+  //    if the owner is not given execute permission, the shared user can't have execute permission as well.
+  //
+  // Notes on permission values, include below permission value
+  //  1. permission code:rwxr-x: owner:read(yes)/write(yes)/execute(yes)|the shared action's user:download(yes)/write(no)/execute(yes), this is default
+  //  2. permission code:rwxr--: owner:read(yes)/write(yes)/execute(yes)|the shared action's user:download(yes)/write(no)/execute(no)
+  //  3. permission code:r-xr-x: owner:read(yes)/write(no)/execute(yes)|the shared action's user:download(yes)/write(no)/execute(yes)
+  //  4. permission code:r-xr--: owner:read(yes)/write(no)/execute(yes)|the shared action's user:download(yes)/write(no)/execute(no)
+  //  5. permission code:r--r--: owner:read(yes)/write(no)/execute(no)|the shared action's user:download(yes)/write(no)/execute(no)
+  //  6. permission code:rw-r--: owner:read(yes)/write(yes)/execute(no)|the shared action's user:download(yes)/write(no)/execute(no)
+  //  7. permission code:rwx--x: owner:read(yes)/write(yes)/execute(yes)|the shared action's user:download(no)/write(no)/execute(yes)
+  //  8. permission code:rwx---: owner:read(yes)/write(yes)/execute(yes)|the shared action's user:download(no)/write(no)/execute(no)
+  //  9. permission code:r-x--x: owner:read(yes)/write(no)/execute(yes)|the shared action's user:download(no)/write(no)/execute(yes)
+  // 10. permission code:r-x---: owner:read(yes)/write(no)/execute(yes)|the shared action's user:download(no)/write(no)/execute(no)
+  // 11. permission code:r-----: owner:read(yes)/write(no)/execute(no)|the shared action's user:download(no)/write(no)/execute(no)
+  // 12. permission code:rw----: owner:read(yes)/write(yes)/execute(no)|the shared action's user:download(no)/write(no)/execute(no)
+  val permissionList = List(
+    defaultPermissions,
+    "rwxr--",
+    "r-xr-x",
+    "r-xr--",
+    "r--r--",
+    "rw-r--",
+    "rwx--x",
+    "rwx---",
+    "r-x--x",
+    "r-x---",
+    "r-----",
+    "rw----")
 
   override val collectionName = "actions"
   override val cacheEnabled = true
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
@@ -218,9 +218,22 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with 
 
         onComplete(checkAdditionalPrivileges) {
           case Success(_) =>
-            putEntity(WhiskAction, entityStore, entityName.toDocId, overwrite, update(user, request) _, () => {
-              make(user, entityName, request)
-            })
+            val operation = if (overwrite) "update" else "create"
+            onComplete(
+              entitlementProvider
+                .checkActionPermissions(
+                  operation,
+                  user,
+                  entityStore,
+                  entityName,
+                  WhiskAction.get,
+                  content.getPermissions())) {
+              case Success(_) =>
+                putEntity(WhiskAction, entityStore, entityName.toDocId, overwrite, update(user, request) _, () => {
+                  make(user, entityName, request)
+                })
+              case Failure(f) => super.handleEntitlementFailure(f)
+            }
           case Failure(f) =>
             super.handleEntitlementFailure(f)
         }
@@ -241,37 +254,43 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with 
    */
   override def activate(user: Identity, entityName: FullyQualifiedEntityName, env: Option[Parameters])(
     implicit transid: TransactionId) = {
-    parameter(
-      'blocking ? false,
-      'result ? false,
-      'timeout.as[FiniteDuration] ? controllerActivationConfig.maxWaitForBlockingActivation) {
-      (blocking, result, waitOverride) =>
-        entity(as[Option[JsObject]]) { payload =>
-          getEntity(WhiskActionMetaData.resolveActionAndMergeParameters(entityStore, entityName), Some {
-            act: WhiskActionMetaData =>
-              // resolve the action --- special case for sequences that may contain components with '_' as default package
-              val action = act.resolve(user.namespace)
-              onComplete(entitleReferencedEntitiesMetaData(user, Privilege.ACTIVATE, Some(action.exec))) {
-                case Success(_) =>
-                  val actionWithMergedParams = env.map(action.inherit(_)) getOrElse action
+    onComplete(
+      entitlementProvider
+        .checkActionPermissions("invoke", user, entityStore, entityName, WhiskAction.get)) {
+      case Success(_) =>
+        parameter(
+          'blocking ? false,
+          'result ? false,
+          'timeout.as[FiniteDuration] ? controllerActivationConfig.maxWaitForBlockingActivation) {
+          (blocking, result, waitOverride) =>
+            entity(as[Option[JsObject]]) { payload =>
+              getEntity(WhiskActionMetaData.resolveActionAndMergeParameters(entityStore, entityName), Some {
+                act: WhiskActionMetaData =>
+                  // resolve the action --- special case for sequences that may contain components with '_' as default package
+                  val action = act.resolve(user.namespace)
+                  onComplete(entitleReferencedEntitiesMetaData(user, Privilege.ACTIVATE, Some(action.exec))) {
+                    case Success(_) =>
+                      val actionWithMergedParams = env.map(action.inherit(_)) getOrElse action
 
-                  // incoming parameters may not override final parameters (i.e., parameters with already defined values)
-                  // on an action once its parameters are resolved across package and binding
-                  val allowInvoke = payload
-                    .map(_.fields.keySet.forall(key => !actionWithMergedParams.immutableParameters.contains(key)))
-                    .getOrElse(true)
+                      // incoming parameters may not override final parameters (i.e., parameters with already defined values)
+                      // on an action once its parameters are resolved across package and binding
+                      val allowInvoke = payload
+                        .map(_.fields.keySet.forall(key => !actionWithMergedParams.immutableParameters.contains(key)))
+                        .getOrElse(true)
 
-                  if (allowInvoke) {
-                    doInvoke(user, actionWithMergedParams, payload, blocking, waitOverride, result)
-                  } else {
-                    terminate(BadRequest, Messages.parametersNotAllowed)
+                      if (allowInvoke) {
+                        doInvoke(user, actionWithMergedParams, payload, blocking, waitOverride, result)
+                      } else {
+                        terminate(BadRequest, Messages.parametersNotAllowed)
+                      }
+
+                    case Failure(f) =>
+                      super.handleEntitlementFailure(f)
                   }
-
-                case Failure(f) =>
-                  super.handleEntitlementFailure(f)
-              }
-          })
+              })
+            }
         }
+      case Failure(f) => super.handleEntitlementFailure(f)
     }
   }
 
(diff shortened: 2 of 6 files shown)


cbickel (changes requested):
Hi,

thanks a lot for your contribution. This PR definitely makes sense.

You are adding the `protection`-parameter as URL parameter. What's the reason for doing it like this?
Today: Everything that is a property of the action is within `WhiskActionPut`. The `overwrite` parameter only says, that this request should overwrite the action if it already exists.
IMHO the `protection` parameter is a property of the action itself and not of that request.

And a second thought that comes into my mind: If we protect the action, should it only be protected against deletion? Or would it make sense to protect it against updates as well? (The trick here would be to allow the update to unlock the action again).

I don't know what your thoughts on these two issues are. If your thoughts are completely different mybe it makes sense to ask the dev-list, if someone has a strong opinion on these two questions on the dev-list.

rabbah:
This is a good start but we should discuss and settle the design first. For example my preference is for Unix style permissions on the action (or asset more generally).

ningyougang:
@cbickel @rabbah ,thanks for your quick reply.

Actually, i opened a pr first here: https://github.com/apache/incubator-openwhisk-cli/pull/372
Solve the update/delete mistake on CLI side, after think deeply, i found, it is better to solve it in backend layer.(So don't merge that pull/372 temporarily)

As you give two suggestions
* You are adding the protection-parameter as URL parameter
Actually, i passed the `protection` as URL parameter` the same as `overwrite`,
If don't pass as URL parameter, has any other good method, how about pass as a Http Header, for example：`OW_PROTECTION`

* Or would it make sense to protect it against updates as well?
I agree, i think this also, yestoday, i push it to here, just check whether upstream guys agree this ` protect delete/update mistake feature`

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -412,7 +432,9 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with
       limits,
       content.version getOrElse SemVer(),
       content.publish getOrElse false,
-      (content.annotations getOrElse Parameters()) ++ execAnnotation(exec))
+      (content.annotations getOrElse Parameters()) ++ execAnnotation(exec) ++ Parameters(
+        "protection",
+        if (protection) "true" else "false"))
```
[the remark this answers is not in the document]
ningyougang: @cbickel ,it seems if save it to annotations, the key/value supports `string` only,
if want to save `protection` as Boolean, should modify `WhiskAction class` to add a `protection` field.
It seems the amount of code changes is relatively large, so balanced, i added the   `protection`  to `annotations`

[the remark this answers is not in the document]
ningyougang: @cbickel ,if i added a protection filed to class WhiskAction's first level, after running controller's main method, report below strange error
```
Exception in thread "main" whisk.core.database.DocumentTypeMismatchException: document type class whisk.core.entity.WhiskTrigger did not match expected type class whisk.core.entity.WhiskAction.
	at whisk.core.database.StoreUtils$.deserialize(StoreUtils.scala:64)
	at whisk.core.database.CouchDbRestStore$$anonfun$19.apply(CouchDbRestStore.scala:235)
	at whisk.core.database.CouchDbRestStore$$anonfun$19.apply(CouchDbRestStore.scala:231)
```
It seems add protection to WhiskAction's annotations may be a good choose.

core/controller/src/main/scala/whisk/core/controller/Actions.scala
```diff
@@ -307,7 +313,21 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with
    * - 500 Internal Server Error
    */
   override def remove(user: Identity, entityName: FullyQualifiedEntityName)(implicit transid: TransactionId) = {
-    deleteEntity(WhiskAction, entityStore, entityName.toDocId, (a: WhiskAction) => Future.successful({}))
+    getEntity(WhiskAction.get(entityStore, entityName.toDocId), Some {
+      action: WhiskAction =>
+        action.annotations.get("protection") match {
+          case Some(value) => {
```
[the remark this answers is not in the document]
ningyougang: @cbickel 
Already modified

core/controller/src/main/scala/whisk/core/controller/Actions.scala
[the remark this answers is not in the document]
ningyougang: I already modify it to Boolean type.

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
```diff
@@ -523,16 +524,32 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with
...
+      //keep original value not changed
+      WhiskAction(
+        action.namespace,
+        action.name,
+        exec,
+        parameters,
+        limits,
+        content.version getOrElse action.version.upPatch,
+        content.publish getOrElse action.publish,
+        (content.annotations getOrElse action.annotations) ++ execAnnotation(exec))
```
dubee: Instead of having the if statement and duplicate code, can you just keep the original `WhiskAction` block and add something like the below to the annotations?

```
++ content.unlock.map(u => Parameters(WhiskAction.lockFieldName, JsBoolean(false))).getOrElse(Parameters())
```

ningyougang: @dubee ,thanks, already fixed.

rabbah:
Sorry for taking so long to get through this --- taking a look now it's a good first step but i think it points to a more general permission set on an action: this is an write permission for the namespace (group in unix terminology), and one can image an execute permission as well next (disables an action from executing for a namespace).

So one of my thoughts is whether to factor the permission more generally to allow for generalization later... or leave it as is and change it later.

Thoughts?

ningyougang:
@rabbah 
Regarding `factor the permission more generally`
it is a good suggestion, let's leave it as is and change it later.

At the same time, we can wait other upsteam guys to give more suggestions,
if we agree on the idea, go on to do this.

mhenke1:
@ningyougang 
Sorry for joining late in this discussion. 
In my last year I have implemented the SPIs for plugging in alternative Authentication and Entitlement providers in Openwhisk. My strong suggestion would be to place any additional code checking for entitlement (in this case: the entitlement to delete entities) into the existing entitlement SPI implementation.
Alternative entitlement implementations (like the one I have implemented for IBM cloud functions) might prevent accidental deletion in other ways like configuring user or user group specific polices 
in a policy system. They have no need of  additional protection. 
Of course the default entitlement provider would benefit from a mechanism like the one you proposed. By moving the functionality in the Entitelemnt-SPI the functionality could be added to the default path, but would not interfere with alternative entitlement implementations.

mhenke1:
@rabbah I would love to. But it uses proprietary code to talk to proprietary endpoints for checking permissions. So it would be no use for any other party. I am contemplating provide a kind of stripped down version with mock up code where proprietary code is used. But I am not sure whether the left-overs would be of enough value. And as said in another place I plan to write a explaining text (blog ?) when time allows.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
```diff
@@ -55,7 +55,8 @@ case class WhiskActionPut(exec: Option[Exec] = None,
                           limits: Option[ActionLimitsOption] = None,
                           version: Option[SemVer] = None,
                           publish: Option[Boolean] = None,
-                          annotations: Option[Parameters] = None) {
+                          annotations: Option[Parameters] = None,
+                          unlock: Option[Boolean] = None) {
```
rabbah, line 59: concretely my suggestion is to change `unlock: Option[Boolean]` to `permissions: Option[Int]` and use bit operations to toggle permissions. The first set of permissions to respect is whether the resource is locked locked or unlocked. This can be abstracted with utility methods.

rabbah (commented):
I think this is close. Sorry that it was neglected for a long time and thank you for bringing it to the dev list. I have a suggestion to how to structure the permissions so that we can extend this in the future. I do not have reservations about the general approach and we should def. get this enhancement in.

docs/actions.md
```diff
@@ -598,6 +598,28 @@ You can clean up by deleting actions that you do not want to use.
   actions
   /guest/mySequence                private sequence
   ```
+## Protect action updated or deleted by mistake
```
rabbah, line 601: I will suggest some updates to this text once the API change and implementation are finalized.

ningyougang:
Hi, guys, i push a commit here in this patch: https://github.com/apache/openwhisk/pull/4058/commits/d77a80fd0348d05f3733f95dbc0e016429e96753

Just check whether my direction(permission unix sytle) is right.
After check in my local, i found that, we can use some parts of unix-sytle. e.g.
for one action, just have 2 type users.
* the action's owner
* if the action is shared,  the user (not the owner) who used the action directly(eg. get, invoke)
As we all already know, unix's permission controller has `3` user types, this is the different point.

Another different  point the permission values is restricted
* for one action, anyway, the user(not the owner) can't write the action, just the owner has the write permission. 
* the owner's permission can affect the permission as well, e.g. if the owner is not given `read` permission, i think the user(not the owner) can't have `read` permission as well.  if the owner is not give `execute` permission, the  user(not the owner) can't have `execute` permission as well.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
ningyougang: I already changed to `unix permission style` to `protect action`

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
```diff
@@ -350,6 +360,31 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
...
+  //    if the owner is not given execute permission, the user(not the owner) can't have execute permission as well.
+  //
+  // Notes on permission values, include below permission value
+  // 1. permission code:rwxr-x: owner:read(yes)/write(yes)/execute(yes)|the shared action's user:read(yes)/write(no)/execute(yes), this is default
+  // 2. permission code:rwxr--: owner:read(yes)/write(yes)/execute(yes)|the shared action's user:read(yes)/write(no)/execute(no)
+  // 3. permission code:r-xr-x: owner:read(yes)/write(no)/execute(yes)|the shared action's user:read(yes)/write(no)/execute(yes)
+  // 4. permission code:r-xr--: owner:read(yes)/write(no)/execute(yes)|the shared action's user:read(yes)/write(no)/execute(no)
+  // 5. permission code:r--r--: owner:read(yes)/write(no)/execute(no)|the shared action's user:read(yes)/write(no)/execute(no)
+  // 6. permission code:rw-r--: owner:read(yes)/write(yes)/execute(no)|the shared action's user:read(yes)/write(no)/execute(no)
+  val permissionList = List(defaultPermissions, "rwxr--", "r-xr-x", "r-xr--", "r--r--", "rw-r--")
```
jiangpengcheng, line 386: does `r` means get codes here?

ningyougang: `r` means `readable`

docs/actions.md
ningyougang: Updated accordingly

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
ningyougang: for the shared user, the `r` means `download the code`

core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/Entitlement.scala
```diff
@@ -231,6 +234,157 @@ protected[core] abstract class EntitlementProvider(
...
+              } else {
+                Future.successful(())
+              }
+          }
+        }
+        .recoverWith {
+          case t: RejectRequest =>
+            Future.failed(t)
+          case _ =>
+            Future.successful(())
```
ningyougang, line 307: In order to make some test cases successfully .e.g
```
 ./gradlew :tests:test --tests="org.apache.openwhisk.core.controller.test.PackageActionsApiTests"
 ./gradlew :tests:test --tests="org.apache.openwhisk.core.controller.test.ActionsApiTests"
```
Need to add above recoverWith logic,  concrete content is
* if the excpetion is `RejectRequest`, return `Future.failed(t)`  

   make normal logic passed
* For some excpetion which generated by test above cases, should return `Future.successful(())`
  
   Make above tests cases passed, e.g. https://github.com/apache/openwhisk/blob/master/tests/src/test/scala/org/apache/openwhisk/core/controller/test/PackageActionsApiTests.scala#L130
   here, it will pass a package paramter to `get(entityStore, entityName.toDocId, DocRevision.empty, true)`'s second param, actually, this method needs `action`, so will throw NoDocument exception, in order to make the test case passed, need to make it return `Future.successful(())`

ningyougang:
Already rebased and solved relative test cases error

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
```diff
@@ -351,6 +361,49 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
   val execFieldName = "exec"
   val requireWhiskAuthHeader = "x-require-whisk-auth"
 
+  // annotation permission key name
+  val permissionsFieldName = "permissions"
```
style95, line 365: I am not quite sure it is a good idea to use annotations for permission management.
I feel like we need a new protocol.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
```diff
@@ -351,6 +361,49 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
...
+  //  9. permission code:r-x--x: owner:read(yes)/write(no)/execute(yes)|the shared action's user:download(no)/write(no)/execute(yes)
+  // 10. permission code:r-x---: owner:read(yes)/write(no)/execute(yes)|the shared action's user:download(no)/write(no)/execute(no)
+  // 11. permission code:r-----: owner:read(yes)/write(no)/execute(no)|the shared action's user:download(no)/write(no)/execute(no)
+  // 12. permission code:rw----: owner:read(yes)/write(yes)/execute(no)|the shared action's user:download(no)/write(no)/execute(no)
+  val permissionList = List(
+    defaultPermissions,
+    "rwxr--",
+    "r-xr-x",
+    "r-xr--",
+    "r--r--",
```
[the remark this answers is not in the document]
ningyougang: Yes, there has an case that the owner doesn't have permission to update or execute the action, e. g. the annotation of permission for that action is:  `r--r--`
in this case, if the owner wants to update the action codes,  need to modify the action's permissions annotation to executeable, e.g. `wsk -i action update $action --annotation permissions rw-r--`, then, user can update their action now.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
ningyougang: yes, i also feel it is not a good idea to use annotation for permission management.
Currently, have no better idea for it.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
```diff
@@ -351,6 +361,49 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
   val execFieldName = "exec"
   val requireWhiskAuthHeader = "x-require-whisk-auth"
 
+  // annotation permission key name
+  val permissionsFieldName = "permissions"
+
+  val defaultPermissions = "rwxr-x"
+
+  // notes on users, just have 2 type users,
```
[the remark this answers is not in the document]
ningyougang: Updated accordingly.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
style95: I think we can store permission information in DB and fetch it along with `Identity` information.
Some proper tools to manage permission are required and the same change should be applied in CLI(`wsk` or at least `wskadmin`) as well.

core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/Entitlement.scala
```diff
@@ -231,6 +234,157 @@ protected[core] abstract class EntitlementProvider(
...
+              s"give error permission code: ${value}, available permission is in ${WhiskAction.permissionList}"
+            Future.failed(RejectRequest(Forbidden, Some(ErrorResponse(errorInfo, transid))))
+          }
+        }
+        .getOrElse(Future.successful(()))
+    } else if (operation == "update") {
+      get(entityStore, entityName.toDocId, DocRevision.empty, true)
+        .flatMap { whiskAction =>
+          val currentPermissions = whiskAction.annotations
+            .get(WhiskAction.permissionsFieldName)
```
style95, line 270: We can depend on permission data in DB rather than relying on action fields.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
jiangpengcheng: I think such permission information is applied to actions instead of `Identity`

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
```diff
@@ -351,6 +361,49 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
...
+  //  5. permission code:r--r--: owner:read(yes)/write(no)/execute(no)|the shared action's user:download(yes)/write(no)/execute(no)
+  //  6. permission code:rw-r--: owner:read(yes)/write(yes)/execute(no)|the shared action's user:download(yes)/write(no)/execute(no)
+  //  7. permission code:rwx--x: owner:read(yes)/write(yes)/execute(yes)|the shared action's user:download(no)/write(no)/execute(yes)
+  //  8. permission code:rwx---: owner:read(yes)/write(yes)/execute(yes)|the shared action's user:download(no)/write(no)/execute(no)
+  //  9. permission code:r-x--x: owner:read(yes)/write(no)/execute(yes)|the shared action's user:download(no)/write(no)/execute(yes)
+  // 10. permission code:r-x---: owner:read(yes)/write(no)/execute(yes)|the shared action's user:download(no)/write(no)/execute(no)
+  // 11. permission code:r-----: owner:read(yes)/write(no)/execute(no)|the shared action's user:download(no)/write(no)/execute(no)
+  // 12. permission code:rw----: owner:read(yes)/write(yes)/execute(no)|the shared action's user:download(no)/write(no)/execute(no)
+  val permissionList = List(
```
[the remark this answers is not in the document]
jiangpengcheng: agree

core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/Entitlement.scala
```diff
@@ -231,6 +234,157 @@ protected[core] abstract class EntitlementProvider(
...
+          permissions match {
+            case Some(value) =>
+              if (!WhiskAction.permissionList.contains(value)) {
+                val errorInfo =
+                  s"give error permission code: ${value}, available permission is in ${WhiskAction.permissionList}"
+                Future.failed(RejectRequest(Forbidden, Some(ErrorResponse(errorInfo, transid))))
+              } else {
+                val passedUpdatePermission = value.charAt(1)
+                if (passedUpdatePermission == 'w') { // make it to modifiable
+                  Future.successful(())
```
jiangpengcheng, line 284: so once passed a `w` of action permission in params, the update will always pass priviledge checking even old update permission is `-`?

core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/Entitlement.scala
```diff
@@ -231,6 +234,157 @@ protected[core] abstract class EntitlementProvider(
...
+   * @return a promise that completes with success iff action operation is feasible
+   */
+  protected[core] def checkActionPermissions(
+    operation: String,
+    user: Identity,
+    entityStore: ArtifactStore[WhiskEntity],
+    entityName: FullyQualifiedEntityName,
+    get: (ArtifactStore[WhiskEntity], DocId, DocRevision, Boolean) => Future[WhiskAction],
+    permissions: Option[String] = None)(implicit transid: TransactionId): Future[Unit] = {
+    if (operation == "create") {
```
jiangpengcheng, line 254: I think
```
operation match {
  case "create" =>
...
```
should be better

core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/Entitlement.scala
```diff
@@ -231,6 +234,157 @@ protected[core] abstract class EntitlementProvider(
...
+    } else { // download the code
+      get(entityStore, entityName.toDocId, DocRevision.empty, true)
+        .flatMap { whiskAction =>
+          val currentPermissions = whiskAction.annotations
+            .get(WhiskAction.permissionsFieldName)
+            .map(value => value.convertTo[String])
+            .getOrElse(WhiskAction.defaultPermissions)
+
+          val errorInfo = s"have no permission to download this shared action"
+          val currentDownloadPermission = currentPermissions.charAt(3)
```
jiangpengcheng, line 367: shouldn't this be `currentPermissions.charAt(0)`? since permission string is like `rwx---` and the first char indicate the "read" permission

jiangpengcheng:
one of my thought is that there is no need to add permissions controls for action's "owner" since he/she can anyway update/remove/invoke his/her action

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
style95: I meant, we can fetch the permission information in a similar way with `Identity`. 
This information can be stored in the cache too.
Permission needs to be checked before actual operations.

core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/Entitlement.scala
ningyougang: It seems it is for `the shared user`
image

core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/Entitlement.scala
ningyougang: image

ningyougang:
> one of my thought is that there is no need to add permissions controls for action's "owner" since he/she can anyway update/remove/invoke his/her action

It seems just need to control the permission for the shared user. (e.g. whether has permission to download code, execute the codes)

But for this pr, our initial target is `avoid the owner update or delete actions by mistake`, it seems if don't add permission to action's owner, can't reach the initial target

core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/Entitlement.scala
ningyougang: Updated accordingly.

ningyougang:
According to review comments

It seems there has another option to do `granular permission management`
#### 1. There is no need to add permissions controls for action's "owner" since he/she can anyway update/remove/invoke his/her action
#### 2. It is necessary to control other shared user's permission (e.g. whether has permission to read/download the codes, whether has permission to execute the codes), this separate document may like below
```
{
  "_id": "whisk.system/helloAction",
  "_rev": "1-1a14818c1083bc12b6d3fd1774fcc956",
  "default": {
      {
           "read" : false
           "execute": false
      }
   }
  "permission": {
      "theSharedUserA": {
           "read" : yes
           "execute": yes
      },
      "theSharedUserB": {
           "read" : yes
           "execute": false
      }
  }
```
Regarding above permission document of `whisk.system/helloAction`
* when create action: `whisk.system/helloAction`, the owner can set default read/execute permission. e.g.
 
```
  "default": {
      {
           "read" : false
           "execute": false
      }
   }
```
* the shared user will inherits the default permissions if not explicitly specified in above document. e.g.

not explicitly specified for theSharedUserC, so theSharedUserC will have default permission

* if explicitly specified in above document for the shared user, the shared user will have, e.g.

already explicitly specified for theSharedUserA, so theSharedUserA's permission is
```
      "theSharedUserA": {
           "read" : yes
           "execute": yes
      }
```

ningyougang:
@style95 , 

Current system permission already existed design solution:
*  the owner have any permissions(read/update/execute) for this own actions.
* if a package is binded to a `shared` package, all users except the owner(we call it shared users) just have read/execute permission for that shared package's actions but do not have update permission.
* if package is not binded to any package, all users(except the owner) do not have any permissions(e.g. read/update/execute) for that package's actions

New design solution: https://github.com/apache/openwhisk/pull/4058#issuecomment-851116337

The difference between the new design solution and `the current system design` is:
* the new design solution can control every shared user's permission

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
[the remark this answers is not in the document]
style95: So even if a owner doesn't have any permission, he can still update the permission annotations?

style95:
@ningyougang Sorry for the late response.
I want to check if we can proceed this again.

Issue #4028, opened 2018-09-18
Uncaught exception in invoker


dubee:
Below is an uncaught exception that I believe occurs when an action's response exceeds the configured HTTP max content-length. We should catch this exception properly.

```
[ERROR] Error during preStart in [akka.http.scaladsl.model.HttpEntity$Limitable@21dd4c89]: EntityStreamSizeException: actual entity size (Some(27711372)) exceeded content length limit (8388608 bytes)! You can configure this by setting `akka.http.[server|client].parsing.max-content-length` or calling `HttpEntity.withSizeLimit` before materializing the dataBytes stream.
akka.http.scaladsl.model.EntityStreamSizeException: EntityStreamSizeException: actual entity size (Some(27711372)) exceeded content length limit (8388608 bytes)! You can configure this by setting `akka.http.[server|client].parsing.max-content-length` or calling `HttpEntity.withSizeLimit` before materializing the dataBytes stream.
	at akka.http.scaladsl.model.HttpEntity$Limitable$$anon$1.preStart(HttpEntity.scala:612)
	at akka.stream.impl.fusing.GraphInterpreter.init(GraphInterpreter.scala:295)
	at akka.stream.impl.fusing.GraphInterpreterShell.init(ActorGraphInterpreter.scala:557)
	at akka.stream.impl.fusing.ActorGraphInterpreter.tryInit(ActorGraphInterpreter.scala:679)
	at akka.stream.impl.fusing.ActorGraphInterpreter.finishShellRegistration(ActorGraphInterpreter.scala:721)
	at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$shortCircuitBatch(ActorGraphInterpreter.scala:736)
	at akka.stream.impl.fusing.ActorGraphInterpreter$$anonfun$receive$1.applyOrElse(ActorGraphInterpreter.scala:761)
	at akka.actor.Actor$class.aroundReceive(Actor.scala:517)
	at akka.stream.impl.fusing.ActorGraphInterpreter.aroundReceive(ActorGraphInterpreter.scala:670)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:588)
	at akka.actor.ActorCell.invoke(ActorCell.scala:557)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:258)
	at akka.dispatch.Mailbox.run(Mailbox.scala:225)
[...]

Issue #3712, opened 2018-05-30
Make it easier to configure max action exec size for a deployment


rabbah:
To change the max size of the action blob, you have to modify:

https://github.com/apache/incubator-openwhisk/blob/22c0807b9eae20af9acc70b91a1943dd35a95092/ansible/roles/nginx/templates/nginx.conf.j2#L13

https://github.com/apache/incubator-openwhisk/blob/d815394ca5201bab7e7d270af4d471341549c843/core/controller/src/main/resources/application.conf#L57

https://github.com/apache/incubator-openwhisk/blob/b3923300638d64b375d8b79dbc268e0ae50a6bf8/common/scala/src/main/scala/whisk/core/entity/Exec.scala#L215

It should be easier to make this change.

csantanapr:
ha! I found nginx.conf.j2, Exec.scala, didn't know where akka.http was

PR #4059, opened 2018-10-08
Increase max-content-length to 50 MB

chetanmeh:
With current setup reading code attachment more than 8 MB results in error (#4028). Increase `max-content-length` to 50 MB allows reading code attachments upto maximum allowed 48 MB

## Description

Akka uses [max-content-length][1] to ensure limits on incoming request entities. By default it is 8 MB while OpenWhisk allows max code upto 48 MB. To enable such large action code as attachment increase the max-content-length to 50 MB

### Basic Test

This PR also adds a system basic test to ensure that such large actions can be created. Due to constrained resource on travis test only create action code upto 10 MB. However it can be increased via config by setting system property `whisk.action.size.max` (in MB)

This test would also help to validate if the proxy and other settings are properly configured. See #3712 for details on which all settings need to be configured

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#4028)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

[1]: https://github.com/akka/akka-http/blob/668d13d094edfb610251994d8154ac1e018a1d9c/akka-http-core/src/main/resources/reference.conf#L414

### DIFF
--- a/tests/src/test/scala/common/WhiskProperties.java
+++ b/tests/src/test/scala/common/WhiskProperties.java
@@ -349,6 +349,10 @@ public class WhiskProperties {
         return osname.equalsIgnoreCase("linux");
     }
 
+    public static int getMaxActionSizeMB(){
+        return Integer.parseInt(getProperty("whisk.action.size.max", "10"));
+    }
+
     /**
      * python interpreter.
      */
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/common/WhiskProperties.java
```diff
@@ -349,6 +349,10 @@ public static boolean onLinux() {
         return osname.equalsIgnoreCase("linux");
     }
 
+    public static int getMaxActionSizeMB(){
+        return Integer.parseInt(getProperty("whisk.action.size.max", "10"));
```
dubee: @chetanmeh, where is `whisk.action.size.max` set?

chetanmeh: Currently its not set and hence defaults to 10 MB. I wish to configure this (via system property) when we run system basic test against out setup with the max value that the cluster supports.

chetanmeh:
@dubee @markusthoemmes Please review

PR #4061, opened 2018-10-09
Ensure CosmosDBArtifactStore return query results within limits

chetanmeh:
`CosmosDBArtifactStore` may return more results than those specified by `limit`

## Description
`CosmosDBArtifactStore` query can return more results than those specified via limit. This would happen for the case involving join operation which can result in more than 1 result for a given input.

This can happen when a subject is queried by name. For example consider a subject having 2 namespaces

```json
{
  "_id": "team1",
  "_rev": "1-c3fa3d69b8474f97dbbee70cde821802",
  "namespaces": [
    {
      "name": "team1",
      "key": "k1",
      "uuid": "uuid1"
    },
    {
      "name": "team1-dev",
      "key": "k2",
      "uuid": "uuid2"
    }
  ],
  "subject": "team1"
}
```

When we query this via 

```
SELECT r AS view 
FROM root r 
JOIN n in r.namespaces 
WHERE 
  (NOT(IS_DEFINED(r.blocked)) OR r.blocked = false) 
    AND (r.subject = 'team1' OR n.name = 'team1')
```
For such queries the limit is not passed on to the db as some of the checks related to subjects are not possible via query. Hence we fetch more results than limit. In this case it would result in 2 entries referring to same `team1`. 

For such cases we need to ensure that stream result in entries under specified `limit`

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStore.scala
+++ b/common/scala/src/main/scala/whisk/core/database/cosmosdb/CosmosDBArtifactStore.scala
@@ -218,7 +218,7 @@ class CosmosDBArtifactStore[DocumentAbstraction <: DocumentSerializer](protected
 
     val publisher =
       RxReactiveStreams.toPublisher(client.queryDocuments(collection.getSelfLink, querySpec, newFeedOptions()))
-    val f = Source
+    val s = Source
       .fromPublisher(publisher)
       .wireTap(Sink.foreach(r => collectMetrics(queryToken, r.getRequestCharge)))
       .mapConcat(asSeq)
@@ -229,6 +229,12 @@ class CosmosDBArtifactStore[DocumentAbstraction <: DocumentSerializer](protected
           .transformViewResult(ddoc, viewName, startKey, endKey, realIncludeDocs, js, CosmosDBArtifactStore.this))
       .mapAsync(1)(identity)
       .mapConcat(identity)
+
+    //If limit is specified then only take that many elements. With join its possible that
+    //number of entries become > limit
+    val s2 = if (limit > 0) s.take(limit) else s
+
+    val f = s2
       .runWith(Sink.seq)
       .map(_.toList)
 
--- a/tests/src/test/scala/whisk/core/database/cosmosdb/CosmosDBTestSupport.scala
+++ b/tests/src/test/scala/whisk/core/database/cosmosdb/CosmosDBTestSupport.scala
@@ -17,7 +17,7 @@
 
 package whisk.core.database.cosmosdb
 
-import com.microsoft.azure.cosmosdb.Database
+import com.microsoft.azure.cosmosdb.{Database, SqlParameter, SqlParameterCollection, SqlQuerySpec}
 import org.scalatest.{BeforeAndAfterAll, FlatSpec}
 import pureconfig.loadConfigOrThrow
 import whisk.core.ConfigKeys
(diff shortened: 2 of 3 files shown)


chetanmeh:
Build is green on [branch][1] where [CosmosDB test do run][2]

[1]: https://travis-ci.org/chetanmeh/incubator-openwhisk/builds/439056450
[2]: https://scans.gradle.com/s/awkidcmlyz5i2/tests

tysonnorris:
Is it possible to replace `(r.subject = 'team1' OR n.name = 'team1')` with `(n.name = 'team1')`? Is there a case where we care to query based on `r.subject = 'team1'`?

chetanmeh:
> Is there a case where we care to query based on r.subject = 'team1'?

@tysonnorris Current query is design to match to current [subject view][1]

```js
function (doc) {
  if(doc.uuid && doc.key && !doc.blocked) {
    var v = {namespace: doc.subject, uuid: doc.uuid, key: doc.key};
    emit([doc.subject], v);
    emit([doc.uuid, doc.key], v);
  }
  if(doc.namespaces && !doc.blocked) {
      var v = {_id: namespace.name + '/limits', namespace: namespace.name, uuid: namespace.uuid, key: namespace.key};
      emit([namespace.name], v);
      emit([namespace.uuid, namespace.key], v);
    });
  }
}
```
From what I understand initially the key, uuid and name were at top level in a subject doc i.e. only one namespace per subject was there. Later design evolved to support multiple namespace per subject.

So the or clause is more for compatibility purpose

[1]: https://github.com/apache/incubator-openwhisk/blob/master/ansible/files/auth_index.json

tysonnorris:
My real question is : if multiple records are returned, is it guaranteed to
be safe to choose the first one? If this query is used for credentials
lookup, and the credentials are in the second record they won’t be found -
is that ok?

(This would be moot if the single record was returned by comparing only the
nested field instead of the top level field)

chetanmeh:
> if multiple records are returned, is it guaranteed to be safe to choose the first one? 

Per docs it appears to be 

https://github.com/apache/incubator-openwhisk/blob/57a38de7a6b23bc4c42f7e777a4cb2da477a355f/common/scala/src/main/scala/whisk/core/entity/Identity.scala#L62-L67

Looking at flow for `Identity#get(AuthStore, EntityName)` it appears to be used 

https://github.com/apache/incubator-openwhisk/blob/57a38de7a6b23bc4c42f7e777a4cb2da477a355f/core/controller/src/main/scala/whisk/core/controller/RestAPIs.scala#L339-L351

So one namespace can have multiple keys. Now CosmosDB 

https://github.com/apache/incubator-openwhisk/blob/57a38de7a6b23bc4c42f7e777a4cb2da477a355f/common/scala/src/main/scala/whisk/core/database/DocumentHandler.scala#L432-L456

would ensure that it picks up the right set of naemsapce from all the namespaces found in the subject. So in current case the 2 returns docs would be same subject and view compute logic would find out the matching namespace entry. Thus returning `team` only.

I have improved the new testcase to also check above case i.e. all returned result have the passed namespace

chetanmeh:
@tysonnorris I have now modified the query to check if key is also defined. So now the where clause looks like

```SELECT r AS view FROM root r JOIN n in r.namespaces WHERE (NOT(IS_DEFINED(r.blocked)) OR r.blocked = false) AND ((r.subject = @name AND IS_DEFINED(r.key)) OR n.name = @name)```

This ensures if no key is defined at top level then it would not lead to duplicate results

PR #4068, opened 2018-10-15
Change config location to work with non root user

Himavanth:
Change config location to work with non root user support which was added in #3579
## Description
With the invoker and controller now having a new user (though invoker is still running as root), we need to put the config file within the new user's home directory to maintain the access when not using the root user.

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/transformEnvironment.sh
+++ b/common/scala/transformEnvironment.sh
@@ -42,10 +42,7 @@ fi
 
 if [ -n "$config" ]
 then
-    location="/config.conf"
-    if [ -n "$NOT_ROOT_USER" ]; then
-        location="/home/$NOT_ROOT_USER/config.conf"
-    fi
+    location="$HOME/config.conf"
     printf "%s" "$config" > "$location"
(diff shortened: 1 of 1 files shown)


common/scala/transformEnvironment.sh
```diff
@@ -43,6 +43,9 @@ fi
 if [ -n "$config" ]
 then
     location="/config.conf"
+    if [ -n "$NOT_ROOT_USER" ]; then
+        location="/home/$NOT_ROOT_USER/config.conf"
+    fi
```
markusthoemmes: That makes sense to me. Should we always write to the current user's home directory to make the script as simple as possible?

Himavanth:
@markusthoemmes Thx for your feedback. I made the change. Please let me know if you think this is good to go.

PR #4073, opened 2018-10-17
Rename the package from whisk to org.apache.openwhisk

houshengbo:
## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


common/scala/src/main/scala/org/apache/openwhisk/core/WhiskConfig.scala
```diff
@@ -186,56 +186,56 @@ object WhiskConfig {
 }
 
 object ConfigKeys {
-  val cluster = "whisk.cluster"
-  val loadbalancer = "whisk.loadbalancer"
-  val buildInformation = "whisk.info"
+  val cluster = "org.apache.openwhisk.cluster"
```
chetanmeh: I do not think we need to change the configKeys?

houshengbo: We need to change everything based on the Apache policy.

dgrove-oss: I'm not convinced we need the extra levels here.  It should be further discussed.

Even if we eventually decide to make the change, it absolutely should be a separate PR.  Changing the name of config variables is significantly higher impact on downstream projects than changing the names of the Scala/Java packages

common/scala/src/main/scala/org/apache/openwhisk/core/WhiskConfig.scala
```diff
@@ -30,7 +30,7 @@ import whisk.common.{Config, Logging}
  * a value, and whose values are default values. A null value in the Map means there is
  * no default value specified, so it must appear in the properties file.
  * @param optionalProperties a set of optional properties (which may not be defined).
- * @param propertiesFile a File object, the whisk.properties file, which if given contains the property values.
+ * @param propertiesFile a File object, the org.apache.openwhisk.properties file, which if given contains the property values.
```
dgrove-oss: The name of the properties file should not be changed.

houshengbo:
RangeError: Array buffer allocation failed","2018-10-25T20:28:08.370376749Z stderr: at Buffer.Uint8Array (native)
This error is occasionally raised by the test file ActionLimitsTests.

PR #4077, opened 2018-10-19
Fix bug of cal manged and blackboxes number

ningyougang:
## Description

Actually, in most cases, there has no problem.
But in some special cases, there has a little problem. e.g. if we assign `fraction = 0.99`
```
  val newSize = 1000
  val fraction = 0.99
  val managed = Math.max(1, Math.ceil(newSize.toDouble * (1 - fraction)).toInt)
  val blackboxes = Math.max(1, Math.floor(newSize.toDouble * fraction).toInt)
```
The manged number in above codes will be `11`
The blackboxes number will be `990`
Obviously, the correct manged number should be `10`.
So i think use `Math.round` may be more reasonable

## Related issue and scope

- [ ] API
- [ ] Controller
- [x] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


rabbah:
You need to rebase your branch to master to pick up a fix for Travis.

rabbah:
This isn't actually a defect - it's true that 11 vs 10 is not perfect, but the reason for ceil/floor vs round is that for small number (say 2 invokers):

```
Let newSize = 2 and blackboxFrac = 0.1.
Current code:  managed = 2   blackboxes = 1
This PR:           managed = 1   blackboxes = 1
```

The point for small N up to the reciprocal of the fraction, both sets overlap.

PR #4078, opened 2018-10-23
Add ability to not store activations for specified namespace.

cbickel:
## Description
With this PR, operators will be able to disable storing of activations into activations store.
The flag is implemented like the per-namespace-limits. It can be set with wskadmin.

## Related issue and scope
- [x] I wrote a mail to the dev list to discuss the proposal

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [x] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
@@ -49,6 +49,7 @@ case class ActivationMessage(override val transid: TransactionId,
                              rootControllerIndex: ControllerInstanceId,
                              blocking: Boolean,
                              content: Option[JsObject],
+                             remainingQuota: RemainingQuota,
                              cause: Option[ActivationId] = None,
                              traceContext: Option[Map[String, String]] = None)
     extends Message {
@@ -68,7 +69,7 @@ object ActivationMessage extends DefaultJsonProtocol {
   def parse(msg: String) = Try(serdes.read(msg.parseJson))
 
   private implicit val fqnSerdes = FullyQualifiedEntityName.serdes
-  implicit val serdes = jsonFormat10(ActivationMessage.apply)
+  implicit val serdes = jsonFormat11(ActivationMessage.apply)
 }
 
 /**
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
@@ -49,7 +49,6 @@ case class ActivationMessage(override val transid: TransactionId,
                              rootControllerIndex: ControllerInstanceId,
                              blocking: Boolean,
                              content: Option[JsObject],
-                             remainingQuota: RemainingQuota,
                              cause: Option[ActivationId] = None,
                              traceContext: Option[Map[String, String]] = None)
     extends Message {
@@ -69,7 +68,7 @@ object ActivationMessage extends DefaultJsonProtocol {
   def parse(msg: String) = Try(serdes.read(msg.parseJson))
 
   private implicit val fqnSerdes = FullyQualifiedEntityName.serdes
-  implicit val serdes = jsonFormat11(ActivationMessage.apply)
+  implicit val serdes = jsonFormat10(ActivationMessage.apply)
 }
 
 /**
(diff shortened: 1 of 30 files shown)


ningyougang:
@cbickel , i also meet below problem when building
```
./tools/build/checkLogs.py: logs is not a directory.
Traceback (most recent call last):
  File "./tools/build/checkLogs.py", line 108, in <module>
    errors = run_file_checks(file_path, checks)
  File "./tools/build/checkLogs.py", line 58, in run_file_checks
    errs = check(file_path)
  File "./tools/build/checkLogs.py", line 45, in database_has_at_most_x_entries
    with open(file) as db_file:
IOError: [Errno 2] No such file or directory: 'logs/db-rules.log'
Directory doesn't exist: /home/travis/build/apache/incubator-openwhisk/logs.
Uploaded Logs with name unit_445493566-master.tar.gz
Time taken for checkAndUploadLogs.sh is 0 secs
```

ningyougang:
@cbickel , is it support `not store activations for specified action`?

may be exist below scenes, for example.
One namepspace, has two actions, one needs to save activations to couchdb, but another action has no 
need to save activation to couchdb.

markusthoemmes:
@ningyougang you might want to answer to the dev-list discussion here: https://lists.apache.org/thread.html/882a13d78cf24f56bae529bb217a096d665feaff2ca761e9ae9572a7@%3Cdev.openwhisk.apache.org%3E

I outlined a similar situation where a more general cascading of our limits might be useful.

common/scala/src/main/scala/whisk/core/entity/Identity.scala
```diff
@@ -32,10 +32,11 @@ import whisk.core.entitlement.Privilege
 case class UserLimits(invocationsPerMinute: Option[Int] = None,
                       concurrentInvocations: Option[Int] = None,
                       firesPerMinute: Option[Int] = None,
+                      disableActivationStore: Option[Boolean] = None,
```
rabbah: I do not like this - it's a brute force hammer and dramatically changes the user's experience: no logs, no activations. In any case, this should be treated as a quota with a limit on storing activations and then disabling all activation records would be equivalent to setting the quota down to 0.

common/scala/src/main/scala/whisk/core/database/ArtifactActivationStore.scala
```diff
@@ -40,9 +40,16 @@ class ArtifactActivationStore(actorSystem: ActorSystem, actorMaterializer: Actor
     implicit transid: TransactionId,
     notifier: Option[CacheChangeNotification]): Future[DocInfo] = {
 
-    logging.debug(this, s"recording activation '${activation.activationId}'")
-
-    val res = WhiskActivation.put(artifactStore, activation)
+    val skipStoring = context.user.limits.disableActivationStore.getOrElse(false)
```
rabbah: This seems awfully late to do and is introducing a hack into the artifact store. IF you want to short circuit the entire activation record, it should be done at the source, not here, in my view... because you've also wasted the effort of gathering logs and creating the record, only to throw it away.

ningyougang:
@cbickel agree

common/scala/src/main/scala/whisk/core/database/ArtifactActivationStore.scala
markusthoemmes: It indeed should be done at source. The option says `disableActivationStore`, so we should in theory circumvent the ActivationStore call as a whole.

common/scala/src/main/scala/whisk/core/entity/Identity.scala
cbickel: That definetly makes sense to me.
What would your proposal of an implementation? Should we have a per-minute limit for these logs and let the controller check if the current activation will be written or not?

common/scala/src/main/scala/whisk/core/database/ArtifactActivationStore.scala
cbickel: Your comment makes sense to me. I was also thinking where to put it and decided to put it here, to match all occurances in one place.
But I can move it up, of cause.

markusthoemmes: @cbickel it only matches all occurrences for this specific implementation of the ActivationStore though. It shouldn't be in the hand of the activation store to decide if this specific quota is enforced or not.

rabbah:
@cbickel what's the plan for this - it has rotted?

markusthoemmes:
This has been replaced by a more general limit-using approach IIRC.

cbickel:
Closing for now, until we want to change the limit from a flag to a numerical limit.

PR #4080, opened 2018-10-24
Catch kafka producer exceptions and recreate the producer

vvraskin:
## Description

When the execute _send()_ on the Kafka producer from the underlying java library fails, there is bunch of Kafka exceptions we do not catch. In some cases the Callback is not executed as a result of the call and it could be the reason of the consumer hangup https://github.com/apache/incubator-openwhisk/issues/4076. Since we still do not fully understand the nature of the failures, the main aim of this PR is to expose the exceptions that happen in the producer. 
## Related issue and scope

- [x] The issues is already here https://github.com/apache/incubator-openwhisk/issues/4076

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
+++ b/common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
@@ -33,7 +33,7 @@ import whisk.utils.Exceptions
 import scala.collection.JavaConverters._
 import scala.concurrent.duration._
 import scala.concurrent.{blocking, ExecutionContext, Future, Promise}
-import scala.util.{Failure, Success}
+import scala.util.{Failure, Success, Try}
 
 class KafkaProducerConnector(
   kafkahosts: String,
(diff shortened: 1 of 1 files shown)


common/scala/src/main/scala/whisk/connector/kafka/KafkaProducerConnector.scala
```diff
@@ -55,12 +55,16 @@ class KafkaProducerConnector(
...
           override def onCompletion(metadata: RecordMetadata, exception: Exception): Unit = {
             if (exception == null) produced.success(metadata)
             else produced.failure(exception)
           }
-        })
+        })).recover {
+          case e: Exception =>
+            logging.error(this, e.getMessage)
+            produced.failure(e)
+        }
```
markusthoemmes: Since you don't do anything with the `Try` value, a simple `try/catch` would suffice here and prevent unnecessary allocations.

The logging is unnecessary I believe, since this will then be fed into the error handling logic below, which logs failures by default as well. The logline would be redundant.

Also, to be absolutely sure we're not creating a new race condition here, please change all the `success`/`failure` calls to `trySuccess`/`tryFailure`. That'll make sure that even if the Callback is executed on a thrown exception (unlikely but better be defensive) the promise resolving code isn't throwing another exception.

PR #4082, opened 2018-10-24
Rename prewarmed containers

kpavel:
Rename prewarmed containers when initialized
## Description

Currently, when prewarmed container moves out from prewarmed pool, it keeps its "prewarmed" name, e.g. wsk00_236_prewarm_nodejs6 while if it wouldn't be taken from prewarmed pool it could have name like wsk00_240_guest_hello which is based on namespace_action name.
In this case, when observing currently running containers, one may have number of containers having default "prewarm" in the name, exceeding the number of prewarmed containers specified in the config. 
As well, it may present issues when it comes to monitoring, as containers name differ from expected
## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
@@ -331,6 +331,19 @@ class ContainerProxy(
   }
 
   /**
+   * Checks whether container name is "prewarmed"
+   *
+   * @param suffix the container name's suffix
+   * @return a unique container name
+   */
+  def isPrewarmed(name: String, instance: InvokerInstanceId, suffix: String): Boolean = {
+    val sanitizedSuffix = suffix.filter(ContainerProxy.isAllowed)
+
+    return name.startsWith(s"${ContainerFactory.containerNamePrefix(instance)}_") && name.endsWith(
+      s"_prewarm_${sanitizedSuffix}")
+  }
+
+  /**
    * Runs the job, initialize first if necessary.
    * Completes the job by:
    * 1. sending an activate ack,
@@ -349,7 +362,15 @@ class ContainerProxy(
     // Only initialize iff we haven't yet warmed the container
     val initialize = stateData match {
       case data: WarmedData => Future.successful(None)
-      case _                => container.initialize(job.action.containerInitializer, actionTimeout).map(Some(_))
+      case _                =>
+        // check if container has generic "prewarm" name
+        if (isPrewarmed(container.name, instance, job.action.exec.kind)) {
+          // now rename container to action based name
+          val newName =
+            ContainerProxy.containerName(instance, job.msg.user.namespace.name.asString, job.action.name.asString)
+          container.rename(newName)
+        }
+        container.initialize(job.action.containerInitializer, actionTimeout).map(Some(_))
     }
 
     val activation: Future[WhiskActivation] = initialize
@@ -450,6 +471,8 @@ object ContainerProxy {
 
   val timeouts = loadConfigOrThrow[ContainerProxyTimeoutConfig](ConfigKeys.containerProxyTimeouts)
 
+  def isAllowed(c: Char): Boolean = c.isLetterOrDigit || c == '_'
+
   /**
    * Generates a unique container name.
    *
@@ -458,8 +481,6 @@ object ContainerProxy {
    * @return a unique container name
    */
   def containerName(instance: InvokerInstanceId, prefix: String, suffix: String): String = {
-    def isAllowed(c: Char): Boolean = c.isLetterOrDigit || c == '_'
-
     val sanitizedPrefix = prefix.filter(isAllowed)
     val sanitizedSuffix = suffix.filter(isAllowed)
 
--- a/core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
@@ -86,7 +86,8 @@ object KubernetesContainer {
  * @param workerIP the ip of the workernode on which the container is executing
  * @param nativeContainerId the docker/containerd lowlevel id for the container
  */
-class KubernetesContainer(protected[core] val id: ContainerId,
+class KubernetesContainer(protected[core] var _name: String,
+                          protected[core] val id: ContainerId,
                           protected[core] val addr: ContainerAddress,
                           protected[core] val workerIP: String,
                           protected[core] val nativeContainerId: String)(implicit kubernetes: KubernetesApi,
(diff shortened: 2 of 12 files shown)


sven-lange-last:
Basically, the idea of renaming pre-warmed Docker action containers when they are inited makes sense. As you pointed out, a simple `docker ps -a --filter 'name=wsk*'` will give you all action containers. With your renaming proposal, all containers containing `prewarm` in their name are pre-warmed while all others are warm - ignoring error conditions for the moment.

The downside of renaming is that it takes time. It will add to the latency of pre-warmed invocations, i.e. Node.js invocations will take longer. Did you run any specific performance benchmarks to understand the effects of renaming or did you compare performance test results of Travis builds?

I only had a very short look at your code changes. You are adding a `rename()` operation to the ContainerProxy workflow - but it seems that you don't test whether renames are properly performed in ContainerProxy tests.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
```diff
@@ -349,7 +362,15 @@ class ContainerProxy(
...
     val initialize = stateData match {
       case data: WarmedData => Future.successful(None)
-      case _                => container.initialize(job.action.containerInitializer, actionTimeout).map(Some(_))
+      case _                =>
+        // check if container has generic "prewarm" name
+        if (isPrewarmed(container.name, instance, job.action.exec.kind)) {
+          // now rename container to action based name
+          val newName =
+            ContainerProxy.containerName(instance, job.msg.user.namespace.name.asString, job.action.name.asString)
+          container.rename(newName)
```
dgrove-oss: I think this is asynchronously renaming the container (since `rename` returns a Future, but that Future object is dropped on the floor).  On the one hand, this is good in that it means we are adding minimal latency to an important hot path.  On the other hand, it is not good in that if the rename fails (a) we don't notice and (b) we don't attempt to retry or otherwise handle the error.

Did I read this correctly

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
```diff
@@ -138,4 +139,9 @@ class KubernetesContainer(protected[core] val id: ContainerId,
       .takeWithin(waitForLogs)
       .map { _.toByteString }
   }
+
+  /** Rename container. */
+  override def rename(name: String)(implicit transid: TransactionId): Future[Unit] =
+    // rename currently not supported
+    Future.successful(Unit)
```
dgrove-oss: Would be easy enough to implement rename by making a call to change the value of the `name` label of the container.  However, this will add some latency, so as above we should decide if we can do it asynchronously and ignore any errors or synchronously so errors can be cleanly detected and handled.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
rabbah: This is a debug feature for operators? I'm trying to get an idea of the motivating use case here. Users of the platform should not see their containers.

rabbah: Are the suspend and resume operations affected by a pending rename?

dgrove-oss: Good point on possible bug with suspend/resume @rabbah.  That argues that the code should be changed to join the two futures for `rename` and `initialize` before it proceeds with moving to the warm state.   We could still overlap the two operations to attempt to hide the latency of the rename behind the initialize, but we should make sure the rename has completed before continuing.

dgrove-oss:
Motivating use case is when we are using the DockerContainerFactory on Kubernetes, where the container names will be visible to operators (who may in fact be users, in the case of deploying OpenWhisk on self-managed kube clusters).  By changing the name of the container, platform-level metric services that include the container name will give the user/operator useful information about which functions are consuming resources.

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
kpavel: > Are the suspend and resume operations affected by a pending rename?
According to docker daemon source code those operations are synchronized by container object. I don't see there any possible conflict. 
From the OpenWhisk perspective, all container operations currently using container UID.

core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala
kpavel: From my brief observations of invoker logs, after running system tests:
rename latency - average 43msec
initialization latency - average 185msec
https://gist.github.com/kpavel/a424db24139d83075abb31b0de365885
https://gist.github.com/kpavel/3464751e26b600a087b603c4767b3fa4

Although rename latency is lower than initialization one, it is too significant to ignore. I think it would be better to make it async instead of increasing initialization latency by 25%

core/invoker/src/main/scala/whisk/core/containerpool/ContainerProxy.scala
rabbah: It should be noted that in the past increasing concurrent operations against docker surfaced deadlocks in the daemon.

rabbah:
Thanks Dave. So this has to happen early on in the container life cycle then?

mrutkows:
@dgrove-oss Even understanding the use case, it seems wrong to track state via name (especially given the concurrency/locking concerns Rodric raised despite appearing to be asynchronous). My main concern is that there is nothing actionable an operator can do from seeing a collection of named containers which may indicate some resource consumption state.  At best, this approach seems to be an attempt at a stop-gap measure for "container watchers", but would require deeper analysis via log/monitoring tools to be actionable. Is it worth going down this road and raising the expectations of operators that name will reflect the actual state with fidelity?

rabbah:
is it ok to do the operation late (after run?) vs early (before init)?

dgrove-oss:
Based on @kpavel's comment that the name doesn't matter for suspend/resume since we use the ContainerID, I would be inclined to do it asynchronously eaerly as is being done now, but with some error handling so that if the rename fails for some reason we at least get an error log.

rabbah:
But if it increases load in the daemon then this could be a concern in production - we limit the max concurrent docker operation iirc.

dgrove-oss:
We should probably understand what versions of docker are actually being used in production by the various operators these days.  For example, IBM is no longer on 1.12.  Maybe that is a dev-list discussion (to go with a discussion of #4090).

sven-lange-last:
@dgrove-oss IBM is partly still at Docker 1.12 - but no more with production environment.

Even if we `docker rename` pre-warmed containers asynchronously in the container proxy, there is a chance that Docker operations on the same container are run consecutively by the Docker daemon. And if not consecutively, there are probably per-container or global locks that prevent fully parallel execution.

We definitely need performance tests to evaluate the effects of a `docker rename`.

dgrove-oss:
Going to close this.  There was not a clear consensus that we wanted to do this, the PR has gotten out of date, and there's a general movement from docker to containerd as the lowlevel engine in Kubernetes.

PR #4087, opened 2018-10-29
Use `latest` as default tag for apigateway

dgrove-oss:
The default tag for all openwhisk project docker images on the master branch should be `latest` so that we are always testing against our own latest code.  Operators can override to pin versions as desired in their own deployments.


dgrove-oss:
@csantanapr @mdeuser -- as discussed in slack thread....test against our own latest by default.

PR #4088, opened 2018-10-29
Make activation polling for blocking invocations configurable

dubee:
## Description

Allows controller activation polling for blocking invocations to be enabled or disable via configuration.

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


dubee:
Changed this back to a WIP state as the two active ack changes do not currently allow DB polling to be disabled.

core/controller/src/main/scala/whisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -593,11 +596,15 @@ protected[actions] trait PrimitiveActions {
     //    in case of an incomplete active-ack (record too large for example).
     activeAckResponse.foreach {
       case Right(activation) => result.trySuccess(Right(activation))
-      case _                 => pollActivation(docid, context, result, i => 1.seconds + (2.seconds * i), maxRetries = 4)
+      case _ if (controllerActivationConfig.pollingFromDb) =>
+        pollActivation(docid, context, result, i => 1.seconds + (2.seconds * i), maxRetries = 4)
```
tysonnorris: I'm looking at a case right now where if this pollActivation is removed, the acks are received in the wrong order in some cases (completion before result) - have you seen anything similar? unfortunately, it is something weird where it only happens on first run of the controller - subsequent runs of same or different actions don't exhibit the problem.  Restarting controller, and the first invocation always gets the acks in wrong order again.

tysonnorris: I think I see the issue causing the out of order messages - the first call to `AcknowledegmentMessage.parse(raw)` is slow, resulting in out of order messages IFF the result + completion are sent at approx same time from invoker. I'll try to figure out a way to force the order of these.

markusthoemmes: @tysonnorris is this related to #4134, which supposedly has been fixed by now?

dubee: After rebasing it looks the two active ack changes no longer required DB polling.

dubee:
@rabbah, rebased and ready to go.

PR #4090, opened 2018-10-29
upgrade from docker 1.12 to 18.06

csantanapr:
docker cli tgz no longer available
https://get.docker.com/builds/Linux/x86_64/docker-12.0.tgz

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -275,7 +275,7 @@ docker:
   image:
     prefix: "{{ docker_image_prefix | default('whisk') }}"
     tag: "{{ docker_image_tag | default('latest') }}"
-  version: 1.12.0-0~trusty
+  version: 18.06.1-ce
   storagedriver: overlay
   port: 4243
   restart:
--- a/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
+++ b/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
@@ -124,7 +124,7 @@ class DockerClient(dockerHost: Option[String] = None,
         }
         .map(ContainerId.apply)
         .recoverWith {
-          // https://docs.docker.com/v1.12/engine/reference/run/#/exit-status
+          // https://docs.docker.com/engine/reference/run/#general-form
           // Exit status 125 means an error reported by the Docker daemon.
           // Examples:
           // - Unrecognized option specified
--- a/tools/travis/setup.sh
+++ b/tools/travis/setup.sh
@@ -31,12 +31,14 @@ function retry() {
 }
 
 #sudo gpasswd -a travis docker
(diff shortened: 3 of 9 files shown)


tools/travis/setup.sh
```diff
@@ -30,12 +30,14 @@ function retry() {
   fi
 }
 
-sudo gpasswd -a travis docker
+#sudo gpasswd -a travis docker
```
dubee: @csantanapr, can this line be deleted, or is it still needed?

csantanapr: not sure if it's need it or note, wanted to try without it first

csantanapr: removed

core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerClient.scala
```diff
@@ -124,7 +124,7 @@ class DockerClient(dockerHost: Option[String] = None,
         }
         .map(ContainerId.apply)
         .recoverWith {
-          // https://docs.docker.com/v1.12/engine/reference/run/#/exit-status
+          // https://docs.docker.com/engine/reference/run/#general-form
```
sven-lange-last: This is the proper link for exit status: https://docs.docker.com/engine/reference/run/#exit-status.

ansible/group_vars/all
```diff
@@ -175,7 +175,7 @@ invoker:
   # Specify if it is allowed to deploy more than 1 invoker on a single machine.
   allowMultipleInstances: "{{ invoker_allow_multiple_instances | default(false) }}"
   # Specify if it should use docker-runc or docker to pause/unpause containers
-  useRunc: "{{ invoker_use_runc | default(true) }}"
+  useRunc: "{{ invoker_use_runc | default(false) }}"
```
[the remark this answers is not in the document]
csantanapr: I agree, I will add he comment. 
I agree with the options we have, right now so don’t have the bandwidth to look into why runc is not working and to repair. 
I’m ok to retired if no one has the bandwidth to investigate and fix

rabbah:
That would come with a significant performance penalty then, no?

csantanapr:
Rodric I think I’m hearing that you would like to volunteer to help look into runc with new version of docker :-)

markusthoemmes:
I think it warrants a dev-list discussion as to why runc isn't working anymore and what the implications of the switch are, no?

csantanapr:
In theory and on paper containerd should work, Sven is already looking and doing research in to that.

csantanapr:
I agree with Markus, in the mean time what do we do with this PR?
The purpose of this PR was to unblock the Travis CI. I can take a copy of the docker CLI 1.12 put it somewhere and fetch it from there.

markusthoemmes:
@csantanapr is the download gone for good or will it eventually come back? We've done things in the past, where we've just based off of our own invoker image to short-circuit the download.

Another alternative is, that we quickly publish our own java/docker1.12 image and build the invoker on top of that in the future. That sounds the quickest and simplest to me tbh.

csantanapr:
I would assume download is gone for good, so having our own copy would be best.

PR #4094, opened 2018-11-02
increase default invoker memory available for user actions to 2GB

csantanapr:
## Description
the performance test stage has being failing lately in travis, intermittently 
running locally with default setting the rps on many runs is always 1 request per second.

increasing the memory I got around 7rps
```
OPENWHISK_HOST="172.17.0.1" USERS="1" REQUESTS_PER_SEC="1" ./gradlew gatlingRun-ColdBlockingInvokeSimulation
```

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [x] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -170,7 +170,7 @@ invoker:
   port: 12001
   heap: "{{ invoker_heap | default('2g') }}"
   arguments: "{{ invoker_arguments | default('') }}"
-  userMemory: "{{ invoker_user_memory | default('1024 m') }}"
+  userMemory: "{{ invoker_user_memory | default('2048 m') }}"
   instances: "{{ groups['invokers'] | length }}"
   # Specify if it is allowed to deploy more than 1 invoker on a single machine.
   allowMultipleInstances: "{{ invoker_allow_multiple_instances | default(false) }}"


sven-lange-last:
@csantanapr would you know how much memory we have in a Travis VM? If user action containers consume too much memory such that the Linux system has memory pressure, the system may decide to OOM kill the largest resident memory consuming process which may be the invoker or controller. We have seen this during tests.

csantanapr:
@sven-lange-last

I understand what the change does the thing is we used to have 16 parallel actions before the change to using memory the ansible config used to be  `16` `busyThreshold: "{{ invoker_busy_threshold | default(16) }}"`

csantanapr:
We should be OK Travis in our case we get 7.5GB of memory 
https://docs.travis-ci.com/user/reference/overview/

csantanapr:
And Travis tests are passing with the change

dgrove-oss:
System test portion of the travis run timed out.  I started it again.  Not sure if this is problematic or just bad luck.  If it fails again, we may need to dig deeper

dgrove-oss:
would cutting the action memory used for the perf test down to 128Kb help?  We could get 8 in 1GB then (none would be pre-warmed, but that probably is ok).

csantanapr:
The Travis failed but the test that failed looks 🐟 

Take a look
```
system.basic.WskActionTests > Whisk actions should blocking invoke an asynchronous action FAILED
    org.scalatest.exceptions.TestFailedException: check failed for activation: ActivationResult(72435352337c4a57835352337c5a57a8,Some(List(2018-11-02T12:34:49.99522953Z  stdout: The message 'this is a test' has 4 words)),ActivationResponse(Some({"count":4}),success,true),2018-11-02T12:34:49.782Z,2018-11-02T12:34:49.999Z,217,None,Some(List({"key":"path","value":"guest/helloAsync"}, {"key":"waitTime","value":33}, {"key":"kind","value":"nodejs:6"}, {"key":"limits","value":{"timeout":60000,"memory":256,"logs":10}}, {"key":"initTime","value":64}))) Some(List("2018-11-02T12:34:49.99522953Z  stdout: The message 'this is a test' has 4 words")) was not equal to Some(List())
```
This test `Whisk actions should blocking invoke an asynchronous action`
Uses `helloAsync` with `helloAsync.js` and assertion is that there are no logs.
But it contains the log `The message 'this is a test' has 4 words`

Which is the output from the previous test `Whisk actions should blocking invoke of nested blocking actions`
On a different action `wc` with `wc.js`

PR #4100, opened 2018-11-06
Always return activation without log on blocking invoke.

cbickel:
If an user invokes an action blockingly today, the invoker executes the activation and sends back the result to the controller, which passes it back to the client. If it is not possible to pass the result back to the controller, it will try to get it from the database. And here's the problem. If the controller receives the result of the activation from the invoker, there are no logs included. If it polls it from the database, the logs are included. This should be unified to never return logs.

## Related issue and scope

- [x] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


rabbah:
LGTM also - I thought we used to do this in the Actions API route but must have lost it in some refactoring. Good catch.

PR #4104, opened 2018-11-08
Cache empty auth results to reduce db load

mhenke1:
Cache negative auth responses to avoid going to the DB with every identical call 

## Description
Since negative auth responses (Identity.get) are not cached, 
all requests with a non existing namespace or an invalid authkey will perform queries to the subject db.
A user can influence OW performance by issuing a high number of those calls.

ThIs PR changes that behavior in the way, that negative auth results are cached as None value.
Subsequent Identical calls will only use the cache (for the 5 minutes cache period).
The external interface and behavior of the Identity.get methods stays unchanged.

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/database/MultipleReadersSingleWriterCache.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/database/MultipleReadersSingleWriterCache.scala
@@ -20,14 +20,12 @@ package org.apache.openwhisk.core.database
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicReference
 
-import scala.concurrent.{ExecutionContext, Future, Promise}
-import scala.util.Failure
-import scala.util.Success
 import com.github.benmanes.caffeine.cache.Caffeine
-import org.apache.openwhisk.common.Logging
-import org.apache.openwhisk.common.LoggingMarkers
-import org.apache.openwhisk.common.TransactionId
+import org.apache.openwhisk.common.{Logging, LoggingMarkers, TransactionId}
 import org.apache.openwhisk.core.entity.CacheKey
+
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.util.{Failure, Success}
 
 /**
  * A cache that allows multiple readers, but only a single writer, at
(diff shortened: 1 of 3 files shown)


rabbah:
Agree. Is 5 minutes too high?

chetanmeh:
Caching such negative lookups may be abused to put lots of of entries in cache and put pressure on heap (causing eviction of valid entries) given our caches are not bounded by size. May be use a separate bounded cache for such entries.

markusthoemmes:
@chetanmeh the cache is "bounded" as in its entries are garbage collectable.

chetanmeh:
@markusthoemmes All caches are currently configured with `softValues` thereby the entries are evicted (apart from expiry) on the basis of jvm garbage collection. Now if an attacker is given a way to add entry to one of the cache in arbitrary manner then it can possibly be abused to add lots of entries in Identity cache. This would put pressure on heap and would possibly trigger evictions of valid data in all other caches also.

If we use a bounded cache in terms of entry count then only identity cache would be impacted but other caches would not get impacted.

mhenke1:
I have now limited the auth cache hard to 100000 entries. Worth to be configurable ?

common/scala/src/main/scala/org/apache/openwhisk/core/database/MultipleReadersSingleWriterCache.scala
```diff
@@ -445,25 +444,37 @@ trait MultipleReadersSingleWriterCache[W, Winfo] {
   }
 
   /** This is the backing store. */
-  private lazy val cache: ConcurrentMapBackedCache[Entry] = evictionPolicy match {
-    case AccessTime =>
-      new ConcurrentMapBackedCache(
-        Caffeine
+  private lazy val cache: ConcurrentMapBackedCache[Entry] = new ConcurrentMapBackedCache(
```
chetanmeh: This can be simplified a bit
```suggestion
    private lazy val cache: ConcurrentMapBackedCache[Entry] = createCache()

  private def createCache() = {
    val b = Caffeine
      .newBuilder()
      .softValues()

    evictionPolicy match {
      case AccessTime => b.expireAfterAccess(5, TimeUnit.MINUTES)
      case _          => b.expireAfterWrite(5, TimeUnit.MINUTES)
    }

    if (fixedCacheSize > 0) b.maximumSize(fixedCacheSize)
    new ConcurrentMapBackedCache(b.build().asMap().asInstanceOf[ConcurrentMap[Any, Future[Entry]]])
  }
```

chetanmeh:
> I have now limited the auth cache hard to 100000 entries. Worth to be configurable ?

@mhenke1 Should be fine to leave it specified in code for now

mhenke1:
[PG 2:3878]

mhenke1:
@chetanmeh I have changed the code as suggested

chetanmeh:
@rabbah @markusthoemmes can you review cache changes once

PR #4109, opened 2018-11-12
update mesos-actor; cleanup orphaned failed task launches

tysonnorris:
## Description

When task launch times out, cleanup tasks that were sent to mesos (or not yet sent).
Also updating mesos-actor which includes a fix for accepting resources that are not assigned to the frameworks role.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/mesos/MesosTask.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/mesos/MesosTask.scala
@@ -37,8 +37,6 @@ import com.adobe.api.platform.runtime.mesos.SubmitTask
 import com.adobe.api.platform.runtime.mesos.TaskDef
 import com.adobe.api.platform.runtime.mesos.User
 import java.time.Instant
-import org.apache.mesos.v1.Protos.TaskState
-import org.apache.mesos.v1.Protos.TaskStatus
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.util.Failure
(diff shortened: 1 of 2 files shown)


common/scala/src/main/scala/org/apache/openwhisk/core/mesos/MesosTask.scala
```diff
@@ -155,7 +158,29 @@ object MesosTask {
...
+    mesosClientActor
+      .ask(DeleteTask(taskId))(taskDeleteTimeout)
+      .andThen {
+        case Success(_) => transid.finished(this, start, logLevel = InfoLevel)
+        case Failure(ate: AskTimeoutException) =>
+          transid.failed(this, start, s"task destroy timed out ${ate.getMessage}", ErrorLevel)
+          MetricEmitter.emitCounterMetric(LoggingMarkers.INVOKER_MESOS_CMD_TIMEOUT(MesosTask.KILL_CMD))
+        case Failure(t) => transid.failed(this, start, s"task destroy failed ${t.getMessage}", ErrorLevel)
+      }
+      .map(_ => {})(ec)
```
chetanmeh: `ec` is implicitly available so we can drop passing it explicitly.

common/scala/src/main/scala/org/apache/openwhisk/core/mesos/MesosTask.scala
```diff
@@ -142,9 +140,14 @@ object MesosTask {
...
         case Failure(ate: AskTimeoutException) =>
-          transid.failed(this, start, ate.getMessage, ErrorLevel)
+          transid.failed(this, start, s"task launch timed out ${ate.getMessage}", ErrorLevel)
           MetricEmitter.emitCounterMetric(LoggingMarkers.INVOKER_MESOS_CMD_TIMEOUT(LAUNCH_CMD))
-        case Failure(t) => transid.failed(this, start, t.getMessage, ErrorLevel)
+          //kill the task whose launch timed out
+          destroy(mesosClientActor, mesosConfig, taskId)
+        case Failure(t) =>
+          //kill the task whose launch timed out
+          destroy(mesosClientActor, mesosConfig, taskId)
```
chetanmeh: Note that the future returned by `destroy` for `andThen` call would be discarded so its more like ask and forget here. Is that the intention or it would be better to have destroy completed by the time `create`s resulting future completes?

tysonnorris: In this case destroy is just cleanup that is mesos-specific, so it should happen independent of the create future; i.e. `create()` timeout should return immediately, but trigger the cleanup of `destroy()`

PR #4114, opened 2018-11-15
Update KindRestrictor to merge namespace and default whitelists

asteed:
## Description
When performing `KindRestrictor.check()`, merge the contents of both the namespace's `allowedKinds` limit with the system's default `whitelist` of allowed kinds.

Originally, the namespace's value for the `allowedKinds` limit was used _instead of_ the value of the default `whitelist` set for the system. However, while crafting a deployment plan enabling whitelist support, we ran into concerns around maintaining these per-namespace whitelists as the system and user/namespace population grows.

The most notable scenario of concern is when a kind is added or removed from the OW system's runtime manifest. After these are added to the manifest, any existing default `whitelist` will need to be updated to include/exclude the new kind. This is an expected maintenance operation. However, in addition to this, _all **existing** limits_ for namespaces' `allowedKinds` additionally have to be modified.

As new kinds enter the system, enabling access to new kinds (which are _already accessible_ for those _without_ an explicit namespace `allowedKinds` limit) will require a batch operation (or similar migration) to update all existing `allowedKinds` namespace limits. This becomes a particularly unnecessary operation that can grow over time.

The solution outlined in this PR is to reduce the need to perform updates per-namespace when the default whitelist can be merged together with the namespace limit.

## Related issue and scope

- [ ] API
- [X] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

- [X] Breaking change (a bug fix or enhancement which changes existing behavior).

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/KindRestrictor.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/KindRestrictor.scala
@@ -46,7 +46,7 @@ case class KindRestrictor(whitelist: Option[Set[String]] = None)(implicit loggin
     })(TransactionId.controller)
 
   def check(user: Identity, kind: String): Boolean = {
-    val kindList = user.limits.allowedKinds.getOrElse(Set()).union(whitelist.getOrElse(Set()))
+    val kindList = user.limits.allowedKinds.getOrElse(Set.empty).union(whitelist.getOrElse(Set.empty))
     kindList.isEmpty || kindList.contains(kind)
   }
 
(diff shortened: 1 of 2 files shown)


core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/KindRestrictor.scala
```diff
@@ -46,10 +46,8 @@ case class KindRestrictor(whitelist: Option[Set[String]] = None)(implicit loggin
     })(TransactionId.controller)
 
   def check(user: Identity, kind: String): Boolean = {
-    user.limits.allowedKinds
-      .orElse(whitelist)
-      .map(allowed => allowed.contains(kind))
-      .getOrElse(true)
+    val kindList = user.limits.allowedKinds.getOrElse(Set()).union(whitelist.getOrElse(Set()))
```
tysonnorris: use `Set.empty` instead of `Set()`

PR #4115, opened 2018-11-15
Ensure, that Result-ack is sent before Completion-ack.

cbickel:
## Description
On invoking an action blocking, a ResultMessage and afterwards a CompletionMessage will be sent from the invoker to the controller. The CompletionMessage is always very small. Currently those two messages are sent asynchronously. If log-collection is very fast, they are sent to Kafka at the same point in time.
This could result into the condition, that the loadbalancer receives the completion-ack before the result ack. This will make the response to the customer slower, as the controller first needs to go against Kafka.

This PR will ensure, that the completion message will be sent after the reult message has been published successfully to Kafka.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
+++ b/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
@@ -497,10 +497,10 @@ class ContainerProxy(
       }
 
     // Sending active ack. Entirely asynchronous and not waited upon.
-    if (job.msg.blocking) {
-      activation.foreach(
+    val sendResult = if (job.msg.blocking) {
+      activation.map(
         sendActiveAck(tid, _, job.msg.blocking, job.msg.rootControllerIndex, job.msg.user.namespace.uuid, false))
-    }
+    } else Future.successful(())
 
     val context = UserContext(job.msg.user)
 
@@ -530,8 +530,16 @@ class ContainerProxy(
     activationWithLogs
       .map(_.fold(_.activation, identity))
       .foreach { activation =>
-        // Sending the completionMessage to the controller asynchronously.
-        sendActiveAck(tid, activation, job.msg.blocking, job.msg.rootControllerIndex, job.msg.user.namespace.uuid, true)
+        // Sending the completionMessage to the controller asynchronously. But not before result message is sent.
+        sendResult.onComplete(
+          _ =>
+            sendActiveAck(
+              tid,
+              activation,
+              job.msg.blocking,
+              job.msg.rootControllerIndex,
+              job.msg.user.namespace.uuid,
+              true))
         // Storing the record. Entirely asynchronous and not waited upon.
         storeActivation(tid, activation, context)
       }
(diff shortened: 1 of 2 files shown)


core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -422,11 +423,12 @@ class ShardingContainerPoolBalancer(
...
-          // If the action was blocking and the Resultmessage has been received before nothing will happen here.
-          // If the action was blocking and the ResultMessage is still missing, we pass the ActivationId. With this Id,
-          // the controller will get the result out of the database.
           // If the action was non-blocking, we will close the promise here.
-          entry.promise.trySuccess(Left(aid))
+          // If the action was blocking we won't close the promise with the result to the user. The ResultMessage
+          // should have arrived earlier and already resolved it. If it didn't arrive yet (what could happen if log
+          // collection is really fast and the result is very large), it will most likely arrive in a few milliseconds.
+          // If that message really got lost, the mechanism of DB-polling will take place.
+          if (!entry.blocking) entry.promise.trySuccess(Left(aid))
```
dubee: There may be a problem with the line  `activations.remove(aid)` as well.  Since the ID is removed from the activations TrieMap, the `processResult` method may not be able to find the ID when `activations.get(aid).map` is called.

cbickel: You are right. That's a very good point.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -497,10 +497,10 @@ class ContainerProxy(
       }
 
     // Sending active ack. Entirely asynchronous and not waited upon.
-    if (job.msg.blocking) {
-      activation.foreach(
+    val sendResult = if (job.msg.blocking) {
+      activation.map(
         sendActiveAck(tid, _, job.msg.blocking, job.msg.rootControllerIndex, job.msg.user.namespace.uuid, false))
```
markusthoemmes: We shoudl reconcile any errors here into a Future.successful or use `onComplete` below. Otherwise you'll end up not sending the second message if the first failed. Unlikely, but just a teeny bit more defensive. I'd go for `onComplete` below, because that makes it more clear on the calling side, that you don't care about errors.

FWIW I like this solution a lot better than the one before.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -530,8 +530,16 @@ class ContainerProxy(
     activationWithLogs
       .map(_.fold(_.activation, identity))
       .foreach { activation =>
-        // Sending the completionMessage to the controller asynchronously.
-        sendActiveAck(tid, activation, job.msg.blocking, job.msg.rootControllerIndex, job.msg.user.namespace.uuid, true)
+        // Sending the completionMessage to the controller asynchronously. But not before result message is sent.
```
rabbah: maybe im confused by the view on github - is the active ack sent twice for blocking invokes?

cbickel:
PG3#3673 is 🔵

cbickel:
@rabbah Thanks a lot for your proposed changes :)

Using these type aliases in more places makes totally sense to me.

PR #4122, opened 2018-11-18
Protect Package Bindings from containing circular references

asteed:
## Description
We ran across an issue where a package binding can contain a reference to itself. When this happens, any attempt at reading the package results in an infinite loop, eventually leading to the controller falling over. This can occur by simply performing the following commands using the `wsk` tool.
```bash
$ wsk package create target
$ wsk package bind target binding
$ wsk package bind binding binding
```
There are two changes contained in this PR
1. Prevent allowing an incoming binding update to point to itself as a reference, which previously created an invalid data record.
2. Prevent following a package's binding reference when it points to itself which causes the infinite loop.

## Related issue and scope

- [ ] API
- [X] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
@@ -299,9 +299,14 @@ trait WhiskPackagesApi extends WhiskCollectionAPI with ReferencedEntities {
       case b: Binding =>
         val docid = b.fullyQualifiedName.toDocId
         logging.debug(this, s"fetching package '$docid' for reference")
-        getEntity(WhiskPackage.get(entityStore, docid), Some {
-          mergePackageWithBinding(Some { wp }) _
-        })
+        if (docid == wp.docid) {
+          logging.error(this, "unexpected package binding refers to itself: $docid")
+          terminate(InternalServerError)
+        } else {
+          getEntity(WhiskPackage.get(entityStore, docid), Some {
+            mergePackageWithBinding(Some { wp }) _
+          })
+        }
     } getOrElse {
       val pkg = ref map { _ inherit wp.parameters } getOrElse wp
       logging.debug(this, s"fetching package actions in '${wp.fullPath}'")
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
@@ -300,7 +300,7 @@ trait WhiskPackagesApi extends WhiskCollectionAPI with ReferencedEntities {
         val docid = b.fullyQualifiedName.toDocId
         logging.debug(this, s"fetching package '$docid' for reference")
         if (docid == wp.docid) {
-          logging.error(this, "unexpected package binding refers to itself: $docid")
+          logging.error(this, s"unexpected package binding refers to itself: $docid")
           terminate(UnprocessableEntity, Messages.packageBindingCircularReference(b.fullyQualifiedName.toString))
         } else {
           getEntity(WhiskPackage.get(entityStore, docid), Some {
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
@@ -243,7 +243,7 @@ trait WhiskPackagesApi extends WhiskCollectionAPI with ReferencedEntities {
       wp.binding map {
         // pre-existing entity is a binding, check that new binding is valid
         b =>
-          checkBinding(b.fullyQualifiedName)
+          checkBinding(binding.fullyQualifiedName)
       } getOrElse {
         // pre-existing entity is a package, cannot make it a binding
         Future.failed(RejectRequest(Conflict, Messages.packageCannotBecomeBinding))
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
@@ -242,7 +242,7 @@ trait WhiskPackagesApi extends WhiskCollectionAPI with ReferencedEntities {
     val validateBinding = content.binding map { binding =>
       wp.binding map {
         // pre-existing entity is a binding, check that new binding is valid
-        b =>
+        _ =>
           checkBinding(binding.fullyQualifiedName)
       } getOrElse {
         // pre-existing entity is a package, cannot make it a binding
@@ -301,7 +301,7 @@ trait WhiskPackagesApi extends WhiskCollectionAPI with ReferencedEntities {
         logging.debug(this, s"fetching package '$docid' for reference")
         if (docid == wp.docid) {
           logging.error(this, "unexpected package binding refers to itself: $docid")
-          terminate(InternalServerError)
+          terminate(UnprocessableEntity, Messages.packageBindingCircularReference(b.fullyQualifiedName.toString))
         } else {
           getEntity(WhiskPackage.get(entityStore, docid), Some {
             mergePackageWithBinding(Some { wp }) _
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/PackageCollection.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/PackageCollection.scala
@@ -99,7 +99,7 @@ class PackageCollection(entityStore: EntityStore)(implicit logging: Logging) ext
           val pkgDocid = binding.docid
           logging.debug(this, s"checking subject has privilege '$right' for bound package '$pkgDocid'")
           if (doc == pkgDocid) {
-            logging.error(this, "unexpected package binding refers to itself: $docid")
+            logging.error(this, s"unexpected package binding refers to itself: $doc")
             Future.failed(
               RejectRequest(
                 UnprocessableEntity,
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/PackageCollection.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/PackageCollection.scala
@@ -98,10 +98,15 @@ class PackageCollection(entityStore: EntityStore)(implicit logging: Logging) ext
           val pkgOwner = namespaces.contains(binding.namespace.asString)
           val pkgDocid = binding.docid
           logging.debug(this, s"checking subject has privilege '$right' for bound package '$pkgDocid'")
-          if (doc == pkgDocid)
-            Future.successful(true)
-          else
+          if (doc == pkgDocid) {
+            logging.error(this, "unexpected package binding refers to itself: $docid")
+            Future.failed(
+              RejectRequest(
+                UnprocessableEntity,
+                Messages.packageBindingCircularReference(binding.fullyQualifiedName.toString)))
+          } else {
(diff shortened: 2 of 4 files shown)


core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
```diff
@@ -243,7 +243,7 @@ trait WhiskPackagesApi extends WhiskCollectionAPI with ReferencedEntities {
       wp.binding map {
         // pre-existing entity is a binding, check that new binding is valid
         b =>
-          checkBinding(b.fullyQualifiedName)
+          checkBinding(binding.fullyQualifiedName)
```
asteed: this was previously checking the _existing_ binding, and not the _incoming_ binding from `content.binding`

rabbah: This also suggests that we need a test along these lines:

```
wsk package create p
wsk package bind p b
wsk package delete p
wsk package create p2
wsk package update p2 b # will fail without your patch, and should succeed with
```

rabbah: the error message for fetching a binding missing its references package is confusing - we should open an issue to also fix that.

currently returns: `The requested resource 'guest/b' does not exist.` instead of `Binding references a package that does not exist.`

core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/PackageCollection.scala
```diff
@@ -98,7 +98,10 @@ class PackageCollection(entityStore: EntityStore)(implicit logging: Logging) ext
           val pkgOwner = namespaces.contains(binding.namespace.asString)
           val pkgDocid = binding.docid
           logging.debug(this, s"checking subject has privilege '$right' for bound package '$pkgDocid'")
-          checkPackageReadPermission(namespaces, pkgOwner, pkgDocid)
+          if (doc == pkgDocid)
```
rabbah: should this be treated as an error also?

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
asteed: @rabbah I don't see how the last line (`wsk package update p2 b`) will succeed currently. As `b` is a binding already, and updating a binding to another binding was already not allowed

#### [`Packages.scala:L206`](https://github.com/apache/incubator-openwhisk/blob/master/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala#L206-L208)
```
// trying to create a new package binding that refers to another binding
case provider if provider.binding.nonEmpty =>
  Future.failed(RejectRequest(BadRequest, Messages.bindingCannotReferenceBinding))
```
While attempting to run your suggested test, I noticed that the command `wsk update p2 b` isn't accepted, and I'm wondering whether you intended to use `wsk bind p2 b` or another operation? Since `p2` is created as a package it shouldn't allow it to be updated as a binding (from my understanding). My apologies if my inference is incorrect here.
```
$ wsk -i package create p
ok: created package p
$ wsk -i package bind p b
ok: created binding b
$ wsk -i package delete p
ok: deleted package p
$ wsk -i package create p2
ok: created package p2
$ wsk -i package update p2 b
error: Invalid argument(s): b. A package name is required.
Run 'wsk --help' for usage.
$ wsk -i package bind p2 b  
error: Binding creation failed: resource already exists (code MjyhYzbZlRWsxjpmNToG4enhbK1j6EYz)
```
> the error message for fetching a binding missing its references package is confusing - we should open an issue to also fix that.

Do you mean that after p is deleted, attempting to fetch `b` is failing with that error?

rabbah: i left out a detail - i edited the cli to send `?overwrite=true` on the `POST`.

rabbah: i left out a detail - i edited the cli to send ?overwrite=true on the POST.
in other words, if this is a REST API test, you can test it.

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
```diff
@@ -299,9 +299,14 @@ trait WhiskPackagesApi extends WhiskCollectionAPI with ReferencedEntities {
       case b: Binding =>
         val docid = b.fullyQualifiedName.toDocId
         logging.debug(this, s"fetching package '$docid' for reference")
-        getEntity(WhiskPackage.get(entityStore, docid), Some {
-          mergePackageWithBinding(Some { wp }) _
-        })
+        if (docid == wp.docid) {
+          logging.error(this, "unexpected package binding refers to itself: $docid")
+          terminate(InternalServerError)
```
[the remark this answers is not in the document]
asteed: Two points about this:
1) I'm not quite sure I 100% agree with a `400 Bad Request` in this situation as there is nothing in fact wrong with the request, however there is entirely something wrong with the stored state of the server. While I also don't agree with a `500 Internal Server Error` I don't see a better alternative for either. While it's use is intended for WebDAV, it almost appears that `508 Loop Detected` would be the most accurate description :)
2) There is the attempt at performing the `get` on the object afterward, and that the test checks it is a 200 response, however you're correct that with the fix for ingest in place it would only be possible to manually craft this bad record (and I'm unable to determine how to do that given the current tests)

core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/PackageCollection.scala
asteed: The way the flow works here, this is a precursor to the call into `mergePackageWithBinding`. My rationalization here was that this is merely a permissions check, and if we're checking the children of self, we've already checked self and that passed or we'd not be here. 

I'm up to suggestions on whether this should fail, however. WDYT?

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
asteed: Did you end up being able to bind a new package to an existing binding? I was unable to do so in my original tests. The logic in `L206` would seem to indicate it would prevent that

asteed: @rabbah note that your example (with `?overwrite=true`) returns 400 as expected today as pointed out in the [source](https://github.com/apache/incubator-openwhisk/blob/master/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala#L206-L208).
```
PUT /api/v1/namespaces/guest/packages/b?overwrite=true HTTP/1.1

{"binding":{"namespace":"guest","name":"p2"}}
```
```
HTTP/1.1 400 Bad Request

{"error":"Binding references a package that does not exist.","code":"HWkcauMms0QCAFNs4mhDCrtx7EbqHBWF"}
```
 Do you still think there is another failure case we should resolve in this PR?

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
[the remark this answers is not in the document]
asteed: @rabbah WDYT?

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
rabbah: @asteed the error "Binding references a package that does not exist" is wrong since `p2` does exist. The controller tried to look up `p` which was deleted and hence the reason the error is wrong. If the error contained the missing package, it would be: 

```bash
> wsk package bind p2 b
error: Binding references a package 'p' that does not exist.
```

which is nonsensical since we're requesting to bind `p2` and `p` is no longer relevant.

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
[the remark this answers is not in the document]
mcdan: Seems like a 422 is pretty spot on "422 UNPROCESSABLE ENTITY"

https://tools.ietf.org/html/rfc4918#section-11.2

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
asteed: @rabbah Thanks, I didn't catch the incorrect error originally, as I took your comments to suggest another area where we could encounter the looping scenario.

Do you think this should be addressed in this PR or filed as a separate bug similar to (or included in) #4130 ?

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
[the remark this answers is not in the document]
asteed: I can agree on `422 Unprocessable Entity` - I'll make those changes.

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
rabbah: i think your patch may fix/hide this bug.
i'm ok with a seperate pr if the test can be written against the new code.

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
```diff
@@ -299,9 +299,14 @@ trait WhiskPackagesApi extends WhiskCollectionAPI with ReferencedEntities {
       case b: Binding =>
         val docid = b.fullyQualifiedName.toDocId
         logging.debug(this, s"fetching package '$docid' for reference")
-        getEntity(WhiskPackage.get(entityStore, docid), Some {
-          mergePackageWithBinding(Some { wp }) _
-        })
+        if (docid == wp.docid) {
+          logging.error(this, "unexpected package binding refers to itself: $docid")
```
rabbah: this won't interpolate correctly - you're missing an s"..." in the log message. 
same below.

core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/PackageCollection.scala
```diff
@@ -98,7 +98,15 @@ class PackageCollection(entityStore: EntityStore)(implicit logging: Logging) ext
           val pkgOwner = namespaces.contains(binding.namespace.asString)
           val pkgDocid = binding.docid
           logging.debug(this, s"checking subject has privilege '$right' for bound package '$pkgDocid'")
-          checkPackageReadPermission(namespaces, pkgOwner, pkgDocid)
+          if (doc == pkgDocid) {
+            logging.error(this, "unexpected package binding refers to itself: $docid")
```
rabbah: same as above.

rabbah: pushed a commit to fix.

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Packages.scala
rabbah: pushed a commit to fix.

chetanmeh:
@asteed Should we also update swagger definition for 422 status code response in package api?

Issue #4134, opened 2018-11-26
the `processAcknowledgement` may process `CompletionMessage` first for blocking incovation even the message is sent ahead of `ResultMessage`


jiangpengcheng:
## Environment details:

* local deployment
* latest master c045b420f71b170f8a0c8e416d8cb19ea2c8e571

## Steps to reproduce the issue:

1. create a simple action
```
function main(){
  return {"payload": "hello world"};
}
```

2.   invoke actions
```
for i in `seq 1 10000`; do wsk -i action invoke hello -r; sleep 1s; done
```

## Provide the expected results and outputs:

I think the expected result is not to fetch activation from database if `ResultMessage` contains that data:
```
[2018-11-26T09:31:51.910Z] [INFO] [#tid_f8c1c812c5052456ad7b026025b222da] POST /api/v1/namespaces/whisk.system/actions/hello blocking=true&result=true
[2018-11-26T09:31:51.911Z] [INFO] [#tid_f8c1c812c5052456ad7b026025b222da] [BasicAuthenticationDirective] authenticate: 789c46b1-71f6-4ed5-8c54-816aa4f8c502
[2018-11-26T09:31:51.912Z] [INFO] [#tid_f8c1c812c5052456ad7b026025b222da] [Identity] [GET] serving from cache: CacheKey(789c46b1-71f6-4ed5-8c54-816aa4f8c502) [marker:database_cacheHit_count:2]
[2018-11-26T09:31:51.915Z] [INFO] [#tid_f8c1c812c5052456ad7b026025b222da] [WhiskActionMetaData] [GET] serving from cache: CacheKey(whisk.system/hello) [marker:database_cacheHit_count:6]
[2018-11-26T09:31:51.915Z] [INFO] [#tid_f8c1c812c5052456ad7b026025b222da] [ActionsApi]  [marker:controller_blockingActivation_start:6]
[2018-11-26T09:31:51.915Z] [INFO] [#tid_f8c1c812c5052456ad7b026025b222da] [ActionsApi] action activation id: d208f976626f429388f976626fb2937a [marker:controller_loadbalancer_start:6]
[2018-11-26T09:31:51.916Z] [INFO] [#tid_f8c1c812c5052456ad7b026025b222da] [ShardingContainerPoolBalancer] posting to 'invoker0/0' with activation id 'd208f976626f429388f976626fb2937a' [marker:controller_kafka_start:7]
[...]

PR #4135, opened 2018-11-26
Ensure ResultMessage is processed

jiangpengcheng:
## Description
As `processCompeltion` and `processResult` is executed asynchronously, it's better to ensure `ResultMessage` will be processed in any case

## Related issue and scope
- [x] I opened an issue to propose and discuss this change (#4134 )

- [ ] API
- [ ] Controller
- [x] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
@@ -267,8 +267,11 @@ class ShardingContainerPoolBalancer(
         val entry = setupActivation(msg, action, invoker)
         sendActivationToInvoker(messageProducer, msg, invoker).map { _ =>
           if (msg.blocking) {
-            blockingPromises.getOrElseUpdate(msg.activationId, Promise[Either[ActivationId, WhiskActivation]]()).future
-          } else entry.promise.future
+            blockingPromises.getOrElseUpdate(msg.activationId, entry.promise)
+          } else {
+            entry.promise.trySuccess(Left(msg.activationId))
+          }
+          entry.promise.future
         }
       }
       .getOrElse {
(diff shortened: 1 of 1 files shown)


cbickel:
@jiangpengcheng Thanks a lot for tackling this issue again.
I also looked at the problem again. What do you think about the following approach of fixing this:
The promise is only used in case the activation was blocking. In case of non-blocking activations, this promise is never used.
What do you think about only closing the promise on getting the ResultMessage?

jiangpengcheng:
@cbickel 
do you mean create an extra `Promise[Either[ActivationId, WhiskActivation]]` , saved it into the new map for blocking activations, and return this `promise.future` in the `publish` method,
and for non-blocking activations, just keep the original way?

or

don't create promise for non-blocking activations, just return an `Future.successful(Left(aid))` in the `publish` method?

cbickel:
Ah wait. One more thing came into my mind. Do we still need to store the promise in the `ActivationEntry` with this change? I think it would be enough to store it in `blockingPromises`, right?

jiangpengcheng:
yes, we don't need that promise

ningyougang:
@cbickel ,can be merged?

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -422,13 +425,9 @@ class ShardingContainerPoolBalancer(
           .foreach(_.releaseConcurrent(entry.fullyQualifiedEntityName, entry.maxConcurrent, entry.memory.toMB.toInt))
         if (!forced) {
           entry.timeoutHandler.cancel()
```
rabbah: the timeout handler (which is the db poller) should be cancelled on the receipt of the active ack.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -262,9 +264,13 @@ class ShardingContainerPoolBalancer(
 
     chosen
       .map { invoker =>
-        val entry = setupActivation(msg, action, invoker)
+        setupActivation(msg, action, invoker)
         sendActivationToInvoker(messageProducer, msg, invoker).map { _ =>
-          entry.promise.future
+          if (msg.blocking) {
+            blockingPromises.getOrElseUpdate(msg.activationId, Promise[Either[ActivationId, WhiskActivation]]()).future
```
rabbah: while not likely, it's plausible the active ack is received before the promise is added to this map (which is why setup activation adds the promise before it's posted to any invoker).

PR #4136, opened 2018-11-27
Reduce invoker health action memory limit to 128MB

SugandhaAgrawal:
## Description
In order to keep the Invoker health test's action with a minimalistic memory, the value of memory limit is changed from the standard memory config to the minimum available memory. This helps in consuming the resource in minimal way possible and would suppress the Rescheduling run message occurrence due to health test.

## Related issue and scope

- [ ] API
- [ ] Controller
- [x] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/InvokerSupervision.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/InvokerSupervision.scala
@@ -276,7 +276,8 @@ object InvokerPool {
       new WhiskAction(
         namespace = healthActionIdentity.namespace.name.toPath,
         name = EntityName(s"invokerHealthTestAction${i.asString}"),
-        exec = CodeExecAsString(manifest, """function main(params) { return params; }""", None))
+        exec = CodeExecAsString(manifest, """function main(params) { return params; }""", None),
+        limits = ActionLimits(memory = MemoryLimit(MemoryLimit.minMemory)))
     }
 }
 


csantanapr:
Should we use this opportunity to set the default min to 64MB?

sven-lange-last:
@csantanapr Reducing the general limit from 128 MB to 64 MB would potentially mean a higher number of concurrent actions per invoker when keeping the same amount of action memory on the invoker. As a consequence the CPU consumption may double. Increasing the potential CPU consumption on the one hand may also require CPU capping for action on the other hand.

At the moment, the Docker container implementation does not support CPU capping. We should definitely look into this topic.

csantanapr:
I had the impression we were already calculating the cpu share by using the fraction of the memory limit proportion since we currently pass the “--cpu-shares” flag. 

So Indont follow, less memory would mean more actions, higher density per invoker, each action with small men gets smaller cpu share.

sven-lange-last:
CPU shares just makes sure that a cgroup receives a minimum amount of CPU if and only if all CPUs are 100% loaded. It has no influence on the maximum CPU share a single cgroup receives.

See https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt and https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt.

cbickel:
@csantanapr I think there are more problems that arise on going down to 64MB (As Sven already mentioned). But the current change will pick up the new minimum value immediately on changing this. So I'll merge this PR as it is for now.

PR #4137, opened 2018-11-28
Remove busy loop in case the invoker is overloaded.

cbickel:
## Description
This PR will remove the busy loop in the invoker, in case it is overloaded. Before this change, the first run message has been processed over and over again, until there was enough space in the pool for the current action. With this change, a new try to execute this action is only done, when there is some more space in the pool.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerPool.scala
```diff
@@ -99,13 +99,16 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
     // their requests and send them back to the pool for rescheduling (this may happen if "docker" operations
     // fail for example, or a container has aged and was destroying itself when a new request was assigned)
     case r: Run =>
-      // Check if the message is resent from the buffer. Only the first message on the buffer can be resent.
-      val isResentFromBuffer = runBuffer.nonEmpty && runBuffer.dequeueOption.exists(_._1.msg == r.msg)
+      // Checks if the current message is the first in the queue
+      val isFirstInQueue = runBuffer.dequeueOption.exists(_._1.msg == r.msg)
+      // A resend is only valid, if the message is the first one in the queue. Otherwise it is a result of a race condition.
+      val isValidResent = r.fromQueue && isFirstInQueue
```
markusthoemmes: Since you already switched to using a `runFirstMessageOfQueue` method, can we simplify our logic towards this:

1. On getting a run message, we always append to the buffer
2. After appending, we'll run the `runFirstMessageOfQueue` method

I imagine that that way, we don't need the `isFirstInQueue` and `isValidResent` checks, as that guarantees that we always send messages off of a queue.

(The `Run` handler could of course optimize, and bypass the queue if it knows it's empty)

WDYT?

cbickel: I also thought about this after I was finished.
The only concern with this that came into my mind was, that one more message is sent for every activation. But I don't think, that this will be a bottleneck.
I think it's worth this change to get the code more clear.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -161,7 +162,7 @@ class ContainerProxy(
 
     // cold start (no container to reuse or available stem cell container)
     case Event(job: Run, _) =>
-      implicit val transid = job.msg.transid
+      implicit val transId: TransactionId = job.msg.transid
```
rabbah: Small nit but we use sid and transid consistently throughout (no camel case) so I wouldn’t make this change (unless it’s done everywhere).

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerPool.scala
rabbah: I don’t understand the comment about a resend by a race. I may be miopic here as I’ve read only the comments so far. 

A resend (which happens when the proxy can’t schedule the message) will get queued at the back of the line.

Do you mean race in the sense that the message was simply resent to the queue?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerPool.scala
```diff
@@ -99,13 +99,23 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
...
 
-      // Only process request, if there are no other requests waiting for free slots, or if the current request is the
-      // next request to process
-      // It is guaranteed, that only the first message on the buffer is resent.
-      if (runBuffer.isEmpty || isResentFromBuffer) {
+      if (!r.isFromQueue) {
+        // If the request is not from the queue, it will be added to the queue always.
+        runBuffer = runBuffer.enqueue(r)
+        runFirstMessageOfQueue()
+      } else if (r.isFromQueue && !isFirstInQueue) {
```
sven-lange-last: If I'm not mistaken, all `Run` messages arriving at a proxy should have `fromQueue = true`. If such a `Run` message would exceed the proxies capacity, it's buffered instead of being run. If a run in the proxy fails, the proxy will destroy the container and send all buffered `Run` messages back to the pool. In that case, this code location may work on a `Run` message sent from a container and NOT being the first message enqueued. The comment states differently...

cbickel:
@sven-lange-last and me decided to not merge this yet, as it is too risky currently (before everyone leaves during christmas).
I'll close the PR for now and I'll reopen after addressing it again.

PR #4156, opened 2018-12-04
Make exec aka code max size configurable

chetanmeh:
Makes the exec hard coded 48 MB limit configurable. 

## Description
In our setups we may want to reduce the max size at controller level hence need to make it configurable. 

Note that for supporting higher value changes in other places would also be needed. See #3712 for more details

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [x] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -234,6 +234,9 @@ whisk {
         std = 1
     }
 
+    # maximum size of the action code
+    exec-size-limit = 48 m
+
     query-limit {
         max-list-limit     = 200  # max number of entities that can be requested from a collection on a list operation
         default-list-limit = 30   # default limit on number of entities returned from a collection on a list operation
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/WhiskConfig.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/WhiskConfig.scala
@@ -239,5 +239,6 @@ object ConfigKeys {
 
   val s3 = "whisk.s3"
   val query = "whisk.query-limit"
+  val execSizeLimit = "whisk.exec-size-limit"
 
 }
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/entity/Exec.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/entity/Exec.scala
@@ -229,7 +229,12 @@ protected[core] case class SequenceExecMetaData(components: Vector[FullyQualifie
 
 protected[core] object Exec extends ArgNormalizer[Exec] with DefaultJsonProtocol {
 
+  val maxSize: ByteSize = 48.MB
   val sizeLimit = loadConfigOrThrow[ByteSize](ConfigKeys.execSizeLimit)
+
+  require(
+    sizeLimit <= maxSize,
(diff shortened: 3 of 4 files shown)


rabbah:
deja vu.... didn't we do this already?

rabbah:
https://github.com/apache/incubator-openwhisk/pull/3757 and https://github.com/apache/incubator-openwhisk/pull/3835 both neither was merged. FYI @cbickel

cbickel:
I wanted to do this in addition to another problem, that I detected, but I never had the time to work on the review.
But what needs to be added to this PR before it can get merged:
- Adapt max allowed request size in ingress
- Adapt the max allowed size of requests in the controller (or at least check, that larger actions are no problem)
- Add a test, that it is at least possible to create an action if it has a size just beyond the limit (?)

If you want, you can also copy and adapt the following test to ensure, that the database is able to store actions with the configured size. https://github.com/apache/incubator-openwhisk/pull/3835/files#diff-522299b4c416d089f4c75c9189447452R207
But if you have a system test, what would be better, you don't need it, of cause.

chetanmeh:
@cbickel  For now our requirement is more about reducing this limit. Hence kept the scope small for this. To enable a simple one config switch to allow larger code size would need more effort (as you did in other  PR).

cbickel:
Ok, I agree, that the risk is much lower on only reducing the limit.
But on making it configurable, someone could also increase it and be disappointed, if it doesn't work for whatever reason.
So I'd suggest to require it to be lower or equal to 48.

But I think adapting the request size in nginx and controller should be part of this PR anyway. Because this is a protection. There is no need for requests with a size of 50MB if it is allowed to create actions with a maximum size of 10MB. This request can already be filtered out in the front door.

chetanmeh:
> So I'd suggest to require it to be lower or equal to 48.

Makes sense

> But I think adapting the request size in nginx and controller should be part of this PR anyway. 

Not sure on that part. If someone is trying to lower it then he can modify the other settings as well on his setup. For e.g. in our case we do not leverage the out of the box ansible books (and probably same things stands for kube deployments). So we can keep the default setting as is and just provide a way to make it configurable with the understanding that a system admin tweaking this also adapt other settings (possibly document it in conf file as a note)

rabbah:
@chetanmeh can you rebase this to pick up conflicting changes?

rabbah:
Oops sorry for the miscommunication - I thought that was a good suggestion and that I didn’t find a require to enforce the max size. Should be a quick addition to the config loader?

chetanmeh:
Sorry for the confusion. Added a check to ensure that configured size is no more than max size limit of 48 MB

```
Caused by: java.lang.IllegalArgumentException: requirement failed: Executable code size limit 52428800 B specified by 'whisk.exec-size-limit' should not be more than max size of 48 MB
	at scala.Predef$.require(Predef.scala:277)
	at org.apache.openwhisk.core.entity.Exec$.<init>(Exec.scala:237)
	at org.apache.openwhisk.core.entity.Exec$.<clinit>(Exec.scala)
	... 12 more
```

PR #4157, opened 2018-12-04
Make query limits configurable

chetanmeh:
Make limits related to list operation against entities configurable. 

## Description

Currently the queries related to collections (used while doing list operation) make use of hard coded limits. 

### Impact due to wsk activation poll

For example if `limit=0` in list operation then it uses 200 has the default value. This is used in `wsk activation poll` which by default generates a GET request like

```
/api/v1/namespaces/_/activations?docs=true&limit=0&since=1542422386538&skip=0
```

Currently the poll command once started "probably" sticks to start time and we have seen user leaving such poll running for days (often left in one of the numerous tabs!) which causes the system to pull similar 200 records over and over.

### CosmosDB and higher resource usage

Further in our case (using CosmosDB) which does client side merging of query results a sorted query gets executed like below

1. Initial ask to fetch 200 records
2. SDK makes call to each partition to fetch query record with limit
    - If limit > default maxItemPerPageCount use maxItemPerPageCount i.e. 100
    - If limit < then use the limit
3. Perform a merge sort on client side

In such a case at times we have seen for listing 200 records SDK fetch ~ 900 records causing higher resource usage on db side

To reduce impact for such cases I would like to have these limits configurable

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -234,6 +234,11 @@ whisk {
         std = 1
     }
 
+    query-limit {
+        max-list-limit     = 200  # max number of entities that can be requested from a collection on a list operation
+        default-list-limit = 30   # default limit on number of entities returned from a collection on a list operation
+    }
+
     mesos {
         master-url = "http://localhost:5050" //your mesos master
         master-public-url = "http://localhost:5050" // if mesos-link-log-message == true, this link will be included with the static log message (may or may not be different from master-url)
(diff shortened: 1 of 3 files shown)


chetanmeh:
Opened apache/incubator-openwhisk-cli#390 to fix the wsk poll logic to change `since` time which should also help db  as it only has to look into recent results to answer the query

PR #4158, opened 2018-12-04
Consolidate to a single tools/ow-utils docker image

dgrove-oss:
Rather than defining separate images for running shell scripts,
running ansible playbooks, running gradle tasks, etc. switch
to an uber utility image that will include wsk, wsk-deploy,
and all the software packages to execute our various scripting
languages.

Note that this image cannot be alpine based because of the DNS
race condition bug (see https://github.com/apache/incubator-openwhisk/pull/4052).
Overall image size is 884MB, which is bigger than one would like but probably
tolerable.


dgrove-oss:
With some advice from @jonpspri cut the image down to 607MB (changes squashed & force-pushed)

jonpspri:
I don't think `--no-install-recommends` made it into the current commit.  I may also have a PR in a little bit to make the `wsk` and `wskdeploy` more multi-arch friendly.

dgrove-oss:
I built images both with and without `--no-install-recommends` and it didn't change the image size.

PR #4159, opened 2018-12-05
Add additional debug information to Loadbalancer.

cbickel:
This PR will add additional information into the loadbalancer which enhances debugging. With these information it will be much easier to debug, which user and which action consumed how much memory.

## Related issue and scope

- [ ] API
- [ ] Controller
- [x] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
@@ -242,8 +242,11 @@ class ShardingContainerPoolBalancer(
   override def publish(action: ExecutableWhiskActionMetaData, msg: ActivationMessage)(
     implicit transid: TransactionId): Future[Future[Either[ActivationId, WhiskActivation]]] = {
 
+    val isBlackboxInvocation = action.exec.pull
+    val actionType = if (!isBlackboxInvocation) "non-blackbox" else "blackbox"
+
     val (invokersToUse, stepSizes) =
-      if (!action.exec.pull) (schedulingState.managedInvokers, schedulingState.managedStepSizes)
+      if (!isBlackboxInvocation) (schedulingState.managedInvokers, schedulingState.managedStepSizes)
       else (schedulingState.blackboxInvokers, schedulingState.blackboxStepSizes)
     val chosen = if (invokersToUse.nonEmpty) {
       val hash = ShardingContainerPoolBalancer.generateHash(msg.user.namespace.name, action.fullyQualifiedName(false))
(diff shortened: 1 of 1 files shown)


core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -265,6 +268,9 @@ class ShardingContainerPoolBalancer(
     chosen
       .map { invoker =>
         setupActivation(msg, action, invoker)
+        logging.info(
+          this,
+          s"activation ${msg.activationId} of $actionType-action '${msg.action.asString}' invoked by namespace '${msg.user.namespace.name.asString}' with memory-limit of ${action.limits.memory.megabytes} MB will be executed on ${invoker.instance}")
```
sven-lange-last: You are adding a new message to be logged for each activation while method `sendActivationToInvoker()` called in next line logs the activation anyway (https://github.com/apache/incubator-openwhisk/pull/4159/files#diff-f749d7d6e4e8ae0b5202a75af1136202R345).

I suggest to extend the existing message - the current proposal may have a performance impact because of the additional log messages. I understand that extending the existing message would require a change of parameters passed to `sendActivationToInvoker()`.

cbickel: I agree, that there is already a message below, but we do not have the information about memory-limit and action-kind there.
And I think it does not make sense to pass these parameters to this method only to be able to write logs.
But my suggestion would be to remove the logmessage there to avoid duplication and only keep the marker there.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -265,6 +268,9 @@ class ShardingContainerPoolBalancer(
     chosen
       .map { invoker =>
         setupActivation(msg, action, invoker)
+        logging.info(
+          this,
+          s"activation ${msg.activationId} of $actionType-action '${msg.action.asString}' invoked by namespace '${msg.user.namespace.name.asString}' with memory-limit of ${action.limits.memory.megabytes} MB will be executed on invoker${invoker.toInt}")
```
rabbah: small tweak suggested:

```
s"activation ${msg.activationId} for '${msg.action.asString}' by namespace '${msg.user.namespace.name.asString}' with memory limit ${action.limits.memory.megabytes}MB assigned to $actionType $invoker"
```

rabbah: small tweak suggested:

```
s"activation ${msg.activationId} for '${msg.action.asString}' by namespace
'${msg.user.namespace.name.asString}' with memory limit 
${action.limits.memory.megabytes}MB assigned to $actionType $invoker"
```

sven-lange-last: While I like your proposal, we should probably not print the action type (managed or blackbox) in a way that it implies it's the invoker type with the background that we have discussions to drop the separation between managed / blackbox invokers. So I suggest that the message should express that the action is either managed or blackbox.

rabbah: Fair although that discussion hasn't resolved and the message is accurate for today's status. Tweaking further for your input nonetheless.

```
s"activation ${msg.activationId} for '${msg.action.asString}' ($actionType) 
by namespace '${msg.user.namespace.name.asString}' with memory limit 
${action.limits.memory.megabytes}MB assigned to $invoker"
```

sven-lange-last:
@cbickel do you have a PG result?

cbickel:
@sven-lange-last Yes I have green PGs.

PR #4161, opened 2018-12-07
Revert "Proposing Lean OpenWhisk (#3886)"

rabbah:
This reverts commit d4a190c62f0aa1cd0bbecc62ad6b9885c70e9f6b.
The reason I'm reverting the commit is because while I can do a local path to mirror some of the changes in the main (fat?) loadbalancer, there is already a fair bit of divergence in the two load balancers because the lean version clones and owns the code and so it's prone to such slippage.

There is a cleaner organization which allows a common LB class which implements much of the logic that is shared between the two implementations. I started to do this on this branch https://github.com/apache/incubator-openwhisk/compare/master...rabbah:patch?expand=1 but it's clear it will take me a bit more time and we shouldn't break all the repos in the meantime.

I further suggest in the refactoring that the lean LB actually implements identity actors for the invoker pool health and feeds so that there's less reason to diverge or special case some of the code. It's best to have one version of this code because it's functionally critical.

@kpavel Sorry that I didn't get to review the code again more recently before it was merged, but I'm happy to help re-introduce this contribution with an improved organization as suggested here if you're agreeable.


rabbah:
For completeness -We can replay the changes in these two PRs https://github.com/apache/incubator-openwhisk/pull/4135 and https://github.com/apache/incubator-openwhisk/pull/4145 for the lean LB but I prefer not to do that, favoring the more principled approach of reducing duplication.

PR #4164, opened 2018-12-11
Add warn log when too many requests happened

ningyougang:
## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/controller/AuthorizedRouteDispatcher.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/controller/AuthorizedRouteDispatcher.scala
@@ -23,18 +23,17 @@ import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
 import scala.concurrent.Future
-
 import akka.http.scaladsl.server.Directives
 import akka.http.scaladsl.model.HttpMethod
 import akka.http.scaladsl.model.StatusCodes.Forbidden
 import akka.http.scaladsl.model.StatusCodes.NotFound
+import akka.http.scaladsl.model.StatusCodes.TooManyRequests
 import akka.http.scaladsl.server.RequestContext
 import akka.http.scaladsl.server.RouteResult
 import akka.http.scaladsl.model.StatusCodes.InternalServerError
 import akka.http.scaladsl.server.Directive1
-
 import org.apache.openwhisk.core.entitlement.Collection
-import org.apache.openwhisk.common.TransactionId
+import org.apache.openwhisk.common.{Logging, TransactionId}
 import org.apache.openwhisk.core.entitlement._
 import org.apache.openwhisk.core.entitlement.Resource
 import org.apache.openwhisk.core.entity._
@@ -64,6 +63,8 @@ trait BasicAuthorizedRouteProvider extends Directives {
   /** Route directives for API. The methods that are supported on entities. */
   protected lazy val entityOps = get
 
+  protected implicit val logging: Logging
+
   /** JSON response formatter. */
   import RestApiCommons.jsonDefaultResponsePrinter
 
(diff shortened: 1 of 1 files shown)


ningyougang:
When too many request happened, through add above log, user can see `which action leads to that and the detail too many request RejectRequst log`

core/controller/src/main/scala/org/apache/openwhisk/core/controller/AuthorizedRouteDispatcher.scala
```diff
@@ -64,6 +65,8 @@ trait BasicAuthorizedRouteProvider extends Directives {
   /** Route directives for API. The methods that are supported on entities. */
   protected lazy val entityOps = get
 
+  private val logger = new PrintStreamLogging()
```
markusthoemmes: Please never use `PrintStreamLogging` in production code. It has severe impact on performance. We always use `AkkaLogging` there.

core/controller/src/main/scala/org/apache/openwhisk/core/controller/AuthorizedRouteDispatcher.scala
```diff
@@ -86,7 +89,11 @@ trait BasicAuthorizedRouteProvider extends Directives {
               case NotFound =>
                 handleEntitlementFailure(
                   RejectRequest(NotFound, Some(ErrorResponse(Messages.resourceDoesntExist(resource.fqname), transid))))
-              case _ => handleEntitlementFailure(t)
+              case code =>
+                if (code == TooManyRequests) {
+                  logger.warn(this, s"${resource.id} invoked failed: ${t}")
+                }
```
markusthoemmes: Since  we don't have a logger at hand here:

1. Should we pass a logger into the `authorizeAndDispatch` method?
2. Can we log outside of the method by inspecting it's return value?

core/controller/src/main/scala/org/apache/openwhisk/core/controller/AuthorizedRouteDispatcher.scala
ningyougang: Already applied AkkaLogging

core/controller/src/main/scala/org/apache/openwhisk/core/controller/AuthorizedRouteDispatcher.scala
ningyougang: Just add `protected implicit val logging: Logging` in `trait BasicAuthorizedRouteProvider` is enough,
because the subclass will implement the logging.

ningyougang:
Regarding `Can you elaborate on the lack of information you have with the current logs?`
it is better to output below warning log
```
[2018-12-11T17:13:31.486Z] [WARN] [#tid_qlbXZYpWuI9uH5UjR5FzXi9AGge2TyPU] [ActionsApi] Unable to invoke action guest/hello, because namespace:guest has exceeded its throttle limit, detail info: RejectRequest(429 Too Many Requests) Too many concurrent requests in flight (count: 0, allowed: 2)
```

markusthoemmes:
@ningyougang that feels like a Debug level log then, certainly not a warning since nothing is wrong with the system in that case?

ningyougang:
@markusthoemmes , yes, you are right, just a debug level log, `this log is saved to controllerN_logs.log` have some benefits, for example. when openwhisk admin received the log alering, he will do some configuration tuning according to `the log alering info`

core/controller/src/main/scala/org/apache/openwhisk/core/controller/AuthorizedRouteDispatcher.scala
rabbah: There was... it's now a debug level statement. @ningyougang does this not suffice? (we can refactor the debug statement so that the concurrent throttle check emits a more specific messages.)

https://github.com/apache/incubator-openwhisk/blob/be1e3a19c02956c9be85023a0bb0ff399c21444d/core/controller/src/main/scala/org/apache/openwhisk/core/entitlement/ActivationThrottler.scala#L41-L50

rabbah: I see. It’s possible to thread the resource through to the rate throttler which would have the name of the action if it’s important.

rabbah:
i'm conflicted on this PR - if this is a debug level message did you consider also looking at the nginx logs which have the status codes in them --- it just feels unnecessary but I'm sympathetic to the goal of adding the debug log message. I can be convinced to take it.

rabbah:
@markusthoemmes can you clear your review or weigh in so we can make a decision and move ahead.

markusthoemmes:
As noted in my comments above, I don't love pulling a logger into this class. As you say, it seems unnecessary. We should be able to log this somewhere outside of this class.

dgrove-oss:
Is this still an active PR?

ningyougang:
Close it

PR #4165, opened 2018-12-11
Update to Kamon 1.1.3 from 0.6 series

chetanmeh:
Updates Kamon to 1.x release (1.1.3) from current 0.6 version

## Description

Going forward I would like to enable metrics reporting via Prometheus. [Kamon Prometheus][1] support was added with Kamon 1.0 series. Hence the need to update Kamon

[Kamon 1.0][2] was a major release involving some breaking changes which needs some [migration work][3]

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

[1]: https://kamon.io/documentation/1.x/reporters/prometheus/
[2]: https://kamon.io/teamblog/2018/01/18/kamon-1.0.0-is-out/
[3]: https://kamon.io/documentation/1.x/recipes/migrating-from-kamon-0.6/

### DIFF
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -19,7 +19,11 @@ akka.http {
 
 #kamon related configuration
 kamon {
-
+    environment {
+      # Identifier for this service. For keeping it backward compatible setting to natch previous
+      # statsd name
+      service = "openwhisk-statsd"
+    }
     metric {
         tick-interval = 1 second
     }
@@ -38,13 +42,6 @@ kamon {
             actor      =  [ "*" ]
             trace      =  [ "*" ]
             dispatcher =  [ "*" ]
-        }
-
-        simple-metric-key-generator {
-            # Application prefix for all metrics pushed to StatsD. The default namespacing scheme for metrics follows
-            # this pattern:
-            #    application.host.entity.entity-name.metric-name
-            application = "openwhisk-statsd"
         }
     }
 
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -45,9 +45,9 @@ kamon {
         }
     }
 
-    modules {
-        kamon-statsd.auto-start = yes
-    }
+    reporters = [
+        "kamon.statsd.StatsDReporter"
+    ]
 }
 
 whisk {
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -43,6 +43,8 @@ kamon {
             trace      =  [ "*" ]
             dispatcher =  [ "*" ]
         }
+
+        metric-key-generator = org.apache.openwhisk.common.WhiskStatsDMetricKeyGenerator
     }
 
     reporters = [
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -20,9 +20,11 @@ package org.apache.openwhisk.common
 import java.io.PrintStream
 import java.time.{Clock, Instant, ZoneId}
 import java.time.format.DateTimeFormatter
+
 import akka.event.Logging._
 import akka.event.LoggingAdapter
 import kamon.Kamon
+import kamon.statsd.{MetricKeyGenerator, SimpleMetricKeyGenerator}
 import org.apache.openwhisk.core.entity.ControllerInstanceId
 
 trait Logging {
@@ -243,6 +245,22 @@ object MetricEmitter {
    */
   private def createName(name: String, metricType: String) = {
     s"$metricType.$name"
+  }
+}
+
+/**
+ * Name generator to make names compatible to pre Kamon 1.0 logic. Statsd reporter "normalizes"
+ * the key name by replacing all `.` with `_`. Pre 1.0 the metric category was added by Statsd
+ * reporter itself. However now we pass it explicitly. So to retain the pre 1.0 name we need to replace
+ * normalized name with one having category followed by `.` instead of `_`
+ */
+class WhiskStatsDMetricKeyGenerator(config: com.typesafe.config.Config) extends MetricKeyGenerator {
+  val simpleGen = new SimpleMetricKeyGenerator(config)
+  override def generateKey(name: String, tags: Map[String, String]): String = {
+    val key = simpleGen.generateKey(name, tags)
+    if (key.contains(".counter_")) key.replace(".counter_", ".counter.")
+    else if (key.contains(".histogram_")) key.replace(".histogram_", ".histogram.")
+    else key
   }
 }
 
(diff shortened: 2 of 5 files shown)


common/scala/src/main/resources/application.conf
```diff
@@ -19,7 +19,11 @@ akka.http {
 
 #kamon related configuration
 kamon {
-
+    environment {
+      # Identifier for this service. For keeping it backward compatible setting to natch previous
+      # statsd name
+      service = "openwhisk-statsd"
```
chetanmeh: Setting this to `openwhisk-statsd` for ensuring that new metric names are backward compatible with previous names as now reporters use this tag to determine the service name and do not expose reporter specific setting

chetanmeh:
Seems like Kamon has changed the metric naming convention. Earlier it used to include the metric type/category as part of name like `openwhisk-statsd.invoker0.counter.database_cacheMiss_count` and now its dropping that which lead to error in our case like 

```
[2018-12-11T15:05:44.284Z] [ERROR] Cannot define metric [database_getDocument_error] as a [Counter], it is already defined as [Histogram] 
java.lang.RuntimeException: Cannot define metric [database_getDocument_error] as a [Counter], it is already defined as [Histogram] 
	at scala.sys.package$.error(package.scala:26)
	at kamon.metric.MetricRegistry.lookupMetric(MetricRegistry.scala:83)
	at kamon.metric.MetricRegistry.counter(MetricRegistry.scala:48)
	at kamon.Kamon$.counter(Kamon.scala:82)
	at kamon.metric.MetricLookup.counter(MetricLookup.scala:41)
	at kamon.metric.MetricLookup.counter$(MetricLookup.scala:40)
	at kamon.Kamon$.counter(Kamon.scala:32)
	at org.apache.openwhisk.common.MetricEmitter$.emitCounterMetric(Logging.scala:222)
```

So we would probably need to do that on our side

chetanmeh:
Discussed this with @ivantopo on [gitter](https://gitter.im/kamon-io/Kamon?at=5c0f880a80986419d572ba8f) and it seems we would need to take care of the naming part in our side as there is a conceptual change in way Kamon modes metrics. 

Would modify `MetricEmitter` to include metric type

chetanmeh:
@vvraskin Can you review and also run a PG for this. There should not be any change in names of metrics so this change should be backward compatible

vvraskin:
I needed to backport the changes to our billing aggregator. Retesting...

vvraskin:
For some reason the metrics do not show up on statsd on our test environment. I will dig deeper on my local environment when I find time later. 
Just to double check you were able to see controller/invoker metrics on statsd? for example by telneting to the admin port @chetanmeh

chetanmeh:
@vvraskin Looks like my earlier test used old docker images somehow. I was able to reproduce the issue and now pushed an update which should fix this. It appears that now we need to invoke `Kamon.loadReportersFromConfig()` for it to load modules which replaces its earlier auto load.

Please try again with updated PR

rabbah:
@vvraskin assigning to you assuming you can approve and merge this.

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
```diff
@@ -210,32 +210,43 @@ object LogMarkerToken {
...
     }
   }
+
+  /**
+   * Kamon 1.0 onwards does not include the metric type in the metric name which cause issue
+   * for us as we use same metric name for counter and histogram. So to be backward compatible we
+   * need to prefix the name with type
+   */
+  private def createName(name: String, metricType: InstrumentType) = {
+    s"${metricType.name.toLowerCase}.$name"
```
rabbah: can/should the name be lower cases once when con'sing  the `metricType` instance? (looks like this can be lowercased repeatedly here otherwise from a cursory glance.)

chetanmeh: Currently the whole metric lookup flow gets invoked for each metric updated. To avoid churn it would be better to have direct reference to metric instance and then increment that. I was thinking to optimize this in a subsequent PR by storing metric instance in `LogMarkerToken` itself and avoid this whole logic on critical path

rabbah: 👍

chetanmeh: Opened #4203 to track this

vvraskin:
Thanks for adding the changes. I've updated the downstream and started another PG, will get back once it is through

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
```diff
@@ -210,32 +210,43 @@ object LogMarkerToken {
...
       }
     }
   }
+
+  /**
+   * Kamon 1.0 onwards does not include the metric type in the metric name which cause issue
+   * for us as we use same metric name for counter and histogram. So to be backward compatible we
+   * need to prefix the name with type
+   */
+  private def createName(name: String, metricType: InstrumentType) = {
```
vvraskin: It seems that this method is not enough to fill the backward compatibility gap... The dots in the resulting metric name are replaced by underscores, e.g. instead of `x.invoker0.counter.y` we see `x.invoker0.counter_y`. 

One possibility to overcome this is to implement a custom metric name generator - `kamon.statsd.MetricKeyGenerator`, I'm not sure how much overhead that would be. Perhaps there are some other options I'm not aware of.

chetanmeh: So I see following difference 

* Pre 1.0 - `openwhisk-statsd.controller0.counter.database_queryView_start`
* Post 1.0 - `openwhisk-statsd.controller0.counter_database_queryView_start`

So one dot gets converted to underscore

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
```diff
@@ -210,32 +212,56 @@ object LogMarkerToken {
...
+ * normalized name with one having category followed by `.` instead of `_`
+ */
+class WhiskStatsDMetricKeyGenerator(config: com.typesafe.config.Config) extends MetricKeyGenerator {
+  val simpleGen = new SimpleMetricKeyGenerator(config)
+  override def generateKey(name: String, tags: Map[String, String]): String = {
+    val key = simpleGen.generateKey(name, tags)
+    if (key.contains(".counter_")) key.replace(".counter_", ".counter.")
+    else if (key.contains(".histogram_")) key.replace(".histogram_", ".histogram.")
+    else key
+  }
```
chetanmeh: @vvraskin Add this custom key generator which would ensure that names are backward compatible. This is invoked at reporting time so should not add much overhead in critcal path

PR #4176, opened 2018-12-13
additional container factory dns configuration

dgrove-oss:
Allow configuration of the --dns-search and --dns-opt (--dns-option)
arguments to the docker run command of the DockerContainerFactory.
Motivated by better support of event providers in Kubernetes when
using the DockerContainerFactory.

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
@@ -27,6 +27,8 @@ import scala.concurrent.Future
 
 case class ContainerArgsConfig(network: String,
                                dnsServers: Seq[String] = Seq.empty,
+                               dnsSearch: Seq[String] = Seq.empty,
+                               dnsOptions: Seq[String] = Seq.empty,
                                extraArgs: Map[String, Set[String]] = Map.empty)
 
 case class ContainerPoolConfig(userMemory: ByteSize, concurrentPeekFactor: Double, akkaClient: Boolean) {
--- a/core/invoker/src/main/resources/application.conf
+++ b/core/invoker/src/main/resources/application.conf
@@ -28,6 +28,7 @@ whisk {
       inspect: 1 minute
       pause: 10 seconds
       unpause: 10 seconds
+      version: 10 seconds
     }
   }
 
@@ -68,7 +69,10 @@ whisk {
   # args for 'docker run' to use
   container-factory.container-args {
     network: bridge
+    # See https://docs.docker.com/config/containers/container-networking/#dns-services for documentation of dns-*
     dns-servers: []
+    dns-search: []
+    dns-options: []
     extra-args: {}   # to pass additional args to 'docker run'; format is `{key1: [v1, v2], key2: [v1, v2]}`
   }
 
--- a/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerClient.scala
+++ b/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerClient.scala
@@ -27,7 +27,7 @@ import akka.actor.ActorSystem
 import scala.collection.concurrent.TrieMap
 import scala.concurrent.blocking
 import scala.concurrent.ExecutionContext
-import scala.concurrent.Future
+import scala.concurrent.{Await, Future}
 import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
@@ -61,6 +61,7 @@ case class DockerClientTimeoutConfig(run: Duration,
                                      ps: Duration,
                                      pause: Duration,
                                      unpause: Duration,
+                                     version: Duration,
                                      inspect: Duration)
 
 /**
@@ -97,6 +98,20 @@ class DockerClient(dockerHost: Option[String] = None,
     val host = dockerHost.map(host => Seq("--host", s"tcp://$host")).getOrElse(Seq.empty[String])
     Seq(dockerBin) ++ host
   }
+
+  // Invoke docker CLI to determine client version.
+  // If the docker client version cannot be determined, an exception will be thrown and instance initialization will fail.
+  // Rationale: if we cannot invoke `docker version` successfully, it is unlikely subsequent `docker` invocations will succeed.
+  protected def getClientVersion(): String = {
+    val vf = executeProcess(dockerCmd ++ Seq("version", "--format", "{{.Client.Version}}"), config.timeouts.version)
+      .andThen {
+        case Success(version) => log.info(this, s"Detected docker client version $version")
+        case Failure(e) =>
+          log.error(this, s"Failed to determine docker client version: ${e.getClass} - ${e.getMessage}")
+      }
+    Await.result(vf, 2 * config.timeouts.version)
+  }
+  val clientVersion: String = getClientVersion()
 
   protected val maxParallelRuns = config.parallelRuns
   protected val runSemaphore =
@@ -196,6 +211,13 @@ class DockerClient(dockerHost: Option[String] = None,
 }
 
 trait DockerApi {
+
+  /**
+   * The version number of the docker client cli
+   *
+   * @return The version of the docker client cli being used by the invoker
+   */
+  def clientVersion: String
 
   /**
    * Spawns a container in detached mode.
--- a/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala
+++ b/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala
@@ -65,6 +65,8 @@ object DockerContainer {
              environment: Map[String, String] = Map.empty,
              network: String = "bridge",
              dnsServers: Seq[String] = Seq.empty,
+             dnsSearch: Seq[String] = Seq.empty,
+             dnsOptions: Seq[String] = Seq.empty,
              name: Option[String] = None,
              useRunc: Boolean = true,
              dockerRunParameters: Map[String, Set[String]])(implicit docker: DockerApiWithFileAccess,
@@ -82,6 +84,8 @@ object DockerContainer {
       case (key, valueList) => valueList.toList.flatMap(Seq(key, _))
     }
 
+    // NOTE: --dns-option on modern versions of docker, but is --dns-opt on docker 1.12
+    val dnsOptString = if (docker.clientVersion.startsWith("1.12")) { "--dns-opt" } else { "--dns-option" }
     val args = Seq(
       "--cpu-shares",
       cpuShares.toString,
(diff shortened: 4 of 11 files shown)


dgrove-oss:
See https://github.com/apache/incubator-openwhisk-deploy-kube/pull/392 for how this would be used.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
```diff
@@ -27,6 +27,8 @@ import scala.concurrent.Future
 
 case class ContainerArgsConfig(network: String,
                                dnsServers: Seq[String] = Seq.empty,
+                               dnsSearch: Seq[String] = Seq.empty,
```
chetanmeh: Going forward we should look into opening up the docker argument configuration in a generic way such that any argument can be configured without requiring code change

core/invoker/src/main/resources/application.conf
```diff
@@ -69,6 +69,8 @@ whisk {
   container-factory.container-args {
     network: bridge
     dns-servers: []
+    dns-search: []
```
chetanmeh: May be add a link in comment here to https://docs.docker.com/config/containers/container-networking/#dns-services

dgrove-oss: Thanks.  Added link.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
dgrove-oss: I think I could have also done this via extra-args, but since dnsServers was already explicit it seemed better to make the other two dns arguments explicit as well.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -93,6 +95,9 @@ object DockerContainer {
       network) ++
       environmentArgs ++
       dnsServers.flatMap(d => Seq("--dns", d)) ++
+      dnsSearch.flatMap(d => Seq("--dns-search", d)) ++
+      // NOTE: --dns-option on modern versions of docker, but is --dns-opt on docker 1.12
+      dnsOptions.flatMap(d => Seq("--dns-opt", d)) ++
```
sven-lange-last: We need to be very careful here. While I understand that OW officially supports Docker 1.12, I know that higher Docker versions are used as well.

I'm not in favour of using the `--dns-opt` option here although we already know it won't work in higher versions. This is weak point that will definitely break in future or when using a higher Docker version today.

dgrove-oss: It annoyingly needs to match the version of the docker client embedded in the invoker Dockerfile.  I could add a comment to the Dockerfile with a pointer to change this line of code when the version is (finally) updated.  I thought about making this more parameterized, but that seemed like overkill.

I do need to get the ndots:5 option into the user actions resolv.conf file to enable user actions to invoke kube services, so we do need to do something.

dgrove-oss:
@sven-lange-last -- How should we progress on this PR?   
1. I could reopen the dev list discussion on upgrading the docker version in the invoker image to something more modern.
2. I could further parametrize this PR so that the invoker was aware of the exact version of the docker client it was talking to and use that to pick the right command line option to to run
3. Other ideas???

markusthoemmes:
This might be a little far-fetched, but anyways: The docker REST API has great backward compatibility IIRC. Would it help our docker client trauma if we switched to using the REST API over a local binary here?

dgrove-oss:
It is easy to read DOCKER_VERSION from the environment and select the option accordingly.  I've added a commit with that check to try to progress on this PR.

sven-lange-last:
While using the `DOCKER_VERSION` variable should basically work, we have to keep in mind that this variable is not generally defined on systems where a `docker` cli client is installed. It's just defined in the OpenWhisk invoker image - see here: https://github.com/apache/incubator-openwhisk/blob/master/core/invoker/Dockerfile#L8.

For this reason, if somebody should use the Docker container factory in an invoker that is not running in a container based on that image, we cannot warrant the variable is defined.

In order to have a more robust solution, I suggest to consider using this command to determine the `docker` cli client version:

```
# docker version --format '{{.Client.Version}}'
17.10.0-ce
```

This requires more code changes but should be much safer. The version info is required when creating a new Docker container, i.e. in `DockerContainer.create()`. Said method receives a `DockerApiWithFileAccess` instance as implicit parameter so that we can extend the `DockerApi` trait and its implementation to determine the `docker` cli command version. It would be good to make version determination a singleton such that the command only needs to be run once. Otherwise, container creation would get slower.

dgrove-oss:
Changed PR as suggested by @sven-lange-last to use `docker version`.

dgrove-oss:
PG2 4068 🏃

dgrove-oss:
PG1 3846 -- LGTM.  A handful of unrelated failures.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerClient.scala
```diff
@@ -98,6 +99,13 @@ class DockerClient(dockerHost: Option[String] = None,
     Seq(dockerBin) ++ host
   }
 
+  // Determine the version of the docker client cli.  Failure to determine the client
+  // version implies a failure to initialize this instance of DockerClient.
+  protected def getClientVersion(): String = {
+    stringSeqToProcess(dockerCmd ++ Seq("version", "--format", "{{.Client.Version}}")).!!
+  }
+  val clientVersion: String = getClientVersion()
```
sven-lange-last: Sorry for being picky here although you already followed my proposal to use the Docker CLI command to determine the version...

Documentation of the `!!` method states: "If the exit code is non-zero, an exception is thrown.". In other words, the `getClientVersion()` invocation may fail with an exception which is uncaught. As a consequence creation of the `DockerClient` instance as part of the invoker initialisation may fail due to the exception. That's a rather high risk for just determining the Docker version.

In addition, `!!` writes the executed command's stderr to the console instead of capturing it. So we have no good diagnostic info available in case of a failure.

I suggest to make `getClientVersion()` fail-safe and return an `Option[String]`: it's `None` upon failure indicating we don't know the version and the version string on success. So `clientVersion` also needs to be an `Option[String]`.

On failure, `getClientVersion()` needs to log an error such that we know about the problem. Probably, we should use the `executeProcess` method instead of `!!`. It already comes with full output capturing and timeout handling.

The rest of the code needs to deal with an unknown version gracefully. If that's not possible, we have to throw a meaningful exception and stop invoker init with a reasonable error message.

core/invoker/src/main/resources/application.conf
```diff
@@ -68,7 +68,10 @@ whisk {
   # args for 'docker run' to use
   container-factory.container-args {
     network: bridge
+    # See https://docs.docker.com/config/containers/container-networking/#dns-services for documentation of dns-*
     dns-servers: []
+    dns-search: []
+    dns-option: []
```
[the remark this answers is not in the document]
dgrove-oss: fixed

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerClient.scala
dgrove-oss: Rewrote to use `executeProcess` and have proper logging of success/failure of command.  As discussed on slack, if we can't figure out the docker cli version it is pretty likely subsequent docker cli operations will fail.  Therefore didn't introduce an Option[String]; left as fail-fast during init instead.

dgrove-oss:
PG1 3865 - ok (on par with other runs without this change)

core/invoker/src/main/resources/application.conf
```diff
@@ -28,6 +28,7 @@ whisk {
       inspect: 1 minute
       pause: 10 seconds
       unpause: 10 seconds
+      version: 1 minute
```
sven-lange-last: While I think it's a great idea to add a timeout for the version command, I also think that a minute is far too long. It's only run once during invoker startup and should never be run again because the version cannot change while the invoker is running.

Even `pause` and `unpause` are limited to 10 seconds which can be tight on a loaded invoker.

I suggest to use at most 10 seconds here. Even in a slow Travis env, this should never take longer.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerClient.scala
```diff
@@ -98,6 +99,20 @@ class DockerClient(dockerHost: Option[String] = None,
...
+  // Failure to determine the client version is a fatal error, because it indicates a high probability
+  // that subsequent docker cli commands will also fail.
+  protected def getClientVersion(): String = {
+    val vf = executeProcess(dockerCmd ++ Seq("version", "--format", "{{.Client.Version}}"), config.timeouts.version)
+      .andThen {
+        case Success(version) => log.info(this, s"Detected docker client version $version")
+        case Failure(e) =>
+          log.error(this, s"Failed to determine docker client version: ${e.getClass} - ${e.getMessage}")
+      }
+    Await.result(vf, Duration.Inf) // The timeout in executeProcess ensures progress; don't create racy double timeout in Await
```
sven-lange-last: I suggest to wait for at most 2 * `config.timeouts.version`. This should never be a race condition and still ensure we have an upper bound.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerClient.scala
```diff
@@ -98,6 +99,20 @@ class DockerClient(dockerHost: Option[String] = None,
     Seq(dockerBin) ++ host
   }
 
+  // Determine the version of the docker client cli.
+  // Failure to determine the client version is a fatal error, because it indicates a high probability
+  // that subsequent docker cli commands will also fail.
+  protected def getClientVersion(): String = {
```
sven-lange-last: I suggest to be more explicit here and state that `getClientVersion()` will throw an exception on failure.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerClient.scala
sven-lange-last: Thanks a lot for your rewrite.

I suggest to add a comment here that a failure to determine the Docker version will lead to an exception when creating an instance of class `DockerClient`.

core/invoker/src/main/resources/application.conf
dgrove-oss: done

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerClient.scala
dgrove-oss: done

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerClient.scala
dgrove-oss: done

dgrove-oss:
Thanks for the review @sven-lange-last.  

I updated to address your comments & then did a squash + rebase to latest master to tidy up for the merge.  I was not planning to run another PG, but I can do one if you think it is necessary.

PR #4185, opened 2018-12-17
recreate http client on resume()

tysonnorris:
In testing concurrent actions (multiple activations in same container) we noticed that when a container is used after `Container.suspend()`, the http connection may be re-created multiple times (and may have side affects of losing connections each time a new one is created.

## Description

This PR addresses this by:
- using a separate function for creating the connection - this will be invoked first during /init
- same function will be invoked again during `Container.resume()` (which means that subclasses must invoke `super.resume()`)

This way the subsequent + multiple calls to /run will arrive after connection has already been recreated.
I extended the ContainerProxyTests to assert connection state (exists or not) after `suspend()` and after `resume()`.
## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/Container.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/Container.scala
@@ -62,6 +62,11 @@ object Container {
     loadConfigOrThrow[ContainerPoolConfig](ConfigKeys.containerPool)
 }
 
+/**
+ * Abstraction for Container operations.
+ * Container manipulation (specifically suspend/resume/destroy) is NOT thread-safe and MUST be synchronized by caller.
+ * Container access (specifically run) is thread-safe (e.g. for concurrent activation processing).
+ */
 trait Container {
 
   implicit protected val as: ActorSystem
@@ -73,7 +78,11 @@ trait Container {
   /** HTTP connection to the container, will be lazily established by callContainer */
   protected var httpConnection: Option[ContainerClient] = None
 
-  /** Stops the container from consuming CPU cycles. */
+  /** maxConcurrent+timeout are cached during first init, so that resuming connections can reference */
+  protected var containerHttpMaxConcurrent: Int = 1
+  protected var containerHttpTimeout: FiniteDuration = 60.seconds
+
+  /** Stops the container from consuming CPU cycles. NOT thread-safe - caller must synchronize. */
   def suspend()(implicit transid: TransactionId): Future[Unit] = {
     //close connection first, then close connection pool
     //(testing pool recreation vs connection closing, time was similar - so using the simpler recreation approach)
@@ -82,8 +91,11 @@ trait Container {
     closeConnections(toClose)
   }
 
-  /** Dual of halt. */
-  def resume()(implicit transid: TransactionId): Future[Unit]
+  /** Dual of halt. NOT thread-safe - caller must synchronize.*/
+  def resume()(implicit transid: TransactionId): Future[Unit] = {
+    httpConnection = Some(openConnections(containerHttpTimeout, containerHttpMaxConcurrent))
+    Future.successful({})
+  }
 
   /** Obtains logs up to a given threshold from the container. Optionally waits for a sentinel to appear. */
   def logs(limit: ByteSize, waitForSentinel: Boolean)(implicit transid: TransactionId): Source[ByteString, Any]
@@ -101,7 +113,8 @@ trait Container {
       LoggingMarkers.INVOKER_ACTIVATION_INIT,
       s"sending initialization to $id $addr",
       logLevel = InfoLevel)
-
+    containerHttpMaxConcurrent = maxConcurrent
+    containerHttpTimeout = timeout
     val body = JsObject("value" -> initializer)
     callContainer("/init", body, timeout, maxConcurrent, retry = true)
       .andThen { // never fails
@@ -132,7 +145,7 @@ trait Container {
       }
   }
 
-  /** Runs code in the container. */
+  /** Runs code in the container. Thread-safe - caller may invoke concurrently for concurrent activation processing. */
   def run(parameters: JsObject, environment: JsObject, timeout: FiniteDuration, maxConcurrent: Int)(
     implicit transid: TransactionId): Future[(Interval, ActivationResponse)] = {
     val actionName = environment.fields.get("action_name").map(_.convertTo[String]).getOrElse("")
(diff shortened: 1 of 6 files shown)


tests/src/test/scala/org/apache/openwhisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -1078,12 +1089,20 @@ class ContainerProxyTests
 
     def runCount = atomicRunCount.get()
     override def suspend()(implicit transid: TransactionId): Future[Unit] = {
+      println("suspending!!!")
```
markusthoemmes: Please remove that extranous logline.

tests/src/test/scala/org/apache/openwhisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -1078,12 +1089,20 @@ class ContainerProxyTests
...
       suspendCount += 1
-      super.suspend()
+      val s = super.suspend()
+      //verify that httpconn is closed
+      httpConnection should be(None)
+      s
     }
-    def resume()(implicit transid: TransactionId): Future[Unit] = {
+    override def resume()(implicit transid: TransactionId): Future[Unit] = {
+      println("resuming!!!")
```
markusthoemmes: Same.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/Container.scala
```diff
@@ -73,6 +73,10 @@ trait Container {
   /** HTTP connection to the container, will be lazily established by callContainer */
   protected var httpConnection: Option[ContainerClient] = None
 
+  /** maxConcurrent+timeout are cached during first init, so that resuming connections can reference */
+  protected var containerHttpMaxConcurrent: Int = 1
+  protected var containerHttpTimeout: FiniteDuration = 60.seconds
```
markusthoemmes: I know this was the case before, but we should add a comment to the trait stating that it's not thread-safe and MUST the caller must ensure synchronization (which we do by using an actor)

tysonnorris:
@markusthoemmes any other comments?

tysonnorris:
@markusthoemmes I think this is ready now

tests/src/test/scala/org/apache/openwhisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -1079,11 +1076,17 @@ class ContainerProxyTests
...
+      httpConnection should be(None)
+      s
     }
-    def resume()(implicit transid: TransactionId): Future[Unit] = {
+    override def resume()(implicit transid: TransactionId): Future[Unit] = {
       resumeCount += 1
-      Future.successful(())
+      val r = super.resume()
+      //verify that httpconn is recreated
+      httpConnection should be('defined)
```
markusthoemmes: Please await the futures of either resume and suspend before checking that the client is there. Right now this is not an issue but once the implementation of the trait changes we're prone to heisenbugs.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/kubernetes/KubernetesContainer.scala
```diff
@@ -104,7 +104,8 @@ class KubernetesContainer(protected[core] val id: ContainerId,
     super.suspend().flatMap(_ => kubernetes.suspend(this))
   }
 
-  def resume()(implicit transid: TransactionId): Future[Unit] = kubernetes.resume(this)
+  override def resume()(implicit transid: TransactionId): Future[Unit] =
+    kubernetes.resume(this).map(_ => super.resume())
```
markusthoemmes: This should be a `flatMap` to also "wait" for the future returned by super.resume.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -174,8 +174,9 @@ class DockerContainer(protected val id: ContainerId,
   override def suspend()(implicit transid: TransactionId): Future[Unit] = {
     super.suspend().flatMap(_ => if (useRunc) runc.pause(id) else docker.pause(id))
   }
-  def resume()(implicit transid: TransactionId): Future[Unit] =
-    if (useRunc) { runc.resume(id) } else { docker.unpause(id) }
+  override def resume()(implicit transid: TransactionId): Future[Unit] = {
+    if (useRunc) { runc.resume(id) } else { docker.unpause(id) }.map(_ => super.resume())
```
markusthoemmes: flatMap

rabbah: when in doubt refactor...

markusthoemmes: good question @sven-lange-last. Better safe than sorry and make an interim value or put parentheses around the if expression.

tysonnorris: I don't think it's needed, but agree it will at least be more clear. Also changed to flatMap

tysonnorris:
Thanks for the feedback, it is always welcome.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/Container.scala
```diff
@@ -204,6 +209,17 @@ trait Container {
         RunResult(Interval(started, finished), response)
       }
   }
+  private def openConnections(timeout: FiniteDuration, maxConcurrent: Int) = {
+    if (Container.config.akkaClient) {
```
ddragosd: I forgot what we're still maintaining the Akka Client and the Apache Client, since the latter was found in https://github.com/apache/incubator-openwhisk/pull/3812 to perform poorly.

rabbah: are we ready to move to the akka client exclusively? there were some reservations about making sure we didn't get bit by some akka issues as in the past.

tysonnorris: I'm ready 😄 
FWIW, we have been using akka client exclusively in prod, with concurrency enabled, since December.
Tests are already using it as well. 
I'm not sure if there are others that are interested in testing it more, but it would be great to retire the apache client IMHO.

markusthoemmes: Let's confirm that via dev-list.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/Container.scala
joey0320: @markusthoemmes Hello. I'm a newbie at scala and having a hard time reading the source code. Can you please explain what thread-safe means above, and why resume() (which I'm guessing that it somehow calls "docker container resume") is not thread safe??

PR #4187, opened 2018-12-18
Handle kafka exception thrown when creating the admin client

mhenke1:
## Description
The KafkaMessagingProvider does not handle KafkaExceptions that are thrown when creating the admin client. Not handling these exceptions kills the invokers main thread (see stack trace below).

```
Exception in thread "main" org.apache.kafka.common.KafkaException: Failed create new KafkaAdminClient
        at org.apache.kafka.clients.admin.KafkaAdminClient.createInternal(KafkaAdminClient.java:370)
        at org.apache.kafka.clients.admin.AdminClient.create(AdminClient.java:52)
        at org.apache.openwhisk.connector.kafka.KafkaMessagingProvider$.ensureTopic(KafkaMessagingProvider.scala:63)
        at org.apache.openwhisk.core.invoker.Invoker$.main(Invoker.scala:156)
        at org.apache.openwhisk.core.invoker.Invoker.main(Invoker.scala)
```

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/connector/kafka/KafkaMessagingProvider.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/connector/kafka/KafkaMessagingProvider.scala
@@ -60,30 +60,35 @@ object KafkaMessagingProvider extends MessagingProvider {
       } getOrElse Map.empty)
 
     val commonConfig = configMapToKafkaConfig(loadConfigOrThrow[Map[String, String]](ConfigKeys.kafkaCommon))
-    val client = AdminClient.create(commonConfig + (AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG -> config.kafkaHosts))
-    val partitions = 1
-    val nt = new NewTopic(topic, partitions, kafkaConfig.replicationFactor).configs(topicConfig.asJava)
+    Try(AdminClient.create(commonConfig + (AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG -> config.kafkaHosts)))
+      .map(client => {
+        val partitions = 1
+        val nt = new NewTopic(topic, partitions, kafkaConfig.replicationFactor).configs(topicConfig.asJava)
 
-    def createTopic(retries: Int = 5): Try[Unit] = {
-      Try(client.createTopics(List(nt).asJava).values().get(topic).get())
-        .map(_ => logging.info(this, s"created topic $topic"))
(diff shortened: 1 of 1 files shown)


common/scala/src/main/scala/org/apache/openwhisk/connector/kafka/KafkaMessagingProvider.scala
```diff
@@ -60,30 +60,35 @@ object KafkaMessagingProvider extends MessagingProvider {
...
-    client.close()
-    result
+        val result = createTopic()
+        client.close()
+      })
+      .recoverWith {
+        case e =>
+          logging.error(this, s"ensureTopic for $topic failed due to $e")
+          Failure(e)
+      }
```
markusthoemmes: I think this breaks the logic of surfacing the exception that might've been thrown by `createTopic`. You should be able to sqash both `Try` into one, by doing something along the lines of:

```scala
val clientTry = Try(AdminClient.create(commonConfig + (AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG -> config.kafkaHosts))

def createTopic(retries: Int = 5): Try[Unit] = clientTry.flatMap { client =>
  ...
}

val result = createTopic()

// Close the client if creating it was successful:
clientTry.foreach(_.close())

createTopic().recoverWith {
  case e => ...
}
```

That way both the error of the client creation and the createTopic error are surfaced properly.

mhenke1: ´createTopic´ does use the admin client. Does it really make sense to continue when we have failed to get the client ?

markusthoemmes: Notice how `createTopic` flatMaps on the `clientTry`. It won't continue if the `clientTry` failed.

mhenke1: @markusthoemmes Fixed by re-adding the lost return statement and flat-mapping the two Tries. Please have look

dubee:
Good one, @mhenke1!

PR #4198, opened 2019-01-03
Enable configuring CosmosDB client per collection

chetanmeh:
Currently same CosmosDB client is used for all collections. This PR enables configuring separate client having different config on per collection basis.

## Description

Depending on requirements it may be required to configure `AsyncDocumentClient` on per collection basis. For e.g. to use different config for each collection use following config

```
cosmosdb {
  endpoint          =  "https://<account>.documents.azure.com:443/"
  key               =  "secret"             # Access key
  db                =  "openwhisk"            # Database name
  throughput        = 1000          # Throughput configured for each collection within this db
  consistency-level = "Session"
  connection-policy {
    max-pool-size = 1000
    # When the value of this property is true, the SDK will direct write operations to
    # available writable locations of geo-replicated database account
    using-multiple-write-locations = false

    # Sets the preferred locations for geo-replicated database accounts e.g. "East US"
    # See names at https://azure.microsoft.com/en-in/global-infrastructure/locations/
    preferred-locations = []
    retry-options {
      # Sets the maximum number of retries in the case where the request fails
      # because the service has applied rate limiting on the client.
      max-retry-attempts-on-throttled-requests = 9

      # Sets the maximum retry time
      # If the cumulative wait time exceeds this SDK will stop retrying and return the
      # error to the application.
      max-retry-wait-time                      = 30 s
    }
  }

  # Specify entity specific overrides below. By default all config values would be picked from top level. To override
  # any config option for specific entity specify them below. For example if multiple region writes need to be enabled
  # for activations then
   collections {
        WhiskActivation {            # Add entity specific overrides here
            connection-policy {
                   using-multiple-write-locations = true
            }
        }
    }
}
```

Above example configures all supported properties. OpenWHisk configures the same defaults so if you need to only change some of the config then you can do it like below

```
cosmosdb {
  endpoint          =  "https://<account>.documents.azure.com:443/"
  key               =  "secret"             # Access key
  db                =  "openwhisk"            # Database name
  # Specify entity specific overrides below. By default all config values would be picked from top level. To override
  # any config option for specific entity specify them below. For example if multiple writes need to be enabled
  # for activations then
   collections {
        WhiskActivation {            # Add entity specific overrides here
            connection-policy {
              using-multiple-write-locations = true
            }
        }
    }
}
```

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -186,7 +186,9 @@ whisk {
         # for activations then
     #   collections {
     #       WhiskActivation {            # Add entity specific overrides here
-    #           using-multiple-write-locations = true
+    #           connection-policy {
+    #              using-multiple-write-locations = true
+    #            }
     #       }
     #   }
     }
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -159,6 +159,8 @@ whisk {
     #   key               =               # Access key
     #   db                =               # Database name
         throughput        = 1000          # Throughput configured for each collection within this db
+        # Select from one of the supported
+        # https://azure.github.io/azure-cosmosdb-java/1.0.0/com/microsoft/azure/cosmosdb/ConsistencyLevel.html
         consistency-level = "Session"
         connection-policy {
             max-pool-size = 1000
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -154,12 +154,42 @@ whisk {
 
     # CosmosDB related configuration
     # For example:
-    # cosmosdb {
-    #     endpoint   =               # Endpoint URL like https://<account>.documents.azure.com:443/
-    #     key        =               # Access key
-    #     db         =               # Database name
-    #     throughput = 1000          # Throughput configure for each collection within this db
-    #}
+    cosmosdb {
+    #   endpoint          =               # Endpoint URL like https://<account>.documents.azure.com:443/
+    #   key               =               # Access key
+    #   db                =               # Database name
+        throughput        = 1000          # Throughput configured for each collection within this db
+        consistency-level = "Session"
+        connection-policy {
+            max-pool-size = 1000
+            # When the value of this property is true, the SDK will direct write operations to
+            # available writable locations of geo-replicated database account
+            using-multiple-write-locations = false
+
+            # Sets the preferred locations for geo-replicated database accounts e.g. "East US"
+            # See names at https://azure.microsoft.com/en-in/global-infrastructure/locations/
+            preferred-locations = []
+            retry-options {
+                # Sets the maximum number of retries in the case where the request fails
+                # because the service has applied rate limiting on the client.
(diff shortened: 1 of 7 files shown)


common/scala/src/main/scala/org/apache/openwhisk/core/database/cosmosdb/CosmosDBArtifactStore.scala
```diff
@@ -21,6 +21,7 @@ import java.io.ByteArrayInputStream
 
 import _root_.rx.RxReactiveStreams
 import akka.actor.ActorSystem
+import akka.event.slf4j.SLF4JLogging
```
tysonnorris: Is there any reason this cannot use the existing `org.apache.openwhisk.common.Logging` instead of `SLF4JLogging`?

common/scala/src/main/resources/application.conf
```diff
@@ -155,12 +155,44 @@ whisk {
...
-    #     endpoint   =               # Endpoint URL like https://<account>.documents.azure.com:443/
-    #     key        =               # Access key
-    #     db         =               # Database name
-    #     throughput = 1000          # Throughput configure for each collection within this db
-    #}
+    cosmosdb {
+    #   endpoint          =               # Endpoint URL like https://<account>.documents.azure.com:443/
+    #   key               =               # Access key
+    #   db                =               # Database name
+        throughput        = 1000          # Throughput configured for each collection within this db
```
tysonnorris: Does this correspond to RU config at azure? (do these need to be in sync?)

common/scala/src/main/resources/application.conf
```diff
@@ -155,12 +155,44 @@ whisk {
...
+        consistency-level = "Session"
```
tysonnorris: comment on this `consistency-level` config?

common/scala/src/main/scala/org/apache/openwhisk/core/database/cosmosdb/CosmosDBArtifactStore.scala
chetanmeh: Makes sense. Switched to default logging

common/scala/src/main/resources/application.conf
chetanmeh: Its used only when a fresh collection is being created. If collection already exists then its throughput is not modified

common/scala/src/main/resources/application.conf
```diff
@@ -155,12 +155,44 @@ whisk {
...
+        connection-policy {
```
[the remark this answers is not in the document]
chetanmeh: Its configurable on per collection basis. See the later example in PR description above. If its specified at collection level then a new `AsyncDocumentClient` instance would be used matching the policy

chetanmeh:
@tysonnorris I have now updated the PR based on review feedback

PR #4211, opened 2019-01-09
Add `binding` annotation to record an action path not resolved

upgle:
## Description

If the user invokes an action in a package binding, they can't know a package binding path from which action is invoked. because the current activation structure only records the resolved action path.

for example, suppose we have the following entities:

- `package` : **pkg1**
   - `action` : pkg1/**action1**
- `package` **pkg2** bind **pkg1**
- `package` **pkg3** bind **pkg1**

If user invokes the `pkg2/action1` action and the `pkg3/action1` action, all path annotations of invoked action in binding packages are saved as **namespace/pkg1/action1**.

```json
    "annotations": [
        {
            "key": "path",
            "value": "namespace/pkg1/action1"
        },
     ]
```

## Suggestion

So i added an annotation to distinguish between resolved action path and invoked action path and opend this PR.

```json
{
    "activationId": "f4250bb2721d4e4ca50bb2721d2e4cdd",
    "annotations": [
        {
            "key": "path",
            "value": "namespace/pkg1/action1"
        },
        {
            "key": "originPath",
            "value": "namespace/pkg2/action1"
        }
    ]
}
```

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
@@ -234,6 +234,27 @@ case class WhiskActionMetaData(namespace: EntityPath,
       None
   }
 
+  def toExecutableWhiskAction(originEntity: FullyQualifiedEntityName) = {
+
+    exec match {
+      case execMetaData: ExecMetaData =>
+        val origin = if (originEntity.path == namespace) None else Some(originEntity.path)
+        Some(
+          ExecutableWhiskActionMetaData(
+            namespace,
+            name,
+            execMetaData,
+            parameters,
+            limits,
+            version,
+            publish,
+            annotations,
+            origin)
+            .revision[ExecutableWhiskActionMetaData](rev))
+
+      case _ => None
+    }
+  }
 }
 
 /**
@@ -265,7 +286,8 @@ case class ExecutableWhiskAction(namespace: EntityPath,
                                  limits: ActionLimits = ActionLimits(),
                                  version: SemVer = SemVer(),
                                  publish: Boolean = false,
-                                 annotations: Parameters = Parameters())
+                                 annotations: Parameters = Parameters(),
+                                 originNamespace: Option[EntityPath] = None)
     extends WhiskActionLike(name) {
 
   require(exec != null, "exec undefined")
@@ -283,7 +305,8 @@ case class ExecutableWhiskAction(namespace: EntityPath,
   }
 
   def toWhiskAction =
-    WhiskAction(namespace, name, exec, parameters, limits, version, publish, annotations).revision[WhiskAction](rev)
+    WhiskAction(namespace, name, exec, parameters, limits, version, publish, annotations)
+      .revision[WhiskAction](rev)
 }
 
 @throws[IllegalArgumentException]
@@ -294,7 +317,8 @@ case class ExecutableWhiskActionMetaData(namespace: EntityPath,
                                          limits: ActionLimits = ActionLimits(),
                                          version: SemVer = SemVer(),
                                          publish: Boolean = false,
-                                         annotations: Parameters = Parameters())
+                                         annotations: Parameters = Parameters(),
+                                         originNamespace: Option[EntityPath] = None)
     extends WhiskActionLikeMetaData(name) {
 
   require(exec != null, "exec undefined")
@@ -303,6 +327,10 @@ case class ExecutableWhiskActionMetaData(namespace: EntityPath,
   def toWhiskAction =
     WhiskActionMetaData(namespace, name, exec, parameters, limits, version, publish, annotations)
       .revision[WhiskActionMetaData](rev)
+
+  def originFullyQualifiedName =
+    originNamespace.map(ns => FullyQualifiedEntityName(ns, name, None))
+
 }
 
 object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[WhiskAction] with DefaultJsonProtocol {
@@ -504,7 +532,9 @@ object WhiskAction extends DocumentFactory[WhiskAction] with WhiskEntityQueries[
         // fully resolved name for the action
         val fqnAction = resolvedPkg.fullyQualifiedName(withVersion = false).add(actionName)
         // get the whisk action associate with it and inherit the parameters from the package/binding
-        WhiskAction.get(entityStore, fqnAction.toDocId) map { _.inherit(resolvedPkg.parameters) }
+        WhiskAction.get(entityStore, fqnAction.toDocId) map {
+          _.inherit(resolvedPkg.parameters)
+        }
       }
     }
   }
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
@@ -274,8 +274,8 @@ case class WhiskActionMetaData(namespace: EntityPath,
  * @param limits the limits to impose on the action
  * @param version the semantic version
  * @param publish true to share the action or false otherwise
- * @param annotations the set of annotations to attribute to the action
- * @param binding the path of the package binding
+ * @param annotation the set of annotations to attribute to the action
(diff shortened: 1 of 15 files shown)


rabbah:
I wonder if instead of modifying the Action type, we can add an alias to `FullyQualifiedName`. This would mean - if it works out - that you don't have to change the activation message and the name of the action and its alias (i.e., package and binding) are coupled more tightly. 

I would use `binding` in the annotation and leave the annotation as is otherwise (shows the package name).

This is a suggestion for discussion.

upgle:
As you mentioned, it would be better to add an alias to `FullyQualifiedName`.  I'll make a commit soon. 👍

upgle:
@rabbah 
I added `binding` field to `FullyQualifiedName` to not change the activation message, and renamed annotation key from `originPath` to `binding` and this annotation contains only entityPath of a package binding.

```json
{
    "activationId": "f4250bb2721d4e4ca50bb2721d2e4cdd",
    "annotations": [
        {
            "key": "binding",
            "value": "whisk.system/p2"
        },
        {
            "key": "path",
            "value": "guest/p1/hello"
        },
    ]
}
```

style95:
@rabbah 
Is it possible to include this change? or any other direction is being considered?
In any case, I hope this issue is addressed so that we can resolve many binding related issues.

rabbah:
Thanks for the nudge. Will get on it.

upgle:
conflict is resolved

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
```diff
@@ -303,6 +327,10 @@ case class ExecutableWhiskActionMetaData(namespace: EntityPath,
   def toWhiskAction =
     WhiskActionMetaData(namespace, name, exec, parameters, limits, version, publish, annotations)
       .revision[WhiskActionMetaData](rev)
+
+  def bindingFullyQualifiedName =
```
rabbah: i suggest adding a type decorator on the method for the return value and a comment, explaining that the result is Some fully qualified name only if there's a binding, else None.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskAction.scala
```diff
@@ -254,7 +274,8 @@ case class WhiskActionMetaData(namespace: EntityPath,
  * @param limits the limits to impose on the action
  * @param version the semantic version
  * @param publish true to share the action or false otherwise
- * @param annotation the set of annotations to attribute to the action
+ * @param annotations the set of annotations to attribute to the action
+ * @param binding the path of the package binding
```
rabbah: -> `@param binding the path of the package binding if any`

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskActivation.scala
```diff
@@ -137,6 +137,7 @@ object WhiskActivation
 
   /** Some field names for annotations */
   val pathAnnotation = "path"
+  val bindingAnnotation = "binding"
```
rabbah: will check down below but this raises the opportunity to have the name in the activation be that of the binding, and the annotation is that of the package.

upgle: I'm sorry, I don't understand this. 

In the example below, I think the name `triple` in the activation can not be the name of the binding `whisk.system/p2`

- there is an action `whisk.system/pkg1/triple`
- and the `whisk.system/p2` is binding the `whisk.system/pkg1`
- and i invoked an action `whisk.system/p2/triple`

```json
{
    "namespace": "whisk.system",
    "name": "triple",
    "version": "0.0.1",
    "subject": "whisk.system",
    "activationId": "b86606b16b8a4371a606b16b8a437182",
    "start": 1552900408086,
    "end": 1552900408090,
    "duration": 4,
    "annotations": [
        {
            "key": "binding",
            "value": "whisk.system/p2"
        },
        {
            "key": "path",
            "value": "whisk.system/pkg1/triple"
        },
...
```

style95:
Any update on this PR?

upgle:
Sorry. I should resolve a conflict. I can handle today.

upgle:
@rabbah  A conflict is resolved. could you please review again?

c.c. @style95

core/controller/src/main/scala/org/apache/openwhisk/core/controller/actions/SequenceActions.scala
```diff
@@ -312,13 +317,15 @@ protected[actions] trait SequenceActions {
...
    * @param futureAction the future which fetches the action to be invoked from the db
+   * @param originAction the invoked fully qualified action name that not rewritten
    * @param accounting the state of the sequence activation, contains the dynamic activation count, logs and payload for the next action
    * @param cause the activation id of the first sequence containing this activations
    * @return a future which resolves with updated accounting for a sequence, including the last result, duration, and activation ids
    */
   private def invokeNextAction(
     user: Identity,
     futureAction: Future[WhiskActionMetaData],
+    originAction: FullyQualifiedEntityName,
```
style95: Is this being used in the call tree?

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
```diff
@@ -595,38 +590,6 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with
     })
   }
 
-  /**
-   * Constructs a WhiskPackage that is a merger of a package with its packing binding (if any).
-   * This resolves a reference versus an actual package and merge parameters as needed.
-   * Once the package is resolved, the operation is dispatched to the action in the package
-   * namespace.
-   */
-  private def mergeActionWithPackageAndDispatch(method: HttpMethod,
```
style95: @upgle Could you elaborate the reason to remove this method a little bit?
It would help people who are not familiar with this change like me to get the intention more easily.

upgle: If an action is resolved before it's dispatched, there is no way to know its original name in the activate method because it's already resolved.
So, I modified it to resolve inside the activate method after it has been dispatched.

```scala
override def activate(user: Identity, entityName: FullyQualifiedEntityName, env: Option[Parameters])(
    implicit transid: TransactionId) = {	    implicit transid: TransactionId)
```

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
```diff
@@ -253,7 +247,7 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with
       'result ? false,
       'timeout.as[FiniteDuration] ? WhiskActionsApi.maxWaitForBlockingActivation) { (blocking, result, waitOverride) =>
       entity(as[Option[JsObject]]) { payload =>
-        getEntity(WhiskActionMetaData.get(entityStore, entityName.toDocId), Some {
+        getEntity(WhiskActionMetaData.resolveActionAndMergeParameters(entityStore, entityName), Some {
```
upgle: @style95 the action is resolved here after it's dispatched.

core/controller/src/main/scala/org/apache/openwhisk/core/controller/actions/SequenceActions.scala
upgle: It seems not using in the call tree. I'll fix it.

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
style95: ok, this is the similar way which is taken for web action.

style95:
@rabbah Do you have any comments on this?

style95:
I would merge this in 72 hours if there is no objection based on silent consent.

PR #4212, opened 2019-01-10
Optimize Kamon metric update flow

### LINKED ISSUE #4203: Reduce overhead in metric update flow
Currently for each Metric update `MetricEmitter` resolves the metric instance everytime. [Kamon docs][1] suggest

> When possible, define and refine your metrics up front, as a static member or instance members on the instrumented components of your services. **This saves a bit of time in looking up instruments.**

As metric updates happen in critical path we should try to reduce this overhead. One way would be to store the metric reference in `LogMarkerToken` itself as all metric instance are function of that and `LogMarkerToken` are explicitly reference thus avoid any lookup overhead.

See also discussion [here][2]

[1]: https://kamon.io/documentation/1.x/core/basics/metrics/
[2]: https://github.com/apache/incubator-openwhisk/pull/4165#discussion_r245491258

chetanmeh:
Fixes #4203. 

## Description
This PR implements the following recommendation for Kamon usage

> When possible, define and refine your metrics up front, as a static member or instance members on the instrumented components of your services. This saves a bit of time in looking up instruments.

Now the Metric instances are stored as part of `LogMarkerToken` and reused for later updates

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#????)

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -187,6 +187,9 @@ case class LogMarkerToken(component: String,
                           tags: Map[String, String] = Map.empty) {
   private var finishToken: LogMarkerToken = _
   private var errorToken: LogMarkerToken = _
+
+  // Using var is safe wrt thread-safety because Kamon makes sure the instances
+  // (given the same key) are always the same, so a missed update is not harmful
   private var _counter: KCounter = _
   private var _histogram: KHistogram = _
 
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -24,6 +24,7 @@ import java.time.format.DateTimeFormatter
 import akka.event.Logging._
 import akka.event.LoggingAdapter
 import kamon.Kamon
+import kamon.metric.{Counter => KCounter, Histogram => KHistogram}
 import kamon.statsd.{MetricKeyGenerator, SimpleMetricKeyGenerator}
 import org.apache.openwhisk.core.entity.ControllerInstanceId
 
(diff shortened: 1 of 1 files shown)


common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
```diff
@@ -184,13 +185,71 @@ case class LogMarkerToken(component: String,
                           state: String,
                           subAction: Option[String] = None,
                           tags: Map[String, String] = Map.empty) {
+  private var finishToken: LogMarkerToken = _
+  private var errorToken: LogMarkerToken = _
+  private var _counter: KCounter = _
+  private var _histogram: KHistogram = _
```
chetanmeh: It would be safe to use `var` here as the variables would get initialized to "same" values. Kamon would ensure that metric instance for given name and tag are always same

markusthoemmes: It'd be nice to add that consideration as a comment here, along the lines of:

```
// Thread-safety is not an issue here, because Kamon makes sure the instances (given the same key) are always the same, so a missed update is not harmful
```

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
```diff
@@ -184,13 +185,71 @@ case class LogMarkerToken(component: String,
...
+    }
+    finishToken
+  }
+
+  def asError: LogMarkerToken = {
+    if (errorToken == null) {
+      errorToken = copy(state = LoggingMarkers.error)
+    }
+    errorToken
+  }
```
markusthoemmes: Is it worth caching `asFinish` or `asError` this way? Are these used multiple times in reality often?

chetanmeh: `asFinish` is being called from `TransactionId.finish`. So that would be frequent

markusthoemmes: Does it make sense to just have `asFinish` and `asError` as a `lazy val` or even just a `val`?

markusthoemmes: Thinking about it: Can all of those be `lazy val`?

chetanmeh: Making it val causes recursion. Lazy val I am trying to avoid as it would cause potencial concurrent access to a locked value.

Hence kept it as lazy initialized var

PR #4216, opened 2019-01-14
Proposing Lean OpenWhisk

kpavel:
This is an improvement of [PR 3886](https://github.com/apache/incubator-openwhisk/pull/3886) inspired by this [suggestion 
](https://github.com/apache/incubator-openwhisk/pull/4161) by @rabbah.

This PR proposes to segregate the logic common to load balancers into a CommonLoadBalancer abstract class. The ShardingContainerPoolBalancer and LeanBalancer (added in this PR) inherit from this class.

The proposition is to have a lean distro of OpenWhisk suitable to small form factors at the edge. The idea is to have Controller and Invoker compiled together into a "lean" controller-invoker (have a Gradle project for that) with controller calling the invoker's method directly via an in-memory object (a queue, which is also loaded via SPI) via a new pluggable load balancer that is selected via SPI using configuration file property setting. For more information here is a [blog](https://medium.com/@davidbr_9022/lean-openwhisk-open-source-faas-for-edge-computing-fb823c6bbb9b)
As well, briefly discussed it in this [thread](https://lists.apache.org/thread.html/516efc0ed9afb09fe52e30d10eb9fe5091af91df53a46c07e1701122@%3Cdev.openwhisk.apache.org%3E).

## Description

The goal is to avoid changes in the core components as much as possible. Rather this PR adds a few components. The main additions are as follows:

- CommonLoadBalancer: provides common LoadBalancer fuctionality

- LeanBalancer SPI: provides LoadBalancer to communicate with Invoker directly without Kafka in the middle (Invoker does not exist as a separate entity, it is built together with Controller)

- LeanMessagingProvider SPI: provides Producer/Consumer implementers to communicate via in-memory queue instead of Kafka.

- Ansible deployment: added controller-lean.yml and other files required for the Ansible deployment. Since lean controller is a combination of the controller + invoker projects, its ansible deploy.yml requires most of the values from both. Currently this problem is resolved by partially replicating variables from the regular invoker's deploy.yml to the tasks/invoker-lean.yml. Probably it would be better to avoid this duplication by having everything we need for lean in the regular deploy.yml.

- Dockerfile: to build Lean controller docker image, which is a merge of the regular invoker + controller 

- Vagrantfile: a small change to make Lean available using environment variable

## Related issue and scope

- [ ] API
- [ ] Controller
- [x] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [x] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


ansible/setup.yml
```diff
@@ -92,3 +92,7 @@
   - name: generate invoker certificates
     when: invoker.protocol == 'https'
     local_action: shell "{{ playbook_dir }}/files/genssl.sh" "{{ invoker.ssl.cn }}" "server" "{{ playbook_dir }}/roles/invoker/files" {{ invoker.ssl.keystore.password }} {{ invoker.ssl.keyPrefix }} "generateKey"
+
+  - name: generate lean controller plugin invoker certificates
```
rabbah: why is this necessary?

core/controller/build.gradle
```diff
@@ -43,6 +43,7 @@ dependencies {
     compile 'com.lightbend.akka.discovery:akka-discovery-kubernetes-api_2.12:0.11.0'
     compile 'com.lightbend.akka.discovery:akka-discovery-marathon-api_2.12:0.11.0'
     compile project(':common:scala')
+    compile project(':core:invoker')
```
rabbah: should this be done only when necessary? at least should have a comment.

core/lean/build.gradle
```diff
@@ -0,0 +1,55 @@
...
+    mavenCentral()
+}
+
+dependencies {
+    compile "org.scala-lang:scala-library:${gradle.scala.version}"
+    compile 'com.lightbend.akka.management:akka-management-cluster-bootstrap_2.12:0.11.0'
+    compile 'com.lightbend.akka.discovery:akka-discovery-kubernetes-api_2.12:0.11.0'
+    compile 'com.lightbend.akka.discovery:akka-discovery-marathon-api_2.12:0.11.0'
+    compile project(':common:scala')
+    compile project(':core:controller')
```
rabbah: actually why not do the invoker build here?

ansible/setup.yml
kpavel: Lean controller ansible setup calls invoker ansible, which requires invoker certificates

core/controller/build.gradle
kpavel: LeanBalancer has dependency on InvokerReactive, it can't be optional dependency

core/lean/build.gradle
kpavel: LeanBalancer has dependency on InvokerReactive. Controller will not compile without this dependency

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanConsumer.scala
```diff
@@ -0,0 +1,48 @@
...
+import java.util.concurrent.BlockingQueue
+import java.util.concurrent.TimeUnit
+
+class LeanConsumer(queue: BlockingQueue[Array[Byte]], override val maxPeek: Int)(implicit logging: Logging)
+    extends MessageConsumer {
+
+  /**
+   */
+  override def peek(duration: FiniteDuration, retry: Int): Iterable[(String, Int, Long, Array[Byte])] = {
+    logging.info(this, s"BBB in peek, Q.size: " + queue.size())
```
dubee: What is `BBB` in the log statements in the file?

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanProducer.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  implicit val ec: ExecutionContext = actorSystem.dispatcher
+
+  override def sentCount(): Long = sentCounter.cur
+
+  /** Sends msg to topic. This is an asynchronous operation. */
+  override def send(topic: String, msg: Message, retry: Int = 3): Future[RecordMetadata] = {
+    implicit val transid = msg.transid
+
+    logging.info(this, s"BBB in send with topic " + topic + " msg " + transid)
```
dubee: More `BBB`s here.

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanMessagingProvider.scala
```diff
@@ -0,0 +1,70 @@
...
+  val queues: Map[String, BlockingQueue[Array[Byte]]] =
+    new TrieMap[String, BlockingQueue[Array[Byte]]]
+
+  def getConsumer(config: WhiskConfig, groupId: String, topic: String, maxPeek: Int, maxPollInterval: FiniteDuration)(
+    implicit logging: Logging,
+    actorSystem: ActorSystem): MessageConsumer = {
+
+    var queue = queues.getOrElseUpdate(topic, new LinkedBlockingQueue[Array[Byte]]())
+
+    logging.info(this, s"AAA getConsumer for topic: " + topic + ", Q: " + queue.size())
```
dubee: What is `AAA` in these log statements?

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanConsumer.scala
kpavel: Removed unneeded logs

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanProducer.scala
kpavel: Removed

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanMessagingProvider.scala
kpavel: Removed

dgrove-oss:
Looks like a merge instead of a rebase has pulled in unrelated/unintended changes to the diff.  Possible to fix to make it easier to see what changes in the PR?

kpavel:
@dgrove-oss Fixed

dgrove-oss:
This looks plausible to me, but I'm not comfortable being the primary reviewer for a loadbalancer change of this magnitude.  @rabbah or @cbickel does one of you have the bandwidth for a careful review?

rabbah:
Thanks @dgrove-oss - I will go through it today.

rabbah:
I deployed lean stack locally, and successfully using ansible. I noticed the playbook generates several ansible warnings such as the ones below (there were at least four instances to fix --- these were fixed in this pr https://github.com/apache/incubator-openwhisk/pull/4092 and possibly some others).

```bash
TASK [controller : pull runtime action images per manifest] ***********************************************************************************************************************************
Monday 21 January 2019  18:04:19 -0500 (0:00:00.036)       0:00:09.207 ******** 
 [WARNING]: The loop variable 'item' is already in use. You should set the `loop_var` value in the `loop_control` option for the task to something else to avoid variable collisions and
unexpected behavior.
...
TASK [controller : pull blackboxes action images per manifest] ********************************************************************************************************************************
Monday 21 January 2019  18:09:12 -0500 (0:04:52.926)       0:05:02.134 ******** 
 [WARNING]: The loop variable 'item' is already in use. You should set the `loop_var` value in the `loop_control` option for the task to something else to avoid variable collisions and
unexpected behavior.
```

It would be convenient to also add `lean` mode to [`redo`](https://github.com/apache/incubator-openwhisk/tree/master/tools/build#how-to-use-redo).

The deployed stack is as expected: nginx, couch, and "lean" container, along with my actions containers.

I will next review the code changes.

ansible/setup.yml
rabbah: is the invoker certificate required though or can it be predicated on `!lean` as well?

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanConsumer.scala
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.openwhisk.common.Logging
+import org.apache.openwhisk.core.connector.MessageConsumer
+import java.util.concurrent.BlockingQueue
+import java.util.concurrent.TimeUnit
+
+class LeanConsumer(queue: BlockingQueue[Array[Byte]], override val maxPeek: Int)(implicit logging: Logging)
+    extends MessageConsumer {
+
+  /**
+   */
```
rabbah: can you add the javadoc for this method - it's a long poll; is it thread safe also?

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanConsumer.scala
```diff
@@ -0,0 +1,46 @@
...
+  override def peek(duration: FiniteDuration, retry: Int): Iterable[(String, Int, Long, Array[Byte])] = {
+    val res = Option(queue.poll(duration.toMillis, TimeUnit.MILLISECONDS))
+      .map(record => Iterable(("", 0, 0L, record)))
+      .getOrElse(Iterable.empty)
+
+    res
+  }
+
+  /**
+   */
```
rabbah: also a comment here - there's no cursor to advance since that's done in the poll above. anything else?

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanMessagingProvider.scala
```diff
@@ -0,0 +1,68 @@
...
+import akka.actor.ActorSystem
+import org.apache.openwhisk.common.Logging
+import org.apache.openwhisk.core.WhiskConfig
+import org.apache.openwhisk.core.connector.MessageConsumer
+import org.apache.openwhisk.core.connector.MessageProducer
+import org.apache.openwhisk.core.connector.MessagingProvider
+import org.apache.openwhisk.core.entity.ByteSize
+
+/**
+ * A simple implementation of MessagingProvider
```
rabbah: please end this line with a `.`

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanMessagingProvider.scala
```diff
@@ -0,0 +1,68 @@
...
+ */
+object LeanMessagingProvider extends MessagingProvider {
+
+  val queues: Map[String, BlockingQueue[Array[Byte]]] =
```
rabbah: can you add a comment here. The key is the topic.
does this actually need to be a triemap though - topics aren't added/removed dynamically and only modified at startup in the current architecture. 

it's likely OK but let's document the implementation choices.

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanMessagingProvider.scala
```diff
@@ -0,0 +1,68 @@
...
+    new TrieMap[String, BlockingQueue[Array[Byte]]]
+
+  def getConsumer(config: WhiskConfig, groupId: String, topic: String, maxPeek: Int, maxPollInterval: FiniteDuration)(
+    implicit logging: Logging,
+    actorSystem: ActorSystem): MessageConsumer = {
+
+    var queue = queues.getOrElseUpdate(topic, new LinkedBlockingQueue[Array[Byte]]())
```
rabbah: note to self: when using kafka, there's an ensure topic pre-processing phase that happens early in the boot process - is the same pre-processing happening here?

rabbah: in fact ensure topic is implement a few lines below... in which case, I'm not sure when a new topic would be dynamically added to the map as done here on a get.

should this be treated such that a warning is generated when new topics are added - does this have an analog when using kafka?

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanProducer.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  implicit val ec: ExecutionContext = actorSystem.dispatcher
+
+  override def sentCount(): Long = sentCounter.cur
+
+  /** Sends msg to topic. This is an asynchronous operation. */
+  override def send(topic: String, msg: Message, retry: Int = 3): Future[RecordMetadata] = {
+    implicit val transid = msg.transid
+
+    var queue = queues.getOrElseUpdate(topic, new LinkedBlockingQueue[Array[Byte]]())
```
rabbah: It seems there are multiple places where the queues could be changing underneath --- might explain the triemap.
My thinking is still that this should be static --- once the topics are ensured, why should new topics be created?

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanProducer.scala
```diff
@@ -0,0 +1,61 @@
...
+
+    val res = Future {
+      queue.put(msg.serialize.getBytes(StandardCharsets.UTF_8))
+      sentCounter.next()
+      null
```
rabbah: why? set type of a `res` to `Future[Unit]` instead and remove `null`.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -285,194 +257,20 @@ class ShardingContainerPoolBalancer(
...
+  override val invokerPool =
     invokerPoolFactory.createInvokerPool(
       actorSystem,
       messagingProvider,
       messageProducer,
       sendActivationToInvoker,
       Some(monitor))
+
+  override protected def releaseInvoker(invoker: InvokerInstanceId, entry: ActivationEntry) = {
+    // guard this
```
rabbah: I don't understand the comment.

tests/build.gradle
```diff
@@ -98,6 +100,16 @@ ext.testSets = [
         "includes" : [
             "system/basic/**"
         ]
+    ],
+    "REQUIRE_LEAN_SYSTEM" : [
+        "includes" : systemIncludes,
+        
+        // Lean OpenWhisk doesn't use Kafka and has no Kafka installed as part of the setup.
+        // Tests suit below validating KafkaConnector so have to be excluded for Lean System tests
```
rabbah: `         // Tests suits below require Kafka so they are excluded for Lean System tests`

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanProducer.scala
kpavel: Probably we can replace "getOrElseUpdate" with "get" instead.
I think we still need a data structure to hold the queues, unless we want to change the MessageProducer interface.
Is TrieMap not good enough?

rabbah: It may be just fine but so is a plain old map because I don't think there are cases where the map keys/values themselves change. The triemap is a concurrent datastructure with locking, which may not be necessary here. Hence the questions.

kpavel: ensureTopic in lean implementation is not called for invokers topic since the regular call for it is from invokers main(). invoker topic is created dynamically in current implementation.

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanProducer.scala
kpavel: It may require to change MessageProducer interface for that. Although, i think org.apache.kafka.clients.producer.RecordMetadata should not be part of the Spi, maybe we should leave it for another PR?

kpavel: And maybe when we change interface, it worth also to change MessageConsumer.peek() to return Message instead of Array[Byte]. It would probably save few cpu cycles for absolutely unneeded message serialization/deserialization in the current lean implementation.

ansible/setup.yml
kpavel: lean predicate will require adding new variable to whisk.conf as well

if it's ok, currently resolved it by removing "generate invoker certificates" block to avoid code duplication. using fact that ansible searches for invoker certificates in ansible/files as well and not only /roles/invoker/files

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanConsumer.scala
kpavel: added javadoc.
yes, according to [BlockingQueue ](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingQueue.html) documentation it is thread safe.

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanConsumer.scala
kpavel: Done.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
kpavel: Wrong comment removed

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanProducer.scala
kpavel: Replaced TrieMap with Map

common/scala/src/main/scala/org/apache/openwhisk/connector/lean/LeanProducer.scala
rabbah: I see - I missed that the type signature of the method - thanks for pointing that out for me. I don't like that we're returning null here in that case as it can lead to an NPE if someone tries to chain the future somewhere. Perhaps we should change the signature of the producer to Future[Option[RecordMetadata]] instead.

@markusthoemmes opinion?

rabbah:
I see one more ansible warning:

```
TASK [controller : (re)start controller] ********************************************************************************************************************
Saturday 26 January 2019  14:09:06 -0500 (0:00:00.232)       0:01:08.513 ****** 
[DEPRECATION WARNING]: Using variables for task params is unsafe, especially if the variables come from an external source like facts. This feature will be 
removed in version 2.6. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.

```

rabbah:
There is still an ansible warning to remove:

```
TASK [controller : pull runtime action images per manifest] *************************************************************************************************
Saturday 26 January 2019  14:11:54 -0500 (0:00:00.035)       0:00:09.117 ****** 
 [WARNING]: The loop variable 'item' is already in use. You should set the `loop_var` value in the `loop_control` option for the task to something else to
avoid variable collisions and unexpected behavior.
```

dgrove-oss:
PG1 3907 🏃

dgrove-oss:
PG1 3921 🏃  (checking again after the last force-push)

rabbah:
The controller image should just be called `controller` not `lean` - why have two different controller images?

The lean image is 302MB. The current controller image is 297MB. I think it's simpler to have one image for both --- is that not feasible?

ansible/roles/controller/tasks/lean.yml
```diff
@@ -0,0 +1,56 @@
...
+- name: update controller docker container params
+  vars:
+    lean_controller_params:
+      userns_mode: "host"
+      pid_mode: "host"
+      privileged: "yes"
+  set_fact:
+    controller_container_params: "{{ controller_container_params | combine(lean_controller_params) }}"
+
+# VIM: let b:syntastic_yaml_yamllint_args="-c '".expand('%:p:h')."../../../yamllint.yml'"
```
rabbah: remove this comment.

ansible/controller-lean.yml
```diff
@@ -0,0 +1,39 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more contributor
```
rabbah: do we really need this file if the image name for the two version of the controller is the same?

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/LeanBalancer.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  override def instance(whiskConfig: WhiskConfig, instance: ControllerInstanceId)(
+    implicit actorSystem: ActorSystem,
+    logging: Logging,
+    materializer: ActorMaterializer): LoadBalancer = {
+
+    new LeanBalancer(whiskConfig, createFeedFactory(whiskConfig, instance), instance)
+  }
+
+  def requiredProperties =
```
rabbah: i think you can remove service port form this - it can't be different from the controller's port and not sure it's used anyway is it?

rabbah:
I'm seeing this ansible warning also which I don't think I noticed previously. If it's not specific to this PR then please defer (will try to confirm on my own).

```
TASK [controller : (re)start controller] ********************************************************************************************************************
Sunday 03 February 2019  02:32:07 +0000 (0:00:00.105)       0:00:29.646 ******* 
[DEPRECATION WARNING]: Using variables for task params is unsafe, especially if the variables come from an external source like facts. This feature will be 
removed in version 2.6. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
changed: [controller0]
```

ansible/controller-lean.yml
kpavel: yes, we need it for several reasons. 
- specify that current setup is actually lean
- specify lean.yml as plugin
would be happy to get rid of it, but can't find a good way to specify those

ansible/roles/controller/tasks/lean.yml
kpavel: > I'm seeing this ansible warning also which I don't think I noticed previously. If it's not specific to this PR then please defer (will try to confirm on my own).
> 
> ```
> TASK [controller : (re)start controller] ********************************************************************************************************************
> Sunday 03 February 2019  02:32:07 +0000 (0:00:00.105)       0:00:29.646 ******* 
> [DEPRECATION WARNING]: Using variables for task params is unsafe, especially if the variables come from an external source like facts. This feature will be 
> removed in version 2.6. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
> changed: [controller0]
> ```

Resolved. This was due to specifying docker_container dynamically. Fixed it by changing approach.

kpavel: Done

ansible/controller-lean.yml
rabbah: as discussed on slack - i think it's possible to nuke this with a task to set lean mode.

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -6,14 +6,15 @@
...
   set_fact:
     controller_name: "{{ name_prefix ~ host_group.index(inventory_hostname) }}"
     controller_index:
       "{{ (controller_index_base|int) + host_group.index(inventory_hostname) }}"
+    image_name: "{{ image_name | default('controller') }}"
 
-- name: "pull the {{ docker.image.tag }} image of controller"
-  shell: "docker pull {{docker_registry}}{{ docker.image.prefix }}/controller:{{docker.image.tag}}"
+- name: pull the {{ image_name }}:{{ docker.image.tag }} image of controller
+  shell: "docker pull {{docker_registry}}{{ docker.image.prefix }}/{{image_name}}:{{docker.image.tag}}"
```
rabbah: this is just controller now no? and you can nuke prev task.

ansible/roles/controller/tasks/deploy.yml
```diff
@@ -296,6 +297,9 @@
     env: "{{ env }}"
     volumes: "{{ controller_volumes }}"
     ports: "{{ ports_to_expose }}"
+    userns_mode: "{{ userns_mode | default('') }}"
```
rabbah: please comment why these are there.

ansible/README.md
```diff
@@ -280,6 +280,17 @@ This is usually not necessary, however in case you want to uninstall all prereqs
 ansible-playbook -i environments/<environment> prereq.yml -e mode=clean
 ```
 
+### Lean Setup
+To have a lean setup (no Kafka, Zookeeper and no Invokers as separate entities):
+
+At [Deploying Using CouchDB](ansible/README.md#deploying-using-cloudant) step, replace:
+```
+ansible-playbook -i environments/<environment> openwhisk.yml
```
rabbah: if you succeed in removing the lean playbook, then we can just a model option here `-e mode=lean` on the deploy controller playbook.

build.gradle
```diff
@@ -87,22 +87,22 @@ subprojects {
                 enabled false
             }
 
-            configurations.archives.artifacts.removeAll {it.file =~ 'zip'}
+            configurations.archives.artifacts.removeAll { it.file =~ 'zip' }
```
rabbah: this looks like all white space changes - nuke this change?

core/controller/Dockerfile
```diff
@@ -8,6 +8,21 @@ ENV UID=1001 \
 ENV SWAGGER_UI_DOWNLOAD_SHA256=3d7ef5ddc59e10f132fe99771498f0f1ba7a2cbfb9585f9863d4191a574c96e7 \
     SWAGGER_UI_VERSION=3.6.0
 
+ENV DOCKER_VERSION=1.12.0 \
```
rabbah: please comment this whole section that it's needed for lean mode where the controller is also an invoker.

tests/performance/preparation/deploy-lean.sh
```diff
@@ -0,0 +1,38 @@
...
+ANSIBLE_CMD="$ANSIBLE_CMD -e limit_invocations_per_minute=999999 -e limit_invocations_concurrent=999999 -e controller_client_auth=false -e userLogs_spi=\"org.apache.openwhisk.core.containerpool.logging.LogDriverLogStoreProvider\""
+
+$ANSIBLE_CMD setup.yml
+
+$ANSIBLE_CMD prereq.yml
+$ANSIBLE_CMD couchdb.yml
+$ANSIBLE_CMD initdb.yml
+$ANSIBLE_CMD wipe.yml
+
+$ANSIBLE_CMD controller-lean.yml
```
rabbah: you'll need to update this if you succeed in nuking the playbook.

tools/build/redo
```diff
@@ -266,6 +266,17 @@ Components = [
                   modes = 'clean',
                   gradle = ':core:invoker'),
 
+    makeComponent('lean-fresh',
```
rabbah: and this --- i think there's a way to add extra ansible parameters to the steps.
if not, i'd be ok removing this option entirely because it can be done from the command line already via `redo`... less is more.

tools/travis/distDocker-lean.sh
```diff
@@ -0,0 +1,32 @@
...
+set -e
+
+# Build script for Travis-CI.
+
+SECONDS=0
+SCRIPTDIR=$(cd $(dirname "$0") && pwd)
+ROOTDIR="$SCRIPTDIR/../.."
+
+cd $ROOTDIR
+TERM=dumb ./gradlew distDocker -PdockerImagePrefix=testing $GRADLE_PROJS_SKIP
```
rabbah: don't need this - only one controller.

ansible/controller-lean.yml
kpavel: Done. controller-lean.yml removed

ansible/roles/controller/tasks/lean.yml
```diff
@@ -0,0 +1,56 @@
...
+  set_fact:
+    controller_env: "{{ env }}"
+    controller_ports: "{{ ports_to_expose }}"
+    inventory_hostname: "invoker0"
+    invoker_index_base: 0
+    name_prefix: "invoker"
+    host_group: "{{ groups['invokers'] }}"
+
+- name: include invoker data
+  include_tasks: "../invoker/tasks/deploy.yml"
```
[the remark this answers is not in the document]
kpavel: We can remove the whole lean plugin file, but it requires to put all its code under "set_fact when lean" in controller deploy.yml. In my opinion, current implementation of using plugin is cleaner, but  I'm ok with whatever will be decided.

ansible/roles/controller/tasks/lean.yml
```diff
@@ -0,0 +1,56 @@
...
+
+- name: populate volumes
+  set_fact:
+    controller_volumes: >-
+      {{ invoker_volumes }} +
+      {{ controller_volumes }}
+
+- name: populate environment variables for LEAN controller
+  vars:
+    lean_env:
```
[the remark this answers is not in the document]
kpavel: we need as well:
- volumes config from invoker
- extended docker container params 
    userns_mode: "host"
    pid_mode: "host"
    privileged: "yes"
- (the first procedure in lean.yml) set inventory_hostname to invoker and save controllers data that can be changed by invoker task

we can move all those under "when lean" in controller deploy.yml if it;s better than having it in separate lean plugin.

ansible/roles/invoker/tasks/deploy.yml
```diff
@@ -26,25 +26,29 @@
 # must include a trailing '/'.
 #
 - name: "pull runtime action images per manifest"
-  shell: "docker pull {{runtimes_registry | default()}}{{item.prefix}}/{{item.name}}:{{item.tag | default()}}"
+  shell: "docker pull {{runtimes_registry | default()}}{{inv_item.prefix}}/{{inv_item.name}}:{{inv_item.tag | default()}}"
```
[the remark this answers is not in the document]
kpavel: it will clash with "item" from controller deploy.yml
we need to include invoker deploy.yml anyway. either from separate plugin or from "when lean" section in controller deploy.yml

ansible/setup.yml
rabbah: i still don't understand the need for these certs in lean mode.
but because the files move as a result of this change, you need to also update the git ignores.

kpavel: removed this change from the setup.yml.
fixed the issue by adding full path to certificates in invoker deploy.yml, similar to how kafkas certificates handled by invokers ansible.

dgrove-oss:
PG4 - 2628 🏃

PR #4219, opened 2019-01-15
Send capacity, system overload metrics for managed and blackbox invokers separately

SugandhaAgrawal:
To send metrics on system overload condition, capacity in flight v/s total and count of healthy, unhealthy, unresponsive and down invokers (managed and blackbox separately) in order to visualise it as graph on Metrics dashboard.

## Related issue and scope

- [ ] API
- [x] Controller
- [x] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -335,7 +335,7 @@ object LoggingMarkers {
   val CONTROLLER_ACTIVATION_BLOCKING_DATABASE_RETRIEVAL =
     LogMarkerToken(controller, "blockingActivationDatabaseRetrieval", count)
 
-  // Time that is needed load balance the activation
+  // Time that is needed to load balance the activation
   val CONTROLLER_LOADBALANCER = LogMarkerToken(controller, loadbalancer, start)
 
   // Time that is needed to produce message in kafka
@@ -382,6 +382,19 @@ object LoggingMarkers {
     LogMarkerToken(invoker, "containerStart", count, Some(containerState), Map("containerState" -> containerState))
   val CONTAINER_CLIENT_RETRIES =
     LogMarkerToken(containerClient, "retries", count)
+
+  val INVOKER_TOTALMEM_BLACKBOX = LogMarkerToken(invoker, "totalCapacityBlackBox", count)
+  val INVOKER_TOTALMEM_MANAGED = LogMarkerToken(invoker, "totalCapacityManaged", count)
+
+  val HEALTHY_INVOKER_MANAGED = LogMarkerToken(invoker, "totalHealthyInvokerManaged", count)
+  val UNHEALTHY_INVOKER_MANAGED = LogMarkerToken(invoker, "totalUnhealthyInvokerManaged", count)
+  val UNRESPONSIVE_INVOKER_MANAGED = LogMarkerToken(invoker, "totalUnresponsiveInvokerManaged", count)
+  val DOWN_INVOKER_MANAGED = LogMarkerToken(invoker, "totalDownInvokerManaged", count)
+
+  val HEALTHY_INVOKER_BLACKBOX = LogMarkerToken(invoker, "totalHealthyInvokerBlackBox", count)
+  val UNHEALTHY_INVOKER_BLACKBOX = LogMarkerToken(invoker, "totalUnhealthyInvokerBlackBox", count)
+  val UNRESPONSIVE_INVOKER_BLACKBOX = LogMarkerToken(invoker, "totalUnresponsiveInvokerBlackBox", count)
+  val DOWN_INVOKER_BLACKBOX = LogMarkerToken(invoker, "totalDownInvokerBlackBox", count)
 
   // Kafka related markers
   def KAFKA_QUEUE(topic: String) = LogMarkerToken(kafka, topic, count)
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -383,18 +383,18 @@ object LoggingMarkers {
   val CONTAINER_CLIENT_RETRIES =
     LogMarkerToken(containerClient, "retries", count)
 
-  val INVOKER_TOTALMEM_BLACKBOX = LogMarkerToken(invoker, "totalCapacityBlackBox", count)
-  val INVOKER_TOTALMEM_MANAGED = LogMarkerToken(invoker, "totalCapacityManaged", count)
+  val INVOKER_TOTALMEM_BLACKBOX = LogMarkerToken(loadbalancer, "totalCapacityBlackBox", count)
+  val INVOKER_TOTALMEM_MANAGED = LogMarkerToken(loadbalancer, "totalCapacityManaged", count)
 
-  val HEALTHY_INVOKER_MANAGED = LogMarkerToken(invoker, "totalHealthyInvokerManaged", count)
-  val UNHEALTHY_INVOKER_MANAGED = LogMarkerToken(invoker, "totalUnhealthyInvokerManaged", count)
-  val UNRESPONSIVE_INVOKER_MANAGED = LogMarkerToken(invoker, "totalUnresponsiveInvokerManaged", count)
-  val DOWN_INVOKER_MANAGED = LogMarkerToken(invoker, "totalDownInvokerManaged", count)
+  val HEALTHY_INVOKER_MANAGED = LogMarkerToken(loadbalancer, "totalHealthyInvokerManaged", count)
+  val UNHEALTHY_INVOKER_MANAGED = LogMarkerToken(loadbalancer, "totalUnhealthyInvokerManaged", count)
+  val UNRESPONSIVE_INVOKER_MANAGED = LogMarkerToken(loadbalancer, "totalUnresponsiveInvokerManaged", count)
+  val DOWN_INVOKER_MANAGED = LogMarkerToken(loadbalancer, "totalDownInvokerManaged", count)
 
-  val HEALTHY_INVOKER_BLACKBOX = LogMarkerToken(invoker, "totalHealthyInvokerBlackBox", count)
-  val UNHEALTHY_INVOKER_BLACKBOX = LogMarkerToken(invoker, "totalUnhealthyInvokerBlackBox", count)
-  val UNRESPONSIVE_INVOKER_BLACKBOX = LogMarkerToken(invoker, "totalUnresponsiveInvokerBlackBox", count)
-  val DOWN_INVOKER_BLACKBOX = LogMarkerToken(invoker, "totalDownInvokerBlackBox", count)
+  val HEALTHY_INVOKER_BLACKBOX = LogMarkerToken(loadbalancer, "totalHealthyInvokerBlackBox", count)
+  val UNHEALTHY_INVOKER_BLACKBOX = LogMarkerToken(loadbalancer, "totalUnhealthyInvokerBlackBox", count)
+  val UNRESPONSIVE_INVOKER_BLACKBOX = LogMarkerToken(loadbalancer, "totalUnresponsiveInvokerBlackBox", count)
+  val DOWN_INVOKER_BLACKBOX = LogMarkerToken(loadbalancer, "totalDownInvokerBlackBox", count)
 
   // Kafka related markers
   def KAFKA_QUEUE(topic: String) = LogMarkerToken(kafka, topic, count)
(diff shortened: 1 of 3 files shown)


SugandhaAgrawal:
PG3 - 3205 ✅

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
```diff
@@ -330,14 +330,14 @@ object LoggingMarkers {
...
-  // Time that is needed load balance the activation
+  // Time that is needed to load balance the activation
   val CONTROLLER_LOADBALANCER = LogMarkerToken(controller, loadbalancer, start)
 
   // Time that is needed to produce message in kafka
   val CONTROLLER_KAFKA = LogMarkerToken(controller, kafka, start)
 
   // System overload and random invoker assignment
-  val SYSTEM_OVERLOAD = LogMarkerToken(controller, "systemOverload", count)
+  def SYSTEM_OVERLOAD(actionType: String) = LogMarkerToken(controller, s"systemOverload$actionType", count)
```
dgrove-oss: Any performance concerns about changing this from a val field to a function that will do a string interpolation on every invocation?

rabbah:
seems like there are a lot more metric being emitted in general relative to existing code.

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
markusthoemmes: Agreed, let's make two static values SYSTEM_OVERLOAD_MANAGED and SYSTEM_OVERLOAD_BLACKBOX to get rid of that work on each invocation.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -179,14 +179,53 @@ class ShardingContainerPoolBalancer(
...
+      totalBBActivationMemory.longValue + totalManagedActivationMemory.longValue)
+    MetricEmitter.emitHistogramMetric(
+      LOADBALANCER_MEMORY_INFLIGHT(controllerInstance, "Blackbox"),
+      totalBBActivationMemory.longValue)
+    MetricEmitter.emitHistogramMetric(
+      LOADBALANCER_MEMORY_INFLIGHT(controllerInstance, "Managed"),
+      totalManagedActivationMemory.longValue)
+    MetricEmitter.emitHistogramMetric(
+      INVOKER_TOTALMEM_BLACKBOX,
+      schedulingState.blackboxInvokers.filter(_.status.isUsable).map(_.id.userMemory.toMB).sum)
```
markusthoemmes: this needs to traverse the whole invoker list 3 times (filter, map, sum). As a general rule, filter + map can always be concatenated into one `collect`, in this case:

```scala
schedulingState.blackboxInvokers.collect { 
  case in if in.status.isUsable => in.id.userMemory.toMB 
}.sum
```

However, in this specific case, we should be able to reduce the traversals to one time.

```scala
schedulingState.blackboxInvokers.foldLeft(0L) { (acc, curr) => if(curr.status.isUsable) {
    acc + curr.id.userMemory.toMB
  } else {
    acc
  }
}
```

(Pseudo code, no IDE installed locally)

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -179,14 +179,53 @@ class ShardingContainerPoolBalancer(
...
+    MetricEmitter.emitHistogramMetric(
+      INVOKER_TOTALMEM_MANAGED,
+      schedulingState.managedInvokers.filter(_.status.isUsable).map(_.id.userMemory.toMB).sum)
+    MetricEmitter.emitHistogramMetric(
+      HEALTHY_INVOKER_MANAGED,
+      schedulingState.managedInvokers.count(_.status.asString == "up"))
```
markusthoemmes: Can we compare the status as is? For example: 

```scala
schedulingState.managedInvokers.count(_.status == Healthy)
```

Nice touch for using `count`

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -298,7 +337,12 @@ class ShardingContainerPoolBalancer(
                               instance: InvokerInstanceId): Future[Either[ActivationId, WhiskActivation]] = {
 
     totalActivations.increment()
-    totalActivationMemory.add(action.limits.memory.megabytes)
+    if (action.exec.pull) {
```
markusthoemmes: Can we pull this into a `isBlackbox` value and use below again? Makes it a bit more readable.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -579,14 +629,14 @@ object ShardingContainerPoolBalancer extends LoadBalancerProvider {
             val random = healthyInvokers(ThreadLocalRandom.current().nextInt(healthyInvokers.size)).id
             dispatched(random.toInt).forceAcquireConcurrent(fqn, maxConcurrent, slots)
             logging.warn(this, s"system is overloaded. Chose invoker${random.toInt} by random assignment.")
-            MetricEmitter.emitCounterMetric(LoggingMarkers.SYSTEM_OVERLOAD)
+            MetricEmitter.emitCounterMetric(LoggingMarkers.SYSTEM_OVERLOAD(actionType))
```
markusthoemmes: I feel like we shouldn't be passing the `actionType` down here but rather extract the overflow logic out of the general scheduler for easier access to that case. It's my bad conflating the two.

SugandhaAgrawal:
@markusthoemmes , @sven-lange-last changes done in code to suffice the comments from above 🎉

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -329,7 +395,8 @@ class ShardingContainerPoolBalancer(
           action.limits.memory.megabytes.MB,
           action.limits.concurrency.maxConcurrent,
           action.fullyQualifiedName(true),
-          timeoutHandler)
+          timeoutHandler,
+          action.exec.pull)
```
markusthoemmes: reuse `isBlackboxInvocation` here.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -426,7 +493,11 @@ class ShardingContainerPoolBalancer(
     activationSlots.remove(aid) match {
       case Some(entry) =>
         totalActivations.decrement()
-        totalActivationMemory.add(entry.memory.toMB * (-1))
+        if (entry.isBlackbox) {
+          totalBlackBoxActivationMemory.add(entry.memory.toMB * (-1))
+        } else {
+          totalManagedActivationMemory.add(entry.memory.toMB * (-1))
+        }
```
markusthoemmes: You can cut down duplication a bit by doing:

```scala
val totalActivationMemory = if(entry.isBlackbox) totalBlackBoxActivationMemory else totalManagedActivationMemory
totalActivationmemory.add(entry.memory.toMB * (-1))
```

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -298,7 +358,13 @@ class ShardingContainerPoolBalancer(
                               instance: InvokerInstanceId): Future[Either[ActivationId, WhiskActivation]] = {
 
     totalActivations.increment()
-    totalActivationMemory.add(action.limits.memory.megabytes)
+    val isBlackboxInvocation = action.exec.pull
+    if (isBlackboxInvocation) {
+      totalBlackBoxActivationMemory.add(action.limits.memory.megabytes)
+    } else {
+      totalManagedActivationMemory.add(action.limits.memory.megabytes)
+    }
```
markusthoemmes: Same comment on cutting down duplication.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -179,14 +180,63 @@ class ShardingContainerPoolBalancer(
...
+    MetricEmitter.emitHistogramMetric(
+      LOADBALANCER_MEMORY_INFLIGHT(controllerInstance, "Managed"),
+      totalManagedActivationMemory.longValue)
+    MetricEmitter.emitHistogramMetric(INVOKER_TOTALMEM_BLACKBOX, schedulingState.blackboxInvokers.foldLeft(0L) {
+      (total, curr) =>
+        if (curr.status.isUsable) {
+          curr.id.userMemory.toMB + total
+        } else {
+          total
+        }
```
markusthoemmes: Not sure if the compile accepts that as a total function, but you can try:

```scala
case (total, curr) if curr.status.isUsable => curr.id.userMemory.toMB + total
case _ => total
```

If the compiler doesn't accept that, it's fine to leave as is anyway.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala
```diff
@@ -244,23 +294,33 @@ class ShardingContainerPoolBalancer(
...
+          if (isOverloaded) {
+            if (isBlackboxInvocation) {
+              MetricEmitter.emitCounterMetric(LoggingMarkers.BLACKBOX_SYSTEM_OVERLOAD)
+            } else {
+              MetricEmitter.emitCounterMetric(LoggingMarkers.MANAGED_SYSTEM_OVERLOAD)
+            }
+          }
+          invoker
+        }
+      }
```
markusthoemmes: As you're only executing a side-effect on the option, please use `foreach` and just return the outer invoker value, like this:

```scala
invoker.foreach {
  case (_, true) =>
    val metric = if(isBlackboxInvocation) LoggingMarkers.BLACKBOX_SYSTEM_OVERLOAD else LoggingMarkers.MANAGED_SYSTEM_OVERLOAD
    MetricEmitter.emitCounterMetric(metric)
  case _ =>
}
invoker.map(_._1)
```

You don't care about the return values of the `emitCounterMetric`, hence they must produce a side-effect. If not for any performance concerns (which is not the case here), one should always use `foreach` to produce those, because `foreach` returns Unit, signifying the side-effect.

tests/src/test/scala/org/apache/openwhisk/core/loadBalancer/test/ShardingContainerPoolBalancerTests.scala
```diff
@@ -86,6 +86,7 @@ class ShardingContainerPoolBalancerTests
   behavior of "ShardingContainerPoolBalancerState"
 
   val defaultUserMemory: ByteSize = 1024.MB
+  val actionType: String = "managed"
```
markusthoemmes: This should be unused now.

tests/src/test/scala/org/apache/openwhisk/core/loadBalancer/test/ShardingContainerPoolBalancerTests.scala
```diff
@@ -288,14 +300,19 @@ class ShardingContainerPoolBalancerTests
...
     result shouldBe expectedResult
 
     // more schedules will result in randomized invokers, but the unhealthy and offline invokers should not be part
     val bruteResult = (0 to 100).map { _ =>
-      ShardingContainerPoolBalancer.schedule(1, fqn, invokers, invokerSlots, 1, index = 0, step = 1).get.toInt
+      ShardingContainerPoolBalancer
+        .schedule(1, fqn, invokers, invokerSlots, 1, index = 0, step = 1)
+        .get
+        ._1
+        .toInt
```
markusthoemmes: This test should also verify that `_2` is actually `true`

SugandhaAgrawal: Thanks for the comments @markusthoemmes. Will work on them 👍

PR #4220, opened 2019-01-16
Make fraction of managed invokers configurable.

cbickel:
This PR makes the fraction of the pool of managed invokers configurable. This also allows to have an overlap between the pool of blackbox invokers and the pool of managed invokers.

## Related issue and scope
The topic has been discussed here: https://lists.apache.org/thread.html/508e10ec9e800bb239363861385818c95d90ad1aa7df64f3b8904580@%3Cdev.openwhisk.apache.org%3E

- [ ] API
- [ ] Controller
- [x] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -60,6 +60,7 @@ limits:
 
 # Moved here to avoid recursions. Please do not use outside of controller-dict.
 __controller_ssl_keyPrefix: "controller-"
+__controller_blackbox_fraction: 0.10
 
 # port means outer port
 controller:
@@ -69,6 +70,8 @@ controller:
   basePort: 10001
   heap: "{{ controller_heap | default('2g') }}"
   arguments: "{{ controller_arguments | default('') }}"
+  managedFraction: "{{ controller_managed_fraction | default(1.0 - (controller_blackbox_fraction | default(__controller_blackbox_fraction))) }}"
+  blackboxFraction: "{{ controller_blackbox_fraction | default(__controller_blackbox_fraction) }}"
   timeoutFactor: "{{ controller_timeout_factor | default(2) }}"
   instances: "{{ groups['controllers'] | length }}"
   localBookkeeping: "{{ controller_local_bookkeeping | default('false') }}"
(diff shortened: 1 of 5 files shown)


core/controller/src/main/resources/reference.conf
```diff
@@ -6,6 +6,7 @@ whisk {
     use-cluster-bootstrap: false
   }
   loadbalancer {
+    managed-fraction: 90%
```
markusthoemmes: This means, that a deployment which adjusts the blackbox-fraction today will have overlapping pools in the future if the managed-fraction is not adjusted?

ansible/group_vars/all
```diff
@@ -69,7 +69,6 @@ controller:
   basePort: 10001
   heap: "{{ controller_heap | default('2g') }}"
   arguments: "{{ controller_arguments | default('') }}"
-  blackboxFraction: "{{ controller_blackbox_fraction | default(0.10) }}"
```
markusthoemmes: Can we keep this default in for now and also define a default "managedFraction" of 1-blackboxFraction to not run at the risk of configuring overlapping pools in existent deployments that do not define a managedFraction?

tests/src/test/scala/org/apache/openwhisk/core/loadBalancer/test/ShardingContainerPoolBalancerTests.scala
```diff
@@ -96,7 +96,7 @@ class ShardingContainerPoolBalancerTests
     IndexedSeq.fill(count)(new NestedSemaphore[FullyQualifiedEntityName](max))
 
   def lbConfig(blackboxFraction: Double) =
-    ShardingContainerPoolBalancerConfig(blackboxFraction, 1)
+    ShardingContainerPoolBalancerConfig(1.0 - blackboxFraction, blackboxFraction, 1)
```
markusthoemmes: Do we need a test to assert that we correctly handle overlapping pools? I think adding a test setting both values to 1 (which should then yield the same pools) should be sufficient.

core/controller/src/main/resources/reference.conf
cbickel: After your proposal of keeping the default in `group_vars/all`, the issue doesn't exists anymore here.

PR #4221, opened 2019-01-16
during api create, correctly handle non-default (i.e. "_") namespaces

mdeuser:
## Description
When creating an API with a specified namespace, the "wsk api create" command will override the namespace with the "__ow_user" value which is not the intended namespace.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [x] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/routemgmt/common/apigw-utils.js
+++ b/core/routemgmt/common/apigw-utils.js
@@ -853,7 +853,8 @@ function updateNamespace(apidoc, namespace) {
     if (apidoc.action) {
       // The action namespace does not have to match the CLI user's namespace
       // If it is different, leave it alone; otherwise use the replacement namespace
-      if (apidoc.namespace === apidoc.action.namespace) {
+      // And only replace when the namespace is the default '_' which needs replacement
+      if (apidoc.namespace === apidoc.action.namespace && apidoc.namespace === '_') {
         apidoc.action.namespace = namespace;
         apidoc.action.backendUrl = replaceNamespaceInUrl(apidoc.action.backendUrl, namespace);      }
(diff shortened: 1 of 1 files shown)


core/routemgmt/common/apigw-utils.js
```diff
@@ -853,7 +853,8 @@ function updateNamespace(apidoc, namespace) {
     if (apidoc.action) {
       // The action namespace does not have to match the CLI user's namespace
       // If it is different, leave it alone; otherwise use the replacement namespace
-      if (apidoc.namespace === apidoc.action.namespace) {
+      // And only replace when the namespace is the default '_' which needs replacement
+      if (apidoc.namespace === apidoc.action.namespace && apidoc.namespace === '_') {
```
csantanapr: hmm having trouble reading this why not instead `if (apidoc.action.namespace == '_' ) {`

mdeuser: good question.  there are two "namespace" values at play here.  one is the .wskprops namespace `apidoc.namespace` and the other is the namespace of the target action `apidoc.action.namespace`.  i believe your suggestion would also work.  the case where the action namespace is `_` while the target namespace is non-default is unusual and i believe not possible with the cli.

rabbah: The cli allows on to set their own namespace. It’s not as useful now because the key is 1-1 with the namespace.

PR #4225, opened 2019-01-21
Change prewarm container

ningyougang:
- Add prewarm container
- Delete prewarm container

Sometimes, the operator wants to add/delete prewarm container to invoker dynamically
For example: under some condition, operator can add the extra runtime perwarm container number(or reduce the number) in advance.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [x] Deployment
- [ ] CLI
- [ ] General tooling
- [x] Documentation

## Types of changes

## Checklist:

- [] I added tests to cover my changes.
- [] My changes require further changes to the documentation.

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -88,6 +88,8 @@ controller:
   authentication:
     spi: "{{ controller_authentication_spi | default('') }}"
   loglevel: "{{ controller_loglevel | default(whisk_loglevel) | default('INFO') }}"
+  username: "{{ controller_username | default(operation.username) }}"
+  password: "{{ controller_password | default(operation.password) }}"
   entitlement:
     spi: "{{ controller_entitlement_spi | default('') }}"
   protocol: "{{ controller_protocol | default('https') }}"
@@ -183,6 +185,8 @@ invoker:
   docker:
     become: "{{ invoker_docker_become | default(false) }}"
   loglevel: "{{ invoker_loglevel | default(whisk_loglevel) | default('INFO') }}"
+  username: "{{ invoker_username | default(operation.username) }}"
+  password: "{{ invoker_password | default(operation.password) }}"
   jmxremote:
     jvmArgs: "{% if inventory_hostname in groups['invokers'] %}
     {{ jmx.jvmCommonArgs }} -Djava.rmi.server.hostname={{ invokerHostname }} -Dcom.sun.management.jmxremote.rmi.port={{ jmx.rmiBasePortInvoker + groups['invokers'].index(inventory_hostname) }} -Dcom.sun.management.jmxremote.port={{ jmx.basePortInvoker + groups['invokers'].index(inventory_hostname) }}
@@ -379,3 +383,8 @@ metrics:
     port: "{{ metrics_kamon_statsd_port | default('8125') }}"
 
 user_events: "{{ user_events_enabled | default(false) }}"
+
+# Operation configuation
+operation:
+    username: "{{ operation_username | default('admin') }}"
+    password: "{{ operation_password | default('admin') }}"
--- a/ansible/roles/controller/tasks/deploy.yml
+++ b/ansible/roles/controller/tasks/deploy.yml
@@ -57,6 +57,18 @@
     dest: "{{ controller.confdir }}/{{ controller_name }}/jmxremote.access"
     mode: 0777
 
+- name: copy controller auth username file
+  template:
+    src: "controllerauth.username.j2"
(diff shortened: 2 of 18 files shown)


markusthoemmes:
Thanks for the contribution. This looks like the inception of an active admin interface into the invoker, correct?

If so, I think we need a broader discussion on how to achieve that and for this case specifically: Do we want different invokers with different prewarming settings? Manipulating a single invoker in this case results in a divergence to the global configuration. If that invoker restarts, what happens?

My gut-feeling in this case is that you'd want a propagation mechanism to roll out a new config to all invokers on the fly.

I think this warrants a discussion on the dev-list.

rabbah:
@ningyougang thank you for this contribution --- having an "admin" interface to the controller and invoker has been something we've been missing. I do think this warrants a bit of discussion. 

It's possible as Markus notes that you can get divergent invokers - I'd imagine one would rolls this out by running some script or playbook to update multiple invokers. A restart doesn't affect correctness, just performance... note however it is possible to have different runtime manifests which could lead to different functional behaviors (if you exclude a runtime entirely from the manifest, then actions allocated to that invoker will fail to run).

I can think of different ways to also approach this - say via the existing protocols for communication between the load balancer and the invokers. In particular, if admin changes were front-doored by the controller, you can keep the manifest consistent in the system. Although you might be deliberately thinking of having a heterogenous mix of stem cells.

This is great to see prototyped and give us something concrete to evaluate and discuss.

ningyougang:
@markusthoemmes @rabbah 
### Background
I think openwhisk needs some operation tools(page or API) to admin, for example:
* real-time status of Action, Controller, Invoker, etc.
* Change prewarm runtime container for invokers

here, we just discuss `Change prewarm container for invokers`

### Requirements

* Admin can add/delete extra prewarm container via controller.
  The flow may like this, admin sends a HTTP request to the controller, controller sends that request info to all invokers through kafka,
  invoker gets the request info, parse the prewarm runtime info, and do `add/delete prewarm container`
  
  Note, the prewarm runtime request info's image should be included in `${OPENWHISK_HOME}/ansible/files/runtimes.json` ,
  if `unknow prewarm runtime image` is sent to invoker, invoker should reject the request.


* Need to support `different prewarm runtime` for different invokers?
In future, invokers may support `group concept`, which means, some invokers's cpu, memory is very good, these invokers can execute some height works.
but some invokers may execute some light works.
In this situation, we may consider `heterogeneous invoker cluster support`, so invokers needs to support  `different prewarm runtime` also.
So admin can send the http request to invoker to add/delete prewarm container.

### Question?
```
 Manipulating a single invoker in this case results in a divergence to the global configuration. If that invoker restarts, what happens?
```
If invoker restarts,  the invoker will read the global runtime.json to create prewarm container( the extra adding prewarm container will be gone, because the invoker can't remember its previous state)

style95:
IMHO with regard to the admin API, I have been requested from users that they want to see the real-time status of their invocation.
For example, one of my users invokes an action with 200 ~ 300 concurrent containers for about 4 ~ 5 mins. In this case, he cannot call it with blocking call, he calls them asynchronously.
And at some point, he wants to make sure all his concurrent invocation finished.
Currently, he needs to indirectly poll activation and count the number of finished invocation.

I think we need to track the invocation status for each users in the future.(Though there could be some differences because invocation duration can be short.)
It can be achieved with controller API as well, but invoker-side API would give us more fine-grained control as containers are actually running on the invoker side.

One more point is what @ningyougang said.
In the future, there will be some needs for different kinds of container pool.
For example, users might need GPU enabled host, and users might expect more powerful hosts or dedicated resources with higher prices. Or OW operator(public cloud) might want to deploy invoker in many different types of host machines to maximize the utilization of computing resources in the cloud. And the heterogenous invoker cluster requires heterogenous prewarm configuration for efficiency and flexbility. 
This is why I think the API should exist in each invoker and it is same for controllers as well.

style95:
And one of my collegues who is public cloud operator said, he noticed that invocation pattern of each runtimes changes over time. For example, python actions mostly invoked during the morning, PHP actions are invoked during the evening and so on. Also invocation patterns change based on the social events such as big conferences, world cup, big news and so on. He is thinking of dynamically changing the prewarm configuration based on the request pattern prediction with machine learning to minimize cold start. I think this feature can be provided outside of OW and per-invoker or per-controller API would be a good building block for this.

ningyougang:
I aready added the logic of `add/delete prewarm container via controller to all managed invokers`.

ningyougang:
Use another pr: https://github.com/apache/openwhisk/pull/4790
So close this pr.

PR #4229, opened 2019-01-22
Add support for TTL at collection level with CosmosDB

chetanmeh:
Enables specifying [Time to live][1] at collection (or container level). 

## Description

For collection like `Activations` its desirable to expire the activations after some time. For CouchDB this is done by running a [cleaner script][2]. For CosmosDB we can make use of the [TTL][1] support which allows auto removal of entries after the TTL time has passed after their last modified time.

So to specify TTL for activations

```
whisk {
    cosmosdb {
        collections {
            WhiskActivation {
                time-to-live = 30 days
            }
        }
    }
}
```

This PR also now sets the default TTL to -1 which allow us to later specify TTL at individual item level if required.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

[1]: https://docs.microsoft.com/en-us/azure/cosmos-db/time-to-live
[2]: https://github.com/apache/incubator-openwhisk/blob/master/tools/db/cleanUpActivations.py

### DIFF
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -171,6 +171,8 @@ whisk {
         consistency-level = "Session"
 
         # TTL duration. By default no TTL is set unless explicitly configured
+        # ENABLING THIS VALUE MEANS YOUR DATA WILL NOT BE PERMANENTLY STORED
+        # Can only be used for `WhiskActivation` for now
     #   time-to-live      = 60 s
         connection-policy {
             max-pool-size = 1000
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -159,9 +159,9 @@ whisk {
     # CosmosDB related configuration
     # For example:
     cosmosdb {
-    #   endpoint          =               # Endpoint URL like https://<account>.documents.azure.com:443/
-    #   key               =               # Access key
-    #   db                =               # Database name
+        # endpoint          =               # Endpoint URL like https://<account>.documents.azure.com:443/
+        # key               =               # Access key
+        # db                =               # Database name
         # Throughput configured for each collection within this db
         # This is configured only if collection is created fresh. If collection
         # already exists then existing throughput would be used
@@ -173,7 +173,7 @@ whisk {
         # TTL duration. By default no TTL is set unless explicitly configured
         # ENABLING THIS VALUE MEANS YOUR DATA WILL NOT BE PERMANENTLY STORED
         # Can only be used for `WhiskActivation` for now
-    #   time-to-live      = 60 s
+        # time-to-live      = 60 s
         connection-policy {
             max-pool-size = 1000
             # When the value of this property is true, the SDK will direct write operations to
@@ -201,13 +201,13 @@ whisk {
         #   - WhiskEntity
         #   - WhiskActivation
         # For example if multiple writes need to be enabled for activations then
-    #   collections {
-    #       WhiskActivation {            # Add entity specific overrides here
-    #           connection-policy {
-    #              using-multiple-write-locations = true
-    #            }
-    #       }
-    #   }
+        # collections {
+        #   WhiskActivation {            # Add entity specific overrides here
+        #     connection-policy {
+        #        using-multiple-write-locations = true
+        #     }
+        #   }
+        # }
     }
 
     # transaction ID related configuration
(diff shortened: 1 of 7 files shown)


tests/src/test/scala/org/apache/openwhisk/core/database/cosmosdb/CosmosDBConfigTests.scala
```diff
@@ -114,7 +114,7 @@ class CosmosDBConfigTests extends FlatSpec with Matchers {
       | }
          """.stripMargin).withFallback(globalConfig)
     val cosmos = CosmosDBConfig(config, "WhiskAuth")
-    cosmos should matchPattern { case CosmosDBConfig("http://localhost", "foo", "openwhisk", _, _, _) => }
+    cosmos should matchPattern { case CosmosDBConfig("http://localhost", "foo", "openwhisk", _, _, _, _) => }
```
tysonnorris: I think some other tests where config is created for each test have the config creation wrapped into a function, so that changing the config sig is a one line change; might be useful here too.

common/scala/src/main/resources/application.conf
```diff
@@ -169,6 +169,9 @@ whisk {
         # Select from one of the supported
         # https://azure.github.io/azure-cosmosdb-java/1.0.0/com/microsoft/azure/cosmosdb/ConsistencyLevel.html
         consistency-level = "Session"
+
+        # TTL duration. By default no TTL is set unless explicitly configured
+    #   time-to-live      = 60 s
```
tysonnorris: It may not be obvious that this example (60s) means documents will only reside in any of these collections for 60s. 
I would 
* emphasize this a bit more (e.g. ENABLING THIS VALUE  MEANS YOUR DATA WILL NOT BE PERMANENTLY STORED)
* disable this for WhiskAuth + WhiskEntity (is there any good use case for these?)
* or maybe only include any sample config only for WhiskActivation collection (this would be a more typical real config right?)

common/scala/src/main/scala/org/apache/openwhisk/core/database/cosmosdb/CosmosDBConfig.scala
```diff
@@ -35,7 +35,8 @@ case class CosmosDBConfig(endpoint: String,
                           db: String,
                           throughput: Int,
                           consistencyLevel: ConsistencyLevel,
-                          connectionPolicy: ConnectionPolicy) {
+                          connectionPolicy: ConnectionPolicy,
+                          timeToLive: Option[Duration] = None) {
```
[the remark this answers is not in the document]
chetanmeh: Want to use default of -1 here but that does not map to `Duration`. Dropped the `None` part now as pureconfig would set it anyway to None

tests/src/test/scala/org/apache/openwhisk/core/database/cosmosdb/CosmosDBSupportTests.scala
```diff
@@ -47,6 +50,26 @@ class CosmosDBSupportTests extends FlatSpec with CosmosDBTestSupport with MockFa
...
+      | }
+         """.stripMargin).withFallback(ConfigFactory.load())
+
+    val cosmosDBConfig = CosmosDBConfig(config, "WhiskActivation")
+    cosmosDBConfig.timeToLive shouldBe Some(60.seconds)
+
+    val testDb = createTestDB()
+    val testConfig = cosmosDBConfig.copy(db = testDb.getId)
+    val (_, coll) = new CosmosTest(testConfig, client, ActivationViewMapper).initialize()
+    coll.getDefaultTimeToLive shouldBe 60.seconds.toSeconds
```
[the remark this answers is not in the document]
chetanmeh: That was more to emphasize that config used 60 secs. Had test config used says minutes then that would help

tests/src/test/scala/org/apache/openwhisk/core/database/cosmosdb/CosmosDBConfigTests.scala
chetanmeh: Problem here is the need to perform assert on created config. Not able to abstract that out

common/scala/src/main/resources/application.conf
chetanmeh: For `WhiskAuth` and `WhiskEntity` yes it does not make sense to modify the default. added checks now to ensure that its only used for `WhiskActivation`

chetanmeh:
Updated indentation

PR #4230, opened 2019-01-23
Complete blocking activations immediately

jiangpengcheng:
Complete blocking activations immediately when a Left(activationId) is returned and controllerActivationConfig.pollingFromDb is false, otherwise this invocation will wait until timeout

## Related issue and scope
- [ ] I opened an issue to propose and discuss this change (#4231)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/controller/actions/PrimitiveActions.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/controller/actions/PrimitiveActions.scala
@@ -599,7 +599,8 @@ protected[actions] trait PrimitiveActions {
       case Right(activation) => result.trySuccess(Right(activation))
       case _ if (controllerActivationConfig.pollingFromDb) =>
         pollActivation(docid, context, result, i => 1.seconds + (2.seconds * i), maxRetries = 4)
-      case _ =>
+      case Left(activationId) =>
+        result.trySuccess(Left(activationId)) // complete the future immediately if it's configured to not poll db for blocking activations
     }
 
     if (controllerActivationConfig.pollingFromDb) {
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/controller/actions/PrimitiveActions.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/controller/actions/PrimitiveActions.scala
@@ -654,7 +654,7 @@ protected[actions] trait PrimitiveActions {
   /** Max atomic action count allowed for sequences */
   private lazy val actionSequenceLimit = whiskConfig.actionSequenceLimit.toInt
 
-  private val controllerActivationConfig =
+  protected val controllerActivationConfig =
     loadConfigOrThrow[ControllerActivationConfig](ConfigKeys.controllerActivation)
 
 }
--- a/tests/src/test/scala/org/apache/openwhisk/core/controller/test/PollingFromDbTests.scala
+++ b/tests/src/test/scala/org/apache/openwhisk/core/controller/test/PollingFromDbTests.scala
@@ -32,7 +32,6 @@ import spray.json._
 import spray.json.DefaultJsonProtocol._
 import org.apache.openwhisk.core.controller.WhiskActionsApi
 import org.apache.openwhisk.core.entity._
-import org.apache.openwhisk.core.database.UserContext
 import org.apache.openwhisk.core.controller.actions.ControllerActivationConfig
 
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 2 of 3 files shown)


core/controller/src/main/scala/org/apache/openwhisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -599,7 +599,8 @@ protected[actions] trait PrimitiveActions {
       case Right(activation) => result.trySuccess(Right(activation))
       case _ if (controllerActivationConfig.pollingFromDb) =>
         pollActivation(docid, context, result, i => 1.seconds + (2.seconds * i), maxRetries = 4)
-      case _ =>
+      case Left(activationId) =>
+        result.trySuccess(Left(activationId)) // complete the future immediately if it's configured to not poll db for blocking activations
```
rabbah: hmm - convince me you actually tested this change.

jiangpengcheng: I created an issue #4231

jiangpengcheng:
@dubee 

> @jiangpengcheng, how were you able to reproduce this problem? I looked at the related issue, seems to involve an activation that is too large for Kafka, is that correct? My understanding is that activations that are too large for Kafka should be truncated by the invoker.

my thought is that once `RecordTooLargeException` is threw and catched, invoker will send a `ResultMessage(tid, Left(activationId))` too Kafka

do you mean that since invoker will truncate the activation if it is too large, the `RecordTooLargeException` will never be threw?

rabbah:
We need a unit test for this. There is a loadbalancer trait in the controller common test harness that will allow you to simulate such a scenario and test what happens when the blocking activation responds in this way.

core/controller/src/main/scala/org/apache/openwhisk/core/controller/actions/PrimitiveActions.scala
markusthoemmes: This change is sound. This block definitely should resolve the promise on every branch in some way or the other. If database polling is disabled, the promise is essentially blackholed.

A RecordTooLargeException results indeed in the fixed `Left` case, see https://github.com/apache/incubator-openwhisk/blob/7ff15502949fa21ab5621825204397ad9eb6dfe5/core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala#L168-L171

jiangpengcheng:
@dubee you are right, I checked that when the activation's response is too large, it will be truncated and  seems the `RecordTooLargeException` will never be threw

rabbah:
This happens when the message that posts to Kafka is too large. It’s a corner case bug which I think we haven’t fixed (didn’t check): because the max response size and the Kafka message size are the same where the latter needs a little bit of slack forbtge metadata.

tests/src/test/scala/org/apache/openwhisk/core/controller/test/PollingFromDbTests.scala
```diff
@@ -0,0 +1,104 @@
...
+    extends DegenerateLoadBalancerService(config) {
+
+  override def activeActivationsFor(namespace: UUID): Future[Int] = Future.successful(0)
+  override def totalActiveActivations: Future[Int] = Future.successful(0)
+
+  override def publish(action: ExecutableWhiskActionMetaData, msg: ActivationMessage)(
+    implicit transid: TransactionId): Future[Future[Either[ActivationId, WhiskActivation]]] = {
+    Future {
+      Future {
+        Left(msg.activationId)
```
rabbah: scala tip: this can be `Future.successful(Future.successful(Left(...))`

tests/src/test/scala/org/apache/openwhisk/core/controller/test/PollingFromDbTests.scala
```diff
@@ -0,0 +1,104 @@
...
+  def aname() = MakeName.next("action_tests")
+  override protected val controllerActivationConfig = ControllerActivationConfig(false)
+
+  override val loadBalancer = new AlwaysReturnLeftLoadBalancerService(whiskConfig)
+
+  it should "complete blocking activation request immediately when a Left is returned and pollingFromDb is set to false" in {
+    implicit val tid = transid()
+    val timeLimit = TimeLimit(1.minute)
+    val action = WhiskAction(namespace, aname(), jsDefault("??"), limits = ActionLimits(timeLimit))
+    val activation = WhiskActivation(
```
rabbah: i think you can get away without having to create this object at all *you're only using the activation start and activation Id.

PR #4234, opened 2019-01-23
Add limit to not store activations for a limitted namespace.

cbickel:
With this PR, operators will be able to disable storing of activations into activations store.
The flag is implemented like the per-namespace-limits. It can be set with wskadmin.

This PR was the initial idea of #4078. In the meantime, the idea in the other PR changed, to implement a throttle instead of a switch. But as this is a completely new type of rate-limit (which does not only allow or deny requests) that's a bit bigger to implement. So I'll go with a staged approach here and implement it as switch first.
If someone needs a throttle instead of completely switching off the writes to the DB, this can be brought up again and the throttle can be built up on the solution we already have.

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [x] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Triggers.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Triggers.scala
@@ -148,28 +148,32 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
             if (activeRules.nonEmpty) {
               val triggerActivationId = activationIdFactory.make()
               logging.info(this, s"[POST] trigger activation id: ${triggerActivationId}")
-              val triggerActivation = WhiskActivation(
-                namespace = user.namespace.name.toPath, // all activations should end up in the one space regardless trigger.namespace
-                entityName.name,
-                user.subject,
-                triggerActivationId,
-                Instant.now(Clock.systemUTC()),
-                Instant.EPOCH,
-                response = ActivationResponse.success(payload orElse Some(JsObject.empty)),
-                version = trigger.version,
-                duration = None)
-              val args: JsObject = trigger.parameters.merge(payload).getOrElse(JsObject.empty)
+              if (user.limits.storeActivations.getOrElse(true)) {
+                val triggerActivation = WhiskActivation(
+                  namespace = user.namespace.name.toPath, // all activations should end up in the one space regardless trigger.namespace
+                  entityName.name,
+                  user.subject,
+                  triggerActivationId,
+                  Instant.now(Clock.systemUTC()),
+                  Instant.EPOCH,
+                  response = ActivationResponse.success(payload orElse Some(JsObject.empty)),
+                  version = trigger.version,
+                  duration = None)
+                val args: JsObject = trigger.parameters.merge(payload).getOrElse(JsObject.empty)
 
-              activateRules(user, args, trigger.rules.getOrElse(Map.empty))
-                .map(results => triggerActivation.withLogs(ActivationLogs(results.map(_.toJson.compactPrint).toVector)))
-                .recover {
-                  case e =>
-                    logging.error(this, s"Failed to write action activation results to trigger activation: $e")
-                    triggerActivation
-                }
-                .map { activation =>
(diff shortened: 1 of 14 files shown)


core/controller/src/main/scala/org/apache/openwhisk/core/controller/Triggers.scala
```diff
@@ -148,28 +148,32 @@ trait WhiskTriggersApi extends WhiskCollectionAPI {
...
-                .map(results => triggerActivation.withLogs(ActivationLogs(results.map(_.toJson.compactPrint).toVector)))
-                .recover {
-                  case e =>
-                    logging.error(this, s"Failed to write action activation results to trigger activation: $e")
-                    triggerActivation
-                }
-                .map { activation =>
-                  activationStore.store(activation, context)
-                }
+              if (user.limits.storeActivations.getOrElse(true)) {
```
markusthoemmes: The `true` default is scattered throughout the codebase now. I'm wondering if there is a good way of centralizing that.

One way is to add a `val/def` to the `UserLimits` class, which defaults to true always.

The other way is to add a system "limit" which makes it possible to enable/disable this for the whole system, therefore acting as the default.

I realize though that the second approach might be a bigger undertaking. Maybe it's fine as-is and we're better off having the default at the actual caller side to later get rid off.

More generally speaking: We should think about building a more generalized cascading system for our limits. I envision something like this:

1. Load the user-limits from the database
2. Merge them with the system limits
3. Pass the merged versions through the code (maybe as part of the identity class)

That'd help us in cases as those and potentially transforms all the Option to actual values before we hit the location of usage.

You can ignore this comment if you want to, just wanted to leave the thoughts here.

core/controller/src/main/scala/org/apache/openwhisk/core/controller/actions/PrimitiveActions.scala
```diff
@@ -569,7 +569,9 @@ protected[actions] trait PrimitiveActions {
       }
     }
 
-    activationStore.store(activation, context)(transid, notifier = None)
+    if (user.limits.storeActivations.getOrElse(true)) {
```
markusthoemmes: Can/should we move this into the ActivationStore itself? The user's identity is available in there as well.

I guess the thought was to not do that, because then each store needs to make sure to implement this if? Can we instead maybe come up with a wrapper function for all `activationStore.store` calls that takes care of a general concern like this?

tools/admin/wskadmin
```diff
@@ -132,6 +132,8 @@ def parseArgs():
     subcmd.add_argument('--firesPerMinute', help='trigger fires per minute allowed', type=int)
     subcmd.add_argument('--concurrentInvocations', help='concurrent invocations allowed for this namespace', type=int)
     subcmd.add_argument('--allowedKinds', help='list of runtime kinds allowed in this namespace', nargs='+', type=str)
+    subcmd.add_argument('--enableStoreActivations', help='enable storing of activations to datastore for this namespace', default=None, dest='storeActivations', action='store_true')
+    subcmd.add_argument('--disableStoreActivations', help='disable storing of activations to datastore for this namespace', default=None, dest='storeActivations', action='store_false')
```
markusthoemmes: Thinking about scriptability here: Would it be easier to just use a value `storeActivations=true/false` instead of two distinct flags? If I were to use `wskadmin` in a script, I much rather do `wskadmin --storeActivations=$VALUE` than replacing the whole flag if that makes sense?

cbickel: I can have another look on that one 👍

core/controller/src/main/scala/org/apache/openwhisk/core/controller/actions/PrimitiveActions.scala
cbickel: Now, I introduced the method `storeAfterCheck`, to do exactly that in ActivationStore, independant of the implementation.

Issue #3861, opened 2018-07-09
is WhiskProperties.useCLIDownload deadcode now?


rabbah:
I don't think we use this flag anymore:

https://github.com/apache/incubator-openwhisk/blob/f6046721acd801ea1777cb5cb040dfaa6922a18e/tests/src/test/scala/common/WskCliOperations.scala#L962-L966

@houshengbo @dubee might you know if this is dead code now?

drcariel:
If this is dead code what would be the default? To always use `WhiskProperties.getCLIPath` AKA `WHISK_HOME/bin/wsk`?

PR #4253, opened 2019-02-07
Remove deadcode.

rabbah:
Closes https://github.com/apache/incubator-openwhisk/issues/3861.

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3861)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/tests/src/test/scala/common/WskCliOperations.scala
+++ b/tests/src/test/scala/common/WskCliOperations.scala
@@ -972,10 +972,5 @@ class CliGatewayOperations(val wsk: RunCliCmd) extends GatewayOperations {
 
 object Wsk {
   val binaryName = "wsk"
-  val defaultCliPath = if (WhiskProperties.useCLIDownload) getDownloadedGoCLIPath else WhiskProperties.getCLIPath
-
-  /** What is the path to a downloaded CLI? **/
-  private def getDownloadedGoCLIPath = {
-    s"${System.getProperty("user.home")}${File.separator}.local${File.separator}bin${File.separator}${binaryName}"
-  }
+  val defaultCliPath = WhiskProperties.getCLIPath
 }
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/common/WskCliOperations.scala
```diff
@@ -972,7 +972,7 @@ class CliGatewayOperations(val wsk: RunCliCmd) extends GatewayOperations {
 
 object Wsk {
   val binaryName = "wsk"
-  val defaultCliPath = if (WhiskProperties.useCLIDownload) getDownloadedGoCLIPath else WhiskProperties.getCLIPath
+  val defaultCliPath = WhiskProperties.getCLIPath
```
dubee: Is the definition below for `getDownloadedGoCLIPath` still needed?

rabbah:
@mdeuser can you please take care of this one - if we don't need this code anywhere downstream let's remove it.

rabbah:
The code for "use.cli.download" should be redone - the path is assumed to be some arbitrary path really, might as well put it where the test harness expects it in openwhisk/bin. As such, since there is no reference in this project otherwise to the code, removing it makes sense to me.

mdeuser:
pg2 4208

rabbah:
An alternative to removing this boolean flag and the arbitrary location for the binary, is to have the ansible parameter point to a specific location instead. I'm not doing this since I see no real value in it or having the flag until someone tells me otherwise.

rabbah:
Thanks @dubee for the update. If we need to preserve the behavior I prefer we do it in an alternate way that's cleaner IMO (the param should point to the binary location).

Issue #4246, opened 2019-02-06
remove older runtimes


rabbah:
The runtimes manifest includes some stale/deprecated runtimes:
- nodejs (node 4)
- python (alias for python 2)
- swift 2?, 3, 3.1.1., 4.1
- php 7.1, 7.2

I propose cleaning up the manifest to remove these older runtimes, and to make node 10 the default or node actions and python 3 for python actions.

cc @akrabat @csantanapr

TPei:
From a development perspective, this would mean adapting the runtime.json, swagger, docs and test/dat files? I feel this might be somewhat beginner-friendly, so I'd like to express an interest in working on that :)

rabbah:
Sorry @TPei I didn't see your note - you'll see that PR fails tests because there's actually some dependence on the actual manifest used for deployment and what the tests are assuming. It turns out we have to add a separate manifest for the test. I'm happy to close my PR if you would like to take this one.

PR #4254, opened 2019-02-07
Remove older runtimes

TPei:
Removes some older runtimes (nodejs, python, swift, swift:3.1.1, swift:4.1, php:7.1, php:7.2) as proposed in #4246 and discussed in the Tech. Interchange Meeting

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#4246)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [x] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


docs/actions-swift.md
```diff
@@ -187,164 +187,7 @@ wsk action update helloSwiftly action-bin.zip --kind swift:4.2
 wsk action invoke helloSwiftly -r
 ```
 
-## Compiling Swift 4.1 packaged actions
```
TPei: should this be removed or does it make sense to update this for 4.2? or maybe replace with https://github.com/apache/incubator-openwhisk-runtime-swift#packaging-an-action-as-a-swift-executable-using-swift-4x ?

core/controller/src/main/resources/apiv1swagger.json
```diff
@@ -1889,21 +1889,16 @@
                         "blackbox",
                         "java",
                         "java:default",
-                        "nodejs:6",
```
rabbah: Node 6 starts.

core/controller/src/main/resources/apiv1swagger.json
```diff
@@ -1889,21 +1889,16 @@
...
                         "nodejs:8",
                         "nodejs:10",
                         "nodejs:default",
-                        "php:7.1",
-                        "php:7.2",
                         "php:7.3",
```
rabbah: Noticed we don’t have php:default let’s add it :) and the other runtimes that should have it too.

tests/dat/actions/unicode.tests/nodejs-6.txt
```diff
@@ -1,8 +0,0 @@
-// Licensed to the Apache Software Foundation (ASF) under one or more contributor
```
rabbah: Stays

docs/actions-swift.md
rabbah: The 3.1 steps should be removed. We can keep 4.x

rabbah: actually there's already a section: `Compiling a single source file for Swift 4.2` and `Compiling dependencies and multi-file projects for Swift 4.2` so what you've removed is fine.

TPei:
ActionsApiTest lines 1563-1612 currently fail
org.apache.openwhisk.core.controller.test.ActionsApiTests > Actions API should report proper error when runtime is deprecated FAILED

As far as I can tell, it was testing with Swift 3, which was deprecated. However, at the moment, there are no more deprecated runtimes, they were all removed.

Though I am not 100% sure what's actually going on there

Any thoughts @rabbah ?

docs/actions-php.md
```diff
@@ -23,15 +23,13 @@ The process of creating PHP actions is similar to that of [other actions](action
 The following sections guide you through creating and invoking a single PHP action,
 and demonstrate how to bundle multiple PHP files and third party dependencies.
 
-PHP actions are executed using PHP 7.3. PHP 7.2 & PHP 7.1 are also available. The specific
+PHP actions are executed using PHP 7.3. The specific
```
rabbah: @akrabat ok to drop older php kinds?

tests/src/test/scala/org/apache/openwhisk/core/containerpool/docker/test/DockerContainerFactoryTests.scala
```diff
@@ -60,7 +60,7 @@ class DockerContainerFactoryTests
 
   it should "set the docker run args based on ContainerArgsConfig" in {
 
-    val image = ExecManifest.runtimesManifest.manifests("nodejs").image
+    val image = ExecManifest.runtimesManifest.manifests("nodejs:10").image
```
rabbah: can we use default?

tests/src/test/scala/org/apache/openwhisk/core/entity/test/ExecHelpers.scala
```diff
@@ -50,52 +47,55 @@ trait ExecHelpers extends Matchers with WskActorSystem with StreamLogging {
   }
 
   protected def jsOld(code: String, main: Option[String] = None) = {
```
rabbah: my ide tell ms this method is not used.

tests/src/test/scala/org/apache/openwhisk/core/entity/test/ExecHelpers.scala
```diff
@@ -50,52 +47,55 @@ trait ExecHelpers extends Matchers with WskActorSystem with StreamLogging {
...
-    CodeExecAsString(RuntimeManifest(NODEJS, imagename(NODEJS), deprecated = Some(true)), trim(code), main.map(_.trim))
+    CodeExecAsString(
+      RuntimeManifest(NODEJS10, imagename(NODEJS10), deprecated = Some(true)),
+      trim(code),
+      main.map(_.trim))
   }
 
   protected def js(code: String, main: Option[String] = None) = {
```
rabbah: neither is this method.

tests/src/test/scala/org/apache/openwhisk/core/entity/test/SchemaTests.scala
```diff
@@ -476,19 +476,18 @@ class SchemaTests extends FlatSpec with BeforeAndAfter with ExecHelpers with Mat
 
   it should "initialize exec manifest" in {
     val runtimes = ExecManifest.runtimesManifest
-    runtimes.resolveDefaultRuntime("nodejs:default").get.kind shouldBe "nodejs:6"
-    runtimes.resolveDefaultRuntime("swift").get.deprecated shouldBe Some(true)
```
rabbah: ok to remove this as i'll add a separate test for deprecated runtime.

jasonpet:
this broke all kinds of actions in downstream repos.  the nodejs 6 runtime image contains many npm modules in it whereas the nodejs 10 one contains no npm modules.

jasonpet:
many install scripts and tests that assume the modules are there by will need to be updated to explicitly set the node.js kind to 6.

jasonpet:
if we want to stop using node.js 6 then the actions will need to be updated to zip actions and the necessary modules will need to be packaged into the zip.

rabbah:
If we pin the downstream repos to nodejs:6 kind that will suffice to resolve the breakage? Node 6 end of life is April.

rabbah:
I see the following fixes:
1.  change the default back to node6 --- the way the catalog is installed it doesn't specify any kinds 😱 and lets the cli assume the default runtime as inferred from the file extension
1. change the utils.sh script to allow for a kind to be specified
1. replace utils.sh with wskdeploy and use a kind explicity
1. modify node 8/10 runtime to install the same package (but perhaps not pin to the patch level to allow for patches to be picked up automatically)

as node6 is reaching end of life, any fix that still uses node 6 as the kind postpones having to do what @jasonpet notes above: either packaging the deps into the action or modifing the runtimes to include at least the dependencies needed by the catalog.

rabbah:
@jasonpet I created https://github.com/apache/incubator-openwhisk-catalog/issues/294. It looks like there is a PR to move to wskdeploy based installation and so now might be a good time to pick it up and push it through.

Issue #4002, opened 2018-09-05
Can I spawn action containers not only use docker?


YellyYU:
Hey guys, I'm planning to use openwhisk for serverless sandbox boot time analysis. I see that openwhisk uses docker containers to bundle and manage user codes, so I suppose somewhere in the server side docker command (or api) like `docker run` is called. Now I'm wondering is it possible for me to spawn containers using other commands other than docker, such as `runc run` command, through several modifications to the source code? I'm a new comer to openwhisk project and not very familiar to the whole picture of openwhisk, so if you can tell me where the codes spawning action containers are, and what's the recommending way to achieve what I want, that would be very nice. Thanks in advance!

markusthoemmes:
Hi @YellyYU,

yes this is definitly possible. In fact, the current implementation even mixes docker/runc commands for performance reasons.

In theory, you'll need just another implementation of the `Container` trait (see [`DockerContainer.scala`](https://github.com/apache/incubator-openwhisk/blob/master/core/invoker/src/main/scala/whisk/core/containerpool/docker/DockerContainer.scala) for an example implementation). The [`RuncClient`](https://github.com/apache/incubator-openwhisk/blob/master/core/invoker/src/main/scala/whisk/core/containerpool/docker/RuncClient.scala) is not fully implemented as of today because we didn't need it, feel free to add whatever you need there.

I hope that helps!

rabbah:
@YellyYU i hope the above answered your question - please reopen if needed.

rabbah:
@sven-lange-last it would be useful to document the assumptions - where is a good place to do that?

sven-lange-last:
@rabbah I opened #4255 to put the documentation directly in the code into the trait that needs to be implemented.

PR #4255, opened 2019-02-07
Improve container factory documentation

sven-lange-last:
Explicitly list some of the expectations and requirements an actual container factory / container implementation needs to satisfy. Addresses #4002.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [x] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
@@ -52,7 +52,23 @@ case class ContainerPoolConfig(userMemory: ByteSize, concurrentPeekFactor: Doubl
  */
 trait ContainerFactory {
 
-  /** create a new Container */
+  /**
+   * Create a new Container
+   *
+   * The created container has to satisfy following requirements:
+   * - The container's file system is based on the provided action image and has a read/write layer on top.
+   * - If the specified image is not available on the system, it is pulled from an image
+   *   repository - for example, Docker Hub.
+   * - The container has a network interface which is able to connect to the internet
+   *   (in particular, to the specified DNS servers). In addition, the invoker must be able
+   *   to connect to the container's interface.
+   * - The IPv4 address of said interface is stored in the created Container instance.
+   * - The default process specified in the action image is run.
+   * - It is desired that all stdout / stderr written by processes in the container is captured such
+   *   that it can be obtained using the logs() method of the Container trait.
+   * - It is desired that the container supports and enforces the specified memory limit and CPU shares.
+   *   In particular, action memory limits rely on the underlying container technology.
+   */
   def createContainer(tid: TransactionId,
                       name: String,
                       actionImage: ExecManifest.ImageName,
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
@@ -56,13 +56,16 @@ trait ContainerFactory {
    * Create a new Container
    *
    * The created container has to satisfy following requirements:
-   * - The container's file system is based on the provided action image and has a read/write layer on top.
+   * - The container's file system is based on the provided action image and may have a read/write layer on top.
+   *   Some managed action runtimes may need the capability to write files.
    * - If the specified image is not available on the system, it is pulled from an image
    *   repository - for example, Docker Hub.
-   * - The container has a network interface which is able to connect to the internet
-   *   (in particular, to the specified DNS servers). In addition, the invoker must be able
-   *   to connect to the container's interface.
-   * - The IPv4 address of said interface is stored in the created Container instance.
+   * - The container needs a network setup - usually, a network interface - such that the invoker is able
+   *   to connect the action container. The container must be able to perform DNS resolution based
+   *   on the settings provided via ContainerArgsConfig. If needed by action authors,
+   *   the container should be able to connect to other systems or even the internet to consume services.
+   * - The IP address of said interface is stored in the created Container instance if you want to use
+   *   the standard init / run behaviour defined in the Container trait.
    * - The default process specified in the action image is run.
    * - It is desired that all stdout / stderr written by processes in the container is captured such
    *   that it can be obtained using the logs() method of the Container trait.


sven-lange-last:
@marc-schwind please have a look.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
```diff
@@ -52,7 +52,23 @@ case class ContainerPoolConfig(userMemory: ByteSize, concurrentPeekFactor: Doubl
...
+   * Create a new Container
+   *
+   * The created container has to satisfy following requirements:
+   * - The container's file system is based on the provided action image and has a read/write layer on top.
+   * - If the specified image is not available on the system, it is pulled from an image
+   *   repository - for example, Docker Hub.
+   * - The container has a network interface which is able to connect to the internet
+   *   (in particular, to the specified DNS servers). In addition, the invoker must be able
+   *   to connect to the container's interface.
+   * - The IPv4 address of said interface is stored in the created Container instance.
```
markusthoemmes: Are we really constrained to IPv4 here? I also don't think that it's necessarily a requirement to store said address in the Container.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
```diff
@@ -52,7 +52,23 @@ case class ContainerPoolConfig(userMemory: ByteSize, concurrentPeekFactor: Doubl
...
 
-  /** create a new Container */
+  /**
+   * Create a new Container
+   *
+   * The created container has to satisfy following requirements:
+   * - The container's file system is based on the provided action image and has a read/write layer on top.
+   * - If the specified image is not available on the system, it is pulled from an image
+   *   repository - for example, Docker Hub.
+   * - The container has a network interface which is able to connect to the internet
```
markusthoemmes: Do we require containers to be able to connect to the internet?

We could instead state that we don't do any work to connect a container to the internet. It either is connected already or it isn't. DNS servers seem not to be known to this interface.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
```diff
@@ -52,7 +52,23 @@ case class ContainerPoolConfig(userMemory: ByteSize, concurrentPeekFactor: Doubl
  */
 trait ContainerFactory {
 
-  /** create a new Container */
+  /**
+   * Create a new Container
+   *
+   * The created container has to satisfy following requirements:
+   * - The container's file system is based on the provided action image and has a read/write layer on top.
```
markusthoemmes: I don't think this is necessarily a requirement. Blackbox containers wouldn't need that.

sven-lange-last: @markusthoemmes I guess you talk about the read/write layer on top? Without that layer, not a single change can be made to existing files / directories and nothing new can be created.

I guess most managed runtime images require creation of new files. And I think that blackbox image authors also expect that they can create files when initialising / running the action.

markusthoemmes: They might from a feature standpoint, yes. It is not required for the system to work in general. From a spec's PoV i'd say we make this a "SHOULD" or "MAY" in that: 

> The container's file system is based on the provided action image and may have a read/write layer on top.

rabbah: A read only file system is nice from a security standpoint: the container image is immutable.

csantanapr: I like the read only and nonroot by default. 

I believe AWS Lambda the file system to the function is read only, and only have access to small of write access in /tmp which I have to guess is just a mount volume in the container.

rabbah: it would be a nice enhancement although the way we "init" some functions wrinkles this.

markusthoemmes: Let's not get ourselves tangled up in enhancements and feature requests here please. For the sake of this documentation I would state:

> The container's file system is based on the provided action image and may have a read/write layer on top.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
sven-lange-last: DNS server settings are part of the the `ContainerArgsConfig`. At least, the container must be able to connect to the DNS servers defined there.

What about this proposal?

_The container needs a network setup - usually, a network interface - such that the invoker is able to connect the action container. The container must be able to perform DNS resolution based on the settings provided via `ContainerArgsConfig`._ If needed by action authors, the container should be able to connect to other systems or even the internet to consume services._

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
sven-lange-last: I support @markusthoemmes' point of view. This is not a feature discussion but a description to clarify how things work today.

Does anybody know whether any of the prominent runtimes NEED a writable layer? If yes, this should be mentioned somehow.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
sven-lange-last: Well, the `Container` trait has built-in connection management based on the stored `ContainerAddress`. The trait also has implementations for `initialize()` and `run()` based on `callContainer()`. You could overwrite all that code - but I'm talking about the default.

I guess we never really tried whether IPv6 would work...

Proposal:

> The IPv4 address of said interface is stored in the created Container instance if you want to use the standard init / run behaviour defined in the Container trait.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
markusthoemmes: Add an "if any" after `ContainerArgsConfig`, then lgtm :)

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
markusthoemmes: Node.js will need it if using zips.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
markusthoemmes: I'm not even positive that it's always an ipv4 address. Let's just simplify to address and go with that?

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
rabbah: i think they all do: node/python for zip files, swift/go for binaries, java/ballerina for jar file.

rabbah: for the feature discussion there is this open issue (which I agree 💯 is orthogonal to the goals of this PR): https://github.com/apache/incubator-openwhisk/issues/890

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
sven-lange-last: Replaced `IPv4 address` with `IP address`.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
sven-lange-last: Updated as discussed.

rabbah:
Thanks Sven!

PR #4263, opened 2019-02-11
Add description for new metrics added

SugandhaAgrawal:
## Description
Add documentation to support addition of new metrics.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [x] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -130,7 +130,11 @@ Following metrics record stats around activation handling within Controller
 Aggregate metrics for inflight activations.
 
 * `openwhisk.histogram.loadbalancer<controllerId>_activationsInflight_count` (histogram) - Records the number of activations being worked upon for a given controller. As a histogram it would give a distribution of inflight activation count within a flush interval.
-* `openwhisk.histogram.loadbalancer<controllerId>_memoryInflight_count` (histogram) - Records the amount of RAM memory in use for in flight activations. This is not actual runtime memory but the memory specified per action limits.
+* `openwhisk.histogram.loadbalancer<controllerId>_memory<actionType>Inflight_count` (histogram) - Records the amount of RAM memory in use for in flight activations. This is not actual runtime memory but the memory specified per action limits. actionType defines whether it is a managed or a blackbox action
+
+Metrics below are for current memory capacity
+
+* `openwhisk.histogram.loadbalance_totalCapacity<invokerType>_count` (histogram) - Current memory capacity for all usable managed and blackbox invokers, total user memory in shard managed by controller. invokerType defines whether it is a managed or a blackbox invoker.
 
 Metrics below are captured within load balancer
 
@@ -144,8 +148,10 @@ Metrics below are captured within load balancer
 
 Metrics below are for invoker state as recorded within load balancer monitoring.
 
-* `openwhisk.counter.loadbalancer_invokerOffline_count` - Records the count of invokers considered offline based on health pings.
-* `openwhisk.counter.loadbalancer_invokerUnhealthy_count` - Records the count of invokers considered unhealthy based on health pings.
+* `openwhisk.counter.loadbalancer_totalHealthyInvoker<actionType>_count` - Records the count of managed invokers considered healthy based on health pings. actionType defines whether it is a managed or a blackbox action
+* `openwhisk.counter.loadbalancer_totalUnresponsiveInvoker<actionType>_count` - Records the count of managed invokers considered unresponsive based on health pings. actionType defines whether it is a managed or a blackbox action
+* `openwhisk.counter.loadbalancer_totalOfflineInvoker<actionType>_count` - Records the count of managed invokers considered offline based on health pings. actionType defines whether it is a managed or a blackbox action
+* `openwhisk.counter.loadbalancer_totalUnhealthyInvoker<actionType>_count` - Records the count of managed invokers considered unhealthy based on health pings. actionType defines whether it is a managed or a blackbox action
 
 #### Invoker metrics
 
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -130,11 +130,11 @@ Following metrics record stats around activation handling within Controller
 Aggregate metrics for inflight activations.
 
 * `openwhisk.histogram.loadbalancer<controllerId>_activationsInflight_count` (histogram) - Records the number of activations being worked upon for a given controller. As a histogram it would give a distribution of inflight activation count within a flush interval.
-* `openwhisk.histogram.loadbalancer<controllerId>_memory<invokerType>Inflight_count` (histogram) - Records the amount of RAM memory in use for in flight activations. This is not actual runtime memory but the memory specified per action limits. `invokerType` defines whether it is a managed or a blackbox action.
+* `openwhisk.histogram.loadbalancer<controllerId>_memory<invokerType>Inflight_count` (histogram) - Records the amount of RAM memory in use for in flight activations. This is not actual runtime memory but the memory specified per action limits. **invokerType** defines whether it is a managed or a blackbox invoker.
 
 Metrics below are for current memory capacity
 
-* `openwhisk.histogram.loadbalancer_totalCapacity<invokerType>_count` (histogram) - Current memory capacity for all usable managed and blackbox invokers, total user memory in shard managed by controller. `invokerType defines whether it is a managed or a blackbox invoker.
+* `openwhisk.histogram.loadbalancer_totalCapacity<invokerType>_count` (histogram) - Current memory capacity for all usable managed and blackbox invokers, total user memory in shard managed by controller. **invokerType** defines whether it is a managed or a blackbox invoker.
 
 Metrics below are captured within load balancer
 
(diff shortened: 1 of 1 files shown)


docs/metrics.md
```diff
@@ -130,7 +130,11 @@ Following metrics record stats around activation handling within Controller
 Aggregate metrics for inflight activations.
 
 * `openwhisk.histogram.loadbalancer<controllerId>_activationsInflight_count` (histogram) - Records the number of activations being worked upon for a given controller. As a histogram it would give a distribution of inflight activation count within a flush interval.
-* `openwhisk.histogram.loadbalancer<controllerId>_memoryInflight_count` (histogram) - Records the amount of RAM memory in use for in flight activations. This is not actual runtime memory but the memory specified per action limits.
+* `openwhisk.histogram.loadbalancer<controllerId>_memory<actionType>Inflight_count` (histogram) - Records the amount of RAM memory in use for in flight activations. This is not actual runtime memory but the memory specified per action limits. actionType defines whether it is a managed or a blackbox action
```
sven-lange-last: Last sentence of paragraph needs to end with full stop.

docs/metrics.md
```diff
@@ -130,7 +130,11 @@ Following metrics record stats around activation handling within Controller
...
+
+Metrics below are for current memory capacity
+
+* `openwhisk.histogram.loadbalance_totalCapacity<invokerType>_count` (histogram) - Current memory capacity for all usable managed and blackbox invokers, total user memory in shard managed by controller. invokerType defines whether it is a managed or a blackbox invoker.
```
sven-lange-last: Typo in metric name: `loadbalance`.

sven-lange-last: I understand that this is just documenting the current code so the question comes too late... why is the metric not per `<controllerId>`?

docs/metrics.md
```diff
@@ -144,8 +148,10 @@ Metrics below are captured within load balancer
 
 Metrics below are for invoker state as recorded within load balancer monitoring.
 
-* `openwhisk.counter.loadbalancer_invokerOffline_count` - Records the count of invokers considered offline based on health pings.
-* `openwhisk.counter.loadbalancer_invokerUnhealthy_count` - Records the count of invokers considered unhealthy based on health pings.
+* `openwhisk.counter.loadbalancer_totalHealthyInvoker<actionType>_count` - Records the count of managed invokers considered healthy based on health pings. actionType defines whether it is a managed or a blackbox action
+* `openwhisk.counter.loadbalancer_totalUnresponsiveInvoker<actionType>_count` - Records the count of managed invokers considered unresponsive based on health pings. actionType defines whether it is a managed or a blackbox action
```
sven-lange-last: If I'm not mistaken, invoker state _unresponsive_ does not base on (missing) health pings from invokers but from (missing) activation responses from invoker.

docs/metrics.md
```diff
@@ -144,8 +148,10 @@ Metrics below are captured within load balancer
 
 Metrics below are for invoker state as recorded within load balancer monitoring.
 
-* `openwhisk.counter.loadbalancer_invokerOffline_count` - Records the count of invokers considered offline based on health pings.
-* `openwhisk.counter.loadbalancer_invokerUnhealthy_count` - Records the count of invokers considered unhealthy based on health pings.
+* `openwhisk.counter.loadbalancer_totalHealthyInvoker<actionType>_count` - Records the count of managed invokers considered healthy based on health pings. actionType defines whether it is a managed or a blackbox action
```
sven-lange-last: These two comments apply to this line and the following 3 as well.

* I understand that this is just documenting the current code so the question comes too late... why is the metric not per `<controllerId>`?
* I think the metric is not per `<actionType>` but per `<invokerType>`.

rabbah: also missing `.` at end of sentence here and below.
i'd use back ticks after "pings." to distinguish the value `<actionType>`  of if it's `<invokerType>`

docs/metrics.md
SugandhaAgrawal: @sven-lange-last - Thanks for the comment. The current grafana display, does work on the basis of different controller. The query in grafana is so designed that we take in result from all the controllers(2 or more) and display them as a total. @chetanmeh I would be interested to know that will this metrics in its current form would also be useful for your current usage of metrics.

docs/metrics.md
SugandhaAgrawal: I will enhance the statements for all the states to make it clear.

docs/metrics.md
SugandhaAgrawal: @sven-lange-last - I would move all the metric to be defined per invokerType. I was just trying to distinguish the names per metrics - probably not such a good idea 😉. Metric is again not defined per `controllerId`, as on the grafana query all the controller results are aggregated.

sven-lange-last:
I just restarted failed Travis job. Since these are only `.md` updates, tests shouldn't be affected anyway.

sven-lange-last:
@chetanmeh anything left from your side? If I don't hear back from you, I'm going to merge.

chetanmeh:
Missed it ... would review by tomorrow

docs/metrics.md
```diff
@@ -130,7 +130,11 @@ Following metrics record stats around activation handling within Controller
 Aggregate metrics for inflight activations.
 
 * `openwhisk.histogram.loadbalancer<controllerId>_activationsInflight_count` (histogram) - Records the number of activations being worked upon for a given controller. As a histogram it would give a distribution of inflight activation count within a flush interval.
-* `openwhisk.histogram.loadbalancer<controllerId>_memoryInflight_count` (histogram) - Records the amount of RAM memory in use for in flight activations. This is not actual runtime memory but the memory specified per action limits.
+* `openwhisk.histogram.loadbalancer<controllerId>_memory<invokerType>Inflight_count` (histogram) - Records the amount of RAM memory in use for in flight activations. This is not actual runtime memory but the memory specified per action limits. `invokerType` defines whether it is a managed or a blackbox action.
```
chetanmeh: I believe `invokerType` would have a pre defined set of possible values. Can you enumerate those in the first reference.

docs/metrics.md
```diff
@@ -144,8 +148,10 @@ Metrics below are captured within load balancer
 
 Metrics below are for invoker state as recorded within load balancer monitoring.
 
-* `openwhisk.counter.loadbalancer_invokerOffline_count` - Records the count of invokers considered offline based on health pings.
-* `openwhisk.counter.loadbalancer_invokerUnhealthy_count` - Records the count of invokers considered unhealthy based on health pings.
+* `openwhisk.counter.loadbalancer_totalHealthyInvoker<invokerType>_count` - Records the count of managed invokers considered healthy based on health pings. `invokerType` defines whether it is a managed or a blackbox action.
```
chetanmeh: Are these values supposed to be recorded as counter or [guages][1]

> Counters: Track the number of times certain event happened. It can only be increased. Useful for counting errors, cache misses, etc.
> Gagues: Track a single value that can be increased, decreased or explicitly set. The common use case for gauges are slowly moving values like disk usage, number of loaded classes or configuration settings.

Per current impl they seem to be recorded as Histogram

https://github.com/apache/incubator-openwhisk/blob/28ae83cfb1bd07f55b12dc91b16f0e113dbfee2f/core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/ShardingContainerPoolBalancer.scala#L186-L209

[1]: https://kamon.io/documentation/1.x/core/basics/metrics/#gauges

SugandhaAgrawal: @chetanmeh Good catch. They are recorded Histogram indeed. Will do the change.

docs/metrics.md
SugandhaAgrawal: invokerType has only `managed` and `blackbox` as possible values. I have made reference to these where ever invokerType is a part of the metrics.

docs/metrics.md
chetanmeh: @SugandhaAgrawal What I meant was that the kind of metrics we record here in code is supposed to be `gauge` and histogram may not be the right way to capture it. So we may need to change the implementation as well

SugandhaAgrawal: @chetanmeh I checked in the code if we have ever used **gauges** before. I saw 3 examples where indeed this could have been gauges but they are implemented as histogram instead. 1) DB Batch size for CouchDB, 2) Client retries in Akka client and 3) Kafka queue size. I would propose to keep Histogram for the above mentioned metrics. Gauge implementation have to be done in code, and once this is done, I propose changing all the incorrect metrics.

chetanmeh: Makes sense. Can you create an issue to track this change? As with current histogram based metric its hard to use these value for any monitoring purpose

SugandhaAgrawal: Sure thing. @chetanmeh can you please merge this one in the meanwhile?

chetanmeh:
#4311 opened to track the switch to Gauge issue

PR #4284, opened 2019-02-13
Add an annotations to inject the API key into the action context.

### LINKED ISSUE #4226: change the default action context to omit api key
Actions receives the API key in the environment even if it is not necessary. This should not be the default behavior. With this issue I'm proposing that we flip the default and provide an annotation on the action to enable the key forwarding to preserve existing behavior.

rabbah:
See https://github.com/apache/incubator-openwhisk/issues/4226. Note that this is a breaking change - actions that are deployed need to be updated to add this annotation.

Closes #4226.

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#4226).

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [x] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
@@ -59,6 +59,33 @@ object WhiskActionsApi {
    * This is the default timeout on a POST request.
    */
   protected[core] val maxWaitForBlockingActivation = 60 seconds
+
+  /**
+   * Amends annotations on an action create/update with system defined values.
+   * This method currently adds the following annotations:
+   * 1. [[WhiskAction.provideApiKeyAnnotationName]] with the value false iff the annotation is not already defined in the action annotations
+   * 2. An [[execAnnotation]] consistent with the action kind; this annotation is always added and overrides a pre-existing value
+   */
+  protected[core] def amendAnnotations(annotations: Parameters, exec: Exec): Parameters = {
+    val newAnnotations = annotations
+      .get(WhiskAction.provideApiKeyAnnotationName)
+      .map(_ => annotations)
+      .getOrElse {
+        annotations ++ Parameters(WhiskAction.provideApiKeyAnnotationName, JsBoolean(false))
+      }
+    newAnnotations ++ execAnnotation(exec)
+  }
+
+  /**
+   * Constructs an "exec" annotation. This is redundant with the exec kind
+   * information available in WhiskAction but necessary for some clients which
+   * fetch action lists but cannot determine action kinds without fetching them.
+   * An alternative is to include the exec in the action list "view" but this
+   * will require an API change. So using an annotation instead.
+   */
+  private def execAnnotation(exec: Exec): Parameters = {
+    Parameters(WhiskAction.execFieldName, exec.kind)
+  }
 }
 
 /** A trait implementing the actions API. */
@@ -416,7 +443,7 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with 
       limits,
       content.version getOrElse SemVer(),
       content.publish getOrElse false,
-      (content.annotations getOrElse Parameters()) ++ execAnnotation(exec))
+      WhiskActionsApi.amendAnnotations(content.annotations getOrElse Parameters(), exec))
   }
 
   /** For a sequence action, gather referenced entities and authorize access. */
@@ -531,7 +558,7 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with 
       limits,
       content.version getOrElse action.version.upPatch,
       content.publish getOrElse action.publish,
-      (content.annotations getOrElse action.annotations) ++ execAnnotation(exec))
+      WhiskActionsApi.amendAnnotations(content.annotations getOrElse action.annotations, exec))
       .revision[WhiskAction](action.docinfo.rev)
   }
 
@@ -683,17 +710,6 @@ trait WhiskActionsApi extends WhiskCollectionAPI with PostActionActivation with 
       }
       totalActionCount
     }
-  }
-
-  /**
-   * Constructs an "exec" annotation. This is redundant with the exec kind
-   * information available in WhiskAction but necessary for some clients which
-   * fetch action lists but cannot determine action kinds without fetching them.
-   * An alternative is to include the exec in the action list "view" but this
-   * will require an API change. So using an annotation instead.
-   */
-  private def execAnnotation(exec: Exec): Parameters = {
-    Parameters(WhiskAction.execFieldName, exec.kind)
   }
 
   /** Max atomic action count allowed for sequences. */
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
@@ -66,13 +66,18 @@ object WhiskActionsApi {
    * 1. [[WhiskAction.provideApiKeyAnnotationName]] with the value false iff the annotation is not already defined in the action annotations
    * 2. An [[execAnnotation]] consistent with the action kind; this annotation is always added and overrides a pre-existing value
    */
-  protected[core] def amendAnnotations(annotations: Parameters, exec: Exec): Parameters = {
-    val newAnnotations = annotations
-      .get(WhiskAction.provideApiKeyAnnotationName)
-      .map(_ => annotations)
-      .getOrElse {
-        annotations ++ Parameters(WhiskAction.provideApiKeyAnnotationName, JsBoolean(false))
-      }
+  protected[core] def amendAnnotations(annotations: Parameters, exec: Exec, create: Boolean = true): Parameters = {
+    val newAnnotations = if (create) {
+      // these annotations are only added on newly created actions
+      // since they can break existing actions created before the
+      // annotation was created
+      annotations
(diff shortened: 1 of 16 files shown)


rabbah:
I know there are actions we're deploying that will need to be updated, waiting for travis to tell me which ones.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -560,7 +560,9 @@ class ContainerProxy(
         }
         val parameters = job.msg.content getOrElse JsObject.empty
 
-        val authEnvironment = job.msg.user.authkey.toEnvironment
+        val authEnvironment = if (job.action.annotations.isTruthy(WhiskAction.provideApiKeyAnnotationName)) {
+          job.msg.user.authkey.toEnvironment
+        } else JsObject.empty
```
markusthoemmes: To avoid a breaking change: Can we assume the absence of the annotation to mean "true" and create all new actions with the annotation set to false?

rabbah: possibly brilliant - yes if the annotation is added server side... will munch on the suggestion - thanks!!

rabbah: implemented this proposal - thanks @markusthoemmes. Can you review the changes now 🙏 ?

rabbah:
I will need to update a few API tests which check the API response and now are missing the new annotation.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
mdeuser: just want to confirm my understanding as well as the non-breaking behavior in an update scenario.  from the dev list discussion
> actions which do not have the annotation will still get the api key injected. The annotation is added by the controller when an action is created or updated unless already present. The default value for the annotation is "false", meaning no key is injected to the action context.
here's the scenario
1. an existing action without annotation is updated, the request does not specify the annotation
2. backend sees action without annotation and updates action with the annotation
    - will this value be `true` or `false` ?
3. some time later, another request to update the same action is made.  as before, the request does not specify the annotation.
    - will the updated action have the annotation set to `true` or `false`?

rabbah: As implemented here, the value for the annotation when added by the system is always `false`. The first update will add the annotation. So in the scenario above, the second update would already have the annotation and the system leaves it in place.

rabbah: The tests here are for what I describe above:

https://github.com/apache/incubator-openwhisk/pull/4284/files#diff-497ffa2cac9fe692a1df3b53e9793a41R1677

https://github.com/apache/incubator-openwhisk/pull/4284/files#diff-497ffa2cac9fe692a1df3b53e9793a41R1686

rabbah: Thanks @mdeuser for taking a look at this proposal.

mdeuser: for the most backwards compatibility, i'm thinking trying to preserve the behavior for existing actions unless explicitly changed by the action owner.

as such, updates to existing actions would have the annotation set to `true` (i.e. retain __ow_api_key behavior) if the annotation did not exist and the annotation is not specified on the request.  if the annotation exists and the annotation was not specified on the update request, retain the annotation existing value.

newly created actions would get the default annotation value of `false` if not specified on the request.

rabbah: > as such, updates to existing actions would have the annotation set to `true`

what if the update request is updating unrelated annotations -- eg to make an action a web action or modifying descriptions?

mdeuser: let's see.  that's an update to an existing action.  so, if the action did not have the api_key annotation, then add it and set it to `true`.  if the action did have the annotation, leave it as-is.

rabbah: take a look at the last commit @mdeuser - implemented your suggestions.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -560,7 +560,13 @@ class ContainerProxy(
         }
         val parameters = job.msg.content getOrElse JsObject.empty
 
-        val authEnvironment = job.msg.user.authkey.toEnvironment
+        // if the action requests the api key to be injected into the action context, add it here;
+        // treat a missing annotation as requesting the api key for backward compatibility
+        val authEnvironment = {
+          if (job.action.annotations.isTruthy(WhiskAction.provideApiKeyAnnotationName, valueForNonExistent = true)) {
```
dubee: 👍

docs/actions.md
```diff
@@ -608,9 +608,9 @@ or set an internal alarm when the action is about to use up its allotted time bu
...
 Node.js, Python, Swift, Java and Docker actions when using the OpenWhisk Docker skeleton.
 
-* `__OW_API_HOST` the API host for the OpenWhisk deployment running this action
-* `__OW_API_KEY` the API key for the subject invoking the action, this key may be a restricted API key
-* `__OW_NAMESPACE` the namespace for the _activation_ (this may not be the same as the namespace for the action)
-* `__OW_ACTION_NAME` the fully qualified name of the running action
-* `__OW_ACTIVATION_ID` the activation id for this running action instance
-* `__OW_DEADLINE` the approximate time when this action will have consumed its entire duration quota (measured in epoch milliseconds)
+* `__OW_API_HOST` the API host for the OpenWhisk deployment running this action.
+* `__OW_API_KEY` the API key for the subject invoking the action, this key may be a restricted API key. This property is absent unless explicitly [requested](./annotations.md#annotations-for-all-actions).
```
dubee: May want to update the docs explaining the reverse compatibility feature you added.

core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
```diff
@@ -59,6 +59,38 @@ object WhiskActionsApi {
...
+      // annotation was created
+      annotations
+        .get(WhiskAction.provideApiKeyAnnotationName)
+        .map(_ => annotations)
+        .getOrElse {
+          annotations ++ Parameters(WhiskAction.provideApiKeyAnnotationName, JsBoolean(false))
+        }
+    } else annotations
+    newAnnotations ++ execAnnotation(exec)
+  }
```
mdeuser: LGTM

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
mdeuser: LGTM.  thanks!

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
mhenke1: 👍 I second to have backwards compatibility here.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -560,7 +560,13 @@ class ContainerProxy(
...
+            job.msg.user.authkey.toEnvironment
+          } else JsObject.empty
```
mhenke1:
```
} else job.msg.user.authkey.toEnvironmentSensitive
```

I do suggest to move the decision what to hide as sensitive information to each `GenericAuthKey` implementation 
by adding a new method `authkey.toEnvironmentSensitive`. This method should return `JsObject.empty`
for the BasicAuthenticationAuthKey. Other `GenericAuthKey` implementations could decide which of the information stored should be made available on request only. 

If we stay backwards compatible , I would be willing to implement this change implement  in a following PR:.

rabbah: 👍

rabbah: Can you send a comment to the dev list with the proposal. It is reasonable to me. Do you have an example you can attach?

docs/actions.md
rabbah: done in last commit.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
mhenke1: I will address it once this PR is merged. No need to delay.

rabbah: Thanks!

PR #4289, opened 2019-02-14
Make ContainerProxyTests more stable.

cbickel:
As the active ack is sent asynchronously in the invoker, it could happen, that the order of them is wrong in the ContainerProxyTest. This has been proven with the following debug information: https://github.com/apache/incubator-openwhisk/pull/4278

This PR makes these tests more stable by not relying on the order of active acks.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/tests/src/test/scala/org/apache/openwhisk/core/containerpool/test/ContainerProxyTests.scala
+++ b/tests/src/test/scala/org/apache/openwhisk/core/containerpool/test/ContainerProxyTests.scala
@@ -18,6 +18,7 @@
 package org.apache.openwhisk.core.containerpool.test
 
 import java.time.Instant
+
 import akka.actor.FSM.{CurrentState, SubscribeTransitionCallBack, Transition}
 import akka.actor.{ActorRef, ActorSystem, FSM}
 import akka.stream.scaladsl.Source
@@ -25,6 +26,7 @@ import akka.testkit.{ImplicitSender, TestKit}
 import akka.util.ByteString
 import common.{LoggedFunction, StreamLogging, SynchronizedLoggedFunction, WhiskProperties}
 import java.util.concurrent.atomic.AtomicInteger
+
 import org.junit.runner.RunWith
 import org.scalatest.junit.JUnitRunner
 import org.scalatest.{BeforeAndAfterAll, FlatSpecLike, Matchers}
@@ -39,6 +41,7 @@ import org.apache.openwhisk.core.entity._
 import org.apache.openwhisk.core.entity.size._
 import org.apache.openwhisk.http.Messages
 import org.apache.openwhisk.core.database.UserContext
+
 import scala.concurrent.Await
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future, Promise}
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/org/apache/openwhisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -304,12 +307,15 @@ class ContainerProxyTests
...
       store.calls should have size 2
 
-      val initRunActivation = acker.calls(0)._2
+      // As the active acks are sent asynchronously, it is possible, that the activation with the init time is not the
+      // first one in the buffer.
+      val initializedActivations =
+        acker.calls.filter(_._2.annotations.get(WhiskActivation.initTimeAnnotation).isDefined)
+      initializedActivations should have size 1
+
+      val initRunActivation = initializedActivations.head._2
```
markusthoemmes: As you already know that the size is 2 and you need both calls, could this be:

```scala
val (initRunActivation, runOnlyActivation) = {
  // false is sorted before true
  val sorted = acker.calls.sortBy(_._2.annotations.get(WhiskActivation.initTimeAnnotation).isEmpty)
  sorted(0)._2, sorted(1)._2
}
```

cbickel: That's a good point. 👍

tests/src/test/scala/org/apache/openwhisk/core/containerpool/test/ContainerProxyTests.scala
```diff
@@ -365,18 +372,18 @@ class ContainerProxyTests
...
-        ._2
-        .annotations
+
+      // As the active acks are sent asynchronously, it is possible, that the activation with the init time is not the
+      // first one in the buffer.
+      val initializedActivations =
+        acker.calls.filter(_._2.annotations.get(WhiskActivation.initTimeAnnotation).isDefined)
+      initializedActivations should have size 1
+
+      initializedActivations.head._2.annotations
```
markusthoemmes: For the instances where you only need one of the calls:

```scala
val initRunActivation = acker.calls.find(_._2.annotations.get(WhiskActivation.initTimeAnnotation).isDefined).getOrElse(fail("..."))
```

?

cbickel: This would keep consistency to the other tests that I've adapted, but then we need another check to verify that only one of the activations contains the init annotation.
So personally I'd prefer to use `filter`, checking the size of the result and using head.

chetanmeh:
Saw another set of failure related to ContainerProxy https://scans.gradle.com/s/36yxkcjvgad6y/tests/failed

cbickel:
@chetanmeh Thanks for the hint. But on the first view it looks to me, like there is a completely different root cause, compared to the problem that has been fixed here.

PR #4299, opened 2019-02-20
Avoid that actions do not stop after action timeout when logging heavily

steffenrost:
## Description
Avoid that actions do not stop after action timeout when logging heavily by not 
waiting for a sentinel but just collecting the log written up to this point in time.
Please find a detailed problem description in https://github.com/apache/incubator-openwhisk/issues/4298

## Related issue and scope

Related issue is https://github.com/apache/incubator-openwhisk/issues/4298

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


steffenrost:
@sven-lange-last could you please review?

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
```diff
@@ -151,7 +155,10 @@ class DockerToActivationFileLogStore(system: ActorSystem, destinationDirectory:
     logs.runWith(combined)._1.flatMap { seq =>
       val possibleErrors = Set(Messages.logFailure, Messages.truncateLogs(action.limits.logs.asMegaBytes))
       val errored = seq.lastOption.exists(last => possibleErrors.exists(last.contains))
-      val logs = ActivationLogs(seq.toVector)
+      val logs = ActivationLogs(
+        if (activation.response.isContainerError)
+          seq.toVector :+ LogLine(Instant.now.toString, "stderr", Messages.logFailure).toFormattedString
+        else seq.toVector)
```
markusthoemmes: This will write the error log only in this path of the code. You might want to concat it to the stream above so all possible branches get this message, like:

```scala
val logsWithPossibleError = if(activation.response.isContainerError) {
  logs.concat(Source.single(ByteString(LogLine(...))))
} else logs
```

And then work off of that. Note that the `errored` check should be short circuited as well in case there was a ContainerError it should always be true and doesn't need to look at the last message "manually".

tests/dat/actions/loggingTimeout.js
```diff
@@ -0,0 +1,100 @@
...
+// following code is only active when being called with nodejs from command-line
+//console.log("process.env: " + JSON.stringify(process.env));
+if (! process.env.__OW_API_HOST) {
+  var params = '{}'
+  if (process.argv.length >= 3) {
+    params = process.argv[2];
+  }
+  info = main(JSON.parse(params));
+  console.log(JSON.stringify(info));
+}
```
markusthoemmes: This function looks extremely complex. Can we tune it down to only do the minimally necessary work here?

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
```diff
@@ -123,7 +123,11 @@ class DockerToActivationFileLogStore(system: ActorSystem, destinationDirectory:
                            container: Container,
                            action: ExecutableWhiskAction): Future[ActivationLogs] = {
 
-    val logs = container.logs(action.limits.logs.asMegaBytes, action.exec.sentinelledLogs)(transid)
+    // wait for a sentinel only if no container (developer) error occurred to avoid
+    // that log collection continues if the action code still logs after timeout
+    val sentinel = action.exec.sentinelledLogs && !activation.response.isContainerError
```
sven-lange-last: Two questions:

1. The `ActivationResponse` is a container error if the response's status code is an `ActivationResponse.DeveloperError`. This status code is set if the action init / run times out. Your change is desired in case of a timeout.
   In addition, a developer error is created for other failures as well. Did you check all code locations where a developer error is created to make sure you do not introduce undesired behaviour for other failures?

2. You are preventing on-going log file reading for developer errors. Did you check whether it could be helpful to change log file reading behaviour for other errors as well?

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
sven-lange-last: I agree - @markusthoemmes has a good finding.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationLogStore.scala
```diff
@@ -76,14 +77,21 @@ class DockerToActivationLogStore(system: ActorSystem) extends LogStore {
...
                            action: ExecutableWhiskAction): Future[ActivationLogs] = {
 
+    // wait for a sentinel only if no container (developer) error occurred to avoid
+    // that log collection continues if the action code still logs after timeout
+    val sentinel = action.exec.sentinelledLogs && !activation.response.isContainerError
+
     container
-      .logs(action.limits.logs.asMegaBytes, action.exec.sentinelledLogs)(transid)
+      .logs(action.limits.logs.asMegaBytes, sentinel)(transid)
       .via(DockerToActivationLogStore.toFormattedString)
```
sven-lange-last: If we change the `DockerToActivationFileLogStore` such that a developer error injects a message that leads to a log collection error, then this log store should have the same behaviour. In that case, we have to inject / append `Messages.logFailure` at this location - or at least before we check for error messages.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +463,30 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
...
+    (wp, assetHelper) =>
+      val name = "NodeJsTestLoggingActionCausingTimeout-" + System.currentTimeMillis()
+      print(s"\n create action ${name} using api host: ${wskprops.apihost}..")
+      assetHelper.withCleaner(wsk.action, name, confirmDelete = true) { (action, _) =>
+        action.create(
+          name,
+          Some(TestUtils.getTestActionFilename("loggingTimeout.js")),
+          timeout = Some(allowedActionDuration))
+      }
+      val run = wsk.action.invoke(name, Map("durationMillis" -> 120000.toJson, "logDeltaMillis" -> 100.toJson))
```
sven-lange-last: The value of `durationMillis` should be based on `allowedActionDuration`. You can find samples in the `ActionLimitTests` class using this approach.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +463,30 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
...
+      withActivation(wsk.activation, run) { result =>
+        withClue("Activation result not as expected:") {
+          result.response.status shouldBe ActivationResponse.messageForCode(ActivationResponse.DeveloperError)
+          result.response.result.get.toString should include("""exceeded its time limits""")
+          result.logs.get.mkString(" ") should include(Messages.logFailure)
```
sven-lange-last: Shouldn't it be sufficient to check the last log line?

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +463,30 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
...
+          val startLogMillis = Instant.parse(result.logs.get.head.split(' ')(0)).toEpochMilli()
+          val endLogMillis = Instant.parse(result.logs.get.last.split(' ')(0)).toEpochMilli()
+          (endLogMillis - startLogMillis).toInt should be < (allowedActionDuration + 60.seconds).toMillis.toInt
```
sven-lange-last: I wonder whether we could use a relative value here instead of hard coding 60 seconds.

ruediger-maass:
@markusthoemmes @sven-lange-last - thanks for your comments! We incoprated your comments into our code changes, test action, and test code. In particular, we found that DeveloperError is not only used in action timeout cases. Therefore we added the function `isConterTimeoutError` in `ActivationResponse` in order to apply our code changes only in timeout cases.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
```diff
@@ -123,7 +123,16 @@ class DockerToActivationFileLogStore(system: ActorSystem, destinationDirectory:
                            container: Container,
                            action: ExecutableWhiskAction): Future[ActivationLogs] = {
 
-    val logs = container.logs(action.limits.logs.asMegaBytes, action.exec.sentinelledLogs)(transid)
+    // wait for a sentinel only if no container (developer) error occurred to avoid
+    // that log collection continues if the action code still logs after timeout
+    val isContainerTimeoutError = activation.response.isContainerTimeoutError(action.limits.timeout.duration)
```
dubee: Can any of these vals reference the same vals from `DockerToActivationLogStore.scala`?

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +463,34 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
...
+          kind = Some("nodejs:8"))
+      }
+      val durationMillis = allowedActionDuration + 3.minutes
+      val checkDurationMillis = allowedActionDuration + 1.minutes
+      val run =
+        wsk.action.invoke(name, Map("durationMillis" -> durationMillis.toMillis.toJson, "delayMillis" -> 100.toJson))
+      withActivation(wsk.activation, run) { result =>
+        withClue("Activation result not as expected:") {
+          result.response.status shouldBe ActivationResponse.messageForCode(ActivationResponse.DeveloperError)
+          result.response.result.get.toString should include("""exceeded its time limits""")
```
dubee: Should not need the triple quotes around the string here.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +463,34 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
       _.response.result.get.fields("error") shouldBe Messages.memoryExhausted.toJson
     }
   }
+
+  /**
+   * Test that a heavy logging action is interrupted within its timeout limits.
+   */
+  it should "interrupt the heavy logging action within its time limits" in withAssetCleaner(wskprops) {
+    (wp, assetHelper) =>
+      val name = "NodeJsTestLoggingActionCausingTimeout-" + System.currentTimeMillis()
```
dubee: Alternatively can do string concatenation like the next line. Ex:
`val name = s"NodeJsTestLoggingActionCausingTimeout-${System.currentTimeMillis()}"`

common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala
```diff
@@ -45,6 +48,15 @@ protected[core] case class ActivationResponse private (val statusCode: Int, val
   def isSuccess = statusCode == ActivationResponse.Success
   def isApplicationError = statusCode == ActivationResponse.ApplicationError
   def isContainerError = statusCode == ActivationResponse.DeveloperError
+  def isContainerTimeoutError(timeout: FiniteDuration) = {
+    statusCode == ActivationResponse.DeveloperError &&
+    (result.get match {
```
markusthoemmes: This will throw an error if the Option is `None`

common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala
```diff
@@ -45,6 +48,15 @@ protected[core] case class ActivationResponse private (val statusCode: Int, val
...
+      case JsObject(fields) =>
+        fields.contains(ActivationResponse.ERROR_FIELD) && fields(ActivationResponse.ERROR_FIELD)
+          .toString() == '"' + Messages.timedoutActivation(timeout, false) + '"'
```
markusthoemmes: Please never use `toString` on JSON. It get's very unwieldy and bypasses typechecks (since your just breaking everything down to a string).

As these are all case-classes, you should be able to greatly simplify the whole check to

```scala
def isContainerTimeoutError(timeout: FiniteDuration) = this == ActivationResponse.developerError(Messages.timedoutActivation(timeout, false))
```

Does that work?

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
```diff
@@ -123,7 +123,16 @@ class DockerToActivationFileLogStore(system: ActorSystem, destinationDirectory:
...
+    val sentinel = action.exec.sentinelledLogs && !isContainerTimeoutError
+
+    val logs = container.logs(action.limits.logs.asMegaBytes, sentinel)(transid)
+    val logsWithPossibleError = if (isContainerTimeoutError) {
+      logs.concat(
+        Source.single(ByteString(LogLine(Instant.now.toString, "stderr", Messages.logFailure).toJson.compactPrint)))
```
markusthoemmes: Does this need another newline at the end?

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +463,34 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
...
+      print(s"\n create action ${name} using api host: ${wskprops.apihost}..")
+      assetHelper.withCleaner(wsk.action, name, confirmDelete = true) { (action, _) =>
+        action.create(
+          name,
+          Some(TestUtils.getTestActionFilename("loggingTimeout.js")),
+          timeout = Some(allowedActionDuration),
+          kind = Some("nodejs:8"))
+      }
```
sven-lange-last: Travis output indicates that the test action could not be created:

```
org.apache.openwhisk.core.limits.ActionLimitsTests > Action limits should interrupt the heavy logging action within its time limits STANDARD_OUT
     create action NodeJsTestLoggingActionCausingTimeout-1550822594899 using api host: 172.17.0.1..Exception occurred during test execution: org.scalatest.exceptions.TestFailedException: 400 was not equal to 200
```

rabbah: Does the kind have to nodejs 8? Is this because the default kind is 6 as of now?

common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala
markusthoemmes: I'd even say: take that piece of code to the caller side if it can be simplified like proposed.

steffenrost:
@markusthoemmes @sven-lange-last @dubee @rabbah  - thanks for all your comments! After discussion with @sven-lange-last we decided to add a timeout flag as annotation to the activation  instead of parsing the `ActivationResponse`. We still need to check why the test action couldn't be created.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationLogStore.scala
```diff
@@ -70,19 +72,37 @@ class DockerToActivationLogStore(system: ActorSystem) extends LogStore {
   override def fetchLogs(activation: WhiskActivation, context: UserContext): Future[ActivationLogs] =
     Future.successful(activation.logs)
 
+  def getLogs(transid: TransactionId,
```
markusthoemmes: The name of this is awfully close to `fetchLogs` and `collectLogs`. Can we name this something like `logStream`? Also please make it protected, it's not supposed to be called from the outside. A piece of ScalaDoc and defining the return value would be the icing on the cake 🙂.

Defining the return value is a good best-practice for non-trivial methods. They make reviewing so much easier since we don't have an IDE in the Github review window just yet.

tests/dat/actions/loggingTimeout.js
```diff
@@ -0,0 +1,39 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more contributor
```
markusthoemmes: Is this file still needed given the switch to python?

steffenrost:
@markusthoemmes @sven-lange-last thanks again for your comments! I hope we now addressed all comments including the latest from @markusthoemmes.
We had some trouble with the logging timeout action. After the first nodejs version worked we streamlined it for nodejs 8 but got error 400 (bad request) for the `create action` request in the system tests. We assumed that the travis openwhisk environment doesn't support nodejs 8 and switched to python but again action creation failed with 400. Why this happens it not yet clear especially as this error occurs only in the travis openwhisk deployment.
However we reverted back to a simplified version of the original nodejs action. The PR now only contains this  nodejs version.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +463,35 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
...
+          name,
+          Some(TestUtils.getTestActionFilename("loggingTimeout.js")),
+          timeout = Some(allowedActionDuration))
+      }
+      val durationMillis = allowedActionDuration + 3.minutes
+      val checkDurationMillis = allowedActionDuration + 1.minutes
+      //val run =
+      //  wsk.action.invoke(name, Map("durationMillis" -> durationMillis.toMillis.toJson, "delayMillis" -> 100.toJson))
+      val run =
+        wsk.action.invoke(name, Map("durationMillis" -> durationMillis.toMillis.toJson, "logDeltaMillis" -> 100.toJson))
```
sven-lange-last: The currently active test code passes parameter `logDeltaMillis` while the test action `loggingTimeout.js` requires a parameter named `delayMillis`.

The commented invocation should be removed.

tests/dat/actions/loggingTimeout.js
```diff
@@ -0,0 +1,35 @@
...
+     now = new Date();
+     if (now - lastLog > delayMillis) {
+        lastLog = now;
+        logLines++;
+        console.log('[' + logLines + '] The quick brown fox jumps over the lazy dog.');
+     }
+     if (now - startMillis >= durationMillis) {
+        break;
+     }
+  }
```
sven-lange-last: This is a busy loop. On a test system with few CPUs, this can be too heavy weight. Can you please use timeouts or intervals instead?

tests/dat/actions/loggingTimeout.js
```diff
@@ -0,0 +1,35 @@
...
+// This action prints log lines to stdout for the duration of durationMillis.
+// The output is throttled by a delay of delayMillis between two log lines in
+// order to keep the log small and to stay within the log size limit.
+
+function doLog(startMillis, durationMillis, delayMillis) {
+  var lastLog = 0;
+  var logLines = 0;
+  while (true) {
+     now = new Date();
+     if (now - lastLog > delayMillis) {
```
sven-lange-last: What happens if no parameter named `delayMillis` is passed? I guess the value is `undefined`?

tests/dat/actions/loggingTimeout.js
```diff
@@ -0,0 +1,35 @@
...
+        break;
+     }
+  }
+  return logLines;
+}
+
+
+function main(args) {
+   const durationMillis = args.durationMillis;
+   const delayMillis = args.delayMillis;
```
sven-lange-last: Please use reasonable default values for `durationMillis` and `delayMillis` if these parameters were not passed. Or fail the action if these values are `undefined`.

steffenrost:
@sven-lange-last agree with your point on the busy loop, we need to go back to the python or nodejs 8 action. @markusthoemmes do you know why the upload of the logs to  https://openwhisk.box.com/v/travis-logs currently does not work, in the travis build we see `Uploading to Box... Posting result <Response [500]> ... Internal Server Error` and the latest logs available under that box folder are from Feb 3, 2019. Having the logs available would help to understand the action creation problem.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +463,33 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
...
+          timeout = Some(allowedActionDuration))
+      }
+      val durationMillis = allowedActionDuration + 3.minutes
+      val checkDurationMillis = allowedActionDuration + 1.minutes
+      val run =
+        wsk.action.invoke(name, Map("durationMillis" -> durationMillis.toMillis.toJson, "delayMillis" -> 100.toJson))
+      withActivation(wsk.activation, run) { result =>
+        withClue("Activation result not as expected:") {
+          result.response.status shouldBe ActivationResponse.messageForCode(ActivationResponse.DeveloperError)
+          result.response.result.get.toString should include("exceeded its time limits")
```
markusthoemmes: Please tighten this up using the defined constants we have. The test above uses

```scala
result.response.result.get.fields("error") shouldBe {
  Messages.timedoutActivation(allowedActionDuration, init = false).toJson
}
```

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +463,33 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
...
+          result.logs.get.last should include(Messages.logFailure)
+          val startLogMillis = Instant.parse(result.logs.get.head.split(' ')(0)).toEpochMilli()
+          val endLogMillis = Instant.parse(result.logs.get.last.split(' ')(0)).toEpochMilli()
+          (endLogMillis - startLogMillis).toInt should be < (checkDurationMillis).toMillis.toInt
```
markusthoemmes: There's quite a bit of deduplication that can be done here. `result.logs.get` is used a couple of times.

What about:

```scala
val logs = result.logs.get
logs.last should include(Messages.logFailure)

val parseLogTime = line => Instant.parse(line.split(' ').head)
val startTime = parseLogTime(logs.head)
val endTime = parseLogTime(logs.last)

Duration.between(startTime, endTime) should be < (checkDuration)
```

You'll need to import [TimeHelpers.scala](https://github.com/apache/incubator-openwhisk/blob/be1e3a19c02956c9be85023a0bb0ff399c21444d/common/scala/src/main/scala/org/apache/openwhisk/utils/TimeHelpers.scala) to implicitly transform the java.time.Duration from Duration.between to a Scala duration.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +463,33 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
...
+      val name = s"PythonTestLoggingActionCausingTimeout-${System.currentTimeMillis()}"
+      print(s"\n create action ${name} using api host: ${wskprops.apihost}..")
+      assetHelper.withCleaner(wsk.action, name, confirmDelete = true) { (action, _) =>
+        action.create(
+          name,
+          Some(TestUtils.getTestActionFilename("loggingTimeout.py")),
+          timeout = Some(allowedActionDuration))
+      }
+      val durationMillis = allowedActionDuration + 3.minutes
+      val checkDurationMillis = allowedActionDuration + 1.minutes
```
markusthoemmes: Please omit the `Millis` here. They have an arbitrary unit.

rabbah:
I checked the box folder and deleted all the old logs to make room. I suspect it was at capacity. Please let me know if your upload succeeds this time.

steffenrost:
@rabbah thanks for the cleanup, but unfortunately the upload still does not work.
```
Size of tarball /tmp/tmpWdN7J7.tar.gz is 285.2 kB
SHA256 hash of tarball is 18241faa9674580274b9ab5633014cec3fb2cd792f93af4dd00bf2ae19da82ca
Uploading to Box...
Posting result <Response [500]>
Internal Server Error
Uploaded Logs with name unit_498200245-sr-action-log-timeout.tar.gz
Time taken for checkAndUploadLogs.sh is 20 secs
```

rabbah:
OK @jonpspri moved the cf app which was running in my account to a new location and so i am not certain where to check what went wrong - i can tell you though i cleaned up the box space which was full. @dgrove-oss do you know where the uploader is running now?

dgrove-oss:
Nope.  No idea how the uploaded is working there days.

steffenrost:
@rabbah @dgrove-oss In `box-upload.py` in `upload_file` the code is using url http://DamCYhF8.mybluemix.net/upload?%s, if the hosting cf app was moved the cf app name has probably changed and hence the url is not valid anymore.

rabbah:
the url is responding so i think the app is fine - i dont know where it's running though (who owns it).
we can move it again to an openwhisk owned account but would prefer to wait for a comment from Jonathan first as he did the last migration.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +465,38 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
...
+          result.response.result.get.fields("error") shouldBe {
+            Messages.timedoutActivation(allowedActionDuration, init = false).toJson
+          }
+          val logs = result.logs.get
+          logs.last should include(Messages.logFailure)
+
+          val parseLogTime = (line: String) => Instant.parse(line.split(' ').head)
+          val startTime = parseLogTime(logs.head)
+          val endTime = parseLogTime(logs.last)
+          between(startTime, endTime).asInstanceOf[FiniteDuration] should be < checkDuration
```
markusthoemmes: Is this `.asInstanceOf[FiniteDuration]` really needed?

steffenrost: We tried without, but got `Cannot resolve reference with such as signature`, so we tried with `asInstanceOf[FiniteDuration]` which worked in intellij, but during compile with grade we get now a compile error, so we still trying to solve this problem.
```
/home/vagrant/github.ibm.com/bluewhisk/open/tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala:498: diverging implicit expansion for type scala.concurrent.duration.FiniteDuration => Comparable[scala.concurrent.duration.FiniteDuration]
starting with method $conforms in object Predef
          between(startTime, endTime).asInstanceOf[FiniteDuration] should be < checkDuration
```

ruediger-maass:
Thanks for your comments, again! We adapted the code and hope that we cover all of your helpful ideas for improvement. The action for the test was a little pain since we did not know that the system test environment only supports nodejs:6 (acc. to https://github.com/apache/incubator-openwhisk/pull/4098).

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +465,38 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
...
+          timeout = Some(allowedActionDuration))
+      }
+      val duration = allowedActionDuration + 3.minutes
+      val checkDuration = allowedActionDuration + 1.minutes
+      val run =
+        wsk.action.invoke(name, Map("durationMillis" -> duration.toMillis.toJson, "delayMillis" -> 100.toJson))
+      withActivation(wsk.activation, run) { result =>
+        withClue("Activation result not as expected:") {
+          result.response.status shouldBe ActivationResponse.messageForCode(ActivationResponse.DeveloperError)
+          result.response.result.get.fields("error") shouldBe {
```
dubee: Shouldn't need curly braces around `Messages.timedoutActivation(allowedActionDuration, init = false).toJson`.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskActivation.scala
```diff
@@ -124,6 +124,11 @@ case class WhiskActivation(namespace: EntityPath,
   }
   def withoutLogs = copy(logs = ActivationLogs()).revision[WhiskActivation](rev)
   def withLogs(logs: ActivationLogs) = copy(logs = logs).revision[WhiskActivation](rev)
+
+  def isTimedoutActivation = {
```
dubee: Can delete curly braces for `isTimedoutActivation` method.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +465,38 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
...
     }
   }
+
+  /**
+   * Test that a heavy logging action is interrupted within its timeout limits.
+   */
+  it should "interrupt the heavy logging action within its time limits" in withAssetCleaner(wskprops) {
+    (wp, assetHelper) =>
+      val name = s"NodeJsTestLoggingActionCausingTimeout-${System.currentTimeMillis()}"
+      print(s"\n create action ${name} using api host: ${wskprops.apihost}..")
```
dubee: Probably don't need this print statement.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -463,4 +465,38 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
...
+          result.response.result.get.fields("error") shouldBe {
+            Messages.timedoutActivation(allowedActionDuration, init = false).toJson
+          }
+          val logs = result.logs.get
+          logs.last should include(Messages.logFailure)
+
+          val parseLogTime = (line: String) => Instant.parse(line.split(' ').head)
+          val startTime = parseLogTime(logs.head)
+          val endTime = parseLogTime(logs.last)
+          between(startTime, endTime) should be < checkDuration.asInstanceOf[Duration]
```
sven-lange-last: You could get rid of the type cast as follows:

```
val duration = between(startTime, endTime)
FiniteDuration(duration.length, duration.unit) should be < allowedActionDuration
```

I checked other usages of `TimingHelpers.between()` and `TimingHelpers.durationOf()`. Actually, it's only used in a few locations. And honestly, I don't understand why these two methods return a `Duration` instead of a `FiniteDuration`. You PR could just change to a `FiniteDuration` - that would make a lot of sense.

The other locations using `TimingHelpers.durationOf()` work around the problem you observed with comparing a `Duration` to a `FiniteDuration` by converting the durations to milliseconds and comparing the resulting values... See https://github.com/apache/incubator-openwhisk/blob/master/tests/src/test/scala/org/apache/openwhisk/core/containerpool/docker/test/DockerContainerTests.scala#L633.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -581,19 +581,30 @@ class ContainerProxy(
               val initRunInterval = initInterval
                 .map(i => Interval(runInterval.start.minusMillis(i.duration.toMillis), runInterval.end))
                 .getOrElse(runInterval)
-              ContainerProxy.constructWhiskActivation(job, initInterval, initRunInterval, response)
+              ContainerProxy.constructWhiskActivation(
+                job,
+                initInterval,
+                initRunInterval,
+                runInterval.duration >= actionTimeout,
```
rabbah: should this be strictly `>`? the action completed here and returned a response.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
rabbah: agree - seems unnecessary.

rabbah: fyi for debugging purposes, there is a verbose flag on the `wsk` class you can flip.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskActivation.scala
```diff
@@ -124,6 +124,11 @@ case class WhiskActivation(namespace: EntityPath,
...
+    annotations.getAs[Boolean](WhiskActivation.timeoutAnnotation).getOrElse(false)
+  }
```
rabbah: This annotation is only added to "invoker" activations - not for compositions which are handled in the controller. Should it be added there also?

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
markusthoemmes: Good call. I wondered about this myself yesterday. Feel free to change the between implementation to FiniteDuration (it's guaranteed to be one anyway)

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskActivation.scala
mhenke1: @rabbah I second your thoughts to to make the time-out annotation consistently available  in the controller. I especially like the added value making the timeout case more visible to the user. Since the underlying problem that triggered this PR is solved with these invoker side changes,  I volunteer to take the controller changes to my list and address it in a subsequent PR.

sven-lange-last: @mhenke1 thanks a lot for volunteering - would you please open a tracking issue for this topic?

I would also highly appreciate to move the annotation work out of this PR to get it done. This PR fixes a problem that should be addressed sooner than later.

@rabbah do you agree with this approach?

mhenke1: @sven-lange-last I have opened #4310 to address the controller side work

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
ruediger-maass: adapted the code to your proposal

tests/dat/actions/loggingTimeout.js
ruediger-maass: We have replaced the test action with a simple nodejs:6 version only using basic parameters (duration and delay)

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
ruediger-maass: We have changed the implementation to use an activation annotation `timeout` (boolean) to indicate action timeouts as proposed by @sven-lange-last.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
ruediger-maass: done; we applied your proposal in the test code

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
ruediger-maass: done, applied your proposal to code

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
ruediger-maass: done, applied your proposal to code

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
ruediger-maass: done, we introduced common code (`logStream`) in `DockerToActivationLogStore` and we added `isTimedoutActivation` function to `WhiskActication`

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
ruediger-maass: done, applied your proposal to the code

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
ruediger-maass: done, applied your proposal to the code

common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala
ruediger-maass: done, handled the case in the new method `isTimedoutActivation` in `WhiskActivation`

common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala
ruediger-maass: done, handled the case in the new method `isTimedoutActivation` in `WhiskActivation`

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
ruediger-maass: in the tests the appendend line `Messages.logFailure` was always looking good, so not sure if an additional newline would be required

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
ruediger-maass: due to the limited runtime support by the system tests this error occurred, we exchanged the action with a plain nodejs:6 version

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationLogStore.scala
ruediger-maass: we applied your proposals

tests/dat/actions/loggingTimeout.js
ruediger-maass: finally, there is only the nodejs:6 version of test action in the change set of the PR

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
steffenrost: We corrected the parameters and removed the commented code.

tests/dat/actions/loggingTimeout.js
steffenrost: Resolved with the new nodejs:6 action.

tests/dat/actions/loggingTimeout.js
steffenrost: The nodejs:6 action now provides default in case of missing arguments. We do not added type checking as this is just test case and we wanted to keep the code simple.

tests/dat/actions/loggingTimeout.js
steffenrost: Done.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
steffenrost: Done according to your proposal.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
steffenrost: Applied.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
steffenrost: Done.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
steffenrost: We finally changed the implementation back to millis because we cannot compare `FiniteDuration` with `should be <`. If someone has an idea how to solve this problem please let us know.

```
> Task :tests:compileTestScala
...
/home/vagrant/github.ibm.com/bluewhisk/open/tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala:496: diverging implicit expansion for type scala.concurrent.duration.FiniteDuration => Comparable[scala.concurrent.duration.FiniteDuration]
starting with method $conforms in object Predef
          between(startTime, endTime) should be < checkDuration
                                                ^
```

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
steffenrost: Done.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskActivation.scala
steffenrost: We deleted the curly braces for `isTimedoutActivation`, we also found other occurrences in `WhiskActivation` with curly braces (not for us) so we cleaned them too.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
steffenrost: Done.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
steffenrost: We changed both methods in `TimingHelpers` to return results using `FiniteDuration` but unfortunately as already mentioned above it seems that we cannot simply compare two `FiniteDuration` values using `should be <` (so we moved back to millis, as said if anyone has an idea how solve please let us know).

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
steffenrost: We tried this, but then the test `KubernetesContainerTests > KubernetesContainer should properly deal with a timeout during run` failed. The testcase using nano seconds whereas the core code uses milli seconds as unit. We talked with @sven-lange-last and decided to revert back to the old implementation based on milli seconds. Sven suggested to not simply change the testcase and keep the lower than comparison for timeout to fix the problem since this change could imply other side effects which should be handled in a separate issue. This issue should also cover changes to use nano seconds for duration instead of milli seconds. 

```
org.apache.openwhisk.core.containerpool.kubernetes.test.KubernetesContainerTests > KubernetesContainer should properly deal with a timeout during run FAILED
    org.scalatest.exceptions.TestFailedException: (Interval(1970-01-01T00:00:00Z,1970-01-01T00:00:01Z),{"result":{"error":"The action did not produce a valid response and exited unexpectedly."},"statusCode":2}) was not equal to (Interval(1970-01-01T00:00:00Z,1970-01-01T00:00:01Z),{"result":{"error":"The action exceeded its time limits of 1000 milliseconds."},"statusCode":2})
```

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
sven-lange-last: @markusthoemmes I think we should accept the latest results and proposal by @steffenrost @ruediger-maass and compare the `Duration`s millis instead of the `Duration`s millis.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
sven-lange-last: @markusthoemmes I think we should accept the latest results and proposal by @steffenrost @ruediger-maass and compare the `Duration`s millis instead of the `Duration`s millis. The Scala compile error from above is hard to solve and - apparently - nobody else solved it so far because the whole project seems to have no `Duration` comparisons with `should be`.

@mhenke1 found a solution where you need to define an implicit in scope to make sure that `should be` is able to compare `Duration`s properly. Not very comfortable to read...

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
rabbah: thank you for the thoughtful explanation.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskActivation.scala
rabbah: > @rabbah do you agree with this approach?

💯, agree.

sven-lange-last:
@steffenrost @ruediger-maass thanks a lot for fixing this invoker issue. From my side, all comments have been addressed properly. You spent pretty much time in troubleshooting problems caused by changes that you made to address review comments. Thanks for your patience with reviewers. Whenever possible with reasonable effort, you followed reviewers' requests and recommendations.

Can you please run a PG with your branch to verify and paste the net result into this PR?

@markusthoemmes @rabbah @dubee please check whether you are ok now with the PR as well.

I suggest to not extend the scope of this PR any more and move all further enhancements out to separate issues. Do you agree with this approach?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
ruediger-maass: additional remark to the comment above regarding "old implementation based on milli seconds": it is meant that we have changed back to `<=` comparisons with timeouts.

sven-lange-last:
PG3 / 3370 running.

steffenrost:
@sven-lange-last PG3 / 3371 run almost fine, only the two non-related testcases (`whisk.core.bluemix.MonitoringServiceAlertAnalysisTests.Pagerduty-observer`) failed.

ruediger-maass:
@dubee Thank you for merging the PR!

mdeuser:
@ruediger-maass - does any doc need updating to describe the new activation annotation?
> you did a great job with introducing a new annotation to mark an activation as timed out. This is much more robust as the previous approach(es) and also provides valuable information to the end user.

ruediger-maass:
@mdeuser @mhenke1 @steffenrost - yes, the doc should be updated. There is one missing piece we wanted to add: activations of sequence functions should also contain the annotation in case of timeouts in order to obtain a consistent behavior. We thought that this could be the right time to "officially" document the timeout annotation.

Issue #4315, opened 2019-03-01
Unify version of various Akka modules


chetanmeh:
While working on PR #4314 which uses [akka-stream-kafka][1] which in turn uses latest Akka 2.5.21 following warning is seen on startup

```
[2019-02-28T20:59:44.394Z] [WARN] Detected possible incompatible versions on the classpath. Please note that a given Akka version MUST be the same across all modules of Akka that you are using, e.g. if you use [2.5.21] all other modules that are released together MUST be of the same version. Make sure you're using a compatible set of libraries. Possibly conflicting versions [2.5.3, 2.5.21, 2.5.12, 2.5.7] in libraries [akka-cluster-metrics:2.5.3, akka-protobuf:2.5.21, akka-actor:2.5.21, akka-slf4j:2.5.12, akka-remote:2.5.7, akka-cluster:2.5.7, akka-distributed-data:2.5.3, akka-stream:2.5.21, akka-stream-testkit:2.5.21, akka-cluster-tools:2.5.3]
```

It seems currently we have 2 different set of Akka version used

* akka-stream - 2.5.12 (above shows 2.5.21 due to alpakka dep)
* akka-cluster - 2.5.7 via mesos-actor dep

We should unify the version and also at same time now switch to latest Akka stack version i.e.

* Akka Stream 2.5.21 (from 2.5.12)
* Akka HTTP - 10.1.7 (from 10.1.1)
* Scala 2.12.8 (from 2.12.7)

[1]: https://doc.akka.io/docs/akka-stream-kafka/current/home.html

chetanmeh:
This warning now does not appear with #4316 changes merged.

Issue #4217, opened 2019-01-15
Update to Java 11


chetanmeh:
We should plan to move to Open JDK 11 which is the next LTS release. This issue is meant to track relevant work required to support that

- [ ] Scala - Per current [compatibility][1] matrix JDK 11 looks like still not supported. See scala/scala-dev#559. 
- [ ] Akka - Would depend on compatible Scala release

May be its fine to use latest Scala if the JPMS module support is not required. That would need to be validated

[1]: https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html

csantanapr:
Maybe is time to upgrade now?

chetanmeh:
Tried to run test based on [this branch][1]. but [build failed][2] for multiple reasons.

Some of the unit tests are failing due to change in behaviour of [Instant][3] which is causing `Instant` in JDK 11 to use microsecond precision while we store the time in db in millis. So when entity instances are reconstructed from db the Instant value differs from one created directly. 

See such failed test [here][4]. Most failure are due to equality test failing for entities ... most likely due to difference in `Instant` values.

Other failures are yet not analyzed!

[1]: https://github.com/chetanmeh/incubator-openwhisk/tree/jdk-11-build
[2]: https://travis-ci.org/chetanmeh/incubator-openwhisk/builds/538141401
[3]: https://github.com/h2database/h2database/issues/1178
[4]: https://scans.gradle.com/s/sf7udehikj6g2/tests/failed

mcdan:
I've proposed a PR that both jumps us to java 11 and a later scala version. All the travis tests pass but perhaps there is something else I should run?

rabbah:
LGTM thanks Dan.

rabbah:
Thanks Dan. I reviewed and approve the PR; it lgtm.

-r

PR #4316, opened 2019-03-01
Update version for Akka dependencies and Scala

chetanmeh:
Updates the following dependencies

* Akka modules - 2.5.12 -> 2.5.22
* Akka HTTP - 10.1.1 -> 10.1.8
* Scala - 2.12.7 - 2.12.8

Required for #4315 and #4217

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#4315)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
@@ -53,8 +53,10 @@ class PoolingRestClient(
 
   //if specified, override the ClientConnection idle-timeout and keepalive socket option value
   private val timeoutSettings = {
-    ConnectionPoolSettings(system.settings.config).withUpdatedConnectionSettings { s =>
-      timeout.map(t => s.withIdleTimeout(t)).getOrElse(s)
+    val s = ConnectionPoolSettings(system.settings.config)
+    timeout match {
+      case Some(t) => s.withMaxConnectionBackoff(t).withUpdatedConnectionSettings(cs => cs.withIdleTimeout(t))
+      case None    => s
     }
   }
 
--- a/common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
@@ -55,7 +55,7 @@ class PoolingRestClient(
   private val timeoutSettings = {
     val s = ConnectionPoolSettings(system.settings.config)
     timeout match {
-      case Some(t) => s.withMaxConnectionBackoff(t).withUpdatedConnectionSettings(cs => cs.withIdleTimeout(t))
+      case Some(t) => s.withMaxConnectionBackoff(t).withUpdatedConnectionSettings(_.withIdleTimeout(t))
       case None    => s
     }
   }
--- a/common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
@@ -54,13 +54,13 @@ class PoolingRestClient(
   //if specified, override the ClientConnection idle-timeout and keepalive socket option value
   private val timeoutSettings = {
     val cps = ConnectionPoolSettings(system.settings.config)
-    timeout match {
-      case Some(t) =>
+    timeout
+      .map { t =>
         cps
           .withMaxConnectionBackoff(cps.maxConnectionBackoff.min(t))
           .withUpdatedConnectionSettings(_.withIdleTimeout(t))
-      case None => cps
-    }
+      }
+      .getOrElse(cps)
   }
 
   // Creates or retrieves a connection pool for the host.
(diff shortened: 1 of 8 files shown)


chetanmeh:
Currently 3 test failures seen. Would need some analysis

chetanmeh:
[2 tests failed][1] due to large request size limit. Seems like as part of akka/akka-http#2137 a new [limit setting][2] was introduced to avoid DOS attack when using gzip encoding.

```
# The maximum number of bytes to allow when reading the entire entity into memory with `toStrict`
    # (which is used by the `toStrictEntity` and `extractStrictEntity` directives)
    max-to-strict-bytes = 8m
```

Now bumped this to 50m also which fixes these 2 failing tests

[1]: https://scans.gradle.com/s/tgjrf2sb6wmwa/tests/failed
[2]: https://github.com/akka/akka-http/blob/d02dbd1043477e371e6bbdea04d3cec58c025c76/akka-http-core/src/main/resources/reference.conf#L435

chetanmeh:
There is [another failure][1] in AkkaContainerClientTests "AkkaContainerClient should retry till timeout on StreamTcpException"

Seems like now Akka HTTP Client has [inbuilt exponential backoff][2] which tries till 2 mins. @tysonnorris should we adapt the test with higher limit or lower the max configured backoff ?

[1]: https://scans.gradle.com/s/fie7qruoqu2ks/tests/xaqugzlgs2zxa-dgsewypul5hrw
[2]: https://github.com/akka/akka-http/commit/13add16f150068a140dcf338c385cc96a9a036d6#diff-b4fbbc15badf58a0b02c9fd961f0aa2f

tysonnorris:
@chetanmeh this seems like it will require a code change in `AkkaContainerClient`; I think we should get the same exception handling behavior, regardless of whether the underlying retry mechanism causes the retry, or we exceed the timeout without catching exceptions. I'm not sure what exact behavior changed there, but we may need to add logic for testing the timeout somewhere besides the ```request.map().recoverWith()...``` that is currently there?

chetanmeh:
@markusthoemmes @tysonnorris Please review changes in `PoolingRestClient`. I have used the provided `timeout` as upper limit for new backoff support. 

So client usage where timeout is not provided would now onwards make use of default exponential backoff strategy (upto max 2 mins) and where we specify timeout (mostly in action invocation) it would be bounded by that

chetanmeh:
@dubee Please run a PG for this

chetanmeh:
@dubee Above log would come as akka cluster dependencies are being pulled via [mesos-actor][1] dependency which is currently bound to an older version of akka and thus causes this warning.

To ensure that we use a consistent version we should probably add an explicit dependency on akka cluster modules in common as controller depends on that feature directly. This would ensure that we do not have to coordinate version across our usage and mesos-actor module.

@tysonnorris @markusthoemmes @dubee @rabbah Thoughts?

PS: [akka 2.5.22](https://akka.io/blog/news/2019/04/03/akka-2.5.22-released) is also now release with some major changes considered stable (Typed actor and Artery implementation for remoting). May be we switch to that also

[1]: https://github.com/adobe-apiplatform/mesos-actor/blob/master/build.gradle#L33

common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
```diff
@@ -53,8 +53,13 @@ class PoolingRestClient(
...
   private val timeoutSettings = {
-    ConnectionPoolSettings(system.settings.config).withUpdatedConnectionSettings { s =>
-      timeout.map(t => s.withIdleTimeout(t)).getOrElse(s)
+    val cps = ConnectionPoolSettings(system.settings.config)
+    timeout match {
+      case Some(t) =>
+        cps
+          .withMaxConnectionBackoff(cps.maxConnectionBackoff.min(t))
+          .withUpdatedConnectionSettings(_.withIdleTimeout(t))
+      case None => cps
```
markusthoemmes: I'd keep the map/getOrElse here

```scala
timeout.map{ t =>
  cps
    .withMaxConnectionBackoff(cps.maxConnectionBackoff.min(t))
    .withUpdatedConnectionSettings(_.withIdleTimeout(t))
}.getOrElse(cps)
```

chetanmeh:
I have now bumped the version to current latest. Also to ensure some of the transitive dependencies from mesos-actor and s3 connector confirm to same version now [constraints](https://docs.gradle.org/4.10.2/userguide/managing_transitive_dependencies.html#sec:dependency_constraints) are placed to enforce that.

With these I do not see warning now

@dubee If needed then please run another PG

@rabbah This PR is now read to be merged!

common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
```diff
@@ -53,9 +53,14 @@ class PoolingRestClient(
 
   //if specified, override the ClientConnection idle-timeout and keepalive socket option value
   private val timeoutSettings = {
-    ConnectionPoolSettings(system.settings.config).withUpdatedConnectionSettings { s =>
-      timeout.map(t => s.withIdleTimeout(t)).getOrElse(s)
-    }
+    val cps = ConnectionPoolSettings(system.settings.config)
+    timeout
+      .map { t =>
+        cps
```
rabbah: this doesn't NPE with the reference to `cps` that's defined on line 56?

chetanmeh: Not getting it ... `cps` is defined at L56 and then referred later. So should be fine?

PR #4325, opened 2019-03-06
Adjust error handling in the Cloudant Rest Store

mhenke1:
## Description
This PR cleans the error handling in the CouchDBRestStore so that metrics and logs are not send twice. It also cleans a case were a success metric was send in an error situation.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/database/ArtifactStoreExceptions.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/database/ArtifactStoreExceptions.scala
@@ -27,7 +27,13 @@ case class DocumentTypeMismatchException(message: String) extends ArtifactStoreE
 
 case class DocumentUnreadable(message: String) extends ArtifactStoreException(message)
 
+case class GetException(message: String) extends ArtifactStoreException(message)
+
 case class PutException(message: String) extends ArtifactStoreException(message)
+
+case class DeleteException(message: String) extends ArtifactStoreException(message)
+
+case class QueryException(message: String) extends ArtifactStoreException(message)
 
 sealed abstract class ArtifactStoreRuntimeException(message: String) extends RuntimeException(message)
 
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/database/CouchDbRestStore.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/database/CouchDbRestStore.scala
@@ -129,7 +129,7 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
             start,
             s"[PUT] '$dbName' failed to put document: '${docinfoStr}'; http status: '${code}'",
             ErrorLevel)
-          throw new Exception("Unexpected http response code: " + code)
+          throw new PutException("Unexpected http response code: " + code)
       }
     }
 
@@ -164,7 +164,7 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
 
         case Left(code) =>
           transid.failed(this, start, s"'$dbName' failed to put documents, http status: '${code}'", ErrorLevel)
-          throw new Exception("Unexpected http response code: " + code)
+          throw new PutException("Unexpected http response code: " + code)
       }
     }
 
@@ -200,7 +200,7 @@ class CouchDbRestStore[DocumentAbstraction <: DocumentSerializer](dbProtocol: St
             start,
             s"[DEL] '$dbName' failed to delete document: '${doc}'; http status: '${code}'",
             ErrorLevel)
-          throw new Exception("Unexpected http response code: " + code)
+          throw new DeleteException("Unexpected http response code: " + code)
       }
     }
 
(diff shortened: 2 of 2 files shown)


core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
```diff
@@ -244,7 +244,8 @@ class InvokerReactive(
                     ActivationResponse.applicationError(Messages.actionRemovedWhileInvoking)
                   case _: DocumentTypeMismatchException | _: DocumentUnreadable =>
                     ActivationResponse.whiskError(Messages.actionMismatchWhileInvoking)
-                  case _ =>
+                  case e =>
+                    logging.error(this, s"${Messages.actionFetchErrorWhileInvoking} Error: ${e.getMessage}")
```
sven-lange-last: Thanks for this helpful addition. I suggest to cover also the other two `case`s because these are system errors as well.

markusthoemmes: Isn't the error already printed by the database layer? If not: Should it be put there?

sven-lange-last: @markusthoemmes are you talking about the `ArtifactStore` implementations - for example: https://github.com/apache/incubator-openwhisk/blob/master/common/scala/src/main/scala/org/apache/openwhisk/core/database/CouchDbRestStore.scala#L217?

If yes - I missed that we already log information there.

Honestly, I don't understand what's going on here: https://github.com/apache/incubator-openwhisk/blob/master/common/scala/src/main/scala/org/apache/openwhisk/core/database/CouchDbRestStore.scala#L242. First, we finished the transaction and later on via `reportFailure()`, we fail it because an exception is thrown?

mhenke1:
I have added a small changes that adjusts the error logging in the CouchDBRestStore to the 
behavior of the other crud methods.

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
rabbah: Every error in the artifact store is already logged as Markus noted. reportFailure is there to make sure the error isn’t lost. It doesn’t modify the future.

rabbah: https://github.com/apache/incubator-openwhisk/blob/be1e3a19c02956c9be85023a0bb0ff399c21444d/common/scala/src/main/scala/org/apache/openwhisk/core/database/StoreUtils.scala#L37

All failured future get logged there. So the earlier change above is redundant isn’t it?

sven-lange-last: @rabbah You are right. I missed that... Can you please have a look here?

My understanding is that we have the following flow:
1. `Left(code)` matches.
2. `transid.finished()` is called.
3. A new exception is thrown, failing Future `f`.
4. `reportFailure()` is called with the failed `f`. Error handling calls `transid.failed()`.

Do your agree? What's the rationale behind finishing the transaction first and then failing it?

https://github.com/apache/incubator-openwhisk/blob/a08175f05cd9bf21e48eea52be62ca0030f9c2c3/common/scala/src/main/scala/org/apache/openwhisk/core/database/CouchDbRestStore.scala#L242-L258

rabbah: your 1-4 is correct... but I don't understand the question about `transid.failed` and specifically "failing it". The `reportFailure` method call to `transid.failed` only serves to log the failed future: https://github.com/apache/incubator-openwhisk/blob/a92d14c3cd685ec8c9caa530606a82bb9aed9405/common/scala/src/main/scala/org/apache/openwhisk/common/TransactionId.scala#L137

sven-lange-last: Well, on the same `transid`, we first call `finished()` and then `failed()`. So for one single `transid`, we have two log entries - one indicating success and a second indicating failure. In addition, two metrics are sent. Correct?

rabbah: thanks for conferring on slack - we agree there is duplication: the `finished` message is logged at `InfoLevel` and `failed` at `ErrorLevel` and both of these generate metrics. The future is not modified in either case.

Given the duplication, and if all the paths in the future are already logging, there is less of a need for `reportFailure`. So I think that should be addressed. With the change @mhenke1 made, now `failed` will be called twice on failure. Further I think the change in the Invoker is not necessary and redundant.

Now back to the existential question: if we agree there should have been an Error level message reported anyone on the failed future, what were you missing from the logs?

rabbah:
@mhenke1 @sven-lange-last what do you want to do about this PR?

After reviewing the comments again, it seems to me the best resolution is to audit all uses of `reportFailure` to remove duplication of `transid` markers and possibly elide uses of `reportFailure` entirely as the artifact store has evolved since the original pattern was introduced.

mhenke1:
@rabbah the PR is still on my list. I hope to start working  on it next week. I agree to get  rid of ‘reportFailure‘ if possible.

mhenke1:
@rabbah @markusthoemmes @sven-lange-last  Since you have already commented, please have a look again (see https://github.com/apache/incubator-openwhisk/pull/4325#issuecomment-486134486 for my thoughts to not remove `reportFailure`).

PR #4329, opened 2019-03-09
Redirect http to https and other nginx config cleanup.

rabbah:
This PR allows SSL certificates for the openwhisk nginx to be copied from an arbitrary path, and provides overrides for the certificate names. I believe these were intended by past changes there were some gaps. For example, the `ssl.path` value was not properly used in the nginx script and the files had to be copied into the ansible directory. Also the certificate names could not be easily overriden (one would have to specify a full json value).

As part of, I removed an unused exposed port, and redirected http to https as well.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [x] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -29,7 +29,7 @@ whisk_api_localhost_name_default: "localhost"
 environmentInformation:
   type: "{{ environment_type | default('local') }}"
 
-hosts_dir: "{{ inventory_dir | default(env_hosts_dir) }}"
+hosts_dir: "{{ inventory_dir | default(env_hosts_dir) }}"
 
 whisk:
   version:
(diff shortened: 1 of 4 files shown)


ansible/group_vars/all
```diff
@@ -209,15 +209,14 @@ nginx:
   version: "{{ nginx_version | default('1.13') }}"
   port:
     http: 80
-    api: 443
+    https: 443
   ssl:
-    path: "{{ openwhisk_home }}/ansible/roles/nginx/files"
-    cert: "openwhisk-server-cert.pem"
-    key: "openwhisk-server-key.pem"
-    password_enabled: false
```
rabbah: it's not necessary to have both password_enabled and the password_file parameter - this chance consolidates both.

ansible/roles/nginx/tasks/deploy.yml
```diff
@@ -16,7 +16,7 @@
 
 - name: copy cert files from local to remote in nginx config directory
   copy:
-    src: "files/{{ item }}"
+    src: "{{ nginx.ssl.path }}/{{ item }}"
```
rabbah: now the certs can be copied from the given location instead.

ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -89,20 +89,27 @@ http {
 {# Turn off sending information about the server to the client #}
     server_tokens off;
 
+    # Redirect all http to https.
+    server {
+        listen 80 default_server;
+        server_name _;
+        return 301 https://$host$request_uri;
```
style95: In some down streams, they might want to use http for some performance reason.
How about making this redirection configurable?

rabbah: I could do that but note that port 80 wasn’t open at all - so one would have had to provide their own nginx anyway?

style95: ok, I just thought they might take advantage of this change.
(by just using upstream default conf)

rabbah: i can remove this if it's controversial -- my thinking was that we have not supported port 80 but the port was exposed... so this change redirect http to https.

i think the reason not to support http in the project is that it wouldn't/shouldn't be used in production, and we've had support for self signed cert for local development, and without a test it would rot.

incidentally, i could add a test for the http redirect.

dgrove-oss:
PG2 4291 🏃

PR #4334, opened 2019-03-12
Feature flag support to turn on/off support for provide-api-key annotation

chetanmeh:
Enables support for feature flags which for now is used to enable or disable support for `provide-api-key` (#4284)

## Description

With #4284 support was added for `provide-api-key` annotation. In some cases we may want to enable this feature later. This PR introduces a feature flag support which allow disabling such feature based on config. To disable this feature use config like below

```
whisk {
    feature-flags {
        require-api-key-annotation = false
    }
}
```

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
@@ -33,7 +33,7 @@ import akka.http.scaladsl.unmarshalling._
 import spray.json._
 import spray.json.DefaultJsonProtocol._
 import org.apache.openwhisk.common.TransactionId
-import org.apache.openwhisk.core.WhiskConfig
+import org.apache.openwhisk.core.{FeatureFlags, WhiskConfig}
 import org.apache.openwhisk.core.controller.RestApiCommons.{ListLimit, ListSkip}
 import org.apache.openwhisk.core.controller.actions.PostActionActivation
 import org.apache.openwhisk.core.database.{ActivationStore, CacheChangeNotification, NoDocumentException}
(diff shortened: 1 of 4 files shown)


core/controller/src/main/scala/org/apache/openwhisk/core/controller/Actions.scala
```diff
@@ -67,7 +67,7 @@ object WhiskActionsApi {
    * 2. An [[execAnnotation]] consistent with the action kind; this annotation is always added and overrides a pre-existing value
    */
   protected[core] def amendAnnotations(annotations: Parameters, exec: Exec, create: Boolean = true): Parameters = {
-    val newAnnotations = if (create) {
+    val newAnnotations = if (create && FeatureFlags.requireApiKeyAnnotation) {
```
rabbah: i suspect that if this is set to false, some tests will fail.

did you consider also updating this for completeness although there may be another place or two that don't use this method that could be affected.

https://github.com/apache/incubator-openwhisk/blob/fb0bab64de6634725b409a63ffbb90fee295e8c1/tests/src/test/scala/org/apache/openwhisk/core/controller/test/ControllerTestCommon.scala#L88-L95

chetanmeh: For now I do not think that we should make test case complex with feature flag support. So test should work assuming all new features enabled (unless using some SPI mechanism). For main code we just validate manually that if feature is disabled then we get back the earlier behavior.

falkzoll:
This looks great :-). Exactly what I was thinking about. Many thanks for providing this implementation. For me this is a good example and I will take some time and look at the code to understand how it has been implemented. 👍

PR #4388, opened 2019-03-22
Adapt limit to query db view as per couch db change

SugandhaAgrawal:
Co-authored-by: steffenrost <srost@de.ibm.com>

CouchDB has introduced a limit check for querying the database view, for the query parameter `limit`. In OpenWhisk it was first set to Scala `Int.Max` value. Since this exceeds the allowed range by couchDB, the query calls using the max value have been reported failing with `http Response: 400 Bad Request`. 

For ref: https://github.com/apache/couchdb/commit/1fdfe466223380d1ee7c82e21283e161290aed87#diff-2bd31f00dcfba780c6c65bc69c256854 

## Description
Changed the query parameter `limit` to max allowed by couchDB. 

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/NamespaceBlacklist.scala
+++ b/core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/NamespaceBlacklist.scala
@@ -47,6 +47,7 @@ class NamespaceBlacklist(authStore: AuthStore) {
   def isBlacklisted(identity: Identity): Boolean = blacklist.contains(identity.namespace.name.asString)
 
   /** Refreshes the current blacklist from the database. */
+  /** Limit query parameter set to 0 for limitless record query. */
   def refreshBlacklist()(implicit ec: ExecutionContext, tid: TransactionId): Future[Set[String]] = {
     authStore
       .query(
@@ -54,7 +55,7 @@ class NamespaceBlacklist(authStore: AuthStore) {
         startKey = List.empty,
         endKey = List.empty,
         skip = 0,
-        limit = Int.MaxValue,
+        limit = 0,
         includeDocs = false,
         descending = true,
         reduce = false,


asfgit:
Can one of the admins verify this patch?

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/NamespaceBlacklist.scala
```diff
@@ -38,6 +38,9 @@ class NamespaceBlacklist(authStore: AuthStore) {
 
   private var blacklist: Set[String] = Set.empty
 
+  // Enforced by couchdb, max limit value for db view query
+  private var couchdb_limit: Int = 268435456
```
markusthoemmes: This should be a `val` and as it's a constant should be written in all caps, like `COUCHDB_LIMIT`.

However, I feel like this should be factored into the couchdb code itself and the `limit` value should have an upper-bound applied there. Technically you do not know in this piece of code that you're handling CouchDB.

SugandhaAgrawal: Fair point!

rabbah:
> In OpenWhisk it was first set to Scala Int.Max value.

Can you point me to it? I thought it was set to `0` not max int.

rabbah:
OK I found this reference https://github.com/apache/incubator-openwhisk/blob/be1e3a19c02956c9be85023a0bb0ff399c21444d/core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/NamespaceBlacklist.scala#L57

This should just be set to 0 which is what the collections do.

steffenrost:
> See my comments - the simpler change is to change Int.Max to 0.

But setting the limit query parameter to 0 (`limit = 0,`)would limit the number of returned blocked namespace in the result set to 0, is that the intended behavior?

rabbah:
The couchdb client in openwhisk will omit the `limit` query parameter when the value is set to `0`:
https://github.com/apache/incubator-openwhisk/blob/be1e3a19c02956c9be85023a0bb0ff399c21444d/common/scala/src/main/scala/org/apache/openwhisk/core/database/CouchDbRestClient.scala#L124

So the caller (here the invoker) should set the value to `0` to indicate no limit.

rabbah:
In general, using "no limit" seems wrong here anyway and the invoker should paginate since it's conceivable there could be a large enough response to both cause a performance problem and a memory problem. In practice, I presume the responses are typically small not to be an issue.

SugandhaAgrawal:
@rabbah - Adapted as our discussion. Please Review!!! TIA

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/NamespaceBlacklist.scala
```diff
@@ -47,14 +47,15 @@ class NamespaceBlacklist(authStore: AuthStore) {
   def isBlacklisted(identity: Identity): Boolean = blacklist.contains(identity.namespace.name.asString)
 
   /** Refreshes the current blacklist from the database. */
+  /** Limit query parameter set to 0 for limitless record query */
```
rabbah: Please terminate the sentence with a period.

asfgit:
Refer to this link for build results (access rights to CI server needed): 
https://builds.apache.org/job/OpenWhisk-Playground/86/
Test FAILed.
Test FAILured.

asfgit:
Refer to this link for build results (access rights to CI server needed): 
https://builds.apache.org/job/OpenWhisk-Playground/87/
Test FAILed.
Test FAILured.

PR #4403, opened 2019-03-27
Add detailed error and reason information to status code

steffenrost:
Co-authored-by: Sugandha Agrawal <sugandha.agrawal18@gmail.com>

This PR is related to https://github.com/apache/incubator-openwhisk/pull/4388. Seeing only a http response 400 Bad Request isn't enough to understand the root cause, but the detailed error and reason information which may exist in the error response will do.

## Description
This code change adds the error and reason info of the error response to the the HTTP status code.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
@@ -110,9 +110,16 @@ class PoolingRestClient(
       if (response.status.isSuccess) {
         Unmarshal(response.entity.withoutSizeLimit).to[T].map(Right.apply)
       } else {
-        // This is important, as it drains the entity stream.
-        // Otherwise the connection stays open and the pool dries up.
-        response.discardEntityBytes().future.map(_ => Left(response.status))
+        Unmarshal(response.entity).to[String].flatMap { body =>
+          val statusCode = response.status
+          val reason =
+            if (body.nonEmpty) s"${statusCode.reason} (details: $body)" else statusCode.reason
+          val customStatusCode = StatusCodes
+            .custom(intValue = statusCode.intValue, reason = reason, defaultMessage = statusCode.defaultMessage)
+          // This is important, as it drains the entity stream.
+          // Otherwise the connection stays open and the pool dries up.
+          response.discardEntityBytes().future.map(_ => Left(customStatusCode))
+        }
       }
     }
 
--- a/tests/src/test/scala/org/apache/openwhisk/http/PoolingRestClientTests.scala
+++ b/tests/src/test/scala/org/apache/openwhisk/http/PoolingRestClientTests.scala
@@ -138,12 +138,15 @@ class PoolingRestClientTests
   }
 
   it should "return a status code on request failure" in {
-    val httpResponse = HttpResponse(NotFound)
+    val body = "Limit is too large, must not exceed 268435456"
+    val httpResponse = HttpResponse(NotFound, entity = body)
     val httpRequest = HttpRequest(entity = HttpEntity(ContentTypes.`application/json`, JsObject.empty.compactPrint))
     val poolingRestClient = new PoolingRestClient("https", "host", 443, 1, Some(testFlow(httpResponse, httpRequest)))
     val request = mkJsonRequest(GET, Uri./, JsObject.empty, List.empty)
 
-    await(poolingRestClient.requestJson[JsObject](request)) shouldBe Left(NotFound)
+    val reqResult = await(poolingRestClient.requestJson[JsObject](request))
+    reqResult shouldBe Left(NotFound)
+    reqResult.left.get.reason shouldBe s"${NotFound.reason} (details: ${body})"
   }
 
   it should "throw an unsupported content-type exception when unexpected content-type is returned" in {


common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
```diff
@@ -110,9 +112,17 @@ class PoolingRestClient(
       if (response.status.isSuccess) {
         Unmarshal(response.entity.withoutSizeLimit).to[T].map(Right.apply)
       } else {
-        // This is important, as it drains the entity stream.
-        // Otherwise the connection stays open and the pool dries up.
-        response.discardEntityBytes().future.map(_ => Left(response.status))
+        Unmarshal(response.entity).to[JsObject].flatMap { responseEntity =>
+          val error = responseEntity.fields.get("error").getOrElse("").toString
+          val reason = responseEntity.fields.get("reason").getOrElse("").toString
```
markusthoemmes: Can you encode this schema into a type? If both of them are optional it'll be like:

```scala
case class Foo(error: Option[String], reason: Option[String])
```

You can then use that in the `Unmarshal(...).to[Foo]` and use the unmarshalled entity directly rather than relying on an implicit schema.

common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
```diff
@@ -110,9 +112,17 @@ class PoolingRestClient(
...
+          val statusCode = response.status
+          val statusCodeReason = statusCode.reason()
+          val customReason = s"$statusCodeReason (error: $error, reason: $reason)"
+          val customStatusCode = StatusCodes.custom(intValue = statusCode.intValue(), reason = customReason, defaultMessage = statusCode.defaultMessage())
```
markusthoemmes: It feels... weird to overload StatusCode with more meaning than is already in the HTTP spec. Should we adjust the future to return `Either[ExpressiveErrorType, T]` rather than just statuscode?

rabbah:
@steffenrost can you clarify what the connection to #4388 is?

steffenrost:
> @steffenrost can you clarify what the connection to #4388 is?
> 
> Is it that the cloudant response is failing and you don't know why - and this change is to increase insight into the http call that's failing?

@rabbah yes that is the idea, in case of #4388 we only saw  that the cloudant view call was failing with 400 (bad request) in the log but had no indication about the root cause, to find out we had to reproduce the issue locally and in addition add temporary log statements. So we think it is better to have detailed error information already in the log if possible (eg `org.scalatest.exceptions.TestFailedException: The future returned an exception of type: java.lang.Exception, with message: Unexpected http response code: 400 Bad Request (error: "query_parse_error", reason: "Limit is too large, must not exceed 268435456")` instead of finding out after the fact.

common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
steffenrost: > It feels... weird to overload StatusCode with more meaning than is already in the HTTP spec. Should we adjust the future to return `Either[ExpressiveErrorType, T]` rather than just statuscode?

But wouldn't that be an interface change implying changes in many classes (including tests) doing REST calls using `requestJson` in `PoolingRestClient` as implementation?

markusthoemmes: It would, yes.

common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
markusthoemmes: Thinking about it, there is no general schema to use here, right? Could we just unmarshal the body as a string and append that somewhere?

common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
steffenrost: Would you both agree that having this info available in the log is valuable? 

Regarding defining an own error type, as this is a breaking change not only in openwhisk, would you prefer it anyway (the implementation only extends the reason field, but the new StatusCode is derived from the orginal one)? Another possibility would be to simply log the valuable info from the response entity in method requestJson in case of an error.

Should be go forward with this PR and which way do you suggest?

rabbah:
Thanks for the explanation @steffenrost. It sounds like having this info in the logs is helpful. So whatever is the minimal change to allow you to proceed, is ok with me.

common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
markusthoemmes: Given all the other comments, I'd say it's okay to move forward this way.

However, I would not parse the response as JSON but as a String. That way you catch many more use-cases than just the very narrow one outlined here. WDYT?

rabbah: +1 there’s no guarantee response is json.

steffenrost: Yes makes sense, I changed the implementation to parse String instead of JSON. For the discussed error case the error now looks as follow:

`Unexpected http response code: 400 Bad Request (details:{"error":"query_parse_error","reason":"Limit is too large, must not exceed 268435456"} ) [marker:database_queryView_error:402:33]`

rabbah: LGTM. you can remove `()` on `.intValue()` and `.defaultMessage()`.

common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
```diff
@@ -110,9 +110,16 @@ class PoolingRestClient(
       if (response.status.isSuccess) {
         Unmarshal(response.entity.withoutSizeLimit).to[T].map(Right.apply)
       } else {
-        // This is important, as it drains the entity stream.
-        // Otherwise the connection stays open and the pool dries up.
-        response.discardEntityBytes().future.map(_ => Left(response.status))
+        Unmarshal(response.entity).to[String].flatMap { errorDetails =>
+          val statusCode = response.status
+          val statusCodeReason = statusCode.reason()
```
markusthoemmes: You can drop the `()` here.

common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
```diff
@@ -110,9 +110,16 @@ class PoolingRestClient(
       if (response.status.isSuccess) {
         Unmarshal(response.entity.withoutSizeLimit).to[T].map(Right.apply)
       } else {
-        // This is important, as it drains the entity stream.
-        // Otherwise the connection stays open and the pool dries up.
-        response.discardEntityBytes().future.map(_ => Left(response.status))
+        Unmarshal(response.entity).to[String].flatMap { errorDetails =>
```
markusthoemmes: `errorDetails` a.k.a `body`?

common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
```diff
@@ -110,9 +110,16 @@ class PoolingRestClient(
...
+          val statusCode = response.status
+          val statusCodeReason = statusCode.reason()
+          val customReason = s"$statusCodeReason (details:$errorDetails)"
```
markusthoemmes: I'd call out that `details:` is actually the response's `body:`. Also please add a space char for good measure.

steffenrost: @markusthoemmes, @rabbah thanks for the final comments!

steffenrost:
> @steffenrost and I agreed on adding a test for the new behavior.
> 
> The change should be around https://github.com/apache/incubator-openwhisk/blob/master/tests/src/test/scala/org/apache/openwhisk/http/PoolingRestClientTests.scala#L140-L147

We extended the test to check in addition for status code reason.

common/scala/src/main/scala/org/apache/openwhisk/http/PoolingRestClient.scala
```diff
@@ -110,9 +110,16 @@ class PoolingRestClient(
       if (response.status.isSuccess) {
         Unmarshal(response.entity.withoutSizeLimit).to[T].map(Right.apply)
       } else {
-        // This is important, as it drains the entity stream.
-        // Otherwise the connection stays open and the pool dries up.
-        response.discardEntityBytes().future.map(_ => Left(response.status))
+        Unmarshal(response.entity).to[String].flatMap { body =>
+          val statusCode = response.status
+          val reason =
+            if (body.length > 0) s"${statusCode.reason} (details: $body)" else statusCode.reason
```
markusthoemmes: Please use `body.nonEmpty`. Computing the whole length of a string (or any datastructure really) **can** be O(n), whereas `isEmpty`/`nonEmpty` is **always** O(1)

tests/src/test/scala/org/apache/openwhisk/http/PoolingRestClientTests.scala
```diff
@@ -138,12 +138,19 @@ class PoolingRestClientTests
   }
 
   it should "return a status code on request failure" in {
-    val httpResponse = HttpResponse(NotFound)
+    val customStatusCodeNotFound = StatusCodes
```
markusthoemmes: As discussed in Slack, this does not actually test what you implemented. It rather tests if a given statuscode is forwarded as is.

Rather, you should add the "Limit is too large.." string as a body to the HttpResponse like so:

```scala
val httpResponse = HttpResponse(NotFound, entity = "Limit is too large, must not exceed 268435456")
```

Then you run that through the PoolingRestClient as is, and in the end you can check that the rest client did it's job to add that info to the StatusCode **it** returns (not the one it got):

```scala
val reqResult = await(poolingRestClient.requestJson[JsObject](request))
reqResult shouldBe Left(NotFound)
reqResult.left.get.reason shouldBe s"${NotFound.reason} (details: Limit is too large, must not exceed 268435456)"
```

You can potentially pull the fixed string into a `val body = "Limit is too large, must not exceed 268435456"` and use it in both cases (HttpResponse building and test check) accordingly.

steffenrost: Agreed, now it should be tested correctly.

PR #4407, opened 2019-03-28
Use static metric names when tag support is enabled

### LINKED ISSUE #3917: Provide option to support static metric names
OpenWhisk currently use dynamic metric names for some of the cases. The dynamic case has 2 variants

1. Finite Bounded options - Some of the metric names have finite states like `openwhisk.counter.invoker_docker.<cmd>_start`
2. Non bounded options - Metric names like those used for Kafka stats `openwhisk.histogram.kafka_invoker0_count` use instance id for the name. As these id can be many the metric names can be many and setup dependent.

Its possible to aggregate such metrics via use of regex or wildcard. However some systems like [Datadog does not support regex][1] which makes it tricky to analyze such metrics. Suggested way is to use Tags which is supported by #3343.

As a solution may be we expose another option for disabling use of dynamic names in unbounded cases and instead use tags.

[1]: https://docs.datadoghq.com/graphing/faq/when-i-query-can-i-use-wildcards-in-metric-names-and-events/

chetanmeh:
As discussed in #3917 in some of the cases we generate dynamic metric names where the possible values are not bounded. For such cases it would be better to capture such dynamic part as tags such that metric system which does not support regex over metric names can work on tags

Fixes #3917 

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3917)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -352,7 +352,10 @@ object LoggingMarkers {
   /*
    * Controller related markers
    */
-  def CONTROLLER_STARTUP(id: String) = LogMarkerToken(controller, s"startup$id", counter)(MeasurementUnit.none)
+  def CONTROLLER_STARTUP(id: String) =
+    if (TransactionId.metricsKamonTags)
+      LogMarkerToken(controller, s"startup", counter, None, Map("controller_id" -> id))(MeasurementUnit.none)
+    else LogMarkerToken(controller, s"startup$id", counter)(MeasurementUnit.none)
 
   // Time of the activation in controller until it is delivered to Kafka
   val CONTROLLER_ACTIVATION =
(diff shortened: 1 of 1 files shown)


common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
```diff
@@ -376,18 +379,38 @@ object LoggingMarkers {
   /*
    * Invoker related markers
    */
-  def INVOKER_STARTUP(i: Int) = LogMarkerToken(invoker, s"startup$i", counter)(MeasurementUnit.none)
+  def INVOKER_STARTUP(i: Int) =
+    if (TransactionId.metricsKamonTags)
+      LogMarkerToken(invoker, s"startup", counter, None, Map("invoker_id" -> i.toString))(MeasurementUnit.none)
```
selfxp: 👍 This would allow grouping by `invoker` in Datadog

rabbah:
Thanks Cosmin.

@SugandhaAgrawal do you have any comments before I merge?

SugandhaAgrawal:
@rabbah - I would need a day to check on this until then *please do not merge*.

SugandhaAgrawal:
@rabbah Thanks for the wait. Please feel free to **merge**. I am good 👋

Issue #4418, opened 2019-04-01
Container was't killed when creating timeout


ysjjovo:
Create container would get timeout,such as mount point to a wrong NFS path,pod will stay in 
'ContainerCreating' state forever.Invoker wait util ready for 1 min by default,
then emit a Exception with Message 'Failed to create pod $name',but LOST pod always stay in k8s cluster.

PR #4424, opened 2019-04-02
Delete pod when creating timeout

ysjjovo:
## Description
Pod creating timeout for some reason,such as mount a nonexistent nfs volume,
both k8s and invoker does't delete the 'ContainerCreating' pod.

## Related issue and scope
- [x] I opened an issue to propose and discuss this change (#4418)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/kubernetes/KubernetesClient.scala
+++ b/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/kubernetes/KubernetesClient.scala
@@ -19,47 +19,37 @@ package org.apache.openwhisk.core.containerpool.kubernetes
 
 import java.io.IOException
 import java.net.SocketTimeoutException
-import java.time.{Instant, ZoneId}
 import java.time.format.DateTimeFormatterBuilder
+import java.time.{Instant, ZoneId}
 
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.Uri
-import akka.http.scaladsl.model.Uri.Path
-import akka.http.scaladsl.model.Uri.Query
-import akka.stream.{Attributes, Outlet, SourceShape}
-import akka.stream.ActorMaterializer
+import akka.http.scaladsl.model.Uri.{Path, Query}
 import akka.stream.scaladsl.Source
 import akka.stream.stage._
+import akka.stream.{ActorMaterializer, Attributes, Outlet, SourceShape}
 import akka.util.ByteString
 import io.fabric8.kubernetes.api.model._
-import pureconfig.loadConfigOrThrow
-import org.apache.openwhisk.common.Logging
-import org.apache.openwhisk.common.TransactionId
-import org.apache.openwhisk.core.ConfigKeys
-import org.apache.openwhisk.core.containerpool.ContainerId
-import org.apache.openwhisk.core.containerpool.ContainerAddress
-import org.apache.openwhisk.core.containerpool.docker.ProcessRunner
-import org.apache.openwhisk.core.entity.ByteSize
-import org.apache.openwhisk.core.entity.size._
-
-import scala.concurrent.duration._
-import scala.concurrent.ExecutionContext
-import scala.concurrent.Future
-import scala.concurrent.blocking
-import scala.util.Failure
-import scala.util.Success
-import scala.util.Try
-import spray.json._
-import spray.json.DefaultJsonProtocol._
-import collection.JavaConverters._
-import io.fabric8.kubernetes.client.ConfigBuilder
-import io.fabric8.kubernetes.client.DefaultKubernetesClient
+import io.fabric8.kubernetes.client.{ConfigBuilder, DefaultKubernetesClient}
 import okhttp3.{Call, Callback, Request, Response}
 import okio.BufferedSource
+import org.apache.openwhisk.common.{Logging, TransactionId}
+import org.apache.openwhisk.core.ConfigKeys
+import org.apache.openwhisk.core.containerpool.docker.ProcessRunner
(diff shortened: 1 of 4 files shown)


core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -170,6 +170,15 @@ class KubernetesClient(
...
       case e =>
+        Future {
+          kubeRestClient
+            .inNamespace(kubeRestClient.getNamespace)
+            .pods()
+            .withName(name)
+            .delete()
+        }.recover {
+          case ex => log.error(this, s"Failed delete pod for '$name': ${ex.getClass} - ${ex.getMessage}")
+        }
```
markusthoemmes: This is not quite how it works. First of all, you'll want to wait for this future to complete before you finish the wrapping Future of the `run`. Otherwise, this will be some random operation, running somewhere in the background uncontrollably.

In this case, the sequence you'll want is as follows:

1. Try to create the pod.
2. If that failed, delete it. Ignore any errors here since we don't know if there's something to delete even.
3. regardless of whether the delete failed or not, return the error from the pod creation.

In this case, you can pull that off like so:

```scala
Future { 
  blocking { // Try to create a pod }
}.andThen { // Once that returns, log if it was an error but pass the original future through
  case Failure(e) => log.error(this, s"Failed create pod for '$name': ${e.getClass} - ${e.getMessage}")
}.recoverWith { // Do something on the error and flatten with the newly created Future
  case createError =>
    Future { 
      blocking { // Try to delete the pod } 
    }.andThen { // If that failed, log. Pass through the original Future (the one from the deleting the pod)
        case Failure(ex) => log.error(this, s"Failed delete pod for '$name': ${ex.getClass} - ${ex.getMessage}")
      }
      .transformWith { _ => // Whether or not pod deletion failed, flatten its Future with a failed Future, which also fails the recoverWith again, returning a failed Future overall
        Future.failed(new Exception(s"Failed to create pod '$name'"))
      }
}
```

This code might not compile as is, but it should get the intent across. Does that make sense?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -170,6 +170,15 @@ class KubernetesClient(
       }
     }.recoverWith {
       case e =>
+        Future {
+          kubeRestClient
```
markusthoemmes: Please wrap this in a `blocking` block if the `kubeRestClient`'s actions are still all blocking. It will instruct the underlying Thread pool to create enough threads to deal with potentially blocked ones gracefully.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/kubernetes/KubernetesClient.scala
silvasong: should you consider creating a failure for other reasons, such as a network timeout, then whether the deletion makes sense?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/kubernetes/KubernetesClient.scala
silvasong: > Please wrap this in a `blocking` block if the `kubeRestClient`'s actions are still all blocking. It will instruct the underlying Thread pool to create enough threads to deal with potentially blocked ones gracefully.
should you consider creating a failure for other reasons, such as a network timeout, then whether the deletion makes sense?

silvasong: should you consider creating a failure for other reasons, such as a network timeout, then whether the delete makes sense?

silvasong: > Please wrap this in a `blocking` block if the `kubeRestClient`'s actions are still all blocking. It will instruct the underlying Thread pool to create enough threads to deal with potentially blocked ones gracefully.

should you consider creating a failure for other reasons, such as network timeouts, whether delete still makes sense

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/kubernetes/KubernetesClient.scala
ysjjovo: I'am scala newbie,You consideration is right,thanks for your advice!I'll do that as you suggested!

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/kubernetes/KubernetesClient.scala
```diff
@@ -168,11 +158,26 @@ class KubernetesClient(
...
+              kubeRestClient
+                .inNamespace(kubeRestClient.getNamespace)
+                .pods()
+                .withName(name)
+                .delete()
+            }
+          }.andThen {
+              case Failure(e) => log.error(this, s"Failed delete pod for '$name': ${e.getClass} - ${e.getMessage}")
+            }
+            .transformWith { _ =>
```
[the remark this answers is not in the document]
ysjjovo: Nice suggestion!,But I can't fix alignment due to Intellij will auto start with a new line with '.transformWith'.

Issue #4336, opened 2019-03-13
Update docker version


style95:
Recently CVE of docker engine is reported.
 https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-5736

Since it has been a long time for OpenWhisk to use `docker-1.12`, I think it would be a good time to update docker version to the latest one not only to deal with the CVE but also to take advantage of the latest docker.
The current latest docker version is `18.09.3`.

For easier maintenance and better interoperability, it would be great to update both the server and the client.
(ex: https://github.com/apache/incubator-openwhisk/blob/master/core/invoker/Dockerfile#L8)

Since it might induce some fundamental changes in docker API, it can be a breaking change.

style95:
closing as #4430 is merged

PR #4430, opened 2019-04-05
Update docker client version to 18.06.3

style95:
## Description
With this change, it is required to install Docker=18.06.3 on target servers.
This is a breaking change.

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#4336)

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [x] Invoker
- [x] Tests
- [x] Deployment
- [ ] CLI
- [ ] General tooling
- [x] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/ansible/environments/jenkins/group_vars/openwhisk-vm1-he-de
+++ b/ansible/environments/jenkins/group_vars/openwhisk-vm1-he-de
@@ -36,3 +36,4 @@ runtimes_enable_concurrency: true
 limit_action_concurrency_max: 500
 
 invoker1_machine: openwhisk-vm3-he-de
+invoker_use_runc: false
--- a/ansible/roles/invoker/tasks/clean.yml
+++ b/ansible/roles/invoker/tasks/clean.yml
@@ -21,9 +21,9 @@
 - name: pause/resume at runc-level to restore docker consistency
   shell: |
         DOCKER_PAUSED=$(docker ps --filter status=paused --filter name=wsk{{ invoker_index }} -q --no-trunc)
-        for C in $DOCKER_PAUSED; do docker-runc pause $C; done
+        for C in $DOCKER_PAUSED; do docker-runc --root {{ invoker.docker.runcdir }} pause $C; done
         DOCKER_RUNNING=$(docker ps --filter status=running --filter name=wsk{{ invoker_index }} -q --no-trunc)
-        for C2 in $DOCKER_RUNNING; do docker-runc resume $C2; done
+        for C2 in $DOCKER_RUNNING; do docker-runc --root {{ invoker.docker.runcdir }} resume $C2; done
         TOTAL=$(($(echo $DOCKER_PAUSED | wc -w)+$(echo $DOCKER_RUNNING | wc -w)))
         echo "Handled $TOTAL remaining actions."
   register: runc_output
--- a/core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
+++ b/core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
@@ -98,11 +98,11 @@ class InvokerReactive(
           "--pids-limit" -> Set("1024")) ++ logsProvider.containerParameters)
   containerFactory.init()
 
-  CoordinatedShutdown(actorSystem).addTask(CoordinatedShutdown.PhaseBeforeActorSystemTerminate, "cleanup runtime containers") {
-    () =>
+  CoordinatedShutdown(actorSystem)
+    .addTask(CoordinatedShutdown.PhaseBeforeActorSystemTerminate, "cleanup runtime containers") { () =>
       containerFactory.cleanup()
       Future.successful(Done)
-  }
+    }
 
   /** Initialize needed databases */
   private val entityStore = WhiskEntityStore.datastore()
--- a/core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
+++ b/core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
@@ -20,7 +20,8 @@ package org.apache.openwhisk.core.invoker
 import java.nio.charset.StandardCharsets
 import java.time.Instant
 
-import akka.actor.{ActorRefFactory, ActorSystem, Props}
+import akka.Done
+import akka.actor.{ActorRefFactory, ActorSystem, CoordinatedShutdown, Props}
 import akka.event.Logging.InfoLevel
 import akka.stream.ActorMaterializer
 import org.apache.kafka.common.errors.RecordTooLargeException
@@ -96,7 +97,12 @@ class InvokerReactive(
           "--ulimit" -> Set("nofile=1024:1024"),
           "--pids-limit" -> Set("1024")) ++ logsProvider.containerParameters)
   containerFactory.init()
-  sys.addShutdownHook(containerFactory.cleanup())
+
+  CoordinatedShutdown(actorSystem).addTask(CoordinatedShutdown.PhaseBeforeActorSystemTerminate, "cleanup runtime containers") {
+    () =>
+      containerFactory.cleanup()
+      Future.successful(Done)
+  }
 
   /** Initialize needed databases */
   private val entityStore = WhiskEntityStore.datastore()
--- a/tools/macos/README.md
+++ b/tools/macos/README.md
@@ -31,8 +31,9 @@ The following are required to build and deploy OpenWhisk from a Mac host:
 - [Scala 2.11](http://scala-lang.org/download/)
 - [Ansible 2.5.2](http://docs.ansible.com/ansible/intro_installation.html)
 
-**Tip1** Versions of Docker and Ansible are lower than the latest released versions, the versions used in OpenWhisk are pinned to have stability during continuous integration and deployment.<br>
-**Tip2** It is required to install Docker >= 18.06.2 because of this [CVE](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-5736)
+**Tips:**
+ 1. Versions of Docker and Ansible are lower than the latest released versions, the versions used in OpenWhisk are pinned to have stability during continuous integration and deployment.<br>
+ 2. It is required to install Docker >= 18.06.2 because of this [CVE](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-5736)
 
 
 [Homebrew](http://brew.sh/) is an easy way to install all of these and prepare your Mac to build and deploy OpenWhisk. The following shell command is provided for your convenience to install `brew` with [Cask](https://github.com/caskroom/homebrew-cask) and bootstraps these to complete the setup. Copy the entire section below and paste it into your terminal to run it.
--- a/tools/macos/README.md
+++ b/tools/macos/README.md
@@ -26,17 +26,18 @@ If you prefer to use Docker-machine, you can follow instructions in [docker-mach
 
 The following are required to build and deploy OpenWhisk from a Mac host:
 
-- [Docker 1.12.0](https://docs.docker.com/docker-for-mac/)
+- [Docker 18.06.3](https://docs.docker.com/docker-for-mac/install/)
 - [Java 8](http://www.oracle.com/technetwork/java/javase/downloads/index.html) or [Open JDK 8](https://adoptopenjdk.net/releases.html#x64_mac)
 - [Scala 2.11](http://scala-lang.org/download/)
 - [Ansible 2.5.2](http://docs.ansible.com/ansible/intro_installation.html)
 
-**Tip** Versions of Docker and Ansible are lower than the latest released versions, the versions used in OpenWhisk are pinned to have stability during continuous integration and deployment.
+**Tip1** Versions of Docker and Ansible are lower than the latest released versions, the versions used in OpenWhisk are pinned to have stability during continuous integration and deployment.<br>
+**Tip2** It is required to install Docker >= 18.06.2 because of this [CVE](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-5736)
 
 
 [Homebrew](http://brew.sh/) is an easy way to install all of these and prepare your Mac to build and deploy OpenWhisk. The following shell command is provided for your convenience to install `brew` with [Cask](https://github.com/caskroom/homebrew-cask) and bootstraps these to complete the setup. Copy the entire section below and paste it into your terminal to run it.
 
-```
+```bash
 echo '
 # install homebrew
 /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
@@ -58,7 +59,7 @@ sudo -H pip install docker==2.2.1 ansible==2.5.2 jinja2==2.9.6 couchdb==1.1 http
 
 The above section of command installs Oracle JDK 8 as the default Java environment. If you would like to install Open JDK 8 instead of Oracle JDK 8, please run the following section:
 
-```
+```bash
 # install for finding alternative versions (Open JDK 8)
 brew tap AdoptOpenJDK/openjdk
 # install Open JDK 8
@@ -67,7 +68,7 @@ brew install adoptopenjdk-openjdk8
 
 instead of
 
-```
+```bash
 # install for finding alternative versions (java8)
 brew tap caskroom/versions
 # install java 8
@@ -79,7 +80,7 @@ with the shell command described above.
 No matter which JDK is used, make sure you correctly configure the environment variable $JAVA_HOME.
 
 # Build
-```
+```bash
 cd /your/path/to/openwhisk
 ./gradlew distDocker
 ```
@@ -92,7 +93,7 @@ Follow instructions in [ansible/README.md](../../ansible/README.md)
 Follow instructions in [Configure CLI](../../docs/cli.md)
 
 ### Use the wsk CLI
-```
+```bash
 bin/wsk action invoke /whisk.system/utils/echo -p message hello --result
 {
     "message": "hello"
--- a/tools/macos/docker-machine/README.md
+++ b/tools/macos/docker-machine/README.md
@@ -27,17 +27,17 @@ You will make provision of a virtual machine with Docker-machine and communicate
 The following are required to build and deploy OpenWhisk from a Mac host:
 
 - [Oracle VM VirtualBox](https://www.virtualbox.org/wiki/Downloads)
-- [Docker 1.12.0](https://docs.docker.com/engine/installation/mac/) (including `docker-machine`)
+- [Docker 18.06.3](https://docs.docker.com/machine/install-machine/) (including `docker-machine`)
 - [Java 8](http://www.oracle.com/technetwork/java/javase/downloads/index.html)
 - [Scala 2.11](http://scala-lang.org/download/)
 - [Ansible 2.5.2](http://docs.ansible.com/ansible/intro_installation.html)
 
-**Tip** Versions of Docker and Ansible are lower than the latest released versions, the versions used in OpenWhisk are pinned to have stability during continues integration and deployment.
-
+**Tip1** Versions of Docker and Ansible are lower than the latest released versions, the versions used in OpenWhisk are pinned to have stability during continues integration and deployment.
+**Tip2** It is required to install Docker >= 18.06.2 because of this [CVE](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-5736)
 
 [Homebrew](http://brew.sh/) is an easy way to install all of these and prepare your Mac to build and deploy OpenWhisk. The following shell command is provided for your convenience to install `brew` with [Cask](https://github.com/caskroom/homebrew-cask) and bootstraps these to complete the setup. Copy the entire section below and paste it into your terminal to run it.
 
-```
+```bash
 echo '
 # install homebrew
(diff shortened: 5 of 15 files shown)


style95:
Todo:
- [ ] Update controller docker cli version
- [ ] performance comparison
- [ ] Update documents accordingly
- [ ] Update Travis build configuration accordingly

style95:
In the latest docker, `runc` does not show any running containers which are created by `docker`.
So container is not paused by `runc` and just removed upon error(container not found).

so I am looking into the way to utilize `runc`.

style95:
The issue was previously container information was stored in `/run/runc` but now it is stored in `/run/docker/runtime-runc/moby`.

After changing mounting directory, it seems working as expected.

style95:
I have discussed this issue with @sven-lange-last.
Let me share the results here.

1. Generally, it takes some times to upgrade docker version for many reasons.
So it would require some time for downstream to use the latest docker version in their production.

2. But at the same time CVE should be properly handled.
This CVE, https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-5736 requires docker >=18.06.2

So if we upgrade the docker version to 18.06.x rather than 18.09, it would give more time to downstream at the same time resolve the CVE.

style95:
I did benchmarking to comparison any performance issue in docker-18.06.3.
It seems there is no big performance difference.

* 17.05: 12,500 TPS
* 18.06: 12,800 TPS

For more information, you can refer to this.
https://cwiki.apache.org/confluence/display/OPENWHISK/Docker+version+comparison

I think it would be good to go with `docker-18.06.3`

style95:
I readded WIP in the title.
Actually there are still few things to handle: https://github.com/apache/incubator-openwhisk/pull/4430#issuecomment-480224431

markusthoemmes:
> So it would require some time for downstream to use the latest docker version in their production.

We should really try to decouple from this. Sure, downstream will always be slower than upstream but we should not need to hold back on innovation and advancement because of that fact.

Is there any possibility for us to have different versions of `DockerClient` that flatten out the differences and make it a deployment/configuration concern which one is used? We could have a deprecation period for older clients to clean up after some time.

style95:
I think we need to differentiate the build step of the invoker Dockerfile to achieve that.
It is mainly related to the docker version inside invoker.

Regarding the downstream version, I will defer to @sven-lange-last as he wanted to make sure this PR does not break IBM production system. 

As the binary name of runc is changed from `docker-runc` to `runc` in docker-18.09, if we decide to upgrade to 18.09, that should be addressed as well.

rabbah:
I like the suggestion from Markus. Did I understand correctly that the conditioning is relatively easier for ansible than the dockerfile?

markusthoemmes:
The Dockerfile only ships docker binaries, right? How much overhead would it be to have multiple versions inside of the Dockerfile that we simply pick through configuration?

style95:
@markusthoemmes 
With the assumption that we will move forward with 18.09 and just including 18.06 for the deprecation period, docker-18.06 would add about 43MB(36 + 7.2 MB) of overhead.

```bash
$ ll -h docker-18.06
total 141M
drwxrwxr-x 2 style95 style95 4.0K Feb 20 11:26 ./
drwxr-xr-x 4 style95 style95 4.0K Apr 15 19:03 ../
-rwxr-xr-x 1 style95 style95  36M Feb 20 11:26 docker*
-rwxr-xr-x 1 style95 style95  26M Feb 20 11:26 docker-containerd*
-rwxr-xr-x 1 style95 style95  15M Feb 20 11:26 docker-containerd-ctr*
-rwxr-xr-x 1 style95 style95 4.0M Feb 20 11:26 docker-containerd-shim*
-rwxr-xr-x 1 style95 style95  51M Feb 20 11:26 dockerd*
-rwxr-xr-x 1 style95 style95 747K Feb 20 11:26 docker-init*
-rwxr-xr-x 1 style95 style95 2.8M Feb 20 11:26 docker-proxy*
-rwxr-xr-x 1 style95 style95 7.2M Feb 20 11:26 docker-runc*
```

@rabbah 
Let me clarify the issues.
There are the following changes.

### Docker 1.12 -> 18.06 or higher
`runc` no more use default(`/run/runc`) directory as a root directory.
It uses `/run/docker/runtime-runc/moby`.

### Docker 18.06 -> 18.09
The name of `runc` is changed from `docker-runc` to `runc`.


IMHO, there are few options to have multiple versions of docker, but I think it would be easier to configure the version at the deployment time.

For example, we can have both versions of docker binaries in the invoker.
It will have two binaries, `docker-runc` and just `runc`.
Based on the configuration, `DockerClient(RuncClient)` will selectively look for one of them.
(Currently, it is hardcoded to `/usr/bin/docker-runc`)

### Keeping docker-1.12
Besides of the above things, If we want to keep docker-1.12, we need to differentiate the ansible script based on the version flag as well.
The mounting directory for `runc` will vary between `/run/runc` and `/run/docker/runtime-runc/moby`.

markusthoemmes:
I see, thanks for the writeup. It looks like moving to 18.06 introduces the least friction for now and makes all parties kinda happy?

If so, it's reasonable to make that move first and think about unblocking future upgrade later.

style95:
I feel the results of system tests are not deterministic.

style95:
There are still some tests which are dependent on docker remote API.
Since we don't need that API in the core components anymore, I am trying to remove them entirely.
It could be an unnecessary security burden.

dgrove-oss:
fwiw, because the dns related command line options are different between docker versions, we added `clientVersion` to the DockerClient a couple of months ago. 
https://github.com/apache/incubator-openwhisk/blob/master/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerClient.scala#L114

style95:
@dgrove-oss Thank you for letting me know.
I am about to overlook it.
I will look into it as well.

style95:
It seems from the Docker 18, it relies on kernel cgroup capability to limit memory swapness.
Only after enabling it, I could successfully run memory exceeding test case in `ActionLimitTests`.

Reference: https://docs.docker.com/install/linux/linux-postinstall/#your-kernel-does-not-support-cgroup-swap-limit-capabilities

style95:
@dgrove-oss Regarding that DNS option, it seems it differentiates DNS option for docker=1.12.
I think it would be better to drop it as Openwhisk will require `docker >= 18.06.2` with this change.
How do you think?

I haven't tested but I think OW will still work with docker=1.12 without runc.
So if the option is required in some downstream, I think we can keep it for some time.

dgrove-oss:
I would suggest leaving the DNS option for 1.12 unchanged for this PR and then we do a separate PR to remove just it.  Will make it  easier in case some downstream needs to dis-entangle.  I agree we should remove the 1.12 specific option soon.

style95:
Since tests have never been passed in Jenkins CI, I am trying to get access to the Jenkins VM.
I will figure out the reason why ConcurrencyTests keep failing.

style95:
I finally got the reason.
Jenkin machine is using Docker-17.05 so it uses `/run/runc` as a root directory.
But this PR will use `/run/docker/runc-runtime/moby` as it is used in Docker-18.06.
So `ConcurrencyTests` are failing because it could not find the container.

```bash
root@openwhisk-vm2-he-de ~ # docker inspect invoker0 | grep runc
                "/run/docker/runtime-runc/moby:/run/runc:rw"
            "Runtime": "runc",
                "Source": "/run/docker/runtime-runc/moby",
                "Destination": "/run/runc",
                "/run/runc": {},
root@openwhisk-vm2-he-de ~ # ls /run/runc
6f9e7c4dbce0fa0a73b2c6ff159e9255a8b4c88c9f600224df1190efacfa1966  e79e510637d3776530043a1b7626ef06818a329031557120fcaabfe4a2b15363
ac8b30ad8be761461b461c85f31d314d747c22bfba6d406d5992d5563806ac4c
```

RuncClient could not find the container.
```
[2019-04-30T08:20:45.180Z] [ERROR] [#tid_sid_invokerNanny] [RuncClient] info: command was unsuccessful, code: 1 (unsuccessful), stdout: , stderr: container "c3aa7acdcb15252a6088025c3102c13aa8e74d816c122b59e6e27cc674501a24" does not exist [marker:invoker_runc.pause_error:1287762:22]
```

dgrove-oss:
Discussed on the openwhisk tech exchange.  The proposed plan is to merge this PR accepting that it will break Jenkins. Then update the docker version on the Jenkins VMs.

rabbah:
Just noting that this doesn't work on macos (which is ok because useRunC is disabled on the mac).

style95:
I have few more things to add.
- [ ] Permission issue in `invoker/clean.yml`
- [ ] CoordinatedShutdown for `ContainerFactoroy.cleanup`

ansible/roles/invoker/tasks/clean.yml
```diff
@@ -21,9 +21,9 @@
 - name: pause/resume at runc-level to restore docker consistency
   shell: |
         DOCKER_PAUSED=$(docker ps --filter status=paused --filter name=wsk{{ invoker_index }} -q --no-trunc)
-        for C in $DOCKER_PAUSED; do docker-runc pause $C; done
+        for C in $DOCKER_PAUSED; do docker-runc --root {{ invoker.docker.runcdir }} pause $C; done
```
style95: This comes from newer version Docker.
In the newer version of runc, it respects an environment variable, `XDG_RUNTIME_DIR`.
(Reference: https://github.com/opencontainers/runc/blob/master/main.go#L67)

Since that environment variable differs based on the current user, non-root users cannot unpause or resume containers with the previous command.

```
style95@ubuntu:~/git/openwhisk/ansible$ docker-runc --help | grep "root dir"
   --root value        root directory for storage of container state (this should be located in tmpfs) (default: "/run/user/1000/runc")
style95@ubuntu:~/git/openwhisk/ansible$ sudo docker-runc --help | grep "root dir"
   --root value        root directory for storage of container state (this should be located in tmpfs) (default: "/run/runc")
```
(Even, the location of root dir in docker 18.06 is `/run/docker/runtime-runc/moby`.)

So I explicitly configured this value.

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
```diff
@@ -96,7 +97,12 @@ class InvokerReactive(
           "--ulimit" -> Set("nofile=1024:1024"),
           "--pids-limit" -> Set("1024")) ++ logsProvider.containerParameters)
   containerFactory.init()
-  sys.addShutdownHook(containerFactory.cleanup())
+
+  CoordinatedShutdown(actorSystem).addTask(CoordinatedShutdown.PhaseBeforeActorSystemTerminate, "cleanup runtime containers") {
```
style95: This also newly comes with the recent version of Docker.
I am not quite sure the reason yet, but this `cleanup` procedure is terminated by `SIGTERM`.
(As you know, `docker stop` sends `SIGTERM` to the PID1 in the container, wait for 10s by default and send `SIGKILL` if the process does not stop)

In the previous version, it seems this procedure was not terminated by `SIGTERM`, but now it is.
So I changed this to `CoordinatedShutdown`.

Since the previous shutdown hook does not guarantee the task run, I think it would provide a better option for us.

>def addShutdownHook(body: ⇒ Unit): ShutdownHookThread
...
>
>Note that shutdown hooks are NOT guaranteed to be run.
(Reference: https://www.scala-lang.org/api/current/scala/sys/index.html#addShutdownHook(body:=%3EUnit):scala.sys.ShutdownHookThread)

One thing to note is, it does not respect the result from `cleanup()` method as it returns nothing(`Unit`).
But I think it provides the same level of guarantee with the previous version, so it would be fine.

style95:
I think it would be better to drop documentation for Docker-machine in a subsequent PR.

style95:
I think it's now ready to merge.
@sven-lange-last Could you confirm if we can merge this?

style95:
@rabbah @markusthoemmes @dgrove-oss 
Please look into this.
It's ready to merge.

sven-lange-last:
@style95 thanks a lot for providing this PR - very thorough work. Particularly, the performance tests are helpful for deciding whether the update to a newer Docker version is a feasible.

Some thoughts around the transient data directory `docker-runc` is using for containers:
* When Docker 18.06 uses `docker-runc`, directory `/run/docker/runtime-runc/moby` is used as base directory.
* `docker-runc` itself has different default directories that are used when not specifying the `--root` option. These are `/run/runc` for root (uid 0) and `/run/user/<uid>/runc` for non-root users.
* Apparently, all uses of `docker-runc` need to make sure that they use the same data directory as Docker - otherwise `docker-runc` won't be able to work on the containers created by Docker before.
* In `ansible/roles/invoker/tasks/clean.yml`, you explicitly use the `--root` option to make sure the proper directory is used.
* You have no changes in the Scala `RuncClient` used by the invoker.
https://github.com/apache/incubator-openwhisk/blob/be1e3a19c02956c9be85023a0bb0ff399c21444d/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/RuncClient.scala#L56-L64
* The Scala `RuncClient` works because you always mount the (configurable) transient data directory location to `/run/runc` in file `ansible/roles/invoker/tasks/deploy.yml` and `docker-runc` is run by root user.
* What happens if the invoker process is run by a non-root user? Won't `docker-runc` invoked by Scala `RuncClient` then fall back to the default transient data directory, i.e. `/run/user/<uid>/runc`?
* I do not know the details of lean setup - but what happens in other setup variants? Do we always make sure that the proper transient data directory location is available at `/run/runc` when running Scala `RuncClient`? Which user runs the process?
* Can we make things more robust by also using the `--root /run/runc` option in Scala `RuncClient`?

sven-lange-last:
@style95 concerning the failing Jenkins test:

> Jenkin machine is using Docker-17.05 so it uses /run/runc as a root directory.
> But this PR will use /run/docker/runc-runtime/moby as it is used in Docker-18.06.
> So ConcurrencyTests are failing because it could not find the container.

Wouldn't it be possible to work around this problem by setting `invoker_runcdir: "/run/runc"` in the [Ansible environment for Jenkins](https://github.com/apache/incubator-openwhisk/tree/master/ansible/environments/jenkins)?

tests/src/test/scala/ha/ShootComponentsTests.scala
```diff
@@ -305,9 +305,8 @@ trait ShootComponentUtils {
     def file(path: String) = Try(new File(path)).filter(_.exists).map(_.getAbsolutePath).toOption
 
     val docker = (file("/usr/bin/docker") orElse file("/usr/local/bin/docker")).getOrElse("docker")
-    val dockerPort = WhiskProperties.getProperty(WhiskConfig.dockerPort)
 
-    Seq(docker, "--host", host + ":" + dockerPort, cmd, component)
+    Seq(docker, cmd, component)
```
sven-lange-last: How is the test supposed to work if there is more than one controller running and the controllers run on different systems?

The original test code determines which host is running the first controller and would ask the remote host's Docker daemon to start, stop and restart the controller container. With the change, the test permanently fails in our test environment that uses multiple systems in parallel.

style95:
@sven-lange-last 
>Wouldn't it be possible to work around this problem by setting invoker_runcdir: "/run/runc" in the Ansible environment for Jenkins?

Thank you for the suggestion, I will try it.

tests/src/test/scala/ha/ShootComponentsTests.scala
style95: I just want to completely disable remote Docker API from the code as we don't use it anymore in our normal use cases.
It seems remote docker API is only required for test cases.

We can selectively enable/disable remote API based on the deployment type such as 
 production or unit tests.
But it depends on the OW operator, there is still a possibility that Docker remote API is enabled without consciousness. And it could induce a security breach.

How about making this test rely on `ssh` command instead?
Then what we need is just to make ssh connection available between target hosts and machine which runs tests.
(Generally, tests are executed in build machines such as jenkins, this requirement would be highly likely already met.)

sven-lange-last: I actually opt for keeping the current approach with running Docker commands against remote systems. From a security perspective, it's a matter of how you run your Docker daemon. In test environments, you can open a TCP listening socket to accept connections from remote systems - and on production systems, you can just disable the TCP listening socket. The rest is how you configure openwhisk.

Today, the `DockerClientWithFileAccess()` is created with an empty `host` parameter anyway such that the local Docker daemon BSD stream socket is used.

style95: Make sense.
I just wanted to clean up this configuration as well: https://github.com/apache/incubator-openwhisk/pull/4430/files#diff-85b6ff9e51a5005c641ac4690b975b2cL55
But that's also a viable option.
I will work on it.

style95:
@sven-lange-last It seems introducing `invoker_runcdir` with Docker 17.05 does not work.
This is because two different versions of runc is not compatible to each other.

```
[2019-05-09T10:04:50.544Z] [ERROR] [#tid_sid_invokerNanny] [RuncClient] info: command was unsuccessful, code: 1 (unsuccessful), stdout: , stderr: json: cannot unmarshal string into Go struct field State.init_process_start of type uint64 [marker:invoker_runc.pause_error:1322355:73]
```

> Yeah, that's because you have two runc versions that are sharing incompatible state. That's not really a supported configuration (though we should be more friendly about not bailing when an error occurs).

https://github.com/opencontainers/runc/issues/1592#issuecomment-328669901

style95:
Let me share the discussion with @sven-lange-last on Slack for history.

He suggested using `invoker_use_runc: false` in Jenkins CI environments and I applied it.

We discussed to add `--root` option in `RuncClient` as well, and it seems it's not required as of now because invoker is running as a root user.
So `RuncClient` client will always have access to `/run/runc` dir.
Since it can be broken at any time if the invoker runs as a non-root user, he also suggested leaving comments about its implications and it is also applied accordingly.

Now let's see what happens in Jenkins CI.

sven-lange-last:
@style95 thanks a lot for the valuable discussion and your recent updates. Jenkins is working now.

We still have two different problems in Travis:
* The code scanner complains about whitespace.
* `services.KafkaConnectorTests > Kafka connector should send and receive a kafka message even after shutdown one of instances` fails because a remote Docker daemon cannot be connected.

The problem with `services.KafkaConnectorTests` is in the last line of this code snippet:

https://github.com/style95/openwhisk/blob/faf765cd9d3fedd3139298a19f911247f127c91a/tests/src/test/scala/services/KafkaConnectorTests.scala#L146-L152

sven-lange-last:
PG 4 / 2840 running...

style95:
Finally!
This is ready to merge! 😄

ansible/environments/jenkins/group_vars/openwhisk-vm1-he-de
```diff
@@ -36,3 +36,4 @@ runtimes_enable_concurrency: true
 limit_action_concurrency_max: 500
 
 invoker1_machine: openwhisk-vm3-he-de
+invoker_use_runc: false
```
sven-lange-last: @houshengbo do you have plans to update Jenkins VMs to a higher Docker version once this PR is merged? This PR will update Docker to a higher version on the CLI side so that we needed to switch off runc usage in the invoker...

style95:
@sven-lange-last Got it.
I would open a confirmation thread (based on the silent assent) in dev list.

tools/macos/README.md
```diff
@@ -26,17 +26,18 @@ If you prefer to use Docker-machine, you can follow instructions in [docker-mach
...
 The following are required to build and deploy OpenWhisk from a Mac host:
 
-- [Docker 1.12.0](https://docs.docker.com/docker-for-mac/)
+- [Docker 18.06.3](https://docs.docker.com/docker-for-mac/install/)
 - [Java 8](http://www.oracle.com/technetwork/java/javase/downloads/index.html) or [Open JDK 8](https://adoptopenjdk.net/releases.html#x64_mac)
 - [Scala 2.11](http://scala-lang.org/download/)
 - [Ansible 2.5.2](http://docs.ansible.com/ansible/intro_installation.html)
 
-**Tip** Versions of Docker and Ansible are lower than the latest released versions, the versions used in OpenWhisk are pinned to have stability during continuous integration and deployment.
+**Tip1** Versions of Docker and Ansible are lower than the latest released versions, the versions used in OpenWhisk are pinned to have stability during continuous integration and deployment.<br>
```
rabbah: do you mind making this:

**Tips:**

  1. ...
  2. ...

tools/macos/docker-machine/README.md
```diff
@@ -27,17 +27,17 @@ You will make provision of a virtual machine with Docker-machine and communicate
...
 - [Oracle VM VirtualBox](https://www.virtualbox.org/wiki/Downloads)
-- [Docker 1.12.0](https://docs.docker.com/engine/installation/mac/) (including `docker-machine`)
+- [Docker 18.06.3](https://docs.docker.com/machine/install-machine/) (including `docker-machine`)
 - [Java 8](http://www.oracle.com/technetwork/java/javase/downloads/index.html)
 - [Scala 2.11](http://scala-lang.org/download/)
 - [Ansible 2.5.2](http://docs.ansible.com/ansible/intro_installation.html)
 
-**Tip** Versions of Docker and Ansible are lower than the latest released versions, the versions used in OpenWhisk are pinned to have stability during continues integration and deployment.
-
+**Tip1** Versions of Docker and Ansible are lower than the latest released versions, the versions used in OpenWhisk are pinned to have stability during continues integration and deployment.
```
rabbah: same as above.

tools/macos/README.md
style95: Updated accordingly

tools/macos/docker-machine/README.md
style95: Updated accordingly

PR #4450, opened 2019-04-17
Fix test cases to also work with nodejs:10 as default nodejs runtime.

falkzoll:
Fix the automated test cases to also work with nodejs:10 as default nodejs runtime.
These changes are to prepare for the upcoming nodejs:6 deprecation.

## Description
When switching to nodejs:10 as the default nodejs runtime, several testcases fail. Reason for this is, that the test cases assume/expect, nodejs:6 is used to execute the actions. With the deprecation of nodejs:6 it will be necessary that the tests will work with a newer nodejs version.
The testcases have now been modified to work with nodejs:6 and with nodejs:10 as the default runtime. Nodejs:8 may also work, but has not been tested.

This PR does not change the default nodejs runtime. It is just to prepare for switching to nodejs:10 as the default runtime.

In most cases the test now explicitly execute on nodejs:10 and also check, that they were executed in nodejs:10.
Some testcases have been modified so that they run on nodejs:6 and nodejs:10.
No tests have been removed, means there is no change in test coverage.

The tests have been executed with nodejs:6 and with nodejs:10 as the default runtime (runtime.json). In both cases the modified tests succeed.

## Related issue and scope

- [ ] see https://github.com/apache/incubator-openwhisk/issues/4265

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/tests/dat/actions/openFiles.js
+++ b/tests/dat/actions/openFiles.js
@@ -22,7 +22,7 @@ function main(params) {
     console.log("opened files = ", openFiles.length);
 
     openFiles.forEach(function(fh) {
-        fs.close(fh);
+        fs.close(fh, (err) => {} );
     })
 
     if (error === undefined) {
--- a/tests/src/test/scala/invokerShoot/ShootInvokerTests.scala
+++ b/tests/src/test/scala/invokerShoot/ShootInvokerTests.scala
@@ -172,6 +172,7 @@ class ShootInvokerTests extends TestHelpers with WskTestHelpers with JsHelpers w
 
   it should "add new parameters and annotations while copying an action" in withAssetCleaner(wskprops) {
     (wp, assetHelper) =>
+      val runtime = "nodejs:default"
       val origName = "origAction"
       val copiedName = "copiedAction"
       val origParams = Map("origParam1" -> "origParamValue1".toJson, "origParam2" -> 999.toJson)
@@ -188,13 +189,12 @@ class ShootInvokerTests extends TestHelpers with WskTestHelpers with JsHelpers w
         JsObject("key" -> JsString("copiedAnnot2"), "value" -> JsBoolean(false)),
         JsObject("key" -> JsString("copiedAnnot1"), "value" -> JsString("copiedAnnotValue1")),
         JsObject("key" -> JsString("origAnnot2"), "value" -> JsBoolean(true)),
-        JsObject("key" -> JsString("exec"), "value" -> JsString("nodejs:6")),
         JsObject("key" -> WhiskAction.provideApiKeyAnnotationName.toJson, "value" -> JsBoolean(false)))
 
       assetHelper.withCleaner(wsk.action, origName) {
         val file = Some(TestUtils.getTestActionFilename("echo.js"))
         (action, _) =>
-          action.create(origName, file, parameters = origParams, annotations = origAnnots)
+          action.create(origName, file, parameters = origParams, annotations = origAnnots, kind = Some(runtime))
       }
 
       assetHelper.withCleaner(wsk.action, copiedName) { (action, _) =>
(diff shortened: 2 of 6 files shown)


falkzoll:
Travis showed some errors in formatting. Will fix them and update this PR.

tests/src/test/scala/invokerShoot/ShootInvokerTests.scala
```diff
@@ -172,6 +172,7 @@ class ShootInvokerTests extends TestHelpers with WskTestHelpers with JsHelpers w
 
   it should "add new parameters and annotations while copying an action" in withAssetCleaner(wskprops) {
     (wp, assetHelper) =>
+      val runtime = "nodejs:10"
```
rabbah: just make this `nodejs:default` and we'll change the default in the runtimes.json.
otherwise we'll be doing this again when node 10 goes out of style.

falkzoll: Right, that would be the best 👍. 
Unfortunately many tests explicitly check the returned runtime string, e.g. for 'nodejs:6'. When we use 'nodejs:default' at action creation, this returned runtime string would be whatever the actual default runtime is.
In this case we would need to either query somehow in the test case what the current default runtime is, which we then can use to compare in the test case.
Or, we could check against a list of possible nodejs runtimes (nodejs:6, nodejs:8, nodejs:10). But then we would also need to adjust this list when we add newer versions.
Or, the other option would be, we do not compare the returned runtime string at all. Which would be the easiest solution,

What do you think?

rabbah: could those tests just check against the pattern `nodejs:*`?

rabbah: i'd think a more generic check is sufficient since there are unit tests that check that `:default` resolves to a specific version of a runtime.

falkzoll: yes, makes sense, I will look for a generic check like `nodejs:*` and see if we can modify the tests accordingly … the travis 'System Tests' also still show a failure I need to analyse and fix… I will get back to you when I am done with that :-)

tests/dat/actions/openFiles.js
```diff
@@ -22,7 +22,7 @@ function main(params) {
     console.log("opened files = ", openFiles.length);
 
     openFiles.forEach(function(fh) {
-        fs.close(fh);
+        fs.close(fh, (err) => {} );
```
markusthoemmes: Should this be `closeSync` instead? If not: Can we safely omit the callback function?

tests/src/test/scala/invokerShoot/ShootInvokerTests.scala
```diff
@@ -188,13 +189,13 @@ class ShootInvokerTests extends TestHelpers with WskTestHelpers with JsHelpers w
         JsObject("key" -> JsString("copiedAnnot2"), "value" -> JsBoolean(false)),
         JsObject("key" -> JsString("copiedAnnot1"), "value" -> JsString("copiedAnnotValue1")),
         JsObject("key" -> JsString("origAnnot2"), "value" -> JsBoolean(true)),
-        JsObject("key" -> JsString("exec"), "value" -> JsString("nodejs:6")),
+//        JsObject("key" -> JsString("exec"), "value" -> JsString(runtime)),
```
markusthoemmes: Please drop this line.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
```diff
@@ -66,7 +66,7 @@ class ActionLimitsTests extends TestHelpers with WskTestHelpers with WskActorSys
 
   val openFileAction = TestUtils.getTestActionFilename("openFiles.js")
   val openFileLimit = 1024
-  val minExpectedOpenFiles = openFileLimit - 15 // allow for already opened files in container
+  val minExpectedOpenFiles = openFileLimit - 20 // allow for already opened files in container
```
rabbah: is this necessary?

tests/dat/actions/openFiles.js
falkzoll: The empty callback was added since this parameter was optional for nodejs:6 but is now required for nodejs:10. Using an empty callback funtion sounded to be the default behaviour when this parameter is not present in nodejs:6. Explicitly defining the empty callback now, works with both nodejs:6 and nodejs:10.
You are right, making this a `closeSync ` call will most likely not change the test behaviour, since this is cleanup code at the end of the test. But I would need to try it first with both node versions.

tests/src/test/scala/invokerShoot/ShootInvokerTests.scala
falkzoll: Sure, forgot that line… will be removed.

tests/src/test/scala/org/apache/openwhisk/core/limits/ActionLimitsTests.scala
falkzoll: The nodejs:10 runtime has some more (+5) already open file handles which reduces the remaining number of open files by 5. The test is `successfully invoke an action when it is within nofile limit` therefore it seems not really necessary to always test for the exact maximum of files that can be opened. With the `-20` the test now runs fine with both versions of the nodejs runtimes.

falkzoll:
From the functional point of view, this PR is now ready for merge. All tests are successful and the testcases also work when nodejs:10 is the default nodejs runtime.

Most comments have been addressed.
Except of moving the check of the exec annotation into a subroutine. This is the check:
```
      .convertTo[Seq[JsObject]]
      .find(_.fields("key").convertTo[String] == "exec")
      .map(_.fields("value"))
      .map(exec => { exec.convertTo[String] should startWith("nodejs:") })
      .getOrElse(fail())
```
The checks (at 7 places) are in multiple classes so we would need something outside to them. I tried to find a good place for such a common subroutine for the tests, but I was not really successful. But I am not that deep in openwhisk, yet, and my Scala skill is not at expert level... yet :-). 
Therefore hints and tricks are highly welcome :-).

Depending on how important it is to merge this PR right now... I could also follow up on the move of this exec annotation check into a common subroutine in a separate PR?

rabbah:
Thanks @falkzoll for the patch and the proposed plan.

PR #4459, opened 2019-04-26
Close the consumer when WakeupExcpetion happened

ningyougang:
## Description

Kafka's consumer.wakeup() is to `shutdown consumer gracefully`.
In KafkaConsumerConnector.scala, has below points, exist `WakupException happend`
* peek(wakup() is called, then execute poll, WakupException happened)
* commit (Actually, when do this action, wakup() is called, WakupException happend)
* metric reporting scheduler (wakup() is called, then execute endOffsets, WakupException happened)

Regarding peek, handle its own way.
Regarding commit and metric reporting scheduler, it is better to close the consumer.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/connector/kafka/KafkaConsumerConnector.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/connector/kafka/KafkaConsumerConnector.scala
@@ -132,6 +132,9 @@ class KafkaConsumerConnector(
         } else {
           throw e
         }
+      case e: WakeupException =>
+        logging.info(this, s"WakeupException happened when do commit action for topic ${topic}")
+        recreateConsumer()
     }
 
   override def close(): Unit = synchronized {
@@ -195,6 +198,8 @@ class KafkaConsumerConnector(
         }
       }
     }.andThen {
+      case Failure(_: WakeupException) =>
+        recreateConsumer()
       case Failure(e) =>
         // Only log level info because failed metric reporting is not critical
         logging.info(this, s"lag metric reporting failed for topic '$topic': $e")


style95:
@ningyougang Thank you for the contribution.

Currently, `wakup()` is only called by this scheduler:
https://github.com/apache/incubator-openwhisk/pull/4459/files#diff-8bf7d87d98238d39ba7f36d2c52def6fR74

So is this to handle `WakeupException` when the scheduler calls `consumer.wakeup()` and unfortunately this exception is thrown while `commit` or `monitoringScheduler` are accessing the underlying consumer?

ningyougang:
Regarding ` this exception is thrown while commit or monitoringScheduler are accessing the underlying consumer?`
yes, right.

common/scala/src/main/scala/org/apache/openwhisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -132,6 +132,10 @@ class KafkaConsumerConnector(
         } else {
           throw e
         }
+      case e: WakeupException =>
+        logging.info(this, s"WakeupException happened when do commit action for topic ${topic}")
+        close()
```
style95: It seems this just leave the consumer closed and there is no logic to recreate a consumer.
Since it propagates the exception to `MessageFeed`, `MessageFeed` will keep failing.

I think we need to 1. catch the exception here and recreate the consumer or 2. throw the exception, catch it in `MessageFeed`, and recreate the consumer.
The same approach should be taken here as well: https://github.com/apache/incubator-openwhisk/pull/4459/files#diff-8bf7d87d98238d39ba7f36d2c52def6fR202

I am inclined to handle this exception in `KafkaConsumerConnector` layer because the same approach is already taken here: https://github.com/apache/incubator-openwhisk/pull/4459/files#diff-8bf7d87d98238d39ba7f36d2c52def6fR115

So in case of error, the consumer will be recreated without notifying `MessageFeed`.

ningyougang: In normal situation, when one consumer is closed, we need to create a new one.

But here, the user has a clear purpose: `close the consumer gracefully using consumer.wakeup` and have no purpose to create a new consumer again.

style95: `consumer.wakeup` is only used to prevent infinite hangs: https://github.com/apache/incubator-openwhisk/pull/4459/files#diff-8bf7d87d98238d39ba7f36d2c52def6fR73
Even there is no public method for `wakeup()` in `KafkaConsumerConnector`.

Is there any part that a user explicitly close this consumer using `wakeup()`?

ningyougang: Oh, in current github master branch, has no any part that user explicitly close this consumer using wakeup().

Maybe one day in future, for some special scene, may expose a public method `wakup()` in KafkaConsumerConnector.

I added this patch,  just for  `future expansion`.

style95: Then how about keeping the current behavior(recreating the consumer) and update it accordingly with the enhancement in the future?

ningyougang: Good idea!

common/scala/src/main/scala/org/apache/openwhisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -132,6 +132,11 @@ class KafkaConsumerConnector(
...
+        recreateConsumer()
```
style95: I think `recreateConsumer()` already includes `consumer.close()`?
https://github.com/apache/incubator-openwhisk/pull/4459/commits/4e9541a16cf26b67cc4e1e5d0a8fcafa0b599d74#diff-8bf7d87d98238d39ba7f36d2c52def6fL171

ningyougang: Oh, good catch. fixed

common/scala/src/main/scala/org/apache/openwhisk/connector/kafka/KafkaConsumerConnector.scala
```diff
@@ -132,6 +132,10 @@ class KafkaConsumerConnector(
         } else {
           throw e
         }
+      case e: WakeupException =>
+        logging.info(this, s"WakeupException happened when do commit action for topic ${topic}")
+        recreateConsumer()
+        throw e
```
style95: If we recreate the consumer do we need to throw the exception?

ningyougang: hm. have no need to throw the exception, because we already recreate the consumer.

common/scala/src/main/scala/org/apache/openwhisk/connector/kafka/KafkaConsumerConnector.scala
dubee: `wakeup()` is called by the `peak()` method above. Wouldn't recreating the consumer here result in issues when we call `wakeup()` from the `peak()` method? Ie, we'd recreate the consumer here while `peak()` is still trying to do work.

style95: @dubee 
Since we open neither `consumer` field nor `wakeup()` method, and currently `wakeup()` method is only called by `wakeupTask` scheduler, I think there would be no case that `WakeupException` thrown by `wakeupTask` is caught here.

In the future, if we expose `wakeup()` method, while `commit()` is called, `wakeup()` method can be called in a different thread, then this change would help to handle it.
Since all accesses to `consumer` are synchronized by `synchronized` keyword, it would be guaranteed that `wakeup()` would "wake up" only one thread at a time.

Issue #4476, opened 2019-05-12
how to remove concurrency limits


YangZhou1997:
Hi community, 

I am trying to run thousands of function instances on OW. However, even I follow this method (https://stackoverflow.com/questions/52210036/openwhisk-increase-number-of-concurrent-requests) to set -e limit_invocations_per_minute=999999 -e limit_invocations_concurrent=999999 when ansible openwhisk.yml, the concurrency limit seems still existing. I find there will still be at most 8 function instances invoked by one invoker concurrently, which corresponds to 500 concurrency invocation per minute.

Any suggestion to remove the concurrency limits?

style95:
@YangZhou1997 Could you share the spec of you invoker machines?
And how much time does it take to run your actions?

Generally, the number of concurrent actions are configured by invoker`s memory.

If you didn't change this value, the default would be 2GB which means 8 concurrent actions if you utilize the default container(256MB)
https://github.com/apache/incubator-openwhisk/blob/master/ansible/group_vars/all#L178

YangZhou1997:
I have four machines hosting four invokers, each with 160GB memory and 40 cores. 
It would take around 40s to run an action. 

Thanks for your answering; that exactly solves my question! I will change the memory setting.

YangZhou1997:
It seems that I cannot directly set the memory to 163840m, since the CPU-share would become lower than 2 and cause container creation crash. Please see the log from invoker: 

`
[2019-05-13T02:46:09.470Z] [INFO] [#tid_437421f4eeb8e1fc5465be9581c85978] [DockerClientWithFileAccess] running /usr/bin/docker run -d --cpu-shares 1 --memory 256m --memory-swap 256m --network openwhiskOverlay -e __OW_API_HOST=https://10.10.1.2:443 --name wsk33_40_whisksystem_fib --pids-limit 1024 --cap-drop NET_RAW --cap-drop NET_ADMIN --ulimit nofile=1024:1024 --log-driver json-file --env __OW_ALLOW_CONCURRENT=True yangzhou1997/python3action:mpi (timeout: 1 minute) [marker:invoker_docker.run_start:8473]
[...]

PR #4477, opened 2019-05-13
Limit the minimum to the cpu-share value

KeonHee:
## Description
The minimum value for `--cpu-shares` is 2.

However, if the `userMemory` is too high or the `reservedMemory` is too low, the result of the previous calculation might be 1. And this issue causes the following issues: https://github.com/apache/incubator-openwhisk/issues/4476#issuecomment-491664067

So I set the minimum value to 2.

## Related issue and scope
- [x] I opened an issue to propose and discuss this change (#4476)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
@@ -24,6 +24,7 @@ import org.apache.openwhisk.core.entity.{ByteSize, ExecManifest, InvokerInstance
 import org.apache.openwhisk.spi.Spi
 
 import scala.concurrent.Future
+import scala.math.max
 
 case class ContainerArgsConfig(network: String,
                                dnsServers: Seq[String] = Seq.empty,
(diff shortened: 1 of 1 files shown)


common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
```diff
@@ -44,7 +44,10 @@ case class ContainerPoolConfig(userMemory: ByteSize, concurrentPeekFactor: Doubl
    */
   private val totalShare = 1024.0 // This is a pre-defined value coming from docker and not our hard-coded value.
   // Grant more CPU to a container if it allocates more memory.
-  def cpuShare(reservedMemory: ByteSize) = (totalShare / (userMemory.toBytes / reservedMemory.toBytes)).toInt
+  def cpuShare(reservedMemory: ByteSize) = {
+    val cpuShare = (totalShare / (userMemory.toBytes / reservedMemory.toBytes)).toInt
```
style95: @KeonHee Thank you for the contribution.
How about using `max`?
Then we may be able to something like this:

```
val cpuShare = max((totalShare / (userMemory.toBytes / reservedMemory.toBytes)).toInt, 2)
```

KeonHee: @style95 Thanks for your review. I think that's better!

PR #4481, opened 2019-05-15
Track Kafka client side metrics via Kamon

chetanmeh:
Tracks Kafka client metrics via Kamon for monitoring

## Description

Currently Kafka metrics are not getting tracked via Kamon. Due to this we do not gain any insight into the Kafka interactions. Out of the box Kafka [tracks quite a few metrics][1] on client side these metrics are exposed via JMX

image

It also support custom [MetricsReporter][2] to listen to such metrics. This PR makes use of same reporter support to publish the metrics to Kamon (based on approach taken in [kamon-metrics-reporter][3])

### Usage

`KamonMetricsReporter` needs to be enabled via config and provided a set of metric names to track.

```
whisk {
  kafka {
    common {
      metric-reporters = "org.apache.openwhisk.connector.kafka.KamonMetricsReporter"
    }
    metrics {
      // Name of metrics which should be tracked via Kamon
      names = [
        // consumer-fetch-manager-metrics
        "records-lag-max", // The maximum lag in terms of number of records for any partition in this window
        "records-consumed-total" // The total number of records consumed
      ]

      report-interval = 10 seconds
    }
  }
}
```

Once enabled those metrics would be pushed to Kamon. For above config following metrics can be seen in Prometheus

```
# TYPE consumer_fetch_manager_metrics_records_consumed_total counter
consumer_fetch_manager_metrics_records_consumed_total{client_id="consumer-completed0"} 2.0
consumer_fetch_manager_metrics_records_consumed_total{client_id="consumer-cacheInvalidation"} 0.0
consumer_fetch_manager_metrics_records_consumed_total{client_id="consumer-health"} 1007.0
```

### Implementation

This PR takes a whitelist approach and does not publishes all metrics by default. As there are more than 300 metrics tracked by Kafka across Producer and Consumer.

For counters Kafka records two types of metrics `total` and `rate`. See [KIP-187][4] for details. So we should ignore metrics ending with `rate` and prefer `total` metrics for Kamon tracking

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

[1]: https://docs.confluent.io/current/kafka/monitoring.html
[2]: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/metrics/MetricsReporter.java
[3]: https://github.com/andreas-schroeder/kamon-kafka-reporter
[4]: https://cwiki.apache.org/confluence/display/KAFKA/KIP-187+-+Add+cumulative+count+metric+for+all+Kafka+rate+metrics

### DIFF
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -95,6 +95,10 @@ whisk {
         common {
             security-protocol = PLAINTEXT
             ssl-endpoint-identification-algorithm = "" // restores pre-kafka 2.0.0 default
+
+            //Enable this for reporting Kafka client metrics
+            //metric-reporters = "org.apache.openwhisk.connector.kafka.KamonMetricsReporter"
+
         }
         producer {
             acks = 1
@@ -153,6 +157,17 @@ whisk {
                 retention-bytes = 1073741824
                 retention-ms    = 3600000
             }
+        }
+
+        metrics {
+            // Name of metrics which should be tracked via Kamon
+            names = [
+                // consumer-fetch-manager-metrics
+                "records-lag-max", // The maximum lag in terms of number of records for any partition in this window
+                "records-consumed-total" // The total number of records consumed
+            ]
+
+            report-interval = 10 seconds
         }
     }
     # db related configuration
--- a/common/scala/src/main/scala/org/apache/openwhisk/connector/kafka/KafkaConsumerConnector.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/connector/kafka/KafkaConsumerConnector.scala
@@ -142,6 +142,7 @@ class KafkaConsumerConnector(
   /** Creates a new kafka consumer and subscribes to topic list if given. */
   private def createConsumer(topic: String) = {
     val config = Map(
+      ConsumerConfig.CLIENT_ID_CONFIG -> s"consumer-$topic",
       ConsumerConfig.GROUP_ID_CONFIG -> groupid,
       ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG -> kafkahost,
       ConsumerConfig.MAX_POLL_RECORDS_CONFIG -> maxPeek.toString) ++
--- a/common/scala/src/main/scala/org/apache/openwhisk/connector/kafka/KamonMetricsReporter.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/connector/kafka/KamonMetricsReporter.scala
@@ -111,7 +111,9 @@ object KamonMetricsReporter {
   }
 
   def kamonName(mn: MetricName): String = {
-    s"${mn.group()}_${mn.name()}"
+    //Drop the `-total` suffix as it results in prometheus metrics ending with total twice
+    val name = if (mn.name().endsWith("-total")) mn.name().dropRight(6) else mn.name()
+    s"${mn.group()}_$name}"
   }
 
   def isCounterMetric(metric: KafkaMetric): Boolean = Try(metric.measurable()) match {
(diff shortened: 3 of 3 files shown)


style95:
@chetanmeh IMHO, we can collect these metrics out of OpenWhisk.
Is there any reason to collect them via OpenWhisk?

Similarly, I am not sure we would collect some metrics in CouchDB via OpenWhisk even if we want to see some CouchDB metrics as well.

chetanmeh:
> IMHO, we can collect these metrics out of OpenWhisk

@style95 Any pointers on that? So far did not find a easy way to get handle to client side metrics in OpenWhisk. Only approach I saw so far was using some form of JMX Exporter which are tricky to setup. Hence using this approach

Note that this PR only enables client side metrics and not broker side ones

style95:
@chetanmeh 
I am using an in-house JMX metric collector to collect major Kafka metrics including client metrics. And it is a standalone java application which is comprised of 300 lines of codes.
I think it is one of the common and natural ways as we can find a similar approach in many places such as Prometheus.
https://github.com/prometheus/jmx_exporter

I meant no offense. I just want to discuss this with you.
I feel like we can take advantage of external components for this kind of case.
For example, we can delegate logs collection to logback or slf4j appender or standalone collection agents such as logstash.
I believe, the more we delegate such logic to external components, the more concise we can keep the core logic.

I have not looked into it deeply yet, but there is also a similar interesting approach.
https://github.com/Segence/kamon-jmx-collector

selfxp:
@chetanmeh I personally think that this extension is very useful. We already have integration with `Kamon` and `Kafka` has been a black box for any type of deployment for a long time. Collecting client side metrics is definitely a step in the right direction. 
But also collecting the `jmx` Kafka metrics would make a lot of sense also, and that could the "external" and "done in house" part that I think @style95 is referring to.

chetanmeh:
@style95 Agreed that JMX exporters can be used. However I find them tricky to setup and each monitoring system has its own way of exporting ([Datadog][1], [Prometheus][2] etc). In most cases it involves running another agent which needs a custom Docker build and for Prometheus results in having 2 http servers running exposing the metrics.

As OpenWhisk uses Kamon which abstract away all such integrations I was looking for a way to route such metrics to Kamon and then sent to various monitoring systems in a uniform way. 

Further note that proposed `KamonMetricsReporter` is optional (disabled by default) and would introduce a minimal overhead in terms of metrics collected and reported

[1]: https://docs.datadoghq.com/integrations/java/
[2]: https://github.com/prometheus/jmx_exporter

rabbah:
Can someone mention this pr on the dev list?

chetanmeh:
Did not got any feedback on [dev thread][1]. Would it be ok to merge this and have this feature disabled by default?

[1]: https://lists.apache.org/thread.html/ff900610c6cf35106d03d27ead17f4f4079817e35c7e60f2b3781914@%3Cdev.openwhisk.apache.org%3E

rabbah:
@chetanmeh LGTM - is this redundant though (when turned on) with the other kafka metric that's emitted by the scheduled actor?

chetanmeh:
> is this redundant though (when turned on) with the other kafka metric that's emitted by the scheduled actor?

This exposes more metrics in addition to lag. So kind of complements what we have in scheduled actor. Going forward we may even think of removing the logic in scheduled actor and make use of approach like done [here][1]

```scala
val server = ManagementFactory.getPlatformMBeanServer
val name = new ObjectName(s"kafka.consumer:type=consumer-fetch-manager-metrics,client-id=$id")
def consumerLag: Long = server.getAttribute(name, "records-lag-max").asInstanceOf[Double].toLong.max(0)
```

[1]: https://github.com/adobe-apiplatform/openwhisk-user-events/blob/master/src/main/scala/com/adobe/api/platform/runtime/metrics/EventConsumer.scala#L126-L133

rabbah:
i like it 👍

PR #4484, opened 2019-05-17
Switch to consistent indexing policy for CosmosDB

chetanmeh:
`activations` collection currently uses `lazy` indexing policy. Per MicrosoftDocs/azure-docs#23875 this is deprecated and should be avoided.

> Lazy mode performs updates to the index at a much lower priority level when the engine is not doing any other work. Because of this query performance can be highly unpredictable. It can also result in higher cost queries because the query engine has to do more work because it can't answer the query from the index. For these reasons we recommend against customers using it.

This PR switches to consistent indexing policy. 

### Disable automatic index policy update logic

Further it also removes logic to automatically update the indexing policy if the collection already exists. Instead it would now log a warning log

```
[2019-05-17T14:47:36.102Z] [WARN] Indexing policy for collection [activations] found to be different.
Expected - {"automatic":true,"indexingMode":"Lazy","includedPaths":[{"path":"/namespace/?","indexes":[{"kind":"Hash","dataType":"String","precision":-1}]},{"path":"/_c/nspath/?","indexes":[{"kind":"Hash","dataType":"String","precision":-1}]},{"path":"/start/?","indexes":[{"kind":"Range","dataType":"Number","precision":-1}]}],"excludedPaths":[{"path":"/"}]}
Existing - {"automatic":true,"indexingMode":"Lazy","includedPaths":[{"path":"/namespace/?","indexes":[]},{"path":"/_c/nspath/?","indexes":[]},{"path":"/start/?","indexes":[]}],"excludedPaths":[{"path":"/"},{"path":"/\"_etag\"/?"}]}
```

Any index update would then need to be done manually by the system administrator

### Dropping Hash index

Now the index docs do not mention of `Hash` index and per [this comment](https://github.com/Azure/azure-cosmosdb-java/issues/93#issuecomment-491417946) those are removed.

Also see [obsolete attributes](https://docs.microsoft.com/en-us/azure/cosmos-db/index-policy#obsolete-attributes)

Due to this now Indexing policy does not use them and always sets the index kind to Range. Now index policy comparison check just check if expected paths are there in included and excluded section

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/database/cosmosdb/CosmosDBSupport.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/database/cosmosdb/CosmosDBSupport.scala
@@ -17,13 +17,14 @@
 
 package org.apache.openwhisk.core.database.cosmosdb
 
+import akka.event.slf4j.SLF4JLogging
 import com.microsoft.azure.cosmosdb._
 import com.microsoft.azure.cosmosdb.rx.AsyncDocumentClient
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable
 
-private[cosmosdb] trait CosmosDBSupport extends RxObservableImplicits with CosmosDBUtil {
+private[cosmosdb] trait CosmosDBSupport extends RxObservableImplicits with CosmosDBUtil with SLF4JLogging {
   protected def config: CosmosDBConfig
   protected def collName: String
   protected def client: AsyncDocumentClient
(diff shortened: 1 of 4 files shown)


common/scala/src/main/scala/org/apache/openwhisk/core/database/cosmosdb/CosmosDBSupport.scala
```diff
@@ -17,13 +17,14 @@
...
 
+import akka.event.slf4j.SLF4JLogging
 import com.microsoft.azure.cosmosdb._
 import com.microsoft.azure.cosmosdb.rx.AsyncDocumentClient
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable
 
-private[cosmosdb] trait CosmosDBSupport extends RxObservableImplicits with CosmosDBUtil {
+private[cosmosdb] trait CosmosDBSupport extends RxObservableImplicits with CosmosDBUtil with SLF4JLogging {
```
markusthoemmes: Why a different logging methodology?

chetanmeh: This is a one time initialization flow which so far was not passing the `Logging` implicit. So thought its fine to use Slf4j directly. Can refactor to pass the `Logging`

ddragosd: Besides consistency, is there a concern for passing `implicit` vs using `SLF4JLogging` ?

PR #4494, opened 2019-05-27
Akka http server responds HEAD request as GET

SugandhaAgrawal:
Akka http client by default treats any http HEAD request as GET unless the configuration sets it otherwise. There are use cases where user expects HEAD to be responded as HEAD and without a response body unlike GET.

https://github.com/akka/akka-http/issues/2088 mentions that the default is set to treat a HEAD as a GET request by the Akka client.

## Description
Update the controller configuration to process HEAD and GET requests as they are expected.

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/tests/dat/actions/echo-web-http-head.js
+++ b/tests/dat/actions/echo-web-http-head.js
@@ -1,0 +1,23 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+function main(params) {
+    return {
+        statusCode: 200,
+        headers: { 'Request-type': params.__ow_method }
+    };
+}
--- a/tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskWebActionsTests.scala
+++ b/tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskWebActionsTests.scala
@@ -309,17 +309,14 @@ class WskWebActionsTests extends TestHelpers with WskTestHelpers with RestUtil w
    * Tests web action for HEAD requests
    */
   it should "create a web action making a HEAD request" in withAssetCleaner(wskprops) { (wp, assetHelper) =>
-    val name = "webaction"
-    val file = Some(TestUtils.getTestActionFilename("echo.js"))
+    val name = "webactionHead"
+    val file = Some(TestUtils.getTestActionFilename("echo-web-http-head.js"))
     val host = getServiceURL()
-    val url = s"$host$testRoutePath/$namespace/default/$name.text/__ow_user"
+    val url = s"$host$testRoutePath/$namespace/default/$name"
 
     assetHelper.withCleaner(wsk.action, name) { (action, _) =>
-      action.create(name, file, web = Some("true"), annotations = Map("require-whisk-auth" -> true.toJson))
+      action.create(name, file, web = Some("true"))
     }
-
-    val unauthorizedResponse = RestAssured.given().config(sslconfig).get(url)
-    unauthorizedResponse.statusCode shouldBe 401
 
     val authorizedResponse = RestAssured
       .given()
@@ -331,6 +328,7 @@ class WskWebActionsTests extends TestHelpers with WskTestHelpers with RestUtil w
 
     authorizedResponse.statusCode shouldBe 200
     authorizedResponse.body.asString() shouldBe ""
+    authorizedResponse.getHeader("Request-type") shouldBe "head"
   }
 
   private val subdomainRegex = Seq.fill(WhiskProperties.getPartsInVanitySubdomain)("[a-zA-Z0-9]+").mkString("-")
--- a/tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskWebActionsTests.scala
+++ b/tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskWebActionsTests.scala
@@ -305,6 +305,34 @@ class WskWebActionsTests extends TestHelpers with WskTestHelpers with RestUtil w
       response.body.asByteArray shouldBe Base64.getDecoder().decode(png)
   }
 
+  /**
+   * Tests web action for HEAD requests.
+   */
+  it should "create a web action making a HEAD request" in withAssetCleaner(wskprops) { (wp, assetHelper) =>
+    val name = "webaction"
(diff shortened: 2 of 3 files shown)


chetanmeh:
@SugandhaAgrawal I believe this is Akka HTTP Server setting and not related to client. 

If we enable this would HEAD request still work in OpenWhisk or would fail. I presume so far they work just that internally executed as GET. However with this change would they stop working if route does not implement HEAD support?

markusthoemmes:
It'd be useful to get some context on the use-cases that trigger this issue.

mbehrendt:
the context of this issue is actually quite straightforward: We have seen users trying to implement a web action that responds differently on HEAD than on GET. However, given `__ow_method` always had the same value (`get`), they weren't able to make that distinction in their action code.

markusthoemmes:
Got it, that makes sense. I guess @chetanmeh was confused (so was I) because the initial description talks about akka-https client not behaving as expected.

@chetanmeh is also right though that this is a breaking change as currently working HEAD requests will now fail.

mbehrendt:
i wouldn't consider to be a breaking change per se, but rather a bug fix. It was just flatout wrong that when a `HEAD` request came in, `__ow_method` showed `GET`. Of course, if people relied on that bug to be true forever, they would break, but that is true for probably most other bug fixes as well :-)

SugandhaAgrawal:
@markusthoemmes I have the same thought as Michael's. I would say this is a bug fix. The breaking change in terms that the users need to change the implementation is fine but this is right way to handle the HEAD request.

SugandhaAgrawal:
@markusthoemmes The initial description does mean that akka http client isnt behaving as expected *by default*. It responds HEAD with a GET and that isnt the right response in my opinion. 
In order to achieve this you need to set explicitly the value to off. My expectation was that the default should have been off or do you not find this as an expected result and correct?

markusthoemmes:
@SugandhaAgrawal the point is that it's the akka http **server**, not the client.

I'm indecisive on the bug fix vs. breaking change. I agree though that the amount of breakage is very small and arguably relying on undocumented/wrong behavior.

If @chetanmeh doesn't have strong feelings against it I'm fine merging this as-is. Otherwise we should take it to the dev-list.

SugandhaAgrawal:
@markusthoemmes Thanks for the correction, server ;)

SugandhaAgrawal:
@chetanmeh I will do include test in the next PR

chetanmeh:
If this fix is not very urgent would it be possible to include the test in this PR itself. I believe you can still enable this flag in your setups via system property

rabbah:
+1 to include a test.

rabbah:
I agree this is a bug - the intended behavior is for the HEAD to reach the function just like OPTIONS. You'll need a _system_ level tests that "curls" the web action. I tested this patch locally and don't see a difference in behavior (a GET is posted with and without the patch). So +1 from me to include a test in this PR.

rabbah:
@chetanmeh I'm not sure if existing actions would fail --- at least not all. I suspect that you need actions which have already opted out of the default CORS response.

rabbah:
@SugandhaAgrawal any luck adding a test for this fix?

SugandhaAgrawal:
@rabbah Unfortunately not. I was on vacation and returned today. Will work on it

rabbah:
welcome back!

tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskWebActionsTests.scala
```diff
@@ -305,6 +305,34 @@ class WskWebActionsTests extends TestHelpers with WskTestHelpers with RestUtil w
...
   }
 
+  /**
+   * Tests web action for HEAD requests
+   */
+  it should "create a web action making a HEAD request" in withAssetCleaner(wskprops) { (wp, assetHelper) =>
+    val name = "webaction"
+    val file = Some(TestUtils.getTestActionFilename("echo.js"))
+    val host = getServiceURL()
+    val url = s"$host$testRoutePath/$namespace/default/$name.text/__ow_user"
```
rabbah: you should not need this here for a HEAD test: `.text/__ow_user`

tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskWebActionsTests.scala
```diff
@@ -305,6 +305,34 @@ class WskWebActionsTests extends TestHelpers with WskTestHelpers with RestUtil w
...
+    val authorizedResponse = RestAssured
+      .given()
+      .config(sslconfig)
+      .auth()
+      .preemptive()
+      .basic(wskprops.authKey.split(":")(0), wskprops.authKey.split(":")(1))
+      .head(url)
+
+    authorizedResponse.statusCode shouldBe 200
+    authorizedResponse.body.asString() shouldBe ""
```
rabbah: you'll want to test that the `__ow_method` received was HEAD not GET.

SugandhaAgrawal:
Screenshot 2019-06-27 at 09 45 52


Travis System Test

tests/dat/actions/echo-web-http-head.js
```diff
@@ -0,0 +1,9 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more contributor
+// license agreements; and to You under the Apache License, Version 2.0.
```
rabbah: Please use the long form license. We have to stop using the short form.

rabbah: it was a recent change.

thanks for the update!

tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskWebActionsTests.scala
```diff
@@ -305,6 +305,35 @@ class WskWebActionsTests extends TestHelpers with WskTestHelpers with RestUtil w
...
+   * Tests web action for HEAD requests
+   */
+  it should "create a web action making a HEAD request" in withAssetCleaner(wskprops) { (wp, assetHelper) =>
+    val name = "webactionHead"
+    val file = Some(TestUtils.getTestActionFilename("echo-web-http-head.js"))
+    val host = getServiceURL()
+    val url = s"$host$testRoutePath/$namespace/default/$name"
+
+    assetHelper.withCleaner(wsk.action, name) { (action, _) =>
+      action.create(name, file, web = Some("true"), annotations = Map("require-whisk-auth" -> true.toJson))
```
rabbah: i think you can skip the require-whisk-auth annotation here.

tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskWebActionsTests.scala
```diff
@@ -305,6 +305,35 @@ class WskWebActionsTests extends TestHelpers with WskTestHelpers with RestUtil w
...
+    }
+
+    val unauthorizedResponse = RestAssured.given().config(sslconfig).head(url)
+    unauthorizedResponse.statusCode shouldBe 401
```
rabbah: and then remove this unless this is important to the test.

SugandhaAgrawal: It is as a pattern also used for the other web action tests above. Hence, I included it. The behaviour is not explicit for head or get request so I can remove it

SugandhaAgrawal:
@rabbah Please merge this, if everything looks good!

rabbah:
Thanks @SugandhaAgrawal 👍

PR #4496, opened 2019-05-29
Settle throttle before running the API GW Tests

mhenke1:
## Description
The API GW Test are able to drive OW into throttling. Though some code existed to 
throttle down these tests these tests were failing constantly on faster machines
with 429 returned by OW.

The test code controlling the throttle was correct in throttling the test execution in isolation,
but was of course not able to overview invocations for the given namspace that do happen 
before tests were executing.

Since this limitations does naturally exist, this change simplifies down the throttling logic
to enhance code maintainability and introduces a wait period before executing the tests to ensure 
that the API GW test do not run into throttling driven by prior invokes in the same namespace.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/tests/src/test/scala/org/apache/openwhisk/core/cli/test/BaseApiGwTests.scala
+++ b/tests/src/test/scala/org/apache/openwhisk/core/cli/test/BaseApiGwTests.scala
@@ -89,8 +89,11 @@ abstract class BaseApiGwTests extends TestHelpers with WskTestHelpers with Befor
    * Create a CLI properties file for use by the tests
    */
   override def beforeAll() = {
-    // Check and settle the throttle so that this test have issues with previous tests
-    checkThrottle(maxInvocationsBeforeThrottle = 1, throttlePercent = 100)
+    //Wait a while to settle the throttle so that the status of throttle algorithm implemented in checkThrottle
+    //is consistent with the reality
+    //Without this waiting time API Tests do fail very consistently on faster build machines
+    //In the longer run the APIGW tests should be moved to a separate namespace
+    Thread.sleep(60.seconds.toMillis)
     cliWskPropsFile.deleteOnExit()
     val wskprops = WskProps(token = "SOME TOKEN")
     wskprops.writeFile(cliWskPropsFile)
--- a/tests/src/test/scala/org/apache/openwhisk/core/cli/test/BaseApiGwTests.scala
+++ b/tests/src/test/scala/org/apache/openwhisk/core/cli/test/BaseApiGwTests.scala
@@ -22,6 +22,7 @@ import java.time.Instant
 
 import scala.collection.mutable.ArrayBuffer
 import scala.concurrent.duration._
+import scala.math.max
 
 import org.junit.runner.RunWith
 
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/org/apache/openwhisk/core/cli/test/BaseApiGwTests.scala
```diff
@@ -56,29 +55,10 @@ abstract class BaseApiGwTests extends TestHelpers with WskTestHelpers with Befor
...
-        s"throttling ${throttlePercent}% of action invocations within last minute = ($numInvocationsToClear) invocations")
-      val throttleTime = 60.seconds.toMillis - (t.toEpochMilli - invocationToClear.toEpochMilli)
-
-      println(s"Waiting ${throttleTime} milliseconds to settle the throttle")
-      Thread.sleep(throttleTime)
-    }
-
-    invocationTimes += Instant.now
+  def throttle(maxInvocationsBeforeThrottle: Int = maxActionsPerMin) = {
+    val throttleTime = (60.seconds.toMillis / maxInvocationsBeforeThrottle * 1.1).toInt
```
markusthoemmes: Doesn't this lack a relation to `invocationsLast60Seconds` from above?

mhenke1: @markusthoemmes The original test code worked as expected in isolation. Buit is way to complicated with seeing the limitation that it only controls the test execution of the API GW tests in  and does not prevent all the error situation we have encountered due to prior invocation in the same namespace.

 In decided to dumb it down to throttle each invocation with a shorter fixed period to increase maintainability.

markusthoemmes: One of the main things I see us losing is time. I agree that functionality-wise this is fine and maybe even better. However, you lose the ability to use all the elapsed time between tests or when not firing any actions (for example when creating them) which the old code took into account.

mhenke1: @markusthoemmes Sorry for answering late. With the latest commit I reverted back to the original throttling code. I just left the additional pause to calm down the throttle before running the tests.

mhenke1:
@markusthoemmes I decided to make the introduced wait time (which is the only remaining change now) even more explicit while explaining its introduction in a comment. 
Since we get in throttling very often with our newer and faster build environment please have a expedited look.

rabbah:
why do these tests not check the API response which actually provides the reason (and counts) for a 409? the tests could parse that error response and settle the throttle as needed instead of these hacks.

rabbah:
i would not bother and implement a namespace change - it's just another hack.

the 409 response from the controller provides useful information to back off the test and retry correctly:

```
  def tooManyRequests(count: Int, allowed: Int) =
    s"Too many requests in the last minute (count: $count, allowed: $allowed)."

  /** Standard message for too many concurrent activation requests within a time window. */
  def tooManyConcurrentRequests(count: Int, allowed: Int) =
    s"Too many concurrent requests in flight (count: $count, allowed: $allowed)."
```

thread.sleep and using another namespace are just piling on technical debt imo.

PR #4503, opened 2019-06-03
Add optional config for appending custom registry to user provided images

duynguyen:
Currently, when `whisk.container-factory.runtimes-registry.url` is set to a specific registry, default runtime images in the manifest are pulled from that registry. Other images provided by users are left intact.
In a multi-region and multi-cloud deployment of OpenWhisk, it is useful for the system operators to force all runtime images (default and blackbox) to be pulled from specific registries.
This requires the registry hostname components be parsed from image names.

## Description

This PR:
- makes ImageName registry aware: differentiate `registry` and `prefix` in ImageName parser.
- adds config to automatically append optional registry hostname to user images, if they do not contain any hostname.

Usage:
* If `whisk.container-factory.runtimes-registry.url` is set, all default runtime images are retrieved from the specified registry, otherwise they are pulled from Docker Hub.
* If `whisk.container-factory.user-images-registry.url` is set, all user images (blackbox) are retrieved from the specified registry, otherwise they are pulled from Docker Hub.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [x] Tests
- [x] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


rabbah:
In your model all the docker actions are stored in a single registry?

duynguyen:
@rabbah it is more about setting the default registry of the system.
With `include-user-images`, we have the option to change the default system registry from Docker Hub (keeping the flag false) to a custom one (turning it to true). As mentioned in the description, it is useful for multi-region deployments (region specific registries), or multi-cloud (AWS ECR or Azure container registry, etc.), or trustable image source, to name a few.

For example:
```
whisk.container-factory.runtimes-registry.url=repo-1.openwhisk.org
```

As an owner of blackbox action, I can use the following images by passing to `--docker`:
1. `openwhisk/nodejs6action` -> invoker pulls `repo-1.openwhisk.org/openwhisk/nodejs6action`
2. `docker.io/openwhisk/nodejs6action` -> the default openwhisk runtime on Docker Hub is pulled
3. `repo-2.openwhisk.org/openwhisk/nodejs6action` -> image is pulled as it is named (from `repo-2`)

Default kind runtimes can still be pulled from different registries by specifying it in the manifest file. Something like:
```
"image": {
    "registry": "local-repo.example.com",
    "prefix": "openwhisk",
    "name": "somekind",
    "tag": "latest"
}
```

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
```diff
@@ -51,7 +51,9 @@ case class ContainerPoolConfig(userMemory: ByteSize, concurrentPeekFactor: Doubl
 
 case class RuntimesRegistryCredentials(user: String, password: String)
 
-case class RuntimesRegistryConfig(url: String, credentials: Option[RuntimesRegistryCredentials])
+case class RuntimesRegistryConfig(url: String,
+                                  includeUserImages: Option[Boolean] = None,
```
chetanmeh: Instea dof using an optional boolean it would be better to not define any value and let the default value (`false`) be defined in application.conf itself.

common/scala/src/main/scala/org/apache/openwhisk/core/mesos/MesosContainerFactory.scala
```diff
@@ -126,7 +126,7 @@ class MesosContainerFactory(config: WhiskConfig,
                                memory: ByteSize,
                                cpuShares: Int)(implicit config: WhiskConfig, logging: Logging): Future[Container] = {
     implicit val transid = tid
-    val image = if (userProvidedImage) {
+    val image = if (userProvidedImage && !runtimesRegistryConfig.includeUserImages.getOrElse(false)) {
```
chetanmeh: This pattern is repeating itself in 4 impls so far. May be we should move this logic to `ContainerFactory` object. Or introduce a new `RuntimesRegistry` object and use that 

```scala
object RuntimesRegistry {
  def resolveImage(actionImage: ExecManifest.ImageName,
                   userProvidedImage: Boolean,
                   config: RuntimesRegistryConfig): Either[ExecManifest.ImageName, String] = {
    if (userProvidedImage && !config.includeUserImages.getOrElse(false)) {
      Left(actionImage)
    } else {
      Right(actionImage.localImageName(config.url))
    }
  }
}
```

ansible/group_vars/all
```diff
@@ -42,6 +42,7 @@ whisk:
 # briefly the parameters are:
 #
 #   runtimes_registry: optional registry (with trailing slack) where to pull docker images from for runtimes and backbox images
+#   runtimes_registry_user_images: optional, this will add runtimes_registry to all images for pulling, blackbox images included
```
rabbah: this might be a separate registry in some cases i'd imagine - would it make sense either now or later to make this a string value (which may be the same as runtime_registry).

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
rabbah: note that this would be removed with a change above and you cons an instance of this class with a different url value.

duynguyen:
sorry I was out of office for a while, and catching up with this PR now. Will make new changes according to review comments.

duynguyen:
@rabbah , @chetanmeh : I have updated the PR according to the comments. Could you please review?

core/invoker/src/main/resources/application.conf
```diff
@@ -98,6 +98,7 @@ whisk {
     }
     runtimes-registry {
       url: ""
+      include-user-images: false # if true, registry hostname is automatically appended to user images which do not contain hostname
```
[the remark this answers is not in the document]
duynguyen: thanks for pointing out, removed the outdated boolean flag and added the string config.

tests/src/test/resources/application.conf.j2
```diff
@@ -86,6 +86,7 @@ whisk {
 
     container-factory.runtimes-registry {
         url = "{{ runtimes_registry | default('') }}"
+        include-user-images = "{{ runtimes_registry_user_images | default('false') }}"
```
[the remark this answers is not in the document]
duynguyen: thanks for pointing out, removed the outdated boolean flag and added the string config.

rabbah:
@duynguyen this test is failing:

```
> org.apache.openwhisk.core.containerpool.docker.test.DockerContainerFactoryTests > DockerContainerFactory should set the docker run args based on ContainerArgsConfig FAILED

org.apache.openwhisk.core.containerpool.docker.test.DockerContainerFactoryTests > DockerContainerFactory should set the docker run args based on ContainerArgsConfig FAILED
    org.scalatest.exceptions.TestFailedException: Unexpected call: <mock-1> DockerApiWithFileAccess.pull(openwhisk/action-nodejs-v10:nightly, #tid_sid_testing)

    Expected:
    inAnyOrder {
      <mock-1> DockerApiWithFileAccess.run(openwhisk/action-nodejs-v10:nightly, List(--cpu-shares, 32, --memory, 10m, --memory-swap, 10m, --network, net1, -e, __OW_API_HOST=, -e, k1=v1, -e, k2=v2, -e, k3=, --dns, dns1, --dns, dns2, --name, testContainer, --extra1, e1, --extra1, e2, --extra2, e3, --extra2, e4), *) once (never called - UNSATISFIED)
      <mock-1> DockerApiWithFileAccess.inspectIPAddress(ContainerId(fakecontainerid), net1, *) once (never called - UNSATISFIED)
      <mock-1> DockerApiWithFileAccess.rm(ContainerId(fakecontainerid), *) once (never called - UNSATISFIED)
      <mock-1> DockerApiWithFileAccess.clientVersion() once (called once)
    }
```

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
```diff
@@ -111,6 +111,20 @@ object ContainerFactory {
   /** include the instance name, if specified and strip invalid chars before attempting to use them in the container name */
   def containerNamePrefix(instanceId: InvokerInstanceId): String =
     s"wsk${instanceId.uniqueName.getOrElse("")}${instanceId.toInt}".filter(isAllowed)
+
+  def resolveRegisterConfig(userProvidedImage: Boolean,
+                            runtimesRegistryConfig: RuntimesRegistryConfig,
+                            userImagesRegistryConfig: RuntimesRegistryConfig): RuntimesRegistryConfig = {
```
sven-lange-last: Wouldn't it make sense to name this function `resolveRegistryConfig()`?

duynguyen: yes that makes sense. In fact that was a typo with `resolveRegisterConfig ` ...
thanks for pointing out.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainerFactory.scala
```diff
@@ -60,10 +62,14 @@ class DockerContainerFactory(instance: InvokerInstanceId,
                                userProvidedImage: Boolean,
                                memory: ByteSize,
                                cpuShares: Int)(implicit config: WhiskConfig, logging: Logging): Future[Container] = {
+    val registryConfig =
+      ContainerFactory.resolveRegisterConfig(userProvidedImage, runtimesRegistryConfig, userImagesRegistryConfig)
+    val image =
+      if (userProvidedImage && registryConfig.url.isEmpty) Left(actionImage)
+      else Right(actionImage.localImageName(registryConfig.url))
```
sven-lange-last: I have the expression that this code can be streamlined as follows to simplify the condition in line 68 (`(userProvidedImage && registryConfig.url.isEmpty)`):

```
   val registryConfig =
      ContainerFactory.resolveRegisterConfig(userProvidedImage, runtimesRegistryConfig, userImagesRegistryConfig)
    val imageRef = actionImage.localImageName(registryConfig.url)
    val image = if (userProvidedImage) Left(imageRef) else Right(imageRef)
```

My understanding is that `actionImage.localImageName()` would just leave out the registry part if no registry urls are specified.

sven-lange-last: Would this lead to the same result - as done in the Kube container factory below?

```
   val registryConfig =
      ContainerFactory.resolveRegisterConfig(userProvidedImage, runtimesRegistryConfig, userImagesRegistryConfig)
    val imageRef = ContainerFactory.resolveImage(actionImage, registryConfig)
    val image = if (userProvidedImage) Left(imageRef) else Right(imageRef)
```

duynguyen: I guess you meant it to be:
```
val registryConfig =
      ContainerFactory.resolveRegisterConfig(userProvidedImage, runtimesRegistryConfig, userImagesRegistryConfig)
val imageRef = actionImage.localImageName(registryConfig.url)
val image = if (userProvidedImage) Left(actionImage) else Right(imageRef)
```
(`actionImage` in Left, not `imageRef`)

However the condition is not the same as before anymore. In the case of `userProvidedImage==true`, the raw `imageName` object is passed into DockerContainer, which will pull the "public image name" (without custom blackbox registry), see [DockerContainer.scala#L105](https://github.com/apache/openwhisk/blob/master/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala#L105). Hence I added a condition for registryConfig being empty or not.

In general this flow does not fit well into the existing flow of DockerContainerFactory. I am working on an enhancement to get rid of `publicImageName`, make `localImageName` (to be renamed) a single image name resolver, and apply it throughout the code base. I will push that change for review.

sven-lange-last: My understanding is that the following behaviour is desired for blackbox (i.e. user provided) images identified by a user provided image reference (see https://github.com/docker/distribution/blob/master/reference/reference.go):

* If the image reference contains a domain, keep the image reference as is to pull from the repository in the image reference.
* If the image reference contains NO domain and NO global user image registry is configured (`registryConfig.url.isEmpty == true`), keep the image reference as is. It's up to the used container factory to convert to a canonical reference including a domain and path.
* If the image reference contains NO domain and a global user image registry is configured (`registryConfig.url.isEmpty == false`), add the global user image registry as domain to the image reference.

I think this is exactly the behaviour that following code implements:

```
   val registryConfig =
      ContainerFactory.resolveRegisterConfig(userProvidedImage, runtimesRegistryConfig, userImagesRegistryConfig)
    val imageRef = ContainerFactory.resolveImage(actionImage, registryConfig)
```

That's why I proposed this code because it already addresses the three cases listed above.

duynguyen: alright, now I got what you meant, thanks for explaining.
My intention with "If the image reference contains a domain", is that if there is a global user image registry, it would overwrite the image domain. I thought of this in a multi-cloud / multi-region setup in which a cluster for a specific cloud / region could force to download *all* images from its desired registry.
But I also agree with you that keeping the domain in image reference (if existed) no matter what global registry is, would be more trivial to users.

sven-lange-last: Yes, since your PR does not provide a list with mappings for registry domains but only a single user image repository domain, all user provided image refs would be overwritten - regardless whether the user provided image ref already contains a domain or not.

This current approach would lead to:

With a user registry config like `uk.icr.io`, this would lead to situations like:

* User registry config empty, user-provided image ref `image:latest` => Docker container factory pulls `docker.io/library/image:latest`.
* User registry config empty, user-provided image ref `docker.io/library/image:latest` => Docker container factory pulls `docker.io/library/image:latest`.
* User registry config `uk.icr.io`, user-provided image ref `image:latest` => Docker container factory pulls `uk.icr.io/image:latest`.
  The user would probably expect to pull `uk.icr.io/library/image:latest` instead - but `docker pull` only inserts the "official" repository path `library` if `docker.io` is specified as domain. See https://github.com/docker/distribution/blob/14b96e55d84c367ebab6caf9f0086de0876eec72/reference/normalize.go#L88-L105.
* User registry config `uk.icr.io`, user-provided image ref `path/image:latest` => Docker container factory pulls `uk.icr.io/path/image:latest`.
* User registry config `uk.icr.io`, user-provided image ref `docker.io/path/image:latest` => Docker container factory pulls `uk.icr.io/path/image:latest`.
  The user would certainly not expect this outcome. If you provide an image ref with a domain, you would not expect that it is replaced.

duynguyen: ok, I think what you said makes sense. thanks @sven-lange-last.
Please see the latest PR version.

ansible/group_vars/all
```diff
@@ -55,7 +55,8 @@ whisk:
 # configuration parameters related to support runtimes (see org.apache.openwhisk.core.entity.ExecManifest for schema of the manifest).
 # briefly the parameters are:
 #
-#   runtimes_registry: optional registry (with trailing slack) where to pull docker images from for runtimes and backbox images
+#   runtimes_registry: optional registry (with trailing slash) where to pull docker images from for default runtimes (in manifest)
+#   user_images_registry: optional registry (with trailing slash) where to pull docker images from for blackbox images
```
chetanmeh: Just to confirm wrt compatibility - If user was earlier configuring `runtimes_registry` which was used for "both" images from for runtimes and backbox images would now have to configure both `runtimes_registry` and `user_images_registry`. 

Or system would default to `runtimes_registry` if `user_images_registry` is not configured

duynguyen: They are separately evaluated.
Earlier there was only one option to add `runtimes_registry` to default images. Blackbox images were only evaluated as is (no system registry applied).
This PR adds `user_images_registry` to apply blackbox images. If this is not set, blackbox images are evaluated by its own name.

chetanmeh:
Thanks @duynguyen  for following up on this PR. Its merged now

duynguyen:
thanks @rabbah , @sven-lange-last and @chetanmeh for reviewing and great suggestions!

Issue #4502, opened 2019-06-03
remove default port from __OW_API_HOST


rabbah:
The __OW_API_HOST can contain the port number (e.g., :443). We should drop that when the port is the default for the protocol. 

Two related references:
- https://github.com/apache/incubator-openwhisk/blob/64cb8ecc60c158226415909a00958c758ca0ebe1/common/scala/src/main/scala/org/apache/openwhisk/core/WhiskConfig.scala#L57

- https://github.com/apache/incubator-openwhisk/blob/64cb8ecc60c158226415909a00958c758ca0ebe1/common/scala/src/main/scala/org/apache/openwhisk/core/WhiskConfig.scala#L156

PR #4504, opened 2019-06-03
Remove default port on api host.

rabbah:
Proposes fix for #4502.

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#4502)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/WhiskConfig.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/WhiskConfig.scala
@@ -19,19 +19,22 @@ package org.apache.openwhisk.core
 
 import java.io.File
 
-import scala.io.Source
+import akka.http.scaladsl.model.Uri.normalize
 import org.apache.openwhisk.common.{Config, Logging}
+
+import scala.io.Source
+import scala.util.Try
 
 /**
  * A set of properties which might be needed to run a whisk microservice implemented
  * in scala.
  *
  * @param requiredProperties a Map whose keys define properties that must be bound to
- * a value, and whose values are default values. A null value in the Map means there is
- * no default value specified, so it must appear in the properties file.
+ *                           a value, and whose values are default values. A null value in the Map means there is
+ *                           no default value specified, so it must appear in the properties file.
  * @param optionalProperties a set of optional properties (which may not be defined).
- * @param propertiesFile a File object, the whisk.properties file, which if given contains the property values.
- * @param env an optional environment to initialize from.
+ * @param propertiesFile     a File object, the whisk.properties file, which if given contains the property values.
+ * @param env                an optional environment to initialize from.
  */
 class WhiskConfig(requiredProperties: Map[String, String],
                   optionalProperties: Set[String] = Set.empty,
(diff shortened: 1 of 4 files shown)


common/scala/src/main/scala/org/apache/openwhisk/core/WhiskConfig.scala
```diff
@@ -54,8 +54,16 @@ class WhiskConfig(requiredProperties: Map[String, String],
   val dockerEndpoint = this(WhiskConfig.dockerEndpoint)
   val dockerPort = this(WhiskConfig.dockerPort)
 
-  val wskApiHost = this(WhiskConfig.wskApiProtocol) + "://" + this(WhiskConfig.wskApiHostname) + ":" + this(
-    WhiskConfig.wskApiPort)
+  val wskApiHost = {
```
chetanmeh: Can possibly be simplified by using `akka.http.scaladsl.model.Uri.normalize`

```suggestion
  val wskApiHost = Uri.normalize(s"${this(WhiskConfig.wskApiProtocol)}://${this(WhiskConfig.wskApiHostname)}:${this(WhiskConfig.wskApiPort)}")
```

rabbah: awesome - thanks for the tip... is there a java equivalent to replace the changes in whisk props below?

chetanmeh: Do not know of any java equivalent and hard to invoke the Scala code from java. So lets retain the current approach for test code

rabbah: the change causes many tests to fail - it appears in a number of unit tests, the value of the string is just `://:` - and I think to guard the values i might end up with more code than the original version so im inclined to just use the explicit removal of the port. wdyt.

chetanmeh: Yeah seems like values are empty which cause the parser to fail. May be an approach like below?

```
  val wskApiHost = Try(
    Uri.normalize(
      this(WhiskConfig.wskApiProtocol) + "://" + this(WhiskConfig.wskApiHostname) + ":" + this(WhiskConfig.wskApiPort)))
    .getOrElse("")
```

common/scala/src/main/scala/org/apache/openwhisk/core/WhiskConfig.scala
```diff
@@ -54,8 +54,12 @@ class WhiskConfig(requiredProperties: Map[String, String],
   val dockerEndpoint = this(WhiskConfig.dockerEndpoint)
   val dockerPort = this(WhiskConfig.dockerPort)
 
-  val wskApiHost = this(WhiskConfig.wskApiProtocol) + "://" + this(WhiskConfig.wskApiHostname) + ":" + this(
-    WhiskConfig.wskApiPort)
+  val wskApiHost: String = scala.util
+    .Try(
+      akka.http.scaladsl.model.Uri.normalize(
```
chetanmeh: Nit. Better to import the package

tests/src/test/scala/common/rest/WskRestOperations.scala
```diff
@@ -764,7 +764,7 @@ class RestActivationOperations(implicit val actorSystem: ActorSystem)
       activation.respBody
     } map {
       Right(_)
-    } getOrElse Left(s"Cannot find activation id from '$activation'")
+    } getOrElse Left(s"No activation record for'$activationId'")
```
rabbah: observed in #4506

PR #4513, opened 2019-06-14
Optional support for queued writes for CosmosDB

chetanmeh:
For activations in case of high volume we are seeing connection pool related error. This PR adds an optional support for having queued writes

## Description

In case of high volume of writes on Invoker we are seeing errors like 

```
[2019-06-11T00:37:10.196Z] [ERROR] Network failureio.reactivex.netty.client.PoolExhaustedException: null    
    at io.reactivex.netty.client.ConnectionPoolImpl.performAquire(ConnectionPoolImpl.java:177)  
    at io.reactivex.netty.client.ConnectionPoolImpl.access$300(ConnectionPoolImpl.java:45)  
    at io.reactivex.netty.client.ConnectionPoolImpl$1.call(ConnectionPoolImpl.java:139) 
    at io.reactivex.netty.client.ConnectionPoolImpl$1.call(ConnectionPoolImpl.java:124) 
    at rx.internal.operators.OnSubscribeLift.call(OnSubscribeLift.java:48)  
    at rx.internal.operators.OnSubscribeLift.call(OnSubscribeLift.java:30)  
```

This was observed with a pool size of 1000.  To prevent such cases it would be better to have a more controlled way of writing documents to db. Earlier for CouchDB similar thing was done via #2812 where batched writes were used. Currently CosmosDB does not provide an easy way to perform batched writes (Azure/azure-cosmosdb-java#182). 

### Design

To enable controlled writes this PR introduces a `QueuedExecutor` (similar in spirit to existing `Batcher` implementation) which ensures that writes are queued and then prcessed in a controlled way. Key points

* The queue size is tracked as a gauge metric
* Implementation ensures that upon close any existing entry in queue should get persisted
* Proper [backpressure is not possible][1]. So in case of very excessive writes entries would be dropped. Other option is to set a higher queue size but that can result in out of memory scenario. 

### Usage

```
  cosmosdb {
    collections {
      WhiskActivation {
        write-queue-config = {
          # Size of in memory queue. If queue gets full then put calls would be rejected
          queue-size = 100000

          # Number of concurrent connections to use to perform writes to db
          concurrency = 500
        }
      }
    }
  }
```

### Future Enhancement

We can possibly optimize the write throughput by implementing the bulk write logic as used in [CosmosDB Bulk Importer](https://chapsas.com/how-the-cosmosdb-bulk-executor-works-under-the-hood/). This would require us to batch the inserts and then sort by partition and send the calls to respective partition.

Another possible option to reduce heap pressure would be to store the object in byte array form in queue and deserialize before passing to `ArtifactStore`

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

[1]: https://github.com/apache/incubator-openwhisk/pull/2812#discussion_r142941162

### DIFF
--- a/common/scala/build.gradle
+++ b/common/scala/build.gradle
@@ -83,7 +83,7 @@ dependencies {
 
     compile 'io.reactivex:rxscala_2.12:0.26.5'
     compile 'io.reactivex:rxjava-reactive-streams:1.2.1'
-    compile ('com.microsoft.azure:azure-cosmosdb:2.4.2')
+    compile ('com.microsoft.azure:azure-cosmosdb:2.4.5')
 
     compile ('com.lightbend.akka:akka-stream-alpakka-s3_2.12:1.0.1') {
         exclude group: 'org.apache.httpcomponents' //Not used as alpakka uses akka-http
--- a/common/scala/src/main/resources/application.conf
+++ b/common/scala/src/main/resources/application.conf
@@ -281,6 +281,14 @@ whisk {
         #     }
         #   }
         # }
+
+        # Optional queued writes mode
+        # write-queue-config = {
+        #    # Size of in memory queue. If queue gets full then elements would be dropped upon addition
+        #    queue-size = 10000
+        #    # Number of concurrent connections to use to perform writes to db
+        #    concurrency = 500
+        # }
     }
 
     # transaction ID related configuration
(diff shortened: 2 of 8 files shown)


chetanmeh:
@markusthoemmes  @cbickel  Would be helpful if you can review the `QueuedExecutor` logic (which is not related to CosmosDB part). Its similar to `Batcher` implementation. Wanted to confirm if Akka Stream semantics are handled properly

chetanmeh:
> gauge I think is never passed to `QueuedExecutor` ctor 

Good catch. That confirms why queue size was coming as zero in the tests. Thats the drawback of using parameters with default values. Fix that now.

chetanmeh:
Closing it for now as we are now pursuing the Activation Persister Service approach #4632

PR #4529, opened 2019-06-30
switch to `nightly` docker image tag for runtimes

dgrove-oss:
Master branch of OpenWhisk should default to using
the most recent runtimes built from the master branch
of the corresponding git repos. This image is now
being published using the tag `nightly` instead of `latest`.

### DIFF
--- a/ansible/files/runtimes.json
+++ b/ansible/files/runtimes.json
@@ -7,7 +7,7 @@
                 "image": {
                     "prefix": "openwhisk",
                     "name": "nodejs6action",
-                    "tag": "latest"
+                    "tag": "nightly"
                 },
                 "deprecated": false,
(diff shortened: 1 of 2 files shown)


ansible/files/runtimes.json
```diff
@@ -71,7 +71,7 @@
                 "image": {
                     "prefix": "openwhisk",
                     "name": "python2action",
-                    "tag": "latest"
+                    "tag": "1.13.0-incubating"
```
dgrove-oss: Note:  build of python2 image was disabled back in January by https://github.com/apache/incubator-openwhisk-runtime-python/pull/48.  Could create a nightly tag, but probably clearer to use the release tag to indicate we stopped building this image.

PR #4535, opened 2019-07-02
Handle deleted limits document for get operation the same way as missing limits document

steffenrost:
This PR is a small consistency fix for the `wskadmin` tool. If the `wskadmin` tool is called to get the currently set limits on a given namespace, it should not print out `Failed to get limits ..` if the limits document is deleted on the database but handle it the same way as if the limits document does not exists (`No limits found, default system limits apply`). This is more consistent as the message shown by `wskadmin` should be the same for both cases. 
In addition, both cases (missing and deleted limits document) shouldn't be treated as error in my opinion and hence the `wskadmin` shouldn't return with a non-zero code, but this is open to discussion.

## Description
This code change changes the messages printed by the `wskadmin` tool for deleted namespace limits documents.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [x] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/tools/admin/wskadmin
+++ b/tools/admin/wskadmin
@@ -576,7 +576,7 @@ def getLimitsCmd(args, props):
             print('No limits found, default system limits apply')
         else:
             print('Failed to get limits (%s)' % res.read().strip())
-        return 1
+            return 1
 
 def deleteLimitsCmd(args, props):
     docId = quote_plus(args.namespace + "/limits")
--- a/tools/admin/wskadmin
+++ b/tools/admin/wskadmin
@@ -572,7 +572,7 @@ def getLimitsCmd(args, props):
                 print('%s = %s' % (limit, givenLimit))
     else:
         error = json.loads(res.read())
-        if error['reason'] == 'missing':
+        if error['reason'] == 'missing' or error['reason'] == 'deleted':
             print('No limits found, default system limits apply')
         else:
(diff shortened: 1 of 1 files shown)


steffenrost:
@rabbah `wskadmin-next` looks good, tested it for a deleted namespace limits document and the tool correctly print outs `No limits found, default system limits apply`.
Regarding the code returned by `wskadmin` in case of a missing or deleted namespace, would it be ok to change this from `1` to `0` to indicate no error?

rabbah:
makes sense to exit 0.

tools/admin/wskadmin
```diff
@@ -576,7 +576,7 @@ def getLimitsCmd(args, props):
             print('No limits found, default system limits apply')
         else:
             print('Failed to get limits (%s)' % res.read().strip())
-        return 1
+            return 1
```
rabbah: can you cover all three exits? return 0 at line 576, and leave return 1 on 579?

steffenrost: I did the change, anything else still missing?

steffenrost:
@rabbah thanks for merging the PR!

PR #4555, opened 2019-07-12
[WIP] [Scheduler] Add Etcd installation & modules

KeonHee:
## Description

This pr covers the [etcd](https://etcd.io/) installation and module implementation. And the etcd will be the coordinator of cluster members and the metadata store in the new scheduler.

We will implement modules based on the order defined in this page and add design documents under the page as well:
https://cwiki.apache.org/confluence/display/OPENWHISK/Component+Design

But I can not write wiki document because I do not have write access to the wiki :(
If you give me write access, I will write a document related to the etcd modules.

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [x] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -387,3 +387,26 @@ metrics:
     port: "{{ metrics_kamon_statsd_port | default('8125') }}"
 
 user_events: "{{ user_events_enabled | default(false) }}"
+
+etcd:
+  version: "{{ etcd_version | default('v3.3.12') }}"
+  client:
+    port: 2379
+  server:
+    port: 2480
+  cluster:
+    token: "{{ etcd_cluster_token | default('openwhisk-etcd-token') }}"
+  dir:
+    data: "{{ config_root_dir }}/etcd"
+  lease:
+    timeout: "{{ etcd_lease_timeout | default(1) }}"
+  quota_backend_bytes: "{{ etcd_quota_backend_bytes | default(0) }}"
+  snapshot_count: "{{ etcd_snapshot_count | default(100000) }}"
+  auto_compaction_retention: "{{ etcd_auto_compaction_retention | default(0) }}"
+  auto_compaction_mode: "{{ etcd_auto_compaction_mode | default('periodic') }}"
+
+etcd_connect_string: "{% set ret = [] %}\
+                      {% for host in groups['etcd'] %}\
+                        {{ ret.append( hostvars[host].ansible_host + ':' + ((etcd.client.port+loop.index-1)|string) ) }}\
+                      {% endfor %}\
+                      {{ ret | join(',') }}"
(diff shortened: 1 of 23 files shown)


mhenke1:
@KeonHee As discussed before, please put the new test code behind a feature toggle in a way that they are not executed as long the "new architecture" is not exploited (you could use the "assume" assertion of Scalatest based on a Ansible property).

mhenke1:
@KeonHee  @style95  Was the usage of another "distributed state store" beside Zookeeper discussed on the mailing list or somewhere else ? On a first glance this feels wrong to add more complexity into the system.

sven-lange-last:
My general comment on this PR is that it unconditionally adds system tests like `EtcdClientTests` that require a deployed etcd. In the dev mailing list discussion, a feature flag was requested for the architecture change: https://lists.apache.org/thread.html/63696c7bee9effb57118fe24513789d5538ea3a64af1204fccde73e8@%3Cdev.openwhisk.apache.org%3E.

By default, etcd should not be deployed and tests requiring etcd should not be run.

style95:
@mhenke1 
Thank you for the feedback.
I have wanted this kind of discussion so that I shared my proposal via the mailing list and bi-weekly meeting.
https://lists.apache.org/list.html?dev@openwhisk.apache.org:lte=1M:Dominic%20Kim
https://www.youtube.com/watch?v=cPQ_UOEK9d8

Regarding etcd and zookeeper, I don't want to list up all the differences between them here because there are already many articles and materials out there.

I have used and loved zookeeper for many years.
But IMHO etcd is more lightweight and easy to operate than zookeeper.

And IIRC, we are using zookeeper only for Kafka.
Since my proposal is based on the assumption that we will rule out the kafka from the critical path in the future, so I believe we can replace zookeeper with etcd at some point.

@sven-lange-last 
Yes, we will definitely add switches to toggle our implementation and all modules will be disabled by default. Most of the modules are written in new files and based on SPI, I think we can easily achieve it.
Thank you for the good suggestion.

ansible/group_vars/all
```diff
@@ -387,3 +387,26 @@ metrics:
...
+    port: 2480
+  cluster:
+    token: "{{ etcd_cluster_token | default('openwhisk-etcd-token') }}"
+  dir:
+    data: "{{ config_root_dir }}/etcd"
+  lease:
+    timeout: "{{ etcd_lease_timeout | default(1) }}"
+  quota_backend_bytes: "{{ etcd_quota_backend_bytes | default(0) }}"
+  snapshot_count: "{{ etcd_snapshot_count | default(100000) }}"
+  auto_compaction_retention: "{{ etcd_auto_compaction_retention | default(0) }}"
```
style95: `retetion` should not be 0?

KeonHee: `retention == 1` is more better. I will update it :)

KeonHee:
@sven-lange-last @mhenke1 
Thank you for the feedback :)

Following yours feedback, I will create a test toggle.

style95:
closing.
We will open a new one when the official procedure for proposals is defined.

Issue #4558, opened 2019-07-16
Allow parameters to be designated as init time properties


rabbah:
As discussed on the [dev list](https://lists.apache.org/thread.html/7e03e6fc965d972882f4c76e34b2e9aa28579a0c3004d63d546e2611@%3Cdev.openwhisk.apache.org%3E), allows parameters to be designated as init time parameters. The result of such a designation is that the corresponding parameters are communicated to the action at initialization time and should be exported to the action's environment accordingly. This change requires updates to the action runtimes for the latter details and will be done separately. Additonal changes are required to the documentation once all the details are in place.

1. [x] implement the core API change
2. [ ] extends all the runtimes
    1. [x] nodejs
    2. [ ] c#
    3. [ ] all action loop related runtimes
3. [ ] update cli (Add `-e`)
4. [ ] update documentation (include examples of using this capability with `wsk` and functions)

PR #4559, opened 2019-07-16
Allow parameters to be designated as init time properties

rabbah:
As discussed on the [dev list](https://lists.apache.org/thread.html/7e03e6fc965d972882f4c76e34b2e9aa28579a0c3004d63d546e2611@%3Cdev.openwhisk.apache.org%3E) this patch allows parameters to be designated as init time parameters. The result of such a designation is that the corresponding parameters are communicated to the action at initialization time and should be exported to the action's environment accordingly. This change requires updates to the action runtimes for the latter details and will be done separately.

This PR does not yet update the documentation since the CLI is also not updated and until we have all the runtime support in place, it doesn't make sense to document the feature yet. 

The commit history shows the implementation evolution from using a heuristic to a parameter modifier. Each commit is self-contained and can be reviewed on its own.

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#4558)

- [x] API
- [x] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/entity/Parameter.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/entity/Parameter.scala
@@ -76,8 +76,8 @@ protected[core] class Parameters protected[entity] (private val params: Map[Para
   }
 
   /** Gets list all defined parameters. */
-  protected[core] def initParameters: Array[String] = {
-    params.keySet filter (params(_).init) map (_.name) toArray
+  protected[core] def initParameters: Set[String] = {
+    params.keySet filter (params(_).init) map (_.name)
   }
 
   protected[core] def toJsArray = {
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/entity/Parameter.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/entity/Parameter.scala
@@ -160,10 +160,10 @@ protected[entity] class ParameterName protected[entity] (val name: String) exten
 protected[entity] case class ParameterValue protected[entity] (private val v: JsValue, val init: Boolean) {
 
   /** @return JsValue if defined else JsNull. */
-  protected[entity] val value = Option(v) getOrElse JsNull
+  protected[entity] def value = Option(v) getOrElse JsNull
 
   /** @return true iff value is not JsNull. */
-  protected[entity] val isDefined = value != JsNull
+  protected[entity] def isDefined = value != JsNull
 
   /**
    * The size of the ParameterValue entity as ByteSize.
@@ -193,7 +193,7 @@ protected[core] object Parameters extends ArgNormalizer[Parameters] {
   protected[core] def apply(p: String, v: String, init: Boolean = false): Parameters = {
     require(p != null && p.trim.nonEmpty, "key undefined")
     Parameters() + (new ParameterName(ArgNormalizer.trim(p)),
-    new ParameterValue(Option(v).map(_.trim.toJson).getOrElse(JsNull), init))
+    ParameterValue(Option(v).map(_.trim.toJson).getOrElse(JsNull), init))
   }
 
   /**
@@ -209,7 +209,7 @@ protected[core] object Parameters extends ArgNormalizer[Parameters] {
   protected[core] def apply(p: String, v: JsValue, init: Boolean): Parameters = {
     require(p != null && p.trim.nonEmpty, "key undefined")
     Parameters() + (new ParameterName(ArgNormalizer.trim(p)),
-    new ParameterValue(Option(v).getOrElse(JsNull), init))
+    ParameterValue(Option(v).getOrElse(JsNull), init))
   }
 
   /**
@@ -224,7 +224,7 @@ protected[core] object Parameters extends ArgNormalizer[Parameters] {
   protected[core] def apply(p: String, v: JsValue): Parameters = {
     require(p != null && p.trim.nonEmpty, "key undefined")
     Parameters() + (new ParameterName(ArgNormalizer.trim(p)),
-    new ParameterValue(Option(v).getOrElse(JsNull), false))
+    ParameterValue(Option(v).getOrElse(JsNull), false))
   }
 
   override protected[core] implicit val serdes = new RootJsonFormat[Parameters] {
@@ -257,11 +257,11 @@ protected[core] object Parameters extends ArgNormalizer[Parameters] {
         _.asJsObject.getFields("key", "value", "init") match {
           case Seq(JsString(k), v: JsValue) =>
             val key = new ParameterName(k)
-            val value = new ParameterValue(v, false)
+            val value = ParameterValue(v, false)
             (key, value)
           case Seq(JsString(k), v: JsValue, JsBoolean(i)) =>
             val key = new ParameterName(k)
-            val value = new ParameterValue(v, i)
+            val value = ParameterValue(v, i)
             (key, value)
         }
       } toMap)
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/entity/Parameter.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/entity/Parameter.scala
@@ -75,9 +75,16 @@ protected[core] class Parameters protected[entity] (private val params: Map[Para
     params.keySet filter (params(_).isDefined) map (_.name)
   }
 
+  /** Gets list all defined parameters. */
+  protected[core] def initParameters: Array[String] = {
+    params.keySet filter (params(_).init) map (_.name) toArray
+  }
+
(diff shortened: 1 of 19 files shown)


tools/actionProxy/invoke.py
```diff
@@ -103,11 +104,18 @@ def init(args):
         contents = None
         binary = False
 
+    value = processPayload(args.env)
```
chetanmeh: Nit - Instead of `value` name it as `env`

common/scala/src/main/scala/org/apache/openwhisk/core/entity/Parameter.scala
```diff
@@ -144,19 +151,19 @@ protected[entity] class ParameterName protected[entity] (val name: String) exten
...
- * @param value the value of the parameter, may be null
+ * @param v the value of the parameter, may be null
+ * @param init if true, this parameter value is only offered to the action during initialization
  */
-protected[entity] class ParameterValue protected[entity] (private val v: JsValue) extends AnyVal {
+protected[entity] case class ParameterValue protected[entity] (private val v: JsValue, val init: Boolean) {
 
   /** @return JsValue if defined else JsNull. */
-  protected[entity] def value = Option(v) getOrElse JsNull
+  protected[entity] val value = Option(v) getOrElse JsNull
```
chetanmeh: Just curious - Any reason to make them `val` instead of `def` as they would then consume small memory

tests/src/test/scala/actionContainers/BasicActionRunnerTests.scala
```diff
@@ -260,6 +283,43 @@ trait BasicActionRunnerTests extends ActionProxyContainerTestUtils {
     })
   }
 
+  it should s"export environment variables before initialization" in {
+    val config = testEnvPartition
+    if (!config.skipTest) {
```
chetanmeh: How can we determine from test run results if some runtime supports a specific feature or not? May be we throw a `TestPendingException` such that test reports call them out as pending

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -770,6 +770,23 @@ object ContainerProxy {
...
+
+  /**
+   * Partitions the activation arguments into two JsObject instances. The first is exported as intended for export
+   * by the action runtime to the environment. The second is passed on as arguments to the action.
+   *
+   * @param content the activation arguments
+   * @param initArgs set of parameters to treat as initialization arguments
+   * @return A partition of the arguments into an environment variables map and the JsObject argument to the action
+   */
+  def partitionArguments(content: Option[JsObject], initArgs: Set[String]): (Map[String, JsValue], JsObject) = {
```
chetanmeh: May be to avoid `content.get` i.e. get on option (though its safe here) we can use

```scala
  def partitionArguments(content: Option[JsObject], initArgs: Set[String]): (Map[String, JsValue], JsObject) = {
    content match {
      case Some(js) if initArgs.empty => (Map.empty, js)
      case Some(js) =>
        val (env, args) = js.fields.partition(k => initArgs.contains(k._1))
        (env, JsObject(args))
      case None =>
        (Map.empty, JsObject.empty)
    }
```

docs/actions-new.md
```diff
@@ -146,7 +146,8 @@ The initialization route is `/init`. It must accept a `POST` request with a JSON
     "name" : String,
     "main" : String,
     "code" : String,
-    "binary": Boolean
+    "binary": Boolean,
+    "env": Map[String, JSON Value]
```
chetanmeh: Ideally it should be a map of `<String, String>`. This would simplify the implementation to just handle the value directly. Otherwise they would need to stringify the value which may vary depending on runtime

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
chetanmeh: Further it would be better if we restrict the env to `Map[String, String]` here i.e. restict/coerce the value to string here

common/scala/src/main/scala/org/apache/openwhisk/core/entity/Parameter.scala
```diff
@@ -144,19 +151,19 @@ protected[entity] class ParameterName protected[entity] (val name: String) exten
...
- * It is a value type (hence == is .equals, immutable and cannot be assigned null).
  * The constructor is private so that argument requirements are checked and normalized
  * before creating a new instance.
  *
- * @param value the value of the parameter, may be null
+ * @param v the value of the parameter, may be null
+ * @param init if true, this parameter value is only offered to the action during initialization
  */
-protected[entity] class ParameterValue protected[entity] (private val v: JsValue) extends AnyVal {
+protected[entity] case class ParameterValue protected[entity] (private val v: JsValue, val init: Boolean) {
```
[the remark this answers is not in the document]
rabbah: logically yes, but there are several operation on parameter names which cons new instances, so i'd have to override equals method (to permit for example removal of a parameter from a map regardless of its init setting); thus attaching the bit on the value treats the setting more opaquely. 

if we reorganize the parameters and annotations to a proper dictionary, we can clean up this whole implementation at some point in the future.

docs/actions-new.md
rabbah: OK - I thought I'd defer this to the runtime so the invoker is doing less work but I can make this change, as your point is valid.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
rabbah: I'll defer this to the `containerInitializer()`.

docs/actions-new.md
rabbah: Thinking about this further, if the value is JSON stringified in a runtime-specific way, it should still be JSON parsed back to the original value. To do this on the invoker side means walking the properties and normalizing each property individually. It would indeed be less work for the runtimes but more work on the invoker. Thoughts?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
rabbah: OK - Since we have to turn this into a JsObject, we eventually still need a Map[String, JsString].

docs/actions-new.md
rabbah: The last commit implement normalization of the env vars to a map[string, string].

rabbah:
Thanks Chetan for the review - much appreciated. I believe I addressed your feedback and questions in additional commits.

docs/actions-new.md
chetanmeh: >  It would indeed be less work for the runtimes but more work on the invoker

Yes but then we would only need to get it right at one place and have consistent behavior for various runtimes

Thanks for addressing this.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -770,6 +770,31 @@ object ContainerProxy {
...
+   * @return A partition of the arguments into an environment variables map and the JsObject argument to the action
+   */
+  def partitionArguments(content: Option[JsObject], initArgs: Set[String]): (Map[String, JsString], JsObject) = {
+    content match {
+      case None                         => (Map.empty, JsObject.empty)
+      case Some(js) if initArgs.isEmpty => (Map.empty, js)
+      case Some(js) =>
+        val (env, args) = js.fields.partition(k => initArgs.contains(k._1))
+
+        (env.map {
```
chetanmeh: Just a minor concern - This logic would be executed for both `/init` and `/run` flow while its only required for `/init`. If we can avoid this for `/run` (which is  the more critical path) then that would reduce unnecessary object creations..

May be break partition into 2 parts - partition and stringify and have stringify logic invoked only for `/init` flows

rabbah: Agreed. I wanted to defer stringify to `containerInitializer().`

rabbah: https://github.com/apache/openwhisk/pull/4559/commits/1fb74dcccd834846975eaf68883875853d03b834 relocated the conversion to the initializer.

PR #4561, opened 2019-07-19
Add more diagnostic information to completion ack processing

sven-lange-last:
Forced completion acks can be a source of overloaded invokers because forced acks free up invoker slots in the load balancer. If the invoker is just "late" with running an activation so that the invoker sends the result / completion ack after the completion ack timeout has been reached in the load balancer, the load balancer may send new activations to the invoker while it is still fully occupied. As a result, these new activations have to wait for a container and can cause forced completion acks in turn. This mechanism can lead to a never-ending chain reaction if an invoker has a constant stream of workload.

Goal of this change is to improve visibility and diagnostic information so that this mechanism can be better understood and fixed in the next step.

* Add metrics for forced completion acks.
* Log more diagnostic information when forcing a completion ack due to timeout.
* Log a warning if a completion ack arrives after it has already been forced - this is an indication that action processing took too long. Today, this situation is only logged as debug message.
* Improve code comments.
* Provide a function for calculating the completion ack timeout. In the long term, this should be unified with the action wait timeout in sequence activations.

The proposed change is related to following discussion thread on the Apache OW dev mailing list: https://lists.apache.org/thread.html/6f0a809584f4bfe712669550ffd99a129202e146235e7c0f75595cd6@%3Cdev.openwhisk.apache.org%3E

## Related issue and scope

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -414,6 +414,28 @@ object LoggingMarkers {
       LogMarkerToken(loadbalancer + controllerInstance.asString, s"memory${actionType}Inflight", counter)(
         MeasurementUnit.none)
 
+  // Counter metrics for completion acks in load balancer
+  sealed trait completionAckType { def asString: String = ??? }
+  case object regularCompletionAck extends completionAckType { override def asString = "regular" }
+  case object forcedCompletionAck extends completionAckType { override def asString = "forced" }
+  case object healthcheckCompletionAck extends completionAckType { override def asString = "healthcheck" }
+  case object regularAfterForcedCompletionAck extends completionAckType { override def asString = "regularAfterForced" }
+  case object forcedAfterRegularCompletionAck extends completionAckType { override def asString = "forcedAfterRegular" }
+
+  def LOADBALANCER_COMPLETION_ACK(controllerInstance: ControllerInstanceId, completionAckType: completionAckType) =
+    if (TransactionId.metricsKamonTags)
+      LogMarkerToken(
+        loadbalancer,
+        "completionAck",
+        counter,
+        None,
+        Map("controller_id" -> controllerInstance.asString, "type" -> completionAckType.asString))(MeasurementUnit.none)
+    else
+      LogMarkerToken(
+        loadbalancer + controllerInstance.asString,
+        "completionAck_" + completionAckType.asString,
+        counter)(MeasurementUnit.none)
+
   // Time that is needed to execute the action
   val INVOKER_ACTIVATION_RUN =
     LogMarkerToken(invoker, "activationRun", start)(MeasurementUnit.time.milliseconds)
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -415,14 +415,14 @@ object LoggingMarkers {
         MeasurementUnit.none)
 
   // Counter metrics for completion acks in load balancer
-  sealed trait completionAckType { def asString: String = ??? }
-  case object regularCompletionAck extends completionAckType { override def asString = "regular" }
-  case object forcedCompletionAck extends completionAckType { override def asString = "forced" }
-  case object healthcheckCompletionAck extends completionAckType { override def asString = "healthcheck" }
-  case object regularAfterForcedCompletionAck extends completionAckType { override def asString = "regularAfterForced" }
-  case object forcedAfterRegularCompletionAck extends completionAckType { override def asString = "forcedAfterRegular" }
+  sealed abstract class CompletionAckType(val name: String) { def asString: String = name }
+  case object RegularCompletionAck extends CompletionAckType("regular")
+  case object ForcedCompletionAck extends CompletionAckType("forced")
+  case object HealthcheckCompletionAck extends CompletionAckType("healthcheck")
+  case object RegularAfterForcedCompletionAck extends CompletionAckType("regularAfterForced")
+  case object ForcedAfterRegularCompletionAck extends CompletionAckType("forcedAfterRegular")
 
-  def LOADBALANCER_COMPLETION_ACK(controllerInstance: ControllerInstanceId, completionAckType: completionAckType) =
+  def LOADBALANCER_COMPLETION_ACK(controllerInstance: ControllerInstanceId, completionAckType: CompletionAckType) =
     if (TransactionId.metricsKamonTags)
       LogMarkerToken(
         loadbalancer,
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -352,6 +352,26 @@ object LoggingMarkers {
   private val containerClient = "containerClient"
 
   /*
+   * The following markers are used to emit log messages as well as metrics. Add all LogMarkerTokens below to
+   * have a reference list of all metrics. The list below contains LogMarkerToken singletons (val) as well as
+   * LogMarkerToken creation functions (def). The LogMarkerToken creation functions allow to include variable
+   * information in metrics, such as the controller / invoker id or commands executed by a container factory.
+   *
+   * When using LogMarkerTokens for emitting metrics, you should use the convenience functions only once to
+   * create LogMarkerToken singletons instead of creating LogMarkerToken instances over and over again for each
+   * metric emit.
+   *
+   * Example:
+   * val MY_COUNTER_GREEN = LoggingMarkers.MY_COUNTER(GreenCounter)
+   * ...
+   * MetricEmitter.emitCounterMetric(MY_COUNTER_GREEN)
+   *
+   * instead of
+   *
+   * MetricEmitter.emitCounterMetric(LoggingMarkers.MY_COUNTER(GreenCounter))
+   */
+
+  /*
    * Controller related markers
    */
   def CONTROLLER_STARTUP(id: String) =
@@ -422,6 +442,7 @@ object LoggingMarkers {
   case object RegularAfterForcedCompletionAck extends CompletionAckType("regularAfterForced")
   case object ForcedAfterRegularCompletionAck extends CompletionAckType("forcedAfterRegular")
 
+  // Convenience function to create log marker tokens used for emitting counter metrics related to completion acks.
   def LOADBALANCER_COMPLETION_ACK(controllerInstance: ControllerInstanceId, completionAckType: CompletionAckType) =
     if (TransactionId.metricsKamonTags)
       LogMarkerToken(
--- a/core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/CommonLoadBalancer.scala
+++ b/core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/CommonLoadBalancer.scala
@@ -84,17 +84,39 @@ abstract class CommonLoadBalancer(config: WhiskConfig,
   override def totalActiveActivations: Future[Int] = Future.successful(totalActivations.intValue())
 
   /**
+   * Calculate the duration within which a completion ack must be received for an activation.
+   *
+   * Calculation is based on the passed action time limit. If the passed action time limit is shorter than
+   * the configured standard action time limit, the latter is used to avoid too tight timeouts.
+   *
+   * The base timeout is multiplied with a configurable timeout factor. This dilution controls how much slack you
+   * want to allow in your system before you start reporting failed activations. The default value of 2 bases
+   * on invoker behavior that a cold invocation's init duration may be as long as its run duration. Higher factors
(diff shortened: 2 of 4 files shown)


sven-lange-last:
image

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
```diff
@@ -414,6 +414,28 @@ object LoggingMarkers {
       LogMarkerToken(loadbalancer + controllerInstance.asString, s"memory${actionType}Inflight", counter)(
         MeasurementUnit.none)
 
+  // Counter metrics for completion acks in load balancer
+  sealed trait completionAckType { def asString: String = ??? }
```
markusthoemmes: I don't think you need to implement the method here? Just define it in the trait, `???` will throw an exception at runtime.

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
```diff
@@ -414,6 +414,28 @@ object LoggingMarkers {
...
+  case object regularCompletionAck extends completionAckType { override def asString = "regular" }
+  case object forcedCompletionAck extends completionAckType { override def asString = "forced" }
+  case object healthcheckCompletionAck extends completionAckType { override def asString = "healthcheck" }
+  case object regularAfterForcedCompletionAck extends completionAckType { override def asString = "regularAfterForced" }
+  case object forcedAfterRegularCompletionAck extends completionAckType { override def asString = "forcedAfterRegular" }
```
style95: This change would be useful to diagnose operational issues.
I have a few comments.

In a small scope, if you don't have any other purpose to use `toString`, I think you can directly override `toString` here.
Then you can just use that like this: `"type" -> completionAckType`.

And I prefer to use the upper case first character for the object and trait according to the scala convention.
https://docs.scala-lang.org/style/naming-conventions.html


In a bigger scope, this is a good way to define subclass(object) but if this trait is only used here I'd rather define them using just `Sting`.

```scala
def LOADBALANCER_COMPLETION_ACK(controllerInstance: ControllerInstanceId, completionAckType: String) =
    if (TransactionId.metricsKamonTags)
      LogMarkerToken(
        loadbalancer,
        "completionAck",
        counter,
        None,
        Map("controller_id" -> controllerInstance.asString, "type" -> completionAckType))(MeasurementUnit.none)
    else
      LogMarkerToken(
        loadbalancer + controllerInstance.asString,
        "completionAck_" + completionAckType,
        counter)(MeasurementUnit.none)
```

If you mind a client uses this with just string, I think you can also define them as a `Constant`.

```scala
val RegularCompletionAck: String = "regular"
.
.
MetricEmitter.emitCounterMetric(
            LoggingMarkers.LOADBALANCER_COMPLETION_ACK(controllerInstance, RegularCompletionAck))
        }
```

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
sven-lange-last: Changed to variant without `???` but only declaring `asString`. Compiler will then complain of a sub-type does not implement `asString`.

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
markusthoemmes: @style95 for posterity: We use `asString` throughout the codebase for places where we explicitly want a conversion to string to happen. `toString` way too often bites you so we make these conversions as explicit as possible.

style95: oh ok, I didn't know the history.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/CommonLoadBalancer.scala
```diff
@@ -248,16 +274,30 @@ abstract class CommonLoadBalancer(config: WhiskConfig,
           // notice here that the activationPromises is not touched, because the expectation is that
           // the active ack is received as expected, and processing that message removed the promise
           // from the corresponding map
+          logging.info(this, s"received completion ack for '$aid', system error=${isSystemError}")(tid)
```
chetanmeh: This info log would be logged for happy path i.e. for all successful ack activations? If yes should we make it debug or it helps in debugging some case?

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/CommonLoadBalancer.scala
```diff
@@ -248,16 +274,30 @@ abstract class CommonLoadBalancer(config: WhiskConfig,
...
           // and complete the promise with a failure if necessary
           activationPromises
             .remove(aid)
             .foreach(_.tryFailure(new Throwable("no completion or active ack received yet")))
+          val actionType = if (entry.isBlackbox) "blackbox" else "managed"
+          val blockingType = if (entry.isBlocking) "blocking" else "non-blocking"
+          val completionAckTimeout = calculateCompletionAckTimeout(entry.timeLimit)
+          logging.warn(
+            this,
+            s"forced completion ack for '$aid', action '${entry.fullyQualifiedEntityName}' (${actionType}), ${blockingType}, mem limit ${entry.memoryLimit.toMB} MB, time limit ${entry.timeLimit.toMillis} ms, completion ack timeout ${completionAckTimeout} from ${invoker}")(
```
chetanmeh: Minor Nit ... In general braces not needed for simple expression like in `${actionType}`

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/CommonLoadBalancer.scala
```diff
@@ -266,17 +306,31 @@ abstract class CommonLoadBalancer(config: WhiskConfig,
...
       case None if !forced =>
-        // Received an active-ack that has already been taken out of the state because of a timeout (forced active-ack).
+        // Received a completion ack that has already been taken out of the state because of a timeout (forced ack).
         // The result is ignored because a timeout has already been reported to the invokerPool per the force.
-        logging.debug(this, s"received completion ack for '$aid' which has no entry")(tid)
+        // Logging this condition as a warning because the invoker processed the activation and sent a completion
+        // message - but not in time.
+        logging.warn(
+          this,
+          s"received completion ack for '$aid' from ${invoker} which has no entry, system error=${isSystemError}")(tid)
```
chetanmeh: Would it be possible to compute the time delta with which this ack missed the forced ack deadline. May be logging the `start` time from `TransactionId` (not sure how that works with sequence. Probably there it may mislead)

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
sven-lange-last: Hm... I introduced `case object`s here for type safety. When using `String`s, you could basically feed in any `String` value. Using pre-defined `String` values is just convenience but no guarantee that "somebody" uses a bogus `String` value you do not expect in your metrics system. Why not use Scala's type system?

I intentionally used `case object`s so they should already be singletons. Or do I miss something here? So there should be no need to define convenience `val`s.

Also see https://github.com/apache/incubator-openwhisk/blob/bf63bb2c31e5907689655d226511cbb99126b734/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala#L43-L52.

@markusthoemmes thanks for jumping in on `asString`. That's exactly why I used `asString` instead of `toString`. If you agree, I would like to keep `asString`.

I agree on the Scala naming convention. Too much golang...

chetanmeh: > I intentionally used case objects so they should already be singletons. Or do I miss something here? So there should be no need to define convenience vals.

`LogMarkerToken` are meant to be treated like static/singleton as they in turn hold reference to actual metric instance like counter. In current approach we create a new instance of token for every call to `emitCounterMetric`. Further that token instance would then inturn lookup the backing `Kamon` metric instance. 

So to avoid this overhead in metric computation it would be better to refer to tokens as vals. This ensures that `MetricEmitter.emitCounterMetric` call just leads to only increment operation and does not create new instance and do lookup.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/CommonLoadBalancer.scala
sven-lange-last: The regular completion ack processing already created an info log before. Unless you have evidence that it should be removed, I would like to keep it because it already helped me in production systems to debug problems.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/CommonLoadBalancer.scala
sven-lange-last: I personally prefer to use the brace syntax (`${}`) even if it's not required. For me, it's more readable. If you think, I should remove, please let me know.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/CommonLoadBalancer.scala
chetanmeh: Missed that aspect. Then we can keep it as is

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/CommonLoadBalancer.scala
chetanmeh: Ack. For me Intellij just highlight thats _The enclosing block is redundant_ which turn the file marked as yellow instead of green! I would probably disable this inspection as this usage is in quite a few places

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
sven-lange-last: I had a private sync with @chetanmeh on his proposal.

* Currently, this PR defines a counter metric convenience function `LOADBALANCER_COMPLETION_ACK` in object `LoggingMarkers` which is called over and over again by `CommonLoadBalancer`. This will create a log of `LogMarkerToken` instances.
* Chetan's proposal is to create `val`s in `CommonLoadBalancer` using `LOADBALANCER_COMPLETION_ACK` and then use these `LogMarkerToken` instances to emit  metrics. This makes a lot of sense!
* At the same time, I will keep `CompletionAckType` and sub-types as well as `LOADBALANCER_COMPLETION_ACK` in object `LoggingMarkers` to define all possible metrics in a single place. With that approach, you do not need multiple steps to look up metrics.
* I will also streamline `CompletionAckType` a little bit to save boilerplate code.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/CommonLoadBalancer.scala
sven-lange-last: I will go through my PR again and remove unnecessary braces (`${}`) to help you guys with IntelliJ.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/CommonLoadBalancer.scala
sven-lange-last: I checked that part... Apparently, this is the code path where the activation entry has already been removed. So we cannot easily provide this info via the activation entry.

Chetan is right with sequences. The transaction starts with the sequence and all but the first action in the sequence will have later start times than the start of the transaction. For this reason, the `tid` cannot be used easily to compute the time delta.

I suggest to better leave out this info for now instead of provide improper information that could be misleading. Even for primitive actions, computing the time delta from the start of the transaction can be improper...

sven-lange-last:
@chetanmeh @style95 @markusthoemmes I hope to addressed all your feedback as discussed above. Please take a look again.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/CommonLoadBalancer.scala
```diff
@@ -214,6 +240,18 @@ abstract class CommonLoadBalancer(config: WhiskConfig,
 
   protected def releaseInvoker(invoker: InvokerInstanceId, entry: ActivationEntry)
 
+  // Singletons for counter metrics related to completion acks
+  val LOADBALANCER_COMPLETION_ACK_REGULAR =
```
chetanmeh: Can be made private

sven-lange-last: Good feedback. I made them `protected`. If a sub-type decided to override `processCompletion()`, it also had access to the `LogMarkerToken` instances.

PR #3685, opened 2018-05-22
Enable test code coverage collection for containers

chetanmeh:
Enables collection of test code coverage for code running within containers like Controller and Invoker

## Description

So far with #3249 we have been collecting test code coverage data from unit and system test for code which need not be running within container. However some of the code paths can only be exercised within a running container. This PR enables collecting test code coverage stats for such code paths.

As explained in #3243 scoverage creates `build/scoverage/scoverage.measurements.xxx` files (xxx maps to threadId) at runtime to record any code path being hit. Instrumented code has compile time configured paths for the data directory. So for code compiled for coverage data collection on travis the instrumented code would try to write the measurement file under _/home/travis/build/apache/incubator-openwhisk/core/controller/build/scoverage_

For normal scenario when code is running outside of container such files are written within `build/scoverage` directory and is accessible when generating reports. However when code runs within container its not possible to access such generated files. To enable that following approach has been taken

### Preparing Instrumented Containers

Docker build now supports a `distDockerCoverage` task which generate a new image for controller and invoker containing instrumented code based on `Dockerfile.cov` file. 

The instrumented container 

* Is tagged with `cov` tag e.g. `whisk/controller:cov`
* Adds scoverage runtime jars and replaces existing openwhisk-common etc jars with those generated by scoverage task
* Adds symlinks in container for data directory path

### Running Instrumented Containers

Further ansible playbooks are modified to look for `GRADLE_COVERAGE` environment variable to enable some specific tasks. Once enabled the new tasks would

1. Create specific directory for controller and invoker like `/tmp/wskcov/controller`
2. Expose these directory as shared volume with path `/coverage`

Once running a path like  _/home/travis/build/apache/incubator-openwhisk/core/controller/build/scoverage_ would be symlinked to `/coverage/controller` which inturn is pointing to `/tmp/wskcov/controller/controller`

### Report Generation

In test module the report generation task would copy all such measurement files from directories under `/tmp/wskcov` to respective modules. The file names would be changed to ensure that they are unique. Post this report would include stats from code running within containers.

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#3243)

## My changes affect the following components

- [ ] API
- [ ] Controller
- [ ] Message Bus (e.g., Kafka)
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Intrinsic actions (e.g., sequences, conductors)
- [ ] Data stores (e.g., CouchDB)
- [x] Tests
- [ ] Deployment
- [ ] CLI
[...]

PR #4567, opened 2019-07-24
Update Gradle to 5.5 version

chetanmeh:
Updates Gradle to 5.5 version and also updates Gradle Scoverage plugin to 3.1.5

## Description

For enabling running and building on JDK 11 (#4217) we need to move to Gradle 5.x version which officially supports JDK 11. In doing so we would need to update the [Gradle Scoverage](https://github.com/scoverage/gradle-scoverage) plugin to 3.1.5 as existing version was not compatible

### Benefits

1. Enables building on JDK 11
2. Gradle 5.x has better support for Maven publishing. We would now be needing that to publish the standalone jar

### Scoverage Version Impact

Newer soverage plugin has a different model of working and does not produce a jar of instrumented classes which impacts our logic for coverage collection for code running within container (#3685). To make it work we now copy the instrumented classes within container and remove existing packaged jars from the lib

## Related issue and scope

- [ ] I opened an issue to propose and discuss this change (#4217 )

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


tests/build.gradle
```diff
@@ -250,8 +255,33 @@ gradle.projectsEvaluated {
...
             exceptionFormat = 'full'
         }
+        maxHeapSize = "1500m"
         outputs.upToDateWhen { false } // force tests to run every time
     }
+    /**
+     * Task to generate coverage xml report. Requires the
+     * tests to be executed prior to its invocation
+     */
+    task reportCoverage(type: ScoverageReport) {
```
chetanmeh: The new version now creates task like `report<Test Task Name>Scoverage` as entry task. This task in turn adapts the compile task to use the instrumented classes. This works fine for conventional cases where tests are part of same module. 

However in our case as we have a separate test module and further to capture coverage of code running within container (as part of integration test) we copy the coverage metrics files manually we cannot use default tasks. Hence the report task duplicate bit of logic from actual task

tests/build.gradle
```diff
@@ -250,8 +255,33 @@ gradle.projectsEvaluated {
             showStandardStreams = true
             exceptionFormat = 'full'
         }
+        maxHeapSize = "1024m" //Gradle 5.5 defaults to 512MB which is low
```
chetanmeh: With Gradle 5.0 it now [defaults to 512M of heap][1]. Due to this OOM was seen in test runs on Travis (which has 7 GB memory). Hence specifying the size explicitly to 1024 MB now

> All workers, including compilers and test executors, now start with 512MB of heap. The previous default was 1/4th of physical memory. Large projects may have to increase this setting on the relevant tasks, e.g. JavaCompile or Test.



[1]: https://docs.gradle.org/current/userguide/upgrading_version_4.html#rel5.0:default_memory_settings

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -17,6 +17,6 @@
 
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.5.1-all.zip
```
chetanmeh: Switching to 5.5.1 and also to `all` mode so as to enable code assists in IDE

settings.gradle
```diff
@@ -43,13 +43,6 @@ gradle.ext.scalafmt = [
     config: new File(rootProject.projectDir, '.scalafmt.conf')
 ]
 
-gradle.ext.scoverage = [
```
chetanmeh: Now scoverage plugin internally adapts the classpath with required dependencies. So no need to explicitly add them

tests/build.gradle
```diff
@@ -315,37 +330,11 @@ def loadWhiskProps(){
     p
 }
 
-/**
- * Aggregates the scoverage xml reports from various modules into a
- * single report
- */
-task aggregateCoverage(type: JavaExec, dependsOn: reportCoverage) {
```
chetanmeh: This task was never used as we rely on [codecov][1] to perform actual aggregation

[1]: https://codecov.io/

chetanmeh:
This PR is now ready for review

chetanmeh:
@dgrove-oss @rabbah Can you review this PR. Once merged we can then work on publishing Maven artifacts from core repo (#4579)

gradlew.bat
```diff
@@ -30,7 +30,7 @@ set APP_BASE_NAME=%~n0
 set APP_HOME=%DIRNAME%
 
 @rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS=-Dfile.encoding=UTF-8
+set DEFAULT_JVM_OPTS=-Dfile.encoding=UTF-8 "-Xmx64m" "-Xms64m"
```
rabbah: is this too low?

rabbah:
No issues on my side using this patch.

gradlew.bat
chetanmeh: Yeah but thats the [current default in there generated files][1]

[1]: https://github.com/gradle/gradle/blob/362226499afd9180ebc19822d44fca5fbaf004e9/gradlew#L47

dgrove-oss: If a different heap size is appropriate, we can change it.  It's fine for us to diverge from the upstream gradlew[.bat] as needed.

chetanmeh: For now build seems to work fine. If we see any issue then we can change. I believe Gradle wrapper internally forks a new jvm process so it needs bare minimum memory for its work

PR #4568, opened 2019-07-24
Adopt full ASF headers for all test files regardless of size

mrutkows:


rabbah:
Thanks @mrutkows.
I suspect some files contained in .zip files might need to be updated. Perhaps we should provide sources for all the zip files and a build script to create the zips used in tests. This should/could be a separate issue of course.

mrutkows:
@rabbah I just did the "tests" exclusion... will look at bin/zip next after some meetings

rabbah:
my instinct is to remove the zip and instead provide a bash script to zip test artifacts in tests/data/actions. This script could be run by the travis jobs but for releases, would result in RC that have no zips.

dgrove-oss:
+1 for the zip removal + scripts to regen it.  I believe we should not release zip files.  Since we want to include tests in the next release, we need to eliminate zip/tars from the checked in source.

Issue #4569, opened 2019-07-25
Kubernetes Container Factory doesn't support extraArgs environment variable


Logicon211:
When deploying openwhisk invokers using the kubernetes container factory we should be able to set the 

```
CONFIG_whisk_containerFactory_containerArgs_extraArgs_env_0
```

environment variable in order to pass along environment variables to whisk runtime containers like we can when using the docker container factory.

dgrove-oss:
Last PR merged in deploy-kube.  Fix completed.

PR #4570, opened 2019-07-25
KCF: propagate cf_ca_extraArgs_env_i into user containers

dgrove-oss:
Partial fix for #4569.

Add logic to KubernetesContainerFactory to propagate the
set of key=value pairs) from containerFactory.containerArgs.extraArgs.envN
into the environment of all action containers.

This fixes the OpenWhisk core code.  After this is merged & new docker images are built,
the Helm chart needs to be augmented to expose this in a nice way to end users.


core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/kubernetes/KubernetesContainerFactory.scala
```diff
@@ -40,12 +41,25 @@ import org.apache.openwhisk.core.{ConfigKeys, WhiskConfig}
...
   config: WhiskConfig,
+  containerArgsConfig: ContainerArgsConfig = loadConfigOrThrow[ContainerArgsConfig](ConfigKeys.containerArgs),
   runtimesRegistryConfig: RuntimesRegistryConfig = loadConfigOrThrow[RuntimesRegistryConfig](
     ConfigKeys.runtimesRegistry))(implicit actorSystem: ActorSystem, ec: ExecutionContext, logging: Logging)
     extends ContainerFactory {
 
   implicit val kubernetes = initializeKubeClient()
 
+  val augmentedEnv: Map[String, String] = {
+    val envArgs = containerArgsConfig.extraArgs.filter((t) => t._1.equals("env")).values
```
markusthoemmes: Isn‘t this a Map anyway? You should be able to use just a „get“ for the „env“ key instead of filtering the map for its key.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/kubernetes/KubernetesContainerFactory.scala
```diff
@@ -40,12 +41,25 @@ import org.apache.openwhisk.core.{ConfigKeys, WhiskConfig}
...
+    envArgs
+      .flatMap(
```
markusthoemmes: This is obsoleted by the above comment I believe.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/kubernetes/KubernetesContainerFactory.scala
```diff
@@ -40,12 +41,25 @@ import org.apache.openwhisk.core.{ConfigKeys, WhiskConfig}
...
+        _.filter(_.contains("="))
+          .map(x => {
+            val tmp = x.split("=")
```
markusthoemmes: You can make use of flatMap to fold this map and the filter into one operation I think:

```scala
val envArgs = ...getOrElse("env", Set.empty[String])
envArgs
  .flatMap(_.split("=") match {
    case Array(key, value) => Some(key -> value)
    case _ => None
  }.toMap
```

sven-lange-last: @markusthoemmes good proposal - this one should work:

```
    val envVars: Set[String] = containerArgsConfig.extraArgs.getOrElse("env", Set.empty[String])
    val augmentedEnv: Map[String, String] = envVars.flatMap {
      _.split("=", 2) match {
        case Array(key, value) => Some(key -> value)
        case _                 => None
      }
    }.toMap
```

dgrove-oss:
I like @sven-lange-last 's proposal:

> Would it make sense to introduce a general concept for injecting static environment variables into action containers. I'm not talking about env variables from the user side but to be injected from the system side. We could introduce a config parameter whisk.container-factory.extra-env-vars that holds a list of env vars. It's up to the container factory implementations to inject the variables.

I wasn't that happy about shoving it through the extra args and looking for the `env` key, but that was the minimal change.  Doing a more general thing makes sense.

For DCF, I would plan to merge `whisk.container-factory.extra-env-vars`  and any additional values provided in the `extra-args` with an `env` key.  Sound plausible?

dgrove-oss:
Revised as suggested by @sven-lange-last.  There was already a unit test for DCF for extra args that I updated to use the new approach.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerFactory.scala
```diff
@@ -30,7 +30,18 @@ case class ContainerArgsConfig(network: String,
...
-                               extraArgs: Map[String, Set[String]] = Map.empty)
+                               extraEnvVars: Seq[String] = Seq.empty,
+                               extraArgs: Map[String, Set[String]] = Map.empty) {
+
+  val extraEnvVarMap: Map[String, String] =
+    extraEnvVars.flatMap {
+      _.split("=", 2) match {
+        case Array(key)        => Some(key -> "")
+        case Array(key, value) => Some(key -> value)
+        case _                 => None
```
dgrove-oss: I think the _ case is unreachable and maybe we don't need the Option type now since a failure to split on `=` can be interpreted as a request to simply set an envvar with an empty value.  But I don't think it hurts to do it this way.

core/invoker/src/main/resources/application.conf
```diff
@@ -93,6 +93,7 @@ whisk {
       dns-servers: []
       dns-search: []
       dns-options: []
+      extra-env-vars: [] # sequence of `key` and/or `key=value` bindings to add to all user action container environments
```
markusthoemmes: If we introduce a fresh thing anyway, why not make it a map from the getgo to alleviate any parsing issues?

dgrove-oss: reasonable point.  I'm on vacation until Wed; will make the change when I get back.

dgrove-oss: I prototyped using a map instead of a seq[k=v] and found an unfortunate interaction with `transformEnvironment.sh` that makes a map cumbersome to use.   The problem is `transformEnvirovnment.sh` rewrites `_` to `.` which makes it quite cumbersome to define environment variables whose name contains an `_` without adding an escape sequence.
For example, consider trying to set `__OW_ALLOW_CONCURRENT=false` via a `CONFIG_` envar.  With a Seq, it is:
```
CONFIG_whisk_containerFactory_containerArgs_extraArgs_env_0=__OW_ALLOW_CONCURRENT=false
```
With a Map, it would be 
```
CONFIG_whisk_containerFactory_containerArgs_extraArgs_env___OW_ALLOW_CONCURRENT=false
```
but we also need to somehow escape all the `_` after `env_` to prevent `transformEnvironment.sh` from converting them to `.`. 

So, I think using Seq to avoid needing to escape `_` is a better trade-off overall.

dgrove-oss:
This is ready for final review

dgrove-oss:
@markusthoemmes @sven-lange-last -- any additional comments?  Would like to get this merged.

PR #4581, opened 2019-08-12
graduation: remove more incubator- references

dgrove-oss:
Remove almost all incubator- cross references to github repos,
leaving only the subset of repos that have not yet been renamed.


dgrove-oss:
this is ready for merge; needs an approving review

PR #4582, opened 2019-08-13
Action container log collection does not wait for sentinel on developer error

sven-lange-last:
* So far, log collection for managed Docker action containers waits for sentinels - even if a developer error occurs. Only if an action exceeded its init or run timeout, sentinel waiting will be disabled.
* This change disables sentinel waiting for all developer errors because a managed action runtime may not be able to write sentinels at all in developer error situations. With this change, missing sentinels are handled in a more robust way.
* Without sentinel waiting, the Docker action container log file is only read until file end is reached. No re-reads are performed if the sentinel has not been found yet. On busy systems, this may lead to a few missing log lines in case of developer errors.

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
@@ -123,8 +123,12 @@ class DockerToActivationFileLogStore(system: ActorSystem, destinationDirectory: 
                            container: Container,
                            action: ExecutableWhiskAction): Future[ActivationLogs] = {
 
-    val isTimedoutActivation = activation.isTimedoutActivation
-    val logs = logStream(transid, container, action, isTimedoutActivation)
+    val logs = logStream(
+      transid,
+      container,
+      action.limits.logs,
+      action.exec.sentinelledLogs,
+      activation.response.isContainerError) // container error means developer error
 
     // Adding the userId field to every written record, so any background process can properly correlate.
     val userIdField = Map("namespaceId" -> user.namespace.uuid.toJson)
@@ -150,8 +154,10 @@ class DockerToActivationFileLogStore(system: ActorSystem, destinationDirectory: 
     val combined = OwSink.combine(toSeq, toFile)(Broadcast[ByteString](_))
 
     logs.runWith(combined)._1.flatMap { seq =>
+      // Messages.logWarningDeveloperError shows up if the activation ends in a developer error.
+      // This must not be reported as log collection error.
       val possibleErrors = Set(Messages.logFailure, Messages.truncateLogs(action.limits.logs.asMegaBytes))
-      val errored = isTimedoutActivation || seq.lastOption.exists(last => possibleErrors.exists(last.contains))
+      val errored = seq.lastOption.exists(last => possibleErrors.exists(last.contains))
       val logs = ActivationLogs(seq.toVector)
       if (!errored) {
         Future.successful(logs)
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
@@ -40,7 +40,6 @@ import org.apache.openwhisk.common.{AkkaLogging, TransactionId}
 import org.apache.openwhisk.core.containerpool.Container
 import org.apache.openwhisk.core.entity.{ActivationLogs, ExecutableWhiskAction, Identity, WhiskActivation}
 import org.apache.openwhisk.core.entity.size._
-import org.apache.openwhisk.http.Messages
 
 import spray.json._
 import spray.json.DefaultJsonProtocol._
@@ -123,12 +122,9 @@ class DockerToActivationFileLogStore(system: ActorSystem, destinationDirectory: 
                            container: Container,
                            action: ExecutableWhiskAction): Future[ActivationLogs] = {
 
-    val logs = logStream(
-      transid,
-      container,
-      action.limits.logs,
-      action.exec.sentinelledLogs,
-      activation.response.isContainerError) // container error means developer error
+    val logLimit = action.limits.logs
+    val isDeveloperError = activation.response.isContainerError // container error means developer error
+    val logs = logStream(transid, container, logLimit, action.exec.sentinelledLogs, isDeveloperError)
 
     // Adding the userId field to every written record, so any background process can properly correlate.
     val userIdField = Map("namespaceId" -> user.namespace.uuid.toJson)
@@ -154,12 +150,8 @@ class DockerToActivationFileLogStore(system: ActorSystem, destinationDirectory: 
     val combined = OwSink.combine(toSeq, toFile)(Broadcast[ByteString](_))
 
     logs.runWith(combined)._1.flatMap { seq =>
-      // Messages.logWarningDeveloperError shows up if the activation ends in a developer error.
-      // This must not be reported as log collection error.
-      val possibleErrors = Set(Messages.logFailure, Messages.truncateLogs(action.limits.logs.asMegaBytes))
-      val errored = seq.lastOption.exists(last => possibleErrors.exists(last.contains))
       val logs = ActivationLogs(seq.toVector)
-      if (!errored) {
+      if (!isLogCollectingError(logs, logLimit, isDeveloperError)) {
         Future.successful(logs)
       } else {
         Future.failed(LogCollectingException(logs))
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationLogStore.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationLogStore.scala
@@ -18,6 +18,7 @@
 package org.apache.openwhisk.core.containerpool.logging
 
 import java.time.Instant
+
 import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.stream.ActorMaterializer
@@ -25,15 +26,13 @@ import akka.stream.scaladsl.Sink
 import akka.stream.scaladsl.Flow
 import akka.stream.scaladsl.Source
 import akka.util.ByteString
-
 import org.apache.openwhisk.common.TransactionId
 import org.apache.openwhisk.core.containerpool.Container
-import org.apache.openwhisk.core.entity.{ActivationLogs, ExecutableWhiskAction, Identity, WhiskActivation}
+import org.apache.openwhisk.core.entity.{ActivationLogs, ExecutableWhiskAction, Identity, LogLimit, WhiskActivation}
 import org.apache.openwhisk.http.Messages
 import org.apache.openwhisk.core.database.UserContext
 
 import scala.concurrent.{ExecutionContext, Future}
-
 import spray.json._
 
 /**
@@ -75,8 +74,19 @@ class DockerToActivationLogStore(system: ActorSystem) extends LogStore {
   /**
    * Obtains the container's stdout and stderr output.
    *
-   * In case of a timed out activation do not wait for a sentinel to appear but instead
-   * collect the log as is and add a message to the log that data might be missing
+   * Managed action runtimes are expected to produce sentinels on developer errors during
+   * init and run. For certain developer errors like process abortion due to unhandled errors
+   * or memory limit exhaustion, the action runtime will likely not be able to produce sentinels.
+   *
+   * In addition, there are situations where user actions (un)intentionally cause a developer error
+   * in a managed action runtime and prevent the production of sentinels. In that case, log file
+   * reading may continue endlessly.
+   *
+   * For these reasons, do not wait for sentinels to appear in log output when activations end up
+   * in a developer error. It is expected that sentinels are filtered in container.logs() even
+   * if they are not waited for.
+   *
+   * In case of a developer error, append a warning message to the logs that data might be missing.
    *
    * @param transid transaction id
    * @param container container to obtain the log from
@@ -87,16 +97,18 @@ class DockerToActivationLogStore(system: ActorSystem) extends LogStore {
    */
   protected def logStream(transid: TransactionId,
                           container: Container,
-                          action: ExecutableWhiskAction,
-                          isTimedoutActivation: Boolean): Source[ByteString, Any] = {
+                          logLimit: LogLimit,
+                          sentinelledLogs: Boolean,
+                          isDeveloperError: Boolean): Source[ByteString, Any] = {
 
-    // wait for a sentinel only if no container (developer) error occurred to avoid
-    // that log collection continues if the action code still logs after timeout
-    val sentinel = action.exec.sentinelledLogs && !isTimedoutActivation
-    val logs = container.logs(action.limits.logs.asMegaBytes, sentinel)(transid)
-    val logsWithPossibleError = if (isTimedoutActivation) {
+    // Wait for a sentinel only if no container (developer) error occurred to avoid
+    // that log collection continues if the action code still logs after developer error.
+    val waitForSentinel = sentinelledLogs && !isDeveloperError
+    val logs = container.logs(logLimit.asMegaBytes, waitForSentinel)(transid)
+    val logsWithPossibleError = if (isDeveloperError) {
       logs.concat(
-        Source.single(ByteString(LogLine(Instant.now.toString, "stderr", Messages.logFailure).toJson.compactPrint)))
+        Source.single(
+          ByteString(LogLine(Instant.now.toString, "stderr", Messages.logWarningDeveloperError).toJson.compactPrint)))
     } else logs
     logsWithPossibleError
   }
@@ -107,15 +119,21 @@ class DockerToActivationLogStore(system: ActorSystem) extends LogStore {
                            container: Container,
                            action: ExecutableWhiskAction): Future[ActivationLogs] = {
 
-    val isTimedoutActivation = activation.isTimedoutActivation
-    val logs = logStream(transid, container, action, isTimedoutActivation)
+    val logs = logStream(
+      transid,
+      container,
+      action.limits.logs,
+      action.exec.sentinelledLogs,
+      activation.response.isContainerError) // container error means developer error
 
     logs
       .via(DockerToActivationLogStore.toFormattedString)
       .runWith(Sink.seq)
       .flatMap { seq =>
+        // Messages.logWarningDeveloperError shows up if the activation ends in a developer error.
+        // This must not be reported as log collection error.
         val possibleErrors = Set(Messages.logFailure, Messages.truncateLogs(action.limits.logs.asMegaBytes))
-        val errored = isTimedoutActivation || seq.lastOption.exists(last => possibleErrors.exists(last.contains))
+        val errored = seq.lastOption.exists(last => possibleErrors.exists(last.contains))
         val logs = ActivationLogs(seq.toVector)
         if (!errored) {
           Future.successful(logs)
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationLogStore.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationLogStore.scala
@@ -88,6 +88,9 @@ class DockerToActivationLogStore(system: ActorSystem) extends LogStore {
    *
    * In case of a developer error, append a warning message to the logs that data might be missing.
    *
+   * TODO: instead of just appending a warning message when a developer error occurs, we should
+   *       have an out-of-band error handling that injects such messages later on.
+   *
    * @param transid transaction id
    * @param container container to obtain the log from
    * @param action action that defines the log limit
@@ -121,6 +124,12 @@ class DockerToActivationLogStore(system: ActorSystem) extends LogStore {
    * If the activation failed due to a developer error, an additional error message is appended.
    * In that case, the second last message indicates whether there was a log collecting error AND
    * the last message MUST be the additional error message mentioned above.
+   *
+   * TODO: this function needs to deal with different combinations of error / warning messages that
+   *       were appended to / injected into the log collecting stream.
+   *       Instead, we should have an out-of-band error handling that does not use log messages to
+   *       detect error conditions but detects errors and appends error / warning messages in
+   *       a different way.
    *
    * @param actLogs the activation logs to check
    * @param logLimit the log limit applying to the activation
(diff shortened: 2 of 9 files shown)


sven-lange-last:
## Assessment

* Following assessment lists all situations where a container / developer error may occur.
* This PR changes log collecting and processing for container / developer errors.
* While this PR makes log collecting and processing more robust when container / developer errors are detected, it has the downside that a few log lines may be lost.

## Conditions for container / developer error

### In `ContainerProxy`:

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala#L280

  * Application-specific error while starting a blackbox container.
  * Examples: image not found in registry or network failure while pulling.
  * The action container does not even start, no log reading is performed at all.

### In `Container`:

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/Container.scala#L140

  * Action timeout is exceeded during init.
  * Example: a Node.js action with a busy loop during init blocks the event loop.
  * Expectation: managed action runtimes do not produce sentinels on timeout during init.

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/Container.scala#L177

  * Action timeout is exceeded during run.
  * Example: action runs too long.
  * Expectation: managed action runtimes do not produce sentinels on timeout during run.

### In `ActivationResult` - `processInitResponseContent()`:

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala#L151

  * POST /init returns a response with status != OK (200), response is not truncated and is a valid JSON. We assume that the managed action runtime detected a problem during init and sent the response.
  * Example: Node.js action code that only contains comment => no valid entry point can be found.
  * Expectation: managed action runtimes should produce sentinels in this situation.

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala#L153

  * POST /init returns a response with status != OK (200), response is not truncated but is no valid JSON. We assume that the managed action runtime detected a problem during init and sent a broken response due to a bug.
  * Should not happen for managed runtimes because such failures should be caught in test. Blackbox containers may return improper responses. Can malicious actions cause sending of a broken response?
  * Example: no known examples.
  * Expectation: managed action runtimes should produce sentinels in this situation.

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala#L157

  * POST /init returns a response with status != OK (200) and response is truncated because it exceeds the size limit (1 MiB). We assume that the managed action runtime detected a problem during init and sent a broken response due to a bug.
  * Should not happen for managed runtimes because such failures should be caught in test. Blackbox containers may return improper responses. Can malicious actions cause sending of a broken response?
  * Example: no known examples.
  * Expectation: managed action runtimes should produce sentinels in this situation.

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala#L161

  * The TCP connection to the action container "fails" or breaks down during POST /init or while waiting for a response AND the OOM check is positive. We assume that the managed action runtime did not have a chance to react on this error situation.
  * Example: action code consumes too much physical memory by including a lot of prereq packages.
  * Expectation: managed action runtimes won't produce sentinels in this situation.

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala#L165

  * The TCP connection to the action container "fails" or breaks down during POST /init or while waiting for a response AND the OOM check is negative. We assume that the managed action runtime did not have a chance to react on this error situation.
  * Example: the action code breaks the managed action runtime process by exiting.
  * Expectation: managed action runtimes won't produce sentinels in this situation. Malicious actions may be able to delay / prevent sentinel production.

### In `ActivationResult` - `processRunResponseContent()`:

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala#L197

  * POST /run returns a response with status != OK (200), response is not truncated and is a valid JSON. We assume that the managed action runtime detected a problem during / after run and sent the response.
  * Example: action runtime caught an exception thrown by action code.
  * Expectation: managed action runtimes should produce sentinels in this situation.

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala#L202

  * POST /run returns a response with status != OK (200), response is not truncated and is a valid JSON - but JSON is not an object (instead: String, Number, ...). We assume that the managed action runtime detected a problem during / after run and and sent a broken response due to a bug.
  * Should not happen for managed runtimes because such failures should be caught in test. Blackbox containers may return improper responses. Can malicious actions cause sending of a broken response?
  * Example: no known examples.
  * Expectation: managed action runtimes should produce sentinels in this situation.

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala#L207

  * POST /run returns a response with status != OK (200), response is not truncated but is no valid JSON. We assume that the managed action runtime detected a problem during / after run and and sent a broken response due to a bug.
  * Should not happen for managed runtimes because such failures should be caught in test. Blackbox containers may return improper responses. Can malicious actions cause sending of a broken response?
  * Example: no known examples.
  * Expectation: managed action runtimes should produce sentinels in this situation.

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala#L211

  * POST /run returns a response with status != OK (200) and response is truncated because it exceeds the size limit (1 MiB). We assume that the managed action runtime detected a problem during / after run and and sent a broken response due to a bug.
  * Should not happen for managed runtimes because such failures should be caught in test. Blackbox containers may return improper responses. Can malicious actions cause sending of a broken response?
  * Example: no known examples.
  * Expectation: managed action runtimes should produce sentinels in this situation.

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala#L215

  * The TCP connection to the action container "fails" or breaks down during POST /run or while waiting for a response AND the OOM check is positive. We assume that the managed action runtime did not have a chance to react on this error situation.
  * Example: action code consumes too much physical memory by creating too many objects.
  * Expectation: managed action runtimes won't produce sentinels in this situation.

* https://github.com/apache/openwhisk/blob/9bef49fcd47f7922e4226ff7a97385f1313f7d64/common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala#L219

  * The TCP connection to the action container "fails" or breaks down during POST /run or while waiting for a response AND the OOM check is negative. We assume that the managed action runtime did not have a chance to react on this error situation.
  * Example: the action code breaks the managed action runtime process by exiting.
  * Expectation: managed action runtimes won't produce sentinels in this situation. Malicious actions may be able to delay / prevent sentinel production.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -257,17 +257,43 @@ class DockerContainer(protected val id: ContainerId,
    * previous activations that have to be skipped. For this reason, a starting position
    * is kept and updated upon each invocation.
    *
-   * If asked, check for sentinel markers - but exclude the identified markers from
-   * the result returned from this method.
+   * There are two possible modes controlled by parameter waitForSentinel:
    *
-   * Only parses and returns as much logs as fit in the passed log limit.
+   * 1. Wait for sentinel: tail container log file until two sentinel markers show up. Complete
```
markusthoemmes: Please break the description to the next line. Indent by one or two spaces if you feel the necessity there.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationLogStore.scala
```diff
@@ -87,16 +97,18 @@ class DockerToActivationLogStore(system: ActorSystem) extends LogStore {
...
+
+    // Wait for a sentinel only if no container (developer) error occurred to avoid
+    // that log collection continues if the action code still logs after developer error.
+    val waitForSentinel = sentinelledLogs && !isDeveloperError
+    val logs = container.logs(logLimit.asMegaBytes, waitForSentinel)(transid)
+    val logsWithPossibleError = if (isDeveloperError) {
       logs.concat(
-        Source.single(ByteString(LogLine(Instant.now.toString, "stderr", Messages.logFailure).toJson.compactPrint)))
+        Source.single(
+          ByteString(LogLine(Instant.now.toString, "stderr", Messages.logWarningDeveloperError).toJson.compactPrint)))
```
markusthoemmes: I'm assuming the log will end with both a logging error and this error, is that correct? If so: maybe we should drop the TimeoutException from the recover below and add that to the the stream here. We can then decide if we want to append the general log failure or this new one.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -257,17 +257,43 @@ class DockerContainer(protected val id: ContainerId,
...
    * @param limit the limit to apply to the log size
    * @param waitForSentinel determines if the processor should wait for a sentinel to appear
-   *
    * @return a vector of Strings with log lines in our own JSON format
    */
   def logs(limit: ByteSize, waitForSentinel: Boolean)(implicit transid: TransactionId): Source[ByteString, Any] = {
+    // Define a time limit for collecting and processing the logs of a single activation.
+    // If this time limit is exceeded, log processing is stopped and declared unsuccessful.
+    // Allow 2 seconds per MB log limit, with a minimum of 10 seconds.
+    val logProcessingTimeout = limit.toMB.toInt.max(5) * 2.seconds
```
markusthoemmes: This is somewhat arbitrary and might need some tweaking. Can you run this code through a loadtest to determine we don't hit this overall timeout under load too often?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala
sven-lange-last: Makes sense. Re-formatting as suggested.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala
```diff
@@ -281,9 +309,11 @@ class DockerContainer(protected val id: ContainerId,
       }
       .via(new CompleteAfterOccurrences(_.containsSlice(DockerContainer.byteStringSentinel), 2, waitForSentinel))
       // As we're reading the logs after the activation has finished the invariant is that all loglines are already
-      // written and we mostly await them being flushed by the docker daemon. Therefore we can timeout based on the time
+      // written and we mostly await them being flushed by the docker daemon. Therefore we can time out based on the time
       // between two loglines appear without relying on the log frequency in the action itself.
       .idleTimeout(waitForLogs)
+      // Apply an overall time limit for this log collecting and processing stream.
+      .completionTimeout(logProcessingTimeout)
```
rabbah: It looks to me like this is the crux of the patch - limit the log extraction until sentinel is found. I don’t know what a reasonable rate should be but up to 2s/1MB strikes me as long. I like Markus’ suggestion and wonder what the drain rate might be under load. That’s especially when waiting is on the critical path.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala
sven-lange-last: Good point. I checked production histogram metrics for `invoker.collect_logs_finish` and the max value for around 20 million action container log reads is below 1.2 seconds. So I conclude that 20 seconds for a 10 MiB log limit should be a reasonable setting.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationLogStore.scala
sven-lange-last: > I'm assuming the log will end with both a logging error and this error, is that correct?

I think no, because:

* If we have a developer error, we won't perform tailing log file reading but only read to the file's current end. In addition, we do not require to find 2 sentinel markers.
* With these conditions, the likelihood to run into idle / completion timeouts should be low.
* All other errors (framing, log limit exceeded) could possibly occur - but there is no correlation to the developer error.

My understanding is that you are thinking about breaking ties when we have two different error conditions like developer error AND timeout / framing error / log limit exceeded?

From my point of view, there are three perspectives on that:

1. As action developer, I use the action logs to understand and debug problems with init'ing and running actions. For the action developer, it probably doesn't matter which error appears in the last log line or whether the two last log lines mention more specific error conditions (`There was an issue ...` and `The action did not initialize or run as expected...`) - as long as there is a hint what happened and that the log is probably truncated.

2. From a container proxy perspective, I want to remove (and not re-use) an action container that had a developer or log collecting and processing error. From that perspective, it doesn't matter which of these two error conditions arrives in the container proxy.

3. As operator, I do not want see indications (system log messages, metrics) for failed log collections if there is a developer error - but log collecting and processing works properly. Only if a timeout / framing error / log limit exceeded occurs, I want a log collecting failure. That's why I excluded `logWarningDeveloperError` as possible error message.

The current variant would not "see" any of the current error messages `There was an issue while collecting your logs. Data might be missing.` and `Logs were truncated because the total bytes size exceeds the limit of ${limit.toBytes} bytes.` in case of a developer error because the last message would always be `The action did not initialize or run as expected. Log data might be missing.`. From that perspective, it would be good append the developer error message after checking the last log line for log collecting / processing errors.

markusthoemmes: Thanks for the exhaustive writeup. My only concern was having multiple error messages in the log file. The missing piece in my head was that we don't tail in the non-sentinel case (I'm hitting my past self for this, I should have introduced a separate path for that which doesn't have the idleTimeout for instance 😂 )

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala
markusthoemmes: It's not really the crux but an additional safeguard as I read it. It's a "catchall".

The crux here is to ignore sentinels (and thus don't tail logs, which guarantees a finish of the stream) in error cases.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala
markusthoemmes: Per @rabbah's comment below, I'm fine with rather going longer than too short on this for now. It's only a safeguard anyway.

rabbah: nice @sven-lange-last thanks for checking.

sven-lange-last: @markusthoemmes @rabbah I explicitly want to check with you on the total timeout setting. Are you ok with the setting above (2s / MiB log limit, minimum 10s) - given that this is "just" a safeguard to address other currently unknown problems? As discussed, the developer error is mainly addressed by changing from sentinel-based tailing to non-sentinel reading to end-of-file without tailing.

markusthoemmes: Yes I am fine with that.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationLogStore.scala
sven-lange-last: @markusthoemmes I pushed a new commit that should properly deal with log collecting errors when there is a developer error. Please have a look. It's no "nice" code though...

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala
rabbah: See my comment above on why I thought/think this is the essence of the patch. Arguably this limit should have already existed when waiting for logs to drain. But I understand the combined effect as well now, thanks to the excellent comments.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
```diff
@@ -123,8 +122,9 @@ class DockerToActivationFileLogStore(system: ActorSystem, destinationDirectory:
                            container: Container,
                            action: ExecutableWhiskAction): Future[ActivationLogs] = {
 
-    val isTimedoutActivation = activation.isTimedoutActivation
-    val logs = logStream(transid, container, action, isTimedoutActivation)
+    val logLimit = action.limits.logs
+    val isDeveloperError = activation.response.isContainerError // container error means developer error
```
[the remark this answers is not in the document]
sven-lange-last: Yes, that's true - it will only read until the end of the log file and not tail until a sentinel log marker is found. And also for more conditions - see here: https://github.com/apache/openwhisk/pull/4582#issuecomment-520706643.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationLogStore.scala
```diff
@@ -75,8 +74,19 @@ class DockerToActivationLogStore(system: ActorSystem) extends LogStore {
...
-   * collect the log as is and add a message to the log that data might be missing
+   * Managed action runtimes are expected to produce sentinels on developer errors during
+   * init and run. For certain developer errors like process abortion due to unhandled errors
+   * or memory limit exhaustion, the action runtime will likely not be able to produce sentinels.
+   *
+   * In addition, there are situations where user actions (un)intentionally cause a developer error
+   * in a managed action runtime and prevent the production of sentinels. In that case, log file
+   * reading may continue endlessly.
+   *
+   * For these reasons, do not wait for sentinels to appear in log output when activations end up
```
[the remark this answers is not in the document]
sven-lange-last: My goal was to quickly complete log collecting if a developer error occurred. That's accomplished with not tailing the log file on developer error.

The total log collecting timeout was "just" meant as a safeguard for further, currently unknown flaws.

If we used a total log collecting timeout also for developer errors, we would give action containers some additional seconds of CPU time.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala
```diff
@@ -55,18 +55,19 @@ protected[core] object ActivationResponse extends DefaultJsonProtocol {
...
 
+  // These constants need to be synchronized with messageForCode() method below
   val Success = 0 // action ran successfully and produced a result
   val ApplicationError = 1 // action ran but there was an error and it was handled
   val DeveloperError = 2 // action ran but failed to handle an error, or action did not run and failed to initialize
   val WhiskError = 3 // internal system error
 
   protected[core] def messageForCode(code: Int) = {
-    require(code >= 0 && code <= 3)
+    require(code >= Success && code <= WhiskError)
```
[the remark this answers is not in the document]
sven-lange-last: I agree - that also came into my mind... But I was already happy to replace numbers with `val`s.

common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationLogStore.scala
```diff
@@ -113,29 +113,61 @@ class DockerToActivationLogStore(system: ActorSystem) extends LogStore {
...
+   * If the activation failed due to a developer error, an additional error message is appended.
+   * In that case, the second last message indicates whether there was a log collecting error AND
+   * the last message MUST be the additional error message mentioned above.
+   *
+   * @param actLogs the activation logs to check
+   * @param logLimit the log limit applying to the activation
+   * @param isDeveloperError did activation fail due to developer error?
+   * @return true if log collecting failed, false otherwise
+   */
+  protected def isLogCollectingError(actLogs: ActivationLogs,
```
markusthoemmes: The logic here kinda what I was fearing earlier 😂 . Should we at least leave a TODO to return errors from the log gathering itself and append the appropriate lines based on these after the fact?

sven-lange-last: Well... I feel ashamed about this piece of code... at the same time, I had the feeling that I would have to improve other areas as well.

For example: the log reading pipeline recovers exceptions with regular stream elements, i.e. log messages that we check for later on. If an action just logged one of these messages as last log line, we would detect this a log collecting error...

I tried a lot of different approaches and one of the problems was that the `DockerToActivationFileLogStore` has a fan-out to different sinks so that the "injection" of error messages needs to be performed at a reasonable stage of the pipeline. That probably requires a custom flow or sink.

I guess it's a good idea to leave a TODO in the comments.

sven-lange-last:
@markusthoemmes @rabbah I think I addressed your review comments. Please check and let me know.

@rabbah concerning your comment / request:

> I think this is a dilation factor we can move to the configuration parameters.

I didn't fully expose this factor as a configuration parameter yet because we were fine so far with also having other timeouts as `val`s local to the class. See https://github.com/apache/openwhisk/blob/b5b88ece0ae08c87948796b38ab18e55ab4b70a7/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/docker/DockerContainer.scala#L176-L178

* That's why I renamed one of the timeouts for more clarity and added `logCollectingTimeoutPerMBLogLimit` to this list.
* I suggest that we should expose all or none of these config parameters and only if there is a need to change these defaults. @rabbah okay to delay loading these config parameters or do you need config loading already now?

rabbah:
>  @rabbah okay to delay loading these config parameters or do you need config loading already now?

Yes, as I wrote, looked good to me already ;)

sven-lange-last:
@markusthoemmes please let me know whether anything is still required from your side.

@style95 please check whether you are ok.

I would like to get this PR merged if everybody is ok...

PR #4593, opened 2019-08-23
Change ContainerPool buffer processing 

tysonnorris:
Change ContainerPool buffer processing to be based on state change messages to reduce cpu burn.

## Description

Existing message processing in ContainerPool can get into loops of immediately resending Run messagings causing cpu burn and crashing in cases where:
- Run messages arrive but cannot be processed so begin queueing
- First message in queue cannot be processed, and is resent constantly

This change alters the buffer handling to:
- Not resend the same Run message during message processing
- Only resend the head of queue Run message when some other event occurs that may free resources: scheduled to container, completed run in container, container is removed

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerPool.scala
+++ b/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerPool.scala
@@ -73,6 +73,8 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
   // buffered here to keep order of computation.
   // Otherwise actions with small memory-limits could block actions with large memory limits.
   var runBuffer = immutable.Queue.empty[Run]
+  // Track the resent buffer head - so that we don't resend buffer head multiple times
+  var resent: Option[Run] = None
   val logMessageInterval = 10.seconds
   //periodically emit metrics (don't need to do this for each message!)
   context.system.scheduler.schedule(30.seconds, 10.seconds, self, EmitMetrics)
@@ -112,6 +114,10 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
       // next request to process
       // It is guaranteed, that only the first message on the buffer is resent.
       if (runBuffer.isEmpty || isResentFromBuffer) {
+        if (isResentFromBuffer) {
+          //remove from resent tracking - it may get resent again, or get processed
+          resent = None
+        }
         val createdContainer =
           // Is there enough space on the invoker for this action to be executed.
           if (hasPoolSpaceFor(busyPool, r.action.limits.memory.megabytes.MB)) {
@@ -184,7 +190,7 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
             // (and a new container would over commit the pool)
             val isErrorLogged = r.retryLogDeadline.map(_.isOverdue).getOrElse(true)
             val retryLogDeadline = if (isErrorLogged) {
-              logging.error(
+              logging.warn(
                 this,
                 s"Rescheduling Run message, too many message in the pool, " +
                   s"freePoolSize: ${freePool.size} containers and ${memoryConsumptionOf(freePool)} MB, " +
@@ -264,10 +270,18 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
 
   /** Resend next item in the buffer, or trigger next item in the feed, if no items in the buffer. */
   def processBufferOrFeed() = {
-    // If buffer has more items, send next one, otherwise get next from feed.
+    // If buffer has more items, and head has not already been resent, send next one, otherwise get next from feed.
     runBuffer.dequeueOption match {
       case Some((run, _)) => //run the first from buffer
-        self ! run
+        implicit val tid = run.msg.transid
+        //avoid sending dupes
+        if (resent.isEmpty) {
+          logging.info(this, s"re-processing from buffer (${runBuffer.length} items in buffer)")
+          resent = Some(run)
+          self ! run
+        } else {
+          //do not resend the buffer head multiple times (may reach this point from multiple messages, before the buffer head is re-processed)
+        }
       case None => //feed me!
         feed ! MessageFeed.Processed
     }
--- a/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerPool.scala
+++ b/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerPool.scala
@@ -107,17 +107,18 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
     case r: Run =>
       // Check if the message is resent from the buffer. Only the first message on the buffer can be resent.
       val isResentFromBuffer = runBuffer.nonEmpty && runBuffer.dequeueOption.exists(_._1.msg == r.msg)
-
       // Only process request, if there are no other requests waiting for free slots, or if the current request is the
       // next request to process
       // It is guaranteed, that only the first message on the buffer is resent.
       if (runBuffer.isEmpty || isResentFromBuffer) {
+        // First grab a warm container if possible
+        val warmContainer = ContainerPool
+          .schedule(r.action, r.msg.user.namespace.name, freePool)
         val createdContainer =
-          // Is there enough space on the invoker for this action to be executed.
-          if (hasPoolSpaceFor(busyPool, r.action.limits.memory.megabytes.MB)) {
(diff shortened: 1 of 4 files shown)


tysonnorris:
@sven-lange-last please review when you have a moment - thanks!

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -555,6 +588,7 @@ class ContainerProxy(factory: (TransactionId,
    *         added to the WhiskActivation
    */
   def initializeAndRun(container: Container, job: Run)(implicit tid: TransactionId): Future[WhiskActivation] = {
+
```
style95: Unnecessary newline?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -479,22 +497,37 @@ class ContainerProxy(factory: (TransactionId,
...
       false
     }
   }
 
+  /** Process buffered items up to the capacity of action concurrency config */
+  def processBuffer(action: ExecutableWhiskAction, newData: ContainerData) = {
+    //send as many buffered as possible
+    var available = action.limits.concurrency.maxConcurrent - activeCount
+    logging.info(this, s"resending up to ${available} from ${runBuffer.length} buffered jobs")
+    while (available > 0 && runBuffer.length > 0) {
```
style95: Small nit.
How about replacing `var` and `while` with something similar to this?

```scala
    val available = action.limits.concurrency.maxConcurrent - activeCount
    logging.info(this, s"resending up to ${available} from ${runBuffer.length} buffered jobs")
    1 to available foreach { _ =>
      runBuffer.dequeueOption match {
        case Some((run, q)) =>
          self ! run
          bufferProcessing = true
          runBuffer = q
        case _ =>
          context.parent ! NeedWork(newData)
      }
    }
```

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerPool.scala
```diff
@@ -199,10 +200,10 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
             }
             if (!isResentFromBuffer) {
               // Add this request to the buffer, as it is not there yet.
-              runBuffer = runBuffer.enqueue(r)
+              runBuffer = runBuffer.enqueue(Run(r.action, r.msg, retryLogDeadline))
             }
-            // As this request is the first one in the buffer, try again to execute it.
-            self ! Run(r.action, r.msg, retryLogDeadline)
+          // As this request is the first one in the buffer, try again to execute it.
+          //self ! Run(r.action, r.msg, retryLogDeadline)
```
style95: Unnecessary comment?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
tysonnorris: Yes this is nicer!

rabbah: nice @style95 👏

tysonnorris:
@rabbah nudge - when you get a chance, please review.

rabbah:
thanks for the nudge - on it.

rabbah:
I've gone through this change once. I think it's sound but want to make a second pass.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerPool.scala
```diff
@@ -199,10 +200,9 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
             }
             if (!isResentFromBuffer) {
               // Add this request to the buffer, as it is not there yet.
-              runBuffer = runBuffer.enqueue(r)
+              runBuffer = runBuffer.enqueue(Run(r.action, r.msg, retryLogDeadline))
             }
-            // As this request is the first one in the buffer, try again to execute it.
-            self ! Run(r.action, r.msg, retryLogDeadline)
+          //buffered items will be processed via processBufferOrFeed()
```
rabbah: can you confirm my understanding that this change is sound because the request will still be in the process run buffer?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerPool.scala
```diff
@@ -173,7 +173,8 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
               // from the buffer
               val (_, newBuffer) = runBuffer.dequeue
               runBuffer = newBuffer
-              runBuffer.dequeueOption.foreach { case (run, _) => self ! run }
```
rabbah: do you mind fixing this comment on line 170
            // Remove the action that get's executed now from the buffer and execute the next one afterwards.
to
            // Remove the action that was just executed from the buffer and execute the next one in the queue.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerPool.scala
tysonnorris: I'm seeing a problem where in some cases messages are processed twice. I will work that out and it may affect this. Will get back to you.

tysonnorris: Sorry thought you were asking something else. Yes, the runBuffer processing works by:
* when an activation does not get scheduled AND it was not already buffered previously, add it to the buffer
* when an activation does not get scheduled AND it was already buffered - do nothing (it remains at buffer head)
* when an activation gets scheduled to a container, if it was previously buffered, remove it from the buffer, and trigger the next buffered item to be resent
* when some container action occurs (NeedsWork/ContainerRemoved) this also triggers next buffered item to be resent - so we need to be careful about sending duplicates by tracking which item has been resent

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerPool.scala
tysonnorris: github is not showing this update 🤷‍♂

tysonnorris:
@rabbah nuther nudge 😄  Let me know if there are other questions?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerPool.scala
```diff
@@ -267,6 +268,25 @@ class ContainerPool(childFactory: ActorRefFactory => ActorRef,
...
+    runBuffer.dequeueOption match {
+      case Some((run, _)) => //run the first from buffer
+        implicit val tid = run.msg.transid
+        //avoid sending dupes
+        if (resent.isEmpty) {
+          logging.info(this, s"re-processing from buffer (${runBuffer.length} items in buffer)")
+          resent = Some(run)
+          self ! run
+        } else {
+          //do not resend the buffer head multiple times (may reach this point from multiple messages, before the buffer head is re-processed)
```
rabbah: here a message is dequeued from the buffer and seemingly dropped - shouldn't this message get put back in the the run buffer?

tysonnorris: Yeah this is confusing - dequeueOption does not alter runBuffer. So if a message is resent, it also remains at the head of buffer until https://github.com/apache/openwhisk/pull/4593/files/f84fa581a910d1d6042a26e5244674251e056fe5#diff-726b36b3ab8c7cff0b93dead84311839R181

This part (head of buffer is stuck there till it can be processed) has not changed, just way in which retries (`self ! run`) are attempted is much fewer. The hyperactive retries were likely also the reason for the "only log the rescheduling message once per 10s" logic, but I tried to leave that log handling in place for now.

tysonnorris:
@rabbah Anything else for this?

tysonnorris:
Agreed there is potential to do more in this area as there are some inherent mismatches between the kafka/buffering/ordering and concurrency support. Will merge this as is, and revisit other items in the future as time allows.

PR #4596, opened 2019-08-28
Allow individual field sets for activation_record and user_log

ruediger-maass:
Introduce individual field set for activation records in user logs.

## Description
We need to handle activation records (i.e., activation metadata) and user logs differently in activation logs. Therefore we need individual field sets for storing these entries in `ActivationFileStorage` that we use in our Activation Store SPI imlementation.

- Controller (for sequences)
- Invoker

## Types of changes
Small change with low risk: add one function `activationToFileExtended`in `org.apache.openwhisk.core.database.ActivationFileStorage` that is basically the same as `activationToFile`but allows to define different sets of additional fields for activation_record and user_log. The proposed change does not replace the existing function `activationToFile` but adds the desired new function in order to not change the interface of `ActivationFileStorage` that might be used by other Activation Store SPI imlementations out there.

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala
@@ -111,8 +111,15 @@ class ActivationFileStorage(logFilePrefix: String,
   def activationToFile(activation: WhiskActivation,
                        context: UserContext,
                        additionalFields: Map[String, JsValue] = Map.empty) = {
-    val transcribedLogs = transcribeLogs(activation, additionalFields)
-    val transcribedActivation = transcribeActivation(activation, additionalFields)
+    activationToFileExtended(activation, context, additionalFields, additionalFields)
+  }
+
+  def activationToFileExtended(activation: WhiskActivation,
+                               context: UserContext,
+                               additionalFieldsForLogs: Map[String, JsValue] = Map.empty,
+                               additionalFieldsForActivation: Map[String, JsValue] = Map.empty) = {
+    val transcribedLogs = transcribeLogs(activation, additionalFieldsForLogs)
+    val transcribedActivation = transcribeActivation(activation, additionalFieldsForActivation)
 
     // Write each log line to file and then write the activation metadata
     Source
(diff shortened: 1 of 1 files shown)


ruediger-maass:
@rabbah thank you for the hint - done. Next I time I'll do that right away.

common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala
```diff
@@ -111,8 +111,16 @@ class ActivationFileStorage(logFilePrefix: String,
...
-    val transcribedActivation = transcribeActivation(activation, additionalFields)
+    activationToFileExtended(activation, context, additionalFields, additionalFields)
+  }
+
+  // used by external ArtifactActivationStore SPI implementation
+  def activationToFileExtended(activation: WhiskActivation,
+                               context: UserContext,
+                               additionalFieldsForLogs: Map[String, JsValue] = Map.empty,
+                               additionalFieldsForActivation: Map[String, JsValue] = Map.empty) = {
+    val transcribedLogs = transcribeLogs(activation, additionalFieldsForLogs)
```
chetanmeh: Minor nit - As its a public method lets specify the return type explicitly. Which would be `Unit` in this case.

ruediger-maass: Sure, done.

ruediger-maass:
> Looks fine. Just to confirm my understanding - You have your own impl similar to `ArtifactWithFileStorageActivationStore` which would now start calling the new method and pass different additional fields for activation and logs

@chetanmeh - exactly, that is correct! Our system has different, very backend-specific requirements. Threrefore we override `ArtifactWithFileStorageActivationStore`.

common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala
ruediger-maass: ok, done

chetanmeh:
Thanks for addressing the feedback!

mhenke1:
I merged this PR now, since all concerns seemed to be addressed by Rüdiger.

PR #4604, opened 2019-09-03
Allow decision about action result inclusion in logs on a per call basis

ruediger-maass:
Add the possibility to decide whether results should be included in the action log on a per-call basis for each activation individually.

## Description
Action results may contain sensible data that should not be logged. There is a system configuration flag (`writeResultToFile`) that can be used to switch on or off the result inclusion in logs. But this is a global switch that holds for all activations. In our case we want to able to decide per activation whether or not the result should be included in the log. In our special case we want results to be included in case of errors (in other words, 'Does the result contain an error field?'). But also other decision logic may be applicable. 

- Controller (for sequences)
- Invoker

## Types of changes
Small, low risk change: add one optional, boolean parameter to the method `activationToFileExtended` of `org.apache.openwhisk.core.database.ActivationFileStorage` that determines the decision whether or not the result of the current activation should be included in the log. The default value of the new parameter is the value of the system configuration flag `writeResultToFile`. The method `activationToFile` is unchanged and works as before. The private method `transcribeActivation` now takes the flag from `activationToFileExtended`.

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala
@@ -91,9 +91,11 @@ class ActivationFileStorage(logFilePrefix: String,
       ByteString(s"${line.compactPrint}\n")
     }
 
-  private def transcribeActivation(activation: WhiskActivation, additionalFields: Map[String, JsValue]) = {
+  private def transcribeActivation(activation: WhiskActivation,
+                                   additionalFields: Map[String, JsValue],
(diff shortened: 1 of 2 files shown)


tysonnorris:
Should this be tested in `ArtifactWithFileStorageActivationStoreTests`?

ruediger-maass:
> Should this be tested in `ArtifactWithFileStorageActivationStoreTests`?

Yes, I added tests for the two states of the flag. Individual sets of additional fields for logs and activation are also tested. The default settings for the flag are indirectly tested by the existing previous tests (`activationToFile` calls `activationToFileExtended`).

PR #4607, opened 2019-09-05
Add asf.yaml to configure github metadata

dgrove-oss:
Experiment with new ASF feature of using a per-repo yaml
file to configure github metadata for a project.

See https://cwiki.apache.org/confluence/display/INFRA/.asf.yaml+features+for+git+repositories for what is supported.   

I suggest we experiment in the main repo to verify it works and get to a core set of tags.  Then we can replicate across all the repos & add additional per-repo tags as well


rabbah:
You beat me to it!

PR #4608, opened 2019-09-05
fix indent on labels in .asf.yaml

dgrove-oss:
See if this works; looks like the example yaml on cwiki
had the wrong indentation level for the labels...


Humbedooh:
FaaS is invalid as it contains uppercase characters. faas would work.

PR #4609, opened 2019-09-06
Add namespace field to activation log

jiangpengcheng:
Add **namespace** field to activation log while using `DockerToActivationFileLogStore`
 
## Description
Currently there is no **namespace** field  but a **namespaceId** in the activation log while save activation logs to a separate file, while I think add **namesapce** will be more convenient for users when they process the log file, such as using Logstash to send logs to different indices(e.g. `openwhisk-%{namespace}`) in ElasticSearch

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala
@@ -84,9 +84,11 @@ class ActivationFileStorage(logFilePrefix: String,
 
   private def transcribeLogs(activation: WhiskActivation, additionalFields: Map[String, JsValue]) =
     activation.logs.logs.map { log =>
-      val line = JsObject(
-        Map("type" -> "user_log".toJson) ++ Map("message" -> log.toJson) ++ Map(
-          "activationId" -> activation.activationId.toJson) ++ additionalFields)
+      val line =
+        JsObject(
+          Map("type" -> "user_log".toJson) ++ Map("message" -> log.toJson) ++ Map(
+            "activationId" -> activation.activationId.toJson) ++ Map(
+            "namespace" -> activation.namespace.asString.toJson) ++ additionalFields)
 
       ByteString(s"${line.compactPrint}\n")
     }
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala
@@ -87,8 +87,7 @@ class ActivationFileStorage(logFilePrefix: String,
       val line =
         JsObject(
           Map("type" -> "user_log".toJson) ++ Map("message" -> log.toJson) ++ Map(
-            "activationId" -> activation.activationId.toJson) ++ Map(
-            "namespace" -> activation.namespace.asString.toJson) ++ additionalFields)
+            "activationId" -> activation.activationId.toJson) ++ additionalFields)
 
       ByteString(s"${line.compactPrint}\n")
     }
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala
@@ -84,10 +84,9 @@ class ActivationFileStorage(logFilePrefix: String,
 
   private def transcribeLogs(activation: WhiskActivation, additionalFields: Map[String, JsValue]) =
     activation.logs.logs.map { log =>
-      val line =
-        JsObject(
-          Map("type" -> "user_log".toJson) ++ Map("message" -> log.toJson) ++ Map(
-            "activationId" -> activation.activationId.toJson) ++ additionalFields)
+      val line = JsObject(
+        Map("type" -> "user_log".toJson) ++ Map("message" -> log.toJson) ++ Map(
+          "activationId" -> activation.activationId.toJson) ++ additionalFields)
 
       ByteString(s"${line.compactPrint}\n")
     }
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/DockerToActivationFileLogStore.scala
@@ -131,7 +131,8 @@ class DockerToActivationFileLogStore(system: ActorSystem, destinationDirectory: 
 
     val additionalMetadata = Map(
       "activationId" -> activation.activationId.asString.toJson,
-      "action" -> action.fullyQualifiedName(false).asString.toJson) ++ userIdField
+      "action" -> action.fullyQualifiedName(false).asString.toJson,
+      "namespace" -> user.namespace.name.asString.toJson) ++ userIdField
 
     val augmentedActivation = JsObject(activation.toJson.fields ++ userIdField)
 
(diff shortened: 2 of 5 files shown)


rabbah:
LGTM but please send to dev list.

jiangpengcheng:
ok, I sent the email

jiangpengcheng:
reopen to trigger the travis check

sven-lange-last:
@jiangpengcheng it seems that Travis is frequently failing on this PR - I guess the system test part?

I restarted only the system test part for you because all other Travis tests have passed.

I also had the situation for a lot of my PRs that 3 of 4 Travis checks passed successfully - but only system test was failing. I wonder whether we have an interference between tests such that the system test will always / very likely fail if the other Travis tests are running as well...

sven-lange-last:
I'm also ok with the changes introduced by this PR.

At the same time, I see that the different classes for providing action container (aka. user) logs are diverging. We have two different strategies to persist user logs in a log store:

1. The `LogStore` implementation extracts user logs from the action container (https://github.com/apache/openwhisk/blob/master/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/LogStore.scala) and returns the extracted logs (`collectLogs`). As a side effect of the extraction, logs are forwarded to the supported log store. This is what `DockerToActivationFileLogStore` is doing.

2. The `LogStore` implementation extracts user logs from the action container (https://github.com/apache/openwhisk/blob/master/common/scala/src/main/scala/org/apache/openwhisk/core/containerpool/logging/LogStore.scala) and returns the extracted logs (`collectLogs`). There is no forwarding to a log store (example: `DockerToActivationLogStore`). The extracted logs are attached to the activation.

  The `ArtifactActivationStore` is responsible for persisting activations (https://github.com/apache/openwhisk/blob/master/common/scala/src/main/scala/org/apache/openwhisk/core/database/ArtifactActivationStore.scala). As a side effect of storing the activation, logs are transformed and forwarded to a log store. With this approach, parts of the activation itself are also forwarded to a log store. Example: `ArtifactWithFileStorageActivationStore`.

This pull request changes `DockerToActivationFileLogStore` while it does not cover `ArtifactWithFileStorageActivationStore` - see this code location:

https://github.com/apache/openwhisk/blob/d4d7ceb627573670f70993d4e31dcf2a6737c3f8/common/scala/src/main/scala/org/apache/openwhisk/core/database/ArtifactWithFileStorageActivationStore.scala#L58

This pull request should probably also cover `ArtifactWithFileStorageActivationStore`.

At the same time, we cannot expect that this pull request addresses the code duplication of diverging log processing approaches...

jiangpengcheng:
@sven-lange-last 
thanks for the review, I wasn't aware of the `ArtifactWithFileStorageActivationStore` before, and seems it does almost the same thing as the `DockerToActivationFileLogStore`

and about the checks, the result is strange because all cases are passed in my local env, I didn;t know the reason.

Since close and reopen PRs seems kind of ugly, I wonder whether it's ok to add some "comment" triggers to re-trigger travis build by contributors themself? e.g. "build system", "build unittest".

jiangpengcheng:
> Will a sequence activation have the same set of extra annotations?

@rabbah  sorry, I don't get it, what's the `annotations` do you mean?

rabbah:
Take a look at `makeSequenceActivation` in `SequenceActions`.

jiangpengcheng:
I think this PR doesn't related to any annotations but just add a `namespace` field while write `activations`/`activation logs` to a separate file, a "sequence activation" will not do that AFAIK

sven-lange-last:
@jiangpengcheng please give me some time to check your latest updates...

common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala
```diff
@@ -84,9 +84,11 @@ class ActivationFileStorage(logFilePrefix: String,
...
   private def transcribeLogs(activation: WhiskActivation, additionalFields: Map[String, JsValue]) =
     activation.logs.logs.map { log =>
-      val line = JsObject(
-        Map("type" -> "user_log".toJson) ++ Map("message" -> log.toJson) ++ Map(
-          "activationId" -> activation.activationId.toJson) ++ additionalFields)
+      val line =
+        JsObject(
+          Map("type" -> "user_log".toJson) ++ Map("message" -> log.toJson) ++ Map(
+            "activationId" -> activation.activationId.toJson) ++ Map(
+            "namespace" -> activation.namespace.asString.toJson) ++ additionalFields)
```
sven-lange-last: Thanks for attending to my feedback and also covering other log stores.

At the same time: can you please move the `namespace` extension over to `ArtifactWithFileStorageActivationStore`? This would be the equivalent class to `DockerToActivationFileLogStore` where you inject `namespace` via `additionalMetadata`.

`ActivationFileStorage` is just a helper that performs writing of logs and activations to a file. The log store implementations that make use of `ActivationFileStorage` should decide which additional metadata they want to add. `ArtifactWithFileStorageActivationStore` is such a log store.

See https://github.com/apache/openwhisk/blob/400a7915115576a363858788a6d080c389a80317/common/scala/src/main/scala/org/apache/openwhisk/core/database/ArtifactWithFileStorageActivationStore.scala#L58-L60

^^ I suggest to inject the additional `namespace` field here.

Hint: IBM is using the `ActivationFileStorage` in a private implementation for IBM's logging system that requires different fields unique to IBM that do not make sense in the openwhisk context. That's why we want to keep `ActivationFileStorage` as open as possible.

sven-lange-last: I guess you also need to update this line in the test to inject the `namespace` field:

https://github.com/apache/openwhisk/blob/37ae815892ae03908c4804c4f826760224f09eea/tests/src/test/scala/org/apache/openwhisk/core/database/ArtifactWithFileStorageActivationStoreTests.scala#L280

jiangpengcheng: but then activation will also has an extra "namespace" field, is that ok?
https://github.com/apache/openwhisk/blob/400a7915115576a363858788a6d080c389a80317/common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala#L116

sven-lange-last: Let's have a look at the code:

https://github.com/apache/openwhisk/blob/400a7915115576a363858788a6d080c389a80317/common/scala/src/main/scala/org/apache/openwhisk/core/database/ArtifactWithFileStorageActivationStore.scala#L55-L62

If you add the `namespace` field to the `additionalFields` map in line 58, it will be persisted to logs via `activationFileStorage.activationToFile()` in line 60.

The `additionalFields` map won't be considered when storing the activation to the database (line 61).

Do you come to the same conclusion?

jiangpengcheng: yes
what I mean is that when write activation to file, it will also has an additional field "namespace"(alghough this just override the original "namespace" filed of activation):
https://github.com/apache/openwhisk/blob/400a7915115576a363858788a6d080c389a80317/common/scala/src/main/scala/org/apache/openwhisk/core/database/ActivationFileStorage.scala#L113-L126

sven-lange-last: We clarified the open questions in a personal Slack communication and agreed on a solution.

jiangpengcheng:
oh, sorry, I missed this file in last commit

jiangpengcheng:
nothing to update here

Issue #4615, opened 2019-09-09
key 'kind' does not exist


jiangpengcheng:
There will be a `key 'kind' does not exist` error while a `generateFallbackActivation` is used as the `generateFallbackActivation` won't add the `kind` annotation to `WhiskActivation`:

```scala
  private def generateFallbackActivation(msg: ActivationMessage, response: ActivationResponse): WhiskActivation = {
    val now = Instant.now
    val causedBy = if (msg.causedBySequence) {
      Some(Parameters(WhiskActivation.causedByAnnotation, JsString(Exec.SEQUENCE)))
    } else None

    WhiskActivation(
      activationId = msg.activationId,
      namespace = msg.user.namespace.name.toPath,
      subject = msg.user.subject,
      cause = msg.cause,
      name = msg.action.name,
      version = msg.action.version.getOrElse(SemVer()),
      start = now,
      end = now,
      duration = Some(0),
      response = response,
      annotations = {
        Parameters(WhiskActivation.pathAnnotation, JsString(msg.action.asString)) ++ causedBy
      })
  }
```

but while send UserEvent, the `ActivationMessage.from` require the `kind` annotation:

```scala
  /** Constructs an "Activation" event from a WhiskActivation */
  def from(a: WhiskActivation): Try[Activation] = {
    for {
      // There are no sensible defaults for these fields, so they are required. They should always be there but there is
      // no static analysis to proof that so we're defensive here.
      fqn <- a.annotations.getAs[String](WhiskActivation.pathAnnotation)
      kind <- a.annotations.getAs[String](WhiskActivation.kindAnnotation)
    } yield {
      Activation(
        fqn,
        a.response.statusCode,
        a.duration.getOrElse(0),
        a.annotations.getAs[Long](WhiskActivation.waitTimeAnnotation).getOrElse(0),
        a.annotations.getAs[Long](WhiskActivation.initTimeAnnotation).getOrElse(0),
        kind,
        a.annotations.getAs[Boolean](WhiskActivation.conductorAnnotation).getOrElse(false),
        a.annotations
          .getAs[ActionLimits](WhiskActivation.limitsAnnotation)
          .map(_.memory.megabytes)
          .getOrElse(0),
        a.annotations.getAs[String](WhiskActivation.causedByAnnotation).toOption)
    }
  }
```

since the `kind` data is not easy to get when generate fallback activation, is it ok to make it optional?

sven-lange-last:
[...]

PR #4617, opened 2019-09-10
Add kind "unknown" to fallback activations

jiangpengcheng:
add kind "unknown" for fallbacked activations

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#4615)

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/entity/Exec.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/entity/Exec.scala
@@ -242,6 +242,9 @@ object Exec extends ArgNormalizer[Exec] with DefaultJsonProtocol {
   protected[core] val SEQUENCE = "sequence"
   protected[core] val BLACKBOX = "blackbox"
 
+  // This is for error cases while cannot get the `kind` of Exec
+  protected[core] val UNKNOWN = "unknown"
+
   private def execManifests = ExecManifest.runtimesManifest
 
   override protected[core] implicit lazy val serdes = new RootJsonFormat[Exec] {
--- a/core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
+++ b/core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
@@ -290,7 +290,11 @@ class InvokerReactive(
       }
   }
 
-  /** Generates an activation with zero runtime. Usually used for error cases */
+  /**
+   * Generates an activation with zero runtime. Usually used for error cases.
+   *
+   * Set the kind annotation to `Exec.UNKNOWN` as user metric requires it
+   */
   private def generateFallbackActivation(msg: ActivationMessage, response: ActivationResponse): WhiskActivation = {
     val now = Instant.now
     val causedBy = if (msg.causedBySequence) {
@@ -309,7 +313,8 @@ class InvokerReactive(
       duration = Some(0),
       response = response,
       annotations = {
-        Parameters(WhiskActivation.pathAnnotation, JsString(msg.action.asString)) ++ causedBy
+        Parameters(WhiskActivation.pathAnnotation, JsString(msg.action.asString)) ++
+          Parameters(WhiskActivation.kindAnnotation, JsString(Exec.UNKNOWN)) ++ causedBy
       })
   }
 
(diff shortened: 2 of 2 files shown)


core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
```diff
@@ -263,7 +263,10 @@ class InvokerReactive(
                 }
 
                 val context = UserContext(msg.user)
-                val activation = generateFallbackActivation(msg, response)
+                val activation = generateFallbackActivation(
+                  msg,
+                  response,
+                  Parameters(WhiskActivation.kindAnnotation, JsString(Exec.UNKNOWN)))
```
rabbah: since it both cases of generating a fall back activation, the action is not available, i favor not changing the function sigature and adding the unknow annotation directly in `generateFallbackActivation` instead of at the two call sites.

sven-lange-last: @rabbah although I originally suggested the approach to extend the signature of `generateFallbackActivation()` to allow for passing additional annotations in #4615, I now agree with your assessment.

`generateFallbackActivation()` is used exactly two times within the same compilation unit in close proximity to its definition. So it's fine to add the `kind = unknown` annotation directly within `generateFallbackActivation()`.

@jiangpengcheng I suggest to add a comment to `generateFallbackActivation()` why the `kind = unknown` annotation is set, i.e. in order to support user metrics because the metrics record requires a kind.

rabbah:
> But all code locations that initially create the ActivationMessage know the action kind. So what about adding the action kind to the ActivationMessage when creating it?

I think the current change is fine as is. In the two cases affected by the change in InvokerReactive. The first case: `get` failed in the second: `namespace` is blacklisted.

What you're suggesting involves sending more information over the message bus. It's OK to do that but there is a cost to continuing to grown the activation message. Is it important enough to extend the ActivationMessage (serialization and deserialization costs) to cover two corner cases?

jiangpengcheng:
@sven-lange-last Never mind, it's totally fine

I also agree with @rabbah that add an extra field to `ActivationMessage` may introduce some costs which are not proportional to the gain

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
```diff
@@ -290,7 +290,11 @@ class InvokerReactive(
       }
   }
 
-  /** Generates an activation with zero runtime. Usually used for error cases */
+  /**
+   * Generates an activation with zero runtime. Usually used for error cases.
+   *
+   * Set the kind annotation to `Exec.UNKNOWN` as user metric requires it
```
rabbah:
```suggestion
   * Set the kind annotation to `Exec.UNKNOWN` since it is not known to the invoker because the action fetch failed.
```

sven-lange-last:
@rabbah @jiangpengcheng you are right - the cost of extending `ActivationMessage` (possibly coming with a performance penalty) is too high for addressing two corner cases. I'm also ok to go with the current proposal.

Do you want me to merge?

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
```diff
@@ -293,7 +293,7 @@ class InvokerReactive(
   /**
    * Generates an activation with zero runtime. Usually used for error cases.
    *
-   * Set the kind annotation to `Exec.UNKNOWN` as user metric requires it
+   * Set the kind annotation to `Exec.UNKNOWN` since it is not known to the invoker because the action fetch failed.
```
sven-lange-last: Thanks for improving the comment.

common/scala/src/main/scala/org/apache/openwhisk/core/entity/Exec.scala
```diff
@@ -242,6 +242,9 @@ object Exec extends ArgNormalizer[Exec] with DefaultJsonProtocol {
   protected[core] val SEQUENCE = "sequence"
   protected[core] val BLACKBOX = "blackbox"
 
+  // This is for error cases while cannot get the `kind` of Exec
```
rabbah:
```suggestion
  // This is for error cases when the action `kind` may not be known.
```

PR #4624, opened 2019-09-13
Combines active ack and slot release when both are available.

### LINKED ISSUE #4614: missing active ack from invoker
There are 5 calls to send a "ack" from an invoker (listed below) - 4 of them are completion messages and only 1 is an active ack with the activation result. The calls from `InvokerReactive` only send a completion message and no active acks. This is a bug.

Is there a reason not to send a combined completion/active ack instead of two messages when the result is ready?

https://github.com/apache/openwhisk/blob/b0d48c87816f2e7c04ad9b9f9b4844a6e064047f/core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala#L268

https://github.com/apache/openwhisk/blob/b0d48c87816f2e7c04ad9b9f9b4844a6e064047f/core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala#L278

https://github.com/apache/openwhisk/blob/b20dffbaf4e8d6ea90da5d8bd233a6b770fa74bb/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala#L289

https://github.com/apache/openwhisk/blob/b20dffbaf4e8d6ea90da5d8bd233a6b770fa74bb/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala#L624

https://github.com/apache/openwhisk/blob/b20dffbaf4e8d6ea90da5d8bd233a6b770fa74bb/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala#L662

### LINKED ISSUE #4636: Refactor active ack method signature to be an interface
In core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala:

>  type ActiveAck = (TransactionId, WhiskActivation, Boolean, ControllerInstanceId, UUID, Boolean) => Future[Any]

Minor observation (or Rant!) - I always struggle in IDE whenever I want to see impl of parameters which are specified as function signature instead of trait (like in ActiveAck). Here things are bit simpler as we specify a type alias which narrows down the search. Otherwise one need to check the whole call hierarchy to understand where is. the actual impl

Having a trait enables easier checking of possible implementations to understand the code flow. This is pre existing stuff ... but may be later we refactor it and use a proper trait for such critical flows.

From @chetanmeh in #4624.

rabbah:
Combines active ack and slot release when both are available. If completion message carries a result, process the active ack.

There are two commits in this patch. The first is semantic preserving/refactoring. The fix is in the second commit.

Closes #4614.
Closes #4636.

## Related issue and scope

- [x] I opened an issue to propose and discuss this change (#4614, #4636)

- [ ] API
- [ ] Controller
- [x] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
@@ -73,7 +73,7 @@ abstract class AcknowledegmentMessage(private val tid: TransactionId) extends Me
   override def serialize: String = AcknowledegmentMessage.serdes.write(this).compactPrint
 
   /** Pithy descriptor for logging. */
-  def name: String
+  def messageType: String
 
   /** Does message indicate slot is free? */
   def isSlotFree: Option[InvokerInstanceId]
@@ -99,23 +99,26 @@ abstract class AcknowledegmentMessage(private val tid: TransactionId) extends Me
 }
 
 /**
- * This message is sent from an invoker to the controller in situtations when the resource slot and the action
+ * This message is sent from an invoker to the controller in situations when the resource slot and the action
  * result are available at the same time, and so the split-phase notification is not necessary. Instead the message
  * combines the `CompletionMessage` and `ResultMessage`. The `response` may be an `ActivationId` to allow for failures
  * to send the activation result because of event-bus size limitations.
+ *
+ * The constructor is private so that callers must use the more restrictive constructors which ensure the respose is always
+ * Right when this message is created.
  */
-case class CombinedCompletionAndResultMessage(override val transid: TransactionId,
-                                              response: Either[ActivationId, WhiskActivation],
-                                              override val isSystemError: Option[Boolean],
-                                              invoker: InvokerInstanceId)
+case class CombinedCompletionAndResultMessage private (override val transid: TransactionId,
+                                                       response: Either[ActivationId, WhiskActivation],
+                                                       override val isSystemError: Option[Boolean],
+                                                       invoker: InvokerInstanceId)
     extends AcknowledegmentMessage(transid) {
-  override def name = "combined"
+  override def messageType = "combined"
   override def result = Some(response)
   override def isSlotFree = Some(invoker)
   override def activationId = response.fold(identity, _.activationId)
   override def toJson = CombinedCompletionAndResultMessage.serdes.write(this)
   override def shrink = copy(response = response.flatMap(a => Left(a.activationId)))
-  override def toString = response.fold(identity, _.activationId).asString
+  override def toString = activationId.asString
 }
 
 /**
@@ -124,12 +127,12 @@ case class CombinedCompletionAndResultMessage(override val transid: TransactionI
  * phase notification to the load balancer where an invoker first sends a `ResultMessage` and later sends the
  * `CompletionMessage`.
  */
-case class CompletionMessage(override val transid: TransactionId,
-                             override val activationId: ActivationId,
-                             override val isSystemError: Option[Boolean],
-                             invoker: InvokerInstanceId)
+protected case class CompletionMessage(override val transid: TransactionId,
+                                       override val activationId: ActivationId,
+                                       override val isSystemError: Option[Boolean],
+                                       invoker: InvokerInstanceId)
     extends AcknowledegmentMessage(transid) {
-  override def name = "completion"
+  override def messageType = "completion"
   override def result = None
   override def isSlotFree = Some(invoker)
   override def toJson = CompletionMessage.serdes.write(this)
@@ -141,10 +144,14 @@ case class CompletionMessage(override val transid: TransactionId,
  * This message is sent from an invoker to the load balancer once an action result is available for blocking actions.
  * This is part of a split phase notification, and does not indicate that the slot is available, which is indicated with
  * a `CompletionMessage`. Note that activation record will not contain any logs from the action execution, only the result.
+ *
+ * The constructor is private so that callers must use the more restrictive constructors which ensure the respose is always
+ * Right when this message is created.
  */
-case class ResultMessage(override val transid: TransactionId, response: Either[ActivationId, WhiskActivation])
+protected case class ResultMessage private (override val transid: TransactionId,
+                                            response: Either[ActivationId, WhiskActivation])
     extends AcknowledegmentMessage(transid) {
-  override def name = "result"
+  override def messageType = "result"
   override def result = Some(response)
   override def isSlotFree = None
   override def isSystemError = response.fold(_ => None, a => Some(a.response.isWhiskError))
@@ -162,11 +169,18 @@ object ActivationMessage extends DefaultJsonProtocol {
 }
 
 object CombinedCompletionAndResultMessage extends DefaultJsonProtocol {
+  // this constructor is restricted to ensure the message is always created with certain invariants
+  private def apply(transid: TransactionId,
+                    activation: Either[ActivationId, WhiskActivation],
+                    isSystemError: Option[Boolean],
+                    invoker: InvokerInstanceId): CombinedCompletionAndResultMessage =
+    new CombinedCompletionAndResultMessage(transid, activation, isSystemError, invoker)
+
   def apply(transid: TransactionId,
             activation: WhiskActivation,
-            invoker: InvokerInstanceId): CombinedCompletionAndResultMessage = {
-    CombinedCompletionAndResultMessage(transid, Right(activation), Some(activation.response.isWhiskError), invoker)
-  }
+            invoker: InvokerInstanceId): CombinedCompletionAndResultMessage =
+    new CombinedCompletionAndResultMessage(transid, Right(activation), Some(activation.response.isWhiskError), invoker)
+
   implicit private val eitherSerdes = AcknowledegmentMessage.eitherResponse
   implicit val serdes = jsonFormat4(
     CombinedCompletionAndResultMessage
@@ -177,13 +191,19 @@ object CompletionMessage extends DefaultJsonProtocol {
   def apply(transid: TransactionId, activation: WhiskActivation, invoker: InvokerInstanceId): CompletionMessage = {
     CompletionMessage(transid, activation.activationId, Some(activation.response.isWhiskError), invoker)
   }
+
   implicit val serdes = jsonFormat4(
     CompletionMessage.apply(_: TransactionId, _: ActivationId, _: Option[Boolean], _: InvokerInstanceId))
 }
 
 object ResultMessage extends DefaultJsonProtocol {
+  // this constructor is restricted to ensure the message is always created with certain invariants
+  private def apply(transid: TransactionId, response: Either[ActivationId, WhiskActivation]): ResultMessage =
+    new ResultMessage(transid, response)
+
   def apply(transid: TransactionId, activation: WhiskActivation): ResultMessage =
-    ResultMessage(transid, Right(activation))
+    new ResultMessage(transid, Right(activation))
+
   implicit private val eitherSerdes = AcknowledegmentMessage.eitherResponse
   implicit val serdes = jsonFormat2(ResultMessage.apply(_: TransactionId, _: Either[ActivationId, WhiskActivation]))
 }
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
@@ -231,19 +231,16 @@ object AcknowledegmentMessage extends DefaultJsonProtocol {
     // If this field is part of the JSON, we try to deserialize into one of these two types,
     // and otherwise to a ResultMessage. If all conversions fail, an error will be thrown that needs to be handled.
     override def read(json: JsValue): AcknowledegmentMessage = {
-      val obj = json.asJsObject
-
-      obj
-        .getFields("invoker")
-        .headOption
-        .map(_ => {
-          obj
-            .getFields("response")
-            .headOption
-            .map(_ => json.convertTo[CombinedCompletionAndResultMessage])
-            .getOrElse(json.convertTo[CompletionMessage])
-        })
-        .getOrElse(json.convertTo[ResultMessage])
+      val JsObject(fields) = json
+      val completion = fields.contains("invoker")
+      val result = fields.contains("response")
+      if (completion && result) {
+        json.convertTo[CombinedCompletionAndResultMessage]
+      } else if (completion) {
+        json.convertTo[CompletionMessage]
+      } else {
+        json.convertTo[ResultMessage]
+      }
     }
   }
 }
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
@@ -187,7 +187,7 @@ trait EventMessageBody extends Message {
 
 object EventMessageBody extends DefaultJsonProtocol {
 
-  implicit def format = new JsonFormat[EventMessageBody] {
+  implicit val format = new JsonFormat[EventMessageBody] {
     def write(eventMessageBody: EventMessageBody) = eventMessageBody match {
       case m: Metric     => m.toJson
       case a: Activation => a.toJson
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
@@ -71,56 +71,87 @@ case class ActivationMessage(override val transid: TransactionId,
 abstract class AcknowledegmentMessage(private val tid: TransactionId) extends Message {
   override val transid: TransactionId = tid
   override def serialize: String = AcknowledegmentMessage.serdes.write(this).compactPrint
+
+  /** Pithy descriptor for logging. */
+  def name: String
+
+  /** Does message indicate slot is free? */
+  def isSlotFree: Option[InvokerInstanceId]
+
+  /** Does message contain a result? */
+  def result: Option[Either[ActivationId, WhiskActivation]]
+
+  /**
+   * Is the acknowledgement for an activation that failed internally?
+   * For some message, this is not relevant and the result is None.
+   */
+  def isSystemError: Option[Boolean]
+
+  def activationId: ActivationId
+
+  /** Serializes the message to JSON. */
   def toJson: JsValue
+
+  /**
+   * Converts the message to a more compact form if it cannot cross the message bus as is or some of its details are not necessary.
+   */
+  def shrink: AcknowledegmentMessage
+}
+
+/**
+ * This message is sent from an invoker to the controller in situtations when the resource slot and the action
+ * result are available at the same time, and so the split-phase notification is not necessary. Instead the message
+ * combines the `CompletionMessage` and `ResultMessage`. The `response` may be an `ActivationId` to allow for failures
+ * to send the activation result because of event-bus size limitations.
+ */
+case class CombinedCompletionAndResultMessage(override val transid: TransactionId,
+                                              response: Either[ActivationId, WhiskActivation],
+                                              override val isSystemError: Option[Boolean],
+                                              invoker: InvokerInstanceId)
+    extends AcknowledegmentMessage(transid) {
+  override def name = "combined"
+  override def result = Some(response)
+  override def isSlotFree = Some(invoker)
+  override def activationId = response.fold(identity, _.activationId)
+  override def toJson = CombinedCompletionAndResultMessage.serdes.write(this)
+  override def shrink = copy(response = response.flatMap(a => Left(a.activationId)))
+  override def toString = response.fold(identity, _.activationId).asString
 }
 
 /**
  * This message is sent from an invoker to the controller, once the resource slot in the invoker (used by the
- * corresponding activation) free again (i.e., after log collection). In some cases, the activation result is
- * ready and the slot is freed at the same time. In such cases, the completion message carries the result as well.
- * This is reflected by the presence of a Right() `response` and the param `result` is set to true.
- * In some cases the `result` is true but the response is Left() if the message was too large for the message bus.
- *
- * There are four possible combinations as described below.
- *
- * - `result` is false and `response` is
- *
- *    1. Left: this is a split phase notification (slot released, active ack sent separately).
- *
- *    2. Right: should not happen and treated the same as Left.
- *
- * - `result` is true and `response` is
- *
- *    3. Right: this combines the active ack and slot release and occurs during system generated
- *               activations for failure scenarios.
- *
- *    4. Left: like the previous case but occurs when the message is too large to cross the event
- *              bus and the sender converts a Right to a Left.
+ * corresponding activation) free again (i.e., after log collection). The `CompletionMessage` is part of a split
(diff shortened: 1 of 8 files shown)


common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
```diff
@@ -64,43 +64,34 @@ case class ActivationMessage(override val transid: TransactionId,
...
+ * corresponding activation, is free again (i.e., after log collection). In some cases, the activation result is
+ * ready and the slot is freed at the same time. In such cases, the completion message carries the result as well.
+ * This is reflected by the of a Right() `response` and the param `result` is set to true.
+ * In some cases the `result` is true but the response is Left() if the message was too large for the message bus.
  */
 case class CompletionMessage(override val transid: TransactionId,
-                             activationId: ActivationId,
-                             isSystemError: Boolean,
+                             response: Either[ActivationId, WhiskActivation],
+                             result: Boolean, // true iff the message is a combined active ack and slot released
```
tysonnorris: Is this almost the same as:
``` response: Option[Either[ActivationId,WhiskActivation]] ```
? I'm not sure that is more or less clear, but want to make sure I'm not missing something.
In other words - is there any case where result==false, and response==Right[WhiskActivation]?

Comment (rabbah) [reply common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala]: When `result` is false:
1. `Left`: this is the current situation and can occur during split phase notification
1. `Right`: cannot happen, would be a bug (this should be an active ack instead)

When `result` is true:
1. `Right`: this combines the active ack and slot release and occurs during system generate activations for failure scenarios
2. `Left`: like the previous case but occurs when the message is too large to cross the event bus and the sender converts the Right to a Left.

Since there are 3 possible scenarios, `Either` by itself is not enough, and so I chose to encode the extra bit with `result` as a boolean.

Comment (rabbah) [reply common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala]: I don't think `response` as option works - there is always `Some` value (the current case is the `ActivationId`). We need a tri-valued typed. 

An alternative would be `Either[ActivationId, Either[ActivationId, WhiskActivation]` where `Left` represents the split phase and `Right(ActivationId)` and `Right(WhiskActivation)` represents the other two cases.

We can add a `require` in the constructor to also prevent the impossible case.

Comment (rabbah) [reply common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala]: @tysonnorris I thought about this further - another option is to introduce a third type. 

1. Active Ack. (as today)
2. Split phase completion. (what Completion message is today, "result is false" above case 1)
3. Combined Completion. (combines active ack and split phase into one message as this PR does, "result is true" above cases 3 and 4).

There would be 1 active ack + 1 split phase (from the container proxy) and 3 combined completions otherwise (for the error cases).

Comment (chetanmeh) [reply common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala]: Would this change cause some compatibility issue when there would be a gradual roll out of new invokers in an existing cluster.

So for we do not have any comparability contract for messages exchange via Kafka

Comment (tysonnorris) [reply common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala]: @rabbah I think it's fine. I don't have a real opinion on any of these options being more clear than the others.  

@chetanmeh I think that is true, but I'm not sure it has been considered for any Message impl changes in the past either? I'm not sure this is explicitly stated anywhere, but I guess it should be, as part of a "system upgrade" documentation?

Comment (rabbah) [reply common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala]: @chetanmeh yes it would since I change one of the property names and the value of the property in the cases now covered by the change may be different (in the corner case, WhiskActivation instead of just ActivationId).

Comment (rabbah) [reply common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala]: > part of a "system upgrade" documentation?

this project doesn't have any yet - are you proposing we start one? ;)

If we introduce a third type, then new invokers could send messages for the bug fix, but an old controller would still ignore them (in fact it would be worse since then the slot release is not recorded)... We could send 2 messages to replicate the split phase notification, but this is wasteful IMO, but then avoid the compatibility issue for a rolling update.

Do you want this PR to address this?

Comment (sven-lange-last) [reply common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala]: @rabbah I do not have the requirement to support rolling updates at the moment. From my side, we can introduce incompatible Kafka message changes requiring that all controllers and invokers need to be updated in parallel.

File: common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala [hunk 2]
```diff
@@ -112,49 +119,51 @@ object CompletionMessage extends DefaultJsonProtocol {
...
-      def read(value: JsValue) = value match {
-        // per the ActivationId's serializer, it is guaranteed to be a String even if it only consists of digits
-        case _: JsString => Left(value.convertTo[ActivationId])
-        case _: JsObject => Right(value.convertTo[WhiskActivation])
-        case _           => deserializationError("could not read ResultMessage")
-      }
-    }
+  private implicit val fqnSerdes = FullyQualifiedEntityName.serdes
+  implicit val serdes = jsonFormat11(ActivationMessage.apply)
+}
```
sven-lange-last: As far as I can see in the diff, you moved the JSON (de-)serialization of `ActivationMessage` down in the source file. I would suggest to keep these JSON (de-)serialization details in close proximity to the case class. I guess `jsonFormat11(ActivationMessage.apply)` won't fail during compile if you add more fields to the case class?

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
sven-lange-last: Based on the discussion above, I would say that three different messages sound like the best option. From a data modelling perspective, it's asymmetric that a `Right( WhiskActivation)` "should not happen" in the `CompletionMessage` if `result = false`.

With the split of the active ack into result / completion ack, we somewhat started a renaming. So when introducing a concept with three messages, I suggest to further follow the new naming and use something like `ResultMessage`, `CompletionMessage` and `CombinedResultCompletionMessage`. To preserve consistency, I would update all comments touched by this PR and replace "active ack" with the new terms.

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
rabbah: Are you referring to the Objects? I found the interleaving problematic and found unnecessary duplication (which I removed in this patch). I suspect the duplication was caused by not actually seeing that the code was very similar because of intervening functions. Now the objects are much more compact. So I like the current flow better of cases classes, then the singletons.

If you change the type of the activation message into something incompatible the code will not compile. `jsonFormatN` is an auto generated serdes and so if the compiler doesn't know what to do, it will complain.

sven-lange-last: Yes, I'm talking about the companion objects. My concern was that developers may miss to update the JSON (de-)serialization of case classes if case classes and their companion objects are at different locations of the source file.

If the compiler would notice mismatches, I have no problem with what you did.

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
rabbah: the `ack` method will have to change to support this. This is the type of the ack method:
```
  type ActiveAck = (TransactionId, WhiskActivation, Boolean, ControllerInstanceId, UUID, Boolean) => Future[Any]
```

The last parameter in this function indicates if the `slot` is free or not. So it is ambiguous to know inside the method if the result is combined or not from this signature alone since every call already receives a WhiskActivation.

So I'm OK with making the change to the message types (2 to 3) but note that the change will cause further refactoring in the invoker code.

sven-lange-last: @rabbah thanks for your openness to adjust your work to our feedback.

chetanmeh: @rabbah As you change the current design I would also like to highlight a potential change I need for my activation persister service work

Per my current understand the `sendActiveAck` flow is like

- `isSlotFree` false - active ack/ResultMessage
    - blocking - `WhiskActivation`
    - non-blocking - NONE
- `isSlotFree` true  - result/CompletionMessage
    - blocking - `CompletionMessage`
    - non-blocking - `CompletionMessage`

Later I would like to have a configurable support for sending `WhiskActivation` for non blocking call as well to a custom topic. Just wanted to make you aware on that

rabbah: @chetanmeh I pushed a commit to implement the new type discussed earlier. I did not see your comment earlier. Do the changes help? I think it makes clearer where split-phase acks are being used and where they aren't.

core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/CommonLoadBalancer.scala
```diff
@@ -204,17 +204,20 @@ abstract class CommonLoadBalancer(config: WhiskConfig,
...
+            acknowledegment.activationId,
+            acknowledegment.transid,
+            forced = false,
+            isSystemError = acknowledegment.isSystemError.getOrElse(false),
+            invoker)
+        }
+
+        acknowledegment.result.foreach { response =>
+          processResult(acknowledegment.activationId, acknowledegment.transid, response)
+        }
```
sven-lange-last: Pretty elegant way to fold all three types of message into a single flow because `acknowledegment.isSlotFree` and `acknowledegment.isSlotFree` are options.

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
```diff
@@ -64,100 +64,165 @@ case class ActivationMessage(override val transid: TransactionId,
...
-      // The field invoker is only part of the CompletionMessage. If this field is part of the JSON, we try to convert
-      // it to a CompletionMessage. Otherwise to a ResultMessage.
-      // If both conversions fail, an error will be thrown that needs to be handled.
+      val obj = json.asJsObject
+
+      obj
         .getFields("invoker")
         .headOption
-        .map(_ => json.convertTo[CompletionMessage])
+        .map(_ => {
```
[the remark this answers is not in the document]
sven-lange-last: From a code maintainability and data modelling perspective, it would be great to also serialise a `type` field that can be used later on to easily de-serialise the JSON. At the same time, this extension increases message size.

I guess the increased size does not really matter whenever a `WhiskActivation` is embedded because a `type` field is pretty small compared to a `WhiskActivation`.

I think the increased size would only matter for the `CompletionMessage`. Today, small JSONs for `CompletionMessage` would look like (146 Byte in compact form):

```
{"transid":"5808a97c269295220a6d8e78508f118b","activationId":"0f3763366aba46a0b763366abae6a0a4","invoker":{"instance":0,"userMemory":"16384 MB"}}
```

When adding a `type` field, we get (166 Byte in compact form):

```
{"type":"completion","transid":"5808a97c269295220a6d8e78508f118b","activationId":"0f3763366aba46a0b763366abae6a0a4","invoker":{"instance":0,"userMemory":"16384 MB"}}
```

This is a size increase of around 14% - which could be compensated by refactoring user memory handling such that it's no more serialised in the `CompletionMessage` - but only in pings.

So I support @chetanmeh's proposal.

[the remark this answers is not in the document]
chetanmeh: We can reduce the size bit further by using `{"t":1}` if needed i.e. encode it as enum. However as overhead is not that big we can keep the more readable string form

> refactoring user memory handling such that it's no more serialised in the CompletionMessage - but only in pings. -> `"invoker":{"instance":0,"userMemory":"16384 MB"}`

Would be good to have a issue for this.

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
```diff
@@ -145,43 +146,34 @@ class InvokerReactive(
...
   })
 
-  /** Sends an active-ack. */
   private val ack: InvokerReactive.ActiveAck = (tid: TransactionId,
                                                 activationResult: WhiskActivation,
                                                 blockingInvoke: Boolean,
                                                 controllerInstance: ControllerInstanceId,
                                                 userId: UUID,
-                                                isSlotFree: Boolean) => {
+                                                acknowledegment: AcknowledegmentMessage) => {
```
sven-lange-last: If the passed `acknowledegment` is a combined completion and result message or result message, it likely contains a `WhiskActivation` instance. So it duplicates the `activationResult` parameter which is only used to emit user metrics.

Can't we drop the `activationResult` parameter and change the condition for sending user metrics? If the `acknowledegment.result` is non-empty and a `Right(WhiskActivation)`, we emit metrics - otherwise not.

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
```diff
@@ -64,100 +64,165 @@ case class ActivationMessage(override val transid: TransactionId,
...
+  /**
+   * Converts the message to a more compact form if it cannot cross the message bus as is or some of its details are not necessary.
+   */
+  def shrink: AcknowledegmentMessage
 }
 
 /**
- * This message is sent from the invoker to the controller, after the slot of an invoker that has been used by the
- * current action, is free again (after log collection)
+ * This message is sent from an invoker to the controller in situtations when the resource slot and the action
```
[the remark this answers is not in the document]
rabbah: that's not a nit - that's a typo :D

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
```diff
@@ -64,100 +64,165 @@ case class ActivationMessage(override val transid: TransactionId,
...
 /**
- * That message will be sent from the invoker to the controller after action completion if the user wants to have
- * the result immediately (blocking activation).
- * When adding fields, the serdes of the companion object must be updated also.
- * The whisk activation field will have its logs stripped.
+ * This message is sent from an invoker to the load balancer once an action result is available for blocking actions.
+ * This is part of a split phase notification, and does not indicate that the slot is available, which is indicated with
+ * a `CompletionMessage`. Note that activation record will not contain any logs from the action execution, only the result.
  */
 case class ResultMessage(override val transid: TransactionId, response: Either[ActivationId, WhiskActivation])
```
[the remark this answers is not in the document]
rabbah: A result message should always contain a Right **except** when it must be shrunk because the message bus size limit on a message doesn't permit it to cross. So unless shrink is going to generate yet another type, I don't think we can remove the `Either` signature.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -628,8 +637,15 @@ class ContainerProxy(
...
-      activation.map(
-        sendActiveAck(tid, _, job.msg.blocking, job.msg.rootControllerIndex, job.msg.user.namespace.uuid, false))
+      activation.map { result =>
+        sendActiveAck(
+          tid,
+          result,
+          job.msg.blocking,
+          job.msg.rootControllerIndex,
+          job.msg.user.namespace.uuid,
+          ResultMessage(tid, result))
```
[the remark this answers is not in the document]
rabbah: good idea - I won't do that here, as you suggest, can be done subsequently.

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
```diff
@@ -54,9 +54,10 @@ object InvokerReactive extends InvokerProvider {
    * @param Boolean is true iff the activation was a blocking request
    * @param ControllerInstanceId the originating controller/loadbalancer id
    * @param UUID is the UUID for the namespace owning the activation
-   * @param Boolean is true this is resource free message and false if this is a result forwarding message
+   * @param AcknowledegmentMessage the acknowledgement message to send
    */
-  type ActiveAck = (TransactionId, WhiskActivation, Boolean, ControllerInstanceId, UUID, Boolean) => Future[Any]
+  type ActiveAck =
```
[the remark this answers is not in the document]
rabbah: echo rant.

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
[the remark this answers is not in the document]
rabbah: I considered these - as well making sure the field names were unique so that we can disambiguate with one field lookup. Will think about it again.

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
rabbah: yes - i considered doing that but because the current types allow for an `Either`, whereas all the call sites do know the activation, it would allow for a weakening should someone construct an ack with a Left where a Right is expected. I don't think there's a way to restrict the constructors (given they are case classes) without adding more types. So I think the current signature is tighter.

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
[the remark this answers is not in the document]
rabbah: This ends up being a bit more extensive than at first appearances so we should defer it to a subsequent patch.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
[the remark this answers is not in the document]
chetanmeh: Opened #4635 to track this

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
```diff
@@ -145,43 +146,38 @@ class InvokerReactive(
...
-        case Failure(t)   => logging.error(this, s"activation event was not sent: $t")
+    if (UserEvents.enabled && acknowledegment.isSlotFree.nonEmpty) {
+      acknowledegment.result.foreach {
+        case Right(activationResult) =>
+          EventMessage.from(activationResult, s"invoker${instance.instance}", userId) match {
+            case Success(msg) => UserEvents.send(producer, msg)
+            case Failure(t)   => logging.error(this, s"activation event was not sent: $t")
+          }
+        case _ => // all acknowledegment messages should have a result
       }
```
rabbah: @sven-lange-last this is an alternate implementation but does not remove the `activation` parameter which turns out to be more extensive refactoring. We can leave this change in place and remove the parameter as part of addressing @chetanmeh's comment about turning this function into a trait implementation.

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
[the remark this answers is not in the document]
rabbah: Note that I pushed a change to tighten the constructors - you cannot cons a result with a Left - only a Right. To convert a Right to a Left you must use `shrink`.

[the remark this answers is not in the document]
chetanmeh: Thats the Right thing to do !!

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
sven-lange-last: Thanks a lot. What do you think about adding a `logging.error()` to the `case _ =>` path - just to make sure that we have a chance to catch the unexpected case if somebody makes a mistake with setting up `AcknowledegmentMessage`?

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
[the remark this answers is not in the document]
rabbah: I'd rather not change the JSON represntation to include the type. I think it's not necessary at this point but I did simplify the implementation to something more compact.

rabbah:
@chetanmeh I refactored the active acker's signature/type.
@sven-lange-last we can't remove `WhiskActivation` - the tests in container proxy rely on it being available and removing it causes a number of issues in testing the proxy (specifically, the tests use non-blocking activations which only generate CompletionMessages and as a result there is no Activation to check as the tests do now.)

rabbah:
I removed WIP label as I think all comments are addressed at this point.

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
```diff
@@ -49,14 +49,21 @@ object InvokerReactive extends InvokerProvider {
...
+   * @param tid the transaction id for the activation
+   * @param activaiton is the activation result
+   * @param blockingInvoke is true iff the activation was a blocking request
+   * @param controllerInstance the originating controller/loadbalancer id
+   * @param userId is the UUID for the namespace owning the activation
+   * @param acknowledegment the acknowledgement message to send
    */
-  type ActiveAck = (TransactionId, WhiskActivation, Boolean, ControllerInstanceId, UUID, Boolean) => Future[Any]
+  trait ActiveAck {
+    def apply(tid: TransactionId,
```
chetanmeh: Nice use of `apply`! Minimizes the impact of switch

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
```diff
@@ -145,44 +152,46 @@ class InvokerReactive(
...
       }
 
-      producer.send(topic = "completed" + controllerInstance.asString, msg).andThen {
-        case Success(_) =>
-          logging.info(
-            this,
-            s"posted ${if (recovery) "recovery" else "completion"} of activation ${activationResult.activationId}")
+      // UserMetrics are sent, when the slot is free again. This ensures, that all metrics are sent.
+      if (UserEvents.enabled && acknowledegment.isSlotFree.nonEmpty) {
+        acknowledegment.result match {
```
rabbah: this change is actually wrong - for the completion message, it will return `None`. 
instead, the method should revert to using the passed in activation.

@sven-lange-last

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
```diff
@@ -145,44 +152,46 @@ class InvokerReactive(
...
+          case Some(Right(activationResult: WhiskActivation)) =>
+            EventMessage.from(activationResult, s"invoker${instance.instance}", userId) match {
+              case Success(msg) => UserEvents.send(producer, msg)
+              case Failure(t)   => logging.error(this, s"activation event was not sent: $t")
+            }
+          case _ =>
+            // all acknowledegment messages should have a result
```
rabbah: this comment is also wrong - completion messages have no result.

PR #4627, opened 2019-09-16
Add descriptions how to update, remove or rename runtime kinds and language families

sven-lange-last:
Removing or renaming runtime kinds and language families by simply changing the runtime manifest breaks existing actions. This change introduces documentation and adds background information to the pre-defined runtime manifests.

For example, a change like #4390 breaks existing actions.

This PR adds a `description` property to all pre-defined runtime manifests without updating the [`ExecManifest`](https://github.com/apache/openwhisk/blob/master/common/scala/src/main/scala/org/apache/openwhisk/core/entity/ExecManifest.scala) tooling. As a consequence, the `description` property is ignored by controllers and invokers and only serves documentation purposes. With that change, I hope to prevent unconscious changes of the runtimes breaking existing actions.

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [x] Documentation
- [x] Pre-defined runtime manifests

## Types of changes

## Checklist:


ansible/files/runtimes.json
```diff
@@ -16,7 +29,7 @@
                 }
             },
             {
-                "kind": "nodejs:8",
+                "kind": "nodejs:18",
```
rabbah: Typo.

sven-lange-last: Shoot... I had that change in the file for a test... missed to remove it.

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+
+1. Deprecate the runtime kind by setting `"deprecated": true` in the [runtimes manifest](actions-new.md#the-runtimes-manifest). This setting prevents that new actions can be created with the deprecated action kind. In addition, existing actions cannot be changed to the deprecated action kind any more.
+2. Ask owners of existing actions with runtime kind to be removed to update their actions to a different action kind.
+3. Create an automated process that identifies all actions with the runtime kind to be removed in the system's action artifact store. Either automatically remove these actions or change to a different runtime kind.
+4. Once the system's action artifact store does no more contain actions with the runtime kind to be removed, remove the runtime kind from the [runtimes manifest](actions-new.md#the-runtimes-manifest).
+5. TODO - is this correct?: Remove the runtime kind from the list of known kinds in the `ActionExec` object of the [controller API's Swagger definition](../core/controller/src/main/resources/apiv1swagger.json).
```
sven-lange-last: @rabbah Is step 5 required? Is anything else missing?

sven-lange-last:
@falkzoll @sciabarracom do you want to take a look at this PR from a runtime maintainer perspective?

docs/actions-update.md
BillZong: It's important to tell the API caller what kinds of runtimes supporting, so I think it's required to do so.

sven-lange-last: I'm not sure where the Swagger is exactly used. The controller itself offers an endpoint where it publishes the supported runtimes - and this seems to base on the runtimes manifest, not the Swagger definition. See for example: https://us-south.functions.cloud.ibm.com/

BillZong: You're right about this, the `Controller.info()` use runtimes manifest.
So the `ActionExec` in `apiv1swagger.json` would only make sense when executing unit tests.

ansible/files/runtimes-nodeonly.json
```diff
@@ -1,4 +1,18 @@
 {
+    "description": [
+        "This file describes the different languages (aka. managed action runtimes) supported by the system",
+        "as well as blackbox images that support the runtime specification.",
+        "Only actions with runtime families / kinds defined here can be created / read / updated / deleted / invoked.",
+        "Define a list of runtime families (example: 'nodejs') with at least one kind per family (example: 'nodejs:10').",
+        "Each runtime family needs a default kind (default: true).",
+        "When removing or renaming runtime families or runtime kinds from this file, all already existing actions",
```
dgrove-oss: Wordsmithing: 
        "When removing or renaming runtime families or runtime kinds from this file, all preexisting actions",
        "with the affected kinds can no longer be read / updated / deleted / invoked. In order to remove or rename

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+
+Over time, you may have the need for change:
+
+* Update runtimes to address security issues - for example, the latest code level of Node.js 10.
+* Remove runtime versions that are no more supported - for example, Node.js 6.
+* Add more languages due to user demand.
+* Remove languages that are no more needed.
+
+While adding and updating languages and runtimes is pretty straightforward, removing or renaming languages and runtimes
+requires some caution to prevent problems with existing actions.
```
dgrove-oss: ...to prevent problems with preexisting actions.

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+
+Follow these steps to update a particular runtime kind:
+
+1. Update the runtimes' container image.
+2. Update the corresponding `image` property in the [runtimes manifest](actions-new.md#the-runtimes-manifest) to use the new container image.
+3. Restart / re-deploy controllers and invokers such that they pick up the changed runtimes manifest.
+
+Already existing actions of the changed runtime kind will immediately use the new container image when the invoker creates a new action container.
+
+Apparently, this approach should only be used if existing actions do not break with the new container image. If a new container image is supposed to break existing actions, introduce a new runtime kind.
```
dgrove-oss: Apparently ==> Obviously ?

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Removing runtimes
+
+Follow these steps to remove a particular runtime kind under the assumption that actions with the runtime kind exist in the system. Apparently, it makes sense to wait some time between accomplishing these steps to give action owners time to react.
```
dgrove-oss: Apparently isn't the right word here.  Perhaps change to "Clearly the steps below should be spaced out in time to give action owners..."

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+
+1. Deprecate the runtime kind by setting `"deprecated": true` in the [runtimes manifest](actions-new.md#the-runtimes-manifest). This setting prevents that new actions can be created with the deprecated action kind. In addition, existing actions cannot be changed to the deprecated action kind any more.
+2. Ask owners of existing actions with runtime kind to be removed to update their actions to a different action kind.
+3. Create an automated process that identifies all actions with the runtime kind to be removed in the system's action artifact store. Either automatically remove these actions or change to a different runtime kind.
+4. Once the system's action artifact store does no more contain actions with the runtime kind to be removed, remove the runtime kind from the [runtimes manifest](actions-new.md#the-runtimes-manifest).
```
dgrove-oss: "does no more" ==> "does not"

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+5. TODO - is this correct?: Remove the runtime kind from the list of known kinds in the `ActionExec` object of the [controller API's Swagger definition](../core/controller/src/main/resources/apiv1swagger.json).
+
+If you remove a runtime kind from the [runtimes manifest](actions-new.md#the-runtimes-manifest), all actions that still use the removed runtime kind can no more be read from the system's action artifact store and thus, can no more be read, updated, deleted or invoked.
```
dgrove-oss: "can no more" ==> "can no longer"

ansible/files/runtimes.json
```diff
@@ -1,4 +1,17 @@
 {
+    "description": [
+        "This file describes the different languages (aka. managed action runtimes) supported by the system",
+        "as well as blackbox images that support the runtime specification.",
+        "Only actions with runtime families / kinds defined here can be created / read / updated / deleted / invoked.",
+        "Define a list of runtime families (example: 'nodejs') with at least one kind per family (example: 'nodejs:10').",
+        "Each runtime family needs a default kind (default: true).",
+        "When removing or renaming runtime families or runtime kinds from this file, all already existing actions",
+        "with the affected kinds can no more be read / updated / deleted / invoked. In order to remove or rename",
```
dgrove-oss: Same suggestions as other copy of this text

ansible/files/runtimes-nodeonly.json
sven-lange-last: Thanks a lot for providing more accurate wording / terms.

docs/actions-update.md
sven-lange-last: Thanks a lot for providing more accurate wording / terms.

sven-lange-last:
@dgrove-oss thanks a lot for reviewing and providing clarifications / improvements. I incorporated your proposals.

docs/actions-update.md
sven-lange-last: Actually, I wonder whether a fixed list of supported runtime kinds in the Swagger API definition makes sense at all...

* We expect that each adopter will have their own set of supported runtime families and kinds resulting in a customised version of the `runtimes.json` file.
* We have a mechanism in Openwhisk that allows adopters to configure their own `runtimes.json` instead of using the default file: https://github.com/apache/openwhisk/blob/06362ad83608335a01ccbb4ef449ec4c0e70dd99/ansible/group_vars/all#L54-L69.
* In order to obtain a sound Swagger API definition, each adopter has to synchronise [resource file `apiv1swagger.json`](../core/controller/src/main/resources/apiv1swagger.json).
  Maybe I'm missing something - but I see no configuration mechanism to replace the resource file provided by Openwhisk. So replacing this file requires some extra operations...
* At the same time, the controller provides a REST endpoint where it publishes all currently supported runtime families and kinds. This information comes from the configured `runtimes.json` file.
* Wouldn't it make sense to remove the runtime kind `enum` from the Swagger API definition (https://github.com/apache/openwhisk/blob/06362ad83608335a01ccbb4ef449ec4c0e70dd99/core/controller/src/main/resources/apiv1swagger.json#L1890-L1915) and add a hint to the description that there is a REST endpoint where the current list can be obtained?
  Does anybody know whether the `enum` data is used somewhere?

falkzoll:
👍 to document how to remove an existing runtime/kind in a controlled way from an already running openwhisk makes a lot of sense. Thanks for providing this guidance.

docs/actions-update.md
BillZong: Don't get any idea about it.

Remove it and run the tests would make sense.
[Local development](https://github.com/apache/openwhisk/tree/master/ansible/README.md) and [Unit Test](https://github.com/apache/openwhisk/tree/master/tests/README.md)

falkzoll: I would support to remove this enum from the swagger documentation and having a pointer to the REST endpoint instead. 
We may think about mentioning two or three samples of stable/longer time supported runtimes, e.g. nodejs:12, nodejs:default, python:3 in this description, just that users get a feeling how the values can look like, but do not specify the whole list.

I have not seen unit tests, that rely on this swagger enum, yet, but in case they exist, I would prefer that we change them to query the REST endpoint, instead of querying the info from the swagger documentation.

dgrove-oss: I agree with @sven-lange-last's  line of thought.  runtimes.json is the source of truth, so we should be using it via the controller's REST API, not an out-of-synch swagger.

dgrove-oss: As discussed on tech exchange, updating the swagger and the implications of that is best done in a separate follow up PR.

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+-->
+## Updating Action Language Runtimes
+
+OpenWhisk supports [several languages and runtimes](actions.md#languages-and-runtimes) that can be made
+available for usage in an OpenWhisk deployment. This is done via the [runtimes manifest](actions-new.md#the-runtimes-manifest).
+
+Over time, you may have the need for change:
+
+* Update runtimes to address security issues - for example, the latest code level of Node.js 10.
+* Remove runtime versions that are no more supported - for example, Node.js 6.
```
rabbah:
```suggestion
* Remove runtime versions that are no longer supported - for example, Node.js 6.
```

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+* Add more languages due to user demand.
+* Remove languages that are no more needed.
```
rabbah:
```suggestion
* Remove languages that are no longer needed.
```

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+## Updating Action Language Runtimes
+
+OpenWhisk supports [several languages and runtimes](actions.md#languages-and-runtimes) that can be made
+available for usage in an OpenWhisk deployment. This is done via the [runtimes manifest](actions-new.md#the-runtimes-manifest).
+
+Over time, you may have the need for change:
```
rabbah:
```suggestion
Over time, you may need to do one or more of the following:
```

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+* Remove languages that are no more needed.
+
+While adding and updating languages and runtimes is pretty straightforward, removing or renaming languages and runtimes
+requires some caution to prevent problems with preexisting actions.
+
+### Updating runtimes
+
+Follow these steps to update a particular runtime kind:
+
+1. Update the runtimes' container image.
```
rabbah:
```suggestion
1. Update the runtime's container image.
```

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+2. Update the corresponding `image` property in the [runtimes manifest](actions-new.md#the-runtimes-manifest) to use the new container image.
+3. Restart / re-deploy controllers and invokers such that they pick up the changed runtimes manifest.
+
+Already existing actions of the changed runtime kind will immediately use the new container image when the invoker creates a new action container.
+
+Obviously, this approach should only be used if existing actions do not break with the new container image. If a new container image is supposed to break existing actions, introduce a new runtime kind.
```
rabbah:
```suggestion
Obviously, this approach should only be used if existing actions do not break with the new container image. If a new container image may break existing actions, consider introducing a new runtime kind instead.
```

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Removing runtimes
+
+Follow these steps to remove a particular runtime kind under the assumption that actions with the runtime kind exist in the system. Clearly, the steps below should be spaced out in time to give action owners time to react.
+
+1. Deprecate the runtime kind by setting `"deprecated": true` in the [runtimes manifest](actions-new.md#the-runtimes-manifest). This setting prevents that new actions can be created with the deprecated action kind. In addition, existing actions cannot be changed to the deprecated action kind any more.
```
rabbah:
```suggestion
1. Deprecate the runtime kind by setting `"deprecated": true` in the [runtimes manifest](actions-new.md#the-runtimes-manifest). This setting prevents new actions from being created with the deprecated action kind. In addition, existing actions cannot be changed to the deprecated action kind any more.
```

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+2. Ask owners of existing actions with runtime kind to be removed to update their actions to a different action kind.
```
rabbah:
```suggestion
2. Ask owners of existing actions affected by the runtime kind to remove or update their actions to a different action kind.
```

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+3. Create an automated process that identifies all actions with the runtime kind to be removed in the system's action artifact store. Either automatically remove these actions or change to a different runtime kind.
```
rabbah: @sven-lange-last if you're done this in the past, do you plan to share the scripts?

docs/actions-update.md
```diff
@@ -0,0 +1,75 @@
...
+4. Once the system's action artifact store does not contain actions with the runtime kind to be removed, remove the runtime kind from the [runtimes manifest](actions-new.md#the-runtimes-manifest).
+5. Remove the runtime kind from the list of known kinds in the `ActionExec` object of the [controller API's Swagger definition](../core/controller/src/main/resources/apiv1swagger.json).
+
+If you remove a runtime kind from the [runtimes manifest](actions-new.md#the-runtimes-manifest), all actions that still use the removed runtime kind can no more be read from the system's action artifact store and thus, can no longer be read, updated, deleted or invoked.
```
rabbah:
```suggestion
If you remove a runtime kind from the [runtimes manifest](actions-new.md#the-runtimes-manifest), all actions that still use the removed runtime kind can no longer be read, updated, deleted or invoked.
```

docs/actions-update.md
sven-lange-last: Sadly, removing something from production takes very long in IBM. That's why we actually haven't removed anything yet. So this step is still ahead of us and I have nothing I could share...

sven-lange-last:
@rabbah @dgrove-oss all feedback should be addressed and I added your suggestions (thanks a lot for improving wording). Can any of you please merge?

dgrove-oss:
Thanks for all your work on this @sven-lange-last  -- merging

BillZong:
@sven-lange-last I'v try the API docs on my Mac. And yes, it would show swagger.json for documentation like **dgrove-oss** said. See those picture:




Hope to get more progress on this, since wrong documentation would mislead the action developers...

PR #4632, opened 2019-09-18
Activation Persister Service

chetanmeh:
This PR introduces a new micro service - `Activation Persister Service` (APS). This is based on discussion done [here][1] and [here][2]. 

## Description

This service fullfills following objective

1. Move out the load of persisting Activation results from Invoker to a new service
2. Enables a more controlled write to backend db in case rate of production of activation is more than the rate at which db can persist them within given capacity

With this service the activation results would now instead written to a Kafka topic and then picked by this service and then saved in db. The approach taken here varies with following parameters

1. Is the activation for a blocking or non blocking call
2. Is the activation result being persisted from Controller (for sequence/composition) or from Invoker
3. Does the setup stores logs as part of activation or store them seprately

### Components

The implementation consist of following high level components

1. Activation Persister Service (APS) - A new service impl in `core/persister` module. It is based on [Alpakka Kafka][3] and reads the activation record from Kafka and saves them to db
2. `KafkaActivationStore` - An extension of `ArtifactActivationStore` which also routes to Kafka via `MessageProducer` connector
3. InvokerReactive - The `send` logic would be modfied (details below)
4. New Kafka topic 
    1. `completed-others` - This topic is meant to be used for setups using `LogDriverLogStore` i.e. where logs are not stored in activation result. Further note that the name of topic confirms to pattern `completed.*` such that Kafka consumer used within ActivationPersisterService can pickup this topic along with other completed topics created per controller
        1. Non blocking activations - Currently non blocking activations are not sent over Kafka. So this topic would be used for them. We did used the existing `completed<controllerId>` as they are strictly used for blocking case
        2. Activations originating from controller - All activations from controller like for trigger, conductor and sequence would be sent via this topic. Again these are currently directly stored in db and not sent on any other kafka topic
    2. `activations` - This topic name would be used when setup stores the log within activation result. Note that in this case if APS used one need to ensure that Kafka can carry messages of size upto 10MB (or configured log limits)

Some key aspects which impact the implementation are

1. Blocking/Non Blocking - Depending on invocation type the nature of activation result sent via Kafka `completed` topic varies
    a. Blocking - The `ResultMessage` contains the activation result without logs
    b. Non Blocking - Only `CompletionMessage` is sent

### Activation Persister Service

This service implements a streaming flow based on [Alpakka Kafka][3]. It runs in 2 modes

#### ActivationResult without Logs

If the setup does not store the activation with logs (uses some `LogDriverLogStore`) then persister service would listen for `ResultMessage` (and CombinedCompletionAndResultMessage post #4624) on all `completed.*` topics. This would also pickup records from `completed-others` topic.

Further how `ActivationStore` gets used would change

1. Within `Controller` - Store all activations to `completed-others` via `KafkaActivationStore`
2. Within `Invoker` - The `store` function would be initialized to a noop function. Here we did not went for a `NoopActivationStore` as that would cause issue when configuring this setup with `StandaloneOpenWhisk` for test setup as such a setup can only have a single `ActivationStore` impl

InvokerReactive send flow would send all WhiskActivation result for blocking calls to existing `completed<instanceId>` topic. In addition non blocking results would also get routed to Kafka to `completed-others`

#### ActivationResult with Logs

In this mode APS would listen for messages on `activations` topic. `KafkaActivationStore` would route all activations to this topic

[1]: https://lists.apache.org/thread.html/7c14d2075e4f8258ec6e909f6b443f4b899b9982103a29b520d963ac@%3Cdev.openwhisk.apache.org%3E
[2]:https://lists.apache.org/thread.html/0f51397cfe6a910e67b8400f8dbbf07ea46dddd8b8c3b31c771510a8@%3Cdev.openwhisk.apache.org%3E
[3]: https://doc.akka.io/docs/alpakka-kafka/current/home.html

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


common/scala/src/main/scala/org/apache/openwhisk/http/BasicRasService.scala
```diff
@@ -36,7 +37,7 @@ trait BasicRasService extends BasicHttpService {
     }
   }
 
-  val ping = path("ping") {
+  def ping: Route = path("ping") {
```
rabbah: this could stay as a `val`...

core/invoker/src/main/resources/application.conf
```diff
@@ -120,5 +120,14 @@ whisk {
 
   invoker {
     protocol: http
+
+    activations {
+      # Enable this if ActivationStore is to be disabled and all activation results are
+      # to be sent to Kafka topic for further processing by ActivationPersisterService
+      activation-store-enabled = true
```
rabbah: this is strangely named: you enable this property to send to kafka instead of the "activation store" (which i'd think is the existing store)...

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
```diff
@@ -195,9 +197,14 @@ class InvokerReactive(
   }
 
   /** Stores an activation in the database. */
-  private val store = (tid: TransactionId, activation: WhiskActivation, context: UserContext) => {
-    implicit val transid: TransactionId = tid
-    activationStore.storeAfterCheck(activation, context)(tid, notifier = None)
+  private val store = if (activationStorageConfig.activationStoreEnabled) {
+    (tid: TransactionId, activation: WhiskActivation, context: UserContext) =>
+      {
```
rabbah: need the `{}` here?

common/scala/src/main/scala/org/apache/openwhisk/core/database/KafkaActivationStore.scala
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.openwhisk.spi.SpiLoader
+import org.apache.openwhisk.core.entity.size._
+import pureconfig.loadConfigOrThrow
+
+import scala.concurrent.Future
+import scala.util.Success
+
+case class KafkaActivationStoreConfig(activationsTopic: String, db: Boolean, maxRequestSize: Option[ByteSize])
+
+class KafkaActivationStore(producer: MessageProducer,
```
rabbah: is this loaded/activated via SPI?

common/scala/src/main/scala/org/apache/openwhisk/core/database/KafkaActivationStore.scala
```diff
@@ -0,0 +1,83 @@
...
+                           actorSystem: ActorSystem,
+                           actorMaterializer: ActorMaterializer,
+                           logging: Logging)
+    extends ArtifactActivationStore(actorSystem, actorMaterializer, logging) {
+
+  override def store(activation: WhiskActivation, context: UserContext)(
+    implicit transid: TransactionId,
+    notifier: Option[CacheChangeNotification]): Future[DocInfo] = {
+    if (config.db) {
+      sendToKafka(activation).flatMap(_ => super.store(activation, context))
```
rabbah: should there be a series of stores that are configured at top level, all of which receive the activation - to keep the db setting from trickling down to subtypes?

common/scala/src/main/resources/reference.conf
```diff
@@ -82,3 +82,22 @@ dispatchers {
...
+    activations-topic = "completed-others"
+
+    # Also send to database via ActivationStore. This mode can be used for initial
+    # trial runs where data is sent to both places
+    db = false
+
+    # Maximum request size. By default it uses the MAX_ACTIVATION_LIMIT as computed from
+    # `whisk.activation.payload`. Bump it to higher value if the activation result also
+    # includes the logs
+    # max-request-size = 2 MB
```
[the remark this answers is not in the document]
chetanmeh: Yes thats possible. Probably I can add a programatic check in `KafkaActivationStore` where we pick the max and log a warning.  Purpose here is to ensure that no activation is rejected due to max payload size error as the fallback route to add to db is disabled

common/scala/src/main/scala/org/apache/openwhisk/http/BasicRasService.scala
chetanmeh: Somehow I thought `val` cannot be overridden. Would fix that

common/scala/src/main/scala/org/apache/openwhisk/core/database/KafkaActivationStore.scala
chetanmeh: Yes via `ActivationStore` SPI

core/invoker/src/main/resources/application.conf
chetanmeh: Description here was wrong. Instead it would be

```
# Disable this if ActivationStore should not be used and all activation results are
# to be sent to Kafka topic for further processing by ActivationPersisterService
```

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
chetanmeh: Technically not as its a single statement but it help in clarifying the boundaries

core/persister/src/main/scala/org/apache/openwhisk/core/database/persister/ActivationConsumer.scala
```diff
@@ -0,0 +1,152 @@
...
+      case CombinedCompletionAndResultMessage(tid, Right(wa), _, _) => Some((tid, wa))
+      case _                                                        => None
+    }
+  }
+
+  private def consumerSettings(): ConsumerSettings[String, Array[Byte]] =
+    ConsumerSettings(system, new StringDeserializer, new ByteArrayDeserializer)
+      .withGroupId(config.groupId)
+      .withBootstrapServers(config.kafkaHosts)
+      .withProperty(ConsumerConfig.CLIENT_ID_CONFIG, config.clientId)
```
style95: Would it be worth to make this blank or to use the unique ID for each instance?
We may want to scale out the microservice at some point.
Then each instance will form one consumer group.

core/persister/src/main/resources/application.conf
```diff
@@ -0,0 +1,43 @@
...
+  kafka-clients {
+    # Disable auto-commit by default
+    enable.auto.commit = false
+
+    # What to do when there is no initial offset in Kafka or if the current offset
+    # does not exist any more on the server (e.g. because that data has been deleted):
+    # - earliest : automatically reset the offset to the earliest offset
+    # - latest   : automatically reset the offset to the latest offset
+    # - none     : throw exception to the consumer if no previous offset is found for the consumer's group
+    auto.offset.reset = "earliest"
```
style95: This implies the multiple processing of the same activation.
Some activation stores would just override existing activations but others may not.

core/persister/src/main/scala/org/apache/openwhisk/core/database/persister/ActivationStorePersister.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  override def persist(act: WhiskActivation)(implicit tid: TransactionId): Future[Done] = {
+    store
+      .store(act, userContext)(tid, None)
+      .recoverWith {
+        //TODO Metric - Conflict counts
+        //Recover for conflict case as its possible in case of unclean shutdown persister need to process
+        // same activation again
+        //Checking the chain as exception may get wrapped
+        case t if Throwables.getRootCause(t).isInstanceOf[DocumentConflictException] => Future.successful(Done)
```
style95: oh here, it would take care of the situation where same activations are processed multiple time.

common/scala/src/main/scala/org/apache/openwhisk/core/database/KafkaActivationStore.scala
```diff
@@ -0,0 +1,89 @@
...
+import scala.util.{Failure, Success}
+
+case class KafkaActivationStoreConfig(activationsTopic: String, db: Boolean, maxRequestSize: Option[ByteSize])
+
+class KafkaActivationStore(producer: MessageProducer,
+                           config: KafkaActivationStoreConfig,
+                           actorSystem: ActorSystem,
+                           actorMaterializer: ActorMaterializer,
+                           logging: Logging)
+    extends ArtifactActivationStore(actorSystem, actorMaterializer, logging) {
```
style95: Is there any possibility that `KafkaActivationStore` can be used with other `ActivationStore` rather than just `ArtifactActivationStore`?

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
jiangpengcheng: I'm getting a little confused, if we set `activationStoreEnabled` to `false`, the `store` will do nothing here(code will not jump to `KafkaActivationStore.store` neither), then how can `activations` be sent to kafka? I see that only blocking activations result will be sent to `completedN` topic, but for non-blocking activations, they will be missing

common/scala/src/main/scala/org/apache/openwhisk/core/database/KafkaActivationStore.scala
jiangpengcheng: I think it's hard for `KafkaActivationStore` to be compatible with other `ActivationStore` as different `ActivationStore` may have different definition for methods `store`, `get`, `delete` and `countActivationsInNamespace`, and so on, and we cannot define these methods except the `store` in the `KafkaActivationStore`

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
chetanmeh: Support for non-blocking activations is currently pending.

core/persister/src/main/resources/application.conf
chetanmeh: Yes the semantics being implemented here are at-least-once processing. So same activation may be processed multiple times

common/scala/src/main/scala/org/apache/openwhisk/core/database/KafkaActivationStore.scala
chetanmeh: Current ActivationPersisterService is also using the ActivationStore for storage. So in theory we can support others. 

I can introduce a config to capture base activation store and then have KafkaActivationStore act as decorator for that. Would give it a try

core/persister/src/main/scala/org/apache/openwhisk/core/database/persister/ActivationConsumer.scala
chetanmeh: Can be done. Currently `client.id` is used to access the lag stats from JMX MBean. Hence made it deterministic via config.

From [docs for client.id](http://kafka.apache.org/documentation.html#producerconfigs) I was not sure of its significance in multi consumer group settings. Currently if you do not specify the id then it gets something like `consumer-0` based on a static counter. So 2 consumer in same group would still end up getting same id.

> An id string to pass to the server when making requests. The purpose of this is to be able to track the source of requests beyond just ip/port by allowing a logical application name to be included in server-side request logging.

common/scala/src/main/resources/reference.conf
```diff
@@ -82,3 +82,22 @@ dispatchers {
...
   }
 }
+
+whisk {
+  kafka-activation-store {
+    # Name of the Kafka topic for sending WhiskActivations
+    # Set to
+    # - `completed-others` - When LogDriverLogStore is used
+    # - `activations` - When other LogStore is used where logs are stored as part of activation result
+    activations-topic = "completed-others"
```
style95: I feel there may be a more intuitive name for this topic(`completed-others`).
It makes me vision activations sent to some misc controllers or else at first sight.

I am not quite sure it's worth to do and I know this is to use the topic pattern, but how about using a more intuitive name? such as `activations-with-logs` and just `activations`? 

Opinion from others would be appreciated.

common/scala/src/main/scala/org/apache/openwhisk/core/database/KafkaActivationStore.scala
chetanmeh: This is implemented now. There is now a config to configure the primary store via `primary-store-provider` which defaults to `ArtifactActivationStoreProvider`

```hocon
whisk {
  kafka-activation-store {
    # Name of the Kafka topic for sending WhiskActivations
    # Set to
    # - `completed-others` - When LogDriverLogStore is used
    # - `activations` - When other LogStore is used where logs are stored as part of activation result
    activations-topic = "completed-others"

    # Primary activation store used for eventual storage and querying of activation records
    primary-store-provider = org.apache.openwhisk.core.database.ArtifactActivationStoreProvider

    # Also store directly to primary ActivationStore. This mode can be used for initial
    # trial runs where data is sent to both primary store and also via Kafka to ActivationPersisterService
    store-in-primary = false

    # Maximum request size. By default it uses the MAX_ACTIVATION_LIMIT as computed from
    # `whisk.activation.payload`. Bump it to higher value if the activation result also
    # includes the logs
    # max-request-size = 2 MB
  }
}
```

So one can configure any `ActivationStore` impl and `KafkaActivationStore` would wrap it and handles the store flow. Rest of the flow is directed back to primary store

common/scala/src/main/scala/org/apache/openwhisk/core/database/KafkaActivationStore.scala
chetanmeh: With recent addition of `ActivationStoreWrapper` this is now handled and `KafkaActivationStore` would now delegate to the configured primary store here if configured

core/invoker/src/main/scala/org/apache/openwhisk/core/invoker/InvokerReactive.scala
bdoyle0182: Is there any update on whether this will support non-blocking activations?

bdoyle0182:
More of a question about activation storage within the invoker and acknowledgment back to the controller. So in the current state it looks like the acknowledgment message on `completed*` that's read by the controller is sent async / simultaneously on the invoker with writing the activation to couchdb or whatever the persistence store. So is there then a race condition that `completed*` gets the acknowledgment, but the activation is then never written to couchdb in the case that producing to `completed*` finishes first, then the invoker goes down before the activation is written to couchdb and the activation is therefore lost. If my understanding is correct, would this new service solve this race condition because it seems like it's just swapping out the store operation in the invoker with a produce to kafka so the ack and new activation store produce to kafka would still be happening simultaneously.

common/scala/src/main/resources/reference.conf
bdoyle0182: Was also thinking this about the name. Would `completed-async` work? It's at the very least much more descriptive than `others`

style95:
@chetanmeh 
Please let me know if you would work on it again.

PR #4660, opened 2019-10-02
Include response size in Activation

chetanmeh:
Enables including the response size in ActivationResponse, WhiskActivation and eventually in Activation object sent to `events` topic

## Description

For better estimating the DB storage requirement with usage it would be helpful to get an estimate of the distribution of activation response size. This PR enables including the size stat with Activation object

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
```diff
@@ -338,7 +339,8 @@ object Activation extends DefaultJsonProtocol {
       "kind",
       "conductor",
       "memory",
-      "causedBy")
+      "causedBy",
+      "size")
```
rabbah: > Enables including the response size in ActivationResponse, WhiskActivation and eventually in Activation object sent to events topic

Why not not just calculate this where needed and avoid changing the message or the interfaces below?

chetanmeh: That would require serializing the response while constructing the Activation object to send the User event.

To avoid that overhead I had to pass the size all the way.

selfxp: Should we also update the [documentation](https://github.com/apache/openwhisk/blob/master/docs/metrics.md#supported-events) with a short description of the new field? E.g.:  
```
...
memory - maximum memory allowed for action container
size - action response size in bytes
```
and 
```
{"body":{"statusCode":0,"duration":3,"name":"whisk.system/invokerHealthTestAction0","waitTime":583915671,"conductor":false,"kind":"nodejs:6","initTime":0,"memory": 256,"size": 128, "causedBy": false},"eventType":"Activation","source":"invoker0","subject":"whisk.system","timestamp":1524476122676,"userId":"d0888ad5-5a92-435e-888a-d55a92935e54","namespace":"whisk.system"}
```

common/scala/src/main/scala/org/apache/openwhisk/core/entity/ActivationResult.scala
```diff
@@ -86,19 +88,25 @@ protected[core] object ActivationResponse extends DefaultJsonProtocol {
...
 
-  protected[core] def success(result: Option[JsValue] = None) = ActivationResponse(Success, result)
-
-  protected[core] def applicationError(errorValue: JsValue) = error(ApplicationError, errorValue)
-  protected[core] def applicationError(errorMsg: String) = error(ApplicationError, JsString(errorMsg))
-  protected[core] def developerError(errorValue: JsValue) = error(DeveloperError, errorValue)
-  protected[core] def developerError(errorMsg: String) = error(DeveloperError, JsString(errorMsg))
-  protected[core] def whiskError(errorValue: JsValue) = error(WhiskError, errorValue)
-  protected[core] def whiskError(errorMsg: String) = error(WhiskError, JsString(errorMsg))
+  protected[core] def success(result: Option[JsValue] = None, size: Option[Int] = None) =
```
rabbah: Why is size an option? Backward compatibility? If so, perhaps should be dealt with in deserializer. This would enforce the size always being Some value.

chetanmeh: Size is only applicable for case where error is based on some response (similar to via result is an Option). In some case we just construct a failed response due to connection error etc and there size is not applicable

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
chetanmeh: Updated the docs to document the `size` metrics

chetanmeh:
This PR is now ready for review

PR #4679, opened 2019-10-10
update travis config to use xenial

dgrove-oss:


dgrove-oss:
travisci builds have been failing for the last 5 days while trying to update pip.  Before trying to make that work again on our `deprecated-2017Q3` group, I think it is worth making an attempt to move to xenial.

dgrove-oss:
closing in favor of #4683

PR #4671, opened 2019-10-04
Standalone OpenWhisk KubernetesContainerFactory support

chetanmeh:
Enables support for configuring `KubernetesContainerFactory` (KCF) with Standalone OpenWhisk for simplified local development for KCF.

## Description

To simplify developing on KCF locally this PR adds support for configuring KCF via `--dev-kcf` flag. 

### Local Port Forwardning

To enable this mode a new config option is added to enable port forwarding

```
wsk {
  kubernetes {
    # Enables forwarding to remote port via a local random port. This mode is mostly useful
    # for development via Standalone mode
    port-forwarding-enabled = false
  }
}
```

If this mode is enabled then after launching a pod a local port forward would be enabled and that port would then be used by `ContainerProxy` to communicate with the pod running with k8s

### Usage

To try out this feature

#### 1. Install and configure Kind

We would use [Kind][1] to setup a local k8s. Follow the steps [here][1] to create a simple cluster

```bash
$ kind create cluster --wait 5m

# Export the kind config for kubectl usage
$ export KUBECONFIG="$(kind get kubeconfig-path)"

# Configure the default namespace
$ kubectl config set-context --current --namespace=default

# See the config path
$ kind get kubeconfig-path
/Users/chetanm/.kube/kind-config-kind
```

#### 2. Launch Standalone 

```bash

# Build the standalone jar
$ ./gradlew :core:standalone:build

# Launch it with `kubeconfig` system property set to kind config
$ java  -Dkubeconfig="$(kind get kubeconfig-path)" -jar bin/openwhisk-standalone.jar --dev-kcf
```

Once started and required `.wskprops` configured (see standalone startup logs for details) create a `hello.js` function

```js
function main(params) {
    greeting = 'hello, world'
    var hello = {payload: greeting}
    var result = {...hello, ...process.env}
    console.log(greeting);
    return result
}
```

```bash
$ wsk action create hello hello.js
$ wsk action invoke hello -br
```

This shows an output like below

```json
{
    "HOME": "/root",
    "HOSTNAME": "wsk0-2-prewarm-nodejs10",
    "KUBERNETES_PORT": "tcp://10.96.0.1:443",
    "KUBERNETES_PORT_443_TCP": "tcp://10.96.0.1:443",
    "KUBERNETES_PORT_443_TCP_ADDR": "10.96.0.1",
    "KUBERNETES_PORT_443_TCP_PORT": "443",
    "KUBERNETES_PORT_443_TCP_PROTO": "tcp",
[...]

PR #4690, opened 2019-10-15
Pod template support with KubernetesContainerFactory

chetanmeh:
This PR adds support for providing a base Pod template to customize the action Pod created by KubernetesContainerFactory (KCF)

## Description

Sometime back we [discussed a usecase on DL][1] to configure a private registry for action containers. This PR now enables that usecase via an optional `pod-template` config

### Usage

Provide the base pod template via config like below

Pod template yaml stored at "/pod-template.yaml` 

```yaml
---
apiVersion: "v1"
kind: "Pod"
metadata:
  annotations:
    allow-outbound : "true"
```

```
whisk {
  spi {
    ContainerFactoryProvider = "org.apache.openwhisk.core.containerpool.kubernetes.KubernetesContainerFactoryProvider"
  }
  kubernetes {
    pod-template = "file:/pod-template.yaml"
  }
}
```

When KCF would create an action container POD it would use the above template as base and add the changes required by OpenWhisk on top of that. 

There are 3 ways to pass this yaml config

#### 1. Inline

```
whisk {
  kubernetes {
    pod-template = """---
apiVersion: "v1"
kind: "Pod"
metadata:
  annotations:
    allow-outbound : "true"
    """
  }
}
```

#### 2. Via File reference

```
whisk {
  kubernetes {
    pod-template = "file:/pod-template.yaml"
  }
}
```
Here the file can be mounted from a config map

#### 3. Via ENV

```
whisk {
  kubernetes {
    pod-template = ${WHISK_POD_TEMPLATE}
  }
}
```

Here the env `WHISK_POD_TEMPLATE` is configured via a ConfigMap value

To see the final POD spec this PR also enables support for logging the pod spec in yaml via `X-OW-EXTRA-LOGGING` header (#3174). 

For example assume that setup has a `hello` action configured (say using #4671). 

```
$ curl -k --dump-header -  -X POST \
      --user "23bc46b1-71f6-4ed5-8c54-816aa4f8c502:123zO3xZCLrMN6v2BKK1dXYFpXlPkccOFqm12CdAsMgRU4VrNZ9lyGVCGuMDGIwP" \
      -H "X-OW-EXTRA-LOGGING: on" \
      "http://localhost:3233/api/v1/namespaces/_/actions/hello?blocking=true&result=true"
```

Then above command would log the pod spec

```yaml
[2019-10-15T16:23:12.659Z] [INFO] [#tid_B7WSlmTXLKHTDCH0TYb6mcf6ttpyKNrZ] [KubernetesClient] Pod spec being created
---
apiVersion: "v1"
kind: "Pod"
metadata:
  annotations:
    allow-outbound: "true"
  labels:
    name: "wsk0-1-guest-hello3"
    invoker: "invoker0"
  name: "wsk0-1-guest-hello3"
  namespace: "default"
spec:
  containers:
  - env:
    - name: "__OW_API_HOST"
      value: "http://host.docker.internal:3233"
    image: "openwhisk/action-nodejs-v10:1.14.0-incubating"
    name: "user-action"
    ports:
    - containerPort: 8080
      name: "action"
    resources:
      limits:
        memory: "256Mi"
      requests: {}
    securityContext:
      capabilities:
        drop:
        - "NET_RAW"
        - "NET_ADMIN"
  nodeSelector: {}
  restartPolicy: "Always"
```

This spec includes the custom annotation configured in base template `allow-outbound`

### Implementation Details

With `pod-template` one can configure any k8s supported config for the `Pod`. Each such pod would always have  a  `container` named `user-action`. Following aspects though would be controlled by KCF 

1. Container memory limit
2. Container port
3. Container env
4. Container securityContext would always have `NET_RAW` and `NET_ADMIN` dropped. One can "add" more capabilities
5. Pod name

[1]: https://lists.apache.org/thread.html/ac5954d64078744fd5b234f973cd29556b77a846116f5793f1ed9b16@%3Cdev.openwhisk.apache.org%3E

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/build.gradle
+++ b/common/scala/build.gradle
@@ -59,7 +59,7 @@ dependencies {
     compile ('com.fasterxml.uuid:java-uuid-generator:3.1.3')
     compile 'com.github.ben-manes.caffeine:caffeine:2.6.2'
     compile 'com.google.code.findbugs:jsr305:3.0.2'
-    compile 'io.fabric8:kubernetes-client:4.4.2'
+    compile "io.fabric8:kubernetes-client:${gradle.kube_client.version}"
     compile ('io.kamon:kamon-core_2.12:1.1.3') {
         exclude group: 'com.lihaoyi'
     }
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/ConfigMapValue.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/ConfigMapValue.scala
@@ -1,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.openwhisk.common
+
+import java.io.File
+import java.net.URI
+import java.nio.charset.StandardCharsets.UTF_8
+
+import org.apache.commons.io.FileUtils
+import pureconfig.ConfigReader
+import pureconfig.ConvertHelpers.catchReadError
+
+class ConfigMapValue private (val value: String)
+
+object ConfigMapValue {
+
+  /**
+   * Checks if the value is a file url like `file:/etc/config/foo.yaml` then treat it as a file reference
+   * and read its content otherwise consider it as a literal value
+   */
+  def apply(config: String): ConfigMapValue = {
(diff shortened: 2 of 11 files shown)


mcdan:
Why not just add the pod template as a config map in the cluster instead on inline-ing the whole thing right into the config?

chetanmeh:
This is still possible by referring to env variable in config which in turn is backed by config map entry.

What's not possible is to refer to file via config map based mount. So would the env based approach be sufficient or we need the file based one also?

dgrove-oss:
Mounting a confgmap in an env var could work fine.  Maybe work through the helm chart change in kube-deploy and see what it would look like?  The thing I'd like to avoid if possible is having to write an awkward string-escaped version of the pod template to change the default value.

chetanmeh:
Added support to enable reading template value from file (see updated doc in PR)

```
whisk {
  kubernetes {
    pod-template = "file:/pod-template.yaml"
  }
}
```

So if `pod-template` value is

1. File reference - String representing a file uri `file:/path/to/template.yaml` then its content would be read and used as value
2. Env variable - `pod-template = ${WHISK_POD_TEMPLATE}` then TypeSafe config would internally read its value and use that
3. Literal inline yaml - Else its value would be considered as a literal yaml value

@dgrove-oss @mcdan Does above implemented approach looks fine?

PR #4698, opened 2019-10-22
Add healthcheck for Invoker -> Action Container 

tysonnorris:
To validate state of containers before use.

## Description
In our deployment case (mesos/kubernetes) there are occasional cases where prewarm/warm containers fail outside of the usage of ContainerProxy, but ContainerProxy/ContainerPool are not aware of these failures. When this happens, when eventually the containers are used, there is a guarantee of failing the activation, since ContainerProxy assumes the prewarm/warm container is running fine, and it just sees a failure for /init or /run. 

This may be more pertinent to ContainerFactory impls that launch containers on hosts other than the invoker (mesos, k8s, yarn), but since it is also possible for docker containers local to invoker to be interrupted by some other system, it may be useful in all cases as a preventative measure. 

This PR introduces a tcp connection health check from invoker to each action container where:
- prewarm containers will be connected periodically
- failure to connect to prewarm will eventually (after configurable number of failures) cause the container to be destroyed and deplaced with a new prewarm container
- warm containers (paused) will be connected immediately after resume, and immediately before usage for /run
- failure to connect to warm will cause the job to be rescheduled and the warm container to be destroyed

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -484,9 +484,17 @@ object LoggingMarkers {
     LogMarkerToken(invoker, "mesos", start, Some(cmd), Map("cmd" -> cmd))(MeasurementUnit.time.milliseconds)
   def INVOKER_MESOS_CMD_TIMEOUT(cmd: String) =
     LogMarkerToken(invoker, "mesos", timeout, Some(cmd), Map("cmd" -> cmd))(MeasurementUnit.none)
-  def INVOKER_CONTAINER_START(containerState: String) =
-    LogMarkerToken(invoker, "containerStart", counter, Some(containerState), Map("containerState" -> containerState))(
-      MeasurementUnit.none)
+  def INVOKER_CONTAINER_START(containerState: String, userNamespace: String, actionNamespace: String, action: String) =
+    LogMarkerToken(
+      invoker,
+      "containerStart",
+      counter,
+      Some(containerState),
+      Map(
+        "containerState" -> containerState,
+        "userNamespace" -> userNamespace,
+        "actionNamespace" -> actionNamespace,
+        "action" -> action))(MeasurementUnit.none)
   val INVOKER_CONTAINER_HEALTH = LogMarkerToken(invoker, "containerHealth", start)(MeasurementUnit.time.milliseconds)
   val INVOKER_CONTAINER_HEALTH_FAILED_WARM =
     LogMarkerToken(invoker, "containerHealthFailed", counter, Some("warm"), Map("containerState" -> "warm"))(
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -519,6 +519,10 @@ object LoggingMarkers {
     LogMarkerToken(containerPool, "prewarmCount", counter)(MeasurementUnit.none)
   val CONTAINER_POOL_PREWARM_SIZE =
     LogMarkerToken(containerPool, "prewarmSize", counter)(MeasurementUnit.information.megabytes)
+  val CONTAINER_POOL_IDLES_COUNT =
+    LogMarkerToken(containerPool, "idlesCount", counter)(MeasurementUnit.none)
+  val CONTAINER_POOL_IDLES_SIZE =
+    LogMarkerToken(containerPool, "idlesSize", counter)(MeasurementUnit.information.megabytes)
 
   val INVOKER_TOTALMEM_BLACKBOX = LogMarkerToken(loadbalancer, "totalCapacityBlackBox", counter)(MeasurementUnit.none)
   val INVOKER_TOTALMEM_MANAGED = LogMarkerToken(loadbalancer, "totalCapacityManaged", counter)(MeasurementUnit.none)
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -487,6 +487,7 @@ object LoggingMarkers {
   def INVOKER_CONTAINER_START(containerState: String) =
     LogMarkerToken(invoker, "containerStart", counter, Some(containerState), Map("containerState" -> containerState))(
       MeasurementUnit.none)
+  val INVOKER_CONTAINER_HEALTH = LogMarkerToken(invoker, "containerHealth", start)(MeasurementUnit.time.milliseconds)
   val CONTAINER_CLIENT_RETRIES =
     LogMarkerToken(containerClient, "retries", counter)(MeasurementUnit.none)
 
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -488,6 +488,12 @@ object LoggingMarkers {
     LogMarkerToken(invoker, "containerStart", counter, Some(containerState), Map("containerState" -> containerState))(
       MeasurementUnit.none)
   val INVOKER_CONTAINER_HEALTH = LogMarkerToken(invoker, "containerHealth", start)(MeasurementUnit.time.milliseconds)
+  val INVOKER_CONTAINER_HEALTH_FAILED_WARM =
+    LogMarkerToken(invoker, "containerHealthFailed", counter, Some("warm"), Map("containerState" -> "warm"))(
+      MeasurementUnit.none)
+  val INVOKER_CONTAINER_HEALTH_FAILED_PREWARM =
+    LogMarkerToken(invoker, "containerHealthFailed", counter, Some("prewarm"), Map("containerState" -> "prewarm"))(
+      MeasurementUnit.none)
   val CONTAINER_CLIENT_RETRIES =
     LogMarkerToken(containerClient, "retries", counter)(MeasurementUnit.none)
 
--- a/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
@@ -484,7 +484,7 @@ object LoggingMarkers {
     LogMarkerToken(invoker, "mesos", start, Some(cmd), Map("cmd" -> cmd))(MeasurementUnit.time.milliseconds)
   def INVOKER_MESOS_CMD_TIMEOUT(cmd: String) =
     LogMarkerToken(invoker, "mesos", timeout, Some(cmd), Map("cmd" -> cmd))(MeasurementUnit.none)
-  def INVOKER_CONTAINER_START(containerState: String, userNamespace: String, actionNamespace: String, action: String) =
+  def INVOKER_CONTAINER_START(containerState: String, invocationNamespace: String, namespace: String, action: String) =
     LogMarkerToken(
       invoker,
       "containerStart",
@@ -492,8 +492,8 @@ object LoggingMarkers {
       Some(containerState),
       Map(
         "containerState" -> containerState,
-        "userNamespace" -> userNamespace,
-        "actionNamespace" -> actionNamespace,
+        "initiator" -> invocationNamespace,
+        "namespace" -> namespace,
         "action" -> action))(MeasurementUnit.none)
   val INVOKER_CONTAINER_HEALTH = LogMarkerToken(invoker, "containerHealth", start)(MeasurementUnit.time.milliseconds)
   val INVOKER_CONTAINER_HEALTH_FAILED_WARM =
--- a/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
+++ b/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
@@ -33,7 +33,6 @@ import akka.io.Tcp.Close
 import akka.io.Tcp.CommandFailed
 import akka.io.Tcp.Connect
 import akka.io.Tcp.Connected
-import akka.pattern.ask
 import akka.pattern.pipe
 import pureconfig._
 import pureconfig.generic.auto._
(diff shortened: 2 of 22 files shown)


core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -820,6 +907,98 @@ object ContainerProxy {
...
+  var pingMarker: StartMarker = null
+  val tcp = IO(Tcp)
+  var resume: Option[Promise[Unit]] = None
+  val addressString = s"${remote.getHostString}:${remote.getPort}"
+  restartPing()
+
+  private def restartPing() = {
+    cancelPing() //just in case restart is called twice
+    scheduledPing = Some(
+      context.system.scheduler.schedule(config.checkPeriod, config.checkPeriod, self, HealthPingSend(false)))
```
style95: Should the initial delay be "0" to send the ping in the Ready state?
https://github.com/apache/openwhisk/pull/4698/files#diff-170ae706abad4ff6500ae12be05d0cbbR518

tysonnorris: I think waiting at least one checkPeriod before checking the first time is fine, since presumably it was just (immediately prior to this) being used, so we don't need the check till we have stopped using it for a bit. If this is important to be 0, or some other value besides checkPeriod, we could expose another config field to allow operators to specify the initial delay. WDYT?

style95: I think I misunderstood the behavior.

It seems it enables Ping when the state become `Ready`.
And after 50ms, it will become `Pausing` and `Paused` state if there is no more activation message. Then it would disable the Ping.
If `checkPeriod` is bigger than 50ms, can it send a ping before the state become `Pausing`?

And when it receives a message and become `Running` again, it would anyway disable the PingActor again. Then can we ping the container before running `/run`?

tysonnorris: @style95 This test already tests what happens (nothing) when the checkPeriod is bigger than pause timeout https://github.com/apache/openwhisk/pull/4698/files#diff-7731f505e6c9b7417c5b0678a2584fd2R1148 
We only ping between Ready and Pausing - not when Paused

When it becomes Running, the first message will be rescheduled on connection failure, unrelated to health ping.

style95: After one invocation is over, it will enable the ping actor by this:
https://github.com/apache/openwhisk/pull/4698/files#diff-170ae706abad4ff6500ae12be05d0cbbR526

If there is no subsequent activation, the container proxy will be `Pausing` state and the ping actor will be disabled again in 50ms(pause grace) by this:
https://github.com/apache/openwhisk/pull/4698/files#diff-170ae706abad4ff6500ae12be05d0cbbR532

Since the initial delay is not 0, no health check will be performed and it will setup `scheduledPing` and disable it immediately(50ms) without any real call.

If the ping actor is only for prewarm containers, do we need to enable the ping actor whenever it becomes a ready state?
Or are you assuming a bigger pause grace than check interval?

tysonnorris: I understand - I have disabled the checking between Ready and Pausing states, so checking only occurs:
* between Started and Running
* on first activation after Paused (this one is not a ping check, but an HTTP failure will cause activation to be rescheduled instead of failed)

tysonnorris:
@style95 @selfxp Any comments on this?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -870,6 +962,71 @@ object ContainerProxy {
   }
 }
 
+object TCPPingClient {
+  def props(containerId:String, config: ContainerProxyHealthCheckConfig, remote: InetSocketAddress, replies: ActorRef) =
+    Props(new TCPPingClient(containerId, remote, replies, config))
+}
+
+class TCPPingClient(containerId:String, remote: InetSocketAddress, listener: ActorRef, config: ContainerProxyHealthCheckConfig)
```
style95: If this `PingActor` is only used to keepalive prewarm containers, how about making it be shared among containers, rather than having dedicated ping actors for each container?
Then this becomes an invoker-wide functionality to keep prewarm containers healthy.
And I think it would not be a big overhead for this actor to send a ping to multiple containers.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -234,19 +247,25 @@ class ContainerProxy(factory: (TransactionId,
...
   var rescheduleJob = false // true iff actor receives a job but cannot process it because actor will destroy itself
   var runBuffer = immutable.Queue.empty[Run] //does not retain order, but does manage jobs that would have pushed past action concurrency limit
   //track buffer processing state to avoid extra transitions near end of buffer - this provides a pseudo-state between Running and Ready
   var bufferProcessing = false
 
   //keep a separate count to avoid confusion with ContainerState.activeActivationCount that is tracked/modified only in ContainerPool
   var activeCount = 0;
+  var healthPingActor: Option[ActorRef] = None //setup after prewarm starts
+  //track the resuming run for easily referring to the action being resumed (it may fail)
+  var resumeRun: Option[Run] = None
```
style95: How about including this in FSMData?
I think that may make this more testable as we can set FSM data directly in the test codes.

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
```diff
@@ -484,8 +484,23 @@ object LoggingMarkers {
...
-    LogMarkerToken(invoker, "containerStart", counter, Some(containerState), Map("containerState" -> containerState))(
+  def INVOKER_CONTAINER_START(containerState: String, userNamespace: String, actionNamespace: String, action: String) =
+    LogMarkerToken(
+      invoker,
+      "containerStart",
+      counter,
+      Some(containerState),
+      Map(
+        "containerState" -> containerState,
+        "userNamespace" -> userNamespace,
```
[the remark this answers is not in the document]
style95: Regarding this, we have been using `invocationNamespace`, I think we need to choose the consistent name for this.
https://github.com/apache/openwhisk/blob/master/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala#L142

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
tysonnorris: Do you mean make it a var on WarmedData?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
tysonnorris: I don't think this will work well. The ContainerProxy actor is already a 1:1 actor per container, and the state in ContainerProxy dictates when it should and should not be active. The overhead of an actor is minimal.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
style95: I meant `Option` data in FSMData(Warmed or not).

tysonnorris: I'm still not sure what you mean, I guess I need more context. I think it only applies on Warmed data, so it could possibly go there. But it doesn't apply always on WarmedData, which is why I kept it out of the state data, and just as part of the actor state.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
style95: Yes, I meant an invoker-wide service that keeps prewarm containers healthy.
So we can have 1:N mapping between PingingService and prewarm containers while still keeping the 1:1 mapping between ContainerProxy and a real container.

With this, PingingService doesn't need to be stopped and created upon the creation/deletion of ContainerProxy.

Each a prewarm ContainerProxy can request this service to enable the ping and PingingService can have a map or list of target endpoints to send a ping.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
style95: What I meant is just to include it in FSMData so that we can set the state and corresponding data with `TestFSMRef.setState()`. It will make this more testable.

If it only applies to `WarmedData` we can add it to `WarmedData`.
If it applies to multiple data, then we can just add it to multiple data.
I feel it is nicer to keep it in FSMData instead of having a `var` state.
We can easily transfer it between each transition with `stay with` directive.

common/scala/src/main/scala/org/apache/openwhisk/common/Logging.scala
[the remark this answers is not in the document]
tysonnorris: I think @selfxp is talking about naming the metrics tags and @style95 is talking about naming scala parameters. I will update both, but these are currently named differently, which I will leave as is for now.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -581,6 +636,25 @@ class ContainerProxy(factory: (TransactionId,
...
+          TCPPingClient
+            .props(c.toString(), healtCheckConfig, new InetSocketAddress(c.addr.host, c.addr.port), self))
+        healthPingActor = Some(hp)
+        hp
+      }
+      hpa ! HealthPingEnabled(true)
+    }
+  }
+  private def disableHealthPing() = {
+    if (healtCheckConfig.enabled) {
```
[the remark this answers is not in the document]
tysonnorris: I removed the double check

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
tysonnorris: I don't think having a separate service will be simpler, since a) it will need to track the internal state of ContainerProxy and b) that state is already tracked inside ContainerProxy and c) we would still create and destroy TCPPingClient and IO(Tcp) actors, which should also be tied to the ContainerProxy state (should be deleted when ContainerProxy is deleted etc)

style95: What I suggested is,

0) the service will be created on the bootstrap of invokers and the ActorRef will be sent to each (prewarm) container.
1) Prewarm containers will send a registration request to the service.
2) When it becomes other than the prewarm state, then it will send a deregistration request to the service.
3) The service will run forever as long as a corresponding invoker is running.
4) Once the service receives the registration request, it can store the endpoint in a map or list.
5) Every interval, it will send ping requests to all endpoints in the map.

So a) it doesn't need to track the internal state of ContainerProxy, b) only prewarm containers will register themselves on the creation and deregister when being warmed. so now the service is independent of the ContainerProxy state. c) it does not need to be removed as there will be always some prewarm containers to keepalive.

With this we can have some benefits:
1) No need to create/delete Pingclient based on the ContainerProxy lifecycle.
2) No need to have multiple actors(even if the overheads for multiple actors are small).

But what I suggested is a small enhancement point regarding performance and it does not have a big difference in terms of the functionality. So I am ok with the current version as well.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
tysonnorris: Current tests do not use TestFSMRef, rather they test the expected messages in and out of the actor. I moved that state to WarmedData and added tests - let me know if you have concerns.

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -346,7 +369,12 @@ class ContainerProxy(factory: (TransactionId,
         .pipeTo(self)
       goto(Running) using PreWarmedData(data.container, data.kind, data.memoryLimit, 1)
 
-    case Event(Remove, data: PreWarmedData) => destroyContainer(data.container)
+    case Event(Remove, data: PreWarmedData) => destroyContainer(data)
```
style95: It seems `Remove` object is only used by `ContainerPool` when removing warmed containers.
It's not related to this PR but is there any case that `ContainerPool` removes prewarm containers?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -429,32 +473,36 @@ class ContainerProxy(factory: (TransactionId,
     case Event(job: Run, data: WarmedData) =>
       implicit val transid = job.msg.transid
       activeCount += 1
-
-      initializeAndRun(data.container, job)
+      val newData = data.withResumeRun(job)
```
style95: Is this to always reschedule the job in case of ContainerHealthError no matter a container is resumed or not?

core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/ContainerProxy.scala
```diff
@@ -565,8 +628,8 @@ class ContainerProxy(factory: (TransactionId,
       .flatMap(_ => container.destroy()(TransactionId.invokerNanny))
       .map(_ => ContainerRemoved)
       .pipeTo(self)
-
-    goto(Removing)
+    println("removing")
```
style95: Should this be removed?

PR #4700, opened 2019-10-23
Make the formula to calculate the action time limit more configurable

steffenrost:
For reasons explained in detail in this *dev.openwhisk.apache.org* discussion thread ([Make the formula to calculate the action time limit more configurable](https://lists.apache.org/thread.html/339d39b40a3078d59573fdc45ba2af10040947dfd557966617aa6fb1@<dev.openwhisk.apache.org>)) it would be beneficial to to have 
the constant duration of 1 minute that is added to the calculated completion ack timeout duration configurable. Motivation is to have a better control over the applied completion ack timeout and by this to avoid the forced completion of activations in situations where it causes trouble.

## Description
This code change removes the constant duration of one minute used in the formula to calculate the completion ack timeout but adds a configurable parameter instead. This parameter defaults to the value of 1 minute, so the code that calculates the duration within a completion ack must be received for an activation behaves the same if the default is used.

## Related issue and scope

- [ ] API
- [x] Controller
- [x] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


ansible/group_vars/all
```diff
@@ -90,7 +90,7 @@ controller:
   managedFraction: "{{ controller_managed_fraction | default(1.0 - (controller_blackbox_fraction | default(__controller_blackbox_fraction))) }}"
   blackboxFraction: "{{ controller_blackbox_fraction | default(__controller_blackbox_fraction) }}"
   timeoutFactor: "{{ controller_timeout_factor | default(2) }}"
-  timeoutAddon: "{{ controller_timeout_addon | default(1) }}"
+  timeoutAddon: "{{ controller_timeout_addon | default('1 m') }}"
```
sven-lange-last: You also need to feed the new setting into the controller deployment like so:

https://github.com/apache/openwhisk/blob/78b2cbd7bd00fb6313ead1d71a316256def8b0c4/ansible/roles/controller/tasks/deploy.yml#L254-L255

@dgrove-oss what's the usual approach to add such configuration parameters to other deployment approaches of Openwhisk - like Kubernetes?

dgrove-oss: @sven-lange-last - Kubernetes uses the same mechanism of setting an environment variable like `CONFIG_whisk_loadbalancer_timeoutFactor` to override the default value that comes from loading `reference.conf`.  To make it easy to configure, you would add an entry into values.yaml (for example like this:https://github.com/apache/openwhisk-deploy-kube/blob/master/helm/openwhisk/values.yaml#L99) and use that entry in the `env` section of controller-pod.yaml (for example like this: https://github.com/apache/openwhisk-deploy-kube/blob/master/helm/openwhisk/templates/controller-pod.yaml#L132)

Issue #4637, opened 2019-09-20
inefficiencies and issues with the actions CRUD API


rabbah:
Actions are represented by the type `WhiskAction`, which includes a property of type `Exec`. The `Exec` instances may include the code inline. The REST API allows the controller to ignore the `Exec` value for the most part, but the current implementation has parallel types `WhiskActionMetadata` and `ExecMetadata` to explicitly drop the `Exec` values.

There is a lot of redundancy in the code - and the two types have drifted and changes are in one type and not the other. There is at least one bug that I'm aware of because of the drift.

Thanks to @chetanmeh's work, we do have uniform support for attaching "code" as separate entities (CouchDB attachments, S3, etc.). The implementation also allows for inlined code up to a certain limit. Because of these changes, we can drop the parallel types and improve the CRUD API's performance.

1. a GET of an action via the REST api today returns a butchered `exec` field with no code... is it OK to actually return the internal representation of the "code" attachment or do we continue to hide it?
1. is it ok to cache inlined code (since they are soft references and can be gced from the cache)?
1. all runtimes should be using attachments in the runtimes manifest, and so we should remove the property for "attachment" from the runtimes manifest
1. action operations continue to fetch the attachment and inline it (... only to discard it later for the "metadata" classes), this is unnecessary overhead

PR #4702, opened 2019-10-23
Fixed to check entityType when deserializing entity

AkimitsuTosha:
## Description
The action can be deleted with wsk trigger delete.

## Related issue and scope

- [ ] I opened an issue to propose and discuss this change ([#4637](https://github.com/apache/openwhisk/issues/4673))

- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskEntity.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskEntity.scala
@@ -117,16 +117,16 @@ object WhiskEntity {
 object WhiskDocumentReader extends DocumentReader {
   override def read[A](ma: Manifest[A], value: JsValue) = {
     val doc = ma.runtimeClass match {
-      case x if x == classOf[WhiskAction]         => WhiskAction.serdes.read(value)
+      case x if x == classOf[WhiskAction] => WhiskAction.serdes.read(value)
       case x if x == classOf[WhiskActionMetaData] => WhiskActionMetaData.serdes.read(value)
-      case x if x == classOf[WhiskPackage]        => WhiskPackage.serdes.read(value)
-      case x if x == classOf[WhiskActivation]     => WhiskActivation.serdes.read(value)
-      case x if x == classOf[WhiskTrigger]        => WhiskTrigger.serdes.read(value)
-      case x if x == classOf[WhiskRule]           => WhiskRule.serdes.read(value)
-      case _                                      => throw DocumentUnreadable(Messages.corruptedEntity)
+      case x if x == classOf[WhiskPackage] => WhiskPackage.serdes.read(value)
+      case x if x == classOf[WhiskActivation] => WhiskActivation.serdes.read(value)
+      case x if x == classOf[WhiskTrigger] => WhiskTrigger.serdes.read(value)
+      case x if x == classOf[WhiskRule] => WhiskRule.serdes.read(value)
+      case _ => throw DocumentUnreadable(Messages.corruptedEntity)
     }
     value.asJsObject.fields.get("entityType") match {
-      case Some( entityType ) => {
+      case Some(entityType) => {
         if (doc.entityType != entityType.convertTo[String]) {
           throw DocumentTypeMismatchException(
             s"document type ${doc.entityType} did not match expected type ${entityType.convertTo[String]}.")
--- a/common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskEntity.scala
+++ b/common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskEntity.scala
@@ -125,13 +125,9 @@ object WhiskDocumentReader extends DocumentReader {
       case x if x == classOf[WhiskRule]           => WhiskRule.serdes.read(value)
       case _                                      => throw DocumentUnreadable(Messages.corruptedEntity)
     }
-    value.asJsObject.fields.get("entityType") match {
-      case Some(JsString(entityType)) => {
-        if (doc.entityType != entityType) {
-          throw DocumentTypeMismatchException(
-            s"document type ${doc.entityType} did not match expected type ${entityType}.")
-        }
-      }
+    value.asJsObject.fields.get("entityType").foreach {
+      case JsString(entityType) if (doc.entityType != entityType) =>
+        throw DocumentTypeMismatchException(s"document type ${doc.entityType} did not match expected type $entityType.")
       case _ =>
     }
     doc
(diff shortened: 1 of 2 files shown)


common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskEntity.scala
```diff
@@ -115,7 +116,8 @@ object WhiskEntity {
 
 object WhiskDocumentReader extends DocumentReader {
   override def read[A](ma: Manifest[A], value: JsValue) = {
-    ma.runtimeClass match {
+    val entityType = value.asJsObject.fields("entityType").convertTo[String]
```
chetanmeh: `entityType` is technically an optional attribute as it was introduced later (#3164). So need to treat that as `Option`

rabbah: That’s a good point. If the property 
 doesn’t exist then fall back to the old code.

rabbah:
Thanks @AkimitsuTosha!

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskEntity.scala
AkimitsuTosha: I will consider this idea

AkimitsuTosha:
I understand. Add unit tests

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskEntity.scala
```diff
@@ -115,15 +116,25 @@ object WhiskEntity {
...
-      case x if x == classOf[WhiskRule]           => WhiskRule.serdes.read(value)
-      case _                                      => throw DocumentUnreadable(Messages.corruptedEntity)
+      case x if x == classOf[WhiskPackage] => WhiskPackage.serdes.read(value)
+      case x if x == classOf[WhiskActivation] => WhiskActivation.serdes.read(value)
+      case x if x == classOf[WhiskTrigger] => WhiskTrigger.serdes.read(value)
+      case x if x == classOf[WhiskRule] => WhiskRule.serdes.read(value)
+      case _ => throw DocumentUnreadable(Messages.corruptedEntity)
     }
+    value.asJsObject.fields.get("entityType") match {
+      case Some(entityType) => {
```
chetanmeh: This can be simplified a bit via destructuring the `JsString` itself and avoid explicit conversions later

```patch
- case Some(entityType) => {
+ case Some(JsString(entityType)) => {
```

tests/src/test/scala/org/apache/openwhisk/core/entity/test/WhiskEntityTests.scala
```diff
@@ -162,4 +147,78 @@ class WhiskEntityTests extends FlatSpec with ExecHelpers with Matchers {
     val trigger = WhiskTrigger(namespace, name)
     assertType(trigger, "trigger")
   }
+
+  behavior of "WhiskDocumentReader"
+
+  it should "check entityType when deserialize" in {
+    def assertType(d: WhiskEntity, entityType: String) = {
```
chetanmeh: Make it part of class such that it can be used by other test also and avoid duplication

tests/src/test/scala/org/apache/openwhisk/core/entity/test/WhiskEntityTests.scala
```diff
@@ -162,4 +147,78 @@ class WhiskEntityTests extends FlatSpec with ExecHelpers with Matchers {
...
+                 |	},
+                 |	"parameters": [],
+                 |	"limits": {
+                 |		"timeout": 60000,
+                 |		"memory": 256
+                 |	},
+                 |	"namespace": "namespace"
+                 |}""".stripMargin.parseJson
+
+    try {
```
chetanmeh: We can avoid try/catch as in case of exception test case itself would fail

tests/src/test/scala/org/apache/openwhisk/core/entity/test/WhiskEntityTests.scala
```diff
@@ -162,4 +147,78 @@ class WhiskEntityTests extends FlatSpec with ExecHelpers with Matchers {
...
+      val action = WhiskDocumentReader.read(manifest[WhiskAction], json)
+      assertType(action.asInstanceOf[WhiskEntity], "action")
+    } catch {
+      case _: DocumentTypeMismatchException =>
+        assert(false)
+    }
+
+    try {
```
chetanmeh: We can simplify this by using [scalatest assertThrows](http://www.scalatest.org/user_guide/using_assertions#interceptedExceptions)

tests/src/test/scala/org/apache/openwhisk/core/entity/test/WhiskEntityTests.scala
```diff
@@ -162,4 +147,78 @@ class WhiskEntityTests extends FlatSpec with ExecHelpers with Matchers {
...
+                 |  },
+                 |  "parameters": [],
+                 |  "limits": {
+                 |    "timeout": 60000,
+                 |    "memory": 256
+                 |  },
+                 |  "namespace": "namespace"
+                 |}""".stripMargin.parseJson
+
+    try {
```
chetanmeh: Again we can avoid try/catch here

common/scala/src/main/scala/org/apache/openwhisk/core/entity/WhiskEntity.scala
```diff
@@ -124,6 +125,16 @@ object WhiskDocumentReader extends DocumentReader {
...
+    value.asJsObject.fields.get("entityType") match {
+      case Some(JsString(entityType)) => {
+        if (doc.entityType != entityType) {
+          throw DocumentTypeMismatchException(
+            s"document type ${doc.entityType} did not match expected type ${entityType}.")
+        }
+      }
+      case _ =>
+    }
+    doc
```
rabbah: Warning Scala Nit ahead. You can simplify this further:

```scala

    value.asJsObject.fields.get("entityType").foreach {
      case JsString(entityType) if (doc.entityType != entityType) =>
        throw DocumentTypeMismatchException(s"document type ${doc.entityType} did not match expected type $entityType.")

      case _ =>
    }

```

The first case match will only apply if the value of `entityType` is a string and it matches the value in the `doc` type. Otherwise, `_` matches all others.

style95:
@AkimitsuTosha 
Thank you for the first contribution 🎉.

PR #4713, opened 2019-11-05
Document optional .wskprops related to provider-specific API Gateway …

mrutkows:
Missing documentation on some optional keys that are needed when using API gateway services specific to some providers.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -64,8 +64,6 @@ To verify your CLI setup, try [creating and running an action](./samples.md).
 
 Some OpenWhisk providers make use of optional properties that can be added to the `.wskprops` file.  The following keys are optional:
 
-- **NAMESPACE** - Optional, provider-relative identifier for a namespace to be used when managing OpenWhisk entities (e.g. Actions, Triggers, Rules, etc.) with the CLI.
-
 - **APIGW_ACCESS_TOKEN** - Optional, provider-specific authorization token for an independently hosted API Gateway service used for managing OpenWhisk API endpoints.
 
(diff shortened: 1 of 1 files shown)


docs/cli.md
```diff
@@ -51,10 +51,25 @@ you can run the following command from your `openwhisk` directory:
...
+- **APIHOST** - Required key for the API host value.
+- **AUTH** - Required key for the Authorization key.
 
 To verify your CLI setup, try [creating and running an action](./samples.md).
 
+### Optional Whisk Properties
+
+Some OpenWhisk providers make use of optional properties that can be added to the `.wskprops` file.  The following keys are optional:
+
+- **NAMESPACE** - Optional, provider-relative identifier for a namespace to be used when managing OpenWhisk entities (e.g. Actions, Triggers, Rules, etc.) with the CLI.
```
rabbah: i dont think we need this at all - in fact i thought we removed recognition of this property from the cli.

mrutkows: I will check... but i had a .wskprops created in July timeframe that added this key (although via the IBM plugin)

mrutkows: NAMESPACE is still in the go-client... 
https://github.com/apache/openwhisk-client-go/blob/master/whisk/wskprops.go

rabbah: but gone/ignored by the cli itself https://github.com/apache/openwhisk-cli/pull/434/files

mrutkows: it appears that the IBM plugin requires it and adds it back in if you delete it...  I will remove from the open docs.

PR #4715, opened 2019-11-07
Extension points for kamon metric key generator and reporter

ruediger-maass:
**This PR is closed since it is not required to introduce new variables to achieve reconfiguration of `application.conf` for `common`. Thank you @chetanmeh for your help!**

--------------------------------------------------------------------
Original request:

# Overview

Extension points for kamon metric key generator and reporter

## Description
We need a different metric key generator and a different reporter in our product. Therefore we request two simple extension points in the `application.conf` for `common` that can be easily configured with two environment variables `KAMON_STATSD_METRIC_KEY_GENERATOR`and `METRICS_KAMON_REPORTER`. If the variables are not defined then Openwhisk will work as before and not change the current behavior. 

So the main changes of this PR relate to `application.conf`for `common` and to the test. This PR also includes changes for exposing the two new environment variables to the deployment. Since this is a special use case and you would normally have to add code for a different generator and reporter it is the question if this PR should omit the changes for the deployment.

Current behavior of OpenWhisk only changes if the aforementioned variables are actively used.
- [ ] API
- [x] Controller
- [ ] Loadbalancer
- [x] Invoker
- [ ] Tests
- [x] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


common/scala/src/main/resources/application.conf
```diff
@@ -61,7 +61,9 @@ kamon {
             dispatcher =  [ "*" ]
         }
 
+        # Extension point for metrics key generator
         metric-key-generator = org.apache.openwhisk.common.WhiskStatsDMetricKeyGenerator
+        metric-key-generator = ${?KAMON_STATSD_METRIC_KEY_GENERATOR}
```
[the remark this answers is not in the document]
ruediger-maass: Thanks, Chetan! Yes, the reporter we need is an extension of the Kamon statsd reporter. If the CONFIG_kamon_... mechanism works for both settings we need in `application.conf` then of course I would not add additional variables - I thought I have to do this when I looked at METRICS_KAMON_TAGS and other variables in `application.conf`. I can check your proposal next Tuesday and come back. Maybe we do not need any code changes then and can withdraw this PR. Thank you for helpful hint!

tests/src/test/scala/org/apache/openwhisk/common/AppConfigForCommonTests.scala
```diff
@@ -0,0 +1,83 @@
...
+    // read application.conf for common
+    val buffer = Source
+      .fromFile(WhiskProperties.getFileRelativeToWhiskHome("common/scala/src/main/resources/application.conf"))
+    var content = buffer.getLines.mkString("\n")
+    // simulate substitution of environment variables in application.conf
+    for (subst <- Seq(("KAMON_STATSD_METRIC_KEY_GENERATOR", generator), ("METRICS_KAMON_REPORTER", reporter))) {
+      content = subst match {
+        case (name, Some(value)) =>
+          // substitute ${name} and ${?name} occurrences with the value
+          val pattern = ("\\$\\{\\??" + name + "\\}").r
```
markusthoemmes: Why even pattern match? These should be exact matches so a simpler `replace("${?" + name + "}", value` should work here? Note that it's still a global replace.

tests/src/test/scala/org/apache/openwhisk/common/AppConfigForCommonTests.scala
```diff
@@ -0,0 +1,83 @@
...
+  def getReporters = {
+    val kamonConfig = Kamon.config.getConfig("kamon")
+    kamonConfig.getStringList("reporters").toArray
+  }
+
+  def configKamon(generator: Option[String], reporter: Option[String]) = {
+    // read application.conf for common
+    val buffer = Source
+      .fromFile(WhiskProperties.getFileRelativeToWhiskHome("common/scala/src/main/resources/application.conf"))
+    var content = buffer.getLines.mkString("\n")
```
markusthoemmes: You need a `var` here because you're mutating the string below, which works. You don't need the mutability here at all though.

Instead, you can run your substitutions through a `foldLeft` based "loop" that applies all the replacements to the base string and passes the replaced thing along for the next iteration, eventually returning it. I'll provide a full example in the overarching comment.

tests/src/test/scala/org/apache/openwhisk/common/AppConfigForCommonTests.scala
```diff
@@ -0,0 +1,83 @@
...
+    // simulate substitution of environment variables in application.conf
+    for (subst <- Seq(("KAMON_STATSD_METRIC_KEY_GENERATOR", generator), ("METRICS_KAMON_REPORTER", reporter))) {
```
markusthoemmes: This looks strongly like key value pairs, so it should probably use a `Map[String, String]`. The neat bit with using a Map here in conjunction with the incoming Options is that you can make Scala not add the values to the `Map` in the first place, if they are none. In this example this would be:

```scala
val replacements = Map.empty[String, String] ++
  generator.map("KAMON_STATSD_METRIC_KEY_GENERATOR" -> _) ++
  reporter.map("METRICS_KAMON_REPORTER" -> _)
```

Which returns a `Map` that only contains any values if the respective Option was a Some.

tests/src/test/scala/org/apache/openwhisk/common/AppConfigForCommonTests.scala
```diff
@@ -0,0 +1,83 @@
...
+      content = subst match {
+        case (name, Some(value)) =>
+          // substitute ${name} and ${?name} occurrences with the value
+          val pattern = ("\\$\\{\\??" + name + "\\}").r
+          pattern.replaceAllIn(content, value)
+        case _ => content // no changes
+      }
+    }
```
markusthoemmes: All three suggestions above together would form a piece of code that'd look like this:

```scala
val replacements = Map.empty[String, String] ++
  generator.map("KAMON_STATSD_METRIC_KEY_GENERATOR" -> _) ++
  reporter.map("METRICS_KAMON_REPORTER" -> _)

val replacedContent = replacements.foldLeft(content) {
  case (s, (k, v)) => s.replace("${?"+ k + "}", v)
}
```

They should be semantically equivalent, but it's a bit more expressive and most importantly keeps the `String` immutable.

tests/src/test/scala/org/apache/openwhisk/common/AppConfigForCommonTests.scala
```diff
@@ -0,0 +1,83 @@
...
+trait Helpers {
+
+  def getKeyGenerator = {
+    val statsDConfig = Kamon.config.getConfig("kamon.statsd")
+    statsDConfig.getString("metric-key-generator")
+  }
+
+  def getReporters = {
+    val kamonConfig = Kamon.config.getConfig("kamon")
+    kamonConfig.getStringList("reporters").toArray
```
markusthoemmes: Arrays are very seldomly used in Scala and only relevant for vastly performance optimized code. I'd use a `toList` here to stay in the "usual" realms.

tests/src/test/scala/org/apache/openwhisk/common/AppConfigForCommonTests.scala
```diff
@@ -0,0 +1,83 @@
...
+      content = subst match {
+        case (name, Some(value)) =>
+          // substitute ${name} and ${?name} occurrences with the value
+          val pattern = ("\\$\\{\\??" + name + "\\}").r
+          pattern.replaceAllIn(content, value)
+        case _ => content // no changes
+      }
+    }
+    // reconfigure kamon with the modified application.conf
+    val newConfig = ConfigFactory.parseString(content).withFallback(ConfigFactory.load()).resolve()
```
markusthoemmes: Is the fallback even wanted at this point?

ruediger-maass:
This PR is closed since it is not required to introduce new variables to achieve reconfiguration of `application.conf` for `common`. Thank you @chetanmeh for your help!

Issue #4718, opened 2019-11-08
Recording metrics for webaction response statusCode in user-events service. 


neerajmangal:
A developer can specify the status code using `statusCode` in response, in case of web actions. 
The top-level activation response captures the activation `statusCode` as 0,1,2, and 3 based on `success`, `application error`, `action developer error` and `whisk internal error`. 

If action developer sending a `statusCode` explicitly in response, it will be stored in activation's response result as `statusCode` (like below as HTTP status code `401` ) 

```
"response": {
        "status": "application error",
        "statusCode": 0,
        "success": false,
        "result": {
            "error": {
                "body": "Authentication credentials are invalid.",
                "statusCode": 401
            }
        }
    }
```

Capturing metrics for the statusCode returned by the action developer in user-events can give insights like
"How many requests failed with which error code".  

Conditions - 

As returning a proper status code, is the responsibility of the developer, if a developer does not specify a return `statusCode` in response the metric will not be present/recoded. 

Please let me know your thoughts. 

PS: - While doing some tests, I see top-level `statusCode` always as 0 in case of all error scenarios.

rabbah:
The status code in the response is orthogonal to the webaction http status code. It is an internal code only. Changing this is an api change as the value 0-3 are part of the activation record schema.

neerajmangal:
> The status code in the response is orthogonal to the webaction http status code. It is an internal code only. Changing this is an api change as the value 0-3 are part of the activation record schema.

Yup, you are right, I confused myself with activation response vs end-user response. 
I see the status code in web actions HTTP status code which makes sense.

```
< HTTP/1.1 401 Unauthorized
< X-Request-ID: ZPyOucUbIuUz2eqskYfGTqJrSROxBOHP
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Methods: OPTIONS, GET, DELETE, POST, PUT, HEAD, PATCH
< Access-Control-Allow-Headers: Authorization, Origin, X-Requested-With, Content-Type, Accept, User-Agent
< x-openwhisk-activation-id: 85daf1091d3940e49af1091d3950e47d
< Server: akka-http/10.1.8
< Date: Fri, 08 Nov 2019 04:21:37 GMT
< Content-Type: text/html; charset=UTF-8
< Content-Length: 39
```

neerajmangal:
Closed by #4726

PR #4726, opened 2019-11-15
Recording metrics for action result response statusCode in user-events service. 

neerajmangal:
This PR is the implementation of #4718, recording action's `statusCode` in Activation `event`, if it is set by the developer in action response. 

## Description
Currently, we can capture the `error` based on `Application`, `Developer` and `System`, with this change, the developer/operator can get visibility of actual status code sent in action response, i.e. 4XX, 5XX, etc by action. 
 
Action `statusCode` will be added in "Activation" `event` as `actionStatusCode` and consumed by user-event service. 

Note - For now `statusCode` present only if `statusCode` field is set by action developer.    

```json
"result": {
            "error": {
                "body": "Authentication credentials are invalid.",
                "statusCode": 401
            }
        }
```

With this change `Activation event`will look like this - 

```json
{
   "body": {
      "actionStatusCode": 404,
      "conductor": false,
      "duration": 46,
      "initTime": 34,
      "kind": "nodejs:10",
      "memory": 256,
      "name": "guest/helloWorld3",
      "size": 53,
      "statusCode": 1,
      "waitTime": 9
   },
   "eventType": "Activation",
   "namespace": "guest",
   "source": "invoker0",
   "subject": "guest",
   "timestamp": 1573811872958,
   "userId": "23bc46b1-71f6-4ed5-8c54-816aa4f8c502"
}
```

**Note** - `actionStatusCode` only present when - 

- Activation response has `error` field with `statusCode` set
- or Activation has `statusCode` set by developer.  
 
Sample Dashboard (added "Action Error Status Code" Panel) - 

Screenshot 2019-11-15 at 3 34 39 PM

TODO
- [ ] Add panel to default dashboards. 

## Related issue and scope

- [x]  I opened an issue to propose and discuss this change (#4718 )

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x]  User events
- [ ] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:


common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
```diff
@@ -293,7 +295,9 @@ case class Activation(name: String,
                       conductor: Boolean,
                       memory: Int,
                       causedBy: Option[String],
-                      size: Option[Int] = None)
+                      size: Option[Int] = None,
+                      actionStatusCode: Option[Int] = None
```
rabbah: The inspection of the result of an action (instead of treating it opaquely) is a down side to this approach where otherwise the result is completely opaque. Naturally it would only work for actions that actually return an http response with a status code, which I don't think you can distinguish at the point of constructing a `Message`. Did you consider capturing the HTTP status code from the edge router instead for a /api/v1/web actions?

tests/src/test/scala/org/apache/openwhisk/core/connector/test/EventMessageTests.scala
```diff
@@ -90,5 +90,31 @@ class EventMessageTests extends FlatSpec with Matchers {
       Activation("ns2/a", 0, toDuration(0), toDuration(0), toDuration(0), "testkind", false, 0, None, Some(42)))
   }
 
+  it should "Transform a activation with error status code" in {
+    val resultWithError =
```
chetanmeh: Also add one more variant with `statusCode` with `error`

tests/src/test/scala/org/apache/openwhisk/core/connector/test/EventMessageTests.scala
```diff
@@ -90,5 +90,31 @@ class EventMessageTests extends FlatSpec with Matchers {
...
+      """
+        |{
+        | "statusCode" : 404,
+        | "body": "Requested resource not found"
+        |}
+        |""".stripMargin
+    val a =
+      fullActivation
+        .copy(response = ActivationResponse.applicationError(resultWithError.parseJson, Some(42)))
+    Activation.from(a) shouldBe Success(
```
chetanmeh: Better to restrict the assertion to `statusCode` value itself instead of checking on whole assertion. Would reduce the noise

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
```diff
@@ -340,7 +345,22 @@ object Activation extends DefaultJsonProtocol {
...
-      "size")
+      "size",
+      "actionStatusCode")
+
+  /** Get "StatusCode" from result response **/
+  def getActivationStatusCode(result: Option[JsValue]): Option[Int] = {
+    val statusCode = JsHelpers.getFieldPath(result.get.asJsObject, ERROR_FIELD, "statusCode")
+
+    statusCode match {
+      case Some(value) => Some(value.convertTo[Int])
```
chetanmeh: Can be simplified  a bit here and also below
```scala
Some(JsNumber(code)) => Some(code.intValue())
```

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
chetanmeh: > The inspection of the result of an action (instead of treating it opaquely) is a down side to this approach where otherwise the result is completely opaque

The `statusCode` field is a documented structure for [web actions][1]. `statusCode` handling when under `error` field though is not documented properly. So current logic only looks for these documented structure to extract the info

> Did you consider capturing the HTTP status code from the edge router instead for a /api/v1/web actions

@rabbah If we capture the metrics at edge router then its tricky to distinguish a 5xx status from system error or web action actually responding with 5xx. Hence need to capture it at activation response level.

[1]: https://github.com/apache/openwhisk/blob/master/docs/webactions.md#handling-http-requests-with-actions

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
```diff
@@ -340,7 +345,22 @@ object Activation extends DefaultJsonProtocol {
...
+      case None =>
+        JsHelpers.getFieldPath(result.get.asJsObject, "statusCode") match {
+          case Some(value) => Some(value.convertTo[Int])
+          case None        => None
+        }
```
rabbah:
```
statusCode
.map(_.intValue)
.orElse(_ => JsHelpers.getFieldPath(result.get.asJsObject, "statusCode").map(_.intValue))
```

i think - warning i didn't vet this in an ide/compiler but see my comment above, i don't think either of these are sufficient.

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
rabbah: the `statusCode` here is a use defined value - it could be of any type, including integer, string or other legal JSON values. Since this function is trying to interpret a use response it needs to be more defensive and also handle  `case _` when there are no matches.

also note that the controller handles string and int status codes.

docs/metrics.md
```diff
@@ -315,6 +315,7 @@ conductor - true for conductor backed actions
 memory - maximum memory allowed for action container
 causedBy - contains the "causedBy" annotation (can be "sequence" or nothing at the moment)
 size - size (in bytes) of the invocation response
+actionStatusCode - status code represents `statusCode` set in result response. (if not set, this field will not be present)
```
selfxp: Should we rename the field into `httpCode` or `httpStatusCode` and point the [statusCode](https://github.com/apache/openwhisk/blob/master/docs/webactions.md#handling-http-requests-with-actions) to the documentation on how the user can set this up?

common/scala/src/main/scala/org/apache/openwhisk/core/connector/Message.scala
neerajmangal: Sure, will try your suggestion.

core/monitoring/user-events/src/main/scala/org/apache/openwhisk/core/monitoring/metrics/KamonRecorder.scala
```diff
@@ -34,6 +34,7 @@ trait KamonMetricNames extends MetricNames {
   val durationMetric = "openwhisk.action.duration"
   val responseSizeMetric = "openwhisk.action.responseSize"
   val statusMetric = "openwhisk.action.status"
+  val actionStatusCodeMetric = "openwhisk.action.errorCode"
```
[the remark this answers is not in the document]
neerajmangal: Thanks for pointing this out @selfxp, I will change it accordingly. 
Initially, we thought to implement it only for web actions error code, but this is now applicable to all type actions if they have status code set in the response.

core/monitoring/user-events/src/main/scala/org/apache/openwhisk/core/monitoring/metrics/PrometheusRecorder.scala
```diff
@@ -46,6 +46,7 @@ trait PrometheusMetricNames extends MetricNames {
   val responseSizeMetric = "openwhisk_action_response_size_bytes"
   val statusMetric = "openwhisk_action_status"
   val memoryMetric = "openwhisk_action_memory"
+  val actionStatusCodeMetric = "openwhisk_action_errorcode"
```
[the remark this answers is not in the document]
neerajmangal: Thanks! I missed it.

neerajmangal:
Thanks @rabbah, @chetanmeh and @selfxp for the feedback. Based on your feedback, I modified below - 

- More defensive checks when parsing the result response. 
  - if the result response contains an `error` field, check if it has `statusCode` defined by the user, as other fields are discarded if the `error` field is present in result response. 
     ```json
     "result": {
            "error": {
                "body": "Authentication credentials are invalid.",
                "statusCode": 401
            }
        }
     ````
  - else check if `statusCode` is set by the user without the `error` field in the result response.
     ```json
     "result": {
                "body": "Authentication credentials are invalid.",
                "statusCode": 401
        }
     ````
  - `statusCode` value is being checked for valid integer (401) and string value ("401").
  -  If there is any issue in getting the `intValue` (i.e. `statusCode` like "i401" or any string "xyz") return set `userDefinedStatusCode as `400` bad request. 
  - Using `value.convertTo[BigInt].intValue()` takes care of `JSNumber` and `JSString` type. I tried alternate approach like handling the status code for webaction https://github.com/apache/openwhisk/blob/master/core/controller/src/main/scala/org/apache/openwhisk/core/controller/WebActions.scala#L234, but for normal action use case it will miss a case where return `statusCode` is a valid integer (like "1") but not a http statusCode.
-  Changed `actionStatusCode` to `userDefinedStatusCode` as `statusCode` is not limited to http status codes. Please let me know if you have any better suggestions. 
-  Other tests and doc changes per feedback. 

Please let me know your feedback.

rabbah:
Changes lgtm. Sorry @neerajmangal for my delayed response. Can you resolve the conflict. Is there anything remaining on this PR?

PR #4728, opened 2019-11-18
Increase toStrictTimeout value to omit intermittent timeout exceptions

steffenrost:
In our tests we encounter every now and then failing tests that fail because the HttpEntity.toStrict timed out after 5 seconds. In order to avoid these timeouts I propose to increase the defined timeout from 5 to 30 seconds.

## Description
This code change increase the timeout value `toStrictTimeout` from 5 to 30 seconds. Intention is to avoid the following error: `org.scalatest.exceptions.TestFailedException: The future returned an exception of type: java.util.concurrent.TimeoutException, with message: HttpEntity.toStrict timed out after 5 seconds while still waiting for outstanding data.`

## Related issue and scope

- [ ] API
- [ ] Controller
- [ ] Loadbalancer
- [ ] Invoker
- [x] Tests
- [ ] Deployment
- [ ] CLI
- [ ] General tooling
- [ ] Documentation

## Types of changes

## Checklist:

### DIFF
--- a/tests/src/test/scala/common/rest/WskRestOperations.scala
+++ b/tests/src/test/scala/common/rest/WskRestOperations.scala
@@ -1131,7 +1131,7 @@ trait RunRestCmd extends Matchers with ScalaFutures with SwaggerValidator {
 
   val protocol: String = loadConfigOrThrow[String]("whisk.controller.protocol")
   val idleTimeout: FiniteDuration = 90 seconds
-  val toStrictTimeout: FiniteDuration = 5 seconds
+  val toStrictTimeout: FiniteDuration = 30 seconds
   val queueSize = 10
   val maxOpenRequest = 1024
   val basePath = Path("/api/v1")


style95:
I feel generally 5s would be enough to finish a request.
Have you checked the root cause for the timeout?
Is this because of the low performance of testing machine?

steffenrost:
> I feel generally 5s would be enough to finish a request.
> Have you checked the root cause for the timeout?
> Can this be because of the low performance of testing machine?

As said this happens sometimes but not always, so it's not a general performance problem. But it is annoying to waste a two hours build because a test is failing because of an improper timeout, at least in this situation. Also `WskRestOperations.scala` is part of the test code, so I don't see why we shouldn't allow a bit more time here. However we can discuss to increase the timeout to only 15 seconds if you think this is more appropriate. @style95 was do think?

style95:
@steffenrost Actually I am ok with this.
I am also occasionally facing an intermittent test failure.
(Even if it is not because of timeout exception.)
Since this can be because of the contention among tests I just thought it would be better to resolve the main cause.

If it really solves your problem, I am ok with this.

